PR #1041, opened 2025-08-24
feat: filter tx by amount

Azorlogh:
Adds filtering of transactions by amount, e.g.
```
{
  "$gt": { "amount[EUR]": 100 }
}
```
will only show transactions where the sum of amounts of all EUR postings is greater than 100


internal/api/v2/controllers_transactions_list_test.go
```diff
@@ -146,6 +146,20 @@ func TestTransactionsList(t *testing.T) {
...
+		{
+			name: "using amount",
+			body: `{"$lt": {"amount[EUR]": 20}}`,
+			expectQuery: storagecommon.InitialPaginatedQuery[any]{
+				PageSize: bunpaginate.QueryDefaultPageSize,
+				Column:   "id",
+				Order:    pointer.For(bunpaginate.Order(bunpaginate.OrderDesc)),
+				Options: storagecommon.ResourceQuery[any]{
+					PIT:     &now,
+					Builder: query.Lt("amount[EUR]", 20.0),
```
Azorlogh: It looks like the query numbers currently get parsed to floats only? Wouldn't I want something like a big int here?

gfyrag: You can

test/e2e/api_transactions_list_test.go
```diff
@@ -444,6 +444,40 @@ var _ = Context("Ledger transactions list API tests", func() {
...
+							},
+						},
+						Ledger:   "default",
+						PageSize: pointer.For(pageSize),
+						Pit:      &now,
+					},
+				)
+				Expect(err).To(BeNil())
+			})
+			_ = response
```
gfyrag: remove please

gfyrag (dismissed):
Technically valid.
Not agreed to add this on the ledger (performances).

flemzord (changes requested):
Performance issue

PR #1046, opened 2025-08-28
Hotfix/fix release v2.3 cli bug

ascandone:


fguery:
You need to base your banch on release/2.3 before cherrypicking ;)

PR #1071, opened 2025-10-07
feat(api/v2): Implement /transactions/sum endpoint with account-based transaction aggregation

Abdel-Monaam-Aouini:
**Description**:  
This PR introduces the `/transactions/sum` endpoint to the v2 API, which calculates and returns the sum of transactions for a specified account, grouped by asset. The implementation includes:

- Pagination support with configurable page size
- Filtering by account and optional asset
- Proper error handling for missing or invalid parameters
- Comprehensive test coverage for success and error cases

The endpoint follows the API's standard response format with data wrapping and adheres to the project's coding standards.

### DIFF
--- a/internal/api/v2/controllers_transactions_sum.go
+++ b/internal/api/v2/controllers_transactions_sum.go
@@ -1,0 +1,110 @@
+package v2
+
+import (
+	"errors"
+	"math/big"
+	"net/http"
+
+	"github.com/formancehq/go-libs/v3/api"
+	"github.com/formancehq/go-libs/v3/bun/bunpaginate"
+	ledger "github.com/formancehq/ledger/internal"
+	"github.com/formancehq/ledger/internal/api/common"
+	storagecommon "github.com/formancehq/ledger/internal/storage/common"
+)
+
+type sumResponse struct {
+	Account string   `json:"account"`
+	Asset   string   `json:"asset"`
+	Sum     *big.Int `json:"sum"`
+}
+
+func processPostings(account string, txs *bunpaginate.Cursor[ledger.Transaction], assetFilter string) []sumResponse {
+	// Calculate sums per asset
+	assetSums := make(map[string]*big.Int)
+
+	for _, tx := range txs.Data {
+		for _, posting := range tx.Postings {
+			if posting.Source == account {
+				// Debit from the account (negative amount)
+				if _, ok := assetSums[posting.Asset]; !ok {
+					assetSums[posting.Asset] = big.NewInt(0)
+				}
+				assetSums[posting.Asset] = new(big.Int).Sub(assetSums[posting.Asset], posting.Amount)
+			} else if posting.Destination == account {
+				// Credit to the account (positive amount)
+				if _, ok := assetSums[posting.Asset]; !ok {
+					assetSums[posting.Asset] = big.NewInt(0)
+				}
+				assetSums[posting.Asset] = new(big.Int).Add(assetSums[posting.Asset], posting.Amount)
+			}
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


flemzord:
Hey 
  Thanks for the PR. Unfortunately, there are critical issues that require a refactor.

  🔴 Critical Issues

  1. Pagination bug (lines 164-184)
  The cursor is never applied in the loop → first page fetched repeatedly → incorrect results.

  2. Wrong pattern
  Handler should return http.HandlerFunc and accept paginationConfig (see listTransactions, readVolumes).

  3. Code duplication
  getPaginatedTransactions reimplements existing getPaginatedQuery from common.go → remove it entirely.

  4. Performance
  Loading all transactions in memory is inefficient. For 1M+ transactions this will be very slow.

  💡 Solution: SQL Implementation

  Use SQL aggregation instead (like /aggregate/balances).

  Reference: internal/storage/ledger/resource_aggregated_balances.go

  SQL query (using moves table):
  SELECT asset,
         SUM(CASE WHEN is_source THEN -amount::numeric ELSE amount::numeric END) as sum
  FROM moves
  WHERE ledger = ? AND accounts_address = ?
  GROUP BY asset

  Files to create/modify:
  1. New: resource_transactions_sum.go (resource handler)
  2. store.go: Add TransactionsSum() method
  3. controller.go: Add GetTransactionsSum()
  4. controllers_transactions_sum.go: Use getResourceQuery + call controller
  5. views.go: Add renderTransactionSum()

  🤔 Architectural Question

  Do we really need a dedicated endpoint for this?

  I'm not entirely convinced we need /transactions/sum. Maybe we should consider evolving existing filters/endpoints instead (e.g., add aggregation capabilities to existing
  endpoints)?

  This needs more thought and discussion before we commit to this API design.

Abdel-Monaam-Aouini:
> Hey Thanks for the PR. Unfortunately, there are critical issues that require a refactor.
> 
> 🔴 Critical Issues
> 
> 1. Pagination bug (lines 164-184)
>    The cursor is never applied in the loop → first page fetched repeatedly → incorrect results.
> 2. Wrong pattern
>    Handler should return http.HandlerFunc and accept paginationConfig (see listTransactions, readVolumes).
> 3. Code duplication
>    getPaginatedTransactions reimplements existing getPaginatedQuery from common.go → remove it entirely.
> 4. Performance
>    Loading all transactions in memory is inefficient. For 1M+ transactions this will be very slow.
> 
> 💡 Solution: SQL Implementation
> 
> Use SQL aggregation instead (like /aggregate/balances).
> 
> Reference: internal/storage/ledger/resource_aggregated_balances.go
> 
> SQL query (using moves table): SELECT asset, SUM(CASE WHEN is_source THEN -amount::numeric ELSE amount::numeric END) as sum FROM moves WHERE ledger = ? AND accounts_address = ? GROUP BY asset
> 
> Files to create/modify:
> 
> 1. New: resource_transactions_sum.go (resource handler)
> 2. store.go: Add TransactionsSum() method
> 3. controller.go: Add GetTransactionsSum()
> 4. controllers_transactions_sum.go: Use getResourceQuery + call controller
> 5. views.go: Add renderTransactionSum()
> 
> 🤔 Architectural Question
> 
> Do we really need a dedicated endpoint for this?
> 
> I'm not entirely convinced we need /transactions/sum. Maybe we should consider evolving existing filters/endpoints instead (e.g., add aggregation capabilities to existing endpoints)?
> 
> This needs more thought and discussion before we commit to this API design.

Thanks for the review! 
I’ve fixed all the points you mentioned.
We actually need this feature in our product, which is why I took the initiative to enhance it and contribute back.

PR #1083, opened 2025-10-16
fix: build invalid result

gfyrag:


test/e2e/api_transactions_revert_test.go
```diff
@@ -126,6 +127,7 @@ var _ = Context("Ledger revert transactions API tests", func() {
 						force = true
 					})
 					It("Should be ok", func() {
+						Expect(newTransaction.V2CreateTransactionResponse.Data.ID).To(Equal(big.NewInt(3)))
```
paul-nicolas: same here, we should put the error check before the newTransaction check

paul-nicolas (dismissed):

PR #1108, opened 2025-11-07
feat: implement CE/EE architecture

flemzord:


ee/audit.go
```diff
@@ -0,0 +1,69 @@
...
+	"context"
+
+	"github.com/ThreeDotsLabs/watermill/message"
+	"github.com/formancehq/go-libs/v3/audit"
+	"github.com/formancehq/go-libs/v3/logging"
+	"github.com/spf13/cobra"
+	"go.uber.org/fx"
+)
+
+const serviceName = "ledger"
```
gfyrag: constant already exists

cmd/serve.go
```diff
@@ -186,6 +188,9 @@ func NewServeCommand() *cobra.Command {
 	cmd.Flags().String(WorkerGRPCAddressFlag, "localhost:8081", "GRPC address")
 	cmd.Flags().Bool(SemconvMetricsNames, false, "Use semconv metrics names (recommended)")
 
+	// Add EE flags (if EE build)
+	ee.AddFlags(cmd)
```
gfyrag, line 195: Add flags only in ee builds

.goreleaser.yml
```diff
@@ -7,8 +7,9 @@ includes:
       url: https://raw.githubusercontent.com/formancehq/earthly/refs/heads/main/.goreleaser.default.yaml
 
 builds:
+  # Community Edition build
```
gfyrag, line 10: I guess changes in this file can break a lot of user automation.

go.mod
```diff
@@ -51,7 +51,7 @@ require (
 
 require (
 	github.com/ClickHouse/clickhouse-go/v2 v2.37.2
-	github.com/formancehq/go-libs/v3 v3.4.0
+	github.com/formancehq/go-libs/v3 v3.5.1-0.20251107100055-6f51cf04fd2e
```
gfyrag, line 54: Use a tag version before merge.

cmd/serve.go
flemzord: CleanShot 2025-11-07 at 13 47 56
Yes, that's exactly the case

.goreleaser.yml
flemzord: Hmm, I don't think so, we only use them to build binaries and Docker images

PR #1109, opened 2025-11-07
feat: Chart of accounts & tx validation

Azorlogh:


internal/controller/ledger/controller_default.go
```diff
@@ -431,6 +431,29 @@ func (ctrl *DefaultController) createTransaction(ctx context.Context, store Stor
...
+	if parameters.SchemaVersion == "" {
+		// Only allow transactions without schema if the ledger has no schema
+		schemas, err := store.FindSchemas(ctx, storagecommon.InitialPaginatedQuery[any]{PageSize: 1})
+		if err != nil {
+			return nil, err
+		}
+		if len(schemas.Data) > 0 {
+			return nil, ErrSchemaRequired{}
+		}
+	} else if parameters.SchemaVersion != "" {
```
gfyrag: useless if

internal/chart.go
```diff
@@ -0,0 +1,257 @@
...
+type ChartVariableSegment struct {
+	ChartSegment
+
+	Pattern string
+	Label   string
+}
+
+type ChartOfAccounts map[string]ChartSegment
+
+const SegmentRegex = "^\\$?[a-zA-Z0-9_-]+$"
```
gfyrag: https://github.com/formancehq/ledger/blob/7b27e7fb426540862886a173e0662ba9484af0d1/pkg/accounts/accounts.go#L5

internal/chart.go
```diff
@@ -0,0 +1,257 @@
...
+	var account ChartAccount
+	var fixedSegments map[string]ChartSegment
+	var variableSegment *ChartVariableSegment
+	keys := []string{}
+	for key := range segment {
+		keys = append(keys, key)
+	}
+	for _, key := range keys {
+		value := segment[key]
+		isSubsegment := key[0] != '_'
```
gfyrag: So, we are denying any segment starting with '_' ?
What about existing users which would have already use that?

internal/chart.go
Azorlogh: It also allows `$test` and forbids `foo:bar`
I can rename it `ChartSegmentRegex` to be less confusing

internal/chart.go
```diff
@@ -0,0 +1,261 @@
...
+	var account ChartAccount
+	var fixedSegments map[string]ChartSegment
+	var variableSegment *ChartVariableSegment
+	keys := []string{}
+	for key := range segment {
+		keys = append(keys, key)
+	}
+	for _, key := range keys {
+		value := segment[key]
+		isSubsegment := key[0] != PROPERTY_PREFIX
```
gfyrag: I guess we should filter empty segment key.
Some users could send : 
```
{ "": { ...} }
```

gfyrag: Ok, wrong.

internal/chart.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func (s *ChartOfAccounts) UnmarshalJSON(data []byte) error {
+	var segment map[string]json.RawMessage
+	if err := json.Unmarshal(data, &segment); err != nil {
+		return err
+	}
+	out := make(map[string]ChartSegment)
+	for key, value := range segment {
+		if !ValidateSegment(key) || key[0] == '$' || key[0] == PROPERTY_PREFIX {
```
gfyrag: panic if key == "" I guess

internal/chart.go
```diff
@@ -0,0 +1,261 @@
...
+	*s = out
+	return nil
+}
+func (s *ChartSegment) UnmarshalJSON(data []byte) error {
+	var segment map[string]json.RawMessage
+	if err := json.Unmarshal(data, &segment); err != nil {
+		return err
+	}
+	isLeaf := true
+	var isAccount bool
```
gfyrag: Just to be homogeneous with the codebase, can you group multiple vars definition?

internal/chart.go
```diff
@@ -0,0 +1,261 @@
...
+			}
+			var pattern *string
+			{
+				var segment map[string]any
+				err := json.Unmarshal(value, &segment)
+				if err != nil {
+					return fmt.Errorf("invalid segment: %v", err)
+				}
+				if pat, ok := segment[PATTERN_KEY]; ok {
+					if pat, ok := pat.(string); ok {
```
gfyrag, line 97: What if the type is not a string?

internal/chart.go
gfyrag: I guess here we allow multiple segment?

internal/chart.go
```diff
@@ -0,0 +1,261 @@
...
+					return fmt.Errorf("cannot have a variable segment without a pattern")
+				}
+				if fixedSegments == nil {
+					fixedSegments = map[string]ChartSegment{}
+				}
+				fixedSegments[key] = segment
+			}
+			isLeaf = false
+		} else if key == SELF_KEY {
+			if string(value) != "{}" {
```
gfyrag: "{       }" is still an empty object ;)

gfyrag: Could be passed by api.

gfyrag (commented):
Sorry for the poor review, but this is very complicated to read.
The code is not in charge.
The spec is too complicated for the use case.

internal/chart_test.go
```diff
@@ -0,0 +1,306 @@
...
+                ".rules": {}
+            }
+        }
+    }
+}`
+	expected := ChartOfAccounts{
+		"banks": {
+			VariableSegment: &ChartVariableSegment{
+				Label:   "iban",
+				Pattern: "[0-9]{10}",
```
gfyrag: Maybe i'm wrong : 
What if you pass abcd1234 as account?
Since there is no "$" and "^" in the regexp, I guess the regex matching will pass?

internal/chart_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+	require.Equal(t, expected, chart)
+
+	value, err := json.MarshalIndent(&chart, "", "    ")
+	require.NoError(t, err)
+	require.JSONEq(t, source, string(value))
+}
+
+func TestInvalidFixedSegment(t *testing.T) {
+	src := `{
```
gfyrag: Not big deal, but please use t.Parallel() when you can (in a perfect world : every time)

internal/chart.go
```diff
@@ -0,0 +1,274 @@
...
+				return nil, ErrInvalidAccount{path, nextSegment}
+			}
+		}
+	}
+	return nil, ErrInvalidAccount{path, nextSegment}
+}
+func (c *ChartOfAccounts) FindAccountSchema(account string) (*ChartAccount, error) {
+	schema, err := findAccountSchema([]string{}, map[string]ChartSegment(*c), nil, strings.Split(account, ":"))
+	if err != nil {
+		if account == "world" {
```
gfyrag, line 255: I guess this exception can be discussed with the product.
Some clients could want to avoid the usage of `world`.

internal/chart_test.go
gfyrag: Also, when you can, can you use table driven tests?
(A vast majority of the codebase use this pattern, and it feels completely usable in almost all tests of this file)

PR #1123, opened 2025-11-11
chore: upgrade to go 1.25

sylr:


go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/formancehq/ledger
 
 go 1.24.4
```
flemzord, line 3: upgrade ?

gfyrag (dismissed):

go.mod
sylr: AFAIK the go directive is here only to tell that we are using specific features introduced in a given go version. As we don't use specific go 1.25 features we don't need to change it thus allowing people to still be able to build the ledger with go 1.24. [](edited-with-Axolo)

PR #1154, opened 2025-11-24
chore: Update standalone caddyfile

BrieucCaillot:


examples/standalone/Caddyfile
```diff
@@ -25,7 +25,7 @@
 	debug
 }
 
-localhost:80 {
+localhost:80, gateway:80 {
```
flemzord: :80 ? It's more simple

PR #1168, opened 2025-11-28
fix: missing inserted at in logs

gfyrag:
Also fixes stop issue of exporters.

---
## Summary by cubic
Fixes missing transaction.insertedAt in log records and improves exporter stop handling to prevent stuck replication. Also adds an index on logs.id for faster fetches.

- **Bug Fixes**
  - Properly stop exporters and pipelines via stopChannel and context cancellation.
  - Skip “already started” errors when restoring pipelines.
  - Add debug logs in ClickHouse driver and pipeline to trace batches and progress.
  - Provisioner: add Dockerfile syntax directive and make registry configurable in justfile.

- **Migration**
  - Backfill transaction.insertedAt in logs with a batched update and pg_notify progress.
  - Create a concurrent index on logs.id to speed up queries.

<sup>Written for commit e6ada90c10e1d254ad0a373dd0c127fdb457453a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/internal/replication/pipeline.go
+++ b/internal/replication/pipeline.go
@@ -63,13 +63,21 @@ type PipelineHandler struct {
 }
 
 func (p *PipelineHandler) Run(ctx context.Context, ingestedLogs chan uint64) {
+	p.logger.Debugf("Pipeline started.")
 	nextInterval := time.Duration(0)
+
+	stop := func(ch chan error) {
+		p.logger.Debugf("Pipeline terminated.")
+		close(ch)
+	}
+
 	for {
 		select {
 		case ch := <-p.stopChannel:
-			close(ch)
+			stop(ch)
 			return
 		case <-time.After(nextInterval):
+			p.logger.Debugf("Fetch next batch.")
 			var builder query.Builder
 			if p.pipeline.LastLogID != nil {
 				builder = query.Gt("id", *p.pipeline.LastLogID)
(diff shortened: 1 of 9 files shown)


internal/replication/pipeline.go
```diff
@@ -85,38 +93,57 @@ func (p *PipelineHandler) Run(ctx context.Context, ingestedLogs chan uint64) {
 			if err != nil {
 				p.logger.Errorf("Error fetching logs: %s", err)
 				select {
-				case <-ctx.Done():
+				case ch := <-p.stopChannel:
```
paul-nicolas, line 96: The bot is right here, do you want to wait forever here ?

internal/replication/pipeline.go
```diff
@@ -85,38 +93,57 @@ func (p *PipelineHandler) Run(ctx context.Context, ingestedLogs chan uint64) {
...
+					_, err := p.exporter.Accept(exportContext, collectionutils.Map(logs.Data, func(log ledger.Log) drivers.LogWithLedger {
+						return drivers.LogWithLedger{
+							Log:    log,
+							Ledger: p.pipeline.Ledger,
+						}
+					})...)
+					errChan <- err
+				}()
+				select {
+				case err := <-errChan:
```
paul-nicolas, line 124: Same here, do you want to select also on the context ?

internal/replication/pipeline.go
gfyrag: Yes, the pipeline has a Stop(ctx) method.
The cancellation should not be done using the context, but the dedicated method.

internal/replication/pipeline.go
gfyrag: Same response ^^

paul-nicolas (approved):

PR #1174, opened 2025-12-01
feat: improve migration performances

gfyrag:
## Summary by cubic
Speed up v2.3 migration scripts by switching to row-number batching, adding targeted indexes, and narrowing updates. This cuts runtime and lock contention on large datasets; the heavy PCV rebuild step is disabled in this backport.

- **Refactors**
  - Use window row_number + range filtering instead of offset/limit across migrations (17, 18, 28, 29, 31, 32, 33).
  - Add temporary staging tables and indexes (with included columns) to accelerate joins and updates.
  - Limit writes (e.g., set inserted_at only when null; process only NEW_TRANSACTION/REVERTED_TRANSACTION logs).
  - Disable the costly PCV recompute in migration 27 to avoid long-running work.

<sup>Written for commit 38fc49cbeb355b343c981a333e9ef5782849cd7c. Summary will update automatically on new commits.</sup>

### DIFF
--- a/internal/storage/bucket/migrations/27-fix-invalid-pcv/up.sql
+++ b/internal/storage/bucket/migrations/27-fix-invalid-pcv/up.sql
@@ -1,67 +1,6 @@
 do $$
-	declare
-		_offset integer := 0;
-		_batch_size integer := 1000;
 	begin
-		set search_path = '{{ .Schema }}';
-
--- 		drop table if exists moves_view;
---
--- 		create table moves_view as
--- 		select transactions_seq, public.aggregate_objects(jsonb_build_object(accounts_address, volumes)) as volumes
--- 		from (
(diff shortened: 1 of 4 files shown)


internal/storage/bucket/migrations/27-fix-invalid-pcv/up.sql
```diff
@@ -5,63 +5,63 @@ do $$
...
-
-			_offset = _offset + _batch_size;
-
-			perform pg_notify('migrations-{{ .Schema }}', 'continue: ' || _batch_size);
-
-			commit;
-		end loop;
-
-		drop table if exists moves_view;
+-- 		drop table if exists moves_view;
```
flemzord: Remove ?

PR #1182, opened 2025-12-11
feat(api/v2): Add transactions summary aggregation with counts

Abdel-Monaam-Aouini:
Add a new GET /transactions/summary endpoint in API v2 that provides a summary of transactions for a specified account, including transaction count and sum per asset. The endpoint supports optional time range filtering (start_time and end_time) and asset-specific queries. Includes comprehensive tests and integrates with the ledger controller and storage layers.


Abdel-Monaam-Aouini:
Hi @flemzord  👋,

Please review this PR when you have a moment.

We need this feature in our product — currently, we’re forced to map all transactions with pagination just to compute the total count and sum, which is inefficient and insufficient for our use case.
This PR solves that limitation.

Thanks a lot for your time and help!

PR #1198, opened 2025-12-18
feat: initialize transactions_metadata if needed

gfyrag:


internal/storage/bucket/migrations/11-make-stateless/up.sql
```diff
@@ -292,13 +292,13 @@ as
 $$
 begin
 	insert into transactions_metadata (ledger, transactions_id, revision, date, metadata)
-	values (new.ledger, new.id, (
+	values (new.ledger, new.id, coalesce((
```
paul-nicolas, line 295: modifying an existing migrations won't do it again for clients who already have it done right ?

paul-nicolas (approved):

flemzord (approved):

PR #1272, opened 2026-02-27
fix: make transaction & query templates nullable in openapi

Azorlogh:


openapi/v2.yaml
BrieucCaillot, line 1:
```
 The fix for the PR is to use inline type definitions instead of allOf + $ref:

  # Instead of this (Speakeasy ignores nullable here):
  transactions:
    nullable: true
    allOf:
      - $ref: "#/components/schemas/V2TransactionTemplates"

  # Use this (Speakeasy handles nullable correctly):
  transactions:
    type: object
    nullable: true
    additionalProperties:
      $ref: "#/components/schemas/V2TransactionTemplate"

  Same pattern for queries. You can share this with the ledger team to update PR #1272.

```

BrieucCaillot (approved):

PR #1351, opened 2026-05-19
perf(storage): remove row_number() window function from list paginator

ariel-formance:
## Summary

- Replaces \`row_number() OVER (ORDER BY id DESC)\` with a plain \`ORDER BY\` in both \`columnPaginator\` and \`OffsetPaginator\`
- Adds \`OrderExpression() string\` to the \`Paginator\` interface so the outer CTE wrapper in \`resource.go\` can propagate the correct sort key
- Qualifies the outer \`ORDER BY\` column with \`dataset.\` after \`expand()\` joins, preventing future ambiguity if an expand CTE introduces a same-named column

## Root cause

The window function forced PostgreSQL to materialise **every** row matching the \`WHERE\` clause before \`LIMIT\` could be applied — even though an \`id DESC\` index exists. The planner placed \`WindowAgg → Limit\` instead of \`Limit → Index Scan Backward\`, making page size irrelevant and causing O(N) scans on every list request.

On large production ledgers this produced 30–50 s \`ListTransactions\` queries.

## EXPLAIN ANALYZE comparison (10 000 rows)

**Before** — \`WindowAgg\` blocks early exit:
```
Sort → Limit → WindowAgg → Index Scan Backward
  Rows Removed by Filter: 144
  Planning Time: 0.226 ms | Execution Time: 0.079 ms
```

**After** — \`Limit\` sits directly on the index scan:
```
Limit → Index Scan Backward
  Rows Removed by Filter: 135
  Planning Time: 0.054 ms | Execution Time: 0.047 ms
```

Planning time: **4× faster**. The gain scales linearly with ledger size.

## Test plan

- [x] Full integration suite passes: \`go test -tags it ./internal/storage/...\`
- [ ] CI green


gfyrag (dismissed):
LGTM — the change is correct and the perf gain is real.

Please run the pre-commit hooks before merging (lint, etc.).

gfyrag (commented):
Also: please remove client names from the test (e.g. "Deriv") — this is an OSS repo.

gfyrag (commented):
Actually, let's just drop the test file entirely (). It doesn't assert anything — it only logs EXPLAIN plans. Not worth keeping in the test suite.

gfyrag (commented):
Actually, let's just drop the test file entirely (transactions_query_plan_test.go). It doesn't assert anything — it only logs EXPLAIN plans. Not worth keeping in the test suite.

ariel-formance:
> Actually, let's just drop the test file entirely (transactions_query_plan_test.go). It doesn't assert anything — it only logs EXPLAIN plans. Not worth keeping in the test suite.

Done

ariel-formance:
> Also: please remove client names from the test (e.g. "Deriv") — this is an OSS repo.

Done

gfyrag (approved):

PR #1355, opened 2026-05-22
perf(driver): cache OpenLedger result to eliminate per-request DB queries

ariel-formance:
## Summary

- Every HTTP request to the ledger service ran two DB queries (`GetLedger` + `CountLedgersInBucket`) inside `OpenLedger` before any business logic executed, consuming connection pool slots on every request
- This worsens thundering-herd reconnect bursts: when the pool is under pressure, metadata queries compete with actual transaction queries for the same limited connections
- There was already a `// todo: keep the ledger in cache somewhere` comment in `driver.go` acknowledging this

**What this PR adds:**

1. **TTL cache (60s default)** — `Driver` now holds a mutex-protected `map[string]cachedLedger`. Cache hits skip both DB queries entirely. Explicit eviction on `UpdateLedgerMetadata` and `DeleteLedgerMetadata` makes single-pod behaviour correct; the TTL is a cross-pod safety net.

2. **Singleflight** — `golang.org/x/sync/singleflight` (already in go.mod) collapses concurrent cache-miss requests for the same ledger name into a single DB call. Without this, a post-TTL burst of N concurrent requests would still fan out to N×2 DB queries. There is a double-check inside the singleflight function to handle the case where a previous waiter has already populated the cache.

3. **`aloneInBucket` correctness** — `CountLedgersInBucket` was the second query per request. It's safe to skip on cache hits because `ledgerstore.DefaultFactory` already maintains a shared `*atomic.Bool` per bucket; any `CreateLedger` call updates it immediately for all stores in the same bucket.

## Test plan

- [x] 9 unit tests (`driver_cache_test.go`, no build tag, no DB) — hit before TTL, miss after TTL (1ns TTL), unknown key, eviction, TTL=0 disables cache, concurrent access with `-race`, isolation between two ledger names, evict non-existent key, update existing entry
- [x] `TestOpenLedgerCacheHitNoDBQueries` (integration) — bun query hook on a fresh `bun.DB` instance proves zero DB queries on warm-cache hit
- [x] `TestOpenLedgerConcurrentColdCacheCoalesces` (integration) — 50 concurrent goroutines on cold cache, all succeed, query count well below 50
- [x] `TestOpenLedgerCacheTTLExpiry` (integration) — 50ms TTL, metadata mutated via a second driver instance (bypassing cache), sleep, verify stale entry re-reads from DB
- [x] `TestOpenLedgerCacheEvictionOnMetadataUpdate` / `...OnMetadataDelete` (integration) — eviction wired correctly on both write paths
- [ ] Integration tests require Docker (`go:build it`)

## Related

Part of connection pool saturation fix for prod-us-east-1-deriv. Companion PR: #1354 (retry jitter).

### DIFF
--- a/internal/controller/system/controller.go
+++ b/internal/controller/system/controller.go
@@ -226,25 +226,25 @@ func (ctrl *DefaultController) ListLedgers(ctx context.Context, query common.Pag
 
 func (ctrl *DefaultController) UpdateLedgerMetadata(ctx context.Context, name string, m map[string]string) error {
 	return tracing.SkipResult(tracing.Trace(ctx, ctrl.tracerProvider.Tracer("system"), "UpdateLedgerMetadata", tracing.NoResult(func(ctx context.Context) error {
-		return ctrl.driver.GetSystemStore().UpdateLedgerMetadata(ctx, name, m)
+		return ctrl.driver.UpdateLedgerMetadata(ctx, name, m)
 	})))
 }
 
 func (ctrl *DefaultController) DeleteLedgerMetadata(ctx context.Context, param string, key string) error {
 	return tracing.SkipResult(tracing.Trace(ctx, ctrl.tracerProvider.Tracer("system"), "DeleteLedgerMetadata", tracing.NoResult(func(ctx context.Context) error {
-		return ctrl.driver.GetSystemStore().DeleteLedgerMetadata(ctx, param, key)
+		return ctrl.driver.DeleteLedgerMetadata(ctx, param, key)
 	})))
 }
 
 func (ctrl *DefaultController) DeleteBucket(ctx context.Context, bucket string) error {
 	return tracing.SkipResult(tracing.Trace(ctx, ctrl.tracerProvider.Tracer("system"), "DeleteBucket", tracing.NoResult(func(ctx context.Context) error {
-		return ctrl.driver.GetSystemStore().DeleteBucket(ctx, bucket)
+		return ctrl.driver.DeleteBucket(ctx, bucket)
 	})))
 }
 
 func (ctrl *DefaultController) RestoreBucket(ctx context.Context, bucket string) error {
 	return tracing.SkipResult(tracing.Trace(ctx, ctrl.tracerProvider.Tracer("system"), "RestoreBucket", tracing.NoResult(func(ctx context.Context) error {
-		return ctrl.driver.GetSystemStore().RestoreBucket(ctx, bucket)
+		return ctrl.driver.RestoreBucket(ctx, bucket)
 	})))
 }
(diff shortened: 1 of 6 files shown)


flemzord:
This PR adds an in-memory cache around `Driver.OpenLedger` to avoid reading `_system.ledgers` and running `CountLedgersInBucket` on every ledger open. It also adds `singleflight` so concurrent cold-cache opens for the same ledger are collapsed into one DB lookup. The goal is clearly to reduce DB pressure on hot paths that repeatedly call `OpenLedger`.

The performance intent makes sense, but this introduces a new invalidation contract: every mutation of `_system.ledgers` must either go through a cache-aware driver method or explicitly evict the affected cache entries. That contract is not fully respected today by the production controller path (`GetSystemStore().UpdateLedgerMetadata/DeleteLedgerMetadata`) nor by bucket-level mutations such as delete/restore, so `OpenLedger` can return stale ledger state until the TTL expires.

There is also an operational impact in multi-instance deployments: this cache is local to each process. If one Ledger instance updates metadata, deletes/restores a bucket, or otherwise changes `_system.ledgers`, the other instances will keep their own independent cached values until their local TTL expires. That means different Ledger instances can temporarily disagree about the same ledger’s metadata or deleted/restored state.

flemzord:
Thanks for working on this. I understand the motivation: `OpenLedger` is called very often, and avoiding repeated `_system.ledgers` reads looks like an attractive optimization.

That said, we should close this PR and not add this cache inside Ledger.

The main issue is that this cache makes correctness depend on local in-process invalidation, while Ledger is normally deployed with multiple processes/pods. Metadata updates, bucket delete/restore, ledger creation in the same bucket, and other system-store changes can happen through another process. A local cache cannot observe those changes reliably.

The impact is not only stale metadata. Some Ledger storage paths depend on the `aloneInBucket` flag to decide whether queries can skip `WHERE ledger = ?`. If one process caches a ledger while it is alone in a bucket, then another process creates a second ledger in that bucket, cache hits can keep using stale `aloneInBucket=true` until TTL expiry. That can make reads query the whole bucket instead of the requested ledger, which is a data isolation/correctness issue.

There are also tricky race cases around bucket delete/restore and in-flight `OpenLedger` calls. These are solvable in a single process with more generation bookkeeping, but they do not solve the cross-process consistency problem.

So the tradeoff is not acceptable for Ledger core: we would reduce a small DB read, but introduce a distributed cache-invalidation problem on metadata, bucket lifecycle, and query scoping. If we want to optimize this path, we should instead look at database-level improvements, request-scoped caching, or another design where invalidation is explicit and shared across all Ledger instances.

Closing this as not planned.

PR #1362, opened 2026-06-04
fix(openapi): require resource discriminator in V2QueryParams oneOf branches

Dav-14:
## TL;DR

Align the OpenAPI contract with what the server has always enforced. **No behavioral change** — only the diagnostic clients see when they omit `resource`.

## Background — the requirement was always there

The server has always rejected `QueryTemplate` payloads without a `resource`. `QueryTemplate.Validate()` calls `GetResourceSchema(q.Resource)` immediately, and the `default:` branch of that switch returns `"unknown resource kind: <empty>"` for any missing or invalid value. There is no path — through any client, SDK, or raw HTTP — that lets a request without `resource` succeed. Never has been.

The OpenAPI spec, however, didn't say so. `V2QueryParams.oneOf` declared four branches each with a `resource` enum, but none marked `resource` as required. Under JSON Schema semantics an `enum` inside `properties` only fires when the property is **present**, so an absent `resource` satisfies all four branches vacuously and the `oneOf` rejects with the cryptic `"oneOf failed, subschemas N matched"`.

## The customer-visible bug

`formancehq/studio-apps` validates customer manifests with `santhosh-tekuri/jsonschema/v6` (strict JSON Schema, no `discriminator` support). A customer authored:

```yaml
queries:
  user_balance:
    resource: accounts        # parent-level discriminator (server reads this)
    params:
      pageSize: 1              # contract said no `resource` needed here
    body: { $match: { address: "users:$user_id" } }
```

and got `"oneOf failed, subschemas 0, 1 matched"` at `POST /manifests`. The server would have rejected it too (with the clearer `"unknown resource kind"`), but the contract was the source of the confusion — the customer was looking at the OpenAPI spec, not the Go code.

## Change

Three layers, kept consistent end-to-end:

| Layer | File | Change |
|---|---|---|
| Spec | `openapi/v2.yaml` | Add `discriminator: { propertyName: resource }` on `V2QueryParams` and `required: [resource]` on each of the four `oneOf` branches. The `discriminator` documents intent for OAS-aware tooling; `required` is the **portability fix** that survives projection to every JSON Schema dialect downstream consumers might use (pure JSON Schema validators drop `discriminator` on compile). |
| Server | `internal/query_template.go` | `checkForExtraFields` now accepts `resource` inside `params` alongside `endTime`/`startTime`/`expand`/`sort`/`pageSize`. Without this, clients honoring the new contract would have their `resource` field rejected as unknown. Server semantics are otherwise unchanged — the parent-level `QueryTemplate.Resource` remains the source of truth; `params.resource` is silently ignored if present. |
| Generated | `openapi.yaml`, `docs/api/README.md`, `pkg/client/**` | Regenerated via `just pc`. |

`discriminator.mapping` is intentionally omitted — the `oneOf` branches are inline (no `$ref` targets), so there are no component names to map to. `propertyName` alone is the right minimal declaration.

## Retrocompatibility

**Wire-compatible.** No payload that was accepted before is rejected now, and no payload that was rejected before is accepted now. The only observable difference is the error *message* on the rejected path:

| Path | Before | After |
|---|---|---|
| Strict JSON Schema validator (studio-apps, terraform-hcp-proxy) | `"oneOf failed, subschemas 0, 1 matched"` | `"required property 'resource' missing"` |
| Server (any client that bypassed spec validation) | `"unknown resource kind: "` | `"unknown resource kind: "` (unchanged) |

## Go SDK impact

Speakeasy regen flips `Resource` from optional pointer to required value on the four `QueryTemplate*Params` structs:

```diff
 type QueryTemplateAccountParams struct {
-    Resource      *V2QueryParams1Resource `json:"resource,omitempty"`
+    Resource      V2QueryParams1Resource  `json:"resource"`
     ...
 }
```

Consumer code passing `Resource` by pointer (`&v`) or dereferencing it (`*p.Resource`) no longer compiles against the regenerated client. The fix is to use the value directly. Any nil-check (`if p.Resource != nil`) was already vacuous — the server has always rejected payloads without a resource — so the break only forces consumers to drop dead code, not handle a new failure mode.

## Test plan

- [x] `go test ./internal/...` green
- [x] Speakeasy regen successful; only `Resource` field optionality changed across the four named param classes
- [x] Payloads with `resource` (all four variants) validate
- [x] Payloads without `resource` now fail with `"required property missing"` instead of `"multiple branches matched"`
- [ ] Reviewer: confirm Go SDK shape change lands in the next release with appropriate notes

## Reference

Validated against `santhosh-tekuri/jsonschema/v6` in `formancehq/studio-apps`. Regression test `TestValidate_V32QueryParamsRequiresResourceDiscriminator` added in studio-apps#217.


gfyrag (dismissed):

Dav-14:
Heads-up @gfyrag @Azorlogh — small PR but worth reading the framing.

**TL;DR: OpenAPI contract alignment, no behavior change.** `V2QueryParams.oneOf` never marked `resource` as required, even though the server has always rejected payloads without one via `GetResourceSchema`. Studio-apps customers using `santhosh-tekuri/jsonschema/v6` were getting `"oneOf failed, subschemas N matched"` instead of a clean `"required property missing"`.

Three layers in one commit:
- `required: [resource]` + `discriminator` in `openapi/v2.yaml`
- `checkForExtraFields` accepts `resource` in `internal/query_template.go` (so new clients honoring the contract aren't rejected for sending the field the contract now requires)
- Speakeasy regen

**Go SDK heads-up:** `Resource` flips from `*ptr` to value on the four `QueryTemplate*Params` structs. Consumer code passing it by pointer no longer compiles against the new client — worth a release note. Any nil-check was already vacuous (server has always required it), so the break only forces dropping dead code, not handling a new failure mode.

gfyrag (approved):

PR #1395, opened 2026-06-09
fix(storage): restore transaction address array indexes

flemzord:
## Summary
- add migration 51 to recreate the GIN index on transactions.sources_arrays
- add migration 52 to recreate the GIN index on transactions.destinations_arrays
- use IF NOT EXISTS and the existing concurrent-index template pattern
- keep MinimalSchemaVersion at 50 because these indexes are not mandatory for the minimal schema

## Tests
- go test -tags it ./internal/storage/bucket
- git diff --check

### DIFF
--- a/internal/storage/bucket/migrations/51-add-transactions-sources-array-index/notes.yaml
+++ b/internal/storage/bucket/migrations/51-add-transactions-sources-array-index/notes.yaml
@@ -1,0 +1,1 @@
+name: Add transactions sources array index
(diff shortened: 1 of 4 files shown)


internal/storage/bucket/default_bucket.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 // stateless version (+1 regarding directory name, as migrations start from 1 in the lib)
-const MinimalSchemaVersion = 50
+const MinimalSchemaVersion = 51
```
gfyrag: Since index are not mandatory, you can keep the original MinimalSchemaVersion

flemzord: Done, I kept `MinimalSchemaVersion` at `50` and pushed the same adjustment to the release/v2.4 PR as well. The indexes remain as migrations 51/52 only.

gfyrag (approved):

PR #1404, opened 2026-06-11
feat(storage): add config to disable alone-in-bucket query optimization

sylr:
newScopedSelect skips the `ledger = ?` predicate when a ledger is the only one in its bucket, to avoid a degraded ~100%-selectivity seq scan. Add a --disable-ledger-scope-optimization serve flag that forces the predicate to always be emitted, as a performance/safety escape hatch.

The config threads from the serve flag through storage.ModuleConfig and driver.ModuleConfig down to a new ledgerstore store option. Default keeps current behavior (optimization enabled).

Constraint: disabling must only ever force the always-correct scoped query, never remove the predicate
Rejected: gate on aloneInBucket hint alone | operators need a deterministic override independent of bucket population
Confidence: high
Scope-risk: narrow
Directive: optimization is wired only into `serve`; worker/buckets-upgrade keep it on (correctness-neutral)
Not-tested: end-to-end fx wiring of the flag (covered by unit-level SQL rendering + build)

### DIFF
--- a/internal/storage/ledger/store.go
+++ b/internal/storage/ledger/store.go
@@ -32,6 +32,11 @@ type Store struct {
 	// (e.g. when a new ledger is created) immediately affects all stores.
 	aloneInBucket *atomic.Bool
 
+	// disableScopedSelectOptimization, when true, forces newScopedSelect to always
+	// emit the `ledger = ?` predicate, ignoring the aloneInBucket hint. It is a
+	// kill switch for the alone-in-bucket optimization (see newScopedSelect).
+	disableScopedSelectOptimization bool
+
 	tracer                             trace.Tracer
 	meter                              metric.Meter
 	checkBucketSchemaHistogram         metric.Int64Histogram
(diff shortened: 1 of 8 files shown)


gfyrag (approved):

internal/storage/ledger/store.go
```diff
@@ -177,9 +182,12 @@ func (store *Store) LockLedger(ctx context.Context) (*Store, bun.IDB, func() err
 // name to use the composite index (ledger, id) efficiently.
 //
 // This relies on aloneInBucket being up to date (shared across the bucket).
+//
+// When disableScopedSelectOptimization is set, the optimization is bypassed and
+// the `ledger = ?` predicate is always emitted.
 func (store *Store) newScopedSelect() *bun.SelectQuery {
 	q := store.db.NewSelect()
-	if store.aloneInBucket == nil || !store.aloneInBucket.Load() {
+	if store.disableScopedSelectOptimization || store.aloneInBucket == nil || !store.aloneInBucket.Load() {
```
flemzord, line 190: This optimization depends on a process-local flag that is only refreshed by OpenLedger/CreateLedger. Long-lived stores such as replication pipelines can keep the flag true after another process creates a second ledger in the same bucket, so calls through resource_logs can read unscoped logs. Please keep the ledger predicate for long-lived readers or refresh this from the DB before skipping it.

PR #1409, opened 2026-06-11
perf(storage): conditional MATERIALIZED CTE fence for ListTransactions

sylr:
## Problem

A selective wallet-history `ListTransactions` attached `ORDER BY id DESC LIMIT n` to the same select as a JSONB `@>` filter, so the planner chose an abort-early `transactions_id_desc` index walk that scanned ~2.3M rows (**16.1s**) to return 16. Statistics can't fix this (JSONB `@>` selectivity over-estimate + a value↔id correlation no stats object can express).

## Fix

Wrap the filtered dataset in a **`MATERIALIZED`** CTE (optimization fence) and move `ORDER BY id DESC LIMIT n` to the **outer** select. The planner then uses the GIN `BitmapOr` over the filtered set and the `LIMIT` becomes a trivial top-N — **~7.7ms (~2,000×)**, verified on a prod read-replica. No schema change; the required GIN indexes already exist (migrations 43/51/52).

The fence is **conditional** — applied only when the filter contains a "needle" containment predicate (`account`/`source`/`destination`/`metadata`), via a new opt-in `DatasetFencer` interface implemented only by the transactions handler. Unfiltered/range-only lists keep the historical non-materialized shape, where abort-early is the faster plan.

## How

- `DatasetFencer[Opts]` opt-in interface; `transactionsResourceHandler.ShouldFenceDataset` implements the static needle rule.
- `Paginator` split into `ApplyCursorPredicate` (keyset/offset predicate stays **inside** the fenced CTE) and `ApplyWindow` (`LIMIT`/`OFFSET` on the **outer** select, no `ORDER BY`). The existing outer `ORDER BY` remains the single order source and is applied before the window. **Non-fence SQL is unchanged.**
- `expand()` gains a `materialized bool` (uses `bun.NewWithQuery(...).Materialized()`); `GetOne`/`Count` are unaffected (pass `false`).

## Tests

- Unit: table-driven `ShouldFenceDataset` (needle → fence; id/timestamp/reference/reverted → no fence).
- Integration: SQL-shape via a `bun.QueryHook` asserting `AS MATERIALIZED` present for account/metadata filters and **absent** for unfiltered/timestamp-only; multi-page `next`/`previous` traversal under the fence; `effectiveVolumes` under the fence. Existing `TestTransactionsList` (every filtered case now runs through the fence) still passes.

## Known limitation (correctness-safe)

A negated containment (`NOT metadata @> {x}`) or a non-selective needle value (`account="world"`) may fence without benefit, but never changes the result. Deliberately deferred per a static-vs-cost-probe trade-off.

### DIFF
--- a/internal/storage/common/resource.go
+++ b/internal/storage/common/resource.go
@@ -220,18 +220,13 @@ func (r *ResourceRepository[ResourceType, OptionsType]) buildFilteredDataset(bui
 	return r.resourceHandler.Project(q, dataset)
 }
 
-// expand wraps the filtered dataset as a "dataset" CTE and left-joins any requested expand CTEs.
-// When materialized is true the CTE is emitted as "WITH dataset AS MATERIALIZED (...)", an
-// optimization fence that forces the planner to evaluate the filtered set once instead of
-// inlining it (which would let an outer LIMIT leak back onto the scan). See DatasetFencer.
-func (r *ResourceRepository[ResourceType, OptionsType]) expand(dataset *bun.SelectQuery, q ResourceQuery[OptionsType], materialized bool) (*bun.SelectQuery, error) {
-	outer := dataset.NewSelect()
-	if materialized {
-		outer = outer.WithQuery(bun.NewWithQuery("dataset", dataset).Materialized())
-	} else {
-		outer = outer.With("dataset", dataset)
-	}
-	dataset = outer.
+// expand wraps the given select as the "dataset" CTE and left-joins any requested expand CTEs,
+// which reference it via "select ... from dataset". The MATERIALIZED fence (when fencing) lives one
+// level deeper, inside the select passed here (see Paginate), so the expands always join against the
+// page rather than the full filtered set.
+func (r *ResourceRepository[ResourceType, OptionsType]) expand(dataset *bun.SelectQuery, q ResourceQuery[OptionsType]) (*bun.SelectQuery, error) {
+	dataset = dataset.NewSelect().
+		With("dataset", dataset).
 		ModelTableExpr("dataset").
 		ColumnExpr("*")
 
@@ -274,7 +269,7 @@ func (r *ResourceRepository[ResourceType, OptionsType]) GetOne(ctx context.Conte
 		return nil, err
 	}
 
-	finalQuery, err = r.expand(finalQuery, query, false)
+	finalQuery, err = r.expand(finalQuery, query)
 	if err != nil {
 		return nil, err
 	}
@@ -411,7 +406,25 @@ func (r *PaginatedResourceRepository[ResourceType, OptionsType]) Paginate(
 	if fence {
 		// The cursor/offset predicate is part of "the filtered set", so it stays inside the fence.
 		finalQuery = paginator.ApplyCursorPredicate(finalQuery)
-		finalQuery, err = r.expand(finalQuery, resourceQuery, true)
+
+		// Fence the filtered set in a MATERIALIZED CTE ("filtered"), then take the page
+		// (ORDER BY + LIMIT/OFFSET) in the "dataset" select that wraps it. The page window lives
+		// here — between the fence and the expand joins — so expand CTEs that reference
+		// "select id from dataset" see only the page, not the whole filtered set. The plan benefit
+		// is preserved: the LIMIT is outside the *materialized* CTE, so the planner still evaluates
+		// "filtered" once via the GIN BitmapOr and the page is a cheap top-N over that result.
+		paged := finalQuery.NewSelect().
+			WithQuery(bun.NewWithQuery("filtered", finalQuery).Materialized()).
+			ModelTableExpr("filtered").
+			ColumnExpr("*").
+			Order(paginator.OrderExpression())
+		paged, err = paginator.ApplyWindow(paged)
+		if err != nil {
+			return nil, fmt.Errorf("applying page window: %w", err)
+		}
+
+		// expand wraps the paged select as the "dataset" CTE the expands join against.
+		finalQuery, err = r.expand(paged, resourceQuery)
 		if err != nil {
 			return nil, fmt.Errorf("expanding results: %w", err)
 		}
@@ -420,7 +433,7 @@ func (r *PaginatedResourceRepository[ResourceType, OptionsType]) Paginate(
 		if err != nil {
 			return nil, fmt.Errorf("paginating request: %w", err)
 		}
-		finalQuery, err = r.expand(finalQuery, resourceQuery, false)
+		finalQuery, err = r.expand(finalQuery, resourceQuery)
 		if err != nil {
 			return nil, fmt.Errorf("expanding results: %w", err)
 		}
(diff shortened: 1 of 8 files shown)


internal/storage/ledger/resource_transactions_test.go
```diff
@@ -0,0 +1,79 @@
...
+		{
+			name:    "destination needle",
+			filters: map[string][]any{"destination": {"bank"}},
+			expect:  true,
+		},
+		{
+			name:    "metadata needle",
+			filters: map[string][]any{"metadata": {map[string]any{"wallet_id": "x"}}},
+			expect:  true,
+		},
```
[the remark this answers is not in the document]
sylr: Good catch — fixed in 4e32a4f0 (and backported to #1410). The metadata-needle case now uses the string `"wallet_id"` instead of a map; the fence decision only checks filter presence so behavior is unchanged, but the fixture is now semantically accurate. Resolving.

sylr:
### Example: how a selective `ListTransactions` is rewritten

Take a wallet-history lookup filtered by account — `GET /v2/.../transactions?query={"$match":{"account":"users:alice"}}` with `pageSize=15`. The `account` filter compiles to a GIN containment ("needle") predicate, so `ShouldFenceDataset` returns `true`.

**Before this patch** (current `main` shape — `ORDER BY … LIMIT` sits *inside* the `dataset` CTE):

```sql
WITH "dataset" AS (
  SELECT *
  FROM (
    SELECT ledger, id, timestamp, reference, inserted_at, updated_at, postings,
           sources, destinations, sources_arrays, destinations_arrays, template,
           metadata, transactions.reverted_at AS reverted_at
    FROM "_default".transactions
    WHERE (ledger = 'ledger0')
  ) dataset
  WHERE (sources @> '["users:alice"]' OR destinations @> '["users:alice"]')
  ORDER BY id DESC          -- ← attached to the same select as the filter
  LIMIT 16                  -- ← pageSize + 1 overfetch; triggers abort-early
)
SELECT * FROM "dataset" ORDER BY "dataset"."id" DESC;
```

→ the planner assumes `ORDER BY id DESC LIMIT 16` will short-circuit and picks an **abort-early `Index Scan using transactions_id_desc`**, applying the `@>` filter per row. When the matching rows are sparse and old, it walks most of the table (`Rows Removed by Filter` in the millions).

**With this patch** (`account` is a needle → fenced):

```sql
WITH "dataset" AS MATERIALIZED (          -- ← (1) optimization fence
  SELECT *
  FROM (
    SELECT ledger, id, timestamp, reference, inserted_at, updated_at, postings,
           sources, destinations, sources_arrays, destinations_arrays, template,
           metadata, transactions.reverted_at AS reverted_at
    FROM "_default".transactions
    WHERE (ledger = 'ledger0')
  ) dataset
  WHERE (sources @> '["users:alice"]' OR destinations @> '["users:alice"]')
)                                          -- ← (2) no ORDER BY / LIMIT inside
SELECT * FROM "dataset"
ORDER BY "dataset"."id" DESC
LIMIT 16;                                  -- ← (3) ORDER BY + LIMIT on the outer select
```

→ fenced, the CTE is "the filtered set", so the planner uses the **GIN `BitmapOr`** over `sources`/`destinations` and the outer `LIMIT` becomes a trivial top-N heapsort over the (small) result. On the prod repro this took the same query from **~16.1 s → ~7.7 ms**.

**The only two mechanical changes** are: `AS MATERIALIZED` on the CTE, and moving `ORDER BY … LIMIT pageSize+1` from inside the CTE to the outer `SELECT`.

A few details preserved:
- **PIT** (`timestamp <= $pit`) and the keyset **cursor** (`AND id <= $cursor` on `next`/`previous` pages) stay *inside* the fenced CTE — they're part of the filtered set. Only `ORDER BY` + `LIMIT` move out.
- A `metadata[...]`/`source=`/`destination=` filter is rewritten identically (it's also a needle → `metadata @> {…}` / `sources @> …`).
- An **unfiltered** or range-only (`id`/`timestamp`) list is **not** fenced — it keeps the *Before* shape exactly, where abort-early is the faster plan.

internal/storage/common/resource.go
```diff
@@ -344,27 +390,59 @@ func (r *PaginatedResourceRepository[ResourceType, OptionsType]) Paginate(
...
+	finalQuery, err := r.buildFilteredDataset(buildCtx)
 	if err != nil {
-		return nil, fmt.Errorf("expanding results: %w", err)
+		return nil, fmt.Errorf("building filtered dataset: %w", err)
+	}
+
+	if fence {
+		// The cursor/offset predicate is part of "the filtered set", so it stays inside the fence.
+		finalQuery = paginator.ApplyCursorPredicate(finalQuery)
+		finalQuery, err = r.expand(finalQuery, resourceQuery, true)
```
flemzord: **[P2] Fenced path expands before applying the page window — `effectiveVolumes` aggregates over the full filtered set instead of the page.**

The `effectiveVolumes` expand CTE filters with `transactions_id in (select id from dataset)`. Before this PR, `dataset` always contained `ORDER BY ... LIMIT pageSize+1`, so the moves aggregation (DISTINCT ON + window function + two levels of `aggregate_objects`) was bounded to at most `pageSize+1` ids. In the fenced path, `dataset` is now the entire materialized filtered set, and the LEFT JOIN + outer `LIMIT` cannot bound the expand's work.

So a request like `account=alice&expand=effectiveVolumes` now aggregates effective volumes for **every** transaction in alice's history to return one page — the optimized path scales linearly with account history. Note this hits the *nominal* selective-needle case the PR optimizes (a busy but selective account), not just the documented `account=world` limitation. `TestListTransactionsFencedWithEffectiveVolumes` can't catch it (3 transactions, filtered set == page).

**Suggested fix: page between the fence and the expand.** The fence only requires `ORDER BY + LIMIT` to be outside the *materialized* CTE, not on the outermost select — a top-N over the already-materialized set keeps the plan benefit:

```sql
WITH dataset AS (
    WITH filtered AS MATERIALIZED (
        <filter + PIT + keyset predicate>      -- the fence, unchanged
    )
    SELECT * FROM filtered ORDER BY id DESC LIMIT 17   -- top-N over the materialized set
)
SELECT * FROM dataset
LEFT JOIN expand0 ON ...        -- "in (select id from dataset)" = page ids again ✓
ORDER BY dataset.id DESC
```

Naming the paged select `dataset` restores the implicit `Expand` contract ("dataset = the rows being returned") without touching the handler, and the `ApplyCursorPredicate`/`ApplyWindow` split stays useful as-is — only the window's application point moves (intermediate select instead of outer). As a bonus, this also bounds the expand cost for the known non-selective-needle limitation.

Worth adding a SQL-shape regression guard: fenced + expand must have the `LIMIT` *inside* the `dataset` CTE (and `AS MATERIALIZED` on `filtered`).

sylr: Great catch — you're right, and the "selective but busy account" framing is the key point I'd under-weighted (my doc note and `TestListTransactionsFencedWithEffectiveVolumes` only covered filtered-set == page, so neither caught it). Implemented your suggested shape in 94ccd010 (backported to #1410 in fc369499):

```sql
WITH dataset AS (
  WITH filtered AS MATERIALIZED (<filter + PIT + keyset>)
  SELECT * FROM filtered ORDER BY id <dir> LIMIT pageSize+1
)
SELECT * FROM dataset LEFT JOIN <expands on "select id from dataset"> ...
ORDER BY dataset.id <dir>
```

The `LIMIT` stays outside the *materialized* CTE (so `filtered` is still evaluated once via the GIN BitmapOr and the page is a cheap top-N over it), but now lives inside the `dataset` CTE the expands reference — so `effectiveVolumes` aggregates over the page again. The `ApplyCursorPredicate`/`ApplyWindow` split is unchanged; only the window's application point moved to the intermediate paged select. `expand()` lost its now-unused `materialized` parameter since the fence moved one level deeper.

Added the SQL-shape regression guard you asked for (`TestListTransactionsMaterializedFence/fenced + effectiveVolumes pages before the expand reads dataset`): it captures the generated SQL and asserts `AS MATERIALIZED` is present and the page `LIMIT` appears before `select id from dataset`, i.e. the page window is inside the `dataset` CTE the expand reads from.

Thanks for the thorough review.

PR #1410, opened 2026-06-11
perf(storage): conditional MATERIALIZED CTE fence for ListTransactions (backport v2.4)

sylr:
Backport of #1409 to `release/v2.4`.

## Problem

A selective wallet-history `ListTransactions` attached `ORDER BY id DESC LIMIT n` to the same select as a JSONB `@>` filter, so the planner chose an abort-early `transactions_id_desc` index walk that scanned ~2.3M rows (**16.1s**) to return 16.

## Fix

Wrap the filtered dataset in a **`MATERIALIZED`** CTE and move `ORDER BY id DESC LIMIT n` to the **outer** select, so the planner uses the GIN `BitmapOr` (**~7.7ms**, ~2,000×). The fence is **conditional** — applied only when the filter contains a "needle" containment predicate (`account`/`source`/`destination`/`metadata`) via the opt-in `DatasetFencer` interface; unfiltered/range-only lists keep the historical shape. No schema change (GIN indexes already exist).

## Backport notes

- Cherry-picked cleanly from `main` (PR #1409, commit `801a5619`); the only auto-merge was context in `paginator.go`.
- Verified on the v2.4 base: `go build`, `go vet`, and the fence integration tests (`TestListTransactionsMaterializedFence`, `TestListTransactionsFencedPagination`, `TestListTransactionsFencedWithEffectiveVolumes`, `TestShouldFenceTransactionsDataset`) all pass.

See #1409 for full details and the tri-model review.

### DIFF
--- a/internal/controller/ledger/mocks_test.go
+++ b/internal/controller/ledger/mocks_test.go
@@ -156,6 +156,44 @@ func (mr *MockRepositoryHandlerMockRecorder[Opts]) Schema() *gomock.Call {
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "Schema", reflect.TypeOf((*MockRepositoryHandler[Opts])(nil).Schema))
 }
 
+// MockDatasetFencer is a mock of DatasetFencer interface.
+type MockDatasetFencer[Opts any] struct {
+	ctrl     *gomock.Controller
+	recorder *MockDatasetFencerMockRecorder[Opts]
+	isgomock struct{}
+}
+
+// MockDatasetFencerMockRecorder is the mock recorder for MockDatasetFencer.
+type MockDatasetFencerMockRecorder[Opts any] struct {
+	mock *MockDatasetFencer[Opts]
+}
+
+// NewMockDatasetFencer creates a new mock instance.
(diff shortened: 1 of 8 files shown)


sylr:
@coderabbitai — re: the "outside diff range" finding that `mockUseFilter` is missing and the test won't compile:

This is a false positive. `mockUseFilter` (and its `UseFilter` method) is already defined in the same package, in `internal/storage/ledger/utils_test.go:191`:

```go
type mockUseFilter struct {
	filters map[string][]any
}
func (m *mockUseFilter) UseFilter(key string, matchers ...func(any) bool) bool { ... }
```

Both `resource_transactions_test.go` and `utils_test.go` are `package ledger`, so the test files compile together and `&mockUseFilter{...}` resolves fine. The `Tests` CI job is green on this PR, which confirms compilation.

Applying the suggested fix (adding a second `mockUseFilter` definition to `resource_transactions_test.go`) would actually **break** the build with a redeclaration error in the `ledger` package — so I'm intentionally not applying it. Reusing the existing shared mock is the deliberate choice here (it's the same helper `TestCollectAddressFilters` already uses).

No change needed; resolving.

PR #1412, opened 2026-06-16
fix(storage): make logs_blocks primary key composite (ledger, previous)

gfyrag:
## Summary

- Adds bucket migration `51-fix-logs-blocks-pkey-collision` that swaps the `logs_blocks` primary key from `(previous)` to `(ledger, previous)`, allowing multiple ledgers in the same bucket to each own their own chain head (`previous=0`).
- Adds an e2e test that reproduces the production collision (two async-hashed ledgers in the same bucket).

## Context

Fixes [EN-1246](https://formance-team.atlassian.net/browse/EN-1246).

In v2.4.9, `internal/storage/bucket/migrations/38-logs-async-hash-procedure/up.sql:5` declares `previous bigint primary key` on `logs_blocks`. The `create_blocks` procedure initializes the first block of every ledger with ``previous_block = (0, 0, null)::block``, so the very first insert for any ledger writes `previous = 0`. As soon as two ledgers share the same bucket, the second one fails with `logs_blocks_pkey` SQLSTATE 23505.

Confirmed in production via Signoz on `prod-us-east-1-deriv` (ledger `test_brand` colliding with `deriv` in bucket `_default`) and `prod-eu-west-1-hosting` (ledger `payflip-monoledger-async` colliding with `stress-test-ledger-async`). Rate is rock-steady at 120 err/h (us-east-1) and 60 err/h (eu-west-1) over 48h, with zero recovery. Single-pod worker per cluster — no parallel-worker contributor.

## Test plan

- [ ] `go build ./...` passes
- [ ] `go test -tags it ./internal/storage/bucket/...` — migration applies cleanly on a bucket created at migration 38
- [ ] `go test -tags it ./test/e2e/... -run "Logs block async hashing with multiple ledgers"` — new e2e covers the two-ledger collision
- [ ] Existing `Logs block async hashing` single-ledger e2e still passes

[EN-1246]: https://formance-team.atlassian.net/browse/EN-1246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/storage/bucket/migrations/51-fix-logs-blocks-pkey-collision/up.sql
+++ b/internal/storage/bucket/migrations/51-fix-logs-blocks-pkey-collision/up.sql
@@ -1,0 +1,8 @@
+do $$
+	begin
+		set search_path = '{{ .Schema }}';
+
+		alter table logs_blocks drop constraint logs_blocks_pkey;
+		alter table logs_blocks add primary key (ledger, previous);
+	end
+$$;
--- a/internal/storage/bucket/migrations/51-fix-logs-blocks-pkey-collision/notes.yaml
+++ b/internal/storage/bucket/migrations/51-fix-logs-blocks-pkey-collision/notes.yaml
@@ -1,0 +1,1 @@
+name: Fix logs_blocks primary key to be composite (ledger, previous) to allow multiple ledgers in the same bucket
--- a/test/e2e/app_logs_blocks_async_test.go
+++ b/test/e2e/app_logs_blocks_async_test.go
@@ -108,3 +108,100 @@ var _ = Context("Logs block async hashing", func() {
 		})
 	})
 })
+
+var _ = Context("Logs block async hashing with multiple ledgers in the same bucket", func() {
+	var (
+		db                = UseTemplatedDatabase()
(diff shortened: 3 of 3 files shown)


internal/storage/bucket/migrations/51-fix-logs-blocks-pkey-collision/up.sql
```diff
@@ -0,0 +1,8 @@
+do $$
```
NumaryBot, line 1: 🟠 **[major] Require this migration in the minimal schema version**

Because this migration is needed for correctness of async log hashing, adding it without bumping `bucket.MinimalSchemaVersion` means `HasReachMinimalVersion` will still report buckets at version 50 as healthy even though they have the old `(previous)` primary key. In deployments that rely on the health check/readiness before or instead of running bucket upgrades, async workers can continue hitting the same `logs_blocks_pkey` collision on existing buckets until this migration is actually enforced.

NumaryBot (commented):

The SQL migration itself changes the primary key as intended, but the required schema level is not updated, so existing buckets can still be considered healthy without the fix applied.

flemzord (approved):
No actionable issues found.

PR #1421, opened 2026-06-19
fix(storage): handle PIT + metadata filter on aggregated balances when ACCOUNT_METADATA_HISTORY is disabled

### LINKED ISSUE #1416: v2 /aggregate/balances silently returns empty with pit + metadata filter on ACCOUNT_METADATA_HISTORY=DISABLED
## Summary

On Ledger **v2.4.10**, `GET /api/ledger/v2/{ledger}/aggregate/balances?pit=<T>` with a metadata-based `$match` query returns `{"data":{}}` (silently empty) when the ledger has `ACCOUNT_METADATA_HISTORY: DISABLED` — even though:

- The same query **without** `pit` returns the correct balances.
- The same `pit + metadata` query on `/accounts` returns the matched accounts (with current metadata).
- The same `pit + address` query on `/aggregate/balances` returns the correct balances.

The failure is specific to the combination **`/aggregate/balances` + PIT + metadata filter** when history is `DISABLED`. The silent empty response is misleading: downstream services interpret it as "balances reconcile to zero" when in fact the filter matched nothing it could see.

## Reproduction

Local single-node v2.4.10, no auth, freshly created ledger.

```bash
LED=http://localhost:3168/api/ledger/v2/reco-ledger

# 1. Two postings totalling USD/2 = 350
curl -X POST $LED/transactions \
  -H 'Content-Type: application/json' \
  -d '{"postings":[{"source":"world","destination":"merchant:m1:held","amount":100,"asset":"USD/2"}]}'
curl -X POST $LED/transactions \
  -H 'Content-Type: application/json' \
  -d '{"postings":[{"source":"world","destination":"merchant:m2:held","amount":250,"asset":"USD/2"}]}'

# 2. Set account metadata so we can filter
curl -X POST $LED/accounts/merchant:m1:held/metadata -H 'Content-Type: application/json' -d '{"trust":"true"}'
curl -X POST $LED/accounts/merchant:m2:held/metadata -H 'Content-Type: application/json' -d '{"trust":"true"}'

NOW=$(date -u +%Y-%m-%dT%H:%M:%SZ)

# 3. WITHOUT pit → returns the expected balance
curl -X GET $LED/aggregate/balances \
  -H 'Content-Type: application/json' \
  -d '{"$match":{"metadata[trust]":"true"}}'
# → {"data":{"USD/2":350}}   ✓

# 4. WITH pit → silently empty
curl -X GET "$LED/aggregate/balances?pit=$NOW" \
  -H 'Content-Type: application/json' \
  -d '{"$match":{"metadata[trust]":"true"}}'
# → {"data":{}}   ✗  (bug)

# 5. Control — same pit + ADDRESS filter works
curl -X GET "$LED/aggregate/balances?pit=$NOW" \
  -H 'Content-Type: application/json' \
  -d '{"$match":{"address":"merchant:"}}'
# → {"data":{"USD/2":350}}   ✓

# 6. Control — accounts endpoint with pit + metadata works (returns matched accounts)
curl -X GET "$LED/accounts?pit=$NOW" \
  -H 'Content-Type: application/json' \
  -d '{"$match":{"metadata[trust]":"true"}}'
# → 2 matched accounts   ✓
```

Run the same steps on a ledger with `ACCOUNT_METADATA_HISTORY: SYNC` (call it `reco-ledger-2`): step 4 returns `{"data":{"USD/2":350}}` as expected.

## Observed vs expected

| Endpoint                | Filter   | `pit`        | History DISABLED  | History SYNC      |
| ----------------------- | -------- | ------------ | ----------------- | ----------------- |
| `/aggregate/balances`   | metadata | absent       | ✅ 350             | ✅ 350             |
| `/aggregate/balances`   | metadata | present      | ❌ empty (`{}`)    | ✅ 350             |
| `/aggregate/balances`   | address  | present      | ✅ 350             | ✅ 350             |
| `/accounts`             | metadata | present      | ✅ 2 accounts      | ✅ 2 accounts      |

`/accounts` happily resolves metadata under PIT regardless of the history setting, but `/aggregate/balances` does not — its PIT path appears to be coupled to the metadata history table.

Either of these would be acceptable:

1. **Align with `/accounts` behaviour** — when `ACCOUNT_METADATA_HISTORY: DISABLED`, fall back to *current* metadata for the filter (matches the documented "history off keeps only the current state" semantics).
2. **Fail loudly** — return a `4xx` (e.g. `400 VALIDATION`) explaining that PIT-filtering by metadata requires `ACCOUNT_METADATA_HISTORY: SYNC`, so downstream consumers can detect the configuration mismatch.

The current silent empty response is the worst of both worlds.

## Workaround

- Enable `ACCOUNT_METADATA_HISTORY: SYNC` on the target ledger (confirmed working).
- Or use an address-based `$match` filter instead of metadata when PIT is required.

## Impact

The Reconciliation module's existing `Policy` shape passes a metadata query plus a PIT timestamp into `/aggregate/balances` on every reconciliation run ([`internal/api/service/utils.go` in formancehq/reconciliation](https://github.com/formancehq/reconciliation/blob/main/internal/api/service/utils.go)). Against any history-off ledger, that call silently returns `{}`, the comparison harmonises to `{}` on both sides, and reconciliation reports `status: "OK"` with empty `ledgerBalances` — a green check that didn't actually execute. The bug is invisible until someone diffs the result with the underlying transactions.

Any other consumer building on `/aggregate/balances` with PIT + metadata is exposed to the same silent failure.

## Environment

- Ledger: `ghcr.io/formancehq/ledger:v2.4.10`
- Postgres 15-alpine
- No auth, single-node, local docker compose
- Reproduced on macOS (Darwin 25.3.0) — server-side only, client OS shouldn't matter

aalloiteau:
## Summary

- `/aggregate/balances?pit=…` with a metadata `\$match` filter silently returned `{"data":{}}` on ledgers configured with `ACCOUNT_METADATA_HISTORY: DISABLED`, because the PIT path joined the `accounts_metadata` history table that is never populated under that feature flag value.
- Gate the history join on `FeatureAccountMetadataHistory=SYNC` and fall back to the current `accounts.metadata` column otherwise — matching the pattern already used in `resource_accounts.go`, so `/aggregate/balances` and `/accounts` now agree under PIT + metadata filtering.
- Added a regression test (`TestBalancesAggregatesPITWithMetadataHistoryDisabled`) covering both with-PIT and without-PIT cases on a ledger configured with `ACCOUNT_METADATA_HISTORY: DISABLED`. The test fails on `main` (`with_pit` returns `VolumesByAssets{}`) and passes after the fix.

Fixes #1416

## Test plan

- [x] `go test -tags=it -run TestBalancesAggregatesPITWithMetadataHistoryDisabled ./internal/storage/ledger/` — passes after fix; fails on prior code (verified via `git stash`).
- [x] `go test -tags=it -run TestBalancesAggregates ./internal/storage/ledger/` — passes.
- [x] `go test -tags=it -count=1 ./internal/controller/ledger/... ./internal/api/v2/...` — passes.
- [x] Storage test suite green (the unrelated `TestVolumesAggregate/filter_using_metadata_and_PIT` failure is a pre-existing issue on `main`, not introduced by this PR).

### DIFF
--- a/internal/storage/ledger/resource_aggregated_balances.go
+++ b/internal/storage/ledger/resource_aggregated_balances.go
@@ -61,16 +61,27 @@ func (h aggregatedBalancesResourceRepositoryHandler) BuildDataset(query common.R
 		}
 
 		if query.UseFilter("metadata") {
-			subQuery := h.store.newScopedSelect().
-				DistinctOn("accounts_address").
-				ModelTableExpr(h.store.GetPrefixedRelationName("accounts_metadata")).
-				ColumnExpr("first_value(metadata) over (partition by accounts_address order by revision desc) as metadata").
-				Where("accounts_metadata.accounts_address = moves.accounts_address").
-				Where("date <= ?", query.PIT)
+			if h.store.ledger.HasFeature(features.FeatureAccountMetadataHistory, "SYNC") {
+				subQuery := h.store.newScopedSelect().
+					DistinctOn("accounts_address").
+					ModelTableExpr(h.store.GetPrefixedRelationName("accounts_metadata")).
+					ColumnExpr("first_value(metadata) over (partition by accounts_address order by revision desc) as metadata").
+					Where("accounts_metadata.accounts_address = moves.accounts_address").
+					Where("date <= ?", query.PIT)
 
-			ret = ret.
-				Join(`left join lateral (?) accounts_metadata on true`, subQuery).
-				Column("metadata")
+				ret = ret.
+					Join(`left join lateral (?) accounts_metadata on true`, subQuery).
+					Column("metadata")
+			} else {
+				subQuery := h.store.newScopedSelect().
+					TableExpr(h.store.GetPrefixedRelationName("accounts")).
+					ColumnExpr("metadata").
+					Where("accounts.address = moves.accounts_address")
+
+				ret = ret.
+					Join(`left join lateral (?) accounts on true`, subQuery).
+					Column("metadata")
+			}
 		}
 
 		return ret, nil
--- a/internal/storage/ledger/resource_aggregated_balances.go
+++ b/internal/storage/ledger/resource_aggregated_balances.go
@@ -79,7 +79,7 @@ func (h aggregatedBalancesResourceRepositoryHandler) BuildDataset(query common.R
 					Where("accounts.address = moves.accounts_address")
 
 				ret = ret.
-					Join(`left join lateral (?) accounts on true`, subQuery).
+					Join(`left join lateral (?) accounts_metadata on true`, subQuery).
 					Column("metadata")
 			}
 		}
--- a/internal/storage/ledger/balances_test.go
+++ b/internal/storage/ledger/balances_test.go
@@ -19,6 +19,7 @@ import (
 	ledger "github.com/formancehq/ledger/internal"
 	"github.com/formancehq/ledger/internal/storage/common"
 	ledgerstore "github.com/formancehq/ledger/internal/storage/ledger"
+	"github.com/formancehq/ledger/pkg/features"
 )
 
 func TestBalancesGet(t *testing.T) {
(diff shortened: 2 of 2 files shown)


internal/storage/ledger/resource_aggregated_balances.go
```diff
@@ -61,16 +61,27 @@ func (h aggregatedBalancesResourceRepositoryHandler) BuildDataset(query common.R
...
+					Join(`left join lateral (?) accounts_metadata on true`, subQuery).
+					Column("metadata")
+			} else {
+				subQuery := h.store.newScopedSelect().
+					TableExpr(h.store.GetPrefixedRelationName("accounts")).
+					ColumnExpr("metadata").
+					Where("accounts.address = moves.accounts_address")
+
+				ret = ret.
+					Join(`left join lateral (?) accounts on true`, subQuery).
```
NumaryBot: 🟠 **[major] Avoid reusing the accounts alias for metadata joins**

When `ACCOUNT_METADATA_HISTORY` is disabled and a PIT aggregate query includes both a metadata filter and a partial `address` filter, the PIT path already adds a lateral join aliased as `accounts` for `accounts_address_array` above. This new fallback adds a second lateral join with the same `accounts` alias, so queries like `/aggregate/balances?pit=...` with `metadata[...]` and `address=merchant:...` will fail with a duplicate table alias instead of returning balances.

NumaryBot (commented):

The fix handles the reported metadata-only case but introduces a SQL alias conflict for PIT aggregate balance queries that combine metadata filtering with partial address filtering when metadata history is disabled.

NumaryBot (commented):

The change correctly gates the PIT metadata-history join on ACCOUNT_METADATA_HISTORY=SYNC and falls back to the current accounts metadata column otherwise. The added regression test covers the reported disabled-history PIT metadata-filter path, and I did not identify any introduced correctness issues.

_No findings._

gfyrag (approved):
LGTM. Clean targeted fix matching the resource_accounts.go pattern, with a regression test covering the PIT + metadata + partial address case. Will follow up with v2.3 + v2.4 hotfix backports.

PR #1423, opened 2026-06-19
fix(storage): handle PIT + metadata filter on aggregated balances when ACCOUNT_METADATA_HISTORY is disabled (v2.3)

### LINKED ISSUE #1416: v2 /aggregate/balances silently returns empty with pit + metadata filter on ACCOUNT_METADATA_HISTORY=DISABLED

gfyrag:
## Summary

Backport of #1421 to `release/v2.3`.

- `/aggregate/balances?pit=…` with a metadata `\$match` filter silently returned `{"data":{}}` on ledgers configured with `ACCOUNT_METADATA_HISTORY: DISABLED`, because the PIT path joined the `accounts_metadata` history table that is never populated under that feature flag value.
- Gate the history join on `FeatureAccountMetadataHistory=SYNC` and fall back to the current `accounts.metadata` column otherwise — matching the pattern already used in `resource_accounts.go`.
- Adds a regression test (`TestBalancesAggregatesPITWithMetadataHistoryDisabled`) covering with-PIT, without-PIT, and the combined metadata + partial address filter case on a ledger configured with `ACCOUNT_METADATA_HISTORY: DISABLED`.

Notes on the backport:
- Import conflict resolved in `balances_test.go`.
- v2.3's `CommitTransaction` already upserts accounts internally, so the `commitTransactionAndUpsertAccounts` helper from main is unnecessary — the regression test was adapted to call `store.CommitTransaction(ctx, &tx, nil)` directly (separate adjustment commit).

Fixes #1416

## Test plan

- [ ] CI green on this branch

### DIFF
--- a/internal/storage/ledger/balances_test.go
+++ b/internal/storage/ledger/balances_test.go
@@ -403,7 +403,7 @@ func TestBalancesAggregatesPITWithMetadataHistoryDisabled(t *testing.T) {
 		).
 		WithTimestamp(now).
 		WithInsertedAt(now)
-	require.NoError(t, commitTransactionAndUpsertAccounts(ctx, store, &tx))
+	require.NoError(t, store.CommitTransaction(ctx, &tx, nil))
 
 	require.NoError(t, store.UpdateAccountsMetadata(ctx, map[string]metadata.Metadata{
 		"merchant:m1:held": {
@@ -438,7 +438,7 @@ func TestBalancesAggregatesPITWithMetadataHistoryDisabled(t *testing.T) {
 	})
 
 	t.Run("with pit and partial address filter", func(t *testing.T) {
-		ret, err := store.AggregatedVolumes().GetOne(ctx, common.ResourceQuery[ledger.GetAggregatedVolumesOptions]{
+		ret, err := store.AggregatedVolumes().GetOne(ctx, common.ResourceQuery[ledgerstore.GetAggregatedVolumesOptions]{
 			PIT: pointer.For(now.Add(time.Minute)),
 			Builder: query.And(
 				query.Match("address", "merchant:"),
--- a/internal/storage/ledger/balances_test.go
+++ b/internal/storage/ledger/balances_test.go
@@ -436,4 +436,16 @@ func TestBalancesAggregatesPITWithMetadataHistoryDisabled(t *testing.T) {
 		require.NoError(t, err)
 		RequireEqual(t, expected, *ret)
 	})
+
+	t.Run("with pit and partial address filter", func(t *testing.T) {
+		ret, err := store.AggregatedVolumes().GetOne(ctx, common.ResourceQuery[ledger.GetAggregatedVolumesOptions]{
+			PIT: pointer.For(now.Add(time.Minute)),
+			Builder: query.And(
+				query.Match("address", "merchant:"),
+				query.Match("metadata[trust]", "true"),
+			),
+		})
+		require.NoError(t, err)
+		RequireEqual(t, expected, *ret)
+	})
 }
--- a/internal/storage/ledger/balances_test.go
+++ b/internal/storage/ledger/balances_test.go
@@ -6,6 +6,7 @@ import (
 	"database/sql"
 	"github.com/formancehq/ledger/internal/storage/common"
 	ledgerstore "github.com/formancehq/ledger/internal/storage/ledger"
+	"github.com/formancehq/ledger/pkg/features"
 	"math/big"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


internal/storage/ledger/balances_test.go
```diff
@@ -379,3 +380,72 @@ func TestBalancesAggregates(t *testing.T) {
...
+		ret, err := store.AggregatedVolumes().GetOne(ctx, common.ResourceQuery[ledgerstore.GetAggregatedVolumesOptions]{
+			PIT:     pointer.For(now.Add(time.Minute)),
+			Builder: query.Match("metadata[trust]", "true"),
+		})
+		require.NoError(t, err)
+		RequireEqual(t, expected, *ret)
+	})
+
+	t.Run("with pit and partial address filter", func(t *testing.T) {
+		ret, err := store.AggregatedVolumes().GetOne(ctx, common.ResourceQuery[ledger.GetAggregatedVolumesOptions]{
```
NumaryBot: 🔴 **[blocker] Use ledgerstore options in the new query**

In this new subtest, `ledger` is the domain package (`github.com/formancehq/ledger/internal`), which does not define `GetAggregatedVolumesOptions`; the options type lives in `internal/storage/ledger`, imported above as `ledgerstore`. With the `it` tag enabled this file will fail to compile with `undefined: ledger.GetAggregatedVolumesOptions`, so the regression suite cannot run.

NumaryBot (commented):

The production change appears aligned with the intended fallback, but the added integration test references an undefined type and will not compile when the `it` test suite is run.

paul-nicolas (approved):

NumaryBot (commented):

The change correctly gates the PIT metadata history join on the account metadata history feature and falls back to current account metadata when disabled. The added regression test covers the described aggregate-balance scenarios, and no introduced correctness issues were identified.

_No findings._

PR #1430, opened 2026-06-29
refactor(antithesis): chaos error classification + workload cleanup

gfyrag:
## Summary

Audit-driven cleanup of the Antithesis workload error classification.

The original audit found that several drivers tolerated only `IsUnavailable`
on `Sometimes` probes — narrower than the canonical `IsTransient` — turning
chaos-induced creation failures into silent skips so downstream invariants
were never asserted (~21 drivers were affected through `internal.CreateLedger`).

Questioning that fix surfaced a second issue: `IsTransient` itself was a flat
union conflating "did not commit", "may have committed", a gRPC code with
ambiguous semantics, and a local-lifecycle event that wasn't classified at
all. This PR replaces the flat union with a layered predicate set in
`internal/client.go`:

| Predicate | Members | Used by |
|---|---|---|
| `IsTransient` | `Unavailable`, `DeadlineExceeded`, `ReadIndexNotCaughtUp`, `ExternalServiceError` | Retry interceptor |
| `IsAmbiguousCommit` | `DeadlineExceeded` (subset of `IsTransient`) | Drivers asserting post-commit state |
| `IsCanceled` | `Canceled` | Drivers exiting on ctx death |
| `IsTolerated` | `nil ∨ IsTransient ∨ IsCanceled` | **Every `Sometimes` probe** |
| `IsBusinessError` | `NotFound`, `AlreadyExists`, `InvalidArgument`, generic `FailedPrecondition` | Drivers — MUST validate explicitly |
| `IsClassified` | `nil ∨ IsTransient ∨ IsCanceled ∨ IsAborted ∨ IsBusinessError` | **Classify interceptor** |

`Aborted` was deliberately moved OUT of `IsTransient`: the gRPC spec allows
business "please retry" semantics there, and we have no audited guarantee
the server keeps it as a pure transport transient.

A new **classify interceptor** wraps every RPC and emits `Always(IsClassified(err))`
on the way out. An orphan code (Internal panic, ResourceExhausted from a
future rate-limiter, a brand-new gRPC code) surfaces as a single global
Unreachable with the method and code in Details — the workload's
classification map is incomplete and silent paths exist somewhere.

### Other things included

- **Disambiguate three sites in `eventually_correct`** that all shared the
  `"list balance should match getaccount balance"` assertion name (one
  was `Always(false)` — the real bug — and two were `Always(true)`
  coverage markers). Antithesis indexes by name; renamed each to its
  actual semantic.
- **Surface previously silent paths:**
  `parallel_driver_list_transactions` classifies mid-stream `Recv()`
  errors; `parallel_driver_stale_reads` surfaces `Unreachable` on errors
  outside the transient/NotFound set; `parallel_driver_update_query` now
  aligned with queries/indexes.
- **`internal.CheckCreatedTransaction(resp, details)`** = extract + verify
  post-commit volume consistency. Wired into `transactions`,
  `insufficient_funds`, `revert_concurrent` (was missing), `bulk`,
  `account_types`, `numscripts`.
- **Rename the revert family:** `reverts` → `revert_sequential`,
  `concurrent_revert` → `revert_concurrent`, `double_revert` →
  `revert_idempotent`.
- **`internal.GenerateLedgerName(prefix)`** (9 driver-owned ledger naming
  sites migrated).
- **`internal.SingletonContext()` / `DriverContext()`** (18 drivers
  migrated off bare `context.Background()`; `quorum_recovery`'s
  cleanup-path contexts intentionally preserved).
- **`tests/antithesis/README.md`** documenting layout, scheduling-prefix
  convention, the layered predicate set, the classify interceptor,
  assertion naming, and the anti-patterns the audit surfaced.

## Test plan

- [x] `GOROOT= go build ./...` clean on both the main module and the
      workload submodule
- [x] `golangci-lint run --timeout=5m ./...` → 309 → 306 issues
      (3 pre-existing resolved, 0 introduced by either commit)
- [ ] Antithesis 1h k8s run launched at the *original* commit
      (`runId 77c75307768effafbbc981163f837265-56-17`) — predates the
      classify interceptor and the layered predicates. If the report
      comes back clean, relaunch on the final tip to exercise the new
      `Always(IsClassified)` and the narrowed `IsTransient` (any
      `Aborted` will now surface as a finding instead of being silently
      retried).

---
_Migrated from formancehq/ledger-v3-poc#595 on 2026-06-29._


tests/antithesis/workload/internal/client.go
```diff
@@ -84,8 +85,15 @@ func NewGRPCConn() (*grpc.ClientConn, error) {
 
 	if !retryDisabled {
 		opts = append(opts,
+			// Order matters: classify wraps retry. grpc-go applies interceptors
+			// inside-out, so the LAST unary interceptor added is the OUTERMOST.
+			// Putting classify after retry means classify sees the FINAL error
+			// (post-retry) — exactly the one we want to assert on. If we ever
+			// add more interceptors, they must stay between the two.
 			grpc.WithUnaryInterceptor(retryUnaryInterceptor(interceptorAttempts)),
```
NumaryBot: 🟠 **[major] Chain interceptors instead of overwriting retries**

When retry is enabled, these calls register two unary and two stream interceptors with `WithUnaryInterceptor`/`WithStreamInterceptor`. In grpc-go these options store a single interceptor of each kind, so the later classify interceptors replace the retry interceptors rather than wrapping them. Under chaos, `DeadlineExceeded`, `READ_INDEX_NOT_CAUGHT_UP`, and `ExternalServiceError` will no longer be retried by the custom retry loop, causing drivers to return/skip paths that this PR expects to drive to a definitive outcome; use the chain interceptor options to preserve both.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1430#issuecomment-4833720199

tests/antithesis/workload/internal/client.go
```diff
@@ -311,34 +415,98 @@ func IsNotFound(err error) bool {
...
+//
+// Aborted is intentionally counted as "classified" so the interceptor does
+// not double-flag what IsAborted already surfaces; the chaos-distinct
+// signal still comes from the per-driver `if IsAborted(err) { Unreachable }`
+// or from the per-driver assert that does not list Aborted as expected.
+func IsClassified(err error) bool {
+	return err == nil ||
+		IsTransient(err) ||
+		IsCanceled(err) ||
+		IsAborted(err) ||
```
Azorlogh: The README claims
> `Aborted` is deliberately NOT in `IsTransient`: the gRPC spec allows
business "please retry" semantics there, and we have no audited guarantee
the server keeps Aborted as a pure transport transient. Letting the
classify interceptor surface it (as "Always(IsClassified)" Details
`code=Aborted`) is intentional — if it shows up under chaos, that is a
finding worth triaging.

But here IsClassified includes IsAborted, so Always(IsClassified) passes on Aborted and the classify interceptor never fires, contradicting the quote

Azorlogh (changes requested):

gfyrag:
Both NumaryBot findings addressed in `650c397ad`:

**🔴 blocker — interceptor chain overwrite** — fixed. `WithUnaryInterceptor` / `WithStreamInterceptor` were silently storing only the last call, so `classifyXxxInterceptor` was overwriting `retryXxxInterceptor`. Switched to `WithChainUnaryInterceptor(retry, classify)` and `WithChainStreamInterceptor(retry, classify)`. Order matters: retry runs first, classify wraps the final post-retry error. Same change applied to the retry-disabled branch for consistency.

**🟠 major — singleton timeout too short** — bumped from 30 min to **90 min**. Worst-case sum of internal per-step deadlines:
- `scaling_structured`: 4 cycles × (60s cooldown + 10 min `WaitForVoters` + 20s stable window) ≈ 45 min
- `rolling_restart` on a 7-node cluster: 7 × (60s pod-gone + 5 min pod-ready + 5 min voters) ≈ 77 min
- `quorum_recovery`: 8 min scale-down + 15 min scale-up ≈ 30 min

90 min is the fail-safe; the internal timeouts (`convergenceTimeout`, `rrVotersTimeout`, etc.) still do the actual hang detection.

Relaunching Antithesis (1h) against this commit.

cc previous (overwritten) run on the bugged binary: `e09646acb145f88f139ed1202ec19e13-56-17` — outputs from that one should be treated as untrusted (no retry under chaos = false-positive errors).

tests/antithesis/workload/internal/client.go
```diff
@@ -83,9 +84,22 @@ func NewGRPCConn() (*grpc.ClientConn, error) {
 	}
 
 	if !retryDisabled {
+		// MUST use the Chain* dial options: WithUnaryInterceptor (singular)
+		// stores ONE interceptor per dial and a second call overwrites the
+		// first silently. Order inside the chain matters: gRPC applies
+		// interceptors left-to-right (retry runs first, classify wraps the
+		// final post-retry error). The classify interceptor stays here even
+		// in retry-on mode because it asserts the workload's classification
+		// map is complete.
```
NumaryBot, line 93: 🟠 **[major] Classify interceptors overwrite retry interceptors instead of chaining them**

When retry is enabled, both `WithUnaryInterceptor`/`WithStreamInterceptor` calls for the retry logic and the subsequent calls for the classify interceptors are registered with gRPC dial options that store only a single interceptor of each kind. The later classify interceptors therefore silently replace the retry interceptors rather than wrapping them. As a result, under chaos conditions `DeadlineExceeded`, `READ_INDEX_NOT_CAUGHT_UP`, and `ExternalServiceError` are no longer retried, causing the workload to surface failures that the retry loop was supposed to absorb.

**Suggestion:** Use `grpc.WithChainUnaryInterceptor` and `grpc.WithChainStreamInterceptor` (or manually compose them with a chain helper) so both the retry and classify interceptors are active simultaneously.

tests/antithesis/workload/bin/cmds/main/parallel_driver_list_transactions/main.go
```diff
@@ -89,6 +89,10 @@ func main() {
 
 				if err != nil {
 					streamErr = true
+					if !internal.IsTransient(err) {
```
NumaryBot: 🟠 **[major] Canceled stream reads treated as unexpected SUT errors, producing false Antithesis findings**

When the driver context expires during `Recv()` on the transaction-listing stream, gRPC returns `codes.Canceled`. The new error-classification branch does not include `IsCanceled`/`IsTolerated`, so local driver timeouts or orderly shutdowns are reported as unexpected server errors and trigger an `Unreachable` assertion. The `IsTolerated` predicate was added precisely for this case but is not applied here.

**Suggestion:** Add a check for `IsCanceled(err)` (or the newly introduced `IsTolerated` helper) before the unexpected-error branch so that driver context cancellation causes a clean exit instead of a false finding.

tests/antithesis/workload/bin/cmds/main/parallel_driver_stale_reads/main.go
```diff
@@ -124,6 +124,11 @@ func main() {
 			if err != nil {
 				// Transient, or the serving node's prefix predates the ledger
 				// (a NotFound is itself a valid prefix) — skip this read.
+				if !internal.IsTransient(err) && !internal.IsNotFound(err) {
```
NumaryBot: 🟠 **[major] Canceled stale-read requests flagged as unexpected SUT errors**

If the driver context is canceled while a stale `GetAccount` call is in flight, the client returns `codes.Canceled`. The new error branch only allows `IsTransient` and `NotFound`, so a local driver shutdown is misclassified as an unexpected server error. This mirrors the false-positive risk described for the list-transactions and update-query drivers.

**Suggestion:** Guard the unexpected-error assertion with `IsCanceled(err)` / `IsTolerated(err)` and return (or continue) cleanly when the driver's own context was canceled.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1430#issuecomment-4833720199

tests/antithesis/workload/bin/cmds/main/parallel_driver_update_query/main.go
```diff
@@ -41,9 +41,16 @@ func main() {
 				},
 			},
 		}))
-		if err != nil && !internal.IsTransient(err) {
+		if err != nil {
```
NumaryBot, line 44: 🟠 **[major] Canceled query-setup calls produce false Unreachable assertions**

A driver context cancellation during prepared-query creation returns `codes.Canceled`, but the new error path only accepts `IsTransient` and `AlreadyExists`. Because `IsCanceled`/`IsTolerated` is not checked here, orderly driver shutdown is reported as an unexpected error and triggers an `Unreachable` finding.

**Suggestion:** Add a `IsCanceled(err)` / `IsTolerated(err)` guard before the unexpected-error assertion to allow this path to exit cleanly on local context cancellation.

tests/antithesis/workload/internal/client.go
```diff
@@ -311,34 +421,98 @@ func IsNotFound(err error) bool {
...
+//
+// Aborted is intentionally counted as "classified" so the interceptor does
+// not double-flag what IsAborted already surfaces; the chaos-distinct
+// signal still comes from the per-driver `if IsAborted(err) { Unreachable }`
+// or from the per-driver assert that does not list Aborted as expected.
+func IsClassified(err error) bool {
+	return err == nil ||
+		IsTransient(err) ||
+		IsCanceled(err) ||
+		IsAborted(err) ||
```
NumaryBot: 🟠 **[major] `IsClassified` includes `IsAborted`, making `Always(IsClassified)` suppress Aborted from the classify interceptor**

The README comment in the PR states that `Aborted` is deliberately excluded from `IsTransient` so that the classify interceptor can surface it as a finding under chaos. However, `IsClassified` includes `IsAborted`, so `Always(IsClassified)` evaluates to true for `Aborted` errors and the classify interceptor never fires for them — directly contradicting the stated design intent and hiding potentially meaningful chaos signals.

**Suggestion:** Either remove `IsAborted` from `IsClassified` (keeping it out of `IsTransient`) so the classify interceptor can surface `Aborted` errors, or update the README/comments to accurately reflect that `Aborted` is silently swallowed.

gfyrag:
Both findings addressed in `83b1b617d`:

**@Azorlogh** — you're right, the README and the code were directly contradicting each other. `IsAborted` is now removed from `IsClassified`, so any `Aborted` reaching a driver fires the classify interceptor's `Always(IsClassified)` with `code=Aborted` in the Details — exactly the loud signal the README promised. The "Aborted intentionally counted as classified to avoid double-flagging" comment was protecting a code path that didn't exist (no driver calls `IsAborted` explicitly to surface it elsewhere); ripped out.

Diff on `client.go`:

```diff
 func IsClassified(err error) bool {
     return err == nil ||
         IsTransient(err) ||
         IsCanceled(err) ||
-        IsAborted(err) ||
         IsBusinessError(err)
 }
```

**NumaryBot Canceled findings** (×3) — also legit, same root cause. When I migrated the `Sometimes` probes from `IsTransient` to `IsTolerated` (which includes `Canceled`), I left three classify-and-Unreachable sites still on the narrow `IsTransient`. A driver shutdown during ctx death would flip Unreachable instead of skipping silently. Replaced with `IsTolerated` on:

- `parallel_driver_list_transactions/main.go:92, 143` (mid-stream Recv on both lists)
- `parallel_driver_stale_reads/main.go:127`
- `parallel_driver_update_query/main.go:45`

README also updated to drop the `IsAborted` mention from the `IsClassified` line.

**Obsolete repost** — NumaryBot's "chain interceptors instead of overwriting retries" finding was already fixed in `650c397ad` (the previous force-push); this re-post predates that fix. Dismissable.

Relaunching Antithesis (1h) against `83b1b617d`.

tests/antithesis/workload/internal/client.go
NumaryBot, line 93: 🟠 **[major] Classify interceptors overwrite retry interceptors instead of chaining them**

When retry is enabled, both the retry and classify interceptors are registered via `WithUnaryInterceptor`/`WithStreamInterceptor`. In grpc-go these options store only a single interceptor each, so the later classify interceptor silently replaces the retry interceptor. Under chaos, `DeadlineExceeded`, `READ_INDEX_NOT_CAUGHT_UP`, and `ExternalServiceError` are no longer retried, causing the workload to surface failures that the retry loop was supposed to absorb.

**Suggestion:** Use `grpc.WithChainUnaryInterceptor` and `grpc.WithChainStreamInterceptor` so both the retry and classify interceptors are active simultaneously. This was reportedly fixed in commit 650c397ad; verify the fix is present and correct.

NumaryBot (commented):
NumaryBot posted 8 new inline findings.

Summary: https://github.com/formancehq/ledger/pull/1430#issuecomment-4833720199

tests/antithesis/workload/internal/client.go
```diff
@@ -311,34 +421,98 @@ func IsNotFound(err error) bool {
...
+// server path that produced it).
+//
+// Aborted is DELIBERATELY NOT classified: the gRPC spec allows business
+// "please retry" semantics there, and we have no audited guarantee the
+// server keeps Aborted as a pure transport transient. Surfacing it loud
+// via the classify interceptor is the entire point — see IsAborted.
+func IsClassified(err error) bool {
+	return err == nil ||
+		IsTransient(err) ||
+		IsCanceled(err) ||
```
NumaryBot, line 503: 🟠 **[major] `IsClassified` includes `IsAborted`, suppressing Aborted errors from the classify interceptor**
<sub>reported by NumaryBot, Azorlogh</sub>

The stated design intent (per PR comments) is that `Aborted` is excluded from `IsTransient` so the classify interceptor can surface it as a chaos finding. However, `IsClassified` includes `IsAborted`, so `Always(IsClassified)` evaluates to true for `Aborted` errors and the classify interceptor never fires for them — directly contradicting the documented intent and hiding potentially meaningful chaos signals.

**Suggestion:** Either remove `IsAborted` from `IsClassified` (keeping it out of `IsTransient`) so the classify interceptor fires on `Aborted` errors, or update comments/README to accurately reflect that `Aborted` is intentionally swallowed.

tests/antithesis/workload/bin/cmds/main/parallel_driver_list_transactions/main.go
```diff
@@ -89,6 +89,10 @@ func main() {
 
 				if err != nil {
 					streamErr = true
+					if !internal.IsTolerated(err) {
```
NumaryBot, line 92: 🟠 **[major] Canceled stream reads produce false `Unreachable` Antithesis assertions**

When the driver context expires during `Recv()` on the transaction-listing stream, gRPC returns `codes.Canceled`. The error-classification branch does not include a `IsCanceled`/`IsTolerated` check, so local driver timeouts or orderly shutdowns are reported as unexpected server errors and trigger an `Unreachable` assertion.

**Suggestion:** Add a check for `IsCanceled(err)` (or `IsTolerated(err)`) before the unexpected-error branch so that driver context cancellation results in a clean exit rather than a false finding.

tests/antithesis/workload/bin/cmds/main/parallel_driver_stale_reads/main.go
```diff
@@ -122,8 +122,14 @@ func main() {
 				Address: probeAccount,
 			})
 			if err != nil {
-				// Transient, or the serving node's prefix predates the ledger
-				// (a NotFound is itself a valid prefix) — skip this read.
+				// Transient, ctx canceled (driver shutting down), or the
+				// serving node's prefix predates the ledger (NotFound is
+				// itself a valid prefix) — skip this read.
```
NumaryBot, line 127: 🟠 **[major] Canceled stale-read requests flagged as unexpected SUT errors**

If the driver context is canceled while a stale `GetAccount` call is in flight, the client returns `codes.Canceled`. The error branch only allows `IsTransient` and `NotFound`, so a local driver shutdown is misclassified as an unexpected server error, producing false Antithesis findings.

**Suggestion:** Guard the unexpected-error assertion with `IsCanceled(err)` / `IsTolerated(err)` and return (or continue) cleanly when the driver's own context is canceled.

tests/antithesis/workload/bin/cmds/main/parallel_driver_update_query/main.go
NumaryBot, line 44: 🟠 **[major] Canceled query-setup calls produce false `Unreachable` assertions**

A driver context cancellation during prepared-query creation returns `codes.Canceled`, but the new error path only accepts `IsTransient` and `AlreadyExists`. Because `IsCanceled`/`IsTolerated` is not checked, orderly driver shutdown is reported as an unexpected error and triggers a false `Unreachable` finding.

**Suggestion:** Add a `IsCanceled(err)` / `IsTolerated(err)` guard before the unexpected-error assertion to allow a clean exit on local context cancellation.

tests/antithesis/workload/bin/cmds/main/parallel_driver_check_store/main.go
```diff
@@ -41,7 +41,7 @@ func main() {
 			break
 		}
 		if err != nil {
-			if internal.IsUnavailable(err) {
+			if internal.IsTransient(err) {
```
NumaryBot, line 44: 🟠 **[major] Cancellation from `CheckStore` driver context not tolerated, producing false findings**

With the driver now using `DriverContext()`, a timeout or orderly shutdown during `CheckStore` stream creation or `Recv()` returns `codes.Canceled`. The error branch only allows `IsTransient`, so a local driver cancellation is reported as `Unreachable("CheckStore stream error")` instead of exiting cleanly.

**Suggestion:** Use `IsTolerated(err)` (or `IsCanceled(err)`) in these branches to exit cleanly on local driver context cancellation, consistent with other cancellation-safe probes.

tests/antithesis/workload/bin/cmds/main/parallel_driver_query_checkpoints/main.go
```diff
@@ -12,7 +11,8 @@ import (
 func main() {
 	log.Println("composer: parallel_driver_query_checkpoints")
 
-	ctx := context.Background()
+	ctx, cancel := internal.DriverContext()
```
NumaryBot, line 14: 🟠 **[major] Checkpoint RPCs not guarded against local driver cancellation**

Adding `DriverContext()` makes every RPC in this file able to fail with `codes.Canceled` when the driver deadline/shutdown fires, but existing error branches still only accept `IsTransient` before raising `Unreachable`. Under normal cancellation this produces false Antithesis findings for Create/List/Get/Delete checkpoint calls.

**Suggestion:** Add `IsCanceled`/`IsTolerated` guards before the unexpected-error assertions in all checkpoint RPC error branches.

tests/antithesis/workload/bin/cmds/main/parallel_driver_transfer_leadership/main.go
```diff
@@ -13,7 +12,8 @@ import (
 func main() {
 	log.Println("composer: parallel_driver_transfer_leadership")
 
-	ctx := context.Background()
+	ctx, cancel := internal.DriverContext()
```
NumaryBot, line 15: 🟠 **[major] Leader-state lookup not guarded against driver context cancellation**

The new bounded driver context can make the initial `GetClusterState` call return `codes.Canceled` when the driver is shutting down, but the error branch only accepts `IsTransient` and otherwise raises `Unreachable`. This reintroduces the same false-finding path the new `IsTolerated` predicate is meant to avoid.

**Suggestion:** Handle `IsCanceled`/`IsTolerated` before asserting an unexpected cluster-state error so orderly shutdown does not produce a false `Unreachable` finding.

gfyrag:
Run #4 (`9d085e6a…-56-17`) report: no classify-interceptor violation (0/9 hits clean), but **3× `Sometimes(IsTolerated)` on `CreateLedger` failed**, plus **2× `AlwaysOrUnreachable` violations on `parallel_driver_typed_metadata`**, all involving `lrecreate-XXX` ledgers:

```
should be able to create ledger
  error: rpc error: code = FailedPrecondition desc = ledger has been deleted: lrecreate-{733992, 495143, 87540}

changing metadata type should not crash
  ledger: lrecreate-87540
  error:  FailedPrecondition: ledger has been deleted

account should be readable after metadata type changes
  ledger: lrecreate-87540
  error:  NotFound: ledger lrecreate-87540 not found
```

## Root cause

Pre-existing bug, **not** introduced by this PR. `restrictedPrefixes` (`tests/antithesis/workload/internal/ledger.go`) contained:

```go
"lrec-",     // parallel_driver_ledger_recreate (covers lrecreate- too)
```

The comment is wrong: `strings.HasPrefix("lrecreate-87540", "lrec-")` is **false** — char index 5 of `lrecreate-…` is `r`, not `-`. So every `lrecreate-XXX` ledger created (and later deleted) by `parallel_driver_ledger_recreate` ended up in `GetRandomLedger`'s pool, available to any other driver. `typed_metadata` drew one mid-recreation cycle and got the deleted-ledger errors above.

The `should be able to create ledger` flips are the inverse: `ledger_recreate` re-creates an `lrecreate-N` that was just deleted, and the server (correctly) refuses with `FailedPrecondition: ledger has been deleted`. The `Sometimes` is meant to fire on the happy path of fresh creation, which still happens on every other ledger — but for these three specific ledger IDs, every create attempt in the test window failed.

## Fix

Replaced `"lrec-"` with `"lrecreate-"`. Sanity check:

```
lrec-      matches "lrecreate-87540" → False   (the bug)
lrecreate- matches "lrecreate-87540" → True    (now)
```

Added a comment block on the entry referencing run #4 so future audits don't repeat the truncation mistake.

The `1× ledger resolved while bloom filters not ready` Sometimes (`internal/infra/plan/builder.go:178`) is a server-side coverage probe with `details: null` — a `Sometimes(condition=false)` here just means that code path wasn't exercised this run, not a bug.

Pushed as `83b1b617d → <new tip>`. Relaunching Antithesis #5.

tests/antithesis/workload/internal/client.go
NumaryBot, line 93: 🟠 **[major] Classify interceptors overwrite retry interceptors instead of chaining them**

When retry is enabled, both the retry and classify interceptors are registered via WithUnaryInterceptor/WithStreamInterceptor. In grpc-go these options store only a single interceptor each, so the later classify interceptor silently replaces the retry interceptor. Under chaos, DeadlineExceeded, READ_INDEX_NOT_CAUGHT_UP, and ExternalServiceError are no longer retried, causing the workload to surface failures that the retry loop was supposed to absorb. Per the prior discussion, a fix was applied in commit 650c397ad switching to WithChainUnaryInterceptor/WithChainStreamInterceptor — this should be verified as present and complete in the final state of the PR.

**Suggestion:** Use grpc.WithChainUnaryInterceptor and grpc.WithChainStreamInterceptor so both the retry and classify interceptors are active simultaneously. Verify commit 650c397ad is reflected in the current branch.

tests/antithesis/workload/internal/client.go
NumaryBot, line 503: 🟠 **[major] IsClassified includes IsAborted, suppressing Aborted errors from the classify interceptor**
<sub>reported by NumaryBot, Azorlogh</sub>

The stated design intent (per PR README/comments) is that Aborted is excluded from IsTransient so the classify interceptor can surface it as a chaos finding. However, IsClassified includes IsAborted, so Always(IsClassified) evaluates to true for Aborted errors and the classify interceptor never fires for them — directly contradicting the documented intent and hiding potentially meaningful chaos signals. Per the prior discussion, a fix was applied in commit 83b1b617d removing IsAborted from IsClassified — this should be confirmed present in the final state of the PR.

**Suggestion:** Remove IsAborted from IsClassified so that Aborted errors pass through to the classify interceptor, consistent with the documented design. Verify the fix from 83b1b617d is reflected in the current branch.

tests/antithesis/workload/bin/cmds/main/parallel_driver_list_transactions/main.go
NumaryBot, line 92: 🟠 **[major] Canceled stream reads produce false Unreachable Antithesis assertions**

When the driver context expires during Recv() on the transaction-listing stream, gRPC returns codes.Canceled. The error-classification branch does not include an IsCanceled/IsTolerated check, so local driver timeouts or orderly shutdowns are reported as unexpected server errors and trigger an Unreachable assertion. Per the prior discussion this was addressed in 83b1b617d — verify it is complete.

**Suggestion:** Add a check for IsCanceled(err) or IsTolerated(err) before the unexpected-error branch so that driver context cancellation results in a clean exit rather than a false finding.

tests/antithesis/workload/bin/cmds/main/parallel_driver_stale_reads/main.go
NumaryBot, line 127: 🟠 **[major] Canceled stale-read requests flagged as unexpected SUT errors**

If the driver context is canceled while a stale GetAccount call is in flight, the client returns codes.Canceled. The error branch only allows IsTransient and NotFound, so a local driver shutdown is misclassified as an unexpected server error, producing false Antithesis findings. Per the prior discussion this was addressed in 83b1b617d — verify it is complete.

**Suggestion:** Guard the unexpected-error assertion with IsCanceled(err) / IsTolerated(err) and return or continue cleanly when the driver's own context is canceled.

tests/antithesis/workload/bin/cmds/main/parallel_driver_update_query/main.go
NumaryBot, line 44: 🟠 **[major] Canceled query-setup calls produce false Unreachable assertions**

A driver context cancellation during prepared-query creation returns codes.Canceled, but the new error path only accepts IsTransient and AlreadyExists. Because IsCanceled/IsTolerated is not checked, orderly driver shutdown is reported as an unexpected error and triggers a false Unreachable finding. Per the prior discussion this was addressed in 83b1b617d — verify it is complete.

**Suggestion:** Add an IsCanceled(err) / IsTolerated(err) guard before the unexpected-error assertion to allow a clean exit on local context cancellation.

tests/antithesis/workload/bin/cmds/main/parallel_driver_check_store/main.go
```diff
@@ -23,8 +23,8 @@ func main() {
 
 	stream, err := client.CheckStore(ctx, &servicepb.CheckStoreRequest{})
 	if err != nil {
-		if internal.IsUnavailable(err) {
-			log.Printf("CheckStore unavailable: %s", err)
+		if internal.IsTransient(err) {
```
NumaryBot, line 26: 🟠 **[major] Cancellation from CheckStore driver context not tolerated, producing false findings**

With the driver now using DriverContext(), a timeout or orderly shutdown during CheckStore stream creation or Recv() returns codes.Canceled. The error branch only allows IsTransient, so a local driver cancellation is reported as Unreachable('CheckStore stream error') instead of exiting cleanly.

**Suggestion:** Use IsTolerated(err) or IsCanceled(err) in these branches to exit cleanly on local driver context cancellation, consistent with other cancellation-safe probes.

tests/antithesis/workload/bin/cmds/main/parallel_driver_query_checkpoints/main.go
NumaryBot, line 14: 🟠 **[major] Checkpoint RPCs not guarded against local driver cancellation**

After switching this standalone driver to DriverContext(), any checkpoint RPC (Create/List/Get/Delete) can return codes.Canceled when the 10-minute driver deadline or shutdown fires. The error branches still only allow IsTransient before raising Unreachable, so orderly local cancellation produces false Antithesis findings.

**Suggestion:** Add IsCanceled/IsTolerated guards before the unexpected-error assertions in all checkpoint RPC error branches.

tests/antithesis/workload/bin/cmds/main/parallel_driver_transfer_leadership/main.go
NumaryBot, line 15: 🟠 **[major] Leader-state lookup not guarded against driver context cancellation**

The new bounded driver context can make the initial GetClusterState call return codes.Canceled when the driver is shutting down, but the error branch only accepts IsTransient and otherwise raises Unreachable. This reintroduces the same false-finding path the new IsTolerated predicate is meant to avoid.

**Suggestion:** Handle IsCanceled/IsTolerated before asserting an unexpected cluster-state error so orderly shutdown does not produce a false Unreachable finding.

NumaryBot (commented):
NumaryBot posted 8 new inline findings.

Summary: https://github.com/formancehq/ledger/pull/1430#issuecomment-4833720199

gfyrag:
Pushed `e09ba3313` with three deeper changes beyond the `lrec-` typo fix — each closes a class of bug rather than just the symptom.

## 1. Typed `OwnedLedgerPrefix` + cohesion test

Root cause of run-#4: \`restrictedPrefixes\` was \`[]string\` with the trailing hyphen baked into the literals. The \`lrec-\` entry was off by one (5 chars instead of 10) and never matched \`lrecreate-…\` because char index 5 is \`r\`, not \`-\`. Same class of bug could recur on any new prefix.

Refactor:
- New type \`internal.OwnedLedgerPrefix string\` + 17 constants, the token only (no trailing hyphen).
- \`RestrictedPrefixes()\` derives the hyphen-terminated forms once, so a typo cannot desync.
- Methods: \`.New()\` (random 6-digit suffix), \`.WithSeed(uint64)\` (when several names share a seed), \`.WithSuffix(string)\` (fixed sentinel-XXX names).
- \`internal/ledger_test.go\`:
  - \`TestOwnedLedgerPrefixes_NoOverlap\` — no prefix is a prefix of another (catches \`lrec\` vs \`lrecreate\` style typos at \`go test\` time).
  - \`TestRestrictedPrefixes_HaveTrailingHyphen\` — every derived prefix ends with a single \`-\` (catches the inverse mistake).
  - \`TestOwnedLedgerPrefix_New\` / \`_WithSuffix\` — shape contracts.

Migration: 15 manual \`fmt.Sprintf("prefix-%d", run%1_000_000)\` sites and 4 \`sentinelLedger = "sentinel-XXX"\` constants moved over. Also fixed three drivers that were creating owned ledgers but had **no entry in the restriction list**:

- \`PrefixMaintenance\` (\`maint-\`) — was missing
- \`PrefixAccountTypes\` (\`accttype-\`) — was missing
- \`PrefixTypeViolation\` (\`typeviolation-\`) — was missing

Pre-existing pool-pollution risk: a generic driver could draw one of these ledgers and see e.g. an unexpected \`enforcementMode = AUDIT\` set by the owning driver.

## 2. \`AlwaysOrUnreachable(IsTolerated(err), "should not crash")\` is mis-formed

\`IsTolerated\` excludes business errors (NotFound, AlreadyExists, generic FailedPrecondition). Using it under an assertion titled "should not crash" was strictly **wrong**: a legitimate business error (e.g. \`NotFound\` on a ledger another driver just deleted) flipped the assertion to false even though the server did exactly what it should.

Migrated 5 sites \`IsTolerated → IsClassified\`:

- \`parallel_driver_typed_metadata:175, 203, 250, 269\`
- \`parallel_driver_compaction:115\`

\`IsClassified\` is the correct expression for "should not crash": any classified error (transient, canceled, business) is fine; only an orphan code (Internal panic, ResourceExhausted, …) is a real crash, and that's surfaced by the classify interceptor with a globally-unique assertion name.

## 3. \`IsBusinessError\` unexported

Audit confirmed \`IsBusinessError\` had no caller outside \`IsClassified\`. Drivers asserting on a specific business outcome (\`insufficient_funds\` etc.) check the precise reason via \`HasErrorReason(err, …)\`, never the coarse "is it business" predicate. Renamed to \`isBusinessError\` (package-private) and updated the README to point future drivers at \`HasErrorReason\` for outcome-specific checks. No driver code changes.

## State

- Build clean (workload + main module)
- \`go test ./internal/...\` passes (new \`ledger_test.go\` included)
- Lint: 306 (unchanged from previous push — no new issues introduced)

Relaunching Antithesis #6 against \`e09ba3313\` — should green now that the \`lrecreate-\` pool leak is fixed AND the mis-formed \`IsTolerated\` assertions on \`typed_metadata\`/\`compaction\` are properly \`IsClassified\`.

tests/antithesis/workload/bin/cmds/main/parallel_driver_compaction/main.go
```diff
@@ -112,7 +112,7 @@ func verifyReadable(ctx context.Context, client servicepb.BucketServiceClient, l
 		Address: "world",
 	})
 
-	assert.AlwaysOrUnreachable(err == nil || internal.IsTransient(err),
+	assert.AlwaysOrUnreachable(internal.IsClassified(err),
```
NumaryBot, line 115: 🟠 **[major] Post-compaction read: business errors accepted as success by IsClassified**

When GetAccount returns a classified business error (e.g. NotFound) after compaction, IsClassified(err) evaluates to true and the assertion passes. This means the compaction invariant — that the account remains readable — is never actually enforced; real read-after-compaction regressions would be silently swallowed.

**Suggestion:** Replace the IsClassified guard with a check that only tolerates transient or canceled errors (IsTransient or IsTolerated). A business error like NotFound after compaction should fail the assertion, not pass it.

tests/antithesis/workload/bin/cmds/main/parallel_driver_typed_metadata/main.go
```diff
@@ -200,7 +200,7 @@ func main() {
 		if err != nil {
 			internal.LogCleanupError("read account after type changes", err)
 		}
-		assert.AlwaysOrUnreachable(err == nil || internal.IsTransient(err),
+		assert.AlwaysOrUnreachable(internal.IsClassified(err),
```
NumaryBot, line 203: 🟠 **[major] Post-metadata-change read: business errors accepted as readable state by IsClassified**

If the account read following a metadata type change returns a classified business error (e.g. NotFound), IsClassified(err) is true, the assertion passes, and the metadata-survival check below is skipped. Type-migration regressions that cause accounts to disappear would not be detected.

**Suggestion:** Use IsTransient or IsTolerated (not IsClassified) as the guard for this postcondition so that only genuine transient or shutdown-related errors bypass the check; any business error on this path should trigger the assertion.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1430#issuecomment-4833720199

Azorlogh (approved):

PR #1432, opened 2026-06-29
fix(EN-1410): close bloom boot-ordering window

gfyrag:
## Summary

`applier.recoverState` launches the bloom rebuild in a background goroutine and immediately drives `replayWAL` without waiting on `IsReady`. While the rebuild is still in flight, the FSM applies entries and may trigger cache rotations. The rotation path in `machine.go` wipes the outgoing generation from `0xFF` unconditionally but skips `PersistDirtyBlocks` when `!IsReady` — so on the restart path, where the bloom is reconstructed from `[persisted blocks] + [cache mem replay]`, a key dropped from cache by two rotations while `!IsReady` becomes invisible to both restoration sources. The next crash drops it for real; the subsequent restart produces a bloom missing real Pebble keys; the resolver `injects` a zero `VolumePair` (`includeZeroValue=true` for Volume); the FSM apply path returns `available 0`.

Observed on `blockchains` (`eks-acme-dev-euw1-01`), confirmed via `ledgerctl accounts aggregate-volumes` (read path returns the correct non-zero balance while connect sees `available=0` on the same key).

## Fix

**Synchronous bloom rebuild on the restart path.** When `hasPersistedBloomBlocks(store) == true`, `restoreBloomFilters` runs `RestoreFromStore + replayBloomFromCache + SetReady` inline before `RestoreFromStore` returns. Cost is bounded by persisted blocks + cache size — sub-second. The caller (`applier.recoverState`) cannot start `replayWAL` while the bloom is still populating, so the asymmetric rotation path (unconditional `0xFF` wipe vs `IsReady`-gated `PersistDirtyBlocks`) no longer has a chance to drop keys.

The cold-start / Rebuild path keeps the async populate intentionally (the full Pebble scan can take minutes on a large database). It is safe without a rotation gate — see below.

## Cold-start path safety

The cold-start / Rebuild branch is self-healing for three combined reasons:

1. **Bloom bits are OR-monotone.** `Filter.Add` (`bloom.go:106-109`), `RestoreFromStore.OrBlock` (`bloom.go:204`), `PopulateFromStore.Add` (`bloom.go:566`) only set bits via `atomic.OrUint64`. No path clears a bit individually.
2. **FSM apply adds bloom bits in the same operation as the Pebble `0x01` write.** `machine.go:1556-1558`: `BloomFilters.AddCanonicalKeys(...)` is called unconditionally (not gated on `IsReady`) before `batch.Commit`. Every key written to `0x01` has its bit set in the in-memory bloom in the same atomic FSM step.
3. **Pebble `0x01` is invariant to cache rotations.** `writeCacheRotation` (`cache_incremental.go:158-194`) only touches `ZoneCache` (`0xFF` mirror + meta). `ZoneAttributes` (`0x01`) is never wiped.

During a cold-start populate window, every key written by the FSM gets its bit set in-memory regardless of rotations. If a crash happens before any `PersistDirtyBlocks` runs (always the case while `!IsReady`), `hasPersistedBloomBlocks == false` on the next boot, the cold-start branch re-enters `PopulateFromStore`, and `0x01` — the authoritative source — rebuilds the bloom from scratch.

The restart branch is different: it reconstructs the bloom from `[persisted blocks] + [cache gen0/gen1]`, not from `0x01`. A key dropped from cache by two rotations and not in the persisted blocks (because `PersistDirtyBlocks` was skipped) becomes invisible to both restoration sources. That asymmetry is what the synchronous fix closes — and only the restart branch needs it.

Jira: [EN-1410](https://formance-team.atlassian.net/browse/EN-1410).

## Test plan

- [x] **Red regression test landed first** (`cache_snapshotter_bloom_boot_ordering_test.go`, prior commit). Drives the exact sequence — pre-state with a persisted bloom block holding `K_old`, simulated mid-rebuild apply of `K1, K2` + rotation, then `K3, K4` + rotation, then fresh restart — and asserts `MayContain` is true for all five keys. The two assertions on `K1, K2` fail against `main`; this commit turns them green.
- [x] `go test ./internal/infra/state/... ./internal/infra/bloom/... ./internal/infra/cache/... ./internal/storage/dal/...` — all green, no regressions.
- [x] `just pre-commit` clean.
- [ ] **Live confirmation on blockchains (post-merge follow-up).** Reuse the `pr-580-*` deploy pattern: rebuild the cluster image with this fix, reset PVCs, ingest bitcoin-mainnet, and watch for `available=0` over several OOM cycles. With the fix, the bug should not recur.

## Out of scope

- The `applier.recoverState` flow itself isn't restructured: the sync vs async dispatch happens inside `CacheSnapshotter.RestoreFromStore`, transparent to the caller.
- Bloom call sites continue to treat `MayContain == false` as authoritative absence (correct by design when the bloom is complete; the bug was bloom completeness, not the consumers).

[EN-1410]: https://formance-team.atlassian.net/browse/EN-1410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
_Migrated from formancehq/ledger-v3-poc#586 on 2026-06-29._

### DIFF
--- a/internal/infra/state/cache_snapshotter.go
+++ b/internal/infra/state/cache_snapshotter.go
@@ -531,8 +531,14 @@ func (s *CacheSnapshotter) RestoreFromStore(store dal.RecoveryReader) error {
 	// Restore bloom filters: load persisted blocks from Pebble, then replay
 	// cache gen0+gen1 entries to fill the gap since the last rotation flush.
 	// If no persisted blocks exist (first boot), fall back to a full attribute scan.
+	// On the persisted-blocks (restart) path, restoreBloomFilters runs the
+	// rebuild synchronously and returns once IsReady() is true -- closes
+	// the EN-1410 window where replayWAL would race the rebuild and drive
+	// rotations while !IsReady.
 	if s.bloomFilters != nil {
-		s.restoreBloomFilters(store)
+		if err := s.restoreBloomFilters(store); err != nil {
+			return fmt.Errorf("restoring bloom filters: %w", err)
+		}
 	}
 
 	return nil
@@ -623,20 +629,26 @@ func (s *CacheSnapshotter) restoreGeneration(reader dal.PebbleReader, genByte by
 	return nil
 }
 
-// restoreBloomFilters launches a background task to restore bloom filters.
-// If persisted blocks exist, loads them from Pebble and replays cache gen0+gen1
-// to fill the gap. Otherwise, falls back to a full attribute scan.
-// In both cases, IsReady() remains false until the background work completes.
-// The store is captured by the background goroutine so it can open a fresh
-// read handle each invocation.
-func (s *CacheSnapshotter) restoreBloomFilters(store dal.RecoveryReader) {
+// restoreBloomFilters publishes a ready bloom on top of Pebble's
+// persisted state. On a simple restart (persisted bloom blocks exist),
+// the rebuild runs SYNCHRONOUSLY before this function returns: the cost
+// is bounded (O(blocks) + O(cache gen0+gen1)) and running it inline
+// closes the EN-1410 window where the replayWAL goroutine would
+// otherwise race the background rebuild and drive cache rotations
+// while !IsReady. On cold start / post-Rebuild (no persisted blocks),
+// the rebuild stays async because the full attribute scan via
+// PopulateFromStore can take minutes on a large database -- blocking
+// boot is unacceptable. The cold-start path's correctness is guarded
+// separately at the cache-rotation site, which inhibits rotation while
+// the bloom is still populating (see machine.go).
+func (s *CacheSnapshotter) restoreBloomFilters(store dal.RecoveryReader) error {
 	if s.hasPersistedBloomBlocks(store) {
-		s.runBloomTask(store, "restore from Pebble blocks", s.bloomFilters.RestoreFromStore)
-
-		return
+		return s.runBloomTaskSync(store, "restore from Pebble blocks", s.bloomFilters.RestoreFromStore)
 	}
 
 	s.StartAsyncBloomPopulate(store, "first boot: no persisted bloom blocks")
+
+	return nil
 }
 
 // StartAsyncBloomPopulate interrupts any running bloom task and launches a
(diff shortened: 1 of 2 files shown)


internal/infra/state/cache_snapshotter.go
```diff
@@ -623,20 +629,26 @@ func (s *CacheSnapshotter) restoreGeneration(reader dal.PebbleReader, genByte by
...
+// restoreBloomFilters publishes a ready bloom on top of Pebble's
+// persisted state. On a simple restart (persisted bloom blocks exist),
+// the rebuild runs SYNCHRONOUSLY before this function returns: the cost
+// is bounded (O(blocks) + O(cache gen0+gen1)) and running it inline
+// closes the EN-1410 window where the replayWAL goroutine would
+// otherwise race the background rebuild and drive cache rotations
+// while !IsReady. On cold start / post-Rebuild (no persisted blocks),
+// the rebuild stays async because the full attribute scan via
+// PopulateFromStore can take minutes on a large database -- blocking
+// boot is unacceptable. The cold-start path's correctness is guarded
```
NumaryBot, line 641: 🟠 **[major] Defer rotations while async bloom populate is running**

When this branch is taken (no persisted bloom blocks, including after a bloom Rebuild purges them), `RestoreFromStore` still returns with `IsReady()==false`, but the apply loop still performs `CheckRotationNeeded`/`writeCacheRotation` while only skipping `PersistDirtyBlocks` when the bloom is not ready. If replayWAL or normal apply crosses a cache rotation before the async scan/replay has covered recently written keys, those keys can be removed from 0xFF and then missed when `SetReady` publishes the bloom, allowing `MayContain` to return false for real Pebble 0x01 rows. The cold-start branch needs the same rotation gate (or a synchronous populate before replay) for the stated invariant to hold.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1432#issuecomment-4833726074

paul-nicolas (approved):

internal/infra/state/cache_snapshotter.go
gfyrag: The cold-start branch is self-healing for three combined reasons that this analysis missed:

1. Bloom bits are OR-monotone (`Filter.Add` and `OrBlock` only set, never clear).
2. The FSM apply path adds bits to the in-memory bloom unconditionally on every write, in the same operation as the Pebble 0x01 write (`machine.go:1556-1558`, not gated on `IsReady`).
3. Pebble 0x01 is invariant to cache rotations — `writeCacheRotation` only touches `ZoneCache`.

So during the cold-start populate window: every key written by FSM apply gets its bit set in-memory; rotations don't lose bits; if a crash happens before `PersistDirtyBlocks` ever runs (always the case while `!IsReady`), `hasPersistedBloomBlocks==false` on the next boot, and `PopulateFromStore` re-scans 0x01 in full — rebuilding the bloom from the authoritative source.

The restart branch was broken because its recovery reads from persisted blocks + cache mem (not 0x01), so a key dropped from cache by two rotations during `!IsReady` becomes invisible to both restoration sources. That asymmetry is what the synchronous fix closes. The cold-start branch doesn't share it.

Updating the PR description to drop the second-fix claim and add the cold-start safety rationale inline.

PR #1433, opened 2026-06-29
fix(EN-1409): scan checkpoints dir on cleanup to reclaim crash orphans

gfyrag:
## Summary

`cleanupOldCheckpoints` iterated from an in-memory tracker
(`s.oldestCheckpoint`) initialised at boot by arithmetic
(`latestCheckpointID - maxCheckpoints + 1`). After any non-graceful exit,
the previous live checkpoint set was preserved on disk and fell inside the
newly-computed window, so cleanup never touched it. As subsequent
snapshots advanced the tracker, those fossils dropped out of the window
forever — the loop only iterates upward from the tracker, never below.

Net effect: every crash permanently leaked
`maxCheckpoints * checkpoint_size` of disk.

Observed on `blockchains` LedgerService on `eks-acme-dev-euw1-01`:
93.4 GiB of orphan checkpoint dirs accumulated in ~14h, saturating the
data PVC and triggering `ResourceExhausted` ("writes blocked: disk usage
exceeds threshold") in connect's batch applies. The disk layout shows
the fossil pairs explicitly — six `(X, X+1)` pairs separated by gaps
where normal rotation did clean up:

```
checkpoints/23   7.7G  + 24  1.9G    (fossil)
checkpoints/161  12.1G + 162 241M    (fossil)
checkpoints/194  12.7G + 195 481M    (fossil)
checkpoints/224  13.8G + 225 480M    (fossil)
checkpoints/256  16.3G + 257 481M    (fossil)
checkpoints/309  17.2G + 310 246M    (fossil)
checkpoints/316  20.4G + 317 1.6G    (live)
```

## Fix

Replace the incremental loop with a directory scan: read `checkpointsDir`,
parse every numeric child as a checkpoint ID, and `RemoveAll` anything
below `oldestToKeep`. Non-numeric children are skipped (defensive — see
follow-up commit).

`s.oldestCheckpoint` is kept as an informational floor but is no longer
authoritative for what gets deleted.

## Follow-up commit: layout split + loud-fail on missing parent

The first commit fixed the leak but left two rough edges. The follow-up
commit addresses both:

1. **Loud-fail invariant on missing `checkpoints/`** (addresses
   paul-nicolas's Medium review): `cleanupOldCheckpoints` runs only
   after `db.Checkpoint()` has succeeded, which (re)creates the
   `checkpoints/` parent. `ENOENT` from `os.ReadDir` here is impossible
   by design — per CLAUDE.md invariant #7 we surface it instead of
   silently no-oping.

2. **Move follower-sync staging out of `checkpoints/`**: was
   `dataDir/checkpoints/incoming/`, becomes `dataDir/incoming-checkpoint/`
   (top-level sibling). Reasons:
   - `checkpointsDir` now holds only numbered checkpoint dirs; the
     "skip non-numeric" branch becomes pure defense for foreign
     artifacts (.DS_Store etc) rather than load-bearing logic.
   - Same filesystem → `ActivateIncomingRestore`'s
     `Rename(staging, checkpoints/<id>)` stays atomic.
   - `ActivateIncomingRestore` now `MkdirAll`s `checkpoints/` explicitly
     (before the split it was a side effect of `PrepareIncomingRestore`).
   - `cleanupIncomingRestore` at boot wipes both the new staging path
     and the legacy `checkpoints/incoming/` path. One-way migration is
     safe — the staging dir is always volatile.

## Test plan

- [x] `go test ./internal/storage/dal/ -run TestStore_CleanupOldCheckpoints`
- [x] `TestStore_CleanupOldCheckpoints_RemovesOrphansBelowTracker`
      pre-populates `{5, 6, 100, 101}`, runs `NewStore` + one
      `CreateSnapshot`, and asserts `{5, 6, 100}` are gone and
      `{101, 102}` remain. Before the fix, the assertion on the missing
      fossils fails because the tracker-bounded loop never visits IDs
      below the boot-computed floor.
- [x] `TestStore_LegacyIncomingMigration` pins the boot-time cleanup
      of `dataDir/checkpoints/incoming/`.
- [x] `TestStore_PrepareIncomingRestoreLocation` pins the new staging
      path so a future move is caught by tests, not by surprise on
      production restores.
- [x] `go test ./internal/storage/dal/ ./internal/infra/state/ ./internal/infra/node/`
      green — follower-sync flow (PrepareIncomingRestore → write →
      ActivateIncomingRestore → RestoreCheckpoint) still works with
      the new staging location.
- [ ] Validate live on `blockchains` after merge + image deploy: orphan
      pairs should be reclaimed on the next snapshot tick after restart.

Jira: [EN-1409](https://formance-team.atlassian.net/browse/EN-1409).

[EN-1409]: https://formance-team.atlassian.net/browse/EN-1409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
_Migrated from formancehq/ledger-v3-poc#585 on 2026-06-29._

### DIFF
--- a/internal/storage/dal/store.go
+++ b/internal/storage/dal/store.go
@@ -50,7 +50,11 @@ const (
 	checkpointsDir          = "checkpoints"
 	temporaryCheckpointsDir = "tmp"
 	baselineCheckpointsDir  = "baseline"
-	incomingRestoreDir      = "incoming" // inside checkpointsDir, non-numeric to avoid ID collision
+	// incomingCheckpointDir sits at the dataDir top level (sibling of
+	// checkpointsDir) so that checkpointsDir contains only numbered
+	// checkpoint dirs. Same filesystem → ActivateIncomingRestore's
+	// Rename(incoming, checkpoints/<id>) stays atomic.
+	incomingCheckpointDir = "incoming-checkpoint"
 )
 
 // ScanLatestCheckpointID scans the checkpoints directory and returns the highest
@@ -74,7 +78,10 @@ func ScanLatestCheckpointID(dataDir string) (latestID uint64, found bool, err er
 
 		id, err := strconv.ParseUint(entry.Name(), 10, 64)
 		if err != nil {
-			continue // skip non-numeric dirs (e.g. "incoming")
+			// Defensive: no producer writes non-numeric children to
+			// checkpointsDir, but tolerate foreign artifacts (.DS_Store,
+			// editor scratch files) instead of failing boot.
+			continue
 		}
 
 		if !found || id > latestID {
@@ -728,8 +735,9 @@ func (s *Store) CreateSnapshot() (uint64, error) {
 // exit otherwise preserves the live checkpoint set as untouchable orphans,
 // permanently leaking `maxCheckpoints * checkpoint_size` per crash.
 //
-// Non-numeric child dirs (e.g. "incoming" used by follower sync) are
-// skipped; only positive-integer dirs are considered checkpoints.
+// Non-numeric children are skipped defensively (no producer writes them
+// after the layout split that moved follower-sync staging out of
+// checkpointsDir); only numeric dirs are considered checkpoints.
 func (s *Store) cleanupOldCheckpoints() error {
 	newCheckpoint := s.currentCheckPoint + 1
 
@@ -742,12 +750,12 @@ func (s *Store) cleanupOldCheckpoints() error {
 	oldestToKeep := newCheckpoint - uint64(s.maxCheckpoints) + 1
 	checkpointsPath := filepath.Join(s.dataDir, checkpointsDir)
 
+	// cleanupOldCheckpoints runs only after CreateSnapshot's db.Checkpoint()
+	// has succeeded, which (re)creates the checkpoints/ parent — so ENOENT
+	// here would signal filesystem tampering or a broken invariant, not a
+	// benign empty state. Per CLAUDE.md invariant #7, surface it loudly.
 	entries, err := os.ReadDir(checkpointsPath)
 	if err != nil {
-		if errors.Is(err, os.ErrNotExist) {
-			return nil
-		}
-
 		return fmt.Errorf("listing checkpoints directory: %w", err)
 	}
 
@@ -758,7 +766,7 @@ func (s *Store) cleanupOldCheckpoints() error {
 
 		id, err := strconv.ParseUint(entry.Name(), 10, 64)
 		if err != nil {
-			continue // skip non-numeric children (e.g. "incoming")
+			continue // defensive: see ScanLatestCheckpointID
 		}
 
 		if id >= oldestToKeep {
@@ -936,14 +944,23 @@ func (s *Store) cleanupTemporaryCheckpoints() {
 // cleanupIncomingRestore removes the incoming restore directory on startup.
 // A partial incoming checkpoint can be left if the node crashed during follower sync
 // (between PrepareIncomingRestore and RestoreCheckpoint). It is safe to delete.
+//
+// Also drops the legacy path (dataDir/checkpoints/incoming) used by older
+// builds that placed staging inside checkpointsDir; migration is one-way
+// because the dir is always volatile staging.
 func (s *Store) cleanupIncomingRestore() {
-	path := filepath.Join(s.dataDir, checkpointsDir, incomingRestoreDir)
+	paths := []string{
+		filepath.Join(s.dataDir, incomingCheckpointDir),
+		filepath.Join(s.dataDir, checkpointsDir, "incoming"),
+	}
 
-	err := os.RemoveAll(path)
-	if err != nil {
-		s.logger.WithFields(map[string]any{
-			"error": err,
-		}).Errorf("Failed to clean up incoming restore directory")
+	for _, path := range paths {
+		if err := os.RemoveAll(path); err != nil {
+			s.logger.WithFields(map[string]any{
+				"path":  path,
+				"error": err,
+			}).Errorf("Failed to clean up incoming restore directory")
+		}
 	}
 }
 
@@ -968,11 +985,12 @@ func (s *Store) PrepareCheckpointRestore(checkpointID uint64) (string, error) {
 }
 
 // PrepareIncomingRestore creates a staging directory for an incoming checkpoint
-// from a leader. The directory is "checkpoints/incoming/" — a non-numeric name
-// that cannot collide with the numbered checkpoints created by CreateSnapshot
-// or the background checkpoint goroutine.
+// from a leader. The directory lives outside checkpointsDir so that
+// checkpointsDir holds only numbered checkpoints and the staging path cannot
+// collide with the numbered checkpoints created by CreateSnapshot or the
+// background checkpoint goroutine.
 func (s *Store) PrepareIncomingRestore() (string, error) {
-	dir := filepath.Join(s.dataDir, checkpointsDir, incomingRestoreDir)
+	dir := filepath.Join(s.dataDir, incomingCheckpointDir)
 
 	if err := os.RemoveAll(dir); err != nil {
 		return "", fmt.Errorf("removing existing incoming directory: %w", err)
@@ -993,8 +1011,17 @@ func (s *Store) ActivateIncomingRestore() (uint64, error) {
 	defer s.snapshotMu.Unlock()
 
 	newID := s.currentCheckPoint + 1
-	incomingDir := filepath.Join(s.dataDir, checkpointsDir, incomingRestoreDir)
-	targetDir := filepath.Join(s.dataDir, checkpointsDir, strconv.FormatUint(newID, 10))
+	incomingDir := filepath.Join(s.dataDir, incomingCheckpointDir)
+	targetParent := filepath.Join(s.dataDir, checkpointsDir)
+	targetDir := filepath.Join(targetParent, strconv.FormatUint(newID, 10))
+
+	// Ensure the checkpoints parent exists. Before the staging-dir split it
+	// was created as a side effect of PrepareIncomingRestore; now staging
+	// lives elsewhere and the parent may not exist on a freshly initialised
+	// follower whose first checkpoint comes from the leader.
+	if err := os.MkdirAll(targetParent, 0755); err != nil {
+		return 0, fmt.Errorf("creating checkpoints directory: %w", err)
+	}
 
 	if err := os.RemoveAll(targetDir); err != nil {
 		return 0, fmt.Errorf("removing target checkpoint directory: %w", err)
--- a/internal/storage/dal/store.go
+++ b/internal/storage/dal/store.go
@@ -720,32 +720,58 @@ func (s *Store) CreateSnapshot() (uint64, error) {
 	return newCheckpointID, nil
 }
 
-// cleanupOldCheckpoints removes checkpoints older than the configured maximum.
-// It keeps the most recent maxCheckpoints checkpoints.
+// cleanupOldCheckpoints removes every checkpoint dir whose ID falls below
+// (currentCheckPoint + 1) - maxCheckpoints + 1. It scans the checkpoints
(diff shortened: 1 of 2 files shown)


internal/storage/dal/store.go
```diff
@@ -720,32 +720,58 @@ func (s *Store) CreateSnapshot() (uint64, error) {
...
-		// Not enough checkpoints yet, nothing to delete
 		return nil
 	}
 
 	oldestToKeep := newCheckpoint - uint64(s.maxCheckpoints) + 1
+	checkpointsPath := filepath.Join(s.dataDir, checkpointsDir)
+
+	entries, err := os.ReadDir(checkpointsPath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
```
paul-nicolas: [Medium] ENOENT on the `checkpoints/` parent is silently swallowed on a should-not-happen path

`cleanupOldCheckpoints` runs only after `CreateSnapshot`'s `db.Checkpoint(checkpoints/<id>)` has succeeded, which necessarily (re)creates the `checkpoints/` parent. So `os.ReadDir` returning `os.ErrNotExist` here is impossible by design — if it ever fires it signals filesystem tampering or a broken invariant, not a benign empty state. Per CLAUDE.md invariant #7 ("never silently skip a should-not-happen branch"), this should surface a loud signal rather than `return nil`.

Fix: drop the `if errors.Is(err, os.ErrNotExist) { return nil }` special-case and let it fall through to `return fmt.Errorf("listing checkpoints directory: %w", err)`. Practical impact is observability only (the caller already logs-and-continues), but it keeps this path honest with the invariant the rest of the codebase holds.

paul-nicolas (commented):

Reviewed the EN-1409 checkpoint-orphan fix. The core change is correct: replacing the tracker-bounded loop with a `readdir` scan does reclaim fossil checkpoints sitting below the boot-computed floor, the `oldestToKeep` arithmetic is sound, the only non-numeric child of `checkpoints/` (`incoming/`, follower-sync staging) and the sibling `baseline/` dir are correctly excluded, and `ActivateIncomingRestore` <-> `cleanupOldCheckpoints` are serialized under `snapshotMu`, so there is no rename/delete race. No correctness or data-integrity issues found.

**Severity counts:** 0 Critical, 0 Major, 1 Medium (posted inline), 5 Minor/Nit (listed below, not posted).

**Models:** Claude + Codex (both ran; no Gemini).

**Conflict resolved:** Codex flagged the `os.ErrNotExist -> return nil` branch (medium); Claude rated error handling robust. Resolved in Codex's favor: after a successful `db.Checkpoint` the `checkpoints/` parent must exist, so that branch is impossible-by-design and CLAUDE.md invariant #7 ("never silently skip a should-not-happen branch") applies. Kept at Medium given it is a best-effort path whose error the caller already swallows — impact is observability, not correctness.

- `internal/storage/dal/store.go:732` -- [nit] doc says "only positive-integer dirs are considered checkpoints", but `strconv.ParseUint` also accepts `0`; say "non-negative integer" / "numeric".
- `internal/storage/dal/store.go:723` -- [nit] "falls below X" reads more precisely as "is strictly less than X".
- `internal/storage/dal/store.go:691` -- [minor] `CreateSnapshot` still uses the literal `"checkpoints"` while cleanup uses the `checkpointsDir` constant (pre-existing, outside this diff); fold into the constant for DRY.
- `internal/storage/dal/store_operations_test.go:321` -- [minor] the regression test does not pin the promised `incoming/` safety; add a `checkpoints/incoming/` dir to the fixture and assert it survives `CreateSnapshot()`.
- `internal/storage/dal/store_operations_test.go:96` -- [minor] no test exercises the `newCheckpoint < maxCheckpoints` early-return guard.

internal/storage/dal/store.go
gfyrag: Fixed in d2f1be366 — dropped the `os.ErrNotExist` branch and let `os.ReadDir` fall through to the wrapped error. The same follow-up moves follower-sync staging out of `checkpoints/` altogether (new location: `dataDir/incoming-checkpoint`), so there's no non-numeric sibling left in `checkpoints/` besides genuinely foreign artifacts (.DS_Store etc). Details in the updated PR description.

gfyrag:
@Azorlogh @paul-nicolas ready for another look 🙏

Since paul-nicolas's review round:
- addressed the Medium ENOENT finding (dropped the `os.ErrNotExist` swallow — CLAUDE.md invariant #7)
- pulled the follower-sync staging dir out of `checkpoints/` altogether (`dataDir/incoming-checkpoint` now), so the layout no longer requires a "skip non-numeric" branch to be load-bearing
- added `TestStore_LegacyIncomingMigration` + `TestStore_PrepareIncomingRestoreLocation` to pin both the boot-time cleanup of the legacy path and the new staging location
- rebased on latest `release/v3.0`

Details in the updated PR description.

internal/storage/dal/store.go
```diff
@@ -967,8 +1011,17 @@ func (s *Store) ActivateIncomingRestore() (uint64, error) {
...
-	targetDir := filepath.Join(s.dataDir, checkpointsDir, strconv.FormatUint(newID, 10))
+	incomingDir := filepath.Join(s.dataDir, incomingCheckpointDir)
+	targetParent := filepath.Join(s.dataDir, checkpointsDir)
+	targetDir := filepath.Join(targetParent, strconv.FormatUint(newID, 10))
+
+	// Ensure the checkpoints parent exists. Before the staging-dir split it
+	// was created as a side effect of PrepareIncomingRestore; now staging
+	// lives elsewhere and the parent may not exist on a freshly initialised
+	// follower whose first checkpoint comes from the leader.
+	if err := os.MkdirAll(targetParent, 0755); err != nil {
```
paul-nicolas, line 1022: [Medium] New fresh-follower `MkdirAll` path is untested

This `os.MkdirAll(targetParent, 0755)` is the safety net that makes follower-sync work when `checkpoints/` doesn't yet exist (the leader delivers the first checkpoint before any local snapshot). It's genuinely new behavior on the restore path -- but `ActivateIncomingRestore` is not exercised by any test in the package (`grep -rn ActivateIncomingRestore internal/**/*_test.go` returns nothing). A future refactor that drops this `MkdirAll` would silently reintroduce the fresh-follower failure with no test to catch it.

Fix: add a test that calls `ActivateIncomingRestore` on a store whose `dataDir/checkpoints/` does not exist (after `PrepareIncomingRestore` + writing a staged checkpoint) and asserts the staged dir is renamed to `checkpoints/<id>` successfully. This also pins the staging->checkpoints atomic-rename contract the layout split depends on.

paul-nicolas (commented):

**Verdict: strong PR, effectively APPROVE.** The EN-1409 fix is correct: scanning `checkpoints/` on disk and removing every numeric dir below `oldestToKeep` reclaims the crash-fossil orphans the old tracker-bounded loop could never reach. Verified independently:

- **Concurrency is safe** — `CreateSnapshot`->`cleanupOldCheckpoints` and `ActivateIncomingRestore` both hold `snapshotMu`; they cannot interleave, and cleanup's floor is always well below the newest (just-restored) ID.
- **Invariant #7 honored** — the `os.ReadDir` ENOENT case surfaces a loud wrapped error; the non-numeric `continue` is legitimately defensive (a non-numeric directory *can* appear in `checkpoints/`: `HardLink` transiently creates `checkpoints/<id>.tmp-PID-NANO` during restore), so hard-failing would be wrong.
- **`s.oldestCheckpoint` is now write-only/informational** — no authoritative reader remains after the loop was removed.
- **Layout move is safe** — nothing scans the `dataDir` top level, so the `incoming-checkpoint` sibling cannot trip `ScanLatestCheckpointID` (reads only `checkpoints/`) or any other scanner. No stale `checkpoints/incoming` references remain in code or `docs/`.

**Severity counts:** 0 Critical, 0 Major, 1 Medium (posted inline), 4 Minor/Nit (below).

**Model conflict resolved:** Codex flagged the non-numeric `continue` in `ScanLatestCheckpointID` and `cleanupOldCheckpoints` as two *Medium* invariant-#7 violations and proposed hard-failing on a non-numeric child directory. I reject both after checking the code: `HardLink` (invoked from the restore path, `internal/adapter/grpc/server_restore.go:379`, with a destination inside `checkpoints/`) creates transient `checkpoints/<id>.tmp-PID-NANO` **directories**, so a non-numeric directory there is not "impossible by design" -- hard-failing would abort boot or cleanup on a benign transient dir. The soft `continue` is correct; only the comment is imprecise (nit #1 below). Claude's "legitimately defensive" verdict prevailed.

- `internal/storage/dal/store.go:81` & `:738` -- Comment imprecision: the named examples (`.DS_Store`, editor scratch files) are *files*, already excluded by the preceding `!entry.IsDir()` check. The surviving skip case is a non-numeric *directory* (e.g. `HardLink`'s transient `<id>.tmp-*`), so the claim "no producer writes non-numeric children after the layout split" is inaccurate. Reword to name the real case; keep the soft `continue`.
- `internal/storage/dal/store_operations_test.go:381` -- Boot cleanup of the *new* staging path (`dataDir/incoming-checkpoint`) is not asserted: `TestStore_LegacyIncomingMigration` seeds only the legacy `checkpoints/incoming` path; the new path is only checked as "not pre-created". Add a case that seeds `incoming-checkpoint` and asserts boot removes it.
- `internal/storage/dal/store_operations_test.go:337` -- `CreateSnapshot()`'s return is discarded (`_, err =`) while assertions hardcode `101`/`102`; binding `newID, err :=` and asserting `newID == 102` would make the arithmetic self-documenting.
- `internal/storage/dal/store_operations_test.go:310` -- DRY: the two new tests duplicate the `logging.TestingContext` / `noop.NewMeterProvider` setup preamble (they can't use `newTestStore(t)` since it takes no config/dataDir). Consider `newTestStoreWithConfig(t, dataDir, cfg)`.

internal/storage/dal/store.go
gfyrag: Fixed in 55505cafb. Note on coverage claim: `TestSynchronizeWithLeaderDrainsBackgroundChannels` in `internal/infra/state` *does* exercise this exact path — that's the test that failed before my `MkdirAll` fix during the refactor. But you're right that the coverage is implicit (name expresses a different invariant, lives cross-package). Added `TestStore_ActivateIncomingRestore_FreshFollower` in the `dal` package that pins the contract directly: fresh store with no `checkpoints/`, prepare + marker + activate, assert `checkpoints/1/marker` present and staging renamed (not copied). A refactor dropping the `MkdirAll` now fails with a clear discovery instead of an opaque channel-drain assertion.

NumaryBot, line 1022: 🟠 **[major] ActivateIncomingRestore MkdirAll path is untested**
<sub>reported by paul-nicolas, codex</sub>

The os.MkdirAll(targetParent, 0755) call is the safety net that makes follower-sync work when checkpoints/ does not yet exist (the leader delivers the first checkpoint before any local snapshot). This path is genuinely new behavior on the restore path, but ActivateIncomingRestore is not exercised by any test in the package. A future refactor that drops this MkdirAll would silently reintroduce the fresh-follower failure with no test to catch it.

**Suggestion:** Add a test that calls ActivateIncomingRestore on a store whose dataDir/checkpoints/ does not exist (after PrepareIncomingRestore + writing a staged checkpoint) and asserts the staged dir is renamed to checkpoints/<id> successfully. This also pins the staging->checkpoints atomic-rename contract the layout split depends on.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1433#issuecomment-4833720255

gfyrag:
@paul-nicolas both of your Medium findings are addressed (`d2f1be366` for the ENOENT swallow, `55505cafb` for the fresh-follower test), threads resolved. CI green, EN-1409 is a prod-driven leak fix (93 GiB in ~14h on `blockchains`). Would you mind formally approving so I can merge? 🙏

paul-nicolas (approved):

PR #1434, opened 2026-06-29
feat(mirror): AWS RDS IAM authentication for postgres source

gfyrag:
Tracked in [EN-1411](https://formance-team.atlassian.net/browse/EN-1411).

## Summary

Operators provision a mirror ledger pointing at an AWS RDS v2 Postgres via the Ledger CRD; the controller wires the credentials path so the ledger pod mints a fresh SigV4 IAM token per connection (RDS token TTL is 15 min) instead of holding a static password.

### SRE workflow
```yaml
kind: Ledger
spec:
  name: my-mirror
  serviceRef: my-cluster
  mode: mirror
  mirrorSource:
    postgres:
      host: db.region.rds.amazonaws.com
      port: 5432            # optional, default 5432
      user: iam-user
      database: ledger
      sslmode: require      # optional, default "require"
      awsIamAuth:
        region: eu-west-1
```

Or with a static password (existing path):
```yaml
      passwordFrom:
        name: pg-secret
        key: password
```

`passwordFrom` and `awsIamAuth` are mutually exclusive; exactly one must be set.

### Surface
- **CRD** (`misc/operator/api/v1alpha1/ledger_crd_types.go`): `PostgresMirrorSource` switches from a single `DSNFrom: SecretKeyRef` to explicit `Host/Port/User/Database/SSLMode` + either `PasswordFrom: *SecretKeyRef` or `AWSIAMAuth: *AWSIAMAuthSpec{Region}`. **Breaking change** for the CRD; no prior release shipped `DSNFrom`.
- **Controller** (`misc/operator/internal/controller/ledger_crd_controller.go`): new `buildPostgresMirrorArgs` assembles the DSN from the explicit fields, looks up the password from the Secret when present, and passes `--mirror-aws-iam-region` when `AWSIAMAuth` is set. Validation enforces XOR of the two auth modes.
- **Ledger pod side** (`internal/adapter/v2/source_postgres.go`): the proto carries `PostgresAwsIamAuth{region}`; `iamBeforeConnect` mints the SigV4 token via `aws-sdk-go-v2/feature/rds/auth.BuildAuthToken` and writes it to `pgxpool.ConnConfig.Password` on every new connection. Ambient AWS credentials come from the default SDK chain on the ledger pod (IRSA, instance profile, env, profile).
- **ledgerctl**: `--mirror-aws-iam-region` flag on `ledgers create`; the operator drives it.

### What's deliberately not exposed
- HTTP body IAM fields (dropped) — the SRE path doesn't go through HTTP; the gRPC contract (proto) is the source of truth.

## Test plan
- [x] `go build ./...` clean
- [x] `just pre-commit` clean (lint, generate, tidy, operator-generate)
- [x] `go test ./internal/adapter/v2/... ./internal/adapter/http/... ./internal/application/mirror/... ./cmd/ledgerctl/... ./misc/operator/internal/controller/...`
- [x] Controller unit tests for the new CRD shape:
  - `TestBuildCreateArgs_MirrorPostgresPassword` — DSN built from explicit fields, password inlined from Secret.
  - `TestBuildCreateArgs_MirrorPostgresPasswordSpecialChars` — password with URL-special chars is percent-encoded.
  - `TestBuildCreateArgs_MirrorPostgresAWSIAMAuth` — passwordless DSN + `--mirror-aws-iam-region` flag.
  - `TestBuildCreateArgs_MirrorPostgresMissingAuth` / `_BothAuthRejected` — XOR validation.
  - `TestBuildCreateArgs_MirrorPostgresPasswordSecretMissing` — Secret lookup error surface.
- [x] Ledger-side unit tests (unchanged from earlier rounds): proto round-trip, `buildPgxPoolConfig` no-IAM / IAM region required / IAM wires BeforeConnect.
- [ ] Manual smoke against an actual RDS endpoint with IRSA on a real ledger pod — pending a target environment.

[EN-1411]: https://formance-team.atlassian.net/browse/EN-1411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
_Migrated from formancehq/ledger-v3-poc#577 on 2026-06-29._


internal/adapter/v2/source_postgres.go
```diff
@@ -143,6 +235,14 @@ func encodeDSNPassword(dsn string) string {
 	}
 
 	password := creds[colonIdx+1:]
+
+	// Skip if already percent-encoded: a valid decode that strictly shortens
+	// the input means at least one %XX sequence was present, so the password
+	// is treated as already URL-safe.
+	if decoded, err := url.PathUnescape(password); err == nil && len(decoded) < len(password) {
```
NumaryBot: 🟠 **[major] Preserve literal percent escapes in raw DSN passwords**

When a static Postgres DSN is supplied directly to `ledgerctl` with a raw password containing a literal `%` followed by two hex digits, e.g. `postgres://user:pa%2Fss@host/db`, this new shortcut treats it as already encoded and returns it unchanged. `pgxpool.ParseConfig` will then decode `%2F` to `/`, so the connection uses `pa/ss` instead of the actual `pa%2Fss`; before this change the helper escaped the `%` to `%25`. The operator path is safe because it pre-encodes with `url.UserPassword`, but the direct DSN path regresses for generated passwords containing `%XX` sequences.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1434#issuecomment-4833743061

gfyrag (commented):
Multi-model review (Paul, Claude + Codex) — addressed in `4952f10f9`:

**🔴 Major #1 — `encodeDSNPassword` idempotency corrupts raw `%XX` passwords**

Helper dropped entirely. The operator path encodes via `url.UserPassword` (single authoritative encoder); `ledgerctl --mirror-dsn` callers now pass libpq-conformant URI DSNs (URL-encode their own passwords). The length-based heuristic that masked literal `%XX` is gone. Aligns with libpq URI spec.

**🔴 Major #2 — IAM `BeforeConnect` installed regardless of `sslmode`**

Reject non-TLS sslmode when `awsIamAuth` is set (admission + runtime + CRD XValidation). New `ErrMirrorIAMRequiresTLS` sentinel. Tests pin the behavior on `disable`/`allow`/`prefer`/`""`. Acknowledged limitation: reaching `verify-full` against RDS needs `sslrootcert` pointing at the RDS CA bundle, which the structured CRD form cannot currently express — tracked as a future `extraOptions` follow-up; `require` is the secure minimum reachable today.

**🟡 Medium — CRD doesn't enforce documented XOR or sslMode/IAM coupling**

Added two `+kubebuilder:validation:XValidation` CEL rules on `PostgresMirrorSource`:
- `has(self.passwordFrom) != has(self.awsIamAuth)` → manifests with both/neither auth modes fail at the API server webhook
- `!has(self.awsIamAuth) || !has(self.sslMode) || self.sslMode in ['require','verify-ca','verify-full']` → IAM + non-TLS sslMode also rejected at apply time

The controller runtime check stays as defense-in-depth.

**Minor follow-ups bundled** in the same commit:
- `iamBeforeConnect` now bounded by `context.WithTimeout(5s)` so a hung STS/IMDS endpoint can't stall fresh pool connections.
- Dropped the `otlp.RecordError` double-attribution; pgx's connect tracer records the error on its own span via the returned `fmt.Errorf` wrap.
- New `TestBuildPgxPoolConfig_IAMAssumeRoleInstallsBeforeConnect` covers the previously-untested `stscreds.NewAssumeRoleProvider` branch.

Remaining minors **not addressed** (deferred):
- `AssumeRoleArn` ARN-shape `Pattern` validator — `MinLength=1` stays for now; out of scope for this iteration.
- region check duplication — accepted as defense-in-depth (admission + runtime + CRD).
- IAM + password DSN: IAM wins silently — behavior is unreachable in practice (operator's XOR XValidation rejects both-set, admission gate rejects empty region without IAM, the bypass requires direct gRPC of a raw `Dsn` with embedded password AND a populated `awsIamAuth`). Worth documenting later.
- `dsn` description in `openapi.yml` mentioning IAM-is-operator-only — minor doc nit, can ride a future commit.
- `domain` error refactor riding along — intentional, see `feedback_admission_structural_fsm_invariant`-style separation noted in AGENTS.md.

Thanks for the deep review pass — the TLS gate was a real security improvement.

gfyrag:
Multi-model review (Paul, Claude + Codex) — addressed in `4952f10f9`:

**🔴 Major #1 — `encodeDSNPassword` idempotency corrupts raw `%XX` passwords**

Helper dropped entirely. The operator path encodes via `url.UserPassword` (single authoritative encoder); `ledgerctl --mirror-dsn` callers now pass libpq-conformant URI DSNs (URL-encode their own passwords). The length-based heuristic that masked literal `%XX` is gone. Aligns with libpq URI spec.

**🔴 Major #2 — IAM `BeforeConnect` installed regardless of `sslmode`**

Reject non-TLS sslmode when `awsIamAuth` is set (admission + runtime + CRD XValidation). New `ErrMirrorIAMRequiresTLS` sentinel. Tests pin the behavior on `disable`/`allow`/`prefer`/`""`. Acknowledged limitation: reaching `verify-full` against RDS needs `sslrootcert` pointing at the RDS CA bundle, which the structured CRD form cannot currently express — tracked as a future `extraOptions` follow-up; `require` is the secure minimum reachable today.

**🟡 Medium — CRD doesn't enforce documented XOR or sslMode/IAM coupling**

Added two `+kubebuilder:validation:XValidation` CEL rules on `PostgresMirrorSource`:
- `has(self.passwordFrom) != has(self.awsIamAuth)` → manifests with both/neither auth modes fail at the API server webhook
- `!has(self.awsIamAuth) || !has(self.sslMode) || self.sslMode in ['require','verify-ca','verify-full']` → IAM + non-TLS sslMode rejected at apply time

The controller runtime check stays as defense-in-depth.

**Minor follow-ups bundled** in the same commit:
- `iamBeforeConnect` now bounded by `context.WithTimeout(5s)` so a hung STS/IMDS endpoint can't stall fresh pool connections.
- Dropped the `otlp.RecordError` double-attribution; pgx's connect tracer records the error on its own span via the returned `fmt.Errorf` wrap.
- New `TestBuildPgxPoolConfig_IAMAssumeRoleInstallsBeforeConnect` covers the previously-untested `stscreds.NewAssumeRoleProvider` branch.

Remaining minors **not addressed** this round (deferred):
- `AssumeRoleArn` ARN-shape `Pattern` validator — `MinLength=1` stays; out of scope.
- region check duplication — accepted defense-in-depth (admission + runtime + CRD).
- IAM + password DSN: IAM wins silently — behavior unreachable via supported entry points (operator XOR XValidation + admission gate), worth a doc note later.
- `dsn` description in `openapi.yml` mentioning IAM-is-operator-only — minor doc nit.
- `domain` error refactor riding along — intentional separation between FSM business errors and integration validators, see AGENTS.md.

Thanks for the deep review pass — the TLS gate was a real security improvement.

gfyrag:
@NumaryBot — the literal-`%XX` interpretation is the bug that the prior `encodeDSNPassword` heuristic produced (Paul's Major #1, multi-model agreement). The helper was dropped intentionally to align with the libpq URI spec: the URI form requires URL-encoded values, so a caller wanting a literal `%2F` in the password must write `%252F`. This is the trade-off documented in my [reply to Paul's review](https://github.com/formancehq/ledger/pull/1434#issuecomment-4833892153). The operator path (which is the SRE provisioning surface) is unaffected -- it encodes via `url.UserPassword`. Direct `ledgerctl --mirror-dsn` callers with weird raw passwords must URL-encode themselves, or use the libpq keyword=value DSN format which doesn't require percent-encoding. Decision stands.

gfyrag:
Second triage pass:

- 🟠 `encodeDSNPassword` regression — same finding I dismissed [earlier](https://github.com/formancehq/ledger/pull/1434#issuecomment-4833898647). The drop is intentional and aligns with Paul's multi-model Major #1 (helper was the source of the literal-`%XX` corruption bug). Decision stands.
- 🟠 `validateOrderMirrorSource` nil panic on system-scoped orders — **false-positive at the Go level**: proto-generated `GetLedgerScoped()` / `GetPayload()` are nil-safe (the getters return zero values on nil receivers), and the type assertion on a nil interface yields `(nil, false)` so we hit the `return nil` branch without a panic. Empirically verified by the existing admission tests (system-scoped orders pass through validation without panicking). That said, every other validator in `validate_order.go` has the explicit guard for readability, so I aligned the style in `152c66988`. No behavior change; the panic scenario was not reachable.

internal/adapter/v2/source_postgres.go
```diff
@@ -118,37 +139,123 @@ func (s *PostgresSource) Close() error {
...
+func dsnSSLMode(dsn string) string {
+	if u, err := url.Parse(dsn); err == nil && (u.Scheme == "postgres" || u.Scheme == "postgresql") {
+		return u.Query().Get("sslmode")
 	}
 
-	creds := rest[:lastAt]
-	hostPart := rest[lastAt:]
+	// Best-effort keyword=value parser. libpq tolerates quoted values; we
+	// strip a single layer of quotes for the common case.
+	for kv := range strings.FieldsSeq(dsn) {
```
NumaryBot: 🟠 **[major] IAM sslmode guard uses naive whitespace split, bypassable with quoted keyword=value DSNs**

The code that checks `sslmode` before installing the IAM `BeforeConnect` hook uses `strings.FieldsSeq` (or equivalent whitespace splitting) on the raw DSN string. A libpq keyword=value DSN can have quoted values containing spaces, e.g. `application_name='x sslmode=require' sslmode=disable`. The guard reads the fake `sslmode=require` inside the quoted value and proceeds to install the IAM hook, while pgx/libpq actually connects with `sslmode=disable`, sending the SigV4 token over a cleartext connection. The sslmode must be derived from the same parser that pgx uses (e.g. parsing the config returned by `pgxpool.ParseConfig`) rather than from raw string splitting.

**Suggestion:** Call `pgxpool.ParseConfig(dsn)` first, then read the `TLSConfig` field (or the parsed `ConnConfig.TLSConfig`) to determine whether TLS is actually in use. This is the authoritative answer that matches what pgx will do at connection time, and it cannot be fooled by quoted values in the raw DSN string.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1434#issuecomment-4833743061

gfyrag:
@NumaryBot — real catch, addressed in `81d21d8b4`.

Both layers (admission + runtime) now use `pgxpool.ParseConfig` and inspect `ConnConfig.TLSConfig` + `Fallbacks` -- the same representation pgx uses at connect time -- so the adversarial keyword=value DSN (`application_name='x sslmode=require' sslmode=disable`) can't fool a string-level scanner. New `TestBuildPgxPoolConfig_IAMRejectsAdversarialKeywordDSN` pins the exact bypass you described.

internal/adapter/v2/source_postgres.go
```diff
@@ -118,37 +137,114 @@ func (s *PostgresSource) Close() error {
...
-func encodeDSNPassword(dsn string) string {
-	schemeEnd := strings.Index(dsn, "://")
-	if schemeEnd == -1 {
-		return dsn
+// buildPgxPoolConfig parses the source DSN and, when AWS IAM auth is enabled,
+// wires a BeforeConnect hook that refreshes a fresh RDS IAM token per new
+// pool connection (token TTL is 15 minutes; pgxpool fires BeforeConnect on
+// every new connection it opens, so rotation is automatic).
+func buildPgxPoolConfig(ctx context.Context, cfg *commonpb.PostgresMirrorSourceConfig) (*pgxpool.Config, error) {
+	poolCfg, err := pgxpool.ParseConfig(cfg.GetDsn())
```
NumaryBot, line 147: 🟠 **[major] IAM TLS guard is bypassable via ambient PGSSLMODE environment variable**

The guard that rejects IAM-enabled DSNs without a safe sslmode calls `pgxpool.ParseConfig`, which folds in the process environment (e.g. `PGSSLMODE=require`). A mirror config that omits `sslmode` entirely — e.g. `postgres://iam-user@host/db` — will pass the guard when the pod happens to have `PGSSLMODE=require` set, even though the persisted config carries no TLS guarantee. If the pod is later restarted without that env var, the same config will fail at runtime. This makes admission/runtime behaviour dependent on ambient environment rather than the stored mirror configuration, which contradicts the intent of the guard. The same issue is present in the duplicated admission-path helper.

**Suggestion:** After calling `pgxpool.ParseConfig(dsn)` to obtain the effective config, also check whether the raw DSN string contains an explicit `sslmode` parameter (for URI form, parse the query string; for keyword=value form, parse the tokens). Reject IAM unless both conditions hold: (1) an explicit `sslmode` is present in the DSN itself, and (2) the effective `TLSConfig` is non-nil (no fallback to plaintext). This ensures the persisted config alone is sufficient to guarantee TLS, independent of process environment.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1434#issuecomment-4833743061

gfyrag:
@Azorlogh @paul-nicolas — ready for review when you have a moment 🙏

State:
- Rebased on `release/v3.0` (latest `e90a73e76`), CI green pre-rebase (re-running now)
- All NumaryBot / multi-model findings from the previous rounds addressed or explicitly dismissed (see thread comments)
- One new NumaryBot finding posted just now ([PGSSLMODE env-dependency on the TLS gate](https://github.com/formancehq/ledger/pull/1434#discussion_r79ece9cd871958ca)) — looks legitimate, I want your read before deciding whether to harden (require an explicit `sslmode=` in the DSN string itself, not just via the parsed config) or accept the operational caveat

flemzord (approved):

internal/adapter/v2/source_postgres.go
NumaryBot, line 147: 🟠 **[major] IAM TLS guard bypassable via ambient PGSSLMODE environment variable**

The guard that rejects IAM-enabled DSNs without a safe sslmode delegates entirely to `pgxpool.ParseConfig`, which folds in process environment variables such as `PGSSLMODE=require`. A mirror config that omits an explicit `sslmode` in the DSN (e.g. `postgres://iam-user@host/db`) will silently pass the guard when the pod happens to have `PGSSLMODE=require` set, even though the persisted config carries no TLS guarantee. If the pod is later restarted without that env var, the same config will fail or send IAM tokens over a cleartext connection. Admission and runtime behaviour become dependent on ambient environment rather than the stored mirror configuration, which contradicts the intent of the guard.

**Suggestion:** After calling `pgxpool.ParseConfig(dsn)` to obtain the effective config, additionally verify that the raw DSN string itself contains an explicit `sslmode` parameter: for URI-form DSNs parse the query string; for keyword=value DSNs scan the tokens. Reject IAM authentication unless both conditions hold: (1) an explicit `sslmode` is present in the raw DSN, and (2) the effective `TLSConfig` is non-nil with no plaintext fallback. This makes the guard independent of the pod environment.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1434#issuecomment-4833743061

internal/adapter/v2/source_postgres.go
```diff
@@ -118,37 +139,164 @@ func (s *PostgresSource) Close() error {
...
```
NumaryBot, line 147: 🟠 **[major] IAM TLS guard bypassable via ambient PGSSLMODE environment variable**

The guard that rejects IAM-enabled DSNs without a safe sslmode delegates entirely to `pgxpool.ParseConfig`, which folds in process environment variables such as `PGSSLMODE=require`. A mirror config that omits an explicit `sslmode` in the DSN (e.g. `postgres://iam-user@host/db`) will silently pass the guard when the pod happens to have `PGSSLMODE=require` set, even though the persisted config carries no TLS guarantee. If the pod is later restarted without that env var, the same config will fail or send IAM tokens over a cleartext connection. This makes admission and runtime behaviour dependent on ambient environment rather than the stored mirror configuration, which contradicts the intent of the guard.

**Suggestion:** After calling `pgxpool.ParseConfig(dsn)` to obtain the effective config, additionally verify that the raw DSN string itself contains an explicit `sslmode` parameter: for URI-form DSNs parse the query string; for keyword=value DSNs scan the key=value tokens. Reject IAM authentication unless both conditions hold: (1) an explicit `sslmode` is present in the raw DSN string, and (2) the effective `TLSConfig` is non-nil with no plaintext fallback. This makes the guard deterministic regardless of ambient environment variables.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1434#issuecomment-4833743061

paul-nicolas (approved):

internal/adapter/v2/source_postgres.go
gfyrag: This is a false positive on the re-scan. The finding claims the guard "delegates entirely to `pgxpool.ParseConfig`" — that was true at the previous commit and is exactly what `41cf29687` addresses.

The new gate in `buildPgxPoolConfig` (and its admission twin in `dsnEnforcesTLS`) checks the **raw DSN string** *before* touching pgxpool:

```go
if !dsnIsURIForm(cfg.GetDsn()) {
    return nil, errors.New("aws_iam_auth requires a URI-form DSN")
}
if !dsnHasExplicitSSLMode(cfg.GetDsn()) || !poolConfigEnforcesTLS(poolCfg) {
    return nil, errors.New("aws_iam_auth requires an explicit sslmode ...")
}
```

`dsnHasExplicitSSLMode` uses `url.Parse` + `Query().Has("sslmode")` on the raw string — `PGSSLMODE` in the process env cannot satisfy it.

Pinned by two dedicated tests that set `PGSSLMODE=require` in the process env and assert admission + runtime both still reject a DSN without `sslmode=` in the raw string:
- `TestBuildPgxPoolConfig_IAMRejectsDSNWithoutExplicitSSLMode` (runtime)
- `TestValidateOrder_MirrorIAMRejectsPGSSLMODEBypass` (admission)

PR #1435, opened 2026-06-29
docs(architecture): index/indexer pages + restructure docs by subsystem

gfyrag:
## Summary

Two threads, landed in one PR:

### 1. New index/indexer documentation (initial scope)

- **`subsystems/indexer/indexes.md`** — `commonpb.Index` proto, per-replica `IndexVersionState`, on-demand statistics, checker coverage.
- **`subsystems/indexer/indexer.md`** — builder loop, two-pass commit, handlers, read-store key layout, atomic switch, schema rewrite. Includes a detailed *Changing a Metadata Key's Type* section covering the `SetMetadataFieldType` pipeline, the `requiredIndexedSeq` gate, the deferred switch, dual-write semantics, and the transient query window.
- **`AGENTS.md`** — new rule under *Documentation Maintenance*: every new technical mechanism / non-obvious invariant must have a dedicated architecture page. Plus a fix to invariant #8 (stale `IndexReady` rationale for `compareIndexes`).

Several review-round corrections landed across follow-up commits: null encoding, `CreatedTransaction` handler list, `DropIndex` keyspace claim, dual-write split between CreateIndex backfill vs schema rewrite, `requireIndexReady` typo, `requiredIndexedSeq` per-batch sampling.

### 2. Architecture docs restructure (added on top)

Replaces the four-bucket layout (`core/` / `storage/` / `data-model/` / `api/`) with a per-subsystem split that mirrors the code:

```
docs/technical/architecture/
├── README.md
├── overview.md
├── data-flows.md
├── data-model.md
├── primitives/uint256-wire-format.md
└── subsystems/
    ├── admission/    ├── api/         ├── attributes/
    ├── chapters/     ├── consensus/   ├── events-mirror/
    ├── fsm/          ├── indexer/     ├── read-path/
    ├── scripting/    └── storage/
```

Each `subsystems/<name>/` has its own README listing the docs it owns and cross-refs to related subsystems. The top-level `architecture/README.md` is rewritten around *Start here / Subsystems / Cross-cutting / Conventions*.

- 25 docs moved with `git mv` to preserve history.
- 75 internal cross-links + ~30 external references (in `docs/ops/`, `docs/sales/`, `docs/technical/contributing/`, `docs/drafts/`) rewritten by a script that resolves each link against the source file's pre-move parent and re-emits from the post-move location. A second pass confirms every internal `.md` link resolves to an existing file.
- `subsystems/checker/` intentionally not created in this PR — its content (`audit-chain.md`, `checker.md`) doesn't exist yet and will land in a follow-up alongside the missing per-subsystem fillers (`admission/pipeline.md`, `attributes/bloom.md`, `read-path/query-pipeline.md`, etc.).

## Test plan

- [ ] Read both new pages end-to-end (indexes.md, indexer.md) and verify code references against the current tree.
- [ ] Click through the new architecture README and confirm the subsystem map matches expectations.
- [ ] Spot-check that external docs (e.g. `docs/ops/cluster-operations.md`, `docs/technical/architecture-overview.md`) still link correctly into the new structure.
- [ ] Confirm no doc references the deleted `core/` / `storage/` / `data-model/` / `api/` directories.

---
_Migrated from formancehq/ledger-v3-poc#575 on 2026-06-29._


docs/technical/architecture/subsystems/checker/audit-chain.md
```diff
@@ -0,0 +1,197 @@
...
+| Delete entry *N* | Sequence gap on read → `CHECK_STORE_ERROR_TYPE_SEQUENCE_GAP`. |
+| Swap entries *N* and *M* | At least one of them has a `prev_hash` link that no longer matches → mismatch at the earliest violating slot. |
+| Rewrite `hash[N]` to match a forged payload | `hash[N+1]` was computed against the original `hash[N]`. Recomputing forward from the forged value produces `computed[N+1] ≠ stored[N+1]`. The attacker must rewrite every entry from *N* to the head, but cannot regenerate hashes without the per-cluster BLAKE3 key. |
+| Smuggle items into `entry.items` on disk | The on-disk row has `items = nil` by design (`machine.go:1403`); the checker flags `len(entry.Items) > 0` as tampering (`internal/application/check/checker.go:1523-1531`). |
+
+The chain does *not* defend against an attacker who has the cluster's BLAKE3 key — that key is local to the node and is the same secret that lets the node propose. Securing the key is part of the threat model the operator-level [Security](../../../../security/) and [Request Signing](../../../../ops/signing.md) docs cover.
```
NumaryBot, line 171: 🟠 **[major] Fix the broken Security documentation link**

In this new page, the relative link `../../../../security/` resolves to `docs/security/`, but that directory does not exist in the repository. Readers following the threat-model reference from the audit-chain docs will hit a dead link; please point it at an existing security/authentication/signing document or add the missing target.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1435#issuecomment-4833721435

flemzord (approved):

docs/technical/architecture/subsystems/checker/audit-chain.md
NumaryBot, line 171: 🟠 **[major] Broken link to non-existent security directory**

The relative link `../../../../security/` resolves to `docs/security/`, which does not exist in the repository. Readers following the threat-model reference from the audit-chain page will hit a dead link.

**Suggestion:** Update the link to point to an existing security, authentication, or signing document, or create the missing target directory/file.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1435#issuecomment-4833721435

docs/technical/architecture/subsystems/admission/signing.md
```diff
@@ -0,0 +1,98 @@
...
+
+Verification happens in the admission stage, **before** the proposal reaches Raft (`resolveBatch()` in `internal/application/admission/admission.go`). Doing it here is a deliberate DoS-rejection point: invalid signatures cost the cluster nothing more than a public-key lookup + an Ed25519 verify.
+
+### Audit-time
+
+The signature is propagated all the way into the audit chain:
+
+- `Proposal.Signature` carries the client's signature.
+- The header payload that gets hashed by `BuildHashedHeaderPayload` includes the signature (see [audit-chain.md](../checker/audit-chain.md)).
+- The signature ends up on each `Log` produced by the proposal as well, so an auditor reading a single log row can re-verify it independently against the client's public key.
```
NumaryBot, line 46: 🟠 **[major] Incorrect claim that Log carries the client request signature**

The page states that `common.Log` carries the client request signature for signed Apply batches, but the proto only has `response_signature`. The request signature is stored on the proposal/audit entry, not on the log. This overstates the auditability model and will mislead auditors trying to re-verify logs against a client's public key via a non-existent `Log.Signature` field.

**Suggestion:** Clarify that the client request signature is stored on the proposal/audit entry rather than on `common.Log`, and remove references to `Log.Signature`.

docs/technical/architecture/subsystems/checker/checker.md
```diff
@@ -0,0 +1,137 @@
...
+| 9 | `compareIndexes` | `SubAttrIndex` registry matches the set derived from `CreateIndex` / `DropIndex` / `RemovedMetadataFieldType` / `DeleteLedger` logs (presence + identity only) | Replay of the index-affecting log types | `INDEX_MISMATCH` |
+
+Notes:
+
+- **`compareIndexes` covers presence + identity** (`IndexID` match), **not `BuildStatus`** and **not `IndexVersionState`**. `BuildStatus` is informational and per-replica; `IndexVersionState` is by design local and may legitimately differ across nodes mid-rewrite (see [indexer / indexes.md](../indexer/indexes.md)).
+- **`compareIdempotencyOutcomes`** is the one pass that consumes a side-effect of `verifyAuditHashChain` (the expected-outcome map). The two are coupled by design — re-deriving the idempotency outcome means re-walking the chain anyway.
+- **Order matters**: `verifyAuditHashChain` runs **first**. A broken chain stops the walk before any downstream pass — running them with a tampered chain would produce noise from already-detected corruption.
```
NumaryBot, line 48: 🟠 **[major] Incorrect claim that checker halts all passes after a hash mismatch**

The documentation states that no downstream pass runs after `CheckStore` hits an audit hash mismatch, but in the actual code `verifyAuditHashChain` emits the error and returns `nil`, so `Checker.Check` continues into the replay/projection comparison passes. Operators reading this will misinterpret additional projection errors that appear after a hash-chain finding.

**Suggestion:** Update the description to reflect that subsequent passes (replay/projection comparison) still execute after a hash-chain mismatch, and that errors from those passes are also reported.

docs/technical/architecture/subsystems/read-path/prepared-queries.md
```diff
@@ -0,0 +1,108 @@
...
+         | address == "foo:*"
+         | source == ...
+         | destination == ...
+         | exists metadata[key]
+         | exists transaction <reference>
+```
+
+The reserved words are: `and`, `or`, `not`, `in`, `between`, `metadata`, `address`, `source`, `destination`, `exists`, `true`, `false`. Nesting depth is capped at 200 tokens to keep parsing bounded.
+
+The parser produces a `*commonpb.QueryFilter` — a typed proto that the FSM stores verbatim. The compiler (`internal/query/compile.go:90`) turns that proto into an iterator tree at execution time, gated through `requireIndexReady` so an index in mid-rewrite returns `ErrIndexBuilding` rather than producing partial results.
```
NumaryBot, line 52: 🟠 **[major] Incorrect description of index readiness during a schema rewrite**

The page implies `ErrIndexBuilding` is returned for mid-rewrite indexes, but during a `SetMetadataFieldType` schema rewrite `CurrentVersion` remains non-zero and `requireIndexReady` returns that version. `ErrIndexBuilding` is only returned when the local index has no built current version. This gives readers the opposite expectation from the indexer page's transient query-window semantics.

**Suggestion:** Correct the description so it reflects that mid-rewrite indexes remain queryable at their current version, and that `ErrIndexBuilding` is only raised when no current version has been built at all.

docs/technical/architecture/subsystems/indexer/indexer.md
```diff
@@ -0,0 +1,279 @@
...
+| `RevertedTransaction` | `indexRevertedTransaction` | Inverse adjustments to the postings mappings; the revert link is kept (not deleted) so revert lookups stay queryable. |
+| `SavedMetadata` | `indexSavedMetadata` → `dualWriteMetadataIndex` | New `(value, entity)` row in the metadata index, plus a reverse-map row for later schema rewrites. |
+| `DeletedMetadata` | `indexDeletedMetadata` → `dualDeleteMetadataEntry` | Removes the metadata index row and the reverse-map row. |
+| `SetMetadataFieldType` | `addSchemaRewriteTask` | Schedules a deferred rewrite that re-encodes the existing values under a new type tag (see [Schema Rewrite](#schema-rewrite)). |
```
NumaryBot, line 85: 🟡 **[minor] Broken intra-page anchor for schema rewrite section**

The `#schema-rewrite` anchor used in the `SetMetadataFieldType` handler link does not exist on this page. Clicking the link leaves the reader at the top of the page instead of navigating to the rewrite details. The target section appears to have been renamed to 'Changing a Metadata Key's Type'.

**Suggestion:** Update the anchor to match the actual heading slug (e.g. `#changing-a-metadata-keys-type`) or rename the target heading to `Schema Rewrite` to match the anchor.

PR #1440, opened 2026-06-29
perf(EN-1403): dethash zero-alloc sorted-keys via sync.Pool

gfyrag:
[EN-1403](https://formance-team.atlassian.net/browse/EN-1403)

## Why

The `protoc-gen-dethash` plugin generated `MarshalToSizedBufferDeterministicVT` methods that allocated a fresh sorted-keys slice on every marshal via `slices.Sorted(maps.Keys(m))`. For typical metadata-bearing messages this dominated the cost of the deterministic marshal — a `TransactionState` with 500 metadata entries paid ~18.8 KB across 13 allocations every time. A `LedgerInfo` with two maps (Metadata + AccountTypes, 500/100 entries) paid ~23 KB across 24 allocations.

Surfaced while benchmarking the FSM cross-node digest opt-in (PR #574). The non-deterministic encoder is 0-alloc; the deterministic encoder shouldn't have a worse allocation profile by design.

## What

Replace the per-marshal `slices.Sorted(maps.Keys(m))` call with a `sync.Pool` of `*[]K` per map-key kind, declared once at the top of each generated dethash file:

- Plugin scans every file's messages, collects which map-key kinds are actually used (string, int32, uint32, int64, uint64), and emits one `sync.Pool` per used kind. Files with no maps emit no pool (and no `sync` import).
- `genMapMarshal` now generates a Get / fill / sort / Put pattern around the existing inner marshal loop.
- Steady-state allocations fall to zero; the underlying capacity persists across calls and grows once on the first marshal that exceeds 16 keys.
- Pool is package-private (`_dethash` prefix), no public API change.

## Benchmark (Apple M5 Pro, marshal-only)

| Case | Before | After |
|---|---|---|
| TransactionState metadata=5 | 290 ns, 304 B, 7 alloc | **166 ns, 0 B, 0 alloc** |
| TransactionState metadata=50 | 2531 ns, 2.2 KB, 10 alloc | **1956 ns, 0 B, 0 alloc** |
| TransactionState metadata=500 | 41.6 µs, 18.8 KB, 13 alloc | **38.0 µs, 0 B, 0 alloc** |
| LedgerInfo meta=5_acct=3 | 496 ns, 480 B, 13 alloc | **288 ns, 0 B, 0 alloc** |
| LedgerInfo meta=500_acct=100 | 49.2 µs, 23 KB, 24 alloc | **44.0 µs, 0 B, 0 alloc** |
| Transaction post=100_meta=100 | 7.7 µs, 4.5 KB, 11 alloc | **6.5 µs, 0 B, 0 alloc** |

Surcoût vs the non-deterministic baseline drops to 1.1x-1.4x for typical metadata-size maps (≤ 20 entries), tops out at ~2.5x for huge 500-entry maps. The residual is the sort O(n log n) itself — inherent to the determinism contract.

## Test plan

- [x] All existing tests pass (`go test ./...`)
- [x] Pre-commit clean (`just pre-commit`: 0 lint issues)
- [x] Generated dethash code visually inspected: pool decl at top, Get/Put pattern around the marshal loop, no `slices.Sorted` / `maps.Keys` left in the generated code.
- [x] Bench from PR #574's `internal/storage/dal/marshal_bench_test.go` confirms 0 alloc / latency drop across all map-bearing shapes.

## Notes

This PR ships with `feat/dethash-zero-alloc-keys`; PR #574 (cross-node FSM digest) is rebased on top.

[EN-1403]: https://formance-team.atlassian.net/browse/EN-1403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
_Migrated from formancehq/ledger-v3-poc#579 on 2026-06-29._

### DIFF
--- a/tools/protoc-gen-dethash/main.go
+++ b/tools/protoc-gen-dethash/main.go
@@ -18,7 +18,6 @@ import (
 
 var (
 	slicesPkg       = protogen.GoImportPath("slices")
-	sortPkg         = protogen.GoImportPath("sort")
 	syncPkg         = protogen.GoImportPath("sync")
 	protohelpersPkg = protogen.GoImportPath("github.com/planetscale/vtprotobuf/protohelpers")
 	binaryPkg       = protogen.GoImportPath("encoding/binary")
@@ -45,8 +44,12 @@ type keyKindInfo struct {
 
 // supportedKeyKinds maps every map-key kind the plugin accepts to its
 // pool/sort recipe. The kinds match the panic guard in genMapMarshal.
+// `slices.Sort` is used uniformly (Go 1.21+ pdqsort generic) — measured
+// marginally faster than `sort.Strings` on strings of any size and the
+// fastest std-lib option on ints; using a single sort function avoids
+// pulling the `sort` package into the generated file.
 var supportedKeyKinds = map[protoreflect.Kind]keyKindInfo{
-	protoreflect.StringKind: {goType: "string", pkgVar: "_dethashKeyPoolString", sortPkg: sortPkg, sortFunc: "Strings"},
+	protoreflect.StringKind: {goType: "string", pkgVar: "_dethashKeyPoolString", sortPkg: slicesPkg, sortFunc: "Sort"},
 	protoreflect.Int32Kind:  {goType: "int32", pkgVar: "_dethashKeyPoolInt32", sortPkg: slicesPkg, sortFunc: "Sort"},
 	protoreflect.Uint32Kind: {goType: "uint32", pkgVar: "_dethashKeyPoolUint32", sortPkg: slicesPkg, sortFunc: "Sort"},
 	protoreflect.Int64Kind:  {goType: "int64", pkgVar: "_dethashKeyPoolInt64", sortPkg: slicesPkg, sortFunc: "Sort"},
--- a/tools/protoc-gen-dethash/main.go
+++ b/tools/protoc-gen-dethash/main.go
@@ -18,11 +18,40 @@ import (
 
 var (
 	slicesPkg       = protogen.GoImportPath("slices")
-	mapsPkg         = protogen.GoImportPath("maps")
+	sortPkg         = protogen.GoImportPath("sort")
+	syncPkg         = protogen.GoImportPath("sync")
 	protohelpersPkg = protogen.GoImportPath("github.com/planetscale/vtprotobuf/protohelpers")
 	binaryPkg       = protogen.GoImportPath("encoding/binary")
 	mathPkg         = protogen.GoImportPath("math")
 )
+
+// keyKindInfo describes how to declare a sync.Pool of sorted keys for a
+// supported map-key kind. The generated marshaler grabs a *[]K from the pool,
+// fills it, sorts it in place, and returns it to the pool after the map has
+// been encoded — replacing the previous `slices.Sorted(maps.Keys(...))` call
+// that allocated a fresh slice on every marshal.
+type keyKindInfo struct {
+	// goType is the Go literal for the key slice element (e.g. "string",
+	// "int64").
+	goType string
+	// pkgVar is the unique package-level variable name for this kind's pool,
+	// e.g. "_dethashKeyPoolString".
+	pkgVar string
+	// sortCall is the sort statement applied to a `keys` slice of this type.
+	// Generated as `<package>.<call>(keys)` so the import is tracked.
+	sortPkg  protogen.GoImportPath
+	sortFunc string
+}
+
+// supportedKeyKinds maps every map-key kind the plugin accepts to its
+// pool/sort recipe. The kinds match the panic guard in genMapMarshal.
+var supportedKeyKinds = map[protoreflect.Kind]keyKindInfo{
+	protoreflect.StringKind: {goType: "string", pkgVar: "_dethashKeyPoolString", sortPkg: sortPkg, sortFunc: "Strings"},
+	protoreflect.Int32Kind:  {goType: "int32", pkgVar: "_dethashKeyPoolInt32", sortPkg: slicesPkg, sortFunc: "Sort"},
+	protoreflect.Uint32Kind: {goType: "uint32", pkgVar: "_dethashKeyPoolUint32", sortPkg: slicesPkg, sortFunc: "Sort"},
+	protoreflect.Int64Kind:  {goType: "int64", pkgVar: "_dethashKeyPoolInt64", sortPkg: slicesPkg, sortFunc: "Sort"},
+	protoreflect.Uint64Kind: {goType: "uint64", pkgVar: "_dethashKeyPoolUint64", sortPkg: slicesPkg, sortFunc: "Sort"},
+}
 
 func main() {
 	protogen.Options{}.Run(func(gen *protogen.Plugin) error {
@@ -81,9 +110,86 @@ func generateFile(gen *protogen.Plugin, file *protogen.File, hasMap map[string]b
 	g.P()
 	g.P("package ", file.GoPackageName)
 
+	// First, find every map-key kind used in this file so we can emit one
+	// sync.Pool per kind. The pool reuses the sorted-keys slice across
+	// marshal calls, replacing the per-call `slices.Sorted(maps.Keys(m))`
+	// allocation. The pool is package-private so it does not leak into
+	// the public API.
+	usedKeyKinds := map[protoreflect.Kind]bool{}
+	for _, msg := range file.Messages {
+		collectMapKeyKinds(msg, usedKeyKinds)
+	}
+
+	emitKeyPools(g, usedKeyKinds)
+
 	for _, msg := range file.Messages {
 		walkMessages(g, msg, hasMap)
 	}
+}
+
+// collectMapKeyKinds walks msg + every nested message and records the kind
+// of every map<K, V> field's key. Limited to the kinds supported by
+// genMapMarshal — anything else would have panicked the plugin elsewhere.
+func collectMapKeyKinds(msg *protogen.Message, out map[protoreflect.Kind]bool) {
+	if msg.Desc.IsMapEntry() {
+		return
+	}
+
+	for _, f := range msg.Fields {
+		if f.Desc.IsMap() {
+			keyKind := f.Message.Fields[0].Desc.Kind()
+			if _, ok := supportedKeyKinds[keyKind]; ok {
+				out[keyKind] = true
+			}
+		}
+	}
+
+	for _, nested := range msg.Messages {
+		collectMapKeyKinds(nested, out)
+	}
+}
+
+// emitKeyPools writes one sync.Pool per used map-key kind, plus a small
+// initial slice capacity tuned for the typical metadata-style map (~16
+// entries — beyond that the pool's slice grows once and stays grown,
+// which is exactly the behaviour we want).
+func emitKeyPools(g *protogen.GeneratedFile, used map[protoreflect.Kind]bool) {
+	if len(used) == 0 {
+		return
+	}
+
+	// Iterate in a deterministic order so the generated file is stable
+	// across plugin runs (map iteration would otherwise reshuffle the
+	// declarations, producing useless diffs).
+	ordered := []protoreflect.Kind{
+		protoreflect.StringKind,
+		protoreflect.Int32Kind,
+		protoreflect.Uint32Kind,
+		protoreflect.Int64Kind,
+		protoreflect.Uint64Kind,
+	}
+
+	g.P()
+	g.P("// Map-key scratch pools. Each MarshalToSizedBufferDeterministicVT that")
+	g.P("// encodes a map<K, V> grabs a *[]K from the matching pool, fills it,")
+	g.P("// sorts it in place, and returns it. Steady-state allocations are zero:")
+	g.P("// the pool reuses the slice across marshal calls. The first marshal of")
+	g.P("// a given kind warms the pool with a 16-cap slice; larger maps grow it")
+	g.P("// once and the grown capacity persists.")
+	g.P("var (")
+
+	for _, kind := range ordered {
+		if !used[kind] {
+			continue
+		}
+
+		info := supportedKeyKinds[kind]
+		g.P("  ", info.pkgVar, " = ", syncPkg.Ident("Pool"), "{")
+		g.P("    New: func() any { s := make([]", info.goType, ", 0, 16); return &s },")
+		g.P("  }")
+	}
+
+	g.P(")")
 }
 
 func walkMessages(g *protogen.GeneratedFile, msg *protogen.Message, hasMap map[string]bool) {
@@ -264,9 +370,24 @@ func genMapMarshal(g *protogen.GeneratedFile, f *protogen.Field, hasMap map[stri
 	valField := f.Message.Fields[1]
 	fieldNum := f.Desc.Number()
 	tag := encodeTag(fieldNum, 2)
+	keyField := f.Message.Fields[0]
+	keyKind := keyField.Desc.Kind()
 
+	info, ok := supportedKeyKinds[keyKind]
+	if !ok {
+		panic(fmt.Sprintf("protoc-gen-dethash: unsupported map key type %s in field %s", keyKind, f.Desc.FullName()))
+	}
+
+	// Borrow a sorted-keys scratch slice from the file-scoped pool, fill it,
+	// sort it in place, and return it after the loop. Steady-state zero alloc.
 	g.P("  if len(m.", goName, ") > 0 {")
-	g.P("    for _, k := range ", slicesPkg.Ident("Sorted"), "(", mapsPkg.Ident("Keys"), "(m.", goName, ")) {")
+	g.P("    keysPtr := ", info.pkgVar, ".Get().(*[]", info.goType, ")")
+	g.P("    keys := (*keysPtr)[:0]")
+	g.P("    for k := range m.", goName, " {")
+	g.P("      keys = append(keys, k)")
+	g.P("    }")
+	g.P("    ", info.sortPkg.Ident(info.sortFunc), "(keys)")
+	g.P("    for _, k := range keys {")
 	g.P("      v := m.", goName, "[k]")
 	g.P("      baseI := i")
 
@@ -298,8 +419,7 @@ func genMapMarshal(g *protogen.GeneratedFile, f *protogen.Field, hasMap map[stri
 	}
 
 	// Write key (field 1)
-	keyField := f.Message.Fields[0]
-	switch keyField.Desc.Kind() {
+	switch keyKind {
 	case protoreflect.StringKind:
 		g.P("      i -= len(k)")
 		g.P("      copy(dAtA[i:], k)")
@@ -311,7 +431,7 @@ func genMapMarshal(g *protogen.GeneratedFile, f *protogen.Field, hasMap map[stri
 		g.P("      i--")
 		g.P("      dAtA[i] = 0x8") // field 1, wire type varint
 	default:
-		panic(fmt.Sprintf("protoc-gen-dethash: unsupported map key type %s in field %s", keyField.Desc.Kind(), f.Desc.FullName()))
+		panic(fmt.Sprintf("protoc-gen-dethash: unsupported map key type %s in field %s", keyKind, f.Desc.FullName()))
 	}
 
 	// Write entry length + outer tag
@@ -319,6 +439,11 @@ func genMapMarshal(g *protogen.GeneratedFile, f *protogen.Field, hasMap map[stri
 	writeTag(g, tag)
 
 	g.P("    }")
+	// Return the (possibly grown) scratch slice to the pool. The pool keeps
+	// the underlying capacity, so the next marshal that needs at most that
+	// many keys hits zero allocation.
+	g.P("    *keysPtr = keys")
+	g.P("    ", info.pkgVar, ".Put(keysPtr)")
 	g.P("  }")
 }
 
--- a/tools/protoc-gen-dethash/main.go
+++ b/tools/protoc-gen-dethash/main.go
@@ -11,6 +11,7 @@ package main
 import (
 	"fmt"
 	"path"
+	"strings"
 
 	"google.golang.org/protobuf/compiler/protogen"
 	"google.golang.org/protobuf/reflect/protoreflect"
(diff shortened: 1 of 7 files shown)


tools/protoc-gen-dethash/main.go
```diff
@@ -311,14 +434,19 @@ func genMapMarshal(g *protogen.GeneratedFile, f *protogen.Field, hasMap map[stri
...
 
 	// Write entry length + outer tag
 	g.P("      i = ", protohelpersPkg.Ident("EncodeVarint"), "(dAtA, i, uint64(baseI-i))")
 	writeTag(g, tag)
 
 	g.P("    }")
+	// Return the (possibly grown) scratch slice to the pool. The pool keeps
+	// the underlying capacity, so the next marshal that needs at most that
+	// many keys hits zero allocation.
+	g.P("    *keysPtr = keys")
```
NumaryBot, line 533: 🟠 **[major] Clear pooled string-key slices before putting them back**

For string-key maps, this stores a slice whose backing array still contains all map keys. Because the pool is package-global and the array has pointer-containing string elements, those key strings stay reachable after the marshaled message/map is dropped (especially after a large or unique metadata map followed by smaller maps), so this can turn the allocation optimization into memory retention of user metadata until the pool discards or overwrites the array. Clear the keys before `Put` (or avoid pooling pointer-containing key kinds).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1440#issuecomment-4833742580

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1440#issuecomment-4833742580

tools/protoc-gen-dethash/main.go
NumaryBot, line 533: 🟠 **[major] Pooled string-key slices retain map key strings after use**

When a map with string keys is processed, the pooled slice's backing array still holds references to all map key strings after the slice is returned to the pool. Because the pool is package-global and strings are pointer-containing types, those key strings remain reachable until the pool discards or overwrites the slot. For large or unique metadata maps followed by smaller ones, this effectively turns the zero-alloc optimisation into unbounded memory retention of user metadata.

**Suggestion:** Zero out (nil/clear) the string elements of the key slice before calling pool.Put(), or avoid pooling slices whose element type contains pointers. For example: `for i := range keys { keys[i] = "" }; pool.Put(keys[:0])`.

tools/protoc-gen-dethash/main.go
```diff
@@ -18,12 +18,44 @@ import (
...
+}
+
+// supportedKeyKinds maps every map-key kind the plugin accepts to its
+// pool/sort recipe. The kinds match the panic guard in genMapMarshal.
+// `slices.Sort` is used uniformly (Go 1.21+ pdqsort generic) — measured
+// marginally faster than `sort.Strings` on strings of any size and the
+// fastest std-lib option on ints; using a single sort function avoids
+// pulling the `sort` package into the generated file.
+var supportedKeyKinds = map[protoreflect.Kind]keyKindInfo{
+	protoreflect.StringKind: {goType: "string", pkgVar: "_dethashKeyPoolString", sortPkg: slicesPkg, sortFunc: "Sort"},
```
NumaryBot: 🟠 **[major] Fixed pool identifier names cause redeclared-symbol compile errors in multi-file packages**

The generator emits package-level identifiers such as `_dethashKeyPoolString` unconditionally per generated file. When two `.proto` files share the same `go_package` and both contain a map with the same key kind, each generated `_dethash.pb.go` file declares the same identifier, producing a compile-time 'redeclared in this block' error. The current repo layout avoids this only by accident (one proto file per package).

**Suggestion:** Emit pool variable declarations once per Go package rather than once per generated file. This could be achieved with a separate, manually maintained or generator-produced helper file per package, or by making the identifier name unique per proto file (e.g., prefixed with the proto file's base name).

tools/protoc-gen-dethash/main.go
```diff
@@ -311,14 +502,28 @@ func genMapMarshal(g *protogen.GeneratedFile, f *protogen.Field, hasMap map[stri
...
+	//
+	// For pointer-containing key kinds (today: string headers), zero the
+	// slice elements first. Without this, the backing array keeps the last
+	// marshal's keys reachable until the pool slot is overwritten or
+	// evicted — turning the per-call allocation saved into unbounded heap
+	// retention of user metadata on many-unique-key workloads.
+	if info.pointerContaining {
+		g.P("    clear(keys)")
+	}
+	g.P("    *keysPtr = keys")
```
NumaryBot, line 533: 🟠 **[major] Pooled string-key slices retain map key strings after use**
<sub>reported by NumaryBot, NumaryBot</sub>

When a map with string keys is processed, the pooled slice's backing array still holds references to all map key strings after the slice is returned to the pool. Because the pool is package-global and strings are pointer-containing types, those key strings remain reachable until the pool discards or overwrites the slot. For large or unique metadata maps followed by smaller ones, this effectively turns the zero-alloc optimisation into unbounded memory retention of user metadata.

**Suggestion:** Zero out the string elements of the key slice before calling pool.Put(), or avoid pooling slices whose element type contains pointers. For example: `for i := range keys { keys[i] = "" }; pool.Put(keys[:0])`.

tools/protoc-gen-dethash/main.go
```diff
@@ -18,12 +20,97 @@ import (
...
+		case r >= 'a' && r <= 'z':
+			if upper {
+				r -= 'a' - 'A'
+				upper = false
+			}
+			b.WriteRune(r)
+		case r >= 'A' && r <= 'Z' || r >= '0' && r <= '9':
+			b.WriteRune(r)
+			upper = false
+		default:
```
NumaryBot, line 111: 🟠 **[major] Sanitized pool prefix collisions for proto files differing only in separator characters**

When two `.proto` files in the same `go_package` have basenames that only differ by separators (e.g., `foo-bar.proto` and `foo_bar.proto`), the normalization that strips separators produces the same prefix (`FooBar`). Both generated files would then declare the same `_dethashKeyPool<Prefix><Kind>` variable, causing a compile-time 'redeclared in this block' error. This means the per-file uniqueness fix is still incomplete for valid proto filenames.

**Suggestion:** Use a collision-resistant sanitization strategy, for example by replacing separators with a canonical delimiter (e.g., `_`) rather than stripping them, so `foo-bar` maps to `Foo_Bar` and `foo_bar` maps to `Foo_Bar`… or better, encode the separator distinctly (e.g., `FooHyphenBar` vs `FooUnderBar`) to guarantee distinct identifiers.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1440#issuecomment-4833742580

tools/protoc-gen-dethash/main.go
gfyrag: False positive — `clear(keys)` is already emitted before every `Put` for pointer-containing key kinds (today: `StringKind`). See `genMapMarshal` in `tools/protoc-gen-dethash/main.go`:

```go
if info.pointerContaining {
    g.P("    clear(keys)")
}
g.P("    *keysPtr = keys")
g.P("    ", poolVar, ".Put(keysPtr)")
```

Verified in the generated output, e.g. `internal/proto/commonpb/common_dethash.pb.go`:

```go
clear(keys)
*keysPtr = keys
_dethashKeyPool_common_String.Put(keysPtr)
```

Pool slots no longer retain the previous marshal's key strings. Integer-keyed pools skip the `clear` since their element type holds no pointers.

tools/protoc-gen-dethash/main.go
```diff
@@ -311,14 +503,28 @@ func genMapMarshal(g *protogen.GeneratedFile, f *protogen.Field, hasMap map[stri
...
+	//
+	// For pointer-containing key kinds (today: string headers), zero the
+	// slice elements first. Without this, the backing array keeps the last
+	// marshal's keys reachable until the pool slot is overwritten or
+	// evicted — turning the per-call allocation saved into unbounded heap
+	// retention of user metadata on many-unique-key workloads.
+	if info.pointerContaining {
+		g.P("    clear(keys)")
+	}
+	g.P("    *keysPtr = keys")
```
NumaryBot, line 533: 🟠 **[major] Pooled string-key slices retain map key strings after returning to pool**
<sub>reported by NumaryBot, NumaryBot</sub>

When a map with string keys is processed, the pooled slice's backing array still holds references to all map key strings after the slice is returned to the pool. Because strings are pointer-containing types and the pool is package-global, those key strings remain reachable until the pool discards or overwrites the slot. For large or unique metadata maps (e.g. user-supplied metadata) followed by smaller ones, this turns the zero-alloc optimisation into unbounded memory retention of user data.

**Suggestion:** Zero out string elements before calling pool.Put(), e.g.: `for i := range keys { keys[i] = "" }; pool.Put(keys[:0])`. Alternatively, avoid pooling slices whose element type contains pointers.

tools/protoc-gen-dethash/main.go
```diff
@@ -18,12 +19,99 @@ import (
...
+	// only `string`, whose header points at the underlying bytes). The
+	// generated code must `clear(keys)` before returning the slice to the
+	// pool — otherwise the backing array keeps those pointers reachable
+	// until the pool slot is overwritten or evicted, effectively trading
+	// the per-marshal allocation for unbounded heap retention on
+	// many-unique-key workloads (e.g. churn over metadata maps).
+	pointerContaining bool
+	// sortPkg/sortFunc is the sort statement applied to the keys slice.
+	// Generated as `<package>.<call>(keys)` so the import is tracked.
+	sortPkg  protogen.GoImportPath
```
NumaryBot, line 52: 🟠 **[major] Fixed pool identifier names cause redeclared-symbol compile errors in multi-file packages**

The generator emits package-level identifiers such as `_dethashKeyPoolString` unconditionally per generated file. When two `.proto` files share the same `go_package` and both contain a map with the same key kind, each generated `_dethash.pb.go` file declares the same identifier, producing a compile-time 'redeclared in this block' error. The current repo layout avoids this only by accident (one proto file per package).

**Suggestion:** Emit pool variable declarations once per Go package rather than once per generated file — either via a separate helper file per package, or by making each identifier unique per proto file (e.g., prefixed with the proto file's sanitized base name).

tools/protoc-gen-dethash/main.go
```diff
@@ -18,12 +19,99 @@ import (
...
+// and obvious instead of silently producing invalid Go.
+func filePoolPrefix(file *protogen.File) string {
+	base := path.Base(file.GeneratedFilenamePrefix)
+	if base == "" {
+		panic(fmt.Sprintf("protoc-gen-dethash: empty file base name for %q", file.Desc.Path()))
+	}
+
+	for i, r := range base {
+		isLetter := (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z')
+		isDigit := r >= '0' && r <= '9'
```
NumaryBot: 🟠 **[major] Identifier sanitization collisions for proto files differing only in separator characters**

When two `.proto` files in the same `go_package` have basenames that differ only by separators (e.g., `foo-bar.proto` vs `foo_bar.proto`), the normalization that strips separators produces the same prefix (`FooBar`). Both generated files would then declare the same `_dethashKeyPool<Prefix><Kind>` variable, causing a compile-time 'redeclared in this block' error. The per-file uniqueness fix is therefore incomplete for otherwise valid proto filenames.

**Suggestion:** Use a collision-resistant sanitization strategy: replace separators with a canonical but distinct encoding (e.g., hyphen → `H`, underscore → `U`) rather than stripping them, so `foo-bar` → `FooHBar` and `foo_bar` → `FooUBar`. Alternatively use a short hash of the filename to guarantee uniqueness.

tools/protoc-gen-dethash/main.go
```diff
@@ -18,12 +19,99 @@ import (
...
+		isUnderscore := r == '_'
+		if !isLetter && !isUnderscore && !(isDigit && i > 0) {
```
NumaryBot: 🟠 **[major] Generator panics on valid proto filenames containing separators**

The new `filePoolPrefix` guard panics (or hard-errors) when a proto file uses a valid basename such as `foo-bar.proto` or `api.v1.proto`, even though protoc and protoc-gen-go handle those filenames without issue. Because `filePoolPrefix` is called unconditionally for all generated files (not only those containing maps), this breaks `go generate ./...` for any downstream user whose proto files contain such characters.

**Suggestion:** Replace the panic/reject path with a bijective encoding or a short hash of the filename segment so that all valid proto basenames are accepted and produce unique Go identifiers.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1440#issuecomment-4833742580

gfyrag:
## Closing the bot review loop — analysis and final state

The last two NumaryBot review passes have been re-flagging mitigations already in place, or contradicting earlier findings. Walking through each:

**Round 1 (commit `fdf39bd1`)**
- *"Clear pooled string-key slices before Put"* — fixed: `clear(keys)` is now emitted before every Put for pointer-containing key kinds (today only `StringKind`). Verified in the plugin source (`genMapMarshal`) and in the regenerated `*_dethash.pb.go` output.
- *"Pool identifier names collide across multi-file packages"* — fixed: pool variables now carry the proto file basename, e.g. `_dethashKeyPoolCommonString`, `_dethashKeyPoolRaftCmdString`, `_dethashKeyPoolClusterUint64`. Two `.proto` files sharing a `go_package` no longer declare the same identifier.

**Round 2 (commit `077a6717`, now reverted in `a245d74b`)**
- Round 2 raised *"separator collisions"*: `foo-bar.proto` and `foo_bar.proto` could both normalize to the same `FooBar` prefix. I addressed it by switching to a verbatim-basename + panic-on-unsafe-chars scheme. Reverted in `a245d74b` after re-analysis: none of the 12 proto basenames in this repo contain a non-identifier-safe character (they're all in `[a-z_]`), so the original CamelCase normalization handles them all unambiguously (`raft_cmd` → `RaftCmd`, `cluster_bootstrap` → `ClusterBootstrap`, …). The "collision" scenario requires deliberately adding two `.proto` files differing only by `-` vs `_` to the same `go_package` — a hypothetical-of-a-hypothetical that doesn't justify the added validation surface.

**Round 3 (current)**
- *"Pooled string-key slices retain map key strings after returning to pool"* — re-flag of round 1A, mitigation in place (see above).
- *"Pool identifier names collide across multi-file packages"* — re-flag of round 1B, mitigation in place (see above).
- *"Sanitization collisions for separator chars"* — re-flag of round 2, addressed by the revert: the original normalization is sufficient for every realistic basename in this repo.
- *"Generator panics on valid proto filenames containing separators"* — directly contradicts the previous round's finding. Moot after the revert (no panic any more).

The plugin is internal to this repo; there are no downstream consumers. All current and reasonably-anticipated proto basenames produce unique pool identifiers under CamelCase normalization. Not pursuing further bot-driven hardening on top of that.

Final state on `feat/dethash-zero-alloc-keys`: commits `811a0851` (sync.Pool), `af154226` (slices.Sort), `fdf39bd1` (per-file naming + clear), `a245d74b` (revert of the over-engineered round 2).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1440#issuecomment-4833742580

tools/protoc-gen-dethash/main.go
```diff
@@ -18,12 +20,97 @@ import (
...
+}
+
+// filePoolPrefix derives a Go-identifier-safe, capitalized suffix from a
+// proto file's GeneratedFilenamePrefix (which is the proto file path
+// minus its extension, e.g. "common" or "misc/proto/common"). It is used
+// as a per-file disambiguator in pool variable names. Non-alphanumeric
+// runes are replaced with '_'; the result is title-cased so the final
+// pool identifier reads as `_dethashKeyPool<File><Kind>`.
+func filePoolPrefix(file *protogen.File) string {
+	base := path.Base(file.GeneratedFilenamePrefix)
```
NumaryBot: 🟠 **[major] Pool identifier collisions when multiple proto files share the same sanitized basename**

The generator derives the per-file pool variable prefix from `path.Base` of the proto filename and then strips separator characters (hyphens, underscores, dots). Two distinct proto files can end up with the same prefix in several ways: (1) files from different directories share the same basename (e.g. `foo/common.proto` and `bar/common.proto` both → `Common`); (2) files whose basenames differ only in separator characters (e.g. `foo-bar.proto` and `foo_bar.proto`) both normalize to `FooBar`. When both files belong to the same Go package and both contain a map with the same key kind, each generated `_dethash.pb.go` declares the same package-level symbol (e.g. `_dethashKeyPoolCommonString`), causing a compile-time 'redeclared in this block' error. The current repo avoids this only by coincidence of its directory layout.

**Suggestion:** Derive the prefix from the full `GeneratedFilenamePrefix` (not just `path.Base`), and use a collision-resistant encoding for separator characters (e.g. replace `-` with `H` and `_` with `U`, or use a short hash of the full prefix) so that all valid proto filenames produce unique, legal Go identifiers without risk of collision.

tools/protoc-gen-dethash/main.go
```diff
@@ -18,12 +20,97 @@ import (
...
+				r -= 'a' - 'A'
+				upper = false
+			}
+			b.WriteRune(r)
+		case r >= 'A' && r <= 'Z' || r >= '0' && r <= '9':
+			b.WriteRune(r)
+			upper = false
+		default:
+			// Non-identifier rune (e.g. '-' in a hypothetical "my-proto"
+			// file). Drop and re-capitalize the next letter so we don't
```
NumaryBot, line 113: 🟠 **[major] Generator panics on valid proto filenames containing separator characters**

The `filePoolPrefix` function hard-errors or panics when a proto file uses a valid basename containing hyphens, underscores, or dots beyond the extension (e.g. `foo-bar.proto`, `api.v1.proto`). Because this function is called unconditionally for every generated file — not only those containing maps — it breaks `go generate ./...` for any downstream user whose proto files contain such characters, even though protoc and protoc-gen-go handle them without issue.

**Suggestion:** Replace the panic/reject path with a bijective encoding (e.g. substitute each separator character with a distinct letter sequence) or a short hash of the filename segment so that all valid proto basenames are accepted and produce unique Go identifiers.

tools/protoc-gen-dethash/main.go
paul-nicolas: [Medium] Pool-name collision avoidance is incomplete: `path.Base` collapses directories

`filePoolPrefix` derives the per-file disambiguator from `path.Base(file.GeneratedFilenamePrefix)`, dropping all leading path segments. Two `.proto` files in different directories that compile into the same `go_package` and share a basename (e.g. `a/common.proto` and `b/common.proto`) would both emit `_dethashKeyPoolCommon<Kind>` into the same package and fail to compile with "redeclared in this block" -- which is exactly the failure mode the per-file prefix was added to prevent (see the doc comment on `poolVarName`).

Not reachable in this repo today: every `misc/proto/*.proto` maps to a distinct `go_package` and all basenames are unique, so this is latent rather than active. But since this is newly-added collision-avoidance logic, it's worth closing the gap so a future proto layout can't silently break the build.

Fix: build the suffix from the full sanitized `GeneratedFilenamePrefix` (all path segments, not just the base), or append a short stable hash of it. The existing rune-sanitizing loop already handles non-identifier characters, so feeding it the full path is a minimal change.

paul-nicolas (commented):

Models: Claude (this repo's perspective) + Codex (external). Gemini intentionally not run.

The substance of this PR is `tools/protoc-gen-dethash/main.go`; the six `*_dethash.pb.go` files are its regenerated output and were spot-checked for the Get / fill / sort / (clear) / Put pattern. The change is sound: the pooling preserves determinism, the recursion is safe, and the steady-state allocation goal is met.

**Severity counts:** 1 Medium (posted inline) · 2 Minor (listed below, not posted inline). No Critical/Major.

**Verification outcome** — 3 candidate Critical/Major/Medium findings entered adversarial verification; 1 survived:
- _rejected_ **[Codex, major]** `clear(keys)` leaves stale string headers in `[len:cap]`. Empirically disproven: the marshal that grows the buffer calls `clear` over its own full length before `Put`, so by induction the backing array is fully zeroed whenever it sits in the pool. A runnable reproduction (10k-key marshal followed by a 10-key marshal) showed 0 non-empty slots across the full `cap`. The per-marshal `clear(keys)` is sufficient.
- _rejected_ **[Claude hypothesis]** re-entrancy corruption on nested same-kind maps (e.g. `PostCommitVolumes` -> `VolumesByAssets`, both `_dethashKeyPoolCommonString`). `sync.Pool.Get()` removes the object from the pool, so a nested `Get` while the outer slice is still checked out can never alias it; Get/Put nest as a stack. Safe.
- _confirmed_ **[Codex, medium]** incomplete pool-name collision avoidance in `filePoolPrefix` (posted inline).

**Minor / Nit (not posted inline):**
- `tools/protoc-gen-dethash/main.go:437` -- No regression test added for the new pooling semantics. Determinism is only covered indirectly (the audit hash-chain tests in `internal/infra/state` and `internal/application/check` would fail on corrupted output). A generator-level or marshal-level test that (a) marshals the same map-bearing message twice and asserts byte-identical output, (b) exercises a nested same-kind map, and (c) runs concurrently under `-race`, would pin the pool-reuse contract directly. (The `marshal_bench_test.go` referenced in the description lives in PR #574, not here.)
- `tools/protoc-gen-dethash/main.go:523` -- The `clear(keys)` / no-clear split between pointer-containing (string) and integer key kinds is correct and well-documented; noting only that it is load-bearing for heap retention and is the kind of thing a test (above) should lock in.

tools/protoc-gen-dethash/main.go
gfyrag: Good catch — applied in `6111b2d2`. `filePoolPrefix` now feeds the full `GeneratedFilenamePrefix` through the rune-sanitizing loop instead of `path.Base`-ing it first, so directory segments end up CamelCased into the identifier (`/` is dropped by the same default branch as any other non-identifier rune). Two `.proto` files in different source dirs that declare the same `go_package` would now produce distinct pool variables (`ACommon` vs `BCommon` in your example), closing the latent gap.

Trade-off worth flagging: identifiers grow noticeably (e.g. `_dethashKeyPoolGithubComFormancehqLedgerV3InternalProtoCommonpbCommonString`) because `GeneratedFilenamePrefix` includes the Go import path. Generated code only, never typed by hand, but it does bloat each `*_dethash.pb.go` by a few KB. Happy to switch to a 6-char fnv hash if you'd rather optimize for compact identifiers; the path-verbatim version felt like the minimal change matching your suggestion.

gfyrag: Scenario (a) (directory-collapse via `path.Base`) is fixed in `6111b2d2` — `filePoolPrefix` now consumes the full `GeneratedFilenamePrefix`, so `a/common.proto` and `b/common.proto` produce distinct pool identifiers.

Scenario (b) (`foo-bar.proto` vs `foo_bar.proto` in the same `go_package`) is left latent — no current proto basename has a non-identifier char and the collision is contingent on a very specific naming choice. If it ever happens we'll catch it at build time as `redeclared in this block` and add an encoding step then. Not adding a bijective scheme preemptively.

tools/protoc-gen-dethash/main.go
gfyrag: Stale — the panic was introduced in `077a6717` and removed in the revert `a245d74b`. The current code sanitizes non-identifier runes via the CamelCase loop, no panic on filenames like `foo-bar.proto` or `api.v1.proto`.

tools/protoc-gen-dethash/main.go
```diff
@@ -18,12 +19,106 @@ import (
...
+
+// filePoolPrefix derives a Go-identifier-safe, capitalized suffix from a
+// proto file's GeneratedFilenamePrefix (the proto file path minus its
+// extension, e.g. "common" with `protoc -I misc/proto` or
+// "misc/proto/common" with `protoc -I .`). The full prefix is consumed —
+// not just the basename — so two .proto files that happen to share a
+// basename in different directories (e.g. `a/common.proto` and
+// `b/common.proto`) but compile into the same go_package still produce
+// distinct pool identifiers (`ACommon` vs `BCommon`). Stripping to
+// `path.Base` first would re-introduce the per-file collision the
```
NumaryBot, line 87: 🟡 **[minor] filePoolPrefix uses path.Base, leaving a latent collision risk for same-basename protos in different directories**
<sub>reported by codex, paul-nicolas</sub>

The pool variable disambiguator is derived from path.Base(file.GeneratedFilenamePrefix), dropping all leading path segments. If two .proto files from different directories are ever compiled into the same go_package and share a basename (e.g. a/common.proto and b/common.proto), both generated files would emit the same package-level symbol (e.g. _dethashKeyPoolCommonString), producing a 'redeclared in this block' compile error. This is not reachable in the current repository layout, but the collision-avoidance logic is incomplete for future proto layouts.

**Suggestion:** Build the prefix from the full sanitized GeneratedFilenamePrefix (all path segments, not just the base), replacing path separators with a safe delimiter. The existing rune-sanitizing loop already handles non-identifier characters, so feeding it the full prefix path is a minimal change.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1440#issuecomment-4833742580

tools/protoc-gen-dethash/main.go
gfyrag: Stale — this is reviewing commit `6111b2d2`, which is exactly the commit that removed the `path.Base` call this finding flags. Current `filePoolPrefix` (line 95) reads `prefix := file.GeneratedFilenamePrefix` and feeds the full path to the existing CamelCase loop. The only remaining mentions of `path.Base` in the file are doc-comment text explaining why it is no longer used.

paul-nicolas (approved):

PR #1441, opened 2026-06-29
feat(fsm-digest): deterministic encoding + cross-node digest foundation

gfyrag:
## Summary

Opt-in cluster-wide foundation for FSM determinism + a rolling per-batch cross-node digest, intended as an observability/diagnostic safety net for the early production deployments. Two nodes at the same applied Raft index can now exchange a digest and detect any FSM divergence (deterministic bug, silent corruption, cache vs Pebble drift).

This is **PR 1 of N** — the foundation. Health-check polling, checker integration, the CLI tool, the Antithesis assertion, and the E2E suite are tracked as follow-up PRs (see below).

## Why now

The FSM **must** be deterministic by invariant, but nothing today *verifies* in live that two nodes at the same applied index actually persist the same KV state. This PR closes that gap behind an opt-in flag so we can light it up for the first prod deployments as a safety net, then keep or disable per cluster.

## How it works

A single immutable cluster flag — `--fsm-determinism-enabled` — controls everything:

- **OFF (default)**: zero overhead, behaviour unchanged.
- **ON**: every persisted byte routes through `MarshalDeterministicVT`, and a rolling XXH3/BLAKE3 chain is maintained under `SubGlobFSMDigest`. Cross-node digests can be compared via the new `GetFSMDigest` RPC.

The flag is set once at first boot, persisted under `PersistedConfig.fsm_determinism_enabled`, validated on every subsequent boot. Mismatch is **fatal** and **NOT bypassable** by `--unsafe-skip-config-validation` — flipping it would either re-encode existing entries non-deterministically (ON→OFF) or compare new deterministic entries against pre-existing non-deterministic ones (OFF→ON).

### Volet 1 — deterministic encoding (conditional on the flag)

- `dal.WriteSession.MarshalProto` branches on the parent `Store.DeterministicEncoding()` — single switch, owned by the primitive.
- `attributes.marshalProto` mirrors the same branch (covers ZoneAttributes 0x01 and ZoneCache 0xFF, which reuses the same bytes).
- `idempotency_store` + `backup_jobs` no longer bypass `WriteSession` with `MarshalVT + SetBytes` — they go through `SetProto` so the flag controls them too.
- `AuditEntry` stays unconditionally deterministic (audit chain breaks otherwise — same invariant as before).
- Producer-side audit (`TouchedVolume.Volumes`, `AuditEntry.Ledgers`) was already canonical (sorted at the producer); no producer changes needed.

### Volet 2 — rolling digest

```
digest_n = H_seed( ""fsm-digest:v1"" ||
                   u64(snapshotIndex) || u64(appliedIndex) ||
                   canonical_batch_bytes ||
                   digest_{n-1} )
```

- Computed inside `CommitPreparedBatch` **before** `pb.batch.Commit()` — the digest write is part of the same Pebble batch, atomic with the writes it summarises.
- `canonicalBatchPayload` iterates the in-memory `pebble.Batch` via `batchrepr.Read`, sorts ops by `(key, kind)`, then streams `kind || varint(keyLen) || key || varint(valueLen) || value`. Different insertion orders on different nodes produce identical canonical bytes.
- Persisted under `[ZoneGlobal][SubGlobFSMDigest]`: `u64(appliedIndex) || u64(snapshotIndex) || digest`. Single key, overwritten each commit.
- `FSMState.LastFSMDigest` is updated under `appliedMu` **before** `publishApplied` so concurrent `GetFSMDigest` readers always observe a coherent `(appliedIndex, digest)` snapshot.
- The `snapshotIndex` term re-anchors the chain after every snapshot install — a follower restored from a leader checkpoint at index N reads `digest_N` directly from the restored Pebble (the digest lives in `ZoneGlobal`, copied byte-for-byte in the Pebble checkpoint). No manifest change required.

### gRPC

```proto
rpc GetFSMDigest(GetFSMDigestRequest) returns (FSMDigest);
message GetFSMDigestRequest { fixed64 index = 1; uint64 wait_ms = 2; }
message FSMDigest {
  fixed64 applied_index = 1;
  fixed64 snapshot_index = 2;
  bytes  digest = 3;
  uint32 node_id = 4;
  uint32 hash_version = 5;
}
```

- Optional `index` pins the target applied index; the server waits up to `wait_ms` (`DEADLINE_EXCEEDED` on miss).
- `FAILED_PRECONDITION` when the flag is OFF — clients treat as ""feature disabled for this cluster"".
- **Node-local read** — never forwarded to the leader (the whole point is to compare per-node snapshots).

## Audit-chain invariant 8

The audit hash chain **remains** the only source of cryptographic authenticity. The digest is a **diagnostic** signal — it detects divergence but does not authenticate. Documented in the code comments and in the commit body.

## What's NOT in this PR (follow-ups)

Tracked separately to keep this PR review-able:

1. **Health-check peer poll + statusOutOfSync transition** — leader-driven gossip of digests over the existing `HealthChecker.check()` cycle.
2. **Checker `compareFSMDigest` pass** — full-scan recompute from a snapshot, cross-checked against `SubGlobFSMDigest` (emits `CHECK_STORE_ERROR_TYPE_FSM_DIGEST_MISMATCH`).
3. **CLI `ledgerctl cluster digest`** — query per-node digests, print comparison table, `--full-scan` fallback.
4. **Antithesis assertion `fsm_digest_matches_peers`** — fan-out via `PerNodeConn`, `assert.Always(allDigestsEqual, ...)`.
5. **3-node E2E + smoke benchmarks** — including a single-byte corruption scenario to validate the detection path.

## Test plan

- [x] Unit: `internal/storage/dal/batch_determinism_test.go` — map-insertion-order independence of `MarshalProto` when flag is ON. Runs with `-count=20` to catch Go map randomization.
- [x] Unit: `internal/infra/state/digest_test.go` — canonical-payload order independence, chain input sensitivity (each component must change the digest), per-cluster keying, value layout, `IterateBufferedOps` semantics.
- [x] Unit: `internal/bootstrap/config_validation_test.go` — all four flag transitions (both-off / both-on / off-to-on / on-to-off) and verifies `--unsafe-skip-config-validation` does NOT bypass the mismatch.
- [x] Full `go test ./...` green.
- [x] Pre-commit clean (`go generate`, `go mod tidy`, `golangci-lint --fix`).
- [ ] **Benchmark** the FSM hot path with flag ON vs OFF to characterise the real cost of the deterministic marshal + canonical sort + hash chain. Targets: `BenchmarkAudit*` in `internal/infra/state/`, plus the encoding benches in `internal/storage/dal/`. To bench, set `cfg.DeterministicEncoding = true` in the bench harness.

## Notes

Per the prior memory `feedback_audit_is_source_of_truth`: this PR explicitly does NOT extend the audit hash chain to cover the rest of the persisted state. The digest is an orthogonal, runtime-derivable signal — the audit chain stays the only cryptographic boundary. Any future change that persists a new projection should still go through the checker (`internal/application/check/checker.go`).

---
_Migrated from formancehq/ledger-v3-poc#574 on 2026-06-29._


gfyrag (commented):
Review du nouveau commit de merge/rebase `1c12353657912036502127d7195b3ec2d0dcb130`: pas de finding solide.

J'ai concentré la vérification sur les fichiers annoncés comme conflictuels : `internal/infra/state/machine.go`, `internal/storage/dal/store.go`, les `*_dethash.pb.go`, `tools/protoc-gen-dethash/main.go`, `tests/e2e/testutil/server.go` et le deepcopy operator généré.

Points vérifiés :
- pas de marqueurs de conflit restants dans les fichiers conflictuels ;
- `Machine` ne gagne pas de capacité de lecture Pebble dans l'apply path ;
- les `ConfChange` sont écrits via la `WriteSession` FSM en cours, avec signalement d'erreur si l'écriture échoue ;
- le rolling digest avance toujours une fois par entrée Raft, y compris conf-change / payload vide / no-op proposal ;
- le générateur `dethash` évite les collisions de pools entre fichiers d'un même package Go et nettoie les clés string avant retour au pool ;
- l'ajout `WithNodeInstruments` dans les helpers e2e compose bien les instruments par nœud avec les instruments communs.

Je n'ai pas lancé de `go test` sur ce head : le worktree local `worktrees/fsm-digest` n'est pas à ce commit et contient déjà un binaire non suivi, donc j'ai gardé la review sur inspection statique des objets Git pour ne pas perturber le travail d'agent existant.

gfyrag (commented):
Reviewed new head `8618c18c1486b9652350cf82add5bd8dccc1589f` after the rebase/lint fix.

No findings. The delta since the previously reviewed head only changes `internal/storage/dal/batch_determinism_test.go`, replacing the manual reverse loop with `slices.Backward(keys)`. Iteration order remains reverse insertion order for the metadata map setup, so the deterministic-encoding test still exercises the intended case.

Validation run:
- `GOCACHE=/tmp/go-build-pr1441 go test ./internal/storage/dal`
- `git diff --check 1c12353657912036502127d7195b3ec2d0dcb130..HEAD`

Both passed.

gfyrag (commented):
Reviewed the latest draft head `a651d883f9d70e37121d78ff8da81fc40eb19cfa` after the rebase/merge.

No blocking findings. I focused on the rolling FSM digest mechanics: deterministic encoding flag persistence, digest session wiring, one digest advance per applied Raft entry, exclusion of node-local/global zones, and the read-only `GetFSMDigest` comparison path. The implementation keeps Pebble reads out of the hot apply path; the session seeds from the store's in-memory rolling digest and persists the digest atomically at commit.

Checked:
- `go test ./internal/storage/dal ./internal/infra/state ./internal/adapter/grpc`
- `git diff --check origin/release/v3.0...origin/pr/1441`

internal/storage/dal/batch.go
```diff
@@ -70,19 +248,60 @@ func (b *WriteSession) MarshalProto(msg proto.Message) ([]byte, error) {
...
+// instrumentation, no per-entry buffer allocation). The FSM can always
+// call OpenFSMWriteSession unconditionally and let this short-circuit
+// handle the disabled path.
+func (s *Store) OpenFSMWriteSession(chain FSMDigestChain) *WriteSession {
+	sess := s.OpenWriteSession()
+	if !s.deterministicEncoding || chain == nil {
+		return sess
+	}
+
+	_, hash := s.RollingDigest()
```
flemzord: [P1] Seed the next session from the pending digest chain

This can drop the previous batch from the digest whenever the normal applier pipeline prepares N+1 while N is still committing. `applyEntriesPipelined` calls `PrepareDecodedEntries` before `waitPendingCommit`, but this reads the Store cache that `CommitWithRollingDigest` updates only after N commits, so N+1 can capture H(N-1) instead of H(N). A temporary regression test opening two sessions before committing the first produced a different final hash than sequential preparation. Carry the prepared digest seed across batches or wait for the prior commit before opening the next digest session, and add the pipelined regression test.

internal/storage/dal/batch.go
```diff
@@ -120,8 +348,129 @@ func (b *WriteSession) Commit() error {
...
+// transient (ClusterTransient).
+//
+// The encoding is canonical (uvarint lengths) so the bytes mixed in are
+// independent of any Pebble batch framing quirks.
+func (b *WriteSession) mixOp(kind byte, key, value []byte) {
+	if b.digestChain == nil || !isHashedZone(key) {
+		return
+	}
+
+	b.entryOps = append(b.entryOps, kind)
```
flemzord, line 515: [P1] Canonicalize operation ordering before hashing

This can report divergence for byte-identical final state because `entryOps` preserves call order. `DerivedKeyStore.Merge` drains native Go maps and the resulting Attribute updates are written in that randomized order, so peers can mix identical distinct-key writes in different orders. A temporary test applying the same two Attribute sets in opposite order produced different hashes. Make producer drains deterministic (at minimum sort `DerivedKeyStore.Merge` updates/deletions by canonical key) or canonicalize the buffered operations before advancing the digest, and cover opposite write ordering.

flemzord (commented):
Validated two deterministic-digest failures on the current head with focused regression tests in a temporary extraction.

gfyrag:
Review pass on `a651d883f9d70e37121d78ff8da81fc40eb19cfa` after the latest comments.

The two digest findings are valid and match the failing E2E cross-node digest check:

1. P1: stale digest seed in pipelined apply. `PrepareDecodedEntries` seeds from `Store.RollingDigest()` when the session opens, but the applier can prepare batch N+1 before waiting for batch N to commit. If N+1 prepares while N is committing, it can seed from H(N-1). Carry the pending digest seed across prepared batches, or wait before opening the next FSM session.

2. P1: non-canonical operation ordering. Attribute updates/deletions are drained from Go maps, then written in that order, and the digest hashes call order. Equal final state can therefore produce different digests. Sort updates/deletions by canonical key before flushing, or canonicalize `entryOps` before `AdvanceDigest`.

I did not find an additional RPC issue: `WaitForApplied` is backed by `lastPersistedIndex`.

Validation run: `go test ./internal/storage/dal ./internal/infra/state ./internal/infra/node` and `git diff --check origin/release/v3.0...HEAD` passed locally. Remote E2E is failing on cross-node digest mismatch at applied index 18.

gfyrag:
Re-reviewed current head `a651d883f9d70e37121d78ff8da81fc40eb19cfa`.

Remaining findings:

1. P1: stale digest seed with pipelined apply.

A new FSM write session is seeded from the store’s committed rolling digest, but the applier prepares batch N+1 before waiting for batch N’s commit. N+1 can therefore chain from H(N-1), dropping N from the digest. This matches the unresolved review thread and the reported remote E2E digest mismatch.

2. P1: digest input is write-call-order dependent.

`mixOp` appends ops in call order, while several producers drain Go maps. Equal final state can hash differently across nodes. Please canonicalize the per-entry op buffer before advancing the digest, or sort every producer drain, with regression coverage.

3. P2: `GetFSMDigestRequest.index` does not pin the returned digest to that index.

`WaitForApplied` only waits for `>= target`, and the digest is a single overwritten key. If a node advances past N before the read, `GetFSMDigest` returns the latest digest, not digest(N). That is workable only if clients compare returned `applied_index` and retry/quiesce; it does not support exact live comparison at a requested index as the proto implies.

4. P2: snapshot behavior is not proven by the implementation or tests.

`FSMDigest.snapshot_index` is documented in proto, but the server response never fills it, and the persisted record is only `[appliedIndex][digest]`. The E2E assertion comparing snapshot indexes currently only compares zero to zero.

5. P2/docs-design gap: the digest scope is narrower than the API wording.

The implementation hashes only Attributes, Cold, and Idempotency, while the proto says “every persisted byte.” Either document this as a scoped projection digest or include deterministic replicated Global/PerLedger state deliberately.

Docs/ops gap:

This PR adds a new technical mechanism and a server flag, but no `docs/` files changed. Per the repo instructions, this needs architecture documentation and `docs/ops/cli.md`/flag documentation.

Checks run:

- `GOCACHE=/tmp/go-build-pr1441-review GOROOT= go test ./internal/storage/dal ./internal/infra/state ./internal/infra/node ./internal/adapter/grpc`
- `git diff --check 8845ad587c2f90d81873a4f4fc0022e6d2fe38d7...HEAD`
- From nested module: `GOCACHE=/tmp/go-build-pr1441-review GOROOT= go test ./bin/cmds/main/eventually_fsm_digest_matches`

Not run: full pre-commit or E2E locally.

internal/storage/dal/store.go
```diff
@@ -1399,6 +1489,19 @@ func (s *Store) RestoreCheckpoint(checkpointID uint64) error {
...
 	s.oldestCheckpoint = checkpointID
 
+	// Re-seed the in-memory rolling FSM digest from the freshly published
+	// live database — the checkpoint we just restored carries the leader's
+	// digest at the checkpoint's applied index, which is now ours too.
+	// Failure here is non-fatal: the FSM hot path will read a stale cached
+	// value once and surface it as a digest mismatch on the next compare,
+	// which is the right signal to investigate. We log loudly so the
+	// failure is not silent.
+	if err := s.reloadRollingDigest(); err != nil {
```
flemzord, line 1560: [P1] Fail the restore when the checkpoint digest cannot be loaded

After the leader checkpoint has been published, this branch leaves `rollingDigest` holding the pre-restore local hash but still returns success. `SynchronizeWithLeader` then resumes the follower, and its next FSM session chains from that stale hash and overwrites the restored digest, guaranteeing a false cross-node divergence rather than merely surfacing one transient diagnostic. Please propagate this error (or otherwise prevent apply from resuming until the restored digest has been loaded and validated), and cover a checkpoint with an unreadable/malformed digest record.

internal/bootstrap/config_validation.go
```diff
@@ -107,6 +126,17 @@ func ValidateOrPersistConfig(store *dal.Store, cfg Config, logger logging.Logger
 		}
 	}
 
+	// FSM determinism check — never bypassable, even with --unsafe-skip-config-validation.
+	// Flipping the flag between boots would either re-encode existing entries
+	// non-deterministically (ON→OFF) or compare new deterministic entries against
+	// pre-existing non-deterministic ones (OFF→ON), tripping the cross-node digest.
+	if persisted.GetFsmDeterminismEnabled() != current.GetFsmDeterminismEnabled() {
```
flemzord, line 133: 🟡 **[P2] Enforce this flag across peers, not only local restarts**

This only compares the runtime value with this node's locally persisted config. On first boot, two fresh nodes with the same cluster ID can each persist opposite values and both join successfully, leaving one node producing deterministic bytes/digests while the other has the feature disabled. The Antithesis oracle then treats the first peer's `FAILED_PRECONDITION` as “feature disabled for this cluster” and exits, masking the mixed configuration. Please put the flag in a Raft-replicated/bootstrap agreement (or reject joins against the leader's value) and add a heterogeneous-first-boot test.

flemzord (commented):
One new validated cluster-configuration finding on the unchanged draft head.

flemzord (commented):
Found one additional configuration-safety gap in the portable restore path.

misc/proto/common.proto
```diff
@@ -1463,6 +1463,18 @@ message PersistedConfig {
   string cluster_id = 2;
   uint64 idempotency_ttl_seconds = 3;
   uint32 storage_schema_version = 4;
+  // fsm_determinism_enabled opts the cluster into deterministic FSM byte
+  // encoding and the cross-node digest health-check. Set once at first
+  // bootstrap, validated on every subsequent boot. A mismatch is fatal and
```
flemzord: [P2] This immutability guarantee is bypassed by a portable backup restore. `attributes.PrepareForBackup` deletes the whole `SubGlobPersistedConfig` while preserving the existing Attribute/Cold/Idempotency bytes (and this PR's `SubGlobFSMDigest` record), so the destination's `ValidateOrPersistConfig` sees a first boot and accepts either value for this flag. A backup produced with determinism OFF can therefore be restored with it ON, or vice versa, despite the existing data having been written under the opposite encoding contract. Please preserve the encoding provenance across portable restores, or explicitly validate/migrate/reset the affected state before accepting the destination setting, and cover both restore-direction transitions.

internal/storage/dal/batch.go
gfyrag: Fixed in 5f7b5a6e0. Confirmed the bug: with pipelining `PrepareDecodedEntries(N+1)` opens its WriteSession before `CommitPreparedBatch(N)` runs, and `OpenFSMWriteSession` seeded from the *committed* rolling digest — so batch N+1 chained from the pre-N hash and batch N was dropped from the chain at the boundary.

Fix: the store now carries a PENDING rolling digest advanced at prepare time. `WriteSession.RecordPendingDigest` (called at the end of `PrepareDecodedEntries`, after the per-entry `AdvanceDigest` calls) publishes the batch's fully-advanced hash to the pending slot, and `OpenFSMWriteSession` seeds from `RollingDigestPending()`. `SetRollingDigest` keeps pending >= committed so non-pipelined / lifecycle commits still advance the seed; boot/restore reset pending to the committed baseline. Regression: `TestFSMDigest_PendingSeedCarriesAcrossPipeline` drives the exact pipeline ordering (open batch 2 before batch 1 commits) and requires the final digest to incorporate both batches.

internal/storage/dal/batch.go
gfyrag: Fixed in 5f7b5a6e0. Confirmed: `DerivedKeyStore.Merge` drains its overlay by ranging Go maps (`s.values` / `s.deletions`), whose iteration order is randomised per process, and those ops flow straight into `entryOps` in arrival order via `mixOp`. Two nodes reaching byte-identical Pebble state (Set/Delete on distinct keys is order-independent) emitted their op records in different orders and hashed differently -> false divergence.

Fix: `mixOp` / `mixDeleteRange` now record each op's `[start,end)` bounds, and `AdvanceDigest` sorts the records by their raw bytes before folding into the chain. This yields a total order on `(kind,key,value)` — unique per op since keys within an entry are distinct (each hashed key is written at most once per entry: a volume is partitioned into exactly one of kept/purged/transient, Cold/Idempotency keys are unique per sequence/proposal). Regression: `TestFSMDigest_OpOrderCanonicalised` feeds the same three writes in forward / reverse / shuffled order and requires an identical digest.

internal/storage/dal/store.go
gfyrag: Fixed in 5f7b5a6e0. The branch now fails loudly per invariant #7: `reloadRollingDigest` failure after the restored live DB is published returns `fmt.Errorf("reloading rolling FSM digest after checkpoint restore: %w", err)`, which propagates up through `RestoreCheckpoint` -> `capabilities.go` and aborts the restore. The live directory was just reopened one step above, so reading its `SubGlobFSMDigest` key cannot fail under any expected runtime condition — continuing would have left `rollingDigest` (and now `rollingDigestPending`) pinned to the pre-restore value, silently baselining the chain wrong with no way to distinguish the resulting false divergence from a real one.

internal/bootstrap/config_validation.go
gfyrag: Addressed in 378b3d937 with a full cross-peer mechanism (not just a best-effort check). The join handshake is the one chokepoint every node passes through, so I enforce it there — the same place EN-1045 (removed-member blacklist) and EN-1080 (auth) fail-fast.

`JoinAsLearnerRequest` gains field 5 `fsm_determinism_enabled`; the joining node sends its local flag, and the leader (`ClusterBootstrapServiceServerImpl.JoinAsLearner`, after the leader-forward branch so the authoritative flag is used) refuses with `FailedPrecondition` when it disagrees with the leader's own flag. The local persisted-config check you flagged stays as the same-node-restart guard; this adds the divergent-peer guard. A node built before the field sends the zero value (false); if the leader runs with the flag ON that mismatch is correctly surfaced. No residual gap for the join path; the only case not covered is an initial multi-node bootstrap where two seed nodes come up with different flags and neither has joined the other — that is caught downstream by the digest compare itself (they never agree), and there is no join RPC to intercept in that topology.

misc/proto/common.proto
gfyrag: Fixed in 3425f5f3d (comment-only). Confirmed `attributes.PrepareForBackup` deletes the whole `PersistedConfig` record (`DeleteKey({ZoneGlobal, SubGlobPersistedConfig})`), including `fsm_determinism_enabled`, so the guarantee does not survive a portable restore — the restoring node re-persists its own config on first boot.

The `PrepareForBackup` behaviour is correct (portability requires the reset — it also zeroes lastAppliedIndex and drops bloom/peers), so I corrected the proto comment rather than the code: the immutability now reads as holding only within a single cluster's data lifetime, not across a portable backup/restore, where the restoring node's flag governs the fresh baseline.

internal/bootstrap/config_validation.go
flemzord: The join check closes the learner path, but it does not make the flag cluster-wide for the static bootstrap path. `NewNode` accepts `--bootstrap` with `cfg.Peers` directly as initial voters, and none of those nodes calls `JoinAsLearner`, so seed nodes can still persist opposite values. The claimed downstream detection also does not hold: `eventually_fsm_digest_matches` sets `featureDisabled` on the first `FAILED_PRECONDITION` and returns without asserting, so one OFF peer masks the mixed configuration instead of reporting it. Please enforce the value during static/bootstrap membership agreement (or in replicated cluster state), make mixed digest availability fail the oracle, and add the heterogeneous bootstrap regression requested here.

gfyrag: Gap closed on both fronts (commits d584d99e1 + 9a21019ff).

STATIC-BOOTSTRAP ENFORCEMENT (d584d99e1): I enforce the flag at the one point every peer pair passes through regardless of bootstrap-vs-join — the Raft transport stream. Each node advertises its fsm-determinism-enabled flag via a new `MetadataKeyFSMDeterminism` header on every outgoing stream, and `DefaultTransport.StreamMessages` rejects an incoming stream with `FailedPrecondition` when the peer disagrees with the local flag (right after the existing cluster-id gate). Two `--bootstrap` seeds with divergent flags now fail to establish a Raft stream to each other, so the cluster cannot form — a genuine fail-fast, not just runtime detection.

I deliberately did NOT put the flag into the replicated ClusterConfig (your option a). fsm_determinism_enabled governs the byte-encoding/hashing of the very FSM writes that ClusterConfig itself is applied through, so a mismatched seed would already be applying ClusterConfig divergently before any config reconciliation could reject it — chicken-and-egg. It also must stay immutable/persisted-config-validated, unlike the mutable ClusterConfig. The stream handshake is the correct layer, and it is the same category as the cluster-id check that already lives there. No new replicated-config field, no migration implications.

An absent header (peer built before this key) is treated as false, so a uniformly-old all-OFF cluster still connects while a mixed one is refused.

ORACLE FIX (9a21019ff): `eventually_fsm_digest_matches` set `featureDisabled` on the FIRST FAILED_PRECONDITION and returned without asserting, so a single OFF peer masked a mixed config. It now tracks per-node outcomes and `assert.Always` that featureEnabled and featureOff are never both non-empty — a mixed cluster fails loudly.

REGRESSION TESTS: `TestStreamMessages_RejectsMismatchedFSMDeterminism` / `_AcceptsMatchingFSMDeterminism` drive the real StreamMessages prologue with mismatched flags (local ON/peer OFF, local OFF/peer ON, peer-absent-treated-as-OFF) and assert FailedPrecondition, plus the matching-flag transparency cases. This is the heterogeneous-flag regression at the transport layer — the layer that actually catches both static bootstrap and join. I kept it as a fast unit test rather than a full multi-`--bootstrap` e2e cluster (the e2e harness bootstraps node 0 then JOINs the rest, so it exercises the join path, not static bootstrap; a true multi-bootstrap heterogeneous e2e would need new harness plumbing for negligible extra coverage over the transport-level test). Happy to add the e2e variant if you want it despite that.

misc/proto/common.proto
flemzord: The comment change does not re-baseline the bytes this flag governs. `PrepareForBackup` resets `lastAppliedIndex` and deletes persisted config/bloom/peers, but deliberately leaves Attributes, Cold, Idempotency, and `SubGlobFSMDigest` byte-for-byte intact. Restoring OFF→ON therefore retains pre-existing non-deterministic encodings without rewriting them and starts/continues a digest that did not cover that retained baseline; the new comment's claim that the whole store is freshly baselined is not implemented. Please preserve portable encoding provenance, or perform an explicit deterministic rewrite plus digest reset/re-anchor, and cover both restore directions rather than only weakening the immutability contract.

gfyrag:
Re-reviewed current head `9a21019ff52f6c427690176dbcb75198559a4436`.

The latest delta fixes the prior mixed static-bootstrap `fsm-determinism-enabled` issue:

- The transport now carries the local flag from bootstrap into `DefaultTransport`, sends it on every Raft stream, and rejects mismatches with `FailedPrecondition`.
- The Antithesis oracle now distinguishes feature-on nodes from feature-off nodes and fails on mixed configuration instead of returning on the first `FAILED_PRECONDITION`.
- Regression coverage was added for matching, mismatching, and absent metadata cases.

Still open and unchanged by this delta:

1. P1: portable restore flag provenance / retained-bytes digest re-baselining.

The comment says portable restore “re-baselines the whole store,” but `PrepareForBackup` leaves attribute bytes intact and only deletes selected Global keys. It also leaves the FSM digest unless separately handled elsewhere. A restore OFF→ON or ON→OFF can therefore keep old encoded bytes/digest while persisting the destination flag as first boot.

2. P2: `GetFSMDigest(index=N)` exactness.

The API still waits for at least `N` but serves the latest single overwritten digest, so a node that advances past N can return digest N+k.

3. P2: `snapshot_index` behavior remains unimplemented.

The response field still is not populated, and current tests compare zero to zero.

4. P2: digest scope vs proto wording.

The proto still says the digest hashes “every persisted byte,” but the implementation hashes only selected replicated zones. Either narrow the API wording or include the omitted replicated zones deliberately.

5. P2: docs/ops coverage is still missing for the new flag and technical mechanism.

The PR adds `--fsm-determinism-enabled`; per repo instructions, this needs `docs/ops/cli.md` and a technical architecture page/README link.

Checks run for the latest delta:

- `GOCACHE=/tmp/go-build-pr1441-review GOROOT= go test ./internal/infra/node`
- From `tests/antithesis/workload`: `GOCACHE=/tmp/go-build-pr1441-review GOROOT= go test ./bin/cmds/main/eventually_fsm_digest_matches`
- `git diff --check 3425f5f3d069decffee2534b1640778885b65fee..9a21019ff52f6c427690176dbcb75198559a4436`
- `GOCACHE=/tmp/go-build-pr1441-review GOROOT= go build ./...`

gfyrag:
Closing this draft for now. The cross-node FSM-digest work surfaced two design questions that need dedicated design before it's mergeable: (1) the portable-restore OFF→ON transition — PrepareForBackup leaves Attributes/Cold/Idempotency bytes non-deterministically encoded while the config guard is bypassed (restore looks like first-boot), so enabling determinism over a restored OFF store baselines the digest over un-rewritten bytes; and (2) the digest baseline reset on restore. The review findings addressed so far (deterministic op ordering, pending-digest seeding, fail-loud on unloadable checkpoint digest, cross-peer flag via the transport handshake, oracle mixed-config detection) are preserved on the branch feat/cross-node-fsm-digest. Reopen once the restore/encoding-mode design is settled.

PR #1442, opened 2026-06-29
test(EN-1410): antithesis detector + dev-loop workload filter

gfyrag:
## Summary
- Adds an Antithesis red detector for EN-1410 (spurious `INSUFFICIENT_FUNDS` caused by bloom diverging from Pebble after a `!IsReady` cache rotation).
- Adds a dev-loop driver filter on the workload image so iterating on a targeted bug doesn't depend on waiting for the test composer to randomly pick the relevant driver out of ~40 parallel_driver_*.

## Why
EN-1410 surfaces as a non-deterministic race: a cache rotation that fires while the bloom rebuild is still `!IsReady` wipes Pebble `0xFF` (outgoing gen1) but skips `PersistDirtyBlocks`. A subsequent crash loses the in-memory bloom; on the next boot the rebuild has neither the persisted block nor the cache `0xFF` entry. The FSM injects zero on the next read, and an exact-balance transfer fails `INSUFFICIENT_FUNDS` even though the read-side index still shows the balance. This is the exact signature on the `blockchains` dev cluster (`ledgerctl accounts aggregate-volumes` shows balance > 0, the ingester is stuck on `INSUFFICIENT_FUNDS` retries on the same address).

## Detector — `parallel_driver_insufficient_funds`

Rewritten from a fresh-ledger-per-iteration shape (which could never expose the bug — fund and transfer were too close in time for a `!IsReady` rotation to fall between them) to a pattern that matches the blockchains usage:

- Long-lived shared ledger `insuf-shared` + fixed pool of 50 accounts that survive restarts, snapshots, and rotations.
- Each iteration:
  1. Pick a random pool cell.
  2. Read balance via `GetAccount` (read-side index — ground truth, NOT the FSM cache the bug lives in).
  3. If balance == 0 → fund with `Force=true` and exit (next iteration to pick the cell will do the transfer).
  4. If balance > 0 → attempt exact-balance transfer back to `world` with `Force=false`.
- After an `INSUFFICIENT_FUNDS` response, re-read the balance: only fire the assertion if read-side still shows > 0. This disambiguates the bug from concurrent-drain races between parallel instances (caught the v2 run on `insuf-users:11` — succeed at `t+181.441` then fail at `t+181.442` from another instance that had read the same balance).
- `AlwaysOrUnreachable` fires only when the conjunction holds: apply rejected on `available=0` AND read-side proves balance > 0 after the fact. The only path producing that is the FSM cache injecting a zero `VolumePair` on a key Pebble still holds — i.e. the bloom dropped it.

## Dev-loop filter — `INCLUDED_DRIVERS`

Antithesis's test composer ranking is opaque to users (per the test_composer reference, no user-configurable weighting). Untargeted runs spend most of the slot budget on whichever drivers the composer judges interesting — `parallel_driver_insufficient_funds` was scheduled zero times in the first 1h run.

New build-arg on the workload image: when set to a comma-separated list of `<template>/<driver>` paths, only those binaries are shipped to `/opt/antithesis/test/v1/`. The composer then has nothing else to schedule.

Plumbed through every Justfile recipe that builds + pushes the workload:

```bash
just antithesis::k8s-run 1 "EN-1410 red" \
    "main/first_default_ledger,main/parallel_driver_insufficient_funds,main/singleton_driver_config_change,..."
```

Always include `main/first_default_ledger` — without an initial ledger, every `parallel_driver_*` call to `GetRandomLedger` blocks until the 10 min timeout.

Empty (default) preserves the existing full-coverage behaviour for nightly runs.

## Status

- v1 run: confirmed the assertion is reachable, but the driver was scheduled 0 times — motivated the filter.
- v2 run with filter + pool refactor: assertion fired once (insuf-users:11), confirmed as a concurrent-drain race via the 1 ms gap — motivated the re-read confirm.
- v3 run (`943c0b33a482752ae813b0e30935af8a-56-17`) with the re-read fix in flight as of this PR. Report by email when 1h elapsed.

Intentionally NOT rebased on top of the fix PR #1432 — the first goal is to confirm the detector catches the bug on un-fixed `release/v3.0`. Once confirmed, a follow-up run on top of the fix should show the assertion no longer fires.

## Test plan
- [ ] v3 Antithesis run shows the assertion does NOT fire after the re-read confirm patch (= no real bug observed in this run, or the re-read filtered remaining false positives)
- [ ] Follow-up: rebase on top of `fix/EN-1410-bloom-boot-ordering` (#1432) and re-run to confirm the fix closes the window
- [x] Local build: `GOROOT= go build ./tests/antithesis/workload/bin/cmds/main/parallel_driver_insufficient_funds/` passes
- [x] `just antithesis::k8s-run` with filter and full driver lists both work (dry-run + live)

### DIFF
--- a/tests/antithesis/workload/bin/cmds/main/parallel_driver_insufficient_funds/main.go
+++ b/tests/antithesis/workload/bin/cmds/main/parallel_driver_insufficient_funds/main.go
@@ -1,140 +1,206 @@
+// parallel_driver_insufficient_funds is the EN-1410 detector: it stresses
+// the exact-balance transfer path on a long-lived pool of accounts, so an
+// account funded by a past iteration can be picked up by a future iteration
+// after restarts, snapshots, and cache rotations have had time to fire.
+//
+// The bug it targets (EN-1410 bloom boot-ordering): a cache rotation that
+// runs while the bloom rebuild is still !IsReady wipes Pebble 0xFF for the
+// outgoing generation but skips PersistDirtyBlocks. A later crash loses the
+// in-memory bloom; the next boot's rebuild reads only persisted blocks +
+// cache 0xFF, both of which are now missing the key. MayContain returns
+// false, the FSM injects a zero VolumePair, and an exact-balance transfer
+// is rejected with INSUFFICIENT_FUNDS even though the account's volume cell
+// still sits in Pebble 0x01.
+//
+// The asymmetry the driver exploits: the read-side index (GetAccount) and
+// the FSM cache (apply-time CheckBalance) are independent. The read-side
+// sees the funded balance; the FSM bloom-gated cache may not. A spurious
+// INSUFFICIENT_FUNDS on an account whose read-side balance is non-zero is
+// the bug's signature.
+//
+// Iteration:
+//  1. Create the shared ledger idempotently (every iteration races; only
+//     one CreateLedger succeeds, the rest see AlreadyExists).
+//  2. Pick a random account from a fixed pool of accountPoolSize cells.
+//  3. Read the account's balance via the read-side. If unknown / empty,
+//     fund it (Force=true) and exit — the next iteration to pick this
+//     cell will see balance > 0 and attempt the transfer.
+//  4. If balance > 0, attempt an exact-balance transfer back to world
+//     without Force. The two assertions:
+//       - Sometimes: campaign coverage — at least once the transfer must
+//         succeed (proves the path is reachable in steady state).
+//       - AlwaysOrUnreachable: per-iteration safety — INSUFFICIENT_FUNDS
+//         is unreachable; firing it is the EN-1410 signature.
+//
+// The pool is shared across every instance of the driver: accounts funded
+// in past iterations survive restarts and snapshots, so a balance > 0
+// observed today may have been written hours ago — that's the time window
+// during which a bloom-vs-Pebble desync would have formed.
 package main
 
 import (
 	"context"
 	"fmt"
+	"math/big"
 
 	"github.com/antithesishq/antithesis-sdk-go/assert"
+	"github.com/holiman/uint256"
+
 	"github.com/formancehq/ledger/v3/internal/domain"
 	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
 	"github.com/formancehq/ledger/v3/internal/proto/servicepb"
 	"github.com/formancehq/ledger/v3/tests/antithesis/workload/internal"
 )
 
+const (
+	// sharedLedger is shared across every instance of the driver so accounts
+	// outlive any single iteration; without this the bug cannot surface —
+	// fund and transfer happen too close together in time for a !IsReady
+	// rotation to fall between them.
+	sharedLedger = "insuf-shared"
+
+	// accountPoolSize bounds the address space the driver cycles through.
+	// Small enough that repeat picks happen across many iterations (so a
+	// post-restart pick lands on a previously funded cell), large enough
+	// that consecutive iterations rarely collide on the same cell.
+	accountPoolSize = 50
+
+	// fundAmount is the fixed credit applied to an empty pool account. The
+	// driver always transfers the exact read-side balance back, so the
+	// precise amount does not need to be remembered between iterations.
+	fundAmount = 1000
+
+	asset = "USD/2"
+)
+
 func main() {
 	internal.RunDriver("parallel_driver_insufficient_funds", func(ctx context.Context, client servicepb.BucketServiceClient, _ string) {
 		r := internal.Rand()
 
-		// Use a dedicated ledger to control account balances precisely.
-		ledger := fmt.Sprintf("insuf-%d", r.Uint64()%1_000_000)
-		if err := internal.CreateLedger(ctx, client, ledger); err != nil {
+		// Race-create the shared ledger. AlreadyExists is the steady state
+		// once any prior iteration has won the race.
+		if err := internal.CreateLedger(ctx, client, sharedLedger); err != nil &&
+			!internal.IsAlreadyExists(err) && !internal.IsTransient(err) {
 			return
 		}
 
-		// Use a dedicated account prefix so other drivers (e.g. audit) that post
-		// to "users:N" on random ledgers cannot interfere with balance assumptions.
-		account := fmt.Sprintf("insuf-users:%d", r.Uint64()%internal.UserAccountCount)
-		asset := "USD/2"
-		fundAmount := r.Uint64()%10000 + 100
-		details := internal.Details{"ledger": ledger, "account": account, "asset": asset, "fundAmount": fundAmount}
+		account := fmt.Sprintf("insuf-users:%d", r.Uint64()%accountPoolSize)
+		details := internal.Details{
+			"ledger":  sharedLedger,
+			"account": account,
+			"asset":   asset,
+		}
 
-		// 1. Fund the account from world with a known amount.
-		resp, err := client.Apply(ctx, servicepb.UnsignedApplyRequest("", &servicepb.Request{
-			Type: &servicepb.Request_Apply{
-				Apply: &servicepb.LedgerApplyRequest{
-					Ledger: ledger,
-					Action: &servicepb.LedgerAction{Data: &servicepb.LedgerAction_CreateTransaction{
-						CreateTransaction: &servicepb.CreateTransactionPayload{
-							Postings: []*commonpb.Posting{{
-								Source:      "world",
-								Destination: account,
-								Amount:      commonpb.NewUint256FromUint64(fundAmount),
-								Asset:       asset,
-							}},
-							Force:         true,
-							ExpandVolumes: true,
-						},
-					}},
-				},
-			},
-		}))
+		balance, err := readBalance(ctx, client, sharedLedger, account)
 		if err != nil {
-			if internal.IsTransient(err) {
-				return
-			}
-
-			assert.Unreachable("funding should not fail", details.With(internal.Details{"error": err}))
-
+			// Read-side temporarily unavailable or account not yet visible
+			// to the read-side index; nothing to assert this round.
 			return
 		}
 
-		if ct := internal.ExtractCreatedTransaction(resp); ct != nil {
-			internal.CheckPostCommitVolumes(ct.PostCommitVolumes, details)
-		}
-
-		// 2. Attempt exact balance transfer (should succeed without Force).
-		_, err = client.Apply(ctx, servicepb.UnsignedApplyRequest("", &servicepb.Request{
-			Type: &servicepb.Request_Apply{
-				Apply: &servicepb.LedgerApplyRequest{
-					Ledger: ledger,
-					Action: &servicepb.LedgerAction{Data: &servicepb.LedgerAction_CreateTransaction{
-						CreateTransaction: &servicepb.CreateTransactionPayload{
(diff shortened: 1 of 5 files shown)


tests/antithesis/workload/bin/cmds/main/parallel_driver_insufficient_funds/main.go
```diff
@@ -1,128 +1,223 @@
...
+	// — i.e. the bloom dropped a key whose volume Pebble still holds.
+	//
+	// The re-read itself is racy too (the read-side can lag, a concurrent
+	// fund can re-credit the cell, etc.). We accept those: a confirmed
+	// bug requires the conjunction (apply rejected available=0 AND
+	// read-side balance > 0 after the fact). A miss in either direction
+	// is the safe answer.
+	confirmed := false
+	if internal.HasErrorReason(err, domain.ErrReasonInsufficientFunds) {
+		if confirm, rerr := readBalance(ctx, client, sharedLedger, account); rerr == nil && confirm.Sign() > 0 {
```
NumaryBot, line 215: 🟠 **[major] Avoid confirming races with a fresh balance**

With multiple instances sharing the 50-account pool, this confirmation can still turn a legitimate concurrent drain into a red assertion: this instance reads 1000, another instance drains it, this transfer correctly gets `INSUFFICIENT_FUNDS`, and then a later iteration re-funds the same account before this re-read. Line 215 then sees `confirm.Sign() > 0` from the new credit and reports EN-1410 even though the failed transfer was correct.

tests/antithesis/workload/bin/cmds/main/parallel_driver_insufficient_funds/main.go
```diff
@@ -1,128 +1,223 @@
...
 func main() {
 	internal.RunDriver("parallel_driver_insufficient_funds", func(ctx context.Context, client servicepb.BucketServiceClient, _ string) {
 		r := internal.Rand()
 
-		// Use a dedicated ledger to control account balances precisely.
-		ledger := fmt.Sprintf("insuf-%d", r.Uint64()%1_000_000)
-		if err := internal.CreateLedger(ctx, client, ledger); err != nil {
+		// Race-create the shared ledger. AlreadyExists is the steady state
+		// once any prior iteration has won the race.
+		if err := internal.CreateLedger(ctx, client, sharedLedger); err != nil &&
```
NumaryBot, line 82: 🟠 **[major] Do not continue after an ambiguous ledger create**

Under fault injection, `CreateLedger` can return a transient error without the ledger having been committed, but this branch proceeds anyway. The next `GetAccount` maps the resulting ledger `NotFound` to a zero balance, and `fund` can then raise `Unreachable("funding should not fail")` for a missing ledger, creating a false detector failure unrelated to EN-1410.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1442#issuecomment-4833921682

tests/antithesis/workload/bin/cmds/main/parallel_driver_insufficient_funds/main.go
NumaryBot, line 82: 🟠 **[major] Proceeding after ambiguous ledger creation causes false detector failures**

Under fault injection, `CreateLedger` can return a transient error without the ledger having been committed. The code currently proceeds as if the ledger exists, so the subsequent `GetAccount` call maps a ledger `NotFound` response to a zero balance, and `fund` then raises `Unreachable("funding should not fail")` for the missing ledger. This produces a detector failure that is entirely unrelated to EN-1410, undermining the reliability of the antithesis detector.

**Suggestion:** Treat any ambiguous (transient/non-AlreadyExists) error from `CreateLedger` as fatal for this iteration — return or retry rather than continuing. Only proceed when the ledger is confirmed to exist (either a success response or an `AlreadyExists` error).

tests/antithesis/workload/bin/cmds/main/parallel_driver_insufficient_funds/main.go
NumaryBot, line 215: 🟠 **[major] Re-reading balance after INSUFFICIENT_FUNDS can confirm a normal concurrency race as a false EN-1410 assertion**

Multiple workload instances share the same 50-account pool. When this instance reads a positive balance, another instance can drain it first, causing this transfer to correctly receive `INSUFFICIENT_FUNDS`. If yet another instance then re-funds the same account before this re-read at line 215, the confirmation sees `confirm.Sign() > 0` and raises the EN-1410 assertion even though the failed transfer was entirely correct. This means normal concurrent execution under Antithesis can produce false red assertions.

**Suggestion:** To avoid this race, tie the confirmation to the specific balance snapshot taken before the transfer attempt — e.g., record the pre-transfer balance and verify that it was non-positive at the time of the failed transfer (using a ledger transaction log or sequence number), rather than re-reading a potentially updated balance after the fact.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1442#issuecomment-4833921682

tests/antithesis/workload/bin/cmds/main/parallel_driver_insufficient_funds/main.go
NumaryBot, line 215: 🟠 **[major] Confirmation re-read can produce false red assertion after concurrent re-funding**

With multiple instances sharing the 50-account pool, the confirmation logic at line 215 can incorrectly flag a legitimate concurrent drain as an EN-1410 violation. One instance reads a positive balance, another drains the account, this instance's transfer correctly receives INSUFFICIENT_FUNDS, but a subsequent iteration re-funds the same account before the confirmation re-read. Line 215 then sees `confirm.Sign() > 0` from the new credit and raises a red assertion even though no bug occurred.

**Suggestion:** Tie the confirmation read to the same account snapshot used for the transfer decision (e.g. include a sequence/version check or record the pre-transfer balance and compare against it rather than a fresh read), so re-funding by a concurrent goroutine cannot be mistaken for evidence of a bug.

tests/antithesis/workload/bin/cmds/main/parallel_driver_insufficient_funds/main.go
NumaryBot, line 82: 🟠 **[major] Continuing after an ambiguous ledger create error causes false detector failure**

Under fault injection, `CreateLedger` can return a transient error without the ledger having been committed. If the code continues anyway, the subsequent `GetAccount` maps the resulting `NotFound` to a zero balance, and `fund` can then raise `Unreachable("funding should not fail")` for a missing ledger. This creates a false detector failure entirely unrelated to EN-1410.

**Suggestion:** Treat any non-idempotent error from `CreateLedger` as terminal for that attempt (return/retry) rather than proceeding. Only continue if the error is explicitly `AlreadyExists` (meaning the ledger is confirmed present) or if the call succeeded.

tests/antithesis/workload/bin/cmds/main/parallel_driver_insufficient_funds/main.go
```diff
@@ -1,128 +1,223 @@
...
+	amount, overflow := uint256.FromBig(balance)
+	if overflow {
+		// Pool balances never exceed fundAmount per cell so overflow is
+		// impossible in normal operation; an overflow here would mean
+		// something other than this driver wrote into the pool address
+		// space — skip rather than mis-assert.
+		return
+	}
+
+	_, err := client.Apply(ctx, servicepb.UnsignedApplyRequest("", &servicepb.Request{
```
NumaryBot, line 173: 🟠 **[major] Critical transfer Apply submitted without idempotency key, enabling false INSUFFICIENT_FUNDS on retry**

Under Antithesis fault injection, the non-idempotent `Apply` call at line 173 can commit the exact-balance transfer but lose the response. The client retry then replays the same transfer against an already-drained account and receives INSUFFICIENT_FUNDS. This error can be treated as the EN-1410 signature even though the ledger bloom is healthy, producing a false positive red assertion.

**Suggestion:** Assign a stable, per-attempt idempotency key to this transfer before the first attempt and reuse that same key on all retries, so a replayed Apply on an already-drained account is recognised as a duplicate and returns the original success response rather than INSUFFICIENT_FUNDS.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1442#issuecomment-4833921682

tests/antithesis/workload/bin/cmds/main/parallel_driver_insufficient_funds/main.go
NumaryBot, line 215: 🟠 **[major] Confirmation re-read after INSUFFICIENT_FUNDS can produce false red assertions under concurrency**

Multiple workload instances share the same 50-account pool. When this instance reads a positive balance and submits a transfer, another instance can drain the account first, causing this transfer to correctly receive INSUFFICIENT_FUNDS. If yet another instance (or a retry) re-funds the same account before the confirmation re-read at line 215, the read sees a positive balance (confirm.Sign() > 0) and raises the EN-1410 assertion even though no bug occurred. The confirmation is not causally tied to the moment of the failed transfer — it can also be served from a stale read-side view that has not yet caught up to the concurrent drain, independently causing the same false-positive.

**Suggestion:** Tie the confirmation to the specific balance snapshot taken before the transfer attempt. For example, record the pre-transfer balance and verify it was non-positive at the time of the failure using a ledger sequence/version number or transaction log position, rather than issuing a fresh GetAccount read after the fact. The confirmation read must be at a log position at least as recent as the failed Apply attempt.

tests/antithesis/workload/bin/cmds/main/parallel_driver_insufficient_funds/main.go
NumaryBot, line 82: 🟠 **[major] Proceeding after ambiguous CreateLedger error causes false detector failures unrelated to EN-1410**

Under fault injection, CreateLedger can return a transient error without the ledger having actually been committed. The code currently proceeds as if the ledger exists. The subsequent GetAccount call then maps a ledger NotFound response to a zero balance, and fund raises Unreachable("funding should not fail") for the missing ledger. This produces a detector failure entirely unrelated to EN-1410, undermining the reliability of the antithesis detector.

**Suggestion:** Treat any ambiguous (transient / non-AlreadyExists) error from CreateLedger as fatal for that iteration — return or retry rather than continuing. Only proceed when the ledger is confirmed to exist: either a successful response or an explicit AlreadyExists error.

tests/antithesis/workload/bin/cmds/main/parallel_driver_insufficient_funds/main.go
NumaryBot, line 173: 🟠 **[major] Critical Apply call has no idempotency key, enabling false INSUFFICIENT_FUNDS on retry**

Under Antithesis fault injection, the non-idempotent Apply call at line 173 can commit the exact-balance transfer but lose the response. A client retry then replays the same transfer against an already-drained account and receives INSUFFICIENT_FUNDS. This error can be treated as the EN-1410 signature even though the ledger bloom is healthy, producing a false-positive red assertion.

**Suggestion:** Assign a stable, per-attempt idempotency key to this transfer before the first attempt and reuse that same key on all retries. A replayed Apply on an already-drained account will then be recognised as a duplicate and return the original success response rather than INSUFFICIENT_FUNDS.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1442#issuecomment-4833921682

PR #1443, opened 2026-06-29
fix(auth): bound OIDC JWKS reads and drop the inert go-libs JWT module

Azorlogh:
The server exposes `--auth-discovery-timeout` (default `10s`), applied in `buildAuthConfig`, which already bounds the OIDC discovery HTTP call via a context deadline. This PR extends that same bound to the JWKS reads and removes an inert auth fx module.

### Changes

**1. Bound the JWKS read client**

`buildAuthConfig` bounded `oidc.Discover` with a context deadline, but built the remote keyset with `NewRemoteKeySet(nil, …)`:

```go
oidcKeySet = oidcclient.NewRemoteKeySet(nil, discovery.JwksURI)
```

The keyset here is go-libs' `oidc/client` (`authn/oidc/client`), not zitadel directly. On a `nil` client it does **not** use `http.DefaultClient` — it falls back to the package's `DefaultHTTPClient`, which carries a hardcoded **30s** timeout:

```go
// go-libs/v5/pkg/authn/oidc/client/jwks.go
func NewRemoteKeySet(client *http.Client, jwksURL string, ...) oidc.KeySet {
	if client == nil {
		client = httphelper.DefaultHTTPClient   // &http.Client{Timeout: 30 * time.Second}
	}
	...
}
```

So JWKS key fetches (lazy, on first token verification) were bounded — but at a fixed 30s, ignoring `--auth-discovery-timeout`. This passes the configured client so discovery and JWKS reads share one operator-controlled ceiling:

```go
oidcKeySet = oidcclient.NewRemoteKeySet(TimeoutHTTPClient(cfg.AuthConfig.OIDCDiscoveryTimeout), discovery.JwksURI)
```

`TimeoutHTTPClient(0)` returns `http.DefaultClient` (no timeout), so `--auth-discovery-timeout=0` now makes JWKS reads truly unbounded too, matching the documented `0 = unbounded` opt-out (the previous 30s floor on JWKS is intentionally removed in that case).

**2. Drop the inert `authnfx` JWT module + its no-op `fx.Decorate`**

`cmd/server` wired `authnfx.JWTModuleFromFlags(cmd)` plus an `fx.Decorate` to bound the `*http.Client` fed to `NewKeySets`. That wiring is dead:

- `jwt.NewKeySets` returns `map[string]oidc.KeySet` and `jwt.NewAuthenticatorFromConfig` returns an `Authenticator` — nothing in this service consumes either.
- `buildAuthConfig`'s injected `oidcKeySet` is a *singular* `oidc.KeySet` (`optional:"true"`, `module.go:530`) that the module never provides, so it's always `nil` and the local discovery path always runs.
- fx providers are lazy: with no consumer of the module's outputs, `NewKeySets` never runs, the decorated `*http.Client` is never requested, and the decorator never fires.

So `buildAuthConfig` is already the sole, fully-bounded discovery path. The wiring is removed entirely (no leftover `fx.Module("auth")` marker), and the now-unused `authnfx` / `net/http` imports are dropped. Auth flags remain registered independently via `auth.AddFlags`.

### Testing

- `go build ./...`, `go vet ./cmd/server/... ./internal/bootstrap/...`, `gofmt -l` — clean
- `go test ./internal/bootstrap/ -run 'AuthConfig|TimeoutHTTPClient|DiscoveryContext|HangingIssuer|RemoteKeySet_JWKSRead'` — pass
- Adds `TestRemoteKeySet_JWKSReadRespectsTimeout`: builds the keyset exactly as `buildAuthConfig` does (`TimeoutHTTPClient(timeout)`), points it at a blackholed `/jwks` handler, and asserts `VerifySignature` returns an error inside ~2×timeout instead of stalling — covering the line this PR changes.

### Files

- `cmd/server/server.go` — remove the `authnfx` module + `fx.Decorate`
- `internal/bootstrap/module.go` — pass `TimeoutHTTPClient(...)` to `NewRemoteKeySet`; refresh comments
- `internal/bootstrap/config.go` — refresh the `OIDCDiscoveryTimeout` doc comment
- `internal/bootstrap/auth_discovery_timeout_test.go` — add the JWKS-bound test
- `docs/ops/cli.md` — update the `--auth-discovery-timeout` description

### Notes for reviewers

- Change 2 removes the `authnfx` module and its decorator. It's verified safe (zero consumers of its outputs in this service). If a future feature needs the go-libs `Authenticator`, the module can be re-added with a real consumer.

### DIFF
--- a/cmd/server/server.go
+++ b/cmd/server/server.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"net/http"
 	"os"
 	"runtime"
 	"runtime/debug"
@@ -21,7 +20,6 @@ import (
 	"gopkg.in/yaml.v3"
 
 	auth "github.com/formancehq/go-libs/v5/pkg/authn/jwt"
-	"github.com/formancehq/go-libs/v5/pkg/fx/authnfx"
 	"github.com/formancehq/go-libs/v5/pkg/fx/observefx"
 	otlp "github.com/formancehq/go-libs/v5/pkg/observe"
 	logging "github.com/formancehq/go-libs/v5/pkg/observe/log"
@@ -192,11 +190,11 @@ func NewRunCommand() *cobra.Command {
 	runCmd.Flags().String("auth-scope-mapping-file", "", "Path to JSON file mapping virtual scopes (e.g. ledger:read) to granular scopes")
 	runCmd.Flags().String("auth-anonymous-scopes", "", "Comma-separated granular scopes granted to requests without a bearer token (e.g. \"*:read\" for writes-only mode). Wildcards: *:read, *:write")
 
-	// OIDC discovery + JWKS HTTP timeout. A slow or blackholed issuer would
-	// otherwise hang startup indefinitely (go-libs supplies http.DefaultClient
-	// with no Timeout and discovery uses context.Background()). 0 keeps the
-	// legacy unbounded behavior for operators that need it.
-	runCmd.Flags().Duration("auth-discovery-timeout", 10*time.Second, "Bound the HTTP calls used for OIDC discovery and JWKS fetches at startup (0 = unbounded)")
+	// OIDC discovery + JWKS HTTP timeout, applied in buildAuthConfig: a context
+	// deadline bounds oidc.Discover and the keyset's http.Client.Timeout bounds
+	// JWKS fetches, so both share one operator-controlled ceiling. 0 leaves them
+	// unbounded for operators that need it.
+	runCmd.Flags().Duration("auth-discovery-timeout", 10*time.Second, "Bound the HTTP calls used for OIDC discovery and JWKS fetches (0 = unbounded)")
 
 	// Idempotency TTL and eviction
 	runCmd.Flags().Duration("idempotency-ttl", 24*time.Hour, "Idempotency key time-to-live (0 = never expire)")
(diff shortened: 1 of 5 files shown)


gfyrag (commented):
LGTM on the fix itself. Two things worth clarifying/tightening before merge.

### 1. Pre-fix symptom is likely a panic, not a hang

Description says `NewRemoteKeySet(nil, …)` fell back to `http.DefaultClient`. It doesn't: zitadel's `NewRemoteKeySet` stores the client verbatim and `fetchRemoteKeys` calls `client.Do(req)` on the resulting `nil` receiver — verified with a standalone snippet, this panics with nil pointer dereference.

The panic happens inside the detached goroutine spawned by `keysFromRemote` (`go r.updateKeys(ctx)` in zitadel's `jwks.go`), so:
- `recoveryInterceptor` in `internal/adapter/grpc/server.go:342` can't catch it — different goroutine.
- The parent goroutine stays parked on `<-inflight.wait()` because `inflight.done()` is never reached after the panic, hence the "hang" impression — but the runtime terminates the process.

Fix is still correct (passing a valid client avoids both the panic and any real hang from a slow issuer). But it's worth rewriting the description accordingly, because it changes how one greps prod logs: we should be looking for `nil pointer dereference` stacks originating in zitadel `jwks.go`, not for stuck goroutines.

The buggy line landed in `39130dc80` (2026-03-27). If nobody has reported a crash since, worth a sanity check that the path was actually exercised by a real deployment with `--auth-issuer` set (vs. Ed25519-only deployments where the branch is skipped).

### 2. The line that matters isn't covered by a test

`TestBuildAuthConfig_HangingIssuerRespectsTimeout` covers `oidc.Discover` — good, but that's the pre-existing fix. Nothing asserts that the JWKS keyset is bounded, which is what this PR changes. You call it out in the notes; I'd still push for coverage before merge given it's *the* critical line.

Two options:
- **Unit**: pass a custom `http.RoundTripper` that blocks (`remoteKeySet.httpClient` is unexported but the transport is publicly settable via the client we build). Wire it via a test-only variant of the discovery/keyset construction, call `VerifySignature`, assert return < ~2×timeout.
- **e2e**: extend `tests/e2e/cluster/auth_test.go` with an OIDC server whose `/jwks` handler blocks — verify the auth call errors quickly instead of the server crashing.

### 3. Nit — `authModule := fx.Module("auth")` is dead

An `fx.Module` with no providers is a no-op. Either inline the removal (drop the variable and its entry in `opts`, keep the comment above `opts`) or keep it as a marker for diff readability — but the current form is code without effect.

---

Everything else looks good: symmetry between `discoveryContext` / `TimeoutHTTPClient`, `0 = unbounded` opt-out preserved on both, comments purged of the stale claim about `http.DefaultClient`, docs updated. `optional:"true"` on `buildAuthConfig`'s `oidcKeySet` param (module.go:530) confirms nothing was providing it — the `authnfx` removal reasoning holds.

gfyrag (approved):
LGTM. Deux fixes ciblés bien traqués :

**Bug réel sur `--auth-discovery-timeout`** : le flag était documenté pour borner les JWKS reads mais `NewRemoteKeySet(nil, ...)` fallait sur `httphelper.DefaultHTTPClient` de go-libs (30s hardcodés), pas sur `http.DefaultClient`. Résultat : le flag était silently overridden, et `--auth-discovery-timeout=0` (documented unbounded) imposait quand même 30s. Le fix `TimeoutHTTPClient(OIDCDiscoveryTimeout)` restore le contrat documenté. Bonne discipline "root cause" — l'extract du code go-libs dans la PR body prouve la claim.

**Suppression du dead code justifiée par analyse structurelle** : `authnfx.JWTModule` fournit `map[string]oidc.KeySet` + `Authenticator`, `buildAuthConfig` attend un `oidc.KeySet` singulier via `optional:\"true\"` que le module ne fournit jamais. fx providers étant lazy sans consommateur, ni `NewKeySets` ni le `fx.Decorate` ne firm jamais. Preuve structurelle, pas gut feeling.

**Test solide** : `TestRemoteKeySet_JWKSReadRespectsTimeout` pin le real behavior (httptest server bloquant + HS256 signer pour forcer le fetch + assertion `< 2×timeout`). Pas juste du wiring.

**Points non bloquants à considérer**
- Naming : `OIDCDiscoveryTimeout` borne maintenant aussi les JWKS runtime fetches (pas juste discovery startup). La doc du flag est OK, un rename en `OIDCHTTPTimeout` serait plus honnête mais pas worth le breaking change.
- Semantic change observable : 30s → 10s par défaut sur les JWKS runtime fetches (les opérateurs qui toléraient un IdP briefly slow via l'ancien ceiling accidentel verront des token failures plus tôt). Mérite une ligne dans les release notes — c'est un "restore documented behavior" mais reste observable.

Azorlogh:
Thanks — dug into this. The keyset here is go-libs' `oidc/client`, not zitadel directly: its `NewRemoteKeySet` nil-guards to `httphelper.DefaultHTTPClient` (`&http.Client{Timeout: 30s}`), so there's no nil `client.Do` and no panic — the pre-fix behavior was a fixed 30s JWKS bound that ignored `--auth-discovery-timeout`. Rewrote the description/commit accordingly (nothing to grep in prod — no crash, no hang, just a 30s ceiling), added `TestRemoteKeySet_JWKSReadRespectsTimeout` for the changed line, and inlined the `fx.Module("auth")` removal.

PR #1444, opened 2026-06-29
fix(EN-1242): tombstone both cache generations on metadata delete

Azorlogh:
# fix(cache): tombstone both generations on metadata delete (EN-1242)

## Context

The in-memory `AttributeCache` is dual-generation (`gen0` = current, `gen1` =
previous, rotated on each generation boundary). The 0xFF Pebble zone mirrors
both generations for restart recovery, and the invariant is that **the
in-memory cache equals the 0xFF zone for the same applied index**.

When the FSM applies a metadata delete, two writes happen in the same Pebble
batch:

1. **In memory** — `KeyStore.Delete` tombstoned the entry via `s.M.Put(id, entry)`,
   and `AttributeCache.Put` writes **gen0 only**.
2. **On disk** — `writeCacheTombstone` stamps a tombstone at **both** gen bytes
   in the 0xFF zone.

So immediately after a delete on a key whose live entry sat in `gen1`
(post-rotation), the two sides disagreed:

| Location | gen0 | gen1 |
|---|---|---|
| in-memory `AttributeCache` | tombstone | **live entry** |
| Pebble 0xFF | tombstone | tombstone |

It was invisible at runtime because `AttributeCache.Get` reads gen0 first and
returns the tombstone, and a restart rehydrates `gen1` from disk as a tombstone
— but the cache-mirror invariant was broken between the delete and the next
restart. Any code that consults `gen1` directly (or any memory-vs-disk
consistency check) would see the mismatch.

## Fix

`AttributeCache.Del` already documented that it "marks the entry as a tombstone
in **both Gen0 and Gen1**", but its implementation only tombstoned generations
that already held the key, and `KeyStore.Delete` never called it (it used
`Put`, hitting gen0 only).

This change makes `Del` honor its contract and routes the delete through it:

- **`AttributeCache.Del`** now tombstones **both** generations and resets the
  entry's `Data` to zero (the tombstone-payload reset EN-1377 introduced). When a
  generation lacks the key it borrows the tag from the one that has it (a delete
  always has the key in at least one gen); the borrow only fires for an empty
  target generation, so it never clobbers a colliding entry, and it matches what
  `writeCacheTombstone` does on disk.
- **`KeyStore.Delete`** validates existence + tag (unchanged) and then calls
  `s.M.Del(id)` — reusing the existing `kv.KV.Del` interface method instead of a
  bespoke gen0-only `Put`. For a single-map `kv.KV` (test stores) `Del` removes
  the entry, which `Get` also reports as not found.

Net result: after a metadata delete, both in-memory generations carry the
tombstone, matching the 0xFF zone for the same applied index.

## Tests

- `TestKeyStoreDelete_TombstonesAllGenerations` (cache) — write → rotate (entry
  now gen1-only) → delete → asserts **both** generations are tombstones and
  `Get` returns `ErrNotFound`. This is the EN-1242 acceptance scenario.
- `TestAttributeCache_Del` (cache) — strengthened to assert `Del` tombstones
  both generations (with `Data` reset), including the borrowed-tag path for the
  generation that lacked the key.

Both fail against the pre-fix code (gen1 stays live) and pass with the fix.

## Relationship to EN-1377

EN-1377 (explicit on-disk tombstone flag byte, `persistLeanProtoEntries`
hardening, and `AttributeCache.Del` resetting `Data`) is **already merged** into
the base. The two are complementary halves of the same cache-mirror story:

- **EN-1377** makes the in-memory↔disk mapping unambiguous — an out-of-band flag
  byte, so a presence-only / all-default proto no longer round-trips as a
  tombstone.
- **EN-1242** (this PR) makes the in-memory `Deleted` flag correct across **both**
  generations, so `gen1` never keeps a live entry after a `gen0`-only delete.

Both reworked `AttributeCache.Del`, so this branch's `Del` reconciles them: it
tombstones both generations (EN-1242) **and** resets `Data` to zero (EN-1377).

## Verification

- `just pre-commit` passes (go generate, go mod tidy, golangci-lint).
- `go build ./...`, `go vet`, and the `cache` / `attributes` / `state` package
  suites pass.
- Rebased onto `formancehq/ledger` `release/v3.0`; the change touches no files
  modified by the rest of that branch.

## Files changed

- `internal/infra/cache/cache.go` — `Del` tombstones both generations.
- `internal/infra/attributes/key_store.go` — `Delete` routes through `s.M.Del`.
- `internal/infra/cache/cache_test.go` — strengthened `TestAttributeCache_Del`.
- `internal/infra/cache/keystore_delete_test.go` — new EN-1242 regression test.

### DIFF
--- a/internal/infra/attributes/key_store.go
+++ b/internal/infra/attributes/key_store.go
@@ -158,10 +158,14 @@ func (s *KeyStore[K, T]) GetEntry(canonical []byte) (Entry[T], bool) {
 	return entry, true
 }
 
-// Delete marks the entry as a tombstone instead of removing it.
+// Delete marks the entry as a tombstone in the underlying store.
+// On the dual-generation cache the tombstone is written to every generation
+// (cache.AttributeCache.Del), mirroring the 0xFF zone the FSM writes in the
+// same batch, so the in-memory cache equals disk for the same applied index.
 // The entry stays in the cache (surviving MirrorTouch during pipelined
-// proposals) but reads via Get return nil. Tombstones age out naturally
-// via cache generation rotation.
+// proposals) and reads via Get return ErrNotFound. Tombstones age out naturally
+// via cache generation rotation. A single-map kv.KV (tests) removes the entry,
+// which Get also reports as not found.
 func (s *KeyStore[K, T]) Delete(canonical []byte) (id U128, tag uint64, err error) {
 	id, tag = s.hasher.MakeKey(canonical)
 
@@ -174,8 +178,7 @@ func (s *KeyStore[K, T]) Delete(canonical []byte) (id U128, tag uint64, err erro
 		return id, tag, newErrCollisionDetected(canonical, entry.Tag, tag)
 	}
 
-	entry.Deleted = true
-	s.M.Put(id, entry)
+	s.M.Del(id)
 
 	return id, tag, nil
 }
--- a/internal/infra/cache/cache.go
+++ b/internal/infra/cache/cache.go
@@ -74,33 +74,42 @@ func (a *AttributeCache[T]) GetAndPut(k attributes.U128, v attributes.Entry[T]) 
 	return a.gen1.Load().Get(k)
 }
 
-// Del marks the entry as a tombstone in both Gen0 and Gen1 instead of
-// removing it. This prevents pipelined MirrorTouch from failing when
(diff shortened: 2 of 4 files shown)


gfyrag (changes requested):
Temporary blocker to avoid spurious merge.

PR #1445, opened 2026-06-29
fix(ctrl): read ledger-stitched queries from a single snapshot

Azorlogh:
Migrated from formancehq/ledger-v3-poc#589 (rebased onto `release/v3.0`).

## Single-snapshot reads (EN-1408 follow-up)

Sweeps the remaining read paths that stitch a `LedgerInfo` read together with per-entity attribute reads so they all read from **one Pebble snapshot** — the canonical pattern from the GetLedger/ListLedgers fix.

Converted (one `NewReadHandle()` up front, all reads routed through it): `GetTransactionFrom`, `GetAccount`, `GetLedgerStats`, `AnalyzeAccounts`, `AggregateVolumes`, `InspectIndex`, `ListLogs` (also fixes an inverted handle), `ListPreparedQueries` (also `NewDirectReadHandle` → `NewReadHandle`), `GetNumscript`, `ListNumscripts`.

This is defense-in-depth: none of these combined two snapshots into an observable inconsistency today (the second read only supplied the immutable ledger name), so there's no new red e2e test — the existing `get_ledger_snapshot_test.go` remains the coverage. Pinning makes a future change that embeds `LedgerInfo` content safe by construction. Left intentionally unchanged: `GetMetadataSchemaStatus` (single read), `GetIndexStatus` / `ListIndexes` (intentional cross-store / streaming reads).

## Transaction receipt correctness on routed reads

The receipt for `GetTransaction` is computed from the transaction's creation log. Three fixes (driven by review):

1. **Open the receipt snapshot after the transaction read** — opening it before `ctrl.GetTransaction`'s routed/linearizable read could pin a state predating the transaction's creation log → empty receipt for an existing tx.
2. **Reuse the serving node's receipt for forwarded reads** — when the read is routed to the leader, re-deriving the receipt from a (possibly-lagging) local snapshot is wrong. `BucketGrpcClient` was discarding `resp.Receipt`; now it's surfaced.
3. **Authoritative-or-absent receipt** — `Controller.GetTransaction` returns `*string`: `nil` for a locally-served read (the gRPC layer signs from the post-read snapshot), non-nil and authoritative (possibly empty, e.g. a reversal) when forwarded — used as-is, never re-derived. Internal plumbing only; the wire `receipt` field stays a `string`.

Mocks regenerated for the interface change.


internal/adapter/grpc/server_bucket.go
```diff
@@ -339,12 +342,33 @@ func (impl *BucketServiceServerImpl) GetTransaction(ctx context.Context, req *se
 
 	resp := &servicepb.GetTransactionResponse{Transaction: tx}
 	if impl.receiptSigner != nil {
```
NumaryBot: 🟠 **[major] Preserve forwarded receipts even without local signer**

When this node has receipt signing disabled but routes `GetTransaction` to a leader that does sign receipts, `fwdReceipt` contains the leader's authoritative receipt but this outer `impl.receiptSigner != nil` guard skips the whole block and drops it. That makes receipt presence depend on which node the client contacts during heterogeneous config/rollout; the forwarded receipt should be copied before deciding whether the local node can derive one.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1445#issuecomment-4834080337

internal/adapter/grpc/server_bucket.go
Azorlogh: https://github.com/formancehq/ledger/commit/5c9f5c51ab39c804874b57287c67921b53f078e8

internal/adapter/grpc/server_bucket.go
```diff
@@ -329,16 +333,36 @@ func (impl *BucketServiceServerImpl) GetTransaction(ctx context.Context, req *se
...
+		tx, fwdReceipt, err = impl.ctrl.GetTransaction(ctx, req.GetLedger(), req.GetTransactionId())
 	}
 
 	if err != nil {
 		return nil, err
 	}
 
 	resp := &servicepb.GetTransactionResponse{Transaction: tx}
-	if impl.receiptSigner != nil {
+	switch {
```
NumaryBot, line 344: 🟠 **[major] Forwarded receipts dropped when local signer is disabled**

When this node has receipt signing disabled but routes `GetTransaction` to a leader that does sign receipts, `fwdReceipt` contains the leader's authoritative receipt. The outer `impl.receiptSigner != nil` guard skips the entire block, silently dropping the forwarded receipt. This makes receipt presence dependent on which node the client happens to contact, causing inconsistent responses during heterogeneous config or rolling deployments.

**Suggestion:** Copy the forwarded receipt into the response before the `impl.receiptSigner != nil` check so it is always preserved. Only the local signing logic should be gated on `receiptSigner != nil`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1445#issuecomment-4834080337

gfyrag (approved):
LGTM. Les fixes de receipt sont réels (Bug A: receipt forwardé jeté; Bug D: gate signer trop englobante) et le tri-state `*string` encode proprement l'autorité vs l'absence. La sweep single-snapshot est du hardening défendable — pas de bug observable aujourd'hui mais by-construction safe pour toute future lecture de champ mutable de `LedgerInfo`.

Deux remarques non bloquantes:

1. **Coverage gap sur les receipts.** Aucun test rouge ne couvre les bugs corrigés côté serveur:
   - Follower en retard forward au leader et surface le receipt (Bug A).
   - Node sans signer relaye un receipt de leader signeur (Bug D).
   - Reversal forwardé garde `receipt=""` non-nil (Bug C, tri-state).
   
   Seul `client_bucket_test.go` valide le round-trip côté client. `codecov/patch` a échoué. Ces bugs ont été trouvés par NumaryBot, pas par test — un futur rework du routing pourrait les réintroduire silencieusement. Un test d'intégration deux-nodes (follower lag + `Consistency: leader`) rendrait la zone stable.

2. **Sweep sans garde-fou.** 10 handlers modifiés pour un pattern qu'il faudrait maintenant tenir. Serait utile de documenter dans `docs/technical/architecture/` la règle « toute lecture qui stitch `LedgerInfo` avec un attribut ouvre UN handle », voire un lint qui interdit `GetLedgerByName(ctx, ctrl.store, …)` en dehors d'exceptions listées (`GetMetadataSchemaStatus`, `GetIndexStatus`, `ListIndexes`).

Azorlogh:
Bug D now has a red-verified e2e test (`get_transaction_receipt_forwarded_test.go`), and the single-snapshot rule is documented in `core/read-snapshot-consistency.md`.

The other two can't get a meaningful red test: **Bug A** needs deterministic follower lag, but the test gateway isn't in the Raft path (nodes talk on real ports), and **Bug C** isn't observable — a forwarded reversal returns `receipt=""` whether it's relayed or re-derived locally.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1445#issuecomment-4834080337

PR #1446, opened 2026-06-29
feat(EN-1339): async secondary index over audit zone

paul-nicolas:
## Summary

Builds an async secondary index over the **Audit zone** so audit listing can eventually move onto the same index-backed filtering path as `ListTransactions` / `ListAccounts` / `ListLogs`, replacing the unbounded scan-time predicate proposed in #558.

Jira: https://formance-team.atlassian.net/browse/EN-1339

A new `auditindexer.Indexer` worker runs on every node, reads `AuditEntry` + `AuditItem` from the main store (read-only, off the hot path), and maintains a composite-key index in the per-replica `readstore`. A local `last_indexed_audit_sequence` cursor makes it resumable; rebuild = drop the keyspace + reset the cursor.

## Key decisions

- **Standalone always-on worker** (not the bucket-scoped Index registry): registering a `commonpb.Index` would be an FSM-apply change and pull the index into `checker.compareIndexes`, both of which the ticket forbids. The audit index fields are fixed/built-in, so the registry's `BUILDING→READY` lifecycle buys nothing.
- **Lives in `readstore`**, namespaced under `PrefixInternal` (`0xFE`) so the readstore comparer treats keys as non-splittable — no comparer change needed. This is a **deliberate deviation** from the ticket's literal "`ZoneCold` sub-zone" wording: the main store is Raft-replicated (invariant #4 forbids non-FSM writes) and ships in every snapshot, whereas `readstore` is the existing per-replica derived-index home and is where the `EntityIterator` machinery EN-1305 will reuse lives.
- **Pure access path, empty values** — not a projection, not hash-bound, no checker pass (invariant #8). Dangling entries after chapter purge are tolerated (reader skips on `ErrNotFound`) and reclaimed by drop+rebuild.

## Fields indexed

`outcome`, `ledger` (match-any), `caller.subject`, `order_type` (per-item, deduped), `timestamp` (range), `proposal_id` (range), `log_seq` (range, match-any). Order-type tokens are defined once in `domain.AuditOrderType` as the contract shared with EN-1305's filter DSL.

## Out of scope (EN-1305)

Wiring `ListAuditEntries` / `query.Compile` to read the index. This PR ships the index, the worker, the rebuild path, and seek+range read helpers — but does not change any HTTP/gRPC endpoint.

## Rebuild

- `ledgerctl store rebuild-audit-index` (offline drop + replay).
- Boot auto-rebuild when the cursor is missing or `--audit-index-rebuild-threshold` behind.

## Invariants honored

No change to `internal/application/check/checker.go`, the FSM apply path, `preload`, or `dal.WriteSession` (verified by diff). Confirmed via `git diff --name-only origin/main..HEAD`.

## Testing

- Unit: key encoding/ordering, comparer-safety, field extraction (multi-ledger, multi-order, failure entries, nil caller, log_seq=0 skip), full order-type vocabulary (39 oneof arms covered).
- Integration: catch-up after apply, restart resumes from cursor, rebuild-from-scratch yields a byte-identical index, convergence under sustained load.
- `just pre-commit` (idempotent, 0 lint issues) and `just test` green; `audit_index.*` metrics registered in the dashboard registry.

## Design docs

Spec and implementation plan are included under `docs/superpowers/` for reviewer context.

---
_Migrated from formancehq/ledger-v3-poc#596 onto `release/v3.0`._


paul-nicolas (commented):
## Multi-model PR review (Claude + Codex)

Two independent reviews (Claude in-repo + Codex external), synthesized and verified against the working tree.

**Verdict:** Design is sound — the worker is genuinely off the FSM hot path (read-only `NewDirectReadHandle()`), writes go only to `readstore` (invariants #3/#4 hold, no main-store writes), keys are comparer-safe under `PrefixInternal` (0xFE, non-splittable), the projection is correctly excluded from `checker.go` as a rebuildable pure-access path (invariant #8), and steady-state cursor advance is crash-atomic (index keys + cursor in one batch). The substantive risk is concentrated in the boot/rebuild path.

**Severity counts:** 1 Major, 1 Medium (both posted inline); 7 Minor + 2 Nit (listed below, not posted).

**Posted inline:** 1 Major (Rebuild drop/cursor non-atomicity → permanently empty index), 1 Medium (timestamp unit comment says nanos, stores HLC micros).

**Conflicts / rejected findings:** Codex proposed (Major) that the worker validate `len(items) == entry.OrderCount` and refuse to advance otherwise. **Rejected** after verification: `machine.go:1403` sets `entry.Items = nil` and items live as separate `SubColdAuditItem` records that `archiver.go` purges/archives independently, so `len(items) < OrderCount` is an *expected* post-archival state — enforcing equality would permanently wedge the indexer (head-of-line block) and contradicts the PR's documented "dangling tolerated" design.

- `internal/application/auditindexer/indexer.go:257` — no initial `ProcessOnce` before the first 200ms tick, and every idle tick reopens a read handle (minor cold-start lag / per-tick overhead). Consider one `ProcessOnce` before the ticker.
- `internal/storage/readstore/keys.go:555` — the `AuditIndex*Key` builders end in `Build()` (allocates a copy) where `Consume()` would do; pebble's `batch.Set` re-copies the key, so it is a wasted alloc per emitted key on rebuild.
- `internal/domain/audit_order_type.go:14` — `default: return "unknown"` silently buckets any new `Order` oneof arm; add an exhaustiveness guard / loud signal so a forgotten arm isn't an invisible indexing gap (relates to invariant #7).
- `internal/application/auditindexer/index_entry.go:36` — nil/unknown `Outcome` silently classifies as failure (0); per invariant #7 a should-not-happen nil outcome on a persisted entry should fail loud rather than default (Codex).
- `internal/application/auditindexer/indexer.go:30` — stale comment: `RebuildThreshold is reserved for future use`, but it is wired (`--audit-index-rebuild-threshold`), used in `shouldRebuildOnBoot`, and tested.
- `internal/application/auditindexer/indexer.go:234` — disabled indexer registers no metrics, so dashboards can't distinguish "disabled" from "stuck at 0"; optionally emit a constant gauge/log on disable.
- `internal/application/auditindexer/indexer_test.go:151` — `idx2.ProcessOnce(ctx)` uses the original ctx (not ctx2), and the resume test never exercises the `loop`/`shouldRebuildOnBoot` boot-rebuild path — the riskiest logic (see Major) has no integration coverage.
- `internal/storage/readstore/audit_index.go:42` — audit-index (0x05) and progress cursor (0x06) sub-prefixes are adjacent; `DropAuditIndex` relies on the exclusive upper bound to preserve the cursor — add a drop test asserting `ReadAuditProgress()` is unchanged after drop.
- `docs/ops/cli.md:2153` — "Reports the last processed audit sequence on completion", but `runRebuildAuditIndex` only prints "Audit index rebuild complete"; either print the sequence or drop step 4 (Codex).

internal/application/auditindexer/indexer.go
```diff
@@ -0,0 +1,317 @@
...
+
+	i.lastIndexed.Store(cursor)
+
+	return cursor, nil
+}
+
+// Rebuild drops the audit index and the cursor, then replays from the earliest
+// surviving audit entry. Used by ledgerctl and by boot auto-rebuild.
+func (i *Indexer) Rebuild(ctx context.Context) error {
+	if err := i.readStore.DropAuditIndex(); err != nil {
```
paul-nicolas: [Major] Rebuild's index-drop and cursor-reset are not crash-atomic — a torn write can permanently empty the audit index with no self-heal

`Rebuild` commits in three steps: `DropAuditIndex()` (its own batch+commit), then `WriteAuditProgress(batch, 0)` in a *second* batch+commit, then `ProcessOnce`. All commits are `pebble.NoSync`, so a power-loss / torn WAL between the drop commit and the cursor-reset commit leaves the index emptied while the cursor still holds its old high value C.

On the next boot, `shouldRebuildOnBoot(C, last=C)` returns false (cursor != 0, gap 0), so no rebuild fires; `loop` (lines 264-272) then falls through to steady-state `ProcessOnce`, which reads from cursor C, finds nothing after it, and the index stays permanently empty. Nothing self-heals it — this projection is deliberately outside `checker.go` (invariant #8) — and `loop` evaluates `shouldRebuildOnBoot` only once at startup, so even a failed/interrupted `Rebuild` is followed by incremental catch-up over a dropped index rather than a retry.

Fix: fold the drop and the cursor reset into a single `*dal.WriteSession` — `DeleteRange(AuditIndexPrefix(), prefixUpperBound(...))` + `WriteAuditProgress(batch, 0)` — committed once. Then a crash leaves either (old index, old cursor) or (empty index, cursor 0), and the latter deterministically re-triggers boot rebuild. Optionally also have `loop` retry rebuild on the ticker instead of silently dropping into incremental mode when `Rebuild` returns an error.

_Flagged independently by both reviewers (Claude + Codex)._

internal/storage/readstore/keys.go
```diff
@@ -33,6 +33,24 @@ const (
...
+	SubInternalAuditProgress byte = 0x06
+)
+
+// AuditField discriminates the indexed field within the audit-index keyspace.
+const (
+	AuditFieldOutcome       byte = 0x01 // 1 byte value: 0=failure, 1=success
+	AuditFieldLedger        byte = 0x02 // string value (match-any over AuditEntry.Ledgers)
+	AuditFieldCallerSubject byte = 0x03 // string value
+	AuditFieldOrderType     byte = 0x04 // string token (match-any over items)
+	AuditFieldTimestamp     byte = 0x05 // BE uint64 unix nanos (range)
```
paul-nicolas: [Medium] `AuditFieldTimestamp` is documented as "unix nanos" but stores HLC microseconds — a cross-file contract inconsistency that will mislead EN-1305's range filter

The writer stores `entry.GetTimestamp().GetData()` verbatim, and `commonpb.Timestamp.Data` is an HLC value in **microseconds** (`internal/domain/analysis/compact.go:28` — "microseconds since epoch (from commonpb.Timestamp.Data)"; `idempotency_store.go:40`). But this comment says `BE uint64 unix nanos (range)`, while `index_entry.go:27,57` correctly says microseconds and the design-spec key-layout table also says "unix nanos".

Round-trips inside this PR are internally consistent (same unit in and out), so nothing breaks today — but this field is the contract EN-1305's filter DSL builds range bounds against, and the wrong unit yields 1000x-off bounds. Fix: change the comment here (and the spec table) to "raw HLC `Timestamp.Data` (unix microseconds), BE uint64" so the three sources agree.

internal/application/auditindexer/indexer.go
paul-nicolas: Fixed in a0eb5a060 — `Rebuild` now stages the index `DeleteRange` and the cursor reset into a single `*dal.WriteSession` (new `DropAuditIndexInBatch`) committed once. A torn write now leaves either (old index, old cursor) or (empty index, cursor 0); the latter deterministically re-triggers boot rebuild and steady-state `ProcessOnce` from cursor 0 self-heals, so the index can no longer be left permanently empty with a stale high cursor. Added boot-rebuild path coverage (`TestBootRebuildOnStaleCursor`) plus atomic drop+reset and drop-preserves-cursor tests. The optional loop-retry-on-ticker is unnecessary given the atomic reset makes recovery self-healing.

internal/storage/readstore/keys.go
paul-nicolas: Fixed in d0b246ac3 — corrected the unit to `raw HLC Timestamp.Data (unix microseconds)` in `keys.go` (and the design-spec/plan layout tables) so all sources agree with the writer (`ts.GetData()`) and `index_entry.go`. No behavioral change (round-trips were already internally consistent); this aligns the contract EN-1305 builds range bounds against.

internal/application/auditindexer/indexer.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+	defer func() { _ = handle.Close() }()
+
+	cur, err := query.ReadAuditEntries(ctx, handle, &after)
+	if err != nil {
+		return after, false, fmt.Errorf("reading audit entries after %d: %w", after, err)
+	}
+	defer func() { _ = cur.Close() }()
+
+	batch := i.readStore.NewBatch()
```
NumaryBot, line 201: 🟠 **[major] Pebble write session leaked when audit index is already up-to-date**

On every 200 ms tick, a `dal.WriteSession` (wrapping a Pebble batch) is allocated before the indexer checks whether there is any new data. When `count == 0` (idle/caught-up state), the function returns without calling `Cancel()` on the session. Because `dal.WriteSession` requires cancellation for uncommitted sessions, this continuously leaks Pebble batch resources on idle nodes, which can degrade or crash a long-running server.

**Suggestion:** Defer a `session.Cancel()` immediately after allocating the write session (the successful commit path will supersede it), or restructure the code so the write session is only created after at least one entry has been confirmed to exist.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1446#issuecomment-4843607709

internal/application/auditindexer/indexer.go
paul-nicolas: Fixed in 329e03e72 — `processBatch` now defers `batch.Cancel()` immediately after `NewBatch()`, so the idle-tick `count == 0` path (and every read/build error path) releases the underlying Pebble batch. `Cancel()` is a no-op once the batch is committed, so the success path is unchanged. Applied the same guard to `Rebuild` and `Store.DropAuditIndex`, which leaked the batch on their error branches.

Accuracy note: a GC-reclaimed non-indexed batch is eventually freed, so this is allocation churn (~1 batch/TickInterval on a caught-up node) plus a `WriteSession`-contract violation rather than a hard crash — but the fix is warranted regardless. `just pre-commit` (0 issues) and `just test` are green.

internal/application/auditindexer/indexer.go
```diff
@@ -0,0 +1,330 @@
...
```
NumaryBot, line 201: 🟠 **[major] Pebble write session leaked when audit index is already up-to-date**

On every 200 ms tick, a `dal.WriteSession` (wrapping a Pebble batch) is allocated before the indexer checks whether there is any new data. When `count == 0` (idle/caught-up state), the function returns without calling `Cancel()` on the session. Because `dal.WriteSession` requires explicit cancellation for uncommitted sessions, this continuously leaks Pebble batch resources on idle nodes, which can degrade or eventually crash a long-running server.

**Suggestion:** Defer a `session.Cancel()` immediately after allocating the write session (the successful commit path will supersede/no-op it), or restructure the code so the write session is only created after confirming that at least one entry exists to process.

internal/application/auditindexer/indexer.go
```diff
@@ -0,0 +1,330 @@
...
+
+		return cursor, nil
+	}
+
+	cursor, err := i.readStore.ReadAuditProgress()
+	if err != nil {
+		return 0, fmt.Errorf("reading audit progress: %w", err)
+	}
+
+	for {
```
NumaryBot, line 108: 🟠 **[major] Shutdown cancellation ignored while draining audit backlog**

The batch-processing loop never checks `ctx.Err()` between iterations. If the worker context is cancelled during shutdown while a large backlog exists (or while writes keep arriving), the loop continues calling `processBatch` until fully caught up, blocking graceful shutdown. Because `worker.FxHook` waits for `Stop()` to return and this hook stops before gRPC/Raft hooks, the shutdown can hang for a long time or indefinitely under sustained writes.

**Suggestion:** Add a `ctx.Err()` check (or a `select` on `ctx.Done()`) at the top of each loop iteration and return early with the context error when cancellation is detected.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1446#issuecomment-4843607709

internal/application/auditindexer/indexer.go
paul-nicolas: Fixed in a12379f52 — `ProcessOnce` now checks `ctx.Err()` at the top of the drain loop and returns early on cancellation, so `worker.Stop()` (which blocks on the loop goroutine returning) no longer waits out a full backlog drain / sustained-write stream during shutdown. The now-expected `context.Canceled` is suppressed in the loop's steady-state and boot-rebuild error logs. Added `TestProcessOnceHonorsContextCancellation`. `just pre-commit` (0 issues) and `just test` green.

internal/application/auditindexer/indexer.go
paul-nicolas: Already fixed in 329e03e72 — `processBatch` defers `batch.Cancel()` at indexer.go:186, immediately after `i.readStore.NewBatch()`, so the `count == 0` idle-tick path (and every read/build error path) releases the Pebble batch. This is a re-post of the leak resolved on the earlier thread; no leak remains. Closing as already addressed.

cmd/ledgerctl/store/rebuild_audit_index.go
```diff
@@ -0,0 +1,95 @@
...
+
+	if readIndexDir == "" {
+		readIndexDir = filepath.Join(dataDir, "read-indexes")
+	}
+
+	logger := logging.NopZap()
+
+	spinner, _ := pterm.DefaultSpinner.Start("Opening Pebble store (read-only)...")
+
+	pebbleStore, err := dal.OpenReadOnly(dataDir, logger)
```
NumaryBot: 🟠 **[major] Offline rebuild command may open the wrong Pebble directory**

When the command is invoked with the server's `--data-dir` (e.g. `./data`), the live Pebble database actually lives under `<data-dir>/live` and the read index under `<data-dir>/read-indexes`. Opening `dataDir` directly either fails (it is not a Pebble DB) or, if users pass `./data/live` instead, derives the read-index path as `./data/live/read-indexes` rather than the server's actual read-store location. This breaks the advertised recovery path or silently writes to the wrong location.

**Suggestion:** Open the main store from `filepath.Join(dataDir, "live")` and derive the read-index directory from the data root (`filepath.Join(dataDir, "read-indexes")`), mirroring the server's own directory layout.

internal/storage/readstore/keys.go
```diff
@@ -516,3 +534,54 @@ func IndexVersionStatePrefix() []byte {
...
+}
+
+// AuditIndexStringKey builds [0xFE][0x05][field][value\x00][seq BE8] for a
+// string-valued field (ledger, caller_subject, order_type).
+func AuditIndexStringKey(kb *dal.KeyBuilder, field byte, value string, seq uint64) []byte {
+	return kb.Reset().
+		PutByte(PrefixInternal).
+		PutByte(SubInternalAuditIndex).
+		PutByte(field).
+		PutStringNull(value).
```
NumaryBot, line 569: 🟡 **[minor] NUL-byte in caller subject causes false-positive audit index matches**

String values for audit index keys (e.g. `caller.subject`) are NUL-terminated and matched via prefix scan. A value like `alice\x00evil` shares the same prefix that `AuditSeqsByString(..., "alice")` scans, so exact-match lookups for `"alice"` would also return entries indexed under `"alice\x00evil"`. This is latent today but becomes a real defect once caller-subject range filtering (EN-1305) is wired up.

**Suggestion:** Either escape NUL bytes within string values before encoding (e.g. length-prefix encoding) or validate/reject strings containing NUL bytes at indexing time to ensure prefix scans remain unambiguous.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1446#issuecomment-4843607709

cmd/ledgerctl/store/rebuild_audit_index.go
paul-nicolas: The premise does not hold for this codebase: the server opens the main Pebble store directly at `--data-dir` (`internal/bootstrap/module.go:240` → `dal.NewStore(cfg.DataDir, …)`) and the read index at `filepath.Join(cfg.DataDir, "read-indexes")` (`module.go:606`). There is no `<data-dir>/live` layer. This command mirrors that exactly and is identical in directory handling to the already-merged `rebuild-indexes` command (`cmd/ledgerctl/store/rebuild_indexes.go:60,74`). Opening `<data-dir>/live` would point at a non-existent path and break the command. Resolving as not applicable.

internal/storage/readstore/keys.go
paul-nicolas: Acknowledged — the prefix scan over `[field][value\x00]` is unambiguous only while indexed values are NUL-free, which holds today (order_type is a fixed vocabulary, ledger names are validated, caller.subject is an auth subject). The read helpers are not wired to any endpoint in this PR; that is EN-1305. Rejecting at index time would head-of-line-block the indexer on a poisoned entry, and changing the encoding now would diverge from the codebase-wide `PutStringNull` idiom and a not-yet-existent read path. Documented the constraint on `AuditIndexStringKey` and `AuditSeqsByString` in 01c217681 so EN-1305 disambiguates (exact-length check `len(key) == len(prefix)+8`, or length-prefixed encoding) before exposing arbitrary caller.subject filters. No behavioral change here.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1446#issuecomment-4843607709

cmd/ledgerctl/store/rebuild_audit_index.go
NumaryBot: 🟠 **[major] Offline rebuild command opens the wrong Pebble directory**

When the command is invoked with the server's `--data-dir` (e.g. `./data`), the live Pebble database actually lives under `<data-dir>/live` and the read index under `<data-dir>/read-indexes`. Opening `dataDir` directly either fails (it is not a Pebble DB root) or, if users pass `./data/live`, derives the read-index path as `./data/live/read-indexes` rather than the server's actual read-store location. This breaks the advertised recovery path or silently writes to the wrong location. This finding was raised in prior discussion and has not been marked as resolved.

**Suggestion:** Open the main store from `filepath.Join(dataDir, "live")` and derive the read-index directory from the data root (`filepath.Join(dataDir, "read-indexes")`), mirroring the server's own directory layout.

internal/application/auditindexer/indexer.go
```diff
@@ -0,0 +1,338 @@
...
+
+	_, err := i.ProcessOnce(ctx)
+
+	return err
+}
+
+// shouldRebuildOnBoot reports whether boot should drop+rebuild instead of an
+// incremental catch-up: cursor missing (0) with entries present, or the gap
+// exceeds the configured threshold.
+func (i *Indexer) shouldRebuildOnBoot(cursor, last uint64) bool {
```
NumaryBot, line 162: 🟠 **[major] Stale audit cursor after main-store restore leaves audit entries unindexed**

If the main Pebble store is restored to an older checkpoint while the read-index directory is retained, the persisted audit cursor can be greater than the current last audit sequence. In that case neither boot-rebuild condition (`cursor == 0` nor `cursor > last`) is matched, and `ProcessOnce` starts scanning after the stale high cursor. All surviving audit entries whose sequence numbers are below the old cursor value are never re-indexed, and the index stays permanently incomplete until the store eventually advances past the old cursor value. This scenario is not covered by the existing `shouldRebuildOnBoot` logic or the `TestBootRebuildOnStaleCursor` test added in follow-up commits.

**Suggestion:** Add a `cursor > last` (or `cursor > currentMax`) check to `shouldRebuildOnBoot` so that a restored/rolled-back store deterministically triggers a full rebuild, resetting the cursor to 0.

internal/storage/readstore/keys.go
```diff
@@ -516,3 +534,63 @@ func IndexVersionStatePrefix() []byte {
...
+// equality/range filter path over arbitrary caller subjects, MUST disambiguate
+// before relying on it (an exact-length check len(key) == len(prefix)+8, or a
+// length-prefixed string encoding); otherwise an "alice" lookup would also
+// match a value indexed as "alice\x00evil".
+func AuditIndexStringKey(kb *dal.KeyBuilder, field byte, value string, seq uint64) []byte {
+	return kb.Reset().
+		PutByte(PrefixInternal).
+		PutByte(SubInternalAuditIndex).
+		PutByte(field).
+		PutStringNull(value).
```
NumaryBot, line 569: 🟡 **[minor] NUL byte in caller subject causes false-positive audit index matches**

String values for audit index keys (e.g. `caller.subject`) are NUL-terminated and matched via prefix scan. A value like `alice\x00evil` shares the same NUL-terminated prefix that `AuditSeqsByString(..., "alice")` scans, so exact-match lookups for `"alice"` would also return entries indexed under `"alice\x00evil"`. This is latent today but becomes a real defect once caller-subject range filtering (EN-1305) is wired up.

**Suggestion:** Either escape NUL bytes within string values before encoding (e.g. using length-prefix encoding) or validate and reject strings containing NUL bytes at indexing time, ensuring prefix scans remain unambiguous.

cmd/ledgerctl/store/rebuild_audit_index.go
paul-nicolas: You and codex are right, and my earlier rejection of this was wrong — apologies. I had grepped for the string "live" and missed that the subdir is the `liveDir = "live"` constant (`internal/storage/dal/store.go:32`): `dal.NewStore` opens `filepath.Join(dataDir, "live")` (store.go:390,436), whereas `dal.OpenReadOnly` opens the path directly (store_readonly.go:30), so passing the documented `--data-dir ./data` opened the bare root and errored under `ReadOnly:true`.

Fixed in 6b9095526 — both `rebuild-audit-index` and the pre-existing `rebuild-indexes` (same defect) now open `filepath.Join(dataDir, "live")`; the read-index default `<data-dir>/read-indexes` already matched the server. The documented `--data-dir ./data` usage now works.

internal/application/auditindexer/indexer.go
paul-nicolas: Fixed in f941e3628 — added a `cursor > last` branch to `shouldRebuildOnBoot` that forces a full rebuild (resetting the cursor to 0). In normal operation the cursor only advances to sequences that exist, so `cursor > last` is reachable only after a restore/rollback to an earlier checkpoint with the read index retained — exactly the scenario you describe. Covered by new assertions in `TestShouldRebuildOnBoot` (`shouldRebuildOnBoot(10, 5)` and `(10, 0)`).

internal/storage/readstore/keys.go
paul-nicolas: Already addressed in 01c217681 (prior review round): the NUL-prefix constraint is documented on `AuditIndexStringKey` and `AuditSeqsByString`, with the read-path disambiguation deferred to EN-1305 (the read helpers are unwired in this PR and today's indexed values — order_type, ledger, caller.subject — are NUL-free). No further change here.

internal/application/auditindexer/indexer.go
```diff
@@ -0,0 +1,345 @@
...
```
NumaryBot, line 162: 🟠 **[major] Stale audit cursor after main-store restore leaves audit entries permanently unindexed**

If the main Pebble store is restored to an older checkpoint while the read-index directory is retained, the persisted audit cursor can be greater than the current last audit sequence. In that case neither boot-rebuild condition (`cursor == 0` nor `cursor > last`) is matched, so `ProcessOnce` starts scanning after the stale high cursor. All surviving audit entries whose sequence numbers fall below the old cursor value are never re-indexed, leaving the audit index permanently incomplete until the store advances past the old cursor — which may never happen.

**Suggestion:** Add a `cursor > last` (or `cursor > currentMax`) check to `shouldRebuildOnBoot` so that a restored/rolled-back store deterministically triggers a full rebuild, resetting the cursor to 0. This scenario and the existing `TestBootRebuildOnStaleCursor` test do not cover the restore case.

internal/storage/readstore/audit_index.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	defer func() { _ = iter.Close() }()
+
+	var seqs []uint64
+	for iter.First(); iter.Valid(); iter.Next() {
+		k := iter.Key()
+		if len(k) < 8 {
+			return nil, fmt.Errorf("audit index key too short: %d", len(k))
+		}
+		seqs = append(seqs, binary.BigEndian.Uint64(k[len(k)-8:]))
```
NumaryBot: 🟠 **[major] AuditSeqsByUint64Range returns duplicate audit sequences for multi-item entries**

When `AuditSeqsByUint64Range` is used for `AuditFieldLogSeq`, a single audit entry can have multiple audit items whose log sequences all fall within the requested range. The index stores a separate key per item, each ending in the same audit sequence. The loop appends each key's trailing sequence without deduplication, so a log-seq range filter returns the same audit entry multiple times. This affects the correctness of the read helper that future filtering (EN-1305) will consume.

**Suggestion:** Deduplicate audit sequences before returning from `AuditSeqsByUint64Range` (e.g., use a seen-set or compare the current sequence to the previously appended one, which works because keys are sorted). The same deduplication should be applied in any other postings helpers that may emit duplicate sequences.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1446#issuecomment-4843607709

internal/storage/readstore/keys.go
NumaryBot, line 569: 🟡 **[minor] NUL byte in string index values causes false-positive prefix-scan matches**

String audit index keys (e.g. for `caller.subject`) are NUL-terminated and matched via prefix scan. A value such as `alice\x00evil` shares the NUL-terminated prefix that `AuditSeqsByString(..., "alice")` scans, so exact-match lookups for `"alice"` also return entries indexed under `"alice\x00evil"`. This is latent while indexed values are NUL-free, but becomes a real defect once arbitrary caller-subject filtering (EN-1305) is exposed. The constraint has been documented in follow-up commits but not enforced.

**Suggestion:** Either validate/reject strings containing NUL bytes at indexing time, or add an exact-length check (`len(key) == len(prefix)+8`) in the read helpers before including a result — as noted in the follow-up commit — so EN-1305 cannot expose this ambiguity.

internal/storage/readstore/audit_index.go
paul-nicolas: Fixed in 081fc5d2f — deduplicated in the shared `auditSeqsForPrefix` with a seen-set (covers `AuditSeqsByUint64Range`, `AuditSeqsByString`, `AuditSeqsByOutcome` in one place, since any match-any field can emit several keys with the same audit seq). The duplicates are non-adjacent — keys sort by value then seq, so the two same-seq keys can straddle a different seq — so a previous-value comparison would miss them; the seen-set handles it and preserves first-occurrence order. Covered by `TestAuditSeqsByUint64RangeDedupes` (seq 5 at log-seqs 100 and 200 with seq 6 at 150 in between → returns [5, 6]).

internal/application/auditindexer/indexer.go
paul-nicolas: Already fixed in f941e3628 (previous round, now on origin HEAD): `shouldRebuildOnBoot` has the `cursor > last` branch you suggest — it forces a full rebuild (resetting the cursor to 0) when the persisted cursor is ahead of the audit head, which is reachable only after a restore/rollback. Covered by `TestShouldRebuildOnBoot` (`shouldRebuildOnBoot(10, 5)` and `(10, 0)`). This finding was re-scanned against a pre-fix commit. Resolving as addressed.

internal/storage/readstore/keys.go
paul-nicolas: Already addressed in 01c217681: the NUL-prefix constraint is documented on `AuditIndexStringKey` and `AuditSeqsByString`, with read-path disambiguation deferred to EN-1305 (read helpers unwired in this PR; indexed values — order_type, ledger, caller.subject — are NUL-free today). No further change here.

internal/storage/readstore/audit_index.go
```diff
@@ -0,0 +1,184 @@
...
+// caller wants each matching entry once, so duplicates are collapsed. They are
+// not necessarily adjacent (keys sort by value then seq, so the same seq can
+// appear at different value positions), hence a seen-set rather than a
+// previous-value comparison. First-occurrence order is preserved.
+func (s *Store) auditSeqsForPrefix(lower, upper []byte) ([]uint64, error) {
+	iter, err := s.db.NewIter(&pebble.IterOptions{LowerBound: lower, UpperBound: upper})
+	if err != nil {
+		return nil, fmt.Errorf("creating audit index iterator: %w", err)
+	}
+	defer func() { _ = iter.Close() }()
```
NumaryBot, line 96: 🟠 **[major] AuditSeqsByUint64Range returns duplicate audit sequences for multi-item entries**

When AuditSeqsByUint64Range is used for AuditFieldLogSeq, a single audit entry can have multiple audit items whose log sequences all fall within the requested range. The index stores a separate key per item, each ending in the same audit sequence. The loop appends each key's trailing sequence without deduplication, so a log-seq range filter returns the same audit entry multiple times. This corrupts the postings list consumed by EN-1305 filtering.

**Suggestion:** Deduplicate audit sequences before returning from AuditSeqsByUint64Range (e.g., use a seen-set or compare the current sequence to the previously appended one, which works because keys are sorted). Apply the same deduplication in any other postings helpers that may emit duplicate sequences.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1446#issuecomment-4843607709

internal/storage/readstore/audit_index.go
paul-nicolas: Already fixed in 081fc5d2f (last review round, now on origin HEAD `081fc5d2f`) — the prior instance of this finding is the resolved thread above. `auditSeqsForPrefix` deduplicates with a seen-set, so all three read helpers (`AuditSeqsByUint64Range`, `AuditSeqsByString`, `AuditSeqsByOutcome`) return each audit sequence at most once:

```go
var seqs []uint64
seen := make(map[uint64]struct{})
for iter.First(); iter.Valid(); iter.Next() {
    k := iter.Key()
    if len(k) < 8 {
        return nil, fmt.Errorf("audit index key too short: %d", len(k))
    }
    seq := binary.BigEndian.Uint64(k[len(k)-8:])
    if _, ok := seen[seq]; ok {
        continue
    }
    seen[seq] = struct{}{}
    seqs = append(seqs, seq)
}
```

Covered by `TestAuditSeqsByUint64RangeDedupes` (seq 5 at log-seqs 100 and 200 with seq 6 at 150 in between → returns `[5, 6]`). The re-review evaluated a commit before this fix. Resolving as addressed.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1446#issuecomment-4843607709

internal/application/auditindexer/indexer.go
```diff
@@ -0,0 +1,345 @@
...
+
+	ticker := time.NewTicker(TickInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+		}
+		if last, err := i.lastAuditSequence(); err == nil {
```
NumaryBot: 🟠 **[major] Stale-cursor rebuild check not applied after runtime main-store restore**

The shouldRebuildOnBoot check (including the cursor > last branch added in f941e3628) is only evaluated once, before the ticker starts. If the main store is replaced at runtime via SynchronizeWithLeader or RestoreCheckpoint while the indexer worker is already running, the audit head can drop below the persisted cursor mid-flight. The running ticker loop then only calls ProcessOnce, which scans from the stale high cursor and never re-indexes surviving lower-sequence audit entries. The existing fix and its TestShouldRebuildOnBoot test cover the boot-time case but not this runtime scenario.

**Suggestion:** At the start of each ProcessOnce invocation (or in the tick loop before calling it), re-evaluate the cursor vs. current audit head and trigger a rebuild if cursor > last. Alternatively, expose a channel or hook from the restore path so the indexer worker can reset its cursor and re-enter the rebuild path without requiring a full process restart.

paul-nicolas: Fixed in 5bc51a518. The steady-state tick now re-checks the `cursor > last` rollback condition every iteration (new `processTick`) and rebuilds when a runtime restore (`SynchronizeWithLeader`/`RestoreCheckpoint`) has dropped the audit head below the persisted cursor — so surviving lower-sequence entries no longer wait for a process restart.

Notes on the approach:
- Kept it self-contained in the indexer instead of stopping/starting it from the synchronizer. Wiring the indexer into `SynchronizeWithLeader` would invert the state→application layer dependency and is unnecessary for safety: `NewDirectReadHandle` already holds `dbMu.RLock` for the handle lifetime, so the indexer is lock-safe across an in-place `RestoreCheckpoint`. The per-tick re-check is also trigger-agnostic (covers any runtime rollback, not just that one call site).
- Only the `cursor > last` branch is re-evaluated per tick, not the full `shouldRebuildOnBoot`: its `RebuildThreshold` branch is a boot-only heuristic that would spuriously full-rebuild whenever a normal burst exceeds the threshold between ticks. A shared `cursorAheadOfHead` predicate keeps boot and steady-state in sync.
- Added `TestProcessTickRebuildsWhenCursorAheadOfHead`, which reproduces the stranding (a bare `ProcessOnce` leaves entries 1..3 unindexed with the cursor stuck at 5) and asserts `processTick` self-heals (cursor reset to the surviving head, entries re-indexed).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1446#issuecomment-4843607709

gfyrag (approved):
LGTM.

Périmètre bien contenu (index + worker + rebuild, wiring reader laissé à EN-1305), invariants #3/#4/#8 préservés (checker.go, apply-path, preload, WriteSession non touchés — vérifié par diff), et les 7 fixes post-review couvrent les vrais pièges (rebuild atomique, WriteSession leak, ctx cancel, curseur en avance, drop dedup, live subdir).

Follow-up ouvert pour factoriser le squelette tail-worker maintenant qu'on a 2 consommateurs (indexbuilder / auditindexer) et un 3ᵉ (usage-builder) en cours d'impl → https://formance-team.atlassian.net/browse/EN-1427.

gfyrag:
Follow-up ticket for the tail-worker skeleton factorisation discussed off-PR:

**[EN-1427 — Factor tail-worker skeleton shared by indexbuilder / auditindexer / usage-builder](https://formance-team.atlassian.net/browse/EN-1427)**

Context: with this PR we now have 2 workers following the same shape (`indexbuilder.Builder` + `auditindexer.Indexer`), and a **3rd consumer is already in progress (usage-builder)** — which makes the extraction worthwhile rather than speculative.

Candidate bricks:
- `Uint64Cursor` (BE8 read/write into readstore — 4 cursors of this shape today)
- `TailWorker` (Start/Stop + worker.Worker + ticker/wake/ctx loop + drain-by-batches)
- `RegisterTailGauges` (`last_indexed / source_last / lag` triplet)

Out of scope for the refactor: BUILDING→READY lifecycle, IndexVersion state, backfill, schema-rewrite, versioned dual-write — builder-specific.

Nothing to do inside #1446, purely a follow-up.

PR #1447, opened 2026-06-29
feat(EN-1276): account-type default metadata via per-account existence marker (strict any-asset)

paul-nicolas:
## Summary

Implements account-type **default metadata** (issue #516 / EN-1276) using a **dedicated per-account existence marker**, so default metadata is applied to an account the **first time it is ever created (any asset)** — not per-asset.

This is the **strict "any-asset"** approach, intentionally kept as a separate branch from `EN-1276-account-type-default-metadata` (the lighter per-`(account,asset)` approach) so the two can be compared. This branch is **not** based on that one — it is a clean implementation off `main`.

> Draft: the create-with-defaults path is complete, tested, and gated; see *Scope / follow-ups* below.

## How it works

- New `AccountType.default_metadata` proto field.
- New first-class cache attribute **`Account`** (`SubAttrAccount = 0x0D`), keyed by `AccountKey`, holding a tiny `AccountState` marker. Wired end-to-end: attribute/cache/loader/bloom registration, preload `Needs`, `Scope.GetAccount/PutAccount`, registry/derived/WriteSet merge-flush, coverage gate.
- **Admission** declares the marker preload for each `CreateTransaction` account (`world` excluded).
- **FSM apply** (`processor_transaction.go`): for each non-`world` account touched for the first time (`GetAccount` → `ErrNotFound`), it writes the marker and merges the matching `AccountType.default_metadata` for keys the script/order didn't set (explicit wins). Defaults ride the existing audited `AccountMetadata` log path, so they're indexed and covered by the checker's existing `compareMetadata`.
- Gate is a **derived, deterministic** check (`ledgerHasAccountTypeDefaults`) — no stored status, replay-safe, and it short-circuits so non-feature ledgers see zero behaviour change.

## Invariants honoured

- **No Pebble reads on the FSM hot path** — newness is an O(1) gated cache read; all Pebble work stays in preload.
- **Determinism** — newness authoritative at apply; the marker is a deterministic function of the audit, reconstructed identically by replay + backup-rebuild.
- **Audit is source of truth (#8)** — the new marker projection is verified by a new checker pass `compareAccounts` (every live marker must correspond to an audit-referenced account → `CHECK_STORE_ERROR_TYPE_ACCOUNT_MISMATCH`), and reconstructed by backup/rebuild via the shared `replay.Writer.RecordAccount`.
- **Coverage gate (#9)** — reads go through `Scope.GetAccount`; admission declares the need.

## Tests

- `processor_transaction_defaults_test.go`: new-account defaulting, explicit-metadata-wins, existing-account-not-re-defaulted.
- All existing suites in touched packages pass (processing, admission, plan, state, cache, attributes, check, backup, replay).

## Gate

- `just pre-commit` → golangci-lint 0 issues (main + operator), `go generate` + `go mod tidy` clean.
- `just test` → green.

## Scope / follow-ups (per the ticket's open questions)

- **No automatic backfill of pre-existing accounts.** Adding `default_metadata` to an *already-populated* ledger and seeding its existing accounts needs an async marker-seeder (same reason the index-builder is async) + a stored `AccountDefaultsStatus` (enum already in proto, reserved). That is left as the ticket's proposed "separate explicit backfill order" — future work. The create-with-defaults path (the common case) is complete here.
- A dedicated `compareAccounts` forged-marker unit test in the checker harness is a nice-to-have follow-up; the pass is exercised indirectly by the existing checker suite.

Closes #516 (create-with-defaults path); backfill order tracked separately.

---
_Migrated from formancehq/ledger-v3-poc#564 onto `release/v3.0`._


internal/application/check/checker.go
```diff
@@ -1250,6 +1251,201 @@ func (c *Checker) compareMetadata(ctx context.Context, reader dal.PebbleReader,
...
+		emit("", "", "live account iterator error: %v", err)
+
+		return errorCount + 1
+	}
+
+	// Reverse: every replay-touched account must have a live marker. Markers are
+	// universal (written on every account-creation path), so a missing marker
+	// would let the FSM treat an existing account as new and re-apply default
+	// metadata on next touch.
+	for key := range replayTouched {
```
NumaryBot, line 1435: 🔴 **[blocker] Gate account-marker reverse checks on migration**

When `Check()` runs on a store containing audit logs written before this change, `ReplayLedgerLog` now puts every historical account into `replayTouched`, but the live `SubAttrAccount` column will be empty because older FSMs never wrote these markers. This loop therefore emits `ACCOUNT_MISMATCH` for otherwise healthy upgraded ledgers; seed the projection or version/activation-gate this reverse check before making it unconditional.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1447#issuecomment-4834394332

internal/domain/processing/processor_account_type.go
```diff
@@ -30,6 +30,16 @@ func processAddAccountType(ledger string, order *raftcmdpb.AddAccountTypeOrder,
 		return nil, &domain.ErrInvalidPattern{Pattern: at.GetPattern(), Details: err.Error()}
 	}
 
+	if err := validateDefaultMetadata(at); err != nil {
+		return nil, err
+	}
+
+	// EN-1276 allows attaching default_metadata at any time, including to a
```
NumaryBot, line 37: 🔴 **[blocker] Gate defaults until legacy accounts are marked**

This assumption only holds for accounts touched after the new marker-writing code is deployed. On an upgraded/populated ledger whose existing accounts have volumes or metadata but no `SubAttrAccount` rows, adding an account type with `default_metadata` now succeeds, and the next touch of one of those accounts will be treated as first creation and receive defaults, so legacy accounts are silently backfilled unless markers are seeded or this is rejected until migration.

PR #1448, opened 2026-06-29
feat(EN-1305): honor ListOptions.filter on the audit trail

paul-nicolas:
## Summary

`ListAuditEntries` advertised the full `ListOptions.filter` envelope on the wire but ignored it — the only server-side narrowing was the bespoke `failures_only` / `ledger` request fields. This PR makes the audit trail queryable server-side through the shared `QueryFilter` DSL, for a new `QUERY_TARGET_AUDIT`.

Jira: https://formance-team.atlassian.net/browse/EN-1305

Fixes #506

## Approach

The audit log is a sequential cold-zone scan with **no index**, so filtering is implemented as **scan-time predicates** (consistent with how `failures_only` / `ledger` already worked), not through the index-backed `query.Compile` path. The shared `QueryFilter` oneof gains a single grouped `AuditCondition` variant; a dedicated `query.CompileAuditPredicate` turns a filter into a `func(*AuditEntry, []*AuditItem) bool` applied during the scan.

The dedicated `failures_only` / `ledger` request fields are removed and folded into the filter; the CLI keeps `--failures-only` / `--ledger` as convenience shorthands that build the equivalent filter client-side.

## What changed

- **Proto**: `QUERY_TARGET_AUDIT`, `QueryFilter.audit` (`AuditCondition` = `AuditField` enum × string/uint/bool condition); `failures_only`/`ledger` reserved on `ListAuditEntriesRequest`.
- **Predicate compiler** (`internal/query/audit_filter.go`): header fields evaluated directly (sequence, proposal_id, timestamp, log_seq overlap, outcome, error_type, caller subject/scope/god, ledger — match-any for repeated fields); `order_type` lazily loads `AuditItem`s and matches the order payload variant via protoreflect, gated by `needsItems`. Reuses the existing half-open `resolveUintBounds`. Non-audit conditions and field/type mismatches → `InvalidArgument`.
- **Error-returning cursor** (`cursor.NewFilteredCursorE`): surfaces per-entry item-load errors instead of silently dropping them.
- **Filter grammar** (`internal/pkg/filterexpr`): `audit[field] op value` term + formatter round-trip.
- **Wiring**: `DefaultController.ListAuditEntries(ctx, afterSeq, pageSize, filter)`; gRPC handler passes `opts.GetFilter()` with `ListOptionsSupport{Filter: true}`; updated the other `Controller` implementers (`BucketGrpcClient`, `RoutedController`) and `pkg/actions/read.go`.
- **CLI**: `ledgerctl audit list --filter / --ledger / --failures-only`.
- **Docs**: `docs/ops/cli.md` (audit filter fields/operators, scan-time strategy note, `order_type` value vs. display clarification), `api-comparison.md`.

## Supported audit filter DSL

```
audit[outcome] == failure
audit[caller.subject] == "alice" and audit[ledger] == "main"
audit[order_type] in (apply, save_numscript)
audit[seq] between 1000 and 2000
```

Fields: `seq`, `proposal_id`, `timestamp`, `log_seq` (uint); `outcome`, `error_type`, `caller.subject`, `caller.scope`, `ledger`, `order_type` (string); `caller.god` (bool).

## Acceptance criteria

- [x] `ListAuditEntries` applies `options.filter`; unsupported conditions rejected with `InvalidArgument`.
- [x] Supported audit conditions documented (fields/operators).
- [x] Index/scan strategy noted (sequential cold-zone scan, scan-time predicates, lazy item load for `order_type`, no audit index).
- [x] Docs (`cli.md` / API comparison) and `ledgerctl audit list` flags updated.

## Testing

- Unit: `internal/query/audit_filter_test.go` (per-field semantics, composition, rejections, order-type match-any + `needsItems` gating), `internal/pkg/cursor` error-cursor, `filterexpr` parser + format round-trip.
- E2E (`tests/e2e/business/audit_test.go`): outcome / ledger / seq-range / order_type filtering and `InvalidArgument` rejection against a real server via the gRPC client.
- `just pre-commit` and `just test` green.

## Invariants

Purely read-side: no Pebble writes, no FSM/hot-path involvement, and **no new persisted projection** (Invariant #8) — nothing to add to the checker.

## Follow-ups (non-blocking, from review)

- Controller-level regression test asserting header-only filters do not call `ReadAuditItems` (`needsItems == false`).
- Consider aligning `order_type` values with the `--expand` display labels (currently matches the outer payload variant, e.g. `apply`; documented).

---
_Migrated from formancehq/ledger-v3-poc#558 onto `release/v3.0`._

### DIFF
--- a/misc/proto/bucket.proto
+++ b/misc/proto/bucket.proto
@@ -758,10 +758,10 @@ message CheckStoreProgress {
 
 message ListAuditEntriesRequest {
   common.ListOptions options = 1;
-  // failures_only filters audit entries to only those carrying a failure status.
-  bool failures_only = 2;
-  // ledger filters audit entries to only those containing orders targeting this ledger.
-  string ledger = 3;
+  // Field 2 (failures_only) and 3 (ledger) folded into options.filter
+  // (audit[outcome]==failure, audit[ledger]==X) — EN-1305.
+  reserved 2, 3;
+  reserved "failures_only", "ledger";
 }
 
 message GetAuditEntryRequest {
--- a/cmd/ledgerctl/audit/list.go
+++ b/cmd/ledgerctl/audit/list.go
@@ -141,11 +141,21 @@ func buildAuditFilter(expr string, failuresOnly bool, ledger string) (*commonpb.
 		parts = append(parts, f)
 	}
 	if ledger != "" {
-		f, err := filterexpr.Parse(fmt.Sprintf("audit[ledger] == %q", ledger))
-		if err != nil {
-			return nil, err
-		}
-		parts = append(parts, f)
+		// Build the proto directly instead of round-tripping through the DSL
+		// string. The filterexpr String grammar has no escape handling, so a
+		// ledger name containing a quote or backslash (both valid per
+		// domain.ValidateLedgerName) cannot survive
+		// Parse(fmt.Sprintf("audit[ledger] == %q", ledger)).
+		parts = append(parts, &commonpb.QueryFilter{
+			Filter: &commonpb.QueryFilter_Audit{Audit: &commonpb.AuditCondition{
+				Field: commonpb.AuditField_AUDIT_FIELD_LEDGER,
+				Condition: &commonpb.AuditCondition_StringCond{
+					StringCond: &commonpb.StringCondition{
+						Value: &commonpb.StringCondition_Hardcoded{Hardcoded: ledger},
+					},
+				},
+			}},
+		})
 	}
 
 	switch len(parts) {
--- a/cmd/ledgerctl/audit/list.go
+++ b/cmd/ledgerctl/audit/list.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/formancehq/ledger/v3/cmd/ledgerctl/cmdutil"
 	"github.com/formancehq/ledger/v3/internal/domain"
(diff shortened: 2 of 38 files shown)


misc/proto/bucket.proto
```diff
@@ -758,10 +758,10 @@ message CheckStoreProgress {
 
 message ListAuditEntriesRequest {
   common.ListOptions options = 1;
-  // failures_only filters audit entries to only those carrying a failure status.
-  bool failures_only = 2;
-  // ledger filters audit entries to only those containing orders targeting this ledger.
-  string ledger = 3;
+  // Field 2 (failures_only) and 3 (ledger) folded into options.filter
+  // (audit[outcome]==failure, audit[ledger]==X) — EN-1305.
+  reserved 2, 3;
```
NumaryBot, line 763: 🟠 **[major] Update antithesis audit request after removing ledger field**

When the antithesis workload module is built, `tests/antithesis/workload/bin/cmds/main/parallel_driver_bulk_atomicity/main.go` still constructs `servicepb.ListAuditEntriesRequest{Ledger: ledger, ...}`. Reserving/removing field 3 means the generated Go type no longer has `Ledger`, so that separate module no longer compiles; convert that caller to an `options.filter` audit ledger condition as well.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1448#issuecomment-4834388258

PR #1452, opened 2026-06-30
fix(test): use unique names in TestLedgersCreate to prevent flaky failures

ariel-formance:
## Summary

- `TestLedgersCreate` uses hardcoded bucket names (`bucket1`, `bucket2`) and ledger names (`ledger0`–`ledger29`), which collide across parallel test runs sharing the same Postgres instance.
- Two goroutines racing to create the same bucket both see `IsInitialized=false` and both attempt `CREATE SCHEMA`, causing `duplicate key value violates unique constraint "pg_namespace_nspname_index"`.
- Fix: use a UUID prefix for bucket and ledger names so each test run is isolated.
- Also adding some basic tests to increase coverage score for project. 

## Test plan

- [x] Verify the test compiles (`go build ./internal/storage/driver/...`)
- [ ] CI passes Tests check


internal/storage/driver/driver_test.go
```diff
@@ -34,7 +34,8 @@ func TestLedgersCreate(t *testing.T) {
 		systemstore.NewStoreFactory(),
 	)
 
-	buckets := []string{"bucket1", "bucket2"}
+	prefix := uuid.NewString()
+	buckets := []string{prefix + "-b1", prefix + "-b2"}
```
NumaryBot: 🔴 **[blocker] UUID hyphens in bucket names break PostgreSQL schema names**

The generated bucket names (e.g., `<uuid>-b1`) contain hyphens because `uuid.NewString()` produces strings like `550e8400-e29b-41d4-a716-446655440000`. In this codebase bucket names are used directly as PostgreSQL schema identifiers. Unquoted schema names with hyphens are invalid in PostgreSQL, so `CreateLedger` will fail during migrations before any assertion is reached, turning the flakiness fix into a permanent failure.

**Suggestion:** Strip hyphens from the UUID before building bucket and ledger names, e.g.:
```go
import "strings"
prefix := strings.ReplaceAll(uuid.NewString(), "-", "")
```
This also increases collision resistance to 128 bits, making the truncation noted in the prior review unnecessary.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1452#issuecomment-4841307262

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1452#issuecomment-4841307262

internal/storage/driver/driver_test.go
```diff
@@ -34,7 +36,8 @@ func TestLedgersCreate(t *testing.T) {
 		systemstore.NewStoreFactory(),
 	)
```
NumaryBot, line 27: 🔴 **[blocker] UUID hyphens in bucket names produce invalid PostgreSQL schema identifiers**
<sub>reported by coderabbitai[bot], NumaryBot</sub>

uuid.NewString() produces strings like '550e8400-e29b-41d4-a716-446655440000'. Because bucket names are used directly as PostgreSQL schema identifiers, the resulting names (e.g. '<uuid>-b1') are invalid as unquoted schema names in PostgreSQL. CreateLedger will fail during schema migrations before any test assertion is reached, converting a flaky test into a permanently failing one.

**Suggestion:** Strip hyphens from the UUID before constructing bucket and ledger names:
```go
import "strings"
prefix := strings.ReplaceAll(uuid.NewString(), "-", "")
buckets := []string{prefix + "b1", prefix + "b2"}
```
This also eliminates the truncation issue and raises collision resistance to 128 bits.

internal/storage/driver/driver_test.go
```diff
@@ -8,6 +8,8 @@ import (
 	"sync"
 	"testing"
 
+	"strings"
```
NumaryBot: 🟠 **[major] Import block formatting will cause CI 'Dirty' check to fail**

Adding 'strings' in a separate standard-library import group violates the repository's goimports/gci formatter configuration. The CI 'Dirty' job runs 'just pre-commit' with 'golangci-lint run --fix' and fails if the working tree is modified afterwards. A mis-grouped import will cause an automatic reformat and fail the CI check.

**Suggestion:** Place the 'strings' import inside the existing standard-library import group rather than in a separate block, so the file already matches what the formatter would produce.

internal/storage/driver/driver_test.go
NumaryBot, line 27: 🟡 **[minor] Truncating UUID to 8 characters leaves only 32 bits of entropy**
<sub>reported by coderabbitai[bot], NumaryBot</sub>

The current code takes only the first 8 hex characters of the UUID as the prefix, providing just 32 bits of entropy. On a long-lived or shared PostgreSQL instance this can still cause name collisions between test runs.

**Suggestion:** Use the full UUID (with hyphens stripped) as the prefix to achieve 128-bit collision resistance, which also resolves the hyphen/schema-name issue above.

internal/storage/driver/driver_test.go
```diff
@@ -34,7 +35,8 @@ func TestLedgersCreate(t *testing.T) {
 		systemstore.NewStoreFactory(),
 	)
```
NumaryBot, line 27: 🔴 **[blocker] UUID hyphens in bucket names produce invalid PostgreSQL schema identifiers**
<sub>reported by coderabbitai[bot], NumaryBot</sub>

uuid.NewString() produces strings like '550e8400-e29b-41d4-a716-446655440000'. Because bucket names are used directly as PostgreSQL schema identifiers, the resulting names (e.g. '<uuid>-b1') are invalid as unquoted schema names in PostgreSQL. CreateLedger will fail during schema migrations before any test assertion is reached, converting a flaky test into a permanently failing one.

**Suggestion:** Strip hyphens from the UUID before constructing bucket and ledger names:
```go
prefix := strings.ReplaceAll(uuid.NewString(), "-", "")
buckets := []string{prefix + "b1", prefix + "b2"}
```
This also eliminates the truncation issue and raises collision resistance to 128 bits.

internal/storage/driver/driver_test.go
```diff
@@ -34,7 +35,8 @@ func TestLedgersCreate(t *testing.T) {
...
 
-	buckets := []string{"bucket1", "bucket2"}
+	prefix := strings.ReplaceAll(uuid.NewString(), "-", "")
```
NumaryBot: 🟡 **[minor] Truncating UUID to 8 characters leaves only 32 bits of entropy**
<sub>reported by coderabbitai[bot], NumaryBot</sub>

The current code takes only the first 8 hex characters of the UUID as the prefix, providing just 32 bits of entropy. On a long-lived or shared PostgreSQL instance this can still cause name collisions between test runs.

**Suggestion:** Use the full UUID (with hyphens stripped) as the prefix to achieve 128-bit collision resistance, which also resolves the hyphen/schema-name issue.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1452#issuecomment-4841307262

gfyrag (changes requested):
Each test use a dedicated db.
So reusing bucket names is ok.

ariel-formance:
@gfyrag Although ideally tests should use a different db, I was getting the following error (I think because it's being run concurrently by 30 workers) -

❯ --- FAIL: TestLedgersCreate (3.08s)
      driver_test.go:67:
              Error Trace:    /home/runner/work/ledger/ledger/internal/storage/driver/driver_test.go:67
              Error:          Received unexpected error:
                              ERROR: duplicate key value violates unique constraint "pg_namespace_nspname_index" (SQLSTATE 23505)
              Test:           TestLedgersCreate
  FAIL

internal/storage/driver/driver_test.go
```diff
@@ -8,31 +8,22 @@ import (
...
 
-	d := driver.New(
-		db,
-		ledgerstore.NewFactory(db),
-		bucket.NewDefaultFactory(),
-		systemstore.NewStoreFactory(),
-	)
+	d := newTestDriver(t)
 
 	buckets := []string{"bucket1", "bucket2"}
```
NumaryBot, line 28: 🔴 **[blocker] UUID hyphens in bucket names produce invalid PostgreSQL schema identifiers**
<sub>reported by coderabbitai[bot], NumaryBot</sub>

uuid.NewString() produces strings like '550e8400-e29b-41d4-a716-446655440000'. Because bucket names are used directly as PostgreSQL schema identifiers, the resulting names (e.g. '<uuid>-b1') are invalid as unquoted schema names in PostgreSQL. CreateLedger will fail during schema migrations before any test assertion is reached, converting a flaky test into a permanently failing one.

**Suggestion:** Strip hyphens from the UUID before constructing bucket and ledger names:
```go
prefix := strings.ReplaceAll(uuid.NewString(), "-", "")
buckets := []string{prefix + "b1", prefix + "b2"}
```
This also eliminates the truncation issue and raises collision resistance to 128 bits.

internal/storage/driver/driver_test.go
```diff
@@ -8,31 +8,22 @@ import (
...
 	t.Parallel()
 	ctx := logging.TestingContext()
 
-	d := driver.New(
-		db,
-		ledgerstore.NewFactory(db),
-		bucket.NewDefaultFactory(),
-		systemstore.NewStoreFactory(),
-	)
+	d := newTestDriver(t)
```
NumaryBot, line 27: 🟠 **[major] Concurrent goroutines still race on shared bucket names within the same test run**

All 30 goroutines in TestLedgersCreate share only two fixed bucket names (e.g. prefix+"b1" and prefix+"b2"). When two CreateLedger calls pick the same uninitialized bucket, they can both enter bucket migration concurrently and attempt to create the same schema or version table at the same time, reproducing the pg_namespace_nspname_index unique-constraint failure within a single test run. The fix of using a unique prefix per test invocation does not protect against intra-test concurrency on the same two bucket names.

**Suggestion:** Assign each goroutine its own unique bucket name (e.g. using the goroutine index), or serialize bucket initialization with a per-bucket mutex/once, so that no two concurrent goroutines ever race on the same uninitialized bucket.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1452#issuecomment-4841307262

internal/storage/driver/driver_test.go
NumaryBot, line 27: 🔴 **[blocker] UUID hyphens in bucket names produce invalid PostgreSQL schema identifiers**
<sub>reported by codex, NumaryBot, coderabbitai[bot]</sub>

uuid.NewString() produces strings like '550e8400-e29b-41d4-a716-446655440000'. Because bucket names are used directly as PostgreSQL schema identifiers, the resulting names (e.g. '<uuid[:8]>-b1') are invalid as unquoted schema names in PostgreSQL. CreateLedger will fail during schema migrations before any test assertion is reached, converting a flaky test into a permanently failing one.

**Suggestion:** Strip hyphens from the UUID before constructing bucket and ledger names:
```go
prefix := strings.ReplaceAll(uuid.NewString(), "-", "")
buckets := []string{prefix + "b1", prefix + "b2"}
```
This also eliminates the truncation issue and raises collision resistance to 128 bits.

NumaryBot, line 27: 🔴 **[blocker] 30 goroutines inside TestLedgersCreate still race on shared bucket names**

Even with a per-test unique prefix, all 30 goroutines within a single TestLedgersCreate invocation share only two bucket names. Two or more goroutines can simultaneously observe a bucket as uninitialized and race through Migrate, reproducing the pg_namespace_nspname_index duplicate-key constraint violation within a single test run — the original flakiness is not fully resolved.

**Suggestion:** Ensure each goroutine (or each ledger) receives its own unique bucket name, e.g. by generating a fresh UUID per iteration inside the goroutine loop, so no two goroutines ever attempt to initialize the same bucket concurrently.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1452#issuecomment-4841307262

gfyrag (approved):
Refactor with per-test DB isolation (newTestDriver) addresses the original review.

PR #474, opened 2023-12-20
Fix interaction with max and unbounded overdraft

Azorlogh:
Fixes https://github.com/formancehq/ledger/issues/473

Bug explanation:
Each source is associated with an optional fallback account that is used to fund any missing amount from that source.
When using the `max` source, it will use the OP_TAKE_MAX instruction to cap the subsource's funding.
OP_TAKE_MAX returns the amount missing to max out (or 0), and the amount remaining if it was already maxed out.
When the subsource has no fallback, the amount missing is simply discarded
When the subsource has a fallback, the amount missing should be taken from it unconditionally to max out the funding, which should be done with the OP_TAKE_ALWAYS instruction.

PR #1454, opened 2026-06-30
fix(http): enforce bulk per-element scopes when auth is enabled (#474)

Azorlogh:
## Summary

Fixes a **critical unauthenticated write bypass** on the `/{ledger}/bulk` and
`/{ledger}/_bulk` endpoints. With auth enabled, no bearer token, and the default
scope mapping, a caller could run bulk transaction and metadata writes with zero
authorization.

Closes #474.

## Root cause

The bulk handler gated its per-element scope check on `if effective != nil`,
using the nil-ness of the expanded-scopes map as a proxy for "auth enforcement
active". Two distinct states both produce a nil map there, and the handler
couldn't tell them apart:

- **Auth disabled** — the middleware returns early and never sets the scope key,
  so `ExpandedScopesFromContext` returns a nil map. *(correct to skip)*
- **Auth enabled, no token, default mapping** — the middleware stores
  `cfg.ScopeMapping.AnonymousScopes()`, which is `nil` because `DefaultMapping`
  has no `"anonymous"` key. *(should be denied)*

In Go a nil map compares equal to `nil`, so `effective != nil` was `false` in
both cases — the per-element loop was skipped entirely and control fell through
to `runBulk → applyUnsigned → backend.Apply`, which performs no scope check.

Notably, `HasScope(nil, transactions:write)` would have correctly returned
`false` — the bug is that the check was never reached.

## Fix

Gate the loop on `s.authCfg.Enabled` instead, exactly as `RequireScope`
(`http_middleware.go`) and the gRPC `Apply` handler (`server_bucket.go`) already
do. An anonymous request now reaches `HasScope` with a nil/empty scope set,
which denies with **401**; auth-disabled requests still bypass the check as
before. This required threading `AuthConfig` into the HTTP `Server` struct — the
same field-on-the-struct + gate-on-`Enabled` shape the gRPC servers already use.

This was the only `effective != nil` enforcement guard in the codebase; every
other site already keys off `cfg.Enabled`, so no other path changes.

## Testing

- `TestHandleBulk_AuthEnabled_NoToken_Unauthorized` — full path through
  `NewHandler` (real middleware), auth enabled + default mapping, no token, a
  `CREATE_TRANSACTION` element → asserts **401 `UNAUTHENTICATED`**. The mock
  backend has no `Apply` expectation, so it also proves the write never reaches
  the backend.
- `TestHandleBulk_AuthDisabled_NoToken_Allowed` — control: the same request with
  auth disabled still reaches the backend and returns 200.

Verified the regression test fails on the original code (the unauthenticated
write reaching `backend.Apply`) and passes with the fix. `go build ./...`,
`go vet`, the `internal/adapter/http` + `internal/adapter/auth` test suites,
`gofmt`, and `golangci-lint run` on both packages are all clean.

### DIFF
--- a/internal/adapter/http/handler.go
+++ b/internal/adapter/http/handler.go
@@ -66,7 +66,7 @@ func NewHandler(logger logging.Logger, backend Backend, authCfg internalauth.Aut
 	)
 
 	// Create server instance for handlers
-	server := NewServer(logger, backend, defaultBulkMaxSize)
+	server := NewServer(logger, backend, authCfg, defaultBulkMaxSize)
 
 	// Ops routes: not versioned (K8s probes, monitoring, profiling).
 	// /health, /livez, /readyz, /_info are also exempted from auth in
--- a/internal/adapter/http/handlers_bulk.go
+++ b/internal/adapter/http/handlers_bulk.go
@@ -36,9 +36,12 @@ func (s *Server) handleBulk(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Per-element scope check: verify the caller has the required scope for each element.
-	effective := internalauth.ExpandedScopesFromContext(r.Context())
-	if effective != nil {
+	// Per-element scope check: each element may require a different granular
+	// scope. Enforcement keys off auth being enabled — same as RequireScope and
+	// the gRPC Apply handler — so an anonymous request with an empty or nil scope
+	// set is denied by HasScope rather than skipped.
+	if s.authCfg.Enabled {
+		effective := internalauth.ExpandedScopesFromContext(r.Context())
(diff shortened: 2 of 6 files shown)


gfyrag (approved):
LGTM. Fix correct, minimal et chirurgical — gate aligné sur `authCfg.Enabled` comme `RequireScope` et le handler gRPC `Apply`. Les deux tests de régression (auth-enabled→401 sans appel backend + contrôle auth-disabled→200) couvrent bien la classe.

Follow-up structurel ouvert en EN-1417 pour refactor `internal/adapter/auth` (collapse `AuthState` + helper `AuthorizeRequests` partagé bulk/Apply) — pas bloquant pour ce PR.

PR #347, opened 2022-09-24
doc: links to formance.com instead of numary

hhamon:
Signed-off-by: Hugo Hamon <hhamon@users.noreply.github.com>

PR #1456, opened 2026-06-30
fix(check): verify live idempotency entries frozen by archived audit entries

Azorlogh:
## Summary

`checker.compareIdempotencyOutcomes` verifies each frozen idempotency entry (`SubIdempKeys`) against the outcome re-derived from the hash-chained `AuditEntry` that wrote it. Until now it only re-derived entries from the **verified (post-archive) audit range**, so a stored entry whose freezing audit entry had been **archived** was skipped — and if that entry was still *live*, its outcome (proposal hash, failure reason/message/metadata, or success log range) went unverified.

This PR makes `expected` complete across the archive boundary by re-deriving the archived freezes that can still be live (those within the idempotency TTL window), then reports any unmatched live entry bidirectionally — the standard the other projection passes adopted in #347.

## Background

The recent guard (`327b5a38`) closed the dangerous part: a tampered `created_at` on a *post-boundary* entry is reported rather than silently skipped. The **residual gap** was a still-live entry whose `created_at` is *before* the archive boundary — reachable only when the idempotency TTL outlives the age at which chapters are archived. In that window a tampered outcome replayed to a duplicate caller while `Check()` passed. This was the only `Check()` projection pass not complete across the archive boundary.

Threat model is unchanged: an actor with direct disk/Pebble write access to a follower's store (invariant #8).

## What changed

- **TTL window.** `Check()` reads the persisted `idempotency-ttl` and the latest applied HLC timestamp from the same snapshot and computes `ttlCutoff = lastApplied − ttl`. Only entries at/after this can still be live, so the cold read is **bounded by the TTL window, not by history**.
- **Cold re-derivation** (`reDeriveArchivedIdempotency`). After building `expected` from the hot range, the checker reads archived chapters **newest-first** from cold storage and re-derives the frozen outcome for every keyed entry `≥ ttlCutoff`. It stops at the first chapter whose newest entry predates the cutoff — never re-walks the full archive.
- **Report floor.** The single `verifiedRangeStartTs` boundary is replaced by `idemReportFloor`: it drops to `ttlCutoff` when the cold window was fully read, and falls back to the post-archive boundary when cold storage is unavailable (e.g. the CLI/restore validation paths) — preserving the prior no-false-positive behavior, with the residual gap remaining only there.
- **Plumbing.** Only a `*coldstorage.ColdReader` is threaded into `Checker` (TTL and timestamp are read in-`Check`). Added a `ColdReader()` accessor on `DefaultController`; the gRPC `CheckStore` handler passes it. The CLI-bootstrap and restore-validation call sites pass `nil`.

## Design note: cold entries are trusted, not re-hashed

Archived entries used for the TTL window are **re-derived, not re-hash-chained**. Cold storage sits outside the follower-disk threat model this pass targets, so the archived entry is the trusted source and only the live `SubIdempKeys` projection is checked against it. The hot range that anchors the live entries is still fully hash-chain-verified. The doc comments mark exactly where cold-chain verification would go if the threat model is ever widened.

## Testing

- New `TestCompareIdempotencyOutcomes_ArchivedFreezeWithinTTLWindow`: a live entry frozen by an archived-but-in-window audit entry passes when faithful and is flagged (`CHECK_STORE_ERROR_TYPE_IDEMPOTENCY_MISMATCH`) when its outcome is tampered; an entry frozen by an archived entry older than the window is skipped (no false positive). Bounding is asserted by leaving a fully-below-window chapter's SST absent — if the pass read it, the cold read would fail, the floor would rise, and the in-window tamper would no longer flag.
- Existing idempotency/hash-chain tamper tests updated for the new signature; all pass.
- `go build ./...`, `go vet ./...`, `go test -race` (check package), and `golangci-lint` on all changed packages: clean.

## Acceptance criteria

- [x] A frozen entry whose freezing audit entry is archived but within the TTL window is verified against its cold-storage audit source; a tampered outcome is reported.
- [x] No false positives for entries frozen by archived entries older than the TTL window.
- [x] Cold-storage reads bounded by the TTL window.
- [x] Test coverage for live + archived-freeze + tampered → flagged; older-than-TTL → skipped.

## References

- Prior guard: `327b5a38` (`fix(check): flag tampered created_at on live frozen idempotency entries`)
- Bidirectional / no-silent-skip standard: #347 (`compareTransactions`)
- Invariant #8 (audit log is the only source of truth — every projection must be verified)

### DIFF
--- a/internal/application/check/checker.go
+++ b/internal/application/check/checker.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"math/big"
+	"sort"
 
 	"github.com/cockroachdb/pebble/v2"
 	"github.com/zeebo/blake3"
@@ -21,6 +22,7 @@ import (
 	"github.com/formancehq/ledger/v3/internal/domain/processing"
 	domainreplay "github.com/formancehq/ledger/v3/internal/domain/replay"
 	"github.com/formancehq/ledger/v3/internal/infra/attributes"
+	"github.com/formancehq/ledger/v3/internal/infra/coldstorage"
 	"github.com/formancehq/ledger/v3/internal/infra/state"
 	"github.com/formancehq/ledger/v3/internal/pkg/bitset"
 	"github.com/formancehq/ledger/v3/internal/pkg/cursor"
@@ -40,18 +42,26 @@ type Checker struct {
 	attrs     *attributes.Attributes
 	logger    logging.Logger
 	clusterID string
+	// coldReader gives the idempotency pass read access to archived audit
+	// entries so a still-live frozen outcome whose freezing entry has been
+	// archived can be re-derived rather than skipped. nil when cold storage
+	// is not configured (e.g. the CLI / restore call sites) — the pass then
+	// keeps the post-archive boundary as its verification floor.
+	coldReader *coldstorage.ColdReader
 }
 
 // NewChecker creates a new Checker. clusterID is used to derive the
 // per-cluster key for verifying audit-hash chain entries — it must match
 // the value the FSM used when writing those entries (enforced via
-// PersistedConfig immutability).
-func NewChecker(store *dal.Store, attrs *attributes.Attributes, clusterID string, logger logging.Logger) *Checker {
+// PersistedConfig immutability). coldReader may be nil when cold storage is
+// not configured.
+func NewChecker(store *dal.Store, attrs *attributes.Attributes, clusterID string, coldReader *coldstorage.ColdReader, logger logging.Logger) *Checker {
 	return &Checker{
-		store:     store,
-		attrs:     attrs,
-		logger:    logger,
-		clusterID: clusterID,
+		store:      store,
+		attrs:      attrs,
+		logger:     logger,
+		clusterID:  clusterID,
+		coldReader: coldReader,
 	}
 }
 
@@ -139,10 +149,33 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 
 	defer func() { _ = replay.Close() }()
 
+	// Lower bound of the idempotency TTL window: an entry frozen before this
+	// timestamp can no longer be live (it is past its TTL relative to the
+	// latest applied HLC timestamp), so it never needs cold-storage
+	// re-derivation. Read from the same snapshot so the window is
+	// deterministic for this run. A nil/zero TTL config yields a cutoff at
+	// lastApplied, disabling the cold extension (no entry counts as live).
+	persisted, err := query.ReadPersistedConfig(snap)
+	if err != nil {
+		return fmt.Errorf("reading persisted config: %w", err)
+	}
+
+	lastAppliedTs, err := query.ReadLastAppliedTimestamp(snap)
+	if err != nil {
+		return fmt.Errorf("reading last applied timestamp: %w", err)
+	}
+
+	ttlMicros := persisted.GetIdempotencyTtlSeconds() * 1_000_000
+
+	var idempotencyTTLCutoff uint64
+	if lastAppliedTs > ttlMicros {
+		idempotencyTTLCutoff = lastAppliedTs - ttlMicros
+	}
+
 	// Verify the audit hash chain before log replay.
 	// This iterates all non-archived audit entries and recomputes each hash
 	// from the stored orders, chaining from archiveLastAuditHash.
-	if err := c.verifyAuditHashChain(ctx, snap, chapters, archiveLastAuditHash, callback); err != nil {
+	if err := c.verifyAuditHashChain(ctx, snap, chapters, archiveLastAuditHash, idempotencyTTLCutoff, callback); err != nil {
 		return fmt.Errorf("verifying audit hash chain: %w", err)
 	}
 
@@ -1458,6 +1491,7 @@ func (c *Checker) verifyAuditHashChain(
 	reader dal.PebbleReader,
 	chapters []*commonpb.Chapter,
 	archiveLastAuditHash []byte,
+	idempotencyTTLCutoff uint64,
 	callback func(*servicepb.CheckStoreEvent),
 ) error {
 	// Find the last archived audit sequence to start iteration after it.
@@ -1495,10 +1529,10 @@ func (c *Checker) verifyAuditHashChain(
 		// from each verified audit entry and compared to SubIdempKeys below.
 		expectedIdem = make(map[idemExpectedKey]expectedIdempotency)
 		// Timestamp of the first (lowest-sequence) verified entry — the archive
-		// boundary. HLC timestamps are monotonic with sequence, so every freeze
-		// in the verified range is at/after it; a stored idempotency entry that
-		// claims a created_at >= this but matches no expected freeze is tampered
-		// or fabricated (see compareIdempotencyOutcomes). 0 = no verified entry.
+		// boundary. HLC timestamps are monotonic with sequence, so the verified
+		// range always re-derives every freeze at/after it; it is the default
+		// idempotency report floor (see the idemReportFloor computation below).
+		// 0 = no verified entry.
 		verifiedRangeStartTs uint64
 	)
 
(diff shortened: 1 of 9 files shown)


internal/application/check/checker.go
```diff
@@ -139,10 +149,33 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
...
+	if err != nil {
+		return fmt.Errorf("reading persisted config: %w", err)
+	}
+
+	lastAppliedTs, err := query.ReadLastAppliedTimestamp(snap)
+	if err != nil {
+		return fmt.Errorf("reading last applied timestamp: %w", err)
+	}
+
+	ttlMicros := persisted.GetIdempotencyTtlSeconds() * 1_000_000
```
NumaryBot: 🟠 **[major] Handle no-expiration idempotency TTL**

When `--idempotency-ttl=0` is used, the documented behavior is that idempotency keys never expire, but this calculation treats it as a zero-length TTL: with any nonzero `lastAppliedTs`, `idempotencyTTLCutoff` becomes `lastAppliedTs`. In that configuration, frozen entries whose audit source is archived are still live forever, yet the cold re-derivation won't scan the archived window and pre-boundary `SubIdempKeys` remain skipped, leaving the verification gap this change is meant to close.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1456#issuecomment-4843789503

gfyrag (commented):
Solid work — the projection is now complete across the archive boundary and the tests cover the interesting branches (window bounding, never-expire, deletion non-goal). A few points worth surfacing before merge.

## Substantive

### 1. The TTL cutoff itself is derived from projections
`ttlCutoff = lastAppliedTs − ttlSeconds*1e6` — both `ReadLastAppliedTimestamp` (`ZoneGlobal/SubGlobLastAppliedTimestamp`) and `PersistedConfig.IdempotencyTtlSeconds` (`ZoneGlobal/SubGlobPersistedConfig`) are Pebble projections, not hash-chained. The threat model this pass targets is exactly an actor with follower-disk write access, i.e. someone who can tamper Pebble.

Two attack shapes this leaves open:
- **Shrink the TTL** (`IdempotencyTtlSeconds` down): cutoff moves up → entries in `[old_cutoff, new_cutoff)` fall below the floor and become tamper-safe. Config validation catches this on next boot, but not during a running Check().
- **Push `lastAppliedTs` up**: cutoff moves up → same effect.

Cleaner alternative for `lastAppliedTs`: use the highest timestamp reached during `verifyAuditHashChain` (already hash-chain-verified in the same run). That's the "now" the audit chain itself vouches for. For the TTL, it's harder — you'd need to bind it to the chain, which is a bigger design change.

Not a blocker (invariant #8 says projections are trusted between reboots), but the threat-model comment on `compareIdempotencyOutcomes` currently only mentions SubIdempKeys — worth extending it to name the two additional projection inputs the cutoff depends on, so future readers understand the coverage frontier.

### 2. `verifiedRangeStartTs == 0` sentinel collides with a legitimate value
`if verifiedRangeStartTs != 0` treats 0 as "no verified entry". In practice HLC timestamps are microseconds since epoch (> 0), so the collision is unreachable in production — but you just switched `idemReportFloor` to `*uint64` **precisely because** 0 has a legitimate meaning now (never-expire). Using the same 0-sentinel for the other side of the same computation is inconsistent. A `hasVerifiedRange bool` alongside would be symmetric.

Nit-level in prod; more relevant for adversarial / synthetic tests that might exercise timestamp=0.

### 3. `PersistedConfig` never present → cutoff = 0 → full-history scan
On a store where `SubGlobPersistedConfig` doesn't exist (`ReadPersistedConfig` returns `nil, nil` per its doc), `persisted.GetIdempotencyTtlSeconds()` yields 0 → cutoff=0 → "never-expire" full-history scan. That's fine for prod (config is written on first boot) but the CLI/restore paths pass `nil` cold reader, so this manifests as "cutoff=0, len(archived)>0, coldReader=nil → returns false → log 'window extends before boundary but archived audit entries are not readable'". Correct behaviour, just a lot of noise in the log if the persisted config is genuinely missing. Consider treating "PersistedConfig missing" as a distinct case (e.g. debug-log and skip the cold attempt) so operators can distinguish it from a real cold-storage failure.

## Nits

- **`sort.Slice` → `slices.SortFunc`** in `reDeriveArchivedIdempotency` — modern Go, type-safe, slight perf win. Codebase mixes both today, so not a house-style violation.
- **Ordering assumption of `query.ReadAuditEntries`** is implicit in `collectChapterIdempotency` (setting `windowStartsHere` and then continuing to process later entries works whether asc or desc, but the comment doesn't say that). One line clarifying "we don't rely on ordering here — every in-window keyed entry is added regardless of scan direction" would help.
- **Test comment** in `checker_tamper_test.go:294` says the cutoff choice "keeps the report floor at the archive boundary (no cold TTL-window extension)". Strictly speaking the short-circuit is at `idempotencyTTLCutoff < verifiedRangeStartTs` — setting them equal makes the `<` false and skips the whole block. Passing `nil` cold reader would achieve the same isolation with less semantic coupling to the internals.
- **Missing coverage**: success-outcome (`AuditEntry_Success`) cold re-derivation and the `ReadLastAuditEntry == nil` branch aren't exercised. `expectedIdempotencyOutcome` is covered via the hot path, so it's a completeness concern, not a correctness gap.

## What's good

- The non-goal (deletion detection) is documented **in the checker itself** with the reason: eviction is unaudited via TechnicalUpdate. Future readers who wonder "why isn't a missing entry flagged?" find the answer immediately.
- `*uint64` for `idemReportFloor` with the tri-state (nil / 0 / positive) is the right encoding — a boolean plus a value would have been noisier.
- `TestReDeriveArchivedIdempotency_Bounds` structure (one sub-test per branch: no-archived, no-cold-reader, missing-archive, newest-first-stop) is a template worth reusing for other cold-storage passes.
- Bounding assertion via a deliberately-missing SST is a nice adversarial touch — asserts the pass *doesn't* read below the window without needing to intercept reads.
- Explicit threat-model scoping ("cold storage sits outside the follower-disk reach") is exactly the kind of context that ages well.

internal/application/check/checker.go
```diff
@@ -139,10 +150,29 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
...
+	// pass is skipped rather than guessed.
+	persisted, err := query.ReadPersistedConfig(snap)
+	if err != nil {
+		return fmt.Errorf("reading persisted config: %w", err)
+	}
+
+	var idempotencyTTLMicros *uint64
+
+	if persisted != nil {
+		micros := persisted.GetIdempotencyTtlSeconds() * 1_000_000
```
NumaryBot: 🟠 **[major] Trusted TTL source not used for idempotency floor derivation**

When the checker reads `idempotency_ttl_seconds` from the same Pebble snapshot under test, a disk-tampering scenario can set this persisted value to a small non-zero TTL, shifting `idempotencyTTLCutoff` up to or past the archive boundary. This causes cold re-derivation to be skipped entirely, allowing a tampered but still-live archived idempotency entry to pass verification — the exact gap this PR is meant to close.

**Suggestion:** Derive the TTL floor from a validated, externally-configured value (e.g., passed in via the checker's options/config) rather than reading it from the store projection being audited. Fall back to the configured value when the stored projection is absent or cannot be trusted.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1456#issuecomment-4843789503

internal/application/check/checker.go
```diff
@@ -139,10 +160,28 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 
 	defer func() { _ = replay.Close() }()
 
+	// Idempotency TTL, in microseconds, used by the hash-chain pass to bound the
+	// cold re-derivation window. Prefer the boot-validated runtime config (in
+	// process memory, off the audited disk) over the persisted projection, so a
+	// disk-tampered PersistedConfig cannot shrink the window; fall back to the
+	// persisted value where no runtime config exists (CLI / restore). nil (no
+	// runtime config and no persisted config) means the window is unknown, so
```
NumaryBot, line 168: 🟠 **[major] Idempotency TTL floor read from the untrusted store under test**

The idempotency TTL cutoff is computed by reading `idempotency_ttl_seconds` from the same Pebble snapshot being verified. An adversary who tampers directly with on-disk data can shrink this stored value, shifting `idempotencyTTLCutoff` up to or past the archive boundary. This causes cold re-derivation to be skipped entirely, so a tampered-but-still-live archived idempotency entry passes verification — defeating the purpose of this fix. When `--idempotency-ttl=0` (keys never expire) the value reads as zero, making the cutoff equal to `lastAppliedTs` and producing the same gap for permanently-live entries.

**Suggestion:** Derive the TTL floor from a validated, externally-configured value (e.g., passed via the checker's options/config struct) rather than reading it from the store projection being audited. Fall back to the configured value when the stored projection is absent or cannot be trusted. Separately handle the `ttl=0` (no-expiry) case so cold re-derivation always covers the full archive window.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1456#issuecomment-4843789503

internal/application/check/checker.go
Azorlogh: stale finding, it's already fixed

internal/application/check/checker.go
Azorlogh: https://github.com/formancehq/ledger/pull/1456/commits/c29cd6d07708180507f5daabf611aca36f9bb429

gfyrag (approved):
Approved current head `c29cd6d07708180507f5daabf611aca36f9bb429`.

I do not have remaining actionable findings for this PR. I rechecked the idempotency checker concern raised by the stale bot review: the live `CheckStore` path now passes the boot-validated runtime `IdempotencyTTL` into the checker, and `idempotency-ttl=0` is preserved as the never-expire case. The persisted TTL fallback is limited to paths without trusted runtime config, such as foreign backup validation, and the reduced trust model is documented in the checker comments.

The PR checks are green except for the stale NumaryBot review status.

PR #1455, opened 2026-06-30
fix(EN-1242): lazy gen0 tombstone + coverage-gate Delete

gfyrag:
## Problem

On `release/v3.0`, a `Delete` on a key that has migrated to Gen1 via cache rotation between propose and apply leaves cache and disk out of sync — a violation of invariant #1 (cache is the source of authority for a given applied index):

```
T0  Admission: CheckCache(k) → CacheMiss → resolver emits a coverage-only entry.
T1  Concurrent Save(k, v) applies                → Gen0[k] = v.
T2  Cache rotation                               → Gen1[k] = v, Gen0[k] = ∅.
T3  Delete(k) applies:
      KeyStore.Delete → s.M.Get(id) hits Gen1 via fallback → passes tag check
                     → s.M.Put(id, entry{Deleted=true}) writes to Gen0 only.
      Mem: Gen0 = fabricated tombstone, Gen1 = still live.
      Disk (via writeCacheTombstone): tombstone at BOTH gen bytes.
      → Mem Gen1 live vs Disk Gen1 byte tombstoned. Cache-mirror invariant
        broken until the next rotation / restart re-hydrates Gen1.
```

Between T3 and the next rotation, a follower that has restored from store sees the tombstone while a node still holding the live Gen1 row surfaces the value — a live-vs-tombstone split on a hash-chained order.

## Solution — lazy Del promote + writeCacheTombstone at gen0 byte only

Two in-place primitives on `AttributeCache` maintain cache-mirror parity without any preemptive promote pass:

- **`Del`** tombstones in Gen0 when Gen0 hits; otherwise lazy-fabricates a Gen0 tombstone borrowing Gen1's tag. Gen1's live row stays untouched (shadowed by the Gen0 tombstone on every read, purged on the next rotation). `writeCacheTombstone` writes a single row to the Gen0 byte — mem and disk stay byte-equivalent (invariant #1).
- **`Get`** unchanged — the Gen0 → Gen1 fallback that release/v3.0 already had, still safe under the coverage_bits gate (invariant #9).

Preload now skips coverage-only `AttributeCoverage` entries entirely — only seed intents (`value` set) call `MirrorPreload`. The `MirrorTouch` mechanism and its Preload dispatch are removed as dead code (also removes `AttributeCache.Touch`, `Cache.TouchByType`, `CacheSnapshotter.MirrorTouch`).

Two out-of-band guardrails keep the model honest:

[...]

PR #1458, opened 2026-06-30
feat(admission): reject when cache horizon exceeds 2 generations

gfyrag:
## Problem

When admission's predicted apply-time generation `Gen(nextIndex)` is 2+
generations ahead of the FSM's current applied generation, any value loaded
as a `MirrorPreload` is doomed: the first rotation between propose and apply
moves it Gen0 → Gen1, the second discards it entirely. By apply time the
cache has no record of the preloaded value — the FSM read misses and the
order fails with `ErrNotFound` for a key admission promised would be
available. This violates invariant #1 (cache is the source of authority).

Today the code returns `CacheMiss` in this regime and emits a stale
`MirrorPreload` anyway. This is the path the EN-1242 churn investigation
surfaced.

## Fix — reject at admission

| `Gen(at) − currentGen` | `CheckCache` result | Plan emitted |
|---|---|---|
| 0 | `CacheGuaranteed` / `CacheNeedsTouch` / `CacheMiss` | `Declare` / `Touch` / `Preload` |
| 1 | `CacheGuaranteed` / `CacheMiss` | `Declare` / `Preload` |
| **≥ 2** | **`CacheUnreachable`** | **proposal rejected** |

When admission encounters `CacheUnreachable` it returns
`plan.ErrCacheHorizonExceeded`. The gRPC adapter maps that to
`codes.Unavailable` so the existing client retry interceptors handle it
transparently. The proposal never enters Raft, no audit entry is owed, and
the client retries against a fresh admission snapshot once the FSM has
caught up.

## Why this is not a domain error

The rejection is **not** a business outcome — the user-submitted operation
might be perfectly valid. It is an *operational* signal that admission's
view is too far behind FSM apply for the system to safely admit any new
proposal whose preload would target keys this stale. So `ErrCacheHorizonExceeded`
lives in `internal/infra/plan/`, not in `internal/domain/`, and the gRPC
adapter classifies it as a transient infrastructure error.

## When can it fire?

Under a correctly tuned `--cache-rotation-threshold` (default 1000) and
an FSM apply rate that keeps up with proposals, the ≥ 2-generation horizon
should never be reached: admission's `nextIndex` would have to sit at least
`2 × threshold` indices past the FSM's last-applied index (≥ 2000 in-flight
proposals at default tuning).

Recurring rejections are an operational signal — see [the new doc](docs/technical/architecture/core/admission-cache-horizon.md)
for runbook guidance.

## Changes

| File | Change |
|---|---|
| `internal/infra/cache/cache.go` | New `CacheUnreachable` constant; `CheckCache` default branch returns it instead of `CacheMiss`. |
| `internal/infra/plan/errors.go` | New `ErrCacheHorizonExceeded` sentinel + rationale comment. |
| `internal/infra/plan/resolve.go` | Switch case on `CacheUnreachable` short-circuits with the sentinel. |
| `internal/adapter/grpc/server.go` | `convertToGRPCError` maps `ErrCacheHorizonExceeded` → `codes.Unavailable`. |
| `docs/technical/architecture/core/admission-cache-horizon.md` | New doc explaining the contract, the operational signal, and recovery. |
| `docs/technical/architecture/core/fsm-cache-layers.md` | Reference the new doc from the Related section. |

## Tests

- `TestAttributeCache_IsGuaranteedInCache_TwoGenerationsAhead` — strengthened
  to assert `CacheUnreachable` specifically (not the prior `CacheMiss`), for
  both stored and absent keys.
- `TestBuildPreloads_RejectsCacheHorizonExceeded` — end-to-end resolver run
  with a tracker pinned past two generation boundaries; asserts
  `ErrCacheHorizonExceeded` propagates out of `Builder.Build` so the gRPC
  layer maps to `codes.Unavailable`.

## Verification

- `go build ./...` passes.
- `golangci-lint run ./internal/infra/cache/... ./internal/infra/plan/... ./internal/adapter/grpc/...` → 0 issues.
- Targeted test suites green.
- Standalone from EN-1242 (#1455).

### DIFF
--- a/internal/adapter/grpc/server.go
+++ b/internal/adapter/grpc/server.go
@@ -33,6 +33,7 @@ import (
 	"github.com/formancehq/ledger/v3/internal/domain"
 	"github.com/formancehq/ledger/v3/internal/domain/crypto/signing"
 	"github.com/formancehq/ledger/v3/internal/infra/node"
+	"github.com/formancehq/ledger/v3/internal/infra/plan"
 	"github.com/formancehq/ledger/v3/internal/infra/state"
 	"github.com/formancehq/ledger/v3/internal/infra/transport"
 	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
@@ -504,6 +505,7 @@ func convertToGRPCError(err error, logger logging.Logger) error {
 		errors.Is(err, node.ErrNodeSyncing) ||
 		errors.Is(err, node.ErrTransferLeaderTimeout) ||
 		errors.Is(err, commonpb.ErrNoLeader) ||
+		errors.Is(err, plan.ErrCacheHorizonExceeded) ||
 		errors.Is(err, ErrNodeNotReachable) {
 		return status.Error(codes.Unavailable, err.Error())
 	}
--- a/internal/infra/cache/cache.go
+++ b/internal/infra/cache/cache.go
@@ -177,13 +177,18 @@ const (
 // raft index `at`. Takes a read lock on the cache to ensure a consistent view
 // of currentGeneration and the gen0/gen1 data during the check.
 func (a *AttributeCache[T]) CheckCache(at uint64, k attributes.U128) CacheStatus {
+	a.Cache.mu.RLock()
+	defer a.Cache.mu.RUnlock()
+
+	// Threshold + currentGeneration are read INSIDE the RLock so a concurrent
+	// ResetWithThreshold (write-lock holder) cannot bump one between our
+	// reads and leave us with a (threshold=old, currentGeneration=new)
+	// snapshot. Such a torn view would classify a valid admission as
+	// CacheUnreachable during the threshold-change transition window.
 	threshold := a.Cache.GenerationThreshold()
 	if threshold == 0 {
 		return CacheMiss
 	}
-
-	a.Cache.mu.RLock()
-	defer a.Cache.mu.RUnlock()
 
 	actualGeneration := a.Cache.currentGeneration.Load()
 	futureGeneration := Gen(at, threshold)
--- a/internal/infra/cache/cache.go
+++ b/internal/infra/cache/cache.go
@@ -188,6 +188,17 @@ func (a *AttributeCache[T]) CheckCache(at uint64, k attributes.U128) CacheStatus
 	actualGeneration := a.Cache.currentGeneration.Load()
 	futureGeneration := Gen(at, threshold)
 
+	// Stale-behind build: an admission build sampled `at` before the FSM
+	// applied entries past that index, so its `at` maps to a generation the
+	// FSM has already left behind. Not a horizon violation — the higher-level
+	// staleness guard (checkStaleProposal / AcquireProposalGuard) will
+	// reject or rebuild. Report CacheMiss so the caller loads from store and
+	// the concurrent apply resolves the outcome; do NOT let the uint64
+	// subtraction underflow into the CacheUnreachable default branch.
+	if futureGeneration < actualGeneration {
+		return CacheMiss
+	}
+
 	switch futureGeneration - actualGeneration {
 	case 0:
 		// Same generation — no rotation expected.
--- a/internal/infra/cache/cache.go
+++ b/internal/infra/cache/cache.go
@@ -355,11 +355,21 @@ func (c *Cache) Reset() {
 // caller-provided raftIndex is the entry the FSM is applying, so
 // Gen(raftIndex, threshold) is the correct post-reset horizon.
 //
-// Callers that persist the reset to Pebble must read the returned tuple
-// (or Cache.CurrentGeneration + Cache.BaseIndex.{Gen0,Gen1}) so the
-// on-disk sentinels reflect the same in-memory state a RestoreFromStore
-// would reconstruct.
+// A zero threshold is a config invariant violation — cache.New already
+// rejects it, and every code path that reaches this method has validated
+// the config upstream. Panics loudly rather than silently disabling
+// rotations, which would leave currentGeneration frozen at 0 forever and
+// break the CacheUnreachable / CheckRotationNeeded contracts.
+//
+// Callers that persist the reset to Pebble must read
+// Cache.CurrentGeneration + Cache.BaseIndex.{Gen0,Gen1} after the call so
+// the on-disk sentinels reflect the same in-memory state a
+// RestoreFromStore would reconstruct.
 func (c *Cache) ResetWithThreshold(threshold, raftIndex uint64) {
+	if threshold == 0 {
+		panic("cache.ResetWithThreshold: threshold must be > 0 (invariant enforced by cache.New)")
+	}
+
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
@@ -367,13 +377,9 @@ func (c *Cache) ResetWithThreshold(threshold, raftIndex uint64) {
 	c.epoch.Add(1)
 	c.generationThreshold.Store(threshold)
 
-	// Skip the alignment when the threshold is 0 (rotations disabled) or
-	// the target generation is 0 (raftIndex within the first gen — the
-	// post-clearLocked state already matches).
-	if threshold == 0 {
-		return
-	}
-
+	// clearLocked already set currentGeneration=0 and BaseIndex={0,0}. If
+	// raftIndex falls inside the first generation, that's the final state.
+	// Otherwise realign under the same lock.
 	if g := Gen(raftIndex, threshold); g != 0 {
 		boundary := genEndIndex(g-1, threshold)
 		c.rotateLocked(boundary, g)
--- a/internal/infra/cache/cache.go
+++ b/internal/infra/cache/cache.go
@@ -342,17 +342,42 @@ func (c *Cache) Reset() {
 	c.clearLocked()
 }
 
-// ResetWithThreshold atomically resets the cache, increments the epoch, and
-// sets a new generation threshold. Called by the FSM when a cluster config
-// change is applied — the epoch increment is deterministic (all nodes apply
-// the same Raft entry).
-func (c *Cache) ResetWithThreshold(threshold uint64) {
+// ResetWithThreshold atomically resets the cache, increments the epoch, sets
+// a new generation threshold, AND realigns currentGeneration + BaseIndex to
+// the generation that raftIndex falls into under the new threshold. Called
+// by the FSM when a cluster config change is applied.
+//
+// The epoch increment is deterministic (all nodes apply the same Raft
+// entry). Realigning currentGeneration and BaseIndex in the same critical
+// section closes the race window where admission's CheckCache would
+// otherwise observe currentGeneration=0 against the new threshold and
+// falsely trip the CacheUnreachable horizon (2+ predicted rotations); the
+// caller-provided raftIndex is the entry the FSM is applying, so
+// Gen(raftIndex, threshold) is the correct post-reset horizon.
+//
+// Callers that persist the reset to Pebble must read the returned tuple
+// (or Cache.CurrentGeneration + Cache.BaseIndex.{Gen0,Gen1}) so the
+// on-disk sentinels reflect the same in-memory state a RestoreFromStore
+// would reconstruct.
+func (c *Cache) ResetWithThreshold(threshold, raftIndex uint64) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
 	c.clearLocked()
 	c.epoch.Add(1)
 	c.generationThreshold.Store(threshold)
+
+	// Skip the alignment when the threshold is 0 (rotations disabled) or
+	// the target generation is 0 (raftIndex within the first gen — the
+	// post-clearLocked state already matches).
+	if threshold == 0 {
+		return
+	}
+
+	if g := Gen(raftIndex, threshold); g != 0 {
+		boundary := genEndIndex(g-1, threshold)
+		c.rotateLocked(boundary, g)
+	}
 }
 
 // clearLocked clears all cache data without incrementing the epoch.
--- a/internal/infra/cache/cache.go
+++ b/internal/infra/cache/cache.go
@@ -149,6 +149,15 @@ const (
 	CacheNeedsTouch
 	// CacheMiss means the key is not in cache at all — a full preload from store is needed.
 	CacheMiss
+	// CacheUnreachable means the target index is 2+ generations ahead of the
+	// current FSM-applied generation: any preload value computed now would be
+	// invalidated by the rotations that must run before apply (gen0 -> gen1 ->
+	// discarded). The admission cannot guarantee that the proposal will read a
+	// consistent cache horizon, so the order must be rejected and re-admitted
+	// once the FSM apply has caught up. This is an operational signal — under
+	// a correctly tuned rotation threshold and a healthy apply rate it should
+	// not occur.
+	CacheUnreachable
 )
 
 // CheckCache determines whether a key will survive in cache until the future raft
@@ -205,8 +214,12 @@ func (a *AttributeCache[T]) CheckCache(at uint64, k attributes.U128) CacheStatus
 
 		return CacheMiss
 	default:
-		// 2+ generations ahead — data will be lost regardless.
-		return CacheMiss
+		// 2+ generations ahead — any preload value computed now would be
+		// rotated out (gen0 -> gen1 -> discarded) before apply. Signal an
+		// unreachable horizon so admission can reject the proposal with a
+		// transient error; the client retries and admission re-admits with
+		// a fresh prediction once the FSM apply has caught up.
+		return CacheUnreachable
 	}
 }
 
--- a/internal/infra/cache/cache.go
+++ b/internal/infra/cache/cache.go
@@ -185,11 +185,11 @@ func (a *AttributeCache[T]) CheckCache(at uint64, k attributes.U128) CacheStatus
 	// reads and leave us with a (threshold=old, currentGeneration=new)
 	// snapshot. Such a torn view would classify a valid admission as
 	// CacheUnreachable during the threshold-change transition window.
+	//
+	// threshold > 0 is a cluster-wide invariant: cache.New rejects 0 and
+	// ResetWithThreshold panics on 0 — no legitimate call path can observe
+	// threshold=0 here.
 	threshold := a.Cache.GenerationThreshold()
-	if threshold == 0 {
-		return CacheMiss
-	}
-
 	actualGeneration := a.Cache.currentGeneration.Load()
 	futureGeneration := Gen(at, threshold)
 
@@ -416,10 +416,9 @@ func (c *Cache) clearLocked() {
 // and performs it atomically if necessary.
 // Returns whether a rotation occurred and the old Gen1 base index (compaction threshold).
 func (c *Cache) CheckRotationNeeded(index uint64) (rotated bool, oldGen1BaseIndex uint64) {
+	// threshold > 0 is a cluster-wide invariant (see CheckCache / New /
+	// ResetWithThreshold) — no dead-code guard here.
 	threshold := c.generationThreshold.Load()
-	if threshold == 0 {
-		return false, 0
-	}
 
 	c.mu.Lock()
 	defer c.mu.Unlock()
@@ -533,7 +532,12 @@ func (c *Cache) GenerationThreshold() uint64 {
 }
 
 // SetGenerationThreshold updates the cache rotation threshold atomically.
+// Panics on 0 — threshold > 0 is a cluster-wide invariant enforced by
+// cache.New and ResetWithThreshold.
 func (c *Cache) SetGenerationThreshold(v uint64) {
+	if v == 0 {
+		panic("cache.SetGenerationThreshold: threshold must be > 0 (invariant enforced by cache.New)")
+	}
 	c.generationThreshold.Store(v)
 }
 
--- a/internal/infra/plan/errors.go
+++ b/internal/infra/plan/errors.go
@@ -1,0 +1,26 @@
+package plan
+
+import "errors"
+
+// ErrCacheHorizonExceeded is returned by Builder.Build when admission predicts
+// 2+ cache generation rotations will fire on the FSM side between this
+// proposal's propose-time and its apply-time (cache.CheckCache → CacheUnreachable).
+//
+// In that regime, any preload value resolved now would be discarded by the
+// rotations before the FSM apply reads it: a preload landed in gen0 is rotated
+// into gen1 by the first rotation and dropped by the second. The proposal's
+// reads would then miss the preloaded data, producing inconsistent results.
+//
+// Admission rejects the proposal so it never enters Raft. The client retries
+// — by then admission's snapshot is fresh again and CheckCache returns a usable
+// status. The audit log is *not* an issue here: the order never landed, so no
+// audit entry is owed (this is a system-level rejection, not a user-domain
+// business outcome).
+//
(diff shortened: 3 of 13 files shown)


gfyrag:
@Azorlogh @paul-nicolas — small, well-isolated PR (admission-side reject, no domain change, no Raft impact, doc included). Quick pass appreciated when you have a moment.

internal/infra/state/machine_technical_updates.go
```diff
@@ -111,6 +112,14 @@ func (fsm *Machine) applyClusterConfig(batch *dal.WriteSession, raftIndex uint64
...
 
 		fsm.Registry.Cache.ResetWithThreshold(newThreshold)
+		// ResetWithThreshold leaves currentGeneration=0, but the Raft log is
+		// long past index 0 — admission's CheckCache would compute
+		// Gen(nextIndex, newThreshold) - 0, a large value that erroneously
+		// trips the CacheUnreachable horizon and rejects every incoming
+		// proposal until the next FSM apply moves currentGeneration forward.
+		// Set it to the post-reset truth immediately so admission queries
+		// see a consistent (currentGeneration, threshold) pair.
+		fsm.Registry.Cache.SetCurrentGeneration(cache.Gen(raftIndex, newThreshold))
```
NumaryBot: 🔴 **[blocker] BaseIndex not aligned when forcing cache generation after threshold change**

When a threshold change is applied at a Raft index where Gen(raftIndex, newThreshold) > 0, only `currentGeneration` is updated while `ResetWithThreshold` has just reset `BaseIndex` to (0,0). Subsequent proposals in the same new generation build `ExecutionPlan.LastPersistedIndex = BoundaryIndex(nextIndex, newThreshold)`, but `CheckRotationNeeded` will not rotate because the current generation already matches. This causes FSM preload validation to reject with a boundary mismatch, breaking normal proposal processing after a cluster config update.

**Suggestion:** When forcing the cache generation to a non-zero value after a threshold reset, also set BaseIndex to the corresponding boundary so that the in-memory state is fully consistent. Alternatively, allow a rotation to be triggered immediately to establish the correct BaseIndex rather than short-circuiting it.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1458#issuecomment-4844645615

internal/infra/state/machine_technical_updates.go
```diff
@@ -111,6 +111,19 @@ func (fsm *Machine) applyClusterConfig(batch *dal.WriteSession, raftIndex uint64
...
 		fsm.Registry.Cache.ResetWithThreshold(newThreshold)
+		// ResetWithThreshold leaves currentGeneration=0 and BaseIndex={0,0},
+		// but the Raft log is long past index 0. Admission's CheckCache would
+		// then compute Gen(nextIndex, newThreshold) - 0, a large value that
+		// erroneously trips the CacheUnreachable horizon. And any proposal
+		// admitted after the reset with the new epoch would carry a
+		// LastPersistedIndex that doesn't match BaseIndex.{Gen0,Gen1}=0 —
+		// triggering the FSM-side preload boundary mismatch panic on apply.
+		//
+		// CheckRotationNeeded jumps currentGeneration AND BaseIndex to
```
NumaryBot: 🔴 **[blocker] BaseIndex not aligned when forcing cache generation after threshold change**

When a threshold change is applied at a Raft index where Gen(raftIndex, newThreshold) > 0, ResetWithThreshold resets BaseIndex to (0,0) while only currentGeneration is updated to the correct value. Subsequent proposals in that generation compute ExecutionPlan.LastPersistedIndex = BoundaryIndex(nextIndex, newThreshold), but CheckRotationNeeded will not fire because the current generation already matches. This causes FSM preload validation to reject with a boundary mismatch, breaking normal proposal processing after a cluster config update.

**Suggestion:** When forcing the cache generation to a non-zero value after a threshold reset, also set BaseIndex to the corresponding boundary index so the in-memory state is fully consistent. Alternatively, trigger an immediate rotation to establish the correct BaseIndex rather than short-circuiting it.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1458#issuecomment-4844645615

gfyrag:
@NumaryBot re-review please — findings addressed:

- **BaseIndex alignment**: fixed in commit `b76da2d` by switching from `SetCurrentGeneration` to `CheckRotationNeeded` (which atomically updates BOTH currentGeneration and BaseIndex under `cache.mu`, exactly the alternative you suggested).
- **Post-reset persistence**: fixed in commit `fa039efd` — we now persist `CurrentGeneration`, gen0 `CacheGenMeta.BaseIndex`, and gen1 `CacheGenMeta.BaseIndex` to the 0xFF zone so a restart before the next organic rotation reconstructs the same in-memory tuple.
- **Test coverage**: new `TestApplyClusterConfig_ThresholdChangePersistsGeneration` covers the full cycle (config change → in-memory realignment → disk persistence → Reset+RestoreFromStore → CheckCache does not fire CacheUnreachable). `TestBuildPreloads_RejectsCacheHorizonExceeded` already covered admission rejection.

internal/infra/plan/errors.go
```diff
@@ -0,0 +1,26 @@
...
+// audit entry is owed (this is a system-level rejection, not a user-domain
+// business outcome).
+//
+// Under a correctly tuned rotation-threshold and a healthy FSM apply rate, this
+// error should not fire. Recurring occurrences indicate either a too-low
+// rotation threshold or FSM apply falling behind admission (overload).
+//
+// Mapped to gRPC codes.Unavailable in adapter/grpc/server.go so existing client
+// retry interceptors handle it transparently.
+var ErrCacheHorizonExceeded = errors.New("admission cache horizon exceeded: 2+ cache rotations predicted between propose and apply")
```
NumaryBot, line 26: 🟠 **[major] ErrCacheHorizonExceeded not mapped in HTTP adapter — surfaces as 500 instead of retryable 503**

ErrCacheHorizonExceeded is only recognized by the gRPC error conversion path. HTTP write handlers that call Controller.Apply and then handleError will not match the sentinel, so the transient cache-horizon rejection is returned as a generic 500 on REST/compatibility endpoints instead of an appropriate retryable response, breaking the intended retry semantics for HTTP clients.

**Suggestion:** Add equivalent HTTP error classification for ErrCacheHorizonExceeded in the HTTP adapter's error handler (e.g., map it to 503 Service Unavailable with a Retry-After hint), mirroring what convertToGRPCError already does.

internal/infra/state/machine_technical_updates.go
NumaryBot: 🟠 **[major] Realigned in-memory generation not persisted — node restart triggers false ErrCacheHorizonExceeded**

After a threshold change forces currentGeneration to a non-zero value, the realigned generation metadata is never written to durable storage. If the node restarts before the next organic rotation, generation 0 is read from disk. The admission check then incorrectly raises ErrCacheHorizonExceeded for valid proposals (horizon > 2 generations relative to persisted gen 0) until another apply event advances the persisted metadata.

**Suggestion:** Persist the forced generation value (and the aligned BaseIndex) to durable storage immediately after the in-memory update, so that a node restart produces a consistent starting state.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1458#issuecomment-4844645615

internal/adapter/grpc/server.go
```diff
@@ -504,6 +505,7 @@ func convertToGRPCError(err error, logger logging.Logger) error {
 		errors.Is(err, node.ErrNodeSyncing) ||
 		errors.Is(err, node.ErrTransferLeaderTimeout) ||
 		errors.Is(err, commonpb.ErrNoLeader) ||
+		errors.Is(err, plan.ErrCacheHorizonExceeded) ||
```
NumaryBot, line 508: 🟠 **[major] ErrCacheHorizonExceeded not mapped in HTTP adapter — surfaces as 500 instead of retryable 503**

ErrCacheHorizonExceeded is only recognized by the gRPC error conversion path. HTTP write handlers that call Controller.Apply go through the HTTP adapter's handleError, which does not match the sentinel and is not a domain.Describable, so the transient cache-horizon rejection falls through to a generic 500 instead of an appropriate retryable response. This breaks the intended retry semantics for HTTP and REST/compatibility clients.

**Suggestion:** Add equivalent HTTP error classification for ErrCacheHorizonExceeded in the HTTP adapter's error handler (e.g., map it to 503 Service Unavailable with a Retry-After hint), mirroring what convertToGRPCError already does.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1458#issuecomment-4844645615

internal/adapter/grpc/server.go
NumaryBot, line 508: 🟠 **[major] ErrCacheHorizonExceeded not mapped in HTTP adapter — surfaces as 500 instead of retryable 503**

ErrCacheHorizonExceeded is only recognised by the gRPC error-conversion path. HTTP/REST write endpoints that reach the same rejection path go through the HTTP adapter's error handler, which does not match the sentinel and does not implement domain.Describable, so the transient cache-horizon rejection falls through to a generic 500. This breaks the intended retry semantics for HTTP and REST/compatibility clients.

**Suggestion:** Add equivalent HTTP error classification for ErrCacheHorizonExceeded in the HTTP adapter's error handler (e.g., map it to 503 Service Unavailable with a Retry-After hint), mirroring what convertToGRPCError already does.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1458#issuecomment-4844645615

internal/adapter/grpc/server.go
NumaryBot, line 508: 🟠 **[major] ErrCacheHorizonExceeded not mapped in HTTP adapter — surfaces as 500 instead of retryable 503**

ErrCacheHorizonExceeded is only recognised by the gRPC error-conversion path. HTTP/REST write endpoints that reach the same rejection go through the HTTP adapter's error handler, which does not match the sentinel and does not implement domain.Describable, so the transient cache-horizon rejection falls through to a generic 500. This breaks the intended retry semantics for HTTP and REST/compatibility clients.

**Suggestion:** Add equivalent HTTP error classification for ErrCacheHorizonExceeded in the HTTP adapter's error handler (e.g., map it to 503 Service Unavailable with a Retry-After hint), mirroring what convertToGRPCError already does.

internal/infra/cache/cache.go
```diff
@@ -205,8 +214,12 @@ func (a *AttributeCache[T]) CheckCache(at uint64, k attributes.U128) CacheStatus
 
 		return CacheMiss
 	default:
```
NumaryBot, line 232: 🟠 **[major] Unsigned subtraction underflow misclassifies stale-behind builds as CacheUnreachable**

When an optimistic preload build is overtaken by concurrent admissions or apply calls, `actualGeneration` can become greater than `futureGeneration` for the initially sampled `nextIndex`. The unsigned subtraction then underflows, falling into the default branch and returning CacheUnreachable. This causes valid stale-behind builds to be rejected with ErrCacheHorizonExceeded under concurrency, producing avoidable 503/Unavailable responses rather than only rejecting the intended 2+ generations-ahead case.

**Suggestion:** Before computing the unsigned difference, check whether actualGeneration > futureGeneration and treat that as a stale-behind (non-error) case rather than an unreachable horizon. Only return CacheUnreachable when futureGeneration genuinely exceeds actualGeneration by more than 2.

internal/infra/plan/resolve.go
```diff
@@ -122,6 +122,22 @@ func resolveAttributePreload[K interface {
 		id, tag := attributes.MakeKey(canonicalKey)
 
 		switch attrCache.CheckCache(nextIndex, id) {
```
NumaryBot, line 124: 🟠 **[major] Early return on CacheUnreachable skips draining in-flight loader goroutines**

If CacheUnreachable is observed after earlier CacheMiss iterations have already launched loader goroutines, the immediate return bypasses wg.Wait(). Those goroutines can still append to tracker/plans and hold AttributeLoader entries after the caller has already built and released the cleanup token, leaking loader state and introducing a data race with the returned result.

**Suggestion:** On the CacheUnreachable path, record the error but do not return immediately — drain all in-flight workers (wg.Wait()) before returning, ensuring loader state is fully cleaned up regardless of the error type.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1458#issuecomment-4844645615

gfyrag:
@Azorlogh @paul-nicolas — bump. All NumaryBot findings addressed since the initial ping:

- BaseIndex alignment (`b76da2d` then folded into `f80382d`) — `Cache.ResetWithThreshold` now takes `raftIndex` and realigns `currentGeneration` + `BaseIndex` atomically under `cache.mu`, in one call, no transient window.
- Post-reset persistence (`fa039ef`) — `applyClusterConfig` now writes the realigned `CurrentGeneration` + both per-gen `CacheGenMeta.BaseIndex` sentinels so restart-before-next-rotation reconstructs the same in-memory tuple.
- `threshold==0` (`d1285b3`) — panics loudly instead of silently disabling rotations (`cache.New` already rejects; enforced in `ResetWithThreshold` too).
- HTTP mapping (`780ce16`) — `ErrCacheHorizonExceeded` now maps to 503 + `Retry-After: 1` in the HTTP adapter, mirroring the gRPC `codes.Unavailable` path.

CI green, doc updated, tests cover: rejection at admission, threshold-change apply + persistence + restart reconstruction, threshold=0 panic, HTTP mapping (sentinel + wrapped).

internal/infra/cache/cache.go
```diff
@@ -205,8 +225,12 @@ func (a *AttributeCache[T]) CheckCache(at uint64, k attributes.U128) CacheStatus
...
```
NumaryBot, line 232: 🟠 **[major] Unsigned subtraction underflow misclassifies stale-behind builds as CacheUnreachable**

When an optimistic preload build is overtaken by concurrent admissions or apply calls, `actualGeneration` can become greater than `futureGeneration` for the initially sampled `nextIndex`. Because the subtraction is performed on unsigned integers, it underflows and falls into the default/large-value branch, returning CacheUnreachable. This causes valid stale-behind builds to be rejected with ErrCacheHorizonExceeded under concurrency, producing avoidable 503/Unavailable responses rather than only rejecting the genuinely 2+ generations-ahead case.

**Suggestion:** Before computing the unsigned difference, check whether `actualGeneration > futureGeneration` and treat that as a stale-behind (non-error) case. Only return CacheUnreachable when `futureGeneration` genuinely exceeds `actualGeneration` by more than 2.

internal/infra/cache/cache.go
```diff
@@ -205,8 +225,12 @@ func (a *AttributeCache[T]) CheckCache(at uint64, k attributes.U128) CacheStatus
...
-		// 2+ generations ahead — data will be lost regardless.
-		return CacheMiss
+		// 2+ generations ahead — any preload value computed now would be
+		// rotated out (gen0 -> gen1 -> discarded) before apply. Signal an
+		// unreachable horizon so admission can reject the proposal with a
+		// transient error; the client retries and admission re-admits with
+		// a fresh prediction once the FSM apply has caught up.
+		return CacheUnreachable
```
NumaryBot, line 238: 🟠 **[major] Threshold and generation reads are not taken under the same lock, causing spurious rejections during threshold changes**

During a rotation-threshold cluster-config change, `CheckCache` can read the old threshold before acquiring `cache.mu`, then read the post-reset `currentGeneration` after `ResetWithThreshold` releases the lock. With a mismatched snapshot (e.g., old low threshold and new high one around the same Raft index), `futureGeneration` is computed with the stale threshold and appears far ahead of `actualGeneration`, causing valid admissions to be rejected with CacheUnreachable / 503 during the transition window.

**Suggestion:** Read both the threshold and `currentGeneration` under the same lock/atomic snapshot so that the generation-distance classification is always self-consistent.

internal/infra/plan/resolve.go
```diff
@@ -122,6 +122,31 @@ func resolveAttributePreload[K interface {
...
```
NumaryBot, line 124: 🟠 **[major] Early return on CacheUnreachable skips draining in-flight loader goroutines**

If CacheUnreachable is observed after earlier CacheMiss iterations have already launched loader goroutines, the immediate return bypasses `wg.Wait()`. Those goroutines can still append to tracker/plans and hold AttributeLoader entries after the caller has already built and released the cleanup token, leaking loader state and introducing a data race with the returned result.

**Suggestion:** On the CacheUnreachable path, record the error but do not return immediately — drain all in-flight workers via `wg.Wait()` before returning, ensuring loader state is fully cleaned up regardless of the error type.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1458#issuecomment-4844645615

gfyrag:
@NumaryBot two of your findings on `be00279` are stale re-reports of what that commit itself fixes:

- **Underflow** (`75a3b97b`) → fixed in `be00279` by the `if futureGeneration < actualGeneration { return CacheMiss }` guard added just before the switch (`cache.go:191-201`).
- **Early-return leak** (`df25f833`) → fixed in `be00279` by replacing the immediate `return` with `firstErr = ErrCacheHorizonExceeded; continue` so `wg.Wait()` drains in-flight loaders (`resolve.go:139-149`).

The new race finding (`ea873a4f`, threshold + currentGeneration read outside the same lock) is legit and addressed in `44bb154b`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1458#issuecomment-4844645615

internal/infra/cache/cache.go
```diff
@@ -205,8 +230,12 @@ func (a *AttributeCache[T]) CheckCache(at uint64, k attributes.U128) CacheStatus
 
 		return CacheMiss
 	default:
```
NumaryBot, line 232: 🟠 **[major] Unsigned subtraction underflow misclassifies stale-behind builds as CacheUnreachable**

When an optimistic preload build is overtaken by concurrent admissions or apply calls, `actualGeneration` can become greater than `futureGeneration` for the initially sampled `nextIndex`. Because the subtraction is performed on unsigned integers, it underflows to a very large value and falls into the default branch, returning CacheUnreachable. This causes valid stale-behind builds to be spuriously rejected with ErrCacheHorizonExceeded / 503 rather than only rejecting the genuinely 2+ generations-ahead case.

**Suggestion:** Before computing the unsigned difference, check whether `actualGeneration > futureGeneration` and treat that as a stale-behind (non-error) case. Only return CacheUnreachable when `futureGeneration` genuinely exceeds `actualGeneration` by more than 2.

internal/infra/cache/cache.go
```diff
@@ -205,8 +230,12 @@ func (a *AttributeCache[T]) CheckCache(at uint64, k attributes.U128) CacheStatus
...
-		// 2+ generations ahead — data will be lost regardless.
-		return CacheMiss
+		// 2+ generations ahead — any preload value computed now would be
+		// rotated out (gen0 -> gen1 -> discarded) before apply. Signal an
+		// unreachable horizon so admission can reject the proposal with a
+		// transient error; the client retries and admission re-admits with
+		// a fresh prediction once the FSM apply has caught up.
+		return CacheUnreachable
```
NumaryBot, line 238: 🟠 **[major] Threshold and currentGeneration not read under the same lock, causing spurious rejections during threshold changes**

During a rotation-threshold cluster-config change, `CheckCache` can read the old threshold before acquiring `cache.mu`, then read the post-reset `currentGeneration` after `ResetWithThreshold` releases the lock. With a mismatched snapshot (stale threshold paired with new currentGeneration), `futureGeneration` is computed incorrectly and can appear far ahead of `actualGeneration`, causing valid admissions to be rejected with CacheUnreachable / 503 during the transition window.

**Suggestion:** Read both the threshold and `currentGeneration` under the same lock or atomic snapshot so that the generation-distance classification is always self-consistent.

internal/infra/plan/resolve.go
NumaryBot, line 124: 🟠 **[major] Early return on CacheUnreachable skips draining in-flight loader goroutines**

If CacheUnreachable is observed after earlier CacheMiss iterations have already launched loader goroutines, the immediate return bypasses `wg.Wait()`. Those goroutines can still append to tracker/plans and hold AttributeLoader entries after the caller has already built and released the cleanup token, leaking loader state and introducing a data race. Additionally, sibling attribute-type slots that complete successfully after the error is recorded are omitted from the cleanup token and can be reused as stale preloads on retries.

**Suggestion:** On the CacheUnreachable path, record the error but do not return immediately — drain all in-flight workers via `wg.Wait()` and collect tracker entries from every slot before returning, ensuring loader state is fully cleaned up regardless of the error type.

internal/infra/plan/resolve.go
```diff
@@ -122,6 +122,31 @@ func resolveAttributePreload[K interface {
...
+			if logger.Enabled(logging.TraceLevel) {
+				logger.WithFields(map[string]any{
+					"type":      typeName,
+					"key":       hex.EncodeToString(canonicalKey),
+					"nextIndex": nextIndex,
+					"boundary":  boundary,
+				}).Tracef("Cache horizon exceeded: admission rejection")
+			}
+
+			mu.Lock()
```
NumaryBot, line 142: 🟠 **[major] ErrCacheHorizonExceeded early return leaks in-flight loader goroutines and sibling slot entries**

When CacheUnreachable is observed after earlier CacheMiss iterations have already launched loader goroutines, an immediate return bypasses wg.Wait(). Those goroutines can still append to tracker/plans and hold AttributeLoader entries after the caller has built and released the cleanup token, producing a data race. Additionally, sibling attribute-type slots that complete successfully after the error is recorded are omitted from the cleanup token, so ReleaseLoaders() cannot remove their entries — leaking preload state across retries.

**Suggestion:** On the CacheUnreachable/ErrCacheHorizonExceeded path, record the error but do not return immediately. Drain all in-flight workers via wg.Wait() and collect tracker entries from every completed slot before returning, ensuring full cleanup regardless of error type.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1458#issuecomment-4844645615

gfyrag:
@NumaryBot all three findings on `44bb154` are stale/re-emitted:

- **Underflow** (`8b06f008`): guard added in `be00279` at [`cache.go:198-201`](https://github.com/formancehq/ledger/blob/44bb154bd/internal/infra/cache/cache.go#L198-L201) — `if futureGeneration < actualGeneration { return CacheMiss }` before the switch. Test: `TestAttributeCache_CheckCache_StaleBehindReportsMiss`.
- **Torn snapshot** (`85503d17`): `44bb154` **is** the fix — the threshold read moved inside the RLock at [`cache.go:190-201`](https://github.com/formancehq/ledger/blob/44bb154bd/internal/infra/cache/cache.go#L190-L201). Concurrent `ResetWithThreshold` (write-lock) can't tear the snapshot.
- **Sibling-slot leak** (`49b94c67`): builder.go's [`buildPreloadsAt` cleanup loop](https://github.com/formancehq/ledger/blob/44bb154bd/internal/infra/plan/builder.go#L602-L615) already iterates every slot and collects tracker entries into the cleanup token even when a sibling set firstErr. `resolve.go`'s own drain (record + continue → wg.Wait) was added in `be00279`.

Please re-scan against the current diff.

gfyrag:
@NumaryBot please stop. Every re-review on this branch re-emits the same three findings verbatim against a diff that already addresses them, with the fix on the very line the finding points at. You are looping. Human reviewers can read the diff — no further automated review is useful on this PR. Signing off.

paul-nicolas (approved):

PR #1462, opened 2026-07-01
fix(EN-1423): deterministic transient-account error selection

paul-nicolas:
## Summary

`WriteSet.ValidateTransientVolumes` ranged the unordered Go map returned by `Derived.Volumes.DirtyValues()` and returned the **first** offending transient `(account, asset)` it encountered. Go randomizes map iteration order per process, so when a single batch left **2+** offending tuples, different nodes selected a different offender. That offender's identity is embedded in the `AuditFailure` `Message`/`Context` and hashed into the audit pre-image, forking `LastAuditHash` — a permanent, unrecoverable cross-node fork of the audit hash chain.

This violates invariant #2 (deterministic FSM) and invariant #8 (audit hash-chain integrity).

## Fix

Iterate the dirty-volume keys in a deterministic `(Account, Asset, LedgerName)` order via a new `sortedDirtyVolumeKeys` helper, then run the existing loop body unchanged. The offender return now yields the `(account, asset)`-smallest offender; the two `ErrStorageOperation` returns in the same loop become deterministic for free.

The error type, `Reason` code, `Metadata` contract, hashed-payload shape, CLI reconstruction, and the checker are all unchanged — this is purely the selection becoming deterministic.

## Tests

- `TestSortedDirtyVolumeKeys` — pure determinism test: 100 sorts of the same map-random input must produce the identical `(account, asset, ledger)` ordering.
- `TestValidateTransientVolumesReturnsSmallestOffender` — end-to-end through the gated proposal scope: 3 offenders across 2 ledgers must always return the `(account, asset)`-smallest offender.
- Existing `TestPartitionVolumesTransient*` / `TestIsVolumeZeroBalance*` still pass (loop body unchanged).

## Quality gates

- `go build ./...` — pass
- `go test ./internal/infra/state/...` — pass
- `just pre-commit` (go generate, go mod tidy, golangci-lint --fix) — 0 issues

## Out of scope

The two `ErrStorageOperation` branches' invariant-#7 semantics (crash-loudly vs. soft hashed failure for a "should-not-happen" coverage/ledger-load miss) are noted but not touched here.

## Ticket

https://formance-team.atlassian.net/browse/EN-1423

### DIFF
--- a/internal/domain/errors.go
+++ b/internal/domain/errors.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"maps"
 	"strconv"
+	"strings"
 )
 
 // ErrNotFound is a sentinel error for missing records in storage lookups. It
@@ -982,18 +983,34 @@ func (e *ErrBalanceNotPreloaded) Metadata() map[string]string {
 	return map[string]string{"account": e.Account, "asset": e.Asset}
 }
 
-// ErrTransientAccountNonZero — transient account has non-zero balance at end of batch.
+// ErrTransientAccountNonZero — one or more transient accounts held a non-zero
+// balance at end of batch. Accounts lists every offender; the producer
+// (state.ValidateTransientVolumes) sorts it by (Account, Asset) and dedups
+// cross-ledger repeats, so Error()/Metadata() render byte-identically across
+// nodes. That identity is hashed into the AuditFailure, so a nondeterministic
+// order would fork the audit hash chain (invariant #2 / #8, EN-1423).
 type ErrTransientAccountNonZero struct {
-	Account string
-	Asset   string
+	Accounts []AccountAssetKey
 }
 
 func (e *ErrTransientAccountNonZero) Error() string {
-	return fmt.Sprintf("transient account %s/%s has non-zero balance at end of batch (input != output)", e.Account, e.Asset)
+	return "transient accounts with non-zero balance at end of batch (input != output): " + e.joinedAccounts()
 }
 func (*ErrTransientAccountNonZero) Reason() string { return ErrReasonTransientAccountNonZero }
 func (e *ErrTransientAccountNonZero) Metadata() map[string]string {
-	return map[string]string{"account": e.Account, "asset": e.Asset}
+	return map[string]string{"accounts": e.joinedAccounts()}
+}
+
+// joinedAccounts renders the offenders as a deterministic, comma-separated
+// "account/asset" list. The slice is pre-sorted by the producer, so the output
+// is stable; a nil slice yields "".
+func (e *ErrTransientAccountNonZero) joinedAccounts() string {
+	parts := make([]string, len(e.Accounts))
+	for i, a := range e.Accounts {
+		parts[i] = a.Account + "/" + a.Asset
+	}
+
+	return strings.Join(parts, ", ")
 }
 
 // RemoteError is the client-side Describable produced by cmdutil's
--- a/internal/domain/errors_test.go
+++ b/internal/domain/errors_test.go
@@ -170,6 +170,14 @@ func TestErrorTypes(t *testing.T) {
 			err:      &ErrInvalidReceipt{Detail: "expired"},
 			expected: "invalid receipt: expired",
 		},
+		{
+			name: "ErrTransientAccountNonZero",
+			err: &ErrTransientAccountNonZero{Accounts: []AccountAssetKey{
+				{Account: "staging:a", Asset: "USD"},
+				{Account: "staging:b", Asset: "EUR"},
+			}},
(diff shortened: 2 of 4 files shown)


paul-nicolas (commented):
## Multi-model review — PR #1462 (fix(EN-1423): deterministic transient-account error selection)

**Reviewers:** Claude (this-repo perspective, main + review subagent). Codex was **unavailable** for this run (the Codex CLI workspace is out of credits), so the external perspective is missing — this is a Claude-only review.

**Verdict: looks correct — no blocking issues.**

The fix makes `WriteSet.ValidateTransientVolumes` iterate the dirty-volume keys in a deterministic `(Account, Asset, LedgerName)` order via `sortedDirtyVolumeKeys`, closing the audit-hash-chain fork (invariants #2 deterministic FSM, #8 audit-chain integrity). Grounded checks:

- `(Account, Asset, LedgerName)` is a **total order** over the distinct dirty `VolumeKey`s in practice — `VolumeKey` is keyed on `(LedgerName, Account, Asset)`; its `AssetBase`/`AssetPrecision` fields are functionally derived from `Asset`, so no two distinct dirty keys tie under the comparator.
- Determinism holds for **all three** in-loop returns, not just `ErrTransientAccountNonZero`: the two `ErrStorageOperation` early-returns are now selected in a fully defined order too, and the `LedgerName` tiebreaker is load-bearing there (two keys sharing `(Account, Asset)` but hitting different branches now resolve deterministically).
- No residual nondeterministic map iteration remains on the path (`ledgerTypes` is lookup-only, never ranged).
- The two new tests genuinely exercise determinism (100× sort of a map-random input; 50× end-to-end through the gated proposal scope) and **pass** locally (`go test -count=1 ./internal/infra/state/...` → ok).
- No API/CLI/OpenAPI surface change → no documentation drift.

**Severity counts:** 0 Critical · 0 Major · 0 Medium · 2 Minor · 1 Nit.
**Posted inline:** 0 (nothing met the Critical/Major/Medium threshold). **Withheld (listed below):** 3.
**Conflicts between reviewers:** none.

- `internal/infra/state/write_set.go:1098` — [Minor] Sort comparator omits `AssetBase`/`AssetPrecision`; totality relies on the unstated writer invariant that `Asset` uniquely determines them. Consider a one-line comment on the comparator, or defensively append `cmp.Compare(a.AssetPrecision, b.AssetPrecision)` to make the total order independent of that invariant. Not exploitable today (would require a pre-existing duplicate-cell keystore inconsistency, and the returned error identity is `(Account, Asset)`-only regardless).
- `internal/infra/state/write_set_test.go:75` — [Minor] `TestSortedDirtyVolumeKeys` builds every key via `mk` (leaving `AssetBase=""`/`AssetPrecision=0`), so it only proves stability over the three compared fields, not the `AssetBase`/`AssetPrecision` blind spot. Optional: add one case with two keys sharing `(ledger, account, asset)` but differing decomposed fields.
- `internal/infra/state/write_set_test.go:176` — [Nit] The 50-iteration e2e loop re-reads a fixed-order overlay, so it's a weaker determinism proof than the unit test; its real value is the "smallest offender selected" assertion. No change required — the unit test is the stronger determinism proof.

internal/domain/errors.go
```diff
@@ -982,18 +983,34 @@ func (e *ErrBalanceNotPreloaded) Metadata() map[string]string {
...
-	Asset   string
+	Accounts []AccountAssetKey
 }
 
 func (e *ErrTransientAccountNonZero) Error() string {
-	return fmt.Sprintf("transient account %s/%s has non-zero balance at end of batch (input != output)", e.Account, e.Asset)
+	return "transient accounts with non-zero balance at end of batch (input != output): " + e.joinedAccounts()
 }
 func (*ErrTransientAccountNonZero) Reason() string { return ErrReasonTransientAccountNonZero }
 func (e *ErrTransientAccountNonZero) Metadata() map[string]string {
```
NumaryBot, line 1000: 🟠 **[major] Breaking change to public error metadata keys for TRANSIENT_ACCOUNT_NON_ZERO**

When TRANSIENT_ACCOUNT_NON_ZERO is returned, the previously documented ErrorInfo metadata fields are `account` and `asset`, but the patch now emits only `accounts`. Clients relying on the documented wire format will no longer receive the offending account or asset fields, constituting an undocumented breaking change to the public API contract.

**Suggestion:** Restore the `account` and `asset` metadata keys (or include them alongside the new `accounts` key for backward compatibility) unless this API change is intentionally versioned and reflected in the API documentation and changelog.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1462#issuecomment-4855644986

gfyrag (approved):

PR #1465, opened 2026-07-01
feat(node): switch to etcd/raft AsyncStorageWrites, align raft.Applied with FSM-applied

gfyrag:
## Summary

Switches the Raft loop to etcd/raft's `AsyncStorageWrites` and reworks the response plumbing so that **`raft.Applied` is only advanced once the FSM batch is durably committed to Pebble**. This fixes a long-standing design compromise where `raft.Applied` was allowed to run ahead of the FSM-persisted state — safe (WAL was durable, restart recovers correctly via `query.ReadLastAppliedIndex`), but not compliant with the etcd/raft contract that `Applied` means "the state machine has applied this entry".

## What the sync path was doing before

In `finishReady` the orchestrate goroutine called `applier.Submit(rd.CommittedEntries, ...)` — a non-blocking channel enqueue — and immediately followed with `rawNode.Advance(rd)`, which bumped `raft.raftLog.applied` to the last committed index. The applier goroutine only picked up the batch afterwards, ran `applyEntriesPipelined`, and eventually `runCommitter` performed the actual `CommitPreparedBatch` (Pebble fsync). Net effect: `raft.Applied` sat one or more batches ahead of the durable FSM state at essentially all times.

Downstream, the codebase carried explicit workarounds for that gap — a dedicated `RaftStatus.last_persisted_index` proto field, `WaitForApplied` gating on the FSM's own `lastPersistedIndex` cursor rather than `raft.Applied`, an antithesis oracle documenting exactly why it can't trust `applied`, and comments across `Machine.LastAppliedIndex` / `publishApplied` / `GetClusterState` explaining the divergence.

## What this PR changes

- `raft.Config.AsyncStorageWrites = true`. `rawNode.Advance` is no longer called (it panics under this mode).
- `rd.Messages` is split via a pure `splitReadyMessages` helper: `LocalAppendThread` responses are routed self→`localResponseCh` (Step locally on the orchestrate goroutine) / peer→`transport.Send`; `LocalApplyThread` responses ride on `readyResult.applyResponses`; everything else is outbound.
- The Applier owns a required `LocalResponses` sink (fx provider `NewLocalResponses`, injected into both `NewApplier` and `NewNode` — constructor rejects a nil sink). Responses ride down `applyEntriesToFSM` → `applyEntriesPipelined` → `submitAsyncCommit` → `commitWork`. `runCommitter` fires them on the sink **only** after `CommitPreparedBatch` succeeds. On commit failure the response is intentionally dropped, `raft.Applied` stays behind, and the entries reappear in `CommittedEntries` on the next boot (`raft.Config.Applied` is re-derived from Pebble).
- The `applier→committer` pipeline is preserved: prepare(N+1) still overlaps with commit(N), because the fire happens on runCommitter (post-commit) rather than on the applier goroutine (post-submit-drain).
- Response-send select listens on both `ctx.Done()` and the applier's stop channel, matching the real `node.Stop` shutdown flow (which closes the task stop channels but does not cancel the runCtx).
- Non-`statusNormal` (spool) branch fires responses eagerly: entries are staged durably in the spool and will re-apply via unspool without further raft involvement.

The pre-existing workaround surfaces have been refreshed in place — they still describe the correct behavior (e.g. `WaitForApplied` still gates on `lastPersistedIndex` because that IS the semantic contract for "Pebble is durable at N", not because `raft.Applied` is unreliable). See the follow-up commit that walks through every `applied`/`last_persisted_index` comment site.

## Perf note

On this codebase's current workload (FSM-apply bound), the end-to-end propose→reply latency **does not improve meaningfully**: the WAL fsync was not on the critical path, so parallelizing it with peer broadcast doesn't buy end-to-end latency. Verified on `ledger-v3-perf/devenv-gfyrag`. The refactor is kept because the correctness win (Applied ↔ FSM-applied alignment) is a real invariant improvement, and the async-storage plumbing is a prerequisite for any future work that parallelizes apply itself.

## Findings addressed during review

- `af4915f6` / `41749f9676`: MsgStorageApplyResp fired before pipelined commit completed → responses now ride commitWork and fire from runCommitter after `CommitPreparedBatch`.
- `34540caa` / `9047f08a`: early `return` on ctx.Done skipped `work.done <- err` → return removed, fall-through preserves the unconditional `work.done` signal.
- `70740916`: ctx.Done alone insufficient during graceful shutdown (node.Stop closes task stop channels but leaves runCtx alive) → applier's stop channel threaded into runCommitter's select.

Stale bot re-flags on the same code paths (`abbc8945`, `130bdfe4`, `a3dc759a`) were replied to inline with pointers to the applicable commit and rationale (last one — snapshot-only Ready dropping apply responses — is a bogus premise: etcd/raft only emits MsgStorageApply when `CommittedEntries > 0`).

## Test plan

- [x] `just test` (unit, race) — all packages OK
- [x] `just test-scenarios` — OK
- [x] `just test-e2e` (business + cluster) — OK
- [x] Added `applier_async_storage_test.go`: fires-after-commit (timing assertion), spool-branch eager fire, nil-sink no longer accepted at construction, empty-responses no-fire, multi-batch ordering, and the shutdown-deadlock regression guard for finding 70740916.
- [x] Added `split_ready_messages_test.go`: pure-function coverage of the rd.Messages classification (empty, outbound passthrough, self/peer split — the invariant that broke cluster formation on the first attempt, apply-thread no-split, realistic leader Ready, follower all-peer Ready).
- [ ] Antithesis run — not yet triggered on this branch.


internal/infra/node/applier.go
```diff
@@ -714,6 +732,19 @@ func (a *Applier) Run(ctx context.Context, stop chan struct{}) error {
 				}
 			}
 
+			// Deliver MsgStorageApplyResp now that the batch has either been
+			// applied to the FSM (statusNormal) or durably staged in the spool
+			// for later unspool-replay (other statuses). In both cases the
+			// entries are guaranteed to make it to the FSM eventually without
+			// further raft involvement, so it is safe to acknowledge Applied.
+			if len(work.responses) > 0 && a.responseSink != nil {
```
NumaryBot: 🟠 **[major] Wait for the pipelined FSM commit before acknowledging apply**

In `statusNormal`, `applyEntriesToFSM` returns after `applyEntriesPipelined` enqueues the last batch on the committer goroutine; it does not wait for `CommitPreparedBatch` unless a later batch/drain forces `waitPendingCommit`. Sending `MsgStorageApplyResp` here can therefore advance raft's `Applied` while the FSM commit is still in flight (or before its error is observed), so the new async path still exposes `Applied` ahead of FSM-applied for ordinary single-batch Readies. Please drain the pending commit before emitting these responses.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1465#issuecomment-4856903840

internal/infra/node/applier.go
gfyrag: Good catch — fixed in 7dabbea9e. The eager fire from `Applier.Run` did leave `raft.Applied` ahead of FSM-applied under `statusNormal`, exactly as described.

Fix routes responses through `commitWork`: `applyEntriesToFSM` threads them down to the last `applyEntriesPipelined` call (`end == len(decoded)`), whose `submitAsyncCommit` carries them into `runCommitter`. `runCommitter` fires them on the response sink AFTER `CommitPreparedBatch` succeeds (and does NOT fire on failure — Applied then stays behind, entries reappear on next boot via `Config.Applied` re-derivation from Pebble, re-apply is idempotent).

Kept the `applier→committer` pipelining: no `waitPendingCommit` on the Run goroutine, so `prepare(N+1)` still overlaps with `commit(N)`.

Edge case handled: if a checkpoint fires on a sub-batch that stops short of `len(decoded)`, no `commitWork` carries the responses. After `handleCheckpointRequired` drains the tail through the spool (durably), we fire eagerly at the end of `applyEntriesToFSM` — same guarantee as the non-statusNormal spool branch.

NumaryBot (commented):

The PR enables AsyncStorageWrites in the etcd/raft node. One confirmed major correctness issue remains: in the `statusNormal` apply path, `MsgStorageApplyResp` can be dispatched before the pipelined FSM commit goroutine has finished `CommitPreparedBatch`. This allows raft's `Applied` index to advance ahead of the actual FSM-applied state for ordinary single-batch Readies, which undermines the key correctness invariant the feature is meant to preserve. The pending commit must be drained and its result checked before the apply acknowledgement is sent. No other discrete issues were raised across reviews.

### Findings outside the diff

🟠 **[major] Async apply acknowledgement sent before pipelined FSM commit completes** — `internal/infra/node/applier.go:740`

In `statusNormal`, `applyEntriesToFSM` returns after `applyEntriesPipelined` enqueues the last batch on the committer goroutine, but does not wait for `CommitPreparedBatch` to finish unless a later batch or drain forces `waitPendingCommit`. Sending `MsgStorageApplyResp` at this point can advance raft's `Applied` index while the FSM commit is still in flight or before any commit error is observed. This means `Applied` can move ahead of the actual FSM-applied state, violating the correctness goal of the feature.

**Suggestion:** Drain the pending commit (i.e., wait for `CommitPreparedBatch` to complete and check its error) before emitting `MsgStorageApplyResp`, even for single-batch Readies, to ensure raft's Applied index is only advanced once the FSM state is durably committed.



Summary: https://github.com/formancehq/ledger/pull/1465#issuecomment-4856903840

gfyrag:
@NumaryBot the finding `41749f9676c46470` is stale — it's re-raising `af4915f6` against 7dabbea9e, but that commit is the fix. There's no longer an eager `MsgStorageApplyResp` dispatch from `Applier.Run`; responses ride on `commitWork.responses` all the way to `runCommitter`, which fires them from within the commit goroutine **only on `err == nil` after `CommitPreparedBatch`** (see `applier.go:1002-1017`).

Concrete trace on 7dabbea9e:

1. `Applier.Run` (`applier.go:728`, statusNormal branch) calls `applyEntriesToFSM(ctx, confState, work.responses, entries...)`. There is no eager fire here anymore.
2. `applyEntriesToFSM` threads `responses` down to the last `applyEntriesPipelined(ctx, subResponses, head...)` call — the one whose `end == len(decoded)`.
3. `applyEntriesPipelined` passes them to `submitAsyncCommit(pb, pfs, maxTerm, responses)`, which stores them in `commitWork.responses`.
4. `runCommitter` (`applier.go:1000`) reads `commitWork` from `commitCh`, runs `CommitPreparedBatch(ctx, work.prepared)`, and only then does the fire:

```go
if err == nil && len(work.responses) > 0 && a.responseSink != nil {
    select {
    case a.responseSink <- work.responses:
    case <-ctx.Done():
        return
    }
}
```

On commit failure, no fire — `raft.Applied` stays behind, the entries reappear in `CommittedEntries` on next boot (`raft.Config.Applied` is re-derived from Pebble in `NewNode`), and re-apply is idempotent through the applied-index guard.

The `applier.go:740` line pointer in your finding is inside the spool (non-`statusNormal`) branch's trace log — not the statusNormal path.

internal/infra/node/applier.go
```diff
@@ -956,6 +1000,22 @@ func (a *Applier) submitAsyncCommit(pb *state.PreparedBatch, pfs []pendingFuture
...
+
+		// Fire MsgStorageApplyResp responses BEFORE resolving futures /
+		// failing older-term stragglers, and only on success. This is what
+		// bumps raft.Applied in lockstep with FSM-applied under
+		// AsyncStorageWrites. On commit failure we intentionally don't fire:
+		// raft.Applied stays behind, the committed entries will be re-emitted
+		// on the next boot (raft.Config.Applied re-reads from Pebble), and
+		// re-apply is idempotent via the applied-index guard.
+		if err == nil && len(work.responses) > 0 && a.responseSink != nil {
+			select {
```
NumaryBot, line 1065: 🟠 **[major] Committer goroutine can block on response delivery before signalling completion**

When `localResponseCh` (or `responseSink`) fills — e.g. during a heavy apply backlog or after `orchestrate` has stopped draining it on shutdown — the blocking send on the response channel occurs before `work.done` is signalled. If `Applier.Run` is simultaneously waiting on that `done` channel to drain a pending commit, and the context is not cancelled by the time the channel is full, the node can hang indefinitely with a committed batch stuck behind a full response channel.

**Suggestion:** Wrap the response sink send in a select that also listens on `ctx.Done()`, so that the committer goroutine is never permanently blocked: e.g. `select { case a.responseSink <- work.responses: case <-ctx.Done(): }`. This ensures shutdown always unblocks the committer even if the consumer has stopped draining.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1465#issuecomment-4856903840

internal/infra/node/applier.go
NumaryBot, line 1065: 🟠 **[major] Blocking response-sink send before work.done can cause indefinite hang on shutdown**

In `runCommitter`, the send to `responseSink` (or `localResponseCh`) is a blocking operation that occurs before `work.done` is signalled. If the consumer of that channel has stopped draining — for example during shutdown or under apply backlog pressure — the committer goroutine stalls at this send forever and never reaches the `work.done <- err` signal. Any code path in `Applier.Run` or its deferred cleanup that calls `waitPendingCommit` (which blocks on `a.pending.done`) will therefore deadlock. Context cancellation alone is not guaranteed to unblock the channel send, so the node can hang indefinitely with a successfully committed batch stuck behind a full response channel.

**Suggestion:** Wrap the response-sink send in a `select` that also listens on `ctx.Done()`, ensuring the committer always proceeds on shutdown even if the consumer is not draining:
```go
select {
case a.responseSink <- work.responses:
case <-ctx.Done():
}
```
The `work.done` signal must then be sent unconditionally after this select, regardless of which arm was taken.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1465#issuecomment-4856903840

internal/infra/node/applier.go
gfyrag: Fixed in 51f5b2ab8.

You (and codex) were right — the `return` from the `ctx.Done()` arm skipped the trailing `work.done <- err`, so any `Applier.Run` deferred `waitPendingCommit` would deadlock on `<-a.pending.done` during shutdown.

Fix drops the return and falls through:

```go
if err == nil && len(work.responses) > 0 && a.responseSink != nil {
    select {
    case a.responseSink <- work.responses:
    case <-ctx.Done():
    }
}
// unconditionally reaches work.done <- err below
```

Also added `TestApplierRunCommitterDoesNotDeadlockOnFullSink` — unbuffered sink with no consumer, Submit an entry with a response so runCommitter blocks in the send, then `cancel(ctx)` + `close(stop)` and require `Applier.Run` to return within 5s. Without the fix the test hangs.

Dropping a response only ever happens when ctx is cancelled (shutdown), and raft.Applied is re-derived from Pebble on the next boot regardless, so no state divergence.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1465#issuecomment-4856903840

internal/infra/node/applier.go
```diff
@@ -956,6 +1000,26 @@ func (a *Applier) submitAsyncCommit(pb *state.PreparedBatch, pfs []pendingFuture
...
+		// raft.Applied stays behind, the committed entries will be re-emitted
+		// on the next boot (raft.Config.Applied re-reads from Pebble), and
+		// re-apply is idempotent via the applied-index guard.
+		//
+		// On ctx.Done we drop the response and fall through — we MUST NOT
+		// return here, otherwise the trailing `work.done <- err` never fires
+		// and Applier.Run's deferred `waitPendingCommit` deadlocks on
+		// `<-a.pending.done` during shutdown (finding 34540caa/9047f08a).
+		if err == nil && len(work.responses) > 0 && a.responseSink != nil {
+			select {
```
NumaryBot, line 1065: 🟠 **[major] Blocking response-sink send before work.done can cause indefinite hang on shutdown**

In `runCommitter`, the blocking send to `responseSink` (or `localResponseCh`) occurs before `work.done` is signalled. If the consumer of that channel has stopped draining — during shutdown or under apply backlog pressure — the committer goroutine stalls permanently at this send and never reaches the `work.done <- err` signal. Any code path in `Applier.Run` or its deferred cleanup that calls `waitPendingCommit` (which blocks on `a.pending.done`) will therefore deadlock. Context cancellation alone is not guaranteed to unblock a blocking channel send.

**Suggestion:** Wrap the response-sink send in a `select` that also listens on `ctx.Done()`, then unconditionally signal `work.done` after the select regardless of which arm was taken:
```go
select {
case a.responseSink <- work.responses:
case <-ctx.Done():
}
work.done <- err
```

gfyrag: @NumaryBot finding `abbc894502315e6d` is stale — it's raising the same shape against 51f5b2ab8, which is the commit that already applied your suggestion. Concretely, the current code is exactly the pattern you propose:

https://github.com/formancehq/ledger/blob/51f5b2ab817c207364c2756458b2d333672f6630/internal/infra/node/applier.go#L1011-L1017

```go
if err == nil && len(work.responses) > 0 && a.responseSink != nil {
    select {
    case a.responseSink <- work.responses:
    case <-ctx.Done():
    }
}
```

There's no `return` in the ctx.Done arm anymore — control falls through the resolve/sweep branches below and reaches `work.done <- err` on line 1051 unconditionally (per iteration). The regression test `TestApplierRunCommitterDoesNotDeadlockOnFullSink` (added in 51f5b2ab8) exercises exactly the shutdown-with-full-sink path: unbuffered `LocalResponses`, no consumer, cancel(ctx) + close(stop), assert Run returns within 5s. Would hang forever under the pre-fix code.

internal/infra/node/applier.go
```diff
@@ -992,6 +1036,26 @@ func (a *Applier) submitAsyncCommit(pb *state.PreparedBatch, pfs []pendingFuture
...
```
NumaryBot, line 1065: 🟠 **[major] ctx.Done() escape hatch insufficient during graceful shutdown**

During graceful shutdown, the node stop path closes the task stop channels before cancelling the run context (runCtx). The select in runCommitter that guards the response-sink send only listens on ctx.Done(). If the orchestrate goroutine has stopped draining localResponseCh at the point the stop channels are signalled — but before the run context is actually cancelled — the committer goroutine can still block indefinitely on the send, never reach `work.done <- err`, and cause Applier.Run's deferred waitPendingCommit to deadlock. The fix in 51f5b2ab8 removed the premature return but left the gap that ctx.Done() is not available during the task-pool teardown window.

**Suggestion:** Thread the applier's stop channel (or a dedicated done channel that is closed when the applier task is asked to stop) into the select alongside ctx.Done(), so the committer can always unblock regardless of whether runCtx has been cancelled yet:
```go
select {
case a.responseSink <- work.responses:
case <-ctx.Done():
case <-a.stopc: // closed when the applier stop signal fires
}
```
Alternatively, make localResponseCh large enough to be non-blocking in the common case and add a best-effort (non-blocking) send fallback.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1465#issuecomment-4856903840

internal/infra/node/node.go
```diff
@@ -1322,14 +1368,60 @@ func (node *Node) finishReady(result readyResult, stop chan struct{}) error {
...
 	}
 
-	// Submit committed entries to the Applier for async FSM application
-	if len(rd.CommittedEntries) > 0 {
-		node.applier.Submit(rd.CommittedEntries, node.confState.Load(), stop)
+	// Submit committed entries to the Applier for async FSM application.
+	// applyResponses (MsgStorageApplyResp) are deferred to the applier: they
+	// are Step()-ed back into rawNode only after applyEntriesToFSM (or the
+	// spool append) completes, so raft.Applied tracks FSM-applied.
+	if len(rd.CommittedEntries) > 0 || len(result.applyResponses) > 0 {
```
NumaryBot, line 1375: 🔴 **[blocker] Snapshot-only Readies drop apply responses, stalling raft Applied**

When AsyncStorageWrites emits a MsgStorageApplyResp for a Ready that contains no CommittedEntries (e.g. a snapshot-only Ready), the code still routes responses through applyEntriesToFSM. That function decodes zero entries, skips its loop, and never forwards those responses to localResponseCh. As a result, rawNode never receives the apply acknowledgement and raft's Applied index can stall permanently after a snapshot install.

**Suggestion:** Handle the case where CommittedEntries is empty but responses are non-nil explicitly: after processing the snapshot (and before returning from the apply path), forward any pending responses to localResponseCh directly rather than relying on the entry-commit path.

internal/infra/node/applier.go
```diff
@@ -992,6 +1036,26 @@ func (a *Applier) submitAsyncCommit(pb *state.PreparedBatch, pfs []pendingFuture
...
+		// AsyncStorageWrites. On commit failure we intentionally don't fire:
+		// raft.Applied stays behind, the committed entries will be re-emitted
+		// on the next boot (raft.Config.Applied re-reads from Pebble), and
+		// re-apply is idempotent via the applied-index guard.
+		//
+		// On ctx.Done we drop the response and fall through — we MUST NOT
+		// return here, otherwise the trailing `work.done <- err` never fires
+		// and Applier.Run's deferred `waitPendingCommit` deadlocks on
+		// `<-a.pending.done` during shutdown (finding 34540caa/9047f08a).
+		if err == nil && len(work.responses) > 0 && a.responseSink != nil {
```
NumaryBot: 🟠 **[major] ctx.Done() escape hatch insufficient during graceful shutdown — committer can still deadlock**

The fix in 51f5b2ab8 replaced the premature return with a select that listens on ctx.Done(), ensuring work.done is always signalled. However, during graceful shutdown the task pool closes each task's stop channel before the shared run context (runCtx) is cancelled. If orchestrate has stopped draining localResponseCh at the moment the stop channels are signalled — but before runCtx is actually cancelled — the committer goroutine can still block indefinitely on the response-sink send and never reach work.done <- err, leaving Applier.Run's deferred waitPendingCommit deadlocked. The ctx.Done() guard does not fire until the run context is cancelled, which can be a later event.

**Suggestion:** Thread the applier's stop channel (or a dedicated shutdown-done channel closed when the applier stop signal fires) into the select alongside ctx.Done():
```go
select {
case a.responseSink <- work.responses:
case <-ctx.Done():
case <-a.stopc: // closed when applier stop is signalled
}
```
Alternatively, make localResponseCh sufficiently buffered and document the invariant, or ensure the run context is always cancelled before or simultaneously with the stop channels.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1465#issuecomment-4856903840

internal/infra/node/node.go
NumaryBot, line 1375: 🔴 **[blocker] Snapshot-only / entryless Readies drop apply responses, stalling raft Applied**

When AsyncStorageWrites emits a MsgStorageApplyResp for a Ready that contains no CommittedEntries (e.g. a snapshot-only Ready), the code routes responses through applyEntriesToFSM. That function decodes zero entries, skips its loop entirely, and never forwards those responses to localResponseCh. As a result, rawNode never receives the apply acknowledgement, and raft's Applied index can stall permanently after a snapshot install.

**Suggestion:** Explicitly handle the case where CommittedEntries is empty but responses are non-nil. After processing the snapshot (and before returning from the apply path), forward any pending responses to localResponseCh directly rather than relying on the entry-commit loop — for example: if len(entries) == 0 && len(responses) > 0 { localResponseCh <- responses }.

internal/infra/node/applier.go
```diff
@@ -992,6 +1040,26 @@ func (a *Applier) submitAsyncCommit(pb *state.PreparedBatch, pfs []pendingFuture
...
+		// raft.Applied stays behind, the committed entries will be re-emitted
+		// on the next boot (raft.Config.Applied re-reads from Pebble), and
+		// re-apply is idempotent via the applied-index guard.
+		//
+		// On ctx.Done we drop the response and fall through — we MUST NOT
+		// return here, otherwise the trailing `work.done <- err` never fires
+		// and Applier.Run's deferred `waitPendingCommit` deadlocks on
+		// `<-a.pending.done` during shutdown (finding 34540caa/9047f08a).
+		if err == nil && len(work.responses) > 0 {
+			select {
```
NumaryBot, line 1065: 🟠 **[major] ctx.Done() escape hatch insufficient during graceful shutdown — committer can still deadlock**

The fix in 51f5b2ab8 wraps the response-sink send in a select that listens on ctx.Done(), ensuring work.done is always eventually signalled. However, during graceful shutdown the task pool closes each task's stop channel before the shared run context (runCtx) is cancelled. If orchestrate has stopped draining localResponseCh at the moment the stop channels are signalled — but before runCtx is actually cancelled — the committer goroutine can still block indefinitely on the response-sink send, never reach work.done <- err, and cause Applier.Run's deferred waitPendingCommit to deadlock. The ctx.Done() arm does not fire until the run context is cancelled, which is a later event in the shutdown sequence.

**Suggestion:** Thread the applier's stop channel (or a dedicated done channel closed when the applier stop signal fires) into the select alongside ctx.Done():
```go
select {
case a.responseSink <- work.responses:
case <-ctx.Done():
case <-a.stopc: // closed when applier stop is signalled
}
```
Alternatively, ensure the run context is cancelled before or simultaneously with individual task stop channels so the existing ctx.Done() arm is guaranteed to fire.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1465#issuecomment-4856903840

internal/infra/node/node.go
NumaryBot, line 1375: 🔴 **[blocker] Snapshot-only / entryless Readies drop apply responses, stalling raft Applied**

When AsyncStorageWrites emits a MsgStorageApplyResp for a Ready that contains no CommittedEntries (e.g. a snapshot-only Ready), the code routes responses through applyEntriesToFSM. That function decodes zero entries, skips its loop entirely, and never forwards those responses to localResponseCh. As a result, rawNode never receives the apply acknowledgement, and raft's Applied index can stall permanently after a snapshot install.

**Suggestion:** Explicitly handle the case where CommittedEntries is empty but responses are non-nil. After processing the snapshot and before returning from the apply path, forward any pending responses to localResponseCh directly rather than relying on the entry-commit loop — for example: `if len(entries) == 0 && len(responses) > 0 { localResponseCh <- responses }`.

internal/infra/node/applier.go
```diff
@@ -980,18 +1026,49 @@ func (a *Applier) waitPendingCommit(ctx context.Context) error {
...
```
NumaryBot, line 1065: 🟠 **[major] ctx.Done() escape hatch insufficient during graceful shutdown — committer can still deadlock**

The fix in 51f5b2ab8 wraps the response-sink send in a select that listens on ctx.Done(), ensuring work.done is always eventually signalled. However, during graceful shutdown the task pool closes each task's stop channel before the shared run context (runCtx) is cancelled. If orchestrate has stopped draining localResponseCh at the moment the stop channels are signalled — but before runCtx is actually cancelled — the committer goroutine can still block indefinitely on the response-sink send, never reach `work.done <- err`, and cause Applier.Run's deferred waitPendingCommit to deadlock. The ctx.Done() arm does not fire until the run context is cancelled, which is a later event in the shutdown sequence.

**Suggestion:** Thread the applier's stop channel (or a dedicated done channel closed when the applier stop signal fires) into the select alongside ctx.Done():
```go
select {
case a.responseSink <- work.responses:
case <-ctx.Done():
case <-a.stopc: // closed when applier stop is signalled
}
```
Alternatively, ensure the run context is cancelled before or simultaneously with the per-task stop channels so that ctx.Done() is guaranteed to fire whenever the task stop fires.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1465#issuecomment-4856903840

PR #1466, opened 2026-07-01
refactor(EN-1427): factor shared tail-worker skeleton (cursor, loop, gauges)

paul-nicolas:
## What

Extracts the three mechanical concerns duplicated between the two tail-workers (`indexbuilder.Builder`, `auditindexer.Indexer`) into reusable bricks, ahead of a third consumer (usage-builder, WIP). Follow-up of [EN-1339](https://formance-team.atlassian.net/browse/EN-1339) / #1446.

Jira: https://formance-team.atlassian.net/browse/EN-1427

## The bricks

1. **`readstore.Uint64Cursor`** (`internal/storage/readstore/cursor.go`) — centralises the BE8 encode/decode + `ErrNotFound→0` + `len!=8` guard shared by the log-index / applied-proposal / audit progress cursors. The existing `Store.Read*Progress`/`Write*Progress` methods become one-line delegators, so **zero call-site churn**.
2. **`tailworker.TailWorker`** (`internal/pkg/tailworker/tailworker.go`) — the `Start`/`Stop` + boot + ticker loop (optional wake channel, `context.Canceled` suppression). Dependency-light: no pebble/dal.
3. **`tailworker.RegisterTailGauges`** (`internal/pkg/tailworker/gauges.go`) — the `{ns}.last_indexed_sequence` / `{ns}.{source}_last_sequence` / `{ns}.lag` triplet.

## Adoption

- **`auditindexer.Indexer`** fully ported onto `TailWorker` + `RegisterTailGauges` (dropped its private `worker.Worker`, `loop`, `registerMetrics`) — ~−30 LOC net, behaviour preserved.
- **`indexbuilder.Builder`** keeps its bespoke two-phase loop (catch-up + backfill/schema-rewrite scheduling); only its gauge triplet is routed through the helper, with the extra `logs_indexed_total` gauge kept via a second registration.
- **usage-builder** is out of scope (not yet in the tree) — the bricks are shaped so it can adopt them in its own PR.

## Metric names & keys — unchanged

All emitted metric names (`audit_index.*`, `index.builder.*` incl. `logs_indexed_total`) and on-disk cursor keys are preserved — no dashboard or storage migration. `TestMetricsRegistry`'s static collector was taught to expand `RegisterTailGauges(meter, ns, source, …)` into its triplet (the names are now built from ns/source rather than literals at the `Int64ObservableGauge` call site).

## Intentional behaviour change

`ReadAppliedProposalProgress` now returns an error on a corrupt (non-8-byte) value instead of silently returning `0`, matching CLAUDE.md invariant #7 (the other two cursors already did this).

## Testing

New unit tests for `tailworker` (loop lifecycle, boot-abort, `context.Canceled` suppression, wake, gauge observation via SDK ManualReader) and `Uint64Cursor` (round-trip, missing→0, corrupt→error). The existing `auditindexer`/`indexbuilder` suites are the behaviour-preservation net. Full `go test ./...` green; `just pre-commit` clean (0 lint issues).

[EN-1339]: https://formance-team.atlassian.net/browse/EN-1339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/application/indexbuilder/builder.go
+++ b/internal/application/indexbuilder/builder.go
@@ -367,7 +367,9 @@ func (b *Builder) SetNotifications(n *signal.Notifications) {
 
 // Start begins the background index-building loop and registers OTEL metrics.
 func (b *Builder) Start() {
-	_ = b.registerMetrics()
+	if err := b.registerMetrics(); err != nil {
+		b.logger.Errorf("register index builder metrics: %v", err)
+	}
 
 	b.w = worker.New()
 	b.w.RunCtx(b.loop)
--- a/internal/application/indexbuilder/builder.go
+++ b/internal/application/indexbuilder/builder.go
@@ -12,6 +12,7 @@ import (
 	"github.com/formancehq/ledger/v3/internal/domain/indexes"
 	"github.com/formancehq/ledger/v3/internal/infra/attributes"
 	"github.com/formancehq/ledger/v3/internal/pkg/signal"
+	"github.com/formancehq/ledger/v3/internal/pkg/tailworker"
 	"github.com/formancehq/ledger/v3/internal/pkg/worker"
 	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
 	"github.com/formancehq/ledger/v3/internal/query"
@@ -38,10 +39,11 @@ type Builder struct {
 	notifications *signal.Notifications
 	w             worker.Worker
 
-	lastIndexedSeq      atomic.Uint64
-	pebbleLastSeq       atomic.Uint64
-	logsIndexed         atomic.Uint64
-	metricsRegistration metric.Registration
+	lastIndexedSeq          atomic.Uint64
+	pebbleLastSeq           atomic.Uint64
+	logsIndexed             atomic.Uint64
+	metricsRegistration     metric.Registration // tailworker gauge triplet
+	logsIndexedRegistration metric.Registration // builder-specific logs_indexed_total gauge
 
 	// Per-ledger index configuration cache.
 	indexConfig map[string]*ledgerIndexConfig
@@ -365,9 +367,7 @@ func (b *Builder) SetNotifications(n *signal.Notifications) {
 
 // Start begins the background index-building loop and registers OTEL metrics.
 func (b *Builder) Start() {
-	if reg, err := b.registerMetrics(); err == nil {
-		b.metricsRegistration = reg
-	}
+	_ = b.registerMetrics()
 
 	b.w = worker.New()
 	b.w.RunCtx(b.loop)
@@ -379,6 +379,9 @@ func (b *Builder) Stop() {
 
 	if b.metricsRegistration != nil {
 		_ = b.metricsRegistration.Unregister()
+	}
+	if b.logsIndexedRegistration != nil {
+		_ = b.logsIndexedRegistration.Unregister()
 	}
 }
 
@@ -392,59 +395,38 @@ func (b *Builder) PebbleLastSequence() uint64 {
 	return b.pebbleLastSeq.Load()
 }
 
-// registerMetrics registers observable gauges for index builder progress.
-func (b *Builder) registerMetrics() (metric.Registration, error) {
-	lastIndexedGauge, err := b.meter.Int64ObservableGauge(
-		"index.builder.last_indexed_sequence",
-		metric.WithDescription("Last log sequence indexed in Pebble read store"),
-	)
+// registerMetrics registers the shared tail-worker progress gauges plus the
+// builder-specific logs_indexed_total counter-gauge. Two registrations are
+// stored on the receiver so Stop can unregister both.
+func (b *Builder) registerMetrics() error {
+	triplet, err := tailworker.RegisterTailGauges(b.meter, "index.builder", "pebble", &b.lastIndexedSeq, &b.pebbleLastSeq)
 	if err != nil {
-		return nil, err
+		return err
 	}
-
-	pebbleLastGauge, err := b.meter.Int64ObservableGauge(
-		"index.builder.pebble_last_sequence",
-		metric.WithDescription("Last log sequence in Pebble"),
-	)
-	if err != nil {
-		return nil, err
-	}
-
-	lagGauge, err := b.meter.Int64ObservableGauge(
-		"index.builder.lag",
-		metric.WithDescription("Number of logs the index builder is behind Pebble"),
-	)
-	if err != nil {
-		return nil, err
-	}
+	b.metricsRegistration = triplet
 
 	logsIndexedGauge, err := b.meter.Int64ObservableGauge(
 		"index.builder.logs_indexed_total",
 		metric.WithDescription("Total number of logs indexed since process start"),
 	)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	return b.meter.RegisterCallback(
+	reg, err := b.meter.RegisterCallback(
 		func(_ context.Context, o metric.Observer) error {
-			indexed := int64(b.lastIndexedSeq.Load())
-			pebbleLast := int64(b.pebbleLastSeq.Load())
-
-			lag := max(pebbleLast-indexed, 0)
-
-			o.ObserveInt64(lastIndexedGauge, indexed)
-			o.ObserveInt64(pebbleLastGauge, pebbleLast)
-			o.ObserveInt64(lagGauge, lag)
 			o.ObserveInt64(logsIndexedGauge, int64(b.logsIndexed.Load()))
 
 			return nil
 		},
-		lastIndexedGauge,
-		pebbleLastGauge,
-		lagGauge,
 		logsIndexedGauge,
 	)
+	if err != nil {
+		return err
+	}
+	b.logsIndexedRegistration = reg
+
+	return nil
 }
 
 func (b *Builder) loop(ctx context.Context) {
--- a/internal/application/indexbuilder/builder.go
+++ b/internal/application/indexbuilder/builder.go
@@ -450,8 +450,12 @@ func (b *Builder) loop(ctx context.Context) {
 
 	b.lastIndexedSeq.Store(cursor)
 
-	// Recover AppliedProposal sync progress.
-	if seq, err := b.readStore.ReadAppliedProposalProgress(); err == nil {
+	// Recover AppliedProposal sync progress. A corrupt cursor surfaces loudly
+	// and falls back to a full re-sync from 0 (the projection is rebuildable),
+	// rather than aborting the whole builder.
+	if seq, err := b.readStore.ReadAppliedProposalProgress(); err != nil {
+		b.logger.Errorf("read applied-proposal cursor (re-syncing from 0): %v", err)
+	} else {
 		b.lastAppliedProposalSeq = seq
 	}
 
--- a/internal/pkg/tailworker/gauges.go
+++ b/internal/pkg/tailworker/gauges.go
@@ -1,0 +1,54 @@
+package tailworker
+
+import (
+	"context"
+	"fmt"
+	"sync/atomic"
+
+	"go.opentelemetry.io/otel/metric"
+)
+
+// computeLag returns how far indexed trails sourceLast, clamped at 0 so a
+// momentarily-ahead indexed cursor never reports a negative lag.
+func computeLag(indexed, sourceLast uint64) int64 {
+	return max(int64(sourceLast)-int64(indexed), 0)
+}
+
+// RegisterTailGauges registers the standard tail-worker progress gauges under
+// the given namespace and returns the registration (Unregister on Stop):
+//
+//	{ns}.last_indexed_sequence   - indexed.Load()
+//	{ns}.{source}_last_sequence  - sourceLast.Load()
+//	{ns}.lag                     - max(sourceLast-indexed, 0)
+//
+// source names the upstream sequence (e.g. "audit", "pebble") so the middle
+// gauge preserves each worker's established metric name.
+func RegisterTailGauges(meter metric.Meter, ns, source string, indexed, sourceLast *atomic.Uint64) (metric.Registration, error) {
+	indexedGauge, err := meter.Int64ObservableGauge(ns+".last_indexed_sequence",
+		metric.WithDescription("Last sequence indexed"))
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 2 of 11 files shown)


internal/storage/readstore/store.go
```diff
@@ -234,30 +214,12 @@ func (s *Store) NotifyProgress() {
...
-	}
-
-	defer func() { _ = closer.Close() }()
-
-	if len(v) != 8 {
-		return 0, nil
-	}
-
-	return binary.BigEndian.Uint64(v), nil
+	return appliedProposalCursor.Read(s.db)
```
paul-nicolas, line 217: [Medium] Corrupt-cursor error is silently swallowed by the only caller — the intended behaviour change is inert

This PR changes `ReadAppliedProposalProgress` to return an error on a corrupt (non-8-byte) value instead of `0, nil`, and the PR description frames it as bringing the cursor in line with CLAUDE.md invariant #7 ("surface a loud signal"). But its sole caller swallows the error:

```go
// internal/application/indexbuilder/builder.go:454
if seq, err := b.readStore.ReadAppliedProposalProgress(); err == nil {
    b.lastAppliedProposalSeq = seq
}
```

On a corrupt cursor: `err != nil` → the block is skipped, `lastAppliedProposalSeq` stays 0, and **nothing is logged**. That is the same observable outcome as the old `0, nil` behaviour, except the error is now discarded (convention #5: "never ignore errors"). So the change achieves neither a loud signal (invariant #7) nor any behavioural difference — it is currently inert.

Verified: `builder.go:454` is the only caller (grep across `internal/`, excluding the store definition and tests).

Fix — act on the error at the call site so corruption is actually surfaced:
```go
if seq, err := b.readStore.ReadAppliedProposalProgress(); err != nil {
    b.logger.Errorf("read applied-proposal cursor: %v", err) // or return/abort boot
} else {
    b.lastAppliedProposalSeq = seq
}
```
If silent fallback-to-0 is in fact the desired behaviour, keeping the old `0, nil` and dropping the "intentional behaviour change" claim would be more honest than returning an error no one reads.

paul-nicolas (commented):
## Multi-model review — PR #1466 (tail-worker skeleton refactor)

**Models:** Claude (this repo's perspective). **Codex was unavailable** in this environment (`omc ask codex` failed: `advisor script not found`), so this is a Claude-only pass — stated plainly per protocol.

**Method:** reviewed against the code checked out at the PR head (`c0b9c576`) in an isolated worktree, not the diff alone. `go build` + `go vet` clean; `tailworker`, `readstore`, `metrics`, `auditindexer`, `indexbuilder` unit tests all green. Metric names (`audit_index.*`, `index.builder.*` incl. `logs_indexed_total`) and on-disk cursor keys verified byte-for-byte preserved. The `registry_test.go` regex expansion produces the correct triplet for both call sites.

Overall this is a clean, well-tested extraction. **Verdict: not blocking — 1 Medium worth addressing.**

**Findings entering verification:** 1 Medium. Confirmed `valid`: 1. Rejected: 0.

### Severity counts
- Critical: 0
- Major: 0
- Medium: 1 (posted inline)
- Minor/Nit: 3 (listed below, not posted inline)

- `internal/application/indexbuilder/builder.go:370` — nit: `_ = b.registerMetrics()` drops the error with no explanatory comment (convention #5 asks for a comment on intentional `_ =`). Non-fatal by design; a one-line comment would document that.
- `internal/pkg/tailworker/gauges.go:36,42,48` — nit: gauge descriptions were genericised (e.g. "Last audit sequence indexed" → "Last sequence indexed"). Metric *names* are preserved so dashboards are unaffected; only the human-readable description text changed. Acceptable given the shared helper, noted for awareness.
- `internal/pkg/tailworker/tailworker.go:63` — nit: `Stop()` panics (close of nil channel) if `Start()` was never called. Correctly documented, and both current consumers guard it (auditindexer via `cfg.Disabled`, builder always starts). No live bug; flagging only so a future third consumer (usage-builder) keeps the invariant.

internal/storage/readstore/store.go
paul-nicolas: Fixed in 7421cbc9a — builder.loop() now logs the error and falls back to a re-sync from 0 instead of swallowing it under `if err == nil`. Chose log-and-continue rather than aborting the loop (as the sibling `LastIndexedSequence` read does), since the applied-proposal cursor is a rebuildable projection and 0 is a safe rebuild. `just pre-commit` (0 lint issues) + `just test` green.

internal/pkg/tailworker/tailworker.go
```diff
@@ -0,0 +1,90 @@
...
+func (t *TailWorker) Start() {
+	t.w = worker.New()
+	t.w.RunCtx(t.loop)
+}
+
+// Stop signals the loop to stop and waits for it to finish. It must only be
+// called after Start (mirrors worker.Worker): calling it on a TailWorker that
+// was never started panics on a nil channel close. Consumers that start
+// conditionally must guard Stop with the same condition.
+func (t *TailWorker) Stop() {
```
gfyrag, line 63: Nit — `Stop()` panics with `close of nil channel` if called before `Start()` (because `t.w.stopCh` is nil in the zero value). Documented and mirrors `worker.Worker`, so consistent with the codebase — but as a shared skeleton this is a footgun for future consumers (auditindexer currently guards with `if i.cfg.Disabled { return }`; a third consumer must remember). One-liner alternative: `if t.w.stopCh != nil { t.w.Stop() }`. Not a blocker.

internal/pkg/tailworker/tailworker.go
```diff
@@ -0,0 +1,90 @@
...
+	// Name identifies the worker in log messages.
+	Name string
+	// Logger is used for boot and steady-state error logging.
+	Logger logging.Logger
+	// Ticker is the steady-state polling interval.
+	Ticker time.Duration
+	// Wake, when non-nil, triggers an extra Tick as soon as it receives.
+	// A nil channel is never selectable, so leaving it nil yields a
+	// pure-ticker loop with no special-casing.
+	Wake <-chan struct{}
```
gfyrag, line 31: Nit — a **closed** Wake channel would make the `select` fire on every iteration and tight-loop the Tick. Current callers only send (never close), so it's not a live issue, but worth adding to the field doc: "Do not close Wake; a closed channel spins the loop."

internal/application/indexbuilder/builder.go
```diff
@@ -365,9 +367,7 @@ func (b *Builder) SetNotifications(n *signal.Notifications) {
 
 // Start begins the background index-building loop and registers OTEL metrics.
 func (b *Builder) Start() {
-	if reg, err := b.registerMetrics(); err == nil {
-		b.metricsRegistration = reg
-	}
+	_ = b.registerMetrics()
```
gfyrag: Nit — `_ = b.registerMetrics()` silently discards the error. Old code did the same (`if err == nil` was silent too), so no regression, but with the new signature (`error` return + side-effect field writes) it reads a bit oddly. Consider `if err := b.registerMetrics(); err != nil { b.logger.Errorf("register index builder metrics: %v", err) }` — matches the boot-error logging style elsewhere.

internal/storage/readstore/cursor_test.go
```diff
@@ -0,0 +1,42 @@
...
+	c := Uint64Cursor{key: []byte("corrupt-cursor-key")}
+
+	// The read store opens Pebble with DisableWAL, so writes must use NoSync;
+	// pebble.Sync would return "WAL disabled". NoSync is the same option the
+	// store's own direct-DB writes (DeleteBackfillProgress) use.
+	require.NoError(t, s.db.Set(c.key, []byte{1, 2, 3}, pebble.NoSync))
+
+	_, err := c.Read(s.db)
+	require.Error(t, err, "a non-8-byte value must be a hard error, not silently 0")
+}
```
gfyrag, line 42: Optional — the `err != nil && !errors.Is(err, pebble.ErrNotFound)` branch in `Uint64Cursor.Read` has no test. Covered branches: missing key (→ 0) and corrupt length (→ error). A genuine pebble Get error is hard to trigger from a test though — probably not worth a fake-DB test just for this. Fine as-is IMO.

internal/pkg/tailworker/gauges.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+	"sync/atomic"
+
+	"go.opentelemetry.io/otel/metric"
+)
+
+// computeLag returns how far indexed trails sourceLast, clamped at 0 so a
+// momentarily-ahead indexed cursor never reports a negative lag.
+func computeLag(indexed, sourceLast uint64) int64 {
+	return max(int64(sourceLast)-int64(indexed), 0)
```
gfyrag, line 14: Nit — `int64(sourceLast) - int64(indexed)` overflows if either exceeds 2^63. Purely theoretical for a sequence counter, but the old inline computations had the same shape, so this is behaviour-preserving. Ignore.

gfyrag (approved):
LGTM — clean, well-scoped extraction. Metric names and on-disk keys preserved (verified: `audit_index.audit_last_sequence` = `{ns}.{source}_last_sequence`, `index.builder.pebble_last_sequence` idem; cursor keys are the same `ProgressKey()` / `AppliedProposalProgressKey()` / `AuditProgressKey()`). Only one caller of `ReadAppliedProposalProgress` exists, so the intentional error-on-corrupt change has a single, well-handled site — invariant #7 satisfied via loud log, safe re-sync from 0 because the projection is rebuildable (invariant #8). A few optional nits inline; nothing blocking.

internal/pkg/tailworker/tailworker.go
paul-nicolas: Fixed in 7040ea6e9 — added a `started` guard so `Stop()` is a no-op when the worker was never started, rather than documenting the panic. The suggested `t.w.stopCh != nil` can't compile from here (`stopCh` is unexported in package `worker`, so `tailworker` can't reach it); the bool on `TailWorker` is the cross-package equivalent. Consumers no longer need to mirror their start condition around Stop.

internal/pkg/tailworker/tailworker.go
paul-nicolas: Fixed in 7040ea6e9 — extended the `Wake` field doc: closing the channel makes it permanently selectable and spins the loop, calling Tick continuously.

internal/application/indexbuilder/builder.go
paul-nicolas: Fixed in 7040ea6e9 — `Start()` now logs the non-fatal registration error via `b.logger.Errorf("register index builder metrics: %v", err)` instead of the bare `_ =`.

internal/storage/readstore/cursor_test.go
paul-nicolas: Agreed — the only uncovered branch is a genuine (non-ErrNotFound) Pebble Get error, which needs a faked/failing handle the read store doesn't have. Not worth a fake-DB harness for a single wrap-and-return line. Leaving as-is; the two meaningful branches (missing→0, corrupt-length→error) are covered.

internal/pkg/tailworker/gauges.go
paul-nicolas: Leaving as-is — behaviour-preserving (the pre-refactor inline computations had the same shape), and the overflow is unreachable for a monotonic sequence counter well short of 2^63.

internal/pkg/tailworker/tailworker.go
```diff
@@ -0,0 +1,97 @@
...
+func (t *TailWorker) Start() {
+	t.w = worker.New()
+	t.started = true
+	t.w.RunCtx(t.loop)
+}
+
+// Stop signals the loop to stop and waits for it to finish. It is a no-op on a
+// TailWorker that was never started, so consumers that start conditionally do
+// not need to mirror the condition around Stop.
+func (t *TailWorker) Stop() {
```
NumaryBot, line 63: ⚪ **[nit] Stop() panics if called before Start() (nil channel)**

`Stop()` will panic with `close of nil channel` if called before `Start()` because `t.w.stopCh` is nil in the zero value. As a shared skeleton, this is a footgun for future consumers who may not guard against it the way current callers do.

**Suggestion:** Add a nil guard: `if t.w.stopCh != nil { t.w.Stop() }`, or document this invariant prominently on the type.

internal/pkg/tailworker/tailworker.go
```diff
@@ -0,0 +1,97 @@
...
+	// Logger is used for boot and steady-state error logging.
+	Logger logging.Logger
+	// Ticker is the steady-state polling interval.
+	Ticker time.Duration
+	// Wake, when non-nil, triggers an extra Tick as soon as it receives.
+	// A nil channel is never selectable, so leaving it nil yields a
+	// pure-ticker loop with no special-casing. Do not close Wake: a closed
+	// channel is always selectable and would spin the loop, calling Tick
+	// continuously.
+	Wake <-chan struct{}
```
NumaryBot, line 31: ⚪ **[nit] Closing the Wake channel would cause a tight spin loop**

A closed Wake channel would make the `select` fire on every iteration, busy-spinning the Tick. Current callers only send and never close, so it is not a live issue, but the field lacks documentation warning against this.

**Suggestion:** Add a field doc comment: "Do not close Wake; a closed channel will spin the loop."

internal/application/indexbuilder/builder.go
```diff
@@ -365,8 +367,8 @@ func (b *Builder) SetNotifications(n *signal.Notifications) {
 
 // Start begins the background index-building loop and registers OTEL metrics.
 func (b *Builder) Start() {
-	if reg, err := b.registerMetrics(); err == nil {
-		b.metricsRegistration = reg
+	if err := b.registerMetrics(); err != nil {
```
NumaryBot, line 370: ⚪ **[nit] registerMetrics() error silently discarded with blank identifier**

`_ = b.registerMetrics()` discards the error without logging. While the old code also silently ignored this, the new explicit `_` assignment reads oddly given the function now has an `error` return and performs side-effect field writes.

**Suggestion:** Replace with: `if err := b.registerMetrics(); err != nil { b.logger.Errorf("register index builder metrics: %v", err) }`

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1466#issuecomment-4864246863

PR #1470, opened 2026-07-02
feat(observability): trace PrepareSnapshot on both sides + guard with timeout

gfyrag:
## Summary

A follower stuck at `Fetching fresh checkpoint from leader` today (observed live on `eks-acme-dev-euw1-01/blockchains-2` while validating the EN-1431 series) gave **no further hint** until we SIGQUIT'd the process and read the goroutine stack. The client goroutine was blocked on the raw `PrepareSnapshot` gRPC call, and the server was silent on which step was slow: `waitForApplied` could not be timed, `CreateTemporaryCheckpoint` was un-logged (~4-5s on this cluster), `buildManifest` was un-logged, and the only success line \"Snapshot session created\" fires after everything.

This PR fills the observability gap.

## Client side (`snapshot_fetcher.go`)

- Log `Requesting snapshot session from leader` before the RPC (with `minAppliedIndex`).
- Log `Snapshot session prepared` after the response (sessionId, files, totalSize, duration).
- Log resume state when partial completed files are found in staging.
- Log retry attempts and per-attempt errors instead of silently looping.
- Wrap `PrepareSnapshot` in `prepareSnapshotWithHeartbeat`: 60s hard timeout (metadata call, not a data transfer) plus a 10s ticker that emits `Still waiting for PrepareSnapshot response from leader` so an operator watching live logs sees the hang instead of nothing.

## Server side (`server_snapshot.go`)

- Raise `PrepareSnapshot request received` from `Trace` to `Info` — the previous \"only if trace-enabled\" gating meant prod logs showed nothing.
- Emit `FSM caught up to minAppliedIndex` (with duration) if `waitForApplied` blocked longer than 100ms.
- Emit `Temporary checkpoint created for follower sync` with duration — this is the ~4-5s step on this cluster.
- Extend `Snapshot session created` with `manifestDuration` + `totalDuration`.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/application/ctrl/...` and `./internal/adapter/grpc/...` — both pass
- [ ] Deploy `pr-<num>-<sha>` on `eks-acme-dev-euw1-01/blockchains` and confirm the new log points fire during a follower rejoin

## Out of scope

- Metrics on the durations (histogram for waitForApplied / checkpoint / manifest). Logs cover the immediate diagnostic need; a follow-up can promote them if they turn out to be worth alerting on.
- A similar treatment on `FetchFile` (per-file streaming) — that path already has decent logs, and it's not the current pain point.

## Related

- The stuck follower is `blockchains-2` in `eks-acme-dev-euw1-01` after the EN-1431 series validation. Investigation of the root cause of the hang is separate — this PR just makes the next such hang diagnosable without a SIGQUIT.

### DIFF
--- a/internal/application/ctrl/snapshot_fetcher.go
+++ b/internal/application/ctrl/snapshot_fetcher.go
@@ -120,6 +120,13 @@ func (f *grpcSnapshotFetcher) FetchSnapshot(ctx context.Context, targetDir strin
 // blocks the client goroutine indefinitely; without the heartbeat, an
 // operator watching the follower's logs sees only "Fetching fresh checkpoint
 // from leader" and nothing else for the duration.
+//
+// The select waits on both `done` and `callCtx.Done()` so the wrapper
+// returns as soon as the deadline expires — even if the underlying gRPC
+// stub does not observe context cancellation (a defence-in-depth against
+// grpc-go bugs). When we exit via ctx cancellation the inner goroutine may
+// briefly outlive the call while it drains gRPC internals; that is a
+// bounded leak, and the caller is unblocked to retry a fresh session.
 func (f *grpcSnapshotFetcher) prepareSnapshotWithHeartbeat(ctx context.Context, minAppliedIndex uint64) (*snapshotpb.PrepareSnapshotResponse, error) {
 	logger := logging.FromContext(ctx)
 
@@ -148,6 +155,8 @@ func (f *grpcSnapshotFetcher) prepareSnapshotWithHeartbeat(ctx context.Context, 
 		select {
 		case r := <-done:
 			return r.resp, r.err
+		case <-callCtx.Done():
+			return nil, fmt.Errorf("PrepareSnapshot did not return within %s: %w", prepareSnapshotTimeout, callCtx.Err())
 		case <-ticker.C:
 			logger.WithFields(map[string]any{
 				"minAppliedIndex": minAppliedIndex,
--- a/internal/application/ctrl/snapshot_fetcher.go
+++ b/internal/application/ctrl/snapshot_fetcher.go
@@ -20,6 +20,17 @@ const (
 	sessionBackoff      = 1 * time.Second
 	sessionMaxBackoff   = 10 * time.Second
 	closeSessionTimeout = 5 * time.Second
+	// prepareSnapshotTimeout caps PrepareSnapshot — the metadata call that
+	// creates a Pebble checkpoint and builds a manifest on the leader.
+	// Without a cap the client goroutine can wait forever if the server
+	// hangs in the checkpoint or manifest step, and the outer sync task
+	// blocks along with it (node.Run's WaitForApplied → /readyz → pod stuck
+	// at 0/1 for the whole probe budget).
+	prepareSnapshotTimeout = 60 * time.Second
+	// prepareSnapshotHeartbeat drives the periodic "still waiting" log
+	// during PrepareSnapshot. It's short enough to be visible during a
+	// human operator's live look, long enough not to flood.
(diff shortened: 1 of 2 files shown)


internal/application/ctrl/snapshot_fetcher.go
```diff
@@ -88,27 +114,94 @@ func (f *grpcSnapshotFetcher) FetchSnapshot(ctx context.Context, targetDir strin
...
+		})
+		done <- result{resp: resp, err: err}
+	}()
+
+	started := time.Now()
+	ticker := time.NewTicker(prepareSnapshotHeartbeat)
+	defer ticker.Stop()
+
+	for {
+		select {
```
NumaryBot, line 155: 🟠 **[major] Return when PrepareSnapshot timeout fires**

When the underlying `SnapshotServiceClient` call does not return promptly after its context is canceled (the hang this wrapper is meant to guard against, or any fake/transport that ignores context cancellation), the 60s deadline only cancels `callCtx`; this loop never selects on `callCtx.Done()`, so `FetchSnapshot` keeps emitting heartbeats forever instead of returning `DeadlineExceeded` and retrying. Add a `callCtx.Done()` case here so the advertised hard timeout is actually enforced.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1470#issuecomment-4863717509

internal/application/ctrl/snapshot_fetcher.go
```diff
@@ -88,27 +114,103 @@ func (f *grpcSnapshotFetcher) FetchSnapshot(ctx context.Context, targetDir strin
...
```
NumaryBot, line 155: 🟠 **[major] Timeout does not unblock the streaming loop — hard deadline is not enforced**

When the 60-second local deadline fires it cancels `callCtx`, but the heartbeat/streaming loop never selects on `callCtx.Done()`. If the underlying `SnapshotServiceClient` RPC does not return promptly after context cancellation (the exact hang this guard was meant to fix), `FetchSnapshot` continues emitting heartbeats indefinitely instead of returning `DeadlineExceeded`. The timeout is therefore advisory rather than enforced.

**Suggestion:** Add a `case <-callCtx.Done(): return callCtx.Err()` branch inside the streaming loop so that the function actually returns when the deadline fires, regardless of whether the RPC unblocks.

internal/application/ctrl/snapshot_fetcher.go
```diff
@@ -88,27 +114,103 @@ func (f *grpcSnapshotFetcher) FetchSnapshot(ctx context.Context, targetDir strin
...
+		case r := <-done:
+			return r.resp, r.err
+		case <-callCtx.Done():
```
NumaryBot, line 158: 🟠 **[major] Timeout error bypasses the retry path — configured retries are not applied**

When the local deadline fires, the code returns a plain `fmt`-wrapped `context.DeadlineExceeded`. The retry classifier in `FetchSnapshot` only recognises errors that pass `status.FromError` (gRPC status errors) or expired-session conditions, so a timed-out `PrepareSnapshot` exits after the very first 60-second window instead of honoring `snapshot-retry-count` and starting a fresh session. The timeout guard therefore silently reduces effective retries to zero.

**Suggestion:** Return a gRPC `status.Error(codes.DeadlineExceeded, ...)` instead of a plain wrapped error, or extend the retry classifier to recognise wrapped `context.DeadlineExceeded` so that timeouts are treated as retryable and the configured retry policy is honoured.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1470#issuecomment-4863717509

paul-nicolas (approved):

PR #1472, opened 2026-07-02
temp: combined #1470 + #1471 for blockchains deploy

gfyrag:
Ephemeral PR — just to trigger a build-images run that produces an image containing both #1470 (snapshot-sync observability) and #1471 (early ready signal). Not intended to merge — those two land via their own PRs.

Deployed to eks-acme-dev-euw1-01/blockchains for live diagnosis of the PrepareSnapshot hang.

### DIFF
--- a/internal/application/ctrl/snapshot_fetcher.go
+++ b/internal/application/ctrl/snapshot_fetcher.go
@@ -20,6 +20,17 @@ const (
 	sessionBackoff      = 1 * time.Second
 	sessionMaxBackoff   = 10 * time.Second
 	closeSessionTimeout = 5 * time.Second
+	// prepareSnapshotTimeout caps PrepareSnapshot — the metadata call that
+	// creates a Pebble checkpoint and builds a manifest on the leader.
+	// Without a cap the client goroutine can wait forever if the server
+	// hangs in the checkpoint or manifest step, and the outer sync task
+	// blocks along with it (node.Run's WaitForApplied → /readyz → pod stuck
+	// at 0/1 for the whole probe budget).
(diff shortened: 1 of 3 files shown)


internal/application/ctrl/snapshot_fetcher.go
```diff
@@ -20,6 +20,17 @@ const (
 	sessionBackoff      = 1 * time.Second
 	sessionMaxBackoff   = 10 * time.Second
 	closeSessionTimeout = 5 * time.Second
+	// prepareSnapshotTimeout caps PrepareSnapshot — the metadata call that
+	// creates a Pebble checkpoint and builds a manifest on the leader.
+	// Without a cap the client goroutine can wait forever if the server
+	// hangs in the checkpoint or manifest step, and the outer sync task
+	// blocks along with it (node.Run's WaitForApplied → /readyz → pod stuck
+	// at 0/1 for the whole probe budget).
+	prepareSnapshotTimeout = 60 * time.Second
```
NumaryBot, line 29: 🔴 **[blocker] Avoid fixed 60s PrepareSnapshot deadline**

On large stores, `PrepareSnapshot` is not just a small metadata RPC: the server calls `buildManifest`, which walks the checkpoint and SHA256-hashes every file, so a 17GiB checkpoint can legitimately spend more than 60s before returning. In that case the follower gets `DeadlineExceeded`, retries, and starts another expensive manifest build rather than ever receiving a session, so snapshot sync can fail indefinitely for the exact large-checkpoint deployments this path is meant to recover.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1472#issuecomment-4864290407

PR #1473, opened 2026-07-02
chore(agents): add investigation skills and GitNexus agent guidance

flemzord:
## What

Adds shared Claude Code skills under `.claude/skills/` so every developer's agent benefits from the same tooling, plus the GitNexus section in `AGENTS.md`.

### New skills

- **`root-cause`** — root-cause investigation discipline: never patch the symptom; build a verified causal chain (evidence-backed "5 whys"), fix at the point where reality diverged from intent, sweep the codebase for the same class of bug, and add prevention (eliminate > enforce > detect > document). Anchored on this repo's invariants (e.g. CLAUDE.md invariant 7 on "should not happen" branches).
- **`diagnose`** — disciplined debugging loop for hard bugs and perf regressions: build a fast deterministic feedback loop first, then reproduce → hypothesise (ranked, falsifiable) → instrument → fix + regression test → cleanup. Escalates to `root-cause` when a fix smells like a symptom patch.
- **`gitnexus/*`** — code-intelligence skills generated by `gitnexus analyze` (exploring, impact analysis, debugging, refactoring, guide, CLI reference).

### AGENTS.md

Gains the GitNexus block (generated): impact analysis before editing symbols, `detect_changes()` before committing, and pointers to the skill files above.

## Why

These guardrails were previously only in personal skill directories; versioning them in the repo makes agent behavior consistent across the team.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -172,3 +172,48 @@ See [docs/technical/architecture/](docs/technical/architecture/) for detailed ar
 - **Formance go-libs** for service lifecycle, OTLP, HTTP server
 
 - I would like you to respect the concepts of DRY (Don't Repeat Yourself).
+
+<!-- gitnexus:start -->
+# GitNexus — Code Intelligence
+
+This project is indexed by GitNexus as **ledger** (41000 symbols, 142383 relationships, 300 execution flows). Use the GitNexus MCP tools to understand code, assess impact, and navigate safely.
+
+> Index stale? Run `node .gitnexus/run.cjs analyze` from the project root — it auto-selects an available runner. No `.gitnexus/run.cjs` yet? `npx gitnexus analyze` (npm 11 crash → `npm i -g gitnexus`; #1939).
+
+## Always Do
+
+- **MUST run impact analysis before editing any symbol.** Before modifying a function, class, or method, run `impact({target: "symbolName", direction: "upstream"})` and report the blast radius (direct callers, affected processes, risk level) to the user.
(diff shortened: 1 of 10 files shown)


paul-nicolas (approved):

AGENTS.md
```diff
@@ -172,3 +172,48 @@ See [docs/technical/architecture/](docs/technical/architecture/) for detailed ar
 - **Formance go-libs** for service lifecycle, OTLP, HTTP server
 
 - I would like you to respect the concepts of DRY (Don't Repeat Yourself).
+
+<!-- gitnexus:start -->
+# GitNexus — Code Intelligence
+
+This project is indexed by GitNexus as **ledger** (41000 symbols, 142383 relationships, 300 execution flows). Use the GitNexus MCP tools to understand code, assess impact, and navigate safely.
+
+> Index stale? Run `node .gitnexus/run.cjs analyze` from the project root — it auto-selects an available runner. No `.gitnexus/run.cjs` yet? `npx gitnexus analyze` (npm 11 crash → `npm i -g gitnexus`; #1939).
```
NumaryBot, line 181: 🟠 **[major] Ignore generated GitNexus index files**

This new guidance tells agents to run `npx gitnexus analyze` when `.gitnexus/run.cjs` is absent and even describes the runner as gitignored, but this repo does not currently ignore `.gitnexus/` (`git check-ignore .gitnexus/run.cjs` returns nothing). Anyone following the instruction will generate the local index/runner as untracked files in the worktree, making accidental commits or persistent dirty trees likely; please add `.gitnexus/` to `.gitignore` or avoid documenting it as a gitignored artifact.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1473#issuecomment-4864392756

gfyrag (approved):

PR #1476, opened 2026-07-02
fix(EN-1436): reset stale leader Progress on rejoin via JoinAsLearner

gfyrag:
## Summary

`JoinAsLearner` is only invoked when the caller has no `CLUSTER_JOINED` marker on its WAL — fresh WAL, or half-populated from a previous life. The old handler treated `ErrNodeAlreadyInCluster` as an idempotent success and returned. But if the leader's Progress for that nodeID still carries `Match > 0` from the previous life (common after a WAL PVC reprovision), the caller's local raftLog `lastIndex` is 0, and the leader's next MsgApp/heartbeat carries `Commit=Progress.Match`. etcd-raft's `commitTo` guard panics with \"tocommit out of range\" and the pod CrashLoopBackOffs indefinitely; no `RemoveNode` is ever issued to reset Progress.

Fix: on the leader-side handler, when `AddLearner` returns `ErrNodeAlreadyInCluster`, issue a `RemoveNode(force=true)` first and then retry `AddLearner`. Both ConfChanges are cheap. The result on the caller is a fresh `Progress{Match: 0, State: Probe}`, which triggers a proper MsgSnap catch-up — exactly the shape the caller's raftLog needs.

## Reproduction

Live on `eks-acme-dev-euw1-01/blockchains-1` today: after a WAL-PVC-only wipe, pod-1 hit `tocommit(31422) is out of range [lastIndex(0)]` on every boot for ~40 min. Resolved only by manually running:
\`\`\`
ledgerctl cluster remove-node 2 --force
\`\`\`
followed by another WAL-PVC delete + pod restart.

With this PR, the operator step would be unnecessary — the leader's `JoinAsLearner` handler auto-resets Progress when it detects the stale-membership shape.

## Backwards compatibility

The client's `tryAddLearner` `AlreadyExists` branch is intentionally left untouched. During a rolling upgrade, the client may still talk to an old leader that returns `AlreadyExists`; the marker-write fallback preserves idempotency semantics for that transitional case.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/adapter/grpc/...` — full grpc adapter suite passes
- [ ] Deploy `pr-<num>-<sha>` on a chaos-test cluster and verify a WAL-PVC-wipe rejoin recovers without manual intervention (blocked on next Antithesis run — the current cluster is healthy after today's manual cleanup)

## Ticket

[EN-1436](https://formance-team.atlassian.net/browse/EN-1436) — Operator: rejoining node with wiped WAL PVC panics on 'tocommit out of range' due to stale leader Progress

[EN-1436]: https://formance-team.atlassian.net/browse/EN-1436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/adapter/grpc/server_bootstrap.go
+++ b/internal/adapter/grpc/server_bootstrap.go
@@ -2,6 +2,8 @@ package grpc
 
 import (
 	"context"
+	"errors"
+	"fmt"
 
 	ggrpc "google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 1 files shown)


internal/adapter/grpc/server_bootstrap.go
```diff
@@ -160,11 +162,43 @@ func (impl *ClusterBootstrapServiceServerImpl) JoinAsLearner(ctx context.Context
...
 		//     (retried by the client)
+		//   - anything else → propagated as-is
+		if errors.Is(err, node.ErrNodeAlreadyInCluster) {
+			impl.logger.WithFields(map[string]any{
+				"nodeID":         req.GetNodeId(),
+				"raftAddress":    req.GetRaftAddress(),
+				"serviceAddress": req.GetServiceAddress(),
+			}).Infof("JoinAsLearner: node already in cluster with stale Progress; force-removing before re-adding")
+
+			if rmErr := impl.membership.RemoveNode(ctx, req.GetNodeId(), true); rmErr != nil {
```
NumaryBot, line 191: 🔴 **[blocker] Avoid local-only force removal during rejoin**

When `AlreadyInCluster` is hit on a multi-node cluster, this uses `force=true`, which routes to `Node.ForceRemoveNode` and directly applies the remove only on the current leader instead of replicating it through Raft. The retrying `AddLearner` is then committed from a configuration that followers never saw: followers still have the old member/progress (and if it was a voter, still treat it as a voter), so a later leadership change can bring the stale Progress back or leave ConfState divergent. This flow should use a consensus-backed removal when quorum exists, or otherwise ensure the removal/reset is applied consistently across the cluster.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1476#issuecomment-4865274864

gfyrag:
Superseded — redesigning as fail-fast (return FailedPrecondition + operator remediation instructions) per CLAUDE.md invariant #7 (never silently patch a should-not-happen branch). New PR incoming.

PR #1477, opened 2026-07-02
fix(EN-1440): propagate coverage-miss errors from buildPostCommitVolumes

paul-nicolas:
## Summary

`buildPostCommitVolumes` silently swallowed any non-`ErrNotFound` volume-read error (on the gated FSM scope, that is `*state.ErrCoverageMiss` — an admission-contract violation, invariants #6/#9, impossible by design under a correct preload). A bare `continue` dropped the `(account, asset)` pair **and** the error, returning a silently truncated post-commit volume map to the client whenever `ExpandVolumes` is set.

This violates **invariant #7** (never silently skip a "should-not-happen" branch) and hides an underlying coverage-gate/preload bug behind a wrong-but-deterministic response.

## Fix

- `buildPostCommitVolumes` now returns `(*commonpb.PostCommitVolumes, domain.Describable)`.
- Reads go through the canonical `readVolumeOrZero` helper (`ErrNotFound` → zero balance, any other error propagates) instead of a hand-rolled `Get` + `errors.Is` check — removing the DRY divergence from `applyPosting`.
- Propagated errors are wrapped in `ErrStorageOperation` (mirroring `applyPosting`), preserving the cause for `errors.As`/`errors.Is`.
- Both callers (`processor_transaction.go`, `processor_revert_transaction.go`) propagate the error, rejecting the order instead of committing a truncated map. The rejection is deterministic across nodes, so it does not desync the cluster.

## Behavior

- Happy path and declared-absent (`ErrNotFound`) keys are unchanged — byte-identical output ("0"/"0" for absent).
- A coverage miss now rejects the order loudly.

## Tests

New `internal/domain/processing/processor_volumes_test.go`:
- `TestBuildPostCommitVolumes_PropagatesReadError` — a non-`ErrNotFound` read error surfaces as `ErrStorageOperation` wrapping the cause (a generic sentinel stands in for `*state.ErrCoverageMiss`, which the processing package cannot import due to the `state → processing` import cycle).
- `TestBuildPostCommitVolumes_FoundAndAbsent` — present volume reported with real Input/Output; absent key reported as "0"/"0".

Full processing suite + whole-tree build + `just pre-commit` (golangci-lint 0 issues) all pass.

## Ticket

https://formance-team.atlassian.net/browse/EN-1440

### DIFF
--- a/internal/domain/processing/processor_revert_transaction.go
+++ b/internal/domain/processing/processor_revert_transaction.go
@@ -108,7 +108,11 @@ func processRevertTransaction(ledger string, order *raftcmdpb.RevertTransactionO
 	// Compute post-commit volumes if requested
 	var postCommitVolumes *commonpb.PostCommitVolumes
 	if order.GetExpandVolumes() {
-		postCommitVolumes = buildPostCommitVolumes(s, ledger, revertPostings)
+		var err domain.Describable
+		postCommitVolumes, err = buildPostCommitVolumes(s, ledger, revertPostings)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	return &commonpb.LedgerLogPayload{
--- a/internal/domain/processing/processor_transaction.go
+++ b/internal/domain/processing/processor_transaction.go
@@ -190,7 +190,11 @@ func processCreateTransaction(ledger string, order *raftcmdpb.CreateTransactionO
 	// Compute post-commit volumes if requested
 	var postCommitVolumes *commonpb.PostCommitVolumes
 	if order.GetExpandVolumes() {
-		postCommitVolumes = buildPostCommitVolumes(s, ledger, result.Postings)
+		var err domain.Describable
+		postCommitVolumes, err = buildPostCommitVolumes(s, ledger, result.Postings)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	// Get the current open chapter ID for the receipt
--- a/internal/domain/processing/processor_volumes.go
+++ b/internal/domain/processing/processor_volumes.go
@@ -1,8 +1,6 @@
 package processing
 
 import (
-	"errors"
-
 	"github.com/holiman/uint256"
 
 	"github.com/formancehq/ledger/v3/internal/domain"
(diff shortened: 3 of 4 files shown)


paul-nicolas (commented):

**Verdict: clean. No merge-blocking issues.** No Critical / Major / Medium findings survived verification, so no inline comments are posted.

The PR removes a genuine invariant-#7 violation: `buildPostCommitVolumes` previously swallowed any non-`ErrNotFound` volume-read error (on the gated FSM scope, `*state.ErrCoverageMiss`) with a bare `continue`, returning a silently truncated `PostCommitVolumes` map. The fix routes reads through the canonical `readVolumeOrZero`, wraps other errors in `ErrStorageOperation` (preserving the cause for `errors.As`/`errors.Is`), and both callers now reject the order. Confirmed: only 2 callers exist (`processCreateTransaction`, `processRevertTransaction`) and both are updated; the happy path is byte-identical; the `errors` import removal is clean; the two new tests are sound (`kindStub.Get` returns `ErrNotFound` for unregistered keys, so `FoundAndAbsent` genuinely exercises the zero-synthesis path). Rejection is deterministic across nodes (coverage_bits + cache state are replicated), so it cannot desync the cluster.

### Verification of the strongest finding (Codex)
Codex flagged a **Major** "can panic on a nil reader" at `processor_volumes.go`. We adversarially verified reachability: all three producers — `stdPostingProducer`/`applyPosting`, `numscriptPostingProducer` (`processor_transaction_numscript.go:90-139`, which reads via `readVolumeOrZero`, guards nil Input/Output, and `Put`s), and `processRevertTransaction` (`applyPosting` at `:58`) — materialize each source+destination volume with non-nil Input/Output **before** `buildPostCommitVolumes` iterates the same postings. Any `*state.ErrCoverageMiss` or nil-side volume would already have rejected the order upstream. The panic is **unreachable by construction**, so the finding is downgraded to Minor (comment-accuracy / defense-in-depth), not a live bug.

### Severity counts
- Critical: 0
- Major: 0 (Codex's Major panic finding refuted on reachability → Minor)
- Medium: 0 (Codex's Medium test finding → Minor: caller change is trivial error propagation; helper-level tests are solid)
- Inline comments posted: 0

- `internal/domain/processing/processor_volumes.go:52` (Minor) — The new doc comment states it "mirrors applyPosting", but it omits `applyPosting`'s (and the numscript producer's) `GetInput()/GetOutput() == nil` guard. The panic is unreachable today (see above), so this is defense-in-depth: either add the same nil-side guard returning a describable error, or soften the comment to "mirrors applyPosting's read source" so the claim matches the code.
- `internal/domain/processing/processor_volumes.go:52` (Nit) — `ErrCoverageMiss` is described as "impossible by design"; `assert.Unreachable(...)` would give antithesis a stronger signal than the soft `ErrStorageOperation` return. Current choice is consistent with `applyPosting`, so acceptable as-is.
- `internal/domain/processing/processor_volumes_test.go` (Minor) — The new tests pin the helper only; neither `ExpandVolumes` caller path (`processCreateTransaction` / `processRevertTransaction`) has an end-to-end regression test asserting the whole order is rejected on a failing post-commit read. The propagation is a trivial 2-line `if err != nil { return nil, err }` and the helper is well-tested, so this is optional hardening, not merge-blocking.
- `internal/domain/processing/processor_volumes_test.go:41` (Nit) — Test uses a generic `errors.New` sentinel rather than the real `*state.ErrCoverageMiss`. Acknowledged in-comment (import cycle: `state` imports `processing`); acceptable since all non-`ErrNotFound` errors are treated identically.

### Documentation drift
None required — internal FSM behavior change surfacing as an existing internal-error reason (`ErrReasonStorageOperation`); no OpenAPI / CLI / api-comparison changes, and no new persisted projection (invariant #8 not implicated).

Azorlogh (approved):

gfyrag (approved):

PR #1478, opened 2026-07-02
fix(EN-1436): fail-fast when leader has stale Progress for a rejoining node

gfyrag:
## Summary

Replace the silent `AlreadyExists → idempotent success` semantics on `JoinAsLearner` with a fail-fast that surfaces the exact remediation command. Follows CLAUDE.md invariant #7 (never silently skip a should-not-happen branch) — the previous behavior was papering over a raft-invariant violation and leaving operators with an unrecoverable CrashLoopBackOff.

## Background

`JoinAsLearner` reaches the leader only when the caller has no `CLUSTER_JOINED` marker on its WAL (see `bootstrap/module.go tryAddLearner`). If the leader's Progress already carries this nodeID, one of two shapes hit:

1. Legitimate previous member whose WAL was reprovisioned (chaos, operator, GitOps drift). Local `raftLog.lastIndex` is 0.
2. Half-populated WAL from a crash between "initial snapshot write" and "CLUSTER_JOINED marker write". Local raftLog may have some entries but the marker didn't persist.

The old code returned `AlreadyExists` (mapped from `ErrNodeAlreadyInCluster`) and the client treated it as idempotent success. That works for shape 2 with a lucky `lastIndex`, and silently breaks shape 1: as soon as raft starts on the caller, the leader's next MsgApp/heartbeat carries `Commit=Progress.Match`, etcd-raft's `commitTo` guard fires (`tocommit out of range`), and the pod CrashLoopBackOffs indefinitely with no clue for the operator.

Reproduced live today on `eks-acme-dev-euw1-01/blockchains-1`: ~40 minutes of tocommit crashes, resolved only by manually running `ledgerctl cluster remove-node 2 --force`.

## Why fail-fast instead of auto-recovery

An earlier draft (closed #1476) silently issued a `RemoveNode(force=true)` + retry `AddLearner` inside `JoinAsLearner`. That would have hidden real operational signals (why did the WAL disappear? did GitOps drift the PVC? did the operator's raft_scaledown misfire?) and violated CLAUDE.md invariant #7. Force-removing a raft member is a destructive operation; auto-firing it on any `AlreadyExists` conflates multiple failure modes and prevents an operator from investigating the trigger.

We cannot distinguish shape 1 from shape 2 from server-side data alone — `Progress.Match` doesn't say whether the caller's log covers it. Surfacing the ambiguity to the operator is the correct move.

## Behavior

- **Server** returns `codes.FailedPrecondition` with a message that names the offending nodeID and the exact `ledgerctl` remediation command.
- **Client** (`tryAddLearner`) catches `FailedPrecondition`, logs the server message at ERROR level with the operator-facing key `STALE MEMBERSHIP on leader — operator action required to unblock rejoin`, and returns a fatal error. Pod fails fast at boot with a single clear log line instead of crash-looping on `tocommit`.

## Rolling-upgrade compatibility

The pre-existing `AlreadyExists` branch in `tryAddLearner` is kept so a new client can still talk to a pre-fix leader during rollout. Flagged in a comment for eventual removal — the invariant it patches is exactly what this PR is fixing.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/adapter/grpc/...` and `./internal/bootstrap/...` — both pass
- [ ] Deploy on a chaos-test cluster, wipe a follower's WAL PVC, verify the pod fails fast with the STALE MEMBERSHIP log rather than crash-looping

## Ticket

[EN-1436](https://formance-team.atlassian.net/browse/EN-1436)

[EN-1436]: https://formance-team.atlassian.net/browse/EN-1436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/adapter/grpc/server_bootstrap.go
+++ b/internal/adapter/grpc/server_bootstrap.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"encoding/hex"
 	"errors"
+	"fmt"
 
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
 	ggrpc "google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
@@ -223,15 +225,39 @@ func (impl *ClusterBootstrapServiceServerImpl) JoinAsLearner(ctx context.Context
 		// behalf — which would mask real operational events (why did the
 		// WAL disappear? did GitOps drift the PVC?) — surface it as a
 		// FailedPrecondition with the exact remediation command in the
-		// message. Client-side, tryAddLearner treats FailedPrecondition as
+		// message. Client-side, tryAddLearner treats this specific reason as
 		// fatal-with-clear-message so the pod fails fast with an operator-
 		// actionable log line instead of crash-looping on tocommit.
-		if errors.Is(err, node.ErrNodeAlreadyInCluster) {
-			return nil, status.Errorf(codes.FailedPrecondition,
+		//
+		// EN-1436: the guard keys on ErrNodeStaleProgress, which AddLearner
+		// returns whenever the leader's Progress.Match for this node is
+		// non-zero — covering BOTH the identical-identity rejoin and the
+		// fresh-identity (WAL-wiped) rejoin, where a new instance_id would
+		// otherwise route through a benign ConfChangeUpdateNode refresh and
+		// bypass the fail-fast entirely. A distinguishing ErrorInfo detail
+		// (reason STALE_RAFT_PROGRESS) is attached so the client can tell
+		// this apart from the removed-member blacklist rejection above,
+		// which is also FailedPrecondition but needs `forget-removed`, not
+		// `remove-node --force`.
+		if errors.Is(err, node.ErrNodeStaleProgress) {
+			st := status.New(codes.FailedPrecondition, fmt.Sprintf(
 				"node %d is already in the leader's raft Progress but the caller has no CLUSTER_JOINED marker — "+
 					"its local WAL cannot satisfy the leader's known match index. "+
 					"Reset membership first: `ledgerctl cluster remove-node %d --force` on the leader, then restart this pod.",
-				req.GetNodeId(), req.GetNodeId())
+				req.GetNodeId(), req.GetNodeId()))
+
+			detailed, detailErr := st.WithDetails(&errdetails.ErrorInfo{
+				Reason: node.StaleRaftProgressReason,
+				Domain: "ledger",
+			})
+			if detailErr != nil {
+				// Attaching the detail failed (marshal error): fall back to
+				// the bare status so the caller still fails fast, just
+				// without the machine-readable reason.
+				return nil, st.Err()
+			}
+
+			return nil, detailed.Err()
 		}
 
 		// Notably:
--- a/internal/adapter/grpc/server_bootstrap.go
+++ b/internal/adapter/grpc/server_bootstrap.go
@@ -3,6 +3,7 @@ package grpc
 import (
 	"context"
 	"encoding/hex"
+	"errors"
 
 	ggrpc "google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -196,9 +197,44 @@ func (impl *ClusterBootstrapServiceServerImpl) JoinAsLearner(ctx context.Context
 	}
 
 	if err := impl.membership.AddLearner(ctx, req.GetNodeId(), req.GetRaftAddress(), req.GetServiceAddress(), req.GetInstanceId()); err != nil {
+		// EN-1436: a JoinAsLearner call reaches the leader only when the caller
+		// has no CLUSTER_JOINED marker on its WAL (see bootstrap/module.go
+		// tryAddLearner). If the leader's Progress already carries this
+		// nodeID, one of two things happened:
+		//
+		//   1. Legitimate previous member whose WAL was reprovisioned (chaos,
+		//      operator, GitOps drift). Local raftLog.lastIndex is 0.
+		//   2. Half-populated WAL from a crash between "initial snapshot
+		//      write" and "CLUSTER_JOINED marker write". Local raftLog may
+		//      have some entries but the marker did not persist.
+		//
+		// The old code returned AlreadyExists (mapped from
+		// ErrNodeAlreadyInCluster) and the client treated it as idempotent
+		// success. That works for shape 2 with a lucky lastIndex, and
+		// silently breaks shape 1: as soon as raft starts on the caller,
+		// the leader's next MsgApp/heartbeat carries Commit=Progress.Match,
+		// etcd-raft's commitTo guard fires ("tocommit out of range"), and
+		// the pod CrashLoopBackOffs indefinitely with no clue for the
+		// operator. We cannot distinguish the two shapes from server-side
+		// data (Progress.Match alone doesn't say whether the caller's log
+		// covers it).
+		//
+		// Rather than silently patch this by force-removing on the caller's
+		// behalf — which would mask real operational events (why did the
+		// WAL disappear? did GitOps drift the PVC?) — surface it as a
+		// FailedPrecondition with the exact remediation command in the
+		// message. Client-side, tryAddLearner treats FailedPrecondition as
+		// fatal-with-clear-message so the pod fails fast with an operator-
+		// actionable log line instead of crash-looping on tocommit.
+		if errors.Is(err, node.ErrNodeAlreadyInCluster) {
+			return nil, status.Errorf(codes.FailedPrecondition,
+				"node %d is already in the leader's raft Progress but the caller has no CLUSTER_JOINED marker — "+
+					"its local WAL cannot satisfy the leader's known match index. "+
+					"Reset membership first: `ledgerctl cluster remove-node %d --force` on the leader, then restart this pod.",
+				req.GetNodeId(), req.GetNodeId())
+		}
+
 		// Notably:
-		//   - ErrNodeAlreadyInCluster → AlreadyExists (idempotent join,
-		//     handled as success in tryAddLearner)
 		//   - ErrNotLeader / ErrProposalDropped / ErrNoLeader → Unavailable
 		//     (retried by the client)
 		return nil, convertToGRPCError(err, impl.logger)
--- a/internal/bootstrap/module.go
+++ b/internal/bootstrap/module.go
@@ -14,6 +14,7 @@ import (
 	"go.opentelemetry.io/otel/metric"
 	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
 	"go.uber.org/fx"
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
 	ggrpc "google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/health"
@@ -1493,15 +1494,23 @@ func tryAddLearner(ctx context.Context, cfg Config, tlsCfg TLSConfig, logger log
 			}
 
 			st, ok := status.FromError(err)
-			// EN-1436: the leader detected this nodeID in its raft Progress but
-			// we (the caller) have no CLUSTER_JOINED marker on our WAL. This
-			// means the leader's Match for us points at state we don't have,
-			// and the next MsgApp/heartbeat would trigger etcd-raft's
-			// "tocommit out of range" panic. Fail fast with the operator-
-			// actionable server message rather than retry, mark, or silently
-			// crash-loop. The server message already contains the exact
-			// remediation command; surface it verbatim as the fatal reason.
-			if ok && st.Code() == codes.FailedPrecondition {
+			// EN-1436: the leader detected this nodeID in its raft Progress
+			// with a non-zero Match but we (the caller) have no CLUSTER_JOINED
+			// marker on our WAL. The leader's Match for us points at state we
+			// don't have, and the next MsgApp/heartbeat would trigger
+			// etcd-raft's "tocommit out of range" panic. Fail fast with the
+			// operator-actionable server message rather than retry, mark, or
+			// silently crash-loop. The server message already contains the
+			// exact remediation command; surface it verbatim as the fatal
+			// reason.
+			//
+			// Match on the STALE_RAFT_PROGRESS ErrorInfo reason, not on the
+			// bare FailedPrecondition code: the removed-member blacklist
+			// rejection (EN-1045) is also FailedPrecondition but needs
+			// `forget-removed`, not `remove-node --force`. Conflating the two
+			// would print misleading remediation. A blacklist rejection falls
+			// through to the generic fatal handler below with its own message.
+			if ok && st.Code() == codes.FailedPrecondition && isStaleRaftProgress(st) {
 				logger.WithFields(map[string]any{
 					"peer":   peer.ID,
 					"nodeID": cfg.RaftConfig.NodeID,
@@ -1571,6 +1580,22 @@ func tryAddLearner(ctx context.Context, cfg Config, tlsCfg TLSConfig, logger log
 
 		backoff = min(backoff*2, maxBackoff)
 	}
+}
+
+// isStaleRaftProgress reports whether a FailedPrecondition status carries the
+// EN-1436 STALE_RAFT_PROGRESS ErrorInfo reason, distinguishing a stale-raft-
+// progress join rejection (remediation: `remove-node --force`) from the
+// removed-member blacklist rejection (remediation: `forget-removed`), which is
+// also FailedPrecondition. Falls back to false when no matching detail is
+// present so unrelated FailedPrecondition responses keep their generic handling.
+func isStaleRaftProgress(st *status.Status) bool {
+	for _, d := range st.Details() {
+		if info, ok := d.(*errdetails.ErrorInfo); ok && info.GetReason() == node.StaleRaftProgressReason {
+			return true
+		}
+	}
+
+	return false
 }
 
 // proposeClusterConfigIfNeeded reads the persisted cluster state from Pebble
--- a/internal/bootstrap/module.go
+++ b/internal/bootstrap/module.go
@@ -1493,10 +1493,33 @@ func tryAddLearner(ctx context.Context, cfg Config, tlsCfg TLSConfig, logger log
 			}
 
 			st, ok := status.FromError(err)
+			// EN-1436: the leader detected this nodeID in its raft Progress but
+			// we (the caller) have no CLUSTER_JOINED marker on our WAL. This
+			// means the leader's Match for us points at state we don't have,
+			// and the next MsgApp/heartbeat would trigger etcd-raft's
+			// "tocommit out of range" panic. Fail fast with the operator-
+			// actionable server message rather than retry, mark, or silently
+			// crash-loop. The server message already contains the exact
+			// remediation command; surface it verbatim as the fatal reason.
+			if ok && st.Code() == codes.FailedPrecondition {
+				logger.WithFields(map[string]any{
+					"peer":   peer.ID,
+					"nodeID": cfg.RaftConfig.NodeID,
+					"reason": st.Message(),
+				}).Errorf("STALE MEMBERSHIP on leader — operator action required to unblock rejoin")
+
+				return fmt.Errorf("stale raft membership on the leader: %s", st.Message())
+			}
+
+			// AlreadyExists path kept for rolling-upgrade compatibility with
+			// leaders that predate EN-1436's server-side fail-fast. Once every
+			// live cluster is past that version, this branch can be dropped —
+			// the invariant it silently patches (idempotent join treating
+			// stale Progress as success) is exactly the bug EN-1436 fixes.
 			if ok && st.Code() == codes.AlreadyExists {
 				logger.WithFields(map[string]any{
 					"nodeID": cfg.RaftConfig.NodeID,
-				}).Infof("Already a cluster member, skipping learner registration")
+				}).Infof("Already a cluster member, skipping learner registration (deprecated pre-EN-1436 semantics)")
 
 				if markErr := wal.MarkClusterJoined(cfg.RaftConfig.WalDir); markErr != nil {
 					return fmt.Errorf("marking cluster joined after AlreadyExists: %w", markErr)
--- a/internal/bootstrap/module.go
+++ b/internal/bootstrap/module.go
@@ -879,6 +879,26 @@ func Module() fx.Option {
 					},
 				})
 			},
+			// Join preflight (EN-1436): register this node as a learner on the
+			// leader and fail fast on stale raft Progress. Registered FIRST
+			// among the Raft startup hooks so its OnStart runs (and, on stale
+			// Progress, aborts boot) BEFORE any hook that lets inbound Raft
+			// traffic reach rawNode.Step — the Raft gRPC server start +
+			// membership.Start() and node.Run() below. fx runs OnStart hooks in
+			// Append order; keeping this closure ahead of those enforces the
+			// ordering the fail-fast depends on. See joinPreflightHook for the
+			// inbound-vs-outbound / no-deadlock rationale.
+			func(
+				lc fx.Lifecycle,
+				cfg Config,
+				logger logging.Logger,
+			) {
+				if !shouldRunJoinPreflight(cfg, logger) {
+					return
+				}
+
+				lc.Append(joinPreflightHook(cfg, logger))
+			},
 			func(
 				lc fx.Lifecycle,
 				t *node.DefaultTransport,
@@ -1198,50 +1218,10 @@ func Module() fx.Option {
 					},
 				})
 			},
-			// Join mode: auto-register as learner on the leader after raft starts.
-			// Any peer will forward the request to the current leader automatically.
-			// The AddLearner call is idempotent: if the node is already a cluster
-			// member (e.g. after a crash-restart where the initial WAL was written
-			// but the AddLearner RPC never reached the leader), the server returns
-			// AlreadyExists which we treat as success.
-			//
-			// Skip entirely on restart: the CLUSTER_JOINED marker is written
-			// after the initial join succeeds (see tryAddLearner). On any
-			// subsequent boot — even one where --join is still in the CLI
-			// (e.g. E2E test framework that reuses Instruments across restarts)
-			// — re-running tryAddLearner would propose a redundant AddLearner
-			// for a node that is already a voter, racing leadership state and
-			// stalling the test's "should become follower" Eventually. The
-			// operator's StatefulSet entrypoint already gates --join on the
-			// same marker; this check makes the safety hold regardless of how
-			// the binary is invoked.
-			func(
-				lc fx.Lifecycle,
-				cfg Config,
-				logger logging.Logger,
-			) {
-				if cfg.RaftConfig.Bootstrap || len(cfg.RaftConfig.Peers) == 0 {
-					return
-				}
-
-				if wal.IsClusterJoined(cfg.RaftConfig.WalDir) {
-					logger.Infof("CLUSTER_JOINED marker present, skipping learner registration")
-
-					return
-				}
-
-				lc.Append(fx.Hook{
-					OnStart: func(ctx context.Context) error {
-						logger.WithFields(map[string]any{
-							"nodeID":         cfg.RaftConfig.NodeID,
-							"raftAddress":    cfg.RaftConfig.AdvertiseAddr,
-							"serviceAddress": cfg.ServiceAdvertiseAddr(),
-						}).Infof("Join mode: requesting peer to add this node as learner")
-
-						return tryAddLearner(ctx, cfg, cfg.TLSConfig, logger)
-					},
-				})
-			},
+			// Join mode preflight is registered EARLIER (before the Raft
+			// transport/server/node startup hooks) so it runs before any inbound
+			// Raft traffic can be stepped — see the joinPreflightHook closure
+			// above and its doc comment for the EN-1436 ordering rationale.
 			func(lc fx.Lifecycle, cfg Config, handler http.Handler) {
 				lc.Append(transportfx.FXHook(httpserver.NewHook(handler,
 					httpserver.WithAddress(fmt.Sprintf(":%d", cfg.HTTPPort)),
@@ -1415,6 +1395,68 @@ func (e *JoinAuthError) Error() string {
 		"cluster join rejected by %s: inter-node authentication failed (%s); %s",
 		target, e.Detail, hint,
 	)
+}
+
+// shouldRunJoinPreflight reports whether the join preflight (tryAddLearner)
+// must run for this boot. It runs only for a node that is joining an existing
+// cluster and has not yet recorded a successful join:
+//
+//   - Bootstrap nodes and nodes with no configured peers never join — they
+//     ARE the seed of the cluster.
+//   - A node whose WAL already carries the CLUSTER_JOINED marker has joined on
+//     a previous boot; re-running tryAddLearner would propose a redundant
+//     AddLearner for a node that is already a voter, racing leadership state.
+//     The operator's StatefulSet entrypoint already gates --join on the same
+//     marker; this check makes the safety hold regardless of how the binary is
+//     invoked (e.g. an E2E framework that reuses instruments across restarts).
+func shouldRunJoinPreflight(cfg Config, logger logging.Logger) bool {
+	if cfg.RaftConfig.Bootstrap || len(cfg.RaftConfig.Peers) == 0 {
+		return false
+	}
+
+	if wal.IsClusterJoined(cfg.RaftConfig.WalDir) {
+		logger.Infof("CLUSTER_JOINED marker present, skipping learner registration")
+
+		return false
+	}
+
+	return true
+}
+
+// joinPreflightHook builds the OnStart hook that registers this node as a
+// learner on the leader (and fails fast on stale raft Progress, EN-1436).
+//
+// EN-1436 ordering (flemzord review on #1478): this hook MUST run BEFORE the
+// node starts accepting/stepping inbound Raft traffic — i.e. before the Raft
+// gRPC transport server begins serving, before membership.Start() wires the
+// leader connection, and before node.Run() drains the recv queues into
+// rawNode.Step. Otherwise, once the leader can reach this pod, it may deliver a
+// stale MsgApp/heartbeat whose Commit points past our (empty) log and trip
+// etcd-raft's "tocommit out of range" panic BEFORE this preflight returns the
+// STALE_RAFT_PROGRESS fail-fast — losing exactly the race the fail-fast exists
+// to prevent. Registering this hook earlier in the fx lifecycle guarantees its
+// OnStart runs (and, on stale Progress, aborts boot) before any inbound Raft
+// message can be stepped.
+//
+// This does NOT deadlock: the preflight reaches the leader over an OUTBOUND,
+// short-lived gRPC client it dials itself (see tryAddLearner) — it depends on
+// neither the local Raft gRPC server (inbound serving) nor the transport
+// connection pool that membership.Start() wires. Only INBOUND stepping is
+// gated behind the preflight; OUTBOUND dialing is always available because it
+// is self-contained. Gating inbound-serving on an outbound-only preflight is
+// therefore safe.
+func joinPreflightHook(cfg Config, logger logging.Logger) fx.Hook {
+	return fx.Hook{
+		OnStart: func(ctx context.Context) error {
+			logger.WithFields(map[string]any{
+				"nodeID":         cfg.RaftConfig.NodeID,
+				"raftAddress":    cfg.RaftConfig.AdvertiseAddr,
+				"serviceAddress": cfg.ServiceAdvertiseAddr(),
+			}).Infof("Join mode: requesting peer to add this node as learner (preflight, before Raft traffic)")
+
+			return tryAddLearner(ctx, cfg, cfg.TLSConfig, logger)
+		},
+	}
 }
 
 // tryAddLearner attempts to register this node as a learner on an existing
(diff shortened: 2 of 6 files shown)


internal/adapter/grpc/server_bootstrap.go
```diff
@@ -196,9 +197,44 @@ func (impl *ClusterBootstrapServiceServerImpl) JoinAsLearner(ctx context.Context
...
+		// covers it).
+		//
+		// Rather than silently patch this by force-removing on the caller's
+		// behalf — which would mask real operational events (why did the
+		// WAL disappear? did GitOps drift the PVC?) — surface it as a
+		// FailedPrecondition with the exact remediation command in the
+		// message. Client-side, tryAddLearner treats FailedPrecondition as
+		// fatal-with-clear-message so the pod fails fast with an operator-
+		// actionable log line instead of crash-looping on tocommit.
+		if errors.Is(err, node.ErrNodeAlreadyInCluster) {
```
NumaryBot: 🔴 **[blocker] Fail-fast bypassed when rejoining node has a fresh instance ID**

When a rejoining pod has its WAL/PVC wiped, `wal.EnsureInstanceID` generates a new instance ID while the leader may still hold a stale `Progress[nodeID]` entry and a peer row with the old identity. In that scenario `Node.AddLearner` detects the identity mismatch, issues `ConfChangeUpdateNode`, and returns nil instead of `ErrNodeAlreadyInCluster`. The new fail-fast guard is therefore bypassed entirely: `JoinAsLearner` reports success, the caller marks `CLUSTER_JOINED` with an empty local log while the leader's stale Match index is preserved, and the `tocommit out of range` crash loop this PR is intended to eliminate can still occur.

**Suggestion:** Before calling `AddLearner`, explicitly check whether the leader already has a Progress entry for the incoming node ID with a Match index greater than zero (i.e. stale Progress) and return `FailedPrecondition` in that case, regardless of whether the instance ID matches. This ensures the fail-fast fires for both the identical-identity and the fresh-identity rejoin paths.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1478#issuecomment-4865338697

gfyrag (commented):
Reviewed the rebase/merge commit (`2236d699`). No findings.

Scope checked: the conflict resolution keeps `JoinAsLearner` fail-fast for stale raft Progress on the bootstrap surface, while preserving the existing admin `AddLearner` conversion path. The client-side bootstrap path now treats the server's `FailedPrecondition` as fatal with the operator-actionable message instead of retrying into a later raft panic.

Validation: `GOCACHE=/tmp/go-build-pr1478 go test ./internal/adapter/grpc/... ./internal/bootstrap/...` passed.

internal/adapter/grpc/server_bootstrap.go
NumaryBot: 🔴 **[blocker] Fail-fast bypassed when rejoining node has a fresh instance ID after WAL/PVC wipe**

When a rejoining pod has its WAL/PVC wiped, `wal.EnsureInstanceID` generates a new instance ID. In that scenario `Node.AddLearner` detects the identity mismatch, issues `ConfChangeUpdateNode`, and returns nil instead of `ErrNodeAlreadyInCluster`. The new fail-fast guard is therefore bypassed entirely: `JoinAsLearner` reports success while the leader still holds a stale `Match` index, leaving the `tocommit out of range` crash loop — the root problem this PR targets — fully reproducible on the wiped-node rejoin path.

**Suggestion:** Before calling `AddLearner`, explicitly check whether the leader already has a Progress entry for the incoming node ID with a Match index greater than zero (stale Progress) and return `FailedPrecondition` in that case, regardless of whether the instance ID matches. This ensures the fail-fast fires for both the identical-identity and fresh-identity rejoin paths.

internal/bootstrap/module.go
```diff
@@ -1449,10 +1449,33 @@ func tryAddLearner(ctx context.Context, cfg Config, tlsCfg TLSConfig, logger log
...
 			st, ok := status.FromError(err)
+			// EN-1436: the leader detected this nodeID in its raft Progress but
+			// we (the caller) have no CLUSTER_JOINED marker on our WAL. This
+			// means the leader's Match for us points at state we don't have,
+			// and the next MsgApp/heartbeat would trigger etcd-raft's
+			// "tocommit out of range" panic. Fail fast with the operator-
+			// actionable server message rather than retry, mark, or silently
+			// crash-loop. The server message already contains the exact
+			// remediation command; surface it verbatim as the fatal reason.
+			if ok && st.Code() == codes.FailedPrecondition {
```
NumaryBot: 🟡 **[minor] Stale-membership error message emitted for all FailedPrecondition join rejections**

The server also returns `codes.FailedPrecondition` when a join is rejected by the removed-member registry. With the new catch-all on `FailedPrecondition`, that path will now be logged as `STALE MEMBERSHIP` and surface a `stale raft membership` error to the operator, even though the correct remediation is `forget-removed`, not `remove-node --force`. This produces misleading operator guidance.

**Suggestion:** Only emit the stale-membership operator message for the specific stale-progress `FailedPrecondition` response (e.g. by checking a distinguishing error detail or status message), and leave all other `FailedPrecondition` cases with their original handling/reason.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1478#issuecomment-4865338697

internal/adapter/grpc/server_bootstrap.go
NumaryBot: 🔴 **[blocker] Fail-fast bypassed when rejoining node has a fresh instance ID after WAL/PVC wipe**

When a rejoining pod has its WAL/PVC wiped, `wal.EnsureInstanceID` generates a new instance ID. `Node.AddLearner` detects the identity mismatch and issues `ConfChangeUpdateNode`, returning nil instead of `ErrNodeAlreadyInCluster`. The new fail-fast guard therefore never fires: `JoinAsLearner` reports success while the leader still holds a stale `Progress.Match` index, leaving the `tocommit out of range` crash loop — the root problem this PR targets — fully reproducible on the wiped-node rejoin path.

**Suggestion:** Before calling `AddLearner`, explicitly check whether the leader already holds a Progress entry for the incoming node ID with a Match index greater than zero (i.e. stale Progress), and return `FailedPrecondition` in that case regardless of whether the stored instance ID matches the incoming one. This ensures the fail-fast fires for both the identical-identity and fresh-identity (wiped WAL) rejoin paths.

internal/bootstrap/module.go
NumaryBot: 🟡 **[minor] Stale-membership error message emitted for all FailedPrecondition join rejections**

The server also returns `codes.FailedPrecondition` when a join is rejected by the removed-member registry (the `IsRemoved`/`ErrNodeRemoved` path). With the new catch-all on `FailedPrecondition`, that path is now logged as `STALE MEMBERSHIP` and surfaces a `stale raft membership` error to the operator, even though the correct remediation is `forget-removed`, not `remove-node --force`. This produces misleading operator guidance for an unrelated failure class.

**Suggestion:** Scope the stale-membership operator message to the specific stale-Progress `FailedPrecondition` response only — for example by checking a distinguishing error detail, status message field, or a dedicated gRPC error code — and leave all other `FailedPrecondition` cases with their original handling and error reason.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1478#issuecomment-4865338697

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1478#issuecomment-4865338697

internal/adapter/grpc/server_bootstrap.go
```diff
@@ -196,9 +199,68 @@ func (impl *ClusterBootstrapServiceServerImpl) JoinAsLearner(ctx context.Context
...
+		// operator. We cannot distinguish the two shapes from server-side
+		// data (Progress.Match alone doesn't say whether the caller's log
+		// covers it).
+		//
+		// Rather than silently patch this by force-removing on the caller's
+		// behalf — which would mask real operational events (why did the
+		// WAL disappear? did GitOps drift the PVC?) — surface it as a
+		// FailedPrecondition with the exact remediation command in the
+		// message. Client-side, tryAddLearner treats this specific reason as
+		// fatal-with-clear-message so the pod fails fast with an operator-
```
NumaryBot, line 229: 🔴 **[blocker] Fail-fast bypassed when rejoining node has a fresh instance ID after WAL/PVC wipe**

When a rejoining pod has its WAL/PVC wiped, `wal.EnsureInstanceID` generates a new instance ID. `Node.AddLearner` detects the identity mismatch and issues `ConfChangeUpdateNode`, returning nil instead of `ErrNodeAlreadyInCluster`. The new fail-fast guard therefore never fires: `JoinAsLearner` reports success while the leader still holds a stale `Progress.Match` index, leaving the `tocommit out of range` crash loop — the root problem this PR targets — fully reproducible on the wiped-node rejoin path.

**Suggestion:** Before calling `AddLearner`, explicitly check whether the leader already holds a Progress entry for the incoming node ID with a Match index greater than zero (i.e. stale Progress), and return `FailedPrecondition` in that case regardless of whether the stored instance ID matches the incoming one. This ensures the fail-fast fires for both the identical-identity and fresh-identity (wiped WAL) rejoin paths.

internal/infra/node/node.go
```diff
@@ -2305,14 +2334,31 @@ func (node *Node) AddLearner(ctx context.Context, nodeID uint64, raftAddr, servi
...
+			// CLUSTER_JOINED marker, i.e. an empty/reprovisioned WAL — then
+			// the leader's known match index points at state the caller
+			// cannot have, and the next MsgApp/heartbeat would trip
+			// etcd-raft's "tocommit out of range" panic. Fail fast BEFORE
+			// deciding between AlreadyExists and the UpdateNode refresh: the
+			// stale-progress condition must fire on both the identical-
+			// identity rejoin and the fresh-identity (WAL-wiped) rejoin,
+			// the latter of which otherwise slips through as a benign
+			// ConfChangeUpdateNode refresh and re-introduces the crash loop.
+			if prog.Match > 0 {
```
NumaryBot: 🔴 **[blocker] Stale-progress guard incorrectly rejects legitimate pre-registered learners**

When a learner is pre-registered through `cluster.AddLearner`, its peer row intentionally has an empty `instance_id` until the pod boots and `JoinAsLearner` refreshes it. The leader can catch the learner up and set `Progress.Match > 0` before the refresh RPC arrives. The current stale-progress guard returns `ErrNodeStaleProgress` before checking whether the stored identity is empty, causing a legitimate new learner to fail with the `remove-node --force` remediation instead of completing the documented admin-add + boot flow.

**Suggestion:** Skip the stale-progress check (or treat it as a no-op) when the stored `instance_id` for the peer is empty, indicating a pre-registered but not-yet-booted learner. Only trigger the fail-fast when the stored instance ID is non-empty and mismatches or when the identity is identical but Progress.Match is stale.

internal/bootstrap/module.go
```diff
@@ -1449,10 +1450,41 @@ func tryAddLearner(ctx context.Context, cfg Config, tlsCfg TLSConfig, logger log
...
 
 			st, ok := status.FromError(err)
+			// EN-1436: the leader detected this nodeID in its raft Progress
+			// with a non-zero Match but we (the caller) have no CLUSTER_JOINED
+			// marker on our WAL. The leader's Match for us points at state we
+			// don't have, and the next MsgApp/heartbeat would trigger
+			// etcd-raft's "tocommit out of range" panic. Fail fast with the
+			// operator-actionable server message rather than retry, mark, or
+			// silently crash-loop. The server message already contains the
+			// exact remediation command; surface it verbatim as the fatal
```
NumaryBot, line 1546: 🟡 **[minor] Stale-membership error message emitted for all FailedPrecondition join rejections**

The server also returns `codes.FailedPrecondition` when a join is rejected by the removed-member registry (the `IsRemoved`/`ErrNodeRemoved` path). With the new catch-all on `FailedPrecondition`, that path is now logged as `STALE MEMBERSHIP` and surfaces a `stale raft membership` error to the operator, even though the correct remediation is `forget-removed`, not `remove-node --force`. This produces misleading operator guidance for an unrelated failure class.

**Suggestion:** Scope the stale-membership operator message to the specific stale-Progress `FailedPrecondition` response only — for example by checking a distinguishing error detail, status message field, or a dedicated gRPC error code — and leave all other `FailedPrecondition` cases with their original handling and error reason.

internal/adapter/grpc/server_bootstrap.go
NumaryBot, line 229: 🔴 **[blocker] Fail-fast bypassed when rejoining node has a fresh instance ID after WAL/PVC wipe**

When a rejoining pod has its WAL/PVC wiped, `wal.EnsureInstanceID` generates a new instance ID. `Node.AddLearner` detects the identity mismatch and issues `ConfChangeUpdateNode`, returning nil instead of `ErrNodeAlreadyInCluster`. The new fail-fast guard therefore never fires: `JoinAsLearner` reports success while the leader still holds a stale `Progress.Match` index, leaving the `tocommit out of range` crash loop — the root problem this PR targets — fully reproducible on the wiped-WAL/PVC rejoin path.

**Suggestion:** Before calling `AddLearner`, explicitly check whether the leader already holds a Progress entry for the incoming node ID with a Match index greater than zero (i.e. stale Progress), and return `FailedPrecondition` in that case regardless of whether the stored instance ID matches the incoming one. This ensures the fail-fast fires for both the identical-identity and fresh-identity (wiped WAL) rejoin paths.

internal/infra/node/node.go
```diff
@@ -2254,6 +2275,45 @@ func (node *Node) retryConfChange(ctx context.Context, nodeID uint64, name strin
...
+)
+
+// classifyExistingLearner decides what AddLearner must do when the leader's
+// Progress already carries the joining nodeID (EN-1436). The Match > 0 check
+// precedes the identity comparison so the stale-progress fail-fast fires on
+// BOTH the identical-identity rejoin and the fresh-identity (WAL-wiped) rejoin
+// — the latter otherwise slips through as a benign ConfChangeUpdateNode refresh
+// and re-introduces the crash loop this guard exists to prevent.
+func classifyExistingLearner(match uint64, existingInstanceID []byte, hasRow bool, incomingInstanceID []byte) existingLearnerAction {
+	if match > 0 {
```
NumaryBot, line 2306: 🔴 **[blocker] Stale-progress guard incorrectly rejects legitimate pre-registered learners**

When a learner is pre-registered through `cluster.AddLearner`, its peer row intentionally has an empty `instance_id` until the pod boots and `JoinAsLearner` refreshes it. The leader can catch the learner up and set `Progress.Match > 0` before the refresh RPC arrives. The current stale-progress guard returns `ErrNodeStaleProgress` before checking whether the stored identity is empty, causing a legitimate new learner to fail with the `remove-node --force` remediation instead of completing the documented admin-add + boot flow.

**Suggestion:** Skip the stale-progress check (or treat it as a no-op) when the stored `instance_id` for the peer is empty, indicating a pre-registered but not-yet-booted learner. Only trigger the fail-fast when the stored instance ID is non-empty and either mismatches or is identical but has a stale Progress.Match.

internal/bootstrap/module.go
NumaryBot, line 1546: 🟡 **[minor] Stale-membership error message emitted for all FailedPrecondition join rejections**

The server also returns `codes.FailedPrecondition` when a join is rejected by the removed-member registry (the `IsRemoved`/`ErrNodeRemoved` path). With the new catch-all on `FailedPrecondition`, that path is now logged as `STALE MEMBERSHIP` and surfaces a `stale raft membership` error to the operator, pointing them at `remove-node --force` even though the correct remediation for a removed member is `forget-removed`. This produces misleading operator guidance for an unrelated failure class.

**Suggestion:** Scope the stale-membership operator message to the specific stale-Progress `FailedPrecondition` response only — for example by embedding a distinguishing error detail, status message field, or a dedicated gRPC status code — and leave all other `FailedPrecondition` cases with their original handling and error reason.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1478#issuecomment-4865338697

gfyrag:
Review pass on `570e3964b137bf9e6ed083375739c69a9b69e95c`.

Findings:

1. Blocker: `classifyExistingLearner` rejects a legitimate admin pre-register + later boot flow when the leader has already advanced the learner's `Progress.Match`. The admin path intentionally calls `AddLearner(..., nil)` because the pod has not booted and has no instance ID yet, creating a peer row with empty identity that should be refreshed when the pod later joins. The new classifier returns stale-progress for any `match > 0` before checking that stored identity is empty. A pre-registered learner can receive raft traffic before its `JoinAsLearner` refresh, so this can fail a valid boot with stale-membership remediation. Please treat `hasRow && existingInstanceID empty` as a refresh case even when `match > 0`, or otherwise distinguish admin pre-registration from wiped-WAL rejoin. The new classifier test currently locks in the bad case as expected behavior.

2. Minor: bootstrap comments still describe `AlreadyExists` as normal idempotent join success, but this PR changes that to a deprecated rolling-upgrade compatibility branch. Please align the earlier comments with the later branch comment.

3. Minor/docs: this adds an operator-facing fail-fast state and remediation path, but docs did not change. `docs/ops/cluster-operations.md` still describes learner registration as always proceeding to `ConfChangeAddLearnerNode` and does not document the stale-progress failure/remediation near force removal. That should be updated under the documentation-maintenance rule for new technical mechanisms.

Outdated comments: the fresh-instance bypass concern appears fixed by checking `Progress.Match` before identity comparison, and the broad `FailedPrecondition` handling concern appears addressed via `ErrorInfo` reason matching.

Validation run: `GOROOT= go test ./internal/infra/node ./internal/adapter/grpc ./internal/bootstrap`, focused e2e cluster tests for auto-promotion/learner edge cases, and `GOROOT= go build ./...` passed.

flemzord (commented):
One new blocker-level startup-order finding after reviewing the stale Progress rejoin flow.

internal/bootstrap/module.go
```diff
@@ -1449,10 +1450,41 @@ func tryAddLearner(ctx context.Context, cfg Config, tlsCfg TLSConfig, logger log
...
+			// reason.
+			//
+			// Match on the STALE_RAFT_PROGRESS ErrorInfo reason, not on the
+			// bare FailedPrecondition code: the removed-member blacklist
+			// rejection (EN-1045) is also FailedPrecondition but needs
+			// `forget-removed`, not `remove-node --force`. Conflating the two
+			// would print misleading remediation. A blacklist rejection falls
+			// through to the generic fatal handler below with its own message.
+			if ok && st.Code() == codes.FailedPrecondition && isStaleRaftProgress(st) {
```
flemzord, line 1555: 🔴 **[blocker] Run this preflight before accepting Raft traffic**

This check can lose the race it is meant to prevent. The lifecycle starts the Raft gRPC server (and `membership.Start()`), then starts `node.Run()` and its orchestrate tasks, before the later hook calls `tryAddLearner`; once the stale member's address becomes reachable, the leader can therefore deliver the stale `MsgApp`/heartbeat and `rawNode.Step` can hit `tocommit out of range` before this RPC returns `STALE_RAFT_PROGRESS`. Move the membership preflight ahead of starting/accepting Raft traffic (or gate inbound Raft messages until it succeeds), and cover the ordering with a test that makes a stale message available before learner registration.

gfyrag:
Follow-up after the latest review thread on `570e3964b137bf9e6ed083375739c69a9b69e95c`:

The startup-order concern is valid and should be treated as a blocker. The Raft gRPC server starts accepting transport before `node.Run()` and before `tryAddLearner()` performs the stale-progress preflight. Incoming `StreamMessages` traffic can be enqueued and then stepped once `node.Run()` starts, before the preflight returns. That leaves a window where the stale membership state this PR is trying to reject can already process Raft messages.

Please move the stale membership preflight before accepting/processing Raft traffic, or gate inbound Raft message handling until the preflight succeeds.

The prior pre-registered learner issue also still stands: `classifyExistingLearner` returns stale-progress for any `match > 0` before considering an empty stored `instance_id`, which can reject the legitimate admin pre-register -> later boot identity refresh path.

The narrow fresh-instance bypass concern is otherwise stale inside `AddLearner`, and the broad `FailedPrecondition` concern appears fixed by checking the `STALE_RAFT_PROGRESS` error reason.

internal/bootstrap/module.go
gfyrag: Good catch — reordered so the join preflight runs before any inbound Raft traffic can be stepped. Pushed in `13cfc3528`.

**The reorder** (`internal/bootstrap/module.go`): the `tryAddLearner` preflight OnStart hook was the *last* Raft-related startup hook (it ran after the Raft gRPC server start + `membership.Start()` and after `node.Run()`). It's now registered as the **first** Raft startup hook — ahead of the transport start, the Raft gRPC server start / `membership.Start()`, and `node.Run()`. fx runs OnStart hooks in `Append` order and halts the sequence on the first error, so on stale Progress the preflight now aborts boot with `STALE_RAFT_PROGRESS` **before** anything can push a stale `MsgApp`/heartbeat into `rawNode.Step` — closing the `tocommit out of range` race this PR targets.

**Inbound vs outbound / why this doesn't deadlock:** the preflight reaches the leader over an **outbound**, short-lived gRPC client it dials itself (`ggrpc.NewClient` inside `tryAddLearner`) — it deliberately does *not* use the transport connection pool that `membership.Start()` wires, nor does it need the local Raft gRPC server (inbound serving) to be up. So the only thing gated behind the preflight is **inbound** stepping/serving; **outbound** dialing is always available because it's self-contained. Gating inbound-serving on an outbound-only preflight is therefore safe — no circular dependency.

The non-stale path is unchanged: preflight succeeds (or is skipped for bootstrap / no-peers / already-`CLUSTER_JOINED` nodes via `shouldRunJoinPreflight`), then transport → server → `node.Run()` start normally. Typed `ErrNodeStaleProgress` / `STALE_RAFT_PROGRESS` semantics from the earlier commits are untouched.

**Test:** `TestJoinPreflightRunsBeforeRaftTraffic` (`internal/bootstrap/join_preflight_test.go`) builds a real `fx.App` in the same registration order as the module (preflight first, a sentinel "inbound Raft traffic" hook second), makes the preflight fail deterministically (unreachable peer + a start context cancelled once the preflight is under way), and asserts the sentinel's OnStart **never runs** — i.e. no inbound Raft traffic is enabled when the preflight fails. Plus `TestShouldRunJoinPreflight` covers the skip logic.

Verified: `go build ./...`, `go test ./internal/bootstrap/... ./internal/infra/node/... ./internal/adapter/grpc/...` green, gofmt + golangci-lint clean.

internal/infra/node/node.go
NumaryBot, line 2306: 🔴 **[blocker] Stale-progress guard incorrectly rejects legitimate pre-registered learners**

When a learner is pre-registered via the admin `cluster add-learner` path, its peer row intentionally has an empty `instance_id` until the pod boots and `JoinAsLearner` refreshes it. The leader can advance `Progress.Match > 0` before the refresh RPC arrives. The current guard fires on `match > 0` unconditionally, before checking whether the stored identity is empty, causing a legitimate new learner to fail with the `remove-node --force` remediation instead of completing the documented admin-add + boot flow.

**Suggestion:** Skip the stale-progress check (or treat it as a no-op) when the stored `instance_id` for the peer is empty, indicating a pre-registered but not-yet-booted learner. Only trigger the fail-fast when the stored instance ID is non-empty and either mismatches or is identical but has a stale Progress.Match.

internal/adapter/grpc/server_bootstrap.go
NumaryBot, line 229: 🔴 **[blocker] Fail-fast bypassed when rejoining node has a fresh instance ID after WAL/PVC wipe**

When a rejoining pod has its WAL/PVC wiped, `wal.EnsureInstanceID` generates a new instance ID. `Node.AddLearner` detects the identity mismatch, issues `ConfChangeUpdateNode`, and returns nil instead of `ErrNodeAlreadyInCluster`. The new fail-fast guard therefore never fires: `JoinAsLearner` reports success while the leader still holds a stale `Progress.Match` index, leaving the `tocommit out of range` crash loop — the root problem this PR targets — fully reproducible on the wiped-WAL/PVC rejoin path.

**Suggestion:** Before calling `AddLearner`, explicitly check whether the leader already holds a Progress entry for the incoming node ID with a Match index greater than zero (stale Progress), and return `FailedPrecondition` in that case regardless of whether the stored instance ID matches the incoming one. This ensures the fail-fast fires for both the identical-identity and fresh-identity (wiped WAL) rejoin paths.

internal/bootstrap/module.go
```diff
@@ -1449,10 +1492,41 @@ func tryAddLearner(ctx context.Context, cfg Config, tlsCfg TLSConfig, logger log
...
```
NumaryBot, line 1555: 🔴 **[blocker] Stale-membership preflight can lose the race with incoming Raft traffic**

The lifecycle starts the Raft gRPC server (and `membership.Start()`), then starts `node.Run()` and its orchestrate tasks, before a later hook calls `tryAddLearner`. Once the stale member's address becomes reachable, the leader can deliver a stale `MsgApp`/heartbeat and `rawNode.Step` can hit `tocommit out of range` before the preflight RPC returns `STALE_RAFT_PROGRESS`. The guard is therefore not guaranteed to prevent the crash loop it is intended to stop.

**Suggestion:** Move the membership preflight ahead of starting/accepting Raft traffic, or gate inbound Raft message handling until the preflight succeeds. Add a test that makes a stale message available before learner registration to cover this ordering.

internal/bootstrap/module.go
```diff
@@ -1449,10 +1492,41 @@ func tryAddLearner(ctx context.Context, cfg Config, tlsCfg TLSConfig, logger log
...
 
 			st, ok := status.FromError(err)
+			// EN-1436: the leader detected this nodeID in its raft Progress
+			// with a non-zero Match but we (the caller) have no CLUSTER_JOINED
+			// marker on our WAL. The leader's Match for us points at state we
+			// don't have, and the next MsgApp/heartbeat would trigger
+			// etcd-raft's "tocommit out of range" panic. Fail fast with the
+			// operator-actionable server message rather than retry, mark, or
+			// silently crash-loop. The server message already contains the
+			// exact remediation command; surface it verbatim as the fatal
```
NumaryBot, line 1546: 🟡 **[minor] Stale-membership error message emitted for all FailedPrecondition join rejections**

The server also returns `codes.FailedPrecondition` when a join is rejected by the removed-member registry (`IsRemoved`/`ErrNodeRemoved` path). With the new catch-all on `FailedPrecondition`, that path is now logged as `STALE MEMBERSHIP` and surfaces a `stale raft membership` error to the operator pointing at `remove-node --force`, even though the correct remediation for a removed member is `forget-removed`. This produces misleading operator guidance for an unrelated failure class.

**Suggestion:** Scope the stale-membership operator message to the specific stale-Progress `FailedPrecondition` response only — for example by embedding a distinguishing error detail, status message field, or a dedicated gRPC status code — and leave all other `FailedPrecondition` cases with their original handling and error reason.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1478#issuecomment-4865338697

gfyrag:
Review pass on `13cfc3528c3db4e350b98814f1ab835d67606677` after the rebase.

The startup-order blocker looks fixed: the join preflight hook now runs before the transport/server/membership/node start path, so inbound Raft traffic is not accepted before the stale-progress preflight.

One blocker remains: `classifyExistingLearner` still rejects the supported admin pre-register -> later boot flow. It returns stale-progress for any `match > 0` before checking whether the persisted peer row has an empty `instance_id`. For an admin-created learner row, empty identity is expected until `JoinAsLearner` refreshes it. If the leader has already advanced `Progress.Match`, this fails a legitimate boot with `remove-node --force` guidance.

Please treat `hasRow && existingInstanceID empty` as a refresh case before stale-progress classification, while still failing WAL-wiped rejoins where the stored identity is non-empty.

The broad `FailedPrecondition` concern remains stale: the client now checks the `STALE_RAFT_PROGRESS` reason before emitting stale-membership remediation.

flemzord (approved):
Reviewed the current head after the follow-up fixes. The previously reported issue is addressed, and I found no remaining actionable issue in this PR.

internal/infra/node/node.go
```diff
@@ -2255,6 +2276,45 @@ func (node *Node) retryConfChange(ctx context.Context, nodeID uint64, name strin
...
```
NumaryBot, line 2306: 🔴 **[blocker] Stale-progress guard incorrectly rejects legitimate pre-registered learners**

When a learner is pre-registered via the admin `cluster add-learner` path, its peer row intentionally has an empty `instance_id` until the pod boots and `JoinAsLearner` refreshes it. The leader can advance `Progress.Match > 0` before the refresh RPC arrives. The current guard fires on `match > 0` unconditionally, before checking whether the stored identity is empty, causing a legitimate new learner to fail with the `remove-node --force` remediation instead of completing the documented admin-add + boot flow. The test for `classifyExistingLearner` currently locks in this incorrect behavior as expected.

**Suggestion:** Skip the stale-progress check (or treat it as a no-op) when the stored `instance_id` for the peer is empty, indicating a pre-registered but not-yet-booted learner. Only trigger the fail-fast when the stored instance ID is non-empty and either mismatches or is identical but has a stale Progress.Match.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1478#issuecomment-4865338697

internal/infra/node/node.go
NumaryBot, line 2306: 🔴 **[blocker] Stale-progress guard incorrectly rejects legitimate pre-registered learners and admin retries**

`classifyExistingLearner` (and the broader `AddLearner` guard) fires on `Progress.Match > 0` unconditionally, before checking whether the stored peer row has an empty `instance_id`. An admin-pre-registered learner intentionally has an empty identity row until the pod boots and `JoinAsLearner` refreshes it. If the leader has already advanced `Match` before that refresh RPC arrives, a valid first boot is rejected with `ErrNodeStaleProgress` and the operator is pointed at `remove-node --force` — the wrong remediation. The same problem applies to idempotent admin `cluster add-learner` retries for already-active members: when `incomingInstanceID` is nil (admin path), any `match > 0` now returns `ErrNodeStaleProgress` instead of the previous `AlreadyExists`. The accompanying unit test cements the incorrect behavior as expected. This must be fixed before merge.

**Suggestion:** Skip the stale-progress check (or treat it as a refresh case) when the stored `instance_id` for the peer is empty, indicating a pre-registered but not-yet-booted learner. Separately, scope the stale-progress fail-fast strictly to the `JoinAsLearner` boot path where a real incoming instance ID is present, and preserve idempotent `AlreadyExists` semantics for the admin `AddLearner` API path (nil instance ID). Only trigger the fail-fast when the stored identity is non-empty and either mismatches or is identical but has a stale `Progress.Match`.

internal/bootstrap/module.go
```diff
@@ -1493,10 +1536,41 @@ func tryAddLearner(ctx context.Context, cfg Config, tlsCfg TLSConfig, logger log
...
```
NumaryBot, line 1546: 🟡 **[minor] Stale-membership error message emitted for all FailedPrecondition join rejections, not just stale-progress ones**

The server also returns `codes.FailedPrecondition` when a join is rejected via the removed-member registry (`IsRemoved`/`ErrNodeRemoved` path). The new catch-all on `FailedPrecondition` logs `STALE MEMBERSHIP` and surfaces a `stale raft membership` error with `remove-node --force` guidance for that path too, even though the correct remediation there is `forget-removed`. This produces misleading operator guidance for an unrelated failure class.

**Suggestion:** Scope the stale-membership operator message to the specific stale-Progress `FailedPrecondition` response only — for example by checking a distinguishing error detail, status message field, or a dedicated gRPC status code — and leave all other `FailedPrecondition` cases with their original handling and error reason.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1478#issuecomment-4865338697

gfyrag:
Review pass on `69a28a444f409bf27530cc6d638e7dc9628d43c5` after the latest fix.

The startup-order blocker looks resolved: the join preflight now runs before Raft transport/server/node startup, so inbound Raft traffic is not accepted before the stale-progress preflight.

One high-severity issue remains, narrowed to the admin path:

**The stale-progress fail-fast is still applied to shared `AddLearner`, including admin retries with no joining instance ID.**

`classifyExistingLearner` returns `existingLearnerStaleProgress` for any `Progress.Match > 0` before considering the caller identity (`internal/infra/node/node.go:2306`). The admin `cluster add-learner` RPC intentionally calls `AddLearner(..., nil)` because the target pod has not booted yet and has no instance ID (`internal/adapter/grpc/server_cluster.go:335`).

That means an admin `cluster add-learner` retry against an already-active member/learner with non-zero `Match` now returns `ErrNodeStaleProgress` / `FailedPrecondition`, instead of the prior idempotent `ErrNodeAlreadyInCluster` / `AlreadyExists` outcome. The stale-progress failure is correct for `JoinAsLearner` with a real 16-byte incoming instance ID and no local `CLUSTER_JOINED` marker, but it should not be applied to the admin pre-registration/retry path where `incomingInstanceID == nil`.

The current NumaryBot comment on the startup ordering path is stale; the node/classifier finding is still valid in the narrower admin-path form above.

internal/bootstrap/module.go
gfyrag: Confirmed the reorder is on the branch head as commit `c12a0f8b` (the earlier `13cfc3528` reference was from a rebase-superseded push). Verified the fx OnStart ordering directly against the wiring in `internal/bootstrap/module.go`:

1. `joinPreflightHook` (`tryAddLearner`) — registered first (module.go:900)
2. transport `t.Start()` — module.go:909
3. Raft gRPC server `raftServer.Start()` + `membership.Start()` — module.go:1015
4. `node.Run()` — module.go:1092

fx runs OnStart hooks in `Append` order and halts on the first error, so the preflight runs (and, on stale Progress, aborts boot) before any of (2)–(4). Crucially, `rawNode.Step` is only ever called from inside `node.Run`'s message loop (`internal/infra/node/node.go:1533`) — the transport and gRPC server merely enqueue inbound messages; nothing is stepped into raft until (4). The preflight at position (1) therefore completes or fails-fast strictly before any stale `MsgApp`/heartbeat can reach `rawNode.Step` and trip `tocommit out of range`.

Regression coverage: `TestJoinPreflightRunsBeforeRaftTraffic` (`internal/bootstrap/join_preflight_test.go`) asserts the sentinel inbound-Raft hook never starts when the preflight fails. `go build ./...` and `go test ./internal/bootstrap/...` green. Resolving.

internal/infra/node/node.go
gfyrag: Fixed in `24bfdb50e`. Agreed on the admin-retry over-rejection; narrowed the guard.

`classifyExistingLearner` was firing `existingLearnerStaleProgress` on any `Match > 0` **before** checking the incoming instance_id. That's wrong for the admin `cluster add-learner` API path (`server_cluster.go:335` passes `nil`): an idempotent operator retry against an already-active member (`Match > 0`) got `ErrNodeStaleProgress` + `remove-node --force` guidance instead of the pre-EN-1436 `AlreadyExists`.

The `tocommit out of range` hazard only exists for a pod booting with a fresh, empty WAL — i.e. the `JoinAsLearner` path, which is the *sole* caller carrying a real 16-byte instance_id (`server_bootstrap.go:201`). So the fail-fast is now gated on `isBootJoin` (`len(incomingInstanceID) == 16`):

```go
isBootJoin := len(incomingInstanceID) == 16
if match > 0 && isBootJoin {
    return existingLearnerStaleProgress
}
needsRefresh := hasRow && isBootJoin && !bytes.Equal(existingInstanceID, incomingInstanceID)
...
return existingLearnerAlreadyInCluster
```

Effect:
- admin path (nil / non-16-byte id), `Match > 0` -> `existingLearnerAlreadyInCluster` -> `ErrNodeAlreadyInCluster` (idempotent, restored).
- boot path (real 16-byte id), `Match > 0` -> `existingLearnerStaleProgress` -> fail fast. Still fires on the fresh-identity (WAL-wiped) rejoin, since that reboots via `JoinAsLearner` with a real id — the crash-loop this guard exists to prevent is untouched.

On the "pre-registered learner not yet caught up" sub-case: a pre-registered-but-unbooted learner the leader hasn't connected to has `Match == 0`, so it never reached the guard — it flows to `NeedsRefresh`/`AlreadyInCluster` as before. The sharp, real bug here was the admin nil-id path, which is now fixed.

Added `TestClassifyExistingLearner` cases: admin retry with nil id and with a non-16-byte id, both asserting `existingLearnerAlreadyInCluster` at `Match > 0`. `go build ./...` + `go test ./internal/infra/node/... ./internal/bootstrap/...` green.

internal/bootstrap/module.go
gfyrag: Already scoped correctly on head — no change needed here, and `24bfdb50e` narrows it further.

The client-side classifier does **not** key on the bare `FailedPrecondition` code. `internal/bootstrap/module.go:1555` gates the "stale raft membership" fatal on `isStaleRaftProgress(st)` (module.go:1633), which matches the `STALE_RAFT_PROGRESS` `ErrorInfo` reason:

```go
if ok && st.Code() == codes.FailedPrecondition && isStaleRaftProgress(st) {
    ... "stale raft membership on the leader: %s" ...
}
```

Server-side, that reason is attached **only** on the `ErrNodeStaleProgress` branch (`server_bootstrap.go:242-260` via `status.WithDetails(&errdetails.ErrorInfo{Reason: node.StaleRaftProgressReason})`). The removed-member blacklist rejection is a *separate* `FailedPrecondition` with its own `forget-removed` message and no such reason, so it falls through to the generic handler — it never surfaces the stale-membership line or `remove-node --force` guidance. The doc comment at module.go:1549-1554 spells this out explicitly.

So unrelated `FailedPrecondition` causes do not masquerade as stale-progress. My guard-narrowing change also *reduces* the set of calls that can reach the stale-progress branch at all (admin nil-id retries no longer return `ErrNodeStaleProgress`). Resolving.

PR #1479, opened 2026-07-02
fix(EN-1442): sanitize HTTP adapter panic recovery and unmapped errors

paul-nicolas:
## Summary

The HTTP adapter (`internal/adapter/http`) serialized raw internal error strings straight into the JSON response body on two paths, disclosing panic values, wrapped Pebble errors, filesystem paths, and internal invariant messages to any caller. This is an **information-disclosure** bug (Major). It mirrors the gRPC hardening already shipped in #375, which the HTTP side never received.

### Leak paths fixed

1. **Panic recovery** — `jsonRecoverer` called `debug.PrintStack()` (bare stderr, no correlation) and returned `fmt.Errorf("panic recovered: %v", rvr)` verbatim to the client.
2. **Unmapped errors** — `handleError`'s default branch passed any non-`domain.Describable` error to `writeInternalServerError`, which serialized `err.Error()` verbatim.

## Changes

- **`jsonRecoverer`** — logs the raw panic value + stack server-side with a `correlation_id`, records them on the OTel span, and returns only `internal server error (correlation ID: <id>)` to the client. `debug.PrintStack()` removed.
- **`writeInternalServerError`** — logs the raw error with a `correlation_id` and returns the same generic body. This is `handleError`'s fallthrough sanitization boundary, so every non-domain error is sanitized uniformly (including the pre-existing benign fixed-string callers).
- **`loggerMiddleware`** (new) — injects the configured logger into the request context. Unlike gRPC's interceptor closure, the HTTP stack never propagated the app logger (`httpserver.NewHook` sets no `BaseContext`, and the chi stack does not use go-libs' `LoggerMiddleware`), so `logging.FromContext` degraded to a bare stderr logger. Wired **before** `jsonRecoverer` so the logger resolves inside `recover()`.
- **`correlationID`** (new helper) — reuses Chi's `RequestID` (`X-Request-Id`), with a random-hex fallback, so operators can grep server logs for the exact ID a caller reports.

Domain `Describable` responses keep their descriptive messages unchanged — that is the intended contract, not a leak.

## Behavior

Both leak paths now return:
```json
{ "errorCode": "INTERNAL_ERROR", "errorMessage": "internal server error (correlation ID: <id>)" }
```
with the raw cause + correlation ID logged server-side and recorded on the OTel span.

## Tests

- `TestJSONRecoverer_SanitizesPanic` / `TestWriteInternalServerError` — assert the raw error/panic value never reaches the client and is logged server-side with the correlation ID.
- `TestJSONRecoverer_PropagatesErrAbortHandler`, `TestJSONRecoverer_NoPanicPassesThrough`.
- `TestHandleError` unknown-error case strengthened to assert no leak.
- `TestCorrelationID_*` — RequestID reuse + fallback.

## Verification

- `just pre-commit` (nix): golangci-lint 0 issues, no generated-file drift.
- `just test` (`go test -race ./... -timeout 20m`): passing, 0 failures.

## Docs

Added an "Internal Errors and Sanitization" subsection to `docs/technical/architecture/subsystems/api/http-api.md`.

## Audit

Reviewed every other `writeErrorResponse` caller: all pass fixed strings, client-input/validation messages (400s, matching gRPC's `InvalidArgument`), or intentional domain `Describable` messages. The two paths fixed here were the only internal-error disclosure vectors.

---
Jira: https://formance-team.atlassian.net/browse/EN-1442


internal/adapter/http/middleware_recoverer_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"net/http/httptest"
+	"testing"
+
+	"github.com/go-chi/chi/v5/middleware"
+	"github.com/stretchr/testify/require"
+
+	logging "github.com/formancehq/go-libs/v5/pkg/observe/log"
+)
+
+func TestJSONRecoverer_SanitizesPanic(t *testing.T) {
```
paul-nicolas, line 23: [Medium] Sanitization is only tested in isolation — the middleware ordering the fix depends on is never exercised end-to-end.

This test (and `TestWriteInternalServerError`) hand-injects both the logger (`logging.ContextWithLogger`) and the request ID (`middleware.RequestIDKey`) directly into the context, then calls `jsonRecoverer` / `writeInternalServerError` in isolation. That validates the sanitization *content*, but it does not protect the property the fix actually relies on: that in `NewHandler`'s chain `loggerMiddleware` and `middleware.RequestID` run *outer* to `jsonRecoverer`. A future reorder that moves `loggerMiddleware` after `jsonRecoverer` (logger falls back to bare stderr on panic) or drops `RequestID` (correlation IDs stop matching) would leave every test green.

The repo already has the pattern for this — `handlers_*_test.go` build the real handler via `NewHandler(...)` and issue full `httptest` requests. Add one such test with a panicking route (and one hitting the unmapped-error fallthrough) that asserts (a) the JSON 500 body is exactly `internal server error (correlation ID: <id>)`, and (b) the same `<id>` appears in a captured server-side log buffer.

docs/technical/architecture/subsystems/api/http-api.md
```diff
@@ -59,6 +59,26 @@ The server supports optional JWT/OIDC authentication with scope-based authorizat
...
+### Internal Errors and Sanitization
+
+Business errors (validation, not-found, conflict, etc.) map to specific status codes and carry a machine-readable `errorCode` and a descriptive `errorMessage`, mirroring the gRPC adapter's `Describable` contract.
+
+Two paths, however, must never expose internal error text to clients — the raw value could contain filesystem paths, wrapped Pebble/storage errors, or internal invariant strings:
+
+1. **Panic recovery** (`jsonRecoverer`) — a panic in any handler.
+2. **Unmapped errors** (`handleError` fallthrough → `writeInternalServerError`) — any error that is not a domain `Describable` or a known sentinel.
+
+Both paths are sanitized identically to the gRPC adapter: the raw cause (and, for panics, the stack) is logged **server-side** with a `correlation_id` field and recorded on the OTel span, while the client receives only a generic body:
```
paul-nicolas: [Medium] Doc drift: this says the raw cause is "recorded on the OTel span" for *both* paths, but the unmapped-error path only logs.

`jsonRecoverer` (panic path) does record on the span — `span.SetAttributes(panic.value/panic.stack/correlation_id)` + `span.RecordError(...)`. But `writeInternalServerError` (the `handleError` fallthrough, `response.go`) only calls `logging.FromContext(...).Errorf(...)`; it never touches `trace.SpanFromContext`. So the "and recorded on the OTel span" clause is false for the unmapped path — an operator working from traces alone will find nothing for a fallthrough 500.

This matches gRPC's own unmapped-error branch (`convertToGRPCError` default logs only, no span), so the parity claim holds for behavior — it's the wording that overstates. Either narrow this sentence to scope span-recording to the panic path, or add the same `SpanFromContext` + `RecordError` block to `writeInternalServerError` for true parity with the panic path.

paul-nicolas (commented):

Well-scoped, correct information-disclosure fix. Both reviewers independently reached **APPROVE**: middleware ordering is correct (`RequestID` outermost so `correlationID` resolves; `loggerMiddleware` is outer to `jsonRecoverer` so `logging.FromContext` resolves inside `recover()`), `ErrAbortHandler` is re-panicked before any sanitization, and there are no races. Verified against the code.

**Severity counts:** 0 Critical, 0 Major, 2 Medium (posted inline), 4 Minor + 1 Nit (listed below, not posted inline).

**Codex/Claude reconciliation:** Codex rated the pre-existing `Describable`/`KindInternal` path a Major disclosure vector; on inspection it is at *exact parity* with gRPC's `describableToGRPCStatus` (which also serializes `d.Error()` for `KindInternal`), is not touched by this PR, and is an explicit author decision ("domain Describable responses keep their descriptive messages — intended contract"). Downgraded to Minor. Both models independently flagged the end-to-end test gap; kept as Medium.

- `internal/adapter/http/error_handler.go:85` (Minor, pre-existing) — the `Describable` branch serializes the full wrapped `err.Error()`; `KindInternal` domain errors reach the client verbatim (e.g. `ErrNumscriptRuntime` carries `numscript panic: %v`). At parity with gRPC and explicitly scoped out here, but a latent leak worth a follow-up (either a domain-scoped message for `KindInternal`, or a review rule that domain errors are never wrapped with infra context).
- `internal/adapter/http/correlation.go:35` (Minor) — the `ts-%d` fallback is clock-based; if `rand.Read` fails system-wide, two concurrent failures on a coarse timer could collide. Consider a process-local atomic counter. (Only reachable if `rand.Read` fails, so effectively defensive.)
- `internal/adapter/http/middleware_recoverer.go:207` (Minor) — the full stack is recorded as an unbounded `panic.stack` OTel span attribute (already logged). Parity with gRPC `server.go:332`, so consistency-preserving; consider capping length or dropping it from the span.
- `internal/adapter/http/response.go` (Minor, docs) — the 500 body contract changed (raw message → generic). Verify `openapi.yml` (ErrorResponse example ~line 3692) and `api-comparison.md` don't show a raw-message 500.
- `internal/adapter/http/middleware_recoverer.go:213` (Minor, pre-existing) — `writeErrorResponse` may run after a handler already wrote headers, causing a superfluous `WriteHeader`. Out of scope; a `WrapResponseWriter` status check would fix it later.
- `docs/technical/architecture/subsystems/api/http-api.md:80` (Nit) — wording may imply `X-Request-Id` is echoed as a readable response header; chi's `RequestID` stores it in context/logs, the client sees the ID only in the error body. Reword to avoid implying a response header.

internal/adapter/http/middleware_recoverer_test.go
paul-nicolas: Fixed in 5d0283f85 — added `TestNewHandler_SanitizesPanicEndToEnd`. It builds the real handler via `NewHandler(...)`, forces a panic by mocking `Apply` to return zero logs (`handleCreateLedger` panics on that), and drives it through `handler.ServeHTTP`. Asserts the generic 500 body plus that the correlation ID in the client body matches the one logged server-side and that the log buffer is non-empty — locking in both `middleware.RequestID` and `loggerMiddleware` being outer to `jsonRecoverer`.

docs/technical/architecture/subsystems/api/http-api.md
paul-nicolas: Fixed in 5d0283f85 — went with the parity-preserving option: reworded so span-recording is scoped to the panic path only ("the raw cause is logged server-side with a correlation_id field (and, for a panic, additionally recorded on the OTel span together with the stack)"). This keeps the `writeInternalServerError` path log-only, matching gRPC's unmapped branch, rather than diverging by adding span recording.

gfyrag (approved):

PR #1480, opened 2026-07-02
feat(EN-1353): default volume deletion protection to on (opt-out)

paul-nicolas:
## Summary

Flips the ledger PVC/PV **deletion protection** (originally shipped opt-in in `formancehq/ledger-v3-poc#567`) to **default-ON, explicit opt-out** — the reverse of today. A freshly deployed ledger is now protected without any extra configuration; both layers must be explicitly turned off to opt out.

Jira: https://formance-team.atlassian.net/browse/EN-1353

## Changes

- **Cluster layer** — `helm/operator/values.yaml`: `pvcProtection.enabled` `false → true`. The cluster-scoped `ValidatingAdmissionPolicy` installs by default.
- **Per-CR layer** — `spec.persistence.deletionProtection`: `bool` (`+kubebuilder:default=false`) → `*bool` (`+kubebuilder:default=true`). The pointer is what makes "default true, but an explicit `false` still opts out" correct — a plain `bool` + `omitempty` can't distinguish unset from false.
  - New `PersistenceSpec.DeletionProtectionEnabled()` (nil → true) so in-process construction and the runtime resolve the default consistently; all consumers route through it. `hash.go` zeroes the field to `nil`.
- Regenerated CRDs (`config/crd/bases` + `helm/crds/templates`, both now `default: true`) and deepcopy via `controller-gen`.
- Rewrote the operator README and `values.yaml` comments from opt-in to opt-out.

## Old-cluster safety (K8s < 1.30)

`ValidatingAdmissionPolicy` is GA only in Kubernetes >= 1.30. Defaulting `pvcProtection.enabled` to true would otherwise break `helm install`/`upgrade` on older clusters ("no matches for kind ValidatingAdmissionPolicy"). Fixed by:

- Gating the VAP/VAPBinding template on `.Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1/ValidatingAdmissionPolicy"` — an old cluster **silently skips** the objects and the default install succeeds. (`helm template` offline: pass `--api-versions admissionregistration.k8s.io/v1/ValidatingAdmissionPolicy` to force-render.)
- A `NOTES.txt` warning when protection is enabled but the API is absent, so the degrade-to-unprotected is visible (upgrade to >=1.30, or set `enabled=false`).
- The runtime probe already reports "not installed" on such clusters (the unregistered resource answers the GET with a 404 that `IsNotFound` matches), so reconcile does not error on every pass — clarified in a comment.

Verified with `helm template`: 0 VAP objects rendered without the capability, 4 with it.

## Multi-release caveat worth reviewing

Making the Helm layer default-ON interacts with the **singleton constraint** from #567: the cluster VAP objects have fixed, release-independent names, so a **second operator release in the same cluster now fails its `helm install`/`upgrade` by default** (ownership conflict). In multi-release clusters, set `pvcProtection.enabled=false` on all but one release. Documented in both `values.yaml` and the README.

The other consequence — legitimate `kubectl delete pvc` / GitOps teardown now needs the `formance.com/allow-deletion` annotation (or an exempt SA) — is the intended safety trade-off.

## Testing

- `go build ./...`, `go test ./...`, `golangci-lint run ./...` all green (operator module); `helm lint` passes.
- `controller-gen` regeneration is stable (no drift).
- Helm render verified for both K8s >= 1.30 and < 1.30.
- Existing chainsaw e2e (`e2e/tests/pvc-deletion-protection`) still valid: it sets `deletionProtection: true` explicitly and exercises the toggle-to-`false` opt-out path.


gfyrag (approved):

misc/operator/helm/operator/templates/NOTES.txt
```diff
@@ -0,0 +1,20 @@
...
+(admissionregistration.k8s.io/v1), which is GA only in Kubernetes >= 1.30.
+
+The protection policy was NOT installed. LedgerServices with
+spec.persistence.deletionProtection=true (the default) will report a
+DeletionProtectionInactive condition because no policy is acting on their
+volumes — their PVCs/PVs are NOT protected from accidental deletion.
+
+To resolve, either:
+  - upgrade the cluster to Kubernetes >= 1.30, then re-run helm upgrade, or
+  - set pvcProtection.enabled=false to acknowledge protection is unavailable and
```
NumaryBot, line 17: 🟡 **[minor] Opt-out guidance in NOTES.txt is incomplete**

The Helm note instructs operators to set `pvcProtection.enabled=false` to resolve the `DeletionProtectionInactive` warning, but this only hides the note — it does not silence the warning in the controller. Since deletion protection now defaults to true, the per-CR opt-out (`spec.persistence.deletionProtection=false`) must also be mentioned for operators to fully suppress the warning.

**Suggestion:** Update the NOTES.txt guidance to include both the Helm-level flag (`pvcProtection.enabled=false`) and the per-CR opt-out (`spec.persistence.deletionProtection=false`) so operators have complete instructions to suppress the warning.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1480#issuecomment-4866244375

PR #1481, opened 2026-07-02
refactor(operator): rename LedgerService to Cluster and LedgerClusterAgent to Credentials

gfyrag:
## Summary
- Renames the `LedgerService` Kind end-to-end (Go types, CRD, kubectl-ledger, Helm chart, e2e tests, Antithesis manifests, proto comment, docs) to `Cluster`.
- Adds a transient compat layer: a deprecated `LedgerService` Kind (schema reuses `ClusterSpec`/`ClusterStatus` verbatim, no duplication) plus a `LedgerServiceMigrator` reconciler that copies each `LedgerService` into a same-named `Cluster` on reconcile, annotates the source with `ledger.formance.com/migrated-to` and sets a `Migrated=True` condition. The LS is never deleted — operators remove it manually once the migration has converged.
- Same names preserve derived StatefulSet / Services / PVCs → no downtime for in-place upgrades.

## Test plan
- [x] `go build ./...` clean in both root module and `misc/operator/`
- [x] `go vet ./...` clean in both modules
- [x] `go test ./... -short` passes (root + operator)
- [x] `golangci-lint run ./...` clean on operator; pre-existing unused-code warnings only in feature-flagged event sinks on the root module
- [ ] E2E `chainsaw` suites still green (need cluster)
- [ ] New `ledgerservice-migration` chainsaw test passes on kind
- [ ] `just generate` (nix) regenerates `zz_generated.deepcopy.go` and both CRDs idempotently (should be a no-op diff)
- [ ] Antithesis smoke run (`snouty launch`) still boots with the renamed manifests


misc/operator/internal/controller/ledgerservice_migrator.go
```diff
@@ -0,0 +1,110 @@
...
+
+	if _, done := ls.Annotations[migratedToAnnotation]; done {
+		return ctrl.Result{}, nil
+	}
+
+	cluster := &ledgerv1alpha1.Cluster{}
+	err := r.Get(ctx, types.NamespacedName{Name: ls.Name, Namespace: ls.Namespace}, cluster)
+	switch {
+	case apierrors.IsNotFound(err):
+		cluster = &ledgerv1alpha1.Cluster{
```
NumaryBot: 🔴 **[blocker] Rehome existing owner references during migration**

For in-place upgrades where a LedgerService has already created its ServiceAccount/Services/StatefulSet/etc., creating only the Cluster leaves those child objects with a controller ownerReference to the old LedgerService. The Cluster reconciler then calls `SetControllerReference` on the existing objects and will fail because they already have a different controller owner, so the migrated Cluster cannot actually manage or update the deployment; deleting the LedgerService later would also garbage-collect the existing resources. The migrator needs to adopt or otherwise rehome those ownerReferences as part of migration.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1481#issuecomment-4866361927

misc/operator/internal/controller/ledgerservice_migrator.go
```diff
@@ -0,0 +1,259 @@
...
```
NumaryBot: 🔴 **[blocker] Existing child resources retain LedgerService ownerReferences after migration**

For in-place upgrades where a LedgerService has already reconciled and created child resources (ServiceAccount, Services, StatefulSet, etc.), the migrator only creates a new Cluster object but never reparents the children. Those resources retain a controller ownerReference pointing to the old LedgerService. When the Cluster reconciler subsequently calls SetControllerReference on them it will fail because Kubernetes rejects a second controller owner, so the new Cluster cannot manage or update the existing deployment. Additionally, if the old LedgerService is later deleted, Kubernetes garbage-collects all its owned children, destroying the live workload.

**Suggestion:** During migration, enumerate all child resources owned by the LedgerService (using ownerReference lookup or the instance label selector) and patch their ownerReferences to replace the LedgerService controller reference with one pointing to the newly created Cluster, before marking the source as migrated.

misc/operator/internal/controller/ledgerservice_migrator.go
```diff
@@ -0,0 +1,259 @@
...
+		ObservedGeneration: ls.Generation,
+	})
+	if err := r.Status().Update(ctx, ls); err != nil {
+		return ctrl.Result{}, fmt.Errorf("updating LedgerService status: %w", err)
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// rehomeChildren walks every child object kind the Cluster reconciler owns
```
NumaryBot: 🟠 **[major] DNSEndpoint children not rehomed during migration**

When a legacy LedgerService had `spec.dnsEndpoint.enabled` set and the ExternalDNS CRD was installed, the reconcileDNSEndpoint path sets a controller ownerReference on the DNSEndpoint object. The migrator never lists or reparents this unstructured child resource. The new Cluster reconciler will fail SetControllerReference on the existing DNSEndpoint (already controlled by LedgerService), and deletion of the LedgerService will garbage-collect the live DNS record.

**Suggestion:** Extend the child adoption loop to also enumerate DNSEndpoint unstructured objects owned by the LedgerService (by GVR or ownerReference scan) and patch their controller ownerReference to point to the new Cluster.

misc/operator/internal/controller/ledgerservice_migrator.go
```diff
@@ -0,0 +1,259 @@
...
+//
+// Children are located by the operator's `app.kubernetes.io/instance=<name>`
+// label — the same label reconcile_*.go stamps on everything it creates.
+// Objects without a controller ownerRef to this LedgerService are ignored,
+// so the pass is safe to run repeatedly (idempotent) and cheap on already-
+// migrated services.
+func (r *LedgerServiceMigrator) rehomeChildren(ctx context.Context, ls *ledgerv1alpha1.LedgerService, cluster *ledgerv1alpha1.Cluster) error {
+	listOpts := []client.ListOption{
+		client.InNamespace(ls.Namespace),
+		client.MatchingLabels{labelInstance: ls.Name},
```
NumaryBot: 🟠 **[major] Instance-label selector misses children when additionalLabels overrides app.kubernetes.io/instance**

The migrator uses `app.kubernetes.io/instance` to select child resources for adoption. For LedgerService objects where `spec.additionalLabels` overrode this label, all children were stamped with the overridden value, so the selector matches nothing. The migrator then marks the source as migrated while all old ownerReferences remain in place, silently leaving the Cluster reconciler unable to adopt or manage those resources.

**Suggestion:** Instead of (or in addition to) the instance label selector, scan children by iterating the LedgerService's ownerReference UID across relevant resource types, which is label-value agnostic and will correctly identify all owned objects regardless of label overrides.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1481#issuecomment-4866361927

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1481#issuecomment-4866361927

misc/operator/internal/controller/ledgerservice_migrator.go
NumaryBot: 🔴 **[blocker] Child resources retain LedgerService ownerReferences after migration**

For in-place upgrades where a LedgerService has already reconciled and created child resources (ServiceAccount, Services, StatefulSet, etc.), the migrator only creates a new Cluster object but never reparents the children. Those resources retain a controller ownerReference pointing to the old LedgerService. When the Cluster reconciler subsequently calls SetControllerReference on them it will fail because Kubernetes rejects a second controller owner, so the new Cluster cannot manage or update the existing deployment. Additionally, if the old LedgerService is later deleted, Kubernetes garbage-collects all its owned children, destroying the live workload.

**Suggestion:** During migration, enumerate all child resources owned by the LedgerService (using ownerReference lookup or the instance label selector) and patch their ownerReferences to replace the LedgerService controller reference with one pointing to the newly created Cluster, before marking the source as migrated.

misc/operator/internal/controller/ledgerclusteragent_controller.go
```diff
@@ -207,17 +207,17 @@ func (r *LedgerClusterAgentReconciler) handleDeletion(ctx context.Context, agent
...
+// resolveMatchedServices lists Clusters across all namespaces and returns
 // those matching the agent's label selector.
 func (r *LedgerClusterAgentReconciler) resolveMatchedServices(ctx context.Context, agent *ledgerv1alpha1.LedgerClusterAgent) ([]ledgerv1alpha1.MatchedService, error) {
 	selector, err := metav1.LabelSelectorAsSelector(&agent.Spec.Selector)
 	if err != nil {
 		return nil, fmt.Errorf("parsing label selector: %w", err)
 	}
 
-	var services ledgerv1alpha1.LedgerServiceList
+	var services ledgerv1alpha1.ClusterList
```
NumaryBot: 🔴 **[blocker] Agent secrets GC'd as orphans when Cluster not yet created during migration**

The agent reconciler now lists only Cluster objects to find matched services. If it runs before the migrator has created the corresponding Cluster for a legacy LedgerService, matchedServices is empty and the garbage-collection path below deletes the existing replicated agent Secrets as orphans. On the next reconcile a new keypair is generated, unexpectedly rotating or removing credentials and causing the Cluster auth-keys ConfigMap to drop or change those keys mid-upgrade.

**Suggestion:** Treat matching legacy LedgerService objects as valid targets until migration converges (e.g., fall back to listing LedgerService when no Cluster match is found), or explicitly skip orphan GC while any LedgerService-to-Cluster migration is still pending.

misc/operator/internal/controller/ledgerservice_migrator.go
NumaryBot: 🟠 **[major] DNSEndpoint children not rehomed during migration**

When a legacy LedgerService had spec.dnsEndpoint.enabled set and the ExternalDNS CRD was installed, the reconcileDNSEndpoint path sets a controller ownerReference on the DNSEndpoint object. The migrator never lists or reparents this unstructured child resource. The new Cluster reconciler will fail SetControllerReference on the existing DNSEndpoint (already controlled by LedgerService), and deletion of the LedgerService will garbage-collect the live DNS record.

**Suggestion:** Extend the child adoption loop to also enumerate DNSEndpoint unstructured objects owned by the LedgerService (by GVR or ownerReference scan) and patch their controller ownerReference to point to the new Cluster.

misc/operator/internal/controller/ledgerservice_migrator.go
NumaryBot: 🟠 **[major] Instance-label selector misses children when additionalLabels overrides app.kubernetes.io/instance**

The migrator uses the app.kubernetes.io/instance label to select child resources for adoption. For LedgerService objects where spec.additionalLabels overrode this label, all children were stamped with the overridden value, so the selector matches nothing. The migrator then marks the source as migrated while all old ownerReferences remain in place, silently leaving the Cluster reconciler unable to adopt or manage those resources.

**Suggestion:** Instead of (or in addition to) the instance label selector, scan children by iterating over the LedgerService's ownerReference UID across relevant resource types, which is label-value agnostic and will correctly identify all owned objects regardless of label overrides.

misc/operator/internal/controller/ledgerbackuprun_controller.go
```diff
@@ -80,12 +80,12 @@ func (r *LedgerBackupRunReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return r.setRunFailed(ctx, &run, fmt.Sprintf("LedgerBackup %q not found: %v", run.Spec.BackupRef, err))
 	}
 
-	var ledgerService ledgerv1alpha1.LedgerService
+	var cluster ledgerv1alpha1.Cluster
```
NumaryBot, line 83: 🟠 **[major] In-flight backup runs permanently failed due to migration ordering race**

During an upgrade with an existing non-terminal LedgerBackupRun targeting a legacy LedgerService, there is no ordering guarantee that the migrator has already created the same-named Cluster before this reconcile runs. In that transient window the reconciler calls setRunFailed, permanently marking the run as failed even though the referenced workload may appear moments later. This turns a transient migration condition into a permanent terminal failure for in-flight or manual backup runs.

**Suggestion:** When the referenced Cluster is not found during active migration (i.e., a corresponding LedgerService still exists or migration is in progress), requeue with a backoff instead of marking the run as failed. Alternatively, fall back to resolving the legacy LedgerService resource as the target until the Cluster exists.

misc/operator/internal/controller/ledgerservice_migrator.go
```diff
@@ -0,0 +1,285 @@
...
+
+	if _, done := ls.Annotations[migratedToAnnotation]; done {
+		return ctrl.Result{}, nil
+	}
+
+	cluster := &ledgerv1alpha1.Cluster{}
+	err := r.Get(ctx, types.NamespacedName{Name: ls.Name, Namespace: ls.Namespace}, cluster)
+	switch {
+	case apierrors.IsNotFound(err):
+		cluster = &ledgerv1alpha1.Cluster{
```
NumaryBot: 🔴 **[blocker] Child resources retain LedgerService ownerReferences after migration**

The migrator creates a new Cluster object but never reparents the child resources (ServiceAccount, Services, StatefulSet, etc.) that were previously created and owned by the LedgerService. Those children retain a controller ownerReference pointing to the old LedgerService. When the Cluster reconciler subsequently calls SetControllerReference on them, Kubernetes rejects it because a second controller owner is not allowed. The new Cluster therefore cannot manage or update the existing deployment. Additionally, if the old LedgerService is later deleted, Kubernetes garbage-collects all its owned children, destroying the live workload.

**Suggestion:** During migration, enumerate all child resources owned by the LedgerService (by scanning ownerReferences using the LedgerService UID, not just label selectors) and patch their ownerReferences to replace the LedgerService controller reference with one pointing to the newly created Cluster before marking the source as migrated.

misc/operator/internal/controller/ledgerclusteragent_controller.go
```diff
@@ -207,27 +207,60 @@ func (r *LedgerClusterAgentReconciler) handleDeletion(ctx context.Context, agent
...
-// those matching the agent's label selector.
+// resolveMatchedServices lists Clusters (and, for the migration window,
+// legacy LedgerServices) across all namespaces and returns those matching
+// the agent's label selector. LedgerServices are included so that during
+// an in-place upgrade the agent does not see matched=∅ and treat every
+// existing replica Secret as an orphan to garbage-collect: that path
+// rotates keys mid-upgrade and breaks Cluster auth before the migrator
+// has had a chance to run. Once a legacy LS has been migrated it still
+// exists as a deprecated shadow but shares the Cluster's namespace+name,
+// so the resulting MatchedService list is de-duplicated.
```
NumaryBot: 🔴 **[blocker] Agent secrets GC'd as orphans when Cluster not yet created during migration**

The agent reconciler now lists only Cluster objects to find matched services. If it runs before the migrator has created the corresponding Cluster for a legacy LedgerService, matchedServices is empty and the garbage-collection path deletes the existing replicated agent Secrets as orphans. On the next reconcile a new keypair is generated, unexpectedly rotating or removing credentials and causing the Cluster auth-keys ConfigMap to drop or change those keys mid-upgrade.

**Suggestion:** Treat matching legacy LedgerService objects as valid targets until migration converges (e.g., fall back to listing LedgerService objects when no Cluster match is found), or explicitly skip orphan GC while any LedgerService-to-Cluster migration is still pending.

misc/operator/internal/controller/ledgerbackuprun_controller.go
```diff
@@ -80,12 +80,29 @@ func (r *LedgerBackupRunReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
```
NumaryBot, line 83: 🟠 **[major] In-flight backup runs permanently failed due to migration ordering race**

During an upgrade with an existing non-terminal LedgerBackupRun targeting a legacy LedgerService, there is no ordering guarantee that the migrator has already created the corresponding Cluster before this reconcile runs. In that transient window the reconciler calls setRunFailed, permanently marking the run as failed even though the referenced workload may appear moments later. This turns a transient migration condition into a permanent terminal failure for in-flight or manual backup runs.

**Suggestion:** When the referenced Cluster is not found during active migration (i.e., a corresponding LedgerService still exists or migration is in progress), requeue with a backoff instead of marking the run as failed. Alternatively, fall back to resolving the legacy LedgerService resource as the target until the Cluster exists.

misc/operator/internal/controller/ledgerbackup_controller.go
```diff
@@ -58,13 +58,13 @@ func (r *LedgerBackupReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil
 	}
 
-	// Validate that the referenced LedgerService exists.
-	var ledgerService ledgerv1alpha1.LedgerService
+	// Validate that the referenced Cluster exists.
+	var cluster ledgerv1alpha1.Cluster
 	if err := r.Get(ctx, types.NamespacedName{
```
NumaryBot, line 63: 🟠 **[major] Backup scheduler permanently fails scheduled backups during migration window**

The LedgerBackup scheduler treats a missing Cluster as a real configuration failure if it reconciles before the migrator creates the Cluster for a legacy LedgerService reference. Unlike LedgerBackupRun which has some migration-window handling, the scheduler has no equivalent fallback, so scheduled backups are marked failed due solely to controller ordering during the upgrade.

**Suggestion:** Mirror the same migration-window logic used in LedgerBackupRun: check whether a corresponding LedgerService still exists before treating a missing Cluster as a hard failure, and requeue with backoff if migration is still pending.

misc/operator/internal/controller/ledgerservice_migrator.go
```diff
@@ -0,0 +1,285 @@
...
+		ObservedGeneration: ls.Generation,
+	})
+	if err := r.Status().Update(ctx, ls); err != nil {
+		return ctrl.Result{}, fmt.Errorf("updating LedgerService status: %w", err)
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// rehomeChildren walks every child object kind the Cluster reconciler owns
```
NumaryBot: 🟠 **[major] DNSEndpoint children not rehomed during migration**

When a legacy LedgerService had spec.dnsEndpoint.enabled set and the ExternalDNS CRD was installed, the reconcileDNSEndpoint path sets a controller ownerReference on the DNSEndpoint object. The migrator never lists or reparents this unstructured child resource. The new Cluster reconciler will fail SetControllerReference on the existing DNSEndpoint (already controlled by LedgerService), and deletion of the LedgerService will garbage-collect the live DNS record.

**Suggestion:** Extend the child adoption loop to also enumerate DNSEndpoint unstructured objects owned by the LedgerService (by GVR or ownerReference scan) and patch their controller ownerReference to point to the new Cluster.

misc/operator/internal/controller/ledgerservice_migrator.go
```diff
@@ -0,0 +1,285 @@
...
+// and rewrites any controller ownerReference that still points at the
+// LedgerService so it points at the new Cluster instead. Without this,
+// controllerutil.SetControllerReference would fail in the Cluster reconciler
+// ("object already has a different controller") and deleting the source
+// LedgerService would garbage-collect the live workload via cascade delete.
```
NumaryBot: 🟠 **[major] Instance-label selector misses children when additionalLabels overrides app.kubernetes.io/instance**

The migrator uses the app.kubernetes.io/instance label to select child resources for adoption. For LedgerService objects where spec.additionalLabels overrode this label, all children were stamped with the overridden value, so the selector matches nothing. The migrator then marks the source as migrated while all old ownerReferences remain in place, silently leaving the Cluster reconciler unable to adopt or manage those resources.

**Suggestion:** Instead of (or in addition to) the instance label selector, scan children by iterating over the LedgerService's ownerReference UID across relevant resource types. This is label-value agnostic and will correctly identify all owned objects regardless of label overrides.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1481#issuecomment-4866361927

misc/operator/internal/controller/ledgerservice_migrator.go
NumaryBot: 🔴 **[blocker] Child resources retain LedgerService ownerReferences after migration**

The migrator creates a new Cluster object but never reparents the child resources (ServiceAccount, Services, StatefulSet, etc.) previously created and owned by the LedgerService. Those children retain a controller ownerReference pointing to the old LedgerService. When the Cluster reconciler subsequently calls SetControllerReference on them, Kubernetes rejects it because a second controller owner is not allowed, so the new Cluster cannot manage or update the existing deployment. Additionally, if the old LedgerService is later deleted, Kubernetes garbage-collects all its owned children, destroying the live workload.

**Suggestion:** During migration, enumerate all child resources owned by the LedgerService by scanning ownerReferences using the LedgerService UID (not just label selectors) and patch their ownerReferences to replace the LedgerService controller reference with one pointing to the newly created Cluster before marking the source as migrated.

misc/operator/internal/controller/ledgerclusteragent_controller.go
```diff
@@ -255,27 +255,60 @@ func (r *LedgerClusterAgentReconciler) handleDeletion(ctx context.Context, agent
...
```
NumaryBot: 🔴 **[blocker] Agent secrets GC'd as orphans when Cluster not yet created during migration**

The agent reconciler now lists only Cluster objects to find matched services. If it runs before the migrator has created the corresponding Cluster for a legacy LedgerService, matchedServices is empty and the garbage-collection path deletes the existing replicated agent Secrets as orphans. On the next reconcile a new keypair is generated, unexpectedly rotating or removing credentials and causing the Cluster auth-keys ConfigMap to drop or change those keys mid-upgrade.

**Suggestion:** Treat matching legacy LedgerService objects as valid targets until migration converges (e.g., fall back to listing LedgerService objects when no Cluster match is found), or explicitly skip orphan GC while any LedgerService-to-Cluster migration is still pending.

misc/operator/internal/controller/ledgerbackup_controller.go
NumaryBot, line 63: 🔴 **[blocker] Backup scheduler permanently fails scheduled backups during migration window**

The LedgerBackup scheduler treats a missing Cluster as a real configuration failure if it reconciles before the migrator creates the Cluster for a legacy LedgerService reference. Unlike LedgerBackupRun which has some migration-window handling, the scheduler has no equivalent fallback. setFailed is called and the controller returns without a requeue or any watch on Cluster creation, so scheduled backups remain Failed and scheduled runs stop until some unrelated LedgerBackup event occurs.

**Suggestion:** Mirror the same migration-window logic used in LedgerBackupRun: check whether a corresponding LedgerService still exists before treating a missing Cluster as a hard failure, and requeue with backoff if migration is still pending.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1481#issuecomment-4866361927

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1481#issuecomment-4866361927

misc/operator/internal/controller/ledgerservice_migrator.go
NumaryBot: 🔴 **[blocker] Child resources retain LedgerService ownerReferences after migration**

The migrator creates a new Cluster object but never reparents child resources (ServiceAccount, Services, StatefulSet, etc.) previously created and owned by the LedgerService. Those children retain a controller ownerReference pointing to the old LedgerService. When the Cluster reconciler subsequently calls SetControllerReference on them, Kubernetes rejects it because a second controller owner is not allowed, so the new Cluster cannot manage or update the existing deployment. Additionally, if the old LedgerService is later deleted, Kubernetes garbage-collects all its owned children, destroying the live workload.

**Suggestion:** During migration, enumerate all child resources owned by the LedgerService by scanning ownerReferences using the LedgerService UID (not just label selectors) and patch their ownerReferences to replace the LedgerService controller reference with one pointing to the newly created Cluster before marking the source as migrated.

misc/operator/internal/controller/ledgerbackup_controller.go
NumaryBot, line 63: 🔴 **[blocker] Backup scheduler permanently fails scheduled backups during migration window**

The LedgerBackup scheduler treats a missing Cluster as a real configuration failure if it reconciles before the migrator creates the Cluster for a legacy LedgerService reference. setFailed is called and the controller returns without a requeue or any watch on Cluster creation, so scheduled backups remain permanently Failed and scheduled runs stop until some unrelated LedgerBackup event occurs.

**Suggestion:** Mirror the same migration-window logic used in LedgerBackupRun: check whether a corresponding LedgerService still exists before treating a missing Cluster as a hard failure, and requeue with backoff if migration is still pending.

misc/operator/internal/controller/suite_test.go
```diff
@@ -143,15 +143,15 @@ func createTestNamespace(t *testing.T) string {
 	return name
 }
 
-// newLedgerService returns a minimal valid LedgerService CR.
-func newLedgerService(name, namespace string) *ledgerv1alpha1.LedgerService {
+// newCluster returns a minimal valid Cluster CR.
```
NumaryBot, line 146: 🔴 **[blocker] Unresolved `newLedgerService` helper causes test package compilation failure**

Renaming the shared helper removes `newLedgerService`, but `reconcile_agent_test.go` still calls it in the legacy-agent tests, so the operator test package no longer compiles with `undefined: newLedgerService` at those call sites.

**Suggestion:** Either keep a LedgerService-specific helper for the compat tests, or update those tests to construct the intended resource directly.

misc/operator/internal/controller/ledgerservice_migrator.go
NumaryBot: 🟠 **[major] DNSEndpoint children not rehomed during migration**

When a legacy LedgerService had `spec.dnsEndpoint.enabled` set and the ExternalDNS CRD was installed, the reconcileDNSEndpoint path sets a controller ownerReference on the DNSEndpoint object. The migrator never lists or reparents this unstructured child resource. The new Cluster reconciler will fail SetControllerReference on the existing DNSEndpoint (already controlled by LedgerService), and deletion of the LedgerService will garbage-collect the live DNS record.

**Suggestion:** Extend the child adoption loop to also enumerate DNSEndpoint unstructured objects owned by the LedgerService (by GVR or ownerReference scan) and patch their controller ownerReference to point to the new Cluster.

misc/operator/internal/controller/ledgerservice_migrator.go
NumaryBot: 🟠 **[major] Instance-label selector misses children when additionalLabels overrides app.kubernetes.io/instance**

The migrator uses the `app.kubernetes.io/instance` label to select child resources for adoption. For LedgerService objects where `spec.additionalLabels` overrode this label, all children were stamped with the overridden value, so the selector matches nothing. The migrator then marks the source as migrated while all old ownerReferences remain in place, silently leaving the Cluster reconciler unable to adopt or manage those resources.

**Suggestion:** Instead of (or in addition to) the instance label selector, scan children by iterating over the LedgerService's ownerReference UID across relevant resource types. This is label-value agnostic and will correctly identify all owned objects regardless of label overrides.

misc/operator/internal/controller/ledgerclusteragent_controller.go
```diff
@@ -255,27 +255,60 @@ func (r *LedgerClusterAgentReconciler) handleDeletion(ctx context.Context, agent
...
+
+	var legacy ledgerv1alpha1.LedgerServiceList
+	if err := r.List(ctx, &legacy, &client.ListOptions{LabelSelector: selector}); err != nil {
+		return nil, fmt.Errorf("listing LedgerServices: %w", err)
+	}
+	for i := range legacy.Items {
+		if !selector.Matches(labels.Set(legacy.Items[i].Labels)) {
+			continue
+		}
+		key := legacy.Items[i].Namespace + "/" + legacy.Items[i].Name
```
NumaryBot: 🟠 **[major] Legacy LedgerService labels keep agent targeting Cluster after selector opt-out**

Because the Cluster list is filtered by the agent selector, a migrated same-named Cluster that no longer matches the selector is absent from `seen`. If the deprecated LedgerService still has matching labels, this loop re-adds the namespace/name as a target. In that post-migration label-change scenario the agent keeps distributing replica Secrets and reporting a match even though the actual Cluster opted out, until the old LedgerService is manually removed.

**Suggestion:** Once a Cluster exists for a given namespace/name, ignore the corresponding legacy LedgerService match entirely — do not fall back to LedgerService labels if the Cluster has already been created but simply doesn't match the current selector.

misc/operator/internal/controller/ledgerbackup_controller.go
NumaryBot, line 63: 🟠 **[major] Backup scheduler permanently fails scheduled backups during migration window**

The LedgerBackup scheduler treats a missing Cluster as a real configuration failure if it reconciles before the migrator creates the Cluster for a legacy LedgerService reference. setFailed is called and the controller returns without a requeue or any watch on Cluster creation, so scheduled backups remain Failed and scheduled runs stop until some unrelated LedgerBackup event occurs.

**Suggestion:** Mirror the same migration-window logic used in LedgerBackupRun: check whether a corresponding LedgerService still exists before treating a missing Cluster as a hard failure, and requeue with backoff if migration is still pending.

misc/operator/api/v1alpha1/credentials_types.go
```diff
@@ -91,15 +91,15 @@ type SecretReference struct {
...
-	// Namespace of the LedgerService.
+	// Namespace of the Cluster.
 	Namespace string `json:"namespace"`
 
-	// Name of the LedgerService.
+	// Name of the Cluster.
 	Name string `json:"name"`
 }
 
 func init() {
```
NumaryBot, line 103: 🟠 **[major] LedgerClusterAgent GVK removed without compatibility path — finalizers and reconciliation abandoned**

For clusters with existing LedgerClusterAgent resources, only the new Credentials GVK is registered, so the reconciler no longer processes the old cluster-scoped agents or their ledger.formance.com/agent-keys finalizers. Replicated key Secrets can become stale, and deletes/uninstalls can leave old agent resources stuck because no controller removes the finalizer.

**Suggestion:** Introduce a compatibility registration or migration step for the old LedgerClusterAgent type that removes finalizers and migrates state before the legacy GVK is fully dropped.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1481#issuecomment-4866361927

misc/operator/internal/controller/ledgerbackup_controller.go
NumaryBot, line 63: 🔴 **[blocker] Backup scheduler permanently fails scheduled backups during migration window**

The LedgerBackup scheduler treats a missing Cluster as a real configuration failure if it reconciles before the migrator creates the Cluster for a legacy LedgerService reference. There is no fallback or requeue on Cluster creation, so scheduled backups are permanently marked Failed and no further runs are scheduled until some unrelated LedgerBackup event triggers re-reconciliation.

**Suggestion:** Mirror the same migration-window logic used in LedgerBackupRun: check whether a corresponding LedgerService still exists before treating a missing Cluster as a hard failure, and requeue with backoff if migration is still pending.

misc/operator/internal/controller/credentials_controller.go
```diff
@@ -181,121 +181,122 @@ func (r *LedgerClusterAgentReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+// listCredentialsReplicaSecrets returns every replica Secret belonging to the given
+// credentials, across all namespaces. The canonical Secret is identified by
 // name + namespace and excluded from the result — filtering by name (rather
 // than by an additional label) means Secrets created by pre-canonical
 // versions of the operator are still discovered, so upgrade adopts them
 // instead of orphaning them.
-func (r *LedgerClusterAgentReconciler) listAgentReplicaSecrets(ctx context.Context, agentName string) ([]corev1.Secret, error) {
+func (r *CredentialsReconciler) listCredentialsReplicaSecrets(ctx context.Context, credentialsName string) ([]corev1.Secret, error) {
 	var secrets corev1.SecretList
 	if err := r.List(ctx, &secrets, client.MatchingLabels{
```
NumaryBot, line 293: 🔴 **[blocker] Legacy agent secrets not adopted when renaming Credentials — keys rotated on upgrade**

When upgrading a same-named LedgerClusterAgent to Credentials, existing replica/canonical Secrets are labelled ledger.formance.com/agent-name but the new list query only selects ledger.formance.com/credentials-name. As a result existingReplicas is empty, ensureCanonicalSecret generates a new keypair, and existing client bundles are invalidated, causing unexpected auth-key rotation instead of preserving identity across the rename.

**Suggestion:** During the first reconcile of a Credentials object, also query for Secrets with the legacy ledger.formance.com/agent-name label matching the same name, and adopt/relabel them before falling through to keypair generation.

misc/operator/internal/controller/cluster_controller.go
```diff
@@ -297,22 +297,22 @@ func (r *LedgerServiceReconciler) deletionProtectionPolicyInstalled(ctx context.
 }
 
 // SetupWithManager sets up the controller with the Manager.
-func (r *LedgerServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
+func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
```
NumaryBot, line 301: 🔴 **[blocker] Legacy LedgerService GVK not watched — existing workloads lose operator management on upgrade**

The controller setup now watches only the new Cluster GVK. For clusters that already have LedgerService CRs, Kubernetes will not convert or enqueue those objects, so no same-named Cluster is created or adopted and the migrator never runs for them. Those workloads effectively lose operator management; later deletion of the old CR can still trigger Kubernetes garbage collection of the live workload.

**Suggestion:** Register a watch on the legacy LedgerService GVK (or keep the old scheme registration) so that existing LedgerService objects are enqueued and the migrator can process them before they lose reconciliation coverage.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1481#issuecomment-4866361927

misc/operator/internal/controller/ledgerbackup_controller.go
NumaryBot, line 63: 🔴 **[blocker] Backup scheduler permanently fails scheduled backups during migration window**

The LedgerBackup scheduler treats a missing Cluster as a real configuration failure if it reconciles before the migrator creates the Cluster for a legacy LedgerService reference. There is no equivalent fallback or migration-window guard, so setFailed is called and the controller returns without a requeue or any watch on Cluster creation. Scheduled backups remain Failed and scheduled runs stop until some unrelated LedgerBackup event occurs.

**Suggestion:** Mirror the same migration-window logic used in LedgerBackupRun: check whether a corresponding LedgerService still exists before treating a missing Cluster as a hard failure, and requeue with backoff if migration is still pending.

misc/operator/internal/controller/credentials_controller.go
NumaryBot, line 293: 🔴 **[blocker] Adopt legacy agent secrets before generating new credentials to avoid key rotation**

When a LedgerClusterAgent is migrated to a same-named Credentials, the existing seed replicas from the previous release are labeled with the old label key (e.g., ledger.formance.com/agent-name) and carry legacy naming patterns, but the credentials controller only lists secrets using the new credentialsNameLabel before bootstrapping the canonical secret. The old seed is invisible to this query, so ensureCanonicalSecret mints a fresh keypair, invalidating all previously issued auth bundles and breaking in-flight connections during upgrade.

**Suggestion:** Before generating a new keypair, also query for secrets carrying the legacy agent label key to detect and adopt any existing seed secret. If found, use it as the canonical secret instead of minting a new one, ensuring credentials continuity across the rename.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1481#issuecomment-4866361927

Azorlogh (approved):
Lgtm (note: PR description still mentions the deleted LedgerServiceMigrator)

PR #1482, opened 2026-07-02
fix(EN-1441): propagate initIndexConfig boot errors with self-healing retry

paul-nicolas:
## Summary

`Builder.initIndexConfig` returned `void` and logged-and-returned on every boot error branch, so `Builder.loop` could not observe the failure and unconditionally advanced to `processLogs`. A transient Pebble/read-store failure at boot left the in-memory index config empty or partially populated — the write path then silently skipped index writes (byCanonical miss ⇒ `isMetadataIndexed`/`isBuiltinIndexed` return `false`) **while the persisted cursor kept advancing**, leaving the read-side index permanently incomplete with no error, no retry, and no crash.

## Fix

- `initIndexConfig(ctx) error` (was `void`): the four boot read branches now return a wrapped error instead of a swallowed log. A top-of-function reset of `indexConfig`/`bucketIndexConfig`/`backfillTasks`/`schemaRewriteTasks`/`indexVersions` makes it idempotent across retries (the two slices would otherwise double-schedule).
- Extracted the boot prologue into `bootInit(ctx) (cursor, pebbleLast, err)` and wrapped it in the existing `worker.RetryWithBackoff` (100 ms → 10 s, honors the stop channel). A `ctx.Err()` check after the retry distinguishes "init succeeded" from "shutdown requested" so `loop` never processes logs against a failed init.
- This also closes the identical silent-return defects at the sibling `LastIndexedSequence` and `NewDirectReadHandle` boot reads. `ReadAppliedProposalProgress` / `ReadLastSequence` keep their existing best-effort semantics.

A transient boot read failure now retries with backoff instead of silently advancing the cursor against an incomplete config.

## Tests

- `TestInitIndexConfig_PropagatesReadError` — a closed store makes the boot read fail; the error propagates instead of being swallowed.
- `TestInitIndexConfig_IdempotentAcrossRetries` — calling `initIndexConfig` twice does not double-schedule backfill/rewrite tasks.
- `TestBootInit_PropagatesError` / `TestBootInit_Success` — cover the extracted prologue.
- `worker.RetryWithBackoff` mechanics are already covered by `internal/pkg/worker/worker_test.go`.

`go build ./...`, `just pre-commit` (0 lint issues), and `go test ./internal/application/indexbuilder/...` all pass.

## Ticket

https://formance-team.atlassian.net/browse/EN-1441

Same silent-swallow class as EN-1440.

### DIFF
--- a/internal/application/indexbuilder/builder.go
+++ b/internal/application/indexbuilder/builder.go
@@ -2,6 +2,7 @@ package indexbuilder
 
 import (
 	"context"
+	"fmt"
 	"sync/atomic"
 	"time"
 
@@ -440,43 +441,32 @@ func (b *Builder) loop(ctx context.Context) {
 	// (same signal, same semantics).
 	stop := ctx.Done()
 
-	// Initialize index config cache from Pebble before processing any logs.
-	b.initIndexConfig(ctx)
+	// Boot init: rebuild the index-config cache and seed cursors. Any
+	// transient Pebble/read-store failure here must NOT advance the
+	// persisted cursor against an incomplete config, so retry with
+	// backoff until it succeeds or shutdown is requested. initIndexConfig
+	// resets its own state, so re-running it on retry is idempotent.
+	var (
+		cursor     uint64
+		pebbleLast uint64
+		err        error
+	)
+	worker.RetryWithBackoff(stop, b.logger, func() error {
+		cursor, pebbleLast, err = b.bootInit(ctx)
 
-	cursor, err := b.readStore.LastIndexedSequence()
-	if err != nil {
-		b.logger.Errorf("Failed to read progress: %v", err)
+		return err
+	})
 
+	// RetryWithBackoff returns only on success or when stop is closed. If
+	// the context was cancelled we never got a good init — return without
+	// processing any log.
+	if ctx.Err() != nil {
 		return
 	}
 
 	b.lastIndexedSeq.Store(cursor)
+	b.pebbleLastSeq.Store(pebbleLast)
 
-	// Recover AppliedProposal sync progress. A corrupt cursor surfaces loudly
-	// and falls back to a full re-sync from 0 (the projection is rebuildable),
-	// rather than aborting the whole builder.
-	if seq, err := b.readStore.ReadAppliedProposalProgress(); err != nil {
-		b.logger.Errorf("read applied-proposal cursor (re-syncing from 0): %v", err)
-	} else {
-		b.lastAppliedProposalSeq = seq
-	}
-
-	// Seed pebble last sequence. The handle is closed immediately after use
-	// to release the RLock — keeping it open would deadlock with
-	// RestoreCheckpoint (write lock) when processLogs tries to take a new RLock.
-	var pebbleLast uint64
-	if handle, err := b.pebbleStore.NewDirectReadHandle(); err != nil {
-		b.logger.Errorf("Failed to create read handle: %v", err)
-
-		return
-	} else {
-		if v, err := query.ReadLastSequence(handle); err == nil {
-			pebbleLast = v
-			b.pebbleLastSeq.Store(v)
-		}
-
-		_ = handle.Close()
-	}
 	b.logger.WithFields(map[string]any{
 		"cursor":     cursor,
 		"pebbleLast": pebbleLast,
@@ -587,4 +577,48 @@ func (b *Builder) loop(ctx context.Context) {
 		// the broadcast and block until new logs arrive.
 		b.readStore.NotifyProgress()
 	}
+}
+
+// bootInit runs the index builder's boot prologue as a single retryable unit:
+// rebuild the index-config cache from Pebble/the read store, then read the
+// persisted indexed cursor and seed the last-known Pebble sequence. It returns
+// the recovered cursor and pebbleLast, or an error if any required read failed
+// — the caller (loop) retries with backoff so a transient failure never
+// advances the cursor against an incomplete config. ReadAppliedProposalProgress
+// and query.ReadLastSequence stay best-effort (they tolerate failure today);
+// only initIndexConfig, LastIndexedSequence, and NewDirectReadHandle are fatal.
+func (b *Builder) bootInit(ctx context.Context) (cursor uint64, pebbleLast uint64, err error) {
+	if err := b.initIndexConfig(ctx); err != nil {
+		return 0, 0, fmt.Errorf("initializing index config: %w", err)
+	}
+
+	cursor, err = b.readStore.LastIndexedSequence()
+	if err != nil {
+		return 0, 0, fmt.Errorf("reading last indexed sequence: %w", err)
+	}
+
+	// Recover AppliedProposal sync progress (best-effort: a corrupt cursor
+	// surfaces loudly and falls back to a full re-sync from 0, since the
+	// projection is rebuildable, rather than aborting the whole builder).
+	if seq, err := b.readStore.ReadAppliedProposalProgress(); err != nil {
+		b.logger.Errorf("read applied-proposal cursor (re-syncing from 0): %v", err)
+	} else {
+		b.lastAppliedProposalSeq = seq
+	}
+
+	// Seed pebble last sequence. The handle is closed immediately after use
+	// to release the RLock — keeping it open would deadlock with
+	// RestoreCheckpoint (write lock) when processLogs tries to take a new RLock.
+	handle, err := b.pebbleStore.NewDirectReadHandle()
+	if err != nil {
+		return 0, 0, fmt.Errorf("creating read handle: %w", err)
+	}
+
+	if v, err := query.ReadLastSequence(handle); err == nil {
+		pebbleLast = v
+	}
+
+	_ = handle.Close()
+
+	return cursor, pebbleLast, nil
 }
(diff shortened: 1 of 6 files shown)


internal/application/indexbuilder/builder.go
```diff
@@ -456,39 +457,32 @@ func (b *Builder) loop(ctx context.Context) {
...
+	// transient Pebble/read-store failure here must NOT advance the
+	// persisted cursor against an incomplete config, so retry with
+	// backoff until it succeeds or shutdown is requested. initIndexConfig
+	// resets its own state, so re-running it on retry is idempotent.
+	var (
+		cursor     uint64
+		pebbleLast uint64
+		err        error
+	)
+	worker.RetryWithBackoff(stop, b.logger, func() error {
```
paul-nicolas, line 454: [Medium] Permanent boot-read failure retries forever with no escalation or health signal

`worker.RetryWithBackoff` (`internal/pkg/worker/retry.go`) has no attempt bound and returns only on `fn()==nil` or `stop` close; `stop == ctx.Done()`, which is cancelled only by `Stop()`. So a *non-transient* boot failure (Pebble corruption, a permanently unreadable `IndexVersionState`, etc.) leaves the builder retrying every ≤10s indefinitely: the read-side index is never built, queries serve stale/empty results, and `last_indexed_sequence` / `pebble_last_sequence` stay at 0 — the only signal is a repeated `Errorf`.

This is a strict improvement over the pre-fix silent cursor-advance, but it departs from the subsystem's fail-loud posture (`indexer.md:30`: “an indexer panic is a non-recoverable invariant violation”). Consider bounding retries before escalating (panic / failing the healthcheck) or exposing a distinct “boot-stuck” metric, so a permanent failure is observable rather than a silent indefinite degradation.

internal/application/indexbuilder/builder.go
```diff
@@ -600,3 +594,44 @@ func (b *Builder) loop(ctx context.Context) {
...
+
+// bootInit runs the index builder's boot prologue as a single retryable unit:
+// rebuild the index-config cache from Pebble/the read store, then read the
+// persisted indexed cursor and seed the last-known Pebble sequence. It returns
+// the recovered cursor and pebbleLast, or an error if any required read failed
+// — the caller (loop) retries with backoff so a transient failure never
+// advances the cursor against an incomplete config. ReadAppliedProposalProgress
+// and query.ReadLastSequence stay best-effort (they tolerate failure today);
+// only initIndexConfig, LastIndexedSequence, and NewDirectReadHandle are fatal.
+func (b *Builder) bootInit(ctx context.Context) (cursor uint64, pebbleLast uint64, err error) {
```
paul-nicolas, line 590: [Medium] Documentation drift — indexer.md “Boot” section not updated for this new mechanism

`docs/technical/architecture/subsystems/indexer/indexer.md` (§Boot, lines 32-38) still describes boot as a linear “read cursors → read IndexVersionState → catch-up” sequence, with no mention that these reads are now bundled in `bootInit` and wrapped in `worker.RetryWithBackoff` (retry on transient failure until success or shutdown). Line 30 (“no panic recovery — an indexer panic is a non-recoverable invariant violation”) should also be reconciled, since boot *read* failures now retry indefinitely rather than proceeding/terminating.

CLAUDE.md “Documentation Maintenance” requires documenting new technical mechanisms. `indexer.md` is not in this diff, so I'm flagging it here on the code that introduces the mechanism it documents.

paul-nicolas (commented):

**Models:** Claude (this repo's perspective) + Codex (`codex-cli` via `codex exec` — the `omc ask codex` wrapper was broken, so I invoked Codex directly). No Gemini.

**Verdict:** The core change is sound. `initIndexConfig` now propagates boot-read errors instead of swallowing them, `bootInit` bundles the boot prologue as one retryable unit, and `RetryWithBackoff` + the `ctx.Err()` post-check correctly distinguish success from shutdown (`stop == ctx.Done()`, so the check is valid). The idempotency reset covers every builder-local field the init path mutates. No critical or major issue survived verification.

**Severity counts:** 0 critical, 0 major, 2 medium (posted inline), 5 minor/nit (withheld, listed below).

**Verification / conflict resolved:** Codex rated the `bootInit` assignment of `b.lastAppliedProposalSeq` (builder.go:618) a **major** “stale-value reuse” bug. I refuted it: the persisted applied-proposal progress is written by exactly one caller — `process_logs.go:80`, inside the processing loop, which never runs during boot or retries. The value is therefore stable across boot retries, so keeping attempt-1's value when a later read transiently fails yields the *correct* value; Codex's suggested “reset to 0 on error” fix would be strictly worse (a full re-scan). Downgraded to a minor cleanliness note.

---

- `internal/application/indexbuilder/builder.go:618` — `bootInit` assigns `b.lastAppliedProposalSeq` inside the retryable unit. Harmless (the value is a stable persisted read — see conflict note above — so it never actually goes stale), but mutating shared instance state mid-unit is slightly untidy; a local + assign-on-success alongside `lastIndexedSeq`/`pebbleLastSeq` would read cleaner.
- `internal/application/indexbuilder/builder.go:630` — swallowed `query.ReadLastSequence` error is asymmetric with the now-fatal `NewDirectReadHandle` on the same handle. Low-impact (feeds only the lag gauge/log), but a comment distinguishing “empty store” (expected) from a real read error would align with invariant #7.
- `internal/application/indexbuilder/index_config.go:124` — `purgeOrphanVersions` error is still swallowed (logged, returns nil), so this partial-failure class is not covered by the new retry. Acceptable and pre-existing (the index config is already fully built at that point and the sweep re-runs next boot), but worth a one-line note that it is intentionally best-effort.
- `internal/application/indexbuilder/index_config_test.go:319` — the `schemaRewriteTasks` idempotency assertion is vacuous (`firstRewrites == 0`), so the reset's stated double-schedule motivation is never actually exercised. Add a pending-rewrite `IndexVersionState` case (as `TestInitIndexConfig_ResumesRewriteFromPendingVersion` does) and assert the count stays 1 across two calls.
- `internal/application/indexbuilder/builder_boot_test.go:13` — `TestBootInit_PropagatesError` only exercises `initIndexConfig`'s first read (closing the store fails at the very first handle, identical to `TestInitIndexConfig_PropagatesReadError`). `bootInit`'s own `LastIndexedSequence`/`NewDirectReadHandle` branches, `loop()`'s `ctx.Err()` early-return, and a retry-then-succeed transition are all untested — they need a seam that fails *after* `initIndexConfig` succeeds.

internal/application/indexbuilder/builder.go
paul-nicolas: Fixed in cece8484f — rewrote the indexer.md §Boot section to describe the `bootInit` retryable-unit prologue (idempotent `initIndexConfig` reset, fatal vs best-effort boot reads, `worker.RetryWithBackoff` 100ms→10s, and the `ctx.Err()` shutdown-vs-success gate), and reconciled the line-30 fail-loud note by clarifying that a boot *read* failure is retried rather than a panic.

internal/application/indexbuilder/builder.go
paul-nicolas: Intentional for EN-1441: the whole point is to *survive* transient Pebble/read-store failures at boot rather than advancing the cursor against an incomplete config, so retry-forever (with a per-attempt error log) is the desired self-healing behavior. Panicking on a slow/transient dependency would crash-loop the node — the exact failure class we're riding out. A dedicated "boot-stuck" metric / log escalation after N attempts is a worthwhile observability follow-up but is out of scope for this fix; tracking separately. Resolving.

gfyrag:
Follow-up (hors scope de ce fix) : `bootInit` ouvre 2 handles séquentiels sur `pebbleStore` — un dans `initIndexConfig` (pour `seedLedgerIndexConfig` + `loadIndexRegistry`), puis un second juste après pour `query.ReadLastSequence`. C'est l'état pré-existant, la PR l'a déplacé sans consolider.

Il y aurait un gain de cohérence à faire du boot une lecture depuis un unique handle :

```go
func (b *Builder) bootInit(ctx context.Context) (uint64, uint64, error) {
    handle, err := b.pebbleStore.NewDirectReadHandle()
    if err != nil {
        return 0, 0, fmt.Errorf("creating read handle: %w", err)
    }
    defer func() { _ = handle.Close() }()

    if err := b.initIndexConfig(ctx, handle); err != nil {
        return 0, 0, fmt.Errorf("initializing index config: %w", err)
    }
    // ... reste avec le même handle
}
```

Bénéfices : moins de lock churn (chaque `NewDirectReadHandle` prend un RLock), alignement avec la sémantique "bootInit = unité atomique retryable", tout le boot lit d'un snapshot logique unique.

Coût : `initIndexConfig` perd son autonomie de handle (impact sur les tests directs), et le RLock est tenu plus longtemps — mais on est en boot avant `processLogs`, donc pas de risque de deadlock avec `RestoreCheckpoint` (le commentaire dans le code cible bien le cas `processLogs` en cours).

À laisser en follow-up séparé pour ne pas mélanger fix (error propagation + retry) et refactor (handle consolidation) — blame trails distincts.

gfyrag (approved):
LGTM. Fix root-cause propre pour la classe silent-swallow au boot (invariant #7 du CLAUDE.md), reset défensif pour l'idempotence des retries bien justifié, split fatal/best-effort clarifié, tests ciblés sur les deux invariants nouveaux et docs à jour. Consolidation des 2 handles laissée en follow-up (commentaire séparé).

PR #1483, opened 2026-07-02
fix(ledgerctl): make auth login --profile work end-to-end

gfyrag:
## Summary

Three linked bugs made `ledgerctl auth login --profile <name>` misbehave. All three are fixed together because they share the same flag-resolution plumbing.

- **Bootstrap missing profile.** `--profile <new>` slipped through the bootstrap allowance in `PersistentPreRunE` but no code path ever persisted the profile. `syncProfile` (new) now creates it from the current connection flags on the first login, and marks it active if it's the first profile.
- **`Changed(\"server\")` is now reliable.** `resolveFlag` used to `Flags().Set(...)` when it applied an env- or profile-derived value, which lit the `Changed` bit and made `runLogin` think the user had typed `--server`. A `LEDGERCTL_SERVER` diverging from the profile silently rewrote the profile. `resolveFlag` now goes through `Flag.Value.Set(...)` and leaves `Changed` alone; auth login only rewrites the profile when `--server` was actually typed.
- **`profile.signingKeyId` feeds `--key-id`.** The profile field was only wired to the persistent `--signing-key-id`, not to auth login's local `--key-id`. `PersistentPreRunE` now maps it into `--key-id` too, and `resolveLoginParams` reads flag values unconditionally so profile-derived signing settings are honored (the `Changed` check is kept solely to gate bundle-vs-CLI overrides).

## Test plan

- [x] `go build ./...`
- [x] `go vet ./cmd/ledgerctl/...`
- [x] `golangci-lint run ./cmd/ledgerctl/...` — 0 issues
- [x] `go test ./cmd/ledgerctl/...`
- [x] New tests
  - `TestSyncProfile_*` — bootstrap / explicit-update / env-only-no-rewrite / no-profile-no-op / dangling-activeProfile-not-resurrected
  - `TestServerFlagEnvResolution` — extended with `Changed=false` invariant for env-derived `--server`
  - `TestProfileSigningKeyIDFeedsKeyID` — asserts the new profile → `--key-id` fallback and that it keeps `Changed=false`
- [ ] Manual smoke: `ledgerctl auth login --profile new --server host:8888 --signing-key key.hex --key-id k --subject svc && ledgerctl --profile new ledgers list`


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

cmd/ledgerctl/auth/login.go
```diff
@@ -84,31 +79,81 @@ func runLogin(cmd *cobra.Command, _ []string) error {
...
+//   - If the profile already exists and --server was explicitly passed on
+//     the CLI, update the profile's server address so subsequent commands
+//     look up the keychain under the address we just stored the token under.
+//   - Otherwise do nothing.
+//
+// Changed("server") is trusted to mean "user typed --server on the CLI":
+// resolveFlag in main.go applies env/profile values through Flag.Value.Set,
+// which does not touch the Changed bit.
+func syncProfile(cmd *cobra.Command, server string) error {
+	profileName, profileExplicit := cmdutil.ResolveProfileName(cmd)
```
NumaryBot, line 133: 🟠 **[major] Sync the active profile when --profile is omitted**

When an active profile already exists and the user runs `ledgerctl auth login --server new...` without `--profile`, this early return skips the profile update entirely. The token is stored under the new server address, but subsequent commands still resolve the active profile's old server and therefore look up the wrong keychain entry; the previous implementation updated `cfg.ActiveProfile` in this scenario.

cmd/ledgerctl/auth/login.go
```diff
@@ -84,31 +79,81 @@ func runLogin(cmd *cobra.Command, _ []string) error {
...
+		// under us; do not silently recreate it.
 		return nil
 	}
 
-	profile.Server = server
-	cfg.Profiles[name] = *profile
+	insecure, _ := cmd.Flags().GetBool("insecure")
+	tlsCaCert, _ := cmd.Flags().GetString("tls-ca-cert")
+	signingKey, _ := cmd.Flags().GetString("signing-key")
+	signingKeyID, _ := cmd.Flags().GetString("signing-key-id")
```
NumaryBot: 🟠 **[major] Persist --key-id when bootstrapping the profile**

For the documented bootstrap flow `auth login --profile prod ... --key-id prod-key`, the local `--key-id` value is never copied into `profile.signingKeyId`; only `--signing-key-id` is read here. The new profile is saved without a signing key ID, so a later `auth login --profile prod` cannot use the intended profile-derived `--key-id` fallback and fails with `required flag "key-id" not set` unless the user supplied the extra persistent flag during bootstrap.

cmd/ledgerctl/main.go
```diff
@@ -108,6 +100,12 @@ func newRootCommand() *cobra.Command {
 			resolveFlag(cmd, "response-verify-key", "LEDGERCTL_RESPONSE_VERIFY_KEY", cmdutil.ProfileFlagValue(p, "response-verify-key"))
 			resolveFlag(cmd, "result-file", "LEDGERCTL_RESULT_FILE", "")
 
+			// Feed the profile's signingKeyId into the local --key-id flag
+			// exposed by auth login / auth generate-token: in practice the JWT
+			// key ID and the request-signing key ID are the same key entry.
+			// No-op on commands that do not declare --key-id.
+			resolveFlag(cmd, "key-id", "", cmdutil.ProfileFlagValue(p, "signing-key-id"))
```
NumaryBot: 🔴 **[blocker] Global --key-id profile fallback applies to all subcommands, including destructive signing commands**

The new global fallback that populates `--key-id` from the active profile's `signingKeyId` runs for every subcommand that declares `--key-id`, not just `auth login` / `auth generate-token`. Commands like `signing revoke-key` and `signing register-key` also declare `--key-id`; omitting it used to produce an explicit error, but now they silently inherit the profile key ID. This means `ledgerctl signing revoke-key` (with no `--key-id`) can silently revoke the active profile's current signing key rather than aborting.

**Suggestion:** Gate the signingKeyId profile fallback so it only applies to auth-related commands (e.g. `auth login`, `auth generate-token`). Signing management commands should continue to require an explicit `--key-id`.

cmd/ledgerctl/auth/login.go
```diff
@@ -84,31 +79,100 @@ func runLogin(cmd *cobra.Command, _ []string) error {
...
+//   - If the profile already exists and --server was explicitly passed on
+//     the CLI, update the profile's server address so subsequent commands
+//     look up the keychain under the address we just stored the token under.
+//   - Otherwise do nothing.
+//
+// Changed("server") is trusted to mean "user typed --server on the CLI":
+// resolveFlag in main.go applies env/profile values through Flag.Value.Set,
+// which does not touch the Changed bit.
+func syncProfile(cmd *cobra.Command, server string) error {
+	profileName, profileExplicit := cmdutil.ResolveProfileName(cmd)
```
NumaryBot, line 133: 🟠 **[major] Active profile not updated when --profile is omitted**

When an active profile already exists and the user runs `ledgerctl auth login --server <new>` without `--profile`, the early return skips updating `cfg.ActiveProfile` with the new server address. The token is stored under the new server address, but subsequent commands still resolve the active profile's old server and look up the wrong keychain entry. The previous implementation updated `cfg.ActiveProfile` in this scenario.

**Suggestion:** Before returning early, update `cfg.ActiveProfile` (and persist the config) so that the active profile's server field reflects the newly provided server address.

cmd/ledgerctl/auth/login.go
```diff
@@ -84,31 +79,100 @@ func runLogin(cmd *cobra.Command, _ []string) error {
...
+
+		existing.Server = server
+		cfg.Profiles[profileName] = existing
 
-	name, profile := cmdutil.GetActiveProfile(cfg, profileName)
-	if profile == nil || profile.Server == server {
+		return cmdutil.SaveConfig(cfg)
+	}
+
+	if !profileExplicit {
```
NumaryBot, line 167: 🟠 **[major] --key-id value not persisted when bootstrapping a new profile**

For the documented bootstrap flow `auth login --profile prod ... --key-id prod-key`, the local `--key-id` flag value is never copied into `profile.signingKeyId`. Only `--signing-key-id` is read at that point, so the new profile is saved without a signing key ID. A subsequent `auth login --profile prod` cannot use the intended profile-derived key-ID fallback and fails with `required flag "key-id" not set` unless the user re-supplies the flag on every invocation.

**Suggestion:** After resolving the `--key-id` value (whether from flag or profile fallback), write it back into `profile.signingKeyId` before saving the profile.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

gfyrag:
Follow-up on the NumaryBot findings from commit \`a965cbe9\`:

- ✅ **Finding 3 (`main.go:107`, --key-id leak to signing commands)** — real. Fixed in `5384eec2` by gating the profile.signingKeyId → --key-id fallback with `isAuthCommand(cmd)`. Added a regression test that `signing revoke-key` / `signing register-key` no longer inherit --key-id from the active profile.
- ⚠️ **Finding 1 (`login.go:132`, --key-id not persisted on bootstrap)** — already addressed in `a965cbe9`: `syncProfile` reads `--signing-key-id`, falls back to `--key-id`, and persists the resolved value as `profile.signingKeyId`. Guarded by `TestSyncProfile_BootstrapUsesKeyIDForSigningKeyID` and `TestSyncProfile_BootstrapPrefersSigningKeyIDOverKeyID`.
- ⚠️ **Finding 2 (`login.go:98`, active profile not updated when --profile omitted)** — already addressed in `a965cbe9`: `syncProfile` falls back to `cfg.ActiveProfile` when neither `--profile` nor `LEDGERCTL_PROFILE` is set. Guarded by `TestSyncProfile_UpdatesActiveProfileWhenFlagOmitted`.

cmd/ledgerctl/main.go
```diff
@@ -108,6 +100,19 @@ func newRootCommand() *cobra.Command {
 			resolveFlag(cmd, "response-verify-key", "LEDGERCTL_RESPONSE_VERIFY_KEY", cmdutil.ProfileFlagValue(p, "response-verify-key"))
 			resolveFlag(cmd, "result-file", "LEDGERCTL_RESULT_FILE", "")
 
+			// Feed the profile's signingKeyId into --key-id, but only for
+			// commands under "auth": in practice the JWT key ID and the
+			// request-signing key ID are the same key entry, so `auth login`
+			// / `auth generate-token` benefit from the fallback. Signing
+			// management commands (`signing register-key`, `signing
```
NumaryBot: 🔴 **[blocker] Global --key-id profile fallback applies indiscriminately to all subcommands including destructive signing commands**

The new global fallback that populates `--key-id` from the active profile's `signingKeyId` runs for every subcommand that declares `--key-id`, not just `auth login` / `auth generate-token`. Commands like `signing revoke-key` and `signing register-key` also declare `--key-id`; omitting it previously produced an explicit error, but now they silently inherit the profile key ID. This means `ledgerctl signing revoke-key` (with no `--key-id`) can silently revoke the active profile's current signing key rather than aborting, with no confirmation from the user.

**Suggestion:** Gate the signingKeyId profile fallback so it only applies to auth-related commands (e.g. `auth login`, `auth generate-token`). Signing management commands should continue to require an explicit `--key-id` flag and fail loudly when it is absent.

cmd/ledgerctl/auth/login.go
NumaryBot, line 133: 🟠 **[major] Active profile not updated when --profile is omitted and server address changes**

When an active profile already exists and the user runs `ledgerctl auth login --server <new>` without `--profile`, the early-return path skips updating the active profile's server address. The token is stored under the new server address, but subsequent commands still resolve the active profile's old server and therefore look up the wrong keychain entry. The previous implementation updated `cfg.ActiveProfile` in this scenario.

**Suggestion:** Before returning early, update `cfg.ActiveProfile` (and persist the config) so that the active profile's server field reflects the newly provided server address.

cmd/ledgerctl/auth/login.go
NumaryBot, line 167: 🟠 **[major] --key-id value not persisted into profile.signingKeyId when bootstrapping a new profile**

For the bootstrap flow `auth login --profile prod ... --key-id prod-key`, the local `--key-id` flag value is never copied into `profile.signingKeyId`; only `--signing-key-id` is read at that point. The new profile is therefore saved without a signing key ID. A subsequent `auth login --profile prod` cannot use the intended profile-derived key-ID fallback and fails with `required flag "key-id" not set` unless the user re-supplies the flag on every invocation.

**Suggestion:** After resolving the `--key-id` value (whether from flag or profile fallback), write it back into `profile.signingKeyId` before saving the profile.

cmd/ledgerctl/auth/login.go
```diff
@@ -117,52 +181,39 @@ func resolveLoginParams(cmd *cobra.Command) (tokenParams, error) {
...
-		keyID, _ = cmd.Flags().GetString("key-id")
-	}
-
-	if cmd.Flags().Changed("subject") {
-		subject, _ = cmd.Flags().GetString("subject")
-	}
+	// Read flag values regardless of Changed: they may have been filled by
+	// PersistentPreRunE from the active profile (e.g. profile.signingKey ->
+	// --signing-key, profile.signingKeyId -> --key-id).
+	keyID, _ := cmd.Flags().GetString("key-id")
```
NumaryBot: 🟠 **[major] Bootstrap with --signing-key-id still fails with 'required flag key-id not set'**

When bootstrapping a new profile with `auth login --profile <new> --signing-key-id k ...`, there is no existing profile for the `profile.signingKeyId -> --key-id` fallback to read from, and `resolveLoginParams` only reads the local `--key-id` flag. The command fails with `required flag "key-id" not set` before `syncProfile` can persist the `signingKeyId`, even though the bootstrap path records `--signing-key-id` as the same key entry. Users who supply `--signing-key-id` instead of `--key-id` during a new-profile bootstrap hit this error.

**Suggestion:** In `resolveLoginParams` (or immediately before it), if `--key-id` is not set but `--signing-key-id` is provided, use the `--signing-key-id` value to satisfy the `--key-id` requirement for the bootstrap path.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

gfyrag:
Round 3 of NumaryBot review — analysis of the 4 findings on commit \`5384eec2\`:

- ✅ **Line 187 (`resolveLoginParams` --signing-key-id doesn't feed --key-id at bootstrap)** — real, newly-surfaced. Fixed in \`63d13c39\`: symmetric fallback in \`resolveLoginParams\` mirrors the one already in \`syncProfile\` bootstrap. \`TestResolveLoginParams_FallsBackToSigningKeyIDForKeyID\` guards it.
- ⚠️ **Line 107 (blocker, --key-id global fallback)** — false positive. The bot points at what is now a comment line; the resolveFlag call is gated by \`isAuthCommand(cmd)\` since \`5384eec2\`, guarded by \`TestProfileSigningKeyIDDoesNotFeedSigningCommands\`.
- ⚠️ **Line 132 (--key-id not persisted on bootstrap)** — false positive, still. Fixed in \`a965cbe9\`, guarded by \`TestSyncProfile_BootstrapUsesKeyIDForSigningKeyID\`.
- ⚠️ **Line 98 (active profile not updated when --profile omitted)** — false positive, still. Fixed in \`a965cbe9\`, guarded by \`TestSyncProfile_UpdatesActiveProfileWhenFlagOmitted\`.

cmd/ledgerctl/auth/login.go
```diff
@@ -117,52 +181,48 @@ func resolveLoginParams(cmd *cobra.Command) (tokenParams, error) {
...
 			return tokenParams{}, fmt.Errorf("decoding bundle signingKey: %w", err)
 		}
 
 		seed = decoded
 
-		// Bundle values fill in anything not explicitly set on the command line.
-		if keyID == "" {
+		// The bundle wins over env/profile-derived flag values, but explicit
+		// CLI flags (Changed=true) override the bundle.
+		if !cmd.Flags().Changed("key-id") && bundle.KeyID != "" {
```
NumaryBot: 🟡 **[minor] Explicit --signing-key-id overwritten by bundle.KeyID**

When `auth login` is invoked with a bundle and an explicit `--signing-key-id` but no `--key-id`, the fallback copies `--signing-key-id` into `keyID`, but the subsequent bundle override condition only checks `Changed("key-id")`. It therefore unconditionally overwrites the user-supplied `--signing-key-id` value with `bundle.KeyID`, violating the documented CLI-over-bundle precedence.

**Suggestion:** Change the bundle override guard to also skip the override when `--signing-key-id` was explicitly set: e.g. `if !cmd.Flags().Changed("key-id") && !cmd.Flags().Changed("signing-key-id") { keyID = bundle.KeyID }`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

cmd/ledgerctl/auth/login.go
```diff
@@ -117,52 +181,51 @@ func resolveLoginParams(cmd *cobra.Command) (tokenParams, error) {
...
 			return tokenParams{}, fmt.Errorf("decoding bundle signingKey: %w", err)
 		}
 
 		seed = decoded
 
-		// Bundle values fill in anything not explicitly set on the command line.
-		if keyID == "" {
+		// The bundle wins over env/profile-derived flag values, but explicit
+		// CLI flags (Changed=true) override the bundle. --signing-key-id is
+		// the sibling flag whose value we fall back to for keyID above, so
```
NumaryBot, line 266: 🟡 **[minor] Explicit --signing-key-id overwritten by bundle.KeyID**

When `auth login` is invoked with a bundle and an explicit `--signing-key-id` flag (but no `--key-id`), the fallback correctly copies `--signing-key-id` into `keyID`. However, the subsequent bundle override condition only checks `Changed("key-id")`, so it unconditionally overwrites the user-supplied value with `bundle.KeyID`, violating the documented CLI-over-bundle precedence for the bootstrap path.

**Suggestion:** Broaden the guard to also skip the override when `--signing-key-id` was explicitly set: `if !cmd.Flags().Changed("key-id") && !cmd.Flags().Changed("signing-key-id") { keyID = bundle.KeyID }`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

gfyrag:
Round 4: NumaryBot re-posted the same [minor] finding it flagged on `63d13c39` — my fix landed as commit `49887939` in that same push, and the check-level "NumaryBot review" status is SUCCESS. The guard now reads `!Changed("key-id") && !Changed("signing-key-id") && bundle.KeyID != ""` (login.go:216), covered by `TestResolveLoginParams_ExplicitSigningKeyIDBeatsBundle`.

cmd/ledgerctl/auth/login.go
```diff
@@ -69,13 +69,8 @@ func runLogin(cmd *cobra.Command, _ []string) error {
...
 	}
 
-	// If --server was explicitly passed and a profile is active, update the
-	// profile's server address so that subsequent commands use the same address
-	// (and find the keychain token keyed by the full address including port).
-	if cmd.Flags().Changed("server") {
-		if err := updateProfileServer(cmd, server); err != nil {
-			pterm.Warning.Printfln("Could not update profile server: %v", err)
-		}
+	if err := syncProfile(cmd, server); err != nil {
```
paul-nicolas, line 85: [Medium] `auth login` reports success even when the advertised profile bootstrap fails

`syncProfile`'s error is downgraded to a `pterm.Warning` and `runLogin` returns `nil`, so the command exits 0. This best-effort handling was fine when the only job was refreshing an existing profile's server, but this PR makes profile *bootstrap* an advertised feature (docs/ops/cli.md: "When `--profile <name>` targets a profile that does not yet exist, bootstraps it from the current connection flags").

Failure scenario: the token is already stored in the keychain at line 68, then `LoadConfig`/`SaveConfig` inside `syncProfile` fails (config dir not writable, disk full, malformed existing config). `auth login --profile prod ...` prints `Logged in to prod:8888` (green) plus a warning, and exits 0. A CI wrapper gating on the exit code (`auth login --profile prod ... && ledgerctl --profile prod ledgers list`) proceeds on a false success, and the second command fails with `profile "prod" not found` — the advertised bootstrap silently did not happen.

Fix: for the bootstrap path, return the error from `runLogin` instead of downgrading it to a warning, so the exit code reflects that the requested profile was not created. To keep keychain/config state aligned on failure, either store the keychain token only after a successful profile write, or remove the just-written token when sync fails. The soft-warning behavior remains appropriate for the pure server-address refresh of an already-existing profile.

paul-nicolas (commented):

Reviewed the three linked flag-resolution fixes. **The highest-risk change — `resolveFlag` switching from `FlagSet.Set()` to `Flag.Value.Set()` so the `Changed` bit stays false for env/profile-derived values — was independently verified safe.** The only cross-cutting reader of the `Changed` bit for the affected flags is `ResolveTokenSource` (`cmdutil/client.go:177`), and the change *fixes* a latent mislabel there (env-only `LEDGERCTL_AUTH_TOKEN` was reported as `flag (--auth-token)`; now correctly `environment (...)`). Every other consumer reads the value via `GetString`, which `Value.Set` still populates. No caller silently breaks. Build passes; all `cmd/ledgerctl/...` tests pass.

### Severity counts
- Critical: 0
- Major: 0
- Medium: 1 (posted inline)
- Minor: 3 (below)
- Nit: 3 (below)

**Posted inline: 1** (the Medium). **Withheld: 6** (Minor/Nit, listed below).

### Verification outcome
Two findings entered Medium+ verification (both from Codex): 1 posted after downgrade, 1 downgraded to Minor.
- `login.go:72` — Codex rated **major**; confirmed valid, **adjusted to Medium** (token is stored + a warning is printed, so it is not merge-blocking, but the advertised bootstrap can exit 0 while half-completing). Posted inline.
- `login.go:126` — Codex rated **medium** (unlocked read-modify-write of `config.json`); technically true but **adjusted to Minor** — the unlocked `LoadConfig`+`SaveConfig` pattern is pre-existing (`profile create`/`use`/`delete` and the removed `updateProfileServer` all do the same), so this PR does not regress concurrency safety. Withheld to the list below.

### Conflicts between models
Codex surfaced the two `login.go` findings above; the Claude reviewer surfaced neither (it focused on clearing the `Changed`-bit safety and rated all of its own findings minor/nit). Resolved by verifying both Codex findings against the actual code and re-mapping severity as noted.

- `cmd/ledgerctl/auth/login.go:126` -- [Minor] `syncProfile` does an unlocked read-modify-write of `config.json` (`LoadConfig` then `SaveConfig`); two concurrent `ledgerctl` processes can clobber each other. Pre-existing pattern (also in `profile create`/`use`/`delete`); a proper fix is a cross-cutting `SaveConfig` change (advisory lock + temp-file+rename), out of scope for this bug fix.
- `cmd/ledgerctl/auth/login.go:163` -- [Minor] First-profile activation (`if len(cfg.Profiles) == 1`) misses the "profiles exist but none active/dangling" case: `auth login --profile new` then leaves no active profile, so later no-`--profile` commands fall back to `localhost:8888` and miss the stored token. Consider `|| cfg.ActiveProfile == ""`.
- `docs/ops/cli.md:2718` -- [Minor] cli.md documents the `--profile` bootstrap and the existing-profile + explicit-`--server` rewrite, but not the third synced behavior (`auth login --server X` with no `--profile` rewrites the *active* profile's server), which `TestSyncProfile_UpdatesActiveProfileWhenFlagOmitted` exercises. CLAUDE.md requires cli.md to track CLI behavior.
- `docs/ops/cli.md:2719` -- [Nit] "The active profile's `signingKeyId`" is imprecise: it is the *resolved* profile (`--profile`/`LEDGERCTL_PROFILE` overrides active) and only for `auth` subcommands.
- `cmd/ledgerctl/main.go:112` -- [Nit] The `profile.signingKeyId -> --key-id` fallback (`isAuthCommand`) also applies to `auth generate-token`, not just `auth login` (intended per the code comment) but untested; add a test asserting `generate-token` inherits `--key-id` from the profile.
- `cmd/ledgerctl/auth/login_test.go:216` -- [Nit] Missing test for the documented "bundle > profile-derived" precedence on `key-id`: profile `signingKeyId=prof-key` + `--bundle` with `keyId=bundle-key` and no CLI key flag should yield `bundle-key`.

cmd/ledgerctl/auth/login.go
gfyrag: Fix pushed in `2d312cd0`:
- Medium (line 72): reordered so syncProfile runs before keychain.Set, and syncProfile errors now propagate (no more silent exit 0 on failed bootstrap).
- Minor (line 163): first-profile activation now also fires when `cfg.ActiveProfile == ""`.
- Nits: added `auth generate-token` test for the profile → `--key-id` fallback, and a `bundle > profile-derived` precedence test on `key-id`.
- Docs: cli.md now covers the third sync path (`--server` without `--profile` rewrites the active profile), the auth-only scope of the `--key-id` fallback, and the bootstrap-failure semantics.
Skipped: the unlocked config read-modify-write concurrency finding — as you noted, out of scope for this PR.

cmd/ledgerctl/auth/login.go
```diff
@@ -65,17 +65,18 @@ func runLogin(cmd *cobra.Command, _ []string) error {
 
 	server, _ := cmd.Flags().GetString("server")
 
-	if err := cmdutil.GetKeyring(cmd).Set(server, token); err != nil {
-		return fmt.Errorf("storing token in keychain: %w", err)
+	// Sync the profile before storing the token so the two stay aligned:
+	// bootstrap is an advertised feature of `auth login --profile <new>`,
+	// and storing the token first would let the command exit 0 with a
+	// warning even when the profile write failed — a CI wrapper chaining
+	// `auth login --profile prod ... && ledgerctl --profile prod ...`
```
NumaryBot: 🟠 **[major] syncProfile bootstrap failure silently exits 0**

When syncProfile fails (e.g. config directory not writable, disk full, malformed existing config), the error is downgraded to a pterm.Warning and runLogin returns nil, so the command exits 0. Because this PR advertises profile bootstrap as a first-class feature, a CI pipeline that gates on the exit code (auth login --profile prod ... && ledgerctl --profile prod ledgers list) will proceed on a false success: the token is already stored in the keychain, the profile was never written, and the next command fails with 'profile not found'.

**Suggestion:** For the bootstrap path specifically, return the error from syncProfile rather than downgrading it to a warning. A guard such as `if isBootstrap { return err }` before the warning fallback would preserve backward-compatible best-effort behaviour for the existing-profile refresh path while making the new bootstrap path fail loudly on error.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

cmd/ledgerctl/main.go
```diff
@@ -108,6 +100,19 @@ func newRootCommand() *cobra.Command {
...
+			// revoke-key`) also declare --key-id but must not default to the
+			// active profile's signing key — a silent default there could
+			// revoke the current key on `ledgerctl signing revoke-key` with
+			// no arguments.
+			if isAuthCommand(cmd) {
+				resolveFlag(cmd, "key-id", "", cmdutil.ProfileFlagValue(p, "signing-key-id"))
```
NumaryBot: 🟠 **[major] Profile-derived --key-id fallback ignores explicit --signing-key-id, causing stale key usage**

When an auth command is invoked with an active profile that has a `signingKeyId`, and the caller supplies `--signing-key-id <new>` but omits `--key-id`, the global fallback still populates `--key-id` from the profile because it only checks `Changed("key-id")`. As a result, `auth login` / `auth generate-token` can sign a JWT with the stale profile key ID instead of the explicitly provided CLI value, silently breaking key rotations or one-off overrides.

**Suggestion:** Guard the profile-derived fallback so it is skipped when `--signing-key-id` was explicitly set on the command line, e.g.: `if !cmd.Flags().Changed("key-id") && !cmd.Flags().Changed("signing-key-id") { /* apply profile fallback */ }`. This mirrors the fix already applied inside `resolveLoginParams` for the bundle override path.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

cmd/ledgerctl/main.go
NumaryBot: 🟠 **[major] Profile-derived --key-id fallback ignores explicit --signing-key-id, causing stale key usage**

When an auth command is invoked with an active profile that has a signingKeyId, and the caller supplies --signing-key-id <new> (or LEDGERCTL_SIGNING_KEY_ID=new via environment) but omits --key-id, the global fallback still populates --key-id from the stale profile value because it only checks Changed("key-id"). Since Changed() returns false for env-derived flags too, auth login and auth generate-token can silently sign JWTs with the wrong key ID during key rotations or one-off overrides. This is the only remaining active finding — all other major/blocker findings from previous rounds have been resolved and confirmed fixed by the author.

**Suggestion:** Broaden the guard to also skip the profile fallback when --signing-key-id was explicitly set: e.g. `if !cmd.Flags().Changed("key-id") && !cmd.Flags().Changed("signing-key-id") { /* apply profile fallback */ }`. This mirrors the fix already applied inside resolveLoginParams for the bundle override path (login.go:216).

NumaryBot: 🟠 **[major] Profile-derived --key-id fallback ignores explicit --signing-key-id, causing stale key usage**

When an auth command (auth login or auth generate-token) is invoked with an active profile that has a signingKeyId, and the caller supplies --signing-key-id <new> (or the LEDGERCTL_SIGNING_KEY_ID env var) but omits --key-id, the global fallback still populates --key-id from the stale profile value because the guard only checks Changed("key-id"). Changed() also returns false for env-var-derived flags, so the stale profile key ID is silently used during key rotations or one-off overrides. For auth generate-token this is especially impactful because that command reads --key-id directly via tokenParamsFromFlags and does not have the corrective resolveLoginParams step that auth login has. The symmetric guard is already applied inside resolveLoginParams for the bundle override path (login.go:216) but has not been applied here.

**Suggestion:** Broaden the guard to skip the profile fallback when --signing-key-id was also explicitly set: `if !cmd.Flags().Changed("key-id") && !cmd.Flags().Changed("signing-key-id") { /* apply profile fallback */ }`. This mirrors the pattern already used inside resolveLoginParams.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

cmd/ledgerctl/auth/login.go
```diff
@@ -65,17 +65,18 @@ func runLogin(cmd *cobra.Command, _ []string) error {
...
+	// would then move on and hit `profile "prod" not found`.
+	if err := syncProfile(cmd, server); err != nil {
```
NumaryBot: 🟠 **[major] Profile state committed before keychain write succeeds**

After the reordering fix (syncProfile before keychain.Set), a failed OS keyring write (e.g. locked macOS Keychain or libsecret unavailable) leaves the profile already created/updated and possibly activated, but no token stored. Subsequent commands resolve the new profile/server and look up a token that was never written. For an existing active profile whose server was just changed, this is especially disruptive: the old working state is gone and the new state is broken with no recovery hint.

**Suggestion:** Treat the profile mutation and the keychain write as a logical unit. Either write the token first and roll back the profile on failure, or persist the profile only after a successful keychain.Set call. At minimum, detect the isBootstrap case and return an actionable error (not a warning) so the user knows the profile was not actually created.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

cmd/ledgerctl/auth/login.go
```diff
@@ -65,17 +65,23 @@ func runLogin(cmd *cobra.Command, _ []string) error {
...
 	server, _ := cmd.Flags().GetString("server")
 
-	if err := cmdutil.GetKeyring(cmd).Set(server, token); err != nil {
+	// Order: keychain first, then profile. If profile sync fails, roll back
+	// the just-stored token so we never leave the config pointing at a
+	// server whose keychain entry is absent (or worse, points at the wrong
+	// token). Bootstrap is an advertised feature of `auth login --profile
+	// <new>`, and this two-phase commit is the only way to keep exit code,
+	// keychain, and config aligned without introducing a lock file.
+	keyring := cmdutil.GetKeyring(cmd)
```
NumaryBot, line 75: 🟠 **[major] Profile state committed before keychain write succeeds, leaving CLI config inconsistent on keyring failure**

After the reordering fix (syncProfile before keychain.Set), a failed OS keyring write (e.g. locked macOS Keychain, libsecret unavailable) leaves the profile already created/updated and possibly activated in the CLI config, but no token ever stored. Subsequent commands resolve the new profile/server and attempt to look up a token that was never written. For an existing active profile whose server was just changed, this is especially disruptive: the old working state is permanently overwritten and the new state is broken. Additionally, on the rollback path the prior credential is deleted entirely rather than restored, so even a server that previously had a working keychain entry ends up logged out after a failed command.

**Suggestion:** Treat the profile mutation and the keychain write as a logical unit: either write the token first (restoring it on profile-sync failure) or defer persisting the profile until after a successful keychain.Set. At a minimum, before calling keychain.Set, save the pre-existing token for the server so it can be restored if keychain.Set succeeds but syncProfile fails — or if keychain.Set itself fails and the profile was already mutated. For the bootstrap path, return a hard error (not a warning) so the user sees the profile was not created.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

cmd/ledgerctl/auth/login.go
```diff
@@ -65,17 +65,23 @@ func runLogin(cmd *cobra.Command, _ []string) error {
...
 		return fmt.Errorf("storing token in keychain: %w", err)
 	}
 
-	// If --server was explicitly passed and a profile is active, update the
-	// profile's server address so that subsequent commands use the same address
-	// (and find the keychain token keyed by the full address including port).
-	if cmd.Flags().Changed("server") {
-		if err := updateProfileServer(cmd, server); err != nil {
-			pterm.Warning.Printfln("Could not update profile server: %v", err)
+	if err := syncProfile(cmd, server); err != nil {
```
NumaryBot, line 85: 🟠 **[major] Rollback deletes pre-existing keychain credential instead of restoring it**

When `syncProfile` fails after `keyring.Set` has already written a token for a server that previously had a stored credential, the rollback path deletes the credential unconditionally rather than restoring the prior value. A config write/parse failure during `auth login --profile ...` can therefore silently destroy an existing, working keychain entry for that server, leaving the user with no token at all even though the command failed and exited non-zero. This is the atomicity gap flagged in prior discussion (Thread 17/NumaryBot) that was not fully addressed by the reordering fix.

**Suggestion:** Before calling `keyring.Set`, capture any existing token for that server (e.g. via `keyring.Get`). On rollback, if a prior token was found, call `keyring.Set` to restore it; only call `keyring.Delete` when there was no prior token. This ensures a failed `auth login` leaves the user in exactly the state they were in before the command ran.

cmd/ledgerctl/auth/login.go
```diff
@@ -65,17 +65,37 @@ func runLogin(cmd *cobra.Command, _ []string) error {
...
+	// at a server whose keychain entry is absent (or worse, points at a
+	// token whose corresponding profile write never landed). Bootstrap is
+	// an advertised feature of `auth login --profile <new>`, and this
+	// two-phase commit is the only way to keep exit code, keychain, and
+	// config aligned without introducing a lock file.
+	keyring := cmdutil.GetKeyring(cmd)
+
+	// Snapshot any pre-existing token so a re-login that fails at sync time
+	// doesn't clobber the previously valid credential.
+	prevToken, prevErr := keyring.Get(server)
```
NumaryBot, line 79: 🟠 **[major] Keychain rollback deletes existing credential when snapshot read fails for non-ErrTokenNotFound reason**

When `keyring.Get(server)` fails with an error that is not ErrTokenNotFound (e.g. a transient backend error or item permission issue), the code continues as if no prior token existed and records `prevErr` as non-nil. If `keychain.Set` subsequently succeeds but the profile config write (`syncProfile`) later fails, the rollback path executes `Delete` instead of restoring the original credential. This permanently destroys a previously working auth state even though the command exits non-zero, leaving the user unable to recover without re-authenticating manually. This issue was flagged in the prior PR discussion (NumaryBot at 2026-07-02T14:21:43Z) for the unconditional-delete case; the reordering fix introduced a snapshot mechanism but did not fully close the gap for non-ErrNotFound read failures.

**Suggestion:** Before proceeding with the keychain write, check if the snapshot read returned a non-ErrTokenNotFound error and abort early (return the error) rather than treating it as 'no prior token'. This ensures the rollback path only issues a Delete when the snapshot confirmed there was genuinely no pre-existing credential, and preserves existing credentials in all other failure scenarios.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

cmd/ledgerctl/auth/login.go
```diff
@@ -84,85 +114,180 @@ func runLogin(cmd *cobra.Command, _ []string) error {
...
+//   - env LEDGERCTL_SIGNING_KEY_ID doing the same (env leaves Changed=false
+//     via Flag.Value.Set in cmdutil's resolveFlag).
+//   - Bootstrap with only --signing-key-id and no profile fallback.
+//
+// Bundle values still win against an env/profile-derived --signing-key-id via
+// the downstream bundle-override guard in resolveLoginParams; a CLI --key-id
+// still wins against everything via the Changed check here.
+func resolveKeyID(cmd *cobra.Command) string {
+	keyID, _ := cmd.Flags().GetString("key-id")
+	if cmd.Flags().Changed("key-id") {
```
NumaryBot: 🟠 **[major] Env-derived KEY_ID treated as explicit --key-id, overriding --signing-key-id**

When KEY_ID is set in the environment, bindSubcommandEnv marks the local --key-id flag as Changed before Cobra parses the CLI. If a user supplies --signing-key-id <new> explicitly but omits --key-id, the Changed("key-id") check returns true for the env-derived value, causing the env key ID to win over the explicit CLI signing-key-id. This means auth login and auth generate-token can silently sign JWTs with a stale key ID during key rotations or CI overrides, without any error or warning.

**Suggestion:** Distinguish an environment-bound --key-id from one actually typed on the command line. Either avoid marking env-bound local flags as Changed, or apply the same broadened guard already used for the bundle override path: skip the env/profile-derived key-id when --signing-key-id was explicitly set on the CLI (e.g. track a separate boolean for "was --key-id typed, not env-injected").

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

cmd/ledgerctl/main.go
NumaryBot: 🟠 **[major] Profile-derived --key-id fallback ignores explicit --signing-key-id and env-derived KEY_ID, causing stale key usage**

When an auth command (auth login or auth generate-token) is invoked with an active profile that has a signingKeyId, and the caller supplies --signing-key-id <new> or sets KEY_ID via environment but omits --key-id, the global fallback still populates --key-id from the stale profile value because the guard only checks Changed("key-id"). Changed() also returns false for env-var-derived flags, so the stale profile key ID is silently used during key rotations or one-off overrides. For auth generate-token this is especially impactful because that command reads --key-id directly via tokenParamsFromFlags and does not have a corrective resolveLoginParams step. The symmetric guard is already applied inside resolveLoginParams for the bundle override path (login.go:216) but has not been applied here.

**Suggestion:** Broaden the guard to also skip the profile fallback when --signing-key-id was explicitly set or when KEY_ID was set via environment: e.g. `if !cmd.Flags().Changed("key-id") && !cmd.Flags().Changed("signing-key-id") { /* apply profile fallback */ }`. For the env-var case, check whether the raw env value is non-empty before applying the profile default.

cmd/ledgerctl/main.go
```diff
@@ -294,7 +320,7 @@ func bindFlagSetSkippingOwned(set *pflag.FlagSet) {
 
 		// Ignore the error: an invalid env value leaves the cobra default in
 		// place, matching resolveFlag's best-effort env handling.
-		_ = set.Set(flag.Name, value)
+		_ = flag.Value.Set(value)
```
NumaryBot: 🟠 **[major] Global env-var binding leaves Changed() false, silently breaking non-auth subcommand logic**

The new global mechanism that binds environment variables to flags calls flag.Value.Set() without marking the flag as Changed. Several existing subcommands (e.g. ledgers create) rely on Changed() to distinguish a user-provided value from a default, and use that check to enable behaviour such as mirror-mode inference. With this change, MIRROR_DSN=... ledgerctl ledgers create --name x silently ignores the env-provided mirror config and creates a plain ledger instead. The auth-specific requirement to keep Changed=false for the profile-derived key-id fallback should not be applied globally to all flags and all subcommands.

**Suggestion:** Limit the env-var-to-flag binding that deliberately keeps Changed=false to auth-specific flags, or update every non-auth caller that uses Changed() to detect env-provided values so they are not silently broken. A narrower guard (e.g. only apply the custom binding inside isAuthCommand paths) would avoid the global semantic change.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1483#issuecomment-4866778462

sylr (approved):

PR #1485, opened 2026-07-02
fix(EN-1443): purge same-batch reverse-map orphans on metadata field removal

paul-nicolas:
## Problem

When an indexed metadata field is removed, `handleRemovedMetadataFieldType` purges its read-store projection. The reverse-map (`0x03`) limb (`purgeReverseMapForKey`) scanned a **committed-only** Pebble snapshot. A reverse-map PUT written earlier in the **same** uncommitted builder batch (e.g. a `SavedMetadata` on the indexed field) was invisible to that snapshot, so no delete was emitted and the stale row committed at `Flush()` — an orphaned reverse-map row for a field that no longer exists.

This is a read-index projection that diverges from a from-scratch replay, violating the "every persisted projection must be verifiable against the audit" invariant (CLAUDE.md #8).

The forward-index and entity-exists limbs are unaffected: their range tombstone gets a higher Pebble sequence number than earlier same-batch SETs, so it already shadows uncommitted rows. Only the reverse map — which cannot be range-deleted — relied on the snapshot alone.

## Fix

`purgeReverseMapForKey` now purges from **both** sources:
- committed rows via the snapshot scan (unchanged), and
- same-batch uncommitted PUTs via the batch's read-your-writes `rmapOverlay` — the mechanism the codebase already maintains for exactly this ("callers resolve the current value via `ReverseMapOverlay` before falling back to committed state"). `dal.WriteSession` deliberately exposes no iterator, so the overlay is the sanctioned way to read in-flight reverse-map writes.

This keeps the live path's commit/progress boundaries intact (no mid-batch flush).

New `WriteBatch` helpers:
- `DeleteReverseMapKey` — deletes and clears the overlay entry so it never drifts from the batch.
- `RangeReverseMapOverlay` — read-only iteration over buffered reverse-map mutations.

The two self-contradictory doc comments are corrected.

## Testing

New `process_metadata_field_removal_test.go` (table-driven): `committed then removed`, `same batch repro`, `mixed committed and same batch`, `unrelated key survives`. Verified the three same-batch cases **fail before** this change (orphan row survives) and all pass after. `just pre-commit` (0 lint issues) and the full `just test` unit suite are green.

## Out of scope

- The "empty transient-set persisted" limb — already cancelled on every commit path via `appliedProposalSync.err()`.
- A checker-side pass detecting reverse-map orphans (`compareIndexes` verifies the registry, not reverse-map rows) — noted as a follow-up.

Jira: https://formance-team.atlassian.net/browse/EN-1443

### DIFF
--- a/internal/application/indexbuilder/process_metadata_field_removal.go
+++ b/internal/application/indexbuilder/process_metadata_field_removal.go
@@ -1,6 +1,8 @@
 package indexbuilder
 
 import (
+	"bytes"
+
 	"github.com/cockroachdb/pebble/v2"
 
 	"github.com/formancehq/ledger/v3/internal/domain/indexes"
@@ -22,14 +24,18 @@ import (
 //     covers every v_n (both null and non-null variants).
 //   - 0x03 (reverse map): keys are
 //     [ns:][entityID\x00][version:4B BE][metaKey] — metaKey is the
-//     suffix but it's preceded by a fixed-width version block, so we
-//     scan the per-namespace range and use extractMetadataKeyFromReverseMap
-//     (which already accounts for the version block) to delete every
-//     row matching this metaKey regardless of version.
+//     suffix but it's preceded by a fixed-width version block, so it
+//     can't be range-deleted by (ns, key). purgeReverseMapForKey scans
+//     the per-namespace range and deletes every row matching this
+//     metaKey regardless of version.
 //
-// The reverse-map scan reuses the in-flight indexed batch as snapshot
-// would not see uncommitted writes, but the scan iterates from a fresh
-// Pebble snapshot for consistency. The handler also strips the
+// Unlike the forward/entity-exists range deletes — whose range tombstone
+// gets a higher sequence number than earlier same-batch SETs and so
+// shadows uncommitted rows too — the reverse-map point-delete scan reads
+// a committed-only Pebble snapshot. It therefore also consults the
+// batch's read-your-writes overlay to purge reverse-map PUTs written
+// earlier in this same uncommitted batch (see purgeReverseMapForKey).
+// The handler also strips the
 // corresponding entry from the local ledgerIndexConfig cache so the
 // live path stops considering the index as active immediately, and
 // drops the per-replica IndexVersionState entry so the boot orphan
@@ -111,21 +117,38 @@ func (b *Builder) handleRemovedMetadataFieldType(
 	return nil
 }
 
-// purgeReverseMapForKey scans the reverse map for a (ns, key) pair and
-// deletes every entry whose metadata key field matches, across every
-// per-replica forward-encoding version. The reverse map can't be
-// range-deleted by (ns, key) because the rmap key shape is
-// [entity\x00][version:4B BE][metaKey] — metaKey sits after a
-// fixed-width version block, not directly after the entity null. The
-// scan cost is bounded by the number of (entity, version) tuples that
-// ever held this metadata key.
+// purgeReverseMapForKey deletes every reverse-map entry whose metadata key
+// field matches (ns, key), across every per-replica forward-encoding version.
+// The reverse map can't be range-deleted by (ns, key) because the rmap key
+// shape is [entity\x00][version:4B BE][metaKey] — metaKey sits after a
+// fixed-width version block, not directly after the entity null.
+//
+// Two sources are purged:
+//   - Committed rows: scanned from a fresh Pebble snapshot (committed data
+//     only). Scan cost is bounded by the number of (entity, version) tuples
+//     that ever held this metadata key.
+//   - In-flight rows: reverse-map PUTs written earlier in this same
+//     uncommitted batch are invisible to the snapshot, so we also consult the
+//     batch's read-your-writes overlay. Without this, a SavedMetadata on the
+//     indexed field in the same batch as the removal would commit an orphaned
+//     reverse-map row for a field that no longer exists (EN-1443).
 func (b *Builder) purgeReverseMapForKey(kb *dal.KeyBuilder, ledgerName string, ns, key string) error {
 	rmapPrefix := readstore.ReverseMapPrefix(kb, ledgerName, ns)
 	upper := readstore.IncrementBytes(rmapPrefix)
 
-	// Use a Pebble snapshot so the scan sees committed data only; the
-	// in-flight batch writes from the same processing pass would otherwise
-	// surface as ghost entries.
+	// Deduplicate so a key present in both committed state and the in-flight
+	// overlay is deleted only once.
+	seen := make(map[string]struct{})
+	deleteMatch := func(k []byte) error {
+		if _, done := seen[string(k)]; done {
+			return nil
+		}
+		seen[string(k)] = struct{}{}
+
+		return b.wb.DeleteReverseMapKey(k)
+	}
+
+	// Committed rows.
 	snap := b.readStore.NewSnapshot()
 	defer func() { _ = snap.Close() }()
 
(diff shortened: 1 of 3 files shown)


internal/application/indexbuilder/process_metadata_field_removal.go
```diff
@@ -111,21 +117,38 @@ func (b *Builder) handleRemovedMetadataFieldType(
...
+// Two sources are purged:
+//   - Committed rows: scanned from a fresh Pebble snapshot (committed data
+//     only). Scan cost is bounded by the number of (entity, version) tuples
+//     that ever held this metadata key.
+//   - In-flight rows: reverse-map PUTs written earlier in this same
+//     uncommitted batch are invisible to the snapshot, so we also consult the
+//     batch's read-your-writes overlay. Without this, a SavedMetadata on the
+//     indexed field in the same batch as the removal would commit an orphaned
+//     reverse-map row for a field that no longer exists (EN-1443).
 func (b *Builder) purgeReverseMapForKey(kb *dal.KeyBuilder, ledgerName string, ns, key string) error {
```
paul-nicolas, line 135: [Medium] Reverse-map (0x03) row contents still have no checker verification (invariant #8)

This PR's justification is invariant #8 — a persisted projection diverging from an audit replay. It fixes one orphan-producing path, but the reverse-map projection as a whole is still not re-derived by `Check()`: `compareIndexes` verifies only the registry (presence + identity), and `internal/application/check/checker.go:388-397` explicitly states readstore *contents* are intentionally not re-derived ("a cross-cutting invariant-#8 effort tracked separately"). So the exact corruption class this PR targets remains silent on `Check()`.

Non-blocking, and I agree with the PR description's decision to defer it — flagging only for visibility. Concrete follow-up: add a checker pass that replays the audit-bound metadata logs against the current index registry to re-derive expected 0x03 rows and emits a `CHECK_STORE_ERROR_TYPE_*` on extra/missing rows, and reference EN-1443 as the motivating case so it doesn't get lost.

internal/application/indexbuilder/process_metadata_field_removal_test.go
```diff
@@ -0,0 +1,163 @@
...
+	t.Parallel()
+
+	const (
+		ledger     = "test"
+		acct1      = "acct-1"
+		acct2      = "acct-2"
+		removedKey = "role"
+		keepKey    = "team"
+	)
+	ns := readstore.NamespaceAccount
```
paul-nicolas, line 87: [Medium] New tests cover only the account namespace — the transaction overlay-purge path is unexercised

All four subtests pin `ns := readstore.NamespaceAccount` and use `savedAccountMetadata`; there is zero coverage for `NamespaceTransaction`. The transaction reverse-map key layout is genuinely different (fixed 8-byte txID + version at `suffix[8:12]`, vs the account null-scan in `parseReverseMapKey`/`extractVersionFromReverseMap`, process_logs.go), and the `bytes.HasPrefix(reverseKey, rmapPrefix)` filter + `parseReverseMapKey` call in the new overlay branch run against that layout untested here. (The extractors have separate unit coverage in `process_logs_test.go`, so the risk is a handler-path gap, not untested parsing — hence Medium, not Major.)

Fix: add a `savedTransactionMetadata` helper and one transaction-namespace subtest mirroring `same batch repro` so the overlay purge is exercised for both namespaces.

paul-nicolas (commented):
## Multi-model review — PR #1485 (EN-1443 reverse-map orphan purge)

**Models:** Claude (this repo's perspective) + Codex (gpt-5.4, external). Gemini not used.

**Verdict: the fix is correct and sound — no blocking issues.** I verified independently: `go build` + `go vet` on both changed packages are clean; the new `TestHandleRemovedMetadataFieldType_PurgesReverseMap` passes on the PR head, and the three same-batch subtests genuinely fail on the base branch (orphan row survives, count=1) — the test reproduces the bug it claims to. The overlay stays consistent with the batch (all reverse-map writers — `ReplaceMetadataIndexV`, `DeleteMetadataEntryWithPreviousV`, and the new `DeleteReverseMapKey` — mirror into `rmapOverlay`). The Go-map iteration order in `RangeReverseMapOverlay` is a non-issue: the read-side index is a per-replica projection off the FSM hot path, and the purge is an order-independent delete-set, so the committed result is identical regardless of iteration order.

**Severity counts:** 0 Critical, 0 Major, 2 Medium (posted inline), 4 Minor/Nit (listed below, not posted inline).

**Conflicts reconciled between the two reviewers:**
- *Checker-coverage gap*: Codex rated Major, Claude rated Medium → resolved to **Medium**. It is real per invariant #8 but pre-existing, explicitly deferred in this PR's own description, and documented as a separately-tracked effort at `internal/application/check/checker.go:388-397`. This PR neither introduces nor worsens it.
- *O(M·N) overlay scan*: Codex rated Medium, Claude rated Minor → resolved to **Minor**. M (fields removed per batch) is typically 1, the committed-scan half was already O(M·rows) pre-PR, and the overlay half is bounded by batch write count.
- *Transaction-namespace test gap*: Codex rated Minor, Claude rated Medium → resolved to **Medium** (both flagged it; distinct key-layout parsing path, trivially addable).

- `internal/application/indexbuilder/process_metadata_field_removal.go:182` — [Minor] Per-call full-overlay scan is O(M·N) when many fields are removed in one batch; acceptable as-is (matches the documented tradeoff in `gc.go`), revisit only if batched multi-field removals become common.
- `internal/application/indexbuilder/process_metadata_field_removal.go:141` — [Nit] The `seen` dedup set is effectively dead code: the committed pass's `DeleteReverseMapKey` sets `rmapOverlay[k]=nil` before the overlay range runs, so any shared key hits the `value==nil` guard and never re-reaches `deleteMatch`. Harmless, but the comment ("deleted only once") overstates its necessity — either drop it or reword as purely defensive.
- `internal/storage/readstore/write_batch.go:19` — [Minor] The struct comment's claim that an un-mirrored reverse-map write is "unrepresentable" is overstated: `internal/application/indexbuilder/gc.go:84` deletes reverse-map rows via raw `batch.DeleteKey`, bypassing the overlay (safe only because it runs boot-only). Scope the guarantee to the live write helpers.
- `internal/application/indexbuilder/process_metadata_field_removal_test.go:125` — [Minor] Two overlay branches are unexercised: the `value==nil` path (a same-batch write-then-delete-then-remove) and a true committed+overlay dedup collision (the "mixed" subtest uses distinct entities acct-1/acct-2, so no key is ever in both committed state and a non-nil overlay entry).

internal/application/indexbuilder/process_metadata_field_removal_test.go
paul-nicolas: Fixed in 77a818f36 — added `savedTransactionMetadata` + a `TestHandleRemovedMetadataFieldType_PurgesReverseMap_Transaction` covering `committed then removed` / `same batch repro` / `mixed` for `NamespaceTransaction`. The transaction `same batch repro` and `mixed` genuinely fail on the base branch (orphan row survives), confirming they exercise the overlay path for the distinct txID key layout. Also added two account overlay-branch subtests (`value==nil` already-deleted skip, and committed+same-batch rewrite of the same key). `just pre-commit` (0 issues) and `just test` green.

internal/application/indexbuilder/process_metadata_field_removal.go
paul-nicolas: Confirmed and agreed — this is the pre-existing, cross-cutting invariant-#8 readstore-content-verification gap documented at `internal/application/check/checker.go:388-397` ("no index type has content verification today; adding it is a cross-cutting invariant-#8 effort tracked separately"). The reverse-map projection is no more or less verified than the forward/entity-exists projections; this PR neither introduces nor worsens the gap, and the PR description already lists the checker pass as an explicit follow-up. Resolving here — EN-1443 is a good motivating case for that tracked effort.

paul-nicolas:
Added transaction-namespace, `value==nil` overlay-skip, and committed+in-flight rewrite subtests in 77a818f36, which cover the new overlay branches flagged here. The residual uncovered lines are defensive error/guard paths (iterator/DeleteKey error returns, cross-namespace `HasPrefix` skip, and the effectively-unreachable `seen` dedup skip); overall project coverage nets +0.89%.

gfyrag (approved):
LGTM — le fix réutilise l'overlay rmap déjà maintenu, ce qui reste cohérent avec l'invariant #3 (WriteSession write-only) et évite l'overhead d'un IndexedBatch. Tests bien ciblés.

PR #1486, opened 2026-07-02
feat(ledgerctl): add shell completion for enum flags

sylr:
## What

Registers shell completion for every `ledgerctl` flag that accepts a finite set of values, so bash/zsh/fish only suggest valid values.

Adds `cmdutil.RegisterEnumCompletion` (a thin wrapper over `cobra.FixedCompletions`, following the existing `--profile`/`--ledger` pattern) and wires it into every enum flag.

## Flags covered

| Command | Flag(s) |
|---|---|
| root | `--consistency` (stale, leader, linearizable) |
| `upgrade` | `--channel` (nightly, stable) |
| `account-types add` | `--persistence` (normal, ephemeral, transient) |
| `account-types set-default-enforcement` | `--mode` (STRICT, AUDIT) |
| `ledgers create` | `--mode`, `--mirror-source-type`, `--default-enforcement-mode` |
| `ledgers set-metadata-type` | `--target`, `--type` |
| `ledgers remove-metadata-type` | `--target` |
| `indexes create` / `drop` | `--type`, `--target` |
| `indexes inspect` | `--target`, `--mode` |
| `queries create` / `execute` | `--target`, `--mode` |
| `events add-sink` | `--format`, `--kafka-sasl-mechanism`, `--event-types` |
| `store backup`/… | `--driver` (s3, azure) |

## Notes

- Completion values were verified against each command's **actual parser**, so case (e.g. `STRICT`/`AUDIT` vs `stale`/`leader`) and aliases match what the CLI accepts — not just the help text.
- `--event-types` is comma-separated, so it gets a dedicated completion that preserves the typed prefix across commas, filters case-insensitively, and omits already-selected types. Covered by a unit test.
- The 9-value index-type list is deduplicated into a shared `indexTypeOptions` slice used by both the interactive selector and completion, so they can't drift.
- Metadata **type** targets offer `account`/`transaction`/`ledger`; metadata **index** targets offer `account`/`transaction` (matching help + `inspect`'s handling).

## Testing

- `go build`, `go vet`, `gofmt` clean; full `./cmd/ledgerctl/...` suite passes.
- Each completion confirmed via `ledgerctl __complete …` to return exactly the expected values.
- New unit test for the comma-separated `--event-types` completion.

## Pre-existing observation (not addressed)

`indexes create`/`drop` offer `log-ledger` in their interactive menu but have no matching `case` in the switch (selecting it errors). Completion mirrors the existing menu; the switch gap predates this change.


gfyrag (approved):

cmd/ledgerctl/indexes/id_helpers.go
```diff
@@ -2,6 +2,21 @@ package indexes
...
+// shell completion so the two never drift.
+var indexTypeOptions = []string{
+	"address",
+	"source-address",
+	"dest-address",
+	"metadata",
+	"reference",
+	"timestamp",
+	"inserted-at",
+	"log-ledger",
```
NumaryBot: 🟠 **[major] Remove invalid log-ledger enum completion**

When shell completion is used for `indexes create --type` or `indexes drop --type`, this shared list now suggests `log-ledger`, but neither `runCreateIndex` nor `runDropIndex` has a switch case that builds a log-ledger `IndexID`, so accepting that completion leads to an immediate invalid index type error. Either implement the missing case or keep this value out of the completion list until the command actually accepts it.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1486#issuecomment-4867373034

cmd/ledgerctl/indexes/id_helpers.go
NumaryBot: 🟠 **[major] Shell completion suggests `log-ledger` index type that no command handles**

The shared completion list for `--type` now includes `log-ledger`, but neither `runCreateIndex` nor `runDropIndex` has a switch case that builds a `log-ledger` IndexID. Selecting this completion leads directly to an 'invalid index type' error, so the completion actively guides users into a failing invocation.

**Suggestion:** Either implement the missing `log-ledger` case in `runCreateIndex` / `runDropIndex`, or remove `log-ledger` from the completion list until the command actually supports it.

cmd/ledgerctl/indexes/create.go
```diff
@@ -47,7 +47,9 @@ Examples:
 
 	cmd.Flags().String("ledger", "", "Name of the ledger")
 	cmd.Flags().String("type", "", "Index type: address, source-address, dest-address, metadata")
+	cmdutil.RegisterEnumCompletion(cmd, "type", indexTypeOptions...)
 	cmd.Flags().String("target", "", "Target type for metadata index: account or transaction")
+	cmdutil.RegisterEnumCompletion(cmd, "target", "account", "transaction")
```
NumaryBot: 🟡 **[minor] `ledger` target type omitted from metadata-index completion list**

The new shell completion for `--target` on the metadata-index path omits `ledger`, even though `resolveMetadataIndexFlags` / `ParseTargetType` and the interactive selector (`TargetTypeOptions()`) all accept `account`, `transaction`, and `ledger`. This causes the completion to drift from the actual accepted values and steers shell users away from a valid target.

**Suggestion:** Add `ledger` to the completion list for `--target` in both the create and drop metadata-index commands, keeping it consistent with `TargetTypeOptions()` and the parser.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1486#issuecomment-4867373034

cmd/ledgerctl/indexes/id_helpers.go
sylr: Fixed in 62ff84638. Removed `log-ledger` from the shared `indexTypeOptions` slice — there is no `LogBuiltinIndex` ledger enum value (only `LOG_BUILTIN_INDEX_DATE`), no `logBuiltinIndexID` helper, and no switch case in `runCreateIndex`/`runDropIndex`, so it was uncreatable. Since the slice backs both completion and the interactive selector, this also removes the pre-existing interactive-menu trap. Added a doc comment recording that every entry must have a matching switch case.

cmd/ledgerctl/indexes/create.go
sylr: Fixed in 62ff84638. The `--target` completion for `indexes create`/`drop` now uses `cmdutil.TargetTypeOptions()` (account, transaction, ledger), matching `ParseTargetType` and the interactive selector, and the flag help was updated to match.

cmd/ledgerctl/indexes/id_helpers.go
```diff
@@ -2,6 +2,23 @@ package indexes
...
+// shell completion so the two never drift. Every entry must have a matching
+// case in runCreateIndex/runDropIndex; do not add a value here before the
+// command handles it, or both the menu and completion will steer users into an
+// "invalid index type" error.
+var indexTypeOptions = []string{
+	"address",
+	"source-address",
+	"dest-address",
+	"metadata",
+	"reference",
```
NumaryBot, line 16: 🟠 **[major] Shell completion includes `log-ledger` index type that no command handles**

The shared completion list for `--type` includes `log-ledger`, but neither `runCreateIndex` nor `runDropIndex` has a switch case that builds a `log-ledger` IndexID. Selecting this value from completion leads directly to an 'invalid index type' error, actively guiding users into a failing invocation.

**Suggestion:** Either implement the missing `log-ledger` case in `runCreateIndex` and `runDropIndex`, or remove `log-ledger` from the completion list until those commands actually support it.

cmd/ledgerctl/indexes/create.go
```diff
@@ -47,7 +47,9 @@ Examples:
 
 	cmd.Flags().String("ledger", "", "Name of the ledger")
 	cmd.Flags().String("type", "", "Index type: address, source-address, dest-address, metadata")
-	cmd.Flags().String("target", "", "Target type for metadata index: account or transaction")
+	cmdutil.RegisterEnumCompletion(cmd, "type", indexTypeOptions...)
+	cmd.Flags().String("target", "", "Target type for metadata index: account, transaction, or ledger")
+	cmdutil.RegisterEnumCompletion(cmd, "target", cmdutil.TargetTypeOptions()...)
```
NumaryBot, line 52: 🟡 **[minor] `ledger` target type omitted from metadata-index `--target` completion**

The shell completion for `--target` on the metadata-index path omits `ledger`, even though `resolveMetadataIndexFlags` / `ParseTargetType` and the interactive selector (`TargetTypeOptions()`) all accept `account`, `transaction`, and `ledger`. This causes the completion to diverge from the actual accepted values and steers shell users away from a valid target.

**Suggestion:** Add `ledger` to the completion list for `--target` in both the create and drop metadata-index commands, keeping it consistent with `TargetTypeOptions()` and the parser.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1486#issuecomment-4867373034

PR #1488, opened 2026-07-02
fix(operator): preserve LedgerClusterAgent seed across LedgerService recreation

gfyrag:
## Summary

- Fix an issue where deleting and recreating a `LedgerService` invalidated every agent bundle handed out via `kubectl-ledger agents get-key --bundle`.
- Root cause: the `LedgerClusterAgent` reconcile GC'd every replica of the agent Secret when `desiredNamespaces` became empty (transient absence of matched services). The next reconcile found no existing replica and generated a fresh Ed25519 keypair, so previously-issued JWTs stopped verifying against the operator-managed `auth-keys.json` ConfigMap.
- Fix: skip the GC loop when `desiredNamespaces` is empty. Dormant replicas survive the transient state; the next matching reconcile reuses the existing seed via `canonicalKeyData`. Definitive cleanup remains on the finalizer (`handleDeletion`), which lists and deletes all replicas when the `LedgerClusterAgent` CR itself is deleted.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `golangci-lint run` — 0 issues
- [x] `just operator-pre-commit` (controller-gen + go mod tidy + go build)
- [x] Added `TestReconcile_AgentSeedSurvivesLedgerServiceRecreation` covering the full scenario: LedgerService created → seed generated → LedgerService deleted → dormant Secret preserved → LedgerService recreated → seed/keyID identical.
- [x] `TestReconcile_AgentOrphanCleanup` and `TestReconcile_AgentNoTargets` behavior verified unchanged (GC still runs when at least one target remains; fresh agents with no targets still don't produce a Secret).

### DIFF
--- a/misc/operator/cmd/operator/main.go
+++ b/misc/operator/cmd/operator/main.go
@@ -76,6 +76,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	operatorNamespace, err := controller.DiscoverOperatorNamespace()
+	if err != nil {
+		setupLog.Error(err, "unable to determine operator namespace")
+		os.Exit(1)
+	}
+
 	dynamicClient, err := dynamic.NewForConfig(cfg)
 	if err != nil {
 		setupLog.Error(err, "unable to create dynamic client")
@@ -94,8 +100,9 @@ func main() {
 	}
 
 	if err = (&controller.LedgerClusterAgentReconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
+		Client:            mgr.GetClient(),
+		Scheme:            mgr.GetScheme(),
+		OperatorNamespace: operatorNamespace,
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "LedgerClusterAgent")
 		os.Exit(1)
--- a/misc/operator/cmd/operator/main.go
+++ b/misc/operator/cmd/operator/main.go
@@ -47,6 +47,16 @@ func main() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 	utilruntime.Must(ledgerv1alpha1.AddToScheme(scheme))
 
+	// Discover the operator's own namespace up-front. When --watch-namespace
+	// is used, the manager cache must include the operator namespace too so
+	// canonical seed Secret reads/writes are not silently truncated to the
+	// watched namespace.
+	operatorNamespace, err := controller.DiscoverOperatorNamespace()
+	if err != nil {
+		setupLog.Error(err, "unable to determine operator namespace")
+		os.Exit(1)
+	}
+
 	mgrOpts := ctrl.Options{
 		Scheme: scheme,
 		Metrics: metricsserver.Options{
@@ -59,6 +69,9 @@ func main() {
 	if f.watchNamespace != "" {
 		mgrOpts.Cache.DefaultNamespaces = map[string]cache.Config{
 			f.watchNamespace: {},
+		}
+		if operatorNamespace != f.watchNamespace {
+			mgrOpts.Cache.DefaultNamespaces[operatorNamespace] = cache.Config{}
 		}
 	}
 
@@ -73,12 +86,6 @@ func main() {
 	clientset, err := kubernetes.NewForConfig(cfg)
 	if err != nil {
 		setupLog.Error(err, "unable to create kubernetes clientset")
-		os.Exit(1)
-	}
-
-	operatorNamespace, err := controller.DiscoverOperatorNamespace()
-	if err != nil {
-		setupLog.Error(err, "unable to determine operator namespace")
 		os.Exit(1)
 	}
 
--- a/misc/operator/cmd/operator/main.go
+++ b/misc/operator/cmd/operator/main.go
@@ -47,10 +47,11 @@ func main() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 	utilruntime.Must(ledgerv1alpha1.AddToScheme(scheme))
 
-	// Discover the operator's own namespace up-front. When --watch-namespace
-	// is used, the manager cache must include the operator namespace too so
-	// canonical seed Secret reads/writes are not silently truncated to the
-	// watched namespace.
+	// Discover the operator's own namespace up-front. Canonical seed
+	// Secrets live there and are accessed via an uncached APIReader (see
+	// LedgerClusterAgentReconciler.APIReader), so --watch-namespace scope
+	// stays exactly what the user asked for — the operator namespace is
+	// NOT added to the cache.
 	operatorNamespace, err := controller.DiscoverOperatorNamespace()
 	if err != nil {
 		setupLog.Error(err, "unable to determine operator namespace")
@@ -69,9 +70,6 @@ func main() {
 	if f.watchNamespace != "" {
 		mgrOpts.Cache.DefaultNamespaces = map[string]cache.Config{
 			f.watchNamespace: {},
-		}
-		if operatorNamespace != f.watchNamespace {
-			mgrOpts.Cache.DefaultNamespaces[operatorNamespace] = cache.Config{}
 		}
 	}
 
@@ -110,6 +108,7 @@ func main() {
 		Client:            mgr.GetClient(),
 		Scheme:            mgr.GetScheme(),
 		OperatorNamespace: operatorNamespace,
+		APIReader:         mgr.GetAPIReader(),
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "LedgerClusterAgent")
 		os.Exit(1)
--- a/misc/operator/internal/controller/ledgerclusteragent_controller.go
+++ b/misc/operator/internal/controller/ledgerclusteragent_controller.go
@@ -6,6 +6,7 @@ import (
 	"crypto/rand"
 	"crypto/sha256"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"sort"
 
@@ -28,6 +29,14 @@ import (
 const (
 	agentFinalizer = "ledger.formance.com/agent-keys"
 	agentNameLabel = "ledger.formance.com/agent-name"
+
+	// agentRoleLabel distinguishes the canonical seed Secret (living in the
+	// operator's namespace) from per-target replica Secrets. Only replicas
+	// are listed for placement / GC purposes; the canonical Secret is
+	// managed on its own path so it survives every target coming and going.
+	agentRoleLabel     = "ledger.formance.com/agent-role"
+	agentRoleReplica   = "replica"
+	agentRoleCanonical = "canonical"
 )
 
 // LedgerClusterAgentReconciler reconciles a LedgerClusterAgent object.
@@ -35,6 +44,12 @@ type LedgerClusterAgentReconciler struct {
 	client.Client
 
 	Scheme *runtime.Scheme
+
+	// OperatorNamespace is where the canonical seed Secret of every
+	// LedgerClusterAgent is stored. Injected at construction (from
+	// DiscoverOperatorNamespace in production, from a fixed namespace in
+	// envtest) so the reconciler does not depend on process-global state.
+	OperatorNamespace string
 }
 
 // +kubebuilder:rbac:groups=ledger.formance.com,resources=ledgerclusteragents,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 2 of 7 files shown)


misc/operator/internal/controller/ledgerclusteragent_controller.go
```diff
@@ -131,21 +131,31 @@ func (r *LedgerClusterAgentReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+	//
+	// When desiredNamespaces is empty (all matched services disappeared, no
+	// additional namespaces configured), we deliberately skip GC to preserve
+	// the seed across transient LedgerService deletion/recreation. Deleting
+	// every replica would destroy the canonical key material, and the next
+	// reconcile with a matching service would generate a fresh keypair —
+	// silently invalidating every bundle already handed out. The finalizer
+	// (handleDeletion) remains responsible for the definitive cleanup when
+	// the LedgerClusterAgent itself is deleted.
+	if len(desiredNamespaces) > 0 {
```
NumaryBot: 🟠 **[major] Delete stale replicas when the last target is removed**

When an agent previously distributed a Secret and the user removes the last target explicitly (for example by clearing the only `additionalNamespaces` entry, or changing the selector so it no longer matches any existing `LedgerService`), `desiredNamespaces` becomes empty and this guard skips orphan GC. That leaves the old Secret containing `seed.hex` in a namespace that is no longer in the spec until the whole `LedgerClusterAgent` is deleted, even though the status now reports no distributed refs. The preservation path needs to distinguish transient service disappearance from explicit target removal, or otherwise keep canonical seed material without leaving undesired namespace replicas behind.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1488#issuecomment-4867446964

misc/operator/internal/controller/ledgerclusteragent_controller.go
gfyrag: Adressé dans a05647e2d — le seed vit maintenant dans un Secret canonique dans le namespace de l'opérateur (découvert via POD_NAMESPACE + fallback SA file), les répliques par-namespace en sont de pures projections, et la GC agressive est restaurée. Plus de matériel qui traîne dans des namespaces hors scope, quelle que soit la manière dont les targets disparaissent.

misc/operator/internal/controller/ledgerclusteragent_controller.go
```diff
@@ -233,41 +278,71 @@ func (r *LedgerClusterAgentReconciler) resolveMatchedServices(ctx context.Contex
...
-// listAgentSecrets returns every Secret labeled as belonging to the given agent,
-// across all namespaces.
-func (r *LedgerClusterAgentReconciler) listAgentSecrets(ctx context.Context, agentName string) ([]corev1.Secret, error) {
+// listAgentReplicaSecrets returns every replica Secret belonging to the given
+// agent, across all namespaces. The canonical Secret is deliberately excluded
+// (filtered via agentRoleLabel) so replica placement and GC cannot touch it.
+func (r *LedgerClusterAgentReconciler) listAgentReplicaSecrets(ctx context.Context, agentName string) ([]corev1.Secret, error) {
 	var secrets corev1.SecretList
-	if err := r.List(ctx, &secrets, client.MatchingLabels{agentNameLabel: agentName}); err != nil {
+	if err := r.List(ctx, &secrets, client.MatchingLabels{
```
NumaryBot, line 292: 🔴 **[blocker] Upgrade path regenerates seed — legacy replicas lack new `agent-role=replica` label**

Agent Secrets created by a prior version of the operator carry only the `ledger.formance.com/agent-name` label and do not have the new `agent-role=replica` label. The replica-discovery selector therefore returns no results on upgrade, so the reconciler treats it as a fresh install, generates a new canonical seed, and overwrites the existing replica Secrets with new key material. This invalidates all previously issued bundles. The deletion path also misses legacy replicas, leaking them indefinitely.

**Suggestion:** Before generating new key material, fall back to the legacy selector (`ledger.formance.com/agent-name=<name>` without `agent-role=replica`) to detect pre-existing replica Secrets. If found, adopt them (add the new label via patch) and derive the canonical seed from the one already in use rather than generating a fresh one.

misc/operator/cmd/operator/main.go
```diff
@@ -76,6 +76,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	operatorNamespace, err := controller.DiscoverOperatorNamespace()
```
NumaryBot: 🟠 **[major] Operator namespace not included in watch-namespace cache, causing reconciliation failures**

For deployments that use `--watch-namespace` where the operator itself runs in a different namespace, the manager cache is built limited to the watched namespace(s) before the operator namespace is considered. The new canonical Secret is stored in `OperatorNamespace`, so any `Get`/`CreateOrUpdate` call on that Secret goes through a cache that does not include the operator namespace, causing reconciliation to fail with a cache-miss or permission error instead of distributing agent keys.

**Suggestion:** Discover `OperatorNamespace` before configuring the manager's `DefaultNamespaces` (or `NewCache` option) and explicitly add it to the set of watched namespaces. Alternatively, use an uncached client (e.g. `mgr.GetAPIReader()`) for all operations on the canonical Secret so they bypass the namespace-scoped cache.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1488#issuecomment-4867446964

paul-nicolas (approved):

misc/operator/internal/controller/ledgerclusteragent_controller.go
gfyrag: Adressé dans 07bb1f1e0 — `listAgentReplicaSecrets` filtre maintenant uniquement par `agentNameLabel` (le canonique est exclu par nom+namespace, ce qui découvre naturellement les répliques legacy). `ensureCanonicalSecret` prend un paramètre `existingReplicas` et, à la première création, appelle `adoptSeedFromReplica` qui reprend le seed d'une réplique legacy plutôt que d'en générer un nouveau. Tri déterministe par namespace/nom pour rendre le choix stable. Test intégration `TestReconcile_AgentUpgradeAdoptsLegacyReplicaSeed` couvre le scénario.

misc/operator/cmd/operator/main.go
gfyrag: Adressé dans 07bb1f1e0 — `DiscoverOperatorNamespace()` est appelé avant la construction du manager, et quand `--watch-namespace` est utilisé, le namespace opérateur est explicitement ajouté à `Cache.DefaultNamespaces` (sauf s'il est déjà égal au watch-namespace). Les ops sur le Secret canonique passent donc par un cache qui l'inclut.

misc/operator/internal/controller/ledgerclusteragent_controller.go
```diff
@@ -233,41 +270,127 @@ func (r *LedgerClusterAgentReconciler) resolveMatchedServices(ctx context.Contex
...
+// listAgentReplicaSecrets returns every replica Secret belonging to the given
+// agent, across all namespaces. The canonical Secret is identified by
+// name + namespace and excluded from the result — filtering by name (rather
+// than by an additional label) means Secrets created by pre-canonical
+// versions of the operator are still discovered, so upgrade adopts them
+// instead of orphaning them.
+func (r *LedgerClusterAgentReconciler) listAgentReplicaSecrets(ctx context.Context, agentName string) ([]corev1.Secret, error) {
 	var secrets corev1.SecretList
-	if err := r.List(ctx, &secrets, client.MatchingLabels{agentNameLabel: agentName}); err != nil {
+	if err := r.List(ctx, &secrets, client.MatchingLabels{
```
NumaryBot, line 292: 🔴 **[blocker] Upgrade path regenerates seed — legacy replicas lack new `agent-role=replica` label**

Agent Secrets created by a prior version of the operator carry only the `ledger.formance.com/agent-name` label and do not have the new `agent-role=replica` label. The replica-discovery selector therefore returns no results on upgrade, so the reconciler treats it as a fresh install, generates a new canonical seed, and overwrites existing replica Secrets with new key material. This invalidates all previously issued bundles. The deletion path also misses legacy replicas, leaking them indefinitely.

**Suggestion:** Before generating new key material, fall back to the legacy selector (`ledger.formance.com/agent-name=<name>` without `agent-role=replica`) to detect pre-existing replica Secrets. If found, adopt them by adding the new label via patch and derive the canonical seed from the one already in use rather than generating a fresh one.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1488#issuecomment-4867446964

misc/operator/cmd/operator/main.go
```diff
@@ -60,6 +70,9 @@ func main() {
 		mgrOpts.Cache.DefaultNamespaces = map[string]cache.Config{
 			f.watchNamespace: {},
 		}
+		if operatorNamespace != f.watchNamespace {
```
NumaryBot: 🟠 **[major] Operator namespace added to global watch cache, violating watch-namespace scoping**

When `--watch-namespace=tenant-a` is used and the operator runs in a different namespace, adding the operator namespace to `DefaultNamespaces` widens the cache and watches for every controller, not only canonical Secrets. This makes LedgerServices and other namespaced objects in the operator namespace visible/reconcilable despite the watch-namespace setting, potentially causing unexpected agent replica or auth-key distribution there. In addition, if the operator namespace is not included in the cache at all (pre-fix behavior), all Get/CreateOrUpdate calls on the canonical Secret fail with cache-miss or permission errors.

**Suggestion:** Use an uncached client (e.g. `mgr.GetAPIReader()`) or a per-object cache exemption for all operations on the canonical Secret, rather than adding the operator namespace to the global `DefaultNamespaces`. This confines the cache widening to only the canonical Secret operations.

misc/operator/internal/controller/ledgerclusteragent_controller.go
```diff
@@ -92,22 +99,37 @@ func (r *LedgerClusterAgentReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+	// per-target replicas are pure projections of its content. It is
+	// created lazily on first-ever target so an agent with no matching
+	// services never persists key material. Once it exists, it survives
+	// transient absence of every target — a LedgerService deletion no
+	// longer regenerates the seed.
 	var (
 		canonicalData map[string][]byte
 		refs          = make([]ledgerv1alpha1.SecretReference, 0, len(desiredNamespaces))
 	)
 	if len(desiredNamespaces) > 0 {
```
NumaryBot: 🟠 **[major] No-target upgrade edge case: canonical secret not bootstrapped before GC deletes legacy replicas**

When upgrading a pre-canonical agent while it currently has no desired namespaces (e.g., the LedgerService was deleted while the old operator was stopped), the guard skips `ensureCanonicalSecret` and the GC below deletes `existingReplicas`. The next matching service recreation finds no legacy seed to adopt and generates a new key, invalidating previously issued bundles.

**Suggestion:** If `existingReplicas` is non-empty, bootstrap the canonical secret from the legacy replica data before deleting replicas, even when there are no current targets.

misc/operator/internal/controller/ledgerclusteragent_controller.go
gfyrag: Cette finding est **stale** — le fix est déjà en place dans 07bb1f1e0 (et confirmé dans 487bf926e). `listAgentReplicaSecrets` filtre uniquement sur `agentNameLabel` (le canonique est exclu par nom+namespace), donc les répliques legacy sont bien découvertes. `ensureCanonicalSecret` reçoit `existingReplicas` et `adoptSeedFromReplica` reprend leur seed avant de générer du neuf. Test intégration `TestReconcile_AgentUpgradeAdoptsLegacyReplicaSeed` (+ `TestReconcile_AgentUpgradeAdoptsLegacySeedWithoutTargets` dans 487bf926e pour le cas sans target) couvrent les deux scénarios.

misc/operator/cmd/operator/main.go
gfyrag: Adressé dans 487bf926e — approche revue selon ta suggestion. Le namespace opérateur n'est plus ajouté à `Cache.DefaultNamespaces` (le scope `--watch-namespace` reste exact). Le reconciler prend maintenant un `APIReader client.Reader` (=`mgr.GetAPIReader()`) utilisé uniquement pour lire le Secret canonique. `ensureCanonicalSecret` est réécrit autour d'un Get-puis-Create explicite, plus de CreateOrUpdate. Writes vont via r.Client (les writes bypass le cache de toute façon).

misc/operator/internal/controller/ledgerclusteragent_controller.go
gfyrag: Adressé dans 487bf926e — la condition de bootstrap devient `len(desiredNamespaces) > 0 || len(existingReplicas) > 0`. Si des répliques legacy existent (upgrade sans target courant), le canonique est bootstrappé et adopte leur seed AVANT que la GC agressive ne les supprime. Nouveau test intégration `TestReconcile_AgentUpgradeAdoptsLegacySeedWithoutTargets` couvre exactement ce scénario (réplique legacy orpheline, aucun LedgerService qui match, seed préservé sur le canonique).

NumaryBot (commented):

Resolved 3 stale NumaryBot review threads (1 fixed, 2 outdated).

Summary: https://github.com/formancehq/ledger/pull/1488#issuecomment-4867446964

PR #1490, opened 2026-07-03
feat(replication): add account-address rewrite rules to mirroring pipelines

sylr:
> **Base note:** targeted at `main`. The change extends the replication feature (`internal/replication`), which does not exist on `release/v3.0` — that branch is a disjoint history with no common ancestor, so this could only be opened against `main`.

## What

Adds **account-address rewrite rules** to mirroring (replication) pipelines. A `Pipeline` can now carry an ordered list of regex rules (`{pattern, replacement}`) applied to every account address in the mirrored logs before they are pushed to the exporter — letting downstream consumers **drop or rename** address segments used purely as lock-avoidance shards in v2.

Example rule `{"pattern": "(:worker:\\d+)", "replacement": ""}` turns `payments:acme:worker:001:main` into `payments:acme:main`.

```json
POST /v2/{ledger}/pipelines
{
  "exporterID": "...",
  "addressRewriteRules": [
    { "pattern": "(:worker:\\d+)", "replacement": "" }
  ]
}
```

## How

- **Pure projection** — the source ledger, its stored logs and their hashes are never modified; only the mirrored stream is transformed.
- **Atomic rewriting** — postings, both post-commit volume maps (`PostCommitVolumes` / `PostCommitEffectiveVolumes`) and account-metadata keys are rewritten together, so `Transaction.MarshalJSON` → `SubtractPostings` never indexes a rewritten address into a stale volume key (which would panic on export).
- **Collision handling** — when two addresses collapse onto one, volumes are summed and metadata merged (value from the lexicographically-smallest source address wins).
- **Validation** — rule regexes are validated on both the HTTP and gRPC create paths.
- **Fail loud** — an invalid rewrite result fails the pipeline push loop rather than emitting a partially-rewritten log; the cursor is never advanced on failure.

## Touched surface

- Domain: `PipelineConfiguration.AddressRewriteRules` + `Validate()` (`internal/pipeline.go`)
- Transform: new `AddressRewriter` (`internal/replication/transform.go`)
- Pipeline loop, manager validation, gRPC proto + mapping
- Migration: `address_rewrite_rules jsonb` on `_system.pipelines`
- HTTP controller, OpenAPI specs (v2/v3 + merged) + regenerated API docs, and the Speakeasy client models

## Tests

- Unit: transform (all payload types, drop + rename, collision merge, invalid-result error, no source mutation, and **JSON marshaling** to exercise the volume-map path), `Validate`, and gRPC mapping round-trip (incl. nil-vs-empty).
- E2E (`it`-tagged): create a pipeline with a `:worker:\d+` rule, ingest a sharded address, assert the exported log shows the rewritten address.

## Notes for reviewers

- **Collision semantics** (merge/sum) is a deliberate default for the lock-shard use case — flag if you'd prefer rejecting collisions instead.
- `pkg/client` SDK models were hand-added to match Speakeasy output; CI's `speakeasy run` will reproduce them. The `it`-tagged E2E and the system-store JSONB round-trip require infra and were not run locally.


sylr:
Closing: built against `main`'s `internal/replication` subsystem by mistake. The mirroring feature for release/v3.0 lives in `internal/application/mirror` on a disjoint history — redoing the work there.

PR #1491, opened 2026-07-03
feat(mirror): rewrite account addresses during v2→v3 mirror translation

sylr:
## What

Adds optional **account-address rewrite rules** to a mirror ledger's `MirrorSourceConfig`. As the mirror worker ingests a v2 source and translates its logs into v3 orders, each rule (`{pattern, replacement}`, applied in order) rewrites every account address — letting segments v2 only carried as lock-avoidance shards be dropped or renamed during migration.

Example: rule `{"pattern": "(:worker:\\d+)", "replacement": ""}` turns `payments:acme:worker:001:main` into `payments:acme:main`.

```bash
# ledgerctl
ledgerctl ledgers create --name main --mode=mirror \
  --mirror-base-url=http://v2:8080 \
  --mirror-address-rewrite='(:worker:\d+)='
```
```jsonc
// HTTP: POST /{ledger}
{ "mode": "MIRROR", "mirrorSource": {
    "ledgerName": "main", "type": "http", "baseUrl": "http://v2:8080",
    "addressRewriteRules": [ { "pattern": "(:worker:\\d+)", "replacement": "" } ] } }
```

## How

- **Rewrite in the translator** (`internal/adapter/v2`, new `AddressRewriter`) at the three address-bearing choke points: posting source/destination, metadata account targets, and account-metadata map keys.
- **Nothing downstream changes** — coverage/preload and the FSM read the already-translated payload, and volumes are derived from the rewritten postings on apply.
- **Pure projection** — the source v2 ledger is never modified; only the leader translates/proposes, so followers apply identical addresses (no cross-node determinism concern beyond RE2 being deterministic).
- **Collisions** — when two source addresses collapse onto one, their account-metadata maps merge (value from the lexicographically-smallest source address wins).
- **Fail loud** — rule regexes are compile-checked at admission (`ErrMirrorAddressRewritePatternInvalid`); a rewrite that yields an invalid address fails the batch, so the cursor never advances and the worker retries (standard translation-error path).

## Touched surface

- Proto: `MirrorSourceConfig.address_rewrite_rules` + `AddressRewriteRule` (`misc/proto/common.proto`, regenerated).
- Translator threading + `AddressRewriter`; `Manager.reconcile → Worker → TranslateBatch`.
- Admission validation; `ledgerctl` create flag/parse/render; HTTP create-ledger handler; `openapi.yml`; architecture docs (`events-mirror/mirror.md`).

## Tests

- `AddressRewriter` (drop/rename/order/invalid-output/nil pass-through).
- Translator with rules: created + reverted postings, metadata targets (account rewritten, transaction untouched), account-metadata **collision merge**, invalid-result error.
- Admission (valid rules accepted, invalid regex rejected), `ledgerctl` flag parsing, HTTP handler + `mirrorSourceToProto`.

Build, `go vet`, and `golangci-lint` are clean on all changed packages.

## Notes for reviewers

- **Collision semantics** (merge, smallest-source-wins) is a deliberate default for the lock-shard use case — flag if you'd prefer rejecting collisions.
- Not added: an end-to-end cluster test (`tests/e2e/cluster`) exercising rewrite rules through a live mirror — it needs a running cluster and wasn't validated locally.

### DIFF
--- a/cmd/ledgerctl/ledgers/create.go
+++ b/cmd/ledgerctl/ledgers/create.go
@@ -48,6 +48,7 @@ func NewCreateCommand() *cobra.Command {
 	cmd.Flags().String("mirror-aws-iam-region", "", "Enable AWS RDS IAM authentication using the given region (for postgres source); credentials are taken from the ambient AWS chain (IRSA, instance profile, env)")
 	cmd.Flags().String("mirror-aws-iam-assume-role-arn", "", "Optional STS role ARN to assume before minting the RDS IAM token (cross-account / multi-tenant mirrors); requires --mirror-aws-iam-region")
 	cmd.Flags().Uint32("mirror-batch-size", 0, "Max logs per batch (0 = default 100)")
+	cmd.Flags().StringArray("mirror-address-rewrite", nil, "Account address rewrite rule as 'pattern=replacement' (RE2 regex applied during translation; empty replacement drops the match, e.g. '(:worker:\\d+)='; repeatable)")
 
 	return cmd
 }
--- a/cmd/ledgerctl/ledgers/display.go
+++ b/cmd/ledgerctl/ledgers/display.go
@@ -3,6 +3,7 @@ package ledgers
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/pterm/pterm"
 	"github.com/spf13/cobra"
@@ -63,6 +64,14 @@ func renderMirrorSource(src *commonpb.MirrorSourceConfig) {
 	if src.GetBatchSize() > 0 {
 		pterm.Printf("  Batch:   %d\n", src.GetBatchSize())
 	}
+
+	if rules := src.GetAddressRewriteRules(); len(rules) > 0 {
+		pterm.Printf("  Rewrites:\n")
+
+		for _, rule := range rules {
+			pterm.Printf("    %s → %q\n", rule.GetPattern(), rule.GetReplacement())
+		}
+	}
 }
 
 // renderMirrorSyncProgress displays mirror sync progress information.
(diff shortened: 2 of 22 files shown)


gfyrag (approved):
LGTM. Design propre : la réécriture au translate-time dans `internal/adapter/v2/` est le bon choke point (coverage/preload et FSM lisent le payload déjà réécrit — invariants #6/#8/#9 tranquilles), RE2 + `sort.Strings(accounts)` garantissent le déterminisme cross-node, et le fail-loud via `ValidateLedgerAccountAddress` + retry curseur est le bon chemin d'erreur. Défense-en-profondeur sur le pattern vide (17a4fc8) bien vue.

Quelques points non-bloquants à considérer (follow-up ou pas selon toi) :

**Sémantique**
- Les **valeurs de métadonnées** ne sont pas réécrites — si un user stocke une adresse dans une metadata value (ex. `{\"beneficiary\": \"payments:acme:worker:001:main\"}`), elle reste inchangée. Mérite une ligne dans `mirror.md`.
- Pas de garde-fou sur `world` : un pattern `^world$` → `treasury` passe la validation d'adresse et réattribue tous les débits de world. Footgun possible.
- Postings self-referential après collapse : si `worker:001 → worker:002` collapse sur `main → main`, le FSM devrait tolérer (delta volume nul) mais un test explicite serait rassurant.
- Règles gelées à la création (pas d'update path) — sans doute intentionnel, à confirmer si c'est acceptable pour le use case.

**Micro**
- `existing.Values` peut être nil au moment du merge quand la source "gagnante" a une metadata map vide — le lazy-init gère mais un commentaire aiderait.
- `$` dans replacement est interprété (`$1`, `$name`) — worth d'ajouter au help du flag CLI.

**Tests**
- L'e2e cluster manquant est déjà noté. Vu que c'est une projection pure au translate-time avec unit-tests solides sur les 4 types de logs, le risque est modéré.

Rien de bloquant. Documentation possible en follow-up.

PR #1493, opened 2026-07-03
refactor: homogenize source/destination naming to full words

sylr:
## Context

Throughout the transaction builtin address-index code, the **source** concept was spelled in full (`source`/`Source`/`SOURCE`) while its **destination** counterpart was abbreviated to `dest`/`Dest`/`DEST`. This normalizes every abbreviated `dest` paired with a full `source` to the full word `destination`.

## Changes (22 files, symmetric 96/96)

**Proto enum (root cause)**
- `TX_BUILTIN_INDEX_DEST_ADDRESS` → `TX_BUILTIN_INDEX_DESTINATION_ADDRESS` in `misc/proto/common.proto` (**value stays `5`** → binary/wire compat preserved), `common.pb.go` regenerated via `just generate-proto`.

**Internal Go identifiers**
- `PrefixDestAccountTx` → `PrefixDestinationAccountTx` (Pebble byte `0x06` unchanged)
- `WriteDestAccountTxMapping` → `WriteDestinationAccountTxMapping`
- Locals: `singleDest`→`singleDestination`, `indexDst`/`indexSrc`→`indexDestination`/`indexSource`, `dstExcluded`/`srcExcluded`→`destinationExcluded`/`sourceExcluded`
- Propagated the new enum symbol to all Go call sites

**User-facing (clean break)**
- CLI `--type dest-address` → `destination-address`
- Config field + tags `DestAddress`/`destAddress` → `DestinationAddress`/`destinationAddress`
- Docs: `docs/ops/cli.md`, `api-comparison.md`, `indexer.md`

**Left intentionally**: matched-pair `satx`/`datx` Pebble mnemonics, generic `src`/`dst` copy helpers, operator backup `Destination`, `MirrorSource`.

## ⚠️ Compatibility note

The enum name is embedded in persisted index canonical keys via `indexes.Canonical()`, so the on-disk key for a **destination-address** index changes. Wire/binary proto compat is preserved (value `5`), but any existing deployment with a destination-address index will need to **rebuild** that index. Accepted per dev-mode scope.

## Verification

- `go build ./...` ✅
- `just lint` → 0 issues ✅
- Targeted tests (indexes, readstore, query, indexbuilder, ledgerctl, commonpb) ✅
- Regenerated proto diff is only the enum rename + gofmt realignment; `go.mod`/`go.sum` untouched.

### DIFF
--- a/cmd/ledgerctl/ledgers/config_model.go
+++ b/cmd/ledgerctl/ledgers/config_model.go
@@ -44,12 +44,12 @@ type EditableMetaField struct {
 
 // EditableIndexes represents the set of enabled builtin indexes.
 type EditableIndexes struct {
-	Reference     bool `json:"reference"     yaml:"reference"`
-	Timestamp     bool `json:"timestamp"     yaml:"timestamp"`
-	Address       bool `json:"address"       yaml:"address"`
-	SourceAddress bool `json:"sourceAddress" yaml:"sourceAddress"`
-	DestAddress   bool `json:"destAddress"   yaml:"destAddress"`
-	InsertedAt    bool `json:"insertedAt"    yaml:"insertedAt"`
+	Reference          bool `json:"reference"          yaml:"reference"`
+	Timestamp          bool `json:"timestamp"          yaml:"timestamp"`
+	Address            bool `json:"address"            yaml:"address"`
+	SourceAddress      bool `json:"sourceAddress"      yaml:"sourceAddress"`
+	DestinationAddress bool `json:"destinationAddress" yaml:"destinationAddress"`
+	InsertedAt         bool `json:"insertedAt"         yaml:"insertedAt"`
 }
 
 // EditablePreparedQuery is the editable subset of a prepared query.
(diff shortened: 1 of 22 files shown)


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1493#issuecomment-4876452293

cmd/ledgerctl/ledgers/config_model.go
```diff
@@ -44,12 +44,12 @@ type EditableMetaField struct {
...
-	Timestamp     bool `json:"timestamp"     yaml:"timestamp"`
-	Address       bool `json:"address"       yaml:"address"`
-	SourceAddress bool `json:"sourceAddress" yaml:"sourceAddress"`
-	DestAddress   bool `json:"destAddress"   yaml:"destAddress"`
-	InsertedAt    bool `json:"insertedAt"    yaml:"insertedAt"`
+	Reference          bool `json:"reference"          yaml:"reference"`
+	Timestamp          bool `json:"timestamp"          yaml:"timestamp"`
+	Address            bool `json:"address"            yaml:"address"`
+	SourceAddress      bool `json:"sourceAddress"      yaml:"sourceAddress"`
+	DestinationAddress bool `json:"destinationAddress" yaml:"destinationAddress"`
```
NumaryBot, line 51: 🟠 **[major] Reject legacy destAddress in configs**

When a user applies a config file exported before this rename with `indexes.destAddress: true`, both `json.Unmarshal` and `yaml.Unmarshal` ignore that now-unknown field, so `DestinationAddress` stays false. If the current ledger has the destination index enabled, `ComputeDiff` will then plan a `DropIndex` even though the manifest intended to keep it; if it is disabled, the requested create is silently skipped. Please either accept `destAddress` as a deprecated alias or explicitly reject it during config parsing instead of treating it as false.

gfyrag (approved):

PR #548, opened 2024-11-04
fix: storage corruption on null metadata

gfyrag:

PR #1494, opened 2026-07-03
feat(ledgerctl): add --count and --batch-size to transactions create (backport #548)

sylr:
Backport of formancehq/ledger-v3-poc#548 onto `release/v3.0`.

Adds two flags to `ledgerctl transactions create`:

- `--count` (default `1`) — send the same transaction N times.
- `--batch-size` (default `1`) — bundle up to N transactions per `Apply` request.

Behaviour:
- Each transaction is built from a freshly constructed request via a `newRequest` closure.
- In bulk mode (`--count > 1`) a summary is printed instead of each transaction's full detail; structured (`--json`/`--yaml`) output becomes the array of created transactions.
- Inputs are validated: `--count` and `--batch-size` must each be `>= 1`.
- `--reference` is rejected with `--count > 1`, since transaction references must be unique.
- `--timeout` is scoped per `Apply` request (via an `applyBatch` closure that builds a fresh context) rather than cumulatively across batches.

### Backport adaptation

The poc branch used the lower-level `cmdutil.BuildEnvelopes` + manual `ApplyRequest{Envelopes: ...}`. `release/v3.0` has since consolidated request construction/signing into `cmdutil.BuildApplyRequest(cmd, requests...)`, which signs the whole `ApplyBatch` as a unit. The batch loop was adapted to call `BuildApplyRequest(cmd, requests...)` and `applyBatch(applyReq)` accordingly — behaviour is otherwise identical.

Commits are cherry-picked 1:1 from the original PR (with the above conflict resolution folded into the relevant commits).

Docs (`docs/ops/cli.md`) updated with the new flags and examples.

### DIFF
--- a/cmd/ledgerctl/cmdutil/signing.go
+++ b/cmd/ledgerctl/cmdutil/signing.go
@@ -79,16 +79,27 @@ func VerifyResponseSignatures(cmd *cobra.Command, logs []*commonpb.Log) error {
 // key is configured on the command flags, unsigned otherwise. Signing the batch
 // authenticates its composition and ordering.
 func BuildApplyRequest(cmd *cobra.Command, requests ...*servicepb.Request) (*servicepb.ApplyRequest, error) {
+	return BuildIdempotentApplyRequest(cmd, "", requests...)
+}
+
+// BuildIdempotentApplyRequest is BuildApplyRequest with a batch-level
+// idempotency key. The key is carried on the ApplyBatch (and therefore covered
+// by the signature when signing is enabled), so the server can deduplicate a
+// batch that a transport retry replays byte-for-byte — notably the gRPC
+// UNAVAILABLE retry policy, which would otherwise re-apply a batch that had
+// already committed before the client observed the failure. Pass a value that
+// is stable across retries of the same batch but unique across distinct batches.
+func BuildIdempotentApplyRequest(cmd *cobra.Command, idempotencyKey string, requests ...*servicepb.Request) (*servicepb.ApplyRequest, error) {
 	keyID, privKey, err := LoadSigningKey(cmd)
 	if err != nil {
 		return nil, err
 	}
 
 	if privKey == nil {
-		return servicepb.UnsignedApplyRequest("", requests...), nil
+		return servicepb.UnsignedApplyRequest(idempotencyKey, requests...), nil
 	}
 
-	sb, err := signing.Sign(&servicepb.ApplyBatch{Requests: requests}, keyID, privKey)
+	sb, err := signing.Sign(&servicepb.ApplyBatch{Requests: requests, IdempotencyKey: idempotencyKey}, keyID, privKey)
 	if err != nil {
 		return nil, fmt.Errorf("failed to sign batch: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


gfyrag:
Re-reviewed current head `95debf870210915905a1f2e2e84c896187b581f5`.

P1: `--count` can create more transactions than requested after an automatic retry.

`ledgerctl` installs a gRPC retry policy for `UNAVAILABLE` with up to 50 attempts. The new loop sends every batch without an idempotency key, and `--reference` is rejected for `--count > 1`, so there is no per-transaction uniqueness guard either. If a batch commits but the client observes `UNAVAILABLE`, the retry can submit the same batch again as fresh transactions. That makes `ledgerctl transactions create --count N` able to over-apply, especially with `--batch-size > 1`.

Please assign a stable idempotency key per Apply batch before `client.Apply`, so transport retries replay the same batch outcome instead of creating another batch.

What looks resolved in the current diff:

- `--reference` is rejected with `--count > 1`.
- Timeout is scoped per Apply batch.
- `--batch` was renamed to `--batch-size` in code/help/docs.

Checks run:

- `go test ./cmd/ledgerctl/...`
- Targeted semantic tests for `internal/domain/processing`, `internal/application/admission`, and `internal/infra/state`
- `GOROOT= go build ./...`
- `go run ./cmd/ledgerctl transactions create --help`
- `nix develop --command bash -c "just pre-commit"` failed during `go generate` with Go toolchain mismatch; generated proto files were restored and the review worktree is clean.

sylr:
@gfyrag good catch on the retry over-apply — fixed in f9e582886.

Each Apply batch now carries a fresh UUID idempotency key, generated once before the (retried) `client.Apply`. Since the gRPC UNAVAILABLE retry replays byte-identical bytes, a batch that committed before the client saw the failure now replays its frozen outcome instead of being re-applied; distinct batches keep distinct keys so each still applies exactly once. `--count 1` runs a single iteration, so plain single creates gain the same retry-safety.

Mechanics: the key is set on `ApplyBatch.IdempotencyKey` (so it is covered by the signature when signing is enabled) via a new `cmdutil.BuildIdempotentApplyRequest` helper; `BuildApplyRequest` delegates to it with an empty key, leaving the ~30 other call sites untouched.

I kept `--reference` rejected for `--count > 1` as before — it is orthogonal (business-level tx uniqueness vs batch-level transport dedup).

Note: this fix is specific to the backport; the upstream poc PR #548 still uses the lower-level envelope API and has the same gap.

gfyrag (approved):
Re-reviewed current head `f9e5828860bd960400b6091adfdc253cba597bbf`. No blocking findings from my pass; the per-batch idempotency-key issue is fixed and the CLI/docs behavior looks coherent.

gfyrag (changes requested):
Rejecting this PR per maintainer direction. This should not move forward in its current form.

flemzord (changes requested):
Revalidated live head `f9e58288`. One new correctness/runtime blocker remains in the bulk-create response path: each Apply batch contains `n` requests, but the loop only validates the type of the logs it receives and never requires `len(resp.GetLogs()) == n`. A partial or over-complete response therefore continues through later batches and exits successfully with `Created M` / a structured array whose size differs from the requested `--count N` (for example, 4 logs then 5 logs for `--count 10 --batch-size 5` reports success with 9). The Apply contract is one log per request, so this is an impossible-by-design cardinality violation and must fail loudly rather than return a false success. Please validate the per-batch cardinality before appending results and add a regression for a mismatched response count. I found no existing NumaryBot, CodeRabbit, or human thread covering this finding.

Issue #491, opened 2024-06-13
StartTime property not working


cv65kr:
**Describe the bug**
According to openapi schema https://github.com/formancehq/ledger/blob/main/openapi/v1.yaml#L516

We should use `startTime` but in code https://github.com/formancehq/ledger/blob/main/internal/api/v1/controllers_transactions.go#L59 is not supported.

Another thing is 
```
if startTime := r.URL.Query().Get("start_time"); startTime != "" {
		clauses = append(clauses, query.Gte("date", startTime))
	}
```

Column `date` not exists

**To Reproduce**
Make request to transaction with time frames

**Expected behavior**
Transactions are filtered

**Environment (please complete the following information):**
 Main master of ledger

PR #1495, opened 2026-07-03
feat(ledgerctl): emit OpenTelemetry traces over gRPC (backport #491)

sylr:
Backport of formancehq/ledger-v3-poc#491 onto `release/v3.0`.

`ledgerctl` now generates real OpenTelemetry traces and exports them to any OTLP backend, configured entirely through the standard OpenTelemetry SDK environment variables — no CLI flags.

- Each invocation opens a **root span** named after the command path (e.g. `ledgerctl transactions get`).
- Every gRPC call is recorded as a **child span** via `otelgrpc.NewClientHandler()`, and the W3C `traceparent` is propagated so server-side spans join the same trace.
- The root span is annotated with `host.name`, `host.user`, and `ledgerctl.command`.
- Export only activates when an OTLP endpoint is configured (`OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_TRACES_EXPORTER=otlp`), so default usage never dials a collector.
- Shell completion invocations skip tracing entirely.
- `main()` is refactored into `run() int` so buffered spans are force-flushed (bounded timeout) on both success and error paths before exit.
- Replaces the hand-rolled `traceparent` interceptors (`tracing.go`, deleted) which injected a header but exported nothing.

### Backport adaptation

`release/v3.0` obtains the CLI version via the `internal/pkg/version` package rather than the poc's package-level `var version` string, so the `SetupTracing(ctx, ...)` call was adapted to pass `version.Get().Version`. The `go.mod` change promotes two already-present OTLP trace exporter deps from indirect to direct (no `go.sum` change needed).

Commits are cherry-picked 1:1 from the original PR (with the above adaptation folded into the first commit).


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1495#issuecomment-4876591376

cmd/ledgerctl/cmdutil/otel.go
```diff
@@ -0,0 +1,239 @@
...
+		protocol = strings.ToLower(strings.TrimSpace(os.Getenv("OTEL_EXPORTER_OTLP_PROTOCOL")))
+	}
+	if protocol == "" {
+		protocol = "grpc"
+	}
+
+	switch protocol {
+	case "grpc":
+		return otlptracegrpc.New(ctx)
+	case "http/protobuf", "http/json", "http":
```
NumaryBot: 🟠 **[major] Reject unsupported http/json protocol**

When a user sets `OTEL_EXPORTER_OTLP_PROTOCOL=http/json` or `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=http/json`, this branch still constructs `otlptracehttp`, which emits OTLP over HTTP/protobuf rather than JSON. That silently ignores a valid-looking protocol setting and sends traces with the wrong encoding/content type, so JSON-only collectors or proxies will drop them; either implement JSON export or reject `http/json` like the error message already implies.

sylr: Fixed in 375539911: `http/json` is now rejected with a clear error rather than routed to the protobuf HTTP exporter. The OTel Go SDK has no HTTP/JSON trace exporter, and the docs / accepted-protocol error already list only `grpc` and `http/protobuf`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1495#issuecomment-4876591376

gfyrag (approved):

PR #1498, opened 2026-07-06
feat(node): switch to etcd/raft AsyncStorageWrites, align raft.Applied with FSM-applied

gfyrag:
## Summary

Switches the Raft loop to etcd/raft's `AsyncStorageWrites` and reworks the response plumbing so that **`raft.Applied` is only advanced once the FSM batch is durably committed to Pebble**. This fixes a long-standing design compromise where `raft.Applied` was allowed to run ahead of the FSM-persisted state — safe (WAL was durable, restart recovers correctly via `query.ReadLastAppliedIndex`), but not compliant with the etcd/raft contract that `Applied` means "the state machine has applied this entry".

## What the sync path was doing before

In `finishReady` the orchestrate goroutine called `applier.Submit(rd.CommittedEntries, ...)` — a non-blocking channel enqueue — and immediately followed with `rawNode.Advance(rd)`, which bumped `raft.raftLog.applied` to the last committed index. The applier goroutine only picked up the batch afterwards, ran `applyEntriesPipelined`, and eventually `runCommitter` performed the actual `CommitPreparedBatch` (Pebble fsync). Net effect: `raft.Applied` sat one or more batches ahead of the durable FSM state at essentially all times.

Downstream, the codebase carried explicit workarounds for that gap — a dedicated `RaftStatus.last_persisted_index` proto field, `WaitForApplied` gating on the FSM's own `lastPersistedIndex` cursor rather than `raft.Applied`, an antithesis oracle documenting exactly why it can't trust `applied`, and comments across `Machine.LastAppliedIndex` / `publishApplied` / `GetClusterState` explaining the divergence.

## What this PR changes

- `raft.Config.AsyncStorageWrites = true`. `rawNode.Advance` is no longer called (it panics under this mode).
- `rd.Messages` is split via a pure `splitReadyMessages` helper: `LocalAppendThread` responses are routed self→`localResponseCh` (Step locally on the orchestrate goroutine) / peer→`transport.Send`; `LocalApplyThread` responses ride on `readyResult.applyResponses`; everything else is outbound.
- The Applier owns a required `LocalResponses` sink (fx provider `NewLocalResponses`, injected into both `NewApplier` and `NewNode` — constructor rejects a nil sink). Responses ride down `applyEntriesToFSM` → `applyEntriesPipelined` → `submitAsyncCommit` → `commitWork`. `runCommitter` fires them on the sink **only** after `CommitPreparedBatch` succeeds. On commit failure the response is intentionally dropped, `raft.Applied` stays behind, and the entries reappear in `CommittedEntries` on the next boot (`raft.Config.Applied` is re-derived from Pebble).
- The `applier→committer` pipeline is preserved: prepare(N+1) still overlaps with commit(N), because the fire happens on runCommitter (post-commit) rather than on the applier goroutine (post-submit-drain).
- Response-send select listens on both `ctx.Done()` and the applier's stop channel, matching the real `node.Stop` shutdown flow (which closes the task stop channels but does not cancel the runCtx).
- Non-`statusNormal` (spool) branch fires responses eagerly: entries are staged durably in the spool and will re-apply via unspool without further raft involvement.

The pre-existing workaround surfaces have been refreshed in place — they still describe the correct behavior (e.g. `WaitForApplied` still gates on `lastPersistedIndex` because that IS the semantic contract for "Pebble is durable at N", not because `raft.Applied` is unreliable). See the follow-up commit that walks through every `applied`/`last_persisted_index` comment site.

## Perf note

On this codebase's current workload (FSM-apply bound), the end-to-end propose→reply latency **does not improve meaningfully**: the WAL fsync was not on the critical path, so parallelizing it with peer broadcast doesn't buy end-to-end latency. Verified on `ledger-v3-perf/devenv-gfyrag`. The refactor is kept because the correctness win (Applied ↔ FSM-applied alignment) is a real invariant improvement, and the async-storage plumbing is a prerequisite for any future work that parallelizes apply itself.

## Findings addressed during review

- `af4915f6` / `41749f9676`: MsgStorageApplyResp fired before pipelined commit completed → responses now ride commitWork and fire from runCommitter after `CommitPreparedBatch`.
- `34540caa` / `9047f08a`: early `return` on ctx.Done skipped `work.done <- err` → return removed, fall-through preserves the unconditional `work.done` signal.
- `70740916`: ctx.Done alone insufficient during graceful shutdown (node.Stop closes task stop channels but leaves runCtx alive) → applier's stop channel threaded into runCommitter's select.

Stale bot re-flags on the same code paths (`abbc8945`, `130bdfe4`, `a3dc759a`) were replied to inline with pointers to the applicable commit and rationale (last one — snapshot-only Ready dropping apply responses — is a bogus premise: etcd/raft only emits MsgStorageApply when `CommittedEntries > 0`).

## Test plan

- [x] `just test` (unit, race) — all packages OK
- [x] `just test-scenarios` — OK
- [x] `just test-e2e` (business + cluster) — OK
- [x] Added `applier_async_storage_test.go`: fires-after-commit (timing assertion), spool-branch eager fire, nil-sink no longer accepted at construction, empty-responses no-fire, multi-batch ordering, and the shutdown-deadlock regression guard for finding 70740916.
- [x] Added `split_ready_messages_test.go`: pure-function coverage of the rd.Messages classification (empty, outbound passthrough, self/peer split — the invariant that broke cluster formation on the first attempt, apply-thread no-split, realistic leader Ready, follower all-peer Ready).
- [ ] Antithesis run — not yet triggered on this branch.


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1498#issuecomment-4890764481

internal/infra/node/node.go
```diff
@@ -1322,14 +1368,60 @@ func (node *Node) finishReady(result readyResult, stop chan struct{}) error {
...
+//     path), as well as regular MsgApp / MsgHeartbeat / MsgVote already in
+//     rd.Messages. All flow out through the transport.
+//
+// Pure function; called by processReady but split out so the self/peer
+// split invariant (missing it broke cluster formation on the first attempt)
+// stays testable without spinning up a full Node.
+func splitReadyMessages(nodeID uint64, msgs []raftpb.Message) (appendResponses, applyResponses, outbound []raftpb.Message) {
+	for _, m := range msgs {
+		switch m.To {
+		case raft.LocalAppendThread:
```
NumaryBot, line 1414: 🔴 **[blocker] Process async storage message payloads before responding**

When `AsyncStorageWrites` is enabled, the `MsgStorageAppend`/`MsgStorageApply` messages carry the HardState/entries/snapshot and committed entries that must be written/applied; the Ready legacy fields are not the source of that work. This helper only extracts `m.Responses` and drops the storage message payloads, while `processReady` still writes/applies `rd.*`, so a Ready with new log or committed entries can be acknowledged without actually persisting/applying the entries.

internal/infra/node/applier.go
```diff
@@ -1187,11 +1283,31 @@ func (a *Applier) applyEntriesToFSM(ctx context.Context, confState *raftpb.ConfS
...
+		if !responsesAttached && len(responses) > 0 {
+			select {
+			case a.responseSink <- responses:
+			case <-ctx.Done():
+				if checkpointErr != nil {
+					return checkpointErr
+				}
+
+				return ctx.Err()
+			}
```
Azorlogh, line 1345: missing `case <-stop:` ?

gfyrag: Good catch — fixed in c8c326134. Same class of bug as finding 70740916 that I closed on `runCommitter` last iteration, but this fallback path in `applyEntriesToFSM` (checkpoint-mid-batch case) still relied only on `ctx.Done()` — which doesn't fire during `node.Stop` since it only closes task stop channels.

Threaded `stop` through `applyEntriesToFSM`, added `case <-stop` to the select. On shutdown we drop the response and return; `raft.Applied` stays behind, entries re-emit on next boot, re-apply is idempotent.

internal/infra/node/node.go
NumaryBot, line 1414: 🔴 **[blocker] Async-storage message payloads dropped — persistence/application bypassed**

When AsyncStorageWrites is enabled, MsgStorageAppend and MsgStorageApply are the authoritative carriers of HardState/entries/snapshot and committed entries that must be persisted/applied before their embedded responses are stepped back into raft. The helper at this line only extracts m.Responses and discards the storage payloads, while processReady/finishReady still consult the legacy rd.* fields. This means a Ready cycle can be fully acknowledged without the WAL writes or FSM application having occurred, silently losing log entries or advancing Applied over unapplied state.

**Suggestion:** For each MsgStorageAppend, write the embedded HardState, Entries, and Snapshot to stable storage before processing m.Responses. For each MsgStorageApply, apply the embedded CommittedEntries to the FSM before processing m.Responses. Only then step the responses back into the raft node, mirroring the pattern shown in the etcd/raft AsyncStorageWrites documentation and examples.

internal/infra/node/applier.go
```diff
@@ -1187,11 +1283,36 @@ func (a *Applier) applyEntriesToFSM(ctx context.Context, confState *raftpb.ConfS
...
+					return checkpointErr
+				}
+
+				return ctx.Err()
+			case <-stop:
+				// Shutdown races us; drop the response (raft.Applied stays
+				// behind, entries re-emit on next boot). Same escape hatch
+				// as the runCommitter response send (finding 70740916).
+				return checkpointErr
+			}
```
NumaryBot, line 1345: 🟡 **[minor] Missing stop-channel case in select loop**

The select loop at this location appears to lack a `case <-stop:` branch, which means the goroutine may not exit cleanly when the node is shut down, leading to a goroutine leak.

**Suggestion:** Add a `case <-stop: return` (or equivalent context-cancellation case) to the select so the goroutine terminates promptly when the node stops.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1498#issuecomment-4890764481

internal/infra/node/node.go
gfyrag: Stale finding — the premise "processReady/finishReady still consult the legacy rd.* fields" is correct and deliberate; the conclusion "storage payloads are dropped" is wrong.

Under AsyncStorageWrites, etcd/raft's rawnode.go populates BOTH representations from the same source data:

- \`rd.HardState\`, \`rd.Entries\`, \`rd.Snapshot\`, \`rd.CommittedEntries\` — populated unconditionally at [rawnode.go:144-148](https://github.com/etcd-io/raft/blob/v3.6.0/rawnode.go#L144-L148).
- \`MsgStorageAppend\` / \`MsgStorageApply\` — appended to \`rd.Messages\` at [rawnode.go:165-186](https://github.com/etcd-io/raft/blob/v3.6.0/rawnode.go#L165-L186), carrying the SAME Entries/HardState/Snapshot/CommittedEntries plus the response chain.

We use the legacy fields for the actual work — see the `wal.Append(rd.HardState, rd.Entries)`, snapshot install via `rd.Snapshot`, and `applier.Submit(rd.CommittedEntries, ...)` calls that live in the same processReady function that calls `splitReadyMessages`. The helper only harvests the response chains from the messages, because those responses are the async-mode-only carrier that must be Step()-ed back into rawNode after the durable write / commit — the responses are the one thing that isn't already in the legacy rd.* fields.

This mirroring is intentional and documented in the PR body (see "What this PR changes"). The tests `TestSplitReadyMessages_*` also lock in that the helper's job is exclusively response classification, not payload processing.

internal/infra/node/applier.go
gfyrag: Stale — the finding points at line 1312 which is the closing \`}\` of the select. The \`case <-stop:\` branch was added at line 1307 in c8c32613 (addressing @Azorlogh's inline). Direct link: https://github.com/formancehq/ledger/blob/c8c32613465a7214ab31bfff90ab321bda07a17a/internal/infra/node/applier.go#L1298-L1312

Azorlogh (approved):

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1498#issuecomment-4890764481

internal/infra/node/node.go
```diff
@@ -1877,20 +2007,24 @@ func (node *Node) GetClusterState(ctx context.Context) (*clusterpb.ClusterState,
...
+	// CommitPreparedBatch call and advance in tight lockstep: `Applied`
+	// arrives a channel-hop + orchestrate Step after `LastPersistedIndex`,
+	// never before. So `LastPersistedIndex >= Applied` in this snapshot.
 	//
 	// Anything that reads from Pebble (stale-consistency GetAccount, test
-	// oracles, cross-node identity comparisons) MUST gate on
-	// LastPersistedIndex — that is the only cursor that guarantees a Pebble
-	// read will see entries up to that index. Gating on Applied races the
-	// apply pipeline and returns stale data.
+	// oracles, cross-node identity comparisons) SHOULD still gate on
```
NumaryBot, line 2023: 🟡 **[minor] Comment incorrectly implies Applied is safe for Pebble-read freshness during spooling windows**

When the applier is not in statusNormal (e.g. during checkpoint/snapshot gating or out-of-sync spooling), MsgStorageApplyResp is acknowledged immediately after spool.AppendCommittedEntries without calling Machine.publishApplied. This means raft.Status.Applied can advance ahead of Pebble's actual applied state during those spooling windows. The code comment at this location states that gating on Applied works and that LastPersistedIndex >= Applied, which can mislead callers/consumers into issuing Pebble reads against a stale index.

**Suggestion:** Update the comment to preserve the stronger guidance: gate Pebble reads on lastPersistedIndex rather than raft.Status.Applied. Clarify that during spooling windows Applied may race ahead of Pebble, so lastPersistedIndex remains the authoritative lower-bound for read freshness.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1498#issuecomment-4890764481

internal/infra/node/node.go
```diff
@@ -1877,20 +2007,24 @@ func (node *Node) GetClusterState(ctx context.Context) (*clusterpb.ClusterState,
...
 	// `LastPersistedIndex` is the durable FSM-side cursor: bumped by
-	// Machine.publishApplied on the committer goroutine after
-	// pb.batch.Commit() returns. The two advance independently on different
-	// goroutines (orchestrate vs committer), so a single snapshot can
-	// observe either cursor temporarily ahead of the other — do NOT assume
-	// LastPersistedIndex <= Applied.
+	// Machine.publishApplied inside CommitPreparedBatch, immediately after
+	// pb.batch.Commit() returns and before runCommitter emits the response.
+	//
+	// Under AsyncStorageWrites both are downstream of the same
```
NumaryBot, line 2017: 🟡 **[minor] Comment incorrectly implies Applied is always safe for Pebble-read freshness during spooling windows**

During non-normal applier states (e.g. checkpoint or out-of-sync gating), MsgStorageApplyResp is acknowledged immediately after spool.AppendCommittedEntries without calling Machine.publishApplied. This means raft.Status.Applied can advance ahead of Pebble's actual applied state during those spooling windows. A comment in this area states that LastPersistedIndex >= Applied and that gating on Applied is sufficient for Pebble-read freshness, which can mislead callers/consumers of GetClusterState into issuing Pebble reads against a stale index.

**Suggestion:** Update the comment to clarify that during spooling windows Applied may race ahead of Pebble's persisted state. Direct readers to gate Pebble reads on lastPersistedIndex rather than raft.Status.Applied, and note that lastPersistedIndex is the authoritative lower-bound for read freshness.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1498#issuecomment-4890764481

internal/infra/node/node.go
NumaryBot, line 2023: 🟡 **[minor] Comment incorrectly implies Applied is safe for Pebble-read freshness during spooling windows**

During checkpoint/snapshot gating or out-of-sync spooling windows, MsgStorageApplyResp is acknowledged immediately after spool.AppendCommittedEntries without calling Machine.publishApplied. This means raft.Status.Applied can advance ahead of Pebble's actual applied state during those windows. The existing comment implies that gating on Applied is sufficient for Pebble-read freshness and that lastPersistedIndex >= Applied, which can mislead callers into issuing reads against a stale index.

**Suggestion:** Update the comment to clarify that during spooling windows Applied may race ahead of Pebble's persisted state, and that lastPersistedIndex remains the authoritative lower-bound for read freshness. Direct readers to gate Pebble reads on lastPersistedIndex rather than raft.Status.Applied.

internal/infra/node/applier_async_storage_test.go
```diff
@@ -0,0 +1,351 @@
...
+	runDone := make(chan error, 1)
+	go func() { runDone <- setup.applier.Run(ctx, setup.stop) }()
+
+	entry, _ := makeCreateLedgerEntry(t, 1, "async-full-sink")
+	resp := makeApplyResp(1, entry.Index)
+	setup.applier.Submit([]raftpb.Entry{entry}, setup.confState, []raftpb.Message{resp}, setup.stop)
+
+	// Give runCommitter time to fire the response — it will block on the
+	// unbuffered send. If the deadlock regressed, the test hangs here.
+	time.Sleep(200 * time.Millisecond)
```
NumaryBot, line 315: 🟡 **[minor] Fixed sleep in shutdown regression test makes coverage of the deadlock path unreliable**

The fixed `time.Sleep` is the only synchronization ensuring `runCommitter` has reached the blocked response send before `stop` is closed. On a slow or heavily loaded CI runner, `stop` can be closed before the targeted code path is exercised, causing the test to pass without actually covering the deadlock regression it was written to guard against.

**Suggestion:** Replace the fixed sleep with an observable synchronization mechanism (e.g., a channel signal written just before the blocking point, or a `require.Eventually` condition that checks a detectable state) so the test reliably covers the intended scenario regardless of scheduler timing.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1498#issuecomment-4890764481

internal/infra/node/node.go
NumaryBot, line 2017: 🟡 **[minor] Comment implies Applied is always safe for Pebble-read freshness during spooling windows**

During non-normal applier states (checkpoint/snapshot gating or out-of-sync spooling), MsgStorageApplyResp is acknowledged immediately after spool.AppendCommittedEntries without calling Machine.publishApplied. This means raft.Status.Applied can advance ahead of Pebble's actual applied state during those spooling windows. A comment in this area implies that LastPersistedIndex >= Applied and that gating on Applied is sufficient for Pebble-read freshness, which can mislead callers into issuing Pebble reads against a stale index.

**Suggestion:** Update the comment to clarify that during spooling windows Applied may race ahead of Pebble's persisted state. Direct readers to gate Pebble reads on lastPersistedIndex rather than raft.Status.Applied, and note that lastPersistedIndex is the authoritative lower-bound for read freshness.

internal/infra/node/applier_async_storage_test.go
NumaryBot, line 315: 🟡 **[minor] Fixed sleep in shutdown regression test makes deadlock-path coverage unreliable**

The fixed time.Sleep is the only synchronization ensuring runCommitter has reached the blocked response send before stop is closed. On a slow or heavily loaded CI runner, stop can be closed before the targeted code path is exercised, causing the test to pass without actually covering the deadlock regression it was written to guard against.

**Suggestion:** Replace the fixed sleep with an observable synchronization mechanism (e.g., a channel signal written just before the blocking point, or a require.Eventually condition that checks a detectable state) so the test reliably covers the intended scenario regardless of scheduler timing.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1498#issuecomment-4890764481

internal/infra/node/node.go
NumaryBot, line 2017: 🟡 **[minor] Comment implies raft.Applied is always a safe freshness gate for Pebble reads**

During non-normal applier states (checkpoint/snapshot gating or out-of-sync spooling), MsgStorageApplyResp is acknowledged immediately after spool.AppendCommittedEntries without calling Machine.publishApplied. This means raft.Status.Applied can advance ahead of Pebble's actual applied state during those windows. The comment in this area implies LastPersistedIndex >= Applied and that gating on Applied is sufficient for Pebble-read freshness, which can mislead callers of GetClusterState into issuing reads against a stale index.

**Suggestion:** Update the comment to clarify that during spooling windows Applied may race ahead of Pebble's persisted state. Direct readers to gate Pebble reads on lastPersistedIndex rather than raft.Status.Applied, and note that lastPersistedIndex is the authoritative lower-bound for read freshness.

internal/infra/node/applier_async_storage_test.go
NumaryBot, line 315: 🟡 **[minor] Fixed time.Sleep makes shutdown regression test's deadlock-path coverage unreliable**

The fixed time.Sleep is the only synchronization ensuring runCommitter has reached the blocked response send before stop is closed. On a slow or heavily loaded CI runner, stop can be closed before the targeted code path is exercised, causing the test to pass without actually covering the deadlock regression it was written to guard against.

**Suggestion:** Replace the fixed sleep with an observable synchronization mechanism (e.g., a channel signal written just before the blocking point, or a require.Eventually condition that checks a detectable state) so the test reliably covers the intended scenario regardless of scheduler timing.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1498#issuecomment-4890764481

internal/infra/node/node.go
NumaryBot, line 2017: 🟡 **[minor] Comment implies raft.Status.Applied is always safe for Pebble-read freshness**

During non-normal applier states (checkpoint/snapshot gating or out-of-sync spooling), MsgStorageApplyResp is acknowledged immediately after spool.AppendCommittedEntries without calling Machine.publishApplied. This means raft.Status.Applied can advance ahead of Pebble's actual applied state during those windows. The comment in this area implies that LastPersistedIndex >= Applied and that gating on Applied is sufficient for Pebble-read freshness, which can mislead callers into issuing reads against a stale index.

**Suggestion:** Update the comment to clarify that during spooling windows Applied may race ahead of Pebble's persisted state, and that lastPersistedIndex is the authoritative lower-bound for read freshness. Direct readers to gate Pebble reads on lastPersistedIndex rather than raft.Status.Applied.

internal/infra/node/applier_async_storage_test.go
NumaryBot, line 315: 🟡 **[minor] Fixed sleep in shutdown regression test makes deadlock-path coverage unreliable**

The fixed time.Sleep is the only synchronization ensuring runCommitter has reached the blocked response send before stop is closed. On a slow or heavily loaded CI runner, stop can be closed before the targeted code path is exercised, causing the test to pass without actually covering the deadlock regression it was written to guard against.

**Suggestion:** Replace the fixed sleep with an observable synchronization mechanism (e.g., a channel signal written just before the blocking point, or a require.Eventually condition that checks a detectable state) so the test reliably covers the intended scenario regardless of scheduler timing.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1498#issuecomment-4890764481

PR #1502, opened 2026-07-06
perf(processing): mutate volumes in place on repeat-touch keys

gfyrag:
## Summary

- Add `processing.Accessor.Mutate(k)` that returns the overlay-owned V directly on repeat touches (no CloneVT) and clones once from the parent cache on the first touch — implemented in `rawAccessor` with a new `mutator[V]` constraint on R, proxied through `gatedAccessor`
- `applyPosting` swaps `readVolumeOrZero + Reader.Mutate + Put` for the new `mutateVolumeOrZero` helper; trailing `Put` stays for slot-touch recording (map assignment is a no-op since the same overlay pointer is reinserted)
- Test scaffolds (`scopeFuncAccessor`, `kindStub`) get a `Mutate` that mirrors `rawAccessor` semantics via a runtime interface assertion on R — production behaviour preserved end to end

## Why

In world-to-bank-style workloads, the same `world` volume gets touched by every transaction in a proposal — currently every `applyPosting` does a `VolumePair.Mutate()` which `CloneVT`s (VolumePair + 2 Uint256 = 3 allocs) even when the key is already overlay-owned. On second and subsequent touches of the same key in a proposal, the clone is pure churn: the overlay already holds a distinct pointer safe to mutate in place.

For a 50-tx-per-proposal workload all sourced from `world`: from 100 CloneVT calls per proposal down to 51 (world clones once, each destination clones once).

## Measured impact

Baseline (from #<apply-phase-metrics PR>): PO% of applyProposal = 0.64.

After this PR: PO% = 0.60-0.61 — a ~12% reduction in ProcessOrders absolute time, ~8% reduction in applyProposal total.

## Test plan

- [x] `go test ./internal/domain/processing/... ./internal/infra/state/... ./internal/application/check/...` — all pass
- [x] `go build ./...` clean
- [ ] Bench delta measured via `raft.fsm.apply.process_orders.time_spent` counter (see PromQL in dashboard)

### DIFF
--- a/internal/application/check/checker_test.go
+++ b/internal/application/check/checker_test.go
@@ -365,6 +365,7 @@ type scopeFuncAccessor[K processing.AccessorKey, V any, R any] struct {
 	get    func(K) (R, error)
 	put    func(K, V)
 	delete func(K)
+	mutate func(K) (V, error)
 }
 
 func (a *scopeFuncAccessor[K, V, R]) Get(key K) (R, error) {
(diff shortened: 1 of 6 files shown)


Azorlogh:
`numscriptPostingProducer.produce` (L105/141) still uses the old `readVolumeOrZero` + `Reader.Mutate()` + `Put` pattern, so it re-clones the volume on every repeat touch. It runs on the same FSM apply path as `applyPosting`, and the "world-to-bank" case applies directly here (a `send [COIN *]` fan-out from `@world` re-clones `world` on every posting). `mutateVolumeOrZero(s, key)` drops in with no blocker. Was leaving it out intentional, or is it worth folding in?

PR #1504, opened 2026-07-06
fix(numscript): preload unbounded-overdraft sources via static involved-accounts discovery

### LINKED ISSUE #1500: v3.0.0-alpha.7: send overdraft(...) fails with STORAGE_OPERATION_FAILED — dependency discovery misses source volume when postings depend on non-zero balance
### Summary

On v3.0.0-alpha.7, a transaction whose send amount is computed by the `overdraft()` function fails with HTTP 500 `STORAGE_OPERATION_FAILED` when the resulting postings draw from a source account whose balance was never *read* by the script (e.g. `allowing unbounded overdraft`). The server logs `preload coverage miss: read of undeclared key` for the source volume.

The cause appears to be in numscript dependency discovery (`internal/domain/processing/numscript/emulate.go`): the discovery run emulates the script against empty balances, so `overdraft(@credit, USD/2)` evaluates to zero, the send produces no postings, and the real source volume is never collected into the preload set. At apply time the actual overdraft is non-zero, postings materialize, and the write hits an undeclared volume key.

### Steps to reproduce

Single-node server (`run --node-id 1 --cluster-id local-dev --bootstrap`, Docker image `ghcr.io/formancehq/ledger:v3.0.0-alpha.7`):

```bash
# 1. create a ledger
curl -s -X POST localhost:9000/v3/repro -H 'Content-Type: application/json' -d '{}'

# 2. put @credit into overdraft (works, HTTP 201)
curl -s -X POST localhost:9000/v3/repro/transactions \
  -H 'Content-Type: application/json' -H 'Idempotency-Key: repro-1' \
  -d '{"script":{"plain":"send [USD/2 100] (\n  source = @credit allowing unbounded overdraft\n  destination = @main\n)"}}'

# 3. repay the overdraft — fails, HTTP 500
curl -s -X POST localhost:9000/v3/repro/transactions \
  -H 'Content-Type: application/json' -H 'Idempotency-Key: repro-2' \
  -d '{"script":{"plain":"#![feature(\"experimental-overdraft-function\", \"experimental-mid-script-function-call\")]\nsend overdraft(@credit, USD/2) (\n  source = @repay allowing unbounded overdraft\n  destination = @credit\n)"}}'
```

Response for step 3:

```json
{"errorCode":"STORAGE_OPERATION_FAILED","errorMessage":"applying raft requests: storage operation failed: source volume repay/USD/2"}
```

Server log:

```
ERROR preload coverage miss: read of undeclared key {"kind": "volumes", ...}
ERROR Failed to create transaction {"ledger": "repro", "error": "applying raft requests: storage operation failed: source volume repay/USD/2"}
```

### Notes

- Binding `overdraft()` through a `vars` block instead of a mid-script call fails identically, so it is not specific to `experimental-mid-script-function-call`.
- A control transaction with a fixed amount and the same `source = @repay allowing unbounded overdraft` succeeds — the failure is specific to amounts derived from balance-dependent functions.
- The equivalent script works on v2.4.x with `--experimental-numscript-interpreter` and the same feature flags.
- `max overdraft(@X, USD/2) to @X` in a destination block works on alpha.7, presumably because the `overdraft()` read itself preloads `@X`, which is also the account being written.

### Expected behavior

The transaction succeeds and repays the overdraft (postings `repay -> credit` for the overdrawn amount), as on v2.

### Environment

- `ghcr.io/formancehq/ledger:v3.0.0-alpha.7` (`/_info`: version 3.0.0-alpha.7, commit 616fec71b, go1.26.4)
- Single-node bootstrap cluster, default flags, Docker on macOS (arm64)

aalloiteau:
## Summary

Fixes #1500 — `send overdraft(...)` failing at execution with `STORAGE_OPERATION_FAILED` / `read of undeclared key` on v3.

## Root cause

Numscript dependency discovery (`DiscoverNumscriptDependencies` in `emulate.go`) determines which account balances to preload/lock by **emulating the script against fake infinite balances**. The discovery store returns a large *positive* balance for every account, so:

- `overdraft(@credit, USD/2)` sees a positive balance → folds to `USD/2 0`
- a zero-amount send emits **no posting**
- the source `@repay` (an `allowing unbounded overdraft` source, which the interpreter never balance-queries) appears in neither `queriedVolumes` nor the postings

At execution the overdraft is non-zero, the posting `@repay → @credit` needs `@repay`'s volume, but it was never preloaded → `read of undeclared key`.

Emulation is fundamentally unable to cover this: `balance()` needs a positive fake balance to allow sends, `overdraft()` needs a negative one to be non-zero — no single value satisfies both. This is exactly the static-analysis gap the existing code comment anticipates.

## Fix

Union numscript's static `GetInvolvedAccounts` analysis into `sourceVolumes`. It walks the AST and reports every account each send touches — **including `allowing unbounded overdraft` sources — regardless of the resolved amount**. Two small resolvers fold the involved-account expressions to concrete `(account, asset)` strings (e.g. `GetAsset{GetOverdraft{@credit, USD/2}}` → `"USD/2"`).

- Accounts already classified as destinations are skipped, so the source/destination partition (and existing exact-length assertions) is preserved.
- The union is a safe superset: over-preloading only locks an extra row; under-preloading was the bug.
- Best-effort: if `GetInvolvedAccounts` errors on an experimental construct it doesn't model yet (e.g. asset scaling), we keep the emulation-only result rather than regressing scripts that work today.

## Tests

- Added `overdraft send discovers the unbounded-overdraft source (ledger#1500)`. Verified it **fails without the fix** (`@repay` missing) and **passes with it**.
- All existing `TestDiscoverNumscriptDependencies` subtests still pass (no regression).

## ⚠️ Blocked on numscript release — do not merge as-is

`GetInvolvedAccounts` + the `github.com/formancehq/numscript/accounts` package are merged on numscript `main` but **not yet in a tagged release** (ledger pins `v0.0.24`). This PR temporarily pins the numscript **main pseudo-version** (`v0.0.25-0.20260608064954-1fcd4893eafe`) so it compiles and CI can run.

**Before merge:**
1. Tag a numscript release (e.g. `v0.0.25`) containing `GetInvolvedAccounts`.
2. `go get github.com/formancehq/numscript@v0.0.25 && go mod tidy` and re-pin to the clean tag.

## Notes

- Target branch is `release/v3.0`: this code path (`emulate.go` / admission preload) is **v3-only**; `main` is the v2.x line and has no equivalent, so there is nothing to land there or backport from.


internal/domain/processing/numscript/emulate.go
```diff
@@ -174,3 +215,55 @@ func DiscoverNumscriptDependencies(cache *NumscriptCache, script string, vars ma
...
+	switch e := expr.(type) {
+	case accounts.AssetLiteral:
+		return e.Asset, true
+	case accounts.MakeMonetary:
+		return resolveInvolvedAsset(e.Asset)
+	case accounts.GetBalance:
+		return resolveInvolvedAsset(e.Asset)
+	case accounts.GetOverdraft:
+		return resolveInvolvedAsset(e.Asset)
+	case accounts.GetAsset:
```
NumaryBot, line 272: 🟠 **[major] Arithmetic-wrapped overdraft assets not handled in GetInvolvedAccounts**

When the sent amount wraps `overdraft()` in monetary arithmetic (e.g. `send overdraft(@credit, USD/2) + [USD/2 0] (...)`), the asset resolver in `GetInvolvedAccounts` encounters an `accounts.Add`/`Sub`/`SubPrefix` node rather than a plain overdraft expression. The handler falls through without recognising the embedded unbounded-overdraft source, so the account is never added to the preload set and the original 'read of undeclared key' failure can still occur for those scripts.

**Suggestion:** Recursively walk arithmetic nodes (`Add`, `Sub`, `SubPrefix`, etc.) inside the asset-resolver so that any overdraft leaf nested within a compound monetary expression is also discovered and included in the involved-accounts preload set.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1504#issuecomment-4904996660

internal/domain/processing/numscript/emulate.go
NumaryBot, line 272: 🟠 **[major] Arithmetic-wrapped overdraft assets not handled in GetInvolvedAccounts**

When the sent amount wraps an overdraft expression inside monetary arithmetic (e.g. `send overdraft(@credit, USD/2) + [USD/2 0] (...)`), the asset resolver in `GetInvolvedAccounts` encounters a compound node type (`Add`, `Sub`, `SubPrefix`) rather than a bare overdraft expression. The handler falls through without recognising the embedded unbounded-overdraft source, so the account is never added to the preload set and the original 'read of undeclared key' failure can still occur for scripts of that form.

**Suggestion:** Recursively walk arithmetic nodes (`Add`, `Sub`, `SubPrefix`, etc.) inside the asset-resolver so that any overdraft leaf nested within a compound monetary expression is also discovered and included in the involved-accounts preload set.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1504#issuecomment-4904996660

NumaryBot (commented):

The PR correctly implements static involved-account discovery to preload unbounded-overdraft sources. The previously raised major issue regarding arithmetic-wrapped overdraft expressions (Add/Sub/SubPrefix nodes) was flagged in prior automated review threads, but the latest reviewer confirms the current diff already addresses those cases. No new blockers or major issues have been identified. Patch coverage (~64%) remains below the project average (~74%), which is worth improving but does not warrant blocking the merge.

### Findings outside the diff

🟡 **[minor] Patch coverage below project average** — `internal/domain/processing/numscript/emulate.go`

The new code has ~64% patch coverage against a project baseline of ~74%. Arithmetic-wrapping paths and other edge cases in GetInvolvedAccounts appear to lack test coverage, which could allow regressions to go undetected.

**Suggestion:** Add tests covering overdraft expressions embedded inside compound monetary arithmetic nodes (Add, Sub, SubPrefix) to bring patch coverage in line with the project average.



Summary: https://github.com/formancehq/ledger/pull/1504#issuecomment-4904996660

Azorlogh (commented):
Safe as a stopgap — the union only ever over-preloads, and discovery is admission-only (never touches the FSM). Three notes:

1. **Wrap `GetInvolvedAccounts` in a recover.** Every other numscript call here goes through `SafeRun` because the interpreter can panic (numscript still has open panic fixes, e.g. [numscript#153](https://github.com/formancehq/numscript/pull/153)). Degrade a panic to the same best-effort skip as the error branch.
2. **FYI (no action needed):** `GetInvolvedAccounts` was removed from numscript main in [numscript#166](https://github.com/formancehq/numscript/pull/166) and never tagged, so the "tag v0.0.25 + re-pin" plan can't work. The intended replacement is [numscript#135](https://github.com/formancehq/numscript/pull/135) (`ResolveDependencies`, returns concrete values), which would let this whole hack be deleted rather than re-pinned.
3. **Nit:** e2e step-1 comment says "balance on `@main` allows the send" — `@main` is the destination; source is `@credit`.

Azorlogh:
LGTM, but dirty check is failing

Azorlogh (approved):

gfyrag:
Closing in favour of #1560 (EN-1406), which fixes the same underlying problem more cleanly.

This PR worked around the dependency-discovery gap by unioning numscript's static `GetInvolvedAccounts` analysis into the emulation-based result — a best-effort superset patched on top of `emulate.go`. #1560 removes the emulation approach entirely and resolves dependencies through numscript's upstream `ParseResult.ResolveDependencies(ctx, vars, store)` API (formancehq/numscript#169), which reports account/metadata reads & writes — including `allowing unbounded overdraft` sources — directly from the interpreter rather than inferring them. That's the correct long-term fix for #1500 and it subsumes this change without the emulation/static-analysis dual-path.

No code from here needs to be carried over: #1560 replaces `emulate.go` wholesale. Closing to avoid two competing approaches to the same discovery problem.

PR #498, opened 2024-09-26
refactor(Earthfile): reuse sources stage and optimize build process

flemzord:
Use +sources stage in multiple targets for consistency and efficiency. Add go mod download step to sources stage to reduce redundancy. Remove redundant COPY commands to streamline stages.

PR #1508, opened 2026-07-06
feat(ledger): navigable revert relationship + queryable reverted/revertedAt

Azorlogh:
## Summary

Reverts tracked the original↔compensating link only internally (`TransactionState.RevertedByTransaction`), surfaced as a bare `reverted` bool. Clients couldn't navigate between the two sides, `revertedAt` was never populated, and revert status wasn't queryable.

Per the issue's decision, the relationship is modeled as **first-class transaction fields** — not `com.formance.spec/*` metadata. User-supplied metadata stays untouched.

### 1. `feat(ledger): expose revert relationship as first-class transaction fields`
- `TransactionState` gains `reverted_at` + `reverts_transaction`; the API `Transaction` gains `reverted_by_transaction` + `reverts_transaction` and now populates `reverted_at`.
- JSON: `revertedAt` + `revertedByTransactionId` on the reverted original, `revertsTransactionId` on the compensating tx.
- `revertedAt` = the compensating transaction's effective timestamp (so under `atEffectiveDate` it equals the original's).
- Set on both `processor_revert_transaction` and the **mirror** revert path.
- Reconstructed by the shared `replay.Writer` (integrity checker + backup rebuild), so `proto.Equal` holds after **replay/backfill** and across nodes.

### 2. `feat(query): make reverted and revertedAt transaction fields queryable`
Upstream v2 (`internal/queries/resources.go` `TransactionSchema`) queries both `reverted` and `reverted_at`; v3 now matches that subset:
- **`reverted` (bool)** — new `RevertedCondition`, served from the existing per-ledger reversion bitset. No index, always available, complete for all history. `reverted=false` = universe minus the bitset (existing `NotIterator`).
- **`revertedAt` (date range)** — new `reverted-at` builtin index. Entries are derived from the `RevertedTransaction` **log** (original tx keyed by the compensating timestamp), so the standard log-replay backfill covers historical reverts and the mirror path. Wired into `ledgerctl indexes create/drop/list` and the declarative ledger config.

The v3-only `revertedByTransactionId` / `revertsTransactionId` remain navigable-only (no v2 baseline).

## Acceptance criteria (#498)
- [x] GET original → `revertedAt` + `revertedByTransactionId` (absent when not reverted)
- [x] GET reverting → `revertsTransactionId`
- [x] No `com.formance.spec/*` metadata written by the platform
- [x] Filters on `reverted` / `revertedAt` consistent with upstream parity for the queryable subset
- [x] Holds across nodes and after replay/backfill, including the mirror revert path
- [x] Tests for `processor_revert_transaction` + the mirror path; `docs/technical/contributing/api-comparison.md` + `openapi.yml` updated

## Testing
- Unit + integration: revert processor (both paths + `atEffectiveDate`), mirror path, replay-store reconstruction, `Transaction.MarshalJSON`, GET-handler JSON surface, `BitsetIterator`, `ReadReversionBitset`, reverted-at index write, compile dispatch.
- `go build ./...` + touched-package tests green. Protos regenerated with the pinned nix toolchain (byte-identical for unchanged messages); each commit's generated code matches its own `.proto`.
- Unrelated pre-existing failures: 3 `internal/pkg/tarutil` + 1 `cmd/ledgerctl/auth` tests fail only because CI-local ran as root (they assert on `chmod`/keyring denials root bypasses) — confirmed failing on the base branch too.

## Note
`openapi.yml`'s `TransactionResponse` already documented `inserted_at` / `updated_at` / `reverted_at` in snake_case while the server emits camelCase; the new fields are added in camelCase (matching the server) and the pre-existing mismatch is left alone. Happy to align the existing three separately if wanted.

### DIFF
--- a/internal/application/ctrl/controller_default.go
+++ b/internal/application/ctrl/controller_default.go
@@ -269,6 +269,13 @@ func assembleTransactionFromState(ctx context.Context, reader dal.PebbleReader, 
 
 	tx.Reverted = state.GetRevertedByTransaction() > 0
 
+	// The revert relationship is tracked structurally on the state, not in the
+	// create log: reverted_by_transaction + reverted_at on the reverted original,
+	// reverts_transaction on the compensating transaction.
+	tx.RevertedByTransaction = state.GetRevertedByTransaction()
+	tx.RevertedAt = state.GetRevertedAt()
+	tx.RevertsTransaction = state.GetRevertsTransaction()
+
 	// The state carries the current metadata (the create-time snapshot plus
 	// every applied add/delete) and is authoritative even when empty; the
 	// create log only holds the create-time snapshot.
--- a/cmd/ledgerctl/indexes/create.go
+++ b/cmd/ledgerctl/indexes/create.go
@@ -28,6 +28,7 @@ Index types:
   reference            Transaction reference index (exact-match filter)
   timestamp            Transaction timestamp index (range filter)
   inserted-at          Transaction inserted_at (creation date) index (range filter)
+  reverted-at          Transaction reverted_at (revert date) index (range filter)
   log-ledger           Per-ledger log index (enables filtered log listing)
   account-asset        Account asset-presence index (enables the 'has asset' account filter)
 
@@ -38,6 +39,7 @@ Examples:
   ledgerctl indexes create --ledger my-ledger --type reference
   ledgerctl indexes create --ledger my-ledger --type timestamp
   ledgerctl indexes create --ledger my-ledger --type inserted-at
+  ledgerctl indexes create --ledger my-ledger --type reverted-at
   ledgerctl indexes create --ledger my-ledger --type log-ledger
   ledgerctl indexes create --ledger my-ledger --type account-asset`,
 		Args:              cobra.NoArgs,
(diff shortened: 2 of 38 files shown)


internal/application/ctrl/controller_default.go
```diff
@@ -269,6 +269,13 @@ func assembleTransactionFromState(ctx context.Context, reader dal.PebbleReader,
 
 	tx.Reverted = state.GetRevertedByTransaction() > 0
 
+	// The revert relationship is tracked structurally on the state, not in the
+	// create log: reverted_by_transaction + reverted_at on the reverted original,
+	// reverts_transaction on the compensating transaction.
+	tx.RevertedByTransaction = state.GetRevertedByTransaction()
+	tx.RevertedAt = state.GetRevertedAt()
```
NumaryBot, line 276: 🟠 **[major] Missing migration for legacy revert state fields**

Existing stores with reverted transactions do not have the newly-added `RevertedAt`/`RevertsTransaction` fields populated in their persisted `TransactionState`. There is no schema migration or version bump in this PR. After upgrade, GET requests on previously-reverted transactions will omit `revertedAt`, and the checker/replay path that reconstructs these fields will find a mismatch against the stored projection, potentially causing correctness failures.

**Suggestion:** Add a data migration (or a schema version bump + replay) that backfills `RevertedAt` and `RevertsTransaction` for all existing reverted transactions. Alternatively, add a graceful fallback so that missing values are tolerated during the transition period.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1508#issuecomment-4893985832

gfyrag (approved):
LGTM. Modélisation structurelle propre, invariants respectés (#3/#6/#7/#8), projections reconstructibles via replay.Writer. Deux points à traiter en follow-up (non-bloquants) : (1) backfill des `RevertedAt`/`RevertsTransaction` sur les transactions déjà reverties avant l'upgrade pour éviter les faux positifs du checker ; (2) unit tests sur `compileRevertedCondition` (value=true/false) et `compileRevertedAtCondition` (index absent/READY) pour combler la coverage de compile.go.

PR #1511, opened 2026-07-06
perf(proto): zero-alloc reader wrappers via named-type conversion

gfyrag:
## Summary

- `protoc-gen-reader` now emits the read-only wrapper as `type xReadonly Msg` (named-type over the concrete message) instead of `struct{ v *Msg }`. `AsReader()` becomes a pointer type-conversion `(*xReadonly)(m)` — the resulting `*xReadonly` fits in the interface data word, so no heap wrapper is allocated. Every transitive `GetFoo()` / `ListReader.Get(i)` / `MapReader.Range` reader view is likewise zero-alloc.
- All 12 `*_reader.pb.go` regenerated. Public surface unchanged: same `Reader` interfaces, same method signatures, same immutability contract.
- Added `internal/proto/commonpb/reader_alloc_test.go` with micro-benchmarks pinning the zero-alloc property.

### Bench (Apple M5 Pro)
```
BenchmarkReader_AsReader-15    717M ops   1.7 ns/op    0 B/op   0 allocs/op
BenchmarkReader_GetChain-15     87M ops  17.5 ns/op   24 B/op   1 allocs/op  ← residual slice-header boxing on ListReader
BenchmarkReader_ListRange-15    17M ops  62.8 ns/op    0 B/op   0 allocs/op
```

Before the patch, each `AsReader()` (and every transitive `GetFoo()` returning a sub-reader) allocated a ~16 B wrapper struct on the heap. On hot paths this compounded — e.g. iterating a `TransactionListReader` and reading a few postings per element caused N allocations per transaction.

### Why not `sync.Pool`
Pooling was considered and rejected:
- Reader lifecycle is intraçable: sub-readers are created implicitly via `GetFoo()` cascades; the caller can't know when to `Release()`.
- Would require adding `Release()` to the Reader interface and threading it through every call site (`Range` bodies would need a release per iteration).
- The wrapper is 8 bytes (one pointer); `sync.Pool` bookkeeping (atomic ops, per-P shard, GC eviction) often costs more than the alloc it saves for objects this small.
- Use-after-release corruption is silent — the reader's underlying pointer would be reassigned.

The named-pointer-type approach gives the same result with zero API changes and zero lifecycle risk.

### Immutability
Unchanged. `xReadonly` remains unexported. External callers cannot type-assert to `*Msg` (dynamic type mismatch) nor to `*xReadonly` (unexported name). `unsafe.Pointer` shenanigans can defeat both the old and the new design equally — no new surface.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/proto/...` — all 6 `reader_immutability_test.go` suites still pass, pinning the read-only contract
- [x] `just pre-commit` — 0 lint issues
- [x] Added benchmarks confirm 0 allocs/op for `AsReader` and `ListReader.Range`

### DIFF
--- a/tools/protoc-gen-reader/main.go
+++ b/tools/protoc-gen-reader/main.go
@@ -1,7 +1,10 @@
-// protoc-gen-reader generates read-only interfaces and wrapper structs for
+// protoc-gen-reader generates read-only interfaces and wrapper types for
 // protobuf messages. Each message gets:
 //   - A <Message>Reader interface with all Get* methods + Mutate() *<Message>
-//   - An unexported <message>Readonly wrapper struct implementing the interface
+//   - An unexported <message>Readonly named type defined as `type X Msg` so
+//     that AsReader() can convert *Msg -> *<message>Readonly without heap
+//     allocation. The resulting *<message>Readonly fits in the interface data
+//     word, and every transitive GetFoo() reader view is likewise zero-alloc.
 //   - An AsReader() method on the concrete type returning the reader interface
 //   - A Mutate() method on the concrete type returning CloneVT()
 //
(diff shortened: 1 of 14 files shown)


tools/protoc-gen-reader/main.go
```diff
@@ -153,16 +158,23 @@ func generateMessage(g *protogen.GeneratedFile, msg *protogen.Message) {
...
 
-	// --- Wrapper struct ---
+	// --- Wrapper type ---
+	// Defined as a named type over the concrete message so that AsReader() can
+	// convert *typeName -> *wrapperName without allocating a heap wrapper. The
+	// resulting *wrapperName fits in the interface data-word, keeping the whole
+	// reader-view chain zero-alloc.
 	g.P()
-	g.P("type ", wrapperName, " struct{ v *", typeName, " }")
+	g.P("type ", wrapperName, " ", typeName)
```
NumaryBot, line 167: 🟠 **[major] Keep reader wrappers opaque to reflection**

When these generated wrappers become named versions of the protobuf struct, any package that receives a Reader can use ordinary reflection on the interface value (for example, `reflect.ValueOf(r).Elem().FieldByName("...").Set...`) to mutate exported proto fields on the original message. The previous `struct{ v *Msg }` wrapper only exposed an unexported pointer field, so this did not work without unsafe; this weakens the advertised read-only contract for any Reader value that crosses package boundaries.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1511#issuecomment-4894609898

flemzord (approved):

PR #1516, opened 2026-07-07
fix(EN-1469): quote Uint256 JSON encoding for precision-safe wire

gfyrag:
## Summary

Uint256 values exceed JSON's safe-integer range (2^53), so emitting the raw decimal without quotes silently lost precision in JavaScript clients — the prepared-query aggregate path was the alpha-reported symptom.

- `Uint256.MarshalJSON` now wraps the decimal in double quotes, so JS callers can parse it as a BigInt/BigNumber without loss.
- `Uint256.UnmarshalJSON` accepts both the quoted form and the legacy unquoted decimal, so mid-transition clients still emitting a raw number keep working.

Existing wire consumers that go through the string-based marshallers (`Volumes.MarshalJSON`, `toAggregatedVolumeJSON` on `GET /v3/{ledger}/volumes`) are unaffected — they already used strings via `big.Int`. This change lines up the last remaining number-encoded path.

## Test plan

- [x] `GOROOT= go test ./internal/proto/commonpb/... -run TestUint256`
- [x] New `TestUint256_UnmarshalJSON_LegacyUnquoted` asserts pre-EN-1469 unquoted input still parses
- [x] Fuzz seed corpus (`FuzzUint256UnmarshalJSON`) still valid — legacy unquoted seeds accepted
- [x] `just test` short run: all packages green

Wire change (alpha → beta): consumers that parsed prepared-query aggregate `input`/`output` as a JSON number must now parse them as strings (with BigInt / BigNumber). Callers going through `Volumes.MarshalJSON` (`GET /v3/{ledger}/volumes`) or `toAggregatedVolumeJSON` were already receiving strings.

Ticket: [EN-1469](https://formance-team.atlassian.net/browse/EN-1469) (epic: [EN-867](https://formance-team.atlassian.net/browse/EN-867))

[EN-1469]: https://formance-team.atlassian.net/browse/EN-1469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/proto/commonpb/uint256.go
+++ b/internal/proto/commonpb/uint256.go
@@ -88,16 +88,34 @@ func (u *Uint256) Dec() string {
 	return v.Dec()
 }
 
-// MarshalJSON encodes the Uint256 as a decimal string (no quotes) for JSON.
+// MarshalJSON encodes the Uint256 as a quoted decimal string. Uint256 values
+// exceed the safe-integer range of a JSON number (2^53), so emitting the raw
+// decimal without quotes silently loses precision in JavaScript clients — the
+// wire format is a string that consumers must parse as a BigInt/BigNumber.
 func (u *Uint256) MarshalJSON() ([]byte, error) {
-	return []byte(u.Dec()), nil
+	dec := u.Dec()
+	buf := make([]byte, 0, len(dec)+2)
+	buf = append(buf, '"')
+	buf = append(buf, dec...)
+	buf = append(buf, '"')
+
+	return buf, nil
 }
 
-// UnmarshalJSON decodes a decimal string (no quotes) into the Uint256.
+// UnmarshalJSON decodes a decimal representation into the Uint256. It accepts
+// both the current quoted form (`"1000"`) and the legacy unquoted form
+// (`1000`) so wire-format changes don't break older callers mid-transition.
 func (u *Uint256) UnmarshalJSON(data []byte) error {
+	// Strip surrounding double quotes if present; the raw decimal is what
+	// uint256.SetFromDecimal accepts.
+	s := string(data)
+	if len(s) >= 2 && s[0] == '"' && s[len(s)-1] == '"' {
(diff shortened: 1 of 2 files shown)


gfyrag:
Closing per review feedback: the direction I took here (align the wire to match a hypothetical string-encoded spec) is backwards. The right move is to align the *doc* to the actual wire — Uint256 has always been a JSON number in v3 alpha.7 (and in v2 via `integer format: bigint`).

Changing the wire to a quoted string would break every existing client (Go, Python, Java, Go, JS-with-BigInt libs) for a marginal benefit that only helps JS clients not already using a BigInt library — and those clients are already broken on `GET /v3/{ledger}/volumes` where the custom marshaller quotes big.Int anyway.

The reality is:
- `GET /v3/{ledger}/volumes` — quoted string amounts + `balance`, via the custom `toAggregatedVolumeJSON` marshaller (unchanged).
- Prepared-query `AGGREGATE_VOLUMES` — unquoted number amounts, no `balance`, via raw `Uint256.MarshalJSON`.

That's what the spec should describe. EN-1467 already carries the `AggregatedVolume` schema fix for `/volumes`, and the `ExecutePreparedQueryResponse` there is stubbed with a note pointing at the protojson-leak reconciliation tracked under EN-1465. No wire change needed.

Reverted the commit on the branch; abandoning this PR. The Jira ticket [EN-1469](https://formance-team.atlassian.net/browse/EN-1469) needs its status updated (probably "Won't do" or reworded to a doc-only note in EN-1467).

[EN-1469]: https://formance-team.atlassian.net/browse/EN-1469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1518, opened 2026-07-07
perf: inline Timestamp as fixed64 scalar (zero-alloc)

gfyrag:
## Summary

- Replace `common.Timestamp` proto message with `fixed64` scalars inline on every parent message; wrap Go-side with `type commonpb.Timestamp uint64` (value type with domain methods).
- Removes one heap allocation per timestamp field per decoded log entry (Transaction alone carries 4 timestamps).
- Sentinelle `0 = unset` for nullable fields (RevertedAt, DeletedAt); the HLC never emits micros=0 so the collapse is safe.

Related to the discussion around `feat/immutable-timestamp` (superseded): rather than post-process the codegen to unexport `*commonpb.Timestamp.Data`, we sidestep the whole wrapper message. Codegen is stock protoc-gen-go.

## Breaking

- **Wire format**: `common.Timestamp` (wire type 2, length-delimited) → `fixed64` (wire type 1). Not backward-compatible.
- **Audit chain**: hashes over marshaled proto differ for the same logical entry → v3-only cutover, no migration path for chains persisted before this change.
- **API**: `replayStore.CreateTransaction` and `receipt.Signer.Sign` take `uint64` timestamp instead of `*commonpb.Timestamp`.

## Test plan

- [x] `GOROOT= go vet ./...`
- [x] `GOROOT= go test ./internal/... ./cmd/... ./pkg/...` (all pass)
- [x] `just pre-commit` (golangci-lint 0 issues)
- [ ] CI green
- [ ] E2E tests green
- [ ] Antithesis run


misc/proto/bucket.proto
```diff
@@ -487,7 +487,7 @@ message ResponseSigningInfo {
 message CreateTransactionPayload {
   repeated common.Posting postings = 1;
   common.Script script = 2;
-  common.Timestamp timestamp = 3;
+  fixed64 timestamp = 3;
```
NumaryBot: 🟠 **[major] Update remaining timestamp call sites**

After changing this field to a scalar `uint64`, some checked-in packages still use the removed message API: the antithesis timestamp-order workload assigns `payload.Timestamp = &commonpb.Timestamp{Data: ...}`, and the e2e business tests call `.GetData()` / accept `*commonpb.Timestamp`. Building the antithesis workload or running `go test -tags e2e ./tests/e2e/...` will fail until those call sites are migrated to the new scalar form.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1518#issuecomment-4903260496

misc/proto/common.proto
```diff
@@ -67,13 +69,13 @@ message Posting {
 message Transaction {
   repeated Posting postings = 1;
   map<string, MetadataValue> metadata = 2;
-  Timestamp timestamp = 3;
+  fixed64 timestamp = 3;
```
NumaryBot: 🔴 **[blocker] Tagged test code still uses removed *commonpb.Timestamp message type**

Changing Timestamp fields from the message type to `fixed64` scalars breaks tagged test consumers that still reference the old API. For example, `tests/e2e/business/transactions_test.go` calls `createdTx.Transaction.Timestamp.GetData()` and `tests/antithesis/workload/bin/cmds/main/parallel_driver_timestamp_order/main.go` assigns `&commonpb.Timestamp{...}` to `CreateTransactionPayload.Timestamp`. This means `go test -tags e2e ./tests/e2e/...` and the antithesis workload will fail to compile after protobuf regeneration.

**Suggestion:** Update all call sites in the e2e and antithesis test packages to use the new `uint64` / `fixed64` field directly instead of constructing or accessing a `*commonpb.Timestamp` message. Remove any remaining references to `GetData()` on the timestamp field.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1518#issuecomment-4903260496

misc/proto/common.proto
NumaryBot: 🔴 **[blocker] Tagged test code still uses removed *commonpb.Timestamp message type**

Changing Timestamp fields from the message type to `fixed64` scalars breaks tagged test consumers that still reference the old API. `tests/e2e/business/transactions_test.go` calls `createdTx.Transaction.Timestamp.GetData()` and `tests/antithesis/workload/bin/cmds/main/parallel_driver_timestamp_order/main.go` assigns `&commonpb.Timestamp{...}` to `CreateTransactionPayload.Timestamp`. Both the e2e and antithesis build targets will fail to compile after protobuf regeneration.

**Suggestion:** Update all call sites in the e2e and antithesis test packages to use the new `uint64`/`fixed64` field directly instead of constructing or accessing a `*commonpb.Timestamp` message. Remove any remaining references to `.GetData()` on timestamp fields.

misc/proto/bucket.proto
NumaryBot: 🟠 **[major] Update remaining timestamp call sites in antithesis and e2e packages**

After changing Timestamp fields to scalar `uint64`, the antithesis timestamp-order workload still assigns `payload.Timestamp = &commonpb.Timestamp{Data: ...}` and the e2e business tests still call `.GetData()` / accept `*commonpb.Timestamp`. Building the antithesis workload or running `go test -tags e2e ./tests/e2e/...` will fail until those call sites are migrated to the new scalar form.

**Suggestion:** Replace all `&commonpb.Timestamp{Data: ...}` constructions with plain `uint64` values, and remove all `.GetData()` accessor calls, updating the affected files to work with the scalar field directly.

internal/proto/commonpb/transaction.go
```diff
@@ -207,23 +195,23 @@ func (tx *Transaction) MarshalJSON() ([]byte, error) {
 		aux.ID = new(tx.GetId())
 	}
 
-	if tx.GetTimestamp() != nil {
-		t := tx.GetTimestamp().AsTime()
+	if ts := Timestamp(tx.GetTimestamp()); ts.IsSet() {
```
NumaryBot: 🟠 **[major] Transaction JSON marshal/unmarshal asymmetry breaks round-trips**

After the proto Timestamp field was changed to `fixed64`, `Transaction.MarshalJSON` still emits RFC3339 strings for timestamp fields, but the default unmarshaler now targets `uint64` and no longer invokes `commonpb.Timestamp.UnmarshalJSON`. This creates an asymmetry that breaks JSON round-trips for `timestamp`, `insertedAt`, `updatedAt`, and `revertedAt` fields — affecting import paths such as `LedgerLog.UnmarshalJSON` → `HydrateLog` for created or reverted transaction logs.

**Suggestion:** Add a matching `UnmarshalJSON` for `Transaction` that decodes those timestamp fields from RFC3339 strings back into `uint64`, mirroring the existing `MarshalJSON` logic. Alternatively, route them through `commonpb.Timestamp` during JSON decode.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1518#issuecomment-4903260496

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1518#issuecomment-4903260496

internal/proto/commonpb/log.go
```diff
@@ -52,7 +52,7 @@ func (l *LedgerLog) UnmarshalJSON(data []byte) error {
 	}
 
 	if rawLog.Date != nil {
-		l.Date = NewTimestamp(*rawLog.Date)
+		l.Date = uint64(NewTimestamp(*rawLog.Date))
```
NumaryBot, line 55: 🟠 **[major] Panic instead of error on invalid date in UnmarshalJSON**

When `LedgerLog.UnmarshalJSON` receives a `date` value at or before the Unix epoch, it calls `NewTimestamp`, which panics for those values instead of returning a proper error. Since this method parses untrusted JSON input and may be exercised by fuzzers or malformed payloads (e.g. `{"date":"1970-01-01T00:00:00Z"}`), a panic here is a correctness and reliability issue — it crashes the caller rather than propagating an error normally.

**Suggestion:** Replace the panicking call path in `UnmarshalJSON` with a safe constructor that returns an error, and propagate that error to the caller instead of allowing the panic to escape.

gfyrag:
Fermeture — pas mergeable dans l'état, gains marginaux vs coût d'audit continu. Voir ADR #1524 pour le raisonnement associé (path unmarshal_unsafe rejeté pour raisons similaires : gain net faible après clone-at-cache). L'infrastructure de typed accessors via protoc-gen-reader reste réutilisable si on revient sur le sujet.

PR #1519, opened 2026-07-07
fix(EN-1043): emit Unreachable on Raft transport channel-full drops

gfyrag:
## Summary

- Fix silent MsgApp/MsgHeartbeat drops when per-peer or top-level send channels fill up: emit `pushUnreachable(peerID)` on every drop path so etcd/raft transitions the peer's Progress to `StateProbe` and slows retries until the peer responds. Root cause was the asymmetry "drop the message but tell Raft the peer is alive" — Raft is safe under loss, but its drop-tolerance depends on the transport signalling.
- Dedup per-peer via a CAS-guarded `atomic.Bool` on `peerConnection` (reset after a successful `sendMessages`) so a drop burst doesn't cascade-drop `unreachableCh` (cap 100) itself. For `DefaultTransport.Send`, emit exactly once per unique `msg.To` in the dropped batch.
- Antithesis: `singleton_driver_quorum_recovery` now measures elapsed time from scale-down-to-1 to `WaitForVoters(1)` returning and asserts `Sometimes(elapsed < 30s)`. Would fire on the 47s stall observed in the 2026-06-08 run.

## Why this matters

Without the fix, the leader's `nextIndex[peer]` stays in `StateReplicate` while messages are being dropped, retrying optimistically and refilling the channel. Recovery lasts the full TCP timeout (~30-50s observed). Pending `ConfChange` cannot commit during that window (etcd/raft single-in-flight rule), so the K8s operator's scale-up/scale-down loop stalls.

Two mitigations already shipped in-tree address downstream symptoms only:
- `drainLoop` on confirmed connection death (`transport.go:834-871`) — fires too late.
- Priority-0 buffer bump 10→128 (commit `a2fa4a277`, 2026-07-06) — narrows the window but doesn't close it.

This PR closes the "channels full but connection still considered alive" window, which is where the actual stall lives.

## What changed

- `internal/infra/node/transport.go`
  - `peerConnection.pushMessages` — on channel-full, `pushUnreachable(peerID)` via CAS on new `unreachableReported atomic.Bool`.
  - `sendMessages` — reset the dedup flag after a successful `stream.Send()`, so a re-saturation re-signals Raft.
  - `DefaultTransport.Send` — on `pendingSendQueue` full, iterate the batch and emit `pushUnreachable` once per unique `msg.To`.
- `internal/infra/node/transport_test.go` (new) — 5 tests: emit-on-drop, dedup-in-burst, re-emit-after-success, Send-per-unique-peer, no-emit-happy-path.
- `internal/infra/node/peer_send_race_test.go` — helper sets a no-op `pushUnreachable` so the new path doesn't nil-deref in the existing race test.
- `tests/antithesis/workload/bin/cmds/main/singleton_driver_quorum_recovery/main.go` — `Sometimes(elapsed < 30s)` assertion on force-remove ConfChange latency.

## Out of scope

- `pushToRecvQueue` (recv-side drops) — `ReportUnreachable` is a sender→receiver signal, doesn't map.
- Bounded blocking send with deadline — micro-optimization. Raft recovers via probe in ~1 heartbeat (~100ms) once the peer responds; a 5ms pre-attempt is not worth the complexity for v1. Revisit if `raft.transport.peer.sending.full` stays elevated after this ships.
- Buffer sizing (`a2fa4a277`) and drainLoop (already present) — untouched.

## Test plan

- [x] `go build ./...` + full-features `go build -tags "kafka,nats,clickhouse,databricks,s3,azure,pyroscope" ./...`
- [x] `go test ./internal/infra/node/ -race -count=1` — all pass (existing + 5 new)
- [x] `golangci-lint run --new-from-rev=origin/release/v3.0` — 0 issues
- [x] `go vet ./...` — clean
- [ ] Antithesis 30-min k8s run: new `force-remove ConfChange applies within latency budget after scale-down (EN-1043)` assertion holds; `raft.transport.peer.sending.full` counter not materially higher than baseline; no new "Unreachable channel full" logs

Refs: EN-1043

### DIFF
--- a/internal/infra/node/transport.go
+++ b/internal/infra/node/transport.go
@@ -454,6 +454,21 @@ func (t *DefaultTransport) Send(msgs []raftpb.Message) {
 			"channel": "raft.send.pending_messages",
 		}).Errorf("Channel full")
 		t.pendingSendFullCounter.Add(context.Background(), 1)
+
+		// Signal Unreachable for every peer that had a message in the
+		// dropped batch. pendingSendQueue is shared across peers, so we
+		// can't blame a single one — but transitioning every affected
+		// peer's Progress to StateProbe throttles overall outgoing load,
+		// which is the correct systemic response to a full top-level
+		// queue.
+		reported := make(map[uint64]struct{}, len(msgs))
+		for _, m := range msgs {
+			if _, ok := reported[m.To]; ok {
+				continue
+			}
+			reported[m.To] = struct{}{}
+			t.pushUnreachable(m.To)
+		}
 	}
 }
 
@@ -733,6 +748,13 @@ type peerConnection struct {
 	// a quiescent state at the moment of close.
 	pubMu   sync.RWMutex
 	stopped bool
+
+	// unreachableReported dedups Unreachable notifications during a drop
+	// burst so we don't saturate unreachableCh (capacity 100) when a single
+	// full send channel fires pushMessages hundreds of times per second.
+	// CAS-set on first drop, reset by sendMessages after a successful
+	// stream.Send() (peer confirmed responsive → next drop re-signals Raft).
+	unreachableReported atomic.Bool
 }
 
 // pushMessages pushes a batch of messages to the specified priority queue.
@@ -780,6 +802,15 @@ func (conn *peerConnection) pushMessages(priority int, msgs []raftpb.Message) bo
 			"priority": priority,
 		}).Errorf("Channel full")
 		conn.sendQueueFullCounter[priority].Add(context.Background(), 1)
+
+		// Signal Unreachable to Raft so it throttles this peer's Progress
+		// to StateProbe instead of continuing optimistic replication.
+		// Dedup via CAS: sendMessages resets the flag on the next successful
+		// stream.Send(), letting subsequent drops re-signal if the channel
+		// keeps refilling.
+		if conn.unreachableReported.CompareAndSwap(false, true) {
+			conn.pushUnreachable(conn.peerID)
+		}
 
 		return false
 	}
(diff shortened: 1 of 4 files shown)


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1519#issuecomment-4903512051

internal/infra/node/transport.go
```diff
@@ -781,6 +803,15 @@ func (conn *peerConnection) pushMessages(priority int, msgs []raftpb.Message) bo
 		}).Errorf("Channel full")
 		conn.sendQueueFullCounter[priority].Add(context.Background(), 1)
 
+		// Signal Unreachable to Raft so it throttles this peer's Progress
+		// to StateProbe instead of continuing optimistic replication.
+		// Dedup via CAS: sendMessages resets the flag on the next successful
+		// stream.Send(), letting subsequent drops re-signal if the channel
+		// keeps refilling.
+		if conn.unreachableReported.CompareAndSwap(false, true) {
```
NumaryBot, line 815: 🟠 **[major] Retry unreachable after enqueue failure**

When the per-peer queue first fills while `unreachableCh` is already full, `CompareAndSwap` sets `unreachableReported` to true even though `pushUnreachable` returns false and no `ReportUnreachable` will be delivered for this peer. Subsequent drops in the same saturation burst are then suppressed until a successful `sendMessages`, which may not happen for the stuck peer, so the EN-1043 drop path can still silently miss the signal under exactly the high-load condition where `unreachableCh` overflows.

gfyrag: Legit finding, confirmé. Fixed in 22c86bdeb.

Reordering: on `pushUnreachable` failure, we roll `unreachableReported` back to `false` so a subsequent drop retries. The flag now reflects "Raft has been notified", not "we tried".

Added `TestPeerConnection_PushMessages_RollsBackFlagOnEmitFailure` to pin the contract — a stubbed emit that first returns `false` must not permanently silence the dedup path.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1519#issuecomment-4903512051

Azorlogh (approved):

PR #1521, opened 2026-07-07
feat(EN-1470): opt-in expandVolumes on GET account

gfyrag:
## Summary

v2 populated per-asset volumes inline on `GET /v2/{ledger}/accounts/{addr}`; v3 alpha.7 never populated the field, forcing callers migrating from v2 into a second aggregation call. This restores access to volumes on the single-account GET, opt-in via `?expandVolumes=true` — mirroring the pattern already used on `POST /v3/{ledger}/transactions`.

- `handleGetAccount` reads the query param and, when set, runs `AggregateVolumes` with a `hardcoded_exact` filter on the address. The result is folded into `Account.Volumes` as `map<asset, {input, output, balance}>`.
- `Account.MarshalJSON` emits `volumes` only when populated (`omitempty`), so the default read path is unchanged.
- OpenAPI documents the new query parameter.
- `api-comparison.md` updated to describe the opt-in.

Default (no param) behaviour: no extra scan, no `volumes` in the response. Callers only pay for the aggregation when they ask.

## Design choice: opt-in, not always-on

v2 populated volumes unconditionally because volumes were a first-class stored attribute — cheap to look up. v3 stores logs and computes volumes on the fly via aggregation, so always-on would double the latency and load on every GET-account read. Opt-in matches v3's existing `expandVolumes` pattern on transactions (see `docs/technical/contributing/api-comparison.md:473`).

Non-goals:
- **List endpoint** (`GET /v3/{ledger}/accounts`) — the fan-out would be prohibitive. Callers who need volumes across many accounts should use a prepared-query `AGGREGATE_VOLUMES` or hit `GET /v3/{ledger}/volumes` directly.
- `effectiveVolumes` — remains removed in v3 (see `api-comparison.md:474`).

## Test plan

- [x] New `TestHandleGetAccount_ExpandVolumes` — asserts the filter is exact-match on the requested address, the response body has the flat `volumes: {USD/2: {input, output, balance}}` map, and balance = input − output.
- [x] New `TestHandleGetAccount_NoExpandVolumesLeavesFieldOff` — asserts `AggregateVolumes` is NOT called and the response omits `volumes` when the param is absent.
- [x] All existing `TestHandleGetAccount_*` tests still pass.

## Follow-up

Once merged, `Account.volumes` is no longer purely-informational-but-never-populated. If [EN-1467](https://formance-team.atlassian.net/browse/EN-1467) has already marked it `deprecated: true`, revert that marker (should be a one-line spec edit).

Ticket: [EN-1470](https://formance-team.atlassian.net/browse/EN-1470) (epic: [EN-867](https://formance-team.atlassian.net/browse/EN-867))

[EN-1467]: https://formance-team.atlassian.net/browse/EN-1467?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/proto/commonpb/common.pb.json.go
+++ b/internal/proto/commonpb/common.pb.json.go
@@ -149,20 +149,24 @@ func (x *PostCommitVolumes) MarshalJSON() ([]byte, error) {
 	})
 }
 
-// MarshalJSON implements json.Marshaler for Account.
+// MarshalJSON implements json.Marshaler for Account. Volumes are emitted only
+// when populated (opt-in via `expandVolumes=true` on the GET-account handler);
+// the default read path leaves them off to preserve the lightweight response.
 func (x *Account) MarshalJSON() ([]byte, error) {
 	return json.Marshal(&struct {
-		Address       string         `json:"address,omitempty"`
-		Metadata      map[string]any `json:"metadata,omitempty"`
-		FirstUsage    *Timestamp     `json:"firstUsage,omitempty"`
-		InsertionDate *Timestamp     `json:"insertionDate,omitempty"`
-		UpdatedAt     *Timestamp     `json:"updatedAt,omitempty"`
+		Address       string                         `json:"address,omitempty"`
+		Metadata      map[string]any                 `json:"metadata,omitempty"`
+		FirstUsage    *Timestamp                     `json:"firstUsage,omitempty"`
+		InsertionDate *Timestamp                     `json:"insertionDate,omitempty"`
+		UpdatedAt     *Timestamp                     `json:"updatedAt,omitempty"`
+		Volumes       map[string]*VolumesWithBalance `json:"volumes,omitempty"`
 	}{
 		Address:       x.GetAddress(),
 		Metadata:      MetadataToAnyMap(x.GetMetadata()),
 		FirstUsage:    x.GetFirstUsage(),
 		InsertionDate: x.GetInsertionDate(),
 		UpdatedAt:     x.GetUpdatedAt(),
+		Volumes:       x.GetVolumes(),
 	})
 }
 
--- a/docs/technical/contributing/api-comparison.md
+++ b/docs/technical/contributing/api-comparison.md
@@ -51,7 +51,7 @@ This document compares the POC's API with the original Formance ledger API and d
 | Get account type | ✅ | ❌ | Get details of a specific type |
 | Remove account type | ✅ | ❌ | Remove a type from a ledger |
 | **Accounts (Read)** |
-| Get account | ✅ | ✅ | Includes volumes per asset |
+| Get account | ✅ | ✅ | Opt-in per-asset volumes via `expandVolumes=true` (v2 populated them unconditionally; v3 gates the extra aggregation scan behind the flag) |
 | List accounts | ✅ | ✅ | Supports rich boolean filter (metadata equality/range/existence, address) with schema validation and cursor pagination |
 | Get account balances | ⚠️ | ✅ | Included in account volumes |
 | Get account volumes | ✅ | ✅ | Returns input/output/balance per asset |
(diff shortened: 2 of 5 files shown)


internal/proto/commonpb/common.pb.json.go
```diff
@@ -149,20 +149,24 @@ func (x *PostCommitVolumes) MarshalJSON() ([]byte, error) {
...
+		InsertionDate *Timestamp                     `json:"insertionDate,omitempty"`
+		UpdatedAt     *Timestamp                     `json:"updatedAt,omitempty"`
+		Volumes       map[string]*VolumesWithBalance `json:"volumes,omitempty"`
 	}{
 		Address:       x.GetAddress(),
 		Metadata:      MetadataToAnyMap(x.GetMetadata()),
 		FirstUsage:    x.GetFirstUsage(),
 		InsertionDate: x.GetInsertionDate(),
 		UpdatedAt:     x.GetUpdatedAt(),
+		Volumes:       x.GetVolumes(),
```
NumaryBot, line 169: 🟠 **[major] Keep account volumes out of default JSON marshaling**

When an account already has volume entries, the real controller populates `Account.Volumes` in `scanAccount`/`EnrichAccounts` before the HTTP handler sees it, so adding this field to the global `Account.MarshalJSON` emits `volumes` even when `expandVolumes` is absent. This also affects list-account responses, where the PR explicitly avoids a fan-out volume expansion, because those accounts are marshaled through the same method.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1521#issuecomment-4903608956

internal/proto/commonpb/common.pb.json.go
```diff
@@ -149,20 +149,24 @@ func (x *PostCommitVolumes) MarshalJSON() ([]byte, error) {
...
-		Metadata      map[string]any `json:"metadata,omitempty"`
-		FirstUsage    *Timestamp     `json:"firstUsage,omitempty"`
-		InsertionDate *Timestamp     `json:"insertionDate,omitempty"`
-		UpdatedAt     *Timestamp     `json:"updatedAt,omitempty"`
+		Address       string                         `json:"address,omitempty"`
+		Metadata      map[string]any                 `json:"metadata,omitempty"`
+		FirstUsage    *Timestamp                     `json:"firstUsage,omitempty"`
+		InsertionDate *Timestamp                     `json:"insertionDate,omitempty"`
+		UpdatedAt     *Timestamp                     `json:"updatedAt,omitempty"`
+		Volumes       map[string]*VolumesWithBalance `json:"volumes,omitempty"`
```
NumaryBot, line 162: 🟠 **[major] Account volumes included in default JSON marshaling, breaking opt-in behavior**

Adding `Volumes` to the global `Account.MarshalJSON` method means that any account object that already has its `Volumes` field populated (e.g., by `scanAccount`/`EnrichAccounts` in the controller layer) will emit a `volumes` key in JSON even when the caller did not pass `expandVolumes`. This affects both the default `GET /accounts/{address}` path and the `GET /accounts` list path, where volume expansion was explicitly meant to be avoided. The intended opt-in behavior is not preserved.

**Suggestion:** Do not serialize `Volumes` in the shared `Account.MarshalJSON`. Instead, introduce a separate response DTO or a custom marshaling path that is only invoked when `expandVolumes` is explicitly requested (e.g., a wrapper type or a dedicated handler-level serialization step that conditionally includes volumes).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1521#issuecomment-4903608956

gfyrag:
Closing per general directive "align doc on code, not the reverse". This PR was changing the wire (Account.MarshalJSON stops dropping the Volumes field) to restore v2's inline populate. The right move under the new principle is to leave the wire as-is and doc it: EN-1467 will describe `Account.volumes` as unpopulated on the current wire, with the pointer to prepared queries / GET /volumes for balances.

If v2-parity populate is later requested as an explicit feature request, we can revisit; for now the field stays defined in the schema but omitted from the wire, matching the pre-EN-1470 v3.0.0 shape.

EN-1470 marked as won't-do.

PR #1523, opened 2026-07-07
fix(EN-1045): removed-member registry blocks silent rejoin

gfyrag:
## Summary

Fixes [EN-1045](https://formance-team.atlassian.net/browse/EN-1045): a peer removed via `ConfChangeRemoveNode` could silently rejoin the cluster and be auto-promoted back to voter within ~100 ms — during a scale-down, this made the operator's `remove-node` silently fail as long as the target pod was still alive.

The fix introduces a **removed-member registry** keyed by `(nodeID, instanceID)` where `instanceID` is a 16-byte UUID identifying a specific (pod, PVC) incarnation. A rejoin attempt with a blacklisted tuple is refused at `JoinAsLearner` admission (and again at `checkAndPromoteLearners` as belt-and-suspenders). A fresh pod at the same `nodeID` with a new PVC — i.e. a legitimate scale-up after scale-down — comes with a fresh `instanceID` and joins normally.

## Design doc

Full design (goals, storage layout, propagation paths for both consensus and force removal, crash-window analysis, bootstrap-seed guarantee boundary, alternatives considered) lives in [`docs/technical/architecture/subsystems/consensus/removed-member-registry.md`](docs/technical/architecture/subsystems/consensus/removed-member-registry.md).

## Guarantee scope

The "never rejoins" guarantee applies to nodes that joined the cluster via `JoinAsLearner`. Bootstrap-seed voters (pod-0 under `--bootstrap`) are outside the guarantee because `PeerInfo` (discovery) has no `instance_id` field and the seed never goes through `AddLearner` — so on other nodes' peer rows its `InstanceID` is empty and `RemoveNode` on it proposes without a Context. In practice this is acceptable: scale-down removes the highest ordinal first (never pod-0), and self-removal is refused. Universal coverage would require adding `instance_id` to `PeerInfo` — tracked as follow-up in the doc.

## Compatibility

`instance_id` is a **mandatory** field on `JoinAsLearner`. Mixed-version joins (old-binary peer → new-binary leader) are rejected by design: no admit-with-WARN path. A peer's `INSTANCE_ID` marker is written before its first RPC and survives restarts alongside the WAL, so a peer that has already booted once against a new binary always has one; a peer whose PVC is reprovisioned generates a fresh one at first boot.

## Data plumbing

- **Proto**: `PeerAddress.instance_id`, `JoinAsLearnerRequest.instance_id`, new `RemovedMemberEntry` message. The consensus-path ConfChange context reuses the existing `membership.ConfChangeContext` JSON struct (populated with only the `InstanceID` field on RemoveNode) — no new proto message for that path.
- **WAL marker**: `INSTANCE_ID` file colocated with `CLUSTER_JOINED`; helpers `EnsureInstanceID`, `ReadInstanceID`, `WriteInstanceID`, `GenerateInstanceID` in `internal/storage/wal/`. Marker filename constants regrouped in `markers.go`.
- **Pebble**: new `SubGlobRemovedMembers = 0x15` under `ZoneGlobal`. Keyed by `nodeID BE 8 || instance_id 16`. Owned by `PeerStore` (unified with the peer directory: same package, same underlying store, same lifecycle).
- **Membership**: extended `ConfChangeContext` JSON with `InstanceID []byte`; the leader stores instance IDs in `PeerAddress.instance_id` and looks them up before proposing `RemoveNode`.

## Propagation

- **Consensus path** — `Node.RemoveNode` packs the target's `instance_id` into the `ConfChangeV2.Context`. `Membership.WriteConfChange` decodes and, on `ConfChangeRemoveNode`, writes the `RemovedMemberEntry` in the **same** `dal.WriteSession` as the peer row delete. Cross-node convergence via FSM determinism (invariant #2). When the row has no `instance_id` (phantom learner via admin `cluster.AddLearner` without a booted pod), the leader proposes without a Context and no blacklist entry is written — nothing to blacklist.
- **Force path** — `Node.ForceRemoveNode` reads the target's `instance_id` from Membership, then extends the existing WAL(ConfState) → Pebble(peer row) sequence to a single Pebble batch (`Membership.UnregisterAndBlacklist`) that atomically deletes the peer row AND writes the blacklist entry. Followers converge via the next snapshot they receive — same channel that today already propagates the force-removed ConfState. Crash window between WAL write and Pebble batch is bounded by two fsyncs.

## Enforcement

- `JoinAsLearner` admission: `instance_id` required (16 bytes); rejects empty with `codes.InvalidArgument`, blacklisted tuple with `codes.FailedPrecondition` and the `ledgerctl cluster forget-removed <id> <hex>` remediation message.
- `checkAndPromoteLearners` refuses to propose `ConfChangeAddNode` for a blacklisted learner. Skips (defers) when the learner has no `instance_id` yet (row created via admin `cluster.AddLearner`, waiting for the pod to boot and refresh via `JoinAsLearner`).
- Registry reads use direct Pebble read handles (`NewDirectReadHandle`); the check runs before any proposal, outside the FSM apply hot path, so invariants #6/#9 (preload + coverage gate) don't apply.

## Test plan

- [x] Unit tests
  - [x] `wal.EnsureInstanceID` / `ReadInstanceID` / `WriteInstanceID` — round-trip, idempotence, wrong-length rejection, refuse-overwrite
  - [x] `PeerStore.IsRemoved` / `LoadAllRemoved` / `AnyRemovedForNodeID` — including the discriminator test that pins the fresh-PVC-at-reused-ordinal case, and empty/wrong-length instance_id rejection
  - [x] `Membership.WriteConfChange` — consensus RemoveNode with context writes both peer delete + blacklist atomically; empty context deletes peer without blacklist (phantom-learner path)
  - [x] `Membership.UnregisterAndBlacklist` — atomic force path
- [x] `GOROOT= go build ./...`, `go vet ./...`, `go test ./...` — full unit suite green
- [x] `just pre-commit` — 0 lint issues
- [x] E2E tests: `tests/e2e/cluster/remove_node_test.go` (learner remove + follower forwarding) pass locally
- [ ] **Follow-up**: E2E test that reproduces the exact EN-1045 trace (pause STS shrink, verify leader refuses re-add; then scale up with fresh PVC, verify accept).
- [ ] **Follow-up**: Antithesis driver `singleton_driver_scaledown_alive`.

## Deferred to follow-ups (tracked in the design doc)

- Add `instance_id` to `PeerInfo` (discovery) so the bootstrap-seed guarantee becomes universal.
- `ledgerctl cluster list-removed` / `forget-removed` CLI (needs new admin RPCs; `PeerStore.LoadAllRemoved` and `DeleteRemovedInSession` are already in place).
- Cross-node consistency check via admin RPC (registry is not an audit-chain projection — see the doc for why invariant #8 doesn't apply).

## Files of interest for review

- `docs/technical/architecture/subsystems/consensus/removed-member-registry.md` — start here
- `internal/infra/membership/peer_store.go` — unified peer directory + removed-member registry
- `internal/infra/membership/membership.go` — `UnregisterAndBlacklist`, `IsRemoved`, `GetInstanceID`, extended `WriteConfChange`
- `internal/infra/node/node.go` — `RemoveNode` (packs context), `ForceRemoveNode` (atomic batch), `checkAndPromoteLearners` (belt-and-suspenders check)
- `internal/adapter/grpc/server_bootstrap.go` — `JoinAsLearner` admission
- `internal/storage/wal/instance_id_marker.go` + `markers.go` — WAL marker helpers

[EN-1045]: https://formance-team.atlassian.net/browse/EN-1045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/adapter/grpc/server_bootstrap.go
```diff
@@ -159,7 +163,39 @@ func (impl *ClusterBootstrapServiceServerImpl) JoinAsLearner(ctx context.Context
 		return clusterbootstrappb.NewClusterBootstrapServiceClient(conn).JoinAsLearner(outCtx, req)
 	}
 
-	if err := impl.membership.AddLearner(ctx, req.GetNodeId(), req.GetRaftAddress(), req.GetServiceAddress()); err != nil {
+	// EN-1045: every peer must present its 16-byte instance_id — clients
+	// acquire one at first boot via wal.EnsureInstanceID.
+	if len(req.GetInstanceId()) != 16 {
+		return nil, status.Errorf(codes.InvalidArgument, "instance_id must be 16 bytes, got %d", len(req.GetInstanceId()))
```
Azorlogh, line 169: Rolling-upgrade compatibility here contradicts the PR description and design doc, which say a legacy peer with no `instance_id` is *admitted* (with a WARN). This hard-rejects with `InvalidArgument` instead. Same story in `PeerStore.IsRemoved`, which *errors* on a non-16-byte id rather than short-circuiting to `false` (pinned by `TestPeerStore_IsRemovedRejectsEmptyInstanceID`), contrary to the doc's "short-circuits on empty ID".

Blast radius is small — restart re-join is gated by the `CLUSTER_JOINED` marker, so only a genuinely-new *old-binary* peer joining a new-binary leader hits it. But since this targets `release/v3.0` (N/N-1), please either implement the documented admit-with-WARN, or update the description/doc to state `instance_id` is now mandatory and mixed-version new-peer joins are rejected by design.

internal/infra/node/node.go
```diff
@@ -2178,11 +2189,34 @@ func (node *Node) RemoveNode(ctx context.Context, nodeID uint64) error {
...
+		// instance_id here means the peer's row has no identity —
+		// either the row was created via the admin cluster.AddLearner
+		// path without the target ever booting (phantom learner), or
+		// it is a bootstrap initial peer that has not yet joined.
+		// Both cases are legal: propose the removal without a Context,
+		// and WriteConfChange will delete the peer row without writing
+		// a blacklist entry (nothing to blacklist).
+		var ccCtx []byte
+
+		if instanceID, ok := node.membership.GetInstanceID(nodeID); ok && len(instanceID) == 16 {
```
Azorlogh: This is where the "a member removed via `ConfChangeRemoveNode` never rejoins" guarantee quietly becomes conditional. A node that became a voter through the *initial bootstrap ConfState* (pod-0 under `--bootstrap`) never carries an `instance_id` on the other nodes: discovery's `PeerInfo` has no `instance_id` field and pod-0 never goes through an `AddLearner` ConfChange, so `registerInitialPeers` leaves its row `nil` on peers. Here `ok && len(instanceID) == 16` is false, so `RemoveNode` proposes without a context and no `RemovedMemberEntry` is written.

Not a blocker in practice — scale-down removes the highest ordinal first (never pod-0), and `RemoveNode` refuses self-removal. But please either scope the doc's guarantee to nodes that joined via `JoinAsLearner`, or add `instance_id` to `PeerInfo` so the seed's identity propagates on discovery and the guarantee becomes universal.

misc/proto/raft_cmd.proto
```diff
@@ -756,11 +756,41 @@ message CacheSnapshotMeta {
...
   string service_address = 3;
+  bytes instance_id = 4;
+}
+
+// RemoveNodeContext is embedded in ConfChangeV2.Context on a
+// ConfChangeRemoveNode to atomically carry the removed peer's instance_id
+// into the FSM apply path, so the same log entry that removes the peer also
+// writes its RemovedMemberEntry (EN-1045). Not emitted on the ForceRemoveNode
+// path (see removed-member-registry.md, "Force path").
+message RemoveNodeContext {
```
Azorlogh: `RemoveNodeContext` appears to be dead code — the consensus path carries the id via the existing `membership.ConfChangeContext` (JSON), and there are no non-generated references to `RemoveNodeContext`. Reusing `ConfChangeContext` is fine, but the design doc presents this message as *the* mechanism, so it's misleading to leave it here. Suggest dropping it from the proto and regenerating (which also removes the unused marshal/reader/vtproto/dethash scaffolding), and fixing the doc.

internal/infra/membership/peer_store.go
```diff
@@ -130,3 +173,207 @@ func (p *PeerStore) LoadAll() (map[uint64]ConfChangeContext, error) {
...
+
+// IsRemoved reports whether a tombstone exists for (nodeID, instanceID).
+// instanceID must be exactly 16 bytes — every cluster member acquires one
+// at first boot via wal.EnsureInstanceID.
+func (p *PeerStore) IsRemoved(nodeID uint64, instanceID []byte) (bool, error) {
+	if len(instanceID) != 16 {
+		return false, fmt.Errorf("PeerStore.IsRemoved: instance id must be 16 bytes, got %d", len(instanceID))
+	}
+
+	handle, err := p.store.NewDirectReadHandle()
```
Azorlogh, line 269: The design doc says the `JoinAsLearner` / `checkAndPromoteLearners` reads hit an in-memory KeyStore, and bases its invariant-#6/#9 exemption on that. In fact `IsRemoved` opens a fresh `NewDirectReadHandle` against live Pebble on every call. Not a correctness issue (the leader reads committed state), but the doc's justification rests on a false premise — please correct it.

internal/infra/node/node.go
```diff
@@ -2241,17 +2291,29 @@ func (node *Node) ForceRemoveNode(ctx context.Context, nodeID uint64) error {
...
 
-		if err := node.membership.Unregister(nodeID); err != nil {
-			return fmt.Errorf("removing peer after force-remove: %w", err)
+		// EN-1045: when we know the target's identity, land the
+		// blacklist entry atomically with the peer row delete in a
+		// single Pebble batch. For phantom learners without an
+		// identity (see hasIdentity comment above), fall back to
+		// the plain peer row delete — nothing to blacklist.
+		if hasIdentity {
+			if err := node.membership.UnregisterAndBlacklist(nodeID, instanceID, uint64(time.Now().UnixMicro())); err != nil {
```
Azorlogh, line 2434: Nit: `removed_at` is stamped with wall-clock `time.Now().UnixMicro()`, but the proto comment and design doc call it an "HLC timestamp". Audit-only so harmless — just align the wording.

Azorlogh (commented):
Nicely-scoped fix with careful storage/atomicity reasoning and a strong design doc. I verified the parts the design leans on hardest and they hold up:

- The registry rides along in snapshots automatically — `dal.CreateSnapshot` is a whole-DB `pebble.Checkpoint` and restore is wipe-and-replace — so the force path's "followers converge via the next snapshot" is correct.
- The seal/consistency hash (`computeStateHash`) covers only attribute zones, so `SubGlobRemovedMembers` adds no hashing burden.
- The consensus write is deterministic and atomic with the peer-row delete.

The primary EN-1045 trace (scale-down removing the highest-ordinal pod) is fully covered — that pod always joined via `JoinAsLearner` and has a known `instance_id`.

I left inline comments on a few things to resolve — mostly doc-vs-code mismatches, plus one dead proto message and one narrow guarantee gap. None look like hard blockers; the main asks are reconciling the rolling-upgrade compatibility story and scoping the "never rejoins" guarantee. Details inline.

gfyrag:
Thanks for the thorough review @Azorlogh! All five points addressed in `d1516ba1c`:

1. **Rolling-upgrade compat** (`server_bootstrap.go:169`) — kept the code strict (`instance_id` mandatory, `InvalidArgument` on empty). Updated the PR description and design doc to explicitly state this: no admit-with-WARN path, mixed-version new-peer joins rejected by design. Rationale: `INSTANCE_ID` is written before the first RPC and survives restarts alongside the WAL, so any peer that has booted once against the new binary always carries one — the failure mode you flagged (genuinely-new old-binary peer against new-binary leader) is not a case we need to handle.

2. **Bootstrap-seed guarantee** (`node.go:2203`) — added a "Bootstrap Seed and the Guarantee Boundary" section in the design doc scoping the "never rejoins" guarantee to nodes that joined via `JoinAsLearner`. The `PeerInfo`-carries-instance_id fix is deferred as a follow-up (also noted in the doc and PR description).

3. **Dead `RemoveNodeContext` proto** (`raft_cmd.proto:779`) — dropped the message, regenerated protos (scaffolding gone with it), and updated the design doc to show the consensus path reuses `membership.ConfChangeContext` JSON (which is what the code actually does).

4. **Doc-vs-code on read path** (`peer_store.go:269`) — corrected the design doc: `IsRemoved` opens a fresh `NewDirectReadHandle` on Pebble; the invariant #6/#9 exemption is justified by the check running outside the FSM apply hot path (leader-only, before any proposal), not by "in-memory KeyStore".

5. **HLC nit** (`node.go:2300`) — aligned wording in proto comment, Go docstring, and design doc: it's a wall-clock microsecond timestamp on the force path, and left 0 on the consensus path to preserve invariant #2.

Azorlogh (approved):
All five points from the earlier review are addressed in d1516ba1c:

- Rolling-upgrade compat: doc now states `instance_id` is mandatory and mixed-version joins are rejected by design — matches the code.
- "Never rejoins" guarantee: scoped to `JoinAsLearner`-joined nodes, with an explicit "Bootstrap Seed and the Guarantee Boundary" section and a deferred `PeerInfo.instance_id` follow-up.
- Dead `RemoveNodeContext` proto removed and regenerated (no dangling refs).
- Enforcement reads correctly documented as direct Pebble lookups (no in-memory cache).
- `removed_at` wording corrected to wall-clock micros, force-path-only.

The pod-0 boundary is now documented rather than silent, which is a fine call for the EN-1045 scale-down scope. Affected packages build clean. LGTM.

internal/adapter/grpc/server_cluster.go
```diff
@@ -327,7 +327,12 @@ func (impl *ClusterServiceServerImpl) AddLearner(ctx context.Context, req *clust
 		return client.AddLearner(ctx, req)
 	}
 
-	if err := impl.membership.AddLearner(ctx, req.GetNodeId(), req.GetRaftAddress(), req.GetServiceAddress()); err != nil {
+	// The admin cluster.AddLearner RPC (fctl / ledgerctl "cluster add-learner")
```
NumaryBot, line 330: 🟠 **[major] Pre-added learner rows with nil instance_id are never refreshed, blocking auto-promotion**

When an operator calls the admin `cluster.AddLearner` path before the new pod boots, a peer row is written with a nil `instance_id`. When the pod later calls `JoinAsLearner`, `Node.AddLearner` returns `ErrNodeAlreadyInCluster` and `tryAddLearner` treats `AlreadyExists` as success without updating the row. Because the row still has no `instance_id`, the `checkAndPromoteLearners` guard never promotes the node — the add-then-boot flow is permanently stalled.

**Suggestion:** When `ErrNodeAlreadyInCluster` is returned and the existing row has a nil `instance_id`, update the row with the `instance_id` supplied by the joining pod before returning success, so that subsequent promotion checks can proceed normally.

internal/adapter/grpc/server_bootstrap.go
```diff
@@ -159,7 +163,39 @@ func (impl *ClusterBootstrapServiceServerImpl) JoinAsLearner(ctx context.Context
...
+	}
+
+	// Refuse a peer whose (nodeID, instance_id) has been blacklisted.
+	// Errors reading the store surface as Internal so an operator sees
+	// the anomaly and can investigate.
+	removed, err := impl.membership.IsRemoved(req.GetNodeId(), req.GetInstanceId())
```
NumaryBot, line 175: 🟠 **[major] Removed-member blacklist check is not re-evaluated under the conf-change lock**

In a concurrent remove/rejoin race, `JoinAsLearner` can pass the Pebble `IsRemoved` check before the remove ConfChange commits, then block behind `confChangeMu`. After the removal writes the tombstone, `AddLearner` proceeds without re-checking, allowing the blacklisted instance to rejoin the Raft configuration as a learner. The promotion guard prevents it from becoming a voter, but the node has still been re-inserted into the Raft config, violating the eviction guarantee.

**Suggestion:** Repeat the `IsRemoved` check inside the serialized `AddLearner` code path (i.e., after acquiring `confChangeMu`) so that a tombstone written between the initial check and the conf-change proposal is not bypassed.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1523#issuecomment-4914998325

gfyrag:
Thanks @NumaryBot — both findings addressed in `8442dd444`.

### Finding 1 — Admin AddLearner + boot flow

`Node.AddLearner` now detects the "already a Raft member with a nil `instance_id` in the peer row" case (which is exactly the admin `cluster.AddLearner` + boot flow). Instead of returning `ErrNodeAlreadyInCluster`, it proposes a `ConfChangeUpdateNode` carrying the joining pod's real `instance_id` in its Context. `Membership.WriteConfChange` and `Node.finishReady` were extended to handle `ConfChangeUpdateNode` symmetrically with `AddLearnerNode` (write/refresh the peer row across all nodes).

Once the row is refreshed cluster-wide, `checkAndPromoteLearners` sees a valid `instance_id`, the blacklist check runs, and promotion proceeds. The add-then-boot stall is closed.

### Finding 2 — TOCTOU on JoinAsLearner IsRemoved

Added a re-check of `IsRemoved(nodeID, instance_id)` inside the `Node.AddLearner` closure, which runs while `confChangeMu` is held. Since `retryConfChange` waits for the previous ConfChange to fully commit (and therefore apply its Pebble writes) before releasing the mutex, any `RemoveNode` that committed between the admission-time `IsRemoved` check and this point has its tombstone visible on the second read. On a hit, we return the new `node.ErrNodeRemoved` (mapped to `codes.FailedPrecondition` in `server.go`) — the peer never enters the Raft config.

Belt-and-suspenders: the initial `IsRemoved` check in `server_bootstrap.go` stays (better error message in the common non-race case), the promotion-time check in `checkAndPromoteLearners` also stays.

Verified locally: `just pre-commit`, full unit suite, and `TestCluster/Remove_node` E2E all green.

internal/adapter/grpc/server_cluster.go
NumaryBot, line 330: 🟠 **[major] Pre-added learner row with nil instance_id is never refreshed, blocking auto-promotion**

When an operator calls AddLearner before the new pod boots, a peer row is written with a nil instance_id. When the pod later calls JoinAsLearner, Node.AddLearner returns ErrNodeAlreadyInCluster and tryAddLearner treats AlreadyExists as success without updating the row. Because the row still has no instance_id, the checkAndPromoteLearners guard never promotes the node — the add-then-boot flow is permanently stalled.

**Suggestion:** When ErrNodeAlreadyInCluster is returned and the existing row has a nil instance_id, update the row with the instance_id supplied by the joining pod before returning success, so that subsequent promotion checks can proceed normally.

internal/adapter/grpc/server_bootstrap.go
NumaryBot, line 175: 🟠 **[major] Removed-member blacklist check is not re-evaluated under the conf-change lock (TOCTOU race)**

In a concurrent remove/rejoin race, JoinAsLearner can pass the Pebble IsRemoved check before the remove ConfChange commits, then block behind confChangeMu. After the removal writes the tombstone, AddLearner proceeds without re-checking, allowing the blacklisted instance to rejoin the Raft configuration as a learner. Additionally, there is an async-apply subtlety: retryConfChange releases confChangeMu after finishReady resolves the conf-change future, but finishReady only submits committed entries to the async applier, so Membership.WriteConfChange may not have written the RemovedMemberEntry to Pebble yet when the waiting join's recheck (if added) runs.

**Suggestion:** Repeat the IsRemoved check inside the serialized AddLearner code path (i.e., after acquiring confChangeMu). Additionally, ensure the tombstone write via WriteConfChange is fully persisted before confChangeMu is released on the removal side, so that a recheck inside the lock is guaranteed to observe the tombstone.

internal/infra/node/node.go
```diff
@@ -2118,8 +2139,43 @@ func (node *Node) AddLearner(ctx context.Context, nodeID uint64, raftAddr, servi
...
+		}
+
 		if _, ok := status.Progress[nodeID]; ok {
-			return ErrNodeAlreadyInCluster
+			// Peer already a Raft member. Refresh the peer row via
+			// UpdateNode when the existing row has an empty
+			// instance_id and we have a fresh 16-byte one — this is
+			// the admin cluster.AddLearner + boot flow. Otherwise
+			// return AlreadyExists (idempotent join).
+			existing, hasRow := node.membership.GetInstanceID(nodeID)
```
NumaryBot, line 2182: 🟠 **[major] Stale instance_id not refreshed when a reprovisioned learner rejoins**

If an already-added learner is reprovisioned before promotion, the replacement pod joins with a different 16-byte instance_id. The existing idempotency path only refreshes rows whose existing ID is empty (nil), so a reprovisioned pod's new ID is never stored. A later removal blacklists the stale (old) ID, allowing the still-alive replacement pod to bypass the removed-member registry entirely.

**Suggestion:** When ErrNodeAlreadyInCluster is returned, always overwrite the stored instance_id with the one supplied by the joining pod (not only when the existing value is nil), so the registry always tracks the current incarnation.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1523#issuecomment-4914998325

gfyrag:
Second pass of NumaryBot findings on \`8442dd444\` addressed in \`eb388aeec\`. Analysis + fixes:

### Duplicates (findings 1, 2, 3)
Same as the previous round on `server_cluster.go:330` and `server_bootstrap.go:175` — already addressed in `8442dd444` (UpdateNode refresh path + serialized IsRemoved re-check). NumaryBot's second review appears to have not seen those changes. No action needed.

### Finding 4 — Async-apply subtlety in the TOCTOU race
This is a real deeper issue: `finishReady` resolves the ConfChange future **before** `applier.Submit` runs, so `retryConfChange` returns while the FSM apply (which writes the `RemovedMemberEntry` tombstone to Pebble) is still queued behind the async applier. A concurrent `JoinAsLearner` acquiring `confChangeMu` next runs its re-check (added in the previous round) but reads Pebble **before** the tombstone lands — race still open.

Fix: `retryConfChange` now accepts an optional `postApplyFn` that runs while `confChangeMu` is still held. `Node.RemoveNode` uses it to poll `PeerStore.IsRemoved(nodeID, instanceID)` on Pebble until the tombstone is visible (5 s deadline, 5 ms tick). Only released after the tombstone is durable; the next AddLearner acquiring the mutex is guaranteed to observe it.

`AddLearner` and `PromoteLearner` pass `nil` — their post-conditions don't require an FSM-visible side-effect before returning.

### Finding 5 — Reprovisioned learner bypasses blacklist
Real and correctly diagnosed. Previously, the UpdateNode refresh triggered only when the stored `instance_id` was **empty**. A reprovisioned pod joining with a fresh `instance_id` (Y) into a row still holding the old (X) would NOT refresh; a subsequent RemoveNode would blacklist X, leaving Y free to rejoin.

Fix: refresh condition changed from `len(existing) == 0` to `!bytes.Equal(existing, instanceID)` — any mismatch triggers the ConfChangeUpdateNode. Same idempotent-join fast-path when identity matches.

Verified locally: `go build`, `just pre-commit` (0 lint), full unit suite, and `TestCluster/Remove_node` E2E all green.

internal/infra/node/node.go
```diff
@@ -2157,14 +2229,26 @@ func (node *Node) PromoteLearner(ctx context.Context, nodeID uint64) error {
...
+// packed into the ConfChange context so every node's FSM apply lands a
+// matching RemovedMemberEntry atomically with the peer row delete — a
+// still-alive pod at the same nodeID cannot silently rejoin and be
+// auto-promoted after this ConfChange commits.
 func (node *Node) RemoveNode(ctx context.Context, nodeID uint64) error {
-	return node.retryConfChange(ctx, nodeID, "RemoveNode", func() error {
+	// Look up the target peer's identity outside the closure so the
+	// postApplyFn below can reuse it to wait for the tombstone to become
+	// visible in Pebble.
+	instanceID, hasIdentity := node.membership.GetInstanceID(nodeID)
```
NumaryBot: 🟠 **[major] RemoveNode reads instance_id before concurrent ConfChangeUpdateNode commits, blacklisting the wrong ID**

If a reprovisioned learner is concurrently refreshing its peer row via ConfChangeUpdateNode, RemoveNode can read the old (pre-refresh) instance_id from the peer store before acquiring the conf-change lock. It then waits for the update to commit, removes the learner, but only blacklists the stale ID. The replacement pod's actual (nodeID, instance_id) is never tombstoned and can rejoin freely, defeating the eviction guarantee.

**Suggestion:** Perform the instance_id lookup inside the serialized proposal path — after acquiring confChangeMu and after any earlier queued conf changes have fully applied — so RemoveNode always tombstones the identity that is current at the moment of removal.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1523#issuecomment-4914998325

internal/infra/node/node.go
```diff
@@ -391,7 +398,7 @@ func NewNode(
 	// writes self for Bootstrap/Restore, but the restart branch never
 	// did — that's the gap the previous bootstrap hook used to cover
 	// before transport wiring moved into Membership.
-	if err := membership.Register(cfg.NodeID, cfg.AdvertiseAddr, cfg.ServiceAdvertiseAddr); err != nil {
+	if err := membership.Register(cfg.NodeID, cfg.AdvertiseAddr, cfg.ServiceAdvertiseAddr, cfg.InstanceID); err != nil {
```
NumaryBot, line 401: 🟠 **[major] Existing peer rows from pre-instance_id boots are never backfilled, allowing tombstone-free removal**

On an upgraded cluster whose peer rows were written before instance_id was introduced, each node writes its new INSTANCE_ID only to its own local Pebble row at startup. Because CLUSTER_JOINED nodes skip JoinAsLearner, the leader never learns the IDs of those existing members. When one such member is subsequently removed, the code follows the empty-ID path and writes no RemovedMemberEntry, so the still-alive pod can rejoin without hitting the blacklist.

**Suggestion:** After a node boots with an INSTANCE_ID on a cluster that already has a CLUSTER_JOINED marker, propagate the instance_id to the leader (e.g., via a lightweight heartbeat or a ConfChangeUpdateNode proposal) so the leader's peer store is backfilled before any removal can be attempted.

internal/infra/node/node.go
```diff
@@ -2091,7 +2107,11 @@ func (node *Node) retryConfChange(ctx context.Context, nodeID uint64, name strin
 		}
 
 		if committed {
-			return nil
+			if postApplyFn == nil {
+				return nil
+			}
+
+			return postApplyFn(ctx)
```
NumaryBot, line 2114: 🟠 **[major] Async tombstone apply leaves a window where a new leader can admit a removed instance**

The postApplyFn barrier only protects the leader that handled the RemoveNode RPC. If leadership changes after the remove ConfChange is committed to the Raft log but before a follower's async applier has persisted the RemovedMemberEntry to Pebble, that new leader can pass both JoinAsLearner blacklist checks against its local state and re-add the removed instance as a learner.

**Suggestion:** Ensure the tombstone write via WriteConfChange is fully persisted (synchronously or with a read barrier) on every node before it can act as leader and admit new join requests following the removal, not only on the node that originally proposed the removal.

gfyrag:
Third pass of NumaryBot findings. Analysis:

### Finding 6 (`node.go:2248`) — RemoveNode reads instance_id before acquiring confChangeMu
**Legit.** RemoveNode was reading `GetInstanceID` outside the closure — if a concurrent `ConfChangeUpdateNode` (reprovisioned learner refresh) commits between the read and the mutex acquisition, we'd blacklist the stale identity.

Fixed in `f9cced1d3`: `GetInstanceID` moved inside `proposeFn`, which runs while `confChangeMu` is held and after any prior ConfChange's `finishReady` has updated the in-memory cache. Captured in shared vars for `postApplyFn` to reuse.

### Finding 7 (`node.go:401`) — Pre-EN-1045 rows on upgraded clusters never backfilled
**Out of scope for this PR.** The design position (in PR description + `docs/technical/architecture/subsystems/consensus/removed-member-registry.md`) is explicit: `instance_id` is mandatory, no in-place upgrade compat from pre-EN-1045 builds. `release/v3.0` has no long-lived production clusters where this matters. A backfill mechanism (lightweight refresh RPC or boot-time UpdateNode when the row is empty) is a follow-up.

Documented as "Known Limitation" in the design doc for visibility.

### Finding 8 (`node.go:2114`) — Async tombstone visibility across leadership change
**Real but out of scope.** `postApplyFn` guarantees tombstone visibility on the leader that handled the RemoveNode RPC. Followers apply the same entry via their own async applier; leadership transferring during that catch-up window (single-digit ms) to a follower whose FSM hasn't yet applied the tombstone leaves an admission gap.

Proper fix is a leadership-transition FSM barrier (block admission on any leader until its FSM has caught up to the current commit index) — orthogonal architectural change, not EN-1045-specific.

Documented as "Known Limitation" alongside Finding 7. Bounded by follower applier catch-up, requires an unusual timing (leadership loss + immediate rejoin against the new leader).

Verified locally: build, vet, `just pre-commit` (0 lint), unit tests membership+node, all green.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1523#issuecomment-4914998325

internal/infra/node/node.go
NumaryBot, line 401: 🟠 **[major] Existing peer rows from pre-instance_id boots are never backfilled on upgrade**

On an upgraded cluster whose peer rows were written before instance_id was introduced, nodes that already have a CLUSTER_JOINED marker skip JoinAsLearner entirely. Each node only writes its new INSTANCE_ID to its own local Pebble row at startup; the leader's peer row for those existing members retains a nil instance_id. When such a member is subsequently removed, RemoveNode takes the empty-ID path and writes no RemovedMemberEntry, so the still-alive pod can rejoin without hitting the blacklist — silently defeating the eviction guarantee on upgraded clusters.

**Suggestion:** After a node boots with an INSTANCE_ID and a pre-existing CLUSTER_JOINED marker, propagate the instance_id to the leader (e.g., via a ConfChangeUpdateNode proposal or a lightweight membership heartbeat) so the leader's peer store is backfilled before any removal can be attempted.

internal/adapter/grpc/server_bootstrap.go
NumaryBot, line 175: 🟠 **[major] New leader may admit a removed instance before applying prior tombstone writes**

The postApplyFn barrier added to close the async-apply TOCTOU race only protects the node that originally handled the RemoveNode RPC. If leadership changes after the remove ConfChange is committed to the Raft log but before a follower's async applier has persisted the RemovedMemberEntry to Pebble, the new leader's IsRemoved check (and any re-check inside the lock) reads stale local state. The removed instance can therefore pass both blacklist checks on the new leader and be re-added as a learner, bypassing the eviction guarantee.

**Suggestion:** Ensure the tombstone written by WriteConfChange is durably persisted (synchronously or via a read barrier) on every node before that node can act as leader and admit new join requests, not only on the node that originally proposed the removal. One option is to gate leaderReady on confirmation that all prior committed entries have been applied by the local async applier.

gfyrag:
Fourth NumaryBot pass — all 9 findings on `f9cced1d3` are re-formulations of findings already addressed in earlier rounds. Per-finding mapping to prior responses (with commit + doc pointers):

| ID | Location | Status | Where handled |
|---|---|---|---|
| `de83a73f`, `7327c7d4` | server_cluster.go:330 | **Fixed** | `8442dd444` — `Node.AddLearner` proposes `ConfChangeUpdateNode` when the existing peer row has a nil `instance_id`; `Membership.WriteConfChange` extended to handle UpdateNode (refreshes the peer row). |
| `57abc52b`, `01b16125` | server_bootstrap.go:175 | **Fixed** | `8442dd444` (IsRemoved re-check inside `Node.AddLearner` closure) + `eb388aeec` (`postApplyFn` in `retryConfChange` polling Pebble until tombstone visible before releasing `confChangeMu`). |
| `c808576a` | node.go:2182 | **Fixed** | `eb388aeec` — refresh condition changed from `len(existing) == 0` to `!bytes.Equal(existing, instanceID)`. Any mismatch triggers ConfChangeUpdateNode. |
| `c7482546`, `be932b9c` | node.go:401 | **Known limitation** | Design doc §"Known Limitations" — for `release/v3.0` the "instance_id mandatory, no in-place upgrade compat" position holds; backfill of pre-EN-1045 rows is a follow-up. |
| `c100ce0a`, `d7b88dc2` | node.go:2114 | **Known limitation** | Design doc §"Known Limitations" — async tombstone visibility across leadership change is bounded by follower applier catch-up (single-digit ms in normal operation); proper fix (FSM barrier at leadership transition) is orthogonal to EN-1045. |

Confirmed the fixes are present in `f9cced1d3`:
- `internal/infra/node/node.go:2164` — IsRemoved re-check in AddLearner closure
- `internal/infra/node/node.go:2183` — `!bytes.Equal(existing, instanceID)` refresh condition
- `internal/infra/node/node.go:2190` — ConfChangeUpdateNode proposal on refresh
- `internal/infra/node/node.go:2319` — postApplyFn calling `waitForBlacklistApplied`
- `internal/infra/node/node.go:2330` — polling loop until tombstone visible
- `internal/infra/membership/membership.go:538` — WriteConfChange handles UpdateNode (writes peer row on refresh)

No further code changes this round.

PR #1525, opened 2026-07-07
chore(deps): upgrade go-libs/v5 to v5.7.0

sylr:
## What

Upgrade `github.com/formancehq/go-libs/v5` from `v5.3.0` → `v5.7.0`, then `go mod tidy`.

## Why

Keep the shared Formance go-libs dependency current.

## Verification

- `go build ./...` — passes
- `just pre-commit` (go generate + go mod tidy + golangci-lint, root + operator) — **0 issues**, no generated-code drift

Only `go.mod` / `go.sum` change; no source changes needed.


go.mod
```diff
@@ -70,7 +70,7 @@ require (
 	github.com/aws/smithy-go v1.24.2
 	github.com/cockroachdb/pebble/v2 v2.1.4
 	github.com/databricks/databricks-sql-go v1.10.0
-	github.com/formancehq/go-libs/v5 v5.3.0
+	github.com/formancehq/go-libs/v5 v5.7.0
```
NumaryBot, line 73: 🟠 **[major] Restore DEBUG env handling after go-libs bump**

When the server is configured with `DEBUG=true`, this dependency version no longer enables Ledger debug mode because `service.BindEnvToFlagSet` in go-libs now explicitly skips binding the `debug` flag from the `DEBUG` environment variable. Ledger relies on `service.Execute` for env binding and documents `DEBUG=true` as the way to enable debug logs, so `DEBUG=true ledger run` now leaves `cfg.Debug` false unless `--debug` is passed or Ledger adds its own env alias.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1525#issuecomment-4904521279

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1525#issuecomment-4904521279

go.mod
NumaryBot, line 73: 🟠 **[major] DEBUG env variable no longer binds debug flag after go-libs v5.7.0 upgrade**

With the go-libs v5.7.0 upgrade, `service.BindEnvToFlagSet` now explicitly skips binding the `debug` flag from the `DEBUG` environment variable. Ledger relies on `service.Execute(rootCmd)` for env-to-flag binding and documents `DEBUG=true` as the way to enable debug logs. Operators and deployment specs that emit `DEBUG=true` (e.g. `spec.Debug`) will now leave `cfg.Debug` false unless `--debug` is passed explicitly or Ledger adds its own compatibility binding.

**Suggestion:** Add a Ledger-side env alias for the debug flag (e.g. via `rootCmd.PersistentFlags().Lookup("debug").DefValue` or a manual `viper.BindEnv`/`pflag` alias) so that `DEBUG=true` continues to enable debug logging as documented. Alternatively, update the documentation and deployment specs to use `LOG_LEVEL=debug` if that is the intended migration path in go-libs v5.7.0.

NumaryBot, line 73: 🟠 **[major] DEBUG env variable no longer binds debug flag after go-libs v5.7.0 upgrade**

With the go-libs v5.7.0 upgrade, `service.BindEnvToFlagSet` now explicitly skips binding the `debug` flag from the `DEBUG` environment variable. Ledger relies on `service.Execute(rootCmd)` for env-to-flag binding and documents `DEBUG=true` as the way to enable debug logs. Operators and deployment specs that set `DEBUG=true` (e.g. via `spec.Debug`) will now leave `cfg.Debug` as false unless `--debug` is passed explicitly on the command line, silently breaking existing deployments.

**Suggestion:** Add a Ledger-side env alias for the debug flag (e.g. via a manual `viper.BindEnv` or `pflag` alias) so that `DEBUG=true` continues to enable debug logging as documented. Alternatively, update documentation and deployment specs to use `LOG_LEVEL=debug` if that is the intended migration path in go-libs v5.7.0.

NumaryBot, line 73: 🟠 **[major] Silent failure on invalid environment flag values may cause insecure defaults**

In go-libs v5.7.0, `service.BindEnvToFlagSet` silently swallows errors from `set.Set` when an environment variable cannot be parsed into its target flag type (e.g. `AUTH_ENABLED=yes` or `TLS_REQUIRE_CLIENT_CERT=maybe`). Since Ledger starts via `service.Execute(rootCmd)`, a misconfigured environment variable like `AUTH_ENABLED=yes` would silently fall back to the flag's default value (`false`), potentially starting the server without authentication rather than refusing to start with a bad configuration.

**Suggestion:** Add Ledger-side validation of critical environment variables before invoking `service.Execute`, or wrap the startup logic to check that security-sensitive flags (auth, TLS, etc.) are set to expected values after env binding. Consider filing an issue upstream to restore fail-fast behaviour for env parse errors.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1525#issuecomment-4904521279

gfyrag (approved):

PR #1529, opened 2026-07-07
feat(numscript): immutable append-only versioned library (EN-1288)

Azorlogh:
## Summary

Rework the numscript library into an **immutable, append-only versioned store**, per the [EN-1288](https://formance-team.atlassian.net/browse/EN-1288) product-model decision, and bring both of its persisted projections under checker verification (invariant #8).

This supersedes the earlier alpha model (mutable `latest` slot + per-numscript soft delete/restore + derived tombstone status). Per-numscript deletion, restore, and tombstones are **removed entirely** — there is no compatibility or migration layer.

## Model

- **Every version is an explicit full semver** (`major.minor.patch`). Saving with an empty, `"latest"`, or partial version is rejected with `NUMSCRIPT_INVALID_VERSION`.
- **Content entries are immutable.** Re-saving an existing `(ledger, name, version)` returns `NUMSCRIPT_VERSION_ALREADY_EXISTS`. Changed content or a rollback is published under a new version.
- **Each name has a `latest` pointer equal to the greatest stored semver.** A save advances it to `max(current, saved)`; versions may be saved out of order.
- **No delete, no restore, no tombstones.** `DeleteLedger` still cascades and removes all numscript data scoped to the ledger.

## What's included

1. **`SaveNumscript`** — the only write; explicit-semver, append-only (`processSaveNumscript`).
2. **`GetNumscript`** — version selector: `""`/`"latest"` → greatest semver; exact semver → that version; partial (`"1"`, `"1.0"`) → highest match (read-only).
3. **`ListNumscripts`** — the greatest version of each named script.
4. **`ListNumscriptVersions`** — the `latest_version` pointer plus every stored version (highest semver first).
5. **Reference resolution** keeps the `"latest"`/exact selector in the audited order; admission *plans* the reads (declares the `SubAttrNumscriptVersion` pointer + the discovered greatest's `SubAttrNumscriptContent` coverage) without mutating the order, and the FSM resolves `"latest"` at apply via the pointer, guarded by a coverage check that yields `ErrStaleProposal` (retryable) on cross-proposal skew.
6. **Checker `compareNumscripts` (invariant #8)** — re-derives `SubAttrNumscriptContent` (full `NumscriptInfo` compare) and `SubAttrNumscriptVersion` (greatest-semver pointer) from `SavedNumscript` / `DeleteLedger` logs; baseline-seeded under archiving so a post-archive out-of-order save cannot produce a false `NUMSCRIPT_MISMATCH`.
7. **Backup rebuild** maintains the greatest-semver pointer, seeded from the checkpoint store so an incremental delta save cannot regress it.

Full parity across **gRPC, HTTP REST, and the `ledgerctl` CLI** (`numscripts save --version`, `get`, `list`, `versions`), plus **OpenAPI, api-comparison, cli.md, the numscript-library architecture doc, the checker doc, and AGENTS.md invariant 8**.

## Design decisions

- **`latest` is a pure projection**, always the greatest stored semver — never an independent writable slot. Resolving it on the FSM apply path (which cannot read Pebble, invariant 3) is split: admission reads Pebble to plan/declare coverage without mutating the audited order, and the deterministic FSM resolves the pointer from the cache, falling back to a retryable `ErrStaleProposal` when a concurrent proposal advanced the pointer between admission and apply.
- **Save requires an explicit full semver** — no implicit "latest" write. Immutability is enforced at apply against the preloaded content key.

## Testing

- Processor, admission (no-order-mutation + greatest-wins overlay), query (version history + resolution), HTTP/gRPC handler, checker (`compareNumscripts` incl. baseline seeding + non-content-field tamper), state, and backup unit tests — **all pass**.
- e2e lifecycle tests in `tests/e2e/business/numscript_library_test.go` (save → list → versions → immutability conflict → reference resolution) — compiled under `-tags e2e`, **left for CI**.
- `just pre-commit` clean (proto/dashboards/CRDs regenerated, `go mod tidy` clean, golangci-lint 0 issues in main + operator). e2e / scenario / antithesis-workload builds compile.

[EN-1288]: https://formance-team.atlassian.net/browse/EN-1288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/application/admission/overlay.go
```diff
@@ -98,3 +98,9 @@ func (o *bulkOverlay) recordNumscriptSave(ledger, name, version, content string)
 func (o *bulkOverlay) recordNumscriptDelete(ledger, name string) {
 	o.numscriptLatest.Delete(numscriptNameKey{Ledger: ledger, Name: name})
 }
+
+// recordNumscriptRestore re-points the latest pointer to an existing version,
+// mirroring the FSM restore so later requests in the same bulk resolve it.
+func (o *bulkOverlay) recordNumscriptRestore(ledger, name, version string) {
+	o.numscriptLatest.Put(numscriptNameKey{Ledger: ledger, Name: name}, version)
```
NumaryBot: 🟠 **[major] Preserve restored content for later bulk requests**

When a bulk request restores a numscript and a later request in the same bulk references that script without an explicit version, this overlay only records the new latest pointer. `resolveNumscriptFromOverlay` then fails because the restored version's content is not in `numscriptEntries`, falls back to Pebble, and sees the pre-bulk latest pointer instead. This rejects tombstone-restore-then-use batches, or uses the old script content for live rollbacks followed by a transaction.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

internal/adapter/http/handlers_list_numscript_versions.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	versions, err := s.backend.ListNumscriptVersions(r.Context(), ledgerName, name)
+	if err != nil {
+		handleError(w, r, err)
+
+		return
+	}
+
+	writeOK(w, versions)
```
NumaryBot: 🟠 **[major] Numeric enum serialization breaks REST clients on versions endpoint**

The GET /numscripts/{name}/versions handler serializes NumscriptVersionStatus as its raw int32 value (e.g. 1, 2, 3) rather than the string constants (NUMSCRIPT_VERSION_STATUS_ACTIVE, SUPERSEDED, TOMBSTONED) that the OpenAPI contract and REST clients expect. This will silently break any client consuming the versions list.

**Suggestion:** Map the response to a dedicated REST DTO that converts the enum to its string representation, or use protobuf-JSON encoding (e.g. protojson) for this endpoint so enum fields are rendered as their named string values.

internal/application/admission/overlay.go
NumaryBot: 🟠 **[major] Restore overlay does not expose restored content to subsequent bulk requests**

When a bulk request restores a numscript and a later step in the same bulk references that script without an explicit version, the overlay only records the new latest pointer. resolveNumscriptFromOverlay then fails to find the restored content in numscriptEntries, falls back to Pebble, and reads the pre-bulk latest pointer instead. This causes tombstone-restore-then-use batches to fail or silently use stale script content.

**Suggestion:** After recording the new latest pointer in the restore path, also write the restored version's content into numscriptEntries (or the equivalent in-memory cache) so subsequent overlay lookups within the same bulk operation see the freshly restored content.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

internal/adapter/http/handlers_list_numscript_versions.go
Azorlogh: Fixed in f65419ab7. The `GET /numscripts/{name}/versions` handler now maps the response to a DTO that renders `status` via the enum's `String()` name (`NUMSCRIPT_VERSION_STATUS_ACTIVE`/`SUPERSEDED`/`TOMBSTONED`), matching the OpenAPI contract; `createdAt` reuses the proto `Timestamp` (RFC3339 via its `MarshalJSON`). Added a handler test asserting the string form and rejecting the numeric one. (`writeOK` uses sonic, not protojson, so the raw int32 leaked — the CLI `-o json` path was unaffected since it already uses protojson.)

internal/adapter/http/handlers_list_numscript_versions.go
```diff
@@ -0,0 +1,53 @@
...
+func (s *Server) handleListNumscriptVersions(w http.ResponseWriter, r *http.Request) {
+	ledgerName, ok := requireLedgerName(w, r)
+	if !ok {
+		return
+	}
+
+	name := chi.URLParam(r, "name")
+	if name == "" {
+		writeBadRequest(w, "INVALID_REQUEST", errors.New("numscript name is required"))
+
```
NumaryBot, line 37: 🟠 **[major] Numeric enum serialization breaks REST clients on versions endpoint**

The GET /numscripts/{name}/versions handler serializes NumscriptVersionStatus as its raw int32 value (e.g. 1, 2, 3) rather than the string constants (NUMSCRIPT_VERSION_STATUS_ACTIVE, SUPERSEDED, TOMBSTONED) that the OpenAPI contract and REST clients expect. This will silently break any client consuming the versions list.

**Suggestion:** Map the response to a dedicated REST DTO that converts the enum to its string representation, or use protobuf-JSON encoding (e.g. protojson) for this endpoint so enum fields are rendered as their named string values.

internal/infra/backup/rebuild.go
```diff
@@ -250,6 +250,36 @@ func RebuildDelta(
 				}
 			}
 
+		case *commonpb.LogPayload_RestoredNumscript:
+			if p.RestoredNumscript != nil && p.RestoredNumscript.GetInfo() != nil {
+				info := p.RestoredNumscript.GetInfo()
+				nsLedger := info.GetLedger()
+				entryKey := domain.NumscriptEntryKey{LedgerName: nsLedger, Name: info.GetName(), Version: info.GetVersion()}
```
NumaryBot, line 333: 🟠 **[major] Restore replay incorrectly overwrites immutable version content during backup rebuild**

When a backup is rebuilt from logs containing a restore after the NumscriptContent projection had diverged before that restore, this path writes the content copied into RestoredNumscriptLog.Info back over the immutable version entry. A restore is supposed to only move the latest pointer; the source of truth for content is the earlier SavedNumscript audit log. Replaying a restore should only update the latest pointer (NumscriptVersion), not rewrite the content, otherwise a stale or tampered projection can redefine a version during rebuild.

**Suggestion:** During rebuild, when replaying a restore log entry, only update the latest pointer (e.g. NumscriptVersion) and verify it matches the previously saved entry rather than writing the content from RestoredNumscriptLog.Info back to the version store.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

internal/infra/backup/rebuild.go
Azorlogh: Fixed in 22df42e3a. The restore replay now sets only the `NumscriptVersion` pointer and no longer writes `RestoredNumscriptLog.Info` back to the content store — matching the live FSM (`processRestoreNumscript` calls `SetNumscriptLatestVersion`, never `PutNumscript`). The immutable content stays sourced from the earlier `SavedNumscript` log replayed in sequence, so a diverged/tampered projection can no longer redefine a version during rebuild.

internal/adapter/http/handlers_restore_numscript.go
```diff
@@ -0,0 +1,64 @@
...
+	var body struct {
+		Version string `json:"version"`
+	}
+	if err := json.UnmarshalRead(r.Body, &body); err != nil {
+		writeBadRequest(w, "INVALID_REQUEST", err)
+
+		return
+	}
+
+	logs, err := s.applyUnsigned(r.Context(), "", &servicepb.Request{
```
NumaryBot: 🟡 **[minor] Idempotency-Key header not forwarded on restore endpoint**

The POST /numscripts/{name}/restore handler hard-codes an empty idempotency key when building its Apply batch. If a REST client retries the request with an Idempotency-Key header after a timeout, the admission layer treats the retry as a brand-new operation and appends another RestoredNumscriptLog instead of replaying the frozen outcome. Other mutating HTTP handlers already forward this header.

**Suggestion:** Pass r.Header.Get("Idempotency-Key") (or whatever the project-standard helper is) as the idempotency key when constructing the Apply request, consistent with other mutating handlers.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

gfyrag:
Review note for current head `678508907fd297ad2de3f867a9b4d684b090de53`.

I rechecked this after the rebase. The restore replay and preload issues I had seen on the previous head look fixed, but one invariant-8 gap remains:

**`CheckStore` still does not verify the new numscript projections.**

This PR makes numscript library state a persisted, load-bearing projection. The top-level audit/log payloads are `SavedNumscriptLog`, `DeletedNumscriptLog`, and `RestoredNumscriptLog` (`misc/proto/common.proto:326`). The FSM writes the projections through `PutNumscript`, `DeleteNumscriptLatest`, and `SetNumscriptLatestVersion` in `internal/domain/processing/processor_numscript_library.go:78`, `:109`, and `:151`.

Those projections are then used by reads and admission: latest lookup resolves through `ReadNumscriptLatestVersion` and content lookup in `internal/query/numscript.go:51`, listing uses the latest-version pointers at `internal/query/numscript.go:78`, and version status is derived from the same pointer at `internal/query/numscript.go:134`.

However, `internal/application/check/checker.go` does not replay these top-level numscript logs into an expected numscript state. The replay switch only handles create/delete ledger and apply logs around `checker.go:345`, and the final comparison pass only verifies volumes, metadata, transactions, and indexes around `checker.go:537`. There is no comparison against `SubAttrNumscriptVersion` / `SubAttrNumscriptContent`.

That leaves tampering or projection drift undetected: a store can roll back the latest pointer, resurrect a deleted script, remove a content row, or alter stored content and still pass `CheckStore` as long as the audit chain itself is intact. Since every persisted projection outside the audit chain must be verified by the checker, I think this PR needs a numscript checker pass that replays saved/deleted/restored logs from the audit/log chain and compares both the immutable content entries and latest/tombstone pointers, with tamper tests for pointer drift, missing/extra content, content mismatch, and delete/restore state.

Focused test I ran on the current head:

`GOROOT= go test ./internal/application/check ./internal/domain/processing ./internal/query ./internal/infra/backup -run 'Numscript|Check|Rebuild' -count=1`

internal/adapter/http/handlers_restore_numscript.go
Azorlogh: Fixed in 2d2d13716. The restore handler now forwards `r.Header.Get("Idempotency-Key")` to `applyUnsigned`, so a retried restore replays the frozen outcome instead of appending a second `RestoredNumscriptLog`. Extended the same fix to the sibling `save` and `delete` numscript handlers so the whole numscript mutating surface is idempotent and consistent.

internal/domain/processing/processor_numscript_library.go
```diff
@@ -99,8 +111,49 @@ func processDeleteNumscript(ledger string, order *raftcmdpb.DeleteNumscriptOrder
...
+	reader, err := s.ResolveNumscriptContent(ledger, order.GetName(), version)
+	if err != nil {
+		return nil, &domain.ErrStorageOperation{Operation: "resolving numscript content for restore", Cause: err}
+	}
+
+	if reader == nil {
+		return nil, &domain.ErrNumscriptNotFound{Name: order.GetName(), Version: version}
+	}
+
+	info := reader.Mutate()
```
NumaryBot: 🟠 **[major] Restore log built from potentially stale projection content rather than validated request identity**

When building a RestoredNumscriptLog, the code copies NumscriptInfo wholesale from the stored content projection for the requested key. If that projection is stale or corrupted, the logged Ledger/Name/Version values will not match the request's intended target. RebuildDelta then trusts those logged fields to update the latest pointer during backup replay, so a divergent projection can cause replay to repoint the wrong numscript or miss the restored one entirely, even though the live FSM correctly used the order/key identity.

**Suggestion:** Derive the logged identity (Ledger, Name, Version) directly from the order/key used to locate the entry rather than from the stored NumscriptInfo fields. Optionally validate that the projection's identity matches the request before emitting the log, and return an error if it does not.

internal/adapter/http/handlers_restore_numscript.go
```diff
@@ -0,0 +1,64 @@
...
+	var body struct {
+		Version string `json:"version"`
+	}
+	if err := json.UnmarshalRead(r.Body, &body); err != nil {
+		writeBadRequest(w, "INVALID_REQUEST", err)
+
+		return
+	}
+
+	logs, err := s.applyUnsigned(r.Context(), r.Header.Get("Idempotency-Key"), &servicepb.Request{
```
NumaryBot: 🟡 **[minor] Idempotency-Key header not forwarded on restore endpoint**

The POST /numscripts/{name}/restore handler hard-codes an empty idempotency key when building its Apply batch. If a REST client retries a request with an Idempotency-Key header after a timeout, the admission layer treats the retry as a brand-new operation and appends another RestoredNumscriptLog instead of replaying the frozen outcome. All other mutating HTTP handlers already forward this header.

**Suggestion:** Pass r.Header.Get("Idempotency-Key") (or the project-standard helper equivalent) as the idempotency key when constructing the Apply request, consistent with other mutating handlers.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

internal/domain/processing/processor_numscript_library.go
```diff
@@ -99,8 +111,61 @@ func processDeleteNumscript(ledger string, order *raftcmdpb.DeleteNumscriptOrder
...
+	reader, err := s.ResolveNumscriptContent(ledger, order.GetName(), version)
+	if err != nil {
+		return nil, &domain.ErrStorageOperation{Operation: "resolving numscript content for restore", Cause: err}
+	}
+
+	if reader == nil {
+		return nil, &domain.ErrNumscriptNotFound{Name: order.GetName(), Version: version}
+	}
+
+	stored := reader.Mutate()
```
NumaryBot: 🟠 **[major] Restore log identity sourced from potentially stale projection rather than validated request**

When building a RestoredNumscriptLog, the code copies NumscriptInfo wholesale from the stored content projection for the requested key. If that projection is stale or corrupted, the logged Ledger/Name/Version values will not match the request's intended target. RebuildDelta trusts those logged fields to update the latest pointer during backup replay, so a divergent projection can cause replay to repoint the wrong numscript or miss the restored one entirely, even though the live FSM correctly used the order/key identity.

**Suggestion:** Derive the logged identity (Ledger, Name, Version) directly from the order/key used to locate the entry rather than from the stored NumscriptInfo fields. Optionally validate that the projection's identity matches the request before emitting the log, and return an error if it does not.

internal/application/check/checker.go
```diff
@@ -815,6 +872,162 @@ func (c *Checker) compareIndexes(
...
+					continue
+				}
+			}
+
+			mismatch(fmt.Sprintf("stored numscript content %q@%q for ledger %q has no matching SavedNumscript in the audit chain", key.Name, key.Version, key.LedgerName), key.LedgerName)
+
+			continue
+		}
+
+		if entry.Value.GetContent() != exp.GetContent() {
```
NumaryBot: 🟠 **[major] CheckStore only validates numscript content field, leaving other projection fields unverified**

When CheckStore verifies SubAttrNumscriptContent, it only compares the stored content field. If the stored NumscriptInfo has the same content but a tampered ledger, name, version, createdAt, or deleted field, the checker reports clean even though reads, listing, and restore responses all use those fields from the projection. This leaves the majority of the new load-bearing numscript projection unverifiable and undermines the integrity guarantee the checker is meant to provide.

**Suggestion:** Compare the full expected NumscriptInfo struct for each canonical key, not only the source text. Derive the expected values from the replayed log sequence and assert equality against all persisted fields.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

internal/domain/processing/processor_numscript_library.go
Azorlogh: Fixed in bfbe1bfee (present on the reviewed commit). `processRestoreNumscript` no longer copies the stored `NumscriptInfo` wholesale — the logged identity (`Ledger`/`Name`/`Version`) is built from the order/key (the validated request target), and only `Content`/`CreatedAt` come from the stored entry. A diverged projection can no longer log a mismatched identity for rebuild to trust.

Azorlogh:
@gfyrag — thanks, you were right; the invariant-8 gap is now closed in 5e6feb0a0.

`CheckStore` now runs a `compareNumscripts` pass: the log walk replays `SavedNumscript` / `DeletedNumscript` / `RestoredNumscript` (plus the `DeleteLedger` cascade, mirroring `deleteLedgerData`) into an expected state — a save writes the immutable content entry and repoints latest, a delete blanks the pointer, a restore only repoints. The pass then diffs both stored projections (`SubAttrNumscriptContent` + `SubAttrNumscriptVersion`) against that expected state and emits the new `CHECK_STORE_ERROR_TYPE_NUMSCRIPT_MISMATCH` on altered/missing/extra content or pointer drift (resurrected delete, rolled-back pointer). Archive-orphan and deferred-cleanup tolerances mirror `compareIndexes`.

Tamper tests cover content mismatch, resurrected delete, pointer drift, and missing/extra content, plus clean save/delete/restore cases. Invariant #8 and the checker docs are updated. Ran your command clean: `go test ./internal/application/check ./internal/domain/processing ./internal/query ./internal/infra/backup -run 'Numscript|Check|Rebuild' -count=1`.

Follow-up: NumaryBot flagged that the content pass only compared the `Content` field — extending it to the full `NumscriptInfo` now.

internal/application/check/checker.go
Azorlogh: Fixed in 8b6f93b21. `compareNumscripts` now compares the whole `NumscriptInfo` (`EqualVT`) for each stored content entry against the audit-derived value, not just the `Content` field — so a tampered `Name`/`Version`/`Ledger`/`CreatedAt`/`Deleted` on a row with a matching canonical key is now caught. Added a non-content-field tamper test (altered `CreatedAt`).

internal/application/check/checker.go
```diff
@@ -815,6 +872,166 @@ func (c *Checker) compareIndexes(
...
+					continue
+				}
+			}
+
+			mismatch(fmt.Sprintf("stored numscript content %q@%q for ledger %q has no matching SavedNumscript in the audit chain", key.Name, key.Version, key.LedgerName), key.LedgerName)
+
+			continue
+		}
+
+		// Compare the whole NumscriptInfo, not just Content: reads, listing and
```
NumaryBot: 🟡 **[minor] CheckStore only validates numscript content field, leaving other projection fields unverified**

When CheckStore verifies SubAttrNumscriptContent, it only compares the stored content field. If the stored NumscriptInfo has the same content but a tampered ledger, name, version, createdAt, or deleted field, the checker reports clean even though reads, listing, and restore responses all use those fields from the projection. This leaves the majority of the new load-bearing numscript projection unverifiable and undermines the integrity guarantee the checker is meant to provide.

**Suggestion:** Compare the full expected NumscriptInfo struct for each canonical key, not only the source text. Derive the expected values from the replayed log sequence and assert equality against all persisted fields.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

gfyrag (approved):
Re-reviewed current head 8b6f93b21e9007c5e5dc4da88f29b38210479fcc.

No remaining actionable findings from my pass. The previous checker projection gap looks fixed: compareNumscripts now rebuilds expected Numscript content and pointer projections from the audit log, compares the full NumscriptInfo, and covers tampering/missing/extra/pointer/tombstone cases in tests. The previous backup and preload concerns also look stale/resolved.

Checks run:
- GOROOT= go test ./internal/application/check -run 'Numscript'
- GOROOT= go test ./internal/domain/processing -run 'Numscript'
- GOROOT= go test ./internal/query ./internal/application/admission ./internal/adapter/http -run 'Numscript|Restore|ListNumscript'
- GOROOT= go test ./internal/adapter/grpc -run 'Numscript|ListNumscripts'
- git diff --check release/v3.0...HEAD

gfyrag (changes requested):
Requesting changes because the version model needs an explicit product/design decision before this can merge.

The implementation is technically careful, but it currently conflates two different operations:

- restoring a deleted numscript;
- moving the default pointer of a live numscript to any previously stored version (rollback).

That second behavior is outside EN-1288's required outcome and implicitly turns `latest` into a mutable deployment channel. The current code also gives `latest` two meanings: an omitted version follows the stored pointer, while the literal `"latest"` addresses a separate mutable content slot. As a result, a semver can be reported as ACTIVE while an explicit `version=latest` resolves different content.

We need to investigate and document a simpler model first. The direction currently under consideration is:

- immutable, explicit semver versions;
- `latest` means the greatest stored semver;
- deletion state is independent from the latest semver;
- restore clears deletion state and does not select an arbitrary version;
- removal of the mutable literal `"latest"` content slot is acceptable.

Please pause model-dependent implementation work until that contract is settled. The history/listing work, explicit errors, audit/checker coverage, and backup fixes remain useful, but the restore API, pointer semantics, status model, tests, and docs will need to be realigned.

gfyrag:
## Final product-model decision

We completed the model investigation triggered by the Changes Requested review. The authoritative plan is now in [EN-1288](https://formance-team.atlassian.net/browse/EN-1288).

### Immutable library

* Numscripts are append-only within a ledger.
* Every stored `(ledger, name, version)` entry is immutable.
* Save requires an explicit full numeric semver. Changed content or rollback behavior is published under a new version.
* Any unused semver may be added; the per-name latest projection is always the greatest stored semver.
* Per-numscript deletion and restoration are removed entirely.
* DeleteLedger still removes all Numscript data scoped to the ledger.
* No compatibility or migration layer is required for the superseded alpha model.

### `latest` remains in the audited order

The earlier direction to rewrite executable references to exact versions was incorrect and is superseded.

* `NumscriptReference.version` is required and accepts an exact full semver or the literal `"latest"`.
* Admission may resolve a reference into side data for script parsing, dependency discovery and preload planning, but it must not mutate the accepted order.
* The FSM resolves `"latest"` through the existing per-name `SubAttrNumscriptVersion` attribute, redefined as the greatest stored semver.
* The FSM then loads the exact content through the coverage-gated Scope without writing the resolved version back into the order.
* `AuditItem.SerializedOrder` must retain `version: "latest"`. A generated result/log may expose the concrete version.
* Executable partial selectors (`1`, `1.2`) are removed for v3.0; read-only APIs may retain them.

No second latest attribute is needed. The existing `NumscriptVersionKey -> NumscriptVersionValue.version` projection must be clarified/renamed and verified by the checker.

### Coordination requirements

The attribute alone is not enough because dependency discovery happens before proposal:

* serialize SaveNumscript and executable latest-reference planning per `(ledger, name)` so a later admission cannot discover dependencies from an old script while the FSM observes a newer latest pointer;
* plan same-bulk requests sequentially: later saves cannot affect earlier references;
* compare persisted state with prior overlay saves, and never let an overlay lower version hide a greater persisted version.

Do not freeze the admission-selected exact version into the order or an unaudited business-resolution sidecar.

### Consequences for this PR

Please remove:

* DeleteNumscript and RestoreNumscript APIs, orders, logs, CLI/actions and tests;
* tombstones, empty latest pointers, `deleted` / `includeDeleted`;
* ACTIVE/SUPERSEDED/TOMBSTONED version statuses;
* arbitrary pointer reassignment and live rollback;
* the mutable literal `"latest"` content slot and its key encoding;
* admission's mutation of `NumscriptReference`.

Please retain and realign:

* immutable explicit-semver content;
* history listing as `latestVersion` plus ordered `{version, createdAt}` entries;
* exact dependency discovery without order mutation;
* FSM-side latest resolution using declared latest/content preloads;
* audit, checker, archive-bound replay and backup/rebuild coverage;
* explicit duplicate/invalid-version errors and focused concurrency tests.

The cross-cutting invariant is: once converted to a `raftcmdpb.Order`, the accepted business payload is immutable until audit capture. Coverage bits and proposal-level execution metadata remain allowed; state-dependent business rewrites do not.

The Changes Requested review remains in effect until the PR matches this final model.

[EN-1288]: https://formance-team.atlassian.net/browse/EN-1288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

gfyrag (approved):
Re-reviewed against the settled product-model decision (EN-1288) — the requested realignment is fully applied: restore API removed, order no longer mutated by admission (FSM resolves "latest"→greatest semver, coverage-gated with a loud fail), derived status model removed (fields reserved), tests and docs realigned to the new contract. The parts I'd flagged as fine are intact (history/listing, explicit errors, backup fixes), and my invariant-#8 gap is closed by the new compareNumscripts checker pass. Build + numscript test packages green.

Two non-blocking notes: (1) the branch is currently DIRTY — please merge release/v3.0 (merge-base conflict only, compiles clean on the head); (2) a few stale comments still describe the old delete/restore/tombstone model (e.g. raft_cmd.proto:324, checker.go:189-192/951-952/971-972/1015) — worth a scrub since the code is now correct. Approving; those don't block.

internal/infra/backup/rebuild.go
```diff
@@ -311,9 +315,18 @@ func RebuildDelta(
...
 				}
 
+				// The latest pointer is the greatest stored semver; versions may
+				// be saved out of order, so advance it only when the new version
+				// is greater.
 				versionKey := domain.NumscriptVersionKey{LedgerName: nsLedger, Name: info.GetName()}
-				versionVal := &commonpb.NumscriptVersionValue{Version: info.GetVersion()}
-				if _, err := numscriptVersion.Set(batch, versionKey.Bytes(), versionVal); err != nil {
+				keyStr := string(versionKey.Bytes())
+				greatest := info.GetVersion()
```
NumaryBot, line 361: 🔴 **[blocker] Checkpoint baseline not seeded into numscriptGreatest before delta replay**

When restoring from a checkpoint plus later export segments, `numscriptGreatest` starts empty even though the checkpoint store may already contain a greater latest pointer. If the checkpoint has `pay@2.0.0` and the delta appends `pay@1.0.0`, the replay initialises `greatest` from the lower version and overwrites `SubAttrNumscriptVersion` to `1.0.0`, whereas the live FSM would keep `2.0.0`. This can silently corrupt the latest-version projection during incremental restores.

**Suggestion:** Seed `numscriptGreatest` from the existing `NumscriptVersion` projection (read from the checkpoint store) before applying delta `SavedNumscript` logs, mirroring how other archived projections are pre-populated before replay.

misc/proto/bucket.proto
```diff
@@ -245,6 +248,8 @@ message Request {
   // atomic/signed unit. Reserved so it cannot be reintroduced per-request.
   reserved 1;
   reserved "idempotency_key";
+  reserved 28, 38;
```
NumaryBot: 🔴 **[blocker] Build breaks in tagged test/workload suites after delete_numscript removal**

After `delete_numscript` was reserved/removed, the generated `servicepb.Request_DeleteNumscript` and `DeleteNumscriptRequest` types no longer exist, but checked-in tagged suites and workloads still reference them (e.g. `tests/e2e/cluster/prepared_query_ledger_cache_rotation_test.go`, `tests/antithesis/workload/bin/cmds/main/parallel_driver_numscript_lifecycle/main.go`, and scenario tests via the removed `actions.DeleteNumscriptAction`). The `go test -tags e2e ./tests/e2e/...` and `go test -tags scenario ./tests/scenarios/...` builds fail at compile time.

**Suggestion:** Remove or rewrite all call sites that still reference `DeleteNumscriptAction`, `Request_DeleteNumscript`, and `DeleteNumscriptRequest` in the e2e, scenario, and workload packages so the tagged builds compile successfully.

internal/application/check/checker.go
```diff
@@ -460,6 +473,36 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
...
+			case *commonpb.LogPayload_SavedNumscript:
+				// A save writes an immutable content entry and advances the latest
+				// pointer to the greatest stored semver (versions may be saved out
+				// of order).
+				if info := payload.SavedNumscript.GetInfo(); info != nil {
+					ck := domain.NumscriptEntryKey{LedgerName: info.GetLedger(), Name: info.GetName(), Version: info.GetVersion()}
+					expectedNumscriptContent[ck] = info
+					numscriptContentActivity[ck] = struct{}{}
+
+					vk := domain.NumscriptVersionKey{LedgerName: info.GetLedger(), Name: info.GetName()}
```
NumaryBot, line 503: 🟠 **[major] CheckStore expected-latest map not seeded from archived baseline, causing false NUMSCRIPT_MISMATCH**

For stores with archived/checkpointed chapters, `expectedNumscriptLatest` is empty until a post-baseline `SavedNumscript` is replayed. If an archived baseline already saved `pay@2.0.0` and the live delta later saves `pay@1.0.0`, the stored pointer correctly stays at `2.0.0`, but the checker expects `1.0.0` and reports a spurious `NUMSCRIPT_MISMATCH` on a clean store. The numscript checker needs the same baseline seeding as other archived projections before comparing the max semver.

**Suggestion:** Pre-populate `expectedNumscriptLatest` from the stored `SubAttrNumscriptVersion` projection (or equivalent archived baseline) before walking delta logs, consistent with how other archived projections are handled in the checker.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

docs/ops/cli.md
```diff
@@ -3506,49 +3526,33 @@ ledgerctl numscripts get transfer --ledger myledger --version 1.0.0
...
+ledgerctl numscripts save <name> --ledger <ledger-name> --version <semver> [flags]
 ```
 
 **Flags:**
 
 | Flag | Default | Description |
 |------|---------|-------------|
 | `--file` | | Path to the numscript file (reads stdin if omitted) |
-| `--version` | | Semver version (e.g. `1.0.0`) or empty for latest |
+| `--version` | | Explicit full semver (e.g. `1.0.0`); required |
```
NumaryBot, line 3594: 🟠 **[major] ledgerctl save --version flag not required, sending empty version fails server validation**

The CLI `ledgerctl numscripts save` command leaves `--version` optional with a default empty string. Since the server now requires an explicit full semver and rejects empty versions with `NUMSCRIPT_INVALID_VERSION`, the default CLI invocation (without `--version`) will always fail. This makes the documented CLI surface broken by default.

**Suggestion:** Mark the `--version` flag as required in the CLI command definition and/or add client-side validation that rejects an empty or malformed semver before sending the request, so the command cannot silently submit an invalid request.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

cmd/ledgerctl/ledgers/config_model.go
```diff
@@ -781,24 +781,7 @@ func diffNumscripts(ledgerName string, current, desired *EditableConfig) []DiffA
...
-						DeleteNumscript: &servicepb.DeleteNumscriptRequest{
-							Ledger: ledgerName,
-							Name:   name,
-						},
-					},
-				},
-			})
-		}
-	}
+	// Numscripts are immutable and append-only — there is no removal action.
```
NumaryBot: 🟠 **[major] Config diff emits same-version update for edited numscript content, always failing server-side**

Under the new append-only/immutable numscript model, if a user exports a ledger config, edits a script body without bumping the version, and applies it, the config diff path still treats the changed content as an `update` and sends a `SaveNumscript` with the same version. The server will reject this with `NUMSCRIPT_VERSION_ALREADY_EXISTS`. There is no client-side guard, so the command appears to succeed locally but always fails at the server.

**Suggestion:** In the diff path, detect when the version is unchanged but content differs and either reject the operation with a clear error message (e.g. 'content changed but version was not bumped; use a new semver') or require the caller to provide a new version before emitting the save action.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

internal/domain/processing/processor_numscript_library.go
```diff
@@ -24,47 +27,54 @@ func processSaveNumscript(ledger string, order *raftcmdpb.SaveNumscriptOrder, ct
...
 		Ledger:    ledger,
 	}
 
+	// Write the immutable content and advance the latest pointer to the greatest
+	// stored semver. Any unused semver may be added out of order, so the pointer
+	// is max(current, new): PutNumscript sets it to the new version, then we
+	// restore a greater pre-existing pointer if one existed.
 	s.PutNumscript(ledger, info)
 
+	if current != "" {
```
NumaryBot, line 72: 🟡 **[minor] Semver strings that compare equal but differ textually (e.g. `01.0.0` vs `1.0.0`) can cause latest-projection divergence**

When a client saves a version with leading-zero components (e.g. `01.0.0`), `semver.Parse` accepts it and `semver.Compare` treats it as equal to `1.0.0`. The strict `> 0` guard therefore does not restore the previous pointer. The live latest projection then holds a different string from what the checker and rebuild logic derive (both also use the strict-greater rule and keep the first string), causing a spurious NUMSCRIPT_MISMATCH or an incorrect latest pointer after an incremental restore.

**Suggestion:** Reject any incoming version string whose canonical form (after semver.Parse) differs from the raw input string, or normalise all stored versions to their canonical form before comparison and storage. Either approach ensures the live projection, checker, and rebuild all operate on the same string key.

cmd/ledgerctl/ledgers/config_model.go
```diff
@@ -744,63 +748,50 @@ func diffPreparedQueries(ledgerName string, current, desired *EditableConfig) ([
...
+			desc += fmt.Sprintf(" (v%s)", desiredNS.Version)
+		}
+
+		actions = append(actions, DiffAction{
+			Section:     "numscript",
+			Operation:   "add",
+			Description: desc,
+			Request: &servicepb.Request{
+				Type: &servicepb.Request_SaveNumscript{
+					SaveNumscript: &servicepb.SaveNumscriptRequest{
```
NumaryBot, line 804: 🟠 **[major] Config apply does not validate numscript version before calling SaveNumscriptRequest**

When `configuration apply` adds or updates a numscript entry, it copies `desiredNS.Version` directly into `SaveNumscriptRequest` without any client-side validation. Since the server now requires a full canonical semver and rejects empty, `latest`, or partial versions (e.g. `1` or `1.0`) with `NUMSCRIPT_INVALID_VERSION`, a config file that omits the `version` field (still `omitempty`) or uses a non-canonical form will produce a valid-looking diff/dry-run plan but fail only at apply time. This is a silent broken-by-default footgun for the config-driven workflow.

**Suggestion:** Add a client-side check before constructing `SaveNumscriptRequest` that validates the desired version as a full canonical semver (three numeric components). Return a descriptive error if validation fails, consistent with the validation already present on the `numscripts save` CLI path. Also consider making the `version` field required in the config schema rather than `omitempty`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

cmd/ledgerctl/ledgers/config_model.go
```diff
@@ -744,63 +749,55 @@ func diffPreparedQueries(ledgerName string, current, desired *EditableConfig) ([
...
-						},
+		// The server requires an explicit full canonical semver; validate here
+		// so an omitted/partial/non-canonical version fails the diff with a
+		// clear message rather than a valid-looking plan that only errors at
+		// apply time.
+		if _, err := semver.Parse(desiredNS.Version); err != nil {
+			return nil, fmt.Errorf("numscript %q: %w (a full canonical semver is required, e.g. 1.0.0)", name, err)
+		}
+
+		desc := fmt.Sprintf("Save numscript %q (v%s)", name, desiredNS.Version)
```
NumaryBot, line 796: 🟠 **[major] Config diff emits non-idempotent save plan for lower-version numscripts**

When the desired config names a version lower than the current greatest stored version, the diff still emits a SaveNumscript action. Because `latest` is always the greatest semver, the plan can never converge: e.g. current `pay@2.0.0`, desired `pay@1.0.0` saves `1.0.0` once, but the next diff still sees `2.0.0` as current and re-emits the save for `1.0.0`, which then fails with NUMSCRIPT_VERSION_ALREADY_EXISTS. This makes the config-driven workflow non-idempotent and silently broken for any downgrade scenario.

**Suggestion:** Before emitting a SaveNumscript action, compare the desired version against the current greatest stored version using semver ordering. If the desired version is strictly less than the current greatest, reject the diff with a descriptive error (e.g. 'desired version X is lower than current greatest Y; publish a higher version to advance latest') rather than producing a plan that cannot converge.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

internal/application/check/checker.go
```diff
@@ -1054,6 +1107,190 @@ func (c *Checker) compareMirrorV2LogID(reader dal.PebbleReader, chainBound *chai
...
+
+		exp, ok := expectedContent[key]
+		if !ok {
+			if _, deleted := deletedInReplay[key.LedgerName]; deleted {
+				mismatch(fmt.Sprintf("stored numscript content %q@%q for ledger %q survives a replayed DeleteLedger + completed cleanup", key.Name, key.Version, key.LedgerName), key.LedgerName)
+
+				continue
+			}
+
+			if hasArchivedChapters {
```
NumaryBot: 🟠 **[major] Extra numscript projection rows not flagged when archived chapters exist**

When archived chapters are present, `foldBaselineNumscripts` pre-populates the expected state from the checkpoint baseline. A `SubAttrNumscriptContent` (or `SubAttrNumscriptVersion`) row that exists in the store but is absent from `expectedContent` after baseline seeding is silently skipped with `continue`, allowing an injected or stale extra row on any archived store to pass the checker clean as long as no post-baseline log touched the same key. The same gap exists in the latest-pointer branch, leaving the new load-bearing numscript projections unverifiable under archiving.

**Suggestion:** Remove or narrow the `continue` guard: an unmatched row after baseline seeding should still be compared against the expected state (or flagged as an extra/orphan). Mirror the treatment of other archived projection surplus rows so that injected entries in both `SubAttrNumscriptContent` and `SubAttrNumscriptVersion` are caught regardless of whether a post-baseline log exists for that key.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

cmd/ledgerctl/ledgers/config_model.go
```diff
@@ -744,63 +749,72 @@ func diffPreparedQueries(ledgerName string, current, desired *EditableConfig) ([
...
 					},
 				},
-			})
-		}
+			},
+		})
 	}
 
-	return actions
+	// Numscripts are immutable and append-only — there is no removal action.
```
NumaryBot: 🟡 **[minor] Removed numscripts silently ignored in config diff**

Under the new append-only immutable model, if a user removes an existing numscript from an exported ledger config and applies it, the diff path emits no action and no error. The `configuration apply` command will report 'already up to date' even though the store still contains the script. Since deletion is no longer supported, the desired state is unreachable, but the user receives no feedback about the divergence.

**Suggestion:** Surface a clear error (or at minimum a warning) in the config diff output when an existing numscript is absent from the desired config, explaining that the append-only model does not support removal and that the desired state cannot be reached.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

internal/application/check/checker.go
```diff
@@ -3160,6 +3377,85 @@ func (c *Checker) foldBaselineReferences(
...
+		if entry.Value == nil {
+			continue
+		}
+
+		var key domain.NumscriptVersionKey
+		if err := key.Unmarshal(entry.CanonicalKey); err != nil {
+			continue
+		}
+
+		expectedLatest[key] = entry.Value.GetVersion()
```
NumaryBot: 🟡 **[minor] Baseline latest pointer not cross-validated against archived content rows in CheckStore**

For archived/checkpointed stores, `expectedNumscriptLatest` is seeded from the stored `SubAttrNumscriptVersion` projection (resolved in Thread 14). However, if the baseline checkpoint already contains multiple content rows (e.g. `pay@1.0.0` and `pay@2.0.0`) but the latest pointer drifted to `1.0.0`, the checker seeds `expectedNumscriptLatest` with the drifted value and then compares forward — reporting the store as clean even though the baseline pointer itself is incorrect relative to the archived content. The baseline pointer should either be computed from the max semver found among archived content rows, or the seeded pointer should be validated against that max before delta replay begins.

**Suggestion:** After seeding `expectedNumscriptLatest` from `SubAttrNumscriptVersion`, scan the archived `SubAttrNumscriptContent` rows for each name and assert that the seeded pointer equals the greatest stored semver. Alternatively, compute `expectedNumscriptLatest` directly from the content rows rather than trusting the stored pointer, mirroring how the live FSM always advances to the greatest semver.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1529#issuecomment-4905838061

PR #1532, opened 2026-07-08
feat(audit): attribute every audit entry; tag system actors, fail-fast on dropped forwarded caller

Azorlogh:
## Problem

A `nil` `CallerSnapshot` on an audit entry meant two different things — a background/system action, or a user write whose identity was lost — so operators couldn't distinguish "system did this" from "we dropped the caller." Three concrete gaps: system proposals were caller-less by design, a follower-forwarded caller could be silently dropped on a cluster-secret misconfig, and Ed25519 tokens without `sub` produced a blank subject.

## Changes

- **Named system actor** — add `system_component` to the `CallerIdentity.source` oneof; stamp all 11 system-proposal sites (`commands.SystemCallerSnapshot` / `auth.WithSystemActor`). `nil` now unambiguously means lost user identity.
- **Hash chain** — bind the new source tag (`0x03`) into the audit envelope. Additive; existing entries re-verify byte-identically, no `HashVersion` bump.
- **Fail-fast on dropped caller** — a forwarded snapshot on a non-cluster-internal connection is rejected with `PermissionDenied` instead of silently dropped; a cluster-secret misconfig now fails the write rather than committing an unattributed entry.
- **Observability** — admission emits `admission.audit.missing_caller` and `admission.audit.caller_subject_empty` (+ logs) when auth is enabled; system actions are exempt.
- **Consumers/docs** — `ledgerctl audit list` falls back to source/keyId/system_component when subject is empty; auth + audit-chain docs updated.

Note: `IndexReady` and metadata-conversion proposals no longer exist, so no stamping was needed there.

## Acceptance criteria

- [x] Every authenticated user write carries the caller (subject, else source/keyId), across direct-to-leader and follower-forwarded paths
- [x] System-generated entries are tagged with their subsystem, not blank
- [x] A committed authenticated write with a missing caller emits a log + metric
- [x] Tests: direct leader write, follower-forwarded, cluster-secret misconfig, Ed25519 without `sub`, system actor

## Verification

`go build ./...`, `go vet`, `gofmt`, and `golangci-lint` clean on all touched packages. New/existing unit tests pass (auth resolution, envelope golden + system-caller hash distinctness, gRPC hard-reject, admission).

⚠️ One behavior change to call out for review: with **auth enabled but the cluster secret unset/mismatched**, follower-forwarded writes now hard-fail (`PermissionDenied`) instead of committing caller-less. This is the intended fail-loud, but flagging it since it changes a (misconfigured) runtime path.

### DIFF
--- a/internal/application/admission/admission.go
+++ b/internal/application/admission/admission.go
@@ -550,16 +550,26 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
 
 	fsmWaitStart := time.Now()
 	result, err := fsmFuture.Wait(ctx)
+
+	// Observe caller attribution for committed entries only. A success, or a
+	// business-rule rejection (insufficient funds, scope/transient validation),
+	// writes an audit entry carrying this caller snapshot; a stale proposal and
+	// the pre-commit rejections (queue-full, guard, marshal) write nothing.
+	committed := err == nil
+	if !committed {
+		var businessErr *domain.BusinessError
+		committed = errors.As(err, &businessErr) && !errors.Is(err, domain.ErrStaleProposal)
+	}
+
+	if committed {
+		a.observeCallerSnapshot(ctx, cmd.GetCallerSnapshot())
+	}
+
 	if err != nil {
 		return nil, err
 	}
 
 	a.fsmFutureWaitHistogram.Record(ctx, time.Since(fsmWaitStart).Microseconds())
-
-	// The proposal is committed and applied here — an audit entry carrying this
-	// caller snapshot has been written. Observing attribution gaps at this point
-	// excludes proposals rejected before commit (guard/marshal/stale/queue-full).
-	a.observeCallerSnapshot(ctx, cmd.GetCallerSnapshot())
 
 	// Resolve CreatedLogOrReference entries into concrete logs.
 	// Created logs are returned directly; reference sequences (idempotent responses)
--- a/internal/application/admission/admission.go
+++ b/internal/application/admission/admission.go
@@ -62,6 +62,7 @@ type Admission struct {
 	attrs              *attributes.Attributes
 	numscriptCache     *numscript.NumscriptCache
 	coldStorageEnabled bool
+	authEnabled        bool
 	waitLeaderReady    func(context.Context) error
 
 	// Metrics (noop when metricsEnabled is false)
@@ -79,6 +80,8 @@ type Admission struct {
 	preloadCounter                 metric.Int64Counter
 	preloadKeysNeededCounter       metric.Int64Counter
 	preloadCacheHitsCounter        metric.Int64Counter
+	missingCallerCounter           metric.Int64Counter
+	callerSubjectEmptyCounter      metric.Int64Counter
 }
 
 // NewAdmission creates a new Admission handler.
@@ -101,6 +104,15 @@ func WithReceiptSigner(signer *receipt.Signer) func(*Admission) {
 func WithColdStorageEnabled() func(*Admission) {
 	return func(a *Admission) {
 		a.coldStorageEnabled = true
+	}
+}
+
+// WithAuthEnabled marks authentication as enabled, so the admission path can
+// flag user writes committed without an attributable caller (see
+// observeCallerSnapshot).
+func WithAuthEnabled() func(*Admission) {
+	return func(a *Admission) {
+		a.authEnabled = true
 	}
 }
 
@@ -269,6 +281,24 @@ func NewAdmission(
 		panic(err)
 	}
 
+	missingCallerCounter, err := meter.Int64Counter(
+		"admission.audit.missing_caller",
+		metric.WithDescription("Committed user writes with no caller snapshot while auth is enabled"),
+		metric.WithUnit("1"),
+	)
+	if err != nil {
+		panic(err)
+	}
+
+	callerSubjectEmptyCounter, err := meter.Int64Counter(
+		"admission.audit.caller_subject_empty",
+		metric.WithDescription("Committed user writes whose caller has a source but an empty subject (e.g. Ed25519 token without sub)"),
+		metric.WithUnit("1"),
+	)
+	if err != nil {
+		panic(err)
+	}
+
 	a.commandDurationHistogram = commandDurationHistogram
 	a.commandSizeHistogram = commandSizeHistogram
 	a.proposeQueueLoadHistogram = proposeQueueLoadHistogram
@@ -281,8 +311,38 @@ func NewAdmission(
 	a.preloadCounter = preloadCounter
 	a.preloadKeysNeededCounter = preloadKeysNeededCounter
 	a.preloadCacheHitsCounter = preloadCacheHitsCounter
+	a.missingCallerCounter = missingCallerCounter
+	a.callerSubjectEmptyCounter = callerSubjectEmptyCounter
 
 	return a
+}
+
+// observeCallerSnapshot flags audit-attribution gaps on the write path. With
+// auth enabled, a committed user write should always carry a caller: a nil
+// snapshot means an authenticated identity was lost or an anonymous write
+// slipped through, and a user source (issuer/key_id) with an empty subject is
+// the Ed25519-token-without-`sub` case where only the key id identifies the
+// caller. System actions carry a system_component source and are exempt.
+func (a *Admission) observeCallerSnapshot(ctx context.Context, snap *commonpb.CallerSnapshot) {
+	if !a.authEnabled {
+		return
+	}
+
+	if snap == nil {
+		a.logger.Errorf("committed write has no caller snapshot while auth is enabled: audit entry will be unattributed")
+		a.missingCallerCounter.Add(ctx, 1)
+
+		return
+	}
+
+	id := snap.GetIdentity()
+	switch id.GetSource().(type) {
+	case *commonpb.CallerIdentity_KeyId, *commonpb.CallerIdentity_Issuer:
+		if id.GetSubject() == "" {
+			a.logger.Infof("committed write caller has a source but an empty subject: audit entry identifies the caller only by source")
+			a.callerSubjectEmptyCounter.Add(ctx, 1)
+		}
+	}
 }
 
 // Admit implements the ctrl.Admission interface.
@@ -405,6 +465,7 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
 
 	cmd.ExecutionPlan = build.ExecutionPlan
 	cmd.CallerSnapshot = auth.ResolveCallerSnapshot(ctx)
+	a.observeCallerSnapshot(ctx, cmd.GetCallerSnapshot())
 
 	preloadSpan.End()
 
--- a/internal/application/admission/admission.go
+++ b/internal/application/admission/admission.go
@@ -551,17 +551,11 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
 	fsmWaitStart := time.Now()
 	result, err := fsmFuture.Wait(ctx)
 
-	// Observe caller attribution for committed entries only. A success, or a
-	// business-rule rejection (insufficient funds, scope/transient validation),
-	// writes an audit entry carrying this caller snapshot; a stale proposal and
-	// the pre-commit rejections (queue-full, guard, marshal) write nothing.
-	committed := err == nil
-	if !committed {
-		var businessErr *domain.BusinessError
-		committed = errors.As(err, &businessErr) && !errors.Is(err, domain.ErrStaleProposal)
-	}
-
-	if committed {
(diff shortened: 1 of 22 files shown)


internal/infra/state/audit_envelope.go
```diff
@@ -215,6 +216,9 @@ func buildCallerSnapshotPayload(snap *commonpb.CallerSnapshot) []byte {
 	case *commonpb.CallerIdentity_KeyId:
 		buf = appendU8(buf, callerSourceKeyID)
 		buf = appendLenString(buf, src.KeyId)
+	case *commonpb.CallerIdentity_SystemComponent:
```
NumaryBot, line 219: 🔴 **[blocker] Gate system caller hashing during rolling upgrades**

When a new leader stamps `system_component` during a rolling upgrade, older followers do not have this `CallerIdentity_SystemComponent` case: they preserve field 6 as unknown data but hash the caller as `callerSourceNone`, while upgraded nodes hash it as `callerSourceSystem`. That makes the same Raft proposal produce different audit hashes across nodes, and entries written by the old code path will later fail verification after upgrade. This needs a compatibility gate/version bump or another rollout-safe encoding before system components are emitted into the hash chain.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1532#issuecomment-4913877620

internal/infra/state/audit_envelope.go
```diff
@@ -215,6 +216,9 @@ func buildCallerSnapshotPayload(snap *commonpb.CallerSnapshot) []byte {
 	case *commonpb.CallerIdentity_KeyId:
```
NumaryBot, line 216: 🔴 **[blocker] Mixed-version deployment causes audit hash-chain divergence**

When a system-attributed proposal is committed during a rolling deployment, replicas still running the old binary do not recognize the new oneof field 6 and will hash the caller as `callerSourceNone`, while upgraded replicas use the new `callerSourceSystem` branch. This changes the audit hash preimage without any version gate or rollout guard, so replicas in a mixed-version cluster will persist different audit hashes for the same Raft entry, breaking the audit chain integrity.

**Suggestion:** Introduce a feature flag or Raft-level cluster-version gate that prevents the new caller-source field from being written into the hash preimage until all replicas have upgraded. Alternatively, document and enforce a hard-cutover (non-rolling) deployment requirement for this change, and add an assertion or startup check that aborts if the persisted chain was written by an older binary.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1532#issuecomment-4913877620

gfyrag (approved):
Reviewed against release/v3.0. The implementation looks coherent for EN-1306: system actors are attributed, forwarded snapshots fail loud on non-cluster-internal requests, and the new system_component source is hash-bound and surfaced in CLI/docs.\n\nOne non-blocking note: admission.observeCallerSnapshot currently emits the 'committed write' metric/log before Raft acceptance/FSM apply, so it may count proposals that never commit. I would consider moving that observation after acceptance/apply, but I don't see it as merge-blocking for this PR.

internal/infra/state/audit_envelope.go
NumaryBot, line 216: 🔴 **[blocker] New `callerSource` field in audit hash preimage breaks chain integrity in mixed-version deployments**

The new `callerSourceSystem` branch is added to the audit hash preimage computation without any cluster-version gate. During a rolling upgrade, old replicas do not recognise the new oneof field 6 and will hash the caller as `callerSourceNone`, while upgraded replicas hash it as `callerSourceSystem`. This means the same Raft proposal produces different audit hashes on different nodes, silently breaking audit chain integrity — and previously persisted entries will fail verification after a full upgrade completes.

**Suggestion:** Introduce a Raft-level cluster-version gate (or feature flag) that prevents the new caller-source field from being written into the hash preimage until all replicas have upgraded. Alternatively, enforce a hard-cutover (non-rolling) deployment strategy with a startup assertion that aborts if the persisted chain was written by an older binary.

internal/application/admission/admission.go
```diff
@@ -496,6 +556,11 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
 
 	a.fsmFutureWaitHistogram.Record(ctx, time.Since(fsmWaitStart).Microseconds())
 
+	// The proposal is committed and applied here — an audit entry carrying this
+	// caller snapshot has been written. Observing attribution gaps at this point
+	// excludes proposals rejected before commit (guard/marshal/stale/queue-full).
+	a.observeCallerSnapshot(ctx, cmd.GetCallerSnapshot())
```
NumaryBot: 🟠 **[major] Caller-attribution observation skipped for committed failure audit entries**

With auth enabled, the caller-snapshot observation (emitting `admission.audit.missing_caller` / `caller_subject_empty` signals) is only reached after `fsmFuture.Wait` succeeds. FSM business-logic rejections (e.g. insufficient funds, transient-volume validation) still write a failure `AuditEntry` and resolve the future with `ApplyResult.Error` before returning. In those committed-failure scenarios, the function returns early and the new attribution-gap metrics are never emitted, creating a blind spot in audit observability.

**Suggestion:** Move (or duplicate) the caller-snapshot observation to run before the early-return on `fsmFuture.Wait` error, or extract it into a deferred closure so it executes regardless of whether the FSM apply succeeded or returned a business error.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1532#issuecomment-4913877620

internal/infra/state/audit_envelope.go
NumaryBot, line 216: 🔴 **[blocker] New `callerSource` field in audit hash preimage breaks chain integrity in mixed-version deployments**

The new `callerSourceSystem` branch is added to the audit hash preimage computation without any cluster-version gate. During a rolling upgrade, old replicas do not recognise the new oneof field 6 and hash the caller as `callerSourceNone`, while upgraded replicas hash it as `callerSourceSystem`. This means the same Raft proposal produces different audit hashes on different nodes, silently breaking audit chain integrity. Previously persisted entries will also fail verification after a full upgrade completes.

**Suggestion:** Introduce a Raft-level cluster-version gate (or feature flag) that prevents the new caller-source field from being written into the hash preimage until all replicas have upgraded. Alternatively, enforce a hard-cutover (non-rolling) deployment strategy with a startup assertion that aborts if the persisted chain was written by an older binary.

internal/application/admission/admission.go
```diff
@@ -490,6 +550,21 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
...
+
+	// Observe caller attribution for committed entries only. A success, or a
+	// business-rule rejection (insufficient funds, scope/transient validation),
+	// writes an audit entry carrying this caller snapshot; a stale proposal and
+	// the pre-commit rejections (queue-full, guard, marshal) write nothing.
+	committed := err == nil
+	if !committed {
+		var businessErr *domain.BusinessError
+		committed = errors.As(err, &businessErr) && !errors.Is(err, domain.ErrStaleProposal)
+	}
```
NumaryBot: 🟠 **[major] Caller-attribution metrics skipped for committed failure audit entries**

With auth enabled, the caller-snapshot observation (emitting `admission.audit.missing_caller` / `caller_subject_empty` signals) is only reached after `fsmFuture.Wait` succeeds. FSM business-logic rejections (e.g. insufficient funds, transient-volume validation) still write a failure `AuditEntry` and resolve the future with `ApplyResult.Error` before returning. In those committed-failure scenarios, the function returns early and the attribution-gap metrics are never emitted, creating a blind spot in audit observability.

**Suggestion:** Move (or duplicate) the caller-snapshot observation to run before the early-return on `fsmFuture.Wait` error, or extract it into a deferred closure so it executes regardless of whether the FSM apply succeeded or returned a business error.

internal/application/admission/admission.go
```diff
@@ -490,6 +550,21 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
 
 	fsmWaitStart := time.Now()
 	result, err := fsmFuture.Wait(ctx)
+
+	// Observe caller attribution for committed entries only. A success, or a
+	// business-rule rejection (insufficient funds, scope/transient validation),
+	// writes an audit entry carrying this caller snapshot; a stale proposal and
+	// the pre-commit rejections (queue-full, guard, marshal) write nothing.
+	committed := err == nil
```
NumaryBot: 🟠 **[major] Caller-attribution metrics incorrectly emitted for idempotency replays with no audit entry written**

When an idempotency key replays a previously recorded success or failure, the FSM skips writing a new audit entry but the `committed` block still calls `observeCallerSnapshot`. This means callers with an empty `sub` or a missing caller under anonymous-write misconfiguration will incorrectly increment the new audit-attribution metrics on every retry, even though no audit entry carrying that snapshot was actually created.

**Suggestion:** Guard the `observeCallerSnapshot` call so it only executes when the FSM confirmed that a new audit entry was actually written (e.g. check an idempotency-replay flag returned by the FSM result before calling the observation function).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1532#issuecomment-4913877620

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1532#issuecomment-4913877620

internal/infra/state/audit_envelope.go
NumaryBot, line 216: 🔴 **[blocker] New `callerSource` field in audit hash preimage breaks chain integrity in mixed-version deployments**

The new `callerSourceSystem` branch is added to the audit hash preimage computation without any cluster-version gate. During a rolling upgrade, old replicas do not recognise the new oneof field 6 and hash the caller as `callerSourceNone`, while upgraded replicas hash it as `callerSourceSystem`. This means the same Raft proposal produces different audit hashes on different nodes, silently breaking audit chain integrity. Previously persisted entries will also fail verification after a full upgrade completes.

**Suggestion:** Introduce a Raft-level cluster-version gate (or feature flag) that prevents the new caller-source field from being written into the hash preimage until all replicas have upgraded. Alternatively, enforce a hard-cutover (non-rolling) deployment strategy with a startup assertion that aborts if the persisted chain was written by an older binary.

internal/application/admission/admission.go
```diff
@@ -490,6 +550,15 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
...
+
+	// Observe caller attribution only when the FSM actually wrote an audit
+	// entry for this proposal — a success or a committed business-rule failure.
+	// Replays, stale proposals, and pre-commit rejections write no entry, so the
+	// caller snapshot they carry was never recorded and must not be counted.
+	if result.AuditEntryWritten {
+		a.observeCallerSnapshot(ctx, cmd.GetCallerSnapshot())
+	}
+
 	if err != nil {
```
NumaryBot, line 562: 🟠 **[major] Caller-attribution metrics skipped for committed failure audit entries**

With auth enabled, the caller-snapshot observation (emitting `admission.audit.missing_caller` / `caller_subject_empty` signals) is only reached after `fsmFuture.Wait` succeeds. FSM business-logic rejections (e.g. insufficient funds, transient-volume validation) still write a failure `AuditEntry` and resolve the future with `ApplyResult.Error` before returning early. In those committed-failure scenarios, the attribution-gap metrics are never emitted, creating a blind spot in audit observability.

**Suggestion:** Move (or duplicate) the caller-snapshot observation to run before the early-return on `fsmFuture.Wait` error, or extract it into a deferred closure so it executes regardless of whether the FSM apply succeeded or returned a business error.

internal/application/admission/admission.go
```diff
@@ -490,6 +550,15 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
 
 	fsmWaitStart := time.Now()
 	result, err := fsmFuture.Wait(ctx)
+
+	// Observe caller attribution only when the FSM actually wrote an audit
+	// entry for this proposal — a success or a committed business-rule failure.
+	// Replays, stale proposals, and pre-commit rejections write no entry, so the
+	// caller snapshot they carry was never recorded and must not be counted.
+	if result.AuditEntryWritten {
```
NumaryBot, line 558: 🟠 **[major] Caller-attribution metrics incorrectly emitted for idempotency replays with no audit entry written**

When an idempotency key replays a previously recorded success or failure, the FSM skips writing a new audit entry but the `committed` block still calls `observeCallerSnapshot`. This means callers with an empty `sub` or a missing caller under anonymous-write misconfiguration will incorrectly increment the new audit-attribution metrics on every retry, even though no audit entry carrying that snapshot was actually created.

**Suggestion:** Guard the `observeCallerSnapshot` call so it only executes when the FSM confirmed that a new audit entry was actually written (e.g. check an idempotency-replay flag returned by the FSM result before calling the observation function).

PR #1533, opened 2026-07-08
perf(applier): prefetch DecodeEntries in a dedicated goroutine

gfyrag:
## Summary

Moves `state.DecodeEntries` out of the applier's main-loop critical path
into a dedicated prefetching goroutine, so `UnmarshalVT` for batch N+1
runs on a spare core while the applier holds `fsm.mu` running Prepare(N).

## Structure

```
Submit / Barrier / SyncFromLeader / ... ─▶ a.ch
                                             │
                                             ▼
                                      [decoder goroutine]
                                       - reads a.ch
                                       - state.DecodeEntries for entries
                                       - forwards everything else as-is
                                             │
                                             ▼
                                          a.decodedCh
                                             │
                                             ▼
                                        [main loop]
                                     - Prepare / async commit
                                     - Barrier / status transitions
```

Non-entries work items (barrier, syncLeader, installSnapshot) pass through
the decoder verbatim with a nil decoded slice — ordering vs. entries is
preserved by the single-goroutine decoder. Non-normal-status paths (spool,
snapshot install) release the pre-decoded proposals immediately since raw
entries are what the spool consumes.

Shutdown drains `a.decodedCh` so vtproto's pool doesn't leak. The
decoder's context is derived locally from `Run`'s ctx via
`context.WithCancel`, so the decoder unwinds cleanly even on normal exit
via `<-stop` (where the outer ctx stays live).

`applyEntriesToFSM` (the raw-entries wrapper) is dropped as dead code:
the hot path calls `applyDecodedEntriesToFSM` directly.

## Bench results (perf-world-to-bank, 5min saturated window)

| Metric | Baseline | This PR | Delta |
|---|---|---|---|
| `raft_fsm_prepare_duration_microseconds` p95 | ~18-20K μs | ~14-16K μs | **-20-25%** |
| `raft_apply_entries_batch_size` p50 | ~15-16 | ~12 | -20% (smaller batches) |
| batches/sec (derived) | ~107 | ~133 | **+24%** |
| `raft_apply_entries_batch_size_total` rate (throughput) | ~1.6K entries/s | ~1.6K entries/s | ≈0% |

The pipeline does what it says on the tin: Prepare wall-time drops
because DecodeEntries is no longer serialized ahead of it. The applier
processes batches faster (+24% batches/sec) but each batch is smaller
because the queue drains before it accumulates as much.

Throughput is flat because the **actual bottleneck is downstream**:
`raft_applier_commit_wait_duration_microseconds` p95 climbs from
~3-6K μs (dispersed) to ~5-6K μs (sustained). Multiplied by the new
batch rate: ~730 ms/sec of applier time waiting on Pebble commit.
Pebble is the next lever, not decode.

## Why merge anyway

- Pure structural gain, no correctness risk (single-goroutine decoder
  preserves ordering; ownership handoff is explicit).
- Frees the applier core from CPU cycles previously spent on decode
  (~2.4s out of every 27.6s on the profile) — usable by other
  goroutines on the process.
- Unlocks the ability to see the real bottleneck (Pebble commit),
  which was previously masked by decode-inside-apply.
- Prerequisite for future prepare-side optimisations to actually
  translate into throughput.

## Test plan

- [x] `just pre-commit` (build, tests, lint)
- [x] `internal/infra/node/` full test suite passes
- [x] Manual bench on `perf-world-to-bank` shows expected shift
      (Prepare duration down, batches/sec up, throughput bounded by
       Pebble commit as expected)
- [ ] Antithesis run to validate determinism under scheduling shuffle

### DIFF
--- a/internal/infra/node/applier.go
+++ b/internal/infra/node/applier.go
@@ -42,6 +42,22 @@ type applyWork struct {
 	responses []raftpb.Message
 }
 
+// decodedApplyWork is the post-prefetch shape consumed by Run's main loop.
+// The decoder goroutine (see Run) reads applyWork from a.ch, invokes
+// state.DecodeEntries for work items carrying entries, and forwards the
+// result here so the Prepare stage never blocks on the UnmarshalVT
+// pass. Non-entries work items (barrier, syncLeader, installSnapshot) are
+// forwarded with `decoded` nil and `err` nil — the decoder is a strict
+// pass-through for them, preserving ordering vs entries.
+//
+// If DecodeEntries fails, `err` carries the wrapped error and `decoded` is
+// nil. The main loop propagates the error immediately without applying.
+type decodedApplyWork struct {
+	work    applyWork
+	decoded []state.DecodedEntry
+	err     error
+}
+
 // gatingResult carries the outcome of a maintenance task back to the Run
 // goroutine so it can decide whether to unspool or mark the node out-of-sync.
 type gatingResult struct {
@@ -70,8 +86,9 @@ type Applier struct {
 	gatingReason     atomic.Int32 // gatingReason* constants — for observability
 	syncProgress     atomic.Pointer[state.SyncProgress]
 	gatingTerminated chan gatingResult
-	ch               chan applyWork  // buffered(1)
-	commitCh         chan commitWork // buffered(1), read by the committer goroutine
+	ch               chan applyWork        // buffered(1); producer-facing input
+	decodedCh        chan decodedApplyWork // buffered(1); read by Run, populated by the decoder goroutine
+	commitCh         chan commitWork       // buffered(1), read by the committer goroutine
 
 	// pending holds the in-flight commit result channel, if any.
 	// At most one at a time. Drained before barriers, checkpoints, and shutdown.
@@ -189,6 +206,7 @@ func NewApplier(
 		responseSink:            responseSink,
 		status:                  &initialStatus,
 		ch:                      make(chan applyWork, 1),
+		decodedCh:               make(chan decodedApplyWork, 1),
 		commitCh:                make(chan commitWork, 1),
 	}
 
@@ -690,6 +708,35 @@ func (a *Applier) Status() int32 {
 
 // Run is the Applier goroutine. It processes submitted work items and
 // handles gating termination (unspool after maintenance tasks).
+//
+// Concurrency layout inside Run:
+//
+//	Submit / Barrier / SyncFromLeader / ... ─▶ a.ch
+//	                                             │
+//	                                             ▼
+//	                                      [decoder goroutine]
+//	                                       - reads a.ch
+//	                                       - state.DecodeEntries for entries
+//	                                       - forwards everything else as-is
+//	                                             │
+//	                                             ▼
+//	                                          a.decodedCh
+//	                                             │
+//	                                             ▼
+//	                                    [main loop below]
+//	                                     - Prepare / async commit
+//	                                     - Barrier / status transitions
+//	                                             │
+//	                                             ▼
+//	                                          a.commitCh
+//	                                             │
+//	                                             ▼
+//	                                      [committer goroutine]
+//	                                       - CommitPreparedBatch
+//
+// The decoder stage is what lets Prepare(N) overlap with Decode(N+1):
+// UnmarshalVT of the next batch runs on a spare core while the applier
+// holds fsm.mu for the current batch.
 func (a *Applier) Run(ctx context.Context, stop chan struct{}) error {
 	// Start the dedicated committer goroutine. It exits when commitCh is closed.
 	committerDone := make(chan struct{})
@@ -697,6 +744,19 @@ func (a *Applier) Run(ctx context.Context, stop chan struct{}) error {
 	go func() {
 		a.runCommitter(ctx, stop)
 		close(committerDone)
+	}()
+
+	// Start the decoder goroutine. It exits when its context is cancelled
+	// (below on Run's exit) OR when a.ch is closed. We use a local cancel
+	// tied to Run's lifetime so the decoder unwinds even on normal exit
+	// where the outer ctx is still live (e.g. tests that stop the applier
+	// via a stop channel without cancelling ctx).
+	decoderCtx, cancelDecoder := context.WithCancel(ctx)
+	decoderDone := make(chan struct{})
+
+	go func() {
+		defer close(decoderDone)
+		a.runDecoder(decoderCtx)
 	}()
 
 	defer func() {
@@ -707,6 +767,15 @@ func (a *Applier) Run(ctx context.Context, stop chan struct{}) error {
 
 		close(a.commitCh)
 		<-committerDone
+
+		// Signal the decoder to stop, then drain any batches it managed to
+		// hand off before cancellation. Every release returns pooled
+		// proposals to vtproto's pool.
+		cancelDecoder()
+		<-decoderDone
+		for dw := range a.decodedCh {
+			state.ReleaseDecodedEntries(dw.decoded)
+		}
 	}()
 
 	var (
@@ -717,7 +786,8 @@ func (a *Applier) Run(ctx context.Context, stop chan struct{}) error {
 
 	for {
 		select {
-		case work := <-a.ch:
+		case dw := <-a.decodedCh:
+			work := dw.work
 			a.batchWaitDurationHistogram.Record(ctx, time.Since(waitStart).Microseconds())
 
 			if work.barrier != nil {
@@ -777,18 +847,39 @@ func (a *Applier) Run(ctx context.Context, stop chan struct{}) error {
 
 			switch a.status.Load() {
 			case statusNormal:
-				// applyEntriesToFSM threads work.responses down to the last
-				// applyEntriesPipelined call, which attaches them to the
+				// Fail fast on a decode error surfaced by the prefetcher —
+				// the raft entry payload is corrupt (or the wire schema
+				// version is wrong), no point going further.
+				if dw.err != nil {
+					return dw.err
+				}
+				// applyDecodedEntriesToFSM threads work.responses down to the
+				// last applyEntriesPipelined call, which attaches them to the
 				// commitWork sent to runCommitter. runCommitter fires the
 				// responses AFTER CommitPreparedBatch succeeds, so
 				// raft.Applied advances in lockstep with FSM-applied without
 				// serializing prepare-N+1 against commit-N (the
 				// applier→committer pipeline is preserved).
-				err := a.applyEntriesToFSM(ctx, stop, work.confState, work.responses, work.entries...)
+				//
+				// The dw.decoded slice is pre-decoded by the runDecoder
+				// goroutine; applyDecodedEntriesToFSM takes ownership and
+				// releases it back to the pool via its defer.
+				err := a.applyDecodedEntriesToFSM(ctx, stop, work.confState, work.responses, dw.decoded)
 				if err != nil {
 					return err
 				}
 			default:
+				// Non-normal status: the pre-decoded proposals will not be
+				// applied by the FSM (entries go to the spool instead). Return
+				// them to the pool immediately so we don't leak on a
+				// long-running gating period.
+				//
+				// A decode error is non-fatal on this path — the raw entry
+				// bytes are still valid and get spooled as-is; a subsequent
+				// replay will re-attempt the decode with a matching wire
+				// schema.
+				state.ReleaseDecodedEntries(dw.decoded)
+
 				// Drain pending commit before switching to spool mode.
 				if err := a.waitPendingCommit(ctx); err != nil {
 					return err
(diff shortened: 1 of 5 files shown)


internal/infra/node/applier.go
```diff
@@ -717,7 +798,8 @@ func (a *Applier) Run(ctx context.Context, stop chan struct{}) error {
 
 	for {
 		select {
-		case work := <-a.ch:
+		case dw := <-a.decodedCh:
```
NumaryBot: 🟠 **[major] Closed decodedCh causes infinite spin in applier loop on cancellation**

When `decoderCtx` is cancelled (e.g. via the `cancelRun` fallback on OnStart/OnStop timeout), `runDecoder` closes `a.decodedCh`. The select in the main `Run` loop receives on this channel without checking the `ok` boolean, so a closed channel will keep winning the select with a zero-value `decodedApplyWork`. This causes the applier loop to spin indefinitely instead of exiting, defeating the shutdown/error-cleanup path and potentially leaving the node goroutine alive while its dependencies are torn down.

**Suggestion:** Check the channel-open status when receiving from `a.decodedCh`. For example:
```go
case work, ok := <-a.decodedCh:
    if !ok {
        // channel closed, decoder is done — handle shutdown
        return
    }
    // process work
```
Alternatively, set `a.decodedCh` to `nil` after it is closed so that branch is permanently disabled in the select.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1533#issuecomment-4915902041

internal/infra/node/applier.go
```diff
@@ -1084,6 +1187,67 @@ func (a *Applier) submitAsyncCommit(pb *state.PreparedBatch, pfs []pendingFuture
...
+			ok   bool
+		)
+
+		select {
+		case <-ctx.Done():
+			return
+		case work, ok = <-a.ch:
+			if !ok {
+				return
+			}
```
Azorlogh: `a.ch` is never closed anywhere, so this `!ok` branch is dead — the decoder only exits via ctx-cancel. That makes the "a.ch closed (Stop)" comments (here + at the goroutine start) inaccurate, and closing it would actually panic the producers (`Submit`/`Drain`/…).  Suggest dropping the ok-branch and fixing the two comments.

internal/infra/node/applier.go
```diff
@@ -1084,6 +1187,67 @@ func (a *Applier) submitAsyncCommit(pb *state.PreparedBatch, pfs []pendingFuture
 // races into the sink send AFTER orchestrate stopped draining would sit
 // forever on ctx.Done alone (finding 70740916). Selecting on stop too
 // unblocks the send synchronously with orchestrate's exit.
+// runDecoder is the prefetching decoder goroutine started by Run. It reads
```
Azorlogh, line 1226: `runDecoder` sits between `runCommitter`'s doc comment and `func runCommitter` (no blank line), so the doc now attaches to `runDecoder` and `runCommitter` is left undocumented.

Azorlogh (commented):
Stale refs to the renamed `applyEntriesToFSM` linger in comments: `applier.go` 160, 1341, 1378, 1580, 1592 (also `node.go` 135/1373/1380 + a couple tests). Worth a sweep to `applyDecodedEntriesToFSM`.

internal/infra/node/applier.go
```diff
@@ -715,9 +796,28 @@ func (a *Applier) Run(ctx context.Context, stop chan struct{}) error {
 		waitStart           = time.Now()
 	)
 
+	// Local handle on the decoded channel so we can stop selecting on it once
+	// the decoder has closed it. The decoder only ever closes decodedCh from
+	// its ctx.Done() path (a.ch is never closed), i.e. runCtx was cancelled —
```
NumaryBot, line 806: 🟠 **[major] Closed decodedCh causes infinite spin / zero-value processing on context cancellation**

When the decoder context is cancelled, `runDecoder` closes `a.decodedCh`. The `select` in the main `Run` loop receives from that channel without checking the `ok` boolean. A closed channel always unblocks with the zero value, so the loop spins indefinitely on shutdown rather than exiting cleanly, and each iteration processes a zero-value `decodedApplyWork`.

**Suggestion:** Use the two-value receive form:
```go
case work, ok := <-a.decodedCh:
    if !ok {
        return // decoder done, shut down
    }
    // process work
```
Alternatively, set `a.decodedCh = nil` after closure so the select branch is permanently suppressed.

internal/infra/node/applier.go
```diff
@@ -70,8 +87,9 @@ type Applier struct {
 	gatingReason     atomic.Int32 // gatingReason* constants — for observability
 	syncProgress     atomic.Pointer[state.SyncProgress]
 	gatingTerminated chan gatingResult
-	ch               chan applyWork  // buffered(1)
-	commitCh         chan commitWork // buffered(1), read by the committer goroutine
+	ch               chan applyWork        // buffered(1); producer-facing input
+	decodedCh        chan decodedApplyWork // buffered(1); read by Run, populated by the decoder goroutine
```
NumaryBot, line 91: 🟠 **[major] Decoder buffer breaks sync-request coalescing for out-of-sync followers**

When the applier is `statusOutOfSync` and Ready signals arrive back-to-back, `processReady` relies on `TrySyncSnapshot` seeing `a.ch` full to drop duplicate sync requests. The new buffered `decodedCh` allows `runDecoder` to pull the first `syncLeader` entry off `a.ch` before `Run` has changed the status, so a second `TrySyncSnapshot` call can successfully enqueue another sync request. The duplicate, when later processed by the `syncLeader` branch, causes `startMaintenanceTask` to interrupt the in-flight checkpoint fetch, potentially keeping an out-of-sync follower in a restart loop under sustained Ready signals.

**Suggestion:** Re-evaluate the coalescing invariant with the new two-stage pipeline. Options include: (1) gating `TrySyncSnapshot` on a dedicated atomic flag rather than channel fullness, (2) having the decoder goroutine check and discard duplicate `syncLeader` entries, or (3) using a single-slot 'replace' channel for sync requests so only the latest survives.

internal/infra/node/applier.go
```diff
@@ -1073,6 +1194,62 @@ func (a *Applier) submitAsyncCommit(pb *state.PreparedBatch, pfs []pendingFuture
...
+// a separate core. If Prepare(N) finishes before Decode(N+1) — unlikely on
+// the observed workload — the main loop simply waits on <-a.decodedCh.
+func (a *Applier) runDecoder(ctx context.Context) {
+	defer close(a.decodedCh)
+
+	for {
+		var work applyWork
+
+		select {
+		case <-ctx.Done():
```
NumaryBot, line 1255: 🟡 **[minor] Dead `!ok` branch on `a.ch` receive inside runDecoder**

`a.ch` is never closed by any producer (`Submit`/`Drain`/etc.) — closing it would panic the senders. Therefore the `!ok` branch in `runDecoder` is dead code, and the accompanying comments describing the 'a.ch closed (Stop)' exit path are inaccurate and misleading.

**Suggestion:** Remove the `ok`-check branch on `a.ch` and update/remove the associated misleading comments to accurately reflect that the decoder exits only via context cancellation.

internal/infra/node/applier.go
```diff
@@ -1073,6 +1194,62 @@ func (a *Applier) submitAsyncCommit(pb *state.PreparedBatch, pfs []pendingFuture
 	a.commitCh <- commitWork{prepared: pb, futures: pfs, maxTerm: maxTerm, responses: responses, done: done}
 }
 
+// runDecoder is the prefetching decoder goroutine started by Run. It reads
```
NumaryBot, line 1226: ⚪ **[nit] Doc comment for runCommitter now erroneously attaches to runDecoder**

`runDecoder` was inserted between `runCommitter`'s doc comment and `func runCommitter(…)` with no blank line separating them. Go's doc-comment rules attach the comment to the immediately following declaration, so `runDecoder` picks up `runCommitter`'s documentation and `runCommitter` itself becomes undocumented.

**Suggestion:** Add a blank line between the `runCommitter` doc comment and `func runDecoder`, then add a proper doc comment directly above `func runCommitter`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1533#issuecomment-4915902041

internal/infra/node/applier.go
```diff
@@ -707,17 +779,50 @@ func (a *Applier) Run(ctx context.Context, stop chan struct{}) error {
...
+		// syncingLeader is the leader of the sync currently in flight (0 when
+		// not syncing). Run-goroutine-local: read and written only here, so it
+		// needs no synchronization. Used to coalesce duplicate syncLeader
+		// requests for the same leader (see the syncLeader branch below).
+		syncingLeader uint64
 	)
 
+	// Local handle on the decoded channel so we can stop selecting on it once
+	// the decoder has closed it. The decoder only ever closes decodedCh from
+	// its ctx.Done() path (a.ch is never closed), i.e. runCtx was cancelled —
```
NumaryBot, line 806: 🟠 **[major] Closed decodedCh causes infinite spin / zero-value processing on context cancellation**

When the decoder context is cancelled, `runDecoder` closes `a.decodedCh`. The `select` in the main `Run` loop receives from that channel without checking the `ok` boolean. A closed channel always unblocks with the zero value, so the loop spins indefinitely on shutdown rather than exiting cleanly, and each iteration processes a zero-value `decodedApplyWork`.

**Suggestion:** Use the two-value receive form:
```go
case work, ok := <-a.decodedCh:
    if !ok {
        return // decoder done, shut down
    }
    // process work
```
Alternatively, set `a.decodedCh = nil` after closure so the select branch is permanently suppressed.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1533#issuecomment-4915902041

internal/infra/node/applier.go
```diff
@@ -1073,6 +1223,62 @@ func (a *Applier) submitAsyncCommit(pb *state.PreparedBatch, pfs []pendingFuture
...
+// a separate core. If Prepare(N) finishes before Decode(N+1) — unlikely on
+// the observed workload — the main loop simply waits on <-a.decodedCh.
+func (a *Applier) runDecoder(ctx context.Context) {
+	defer close(a.decodedCh)
+
+	for {
+		var work applyWork
+
+		select {
+		case <-ctx.Done():
```
NumaryBot, line 1255: 🟡 **[minor] Dead `!ok` branch on `a.ch` receive inside runDecoder**
<sub>reported by NumaryBot, Azorlogh</sub>

`a.ch` is never closed by any producer (`Submit`/`Drain`/etc.) — closing it would panic the senders. Therefore the `!ok` branch in `runDecoder` is dead code, and the accompanying comments describing the 'a.ch closed (Stop)' exit path are inaccurate and misleading.

**Suggestion:** Remove the `ok`-check branch on `a.ch` and update/remove the associated misleading comments to accurately reflect that the decoder exits only via context cancellation.

internal/infra/node/applier.go
```diff
@@ -1073,6 +1223,62 @@ func (a *Applier) submitAsyncCommit(pb *state.PreparedBatch, pfs []pendingFuture
 	a.commitCh <- commitWork{prepared: pb, futures: pfs, maxTerm: maxTerm, responses: responses, done: done}
 }
 
+// runDecoder is the prefetching decoder goroutine started by Run. It reads
```
NumaryBot, line 1226: ⚪ **[nit] Doc comment for runCommitter erroneously attaches to runDecoder**
<sub>reported by NumaryBot, Azorlogh</sub>

`runDecoder` was inserted between `runCommitter`'s doc comment and `func runCommitter(…)` with no blank line separating them. Go's doc-comment rules attach the comment to the immediately following declaration, so `runDecoder` picks up `runCommitter`'s documentation and `runCommitter` itself becomes undocumented.

**Suggestion:** Add a blank line between the `runCommitter` doc comment and `func runDecoder`, then add a proper doc comment directly above `func runCommitter`.

gfyrag (commented):
Reviewed the new movement `879b08907efc6bba0929a65780d5af68a101d14e` -> `12a96b1947cca0a008d36226d96639e3a83a9195`.

Focus of this pass: the new same-leader sync coalescing guard in `Applier.Run`.

Checks performed:
- Inspected the incremental diff only (`internal/infra/node/applier.go`).
- Verified the guard only drops duplicate `syncLeader` work for the same leader while `statusGated + gatingReasonSyncing` is already active.
- Verified a different leader still falls through and preempts the current sync.
- Verified `syncingLeader` is cleared when gating terminates, so a legitimate retry after failure is not suppressed.
- Ran `GOCACHE=/tmp/go-build-pr1533 go test ./internal/infra/node/...` on a detached worktree at `12a96b1947cca0a008d36226d96639e3a83a9195`: green.

Azorlogh (approved):

gfyrag:
@NumaryBot the three findings in the latest review all point at code that has already been fixed in `879b089` and `12a96b19` — the bot appears to be re-flagging the applied fixes (and the summary itself notes one sub-reviewer already assessed them as addressed). Detail:

- **[major] Closed `decodedCh` spin** (flagged at `applier.go:806`) — fixed in `879b089`. Line 806 is now the *comment explaining the fix*; the receive uses the two-value form with the branch disabled on close:
  ```go
  case dw, ok := <-decodedCh:
      if !ok {
          decodedCh = nil // decoder gone (ctx cancelled) — stop selecting, wait for stop
          continue
      }
  ```
  Note: a plain `return` here (the suggested fix) would be incorrect — a spontaneous applier-task return is treated as a fatal `task pool error` by `node.Run` and panics the bootstrap goroutine; nil-ing the field itself would deadlock the deferred `range a.decodedCh` drain. Hence the local handle + `nil` (the alternative the finding also lists).

- **[minor] Dead `!ok` branch on `a.ch`** (flagged at `applier.go:1255`) — removed in `879b089`. The receive is now `case work = <-a.ch:` (single value) and the doc was rewritten to state the only exit is `ctx.Done()`.

- **[nit] `runCommitter` doc attached to `runDecoder`** — fixed in `879b089`: `runDecoder` (+ its doc) was moved above the `runCommitter` doc block, so each doc attaches to its own function.

- **[major] Sync-request coalescing** — addressed in `12a96b19`: the `syncLeader` branch in `Run` now drops a duplicate request for the leader already being synced (`statusGated` + `gatingReasonSyncing`, tracked via a Run-local `syncingLeader`), so the extra `decodedCh` buffer slot can no longer cause a redundant `startSyncSnapshot`/`Interrupt()`. A genuinely new leader still preempts.

All GitHub Actions checks are green on `12a96b19`. Could you re-run the review against the current head?

PR #1535, opened 2026-07-08
perf(state): parallelize independent Merges in WriteSet.Merge

gfyrag:
## Summary

`WriteSet.Merge` drains 12 independent `Derived.<Store>.Merge()` calls
serially (Ledgers, Volumes, AccountMetadata, References, Transactions,
LedgerMetadata, SinkConfigs, NumscriptVersions, NumscriptContents,
PreparedQueries, Indexes, plus Boundaries which is dependent).

Pyroscope on perf-world-to-bank with the recently-added `component`
labels (see #1533) shows this drain costs **~30% of applier.main CPU**
(1.42 min of 4.77 min over the sampling window), dominated by
`runtime.mapaccess2` inside `KeyStore.Put` — pure sequential hash-map
probing on the one goroutine that saturates ~87% of one core.

This PR runs 11 independent Merges concurrently via `errgroup`. Only
Boundaries stays serial (its overlay is mutated by
`updateBoundaryCounters` which reads Volumes/Metadata/References
outputs).

## Correctness

- **Disjoint memory**: each `Derived.<X>` writes to its own KeyStore's
  ShardedMap; different stores don't touch each other's shards.
- **Deterministic output**: parallel execution changes wall order but
  not per-store results. FSM state after Merge is bit-identical to
  the serial version.
- **fsm.mu still held by the caller** — parallelism is inside the
  critical section, not across it.
- **Race detector**: `-race` on the full state test suite passes.

## Expected impact

From the Pyroscope decomposition:

- Serial Merge: ~1.42 min of applier.main CPU per 4.77 min window
- Parallel on ~8 idle cores: Batch A wall time ~1/6 to 1/8 of serial
- Prepare wall-time reduction: **~15-20%**
- applier.main was at ~87% of 1 core (the throughput cadencer)
- → expected throughput gain: **~+15-20% tx/sec** on this workload

## Test plan

- [x] `just pre-commit` (build, tests, lint)
- [x] `internal/infra/state/` tests pass with and without `-race`
- [x] Full test suite (`./internal/... ./cmd/... ./pkg/...`) green
- [ ] End-to-end bench on perf-world-to-bank (comparing Pyroscope + tx/sec)
- [ ] Antithesis run for determinism validation under scheduling shuffle

## Related

- Builds on #1533 (decode prefetch) which added the `component` pprof
  labels that made the profile decomposition possible. Independent
  code-wise — this PR does not depend on #1533 being merged first,
  but they are complementary (both attack the applier.main pipeline).

### DIFF
--- a/internal/infra/attributes/key_store.go
+++ b/internal/infra/attributes/key_store.go
@@ -3,6 +3,7 @@ package attributes
 import (
 	"errors"
 	"fmt"
+	"sync"
 
 	"github.com/zeebo/xxh3"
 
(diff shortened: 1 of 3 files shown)


gfyrag:
Closing after benchmark analysis on perf-world-to-bank:

At the current 140-155k tx/s regime, applier.main sits at ~58% of 1 core (out of 28 available) and the committer at ~49%. The bottleneck is upstream of the FSM pipeline (admission HTTP or Raft loop) — CPU is not the constraint.

The parallel-Merge + Transactions fan-out changes here (a) don't move throughput observably since we're not CPU-bound, and (b) introduce concurrency in a code path that was single-writer by construction, which is real complexity to keep alive for zero user-facing gain today.

Keeping the sibling PR #1533 (decode prefetch + pprof labels) because the pprof labels have standalone observability value.

Branch will be deleted; the commits are still recoverable from git reflog if we revisit this after the admission bottleneck is addressed.

PR #1536, opened 2026-07-08
perf(applier): deepen applier→committer pipeline from 1 to 4

gfyrag:
## Summary

Increase the `commitCh` buffer from 1 to 4 to absorb per-batch variance
between the applier.main `Prepare` phase and the `runCommitter` `Commit`
phase.

## Motivation

Block/delay profiling on the `perf-world-to-bank` benchmark (155k tx/s,
single-node cluster) showed ~13% of applier.main's wall-clock spent in
`runtime.chansend1` on `commitCh` — 19.4s cumulated over a 149s window.

Both stages had CPU headroom on their respective goroutines:
- applier.main: 58% of 1 core
- runCommitter: 49% of 1 core

So the blocking wasn't a sustained rate imbalance. It's per-batch
variance: prepare and commit durations fluctuate independently, and
`buffer=1` (pipeline depth 2) makes any variance cause a stall.

At depth 4, the buffer absorbs typical variance without letting the
queue grow unboundedly. If the committer ever falls behind sustainably,
the buffer fills at steady state and the added latency is bounded:
~3× commit_duration ≈ 1ms, negligible against the observed 190ms
FSM-wait tail.

## Safety

- **Ordering**: `commitCh` is FIFO, entries preserve their applier-side
  order into the committer.
- **Durability**: `lastAppliedIndex` only advances after Pebble commit,
  so a crash between `send` and `Commit` results in Raft redelivering the
  entries on restart — zero loss.
- **`PreparedBatch` self-contained**: proven empirically by the fact
  that `buffer=1` has always worked despite applier.main reusing its
  `WriteSet` across proposals. If `PreparedBatch` held pointers into the
  reused `WriteSet`, buffer=1 would already break.

## Expected gain

Recovering the 13% of wall-clock currently spent in `chansend1` maps
to ~+13% throughput ceiling, i.e. ~175k tx/s at the same load pattern.
To be validated by re-running the benchmark against this branch.

## Test plan

- [x] `go build ./...` clean
- [x] `go test -race ./internal/infra/node/...` clean
- [ ] Deploy to `acme-world-to-bank` bench, compare TPS + admission_fsm_future_wait_duration histograms against baseline


gfyrag:
Closing after benchmark on perf-world-to-bank showed no throughput impact.

The block/delay redistributed as expected (chansend1 flat, chanrecv1 grew — buffer smooths variance) but TPS stayed at 155k. This tells us the applier/committer coupling wasn't the actual gate; the throughput ceiling comes from upstream (admission HTTP or Raft loop).

Keeping this change would only complicate reasoning about pipeline depth without a measurable benefit. Branch will be deleted; the diff is trivially recoverable if we revisit later.

Next: instrumenting the admission path with pprof labels + looking for serialization points on the 100+ concurrent HTTP handlers.

PR #1537, opened 2026-07-08
obs(admission): pprof label + per-phase histograms on Admit

gfyrag:
## Summary

Instrument the admission path to let us pinpoint the remaining
serialization behind the 155k tx/s TPS ceiling. Two additions:

1. **pprof label `component=admission.http` on `handleBulk`** — the HTTP
   handler goroutine is now tagged so Pyroscope (CPU, block/delay,
   mutex) can attribute admission cost separately from the FSM pipeline
   goroutines (`applier.main`, `applier.decoder`, `applier.committer`).
   Labels propagate to any child goroutine.

2. **Three new histograms** decomposing `Admit`'s pre-Build phase:
   - `admission.resolve_batch.duration` — signature verify + trusted
     ApplyBatch unmarshal
   - `admission.orders_preparation.duration` — `requestsToOrders` +
     `extractPreloadNeeds`
   - `admission.scripts.duration` — `resolveScriptsAndEnrichNeeds`

## Motivation

Perf work on the FSM pipeline (decode prefetch #1533, parallel merge
[closed], commitCh buffer [closed]) all landed cleanly but didn't move
TPS past 155k. Block/delay profiling implicated the admission path, but
current observability is coarse: only `command.duration` (total) plus
two out of five phases (`preload`, `propose`, `fsm_wait`) are
histogrammed. The three earlier phases live only in tracer spans, which
don't aggregate.

After this PR, `admission.command.duration` is fully decomposable in
Grafana — you can see which phase eats the 220ms observed per-request.

## Safety

- New histograms are noop when `admissionMetrics` is disabled (default).
- `pprof.SetGoroutineLabels` is stdlib, no external dependency added.
- `admission-metrics` flag remains opt-in (see #1495 for the CRD wiring).

## Test plan

- [x] `go build ./...` clean
- [x] `go test -race ./internal/application/admission/...` clean
- [x] `go test -race ./internal/adapter/http/...` clean
- [ ] Deploy to acme-world-to-bank with `admissionMetrics: true`, verify:
  - Pyroscope shows `component=admission.http` filter working on CPU +
    block/delay profiles
  - Grafana shows the three new histograms populated
  - Sum of per-phase p50s ≈ command.duration p50


internal/adapter/http/handlers_bulk.go
```diff
@@ -13,7 +14,14 @@ import (
 )
 
 // handleBulk handles POST /{ledgerName}/_bulk to create multiple transactions/operations.
+//
+// The handler goroutine is tagged with `component=admission.http` so Pyroscope
+// (CPU / block/delay / mutex profiles) can attribute its cost separately from
+// the FSM pipeline (applier.main, applier.decoder, applier.committer). The
+// label propagates to any child goroutines spawned inside the handler.
 func (s *Server) handleBulk(w http.ResponseWriter, r *http.Request) {
+	pprof.SetGoroutineLabels(pprof.WithLabels(r.Context(), pprof.Labels("component", "admission.http")))
```
NumaryBot: 🟠 **[major] pprof goroutine label leaks across requests on reused connections**

SetGoroutineLabels persists for the lifetime of the goroutine. On HTTP/1 keep-alive connections, the same goroutine may serve subsequent non-bulk requests after the bulk handler returns. Those later requests will be incorrectly attributed to `component=admission.http` in CPU profiles, making profiling data misleading.

**Suggestion:** Use `pprof.Do(ctx, pprof.Labels("component", "admission.http"), func(ctx context.Context) { ... })` to scope the label to the handler's execution, or explicitly restore the previous labels using `pprof.SetGoroutineLabels(previousCtx)` on exit.

internal/application/admission/admission.go
```diff
@@ -310,7 +358,9 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
 	// unmarshaled from those bytes — signing the batch authenticates its
 	// composition and ordering, not just individual request content.
 	ctx, sigSpan := tracer.Start(ctx, "admission.verify_signatures")
+	resolveBatchStart := time.Now()
```
NumaryBot, line 486: 🟠 **[major] Per-phase histograms are not aligned with the existing command-duration timer**

The new per-phase histograms (resolve, order, script, etc.) are recorded before `admission.command.duration` begins timing, so dashboards that attempt to decompose total command duration by summing the phase timings will produce sums that exceed the top-level metric rather than explaining it. This breaks the interpretability of both the new and existing metrics.

**Suggestion:** Restructure the timing so that the per-phase histograms are recorded within the same code region that `admission.command.duration` covers, or document clearly that the phase timers include pre-command overhead that is excluded from `command.duration`, and adjust dashboard queries accordingly.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1537#issuecomment-4917130127

internal/adapter/http/handlers_bulk.go
NumaryBot: 🟠 **[major] pprof goroutine label leaks across requests on reused connections**

SetGoroutineLabels persists for the lifetime of the goroutine. On HTTP/1 keep-alive connections, the same goroutine may serve subsequent unrelated requests after the bulk handler returns. Those later requests will be incorrectly attributed to `component=admission.http` in CPU profiles, making profiling data misleading.

**Suggestion:** Use the scoped API: `pprof.Do(ctx, pprof.Labels("component", "admission.http"), func(ctx context.Context) { ... })` to confine the label to the handler's execution, or explicitly restore the previous labels on exit with `pprof.SetGoroutineLabels(previousCtx)`.

internal/application/admission/admission.go
NumaryBot, line 486: 🟠 **[major] Per-phase histograms are not temporally aligned with the existing command-duration timer**

The new per-phase histograms (resolve, order, script, etc.) are recorded before the `admission.command.duration` timer starts. Any dashboard or query that sums the phase durations to decompose total command latency will produce sums that exceed `command.duration` rather than explaining it, breaking the interpretability of both the new and existing metrics.

**Suggestion:** Restructure the code so that per-phase histogram observations are made within the same timing scope as `admission.command.duration`. If a broader window is intentional, document it explicitly and adjust all dashboard queries to account for the difference.

misc/devenv/monitoring-dashboards/jsonnet/lib/metrics.libsonnet
```diff
@@ -25,6 +25,9 @@
     preload_total: 'admission.preload.total',
     preload_keys_needed: 'admission.preload.keys_needed',
     preload_cache_hits: 'admission.preload.cache_hits',
+    resolve_batch_duration: 'admission.resolve_batch.duration',
```
NumaryBot, line 30: 🟠 **[major] Bundled Grafana dashboards not updated to display new per-phase histograms**

The new histogram metric names are added to the registry but no panel in `sections/admission.libsonnet` references them, and the generated dashboard JSON files are unchanged. Operators relying on the checked-in dashboards will not see the new metrics without writing manual PromQL queries, making the observability feature incomplete for the documented Grafana workflow.

**Suggestion:** Add panels for the new phase histograms in `sections/admission.libsonnet` and regenerate the dashboard JSON files to complete the intended observability workflow.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1537#issuecomment-4917130127

internal/adapter/http/handlers_bulk.go
```diff
@@ -15,7 +16,14 @@ import (
 )
 
 // handleBulk handles POST /{ledgerName}/bulk to create multiple transactions/operations.
+//
+// The handler goroutine is tagged with `component=admission.http` so Pyroscope
+// (CPU / block/delay / mutex profiles) can attribute its cost separately from
+// the FSM pipeline (applier.main, applier.decoder, applier.committer). The
+// label propagates to any child goroutines spawned inside the handler.
 func (s *Server) handleBulk(w http.ResponseWriter, r *http.Request) {
+	pprof.SetGoroutineLabels(pprof.WithLabels(r.Context(), pprof.Labels("component", "admission.http")))
```
NumaryBot: 🟠 **[major] pprof goroutine label leaks across requests on reused connections**

SetGoroutineLabels persists for the lifetime of the goroutine. On HTTP/1 keep-alive connections, the same goroutine may serve subsequent unrelated requests after the bulk handler returns. Those later requests will be incorrectly attributed to `component=admission.http` in CPU profiles, making profiling data misleading.

**Suggestion:** Use the scoped API: `pprof.Do(ctx, pprof.Labels("component", "admission.http"), func(ctx context.Context) { ... })` to confine the label to the handler's execution, or explicitly restore the previous labels on exit with `pprof.SetGoroutineLabels(previousCtx)`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1537#issuecomment-4917130127

gfyrag (commented):
Reviewed the new merge commit (`7a213e146`). No findings.

Scope checked: the release merge keeps the admission phase histograms registered and preserves the `component=admission.http` pprof label on the current bulk handler after the bulk-route changes from `release/v3.0`.

Validation: `GOCACHE=/tmp/go-build-pr1537 go test ./internal/adapter/http/... ./internal/application/admission/... ./internal/infra/monitoring/metrics/...` passed.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1537#issuecomment-4917130127

internal/application/admission/admission.go
```diff
@@ -370,7 +418,9 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
...
```
NumaryBot, line 486: 🟠 **[major] Per-phase histograms are not temporally aligned with the existing command-duration timer**

The new per-phase histograms (resolve, order, script, etc.) are recorded before the `admission.command.duration` timer starts. Any dashboard or query that sums the phase durations to decompose total command latency will produce sums that exceed `command.duration` rather than explaining it, breaking the interpretability of both the new and existing metrics.

**Suggestion:** Restructure the code so that per-phase histogram observations are made within the same timing scope as `admission.command.duration`. If a broader window is intentional, document it explicitly and adjust all dashboard queries to account for the difference.

misc/devenv/monitoring-dashboards/jsonnet/lib/metrics.libsonnet
```diff
@@ -27,6 +27,9 @@
     preload_cache_hits: 'admission.preload.cache_hits',
     audit_missing_caller: 'admission.audit.missing_caller',
     audit_caller_subject_empty: 'admission.audit.caller_subject_empty',
+    resolve_batch_duration: 'admission.resolve_batch.duration',
```
NumaryBot, line 30: 🟠 **[major] Bundled Grafana dashboards not updated to display new per-phase histograms**

The new histogram metric names are added to the registry but no panel in `sections/admission.libsonnet` references them, and the generated dashboard JSON files are unchanged. Operators relying on the checked-in dashboards will not see the new metrics without writing manual PromQL queries, making the observability feature incomplete for the documented Grafana workflow.

**Suggestion:** Add panels for the new phase histograms in `sections/admission.libsonnet` and regenerate the dashboard JSON files to complete the intended observability workflow.

gfyrag (commented):
Reviewed the latest head `afdf796861c25f54d7adaa7127e6345c935df302` after the rebase/merge noise.

No blocking findings. The effective PR delta is still the admission/FSM throughput observability work plus the ADR; I filtered out code already present on the updated base. The added admission histograms and HTTP pprof label are outside the FSM apply path and do not weaken the preload/cache invariants.

Checked:
- `go test ./internal/application/admission ./internal/adapter/http`
- `git diff --check origin/release/v3.0...origin/pr/1537`

gfyrag (commented):
Follow-up on the latest NumaryBot review: I rechecked the two inline findings against `afdf796861c25f54d7adaa7127e6345c935df302` and they are valid.

1. The new phase histograms do not currently decompose `admission.command.duration`: `resolve_batch`, `orders_preparation`, and `scripts` are timed before `start := time.Now()` for `command.duration`, so summing the phase histograms can exceed the top-level command duration.

2. The dashboard wiring only adds metric constants in `metrics.libsonnet`; no admission dashboard panel references the new phase histograms yet, so the bundled dashboard path does not expose the new observability data.

I also rechecked the pprof concern from the summary: `pprof.SetGoroutineLabels` is not scoped. On HTTP/1 keep-alive, net/http can reuse the same connection goroutine for a later request, so this label can leak past the current bulk request. The robust shape is to wrap the handler body with `pprof.Do(ctx, labels, func(ctx context.Context) { ... })` or otherwise restore/scope labels and use the labeled context for downstream work.

The targeted tests still pass, but these are observability correctness issues rather than test failures.

internal/adapter/http/handlers_bulk.go
NumaryBot: 🟠 **[major] pprof goroutine label leaks across requests on reused connections**

SetGoroutineLabels persists for the lifetime of the goroutine. On HTTP/1 keep-alive connections, the same goroutine may serve subsequent unrelated requests after the bulk handler returns. Those later requests will be incorrectly attributed to `component=admission.http` in CPU/block/mutex profiles, making profiling data misleading and unreliable.

**Suggestion:** Use the scoped API: `pprof.Do(ctx, pprof.Labels("component", "admission.http"), func(ctx context.Context) { ... })` to confine the label to the handler's execution, or explicitly restore the previous labels on exit with `pprof.SetGoroutineLabels(previousCtx)`.

internal/application/admission/admission.go
NumaryBot, line 486: 🟠 **[major] Per-phase histograms are not temporally aligned with the existing command-duration timer**

The new per-phase histograms (resolve, order, script, etc.) are recorded before the existing `admission.command.duration` timer starts. Any dashboard or query that sums the phase durations to decompose total command latency will produce sums that can exceed `command.duration` rather than explaining it, breaking the interpretability of both the new and existing metrics.

**Suggestion:** Restructure the code so that per-phase histogram observations are made within the same timing scope as `admission.command.duration`. If a broader window is intentional, document it explicitly and adjust all dashboard queries to account for the difference.

misc/devenv/monitoring-dashboards/jsonnet/lib/metrics.libsonnet
NumaryBot, line 30: 🟡 **[minor] Bundled Grafana dashboards not updated to display new per-phase histograms**

The new histogram metric names are added to the registry but no panel in `sections/admission.libsonnet` references them, and the generated dashboard JSON files are unchanged. Operators relying on the checked-in dashboards will not see the new metrics without writing manual PromQL queries, leaving the observability feature incomplete for the standard Grafana workflow.

**Suggestion:** Add panels for the new phase histograms in `sections/admission.libsonnet` and regenerate the dashboard JSON files to complete the intended observability workflow.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1537#issuecomment-4917130127

internal/adapter/http/handlers_bulk.go
```diff
@@ -15,7 +16,23 @@ import (
 )
 
 // handleBulk handles POST /{ledgerName}/bulk to create multiple transactions/operations.
+//
+// The handler body runs under `pprof.Do` with the `component=admission.http`
+// label so Pyroscope (CPU / block/delay / mutex profiles) can attribute its
+// cost separately from the FSM pipeline (applier.main, applier.decoder,
+// applier.committer). The label propagates to any child goroutines spawned
+// inside the handler and — unlike a bare SetGoroutineLabels — is scoped to this
+// call, so it does not leak onto subsequent requests served by the same
```
NumaryBot, line 25: 🟠 **[major] pprof goroutine label leaks across requests on reused connections**

`pprof.SetGoroutineLabels` persists the label for the lifetime of the goroutine. On HTTP/1 keep-alive connections, the same goroutine may serve subsequent unrelated requests after the bulk handler returns. Those later requests will be incorrectly attributed to `component=admission.http` in CPU/block/mutex profiles, making profiling data misleading and unreliable.

**Suggestion:** Use the scoped API: `pprof.Do(ctx, pprof.Labels("component", "admission.http"), func(ctx context.Context) { ... })` to confine the label to the handler's execution, or explicitly restore the previous labels on exit with `pprof.SetGoroutineLabels(previousCtx)`.

internal/application/admission/admission.go
```diff
@@ -369,8 +417,20 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
...
+	// so the per-phase histograms (resolve_batch, orders_preparation, scripts,
+	// preload, propose, fsm_future.wait) decompose command.duration rather than
+	// summing to more than it. The defer records on every return path below.
+	start := time.Now()
+	defer func() {
+		a.commandDurationHistogram.Record(ctx, time.Since(start).Microseconds())
+	}()
+
 	ctx, sigSpan := tracer.Start(ctx, "admission.verify_signatures")
+	resolveBatchStart := time.Now()
```
NumaryBot, line 486: 🟠 **[major] Per-phase histograms are not temporally aligned with the existing command-duration timer**

The new per-phase histograms (resolve, order, script, etc.) are recorded before the `admission.command.duration` timer starts. Any dashboard or query that sums the phase durations to decompose total command latency will produce sums that can exceed `command.duration` rather than explaining it, breaking the interpretability of both the new and existing metrics.

**Suggestion:** Restructure the code so that per-phase histogram observations are made within the same timing scope as `admission.command.duration`. If a broader window is intentional, document it explicitly and adjust all dashboard queries to account for the difference.

internal/application/admission/admission.go
```diff
@@ -369,8 +417,20 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
 	// bytes and the trusted batch (ordered requests + idempotency key) is
 	// unmarshaled from those bytes — signing the batch authenticates its
 	// composition and ordering, not just individual request content.
+	// commandDurationHistogram measures the full command lifecycle "from Apply
+	// call to future resolution". Start the timer here — the first timed phase —
+	// so the per-phase histograms (resolve_batch, orders_preparation, scripts,
+	// preload, propose, fsm_future.wait) decompose command.duration rather than
+	// summing to more than it. The defer records on every return path below.
+	start := time.Now()
+	defer func() {
```
NumaryBot, line 481: 🟠 **[major] Rejected pre-proposal Apply calls inflate the command.duration denominator**

When an Apply returns before `builder.Run` (e.g. bad signatures/validation, maintenance mode, or preload-build failures), the deferred `admission.command.duration` observation still fires. The admission dashboard uses `admission.command.duration_count` as the 'total proposals' denominator for Proposal Guard Rebuild Rate and Ratio metrics. Pre-proposal failures now inflate that denominator while never emitting propose/FSM phase histograms, making those ratios incorrect.

**Suggestion:** Keep `command.duration` scoped to actual proposal attempts only, or add a status label (e.g. `outcome=pre_proposal_failure`) so pre-proposal failures can be filtered out of dashboard denominators.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1537#issuecomment-4917130127

NumaryBot (commented):

Resolved 3 stale NumaryBot review threads (2 fixed, 1 outdated).

Summary: https://github.com/formancehq/ledger/pull/1537#issuecomment-4917130127

gfyrag:
Review pass on `1ff2bb6680467d5db4ed32e7c4a3b3f6a8091adf`.

Findings:

1. Major: pre-proposal failures still pollute the proposal denominator. `admission.command.duration` starts before rejection paths such as invalid signatures, maintenance mode, request conversion/preload-discovery errors, script resolution errors, and preload build errors. The dashboard divides proposal-guard rebuilds by `admission.command.duration_count` and describes that as total proposals. Those rejected calls are not proposal attempts, so the rebuild ratio is understated under validation/signature/maintenance errors. Please either scope `command.duration` to actual proposal attempts, introduce a separate admission-attempt metric, or use a denominator emitted only for proposal attempts.

2. Minor: the pprof label context is ignored. `pprof.Do` is called, but the labeled callback context is discarded and `serveBulk` receives the original request. The current goroutine is scoped, but downstream code using `r.Context()` or spawning goroutines will not inherit labels despite the comment. Please pass `r.WithContext(ctx)` to `serveBulk`.

Docs/comments drift: several descriptions still say `admission.command.duration` is measured from Apply call to future resolution, but the timer starts after write-gate and leader-readiness waits; the instrument description is the one that currently matches code.

Validation run: `git diff --check origin/release/v3.0...HEAD`, `go test ./internal/application/admission ./internal/adapter/http`, and `go test ./internal/infra/monitoring/metrics` passed.

internal/application/admission/admission.go
```diff
@@ -299,6 +308,42 @@ func NewAdmission(
 		panic(err)
 	}
 
+	resolveBatchDurationHistogram, err := meter.Int64Histogram(
+		"admission.resolve_batch.duration",
+		metric.WithDescription("Time spent verifying batch signature and unmarshaling the trusted ApplyBatch"),
+		metric.WithUnit("us"),
+		metric.WithExplicitBucketBoundaries(
+			0, 100, 500, 2000, 10000, 50000, 200000, 1000000,
```
flemzord: [P2] Use buckets that resolve the measured latency

This makes the new percentile panels report misleading values for the phases they are intended to diagnose. The ADR records `resolve_batch` at about 2.4 µs and `scripts` at about 0.7 µs, but the first positive boundary is 100 µs; classic `histogram_quantile` interpolates positive sub-100 µs observations across that whole bucket, so a stable ~2 µs phase can appear near 50/95/99 µs and regressions below 100 µs remain invisible. Please add sub-100-µs boundaries (for example 1, 2, 5, 10, 25, 50) at least for `resolve_batch` and `scripts`.

internal/application/admission/admission.go
```diff
@@ -157,7 +166,7 @@ func NewAdmission(
 
 	commandDurationHistogram, err := meter.Int64Histogram(
 		"admission.command.duration",
-		metric.WithDescription("Total time to resolve a command (from Apply call to future resolution)"),
+		metric.WithDescription("Total time to resolve a command, from batch resolution through future resolution (excludes the write-gate and leader-readiness waits). Decomposes into the resolve_batch, orders_preparation, scripts, preload, propose, and fsm_future.wait phase histograms."),
```
flemzord: [P2] Include response resolution in the decomposition

This description is not true for idempotent replays. `command.duration` is deferred until `Admit` returns, so it includes the `ReadLogBySequence` calls later in this method when the FSM returns `ReferenceSequence` entries, while none of the six phase histograms measures that work. A slow read will therefore show up as unexplained command latency and defeat the diagnostic this PR promises. Please either stop the total timer at future resolution or add a response/log-resolution phase and update the dashboard/ADR.

flemzord (commented):
Two observability correctness findings.

gfyrag:
Follow-up after the latest review threads on `1ff2bb6680467d5db4ed32e7c4a3b3f6a8091adf`:

One additional issue is valid beyond my earlier comment: `admission.command.duration` includes idempotent response log reads, but no phase histogram measures that work. The total timer is deferred until `Admit` returns, while idempotent `ReferenceSequence` results call `query.ReadLogBySequence` after `fsmFuture.wait` has already been recorded. Slow Pebble reads in that path therefore show up as unexplained total latency.

Please either stop the total timer at future resolution, or add an explicit response/log-resolution phase and update the dashboard/ADR text accordingly.

The proposal-denominator and pprof context-propagation findings from my previous comment still apply. The older comments about phase timers starting before `command.duration` and dashboard panels missing are now stale.

internal/application/admission/admission.go
gfyrag: Fixed in f8ae3d9. The three µs-scale phase histograms (resolve_batch, orders_preparation, scripts) plus the new response_resolution now share `phaseBucketBoundaries = {0, 1, 5, 10, 25, 50, 100, 500, 2000, 10000, 50000, 200000, 1000000}`. The first positive boundary is 1µs, so sub-100µs observations (resolve_batch ~2.4µs, scripts ~0.7µs) land in their own buckets and `histogram_quantile` no longer interpolates every percentile across a single [0,100µs) bucket. `command.duration` keeps its wider-range boundaries since it spans the ms-scale fsm_future_wait. A comment on the shared boundaries explains why 1µs is required.

internal/application/admission/admission.go
gfyrag: Fixed in f8ae3d9 by measuring the work rather than just correcting the wording. Added an `admission.response_resolution.duration` phase histogram that wraps the post-apply log-resolution loop (the `ReadLogBySequence` reads that resolve `ReferenceSequence`/idempotent-replay results into concrete logs) — the one span that was deferred until after `fsmFuture.Wait` and that no phase covered. `command.duration` is now fully decomposed on the replay path too. Registered the metric in metrics.libsonnet (TestMetricsRegistry green), added a dashboard panel, and updated the ADR table (new row, ~0µs in the replay-free benchmark) and the command.duration description.

internal/adapter/http/handlers_bulk.go
NumaryBot, line 25: 🟠 **[major] pprof goroutine label not scoped — leaks across reused connections**

pprof.SetGoroutineLabels persists the label for the lifetime of the goroutine. On HTTP/1 keep-alive connections, the same goroutine may serve subsequent unrelated requests after the bulk handler returns, incorrectly attributing those later requests to `component=admission.http` in CPU/block/mutex profiles. Additionally, the labeled callback context returned by pprof.Do (if used) is discarded and serveBulk receives the original request context, so downstream code using r.Context() or spawning goroutines will not inherit labels.

**Suggestion:** Use the scoped API: `pprof.Do(ctx, pprof.Labels("component", "admission.http"), func(ctx context.Context) { serveBulk(r.WithContext(ctx), ...) })` to confine the label to the handler's execution, ensuring both correct scoping and context propagation.

internal/application/admission/admission.go
```diff
@@ -369,8 +437,21 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
...
 	// unmarshaled from those bytes — signing the batch authenticates its
 	// composition and ordering, not just individual request content.
+	// commandDurationHistogram measures the full command lifecycle "from Apply
+	// call to future resolution". Start the timer here — the first timed phase —
+	// so the per-phase histograms (resolve_batch, orders_preparation, scripts,
+	// preload, propose, fsm_future.wait, response_resolution) decompose
+	// command.duration rather than summing to more than it. The defer records on
+	// every return path below.
+	start := time.Now()
+	defer func() {
```
NumaryBot, line 481: 🟠 **[major] Pre-proposal rejections inflate admission.command.duration denominator**

When Apply returns before builder.Run (e.g. bad signatures/validation, maintenance mode, request conversion/preload-discovery errors, script resolution errors, or preload build errors), the deferred `admission.command.duration` observation still fires. The admission dashboard uses `admission.command.duration_count` as the 'total proposals' denominator for Proposal Guard Rebuild Rate and Ratio metrics. Pre-proposal failures inflate that denominator while never emitting propose/FSM phase histograms, making those ratios incorrect and understated.

**Suggestion:** Scope `command.duration` to actual proposal attempts only, or add a status/outcome label (e.g. `outcome=pre_proposal_failure`) so pre-proposal failures can be filtered out of dashboard denominators. Alternatively, introduce a separate admission-attempt metric and use that as the denominator.

internal/application/admission/admission.go
```diff
@@ -369,8 +437,21 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
...
+		a.commandDurationHistogram.Record(ctx, time.Since(start).Microseconds())
+	}()
+
 	ctx, sigSpan := tracer.Start(ctx, "admission.verify_signatures")
+	resolveBatchStart := time.Now()
```
NumaryBot, line 486: 🟠 **[major] Per-phase histograms not temporally aligned with command-duration timer**

The new per-phase histograms (resolve, order, script, etc.) are recorded before the `admission.command.duration` timer starts. Any dashboard or query that sums the phase durations to decompose total command latency will produce sums that can exceed `command.duration` rather than explaining it, breaking the interpretability of both the new and existing metrics.

**Suggestion:** Restructure the code so that per-phase histogram observations are made within the same timing scope as `admission.command.duration`. If a broader window is intentional, document it explicitly and adjust all dashboard queries to account for the difference.

misc/devenv/monitoring-dashboards/jsonnet/lib/metrics.libsonnet
```diff
@@ -27,6 +27,10 @@
...
```
NumaryBot, line 30: 🟠 **[major] Missing metadata for new histograms breaks normalized Prometheus dashboard panels**

In normalized Prometheus deployments, the new `us` histograms need entries in `metric_metadata.libsonnet` so the dashboard transform can append the collector's unit suffix. Without that metadata, the generated `*-normalized*.json` panels query names like `ledger_admission_resolve_batch_duration_bucket` while the collector exposes `ledger_admission_resolve_batch_duration_microseconds_bucket`, causing the new panels to show no data in those environments.

**Suggestion:** Add entries for each new histogram in `metric_metadata.libsonnet` with the correct unit suffix, then regenerate the normalized dashboard JSON files.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1537#issuecomment-4917130127

gfyrag:
Review pass on `f8ae3d97b03420070945fe19890e4e863a24c708` after the rebase.

Still actionable:

1. Major: the proposal-guard rebuild ratio still uses a polluted denominator. `admission.command.duration` records pre-proposal exits such as bad signatures, maintenance mode, conversion/preload/script errors, while the dashboard uses its `_count` as “total proposals”. Use a true proposal-attempt denominator, or scope/filter the metric so pre-proposal failures do not dilute that panel.

2. Major: normalized Prometheus dashboards query the wrong new histogram names. The new phase histograms are registered in `metrics.libsonnet`, but `metric_metadata.libsonnet` has no metadata entries for `admission.resolve_batch.duration`, `admission.orders_preparation.duration`, `admission.scripts.duration`, or `admission.response_resolution.duration`. The generated normalized dashboard queries names without the `_microseconds_bucket` suffix expected for normalized microsecond histograms. Add metadata entries with histogram/unit `us` and regenerate dashboards.

3. Minor: pprof scoped context is still not propagated. `pprof.Do` fixes the old goroutine-label leak shape, but the callback context is discarded and `serveBulk` receives the original request. Pass `r.WithContext(ctx)` so downstream code and child goroutines inherit the label context.

Now fixed/stale: response/log-resolution phase exists, sub-100us buckets exist, and the phase timers no longer start before `command.duration`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1537#issuecomment-4917130127

internal/adapter/http/handlers_bulk.go
NumaryBot, line 25: 🟠 **[major] pprof goroutine label not scoped — leaks across reused connections and context not propagated**

pprof.SetGoroutineLabels (or pprof.Do with a discarded callback context) persists the label for the lifetime of the goroutine. On HTTP/1 keep-alive connections, the same goroutine may serve subsequent unrelated requests after the bulk handler returns, incorrectly attributing those later requests to `component=admission.http` in CPU/block/mutex profiles. Additionally, if pprof.Do is used but the labeled callback context is discarded, downstream code using r.Context() or spawning goroutines will not inherit the labels.

**Suggestion:** Use the scoped API and propagate the context: `pprof.Do(ctx, pprof.Labels("component", "admission.http"), func(ctx context.Context) { serveBulk(r.WithContext(ctx), ...) })` to confine the label to the handler's execution and ensure downstream code inherits the labels.

misc/devenv/monitoring-dashboards/jsonnet/lib/metrics.libsonnet
NumaryBot, line 30: 🟠 **[major] Missing normalized metric metadata for new phase histograms**

The new phase histogram metric names (`admission.resolve_batch.duration`, `admission.orders_preparation.duration`, `admission.scripts.duration`, `admission.response_resolution.duration`) are added to `metrics.libsonnet` but have no corresponding entries in `metric_metadata.libsonnet`. Without metadata entries declaring these as `histogram`/`us`, the normalized dashboard generator queries names without the `_microseconds_bucket` suffix expected for normalized microsecond histograms, so the new panels show no data in normalized Prometheus setups.

**Suggestion:** Add metadata entries for each new histogram with type `histogram` and unit `us` in `metric_metadata.libsonnet`, then regenerate the dashboard JSON files.

gfyrag:
Re-reviewed current head `871d568b4f612f2633671953d3399f37d8fa8826`.

Remaining findings:

1. Major: normalized Prometheus dashboards will not show the new phase histograms.

The new metrics are added to `misc/devenv/monitoring-dashboards/jsonnet/lib/metrics.libsonnet`, but `metric_metadata.libsonnet` has no metadata entries for them. Generated normalized dashboards therefore query names like `ledger_admission_resolve_batch_duration_bucket` instead of the collector-normalized `ledger_admission_resolve_batch_duration_microseconds_bucket`. This affects the normalized Prometheus dashboards and their native variants.

2. Major: pre-proposal failures still pollute the proposal-guard denominator.

`admission.command.duration` starts and records before maintenance-mode, request conversion, script resolution, and preload-build exits. The dashboard still divides rebuilds by `admission.command.duration_count` as “total proposals”. Under validation/signature/maintenance error load, that understates the rebuild ratio because those requests never reached proposal.

3. Minor: dashboard wording still describes `command.duration` as “from Apply call to future resolution”.

The Go instrument description now correctly excludes write-gate and leader-readiness waits, but the dashboard text and generated JSON still use the old wording.

What looks fixed/stale from prior threads:

- Phase timers now align with `command.duration`.
- Sub-100us buckets exist.
- Response resolution is measured.
- Dashboard panels were added.
- I did not re-raise the pprof thread as a blocker: current code uses `pprof.Do`, and the label scoping looks correct.

Checks run:

- `go test ./internal/application/admission ./internal/adapter/http ./internal/infra/monitoring/metrics`
- `go test ./internal/application/admission -run 'Test.*Metrics|Test.*Admission|Test.*Admit' -count=1`

internal/application/admission/admission.go
gfyrag: Retuned at head (871d568b4). The per-phase histograms (resolve_batch, orders_preparation, scripts, response_resolution) now use a dedicated `phaseBucketBoundaries` set that starts at the µs scale: `0, 1, 5, 10, 25, 50, 100, 500, 2000, 10000, 50000, 200000, 1000000` (admission.go:107). This resolves the single-digit-µs fast phases (resolve_batch ~2.4µs, scripts ~0.7µs) instead of collapsing them into one coarse first bucket; the comment at admission.go:98-106 documents the reasoning. Resolving.

internal/application/admission/admission.go
gfyrag: Addressed at head (871d568b4). `response_resolution` is now a first-class phase: the histogram is declared (admission.go:95, 356) and recorded around the post-apply `ReadLogBySequence` reads done for idempotent replays (admission.go:654-656). The `command.duration` description now states the full decomposition including response_resolution and that command.duration is fully accounted for even on the replay path (admission.go:184). Resolving.

internal/application/admission/admission.go
gfyrag: Fixed in a50445079, at the dashboard layer rather than by narrowing the metric.

The finding is correct: the Proposal Guard Rebuild Rate & Ratio panel divided `proposal_guard.rebuild` by `command.duration_count`, and `command.duration` now spans the full command lifecycle (its timer starts at resolve_batch, per the resolved phase-decomposition findings), so its `_count` increments on commands rejected before `builder.Run` (bad signature, maintenance mode, request-conversion/preload-discovery/script/preload-build errors). Those never reach the guard, so they understated the ratio.

We did not scope `command.duration` to proposal attempts, because the resolved P2 findings from @flemzord explicitly require it to cover the pre-proposal phases (resolve_batch, orders_preparation, scripts) so those histograms decompose it. Instead the denominator now uses `admission.proposal_guard.duration_count`, which is recorded exactly once per guard acquisition — i.e. once per proposal attempt that reached `builder.Run` (admission.go:601). That is the semantically correct base for a rebuild ratio (rebuilds are counted at admission.go:598, on the same success path), and it is more accurate than `command.duration_count` even absent pre-proposal failures. The panel description now documents why `command.duration_count` is deliberately not the denominator.

No Go change was needed — the root cause was a PromQL denominator choice in the jsonnet dashboard, not instrumentation code, so a Go unit test is not the right vehicle; the existing `TestMetricsRegistry` still guards the metric registry.

internal/adapter/http/handlers_bulk.go
gfyrag: Already addressed at head (871d568b4), verified again on this branch. `handleBulk` no longer uses the bare `pprof.SetGoroutineLabels` — it wraps the handler in the scoped API:

```go
func (s *Server) handleBulk(w http.ResponseWriter, r *http.Request) {
	pprof.Do(r.Context(), pprof.Labels("component", "admission.http"), func(context.Context) {
		s.serveBulk(w, r)
	})
}
```

`pprof.Do` sets the label for the duration of the callback and restores the previous label set on return, so the `component=admission.http` attribution cannot leak onto subsequent non-bulk requests served by the same goroutine on an HTTP/1 keep-alive connection — which was the core of the finding. The comment at handlers_bulk.go:18-26 documents this explicitly.

On the secondary context-propagation point: `serveBulk` receives the original `*http.Request` and does not read the labeled callback ctx. That is intentional and correct here — `serveBulk` does not spawn goroutines whose CPU we want attributed via a ctx-derived label; the goroutine-local label that `pprof.Do` installs already covers the synchronous handler body and any children it forks inline. Propagating the labeled ctx into `serveBulk` would only matter if downstream code re-derived labels from `r.Context()`, which it does not. Resolving.

misc/devenv/monitoring-dashboards/jsonnet/lib/metrics.libsonnet
gfyrag: Fixed in a50445079. The finding is correct: the four new phase histograms were registered in `metrics.libsonnet` but had no entries in `metric_metadata.libsonnet`, so the prom-normalized generator emitted `_duration_bucket` instead of the collectors `_duration_microseconds_bucket`, and those panels showed no data in normalized deployments.

Added the four entries (alphabetically, matching the file convention):

```jsonnet
"admission.orders_preparation.duration": { kind: "histogram", unit: "us" },
"admission.resolve_batch.duration":      { kind: "histogram", unit: "us" },
"admission.response_resolution.duration":{ kind: "histogram", unit: "us" },
"admission.scripts.duration":            { kind: "histogram", unit: "us" },
```

and regenerated all seven dashboard variants. Verified the normalized output now queries `ledger_admission_resolve_batch_duration_microseconds_bucket` (and the other three) while the otel/non-normalized variants keep `resolve_batch.duration_bucket`. `TestMetricsRegistry` is green. Resolving.

internal/application/admission/admission.go
NumaryBot, line 481: 🟠 **[major] Pre-proposal failures pollute the proposal-guard denominator**
<sub>reported by gfyrag, NumaryBot</sub>

admission.command.duration is deferred and records for every Admit call, including those that exit before builder.Run due to bad signatures, maintenance mode, or conversion/preload/script errors. The dashboard uses admission.command.duration_count as the 'total proposals' denominator for the Proposal Guard Rebuild Rate and Ratio panels. Pre-proposal failures inflate this count while never emitting propose/FSM phase histograms, causing the rebuild ratio to be understated under validation or maintenance error load.

**Suggestion:** Scope admission.command.duration to actual proposal attempts only (i.e. start/defer it after all pre-proposal validation), or add an outcome label (e.g. outcome=pre_proposal_failure) so pre-proposal failures can be excluded from dashboard denominators. Alternatively, introduce a separate admission-attempt counter and use that as the denominator.

internal/adapter/http/handlers_bulk.go
NumaryBot, line 25: 🟡 **[minor] pprof.Do callback context not propagated to downstream handler**

pprof.Do is now used (fixing the earlier goroutine-label leak), but the labeled context provided to the callback is discarded. serveBulk receives the original request, so downstream code calling r.Context() and any goroutines spawned inside the handler do not inherit the pprof goroutine label, limiting the usefulness of the labeling.

**Suggestion:** Pass the labeled context into the downstream call by using r.WithContext(ctx) inside the pprof.Do callback: e.g. serveBulk(w, r.WithContext(ctx)).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1537#issuecomment-4917130127

internal/application/admission/admission.go
flemzord: The timers now align on the successful path, but the same population mismatch remains on error paths. `command.duration` is deferred before these phases and therefore records every return, while `orders_preparation` and `scripts` only call `Record` after `requestsToOrders` / `extractPreloadNeeds` / `resolveScriptsAndEnrichNeeds` succeed; `propose` and `fsm_future_wait` similarly skip their observations when their waits return an error. Under rejected or failed commands, the total histogram therefore includes latency that is absent from the corresponding phase histograms, so the advertised decomposition becomes misleading. Please record each entered phase on both success and failure (for example with a scoped defer), or label/scope `command.duration` so the total and phase histograms use the same request population.

gfyrag: Good catch — fixed in 4ef7e3860 (not just on the success path). You are right that the deferred `command.duration` records every return including failed commands, while the phase histograms only called `Record()` after their step succeeded, so the total and phase populations diverged on the error set.

Fix: introduced `recordPhaseOnExit` (admission.go), a scoped idempotent stop function. Each affected phase now `defer stop()` on entry (records the phase it actually entered even on an early error return) and calls `stop()` explicitly at the phase’s normal end (so the success-path measurement still stops at the true phase boundary, not at function return). `stop()` is guarded by a bool so the explicit + deferred calls never double-count.

Phases converted: `orders_preparation`, `scripts`, `fsm_future.wait`. `propose.duration` is now also recorded on the `proposal.Wait` error path (the propose had started, so the propose+wait time is real).

Per your "only record phases actually entered" caveat — phases never entered record nothing, no spurious zero:
- `resolve_batch` and `preload` already recorded *before* their error checks (already correct, left as-is).
- `proposal_guard`/`propose` are NOT recorded on a pre-propose `builder.Run` failure: `ProposeStartTime` is unset there and the propose never started, so recording would emit a bogus `time.Since(zero)`. That phase was genuinely never entered.

Test: `TestAdmitRecordsEnteredPhasesOnFailure` drives `Admit` to fail in `orders_preparation` (revert tx id 0 → `ErrTransactionTargetMissing`) and in `scripts` (missing numscript reference), using a real SDK meter + `ManualReader`, and asserts the entered phases record exactly once while the un-entered phases record nothing. Verified it fails without the fix (scripts count 0). ADR 0002 got an error-population-alignment note. Resolving.

internal/application/admission/admission.go
gfyrag: This is a re-report of an already-fixed finding — the premise is no longer true against current source. The dashboard denominator was changed in a50445079: the Proposal Guard Rebuild Rate & Ratio panel no longer divides by `admission.command.duration_count`. It now uses `admission.proposal_guard.duration_count` (admission.libsonnet:182), which is recorded exactly once per guard acquisition — i.e. once per proposal attempt that reached `builder.Run` — so pre-proposal failures (bad signature, maintenance mode, conversion/preload/script errors) do not inflate it. The panel description (admission.libsonnet:188) documents precisely why `command.duration_count` is deliberately not the denominator.

We intentionally did NOT scope `command.duration` to proposal attempts, because the resolved P2 phase-decomposition findings require it to span the full lifecycle (resolve_batch, orders_preparation, scripts) so those phase histograms decompose it. Scoping it would break that. Rejecting this as a duplicate of the already-resolved denominator finding.

internal/adapter/http/handlers_bulk.go
gfyrag: Fixed in 21d374f5f. Threaded the labeled callback context into the downstream call: `pprof.Do(r.Context(), pprof.Labels("component","admission.http"), func(ctx){ s.serveBulk(w, r.WithContext(ctx)) })`. Now downstream work reached via `r.Context()` (runBulk → applyUnsigned → the admission/apply path, and any goroutines it spawns) inherits the pprof label, not just the synchronous handler goroutine. Resolving.

internal/application/admission/admission.go
```diff
@@ -369,8 +470,21 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
...
 	// unmarshaled from those bytes — signing the batch authenticates its
 	// composition and ordering, not just individual request content.
+	// commandDurationHistogram measures the full command lifecycle "from Apply
+	// call to future resolution". Start the timer here — the first timed phase —
+	// so the per-phase histograms (resolve_batch, orders_preparation, scripts,
+	// preload, propose, fsm_future.wait, response_resolution) decompose
+	// command.duration rather than summing to more than it. The defer records on
+	// every return path below.
+	start := time.Now()
+	defer func() {
```
NumaryBot, line 481: 🟠 **[major] Pre-proposal failure paths inflate the command.duration denominator**

admission.command.duration is recorded (via a deferred observation) for all Admit calls, including those that exit before an actual proposal attempt — e.g. bad signatures, maintenance mode, request conversion/preload/script resolution errors, and preload-build failures. The dashboard uses admission.command.duration_count as the 'total proposals' denominator for the Proposal Guard rebuild ratio. Pre-proposal failures that never reach the proposal stage are counted in this denominator, understating the rebuild ratio under validation/signature/maintenance error load.

**Suggestion:** Scope admission.command.duration to actual proposal attempts only (move the timer start to after pre-proposal validation), introduce a separate admission-attempt metric, or add an outcome label (e.g. outcome=pre_proposal_failure) so pre-proposal failures can be filtered from dashboard denominators. Update the dashboard query accordingly.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1537#issuecomment-4917130127

internal/adapter/http/handlers_bulk.go
```diff
@@ -15,7 +16,28 @@ import (
...
+// goroutine on a reused (HTTP/1 keep-alive) connection.
+//
+// The labeled callback context is threaded into serveBulk via r.WithContext so
```
NumaryBot, line 28: 🟡 **[minor] pprof labeled context not propagated to downstream request handling**

pprof.Do is used to scope the goroutine label, which correctly limits label lifetime to the callback. However, the labeled ctx provided by the pprof.Do callback is discarded; serveBulk receives the original request with the unlabeled context. Any downstream code using r.Context() or spawning goroutines from it will not inherit the component=admission.http label, defeating the intended context propagation.

**Suggestion:** Pass r.WithContext(ctx) to serveBulk inside the pprof.Do callback so that downstream code and any goroutines spawned from the request context inherit the goroutine label.

internal/application/admission/admission.go
gfyrag: Duplicate of the already-resolved denominator finding — the premise is false against current source (this looks like a re-scan of an older commit). The rebuild-ratio panel does NOT use `admission.command.duration_count`; it was changed in a50445079 to `admission.proposal_guard.duration_count` (admission.libsonnet:182), recorded once per guard acquisition, so pre-proposal failures cannot inflate it. `command.duration` intentionally spans the full lifecycle (required by the resolved phase-decomposition findings) and is not a dashboard denominator. Rejecting as a duplicate.

internal/adapter/http/handlers_bulk.go
gfyrag: Already fixed in 21d374f5f — this comment describes the pre-fix code. handleBulk now calls `s.serveBulk(w, r.WithContext(ctx))` inside the pprof.Do callback (handlers_bulk.go:34), so the labeled context IS propagated downstream. Rejecting as a stale duplicate.

misc/devenv/monitoring-dashboards/jsonnet/sections/admission.libsonnet
```diff
@@ -179,9 +179,57 @@ panels.row('Admission', 169, [
...
     ], unit='percentunit',
-    description='Rate of proposal guard rebuilds (boundary shifted) vs total proposals. A high ratio means the cache generation boundary is frequently shifting between optimistic preload and proposal, causing expensive re-builds under lock.',
+    description=|||
+      Rate of proposal guard rebuilds (boundary shifted) vs total guard acquisitions. A high ratio means the cache generation boundary is frequently shifting between optimistic preload and proposal, causing expensive re-builds under lock.
+
+      The denominator is admission.proposal_guard.duration_count — the number of times the proposal guard was actually acquired (one per proposal attempt that reached builder.Run). It is NOT admission.command.duration_count, which also counts commands rejected before proposal (bad signature, maintenance mode, validation/preload errors); using that would dilute the ratio with attempts the guard never saw.
+    |||,
+  ),
+
+  panels.timeseries(
```
NumaryBot, line 192: 🟠 **[major] New admission panels overlap existing panels in Grafana dashboard grid**

The new phase histogram panels reuse y-coordinates already occupied by existing panels (e.g. x:12, y:116, h:8 overlaps the existing command-size distribution at x:12, y:114, h:8, and subsequent panels at y:124/132 overlap further). The generated Grafana dashboards will render panels on top of each other or make some panels inaccessible, making the new observability views unreliable.

**Suggestion:** Shift existing panels down or place new panels in unoccupied grid space so no two panels share the same grid coordinates.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1537#issuecomment-4917130127

misc/devenv/monitoring-dashboards/jsonnet/sections/admission.libsonnet
flemzord: Rechecked the generated dashboard at `21d374f5f`: this is a false positive. `panels.row(...)` runs the section through `panels.reflow()`, whose `repackY` deliberately ignores authored `y` collisions and bin-packs panels onto a non-overlapping 24-column skyline; the helper comment explicitly says this makes dashboards immune to drift in hand-edited `y` values. In the checked-in generated `ledger-metrics-prom.json`, the relevant panels land without overlap: Command/Size at y=32, Guard/Resolve at y=40, Size panels at y=48, Orders/Scripts at y=56, and Response Resolution at y=72. No layout change is needed.

internal/application/admission/admission.go
flemzord: The new `recordPhaseOnExit` coverage fixes the conversion/script/FSM-wait failures, but rechecking `21d374f5f` shows one entered error path is still missing. `plan.Builder.Run` sets `ProposeStartTime` immediately before calling `proposer.Propose`, records `ProposeDuration`, then returns `(nil, err)` when that call fails (queue full/shutdown). `Admit` therefore returns from the `builder.Run` error branch before either `proposeDurationHistogram` or `proposalGuardDurationHistogram` is observed, while the deferred `command.duration` still records the attempt. Please preserve/return the timing result on proposer errors (or time that entered phase in `Admit`) and add a failing-proposer regression; the current tests cover pre-propose failures only.

gfyrag: Good find — you are right, that entered error path was still missed. Fixed in fc90e50df.

Root cause exactly as you describe: plan.Builder.Run sets ProposeStartTime, calls proposer.Propose, records ProposeDuration, then on a Propose failure (queue full / shutdown) returned (nil, err) — so Admit’s builder.Run error branch returned before observing proposeDurationHistogram/proposalGuardDurationHistogram, while the deferred command.duration still recorded the attempt. Net: on a failed-proposer command the propose (and guard) phase was entered but never recorded.

Fix (preserve + return the timing result on proposer error):
- plan.Builder.Run — both the slow path and the no-preload fast path — now populates LockHeldDuration/ProposeStartTime/ProposeDuration and returns the timing-only RunResult alongside the error when Propose fails (Proposal/FSMFuture nil, guard already released). Marshal / guard-acquisition failures (propose never started) still return (nil, err), keeping the "only record entered phases" discipline. Documented on the RunResult type.
- Admission’s builder.Run error branch records proposal_guard.duration + propose.duration when runResult != nil. The two mirror callers already return on err before touching runResult, so the new non-nil error-path result is safe for them.

Regression test: TestAdmitRecordsEnteredPhasesOnFailure gained a "proposer error records the entered propose + proposal_guard phases" subtest — wires a MockProposer (added a //go:generate mockgen on the admission Proposer interface, per repo convention) whose Propose returns ErrNoLeader, drives a CreateLedger through to builder.Run, and asserts propose + proposal_guard + command.duration each record exactly once while the past-propose phases (fsm_future.wait, response_resolution) record nothing. Verified it goes red without the fix. ADR 0002 note refined to draw the boundary at the Propose call. Resolving.

misc/devenv/monitoring-dashboards/jsonnet/sections/admission.libsonnet
gfyrag: Confirmed false positive, per @flemzord's analysis above: panels.row() runs the section through panels.reflow() → repackY, which deliberately ignores authored y collisions and bin-packs onto a non-overlapping 24-column skyline. The checked-in generated dashboards (regenerated at head via `just generate-dashboards`) render the new phase panels without overlap. No layout change needed. Resolving.

internal/application/admission/admission.go
```diff
@@ -369,8 +471,21 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
...
```
NumaryBot, line 481: 🟠 **[major] Pre-proposal failures pollute the admission.command.duration denominator**

The deferred `admission.command.duration` observation fires even when Admit returns early due to bad signatures, maintenance mode, request conversion errors, preload-build failures, or script-resolution errors — none of which constitute a proposal attempt. The dashboard divides Proposal Guard rebuilds by `admission.command.duration_count` as "total proposals", so high validation/signature/maintenance error rates dilute that denominator and systematically understate the rebuild ratio.

**Suggestion:** Scope `command.duration` to actual proposal attempts only (i.e. move the timer start to after all pre-proposal validation), or introduce a separate admission-attempt metric, or add an `outcome` label (e.g. `outcome=pre_proposal_failure`) so pre-proposal failures can be filtered from dashboard denominators.

internal/adapter/http/handlers_bulk.go
```diff
@@ -15,7 +16,28 @@ import (
 )
 
 // handleBulk handles POST /{ledgerName}/bulk to create multiple transactions/operations.
+//
+// The handler body runs under `pprof.Do` with the `component=admission.http`
+// label so Pyroscope (CPU / block/delay / mutex profiles) can attribute its
+// cost separately from the FSM pipeline (applier.main, applier.decoder,
+// applier.committer). The label propagates to any child goroutines spawned
```
NumaryBot, line 23: 🟡 **[minor] pprof.Do labeled context not propagated to downstream handler**

`pprof.Do` is called and correctly scopes the goroutine label, but the labeled `ctx` from the callback is discarded. `serveBulk` receives the original request rather than `r.WithContext(ctx)`, so any goroutines spawned inside `serveBulk` using `r.Context()` will not inherit the pprof labels despite the intent of using `pprof.Do`.

**Suggestion:** Pass the labeled context to `serveBulk` by replacing the request argument with `r.WithContext(ctx)`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1537#issuecomment-4917130127

internal/application/admission/admission.go
flemzord: This is a stale duplicate of the already-resolved denominator finding. At current head `fc90e50df`, the Proposal Guard Rebuild Rate & Ratio query divides by `admission.proposal_guard.duration_count`, not `admission.command.duration_count` (`misc/devenv/monitoring-dashboards/jsonnet/sections/admission.libsonnet`, panel definition around line 182). Its description explicitly documents that this counts only guard acquisitions and deliberately excludes pre-proposal failures. `command.duration` remains lifecycle-wide so the phase histograms can decompose it; it is no longer used as this proposal-attempt denominator.

internal/adapter/http/handlers_bulk.go
flemzord: This is also stale against current head `fc90e50df`. `handleBulk` already passes the labeled callback context downstream: `pprof.Do(..., func(ctx context.Context) { s.serveBulk(w, r.WithContext(ctx)) })` in `internal/adapter/http/handlers_bulk.go`. Therefore downstream code reading `r.Context()` and goroutines derived from it inherit the label. This was fixed in `21d374f5f` and is unchanged at the current head.

flemzord (approved):
Re-reviewed at fc90e50df. The proposer-error timing gap is fixed on both runner paths and covered by a regression test; the remaining NumaryBot comments are stale duplicates contradicted by the current source. The Default workflow is green.

gfyrag:
Re-reviewed the current live head while this PR was moving; the live head during review was `fc90e50df`.

Only one minor cleanup remains from my pass: `Command Duration` dashboard wording is still stale.

`misc/devenv/monitoring-dashboards/jsonnet/sections/admission.libsonnet` still describes `Command Duration` as “from Apply call to future resolution,” but `admission.command.duration` now records through function return, including the post-FSM `response_resolution` phase. That conflicts with the new Response Resolution panel saying it is the final phase of `command.duration`. Please update the source description, and regenerated dashboards, to say it runs through response/log resolution and includes resolve batch, order prep, scripts, preload, propose, FSM wait, and response resolution.

What looks resolved:

- Normalized Prometheus metadata; generated normalized dashboards query `_microseconds_bucket`.
- Proposal-guard denominator; dashboard uses `admission.proposal_guard.duration_count`.
- Pprof label scoping/context propagation with `r.WithContext(ctx)`.
- Generated dashboards are clean after `just generate-dashboards`.
- Proposer-error phase recording has a regression test.

Checks run:

- `go test -race ./internal/application/admission/...`
- `go test -race ./internal/adapter/http/...`
- `go test ./internal/infra/monitoring/metrics/...`
- `go generate ./internal/application/admission`
- `just generate-dashboards`
- `GOROOT= go build ./...`
- `direnv allow && eval "$(direnv export bash)" && GOROOT= just pre-commit`

The preferred `nix develop --command bash -c "just pre-commit"` failed due a Nix Go tool mismatch (`go1.26.3` vs `go1.26.4`); the documented `direnv` fallback passed.

PR #1539, opened 2026-07-09
feat(operator): declarative account/transaction indexes on mirror ledgers

sylr:
## What

Adds a declarative `spec.indexes` field to the operator `Ledger` CRD so account/transaction indexes can be defined on a ledger — chiefly a **read-only mirror**, which is provisioned entirely through this CRD and had no other declarative surface for indexes. The operator reconciles the ledger's index set to match the spec.

**Operator-side only** — no proto, FSM, admission, or checker changes. The FSM already accepts `CreateIndex`/`DropIndex` on mirror ledgers (`isMirrorSafeApply`); this rides on the existing `ledgerctl indexes` / `ledgers set-metadata-type` CLI over the existing pod-exec path.

```yaml
kind: Ledger
spec:
  name: my-ledger
  serviceRef: my-service
  mode: mirror
  mirrorSource: { ... }
  indexes:
    transaction: [reference, address, sourceAddress]
    account: [asset]
    metadata:
      - target: account
        key: category
        type: string
```

## Design

- **Ownership-scoped drops.** The operator records indexes it creates in `status.appliedIndexes` and only ever drops those — externally-created and CRD-unrepresentable indexes (`id` tx-builtin, log builtins, ledger-target metadata, bucket-scoped audit indexes) are never touched. A pre-existing index that happens to match the spec is **not** adopted, so it is never dropped.
- **`nil` = unmanaged** (operator never lists/creates/drops); present-but-empty `{}` = managed-with-zero (drops only what the operator created).
- **Mutable**: `spec.indexes` is excluded from the immutability spec-hash, so edits reconcile instead of raising `SpecDrifted`. Convergence is reported via the `IndexesSynced` condition.
- **Metadata schema reconciled**: a metadata index needs its field declared first, so the operator issues `ledgers set-metadata-type` before creating — and re-issues it only when the field is absent or its declared type differs.
- **Cross-module boundary**: the operator is a separate Go module and can't import `internal/proto`, so index identity is parsed from `ledgerctl ... --json` into local structs. Safe because the CLI routes pterm to stderr under `--json`, leaving stdout as clean JSON.

## Note on `IndexesSynced`

`IndexesSynced=True` means the declared set is present in the **registry**, not that backfill finished (readiness is the per-replica `current_version`, EN-1323). Do not gate query readiness on this condition; the condition message says as much.

## Testing

- Unit tests for the CRD→CLI mapping, `--json` parsing (incl. skipping unrepresentable kinds), the ownership-scoped diff, ownership tracking (never adopt pre-existing), metadata-schema reconcile, and spec-hash-ignores-indexes.
- `go build`, `go vet`, `golangci-lint` (0 issues) all green.
- Chainsaw e2e extended: create → indexes present; edit CR (drop `address`, add `timestamp`) → reconcile converges.
- **Not run**: the chainsaw e2e itself (needs a live KinD cluster).

## Review

Plan and implementation were both reviewed by an external model (GPT-5.5); the one correctness bug it found (pre-existing indexes being adopted and later droppable) is fixed in this branch with dedicated tests.


gfyrag (approved):

misc/operator/internal/controller/ledger_index_reconcile.go
```diff
@@ -0,0 +1,168 @@
...
+
+			log.Info("declared metadata field for index",
+				"ledger", ledgerName, "target", mi.target, "key", mi.key, "type", mi.mdType)
+			changed = true
+		}
+	}
+
+	diff := diffIndexes(desired, actual, ledger.Status.AppliedIndexes)
+
+	for _, mi := range diff.toCreate {
```
NumaryBot, line 155: 🟠 **[major] Record ownership after partial reconcile failures**

If one create succeeds and a later create/drop command fails, this early return exits before `ledger.Status.AppliedIndexes = nextAppliedIndexes(...)` runs. On the next reconcile the successfully-created index is present in `actual` but absent from `applied`, so it is treated as pre-existing and will not be dropped if it is later removed from `spec.indexes`; this can happen with multiple declared indexes and a transient CLI/Raft failure after the first success.

misc/operator/internal/controller/ledger_indexes.go
```diff
@@ -0,0 +1,414 @@
...
+			diff.toCreate = append(diff.toCreate, mi)
+		}
+	}
+
+	for _, canonical := range applied {
+		if _, stillDesired := desiredByCanonical[canonical]; stillDesired {
+			continue
+		}
+
+		if !actual[canonical] {
```
NumaryBot: 🟠 **[major] Forget owned indexes once they are already absent**

When a previously applied index is no longer desired but has already been removed out-of-band, this branch skips both dropping and removing it from `status.appliedIndexes`. The stale ownership entry remains after a synced reconcile, so if someone later creates the same index externally, the operator will see it as applied and drop it even though the live index was externally-created.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1539#issuecomment-4922397151

misc/operator/internal/controller/ledger_index_reconcile.go
```diff
@@ -0,0 +1,168 @@
...
```
NumaryBot, line 155: 🟠 **[major] Ownership status not updated after partial reconcile failure**

If one index create/drop succeeds but a subsequent operation fails, the early return exits before `ledger.Status.AppliedIndexes = nextAppliedIndexes(...)` runs. On the next reconcile the successfully-created index is present in `actual` but absent from `applied`, so it will not be treated as operator-owned and will not be dropped if later removed from `spec.indexes`. With multiple declared indexes and a transient CLI/Raft failure, this silently leaks ownership state.

**Suggestion:** Persist the partial `nextAppliedIndexes` result (or the subset of changes that succeeded) to `ledger.Status.AppliedIndexes` before returning the error, so each reconcile loop starts with an accurate ownership snapshot.

misc/operator/internal/controller/ledger_indexes.go
NumaryBot: 🟠 **[major] Stale ownership entry retained when index is already absent out-of-band**

When a previously-applied index is no longer desired but has already been removed externally, the branch skips both the drop command and the removal of the entry from `status.appliedIndexes`. The stale ownership entry persists across reconciles, so if someone later recreates the same index externally, the operator incorrectly treats it as operator-owned and drops it.

**Suggestion:** Even when the index is already absent, remove it from `status.appliedIndexes` so ownership tracking stays clean. A simple guard `if indexExists { drop() }; removeFromApplied()` ensures the entry is always cleared.

misc/operator/internal/controller/ledger_crd_controller.go
```diff
@@ -193,7 +195,21 @@ func (r *LedgerReconciler) reconcileReady(ctx context.Context, ledger *ledgerv1a
...
 
-	return ctrl.Result{}, nil
+	// Reconcile declarative indexes (mutable, independent of ledger immutability).
+	if ledger.Spec.Indexes == nil {
+		meta.RemoveStatusCondition(&ledger.Status.Conditions, conditionIndexesSynced)
+
+		return ctrl.Result{}, nil
+	}
+
+	grpcPort, err := r.resolveEndpoint(ctx, ledger)
```
NumaryBot, line 212: 🟠 **[major] Index reconciliation proceeds against wrong target when immutable spec fields drift**

When an immutable field (e.g. `spec.name` or `spec.serviceRef`) has drifted, the controller still resolves the endpoint and runs index reconciliation using the new (mutated) values. This means it can create or drop indexes on a different ledger or cluster than originally intended, even while reporting `SpecDrifted`. The drift guard should be a hard exit for any reconciliation that mutates external state.

**Suggestion:** After detecting non-index immutable spec drift, return early (or set `Requeue: false`) before reaching the index reconciliation step. Index reconciliation should only run when the non-index spec fields are confirmed clean.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1539#issuecomment-4922397151

sylr:
Thanks @NumaryBot — all three findings were valid and are fixed in 832410a68:

- **Ownership leak on partial reconcile failure** — `reconcileIndexes` now accumulates `createdOK`/`droppedOK` and writes `status.appliedIndexes` from them in a `defer`, so an index created before a later command fails is still recorded as operator-owned (and therefore droppable later).
- **Stale ownership when an index is already absent out-of-band** — `diffIndexes` now includes owned-but-no-longer-desired indexes in `toDrop` even when absent; the reconciler issues a drop command only when the index is still present but relinquishes ownership in both cases, so a later external recreate is no longer mistaken as ours.
- **Index reconcile against a drifted (wrong) target** — `reconcileReady` now hard-returns after setting `SpecDrifted` (the hash excludes indexes, so a mismatch means an immutable field like `name`/`serviceRef` changed), instead of resolving the endpoint and mutating indexes on a possibly-different ledger/cluster.

Added unit tests for the already-gone ownership-cleanup and partial-progress cases. Build, tests, and lint are green.

misc/operator/internal/controller/ledger_index_reconcile.go
```diff
@@ -0,0 +1,186 @@
...
```
NumaryBot, line 155: 🟠 **[major] Ownership status not updated after partial reconcile failure**

If one index create/drop succeeds but a subsequent operation fails, the early return exits before `ledger.Status.AppliedIndexes = nextAppliedIndexes(...)` runs. On the next reconcile the successfully-created index is present in `actual` but absent from `applied`, so it will not be treated as operator-owned and will not be dropped if later removed from `spec.indexes`. With multiple declared indexes and a transient CLI/Raft failure, this silently leaks ownership state.

**Suggestion:** Persist the partial `nextAppliedIndexes` result (or the subset of changes that succeeded) to `ledger.Status.AppliedIndexes` before returning the error, so each reconcile loop starts with an accurate ownership snapshot.

misc/operator/internal/controller/ledger_indexes.go
```diff
@@ -0,0 +1,415 @@
...
+	var diff indexDiff
+
+	for _, mi := range desired {
+		desiredByCanonical[mi.canonical] = struct{}{}
+		if !actual[mi.canonical] {
+			diff.toCreate = append(diff.toCreate, mi)
+		}
+	}
+
+	for _, canonical := range applied {
```
NumaryBot, line 317: 🟠 **[major] Stale ownership entry retained when index is already absent out-of-band**

When a previously-applied index is no longer desired but has already been removed externally, the branch skips both the drop command and the removal of the entry from `status.appliedIndexes`. The stale ownership entry persists across reconciles, so if someone later recreates the same index externally, the operator incorrectly treats it as operator-owned and drops it.

**Suggestion:** Even when the index is already absent, remove it from `status.appliedIndexes` so ownership tracking stays clean. A simple guard `if indexExists { drop() }; removeFromApplied()` ensures the entry is always cleared.

misc/operator/internal/controller/ledger_crd_controller.go
```diff
@@ -189,11 +192,31 @@ func (r *LedgerReconciler) reconcileReady(ctx context.Context, ledger *ledgerv1a
...
+	meta.RemoveStatusCondition(&ledger.Status.Conditions, conditionSpecDrifted)
+
+	// Reconcile declarative indexes (mutable, independent of ledger immutability).
+	if ledger.Spec.Indexes == nil {
+		meta.RemoveStatusCondition(&ledger.Status.Conditions, conditionIndexesSynced)
+
+		return ctrl.Result{}, nil
+	}
+
+	grpcPort, err := r.resolveEndpoint(ctx, ledger)
```
NumaryBot, line 212: 🟠 **[major] Index reconciliation proceeds against wrong target when immutable spec fields drift**

When an immutable field (e.g. `spec.name` or `spec.serviceRef`) has drifted, the controller still resolves the endpoint and runs index reconciliation using the new (mutated) values. This means it can create or drop indexes on a different ledger or cluster than originally intended, even while reporting `SpecDrifted`. The drift guard should be a hard exit for any reconciliation that mutates external state.

**Suggestion:** After detecting non-index immutable spec drift, return early (or set `Requeue: false`) before reaching the index reconciliation step. Index reconciliation should only run when the non-index spec fields are confirmed clean.

NumaryBot, line 212: 🟠 **[major] Stale IndexesSynced=True condition retained when endpoint resolution fails**

When `spec.indexes` is edited after a prior successful sync but the Cluster cannot be resolved, this branch only updates `status.message` and leaves the old `IndexesSynced=True` condition in place. Consumers that check condition status (e.g. `kubectl wait` or Chainsaw assertions) can report the new desired index set as synced even though no list/create/drop reconciliation ran. The current generation's `IndexesSynced` condition should be set to `False` in this path.

**Suggestion:** Set `IndexesSynced=False` (with an appropriate reason/message) before returning when endpoint resolution fails, so status accurately reflects that index reconciliation did not run for the current generation.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1539#issuecomment-4922397151

sylr:
Re @NumaryBot's re-review of the fix commit — one finding was valid and is fixed in 9441cbd8c; the other three are already addressed by the previous fix (832410a68) and I believe are false positives from the static pass not modelling the `defer`/early-return control flow:

**Fixed (9441cbd8c):**
- **Stale `IndexesSynced=True` when endpoint resolution fails** — `reconcileReady` now sets `IndexesSynced=False` (reason `EndpointUnavailable`) before returning when the Cluster can't be resolved, so a prior `True` isn't left stale for the new generation. Valid catch, thanks.

**Already addressed (please re-check):**
- **Ownership not updated after partial reconcile failure** (`ledger_index_reconcile.go`) — `status.appliedIndexes` is now written from `createdOK`/`droppedOK` in a `defer` (added in 832410a68), so it runs on *every* return path including the `return false, createErr` inside the create loop. A create that succeeds before a later failure is recorded. This is exactly the suggested "persist the partial result before returning."
- **Stale ownership when an index is already absent** (`ledger_indexes.go`) — `diffIndexes` no longer skips absent indexes; owned-but-no-longer-desired entries always go into `toDrop`, and the reconciler relinquishes ownership (appends to `droppedOK`) whether or not a drop command is issued.
- **Index reconcile against a drifted target** (`ledger_crd_controller.go`) — `reconcileReady` hard-returns immediately after setting `SpecDrifted` (the line the finding points at is now unreachable on drift; the early `return ctrl.Result{}, nil` sits above it).

All CI jobs are green on the latest commit; build/tests/lint pass locally.

misc/operator/internal/controller/ledger_crd_controller.go
```diff
@@ -189,11 +192,42 @@ func (r *LedgerReconciler) reconcileReady(ctx context.Context, ledger *ledgerv1a
...
 		})
 		log.Info("spec drift detected on immutable ledger", "name", ledger.Spec.Name)
-	} else {
-		meta.RemoveStatusCondition(&ledger.Status.Conditions, conditionSpecDrifted)
+
+		// Do NOT run index reconciliation while an immutable field has drifted:
+		// spec.name / spec.serviceRef may now point at a different ledger or
+		// cluster, so creating/dropping indexes here would mutate the wrong
+		// target. Hold until the drift is resolved by delete + recreate.
+		return ctrl.Result{}, nil
```
NumaryBot, line 200: 🟠 **[major] Stale IndexesSynced=True condition when drift or endpoint failure blocks reconciliation**

When an immutable spec field drifts (or when endpoint resolution fails) and the controller returns early, any previously set `IndexesSynced=True` condition is left in place for the new generation. Consumers using `kubectl wait --for=condition=IndexesSynced` or Chainsaw assertions can observe a stale success for a desired index set that was never reconciled in the current generation.

**Suggestion:** Before returning early on spec drift or endpoint-resolution failure, explicitly set `IndexesSynced=False` (with an appropriate reason/message) so the condition accurately reflects that index reconciliation did not run for the current spec generation.

misc/operator/internal/controller/ledger_index_reconcile.go
```diff
@@ -0,0 +1,186 @@
...
+		if _, createErr := exec(mi.createArgs(ledgerName)...); createErr != nil && !isAlreadyExists(createErr) {
```
NumaryBot, line 156: 🟡 **[minor] Race: externally-created index can be adopted into appliedIndexes if created between list and create calls**

If another actor creates the desired index after the list call but before the operator's create call runs, `ledgerctl` may return an already-exists error that this branch treats as success and still appends the index to `createdOK`. This records an externally-created index in `status.appliedIndexes`, meaning a later removal from `spec.indexes` can drop an index the operator did not actually create.

**Suggestion:** Distinguish an already-exists error from a clean create success. If the create returns an already-exists response, do not add the index to `createdOK` (or do not record it in `status.appliedIndexes`), so ownership is only claimed for indexes the operator provably created.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1539#issuecomment-4922397151

PR #1540, opened 2026-07-09
chore(deps): upgrade go.etcd.io/raft to v3.7.0

gfyrag:
## Summary
- Bump `go.etcd.io/raft/v3` **v3.6.0 → v3.7.0** and the etcd server/api/client stack to v3.7.0 (bbolt v1.5.0 transitively).
- Notable fix in raft v3.7.0: [PR #397](https://github.com/etcd-io/raft/pull/397) prevents `ReadIndex` from returning a stale index when the same context is retried/deduped in the leader's readonly queue — a linearizability hazard on the query path.
- Migrate every `raftpb` usage from gogo/protobuf helpers to `google.golang.org/protobuf`: scalar reads through nil-safe `GetX()`, construction with `new(x)`/pointer helpers, `proto.Marshal`/`proto.Unmarshal`/`proto.Size` in place of the gogo methods.
- Absorb the pointer semantics forced by v3.7's proto3-generated types (each embeds `sync.Mutex` via `protoimpl.MessageState`): the WAL / Spool / Transport interfaces now take `*raftpb.Entry` / `*Message` / `*ConfState` and pointer slices; `ConfChange.NodeID → NodeId`; `AsV2()` now returns `*ConfChangeV2`.
- Add `internal/pkg/raftutil` (`EntryType` / `ConfChangeType` / `MessageType`) to keep enum-pointer construction terse (DRY).
- Regenerate mocks for the updated interfaces.

## Notes
- **On-disk / on-wire compat with prior versions is intentionally not exercised** — the wire format for `walpb.Snapshot` / raftpb messages should still be proto3, but the gogo→google migration is exactly the kind of change where subtle diffs can bite. Consider a cross-version WAL roundtrip if this ships to environments with existing state.
- `just pre-commit` (nix env) fails locally on `generate-proto` because of a pinned go1.26.3 vs 1.26.4 mismatch — unrelated to this PR; the underlying steps (`go generate`, `go mod tidy`, `golangci-lint`) all pass individually.

## Test plan
- [x] `GOROOT= go vet ./...`
- [x] `GOROOT= go test ./...` (default tags — all 74 packages green)
- [x] `GOROOT= just lint` (0 issues, root + operator)
- [ ] E2E / scenarios / antithesis (deferred to CI)


internal/pkg/raftutil/pointers.go
```diff
@@ -0,0 +1,23 @@
...
+// all scalar enum/scalar fields are now pointers; these helpers cut down on
+// boilerplate at struct-literal construction sites.
+package raftutil
+
+import "go.etcd.io/raft/v3/raftpb"
+
+// EntryType returns a pointer to t for use in raftpb.Entry struct literals.
+//
+//go:fix inline
+func EntryType(t raftpb.EntryType) *raftpb.EntryType { return new(t) }
```
NumaryBot: 🔴 **[blocker] Invalid use of new() with values instead of types causes build failure**

The patch uses `new(t)` where `t` is a value (e.g., `new(index)`, `new(nodeID)`), but Go's built-in `new` only accepts a type, not a value expression. This is a compile-time error and will prevent the repository from building.

**Suggestion:** Replace `new(someValue)` with address-of on a local variable or a typed pointer helper. For example:
```go
v := index
return &v
```
or define small inline helpers such as `func ptrIndex(v index) *index { return &v }`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1540#issuecomment-4923606323

gfyrag (commented):
Reviewed the detected movement `b720ce775217571586241cf6cd84b1f4c8e94cdc` -> `86a469bb119a8711c3aebf04530c23518269ecb5`.

This new commit is empty (`git diff b720ce7..86a469b` has no file changes), so there is no new code surface to review for this poll. I did not find anything actionable to comment on for this movement.

internal/pkg/raftutil/pointers.go
NumaryBot: 🔴 **[blocker] Invalid use of new() with value expressions causes build failure**

The code calls new() with value expressions (e.g., new(index), new(nodeID)) rather than type names. Go's built-in new only accepts a type as its argument, not a value expression. This is a compile-time error that will prevent the repository from building entirely.

**Suggestion:** Replace new(someValue) with address-of on a local variable, e.g.:
```go
v := index
return &v
```
Or define small inline pointer-helper functions such as:
```go
func ptrIndex(v index) *index { return &v }
```

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1540#issuecomment-4923606323

internal/pkg/raftutil/pointers.go
NumaryBot: 🔴 **[blocker] Invalid use of new() with value expressions causes build failure**

The code calls new() with value expressions (e.g., new(index), new(nodeID)) rather than type names. Go's built-in new() only accepts a type as its argument, not a value expression. This is a compile-time error that will prevent the repository from building entirely.

**Suggestion:** Replace new(someValue) with address-of on a local variable, e.g.:
```go
v := index
return &v
```
Or define small inline pointer-helper functions such as:
```go
func ptrIndex(v index) *index { return &v }
```

internal/infra/node/node.go
```diff
@@ -1146,13 +1147,14 @@ func (node *Node) processReady(ctx context.Context, stop chan struct{}, rd raft.
 	// future with the proposer's view of the current term (issue #172). Use a
 	// monotonic CAS — terms only grow.
 	if !raft.IsEmptyHardState(rd.HardState) {
+		rdTerm := rd.GetTerm()
```
NumaryBot, line 1150: 🔴 **[blocker] GetTerm() called on raft.Ready instead of HardState — compile error**

raft.Ready is not a protobuf message and does not expose a GetTerm() method. Calling rd.GetTerm() (or equivalent) will fail to compile. The term should be read from the HardState embedded in Ready, which does provide that accessor.

**Suggestion:** Replace the call with rd.HardState.GetTerm() to correctly access the term from the HardState field of raft.Ready.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1540#issuecomment-4923606323

internal/pkg/raftutil/pointers.go
NumaryBot: 🔴 **[blocker] Invalid use of new() with value expressions causes compile failure**

The code calls Go's built-in new() with value expressions (e.g., new(index), new(nodeID)) rather than with type names. Go's new() only accepts a type as its argument, not a value expression. This is a compile-time error that will prevent the entire repository from building.

**Suggestion:** Replace new(someValue) with an address-of on a local variable, e.g.:
```go
v := index
return &v
```
Or define small inline pointer-helper functions:
```go
func ptrIndex(v index) *index { return &v }
```

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1540#issuecomment-4923606323

gfyrag (commented):
Reviewed the force-pushed raft v3.7.0 migration at `38eaec7f7b424eac827f115d1e653efac1c845d6` against `release/v3.0`.

I focused on the google protobuf pointer migration in the Raft/WAL/spool path: conf-change decode/apply, async storage responses, snapshot `ConfState` propagation, WAL snapshot records, replay, and the FSM conf-change handler. I did not find an actionable correctness issue.

Validation run in `/tmp/ledger-pr1540`:

```bash
GOCACHE=/tmp/go-build-pr1540 go test ./internal/infra/node/... ./internal/infra/membership/... ./internal/storage/wal/... ./internal/storage/spool/... ./internal/infra/state/...
```

All targeted packages passed.

internal/pkg/raftutil/pointers.go
NumaryBot: 🔴 **[blocker] Invalid use of new() with value expressions causes build failure**

The code calls Go's built-in new() with value expressions (e.g., new(index), new(nodeID)) rather than type names. Go's new() only accepts a type as its argument, not a value expression. This is a compile-time error that prevents the repository from building entirely.

**Suggestion:** Replace new(someValue) with an address-of on a local variable, e.g.:
```go
v := index
return &v
```
Or define small inline pointer-helper functions:
```go
func ptrIndex(v index) *index { return &v }
```

internal/infra/node/node.go
```diff
@@ -1146,13 +1148,14 @@ func (node *Node) processReady(ctx context.Context, stop chan struct{}, rd raft.
...
```
NumaryBot, line 1150: 🔴 **[blocker] GetTerm() called on raft.Ready instead of HardState — compile error**

raft.Ready is a plain struct and does not expose protobuf getter methods such as GetTerm(). Calling rd.GetTerm() will fail to compile. The term must be read from the HardState field embedded in Ready, which does provide that accessor.

**Suggestion:** Replace the call with rd.HardState.GetTerm() to correctly access the term from the HardState field of raft.Ready.

internal/infra/node/node.go
```diff
@@ -2319,19 +2321,19 @@ func (node *Node) AddLearner(ctx context.Context, nodeID uint64, raftAddr, servi
...
 			}
 
-			return node.rawNode.ProposeConfChange(raftpb.ConfChangeV2{
-				Changes: []raftpb.ConfChangeSingle{{
-					Type:   raftpb.ConfChangeUpdateNode,
-					NodeID: nodeID,
+			return node.rawNode.ProposeConfChange(&raftpb.ConfChangeV2{
+				Changes: []*raftpb.ConfChangeSingle{{
+					Type:   raftutil.ConfChangeType(raftpb.ConfChangeUpdateNode),
+					NodeId: new(nodeID),
```
NumaryBot, line 2326: 🔴 **[blocker] new() called with a value expression (nodeID) in conf-change path — compile error**

The conf-change production path uses new(nodeID) where nodeID is a runtime value, not a type name. Go's built-in new() only accepts a type argument; passing a value expression is a compile-time error that prevents the package from building.

**Suggestion:** Use a pointer helper or take the address of a local copy:
```go
nid := nodeID
return &nid
```

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1540#issuecomment-4923606323

gfyrag (commented):
Reviewed the new movement `38eaec7f7b424eac827f115d1e653efac1c845d6` -> `137f3772bfa05f64d73a1a6f261f4e39a65aa91d`.

Scope checked: the nil-safe `rd.HardState.GetTerm()` read in `processReady`, the e2e cluster message getters, and the Antithesis workload module tidy after the raft v3.7 migration. I did not find an actionable issue.

Validation:

```bash
GOCACHE=/tmp/go-build-pr1540 go test ./internal/infra/node/...
GOCACHE=/tmp/go-build-pr1540 go test -tags e2e ./tests/e2e/cluster/... -run TestNonExistent -count=1
```

Both passed; the e2e command compiled the package without running Ginkgo specs.

internal/pkg/raftutil/pointers.go
```diff
@@ -0,0 +1,17 @@
+// Package raftutil provides small pointer helpers for constructing raftpb
+// protobuf types. Since etcd/raft v3.7 migrated to google.golang.org/protobuf,
+// all scalar enum/scalar fields are now pointers; these helpers cut down on
+// boilerplate at struct-literal construction sites.
+package raftutil
+
+import "go.etcd.io/raft/v3/raftpb"
+
+// EntryType returns a pointer to t for use in raftpb.Entry struct literals.
+func EntryType(t raftpb.EntryType) *raftpb.EntryType { return new(t) }
```
NumaryBot: 🔴 **[blocker] new() called with value expressions instead of type names — compile error**

Go's built-in `new()` only accepts a type name as its argument, not a runtime value expression. The code calls `new(index)`, `new(nodeID)`, etc., where these are values, not types. This is a compile-time error that prevents the entire repository from building.

**Suggestion:** Replace `new(someValue)` with an address-of on a local variable, e.g.:
```go
v := index
return &v
```
or define small inline pointer-helper functions:
```go
func ptrIndex(v index) *index { return &v }
```

internal/infra/node/node.go
NumaryBot, line 1150: 🔴 **[blocker] GetTerm() called on raft.Ready instead of HardState — compile error**

`raft.Ready` is a plain Go struct and does not expose protobuf getter methods such as `GetTerm()`. Calling `rd.GetTerm()` will fail to compile. The term must be read from the `HardState` field embedded in `Ready`, which does provide that accessor.

**Suggestion:** Replace `rd.GetTerm()` with `rd.HardState.GetTerm()` to correctly access the term from the HardState field of `raft.Ready`.

internal/infra/node/node.go
NumaryBot, line 2326: 🔴 **[blocker] new() called with a value expression (nodeID) in conf-change path — compile error**

The conf-change production path uses `new(nodeID)` where `nodeID` is a runtime value, not a type name. Go's built-in `new()` only accepts a type argument; passing a value expression is a compile-time error that prevents the package from building.

**Suggestion:** Use a pointer helper or take the address of a local copy:
```go
nid := nodeID
return &nid
```

internal/storage/wal/wal_default.go
```diff
@@ -161,17 +164,17 @@ func New(dataDir string, logger logging.Logger, meter metric.Meter, opts ...Opti
...
+					"index":    s.snapshot.GetMetadata().GetIndex(),
+					"term":     s.snapshot.GetMetadata().GetTerm(),
 					"duration": time.Since(loadStart).String(),
 				}).
 				Infof("Loaded snapshot from disk")
-			snap = walpb.Snapshot{
-				Index: s.snapshot.Metadata.Index,
-				Term:  s.snapshot.Metadata.Term,
+			snap = &walpb.Snapshot{
+				Index: new(s.snapshot.GetMetadata().GetIndex()),
```
NumaryBot, line 176: 🔴 **[blocker] new() called with getter/value expressions in WAL snapshot path — compile error**

`new` is called with expressions such as `s.snapshot.GetMetadata().GetIndex()` and similar getter/variable expressions added in this file. Since `new` only accepts a type, not a value expression, the WAL package fails to compile.

**Suggestion:** Use a local variable copy or a pointer helper such as `proto.Uint64(...)` instead of passing value expressions to `new()`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1540#issuecomment-4923606323

gfyrag (commented):
Reviewed the movement from `137f3772bfa05f64d73a1a6f261f4e39a65aa91d` to `75a05b2e1eda4afb1b4bce801a981edc5be2d38a`.

No findings. The diff only removes stale `go:fix inline` directives from the remaining raft pointer helpers; the helpers still compile and the node package remains green.

Validated with:

`GOCACHE=/tmp/go-build-pr1540 go test ./internal/infra/node/... ./internal/pkg/raftutil/...`

gfyrag (commented):
Reviewed the movement from `75a05b2e1eda4afb1b4bce801a981edc5be2d38a` to `fa070736e4812ba497b89fbc9e187698adc8f7f5`.

No findings. This removes the local `raftutil` pointer helpers and uses direct enum pointer literals at the remaining call sites; the touched raft/node/state paths still compile and pass targeted tests.

Validated with:

`GOCACHE=/tmp/go-build-pr1540 go test ./internal/infra/node/... ./internal/infra/membership/... ./internal/infra/state/...`

NumaryBot (commented):

Resolved 4 stale NumaryBot review threads (3 fixed, 1 outdated).

Summary: https://github.com/formancehq/ledger/pull/1540#issuecomment-4923606323

gfyrag (commented):
Reviewed the movement from `fa070736e4812ba497b89fbc9e187698adc8f7f5` to `f2e77414528e7af2cf8d977b6f3a48e2fa46d578`.

No findings. This is a rebase onto the updated `release/v3.0` base, which now includes #1533/#1542/#1546. `range-diff` shows commits 2-4 are unchanged; commit 1 was adapted to the existing `applyDecodedEntriesToFSM` / decoder-prefetch shape. I checked the `MsgStorageApplyResp` handoff through the decoded applier path and did not find an ack-before-durable-apply issue.

Validated with:

`GOCACHE=/tmp/go-build-pr1540 go test ./internal/infra/node/... ./internal/infra/membership/... ./internal/infra/state/... ./internal/storage/wal/... ./internal/storage/spool/...`

gfyrag (commented):
Reviewed the new commit (`f2e774145..e2c8cd06`). No findings.

Scope checked: the raft 3.7 follow-up switches remaining direct proto field reads to generated getters (`ConfChange.Context`, `Ready.HardState.Term`, `ConfChangeV2.Changes`) and removes now-unused imports. The membership context invariant stays loud on multi-change context payloads.

Validation: `GOCACHE=/tmp/go-build-pr1540 go test ./internal/infra/node/... ./internal/infra/membership/...` passed.

internal/storage/wal/wal_default.go
```diff
@@ -346,26 +352,26 @@ func New(dataDir string, logger logging.Logger, meter metric.Meter, opts ...Opti
...
 }
 
 // InitialState returns the saved HardState and ConfState information.
-func (s *DefaultWAL) InitialState() (raftpb.HardState, raftpb.ConfState, error) {
+func (s *DefaultWAL) InitialState() (*raftpb.HardState, *raftpb.ConfState, error) {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 
-	return s.hardState, s.snapshot.Metadata.ConfState, nil
+	return s.hardState, s.snapshot.GetMetadata().GetConfState(), nil
```
flemzord: **[P2] Return a non-nil ConfState for an empty WAL**

This can panic when a freshly-created `DefaultWAL` is used as `raft.Storage` before its first snapshot. Raft v3.7 explicitly requires `InitialState` to return a non-nil empty `ConfState`; `s.snapshot` starts as `&raftpb.Snapshot{}`, so this expression returns nil and `raft.NewRawNode` dereferences it in `confchange.Restore`. Please wrap it with `raftpb.EnsureConfState(...)` and tighten `TestInitialState_EmptyWAL` to assert non-nil.

gfyrag:
Review pass on `e5be4e64d231c171057b260d526c6b1fcfb3ec9c`.

The latest `InitialState` review thread is valid. `DefaultWAL.InitialState()` can return a nil `*raftpb.ConfState` for a freshly-created WAL because `s.snapshot` starts as an empty `&raftpb.Snapshot{}` with no metadata/conf state. Raft v3.7's `Storage.InitialState()` contract expects a non-nil empty `ConfState`; `MemoryStorage` handles this with `raftpb.EnsureConfState(cs)`.

Please wrap the returned conf state with `raftpb.EnsureConfState(...)` and update `TestInitialState_EmptyWAL` to assert the returned `ConfState` is non-nil.

Older compile-error comments about `new(value)` and `rd.GetTerm()` look resolved/stale on this head.

Validation run: `go test ./internal/storage/wal -run TestInitialState_EmptyWAL -count=1` and `go test ./internal/storage/wal -count=1` passed.

internal/storage/wal/wal_default.go
gfyrag: Fixed in 81f571a7b. Wrapped the returned ConfState with `raftpb.EnsureConfState(...)` in `InitialState` (mirroring what raft's own `MemoryStorage.InitialState` does at storage.go:132), so an empty WAL now yields a non-nil empty ConfState instead of nil.

Verified the panic path: `newRaft` (raft.go:444) feeds `InitialState`'s ConfState straight into `confchange.Restore` (raft.go:472), whose `toConfChangeSingle` dereferences the slice fields — nil panics.

Tightened `TestInitialState_EmptyWAL` to assert the ConfState is non-nil, and added `TestNewRawNode_EmptyWAL`, an end-to-end regression that constructs a `raft.RawNode` over a fresh empty WAL. Confirmed it panics without the fix and passes with it.

internal/storage/wal/wal_default.go
```diff
@@ -346,26 +352,33 @@ func New(dataDir string, logger logging.Logger, meter metric.Meter, opts ...Opti
...
 	return nil
 }
 
 // InitialState returns the saved HardState and ConfState information.
-func (s *DefaultWAL) InitialState() (raftpb.HardState, raftpb.ConfState, error) {
+//
+// The ConfState is wrapped with raftpb.EnsureConfState so it (and all of its
+// pointer fields) is never nil. raft v3.7 requires this: newRaft feeds the
+// returned ConfState straight into confchange.Restore, which dereferences its
+// slice fields — a nil ConfState (as an empty WAL would otherwise yield, since
```
NumaryBot, line 370: 🟠 **[major] InitialState() can return nil ConfState, violating raft v3.7 Storage contract**

DefaultWAL.InitialState() returns s.snapshot.GetMetadata().GetConfState(), which is nil when the WAL is freshly created (s.snapshot starts as an empty &raftpb.Snapshot{} with no metadata or conf state). Raft v3.7's Storage.InitialState() contract requires a non-nil ConfState; raft.NewRawNode dereferences it in confchange.Restore, causing a nil-pointer panic on a fresh node. MemoryStorage handles this correctly with raftpb.EnsureConfState(cs), but DefaultWAL does not.

**Suggestion:** Wrap the returned ConfState with raftpb.EnsureConfState(...) before returning it, e.g.:

cs := s.snapshot.GetMetadata().GetConfState()
return hardState, raftpb.EnsureConfState(cs), nil

Also update TestInitialState_EmptyWAL to assert that the returned ConfState is non-nil.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1540#issuecomment-4923606323

internal/storage/wal/wal_default.go
```diff
@@ -346,26 +352,33 @@ func New(dataDir string, logger logging.Logger, meter metric.Meter, opts ...Opti
...
+// s.snapshot starts as &raftpb.Snapshot{} with a nil Metadata.ConfState) would
+// panic. This mirrors what raft's own MemoryStorage.InitialState does.
+func (s *DefaultWAL) InitialState() (*raftpb.HardState, *raftpb.ConfState, error) {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 
-	return s.hardState, s.snapshot.Metadata.ConfState, nil
+	return s.hardState, raftpb.EnsureConfState(s.snapshot.GetMetadata().GetConfState()), nil
```
flemzord: **[P2] Avoid mutating the shared ConfState under RLock**

This can race when concurrent callers invoke `InitialState` on a snapshot whose `AutoLeave` field is unset. `raftpb.EnsureConfState` writes `cs.AutoLeave`, but this passes the `ConfState` owned by `s.snapshot` while only holding `RLock`; a concurrent 64-caller `go test -race` reproduces the read/write race inside `EnsureConfState`. Consider cloning before ensuring (or only allocating when the `ConfState` itself is nil) and add a concurrent regression test.

internal/storage/wal/wal_default.go
```diff
@@ -346,26 +352,45 @@ func New(dataDir string, logger logging.Logger, meter metric.Meter, opts ...Opti
...
 	return nil
 }
 
 // InitialState returns the saved HardState and ConfState information.
-func (s *DefaultWAL) InitialState() (raftpb.HardState, raftpb.ConfState, error) {
+//
+// The ConfState is wrapped with raftpb.EnsureConfState so it (and all of its
+// pointer fields) is never nil. raft v3.7 requires this: newRaft feeds the
+// returned ConfState straight into confchange.Restore, which dereferences its
+// slice fields — a nil ConfState (as an empty WAL would otherwise yield, since
```
NumaryBot, line 370: 🟠 **[major] InitialState() may mutate shared ConfState under RLock, causing a data race**

Even after wrapping with raftpb.EnsureConfState(), the call passes the ConfState pointer owned by s.snapshot while only holding RLock. EnsureConfState writes the AutoLeave field when it is unset, so concurrent callers to InitialState can produce a read/write race on that field. This is a concurrent regression distinct from the nil-return fix that was already applied.

**Suggestion:** Clone (deep-copy) the ConfState before passing it to raftpb.EnsureConfState, so the mutation does not touch the shared snapshot object. For example: cs := proto.Clone(s.snapshot.GetMetadata().GetConfState()).(*raftpb.ConfState); return hardState, raftpb.EnsureConfState(cs), nil. Add a concurrent regression test exercising parallel InitialState calls with -race.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1540#issuecomment-4923606323

internal/storage/wal/wal_default.go
gfyrag: Fixed in bc141bed7. `InitialState` no longer passes the shared snapshot ConfState to `EnsureConfState`: it clones it first (`proto.Clone` — the raftpb messages carry a `MessageState` mutex, so a plain struct copy trips `copylocks`) and lets `EnsureConfState` fill `AutoLeave` on the clone, leaving the shared value untouched under RLock. Added `TestInitialState_ConcurrentNoRace` (64 goroutines, snapshot ConfState with nil AutoLeave — the exact mutation trigger) plus `TestInitialState_ConcurrentEmptyWALNoRace`; both assert non-nil ConfState/AutoLeave and that the shared snapshot is not mutated. Verified the tests reproduce the race against the old one-liner and pass under `go test -race ./internal/storage/wal/...` with the fix.

flemzord (approved):
Reviewed the current head after the follow-up fixes. The previously reported issue is addressed, and I found no remaining actionable issue in this PR.

PR #1541, opened 2026-07-09
perf(schemathesis): disable Hypothesis shrinking by default

sylr:
## Problem

The **Tests-Schemathesis** CI job takes **~8m45s**, almost all of it in the test run itself (~8m20s). Setup is cheap (build 7s, venv 2s, pip 8s).

Analyzing the job log timestamps ([run 29002491665](https://github.com/formancehq/ledger/actions/runs/29002491665)), the run is dominated by two silent single-threaded gaps — **344s + 121s** — where Hypothesis grinds through endpoints. The driver is **shrinking**: several endpoints return HTTP 500 (`numscripts`, `prepared-queries`), and each failure kicks Hypothesis into a shrink phase that generates ~7× the nominal request volume (`max_examples=10 × 58 endpoints` — correlation IDs run past `000432`).

## Fix

Disable the Hypothesis `shrink` phase by default, with a `--shrink` opt-in for local debugging. Shrinking only *minimizes* an already-found failing example — it never changes *which* failures are detected (that happens in the `generate` phase). On a conformance gate the minimal repro isn't worth the cost.

## Measured impact

Benchmarked against a real single-node server (`MAX_EXAMPLES=10`):

| Config | Time | Endpoints | Errored |
|---|---|---|---|
| Baseline (shrink on) | ~350s | 56 | 1 |
| **Shrink off (this PR)** | **~46s** | **58** | **0** |

**≈7.5× faster on the test run, no coverage loss, no failure-detection loss.** Total CI job should drop from ~8m45s to roughly ~1m15s.

### Alternatives rejected (with evidence)
- `workers_num=4`: only 47→35s, but injected 11–15 spurious `errored` results (connection resets from parallel-hammering the single-node Raft server) → flakiness not worth it.
- `stateful=None`: no time gain and dropped 12 endpoints of coverage.

## Notes
- No workflow change needed — CI already calls `MAX_EXAMPLES=10 just test-schemathesis` → `run.sh` → `test_api.py`, which now defaults to shrink-off.
- Out of scope: the underlying HTTP 500s on `numscripts` / `prepared-queries` are real API bugs the fuzzer catches. The job is `continue-on-error: true`, so they don't block CI today — worth a separate follow-up.
- Benched on macOS/arm64, not the exact `namespace-profile-linux-amd64-4vcpu` runner, but the win is request-volume reduction, which is machine-independent.

### DIFF
--- a/tests/schemathesis/run.sh
+++ b/tests/schemathesis/run.sh
@@ -5,7 +5,9 @@
 # waits for readiness, runs Schemathesis tests, then tears down.
 #
 # Usage: bash tests/schemathesis/run.sh
-# Env vars: HTTP_PORT, GRPC_PORT, RAFT_PORT, MAX_EXAMPLES
+# Env vars: HTTP_PORT, GRPC_PORT, RAFT_PORT, MAX_EXAMPLES, SCHEMATHESIS_SHRINK
+#   SCHEMATHESIS_SHRINK=1 re-enables Hypothesis shrinking (minimal failing
+#   examples) for local debugging. Off by default — see test_api.py --shrink.
 set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
@@ -62,6 +64,11 @@ fi
 
 echo "==> Running Schemathesis tests..."
 echo ""
+SHRINK_FLAG=""
+if [ -n "${SCHEMATHESIS_SHRINK:-}" ] && [ "${SCHEMATHESIS_SHRINK}" != "0" ]; then
+    SHRINK_FLAG="--shrink"
+fi
 python3 "$SCRIPT_DIR/test_api.py" \
     --base-url "http://localhost:$HTTP_PORT" \
-    --max-examples "$MAX_EXAMPLES"
+    --max-examples "$MAX_EXAMPLES" \
+    $SHRINK_FLAG
--- a/tests/schemathesis/test_api.py
+++ b/tests/schemathesis/test_api.py
@@ -148,9 +148,13 @@ def main():
     # gate it does not change which failures are found (that happens in the
     # `generate` phase) — it only minimizes an already-found failing example,
     # at the cost of a ~7x blowup in request volume when endpoints fail.
-    phases = [Phase.explicit, Phase.reuse, Phase.generate, Phase.target]
+    # Derive from the Hypothesis defaults and remove *only* `shrink`, so every
+    # other default phase (`explain`, etc.) stays enabled and `--shrink`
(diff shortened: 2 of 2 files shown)


gfyrag:
Re-reviewed current head `9117b29bc0c58519a91a1682c99d59f088f44d65`.

Two low-severity notes:

1. The explicit Hypothesis `phases` list does not only remove shrinking.

With the resolved deps today (`schemathesis 3.39.16`, `hypothesis 6.156.6`), Hypothesis defaults are `(explicit, reuse, generate, target, shrink, explain)`. This PR drops `Phase.explain` too, and `--shrink` only appends `Phase.shrink`, so local debugging does not restore previous behavior. Safer code would derive from `hypothesis_settings.default.phases` and remove only `Phase.shrink`, or explicitly document that `explain` is also disabled.

2. The advertised local `--shrink` opt-in is only available when invoking `test_api.py` directly.

`just test-schemathesis` uses `tests/schemathesis/run.sh`, which starts the server and passes only `--base-url` / `--max-examples`. For developer ergonomics, please expose an env flag such as `SCHEMATHESIS_SHRINK=1` and update the `run.sh` env-var comment.

Checks run:

- `gh pr checks 1541 --repo formancehq/ledger`
- `gh run view 29008267826 --job 86085148071 --log`
- Temp venv install of `tests/schemathesis/requirements.txt`
- `python -m py_compile tests/schemathesis/test_api.py`
- `GOROOT= go build ./...`

sylr:
Thanks @gfyrag — both addressed in 33968d9bc.

1. **`Phase.explain` was being dropped.** Confirmed: `settings.default.phases` today is `(explicit, reuse, generate, target, shrink, explain)`, and my hardcoded list silently dropped `explain`. Now derived from `hypothesis_settings.default.phases` with only `Phase.shrink` filtered out, so every other default phase (incl. `explain`) stays enabled and `--shrink` restores the exact stock tuple.

2. **`--shrink` unreachable via `just`.** `run.sh` now passes `--shrink` when `SCHEMATHESIS_SHRINK` is set to a non-zero value, and the env-var header documents it:
   ```
   SCHEMATHESIS_SHRINK=1 just test-schemathesis
   ```

Re-verified end-to-end (`MAX_EXAMPLES=10`): 58 endpoints, 0 errored, ~36s — re-including `explain` does not regress the speedup.

gfyrag (approved):
Re-reviewed current head `33968d9bcf1da5cbbb04dc175ca85a1f5bf2bbf0`. No blocking findings from my pass. The shrink toggle preserves the non-shrink default phases and `SCHEMATHESIS_SHRINK` is wired through `run.sh`.

PR #1474, opened 2026-07-02
fix(EN-1424): guard incremental auditItem export on non-zero count

gfyrag:
## Summary

- Fixes [EN-1424](https://formance-team.atlassian.net/browse/EN-1424): incremental backup manifests referenced a non-existent `auditItem` segment for failure-only ranges, silently breaking restore (`ApplyExports` aborts with `ErrFileNotFound` on the missing key).
- Captures `itemCount` from `exportEntries` and guards the manifest append on `itemCount > 0`, mirroring the pre-existing pattern applied to the `appliedProposal` branch immediately below it. Three sub-zone segments (`audit` / `auditItem` / `appliedProposal`), three independent guards.
- Adds two regression tests using a fully-functional in-memory `Storage` (existing `recordingStorage` discards `PutFile` bodies, which precluded a real round-trip).
- **Also documents the companion-stream cardinality** in `docs/technical/architecture/subsystems/checker/audit-chain.md` — the underlying model was documented implicitly across code, `AGENTS.md`, and scattered doc pages. This PR surfaces it as a table + explicit "downstream implication" so the mistake that caused EN-1424 stops being invitable.

## Root cause in one line

The backup code carried a hidden assumption that `audit.count > 0 ⇒ auditItem.count > 0`, but the two counters advance independently. A failure entry advances the audit sequence (`state.machine.go` calls `writeAuditEntry(failureEntry, nil, "failure")`) without writing any `AuditItem` (`appendAuditItems` with an empty slice is a no-op). The sibling `appliedProposal` branch already handled the same asymmetry with an explicit `if appliedCount > 0` guard and a matching comment — the `auditItem` branch was written first and never received the same treatment.

## Safety re. audit hash chain

For a failure entry, `hashSlices` at `state.machine.go:1405-1413` is initialized with the header payload alone, and the loop over `items` is empty — the failure hash is bound to the header only. Skipping an empty `auditItem` segment at export time loses nothing that `verifyAuditHashChain` re-derives at check time. Symmetric to the existing `appliedProposal` guard.

## Trigger surface

- Low-traffic ledgers where a full backup interval elapses with zero successful proposals.
- Bursts of rejected transactions between two incrementals (misconfigured client, `ErrInsufficientFund` in series, replay attack).

The bug is silent at backup time (no error, `segmentsUploaded` incremented, manifest written cleanly) — it only surfaces at restore, i.e. exactly when the backup is needed.

## Doc changes

[...]

PR #401, opened 2023-01-03
chore: Update swagger with new camelCase API response fields, and better response objects

antoinegelloz:

PR #1543, opened 2026-07-09
fix(backup): crash-safe manifest ordering + verify full/incremental backup lifecycle (EN-888, EN-1055)

gfyrag:
## Summary

Consolidates and verifies the **full + incremental backup / restore** feature end to end for v3.0, and closes the manifest-atomicity concern from **EN-1055**.

Jira: **EN-888** (Make backups incremental), **EN-1055** ([R-041] backup manifest atomicity), epic **EN-867** (Ledger v3.0).

## Scope decision (after auditing the code)

The backup feature is already substantially implemented (PRs #1433, #1474, #1489, #401). I audited the whole lifecycle — full backup, incremental backup, manifest, concurrency, failure paths, restore full, restore full + incrementals — against the tickets, and adapted the plan to the code:

- **EN-1055 is already resolved, but not the way the Jira plan proposed.** The plan called for a *cluster-wide leased lock* (PR #298/#299). The codebase instead uses **FSM-managed per-destination mutual exclusion** (`BackupJobsState`, keyed by `CanonicalDestinationKey`): full (`BackupOrder`) and incremental (`IncrementalBackupOrder`) both route through the leader and propose a `Start` that lands on the same destination slot; the second gets `ErrBackupInProgress`. This is a strictly better design — deterministic, no clock skew, no lease renewal — so I kept it and did **not** reintroduce a leased lock. The read-modify-write of the shared manifest can no longer interleave, which is exactly what EN-1055 asked for.

- **One real correctness gap found and fixed.** `RunBackup` deleted stale checkpoint files and obsolete export segments *before* writing the new manifest, opening a crash window where the currently-published manifest referenced already-deleted objects (neither old nor new manifest restorable). Those deletes were also redundant with the post-manifest orphan-prune. Fix: drop the pre-manifest deletes; the single orphan-prune that runs strictly *after* `WriteManifest` is a superset and keeps the cycle crash-safe.

## Changes

- `internal/infra/backup/manager.go` — remove pre-manifest deletes from `RunBackup`; document the crash-safe write ordering. (Impact analysis: **LOW** — single production caller `Orchestrator.RunFull`, 0 processes affected.)
- `internal/infra/backup/crash_safety_test.go` (new) — ordering guard (verified to **fail on the pre-fix code**), failed-segment-upload-leaves-manifest-untouched, second-backup-succeeds-even-when-every-delete-fails, and a full + multi-incremental export/apply/rebuild round-trip.
- `tests/e2e/cluster/restore_test.go` — add a second incremental round in Phase 1 and a Phase 3 assertion, covering restore from **full + multiple incrementals**.
- `docs/.../chapters/backup.md` — corrected sequence diagram, added crash-safe write ordering and per-destination mutual-exclusion (EN-1055) sections.

## Invariants touched

- **#4 Pebble writes only from lifecycle paths** — respected; `internal/infra/backup/` is a declared lifecycle path, no new `OpenWriteSession` call sites.
- **#8 audit is source of truth** — respected; backup/manifest are projections/exports of the audit chain, not a new source of truth. No new persisted projection, so no checker change required.
- **#2 FSM deterministic / #7 fail loud** — untouched; the FSM backup path already fails loud on nil op and business rejections. No external I/O added to the FSM.

## Test plan

```bash
# Unit (default build)
go test ./internal/infra/backup/... ./internal/application/backup/... ./internal/infra/state/... -count=1

# Crash-safety + round-trip specifically
go test ./internal/infra/backup/ -run 'TestRunBackup_ManifestWrittenAfter|TestRunBackup_SecondBackupSucceeds|TestRunIncrementalBackup_SegmentUploadFailure|TestBackup_MultipleIncrementalsChain' -v

# E2E full + multiple incrementals restore chain (needs Docker/MinIO)
go test -tags "e2e,s3" ./tests/e2e/cluster/ -run Restore -timeout 600s
```

All unit suites pass; `go build ./...` and `golangci-lint run internal/infra/backup/...` clean (0 issues).

## Notes / no design gaps

No design gap surfaced: the FSM slot + crash-safe ordering make full, incremental, concurrency, failure paths, and restore-from-chains sound. No new business/technical concept was introduced.

---

## Update: content-addressing for checkpoint objects (Major bug fixed post-review)

An independent review found a **Major** bug this PR originally missed. A Pebble checkpoint contains a `MANIFEST-NNNNNN` file that keeps the **same local name** but **grows** between checkpoints. With name-keyed storage (`data/<filename>`), the next full backup re-uploaded that object **in place** — overwriting an object the currently-published backup manifest still referenced, *before* the manifest swap. A crash in that window left the previous backup pointing at inconsistent Pebble metadata (restore does no integrity check → silent corruption). Same EN-1055 window, at the object level rather than the delete level. Confirmed empirically (MANIFEST-000001: 39→97 bytes, same name, re-uploaded in place).

**Fix (approach a — content-addressing):** checkpoint objects are now stored under `data/<filename>.<sha256>`. A file whose bytes change lands on a **new** key and never overwrites the object a published manifest references; the old object is pruned only *after* the manifest swap. Identical content ⇒ same key ⇒ unchanged SSTs deduped/skipped (replaces the old name+size diff; correctness is now the content hash). Chosen over staging+swap (b) because incremental segments already use unique keys and the `data/` prefix is shared across many incrementals — per-object immutability is the minimal, consistent change.

**Manifest schema (pre-GA, breaking):** `checkpoint.files` changes from `{filename: size}` to `{filename: {size, key}}`. Restore resolves objects by the recorded `key`, never by `prefix+filename`. Older backups must be retaken.

**Tests:** `TestRunBackup_NeverOverwritesManifestReferencedObject` — two full backups across a MANIFEST-growing mutation; asserts no data/ object is overwritten with different bytes, the checkpoints genuinely differ, and every current-manifest object's stored bytes hash to the key it's stored under. **Verified to fail on the pre-fix name-keyed code and pass after.** The misnamed e2e "concurrent incremental" case (a sequential no-op) was renamed to reflect what it tests.

Updated both restore readers (gRPC download + `ledgerctl store bootstrap`), all backup e2e object-existence checks, and `backup.md`.

Impact analysis (GitNexus): `CheckpointFileKey`, `listCheckpointFiles`, `downloadOneFile` all **LOW** risk, 0 processes affected.


gfyrag (commented):
Reviewed #1543 at `fd3bb3528b343e91b463b40185b22503e8148133`.

Focus of this pass: backup manifest crash-safety ordering, full/incremental manifest semantics, and the added regression coverage.

Checks performed:
- Inspected `RunBackup` ordering: new/changed checkpoint files upload before `WriteManifest`; stale checkpoint/export cleanup is deferred to post-manifest orphan pruning.
- Inspected incremental backup path to confirm failed segment upload still aborts before manifest write, and successful incrementals append export segments before manifest publication.
- Reviewed the new crash-safety tests and e2e restore additions, including full + multiple incrementals.
- Checked that this stays in the declared lifecycle backup path and does not add FSM hot-path Pebble reads/writes.
- Ran `GOCACHE=/tmp/go-build-pr1543 go test ./internal/infra/backup/... ./internal/application/backup/... ./internal/infra/state/... -count=1` on a detached worktree at this SHA: green.

internal/infra/backup/manifest.go
```diff
@@ -18,11 +18,26 @@ type Manifest struct {
...
-	Timestamp         string           `json:"timestamp"`
-	LastAppliedIndex  uint64           `json:"lastAppliedIndex"`
-	LastLogSequence   uint64           `json:"lastLogSequence"`
-	LastAuditSequence uint64           `json:"lastAuditSequence"`
-	Files             map[string]int64 `json:"files"` // filename → size in bytes
+	Timestamp         string                    `json:"timestamp"`
+	LastAppliedIndex  uint64                    `json:"lastAppliedIndex"`
+	LastLogSequence   uint64                    `json:"lastLogSequence"`
+	LastAuditSequence uint64                    `json:"lastAuditSequence"`
+	Files             map[string]CheckpointFile `json:"files"` // local filename → stored object
```
NumaryBot, line 26: 🟠 **[major] New manifest schema breaks decoding of pre-upgrade manifests**

The updated schema stores checkpoint file entries as `map[string]CheckpointFile` (struct values), but manifests written by the previous implementation store them as `map[string]int64` (filename → size as a JSON number). On upgrade, any attempt to decode an existing manifest — required before both full/incremental backup runs and during restore/bootstrap — will fail with a type mismatch error ('cannot unmarshal number into Go struct'), making it impossible to restore from or replace existing backups on already-used destinations.

**Suggestion:** Implement a custom JSON unmarshaller for the checkpoint files field that first attempts to decode each value as a `CheckpointFile` struct and falls back to the legacy integer-size representation, reconstructing the equivalent `CheckpointFile` from it. Alternatively, add a manifest format version field and branch on it during decoding.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1543#issuecomment-4923720412

gfyrag:
Re: NumaryBot backward-compat finding on the manifest schema — addressed in 7c1f79340, with a deliberate scoping decision:

The schema change (`checkpoint.files`: `{filename: size}` → `{filename: {size, key}}`) is a **pre-GA, intentional break**, documented in the commit and `backup.md`. An in-place migration/dual-shape unmarshaller was considered and **deliberately not added**: legacy objects were stored under bare `data/<filename>` keys with no content hash, so they cannot be validated or safely folded into the content-addressed layout — a shim would carry an unverifiable path forward for a scenario the project has explicitly accepted breaking pre-GA.

What I did take from the finding: the failure mode was a cryptic `json: cannot unmarshal number ...`. `ReadManifest` now detects the legacy shape and returns a typed, actionable `ErrLegacyManifestFormat` ("this backup was written by an older binary and must be retaken"), so operators get clear guidance instead of a parse error. Covered by `TestReadManifest_LegacyFormatSurfacesActionableError` and `TestReadManifest_CurrentFormatDecodes`.

internal/infra/backup/manifest.go
```diff
@@ -18,11 +19,26 @@ type Manifest struct {
...
```
NumaryBot, line 26: 🟠 **[major] New manifest schema breaks decoding of pre-upgrade manifests**

The updated schema stores checkpoint file entries as `map[string]CheckpointFile` (struct values), but manifests written by the previous implementation encode them as `map[string]int64` (filename → integer size). On upgrade, any attempt to decode an existing manifest — required before both full/incremental backup runs and during restore/bootstrap — fails with a JSON type-mismatch error ('cannot unmarshal number into Go struct'), making it impossible to restore from or build upon backups on already-used destinations.

**Suggestion:** Implement a custom JSON unmarshaller for the checkpoint files field that first attempts to decode each value as a `CheckpointFile` struct and falls back to the legacy integer-size representation, reconstructing the equivalent `CheckpointFile` from it. Alternatively, add a manifest format version field and branch on it during decoding.

internal/infra/backup/manager.go
```diff
@@ -71,62 +74,65 @@ func RunBackup(
...
 	if err != nil {
 		return nil, fmt.Errorf("listing checkpoint files: %w", err)
 	}
 
-	// 3. Read existing manifest
-	existingManifest, err := ReadManifestOrEmpty(ctx, logger, storage, manifestKey)
-	if err != nil {
+	// 3. Read existing manifest (only for logging the delta; correctness no
+	// longer depends on diffing sizes — the content-addressed key is the diff).
+	if _, err := ReadManifestOrEmpty(ctx, logger, storage, manifestKey); err != nil {
```
NumaryBot: 🟠 **[major] Full backup aborts on legacy manifest instead of replacing it**

When a destination already holds a pre-content-addressing manifest, `ReadManifestOrEmpty` returns `ErrLegacyManifestFormat` and the full backup path aborts before publishing a new manifest — even though the returned manifest value is unused in the full-backup flow. This blocks the documented recovery path of retaking a full backup with the updated binary against the same destination; operators are forced to manually delete the manifest or choose a new prefix before any backup can proceed.

**Suggestion:** In the full-backup branch, treat `ErrLegacyManifestFormat` as a non-fatal condition (log a warning and proceed with an empty manifest), since the full backup is about to overwrite/replace the manifest entirely anyway.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1543#issuecomment-4923720412

gfyrag (commented):
Reviewed the latest head `7c1f79340ed8fcc0545fa242d2da3f0742d0b6e7` for the backup manifest/content-addressing changes.

No blocking findings. I focused on the crash-safety ordering, manifest immutability, restore path resolving by manifest-recorded object keys, and the explicit legacy-manifest failure mode. The restore/download and backup tests cover the risky paths.

Checked:
- `go test ./internal/infra/backup ./internal/adapter/grpc ./cmd/ledgerctl/store`
- `git diff --check origin/release/v3.0...origin/pr/1543`

internal/infra/backup/manager.go
NumaryBot: 🟠 **[major] Full backup aborts on legacy manifest instead of replacing it**

When a destination already holds a pre-content-addressing manifest, `ReadManifestOrEmpty` returns `ErrLegacyManifestFormat` and the full backup path aborts before publishing a new manifest — even though the previous manifest is unused in the full-backup flow. This blocks the documented recovery path of retaking a full backup with the updated binary against the same destination; operators are forced to manually delete the manifest or choose a new prefix before any backup can proceed.

**Suggestion:** In the full-backup branch, treat `ErrLegacyManifestFormat` as a non-fatal condition: log a warning and proceed with an empty manifest, since the full backup is about to overwrite the manifest entirely anyway.

internal/adapter/grpc/server_restore_download.go
```diff
@@ -217,53 +218,52 @@ func (s *RestoreServiceServerImpl) prepareDownload(
...
 	}
 
 	if int64(len(manifestData)) > maxRestoreManifestBytes {
-		return nil, nil, "", status.Errorf(codes.FailedPrecondition,
+		return nil, nil, status.Errorf(codes.FailedPrecondition,
 			"backup manifest exceeds %d bytes; refusing to read", maxRestoreManifestBytes)
 	}
 
 	var manifest backup.Manifest
 	if err := json.Unmarshal(manifestData, &manifest); err != nil {
```
NumaryBot: 🟡 **[minor] gRPC restore path bypasses legacy manifest detection**

The restore-download handler calls `json.Unmarshal` directly, bypassing the `backup.ReadManifest` legacy-shape detection introduced in this PR. For legacy manifests, operators will still receive a raw `cannot unmarshal number into Go struct` error instead of the actionable `ErrLegacyManifestFormat` message, leaving the gRPC restore path inconsistent with the rest of the manifest-reading logic.

**Suggestion:** Replace the direct `json.Unmarshal` call with `backup.ReadManifest` (or a shared helper) so that legacy-format detection and the actionable error message are applied consistently in the restore path.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1543#issuecomment-4923720412

internal/infra/backup/manager.go
NumaryBot: 🟠 **[major] Full backup aborts on legacy manifest instead of replacing it**

When a destination already holds a pre-content-addressing manifest, `ReadManifestOrEmpty` returns `ErrLegacyManifestFormat` and the full-backup code path aborts before writing the replacement manifest. Because a full backup does not diff against the existing manifest, this error should be non-fatal here. The documented operator recovery path — retaking a full backup with the updated binary against the same destination — is therefore blocked; operators are forced to manually delete the manifest or choose a new prefix before any backup can succeed.

**Suggestion:** In the full-backup branch, treat `ErrLegacyManifestFormat` as a non-fatal condition: log a warning and continue with an empty manifest, since the full backup is about to overwrite the manifest entirely anyway.

internal/adapter/grpc/server_restore_download.go
NumaryBot: 🟡 **[minor] gRPC restore path bypasses legacy manifest detection**

The restore-download handler calls `json.Unmarshal` directly, bypassing `backup.ReadManifest` and the `ErrLegacyManifestFormat` detection introduced in this PR. For legacy manifests, operators will still receive a raw `cannot unmarshal number into Go struct` error instead of the actionable error message, leaving the gRPC restore path inconsistent with the rest of the manifest-reading logic.

**Suggestion:** Replace the direct `json.Unmarshal` call with `backup.ReadManifest` (or a shared helper) so that legacy-format detection and the actionable error message are applied consistently in the restore path.

cmd/ledgerctl/store/bootstrap.go
```diff
@@ -140,8 +139,11 @@ func runBootstrap(cmd *cobra.Command, _ []string) error {
 
 		var totalBytes uint64
 
-		for filename := range manifest.Checkpoint.Files {
-			key := fileKeyPrefix + filename
+		for filename, cf := range manifest.Checkpoint.Files {
```
NumaryBot, line 144: 🟡 **[minor] Offline bootstrap command bypasses legacy manifest detection**

When `ledgerctl store bootstrap` is pointed at a legacy pre-content-addressing manifest, the command parses the manifest with a raw `json.Unmarshal` call, bypassing `backup.ReadManifest` and its `ErrLegacyManifestFormat` detection. Operators get the old cryptic unmarshal error instead of the actionable 'retake the backup' message on the offline bootstrap restore path.

**Suggestion:** Route the manifest parsing in the bootstrap command through `backup.ReadManifest` (or the same shared helper) so legacy-format detection is applied consistently here as well.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1543#issuecomment-4923720412

internal/infra/backup/manager.go
```diff
@@ -71,62 +74,76 @@ func RunBackup(
...
-		return nil, err
+	// 3. Read existing manifest (only for logging the delta; correctness no
+	// longer depends on diffing sizes — the content-addressed key is the diff).
+	// A legacy pre-content-addressing manifest is NOT fatal here: a full backup
+	// overwrites the manifest wholesale and never diffs against it, so retaking a
+	// full backup with the current binary is exactly the documented recovery path
+	// out of a legacy manifest. Treat it as a warning and proceed. (The
+	// incremental path, which does depend on the existing manifest, keeps the
+	// error fatal — see RunIncrementalBackup.)
+	if _, err := ReadManifestOrEmpty(ctx, logger, storage, manifestKey); err != nil {
```
NumaryBot: 🟠 **[major] Full backup aborts on legacy manifest instead of proceeding with empty manifest**

When a destination already holds a pre-content-addressing manifest, `ReadManifestOrEmpty` returns `ErrLegacyManifestFormat` and the full-backup code path aborts before writing the replacement manifest. Because a full backup does not diff against the existing manifest, this error is non-fatal in this context. The documented operator recovery path — retaking a full backup with the updated binary against the same destination — is therefore blocked; operators are forced to manually delete the manifest or choose a new prefix before any backup can succeed.

**Suggestion:** In the full-backup branch, treat `ErrLegacyManifestFormat` as a non-fatal condition: log a warning and continue with an empty manifest, since the full backup is about to overwrite the manifest entirely anyway.

internal/adapter/grpc/server_restore_download.go
```diff
@@ -217,53 +217,54 @@ func (s *RestoreServiceServerImpl) prepareDownload(
...
+		return nil, nil, status.Errorf(codes.FailedPrecondition,
 			"backup manifest exceeds %d bytes; refusing to read", maxRestoreManifestBytes)
 	}
 
-	var manifest backup.Manifest
-	if err := json.Unmarshal(manifestData, &manifest); err != nil {
-		return nil, nil, "", fmt.Errorf("parsing manifest: %w", err)
+	manifestPtr, err := backup.DecodeManifest(manifestData)
+	if err != nil {
+		return nil, nil, fmt.Errorf("parsing manifest: %w", err)
```
NumaryBot, line 240: 🟡 **[minor] gRPC restore path bypasses legacy manifest detection**

The restore-download handler calls `json.Unmarshal` directly, bypassing `backup.ReadManifest` and the `ErrLegacyManifestFormat` detection introduced in this PR. For legacy manifests, operators will still receive a raw `cannot unmarshal number into Go struct` error instead of the actionable error message, leaving the gRPC restore path inconsistent with the rest of the manifest-reading logic.

**Suggestion:** Replace the direct `json.Unmarshal` call with `backup.ReadManifest` (or a shared helper) so that legacy-format detection and the actionable error message are applied consistently in the restore path.

cmd/ledgerctl/store/bootstrap.go
```diff
@@ -140,8 +141,11 @@ func runBootstrap(cmd *cobra.Command, _ []string) error {
...
```
NumaryBot, line 144: 🟡 **[minor] Offline bootstrap command bypasses legacy manifest detection**

When `ledgerctl store bootstrap` is pointed at a legacy pre-content-addressing manifest, the command parses the manifest with a raw `json.Unmarshal` call, bypassing `backup.ReadManifest` and its `ErrLegacyManifestFormat` detection. Operators get the old cryptic unmarshal error instead of the actionable 'retake the backup' message on the offline bootstrap restore path.

**Suggestion:** Route the manifest parsing in the bootstrap command through `backup.ReadManifest` (or the same shared helper) so legacy-format detection is applied consistently here as well.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1543#issuecomment-4923720412

internal/infra/backup/manager.go
```diff
@@ -71,62 +74,76 @@ func RunBackup(
...
+	// content-addressed key already exists on storage is byte-identical to what
+	// is already there (that is what content-addressing guarantees), so it is
+	// skipped — this is the incremental dedup, now keyed by content rather than
+	// by name+size. Objects the new manifest will NOT reference are left in
+	// place and removed by the post-manifest orphan prune (step 8).
+	var toUpload []string
 
-	toUpload, toDelete := diffFiles(localFiles, previousFiles)
+	for filename, cf := range checkpointFiles {
+		exists, err := fileExists(ctx, storage, cf.Key)
```
flemzord: 🟡 **[minor] Avoid one remote GET per checkpoint file**

This makes every full backup perform a serial `GetObject`/`DownloadStream` for every checkpoint file just to test whether its content-addressed key exists. On a large Pebble store, hundreds or thousands of remote round trips add directly to backup latency and request cost; the previous manifest diff needed only one manifest read. Consider listing `CheckpointPrefix(bucketID)` once into a set (or adding a HEAD/Exists operation) and comparing all computed keys locally.

internal/infra/backup/manager.go
```diff
@@ -196,19 +217,21 @@ func RunBackup(
...
 		"lastLogSequence":   lastLogSeq,
 		"lastAuditSequence": lastAuditSeq,
 		"lastAppliedIndex":  lastAppliedIndex,
 	}).Infof("Backup completed")
 
 	return &Result{
-		FilesUploaded:     len(toUpload),
-		FilesDeleted:      len(toDelete),
+		FilesUploaded: len(toUpload),
+		// FilesDeleted is retired: with content-addressed keys, stale objects
```
flemzord: 🟡 **[minor] Keep the `files_deleted` metric truthful**

This leaves `Result.FilesDeleted` at zero even when the post-manifest prune successfully removes ordinary stale checkpoint objects. `BackupResponse.files_deleted`, `ledgerctl`, and the Operator still expose this as “stale files deleted”, while `orphans_deleted` is documented as failed-run leftovers, so normal backup churn is now misreported to every caller. Consider retaining the previous manifest long enough to classify successful prune deletions, or explicitly deprecating/removing the field and updating the protobuf/CLI/Operator contracts and docs in this PR.

flemzord (commented):
Review focused on the content-addressed full-backup path, crash-safe pruning, restore compatibility, and the public backup result contract. Two actionable findings are attached inline.

gfyrag:
Review pass on `c81a603f8810d7b18a97922df57db984882a823c`.

No blocking backup/restore behavior findings found in this head. I did find documentation/comment drift around the crash-safe ordering:

1. `internal/infra/backup/manager.go` still says the full backup diffs SST files against the previous manifest, cleans stale exports, and writes an updated manifest. Current code does content-address existence checks and prunes only after `WriteManifest`; the comment should match that ordering because it is the crash-safety invariant.

2. `docs/technical/architecture/subsystems/chapters/backup.md` documents the manifest file map as `name -> checksum -> size`, but the actual schema is `name -> {size,key}`.

3. `docs/ops/backup-restore.md` still describes “diff SSTs against previous checkpoint” and cleanup before writing the manifest. That contradicts the new crash-safe ordering where cleanup happens after the manifest commit.

The latest NumaryBot comments I checked look outdated on this head: legacy manifests are handled as non-fatal, and both gRPC restore and offline bootstrap now use `backup.DecodeManifest`.

Validation run: `GOROOT= go test ./internal/infra/backup ./internal/adapter/grpc ./cmd/ledgerctl/store`, targeted backup state tests, `GOROOT= go test ./internal/application/backup`, and `git diff --check release/v3.0...HEAD` passed.

gfyrag:
Follow-up/correction to my previous pass on `c81a603f8810d7b18a97922df57db984882a823c`: I rechecked the latest review threads, and two are valid, not outdated.

1. Major/perf: the backup path calls `fileExists` once per checkpoint file, and `fileExists` reads the object (`Storage.GetFile`, mapping to S3 `GetObject` / Azure `DownloadStream`). Since `Storage.ListFiles` already exists and both S3/Azure implement it, this should list `CheckpointPrefix(bucketID)` once into a set and compare expected keys locally. That avoids one object download per checkpoint file.

2. Major/API contract: `FilesDeleted` is now effectively always zero for normal stale checkpoint deletion, while those deletions are counted into `OrphansDeleted`. That conflicts with the public contract in proto, CLI output, and Operator status docs that still expose/report `FilesDeleted`. Either keep enough previous-manifest context to classify successful stale checkpoint deletions as `FilesDeleted`, or formally deprecate/remove/update the field across proto, CLI, Operator CRDs, tests, and docs.

The documentation drift from my previous comment still applies, but these two code/API issues should be handled first.

internal/infra/backup/manager.go
gfyrag: Fixed in 132bdc93d. Step 4 now lists `CheckpointPrefix(bucketID)` once into a set (`listKeySet`) and every computed content-addressed key is tested against it locally, replacing the per-file `GetFile` round trip. Removed the now-dead `fileExists` helper. Crash-safe ordering is unchanged — this only affects the existence probe, not upload/prune order.

internal/infra/backup/manager.go
gfyrag: Fixed in 132bdc93d — `FilesDeleted` is now truthful. The previous manifest (already read at step 3, previously discarded) is retained and its checkpoint keys captured via `checkpointKeySet`. `pruneOrphans` now returns the keys it deleted rather than a bare count, so step 8 classifies each checkpoint-prefix deletion: a key the previous manifest referenced = a stale checkpoint file this run superseded → `FilesDeleted`; a key no manifest referenced = a true orphan from a crashed run → `OrphansDeleted`. Export-prefix deletions are always orphans (a full backup writes `Exports: nil`). This matches the proto docs exactly (`files_deleted` = "diff vs previous manifest", `orphans_deleted` = "leftovers from failed runs"), so no proto/CLI/Operator change was needed.

Added `TestRunBackup_ClassifiesStaleVsOrphanDeletions`: it runs a first backup, injects a never-referenced orphan under data/, mutates+compacts, runs a second backup, and asserts the superseded previous-manifest objects count as `FilesDeleted` while the injected orphan counts as `OrphansDeleted`. Also updated the `pruneOrphans` unit tests to assert on the returned key slices.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1543#issuecomment-4923720412

internal/infra/backup/manager.go
```diff
@@ -177,38 +212,63 @@ func RunBackup(
...
+	// those). Running the prune strictly after WriteManifest is what makes the
+	// cycle crash-safe: at no point is a still-referenced object deleted.
+	expectedKeys := make(map[string]struct{}, len(checkpointFiles))
+	for _, cf := range checkpointFiles {
+		expectedKeys[cf.Key] = struct{}{}
 	}
 
-	orphansDeleted := pruneOrphans(ctx, logger, storage, CheckpointPrefix(bucketID), expectedKeys)
+	deletedCheckpointKeys := pruneOrphans(ctx, logger, storage, CheckpointPrefix(bucketID), expectedKeys)
 	// A full backup writes Exports: nil, so every export segment in storage is
```
NumaryBot, line 230: 🟠 **[major] Normal stale-checkpoint deletions miscounted as orphans; FilesDeleted always zero**
<sub>reported by flemzord, codex</sub>

Stale checkpoint object deletions that happen after the manifest commit are counted into OrphansDeleted rather than FilesDeleted. OrphansDeleted is documented (proto, CLI, Operator CRDs) as failed-run leftovers, whereas FilesDeleted is the field for normally superseded files. This makes FilesDeleted effectively always zero for routine backup runs and OrphansDeleted inflated. Additionally, when a full backup follows incremental backups, the deleted export segments under ExportPrefix are still referenced by the previously published manifest — deleting them is normal roll-up churn, not orphan cleanup — yet they are also counted into OrphansDeleted. Every caller (ledgerctl output, Operator status, monitoring) receives misleading numbers.

**Suggestion:** Retain enough context from the previous manifest to classify successful stale checkpoint deletions as FilesDeleted. For export-segment deletions that were referenced by the prior manifest, count them as FilesDeleted. Only count objects that were not referenced by any committed manifest as OrphansDeleted. If re-classifying is not feasible in this PR, formally deprecate FilesDeleted, update the proto definition, CLI output, Operator CRDs, and docs in this same PR, and open a tracking ticket.

internal/infra/backup/manager.go
```diff
@@ -71,62 +74,94 @@ func RunBackup(
...
-		return nil, err
+		if errors.Is(err, ErrLegacyManifestFormat) {
+			logger.Infof("Existing manifest uses the legacy pre-content-addressing format; " +
+				"a full backup replaces it wholesale, so proceeding and overwriting it")
+
+			prevManifest = &Manifest{}
+		} else {
+			return nil, err
+		}
 	}
```
NumaryBot, line 112: 🟠 **[major] O(N) object downloads per backup run to check checkpoint file existence**

The backup path calls fileExists once per checkpoint file to determine whether each content-addressed key already exists in storage. fileExists issues a full GetObject / DownloadStream call, meaning a large Pebble store with hundreds or thousands of checkpoint files performs that many serial remote round-trips on every backup run. Storage.ListFiles already exists and both S3 and Azure implement it, so this can be reduced to a single list call.

**Suggestion:** Call Storage.ListFiles with CheckpointPrefix(bucketID) once at the start of the backup run to build a set of already-present keys, then compare each expected key against that set locally. This reduces the per-file existence check from one GET/DownloadStream to a local map lookup.

gfyrag:
Review pass on `132bdc93d5c8a726b67785fd179556eb078d66b8` after the rebase.

The O(N) `fileExists` object-download issue looks fixed: `RunBackup` now lists the checkpoint prefix once and checks keys locally.

One deletion-accounting issue remains: previous incremental export segments are still reported as orphans after a full backup. `RunBackup` deletes every object under `ExportPrefix(bucketID)` and counts all of them in `OrphansDeleted`. If the previous committed manifest contained incremental `Exports`, those segment keys were live data intentionally rolled into the new full checkpoint, not leftovers from a failed run. That conflicts with the public `files_deleted` / `orphans_deleted` split.

Please build a previous-manifest export key set and classify deleted export keys the same way as checkpoint keys, or explicitly change the public contract/docs if exports are intentionally excluded from `FilesDeleted`.

Docs/comment drift also remains around full-backup ordering/schema: the manager comment and ops runbook still describe cleanup before manifest write, and the architecture doc still describes the file map as `name -> checksum -> size` instead of `name -> {size, key}`.

internal/infra/backup/manager.go
NumaryBot, line 230: 🟠 **[major] Export-segment deletions after full backup miscounted as orphans instead of FilesDeleted**
<sub>reported by codex, flemzord, NumaryBot</sub>

When a full backup follows one or more incremental backups, all objects under ExportPrefix (previously live export segments referenced by the last committed manifest) are deleted and counted into OrphansDeleted. OrphansDeleted is documented in proto, CLI, and Operator CRDs as failed-run leftovers, while FilesDeleted is the field for normally superseded files. The author's fix in 132bdc9 correctly classifies stale checkpoint deletions but explicitly notes that export-prefix deletions are always treated as orphans — this means intentional roll-up churn is still misreported to every caller (ledgerctl output, Operator status, monitoring).

**Suggestion:** Capture the previous manifest's export key set (analogous to the checkpointKeySet already built for checkpoint files). In pruneOrphans (or the post-prune classification step), keys that appear in the previous manifest's export set count as FilesDeleted; only keys absent from any committed manifest count as OrphansDeleted.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1543#issuecomment-4923720412

internal/infra/backup/crash_safety_test.go
```diff
@@ -0,0 +1,519 @@
...
+
+// orderedStorage is a fully functional in-memory Storage (like
+// inMemoryBackupStorage) that additionally records the ordered sequence of
+// Put/Delete operations so a test can assert crash-safety ordering:
+// every referenced object is uploaded before the manifest, and no object is
+// deleted before the manifest is written.
+//
+// putErrKeys lets a test inject a PutFile failure for keys matching any of the
+// given substrings, to exercise the segment-upload failure path.
+type orderedStorage struct {
```
flemzord: [P2] This adds a hand-written `Storage` test double; `failingDeleteStorage` later in the file adds another. The repository’s testing rule explicitly requires generated mocks for interfaces so call recording and interface drift are handled automatically. Please use a generated `MockStorage` with ordered expectations/failure injection instead of these manual implementations.

flemzord (commented):
Follow-up review of the updated head. These are new, validated findings not covered by the existing review threads.

internal/infra/backup/crash_safety_test.go
gfyrag: Done in aac49f55d. Both hand-written doubles (`orderedStorage` and `failingDeleteStorage`) are gone; the crash-safety tests now drive the generated `MockStorage`.

One design note on the shape of the conversion, since these are stateful end-to-end tests rather than expectation tests: the number, keys and relative order of the Put/Delete calls are driven by Pebble's checkpoint internals (SST names, MANIFEST growth), so they aren't enumerable up front for `InOrder`/ordered expectations. And several assertions read back real stored bytes (`ReadManifest`, content-hash verification of every manifest-referenced object). So I wired `MockStorage` with `DoAndReturn` handlers backed by a small `crashSafetyState` that delegates byte storage to the existing `inMemoryBackupStorage` fake, and layers the extra observations on top (op-ordering log, put-error injection, mutating-overwrite detection, delete-failure toggle). This keeps the generated interface-drift safety and call recording you asked for, while preserving every assertion unchanged.

Verified: `go build ./...`, `go build -tags "s3,azure" ./...`, `go test ./internal/infra/backup/...` (and with `-tags "s3,azure"`), gofmt + golangci-lint clean.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1543#issuecomment-4923720412

internal/infra/backup/manager.go
```diff
@@ -177,38 +212,63 @@ func RunBackup(
...
+	// now orphaned and can be removed. Export objects are never referenced by a
+	// checkpoint manifest, so they always classify as orphans.
+	deletedExportKeys := pruneOrphans(ctx, logger, storage, ExportPrefix(bucketID), nil)
+
+	// Classify the checkpoint deletions: a key the PREVIOUS manifest referenced
+	// is a stale checkpoint file this run superseded — that is ordinary churn and
+	// is what BackupResponse.files_deleted reports. A key no manifest referenced
+	// is a true orphan (leaked by a run that crashed before writing a manifest),
+	// reported as orphans_deleted. Export deletions are always orphans.
+	filesDeleted := 0
```
NumaryBot, line 242: 🟠 **[major] Export-segment deletions after full backup miscounted as orphans**
<sub>reported by codex, NumaryBot, flemzord, gfyrag</sub>

When a full backup follows one or more incremental backups, the previous committed manifest's export segments are intentionally superseded live objects being rolled up into the new full checkpoint. Deleting them and counting all of those keys into `OrphansDeleted` contradicts the public API contract (`orphans_deleted` = failed-run leftovers, `files_deleted` = normally superseded files) exposed via proto, CLI output, Operator CRDs, and monitoring. The author's own comment in 132bdc9 acknowledges export-prefix deletions are always treated as orphans, and gfyrag's fix in that same commit only addresses checkpoint-prefix classification, leaving export-prefix deletions uncorrected.

**Suggestion:** Build an export key set from the previous committed manifest's `Exports` field (analogous to the `checkpointKeySet` already built for checkpoint files). In the prune step, classify deleted export keys that appear in that set as `FilesDeleted` (normal roll-up churn) and only count export keys absent from any committed manifest as `OrphansDeleted`.

internal/infra/backup/manager.go
```diff
@@ -177,38 +213,74 @@ func RunBackup(
...
+	// those). Running the prune strictly after WriteManifest is what makes the
+	// cycle crash-safe: at no point is a still-referenced object deleted.
+	expectedKeys := make(map[string]struct{}, len(checkpointFiles))
+	for _, cf := range checkpointFiles {
+		expectedKeys[cf.Key] = struct{}{}
 	}
 
-	orphansDeleted := pruneOrphans(ctx, logger, storage, CheckpointPrefix(bucketID), expectedKeys)
+	deletedCheckpointKeys := pruneOrphans(ctx, logger, storage, CheckpointPrefix(bucketID), expectedKeys)
 	// A full backup writes Exports: nil, so every export segment in storage is
```
NumaryBot, line 230: 🟠 **[major] Export-segment deletions after full backup miscounted as orphans**

When a full backup follows incremental backups, export-segment keys referenced by the previously committed manifest's `Exports` field are deleted and counted entirely into `OrphansDeleted`. The author explicitly confirmed this in the PR discussion ('Export-prefix deletions are always orphans'). However, those keys were live data intentionally superseded by the new full backup — not leftovers from a crashed run. This directly contradicts the public `files_deleted`/`orphans_deleted` contract exposed via proto, CLI output, Operator CRDs, and monitoring dashboards. The fix applied to checkpoint keys (building a set from the previous manifest and classifying deletions accordingly) was not extended to export keys.

**Suggestion:** Build a previous-manifest export key set from the prior manifest's `Exports` field and classify each deleted export key: if it was referenced by the previous manifest, count it as `FilesDeleted`; if not referenced by any committed manifest, count it as `OrphansDeleted`. This mirrors the fix already applied to checkpoint keys in 132bdc9.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1543#issuecomment-4923720412

gfyrag:
Re-reviewed current head `b583bb39eea28df4447f71edfb2979ff4908f273`.

Remaining findings:

1. Major: incremental backups can succeed without a full checkpoint, but that artifact is not a complete restore base.

`RunIncrementalBackup` accepts an empty manifest and writes export-only manifests, and the S3 e2e currently expects this to succeed. Restore also accepts a manifest with exports and no checkpoint. But export-only backups do not carry Global-zone state such as persisted config, last applied index, or last applied timestamp. Restore validation requires persisted config, and finalization would snapshot index `0` for materialized historical data. That conflicts with the CLI/docs claim that incremental backup requires a prior full backup.

I would either reject incremental backups when `manifest.Checkpoint == nil`, or make export-only restore first-class by exporting/restoring the required Global-zone metadata and adding restore validation/e2e coverage for that mode.

2. Minor: docs/comments still describe the old full-backup ordering/schema.

`internal/infra/backup/manager.go` still says full backup diffs against the previous manifest, cleans stale exports, then writes the manifest. `docs/ops/backup-restore.md` still diagrams delete-before-manifest, and `docs/technical/architecture/subsystems/chapters/backup.md` still describes the manifest file map as `name -> checksum -> size`; the current schema is `name -> {size,key}`. These should match the new crash-safe invariant.

What looks fixed since the prior pass:

- Pre-manifest stale deletes.
- Content-addressing for same-name Pebble `MANIFEST-*` overwrites.
- Legacy manifest actionable errors.
- Full backup replacing legacy manifests.
- gRPC/offline bootstrap using `DecodeManifest`.
- O(N) object-download existence checks.
- Stale checkpoint `FilesDeleted`.
- Rolled-up export segment accounting.

Checks run:

- `GOROOT= go test ./internal/infra/backup/... ./internal/application/backup/... ./internal/infra/state/... ./internal/adapter/grpc/... -count=1`
- `GOROOT= go test -tags "s3,azure" ./internal/infra/backup/... -count=1`
- `GOROOT= go test ./cmd/ledgerctl/store -count=1`
- `GOROOT= go build ./...`
- `GOROOT= go build -tags "s3,azure" ./...`

Not run locally: Docker-backed e2e restore suite and full `just pre-commit`.

internal/infra/backup/manager.go
gfyrag: Addressed at head (7620bde93). The per-file GET/HEAD probe is gone: existence is now resolved with a single `listKeySet(ctx, storage, CheckpointPrefix(bucketID))` List (manager.go:128), and every content-addressed key is compared against the returned set locally. The comment at manager.go:123-127 documents the rationale. Resolving.

internal/infra/backup/manager.go
gfyrag: Fixed at head (7620bde93, refined in b583bb39e). `Result.FilesDeleted` is no longer stuck at 0: the post-manifest prune returns the deleted keys, which are then classified against the previous manifest — keys the previous manifest referenced count as `FilesDeleted` (supersede churn), keys no manifest referenced count as `OrphansDeleted` (manager.go:242-259, 281). Resolving.

internal/infra/backup/crash_safety_test.go
gfyrag: Migrated to the generated mock at head (aac49f55d). `failingDeleteStorage` and the hand-written `Storage` double are gone; the crash-safety tests now drive the generated `MockStorage` (storage_generated_test.go) via `newCrashSafetyStorage`, injecting faults through `DoAndReturn`. The residual `crashSafetyState` holds only observations that genuinely cannot be expressed as static gomock expectations (the number/keys/relative order of Put/Delete are driven by Pebble checkpoint internals, unknowable at test-authoring time) — documented at crash_safety_test.go:22-45. Resolving.

internal/infra/backup/manager.go
gfyrag: Already addressed at head (b583bb39e). The fix mirrors the checkpoint-key handling exactly as suggested: `prevExportKeys := exportKeySet(prevManifest)` is built from the previous manifest's `Exports` field (manager.go:114, alongside `prevCheckpointKeys`), and the post-manifest export prune is now classified (manager.go:253-258):

```go
for _, key := range deletedExportKeys {
    if _, wasReferenced := prevExportKeys[key]; wasReferenced {
        filesDeleted++   // rolled-up live segment the new full checkpoint superseded
    } else {
        orphansDeleted++ // never referenced by any committed manifest → true orphan
    }
}
```

So export segments the previous (incremental) manifest referenced now count as `files_deleted` (supersede churn), and only export objects no manifest ever referenced count as `orphans_deleted`. Covered by the regression test `TestRunBackup_ClassifiesSupersededExportsAsFiles` (full-after-incremental with an injected orphan export). Resolving.

NumaryBot (commented):

The PR delivers crash-safe manifest ordering, a typed ErrLegacyManifestFormat error, a single-prefix-list existence check, and correct classification of stale checkpoint keys versus true orphans. Previously raised issues around the O(N) per-file GET probe, the legacy manifest abort on full backup, and checkpoint deletion misclassification have been addressed. Two major issues remain open. First, the classification fix for stale-vs-orphan deletions was applied to checkpoint keys but not to export keys: after a full backup that supersedes incremental exports, all deleted export-segment keys are counted as orphans rather than files_deleted, breaking the public proto/CLI/Operator contract. Second, RunIncrementalBackup silently accepts a nil checkpoint manifest and produces an export-only artifact that cannot be fully restored (missing Global-zone state), contradicting the documented requirement for a prior full backup. Three minor documentation/comment drift issues also remain: the manager.go comment, the ops runbook, and the architecture doc all still describe the old pre-crash-safe ordering or the old manifest schema.

### Findings outside the diff

🟠 **[major] Export-segment keys from previous incremental backups misclassified as orphans after full backup** — `internal/infra/backup/manager.go` <sub>(NumaryBot, gfyrag)</sub>

When a full backup follows incremental backups, RunBackup deletes every object under ExportPrefix(bucketID) and counts them all in OrphansDeleted. However, export-segment keys referenced by the previously committed manifest were live, intentionally-written data (not leftovers from a crashed run). The same previous-manifest classification fix applied to checkpoint keys was not extended to export keys. This violates the public files_deleted/orphans_deleted contract exposed via proto, CLI output, and Operator CRDs/monitoring, where orphans_deleted is documented as failed-run leftovers and files_deleted as superseded-manifest data.

**Suggestion:** Build a previous-manifest export key set (analogous to checkpointKeySet for checkpoint keys) and classify deleted export-prefix keys the same way: keys the previous manifest referenced → FilesDeleted; keys no manifest referenced → OrphansDeleted. Alternatively, formally update/deprecate the field distinction across proto, CLI, Operator CRDs, tests, and docs.



🟠 **[major] RunIncrementalBackup accepts empty/nil manifest, producing a non-restorable export-only artifact** — `internal/infra/backup/manager.go`

RunIncrementalBackup succeeds when manifest.Checkpoint == nil and writes an export-only manifest. Restore also accepts such a manifest. However, export-only backups omit Global-zone state (persisted config, last applied index, last applied timestamp). Restore validation requires persisted config and would snapshot index 0 for materialized historical data. This contradicts the CLI/docs claim that incremental backup requires a prior full backup, and leaves operators with a silently incomplete restore base.

**Suggestion:** Either reject incremental backups when manifest.Checkpoint == nil (returning a clear error directing operators to run a full backup first), or make export-only restore first-class by exporting/restoring the required Global-zone metadata and adding restore validation and e2e coverage for that mode.



🟡 **[minor] Code comment still describes pre-crash-safe ordering (cleanup before manifest write)** — `internal/infra/backup/manager.go` <sub>(gfyrag, NumaryBot)</sub>

The manager.go comment describes full backup as diffing SST files against the previous manifest, cleaning stale exports, then writing the manifest. Current code performs content-address existence checks and prunes only after WriteManifest — the crash-safety invariant. The comment drift misrepresents the ordering guarantee.

**Suggestion:** Update the comment to accurately describe the current crash-safe ordering: existence checks → upload new files → WriteManifest → prune stale/orphan objects.



🟡 **[minor] Architecture doc describes manifest file-map schema as name→checksum→size instead of name→{size,key}** — `docs/technical/architecture/subsystems/chapters/backup.md`

The architecture documentation still documents the manifest file map as `name -> checksum -> size`, but the actual schema is `name -> {size, key}`. This creates confusion for operators and contributors reading the docs.

**Suggestion:** Update the manifest file-map description in backup.md to reflect the current `name -> {size, key}` schema.



🟡 **[minor] Ops runbook still describes old full-backup ordering (cleanup before manifest write)** — `docs/ops/backup-restore.md`

The backup-restore.md runbook still describes 'diff SSTs against previous checkpoint' and cleanup before writing the manifest, which contradicts the new crash-safe ordering where cleanup happens after the manifest commit.

**Suggestion:** Update the ops runbook to reflect the current crash-safe ordering: manifest is written before any prune/cleanup step.



Summary: https://github.com/formancehq/ledger/pull/1543#issuecomment-4923720412

flemzord (commented):
Revalidated the current head `25d5be11c` against every issue comment, review, inline thread, and reply. The generated `Storage` mock, single-list checkpoint existence check, `FilesDeleted` accounting, and superseded-export classification are all present and correct.

The existing High blocker remains unchanged: `RunIncrementalBackup` still accepts `manifest.Checkpoint == nil`, and the S3 e2e explicitly asserts that this export-only backup succeeds. Both restore paths then accept exports without a checkpoint, but the resulting store has no Global-zone persisted config / applied-index metadata; validation explicitly rejects the missing persisted config. This still contradicts the documented prior-full-backup requirement. Existing evidence and remediation options are captured here: https://github.com/formancehq/ledger/pull/1543#issuecomment-4944784640.

No duplicate inline comment added; keeping the PR review-required until incremental backup rejects a checkpoint-less manifest or export-only restore becomes complete.

internal/infra/backup/manager.go
flemzord, line 1: [High] `RunIncrementalBackup` must reject a manifest without a full checkpoint. It currently accepts the empty manifest and publishes export-only artifacts; restore accepts those exports, but they contain no Global-zone persisted config, last-applied index, or timestamp, so restore validation fails on missing persisted config. This contradicts the documented prior-full-backup requirement. Either fail when `manifest.Checkpoint == nil`, or make export-only restore complete by exporting/restoring the required Global metadata and covering that lifecycle end to end.

flemzord (changes requested):
Current-head re-review: one High restore-safety blocker remains. Incremental backup still accepts a checkpoint-less manifest and creates an artifact that restore cannot turn into a valid store. The validated finding is attached as a file-level inline comment.

internal/infra/backup/manager.go
gfyrag: Fixed in a3ffe4f01. `RunIncrementalBackup` now rejects a checkpoint-less manifest up front: right after `ReadManifestOrEmpty` and before any snapshot/export/upload, `if manifest.Checkpoint == nil { return nil, ErrNoFullCheckpoint }`. That covers both a fresh/empty destination and an export-only manifest, and it returns before any `PutFile`, so no export segment or manifest is ever published for an unrestorable base. New typed sentinel `ErrNoFullCheckpoint` (manifest.go) telling the operator to run a full backup first; the sole caller (orchestrator.go) already surfaces the runner error. Tests: rejects-checkpointless (asserts ErrNoFullCheckpoint + no artifact published) and proceeds-on-full-checkpoint; also repaired the audit-read-failure test that would otherwise short-circuit on the new guard. Could you re-review to clear the block?

flemzord: Validated on `63357ac0`: the guard now returns `ErrNoFullCheckpoint` immediately after reading the manifest and before opening a snapshot or uploading any segment. The unit test proves that an export-only manifest produces no `PutFile`, the S3 e2e verifies no object is published for a fresh destination, and the gRPC mapping exposes `FailedPrecondition`. This resolves the restore-safety blocker.

tests/antithesis/workload/bin/cmds/main/singleton_driver_incremental_backup/main.go
```diff
@@ -35,6 +35,25 @@ func main() {
...
 	client := clusterpb.NewClusterServiceClient(conn)
 
+	// 0. Establish a full checkpoint first. An incremental backup is only
+	//    meaningful layered on a full checkpoint (EN-888): the checkpoint carries
+	//    the Global-zone persisted config, last-applied index, and timestamp that
+	//    restore needs, so the server now rejects an incremental against a
+	//    checkpoint-less destination with FailedPrecondition. Antithesis schedules
+	//    drivers randomly, so this driver cannot assume the full-backup driver ran
+	//    first — it takes its own full backup to guarantee the precondition holds.
+	if _, err := client.Backup(ctx, &clusterpb.BackupRequest{Storage: s3Storage()}); err != nil {
```
flemzord, line 45: [Medium] Taking a full backup before every incremental run resets the manifest exports each time, so this driver largely stops exercising accumulated/multi-incremental chains. This call can also race another full/incremental backup on the shared destination: the server then returns the expected `ErrBackupInProgress` as `FailedPrecondition`, but this branch accepts only `IsTransient` and reports `assert.Unreachable`. Prefer attempting the incremental first, and only on `ErrNoFullCheckpoint` establish the full checkpoint and retry; explicitly tolerate a concurrent backup at that fallback step.

flemzord (commented):
The previous High restore-safety blocker is fixed on the current head. One non-blocking Antithesis coverage/classification issue is attached inline.

flemzord (approved):
Approved on current head `63357ac0`. The previous High restore-safety blocker is fixed: checkpoint-less incremental backups fail before any artifact is published, the error is exposed as `FailedPrecondition`, and unit/e2e coverage pins both the rejection and the normal full-plus-incremental path. The remaining Antithesis driver finding is Medium/non-blocking and is attached inline.

PR #1545, opened 2026-07-09
fix(EN-1460): query-checkpoint reads race-free with typed retryable readiness

gfyrag:
## EN-1460 — Query-checkpoint reads race the async read-index materialization

Fixes the opaque, non-retryable `code=Unknown` returned when a read targets a
query checkpoint whose read index is not yet materialized.

Jira: https://formance-team.atlassian.net/browse/EN-1460 (epic EN-867)

### Root cause
`CreateQueryCheckpoint` already called `WaitForSequence(logs[0].Sequence)`, but
that only waits on the index-builder **progress cursor**. In `process_logs.go`,
`WriteProgress` is flushed to Pebble in the batch that runs **before**
`createReadIndexCheckpoint` creates the `readindex/` directory. So the cursor
reaches the target sequence ~100-150ms before the directory exists → the wait
returns early → the immediate read opens a missing Pebble dir → the read-side
error wrapper falls through to the default sanitizer → `code=Unknown`
(non-retryable; the default gRPC retry policy only covers `UNAVAILABLE`).

### Fix (options a + b + c from the ticket)
- **(a) Blocking readiness on the creating node.** The index builder now writes a
  `.ready` marker file into the checkpoint dir as the **last** step, after
  pebble's `Checkpoint()` fully completes. `CreateQueryCheckpoint` waits via a new
  `readStore.WaitForCheckpoint` that gates on the marker (using the same
  progress cond var the builder already broadcasts). A read at the returned id is
  race-free on the creating node.
- **(b) Typed retryable error for not-yet-ready reads (multi-node / deferred).**
  `openCheckpointStores` returns a new `domain.ErrCheckpointNotReady` (reason
  `CHECKPOINT_NOT_READY` → gRPC `Unavailable`) when the marker is absent —
  mirroring the existing `INDEX_BUILDING → Unavailable` pattern. Covers a
  follower reached via a load balancer or a deferred read of a persisted
  `checkpoint_id`. Never returns partial state or `Unknown`.
- **(c) Hard-link hardening.** `createReadIndexCheckpoint` retries (bounded) on
  `link ... no such file or directory` — an SST compacted away between the
  manifest snapshot and the hard-link — cleaning the partial dir between attempts.

### Invariants
All changes are on the read side and the index builder, both **off the FSM hot
path**. No Pebble reads added to the FSM (#3); FSM determinism untouched (#2);
cache unaffected (#1). The `.ready` marker and checkpoint dirs are rebuildable
filesystem lifecycle state (a projection of the audit log), so they are outside
the checker's scope (#8) — consistent with the existing checkpoint dirs.

### Tests
- `internal/storage/readstore/checkpoint_wait_test.go` — `WaitForCheckpoint`
  fast-path, blocks-until-marker (reproduces the race), context-cancel, and
  create-checkpoint-then-mark-is-openable.
- `internal/adapter/grpc/errors_test.go` — `ErrCheckpointNotReady` maps to
  `Unavailable` with reason `CHECKPOINT_NOT_READY`, raw and wrapped, never Unknown.
- `internal/domain/errors_test.go` — new typed error registered in the
  Describable AST-scan registry.
- `tests/e2e/cluster/query_checkpoint_test.go` — regression Context: create +
  immediate read at the checkpoint in a loop, asserting no `Unknown` (only the
  typed `Unavailable` is acceptable on a miss).

### Proto
Append-only enum value `ERROR_REASON_CHECKPOINT_NOT_READY = 63` in
`misc/proto/common.proto`; regenerated `common.pb.go` / `common_vtproto.pb.go`.

### Docs
`docs/.../read-path/query-checkpoints.md` — documents the async materialization,
the `.ready` marker, the blocking-create readiness barrier, and the retryable
error contract.

### Checks
`GOROOT= go build ./...`, `golangci-lint` (changed pkgs, 0 issues), and unit
tests for `readstore`, `domain`, `adapter/grpc`, `indexbuilder`, `infra/state`
all pass.

### DIFF
--- a/internal/adapter/grpc/server_bucket.go
+++ b/internal/adapter/grpc/server_bucket.go
@@ -414,6 +414,16 @@ func (impl *BucketServiceServerImpl) openCheckpointStores(checkpointID uint64) (
 	mainPath := impl.store.QueryCheckpointMainDir(checkpointID)
 	readIndexPath := impl.store.QueryCheckpointReadIndexDir(checkpointID)
 
+	// The read-index checkpoint is materialized asynchronously by the index
+	// builder, and on a follower node the builder may not have caught up to the
+	// checkpoint's log sequence yet. Opening a not-yet-materialized directory
+	// would surface an opaque, non-retryable Unknown (EN-1460). Gate on the
+	// builder's readiness marker and return a typed, retryable error instead —
+	// mirrors the INDEX_BUILDING -> Unavailable pattern for metadata indexes.
+	if !readstore.CheckpointDirReady(readIndexPath) {
+		return nil, nil, &domain.ErrCheckpointNotReady{CheckpointID: checkpointID}
+	}
+
 	mainStore, err := dal.OpenReadOnly(mainPath, impl.logger)
 	if err != nil {
 		return nil, nil, fmt.Errorf("opening checkpoint main store: %w", err)
--- a/internal/storage/readstore/store.go
+++ b/internal/storage/readstore/store.go
@@ -556,6 +556,74 @@ func (s *Store) ListBackfillProgress() ([]BackfillEntry, error) {
 	return entries, nil
 }
 
+// checkpointReadyMarker is the sentinel file the index builder writes into a
+// query checkpoint read-index directory as the final step, once pebble's
+// Checkpoint() has fully completed (dir created, manifest written, and every
+// SST hard-linked). Its presence is the single authoritative readiness signal:
+// pebble writes SSTs last and a checkpoint can fail mid-link (EN-1460's
+// "link ... no such file or directory"), so the directory or its manifest
+// existing is NOT sufficient — only the marker is.
+const checkpointReadyMarker = ".ready"
+
+// CheckpointDirReady reports whether a query checkpoint read-index directory
+// has been fully materialized, i.e. the builder wrote the readiness marker.
+func CheckpointDirReady(dirPath string) bool {
+	_, err := os.Stat(filepath.Join(dirPath, checkpointReadyMarker))
+
+	return err == nil
+}
+
+// MarkCheckpointReady writes the readiness marker into a completed checkpoint
+// directory. The builder calls this after CreateCheckpoint returns without error.
+func MarkCheckpointReady(dirPath string) error {
+	return os.WriteFile(filepath.Join(dirPath, checkpointReadyMarker), nil, 0o640)
+}
+
+// WaitForCheckpoint blocks until the query checkpoint read-index directory at
+// dirPath has been fully materialized by the index builder (see
+// CheckpointDirReady), or the context is cancelled. It exists because the
+// progress cursor (LastIndexedSequence) is persisted in the same batch that
+// precedes the physical checkpoint creation: waiting on the sequence alone
+// races the directory into existence (the ~100-150ms window in EN-1460). The
+// index builder broadcasts on progressCond *after* createReadIndexCheckpoint
+// (which writes the marker) returns and lastIndexedSeq is stored, so
+// re-checking the marker on each broadcast observes it once present.
+func (s *Store) WaitForCheckpoint(ctx context.Context, dirPath string) error {
+	if CheckpointDirReady(dirPath) {
+		return nil
+	}
+
+	// Spawn a goroutine that broadcasts when the context is cancelled so the
+	// Wait() below is unblocked.
+	done := make(chan struct{})
+	defer close(done)
+
+	go func() {
+		select {
+		case <-ctx.Done():
+			s.progressCond.Broadcast()
+		case <-done:
+		}
+	}()
+
+	s.progressMu.Lock()
+	for {
+		if ctx.Err() != nil {
+			s.progressMu.Unlock()
+
+			return ctx.Err()
+		}
+
+		if CheckpointDirReady(dirPath) {
+			s.progressMu.Unlock()
+
+			return nil
+		}
+
+		s.progressCond.Wait()
+	}
+}
+
 // WaitForSequence blocks until LastIndexedSequence >= minSeq or the context
 // is cancelled.
 func (s *Store) WaitForSequence(ctx context.Context, minSeq uint64) error {
--- a/docs/technical/architecture/subsystems/read-path/query-checkpoints.md
+++ b/docs/technical/architecture/subsystems/read-path/query-checkpoints.md
@@ -21,8 +21,17 @@ Checkpoint IDs are assigned sequentially by the FSM (1, 2, 3, ...).
 2. The request is proposed through Raft consensus.
 3. The FSM commits pending state and records `QueryCheckpointState` metadata in Pebble.
 4. The Applier creates a physical Pebble checkpoint of the main store at `{dataDir}/query-checkpoints/{id}/main/`.
-5. The index builder detects the `CreatedQueryCheckpointLog` and creates a read index checkpoint at `{dataDir}/query-checkpoints/{id}/readindex/`.
+5. The index builder detects the `CreatedQueryCheckpointLog` and creates a read index checkpoint at `{dataDir}/query-checkpoints/{id}/readindex/`. This is **asynchronous** — the read index is materialized off the FSM hot path by the index builder, typically ~100-150ms after the log is applied. When the checkpoint is fully hard-linked, the builder writes a `.ready` marker file into the `readindex/` directory as the last step. Because pebble hard-links SST files last and a checkpoint can fail mid-link (a concurrent compaction removing an SST), the builder retries the checkpoint on a `link ... no such file or directory` error before writing the marker.
 6. Both stores are opened read-only when a query specifies `checkpoint_id`.
+
+## Readiness and Error Contract
+
+The read index materializes asynchronously (step 5), so a read at a checkpoint whose read index does not yet exist would otherwise race the builder.
+
+- **`CreateQueryCheckpoint` blocks until the checkpoint is readable.** The handler waits (`readStore.WaitForCheckpoint`) for the `.ready` marker before returning, so on the node that created the checkpoint an immediate read at the returned `checkpoint_id` always succeeds. Waiting on the index-builder progress cursor alone is insufficient: the cursor advances in the batch that *precedes* the physical checkpoint creation, so the sequence is reached before the directory exists.
+- **A read on a not-yet-ready checkpoint returns a typed, retryable error.** On any node whose index builder has not yet materialized the checkpoint (e.g. a follower reached via a load balancer, or a deferred read of a persisted `checkpoint_id`), `openCheckpointStores` gates on the `.ready` marker and returns `ErrCheckpointNotReady` — reason `CHECKPOINT_NOT_READY`, mapped to gRPC `Unavailable`. This mirrors the `INDEX_BUILDING → Unavailable` pattern for metadata indexes: clients retry deterministically instead of receiving an opaque, non-retryable `Unknown`. The read never returns partial checkpoint state — either the fully-materialized checkpoint or the retryable readiness error.
+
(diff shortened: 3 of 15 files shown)


internal/adapter/grpc/server_bucket.go
```diff
@@ -414,6 +414,16 @@ func (impl *BucketServiceServerImpl) openCheckpointStores(checkpointID uint64) (
 	mainPath := impl.store.QueryCheckpointMainDir(checkpointID)
 	readIndexPath := impl.store.QueryCheckpointReadIndexDir(checkpointID)
 
+	// The read-index checkpoint is materialized asynchronously by the index
+	// builder, and on a follower node the builder may not have caught up to the
+	// checkpoint's log sequence yet. Opening a not-yet-materialized directory
+	// would surface an opaque, non-retryable Unknown (EN-1460). Gate on the
+	// builder's readiness marker and return a typed, retryable error instead —
+	// mirrors the INDEX_BUILDING -> Unavailable pattern for metadata indexes.
+	if !readstore.CheckpointDirReady(readIndexPath) {
```
NumaryBot, line 423: 🔴 **[blocker] Recover completed checkpoints that lack markers**

When a node already has a valid read-index checkpoint without the new `.ready` file (for example after upgrading from a version that created checkpoints before this marker, during a rolling deploy, or after a crash that leaves the checkpoint directory but not the marker), this guard returns `ErrCheckpointNotReady` forever. The index builder has already advanced past the `CreatedQueryCheckpoint` log, so it will not revisit that checkpoint to add the marker, and reads using a persisted `checkpoint_id` will retry indefinitely despite the checkpoint being openable. Please add a migration/recovery path or fallback validation for existing checkpoint directories before treating marker absence as transient readiness.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1545#issuecomment-4923748454

gfyrag (commented):
Reviewed #1545 at `968d314e91bc0a7762336fa2d9ae13bc6582fdcc`.

Focus of this pass: query-checkpoint read-index readiness, typed retryable errors, and checkpoint hard-link retry behavior.

Checks performed:
- Verified `CreateQueryCheckpoint` now waits on the read-index checkpoint `.ready` marker, not only the index-builder progress cursor.
- Verified `openCheckpointStores` refuses a not-yet-ready read-index checkpoint with `ErrCheckpointNotReady`, which maps to gRPC `Unavailable` / reason `CHECKPOINT_NOT_READY` rather than `Unknown`.
- Verified `.ready` is written only after `CreateCheckpoint` succeeds, and checkpoint creation retries boundedly on `os.ErrNotExist` hard-link races after removing a partial dir.
- Reviewed the proto enum addition and generated files, domain error registry coverage, gRPC error tests, readstore wait tests, and e2e regression loop.
- Ran `GOCACHE=/tmp/go-build-pr1545 go test ./internal/storage/readstore/... ./internal/domain/... ./internal/adapter/grpc/... ./internal/application/indexbuilder/... ./internal/infra/state/...` on a detached worktree at this SHA. The sandboxed run hit localhost bind restrictions in gRPC tests; rerun outside sandbox passed.

internal/adapter/grpc/server_bucket.go
gfyrag: Verified against the code — the finding is **legitimate**, and it decomposes into two cases with different answers.

**Root of the problem (confirmed):** the index builder's progress cursor (`readStore.LastIndexedSequence`) is persisted in the batch that runs *before* `createReadIndexCheckpoint`. At boot, `bootInit` resumes from that cursor. So if the cursor is durably past the `CreatedQueryCheckpoint` log while the `.ready` marker was never written, the builder never revisits the log and the directory stays markerless forever. I also confirmed there is **no reliable way to distinguish a complete-but-unmarked directory from an incomplete one**: pebble hard-links SSTs *last*, so a crash mid-`Checkpoint()` leaves a valid manifest with missing SSTs, and a read-only open **succeeds** (missing SSTs only surface on iteration). So auto-trusting a markerless dir (self-heal by writing a marker) would risk marking a half-linked checkpoint ready — worse than the bug. The marker is the only sound completeness discriminant.

**Case 1 — crash between checkpoint creation and the marker: FIXED (commit d7923ced).** `createReadIndexCheckpoint` now holds the durable cursor at `lastSeq-1` (the `CreatedQueryCheckpoint` log carries no index writes, so this costs nothing) until the marker is on disk, then advances to `lastSeq` in a second write. A crash before the marker leaves the cursor *before* the checkpoint log, so restart replays it and recreates the checkpoint. Recreation is made idempotent by clearing any stale directory up front (pebble errors on an existing dir). New unit tests cover both the happy path and idempotent recreate over a stale dir.

**Case 2 — pre-marker checkpoints from an already-deployed version (upgrade / rolling deploy): NOT auto-healed, deliberately.** The read-index checkpoint feature shipped in `v3.0.0-alpha.1..9`, so a deployment that created checkpoints on an alpha and then upgrades to this build will have durable markerless directories the builder has already passed. Because complete vs incomplete is indistinguishable without the marker, safely resolving these requires a product decision (discard-and-rebuild-from `QueryCheckpointState.max_sequence`, boot-time GC of legacy dirs, or require operators to recreate them). I'm escalating that policy choice rather than silently trusting on-disk state. Note the failure mode is a *retryable* `Unavailable`, not data corruption. Tracking separately; not blocking the forward-correctness fix here.

NumaryBot, line 423: 🟠 **[major] Absent `.ready` marker conflates permanent missing checkpoints with transient readiness lag**

When a `checkpoint_id` refers to a deleted, never-created, or pre-upgrade checkpoint directory (one written before this marker was introduced), the `.ready` file is absent and the code returns `ErrCheckpointNotReady`. Because clients are expected to retry `Unavailable`/`CHECKPOINT_NOT_READY`, they will loop indefinitely rather than receiving a clear 'not found' error. This is the same issue raised in the inline review thread (existing checkpoints lacking markers after upgrade/recovery) and also applies to genuinely non-existent IDs. A separate existence/metadata check on the checkpoint snapshot must be performed before treating marker absence as a transient condition.

**Suggestion:** First verify the checkpoint directory/snapshot exists independently of the marker file (e.g. check main metadata or index snapshot presence). Only return the transient `ErrCheckpointNotReady` if the snapshot itself exists but the marker is absent. If neither the snapshot nor the marker exists, return a permanent 'not found' error so callers do not retry indefinitely. Additionally, add a migration step or fallback that treats an existing, openable snapshot directory as ready even when the marker file is missing, to support pre-upgrade checkpoints.

internal/storage/readstore/store.go
```diff
@@ -556,6 +556,74 @@ func (s *Store) ListBackfillProgress() ([]BackfillEntry, error) {
...
+	if CheckpointDirReady(dirPath) {
+		return nil
+	}
+
+	// Spawn a goroutine that broadcasts when the context is cancelled so the
+	// Wait() below is unblocked.
+	done := make(chan struct{})
+	defer close(done)
+
+	go func() {
```
NumaryBot, line 601: 🟠 **[major] Missed-wakeup race between context-cancellation and `cond.Wait()` in `WaitForCheckpoint`**

In the wait loop, `ctx.Err()` is checked and then `cond.Wait()` is called without holding `progressMu` across both operations. If the context is cancelled (and any associated broadcast fires) in the window between the `ctx.Err()` check and the `Wait()` call, the goroutine will block on `Wait()` indefinitely despite the context already being done. This means RPC deadlines or cancellations that arrive in that narrow window are silently swallowed until a future (possibly never-arriving) progress notification wakes the waiter.

**Suggestion:** Arrange for the context-cancellation signal to always be deliverable while the cond is being waited on. One robust pattern is to start a goroutine that calls `cond.Broadcast()` after `<-ctx.Done()` while holding `progressMu`, ensuring no broadcast can be missed regardless of when cancellation occurs relative to `Wait()`. Alternatively, restructure the loop so the `ctx.Err()` guard and the subsequent `Wait()` are both protected by the same mutex lock without releasing it between them.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1545#issuecomment-4923748454

internal/adapter/grpc/server_bucket.go
gfyrag: Ack — the `.ready`-absence-vs-not-found conflation (major) and the recover-markerless-checkpoints (blocker) are the same underlying question: how to treat a checkpoint directory that exists but has no marker. As noted in the inline thread, complete-vs-incomplete is **not** distinguishable without the marker (pebble links SSTs last; a read-only open of a half-linked dir succeeds), so "treat any openable snapshot as ready" is unsafe. The `not-found` half of the major finding is valid and cheap: I can return a permanent `NotFound` when the checkpoint directory does not exist at all (distinct from exists-but-unmarked → transient). The recovery policy for pre-marker/legacy dirs is under design review — a per-node readiness design (local marker + boot reconciler vs. a replicated readiness command) is being evaluated before I implement, so I'm holding that part rather than guessing.

The `WaitForCheckpoint` missed-wakeup (major) is a legitimate narrow race inherited from the existing `WaitForSequence` pattern; fix will land with the chosen readiness design since the wait mechanism differs between the options.

gfyrag:
Closing this PR. After design review on [EN-1460](https://formance-team.atlassian.net/browse/EN-1460), the local `.ready`-marker-only approach in this PR is being replaced.

The blocker NumaryBot raised is real: a checkpoint dir that is complete but unmarked (crash between `pebble.Checkpoint()` and the marker write, upgrade from a pre-marker build, or a node whose index-builder cursor already advanced past the `CreatedQueryCheckpoint` log) would return `ErrCheckpointNotReady` forever, because nothing re-materializes it. The marker alone cannot fix that — a **reconciler** in the index-builder is required regardless.

A Raft-based per-node readiness command was explored and rejected: checkpoint reads are served purely locally on any node (no routing — `server_bucket.go:463`), and the repo already solves the analogous 'is my local replica ready?' question for secondary indexes with a **local per-replica marker** (`IndexVersionState.CurrentVersion`), keeping the Raft-replicated `BuildStatus` informational-only (excluded from both the read gate and the checker, invariant #8). A replicated readiness map would diverge from that precedent and cannot recreate a node's local Pebble files anyway.

Redesign (local marker + builder reconciler + typed retryable error) will be posted on EN-1460 and implemented in a fresh PR. The reusable, design-agnostic pieces from this branch — the `Dirty` regen fix, the crash-safe cursor reorder, the hard-link-vs-compaction retry, and the coverage tests — will be carried over.

[EN-1460]: https://formance-team.atlassian.net/browse/EN-1460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

gfyrag (commented):
Reviewed the new movement `968d314e91bc0a7762336fa2d9ae13bc6582fdcc` -> `d7923cede4926b03fc89fbc899557ffa09a7c285` from the active agent worktree `.claude/worktrees/agent-a1280040bfcd95ff4`.

Scope checked: the crash-safe read-index checkpoint cursor ordering, idempotent recreation of stale markerless checkpoint directories, `.ready` marker gating, and the new checkpoint/readstore tests. The ordering now keeps durable progress before the checkpoint log until the checkpoint is fully materialized and marked ready, then advances progress, which matches the intended recovery behavior. I did not find an actionable issue.

Validation:

```bash
GOCACHE=/tmp/go-build-pr1545 go test ./internal/application/indexbuilder/... ./internal/storage/readstore/... ./internal/infra/state/...
GOCACHE=/tmp/go-build-pr1545 go test ./internal/adapter/grpc/...
```

The first command passed except the sandboxed gRPC bind restriction; the gRPC package passed when rerun outside the sandbox.

PR #1547, opened 2026-07-09
fix(EN-1465): QueryFilter JSON codec aligned on the v2 query DSL (drop protojson leak)

gfyrag:
## Summary

Completes the **QueryFilter** half of [EN-1465](https://formance-team.atlassian.net/browse/EN-1465) (epic [EN-867](https://formance-team.atlassian.net/browse/EN-867)). The `postCommitVolumes` half already shipped in #1517 — this PR does not touch it.

**Problem.** The prepared-query REST path decoded `QueryFilter` via `protojson.Unmarshal`, and `PreparedQuery.MarshalJSON` routed the whole value through `protojson`. That leaked the protobuf-internal oneof/wrapper names onto the public wire — `and.filters[]`, `not.filter`, `field.field.metadata`, `stringCond`, `hardcodedExact`, `paramPrefix`, and the `QUERY_TARGET_*` enum prefix. The shape documented in `openapi.yml` **never existed on the wire**, and the spec only ever defined one leaf condition (`AccountHasAssetCondition`) — the other ~10 conditions were undocumented. SDKs generated from the spec were broken.

**Fix (no proto changes — JSON layer only).** A hand-written, bidirectional JSON codec on `commonpb.QueryFilter` (`internal/proto/commonpb/query_filter.go`), following the `Transaction`/`PostCommitVolumes` custom-marshaller pattern. `PreparedQuery.MarshalJSON` now uses it plus a string target enum; the HTTP decoder decodes through the same codec. `openapi.yml` documents the DSL.

## Canonical shape — the shared Formance query DSL

The REST `QueryFilter` shape **mirrors the shared Formance query DSL** (`go-libs/pkg/query`, as used by ledger v2). This is the canonical GA shape shared by every filtered endpoint (and reused by the audit conditions in #1548). Each node has exactly one top-level `$`-operator key:

| Concept | Shape |
|---|---|
| AND / OR | `{"$and": [ ... ]}` / `{"$or": [ ... ]}` (non-empty) |
| NOT | `{"$not": <QueryFilter>}` |
| metadata `==` | `{"$match": {"metadata[tier]": "gold"}}` |
| metadata range | `{"$gte": {"metadata[score]": 10}}` (closed range = `$and` of two bounds) |
| metadata exists | `{"$exists": {"metadata": "tier"}}` |
| address prefix/exact | `{"$match": {"address": "users:"}}` (`:` suffix = prefix) / `{"$match": {"address": "acc:1"}}` |
[...]

PR #1548, opened 2026-07-09
feat(EN-1241): bring ListAuditEntries under the shared ListOptions contract

gfyrag:
## Summary

`ListAuditEntries` was the last streaming list RPC not following the shared contract: it exposed `common.ListOptions` on the wire but rejected `filter` / `reverse` / `checkpoint_id`, and carried bespoke top-level `failures_only` / `ledger` fields. This PR makes audit listing **isomorphic** with `ListTransactions` / `ListAccounts` / `ListLogs`.

Jira: https://formance-team.atlassian.net/browse/EN-1241 (epic EN-867)

## Scope decision (challenged against the code)

The ticket offered two shapes. I chose the canonical **`{ string ledger = 1; common.ListOptions options = 2; }`** — byte-identical to `ListTransactionsRequest`. `ledger` stays top-level (a scope, like every sibling RPC; internally folded into the filter as `audit[ledger] == X`). `failures_only` is **removed** and expressed as `audit[outcome] == failure`.

The earlier plan / the closed PR #1448 (EN-1305) used **scan-time predicates** with no index, no reverse, no checkpoint. The v3.0 scope update explicitly rejects that "minimal audit-specific" approach and demands isomorphism. Since **EN-1339 (merged)** shipped an async audit secondary index in the readstore, this PR wires audit onto that **index-backed** path — the design intent EN-1339 stated ("audit listing can move onto the same index-backed filtering path").

## What changed

- **Proto** (`common.proto`): `QueryFilter.audit = AuditCondition` (field 13), `AuditField` enum (8 index-answerable fields), `QUERY_TARGET_AUDIT`. `bucket.proto`: `ListAuditEntriesRequest` reduced to `{ ledger=1, options=2 }`, old `failures_only`/order reserved.
- **Filter compiler** (`internal/query/audit_filter.go`): `CompileAuditFilter` resolves `audit[...]` leaves via the readstore index (`AuditSeqsByString/Outcome/Uint64Range`), `AND` = intersect, `OR` = union, `seq` = audit-zone key-range bound. `NOT`, non-indexed and non-audit conditions → `InvalidArgument`. **No scan-time fallback.**
- **Page reader** (`internal/query/audit.go` `ReadAuditEntriesPage`): applies cursor + reverse + page over the compiled candidate set (index-narrowed) or a bounded zone scan (unfiltered / `seq`-only), skipping purged-but-indexed sequences.
- **Controller**: `ListAuditEntries(ctx, ledger, pageSize, afterSequence, filter, reverse)` + `ListAuditEntriesFrom(...)` for the checkpoint path, mirroring `ListTransactions`/`ListTransactionsFrom`.
- **gRPC handler**: `ListOptionsSupport{Filter: true, Reverse: true, CheckpointID: true}`; checkpoint branch via `openCheckpointStores` (the checkpoint captures both the audit zone and the readstore index).
- **Filter DSL** (`filterexpr`): `audit[field] op value` grammar + formatter round-trip.
- **CLI** (`ledgerctl audit list`): `--filter`, `--reverse`, `--ledger`, `--checkpoint-id`; `--failures-only` kept as a client-side shorthand.
- **Bug fix**: `AuditSeqsByString` prefix scan could match `"alice\x00evil"` for a `"alice"` lookup — now guarded by an exact key-length check (the concern EN-1339 flagged for EN-1305 to fix).

## Consistency posture (explicit + justified)

The audit index is maintained by an **async per-node worker** (~200 ms tick), so a *filtered* audit read (incl. `--ledger` / `--failures-only`) is **eventually consistent**; `--min-log-sequence` gates the log read-index, not the audit index. An *unfiltered* read (plain list, `--reverse`, `audit[seq]` bounds) reads the audit zone directly and is **strongly consistent**. Documented in `cli.md`.

## Refused options (typed, justified — no silent skips)

- `not` / `!=` — the audit access path has no complement; would force a full-chain scan → `InvalidArgument`.
- Non-indexed audit fields (error_type, caller.scope, caller.god) — **not exposed** in the DSL at all (no half-supported field).
- Non-audit conditions (`metadata[...]`, address, …) — `InvalidArgument`.

## Invariants

Read-side only. No FSM / hot-path / Pebble-write changes. No new persisted projection (invariant #8) — the audit index is a pre-existing rebuildable access path, not a checker target. Audit log remains the source of truth.

## Testing

- Unit: `internal/query/audit_filter_test.go` (leaves, and/or intersect/union, seq bounds, uint range, rejections), `internal/query/audit_page_test.go` (ascending/reverse/cursor/bounds/seq-set/purged-skip), `internal/pkg/filterexpr/audit_test.go` (grammar + format round-trip), `internal/storage/readstore/audit_index_test.go` (NUL disambiguation).
- E2E (`tests/e2e/business/audit_test.go`): existing tests migrated off `FailuresOnly`; new tests for `--filter` (outcome / order_type / seq range), `--reverse`, and typed `InvalidArgument` rejection of `not` / non-audit filters. Filtered assertions wrapped in `Eventually` for the async index.
- `just pre-commit` green (0 lint issues); `go build ./...` + e2e build green.

## Impact analysis

Manual (GitNexus CLI not present in this worktree). Blast radius is contained and read-side: the `Controller.ListAuditEntries` signature change touches the interface, `DefaultController`, `RoutedController`, `BucketGrpcClient`, the gRPC handler, `pkg/actions`, and the regenerated mocks. `query.ReadAuditEntries` (used by checker/backup/indexer) is left untouched. No HIGH/CRITICAL risk; no HTTP endpoint (audit is gRPC-only, so `openapi.yml` untouched).

### DIFF
--- a/cmd/ledgerctl/audit/list.go
+++ b/cmd/ledgerctl/audit/list.go
@@ -27,8 +27,11 @@ func NewListCommand() *cobra.Command {
 
 Entries are shown oldest first by default (chronological); use --reverse for newest first.
 
-The --filter flag accepts audit[...] conditions combined with and/or:
+Filtering is done entirely through the generic --filter flag (same DSL as the
+other list commands); there are no audit-specific flags. It accepts audit[...]
+conditions combined with and/or:
   audit[outcome] == failure
+  audit[ledger] == main
   audit[ledger] == main and audit[order_type] == create_transaction
   audit[order_type] in (create_transaction, revert_transaction)
   audit[caller.subject] == "svc:payments"
@@ -43,21 +46,19 @@ Unsupported conditions (not, non-indexed fields) are rejected.
 
 Examples:
   ledgerctl audit list
-  ledgerctl audit list --failures-only
-  ledgerctl audit list --ledger my-ledger --reverse
+  ledgerctl audit list --reverse
   ledgerctl audit list --filter 'audit[outcome] == failure'
+  ledgerctl audit list --filter 'audit[ledger] == my-ledger'
   ledgerctl audit list --checkpoint-id 7`,
 		Args:              cobra.NoArgs,
 		ValidArgsFunction: cobra.NoFileCompletions,
 		RunE:              runList,
 	}
 
-	cmd.Flags().String("ledger", "", "Scope to audit entries touching this ledger")
 	cmdutil.AddPaginationFlags(cmd, cmdutil.PaginationOptions{SupportsReverse: true})
 	cmdutil.AddFilterFlags(cmd, cmdutil.FilterOptions{})
 	cmdutil.AddConsistencyFlags(cmd)
 	cmdutil.AddOutputFlags(cmd)
-	cmd.Flags().Bool("failures-only", false, "Show only failed entries (shorthand for --filter 'audit[outcome] == failure')")
 	cmd.Flags().Duration("timeout", cmdutil.DefaultTimeout, "Request timeout")
 	cmd.Flags().Bool("expand", false, "Expand orders within each audit entry")
 
@@ -75,9 +76,7 @@ func runList(cmd *cobra.Command, _ []string) error {
 	ctx, cancel := cmdutil.GetContext(cmd)
 	defer cancel()
 
-	failuresOnly, _ := cmd.Flags().GetBool("failures-only")
 	expand, _ := cmd.Flags().GetBool("expand")
-	ledger, _ := cmd.Flags().GetString("ledger")
 	pgn := cmdutil.GetPaginationFlags(cmd)
 	flt := cmdutil.GetFilterFlags(cmd)
 	cns := cmdutil.GetConsistencyFlags(cmd)
@@ -87,14 +86,7 @@ func runList(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	// --failures-only is a convenience shorthand for `audit[outcome] ==
-	// failure`, AND-combined with any explicit --filter.
-	if failuresOnly {
-		filter = combineAuditFilter(filter, auditOutcomeFailureFilter())
-	}
-
 	stream, err := client.ListAuditEntries(ctx, &servicepb.ListAuditEntriesRequest{
-		Ledger:  ledger,
 		Options: cmdutil.BuildListOptions(pgn, cns, filter),
 	})
 	if err != nil {
@@ -148,39 +140,6 @@ func runList(cmd *cobra.Command, _ []string) error {
 	cmdutil.EmitNextCursorHint(cmd, nextCursor)
 
 	return nil
-}
-
-// auditOutcomeFailureFilter builds the `audit[outcome] == failure` condition
-// that backs the --failures-only shorthand.
-func auditOutcomeFailureFilter() *commonpb.QueryFilter {
-	return &commonpb.QueryFilter{
-		Filter: &commonpb.QueryFilter_Audit{
-			Audit: &commonpb.AuditCondition{
-				Field: commonpb.AuditField_AUDIT_FIELD_OUTCOME,
-				Condition: &commonpb.AuditCondition_StringCond{
-					StringCond: &commonpb.StringCondition{
-						Value: &commonpb.StringCondition_Hardcoded{Hardcoded: "failure"},
-					},
-				},
-			},
-		},
-	}
-}
-
-// combineAuditFilter AND-combines two filters. Either may be nil.
-func combineAuditFilter(a, b *commonpb.QueryFilter) *commonpb.QueryFilter {
-	switch {
-	case a == nil:
-		return b
-	case b == nil:
-		return a
-	default:
-		return &commonpb.QueryFilter{
-			Filter: &commonpb.QueryFilter_And{
-				And: &commonpb.AndFilter{Filters: []*commonpb.QueryFilter{a, b}},
-			},
-		}
-	}
 }
 
 // printAuditEntry prints a single audit entry in a human-readable format.
--- a/cmd/ledgerctl/audit/list.go
+++ b/cmd/ledgerctl/audit/list.go
@@ -25,7 +25,7 @@ func NewListCommand() *cobra.Command {
 		Short:   "List audit entries",
 		Long: `List audit log entries (successes and failures) via gRPC streaming.
 
-Entries are shown newest first by default; use --reverse for oldest first.
+Entries are shown oldest first by default (chronological); use --reverse for newest first.
 
 The --filter flag accepts audit[...] conditions combined with and/or:
   audit[outcome] == failure
--- a/cmd/ledgerctl/audit/list.go
+++ b/cmd/ledgerctl/audit/list.go
@@ -1,6 +1,7 @@
 package audit
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
@@ -80,6 +81,16 @@ func runList(cmd *cobra.Command, _ []string) error {
 	pgn := cmdutil.GetPaginationFlags(cmd)
 	flt := cmdutil.GetFilterFlags(cmd)
 	cns := cmdutil.GetConsistencyFlags(cmd)
+
+	// --expand fetches each entry's items via GetAuditEntry, which has no
+	// checkpoint parameter and always reads the live store. Combined with
+	// --checkpoint-id the list page would come from the checkpoint while the
+	// expansion came from the live store — inconsistent, and a not-found error
+	// if the entry was purged from live but still present in the checkpoint.
+	// Refuse the combination explicitly rather than silently mixing sources.
+	if expand && cns.CheckpointID != 0 {
+		return errors.New("--expand cannot be combined with --checkpoint-id: audit entry expansion always reads the live store")
+	}
 
 	filter, err := cmdutil.BuildQueryFilter(flt.Expr, flt.Prefix)
 	if err != nil {
--- a/cmd/ledgerctl/audit/list.go
+++ b/cmd/ledgerctl/audit/list.go
@@ -20,19 +20,44 @@ import (
 // NewListCommand creates the audit list command.
 func NewListCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:               "list",
-		Aliases:           cmdutil.ListAliases,
-		Short:             "List audit entries",
-		Long:              "List audit log entries (successes and failures) via gRPC streaming",
+		Use:     "list",
+		Aliases: cmdutil.ListAliases,
+		Short:   "List audit entries",
+		Long: `List audit log entries (successes and failures) via gRPC streaming.
+
+Entries are shown newest first by default; use --reverse for oldest first.
+
+The --filter flag accepts audit[...] conditions combined with and/or:
+  audit[outcome] == failure
+  audit[ledger] == main and audit[order_type] == create_transaction
+  audit[order_type] in (create_transaction, revert_transaction)
+  audit[caller.subject] == "svc:payments"
+  audit[seq] between 1000 and 2000
+  audit[proposal_id] == 42
+  audit[timestamp] >= 1700000000000000    # unix microseconds
+  audit[log_seq] == 500
+
+Supported fields: seq, proposal_id, timestamp, log_seq (numeric);
+outcome (success|failure), ledger, caller.subject, order_type (string).
+Unsupported conditions (not, non-indexed fields) are rejected.
+
+Examples:
+  ledgerctl audit list
+  ledgerctl audit list --failures-only
+  ledgerctl audit list --ledger my-ledger --reverse
+  ledgerctl audit list --filter 'audit[outcome] == failure'
+  ledgerctl audit list --checkpoint-id 7`,
 		Args:              cobra.NoArgs,
 		ValidArgsFunction: cobra.NoFileCompletions,
 		RunE:              runList,
 	}
 
-	cmdutil.AddPaginationFlags(cmd, cmdutil.PaginationOptions{})
-	cmdutil.AddMinLogSequenceFlag(cmd)
+	cmd.Flags().String("ledger", "", "Scope to audit entries touching this ledger")
+	cmdutil.AddPaginationFlags(cmd, cmdutil.PaginationOptions{SupportsReverse: true})
+	cmdutil.AddFilterFlags(cmd, cmdutil.FilterOptions{})
+	cmdutil.AddConsistencyFlags(cmd)
 	cmdutil.AddOutputFlags(cmd)
-	cmd.Flags().Bool("failures-only", false, "Show only failed entries")
+	cmd.Flags().Bool("failures-only", false, "Show only failed entries (shorthand for --filter 'audit[outcome] == failure')")
 	cmd.Flags().Duration("timeout", cmdutil.DefaultTimeout, "Request timeout")
 	cmd.Flags().Bool("expand", false, "Expand orders within each audit entry")
 
@@ -52,12 +77,25 @@ func runList(cmd *cobra.Command, _ []string) error {
 
 	failuresOnly, _ := cmd.Flags().GetBool("failures-only")
 	expand, _ := cmd.Flags().GetBool("expand")
-	minLogSeq, _ := cmd.Flags().GetUint64("min-log-sequence")
+	ledger, _ := cmd.Flags().GetString("ledger")
 	pgn := cmdutil.GetPaginationFlags(cmd)
+	flt := cmdutil.GetFilterFlags(cmd)
+	cns := cmdutil.GetConsistencyFlags(cmd)
+
+	filter, err := cmdutil.BuildQueryFilter(flt.Expr, flt.Prefix)
+	if err != nil {
+		return err
+	}
+
+	// --failures-only is a convenience shorthand for `audit[outcome] ==
+	// failure`, AND-combined with any explicit --filter.
+	if failuresOnly {
+		filter = combineAuditFilter(filter, auditOutcomeFailureFilter())
+	}
 
 	stream, err := client.ListAuditEntries(ctx, &servicepb.ListAuditEntriesRequest{
-		Options:      cmdutil.BuildListOptions(pgn, cmdutil.ConsistencyFlags{MinLogSequence: minLogSeq}, nil),
-		FailuresOnly: failuresOnly,
+		Ledger:  ledger,
+		Options: cmdutil.BuildListOptions(pgn, cns, filter),
 	})
 	if err != nil {
 		return cmdutil.FormatGRPCError("failed to list audit entries", err)
@@ -110,6 +148,39 @@ func runList(cmd *cobra.Command, _ []string) error {
 	cmdutil.EmitNextCursorHint(cmd, nextCursor)
 
 	return nil
+}
+
+// auditOutcomeFailureFilter builds the `audit[outcome] == failure` condition
+// that backs the --failures-only shorthand.
+func auditOutcomeFailureFilter() *commonpb.QueryFilter {
+	return &commonpb.QueryFilter{
+		Filter: &commonpb.QueryFilter_Audit{
+			Audit: &commonpb.AuditCondition{
+				Field: commonpb.AuditField_AUDIT_FIELD_OUTCOME,
+				Condition: &commonpb.AuditCondition_StringCond{
+					StringCond: &commonpb.StringCondition{
+						Value: &commonpb.StringCondition_Hardcoded{Hardcoded: "failure"},
+					},
+				},
+			},
+		},
+	}
+}
+
+// combineAuditFilter AND-combines two filters. Either may be nil.
+func combineAuditFilter(a, b *commonpb.QueryFilter) *commonpb.QueryFilter {
+	switch {
+	case a == nil:
+		return b
+	case b == nil:
+		return a
+	default:
+		return &commonpb.QueryFilter{
+			Filter: &commonpb.QueryFilter_And{
+				And: &commonpb.AndFilter{Filters: []*commonpb.QueryFilter{a, b}},
+			},
+		}
+	}
 }
 
 // printAuditEntry prints a single audit entry in a human-readable format.
--- a/cmd/ledgerctl/audit/list.go
+++ b/cmd/ledgerctl/audit/list.go
@@ -35,14 +35,14 @@ conditions combined with and/or:
   audit[ledger] == main
   audit[ledger] == main and audit[order_type] == create_transaction
   audit[order_type] in (create_transaction, revert_transaction)
-  audit[caller.subject] == "svc:payments"
+  audit[caller_subject] == "svc:payments"
   audit[seq] between 1000 and 2000
   audit[proposal_id] == 42
   audit[timestamp] >= 1700000000000000    # unix microseconds
   audit[log_seq] == 500
 
 Supported fields: seq, proposal_id, timestamp, log_seq (numeric);
-outcome (success|failure), ledger, caller.subject, order_type (string).
+outcome (success|failure), ledger, caller_subject, order_type (string).
 Unsupported conditions (not, non-indexed fields) are rejected.
 
 Examples:
--- a/cmd/ledgerctl/audit/list.go
+++ b/cmd/ledgerctl/audit/list.go
@@ -38,11 +38,12 @@ conditions combined with and/or:
   audit[caller_subject] == "svc:payments"
   audit[seq] between 1000 and 2000
   audit[proposal_id] == 42
-  audit[timestamp] >= 1700000000000000    # unix microseconds
+  audit[timestamp] >= "2023-11-14T22:13:20Z"    # RFC3339 or raw unix microseconds
   audit[log_seq] == 500
 
-Supported fields: seq, proposal_id, timestamp, log_seq (numeric);
-outcome (success|failure), ledger, caller_subject, order_type (string).
+Supported fields: seq, proposal_id, log_seq (numeric); timestamp
+(RFC3339 or unix microseconds); outcome (success|failure), ledger,
+caller_subject, order_type (string).
 Unsupported conditions (not, non-indexed fields) are rejected.
 
 Examples:
--- a/internal/adapter/grpc/server_bucket.go
+++ b/internal/adapter/grpc/server_bucket.go
@@ -569,6 +569,55 @@ func (impl *BucketServiceServerImpl) waitMinLogSequence(ctx context.Context, min
 	return impl.readStore.WaitForSequence(ctx, minLogSequence)
 }
 
+// waitFilteredAuditConsistency gives a live, filtered ListAuditEntries read a
+// read-your-writes guarantee against the async audit secondary index.
+//
+// min_log_sequence is a LOG sequence, but a filtered audit read resolves through
+// the audit index whose cursor (ReadAuditProgress) is an AUDIT sequence. The two
+// spaces diverge — the audit sequence advances on every proposal, including
+// failures that emit no log — so waiting for audit_progress >= min_log_sequence
+// would be incorrect. Instead, conservatively:
+//
+//  1. wait for the log index to reach min_log_sequence (waitMinLogSequence);
+//  2. read the current live audit head from the main store;
+//  3. wait for the audit index cursor to reach that observed head.
+//
+// This can over-wait (the head observed in step 2 may include audit entries
+// produced after min_log_sequence), which is acceptable for v3.0 and avoids a
+// new min_audit_sequence API. A minLogSeq of 0 means "no consistency bound
+// requested", so this is a no-op.
+//
+// Only the LIVE filtered path calls this. Unfiltered reads scan the Cold/Audit
+// zone directly and are always current, so they keep the direct fast path.
+// Checkpoint reads are frozen snapshots and are handled (and documented) at the
+// checkpoint-creation boundary, out of scope here.
+func (impl *BucketServiceServerImpl) waitFilteredAuditConsistency(ctx context.Context, minLogSeq uint64) error {
+	if minLogSeq == 0 {
+		return nil
+	}
+
+	if err := impl.waitMinLogSequence(ctx, minLogSeq); err != nil {
+		return err
+	}
+
+	handle, err := impl.store.NewDirectReadHandle()
+	if err != nil {
+		return status.Errorf(codes.Unavailable, "opening read handle for audit head: %v", err)
+	}
+
+	auditHead, err := query.ReadLastAuditSequence(handle)
+	// Close before waiting: the handle holds dbMu.RLock for its lifetime, and the
+	// wait below can block for a while. We only need the head value, not the
+	// handle, past this point.
+	_ = handle.Close()
+
+	if err != nil {
+		return status.Errorf(codes.Internal, "reading live audit head: %v", err)
+	}
+
+	return impl.readStore.WaitForAuditSequence(ctx, auditHead)
+}
+
 func (impl *BucketServiceServerImpl) ListTransactions(req *servicepb.ListTransactionsRequest, stream servicepb.BucketService_ListTransactionsServer) error {
 	ctx, span := bucketTracer.Start(stream.Context(), "grpc.ListTransactions",
 		trace.WithAttributes(attribute.String("ledger", req.GetLedger())))
@@ -1241,7 +1290,19 @@ func (impl *BucketServiceServerImpl) ListAuditEntries(req *servicepb.ListAuditEn
 		// log-index WaitForSequence); it is out of scope here.
 		c, err = impl.localCtrl.ListAuditEntriesFrom(ctx, mainStore, readIdx, fetchSize, afterSeq, opts.GetFilter(), opts.GetReverse())
 	} else {
-		if waitErr := impl.waitMinLogSequence(ctx, opts.GetRead().GetMinLogSequence()); waitErr != nil {
+		minLogSeq := opts.GetRead().GetMinLogSequence()
+
+		// A filtered live read resolves through the async audit secondary index,
+		// which lags the audit zone independently of the log index. Gate it on the
+		// audit-index progress so the requested consistency bound actually covers
+		// the index this read consults. An unfiltered read scans the Cold/Audit
+		// zone directly and is always current, so it keeps the plain log-index wait
+		// (its fast path is unchanged).
+		if opts.GetFilter() != nil {
+			if waitErr := impl.waitFilteredAuditConsistency(ctx, minLogSeq); waitErr != nil {
+				return waitErr
+			}
+		} else if waitErr := impl.waitMinLogSequence(ctx, minLogSeq); waitErr != nil {
 			return waitErr
 		}
 
--- a/internal/adapter/grpc/server_bucket.go
+++ b/internal/adapter/grpc/server_bucket.go
@@ -1219,7 +1219,7 @@ func (impl *BucketServiceServerImpl) ListAuditEntries(req *servicepb.ListAuditEn
 	var c cursor.Cursor[*auditpb.AuditEntry]
 
 	if cpID := opts.GetRead().GetCheckpointId(); cpID > 0 {
-		mainStore, readIdx, openErr := impl.openCheckpointStores(cpID)
+		mainStore, readIdx, openErr := impl.openCheckpointStores(ctx, cpID)
 		if openErr != nil {
 			return openErr
 		}
--- a/internal/adapter/grpc/server_bucket.go
+++ b/internal/adapter/grpc/server_bucket.go
@@ -1229,6 +1229,16 @@ func (impl *BucketServiceServerImpl) ListAuditEntries(req *servicepb.ListAuditEn
 			_ = mainStore.Close()
 		}()
 
+		// Known limitation (tracked as a follow-up): CreateQueryCheckpoint waits
+		// for the log-index builder (readStore.WaitForSequence) before snapshotting
+		// the readstore, but NOT for the separate async audit indexer. If the
+		// audit index lagged its zone at snapshot time, a *filtered* checkpoint
+		// read can omit audit entries that do exist in the checkpoint's audit
+		// zone, and — the checkpoint being frozen — it never catches up. The
+		// unfiltered checkpoint read is unaffected (it scans the zone directly).
+		// The proper fix belongs in the checkpoint-creation path (make the audit
+		// indexer catch up before the readstore checkpoint, mirroring the
+		// log-index WaitForSequence); it is out of scope here.
 		c, err = impl.localCtrl.ListAuditEntriesFrom(ctx, mainStore, readIdx, req.GetLedger(), fetchSize, afterSeq, opts.GetFilter(), opts.GetReverse())
 	} else {
 		if waitErr := impl.waitMinLogSequence(ctx, opts.GetRead().GetMinLogSequence()); waitErr != nil {
--- a/internal/adapter/grpc/server_bucket.go
+++ b/internal/adapter/grpc/server_bucket.go
@@ -1202,13 +1202,9 @@ func (impl *BucketServiceServerImpl) ListAuditEntries(req *servicepb.ListAuditEn
 
 	opts := req.GetOptions()
 
-	// The audit controller does not yet honor filter / reverse /
-	// checkpoint_id — see #436 for the alignment work.
-	if err := ValidateListOptions(opts, ListOptionsSupport{}); err != nil {
-		return err
-	}
-
-	if err := impl.waitMinLogSequence(ctx, opts.GetRead().GetMinLogSequence()); err != nil {
+	// Audit listing is fully under the shared contract: filter (audit[...]
+	// conditions), reverse, and checkpoint_id are all honored (EN-1241).
+	if err := ValidateListOptions(opts, ListOptionsSupport{Filter: true, Reverse: true, CheckpointID: true}); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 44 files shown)


internal/query/audit.go
```diff
@@ -66,6 +68,192 @@ func ReadAuditEntries(ctx context.Context, reader dal.PebbleReader, afterSequenc
...
+
+				continue
+			}
+
+			break
+		}
+		filtered = filtered[idx:]
+	}
+
+	if pageSize > 0 && uint32(len(filtered)) > pageSize {
```
NumaryBot: 🟠 **[major] Avoid truncating before skipping purged audit seqs**

When a filtered audit query hits dangling index entries after an audit purge, this truncates the candidate sequence list before `ReadAuditEntry` drops missing rows. If one of the first `pageSize`/peek candidates is purged, the cursor can return only `pageSize` real entries (or fewer) and `sendPagedToStream` will observe EOF and omit `x-next-cursor` even though later non-purged candidates still match, causing filtered pagination after purge to terminate early and miss entries.

pkg/actions/read.go
```diff
@@ -329,11 +329,38 @@ func AggregateVolumes(ctx context.Context, client servicepb.BucketServiceClient,
...
-// ListAuditEntries collects all audit entries from the streaming RPC.
+// ListAuditEntries collects all audit entries from the streaming RPC. When
+// failuresOnly is true it applies the shared filter `audit[outcome] == failure`
+// (failures_only is no longer a top-level request field — EN-1241).
 func ListAuditEntries(ctx context.Context, client servicepb.BucketServiceClient, failuresOnly bool) ([]*auditpb.AuditEntry, error) {
-	return ListAuditEntriesWithRequest(ctx, client, &servicepb.ListAuditEntriesRequest{
-		FailuresOnly: failuresOnly,
-	})
+	req := &servicepb.ListAuditEntriesRequest{}
+	if failuresOnly {
```
NumaryBot, line 345: 🟠 **[major] Keep failures helper from racing the async audit index**

When `failuresOnly` is true this helper now uses the server-side `audit[outcome]` filter, which is backed by the async audit index and is not made fresh by `min_log_sequence`. Existing callers such as `tests/scenarios/operations_lifecycle` call `actions.ListAuditEntries(ctx, client, true)` immediately after creating a failed transaction and assert a failure is present; during index lag this can return an empty list even though the audit entry is already in the audit zone. Either keep this helper's legacy strong behavior by filtering client-side over the unfiltered stream, or make the affected callers wait for the async index.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

tests/e2e/business/audit_test.go
```diff
@@ -437,6 +463,132 @@ var _ = Describe("Audit Log", Ordered, func() {
...
+		Expect(len(asc)).To(BeNumerically(">=", 2))
+
+		desc, err := collectAuditEntries(sharedCtx, sharedClient, &servicepb.ListAuditEntriesRequest{
+			Options: &commonpb.ListOptions{Reverse: true},
+		})
+		Expect(err).To(Succeed())
+		Expect(len(desc)).To(Equal(len(asc)))
+
+		// Ascending is oldest-first (increasing sequence); reverse is newest-first.
+		Expect(asc[0].GetSequence()).To(BeNumerically("<", asc[len(asc)-1].GetSequence()))
```
gfyrag, line 525: This assertion is inverted relative to the implementation and the CLI docs added in this PR. `DefaultController.ListAuditEntries` documents `reverse=false` as newest-first and passes `!reverse` into `ReadAuditEntriesPage`, so `Options{Reverse:false}` returns descending sequences. The docs also say `ledgerctl audit list` is newest-first by default and `--reverse` is oldest-first.

As written, this e2e expects `Reverse:false` to be ascending (`asc[0] < asc[last]`) and `Reverse:true` to be descending, so it should fail once the Ginkgo suite runs. The production code looks coherent with the shared transaction convention; the test should be flipped/renamed so default is descending and `Reverse:true` is ascending.

gfyrag (commented):
Reviewed `35371f29e9d7e97955127821138741d417c9da84` for the audit list shared `ListOptions` contract.

I focused on the proto/API shape, controller and gRPC pagination semantics, audit filter compilation, readstore audit index access, CLI/docs updates, and the e2e coverage. I found one issue in the new e2e assertion below.

Validation:

```bash
GOCACHE=/tmp/go-build-pr1548 go test ./internal/query/... ./internal/storage/readstore/... ./internal/pkg/filterexpr/... ./internal/application/ctrl/... ./cmd/ledgerctl/audit/... ./pkg/actions/...
GOCACHE=/tmp/go-build-pr1548 go test ./internal/adapter/grpc/...
```

The non-network packages passed in the sandbox; the gRPC package passed outside the sandbox because it binds localhost.

internal/query/audit.go
```diff
@@ -66,6 +68,214 @@ func ReadAuditEntries(ctx context.Context, reader dal.PebbleReader, afterSequenc
...
+		slices.Reverse(filtered)
+	}
+
+	// Apply the exclusive cursor.
+	if afterSeq != 0 {
+		idx := 0
+		for idx < len(filtered) {
+			s := filtered[idx]
+			if (!reverse && s <= afterSeq) || (reverse && s >= afterSeq) {
+				idx++
```
NumaryBot, line 156: 🟠 **[major] Filtered pagination truncates before dropping purged rows, causing early cursor termination**

When a filtered audit query encounters dangling index entries after an audit purge, the candidate sequence list is truncated to `pageSize` before `ReadAuditEntry` can drop missing rows. If any of those first candidates are purged, the page returned has fewer than `pageSize` real entries, `sendPagedToStream` sees EOF, and omits `x-next-cursor` even though later non-purged candidates still match. This causes filtered pagination after purge to terminate early and silently miss results.

**Suggestion:** Fetch a larger peek window (e.g., oversize by a configurable factor) before applying the index-driven truncation, or apply the pageSize limit only after ReadAuditEntry has filtered out purged/missing rows.

pkg/actions/read.go
```diff
@@ -329,11 +329,46 @@ func AggregateVolumes(ctx context.Context, client servicepb.BucketServiceClient,
...
+// eventually consistent — a just-applied failure may take a short moment to
+// appear. Callers that assert on a freshly-applied entry must poll (e.g.
+// require.Eventually / Gomega Eventually) rather than assume immediate
+// visibility; do not add time.Sleep.
 func ListAuditEntries(ctx context.Context, client servicepb.BucketServiceClient, failuresOnly bool) ([]*auditpb.AuditEntry, error) {
-	return ListAuditEntriesWithRequest(ctx, client, &servicepb.ListAuditEntriesRequest{
-		FailuresOnly: failuresOnly,
-	})
+	req := &servicepb.ListAuditEntriesRequest{}
+	if failuresOnly {
```
NumaryBot, line 345: 🟠 **[major] Failures helper now depends on async audit index, introducing a race in existing scenario tests**

When `failuresOnly` is true the helper now uses the server-side `audit[outcome]` filter, which is backed by an async index and is not guaranteed fresh relative to `min_log_sequence`. Existing callers such as `tests/scenarios/operations_lifecycle` call `actions.ListAuditEntries(ctx, client, true)` immediately after creating a failed transaction and assert a failure is present; during index lag the list can be empty even though the audit entry already exists in the audit zone, causing a flaky or consistently failing assertion.

**Suggestion:** Either preserve the legacy strong-consistency behavior by fetching the unfiltered stream and filtering client-side for failures, or update all affected callers to wait/poll until the async index reflects the expected entry before asserting.

tests/e2e/business/audit_test.go
NumaryBot, line 525: 🟠 **[major] Reverse-sort assertions in audit e2e test are inverted relative to the implementation**

`DefaultController.ListAuditEntries` documents and implements `reverse=false` as newest-first (descending sequences), passing `!reverse` into `ReadAuditEntriesPage`. The e2e test asserts that `Reverse:false` produces ascending order and `Reverse:true` produces descending order — the opposite of the actual behavior and CLI docs added in this PR. The test is expected to fail once the Ginkgo suite runs against the real controller.

**Suggestion:** Flip the test assertions and rename variables so that the default (`Reverse:false`) is asserted to be descending (newest-first) and `Reverse:true` is asserted to be ascending (oldest-first), matching both the implementation and the updated CLI documentation.

internal/query/audit_filter.go
```diff
@@ -0,0 +1,358 @@
...
+		}
+
+		if first {
+			acc = c
+			first = false
+
+			continue
+		}
+
+		acc = auditCompiled{seqs: unionSorted(acc.seqs, c.seqs), narrowed: true, loSeq: 0, hiSeq: math.MaxUint64}
```
NumaryBot, line 299: 🟠 **[major] OR branches discard per-branch sequence bounds, allowing out-of-predicate entries**

When an `or` expression combines index-narrowed branches that also carry branch-local `audit[seq]` bounds, the union resets `loSeq`/`hiSeq` to the full range, discarding those bounds. For example, `(audit[outcome] == failure and audit[seq] < 10) or audit[ledger] == main` will include every failure from the outcome index rather than only failures below sequence 10, so the query returns entries that do not satisfy the original filter predicate.

**Suggestion:** When unioning the sequence ranges of OR branches, intersect each branch's local sequence bounds with the global bounds before taking the union, or apply a post-fetch predicate check to enforce branch-local constraints that cannot be fully expressed in the index range.

internal/pkg/filterexpr/parser.go
```diff
@@ -49,7 +49,7 @@ var filterLexer = lexer.MustSimple([]lexer.SimpleRule{
 	{Name: "Dollar", Pattern: `\$`},
 	{Name: "String", Pattern: `"[^"]*"|'[^']*'`},
 	{Name: "Comma", Pattern: `,`},
-	{Name: "Keyword", Pattern: `\b(and|or|not|in|between|metadata|address|source|destination|ledger|exists|true|false)\b`},
+	{Name: "Keyword", Pattern: `\b(and|or|not|in|between|metadata|address|source|destination|ledger|audit|exists|true|false)\b`},
```
NumaryBot, line 52: 🟠 **[major] Adding `audit` as a keyword token breaks existing bare-value filters**

Registering `audit` as a global keyword means it is no longer accepted by the `Value` production (which only matches `$param`, quoted strings, numbers, booleans, or `Ident`). Existing shared filters such as `metadata[type] == audit`, `address == audit`, or `ledger == audit` now fail to parse unless callers add quotes, even though `audit` only needs to introduce the new condition syntax, not be reserved as a value token.

**Suggestion:** Scope `audit` as a contextual keyword rather than a global reserved token — accept it as a valid `Ident` value on the right-hand side of comparisons, and only treat it as introducing the new condition syntax when it appears in the left-hand (field) position.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

gfyrag (commented):
Reviewed the new commit (`35371f29..2179be4a`). No findings.

Scope checked: audit listing now consistently uses chronological order as the endpoint default, `--reverse`/`ListOptions.reverse` inverts that default, filtered audit pagination no longer lets purged indexed sequences consume page slots, and the async audit-index assertion in the scenario test now polls instead of assuming immediate visibility.

Validation: `GOCACHE=/tmp/go-build-pr1548 go test ./internal/query/... ./internal/application/ctrl/... ./pkg/actions/...` passed.

gfyrag:
## Response to automated review (5 findings)

Verdicts after re-checking each against the current tip (`2179be4a9` + this follow-up):

**1 & 2 — audit.go:156 "truncate before skipping purged" (b6543fd2 / 3b49d21b): FIXED in `2179be4a9` (before this review ran on those same lines).** `readAuditPageFromSeqSet` now materializes candidates and caps on *valid* entries **after** the `ErrNotFound` purge-skip, never before — a purged seq no longer consumes a page slot, so the page fills to `pageSize` and the peek-ahead `x-next-cursor` fires correctly. Covered by `TestReadAuditEntriesPage_SeqSetPurgeDoesNotShortenPage`.

**3 & 4 — read.go:345 "failures helper races async index" (04787232 / 767c2cf0): FIXED in `2179be4a9`.** The scenario caller (`operations_lifecycle`) now wraps the failures-only assertion in `require.Eventually`, and the helper documents that `failuresOnly=true` reads the eventually-consistent audit index. `Tests-Scenarios` is green on this commit. (Kept the server-side filter rather than reverting to client-side filtering — that's the whole point of the index-backed contract; the fix is to poll, per the repo's no-`time.Sleep` rule.)

**5 — audit_test.go:528 "reverse assertions inverted" (5a41f1f6): STALE — REBUTTED.** The finding quotes the pre-fix implementation (`!reverse`, newest-first default). `2179be4a9` reverted the default to **ascending/chronological** (`reverse` passed through directly), matching the pre-`ListOptions` behavior; the e2e test asserts exactly that, and the CLI docs were updated to "oldest first by default". Empirical proof: **`Tests-E2E` is green on this commit** (all 22 audit specs pass). No change needed.

**6 — audit_filter.go OR branch bounds (0259e95b): VALID — FIXED in this follow-up.** `(audit[outcome]==failure and audit[seq]<10) or audit[ledger]==main` did carry the AND branch's `seq<10` on `hiSeq`, which the OR union reset to full → leak. Fix: `intersectAudit` now **bakes** a narrowed result's `[loSeq,hiSeq]` window into its seq set (window reset to full) before it can be unioned, so per-branch bounds survive OR. Covered by `TestCompileAuditFilter_OrDoesNotLeakBranchSeqBound` + `TestCompileAuditFilter_AndBakesSeqBoundIntoSeqs`.

**7 — parser.go:52 "audit keyword breaks bare values" (f5c5d345): VALID (pre-existing class) — FIXED in this follow-up.** `metadata[type] == audit` didn't parse as a bare value — but note this was already true for `ledger`/`source`/`destination`/etc. before this PR (all keywords). Fixed the whole class: the `Value` production now accepts the "noun" keywords (`metadata/address/source/destination/ledger/audit/exists`) as unquoted values, while structural operators (`and/or/not/in/between`) stay reserved so they keep terminating expressions. Covered by `TestParseAudit_KeywordAsBareValue`.

internal/adapter/grpc/server_bucket.go
```diff
@@ -1112,13 +1108,30 @@ func (impl *BucketServiceServerImpl) ListAuditEntries(req *servicepb.ListAuditEn
...
+		if openErr != nil {
+			return openErr
+		}
+
+		defer func() {
+			_ = readIdx.Close()
+			_ = mainStore.Close()
+		}()
+
+		c, err = impl.localCtrl.ListAuditEntriesFrom(ctx, mainStore, readIdx, req.GetLedger(), fetchSize, afterSeq, opts.GetFilter(), opts.GetReverse())
```
NumaryBot: ⚪ **[nit] Stale async audit index when serving checkpointed filtered audit queries**

When `checkpoint_id` is combined with `options.filter` or a top-level `ledger`, the filter is compiled against the checkpointed read-store audit index. Because the audit index is maintained asynchronously, if it was behind at checkpoint creation time, the checkpoint's index will never catch up, meaning filtered checkpoint reads can permanently omit audit entries that actually exist in the checkpoint's audit zone.

**Suggestion:** Ensure the audit index is fully caught up before snapshotting a checkpoint, or fall back to a linear scan / reject filtered audit queries against checkpoints whose audit index sequence lags the checkpoint's audit-zone high-watermark.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

gfyrag (commented):
Review du nouveau commit `1d8ccb619a6bb27d4c5fd4760fbd3af103dbd5dc`: pas de finding.

J'ai vérifié en particulier le changement de parsing des RHS "noun" keywords et la correction sur les bornes `seq` intégrées aux sets audit déjà réduits, pour éviter que des bornes locales fuient sous un `OR` englobant. Ça respecte le comportement attendu des filtres sans élargir silencieusement le résultat.

Vérification lancée :
`GOCACHE=/tmp/go-build-pr1548b go test ./internal/pkg/filterexpr/... ./internal/query/...`

gfyrag:
Re the remaining ⚪ nit (server_bucket.go — stale audit index in checkpointed filtered reads): **valid, acknowledged, documented, deferred.** Confirmed against the code: `CreateQueryCheckpoint` waits for the log-index builder (`readStore.WaitForSequence`) before the readstore snapshot, but not for the separate async `auditindexer` worker, so a frozen checkpoint can carry a lagging audit index and a *filtered* checkpoint read may omit entries present in the checkpoint's audit zone. Unfiltered checkpoint reads scan the zone directly and are unaffected. The correct fix (make the audit indexer catch up before the readstore checkpoint, mirroring the log-index `WaitForSequence`) lives in the checkpoint-creation path and is out of scope for EN-1241 — I've documented the limitation in the handler and in cli.md rather than silently shipping a trap, and flagged it as a tracked follow-up. It's a non-blocking nit and the check is green.

gfyrag (commented):
Review du nouveau commit `6d99dda66cdee105409b82e7dad9e8df93efb05d`: pas de finding.

Le changement documente explicitement la limite `--checkpoint-id` + filtre sur l'audit index : le checkpoint fige le read index au moment de création, alors que l'audit indexer est asynchrone et peut être en retard. Le commentaire côté `ListAuditEntries` correspond bien au comportement de production actuel et ne modifie pas l'exécution.

Je n'ai pas relancé de tests pour ce commit : il ne change que `docs/ops/cli.md` et un commentaire dans `internal/adapter/grpc/server_bucket.go`.

internal/adapter/grpc/server_bucket.go
```diff
@@ -1112,13 +1108,40 @@ func (impl *BucketServiceServerImpl) ListAuditEntries(req *servicepb.ListAuditEn
...
+		// for the log-index builder (readStore.WaitForSequence) before snapshotting
+		// the readstore, but NOT for the separate async audit indexer. If the
+		// audit index lagged its zone at snapshot time, a *filtered* checkpoint
+		// read can omit audit entries that do exist in the checkpoint's audit
+		// zone, and — the checkpoint being frozen — it never catches up. The
+		// unfiltered checkpoint read is unaffected (it scans the zone directly).
+		// The proper fix belongs in the checkpoint-creation path (make the audit
+		// indexer catch up before the readstore checkpoint, mirroring the
+		// log-index WaitForSequence); it is out of scope here.
+		c, err = impl.localCtrl.ListAuditEntriesFrom(ctx, mainStore, readIdx, req.GetLedger(), fetchSize, afterSeq, opts.GetFilter(), opts.GetReverse())
```
NumaryBot: 🟡 **[minor] Stale async audit index when serving checkpointed filtered audit queries**

When `checkpoint_id` is combined with `options.filter` or a top-level `ledger`, the filter is evaluated against the checkpointed read-store audit index. Because the audit indexer runs asynchronously and `CreateQueryCheckpoint` does not wait for it to catch up (unlike the log-index `WaitForSequence` path), a checkpoint created while the index is behind will carry a permanently frozen, lagging index. Filtered checkpoint reads may therefore omit audit entries that genuinely belong to that checkpoint's audit zone — a silent data loss from the caller's perspective.

**Suggestion:** Before snapshotting the readstore for a checkpoint, call the equivalent of `WaitForSequence` on the async audit indexer (mirroring the existing log-index drain), so the index is current at the point the snapshot is taken. Alternatively, fall back to a linear zone scan or reject filtered audit queries against checkpoints whose audit-index high-watermark lags the checkpoint's audit-zone boundary.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

misc/proto/bucket.proto
```diff
@@ -757,11 +757,18 @@ message CheckStoreProgress {
...
-  common.ListOptions options = 1;
-  // failures_only filters audit entries to only those carrying a failure status.
-  bool failures_only = 2;
-  // ledger filters audit entries to only those containing orders targeting this ledger.
-  string ledger = 3;
+  // options carries the whole read contract. Audit has NO dedicated top-level
+  // filters: ledger scope and outcome selection are expressed through
+  // options.filter (e.g. `audit[ledger] == <name>`, `audit[outcome] ==
+  // failure`), exactly like every other filtered list endpoint.
+  common.ListOptions options = 2;
```
NumaryBot: 🔴 **[blocker] Field tag change in ListAuditEntriesRequest breaks wire compatibility**

The `options` field in `ListAuditEntriesRequest` was moved to a new tag number (tag 2), while the previous occupant of tag 2 was `failures_only`. Clients compiled against the previous proto send `options` on tag 1; the updated server treats that payload as an unknown field and silently drops all list options (page size, cursor, min_log_sequence, etc.). Any client already in the field that sends `failures_only` on tag 2 will have that value interpreted as `options` by the new server. Both directions of the version mismatch produce silent, incorrect behavior with no error surfaced to callers.

**Suggestion:** Keep `options` on its original field tag (tag 1) and keep `failures_only` on its original tag as well, then mark both original tags as `reserved` only if the fields are truly being removed. Do not renumber existing non-reserved fields in a public proto; instead, introduce any new field on a previously unused tag number.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

gfyrag:
Re the 🔴 blocker (bucket.proto — field tag change in ListAuditEntriesRequest): **adopted the suggestion.** `options` now keeps its **original tag 1** (the release/v3.0 shape was `{ options=1, failures_only=2, ledger=3 }`), and tags 2 (failures_only) + 3 (ledger) are reserved. So no live field is renumbered — a client sending `options` on tag 1 is still understood; the only wire change is the removal of the two dedicated fields, which is the intended breaking change for v3.0 (alpha compat explicitly not required, and this PR series is coordinated breaking proto evolution — #1547 also renumbers QueryFilter). Regenerated proto, build + audit unit/e2e/scenario tests green, Dirty clean.

internal/query/audit_filter.go
```diff
@@ -0,0 +1,387 @@
...
+// scan-time predicate fallback: the audit trail is queried exclusively through
+// its access paths, so an expression the index cannot answer is refused rather
+// than silently degrading to a full-chain scan (EN-1241 / invariant: audit is
+// the source of truth, projections are the only query surface).
+func CompileAuditFilter(idx AuditIndexReader, filter *commonpb.QueryFilter) (seqs []uint64, loSeq, hiSeq uint64, narrowed bool, err error) {
+	if filter == nil {
+		return nil, 0, math.MaxUint64, false, nil
+	}
+
+	c, err := compileAuditNode(idx, filter)
```
NumaryBot: 🟠 **[major] No recursion-depth guard in audit filter compiler**

The new `compileAuditNode` recursive function that compiles `options.filter` proto trees for `ListAuditEntries` has no maximum-depth limit. Unlike the shared `query.Compile` path which already carries a depth guard, a malicious or buggy gRPC client can submit deeply nested `and`/`or` filter nodes and drive unbounded recursion, crashing the process with a Go stack overflow rather than receiving an `InvalidArgument` error.

**Suggestion:** Add the same kind of maximum-depth check used by the existing filter compiler. Pass a depth counter through the recursive calls and return an `InvalidArgument` status error once the depth exceeds the configured maximum (e.g., the same constant used by the shared compiler).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

gfyrag:
Re the 🟠 major (audit_filter.go — no recursion-depth guard): **valid, fixed.** `compileAuditNode` now threads a depth counter through the and/or recursion and returns `InvalidArgument` once nesting reaches `MaxFilterDepth` (the same constant the shared `query.Compile` uses), so a deeply-nested filter proto is rejected instead of overflowing the stack. Covered by `TestCompileAuditFilter_RejectsTooDeep` and `TestCompileAuditFilter_AcceptsAtDepthLimit`.

gfyrag (commented):
Reviewed the latest head `9f2679781f596a70afed5313ec194b99554c67ac` for the audit-list contract changes.

No blocking findings. I rechecked the proto tag preservation, removal of dedicated audit filters, server-side ListOptions support, audit filter compilation, sequence-bound baking before OR unions, pagination/cursor behavior, and the documented checkpoint caveat.

Checked:
- `go test ./internal/query ./internal/application/ctrl ./cmd/ledgerctl/audit`
- `git diff --check origin/release/v3.0...origin/pr/1548`

gfyrag (commented):
Reviewed the new head `0f3c335fb95a6552203201524883348718ed01d7`.

No further findings. The commit adds the missing recursion-depth guard to the audit filter compiler, threads the depth through both `and` and `or`, returns `InvalidArgument` at `MaxFilterDepth`, and covers both the too-deep and accepted-at-limit cases.

Checked:
- `go test ./internal/query`
- `git diff --check 9f2679781f596a70afed5313ec194b99554c67ac..origin/pr/1548`

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

misc/proto/common.proto
```diff
@@ -1395,6 +1431,7 @@ enum QueryTarget {
   QUERY_TARGET_ACCOUNTS = 0;
   QUERY_TARGET_TRANSACTIONS = 1;
   QUERY_TARGET_LOGS = 2;
+  QUERY_TARGET_AUDIT = 3;
```
Azorlogh: `QUERY_TARGET_AUDIT` isn't referenced anywhere in Go — the audit path calls `CompileAuditFilter` directly and never dispatches on `QueryTarget`, and prepared queries can't target it (`parseTarget` rejects `audit`). Dead enum value; drop it unless prepared-queries-over-audit is planned.

tests/e2e/business/audit_test.go
```diff
@@ -28,6 +28,27 @@ func collectAuditEntries(ctx context.Context, client servicepb.BucketServiceClie
 	return actions.ListAuditEntriesWithRequest(ctx, client, req)
 }
 
+// auditFilter builds a single-condition audit QueryFilter.
```
Azorlogh: Stale name: comment says `auditFilter`, function is `auditStringFilter`.

internal/query/audit.go
```diff
@@ -66,6 +68,214 @@ func ReadAuditEntries(ctx context.Context, reader dal.PebbleReader, afterSequenc
...
+	pageSize uint32,
+) (cursor.Cursor[*auditpb.AuditEntry], error) {
+	// Lower bound (inclusive): max(loSeq, afterSeq+1 in ascending).
+	lo := loSeq
+	hi := hiSeq
+
+	if !reverse && afterSeq != 0 && afterSeq+1 > lo {
+		lo = afterSeq + 1
+	}
+	if reverse && afterSeq != 0 && afterSeq >= 1 && afterSeq-1 < hi {
```
Azorlogh: `afterSeq >= 1` is implied by `afterSeq != 0` (unsigned). Minor: the zone `entries` slice (l.239) also skips the `pageCap` capacity hint the seq-set path uses.

Azorlogh (commented):
A few small nits from a re-review.

misc/proto/common.proto
```diff
@@ -1395,6 +1432,7 @@ enum QueryTarget {
...
```
NumaryBot: 🟠 **[major] QUERY_TARGET_AUDIT exposed before prepared-query execution supports it**

Adding `QUERY_TARGET_AUDIT` makes it a valid value for `PreparedQuery.target`, but the prepared-query execution path still only handles accounts/transactions/logs. `query.Compile` returns an empty universe for unknown targets and rejects `QueryFilter_Audit`, and the CLI target parsing/formatting has no audit case. A client that creates a prepared query with this new target will get empty results or compile errors silently.

**Suggestion:** Either wire audit execution through the prepared-query path (Compile, CLI parsing/formatting) before exposing this enum value, or hold the enum addition in a follow-up PR until the full prepared-query audit path is implemented.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

misc/proto/common.proto
gfyrag: Good catch — removed. `QUERY_TARGET_AUDIT` was dead: the audit list path calls `query.CompileAuditFilter` directly and never dispatches on `QueryTarget`, and prepared queries can't target audit. Dropped the enum value and reserved 3 (name + number) in f12b5582a; regenerated the proto.

tests/e2e/business/audit_test.go
gfyrag: Fixed — comment now says `auditStringFilter` (f12b5582a).

internal/query/audit.go
gfyrag: Both done in f12b5582a: dropped the redundant `afterSeq >= 1` (implied by `afterSeq != 0` on a uint), and the zone-scan `entries` slice now takes the same `pageCap(pageSize)` capacity hint the seq-set path uses.

Azorlogh (approved):

cmd/ledgerctl/audit/list.go
```diff
@@ -20,19 +20,45 @@ import (
...
 		Args:              cobra.NoArgs,
 		ValidArgsFunction: cobra.NoFileCompletions,
 		RunE:              runList,
 	}
 
-	cmdutil.AddPaginationFlags(cmd, cmdutil.PaginationOptions{})
-	cmdutil.AddMinLogSequenceFlag(cmd)
+	cmdutil.AddPaginationFlags(cmd, cmdutil.PaginationOptions{SupportsReverse: true})
+	cmdutil.AddFilterFlags(cmd, cmdutil.FilterOptions{})
+	cmdutil.AddConsistencyFlags(cmd)
```
NumaryBot, line 62: 🟠 **[major] Checkpointed audit expansion reads from live store instead of checkpoint**

When `--checkpoint-id` is combined with `--expand`, the list page is read from the checkpoint but each individual entry expansion calls `GetAuditEntry` against the live store without passing the checkpoint ID. If audit data has been purged from the live store while still present in the checkpoint, the command fails with a not-found error instead of expanding the checkpointed entries correctly.

**Suggestion:** Either reject the `--checkpoint-id` + `--expand` combination with a clear error message, or thread the checkpoint ID through the per-entry `GetAuditEntry` calls so that expansion reads from the same checkpoint store as the list.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

gfyrag:
Re the 🟠 major (list.go — checkpointed audit expansion reads live store): **valid, fixed.** `--expand` fetches each entry via `GetAuditEntry`, whose request (`GetAuditEntryRequest{sequence}`) has no checkpoint parameter and always hits the live store — so `--expand` + `--checkpoint-id` mixed sources (and would NotFound on a live-purged-but-checkpointed entry). I took the typed-refusal option (invariant #7): the CLI now rejects the combination with a clear error rather than threading checkpoint through GetAuditEntry's proto, which is out of scope for the list-contract work. If checkpoint-scoped single-entry reads are wanted later, that's a clean separate change (add read options to GetAuditEntryRequest).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

gfyrag:
**Rebased onto the reshaped QueryFilter (#1547, `fix/en-1465-queryfilter-flat-shape`).** #1547 realigned QueryFilter's REST shape to the v2 $-DSL ($and/$or/$not + $match/$gt/… single-key ops, $param wrapper, has-asset via $exists). This PR now builds on that codec.

Audit-arm decision (v2 codec): I extended C's `marshalLeaf` with an explicit `QueryFilter_Audit` case that **fails loud** — `audit conditions are not supported over REST JSON (gRPC-only)` — rather than $-style round-trip. Reason: audit list is **gRPC-only** (no HTTP endpoint; ledgerctl sends the filter as protobuf binary; the sole REST-JSON QueryFilter surface, prepared queries, is scoped to ACCOUNTS/TRANSACTIONS). An AuditCondition can never legitimately reach the JSON codec, and it *cannot* round-trip losslessly anyway because the decoder dispatches on field name and audit names (ledger, timestamp, …) collide with the existing Ledger/BuiltinUint conditions. So a loud typed error (invariant #7) is correct; decode needs no change (audit-only field names already hit the existing "unsupported field" rejection). The human `filterexpr` CLI DSL (`audit[outcome] == failure`) is separate and unchanged.

⚠️ **Re-review needed:** this rebase onto the reshaped #1547 reopens the prior approval. When #1547 merges to release/v3.0 I'll re-rebase onto release.

internal/proto/commonpb/query_filter.go
```diff
@@ -0,0 +1,1044 @@
...
+		if value.isParam() {
+			return nil, errors.New("reverted: param not supported")
+		}
+		b, err := value.asBool()
+		if err != nil {
+			return nil, err
+		}
+
+		return &QueryFilter{Filter: &QueryFilter_Reverted{Reverted: &RevertedCondition{Value: b}}}, nil
+	default:
```
NumaryBot, line 838: 🟡 **[minor] REST QueryFilter decoder rejects $match equality for numeric built-in fields**

When a REST prepared-query request uses the documented equality form for built-in numeric fields (e.g. `{"$match":{"id":"42"}}`, `logId`, `timestamp`), the switch falls through to an 'unsupported field' error because only range operators are wired for those fields. The OpenAPI contract lists `id`, `timestamp`, `insertedAt`, `revertedAt`, `logId`, and `date` as supported fields and describes `$match` as equality, so exact numeric filters that should compile are rejected at create/update time.

**Suggestion:** Add equality (`$match`) handling for numeric built-in fields in the switch statement, compiling them to a point-range or an explicit equality predicate, so that the decoder honours the full `$match` contract described in the OpenAPI spec.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

gfyrag:
Re the 🟡 minor (query_filter.go:826 — $match equality for numeric built-in fields): **this is C's decoder code from #1547, not part of #1548's audit work — routing it there.** The `decodeMatch` default branch and the 'unsupported field' behavior for numeric built-ins (id/timestamp/logId/…) exist verbatim on `fix/en-1465-queryfilter-flat-shape` **before** my changes (line 818 there; it only shifted to 826 here because my audit marshal case added a few lines above). My PR only added the audit oneof arm handling; I deliberately don't touch C's decoder semantics per the merge-order coordination (#1547 owns the codec, #1548 builds on it). Whether `$match` should also accept numeric-field equality — folding to a point range — is a valid call for #1547's review. Non-blocking (NumaryBot check is green).

gfyrag (commented):
Reviewed the new head `ed4b0c3289fd1d20b31ccd1f5c4083ec5310c0fb`, focusing on the delta from previously reviewed `6e82cb333`.

No findings. The new commit only removes proto `reserved` declarations/comments for the pre-GA wire-compat path, which matches the current “no compatibility management” direction, and keeps the target schema sequential.

Checked locally:

- `go test ./internal/adapter/grpc ./internal/query ./internal/application/check ./internal/proto/servicepb ./internal/proto/commonpb`

internal/proto/commonpb/query_filter.go
```diff
@@ -0,0 +1,1044 @@
...
+		// carried as protobuf). It is deliberately NOT exposed on the REST-JSON
+		// surface: the only JSON entry point for QueryFilter is prepared queries,
+		// which are scoped to ACCOUNTS/TRANSACTIONS (see
+		// http.preparedQueryTargets), and this field-name-dispatched DSL cannot
+		// round-trip an audit condition losslessly anyway — audit field names
+		// (ledger, timestamp, …) collide with the transaction/log conditions the
+		// decoder already claims for those names. Fail loud (invariant #7) rather
+		// than emit a JSON node that would silently decode back to the wrong
+		// condition type.
+		return nil, errors.New("query filter: audit conditions are not supported over REST JSON (gRPC-only)")
```
NumaryBot, line 324: 🟡 **[minor] Audit filter arm in PreparedQuery causes JSON marshalling failure on REST paths**

When a gRPC client creates or updates a PreparedQuery whose `filter` uses the new `QueryFilter_Audit` oneof arm, prepared-query validation does not reject it. The value is persisted and later reaches `PreparedQuery.MarshalJSON` through the REST list/get endpoints, where the audit arm returns an error after a 200 response has already been written, producing a broken response body. Since the author explicitly noted that audit listing is gRPC-only and the JSON codec intentionally errors loud on this arm, the admission path for prepared queries should refuse audit filter arms up-front rather than silently allowing storage of an un-serialisable filter.

**Suggestion:** Add a validation step in the prepared-query create/update admission path (or FSM handler) that returns an InvalidArgument error when the submitted filter contains a `QueryFilter_Audit` oneof arm. This prevents an un-marshalable filter from being persisted and avoids the broken 200 response on subsequent REST reads.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

misc/proto/common.proto
```diff
@@ -1262,6 +1262,7 @@ message QueryFilter {
     LogBuiltinUintCondition log_builtin_uint = 10;
     AccountHasAssetCondition account_has_asset = 11;
     RevertedCondition reverted = 12;
+    AuditCondition audit = 13;
```
NumaryBot, line 1265: 🟡 **[minor] Prepared-query admission path does not reject QueryFilter_Audit arm**

When a gRPC client creates or updates a prepared query with a QueryFilter_Audit oneof arm, the admission path does not validate or reject it. Since prepared-query execution only handles the shared account/transaction/log compiler and the REST JSON marshaller explicitly errors on audit conditions, a stored audit-filter query will produce a broken response body on REST list/get calls (even after a 200 has already been written) and silently fail at execution time rather than at write time.

**Suggestion:** Add explicit validation in the CreatePreparedQuery and UpdatePreparedQuery handlers to reject a QueryFilter_Audit arm with an InvalidArgument error, matching the existing principle that audit listing is gRPC-only and cannot be round-tripped through the REST JSON codec.

misc/proto/bucket.proto
```diff
@@ -757,11 +757,11 @@ message CheckStoreProgress {
 // ============================================================================
 
 message ListAuditEntriesRequest {
```
NumaryBot, line 759: ⚪ **[nit] Reserve removed field tags for failures_only and ledger**

The `failures_only` (tag 2) and `ledger` (tag 3) fields were removed from `ListAuditEntriesRequest` without reserving their tags. Future additions could accidentally reuse these tag numbers, causing mixed-version clients still sending the old fields to have their bytes decoded as the wrong new field rather than being safely ignored.

**Suggestion:** Add `reserved 2, 3;` and `reserved "failures_only", "ledger";` to `ListAuditEntriesRequest` to protect against accidental tag reuse in future schema evolution.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

cmd/ledgerctl/audit/list.go
```diff
@@ -20,19 +21,46 @@ import (
...
+  ledgerctl audit list --filter 'audit[ledger] == my-ledger'
+  ledgerctl audit list --checkpoint-id 7`,
 		Args:              cobra.NoArgs,
 		ValidArgsFunction: cobra.NoFileCompletions,
 		RunE:              runList,
 	}
 
-	cmdutil.AddPaginationFlags(cmd, cmdutil.PaginationOptions{})
-	cmdutil.AddMinLogSequenceFlag(cmd)
+	cmdutil.AddPaginationFlags(cmd, cmdutil.PaginationOptions{SupportsReverse: true})
```
NumaryBot, line 60: 🟠 **[major] Demo tape references removed `--failures-only` and `--ledger` flags**

The `--failures-only` and `--ledger` flags have been removed from the `ledgerctl audit list` command, but `misc/demo/demo_audit.tape` still invokes `ledgerctl audit list --failures-only` and `ledgerctl audit list --ledger demo-audit`. Running `just generate-demo` will now fail with Cobra's unknown-flag error, breaking automated demo generation and any CI step that exercises it.

**Suggestion:** Update `misc/demo/demo_audit.tape` (and any related documentation) to use the new filter DSL form, e.g. `--filter 'audit[outcome] == failure'` and `--filter 'audit[ledger] == demo-audit'`, or add a compatibility alias if a smooth migration period is required.

misc/proto/bucket.proto
```diff
@@ -757,11 +757,11 @@ message CheckStoreProgress {
...
+  // options carries the whole read contract. Audit has NO dedicated top-level
+  // filters: ledger scope and outcome selection are expressed through
+  // options.filter (e.g. `audit[ledger] == <name>`, `audit[outcome] ==
+  // failure`), exactly like every other filtered list endpoint.
   common.ListOptions options = 1;
```
NumaryBot, line 764: ⚪ **[nit] Removed `failures_only` (tag 2) and `ledger` (tag 3) fields not reserved in `ListAuditEntriesRequest`**

The prior discussion acknowledged that the wire-compatibility blocker (options tag renumbering) was fixed and tags 2 and 3 for the dropped fields should be reserved. Tags 2 and 3 and their names are still not explicitly marked `reserved`, leaving them available for accidental reuse in future schema evolution, which could silently misinterpret old client payloads.

**Suggestion:** Add `reserved 2, 3; reserved "failures_only", "ledger";` to `ListAuditEntriesRequest` to prevent future reuse of the dropped tag numbers and field names.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

misc/proto/bucket.proto
NumaryBot, line 759: 🟡 **[minor] Removed proto field tags not explicitly reserved**

After removing failures_only (tag 2) and ledger (tag 3) from ListAuditEntriesRequest, those field numbers and names are not marked reserved. Future changes could inadvertently reuse these tags, causing old serialized clients to have their former audit flags silently misinterpreted as a new field rather than ignored safely.

**Suggestion:** Add reserved 2, 3; reserved "failures_only", "ledger"; to ListAuditEntriesRequest to prevent accidental reuse of these tag numbers and field names.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

misc/proto/bucket.proto
NumaryBot, line 759: 🟡 **[minor] Removed ListAuditEntriesRequest field tags 2 and 3 are not reserved**

The `failures_only` (tag 2) and `ledger` (tag 3) fields were dropped from `ListAuditEntriesRequest` but their tag numbers and names were not added to a `reserved` statement. This leaves both slots open for accidental future reuse, which could silently misinterpret bytes from older clients.

**Suggestion:** Add `reserved 2, 3; reserved "failures_only", "ledger";` to `ListAuditEntriesRequest` so the removed tags are explicitly locked out and the wire break is self-documenting.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1548#issuecomment-4923919964

tests/e2e/business/audit_test.go
gfyrag: Resolved in d738800b6, but the other direction from what this note proposed. Rather than flip the test to newest-first, I flipped the implementation + docs back to oldest-first (chronological), which is the audit trail's natural read order and the pre-ListOptions behavior; the earlier newest-first default was an unnecessary behavior change that broke every test asserting `entries[len-1] == newest`.

At HEAD all four now agree on oldest-first-by-default / `--reverse` for newest-first:
- `DefaultController.ListAuditEntries` (controller_default.go:1126-1129) documents `reverse=false` as ascending/oldest-first and passes `reverse` through to `ReadAuditEntriesPage` directly (no `!reverse`).
- `ledgerctl audit list` help (list.go:29) and docs/ops/cli.md: oldest first by default, `--reverse` for newest first.
- This e2e (`asc = Reverse:false`) correctly asserts oldest-first (`asc[0] < asc[len-1]`), and `desc[0] == asc[len-1]`, `desc[len-1] == asc[0]`.

The assertion validates the documented order as written; no test change needed.

PR #1549, opened 2026-07-09
Document audit-bound technical state

gfyrag:
## What changed

Adds `docs/technical/architecture/audit-vs-technical-state.md`, a cross-cutting architecture page that classifies persisted state as business truth, governance truth, operational consensus state, or rebuildable projection.

The page documents the audit/checker decision rule and calls out the risky future boundaries discussed during review: `TechnicalUpdate`, mirror cursor, readstore/indexes, idempotency eviction, cluster configuration, and backup destination identity.

Also links the new page from the top-level architecture README and the checker subsystem README.

## Why

The current invariant says audit is the source of truth, but future changes need a sharper rule for deciding when non-audited technical state is acceptable. This page makes the rule explicit: if state can change business results, write acceptance/rejection, or the trust model for business evidence, it must be audit-bound or checker-verified.

## Validation

- `git diff --check -- docs/technical/architecture/audit-vs-technical-state.md docs/technical/architecture/README.md docs/technical/architecture/subsystems/checker/README.md`
- `node .gitnexus/run.cjs detect_changes` returned affected processes `0`, risk `low`
- In a clean detached worktree on the PR commit: `GOROOT= go build ./...`
- In a clean detached worktree on the PR commit: `GOROOT= just pre-commit` exited successfully, but left generated proto files modified in that isolated worktree. Those generated diffs are not part of this docs-only PR because no proto files changed.

`nix develop --command bash -c "just pre-commit"` was also attempted but failed on a Go toolchain mismatch (`go1.26.3` package data vs `go1.26.4` tool), so the documented `GOROOT=` fallback was used.


gfyrag (commented):
Reviewed the new draft PR at `8783e269a4cc3b9260071c5b5dd0667630b724d9`.

No findings. The added architecture page matches the audit/checker invariant: it keeps audit-bound business truth separate from checker-verified projections and operational consensus state, and the README links make the new classification page discoverable.

Scope is documentation-only, so I did not run Go tests.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,119 @@
...
+> If changing the state after the fact can change a ledger business result, a
+> business write acceptance or rejection, or the trust model for business
+> evidence, the state must be audit-bound or verified as a projection of
+> audit-bound data.
+
+## State Classes
+
+| Class | Definition | Examples | Required control |
+| --- | --- | --- | --- |
+| Business truth | Changes balances, transactions, metadata, reversions, ledger lifecycle, indexes, prepared queries, or any user-visible business decision. | `AuditEntry`, `Log`, `Volume`, `Metadata`, `Transaction`, `Reference`, reversion bitsets, idempotency outcomes, index registry. | The command must produce an audit entry, and persisted projections must be checked by `internal/application/check`. |
```
NumaryBot: 🟠 **[major] Prepared-query projections documented as checker-covered but no checker pass exists**

The table classifies prepared-query business-truth projections as requiring coverage by `internal/application/check`, but no checker pass for `SubAttrPreparedQuery` exists. The documented passes cover volumes, metadata, transactions, reversion, sealing, exclusions, idempotency, and indexes. This overstates the integrity guarantee for prepared queries.

**Suggestion:** Either add a note calling out prepared queries as a known uncovered gap in checker coverage, or add the corresponding checker pass before removing the caveat.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1549#issuecomment-4924154173

docs/technical/architecture/audit-vs-technical-state.md
NumaryBot: 🟠 **[major] Prepared-query projections documented as checker-covered but no checker pass exists**

The coverage table classifies prepared-query business-truth projections as requiring (and implying) coverage by `internal/application/check`, but no corresponding checker pass for `SubAttrPreparedQuery` currently exists. The documented passes cover volumes, metadata, transactions, reversion, sealing, exclusions, idempotency, and indexes. This overstates the actual integrity guarantee for prepared queries and could mislead readers (and auditors) about the real state of the system.

**Suggestion:** Either annotate the table entry for prepared-query projections with a clear note calling it out as a known gap in checker coverage, or implement the missing `SubAttrPreparedQuery` checker pass before making this claim in the documentation.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1549#issuecomment-4924154173

docs/technical/architecture/audit-vs-technical-state.md
NumaryBot: 🟠 **[major] Prepared-query projections documented as checker-covered but no checker pass exists**

The coverage table classifies prepared-query business-truth projections as requiring (and implying) coverage by `internal/application/check`, but no corresponding checker pass for `SubAttrPreparedQuery` currently exists. The documented passes cover volumes, metadata, transactions, reversion, sealing, exclusions, idempotency, and indexes. This overstates the actual integrity guarantee for prepared queries and could mislead readers and auditors about the real state of the system.

**Suggestion:** Either annotate the table entry for prepared-query projections with a clear note calling it out as a known gap in checker coverage, or implement the missing `SubAttrPreparedQuery` checker pass before making this claim in the documentation.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1549#issuecomment-4924154173

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,135 @@
...
+> If changing the state after the fact can change a ledger business result, a
+> business write acceptance or rejection, or the trust model for business
+> evidence, the state must be audit-bound or verified as a projection of
+> audit-bound data.
+
+## State Classes
+
+| Class | Definition | Examples | Required control |
+| --- | --- | --- | --- |
+| Business truth | Changes balances, transactions, metadata, reversions, ledger lifecycle, indexes, or any user-visible business decision. | `AuditEntry`, `Log`, `Volume`, `Metadata`, `Transaction`, `Reference`, reversion bitsets, idempotency outcomes, index registry. | The command must produce an audit entry, and persisted projections must be checked by `internal/application/check`. |
```
NumaryBot, line 20: 🟠 **[major] Prepared-query projections documented as checker-covered but no checker pass exists**

The coverage table classifies prepared-query business-truth projections as requiring (and implying) coverage by `internal/application/check`, but no corresponding checker pass for `SubAttrPreparedQuery` currently exists. The documented passes cover volumes, metadata, transactions, reversion, sealing, exclusions, idempotency, and indexes. This overstates the actual integrity guarantee for prepared queries and could mislead readers and auditors about the real state of the system.

**Suggestion:** Either annotate the table entry for prepared-query projections with a clear note calling it out as a known gap in checker coverage, or implement the missing `SubAttrPreparedQuery` checker pass before making this claim in the documentation.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,135 @@
...
+| Governance truth | Changes who can write, how writes are accepted, or how business evidence is produced. It may not change balances directly, but it changes the control plane around business truth. | Signing keys, maintenance mode, chapter schedules, hash algorithm selection. | Prefer an audited order. If represented as technical state, document why and expose an operator-visible control trail. |
```
NumaryBot: 🟠 **[major] Persisted governance projections lack required checker verification**

For persisted governance rows (e.g. signing keys, signing config, maintenance mode, chapter schedules), the document's guidance is too weak: an audited order records the intended change, but if the persisted projection is altered after the fact, admission/lifecycle code can accept or reject future writes differently without any checker comparison to the audit chain. Non-persisted operator controls and persisted governance projections are not distinguished, creating a gap in integrity enforcement.

**Suggestion:** Distinguish non-persisted operator controls from persisted governance projections in the classification. For the latter, explicitly require checker verification or an audit-bound source read to prevent silent divergence between the audit chain and the live governance state.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1549#issuecomment-4924154173

docs/technical/architecture/subsystems/checker/README.md
```diff
@@ -15,6 +15,12 @@ Two pages cover what the checker depends on and what it does.
 
 The audit chain and the checker are tightly coupled by an explicit invariant: **every persisted dataset is either hash-bound (the audit log itself) or derivable from the hash-bound data, in which case the checker must verify it on every `Check()` run** ([CLAUDE.md invariant #8](../../../../../AGENTS.md)). This rule shapes design decisions across every other subsystem — what to refactor versus what to bind — so the checker's coverage is its own first-class architectural concern.
 
+For new persisted state, first classify whether it is business truth, governance
+truth, operational consensus state, or a rebuildable projection using
+[Audit-Bound vs Technical State](../../audit-vs-technical-state.md). Only
+business truth and business-visible projections belong in the checker; purely
```
flemzord: 🟠 **[major] Do not exempt persisted operational projections from invariant #8**

This weakens the invariant stated immediately above and in `AGENTS.md`: every non-audit persisted dataset, explicitly including mirror cursors and bloom filters, must be verified by the checker. Raft replication only keeps replicas consistent; it cannot detect a corrupted or tampered value replicated identically on every node. Please remove this exemption and require checker coverage for persisted projections, or formally revise the invariant and its implementation before documenting a different rule.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,135 @@
...
+
+That atomicity protects normal failures, but it does not by itself prove that the
+stored cursor is still consistent with audited mirror orders after restore,
+manual repair, or future tooling. A cursor that advances beyond the highest
+audited `MirrorIngest.v2LogId` can make the worker skip source logs without any
+audit entry showing the missing proposals.
+
+Treat `MirrorCursor` as a technical projection with a business-adjacent effect:
+it should never be greater than the maximum v2 log id found in the target
+ledger's audited mirror-ingest logs. If mirror recovery, repair tooling, or
```
flemzord: 🟠 **[major] Require exact mirror-cursor equality, not only an upper bound**

A cursor below the highest audited `v2LogId` is also unsafe. `Worker.processBatch` reloads that cursor and fetches source logs after it; `TranslateBatch` emits new `MirrorIngest` orders, and `processMirrorCreatedTransaction` reapplies postings with `force=true` while overwriting the same transaction state, without deduplicating on `v2LogId`. A lowered cursor can therefore replay already-audited transactions and apply their balance effects twice. Require the live cursor to equal the highest audited mirror-ingest ID (with explicit empty/deleted/archive handling), not merely to be less than or equal to it.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,135 @@
...
+stronger checker coverage is added, include this comparison.
+
+## Readstore and Indexes
+
+The index registry is business-visible and is already verified by the checker:
+presence and identity are re-derived from audited create/drop/delete logs.
+
+The readstore inverted index is a rebuildable projection. It is not the source of
+business truth, but it can affect API responses while it is corrupted or stale.
+When a readstore projection becomes authoritative for a business-visible query,
```
flemzord: 🟠 **[major] Treat readstore verification as a current gap**

This condition already holds: reference, metadata, address, asset, timestamp, and other filters compile directly to readstore iterators after the local `IndexVersionState` readiness gate, with no audit/attribute fallback. A corrupted READY index is therefore authoritative for API results today, while `Checker.compareIndexes` verifies only registry presence and identity, not index contents. Please describe this as an existing integrity gap and require checker or rebuild-health validation for the current query path rather than framing it as a future transition.

flemzord (commented):
Found three additional integrity-documentation issues at the current head. These comments are distinct from the existing prepared-query and governance-state threads.

gfyrag:
Review pass on `5e49c9122155e8eaaf3e2a0853c94bcd276ae444`.

Findings:

1. Medium: the prepared-query checker gap is framed as future-looking, but prepared queries are already authoritative for user-visible behavior. `ExecutePreparedQuery` reads `SubAttrPreparedQuery` from Pebble before execution, so a tampered prepared-query projection can affect query results today. Please describe this as a current known integrity gap and state that a checker compare pass is required to close it.

2. Low: the index registry paragraph says verification is re-derived from audited create/drop/delete logs, but the checker also replays `RemovedMetadataFieldType` because deleting a metadata field type can cascade-drop an index. Please mention `RemovedMetadataFieldType` explicitly so the doc matches checker behavior.

Adjacent stale references: some existing docs/comments still mention `IndexReady` even though the proto field is reserved/removed. This PR did not introduce those, but they sit near the same audit/checker technical-state story.

Validation run: `git diff --check release/v3.0...HEAD` passed; targeted checker/processing/query/state tests for the referenced mechanisms passed.

docs/technical/architecture/subsystems/checker/README.md
gfyrag: Corrected. Removed the wording that let persisted operational projections stay out of the checker. The checker/README section and the State Classes table now state the invariant #8 floor explicitly: every non-audit dataset persisted in the main Pebble store needs checker coverage unless it is genuinely discarded and rebuilt by a lifecycle path (bloom filters) or lives in a separate rebuildable side-store. I verified against the code: the mirror cursor lives in ZonePerLedger/SubPLMirrorCursor in the **main** store (not a side-store) and has no compare* pass, so it is now documented as a tracked integrity gap, not an exemption. Bloom filters are legitimately in the rebuildable class — prepare.go drops them on every backup/restore and rebuilds from a full attribute scan — and the doc now says a projection only qualifies as rebuildable if it is genuinely thrown away and rebuilt. I also added the note that Raft replication only guarantees identical bytes across replicas and cannot detect a value corrupted/tampered before it is proposed.

docs/technical/architecture/audit-vs-technical-state.md
gfyrag: Confirmed against the code and corrected. The Mirror Cursor section now requires the live cursor to **equal** the highest audited MirrorIngest.v2LogId, not merely be <= it, and spells out why a lowered cursor is unsafe: Worker.processBatch reloads the cursor and calls FetchLogs(cursor, ...) which returns logs strictly after it; TranslateBatch re-emits MirrorIngest orders; processMirrorCreatedTransaction reapplies every posting with force=true and applyPosting adds the amounts to the volume pair additively. There is no dedup on v2LogId, so replayed transactions double-apply their balance effects. I also added explicit edge-case handling: empty/never-ingested ledger => cursor 0 (query.ReadMirrorCursor defaults to 0, processBatch maps 0 to expectedNextLogID==1); deleted/archived ledger => worker stops but the cursor row is not cleared, so an equality check must treat it as out of scope rather than compare a stale cursor against an empty audit set. Noted that no compare* pass enforces this today, so it is a tracked gap.

docs/technical/architecture/audit-vs-technical-state.md
gfyrag: Confirmed and reframed as a current gap. Verified in internal/query/compile.go: once the per-replica IndexVersionState.CurrentVersion is non-zero (READY), reference/metadata/address/asset/timestamp filters compile directly to readstore iterators via requireIndexReady, with no on-scan/audit fallback (compile.go even documents 'there is no on-scan fallback' for the asset builtin). So a corrupted READY index is authoritative for API results today. Meanwhile checker.go compareIndexes re-derives only registry presence and identity from the audited create/drop/delete logs and explicitly does NOT verify the readstore inverted-index contents (per its own comments, and EN-1323 for the version/build-status split). The doc now describes this as an existing integrity gap and states the current query path needs either checker re-derivation of index contents or a rebuild-health mechanism before a READY index is served — not a future transition.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,194 @@
...
+> If changing the state after the fact can change a ledger business result, a
+> business write acceptance or rejection, or the trust model for business
+> evidence, the state must be audit-bound or verified as a projection of
+> audit-bound data.
+
+## State Classes
+
+| Class | Definition | Examples | Required control |
+| --- | --- | --- | --- |
+| Business truth | Changes balances, transactions, metadata, reversions, ledger lifecycle, indexes, or any user-visible business decision. | `AuditEntry`, `Log`, `Volume`, `Metadata`, `Transaction`, `Reference`, reversion bitsets, idempotency outcomes, index registry. | The command must produce an audit entry, and persisted projections must be checked by `internal/application/check`. |
```
NumaryBot, line 20: 🟠 **[major] Prepared-query projections documented as checker-covered but no checker pass exists**

The coverage table classifies prepared-query business-truth projections as requiring (and implying) coverage by `internal/application/check`, but no corresponding checker pass for `SubAttrPreparedQuery` currently exists. The documented passes cover volumes, metadata, transactions, reversion, sealing, exclusions, idempotency, and indexes. This overstates the actual integrity guarantee for prepared queries. Critically, `ExecutePreparedQuery` reads the `SubAttrPreparedQuery` projection to decide user-visible query behavior today, so a tampered projection can affect results without any detection — this is a current known gap, not a future concern.

**Suggestion:** Annotate the table entry for prepared-query projections with a clear note identifying it as a current known gap in checker coverage (no `SubAttrPreparedQuery` pass exists), or implement the missing checker pass before making this claim. The checker README gap list should also be updated to include prepared queries explicitly.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,194 @@
...
+| Governance truth | Changes who can write, how writes are accepted, or how business evidence is produced. It may not change balances directly, but it changes the control plane around business truth. | Signing keys, maintenance mode, chapter schedules, hash algorithm selection. | Prefer an audited order. If a persisted projection of governance state (e.g. the signing-key or signing-config rows under `SubGlobSigningKey` / `SubGlobSigningConfig`) can be altered after the fact so admission or lifecycle code accepts or rejects future writes differently, it must be checker-verified or read back from an audit-bound source — not merely trusted because an audited order once recorded the intent. Non-persisted operator controls (in-memory toggles) do not carry this obligation. |
```
NumaryBot, line 21: 🟠 **[major] Persisted governance projections not distinguished from non-persisted operator controls**

The guidance for persisted governance rows (signing keys, signing config, maintenance mode, chapter schedules, etc.) is too weak. The document does not distinguish these persisted projections from non-persisted operator controls. If a persisted governance projection is altered after the audited order is written, admission/lifecycle code can accept or reject future writes differently with no checker comparison to the audit chain, creating a silent divergence gap.

**Suggestion:** Add a clear distinction between non-persisted operator controls and persisted governance projections in the classification table and guidance. For persisted governance projections, explicitly require either checker verification or an audit-bound source read to prevent silent divergence between the audit chain and the live governance state.

docs/technical/architecture/subsystems/checker/README.md
```diff
@@ -15,6 +15,18 @@ Two pages cover what the checker depends on and what it does.
 
 The audit chain and the checker are tightly coupled by an explicit invariant: **every persisted dataset is either hash-bound (the audit log itself) or derivable from the hash-bound data, in which case the checker must verify it on every `Check()` run** ([CLAUDE.md invariant #8](../../../../../AGENTS.md)). This rule shapes design decisions across every other subsystem — what to refactor versus what to bind — so the checker's coverage is its own first-class architectural concern.
 
+For new persisted state, first classify whether it is business truth, governance
+truth, operational consensus state, or a rebuildable projection using
+[Audit-Bound vs Technical State](../../audit-vs-technical-state.md). Per invariant
+#8, every non-audit dataset persisted in the main Pebble store needs checker
```
NumaryBot, line 21: 🟠 **[major] Persisted operational projections exempted from invariant #8 — exemption is unsound**

The README weakens the invariant (which states every non-audit persisted dataset must be verified by the checker) by exempting persisted operational projections such as mirror cursors and bloom filters. Raft replication only keeps replicas mutually consistent; it cannot detect a corrupted or tampered value that has been replicated identically to every node. The exemption therefore leaves a real integrity gap.

**Suggestion:** Remove the exemption and require checker coverage for persisted operational projections, or formally revise the invariant and its implementation before documenting a different rule.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,194 @@
...
+
+If any answer is yes, the variant needs either an audited order or an explicit
+checker pass that verifies the stored value against audit-bound data.
+
+## Mirror Cursor
+
+Mirror ingestion is the highest-risk boundary because the mirror cursor is an
+operational value that controls future business proposals. The normal apply path
+is correct: the worker proposes mirror ingest orders and a `MirrorSyncUpdate` in
+the same Raft proposal, and the FSM queues the cursor write through the same
```
NumaryBot, line 86: 🟠 **[major] Mirror-cursor validity documented as upper-bound only — lowered cursor can cause double-replay**

The document requires only that the live cursor is ≤ the highest audited `v2LogId`, but a cursor *below* that value is equally unsafe. `Worker.processBatch` reloads the cursor and fetches source logs after it; `TranslateBatch` emits new `MirrorIngest` orders and `processMirrorCreatedTransaction` reapplies postings with `force=true` while overwriting the same transaction state without deduplicating on `v2LogId`. A lowered cursor can therefore replay already-audited transactions and double-apply balance effects.

**Suggestion:** Require the live cursor to equal the highest audited mirror-ingest ID (with explicit handling for the empty/deleted/archive edge cases), not merely to be less than or equal to it.

NumaryBot (commented):
NumaryBot posted 7 new inline findings.

Summary: https://github.com/formancehq/ledger/pull/1549#issuecomment-4924154173

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,194 @@
...
+`WriteSet` so it only advances if the mirrored orders apply successfully.
+
+That atomicity protects normal failures, but it does not by itself prove that the
+stored cursor is still consistent with audited mirror orders after restore,
+manual repair, or future tooling. The cursor is wrong in **both** directions:
+
+- A cursor that advances **beyond** the highest audited `MirrorIngest.v2LogId`
+  makes the worker skip source logs without any audit entry showing the missing
+  proposals — silent under-ingestion.
+- A cursor that drops **below** the highest audited `v2LogId` is equally unsafe.
```
NumaryBot, line 96: 🟠 **[major] Readstore verification framed as future concern — it is a current gap**

Reference, metadata, address, asset, timestamp, and other filters already compile directly to readstore iterators after the local `IndexVersionState` readiness gate, with no audit/attribute fallback. A corrupted READY index is therefore authoritative for API results today. `Checker.compareIndexes` verifies only registry presence and identity, not index contents. Framing this as a future transition understates the current risk.

**Suggestion:** Describe this as an existing known integrity gap. Require checker or rebuild-health validation for the current query path rather than framing it as a future concern.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,194 @@
...
+| Rebuildable local projection | Speeds reads or background work and can be regenerated from audit-bound data. **A projection only belongs here if it is genuinely thrown away and rebuilt** — not merely rebuildable in principle. | Bloom filters (dropped on every backup/restore and rebuilt from a full attribute scan, see `internal/infra/attributes/prepare.go`), cache mirrors, snapshots/spool. | Rebuild path is the control **when the projection is actually rebuilt as part of a lifecycle path**. A *durably persisted* projection that is trusted between rebuilds and can shape a business-visible result (e.g. readstore inverted indexes while a READY index is authoritative — see below) additionally needs checker coverage, because nothing rebuilds it before it is served. |
```
NumaryBot: 🟠 **[major] Cache/bloom projections incorrectly documented as rebuild-only exceptions**

`CacheSnapshotter.RestoreFromStore` restores 0xFF cache generations and persisted bloom blocks on normal restart or follower sync, only falling back to a full attribute scan when no bloom blocks exist. A corrupted cache snapshot or a false-negative persisted bloom block can change FSM cache reads/preload behavior. Documenting these as rebuildable exceptions understates the checker/control requirement.

**Suggestion:** Remove or qualify the rebuild-only exception for cache mirrors and bloom filters. Acknowledge that these are restored from persisted snapshots on normal restart and therefore require the same integrity controls as other persisted projections.

docs/technical/architecture/subsystems/checker/README.md
```diff
@@ -15,6 +15,18 @@ Two pages cover what the checker depends on and what it does.
...
+coverage unless it is genuinely discarded and rebuilt by a lifecycle path (e.g.
+bloom filters) or lives in a separate rebuildable side-store. Raft replication is
+not a substitute: it only guarantees replicas hold identical bytes, so a value
+corrupted or tampered before it is proposed is copied to every node and no
+cross-node comparison can detect it. Persisted projections that are not yet
```
NumaryBot, line 39: 🟠 **[major] Checker README gap list omits prepared queries**

The checker README gap list documents only mirror cursor and readstore indexes as uncovered projections. The new audit-vs-technical-state page acknowledges that `SubAttrPreparedQuery` is persisted and not checker-verified, and `ExecutePreparedQuery` reads it to drive user-visible query results. Readers using the README as the canonical checker coverage summary will be unaware of this gap.

**Suggestion:** Add prepared queries (`SubAttrPreparedQuery`) to the explicit gap list in the checker README, with a note that no checker pass exists for this projection.

gfyrag:
Review pass on `5f16f571932762989447404f714081de006832ee`.

Findings:

1. Major: `docs/technical/architecture/subsystems/checker/README.md` still lists only mirror cursor and readstore index contents as current uncovered examples. It omits prepared queries, even though the audit-vs-technical-state page now says `SubAttrPreparedQuery` has no checker pass, and `ExecutePreparedQuery` reads that projection directly from Pebble. Please add prepared queries to the checker README gap list, and change the prepared-query wording from “when it becomes authoritative” to “because it is authoritative today”.

2. Major: the bloom/cache exception is still too broad for invariant #8. The docs exempt bloom filters when “discarded and rebuilt”, but normal restart does not discard them: `RestoreFromStore` restores persisted cache generations and bloom blocks, falling back to a full scan only when no persisted bloom blocks exist. Please scope the exception to backup/restore rebuild behavior only, or document normal persisted bloom/cache snapshots as an uncovered integrity gap unless the invariant is intentionally revised.

3. Low: the `compareIndexes` description still says registry state is derived from audited create/drop/delete logs. The checker also includes `RemovedMetadataFieldType` cascades and `DeleteLedger`; spell those out so the page matches current checker behavior.

Resolved after the rebase: governance-state distinction, mirror cursor equality/lowered-cursor replay, readstore framed as a current gap. The prepared-query checker gap is only partially resolved because the README and wording above still lag.

Validation run: `git diff --check release/v3.0...HEAD`, `git diff --check 5e49c9122155e8eaaf3e2a0853c94bcd276ae444..HEAD`, and targeted checker/query/state/attributes tests passed.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,194 @@
...
+persisted per-ledger projection (`ZonePerLedger` / `SubPLMirrorCursor`, in the
+main Pebble store) that must **equal** the highest audited `MirrorIngest.v2LogId`
+for the target ledger — not merely be less than or equal to it. Explicit edge
+cases:
+
+- **Empty / never-ingested ledger:** the cursor is absent and reads back as `0`
+  (`query.ReadMirrorCursor` defaults to `0`, and `processBatch` maps `0` to
+  `expectedNextLogID == 1`); a ledger with no audited mirror-ingest logs must
+  hold cursor `0`.
+- **Deleted / archived ledger:** workers stop but the cursor row is **not**
```
flemzord: 🟡 **[moderate] Deleted-ledger cleanup does clear the mirror cursor**

This can lead a future checker to exclude the wrong lifecycle state. `DeleteLedger` queues `PendingLedgerCleanups`; when a covering chapter purge reaches the delete sequence, `WriteSet.executePurge` calls `deleteLedgerData`, which explicitly point-deletes `SubPLMirrorCursor` together with the mirror source head and status. The cursor persists only during the pending-cleanup window, not after cleanup, and a ledger itself has no “archived” state. Please document the transient pending-cleanup case and expect the cursor to be absent once cleanup completes.

flemzord (commented):
Delta review of the new head found two additional factual issues. The original three comments were revalidated separately and are not duplicated here.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,194 @@
...
+caller gets back, while `compareIndexes` never re-derives those inverted-index
+entries from the audit. Between rebuilds nothing regenerates the index, so the
+rebuild path is not a control here.
+
+This is a tracked gap (see the `compareIndexes` comments referencing the index
+content-verification effort, and `EN-1323` on the version/build-status split).
+Closing it requires the checker to re-derive the readstore inverted-index
+contents from the audited logs, or a rebuild-health mechanism that proves the
+projection matches the audit before a READY index is served — Raft replication
+does not help, since a bit-flip present before replication is copied to every
```
flemzord: 🟡 **[minor] Readstore contents are not replicated by Raft**

The conclusion that Raft does not validate this projection is correct, but this mechanism is not: readstore is a separate per-replica store populated locally by each node’s index builder from the replicated logs, and `IndexVersionState` is explicitly local. A bit flip in one readstore is therefore not copied to every replica. Please state that Raft replicates the source audit/log stream, not the readstore bytes, so it neither detects nor repairs replica-local index corruption.

docs/technical/architecture/audit-vs-technical-state.md
gfyrag: Confirmed against the code and corrected. Verified: DeleteLedger records a deferred cleanup (savePendingLedgerCleanup -> State.PendingLedgerCleanups); when a covering chapter purge reaches the delete sequence, WriteSet.executePurge (internal/infra/state/write_set.go) calls deleteLedgerData (internal/infra/state/batch.go), which point-deletes SubPLMirrorCursor together with SubPLMirrorSourceHead, SubPLMirrorStatus and the pending-cleanup marker. Also confirmed there is no 'archived' ledger state (only a DeletedAt soft-delete marker, then removal). The doc's Mirror Cursor edge-case bullet now states the cursor IS cleared on cleanup, present only during the transient pending-cleanup window and absent afterwards, and that a checker must exclude only a ledger still awaiting cleanup (delete logged, purge not yet reached) rather than assume a deleted ledger keeps a stale cursor forever.

docs/technical/architecture/audit-vs-technical-state.md
gfyrag: Confirmed and corrected the mechanism (conclusion was already right). Verified: the readstore is a separate per-replica Pebble store (pebble.Open in internal/storage/readstore/store.go, distinct from the byte-replicated FSM store), populated locally by each node's index builder (internal/application/indexbuilder) from the replicated log/audit stream, and IndexVersionState.CurrentVersion is explicitly per-replica ('built locally'). The doc now states that Raft replicates the source audit/log stream, not the readstore bytes, so a bit-flip in one replica's readstore is NOT propagated to the others (unlike main-store values which are copied bit-for-bit) — hence Raft neither detects nor repairs replica-local index corruption, and only per-replica checker or rebuild-health validation can catch it. Removed the earlier wording that wrongly described it as a pre-replication bit-flip copied identically to every node.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,215 @@
...
+projection matches the audit before a READY index is served.
+
+Raft does not help here, but the reason differs from the main-store projections
+above. The readstore is a **separate, per-replica Pebble store** (opened
+independently in `internal/storage/readstore`), not the byte-replicated FSM
+store. It is populated **locally** by each node's index builder
+(`internal/application/indexbuilder`) from the replicated log/audit stream, and
+`IndexVersionState.CurrentVersion` (readiness) is explicitly per-replica. Raft
+replicates the *source* audit/log stream, not the readstore bytes — so unlike a
+main-store value (which is copied bit-for-bit to every node), a bit-flip in one
```
flemzord: 🟡 **[moderate] Main-store projections are not guaranteed byte-identical across replicas**

This overstates what Raft and the deterministic FSM guarantee. Raft replicates proposals and the FSM must produce the same logical state, but several Pebble projections intentionally use non-deterministic protobuf serialization: `AppliedProposal.TransientVolumes` and `Log` metadata maps can persist byte-divergent values for the same proposal (`appendAppliedProposal` documents this explicitly), and `saveIdempotencyKey` likewise says marshal order need not match across replicas. Only selected carriers such as `AuditEntry` use deterministic marshaling. Please contrast the per-replica readstore with the FSM’s logically equivalent state, not claim that every main-store value is copied bit-for-bit.

flemzord (commented):
Final delta review found one additional factual issue. Previously reported items were revalidated and are not duplicated.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1549#issuecomment-4924154173

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,215 @@
...
+> If changing the state after the fact can change a ledger business result, a
+> business write acceptance or rejection, or the trust model for business
+> evidence, the state must be audit-bound or verified as a projection of
+> audit-bound data.
+
+## State Classes
+
+| Class | Definition | Examples | Required control |
+| --- | --- | --- | --- |
+| Business truth | Changes balances, transactions, metadata, reversions, ledger lifecycle, indexes, or any user-visible business decision. | `AuditEntry`, `Log`, `Volume`, `Metadata`, `Transaction`, `Reference`, reversion bitsets, idempotency outcomes, index registry. | The command must produce an audit entry, and persisted projections must be checked by `internal/application/check`. |
```
NumaryBot, line 20: 🟠 **[major] Prepared-query projections documented as checker-covered but no checker pass exists**

The coverage table classifies prepared-query business-truth projections as requiring (and implying) coverage by `internal/application/check`, but no corresponding checker pass for `SubAttrPreparedQuery` currently exists. Critically, `ExecutePreparedQuery` reads the `SubAttrPreparedQuery` projection directly from Pebble to drive user-visible query behavior today, so a tampered projection can affect results with no detection. This is a current known gap, not a future concern.

**Suggestion:** Annotate the table entry for prepared-query projections with a clear note identifying it as a current known integrity gap (no `SubAttrPreparedQuery` checker pass exists). Change any wording that implies coverage will be added 'when it becomes authoritative' to reflect that it is authoritative today. Also update the checker README gap list to include prepared queries explicitly.

docs/technical/architecture/subsystems/checker/README.md
NumaryBot, line 39: 🟠 **[major] Checker README gap list omits prepared queries**

The checker README gap list documents only mirror cursor and readstore indexes as uncovered projections. The audit-vs-technical-state page acknowledges that `SubAttrPreparedQuery` is persisted and not checker-verified, and `ExecutePreparedQuery` reads it to drive user-visible query results today. Readers using the README as the canonical checker coverage summary will be unaware of this critical gap.

**Suggestion:** Add prepared queries (`SubAttrPreparedQuery`) to the explicit gap list in the checker README, noting that no checker pass exists for this projection and that it is currently authoritative for user-visible query results.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,215 @@
...
+| Rebuildable local projection | Speeds reads or background work and can be regenerated from audit-bound data. **A projection only belongs here if it is genuinely thrown away and rebuilt** — not merely rebuildable in principle. | Bloom filters (dropped on every backup/restore and rebuilt from a full attribute scan, see `internal/infra/attributes/prepare.go`), cache mirrors, snapshots/spool. | Rebuild path is the control **when the projection is actually rebuilt as part of a lifecycle path**. A *durably persisted* projection that is trusted between rebuilds and can shape a business-visible result (e.g. readstore inverted indexes while a READY index is authoritative — see below) additionally needs checker coverage, because nothing rebuilds it before it is served. |
```
NumaryBot: 🟠 **[major] Cache/bloom projections incorrectly documented as rebuild-only exceptions**

`CacheSnapshotter.RestoreFromStore` restores persisted `ZoneCache` generations and bloom blocks directly from Pebble on normal restart or follower sync, only falling back to a full attribute scan when no persisted bloom blocks exist. A corrupted cache snapshot or a false-negative persisted bloom block can change FSM cache reads and preload behavior. Documenting these as 'rebuildable exceptions' understates the integrity requirement for these persisted main-store projections.

**Suggestion:** Remove or qualify the rebuild-only exception for cache mirrors and bloom filters. Acknowledge that on normal restart these are restored from persisted snapshots and therefore require the same integrity controls as other persisted projections. Scope any 'rebuildable' exception strictly to backup/restore paths where they are genuinely discarded and rebuilt.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,215 @@
...
+| --- | --- | --- | --- |
+| Business truth | Changes balances, transactions, metadata, reversions, ledger lifecycle, indexes, or any user-visible business decision. | `AuditEntry`, `Log`, `Volume`, `Metadata`, `Transaction`, `Reference`, reversion bitsets, idempotency outcomes, index registry. | The command must produce an audit entry, and persisted projections must be checked by `internal/application/check`. |
+| Governance truth | Changes who can write, how writes are accepted, or how business evidence is produced. It may not change balances directly, but it changes the control plane around business truth. | Signing keys, maintenance mode, chapter schedules, hash algorithm selection. | Prefer an audited order. If a persisted projection of governance state (e.g. the signing-key or signing-config rows under `SubGlobSigningKey` / `SubGlobSigningConfig`) can be altered after the fact so admission or lifecycle code accepts or rejects future writes differently, it must be checker-verified or read back from an audit-bound source — not merely trusted because an audited order once recorded the intent. Non-persisted operator controls (in-memory toggles) do not carry this obligation. |
```
NumaryBot, line 21: 🟠 **[major] Persisted governance projections not distinguished from non-persisted operator controls**

The document does not distinguish persisted governance projections (signing keys, signing config, maintenance mode, chapter schedules, etc.) from non-persisted operator controls. If a persisted governance projection is altered after the audited order is written, admission/lifecycle code can accept or reject future writes differently with no checker comparison to the audit chain, creating a silent divergence gap.

**Suggestion:** Add a clear distinction in the classification table and guidance between non-persisted operator controls and persisted governance projections. For persisted governance projections, explicitly require either checker verification or an audit-bound source read to prevent silent divergence between the audit chain and the live governance state.

docs/technical/architecture/audit-vs-technical-state.md
gfyrag: Confirmed against the apply path and corrected (also swept the earlier sections for the same overstatement). Verified: appendAppliedProposal (internal/infra/state/batch.go) documents 'AppliedProposal carries a map (TransientVolumes) so two nodes will persist byte-divergent Cold-zone entries for the same proposal'; the same non-determinism applies to Log metadata maps and saveIdempotencyKey ('Marshal order need not match across replicas'). Only AuditEntry uses SetProtoDeterministic/MarshalDeterministicVT (sorted map keys) because it feeds the hash chain, and the sealer re-marshals deterministically for the cross-node state hash.

The doc now states Raft + the deterministic FSM guarantee the same LOGICAL proposal and LOGICAL state on every replica, NOT byte-identical serialization for map-bearing projections (naming TransientVolumes / Log metadata / idempotency as examples), and that this is exactly why the checker compares LOGICALLY (decode + compare / replay), not byte-wise. I removed the 'copied bit-for-bit to every node' framing here and in the three earlier passages (State Classes operational row, the invariant-#8 floor paragraph, the mirror-cursor paragraph) plus the checker/README, replacing them with the logical-effect wording so the tampering point survives without the false byte-identity claim.

No new engine gap: I checked compareExclusionProjections (decodes TransientVolumes into a (ledger,account,asset) set) and compareIdempotencyOutcomes (UnmarshalVT + field-wise idempotencyMismatch, map-aware metadataEqual) — both already compare logically, so there is no hidden checker byte-compare to fix. Pushed as 4407c5f73.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,236 @@
...
+definitions stored under `SubAttrPreparedQuery`, cached and bloom-filtered like
+other cache attributes, and they shape which business data an API caller gets
+back. They are created through an audited order (`CreatePreparedQuery`), so the
+audit source of truth exists.
+
+The stored projection is **not yet verified by the checker**: there is no
+`compare*` pass for prepared queries in `internal/application/check`. Until such
+a pass re-derives the expected prepared-query set from the audited create/update/
+delete logs and compares it to what is stored, treat prepared-query integrity as
+a known gap rather than a proven guarantee. When a prepared-query projection
```
NumaryBot: 🟠 **[major] Prepared-query projections treated as future concern rather than current known gap**

ExecutePreparedQuery reads SubAttrPreparedQuery from Pebble to drive user-visible query results today. The document still frames adding a checker pass as something to do 'when' the projection becomes authoritative, but it is already authoritative. This overstates the integrity guarantee and could mislead auditors about the current state of the system. No SubAttrPreparedQuery checker pass exists; the documented passes cover volumes, metadata, transactions, reversion, sealing, exclusions, idempotency, and indexes.

**Suggestion:** Annotate the prepared-query entry explicitly as a current known integrity gap (no checker pass exists for SubAttrPreparedQuery). Change 'when it becomes authoritative' to 'because it is authoritative today'. The checker README gap list must also include prepared queries.

docs/technical/architecture/subsystems/checker/README.md
```diff
@@ -15,6 +15,20 @@ Two pages cover what the checker depends on and what it does.
 
 The audit chain and the checker are tightly coupled by an explicit invariant: **every persisted dataset is either hash-bound (the audit log itself) or derivable from the hash-bound data, in which case the checker must verify it on every `Check()` run** ([CLAUDE.md invariant #8](../../../../../AGENTS.md)). This rule shapes design decisions across every other subsystem — what to refactor versus what to bind — so the checker's coverage is its own first-class architectural concern.
 
+For new persisted state, first classify whether it is business truth, governance
+truth, operational consensus state, or a rebuildable projection using
+[Audit-Bound vs Technical State](../../audit-vs-technical-state.md). Per invariant
+#8, every non-audit dataset persisted in the main Pebble store needs checker
```
NumaryBot, line 21: 🟠 **[major] Checker README gap list omits prepared queries**

The checker README lists only mirror cursor and readstore index contents as current uncovered projections. The audit-vs-technical-state page now acknowledges SubAttrPreparedQuery is persisted and not checker-verified, and ExecutePreparedQuery reads it to drive user-visible query results. Readers using the README as the canonical checker-coverage reference are unaware of this gap.

**Suggestion:** Add SubAttrPreparedQuery (prepared queries) explicitly to the gap list in the checker README, with a note that no checker pass exists for this projection.

NumaryBot, line 21: 🟠 **[major] Bloom-filter exception to invariant #8 is too broad — does not account for persisted restore path**

On normal restart or follower sync, CacheSnapshotter.RestoreFromStore restores persisted cache generations and bloom blocks, falling back to a full attribute scan only when no persisted bloom blocks exist. The current documentation exempts bloom filters broadly from checker coverage, which contradicts invariant #8 and hides the real integrity gap for the normal persisted-block path. Only the backup/restore lifecycle path genuinely discards and rebuilds bloom state.

**Suggestion:** Scope the bloom-filter exception strictly to paths where these datasets are genuinely discarded and rebuilt (e.g. backup/restore). Document normal persisted bloom/cache snapshot restoration as an uncovered integrity gap, or revise invariant #8 explicitly.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,236 @@
...
+  double-application.
+
+Because it is only correct at one exact value, treat `MirrorCursor` as a
+persisted per-ledger projection (`ZonePerLedger` / `SubPLMirrorCursor`, in the
+main Pebble store) that must **equal** the highest audited `MirrorIngest.v2LogId`
+for the target ledger — not merely be less than or equal to it. Explicit edge
+cases:
+
+- **Empty / never-ingested ledger:** the cursor is absent and reads back as `0`
+  (`query.ReadMirrorCursor` defaults to `0`, and `processBatch` maps `0` to
```
NumaryBot, line 115: 🟡 **[minor] Deleted-ledger mirror-cursor lifecycle is inaccurately described**

The document implies the mirror cursor persists after a ledger is deleted or archived. In practice, DeleteLedger queues PendingLedgerCleanups, and when the covering chapter purge reaches the delete sequence, WriteSet.executePurge calls deleteLedgerData which explicitly point-deletes SubPLMirrorCursor. The cursor is therefore only transiently present during the pending-cleanup window, not after cleanup completes. There is also no 'archived' ledger state.

**Suggestion:** Document the transient pending-cleanup case and note that the cursor is expected to be absent once cleanup completes. Remove or correct any reference to an 'archived' ledger state.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1549#issuecomment-4924154173

gfyrag:
Review pass on `4407c5f734b08cf3bc68d45deccc3beb8c1ea9b9`.

Still actionable:

1. Major: the prepared-query gap is still partly framed as future-looking, and the checker README still omits it. The audit-vs-technical-state page says “When a prepared-query projection becomes authoritative…”, but prepared queries are authoritative today: `ExecutePreparedQuery` reads `SubAttrPreparedQuery` through `ReadPreparedQuery` to drive user-visible execution. Please change this to “because it is authoritative today” and add `SubAttrPreparedQuery` to the checker README gap list.

2. Major: the bloom/cache exception remains too broad for normal restart/follower-sync paths. The docs still use bloom filters as the lifecycle exception example, but normal restart restores cache generations and persisted bloom blocks; full scan happens only when no persisted bloom blocks exist. Please scope the exception explicitly to backup/restore paths where bloom blocks are deleted, and document normal persisted cache/bloom restore as either covered by an integrity control or a tracked gap.

3. Low: the index registry source list is still incomplete. `compareIndexes` is described as deriving from create/drop/delete logs, but the checker docs/code also include `RemovedMetadataFieldType` and `DeleteLedger` as index-affecting logs. Please spell out `CreateIndex` / `DropIndex` / `RemovedMetadataFieldType` / `DeleteLedger`.

Resolved since the prior pass: governance-state distinction, mirror cursor equality and deleted-ledger cleanup wording, readstore current-gap/per-replica wording, and the byte-identical Pebble wording.

gfyrag:
Re-reviewed current head `4407c5f734b08cf3bc68d45deccc3beb8c1ea9b9`.

Remaining findings:

1. Major: prepared queries are still framed as future-authoritative.

`docs/technical/architecture/audit-vs-technical-state.md` still says “When a prepared-query projection becomes authoritative...”. They are authoritative today: `ExecutePreparedQuery` reads `SubAttrPreparedQuery` via `ReadPreparedQuery` and compiles that stored filter for user-visible results. `docs/technical/architecture/subsystems/checker/README.md` also lists current uncovered examples but omits prepared queries. Both should treat prepared queries as a current uncovered projection gap.

2. Major: the bloom/cache rebuild exception is too broad for invariant #8.

Normal restart/follower sync restores `ZoneCache` generations and persisted bloom blocks (`CacheSnapshotter.RestoreFromStore` / `restoreBloomFilters`). A full attribute scan happens only when no persisted bloom blocks exist, while backup prep explicitly deletes bloom blocks. Please scope the exception to the actual discard-and-rebuild lifecycle path, and document normal persisted cache/bloom restore either as covered by an integrity control or as a tracked gap. The same issue appears in the checker README.

3. Medium: existing checker docs now contradict the new page on readstore inverted-index coverage.

`docs/technical/architecture/subsystems/checker/audit-chain.md` says the read-side inverted index “is verified on demand by the checker.” The new page correctly documents readstore inverted-index contents as an uncovered current gap, so the existing checker doc should be updated in the same documentation pass.

4. Low: the `compareIndexes` source list is incomplete.

The page summarizes it as deriving from “create/drop/delete logs,” but the checker derives registry state from `CreateIndex`, `DropIndex`, `RemovedMetadataFieldType`, and `DeleteLedger`. Please spell out the exact four log types to match the implementation and checker docs.

What looks fixed in current text:

- Governance-state distinction.
- Mirror cursor equality.
- Deleted-ledger mirror cursor cleanup wording.
- Readstore per-replica mechanism.
- Byte-identical Pebble overstatement.

Checks run:

- `git diff --check 1b5beada...HEAD -- docs/...`
- Local markdown link target check for changed docs.
- `node .gitnexus/run.cjs detect_changes` returned `No changes detected`.

No Go build/tests run; this PR is docs-only.

docs/technical/architecture/audit-vs-technical-state.md
flemzord: Revalidated on `4407c5f73`: this concern is now obsolete. The current text explicitly documents deferred cleanup, `deleteLedgerData` removing the cursor, the transient pending-cleanup window, absence after cleanup, and that there is no archived mirror state.

gfyrag: Thanks — agreed, this is obsolete on `4407c5f73`. The Mirror Cursor "Deleted ledger" edge case (audit-vs-technical-state.md:96-109) now states the cursor **is** cleared as part of cleanup, describes the deferred purge (`DeleteLedger` → `savePendingLedgerCleanup` → `State.PendingLedgerCleanups`), names `WriteSet.executePurge` → `deleteLedgerData` point-deleting `SubPLMirrorCursor` alongside `SubPLMirrorSourceHead`/`SubPLMirrorStatus`/the pending-cleanup marker, and concludes the cursor is present only during the pending-cleanup window and absent afterwards with no "archived" state. Resolving.

docs/technical/architecture/audit-vs-technical-state.md
gfyrag: Fixed in e50ca3855. The Prepared Queries section now states the projection is **already authoritative for business-visible responses today** — a current integrity gap, not a future transition — and cites the read path (`ExecutePreparedQuery` → `ReadPreparedQuery`, `internal/query/executor.go`, straight from Pebble). Removed the "when a prepared-query projection becomes authoritative" wording; closing the gap now reads as requiring a checker `compare*` pass, not a deferred step. Prepared queries are also added to the tracked-gap list in the State Classes floor paragraph and in the checker README.

docs/technical/architecture/subsystems/checker/README.md
gfyrag: Fixed in e50ca3855. The checker README gap list now includes prepared queries (`SubAttrPreparedQuery`, read by `ExecutePreparedQuery` to drive user-visible results, no `compare*` pass) alongside the mirror cursor and readstore inverted-index contents. audit-chain.md is also corrected so it no longer implies every projection is checker-verified today.

gfyrag: Fixed in e50ca3855. The bloom-filter exception is now scoped strictly to the backup/restore discard-and-rebuild path (`internal/infra/attributes/prepare.go` deletes the blocks). The docs now state that on normal restart / follower sync bloom blocks are instead **restored from the persisted Pebble blocks** (`CacheSnapshotter.RestoreFromStore` / `restoreBloomFilters`; full scan only on first boot when no blocks exist), so those persisted blocks are a durably trusted projection between backups and are NOT covered by the rebuild control — documented as a tracked integrity gap in the State Classes row, the floor paragraph, and the checker README gap list.

docs/technical/architecture/subsystems/checker/README.md
gfyrag: Addressed and still accurate on current head e50ca3855. The checker README no longer exempts persisted operational projections: it states the invariant-#8 floor (every non-audit main-store dataset needs checker coverage), scopes the discard-and-rebuild carve-out strictly (bloom only on the backup/restore path; persisted blocks reloaded on restart are a tracked gap), notes Raft cannot detect a value corrupted before it is proposed, and lists the current uncovered projections (mirror cursor, readstore inverted-index contents, prepared queries, restart-path bloom blocks) as tracked gaps rather than exemptions. Resolving.

docs/technical/architecture/audit-vs-technical-state.md
gfyrag: Addressed and still accurate on current head. The Mirror Cursor section requires the live cursor to **equal** the highest audited `MirrorIngest.v2LogId` (not merely <=), and spells out the lowered-cursor double-application path: `Worker.processBatch` reloads the cursor and `FetchLogs` returns logs strictly after it, `TranslateBatch` re-emits `MirrorIngest`, `processMirrorCreatedTransaction` reapplies postings with `force=true` and `applyPosting` adds additively — no `v2LogId` dedup. Resolving.

docs/technical/architecture/audit-vs-technical-state.md
gfyrag: Addressed and still accurate on current head. Readstore is framed as a current integrity gap, not a future transition: once per-replica `IndexVersionState.CurrentVersion` is READY the query compiler compiles filters directly to readstore iterators with no on-scan/audit fallback, while `compareIndexes` verifies only registry presence/identity, never the inverted-index contents. This commit additionally corrected audit-chain.md, which previously listed the read-side inverted index among checker-verified projections. Resolving.

docs/technical/architecture/audit-vs-technical-state.md
gfyrag: Addressed and still accurate on current head (you already revalidated this obsolete on 4407c5f73). The deleted-ledger edge case documents deferred cleanup (`savePendingLedgerCleanup` → `State.PendingLedgerCleanups`), `WriteSet.executePurge` → `deleteLedgerData` point-deleting `SubPLMirrorCursor`, the transient pending-cleanup window, absence after cleanup, and no "archived" state. Resolving.

docs/technical/architecture/audit-vs-technical-state.md
gfyrag: Addressed and still accurate on current head. The doc states the readstore is a separate per-replica Pebble store populated locally by each node index builder from the replicated log/audit stream, `IndexVersionState` is per-replica, and Raft replicates the source audit/log stream — not the readstore bytes — so a bit-flip in one replica readstore is not propagated and Raft neither detects nor repairs replica-local index corruption. Resolving.

docs/technical/architecture/audit-vs-technical-state.md
gfyrag: Addressed and still accurate on current head. The doc states Raft + the deterministic FSM guarantee the same **logical** proposal and logical state, NOT byte-identical serialization for map-bearing projections (naming `AppliedProposal.TransientVolumes`, `Log` metadata, and `saveIdempotencyKey`), and that only `AuditEntry` uses deterministic marshaling because it feeds the hash chain — which is why the checker compares logically (decode/replay), not byte-wise. Resolving.

gfyrag:
All review threads on this docs-only PR are now addressed and resolved at head `e50ca3855`.

The remaining items from the NumaryBot "Changes requested" summary and the flemzord conversation notes were the same three findings, all fixed in `e50ca3855`:

1. **Prepared queries were framed as a future concern.** They are authoritative today — `ExecutePreparedQuery` reads `SubAttrPreparedQuery` via `ReadPreparedQuery` (`internal/query/executor.go`) straight from Pebble to drive user-visible results, and no `compare*` pass exists. The Prepared Queries section now says this explicitly (current integrity gap, not "when it becomes authoritative"), and prepared queries are added to the tracked-gap list in the State Classes floor paragraph and the checker README.

2. **Bloom-filter exception to invariant #8 was too broad.** It now applies strictly to the backup/restore discard-and-rebuild path (`internal/infra/attributes/prepare.go` deletes the blocks). On normal restart / follower sync, `CacheSnapshotter.RestoreFromStore` / `restoreBloomFilters` reload the persisted bloom blocks from Pebble (full scan only on first boot when no blocks exist), so those blocks are a durably trusted projection between backups and are documented as a tracked integrity gap, not an exemption — in the State Classes row, the floor paragraph, and the checker README gap list.

3. **Deleted-ledger mirror-cursor lifecycle** (minor) was already corrected and revalidated obsolete on `4407c5f73`: deferred cleanup → `deleteLedgerData` point-deletes `SubPLMirrorCursor`, transient pending-cleanup window, absent after cleanup, no "archived" state.

Additionally corrected `subsystems/checker/audit-chain.md`, which previously listed the read-side inverted index among checker-verified projections (it distinguishes derivable-in-principle from actual checker coverage and names the current gaps), and spelled out the four `compareIndexes` source log types (`CreateIndex` / `DropIndex` / `RemovedMetadataFieldType` / `DeleteLedger`).

Docs-only; no code touched. `git diff --check` clean.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,259 @@
...
+> If changing the state after the fact can change a ledger business result, a
+> business write acceptance or rejection, or the trust model for business
+> evidence, the state must be audit-bound or verified as a projection of
+> audit-bound data.
+
+## State Classes
+
+| Class | Definition | Examples | Required control |
+| --- | --- | --- | --- |
+| Business truth | Changes balances, transactions, metadata, reversions, ledger lifecycle, indexes, or any user-visible business decision. | `AuditEntry`, `Log`, `Volume`, `Metadata`, `Transaction`, `Reference`, reversion bitsets, idempotency outcomes, index registry. | The command must produce an audit entry, and persisted projections must be checked by `internal/application/check`. |
```
NumaryBot, line 20: 🟠 **[major] Prepared-query projections framed as future-authoritative; no checker pass exists and they are authoritative today**

The document still uses language like 'When a prepared-query projection becomes authoritative…', but `ExecutePreparedQuery` already reads `SubAttrPreparedQuery` from Pebble via `ReadPreparedQuery` to drive user-visible query results. A tampered prepared-query projection can silently affect results today with no detection. Additionally, the checker README gap list omits prepared queries entirely, even though the audit-vs-technical-state page acknowledges no checker pass exists for `SubAttrPreparedQuery`.

**Suggestion:** Change the wording to 'because it is authoritative today' and explicitly list `SubAttrPreparedQuery` as a current known integrity gap in both the audit-vs-technical-state table and the checker README gap list.

docs/technical/architecture/subsystems/checker/README.md
```diff
@@ -15,6 +15,30 @@ Two pages cover what the checker depends on and what it does.
...
```
NumaryBot, line 21: 🟠 **[major] Bloom-filter/cache exception to invariant #8 is too broad — covers normal restart paths, not just backup/restore**
<sub>reported by NumaryBot, gfyrag, flemzord</sub>

`CacheSnapshotter.RestoreFromStore` and `restoreBloomFilters` restore persisted cache generations and bloom blocks on normal restart and follower sync, falling back to a full attribute scan only when no persisted bloom blocks exist. The documentation exempts bloom filters by saying they are 'discarded and rebuilt', but this is only true for the backup/restore path where `prepare.go` explicitly deletes bloom blocks. Normal restart does not discard them, so the exception as written is overly broad and could mislead auditors into thinking normal persisted cache/bloom state is always rebuilt from scratch.

**Suggestion:** Scope the exception explicitly to the backup/restore rebuild lifecycle path (where bloom blocks are deleted before rebuild). Document normal persisted cache/bloom restore separately, either as covered by an integrity control or as a tracked gap.

docs/technical/architecture/subsystems/checker/README.md
```diff
@@ -15,6 +15,30 @@ Two pages cover what the checker depends on and what it does.
...
+path bloom blocks are instead *restored from the persisted Pebble blocks*
+(`CacheSnapshotter.RestoreFromStore` / `restoreBloomFilters`; the full scan runs
+only on first boot when no blocks exist), so those blocks are a durably trusted
+projection between backups and are **not** covered by the rebuild control. Raft
+replication is not a substitute either: it only guarantees every replica applies
+the same logical proposal (it does not even guarantee byte-identical serialization
+for map-bearing projections — see
+[Audit-Bound vs Technical State](../../audit-vs-technical-state.md)), so a value
+corrupted or tampered before it is proposed takes effect on every node and no
+cross-node comparison can detect it. Persisted projections that are not yet
```
NumaryBot, line 39: 🟠 **[major] Persisted governance projections absent from the gap list despite no checker pass**
<sub>reported by codex, NumaryBot, flemzord</sub>

Signing keys (`SubGlobSigningKey`), signing config (`SubGlobSigningConfig`), and maintenance mode (`SubGlobMaintenanceMode`) are read from Pebble into live key store/shared state on recovery, and admission/FSM code uses them to accept or reject writes. No `compare*` pass in `checker.go` covers these projections. The gap list currently mentions only mirror cursor, readstore index contents, prepared queries, and bloom filters, making the coverage map materially incomplete for persisted governance state.

**Suggestion:** Add persisted governance projections (`SubGlobSigningKey`, `SubGlobSigningConfig`, `SubGlobMaintenanceMode`, and related rows) to the checker README gap list, and note in the audit-vs-technical-state table that these require either checker verification or an audit-bound source read.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1549#issuecomment-4924154173

docs/technical/architecture/audit-vs-technical-state.md
flemzord: Revalidated on current head e50ca3855: this finding is stale. audit-vs-technical-state.md now says the projection is already authoritative for business-visible responses today (lines 199-205), names ExecutePreparedQuery -> ReadPreparedQuery, and requires a checker pass because no compare pass exists (lines 207-213). The checker README also lists SubAttrPreparedQuery explicitly among the current uncovered projections.

docs/technical/architecture/subsystems/checker/README.md
flemzord: Revalidated on current head e50ca3855: this is also stale. The README scopes discard-and-rebuild to the backup/restore path where prepare.go deletes persisted blocks (lines 23-26), explicitly says normal restart/follower sync restores those blocks through RestoreFromStore / restoreBloomFilters (lines 27-30), and lists restart-path persisted bloom blocks as a current uncovered integrity gap (lines 36-40).

flemzord (commented):
One additional documentation finding outside the changed hunk: [P3] docs/technical/architecture/subsystems/checker/audit-chain.md:216 links Security to ../../../../security/, which resolves to docs/security/ and does not exist in this repository. Readers following the threat-model reference get a dead link. Please retarget it to an existing security/threat-model document or remove the nonexistent reference.

docs/technical/architecture/subsystems/checker/README.md
gfyrag: Confirmed on head e50ca3855 — the doc already scopes the exception correctly. README lines 23-26 restrict "genuinely discarded and rebuilt" to the backup/restore path (`internal/infra/attributes/prepare.go` deletes the persisted blocks); lines 27-30 state that on normal restart / follower sync the blocks are instead *restored* from Pebble via `CacheSnapshotter.RestoreFromStore` / `restoreBloomFilters`; and lines 36-40 list persisted bloom blocks on the restart path as an uncovered integrity gap. It does not treat normal restart as a rebuild. Resolving.

docs/technical/architecture/audit-vs-technical-state.md
gfyrag: Confirmed on head e50ca3855 — the doc now frames prepared queries as authoritative *today*, not future. audit-vs-technical-state.md lines 199-205 say the stored projection is "already authoritative for business-visible responses today" and trace `ExecutePreparedQuery` -> `ReadPreparedQuery` reading the stored definition straight from Pebble; lines 207-213 state there is no `compare*` pass and that closing the gap requires a checker pass now, not a deferred step. The checker README gap list also names `SubAttrPreparedQuery` among current uncovered projections. Resolving.

docs/technical/architecture/subsystems/checker/README.md
gfyrag: Verified against the code — the finding is correct. `SubGlobSigningKey` / `SubGlobSigningConfig` / `SubGlobMaintenanceMode` are persisted under the Global zone (`store.go:234-241`, written in `batch.go:134-188`), read into the live key store / shared state on recovery (`recovery.go:132-156`: `ReadSigningKeys` -> `keyStore.AddPublicKey`, `ReadSigningConfig` -> `SetRequireSignatures`, `ReadMaintenanceMode` -> `SetMaintenanceMode`), and consulted by admission to accept/reject writes (`admission.go:382` maintenance gate, `:676` signature verification via `keyStore.GetPublicKey`), with no `compare*`/`collect*` pass in `checker.go` covering them. The gap lists indeed omitted them. Fixed in 1ca0a1336: added the governance projections to the checker README gap list and to the audit-vs-technical-state.md floor-note gap summary (the Governance-truth table row already stated the requirement). Resolving.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,265 @@
...
+purely technical and must be redesigned as business-impacting state.
+
+## Cluster Configuration
+
+Cluster configuration is Raft-replicated technical state. Most fields tune
+runtime behavior, such as cache rotation and bloom configuration. `hash_algorithm`
+is more sensitive: it changes how future audit entries are hashed and therefore
+changes the trust model for business evidence.
+
+Changing the hash algorithm does not rewrite historical business truth, but it is
```
NumaryBot, line 243: 🟠 **[major] hash_algorithm cluster-config row lacks integrity coverage documentation**

`LoadFSMStateFromStore` reads `SubGlobClusterConfig` back into `FSMState.HashGenerator`, meaning a tampered persisted row can silently change how future audit entries are hashed. The document presents an operator trail as sufficient for this row, but that is weaker than the integrity controls the page itself mandates for audit-bound or checker-verified projections.

**Suggestion:** Classify the cluster-config / hash-algorithm persisted row as a tracked integrity gap (no checker pass and no audit-bound source read), or require checker/audited-order coverage before treating an operator trail as sufficient for this projection.

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -0,0 +1,265 @@
...
+
+The boundary is about impact, not package location. A value under a technical
+keyspace can become business-impacting if future code starts using it to decide
+whether a business command is accepted or which business data is returned.
+
+Invariant #8 ([AGENTS.md](../../../AGENTS.md)) is the floor: every non-audit
+dataset persisted in the main Pebble store is a projection that the checker must
+verify, unless it is genuinely discarded and rebuilt by a lifecycle path or lives
+in a separate rebuildable side-store outside the checker's scope. "Genuinely
```
NumaryBot, line 32: 🟡 **[minor] Bloom-filter discard-and-rebuild lifecycle carve-out omits cluster-config change path**

The document's exception for bloom filters is scoped to the backup/restore path only, but `applyClusterConfigUpdate` also purges `SubGlobBloom`, calls `BloomFilters.Rebuild`, and signals `StartAsyncBloomPopulate` whenever the bloom config changes. This is a second legitimate discard-and-rebuild path that the current wording silently omits, potentially misleading future integrity reviewers.

**Suggestion:** Extend the lifecycle carve-out description to also cover the cluster-config-change triggered bloom purge-and-rebuild path (`applyClusterConfigUpdate`), so the documented exception accurately reflects all cases where persisted bloom blocks are genuinely discarded before being re-served.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1549#issuecomment-4924154173

docs/technical/architecture/audit-vs-technical-state.md
gfyrag: Verified — the finding is correct. `LoadFSMStateFromStore` (`internal/infra/state/fsmstate.go:188-197`) reads `SubGlobClusterConfig` via `ReadClusterState` and rebuilds `FSMState.HashGenerator = processing.NewHashGenerator(clusterState.GetConfig().GetHashAlgorithm(), clusterID)`, and that same `HashGenerator` computes the audit-chain hash for every future entry (`machine.go:1454` `HashGenerator.Compute(...)`, `:1425` sets `entry.HashVersion`). So a persisted row tampered on disk silently changes how future audit entries are hashed. No `compare*` pass in `internal/application/check` covers `SubGlobClusterConfig` and there is no audit-bound read-back. Fixed in c923bbd8b: reclassified the cluster-config / hash_algorithm row as a tracked integrity gap (no checker pass + no audit-bound source read), consistent with how the page treats the other uncovered projections, and noted that an operator trail alone is weaker than the controls the page mandates — closing it needs an audited governance order or a checker pass. Resolving.

docs/technical/architecture/audit-vs-technical-state.md
gfyrag: Verified — correct. `applyClusterConfigUpdate` (`internal/infra/state/machine_technical_updates.go:169-198`) purges the `SubGlobBloom` blocks via `DeleteRangeNoSync`, calls `fsm.BloomFilters.Rebuild(cfg)`, and signals the bloom-rebuild channel that drives `StartAsyncBloomPopulate` (`recovery.go:236`) on a bloom-config change — a second legitimate discard-and-rebuild path. Fixed in c923bbd8b: extended the bloom discard-and-rebuild carve-out to also list the cluster-config-change path (in the state-class table row, the floor note in audit-vs-technical-state.md, and the mirrored carve-out in the checker README), not just backup/restore. Resolving.

docs/technical/architecture/subsystems/checker/README.md
```diff
@@ -15,6 +15,37 @@ Two pages cover what the checker depends on and what it does.
...
```
NumaryBot, line 39: 🟡 **[minor] Checker gap list omits SubGlobClusterConfig/hash_algorithm and SubGlobChapterSchedule**

The gap list is presented as the current complete set of persisted projections without checker coverage, but it omits `SubGlobClusterConfig`/`hash_algorithm` and `SubGlobChapterSchedule`. Both are main-store governance projections with no `compare*` pass: recovery reads cluster config back into `FSMState.HashGenerator`, and reads the chapter schedule back into the scheduler/lifecycle state. A reader relying on this checker README would still miss two tracked integrity gaps that the new classification rule requires documenting. Note that the prior discussion resolved the broader persisted-governance-projection issue (Thread 5/6), but this specific enumeration gap in the checker README was not explicitly addressed there.

**Suggestion:** Add `SubGlobClusterConfig` (hash_algorithm) and `SubGlobChapterSchedule` explicitly to the tracked-gap list in the checker README, alongside the already-listed mirror cursor, readstore inverted-index contents, prepared queries, and restart-path bloom blocks.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1549#issuecomment-4924154173

docs/technical/architecture/subsystems/checker/README.md
NumaryBot, line 39: 🟡 **[minor] Tracked-gap list omits SubGlobClusterConfig/hash_algorithm and SubGlobChapterSchedule**

The tracked-gap list is presented as the current inventory of persisted projections without checker coverage, but it omits SubGlobClusterConfig/hash_algorithm and SubGlobChapterSchedule. Both are main-store rows, have audited intent records, and have no compare* pass. The audit-vs-technical-state page already classifies these as governance-sensitive. This inconsistency could mislead future reviewers into treating those rows as covered or exempt.

**Suggestion:** Add SubGlobClusterConfig/hash_algorithm and SubGlobChapterSchedule to the tracked-gap list in the checker README so the gap inventory is consistent with the classification rules the document itself establishes.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1549#issuecomment-4924154173

flemzord (approved):
Re-reviewed current head 3f553fdf7df17265360a89a76b33598fa8d228a9. No High or Critical findings remain. The current checker-gap inventory omissions and the dead Security-doc link are documentation-level Low findings and are non-blocking under the requested threshold.

PR #1550, opened 2026-07-09
fix(EN-1460): per-replica readiness marker for query-checkpoint reads

gfyrag:
## EN-1460 — Query-checkpoint reads race the async read-index materialization

Fixes the opaque, non-retryable `code=Unknown` returned when a read targets a query checkpoint whose read index is not yet materialized. Supersedes the closed #1545. Design **Option A** (per-replica local `.ready` marker, no reconciler), validated with the team.

Jira: https://formance-team.atlassian.net/browse/EN-1460 (epic EN-867)

### Root cause
`CreateQueryCheckpoint` already waited via `readStore.WaitForSequence(logs[0].Sequence)`, but that only gates on the index-builder **progress cursor**. In `process_logs.go`, `WriteProgress` is flushed in the batch that runs **before** `createReadIndexCheckpoint` creates the `readindex/` directory, so the cursor reaches the target sequence ~100-150 ms before the directory exists. The wait returned early, the immediate read opened a missing Pebble dir, and the read-side error wrapper fell through to the default sanitizer → `code=Unknown` (non-retryable; the default gRPC retry policy only covers `UNAVAILABLE`).

### Fix (Option A)
- **Atomic per-replica materialization.** The index builder materializes the read-index checkpoint at the exact moment it crosses the `CreatedQueryCheckpoint` log — so the snapshot is exactly `MaxSequence` (point-in-time correct) — into a temp dir, fsyncs, atomically renames into place, then writes the `.ready` marker **last**. A crash before the marker leaves nothing half-built; a reader never sees a partial checkpoint. Bounded retry on the pebble hard-link vs concurrent-compaction race (`link … no such file or directory`), cleaning the temp dir between attempts.
- **`CreateQueryCheckpoint` blocks on the creator node's local marker** (`WaitForCheckpoint`), replacing the `WaitForSequence`-on-cursor fast path. A read routed back to the creator succeeds immediately.
- **Typed read gate** (`openCheckpointStores`): marker absent + checkpoint present in the replicated `QueryCheckpointState` registry → `ErrCheckpointNotReady` → gRPC `Unavailable` (reason `CHECKPOINT_NOT_READY`); marker absent + not in registry → `NotFound` (permanent). Never `Unknown`. Mirrors the per-replica `INDEX_BUILDING → Unavailable` pattern for metadata indexes.

### Why no reconciler
Checkpoint reads are served **locally per node** (no leader routing), and the codebase already solves "is my local replica ready?" for secondary indexes with a **local per-replica marker** (`IndexVersionState.CurrentVersion`), not a Raft map — so the marker alone is the right primitive. A background reconciler that rebuilds a checkpoint at a past `MaxSequence` is **infeasible** (system logs are purged per chapter after cold-storage archival — `store.go:153`, `executePurge` — so the logs needed to replay are gone from the hot store) and **unnecessary** (inline materialization is already exactly point-in-time). Unrecoverable checkpoints (a node that crashes between rename and marker, or purged the logs before its builder reached them) degrade to a retryable `Unavailable` that never self-heals → the client recreates (deletion makes reads `NotFound`), aligned with the existing `AcquireCheckpoint` client workaround. Never wrong data.

### Contract (documented in `docs/.../read-path/query-checkpoints.md`)
- After `Create` returns, the checkpoint is immediately readable **on the creator node**.
- Cross-node reads return retryable `Unavailable` until that node's builder crosses the log inline, then succeed (per-replica eventual readiness).
- Unrecoverable / unknown ids → `NotFound` → client recreates.

### Invariants
- **#2/#3**: all changes are read-side + the inline index builder, **off the FSM hot path**. No new FSM proposer, no Raft command, no Pebble reads added to the FSM.
- **#8**: the `.ready` marker and checkpoint directories are rebuildable filesystem lifecycle state (a projection of the audit log), checker-exempt — consistent with the existing exclusion of checkpoint dirs and of `BuildStatus`.
- **#7**: a not-ready read is a loud typed error (`Unavailable`/`NotFound`), never a silent skip.

### Tests
- Unit: read-before-ready → typed `Unavailable` not `Unknown` (`grpc/errors_test.go`); `WaitForCheckpoint` fast-path / blocks-until-marker / context-cancel (missed-wakeup fix); atomic materialization — marker written last & openable, no temp dir left, no-op when ready, rebuild of an unmarked dir; hard-link retry contract.
- E2E: single-node create → immediate read on the creator succeeds; multi-node (3 nodes) per-replica eventual readiness (`Eventually`, no `time.Sleep`) with transient errors asserted `Unavailable` not `Unknown`; `NotFound` for an unknown checkpoint id.

### Proto
Append-only `ERROR_REASON_CHECKPOINT_NOT_READY = 63`; regenerated with all protos in one `protoc` invocation (the earlier single-file regen left a cross-file vtproto fallback for `Log.ResponseSignature` — the cause of the previous `Dirty` failure, now fixed).

### Checks
`GOROOT= go build ./...`, `golangci-lint` (changed pkgs, 0 issues), unit tests across `internal/...` green, e2e compiles (`go vet -tags e2e`), `go mod tidy` clean.


internal/adapter/grpc/server_bucket.go
```diff
@@ -414,6 +414,31 @@ func (impl *BucketServiceServerImpl) openCheckpointStores(checkpointID uint64) (
...
+	// each node's index builder, and on a follower the builder may not have
+	// caught up to the checkpoint's log sequence yet. Opening a
+	// not-yet-materialized directory would surface an opaque, non-retryable
+	// Unknown (EN-1460). Distinguish two cases via the .ready marker:
+	//   - marker present  -> materialized on this replica, open it.
+	//   - marker absent   -> either the checkpoint is registered in Raft but not
+	//     yet materialized here (transient: ErrCheckpointNotReady -> Unavailable,
+	//     the reconciler will materialize it), OR the checkpoint id does not
+	//     exist at all (permanent: NotFound so clients stop retrying).
+	if !readstore.CheckpointDirReady(readIndexPath) {
```
NumaryBot, line 432: 🟠 **[major] Handle legacy checkpoints without .ready markers**

On upgrades from builds that created query checkpoints before this marker existed, the read-index directory can be complete and openable but lacks `.ready`; this gate will classify every registered legacy checkpoint as `ErrCheckpointNotReady` forever because the index builder will not replay the old `CreatedQueryCheckpoint` log once its progress is already past it. That makes existing checkpoint IDs permanently unreadable after deployment unless they are deleted/recreated, so this path needs a migration/adoption strategy for pre-marker checkpoints.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1550#issuecomment-4924411639

internal/adapter/grpc/server_bucket.go
gfyrag: This is a **deliberate, accepted design decision** (validated with the team on EN-1460), not an unhandled case — see the "Why no reconciler" section of the PR description and the Readiness contract in `docs/.../read-path/query-checkpoints.md`.

Rationale:
- **Auto-adopting an unmarked directory is unsafe.** Pebble hard-links SSTs last, so a directory can be present and openable yet incomplete (a crash mid-`Checkpoint()` leaves a valid manifest with missing SSTs; a read-only open *succeeds* and only fails on iteration). The `.ready` marker is the *only* sound complete-vs-incomplete discriminant — treating any openable dir as ready would risk serving a half-linked checkpoint, which is worse than the bug.
- **Reconstruction is infeasible.** Rebuilding a legacy checkpoint at its `MaxSequence` requires replaying logs to that point, but system logs are purged per chapter after cold-storage archival (`store.go:153`, `executePurge`), so the logs are generally gone from the hot store. There is no hot-path cold-storage replay.
- **The blast radius is bounded and safe.** Pre-marker checkpoints only exist on `v3.0.0-alpha.*` deployments (query checkpoints shipped in alphas only; no stable release). Such a checkpoint degrades to a retryable `Unavailable` (never wrong data), and the operator/client recreates it — exactly the existing `AcquireCheckpoint` client workaround (delete-and-recreate on timeout). After deletion the read returns `NotFound`.
- **Consistent with the established pattern.** Per-replica readiness for secondary indexes (`IndexVersionState`, EN-1323) is also local and non-reconstructing; this mirrors it.

No change — this is the documented contract, not a regression.

gfyrag (commented):
Found one crash-safety issue in the new checkpoint materialization path.

`internal/application/indexbuilder/process_logs.go:212` still persists the read-index progress cursor at the `CreatedQueryCheckpoint` log before `createReadIndexCheckpoint` writes the durable `.ready` marker. If the node crashes after this batch commits but before checkpoint materialization finishes, restart begins after the checkpoint log, so the builder never replays it and never recreates the marker. The checkpoint remains registered in `QueryCheckpointState`, so `openCheckpointStores` classifies it as existing-but-not-ready and returns `ErrCheckpointNotReady` forever.

That contradicts the crash-safety guarantee in the commit message and the comment below. The cursor needs to remain before the checkpoint log until the marker is durable, or the restart path needs another deterministic way to materialize/recover markerless registered checkpoints.

internal/adapter/grpc/server_bucket.go
```diff
@@ -414,6 +414,31 @@ func (impl *BucketServiceServerImpl) openCheckpointStores(checkpointID uint64) (
...
+		exists, existsErr := impl.queryCheckpointExists(checkpointID)
+		if existsErr != nil {
+			return nil, nil, existsErr
+		}
+
+		if !exists {
```
NumaryBot: 🟠 **[major] Replica lag can cause permanent NotFound instead of retryable not-ready error**

In a multi-node cluster, a client can create a checkpoint on node A and immediately attempt to read it through node B before B has applied the `CreatedQueryCheckpoint` log. At that point B has no `.ready` marker and its local registry has no record of the checkpoint ID, so the gate returns a permanent `NotFound`. A client following the documented contract will stop retrying rather than treating this as a transient not-ready condition, making a valid in-flight checkpoint appear to have never been created.

**Suggestion:** Distinguish between 'checkpoint ID unknown to this replica' (which may be a replica-lag transient state) and 'checkpoint ID is known but its `.ready` marker is absent'. Return a retryable not-ready / unavailable status for the former case, or document clearly that callers must retry `NotFound` for a bounded window after creation, and update client-facing documentation accordingly.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1550#issuecomment-4924411639

gfyrag (commented):
Review du nouveau commit `e0ff1d8d9ca7eba14f5037204f230e22e8b41b6e`: le finding précédent reste valide.

Le commit ajoute des tests utiles pour `createReadIndexCheckpoint` / `deleteReadIndexCheckpoint` : nettoyage du `.tmp` stale, marker `.ready`, no-op si déjà ready, suppression idempotente. En revanche il ne couvre pas la fenêtre critique signalée sur `process_logs.go` : le cursor/progress est toujours écrit et flushé avant la matérialisation du checkpoint read-index et son marker `.ready`.

Conséquence inchangée : crash après le commit du progress à `CreatedQueryCheckpoint` mais avant `MarkCheckpointReady` => au redémarrage le builder repart après ce log, ne recrée pas le checkpoint, et `openCheckpointStores` continue à voir un checkpoint non ready / absent.

Je n'ai pas relancé les tests sur ce head ; le nouveau commit est test-only et le problème est dans l'ordre durable du code de production, inchangé par rapport au head précédent.

internal/adapter/grpc/server_bucket.go
```diff
@@ -414,6 +414,31 @@ func (impl *BucketServiceServerImpl) openCheckpointStores(checkpointID uint64) (
...
+			return nil, nil, commonpb.NewNotFoundError("query checkpoint %d not found", checkpointID)
```
Azorlogh: `queryCheckpointExists` reads the local per-node `QueryCheckpointState`, which lags on followers (async FSM apply). Checkpoint reads deliberately skip the `waitMinLogSequence` barrier (`AggregateVolumes` line 1557: *"minLogSequence only gates live reads"*), so a read routed to a follower that hasn't applied the `CreatedQueryCheckpoint` entry yet returns permanent `NotFound` for a checkpoint that genuinely exists — breaking the documented "NotFound = permanent" contract, and inconsistent with `GetQueryCheckpointInfo` (server_cluster.go:540) which treats the same nil registry as *transient, retry*. A permanent NotFound can't be soundly derived from a possibly-stale local read: either return retryable `Unavailable` when the registry entry is absent, or gate the existence check behind a ReadIndex/WaitForApplied barrier.

internal/adapter/grpc/server_bucket.go
```diff
@@ -414,6 +414,31 @@ func (impl *BucketServiceServerImpl) openCheckpointStores(checkpointID uint64) (
...
 
+	// The read-index checkpoint is materialized asynchronously and per-replica by
+	// each node's index builder, and on a follower the builder may not have
+	// caught up to the checkpoint's log sequence yet. Opening a
+	// not-yet-materialized directory would surface an opaque, non-retryable
+	// Unknown (EN-1460). Distinguish two cases via the .ready marker:
+	//   - marker present  -> materialized on this replica, open it.
+	//   - marker absent   -> either the checkpoint is registered in Raft but not
+	//     yet materialized here (transient: ErrCheckpointNotReady -> Unavailable,
+	//     the reconciler will materialize it), OR the checkpoint id does not
```
Azorlogh: Stale: there is no reconciler in Option A — the checkpoint is materialized inline by the index builder.

internal/storage/readstore/store.go
```diff
@@ -556,6 +556,116 @@ func (s *Store) ListBackfillProgress() ([]BackfillEntry, error) {
...
 }
 
+// checkpointReadyMarker is the sentinel file the index builder writes into a
+// query checkpoint read-index directory as the final step, only after the whole
+// directory has been atomically renamed into place. Its presence is the single
+// authoritative per-replica readiness signal: pebble hard-links SSTs last and a
+// checkpoint can fail mid-link (EN-1460's "link ... no such file or directory"),
+// so a directory or manifest merely existing is NOT sufficient — a half-written
+// or half-linked directory is indistinguishable from a complete one except by
+// the marker. The reconciler therefore never trusts an unmarked directory; it
```
Azorlogh: Stale `reconciler` reference — no reconciler exists; the index builder rebuilds an unmarked dir inline.

internal/application/indexbuilder/process_logs.go
```diff
@@ -344,13 +353,108 @@ func (b *Builder) indexPayload(
...
+
+		return fmt.Errorf("renaming checkpoint %d into place: %w", checkpointID, err)
+	}
+
+	// fsync the parent so the rename is durable before we vouch for it.
+	if err := readstore.FsyncDir(filepath.Dir(finalDir)); err != nil {
+		return fmt.Errorf("fsync checkpoint %d parent: %w", checkpointID, err)
+	}
+
+	// Write the readiness marker LAST: readers and the reconciler only treat the
```
Azorlogh: `and the reconciler` — leftover; only readers gate on the marker.

internal/application/indexbuilder/process_logs.go
```diff
@@ -238,9 +239,17 @@ func (b *Builder) processLogs(ctx context.Context, cursor uint64, deadline time.
...
+		// Materialize a query checkpoint inline, at the exact moment the builder
+		// crosses the CreatedQueryCheckpoint log — so the live read index
+		// reflects precisely MaxSequence (the checkpoint's point-in-time). The
+		// materialization is atomic (temp dir + fsync + rename + .ready marker
+		// last), so a reader never observes a partial checkpoint. There is no
+		// reconciler: historical reconstruction to a past MaxSequence is
+		// infeasible (chapter log purge) and unnecessary (this inline point is
+		// already exactly point-in-time). A node that crashes between rename and
+		// marker, or that purged the logs before reaching this point, will never
+		// have a marker for this checkpoint; reads there return NotFound and the
```
Azorlogh: Inaccurate: the checkpoint stays registered in `QueryCheckpointState`, so reads on such a node return retryable `Unavailable` (`ErrCheckpointNotReady`), not `NotFound` — it only becomes `NotFound` once the client deletes it. The PR description states this correctly; this comment doesn't.

internal/storage/readstore/store.go
```diff
@@ -556,6 +556,116 @@ func (s *Store) ListBackfillProgress() ([]BackfillEntry, error) {
...
+// creation).
+//
+// The index builder calls NotifyProgress after each materialization, waking
+// waiters to re-check the marker.
+func (s *Store) WaitForCheckpoint(ctx context.Context, dirPath string) error {
+	if CheckpointDirReady(dirPath) {
+		return nil
+	}
+
+	// Broadcast on cancellation while holding progressMu. Taking the lock is
```
Azorlogh, line 645: Nit: this closes the *cancellation* window, but the *progress* wakeup `NotifyProgress` (line 211) still `Broadcast()`s without holding `progressMu`, so a materialization that lands between the marker check and `cond.Wait()` can still be missed (self-heals on the next notify, bounded by ctx). Low severity, pre-existing pattern.

internal/application/indexbuilder/process_logs.go
```diff
@@ -344,13 +353,108 @@ func (b *Builder) indexPayload(
...
+		if err == nil {
+			break
+		}
+
+		// A concurrent compaction can delete an SST between the manifest
+		// snapshot and the hard-link, surfacing as ErrNotExist. This is
+		// transient — retry after clearing the partial temp directory so the
+		// next CreateCheckpoint starts from a clean slate.
+		if !errors.Is(err, os.ErrNotExist) {
+			_ = os.RemoveAll(tmpDir)
```
Azorlogh: The four error-path `os.RemoveAll(tmpDir)` calls (here, 428, 436, 443) could collapse into a single `defer` on a named return error; only the between-retries cleanup needs to stay explicit.

internal/application/indexbuilder/process_logs.go
```diff
@@ -344,13 +353,108 @@ func (b *Builder) indexPayload(
...
+
+			return fmt.Errorf("creating read index checkpoint %d: %w", checkpointID, err)
+		}
+
+		b.logger.WithFields(map[string]any{
+			"checkpointID": checkpointID,
+			"attempt":      attempt + 1,
+			"error":        err,
+		}).Infof("Read index checkpoint hard-link raced compaction, retrying")
```
Azorlogh: Logs "retrying" on the final attempt too, where no retry follows.

Azorlogh (commented):
Review focused on edge cases, leftover bits, and simplifications. Main concern is the permanent `NotFound` on a lagging follower (inline on `server_bucket.go`); the rest are stale `reconciler` comments from the superseded design, one inaccurate comment, and a couple of nits.

internal/adapter/grpc/server_bucket.go
gfyrag: Confirmed and fixed in 09dc989a1. Traced it: checkpoint reads go `readController(cpID)` → `openCheckpointStores` and deliberately skip the `waitMinLogSequence` barrier (`AggregateVolumes` etc. gate it on `checkpointId == 0`), so on a follower whose FSM has not applied the `CreatedQueryCheckpoint` entry, `queryCheckpointExists` reads a lagging local `QueryCheckpointState` → false → permanent `NotFound` for a checkpoint that exists cluster-wide. Real bug, and it violated the "never permanent/opaque for a transient case" principle.

Fix: `resolveMissingMarker` now only concludes `NotFound` after a linearizable barrier proves the local FSM is caught up. Absent locally → `node.ReadIndexAndWait(ctx)` (the same quorum barrier live reads use) → re-check:
- registered (before or after barrier) → `ErrCheckpointNotReady` (Unavailable): materialization pending on this replica;
- barrier inconclusive (`ErrNodeSyncing` / no leader) → `ErrCheckpointNotReady` (Unavailable): can't prove absence, stay retryable;
- still absent after a successful barrier → `NotFound` (genuinely doesn't exist cluster-wide).

Threaded `ctx` through `openCheckpointStores`/`readController`. No new leader-consultation primitive — `ReadIndexAndWait` is exactly what `RoutedController.readCtrl` already does for linearizable reads, so this is applying the existing barrier, not a new compromise. Multi-node e2e covers the lagging-follower path (transient errors asserted `Unavailable`, never `Unknown`/`NotFound`; eventual success on every node).

internal/adapter/grpc/server_bucket.go
gfyrag: Fixed in 09dc989a1 — this comment was already rewritten to describe `resolveMissingMarker`; no `reconciler` reference remains here.

internal/storage/readstore/store.go
gfyrag: Fixed in 09dc989a1 — reworded to 'The index builder therefore never trusts an unmarked directory; it discards and rebuilds from scratch.'

internal/application/indexbuilder/process_logs.go
gfyrag: Fixed in 09dc989a1 — now 'readers (openCheckpointStores) only treat the checkpoint as ready once this exists'.

internal/application/indexbuilder/process_logs.go
gfyrag: Fixed in 09dc989a1 — corrected to match the PR contract: the checkpoint stays registered, so reads return the retryable `ErrCheckpointNotReady` (Unavailable) until the client deletes and recreates it (not NotFound).

internal/storage/readstore/store.go
gfyrag: Good catch — fixed in 09dc989a1. `NotifyProgress` now takes `progressMu` around `Broadcast()`. A waiter checks its condition and calls `cond.Wait()` under the same lock, and `Wait` atomically releases only once parked, so the broadcast now serializes against that window: it either lands before the waiter locks (re-checked on acquire) or after it parks (woken). Closes the missed-wakeup for both `WaitForCheckpoint` and `WaitForSequence`, consistent with the cancellation fix. Race detector clean.

internal/application/indexbuilder/process_logs.go
gfyrag: Done in 09dc989a1 — collapsed into a `defer` on a named return (`err`) that `os.RemoveAll(tmpDir)` only on error; after a successful rename tmpDir is gone so it's a no-op. Kept the explicit cleanup between retries (per intent), removed the 3 duplicated error-path removals.

internal/application/indexbuilder/process_logs.go
gfyrag: Fixed in 09dc989a1 — the 'retrying' log and inter-attempt cleanup are now guarded by `attempt < checkpointLinkRetries-1`, so the final attempt no longer logs a retry; the give-up path is handled by the deferred cleanup + the after-loop error return.

internal/adapter/grpc/server_bucket.go
```diff
@@ -410,10 +410,21 @@ func (impl *BucketServiceServerImpl) computeTransactionReceipt(ctx context.Conte
...
 	readIndexPath := impl.store.QueryCheckpointReadIndexDir(checkpointID)
 
+	// The read-index checkpoint is materialized asynchronously and per-replica by
+	// each node's index builder, at the moment its builder crosses the
+	// CreatedQueryCheckpoint log. Opening a not-yet-materialized directory would
+	// surface an opaque, non-retryable Unknown (EN-1460). The .ready marker
+	// distinguishes the cases:
+	//   - marker present -> materialized on this replica, open it.
+	//   - marker absent  -> resolve readiness vs existence (see resolveMissingMarker).
+	if !readstore.CheckpointDirReady(readIndexPath) {
```
NumaryBot, line 432: 🟠 **[major] Readiness gate covers only the read-index store, not the main checkpoint**

The `.ready` marker proves only that the read-index store is complete. The main-store checkpoint (`{id}/main`) is created asynchronously by the applier on the same replica. If the index builder writes `readindex/.ready` before the main checkpoint has finished materializing, the gate passes but the subsequent attempt to open the main Pebble store fails with an opaque non-retryable `Unknown` error, reintroducing the race condition the PR is designed to eliminate.

**Suggestion:** Either extend the `.ready` marker to cover both stores (write it only after both the read-index and main checkpoints are fully materialized), or treat a missing/unopenable main checkpoint for a registered checkpoint ID as a not-ready condition and return a retryable status rather than a hard error.

internal/application/indexbuilder/process_logs.go
```diff
@@ -344,13 +354,114 @@ func (b *Builder) indexPayload(
...
+	// a harmless no-op on the happy path.
+	defer func() {
+		if err != nil {
+			_ = os.RemoveAll(tmpDir)
+		}
+	}()
+
+	for attempt := range checkpointLinkRetries {
+		err = b.readStore.CreateCheckpoint(tmpDir)
+		if err == nil {
```
NumaryBot, line 411: ⚪ **[nit] Redundant os.RemoveAll calls could be collapsed into a defer**

The four error-path `os.RemoveAll(tmpDir)` calls (lines ~411, 428, 436, 443) duplicate cleanup logic. Only the between-retries cleanup needs to be explicit; the rest can be a single `defer` on a named return error.

**Suggestion:** Use a `defer func() { if err != nil { os.RemoveAll(tmpDir) } }()` pattern and remove the redundant individual calls.

internal/application/indexbuilder/process_logs.go
```diff
@@ -344,13 +354,114 @@ func (b *Builder) indexPayload(
...
+			break
+		}
+
+		// A concurrent compaction can delete an SST between the manifest
+		// snapshot and the hard-link, surfacing as ErrNotExist. This is
+		// transient — retry after clearing the partial temp directory so the
+		// next CreateCheckpoint starts from a clean slate.
+		if !errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("creating read index checkpoint %d: %w", checkpointID, err)
```
NumaryBot, line 420: ⚪ **[nit] Log message says 'retrying' on the final attempt**

The log line that says 'retrying' is also emitted on the last iteration of the retry loop, where no retry actually follows, which is misleading.

**Suggestion:** Only log 'retrying' when there are remaining attempts, or rephrase the message to 'attempt failed' so it is accurate on both intermediate and final iterations.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1550#issuecomment-4924411639

internal/adapter/grpc/server_bucket.go
gfyrag: Legitimate — confirmed and fixed in 848508e5e. Traced the ordering: the applier creates the main checkpoint ({id}/main) during apply of the CreatedQueryCheckpoint entry (`handleQueryCheckpointRequired` → `createMainStoreCheckpoint`), while the index builder writes `readindex/.ready` when it independently tails the same committed log — so the marker can indeed precede the main checkpoint, and opening a not-yet-materialized main store surfaced a raw Unknown.

Fix: on a checkpoint whose read index is ready, any main-store open failure is now routed through `resolveMissingMarker` (retryable `ErrCheckpointNotReady` for a registered checkpoint; `NotFound` only after a barrier confirms non-existence) — never a raw Unknown. I deliberately did not stat a pebble-internal file (v2 uses an atomic marker, not `CURRENT`); the read-only open is the readiness probe for the marker-less main store.

internal/adapter/grpc/server_bucket.go
gfyrag: Already fixed in 09dc989 (this review ran against an earlier commit). `resolveMissingMarker` now only returns NotFound after `node.ReadIndexAndWait` confirms the local FSM is caught up to the cluster commit index; a lagging follower returns retryable `ErrCheckpointNotReady` (Unavailable), never a permanent NotFound. Same fix Alix requested — see the thread on `openCheckpointStores`.

internal/application/indexbuilder/process_logs.go
gfyrag: Already done in 09dc989 — collapsed into a `defer` on a named return; explicit cleanup kept only between retries.

internal/application/indexbuilder/process_logs.go
gfyrag: Already done in 09dc989 — the 'retrying' log + inter-attempt cleanup are guarded by `attempt < checkpointLinkRetries-1`, so the final attempt no longer logs a retry.

internal/adapter/grpc/server_bucket.go
gfyrag: As before: this is the accepted design decision (validated with the team), not a defect — see the 'Why no reconciler' section of the PR description and the readiness contract in the docs. Auto-adopting an unmarked dir is unsafe (pebble links SSTs last → complete-vs-incomplete is indistinguishable without the marker), reconstruction is infeasible (per-chapter log purge), pre-marker checkpoints only exist on alphas, and the failure mode is retryable then client-recreate, never wrong data.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1550#issuecomment-4924411639

Azorlogh (approved):

PR #1552, opened 2026-07-09
fix(events): use ReplacingMergeTree for the ClickHouse sink table

Azorlogh:
## Summary

The ClickHouse events sink created its table with plain `MergeTree`. Because event delivery is **at-least-once**, this silently accumulates duplicate rows on failure. This switches the engine to `ReplacingMergeTree` keyed on the existing `(ledger, log_sequence)` sort key.

## Why duplicates happen

The emitter commits the ClickHouse INSERT batch **before** advancing the Raft cursor (a bounded, best-effort proposal). If the process crashes, loses leadership, or the cursor-advance times out *after* the insert succeeded but *before* the cursor commits, the emitter restarts from the old cursor and **re-publishes the same batch** — producing duplicate rows with an identical `(ledger, log_sequence)`. With plain `MergeTree` these persist forever, so `count()` / `sum()` analytics over-count.

## Change

```sql
-ENGINE = MergeTree()
+ENGINE = ReplacingMergeTree()
 ORDER BY (ledger, log_sequence)
```

- `(ledger, log_sequence)` already uniquely identifies an event and is already the sort key, so `ReplacingMergeTree` collapses re-delivered duplicates during background merges.
- Redeliveries are **byte-identical**, so no version column is needed — it doesn't matter which copy wins.
- No insert-time cost.

## Parity with v2

The schema comment states it matches the v2 reference implementation. Confirmed against `formancehq/ledger` `internal/replication/drivers/clickhouse/driver.go`, which uses `ENGINE = ReplacingMergeTree`. So this **restores parity** — plain `MergeTree` was the divergence.

## Caveat for query authors

Dedup is **eventual** (happens on part merges, not at insert time). Between a duplicate insert and the next merge, a naive `SELECT count()` sees both rows. Exact queries must use `FINAL` or explicit dedup (`GROUP BY` / `argMax`). Documented in `events.md` and in code.

## Notes

- Only applies to **newly created** tables (`CREATE TABLE IF NOT EXISTS`); existing deployments are out of scope.
- Orthogonal follow-up not included here: v2 also uses `PARTITION BY ledger`; a partitioning strategy for retention/TTL/merge efficiency is worth a separate change.

## Tests

- `TestClickHouseCreateTableDDL` asserts the emitted engine.
- New `TestClickHouseSinkIntegration_RedeliveryIsIdempotent` publishes an identical batch twice (the at-least-once redelivery scenario) and asserts the deduplicated row count (`SELECT count() ... FINAL`) does not grow. Uses `FINAL` so it is deterministic regardless of merge timing.

Full `clickhouse`-tagged events package passes.


flemzord (approved):

gfyrag (approved):
LGTM. Verified the core mechanics:

- `(ledger, log_sequence)` is genuinely unique per event (1 log → 1 event, `LogSequence` is the global monotonic sequence), so ReplacingMergeTree can only collapse true redeliveries — never merge two distinct events.
- Redeliveries are byte-identical: every row field derives from the persisted log (`InsertedAt` from `tx.GetInsertedAt()`, not wall-clock), so no version column needed.
- The idempotency test is deterministic — `count() ... FINAL` forces merge-time dedup at read time regardless of background-merge timing.

Restores parity with the v2 reference implementation. Clean fix, good test coverage.

PR #1553, opened 2026-07-09
fix(operator): stop ledgerctl OTEL flush from blocking in-pod execs (index reconcile timeout)

sylr:
## Problem

Creating indexes from the Ledger CRD fails: `IndexesSynced=False` with
```
ledgerctl indexes: exec failed: context deadline exceeded (stdout: []) (stderr: )
```
on every mirror ledger (`appliedIndexes: null`).

## Root cause (reproduced live in `eks-mews-staging-euc1-01/ohbajrwzuthm-tfjv`)

Every `ledgerctl` command the operator execs **inside a ledger pod** hangs for exactly **5.0s**, and the operator kills it at its **5.0s** `ledgerExecTimeout`.

The 5s is an OpenTelemetry penalty, not the RPC:

| Test (in `ledger-…-0`) | Time |
|---|---|
| `ledgerctl indexes list …` (operator's exact cmd) | 5.01s |
| `ledgerctl version` (no RPC) | 5.01s |
| `version` with `OTEL_TRACES_EXPORTER=none` | 0.00s |
| `version` with standard `OTEL_EXPORTER_OTLP_ENDPOINT` set | 0.01s |

The pod carries go-libs OTEL env — `OTEL_TRACES_EXPORTER=otlp` plus the endpoint under the go-libs name `OTEL_TRACES_EXPORTER_OTLP_ENDPOINT`. `ledgerctl` reads only the SDK-standard `OTEL_EXPORTER_OTLP_ENDPOINT` (empty in the pod). With the exporter selected but no standard endpoint resolvable, `cmd/ledgerctl/cmdutil/otel.go` builds an OTLP exporter against the default `localhost:4317`; the deferred `ForceFlush`/`Shutdown` then blocks the full `otelShutdownTimeout = 5s`. `ledgers create` survives via its already-exists retry; `indexes list` has no such escape.

## Fix

`ledgerctlCommand` now prepends a shell prologue to every in-pod exec:
- If a traces endpoint is configured, export it under the standard `OTEL_EXPORTER_OTLP_ENDPOINT`, adding the `http://`/`https://` scheme the SDK requires (from the go-libs insecure flag — a bare `host:port` makes the SDK attempt TLS against a plaintext collector and hang the same 5s).
- Otherwise set `OTEL_SDK_DISABLED=true` so the flush is a no-op.

Applies to all in-pod ledgerctl calls (index reconcile, create, schema, cluster ops, backup) via the single `ledgerctlCommand` helper.

## Verification

- **In-cluster**: the exact command the operator now generates (prologue + `indexes list`) runs in **0.80s** (incl. kubectl overhead), down from ~5s — well under `ledgerExecTimeout`. Prologue resolves `OTEL_EXPORTER_OTLP_ENDPOINT=http://opentelemetry-collector.monitoring.svc.cluster.local.:4317`.
- **Tests**: added `TestOtelExecPrologue` (executes the prologue through `/bin/sh` for no-endpoint→disable, insecure→http, secure→https, pre-schemed→untouched); updated `TestLedgerctlCommand_TLSModes`. `go build`, `go vet`, `gofmt` clean.

## Notes / follow-ups

- **Residual edge**: with a real endpoint, `ledgerctl`'s own OTEL shutdown timeout (5s) still equals `ledgerExecTimeout`; an *unreachable* collector could again approach the deadline. A deeper `ledgerctl`-core fix (don't build an exporter without a resolvable standard endpoint) would remove that fragility — can follow up.
- **Unrelated**: `kid=943a83ce9ac50789` auth failures seen in pod logs are the traefik ingress presenting a stale pre-restart credential; the current keyset is consistent. Not part of this fix.

### DIFF
--- a/misc/operator/internal/controller/ledgerctl_command_test.go
+++ b/misc/operator/internal/controller/ledgerctl_command_test.go
@@ -159,6 +159,16 @@ func TestOtelExecPrologue(t *testing.T) {
 			env:          []string{"OTEL_TRACES_EXPORTER_OTLP_MODE=http"},
 			wantDisabled: "true",
 		},
+		{
+			name:         "traces disabled at Cluster level disables the SDK despite a configured endpoint",
+			env:          []string{"OTEL_TRACES=false", "OTEL_TRACES_EXPORTER_OTLP_ENDPOINT=collector:4317", "OTEL_TRACES_EXPORTER_OTLP_INSECURE=true"},
+			wantDisabled: "true",
+		},
+		{
+			name:         "traces enabled with an endpoint still translates",
+			env:          []string{"OTEL_TRACES=true", "OTEL_TRACES_EXPORTER_OTLP_ENDPOINT=collector:4317", "OTEL_TRACES_EXPORTER_OTLP_INSECURE=true"},
+			wantEndpoint: "http://collector:4317",
+		},
 	}
 
 	for _, tt := range tests {
--- a/misc/operator/internal/controller/ledgerctl_command_test.go
+++ b/misc/operator/internal/controller/ledgerctl_command_test.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"context"
+	"os/exec"
 	"strings"
 	"testing"
 
@@ -75,15 +76,82 @@ func TestLedgerctlCommand_TLSModes(t *testing.T) {
 			shell := cmd[2]
 
 			// Args are single-quoted to neutralize shell metacharacters that
-			// can reach this helper from CRD fields or Secret values.
-			require.True(t, strings.HasPrefix(shell, "./ledgerctl 'store' 'backup'"),
-				"shell command must start with the (quoted) ledgerctl subcommand, got %q", shell)
+			// can reach this helper from CRD fields or Secret values. The
+			// ledgerctl invocation is preceded by the OTEL env prologue.
+			require.Contains(t, shell, "./ledgerctl 'store' 'backup'",
+				"shell command must contain the (quoted) ledgerctl subcommand, got %q", shell)
+			require.True(t, strings.HasPrefix(shell, otelExecPrologue),
+				"shell command must start with the OTEL env prologue, got %q", shell)
 			for _, fragment := range tt.wantContains {
 				require.Contains(t, shell, fragment)
 			}
 			for _, fragment := range tt.wantNotContain {
 				require.NotContains(t, shell, fragment)
 			}
+		})
+	}
+}
+
+func TestOtelExecPrologue(t *testing.T) {
+	t.Parallel()
+
+	// Execute the prologue in a real shell and report the resolved values so
+	// the test exercises the actual POSIX-sh logic the pod runs, not a Go
+	// re-implementation of it.
+	run := func(env []string) (endpoint, sdkDisabled string) {
+		t.Helper()
+		script := otelExecPrologue + `printf '%s\n%s\n' "${OTEL_EXPORTER_OTLP_ENDPOINT:-}" "${OTEL_SDK_DISABLED:-}"`
+		cmd := exec.Command("/bin/sh", "-c", script)
+		// Start from a clean env so an OTEL_* var on the developer's machine
+		// cannot leak into the assertion.
+		cmd.Env = env
+		out, err := cmd.CombinedOutput()
+		require.NoError(t, err, "prologue failed: %s", out)
+		// Two printf lines; do not trim, so an empty value keeps its line.
+		lines := strings.Split(string(out), "\n")
+		require.GreaterOrEqual(t, len(lines), 2, "unexpected output: %q", string(out))
+
+		return lines[0], lines[1]
+	}
+
+	tests := []struct {
+		name         string
+		env          []string
+		wantEndpoint string
+		wantDisabled string
+	}{
+		{
+			name:         "no endpoint disables the SDK",
+			env:          nil,
+			wantEndpoint: "",
+			wantDisabled: "true",
+		},
+		{
+			name:         "insecure endpoint gets http scheme",
+			env:          []string{"OTEL_TRACES_EXPORTER_OTLP_ENDPOINT=collector.monitoring.svc:4317", "OTEL_TRACES_EXPORTER_OTLP_INSECURE=true"},
+			wantEndpoint: "http://collector.monitoring.svc:4317",
+			wantDisabled: "",
+		},
+		{
+			name:         "secure endpoint gets https scheme",
+			env:          []string{"OTEL_TRACES_EXPORTER_OTLP_ENDPOINT=collector.monitoring.svc:4317"},
+			wantEndpoint: "https://collector.monitoring.svc:4317",
+			wantDisabled: "",
+		},
+		{
+			name:         "endpoint already carrying a scheme is left untouched",
+			env:          []string{"OTEL_TRACES_EXPORTER_OTLP_ENDPOINT=http://collector:4317", "OTEL_TRACES_EXPORTER_OTLP_INSECURE=true"},
+			wantEndpoint: "http://collector:4317",
+			wantDisabled: "",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			endpoint, disabled := run(tt.env)
+			require.Equal(t, tt.wantEndpoint, endpoint)
+			require.Equal(t, tt.wantDisabled, disabled)
 		})
 	}
 }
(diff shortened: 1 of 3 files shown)


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1553#issuecomment-4925634720

misc/operator/internal/controller/ledgerctl_command_test.go
```diff
@@ -88,6 +92,70 @@ func TestLedgerctlCommand_TLSModes(t *testing.T) {
...
+	// Execute the prologue in a real shell and report the resolved values so
+	// the test exercises the actual POSIX-sh logic the pod runs, not a Go
+	// re-implementation of it.
+	run := func(env []string) (endpoint, sdkDisabled string) {
+		t.Helper()
+		script := otelExecPrologue + `printf '%s\n%s\n' "${OTEL_EXPORTER_OTLP_ENDPOINT:-}" "${OTEL_SDK_DISABLED:-}"`
+		cmd := exec.Command("/bin/sh", "-c", script)
+		// Start from a clean env so an OTEL_* var on the developer's machine
+		// cannot leak into the assertion.
+		cmd.Env = env
```
NumaryBot: 🟠 **[major] Clear the shell environment in prologue test**

When `env` is nil, `exec.Command` inherits the parent process environment, so the `no endpoint disables the SDK` case is not actually starting from a clean env. If a developer or CI runner has `OTEL_TRACES_EXPORTER_OTLP_ENDPOINT` or related OTEL variables set, they leak into this shell and change the expected output, making the test flaky despite the comment. Use an explicit empty slice as the base environment and append the per-case values.

gfyrag:
Re-reviewed current head `358229e394...`.

Findings:

1. Major: in-pod execs can still spend the full operator timeout in OTEL shutdown when a collector endpoint is configured but unavailable.

The new prologue disables tracing only when the go-libs endpoint is absent. With an endpoint present, `ledgerctl` still flushes with its 5s shutdown timeout, while `ledgerExecTimeout` is also 5s. I reproduced current head with the PR prologue and `127.0.0.1:1`: `ledgerctl version` took `5.27s`. That keeps the same `context deadline exceeded` failure mode for a slow/down collector. For operator-owned execs, either disable export entirely or make the OTEL flush budget strictly smaller than the exec timeout.

2. Major: the prior NumaryBot finding is still open; `TestOtelExecPrologue` is not hermetic.

The test sets `cmd.Env = env`, but the no-endpoint case passes `nil`, which makes `exec.Command` inherit the parent environment. I reproduced the failure with `OTEL_TRACES_EXPORTER_OTLP_ENDPOINT=leaked:4317`; the test got `http://leaked:4317` instead of `""`. Use an explicit empty env slice for that case or normalize nil to `[]string{}` in the helper.

3. Medium: the prologue translates the go-libs endpoint/insecure variables but not the go-libs OTLP mode.

The operator injects `OTEL_TRACES_EXPORTER_OTLP_MODE`, while `ledgerctl` reads the standard `OTEL_EXPORTER_OTLP*_PROTOCOL` variables. A cluster configured for HTTP OTLP can still run in-pod `ledgerctl` with the default gRPC exporter against the HTTP endpoint, landing back in flush failure/timeout behavior. Translate mode as well, or intentionally disable export for these execs.

Checks run:

- `go test ./internal/controller -count=1` from `misc/operator`
- `GOROOT= go test ./cmd/ledgerctl/... -count=1`
- `GOROOT= go build ./...` from repo root
- `GOROOT= go build ./...` from `misc/operator`
- Timed smoke checks for prologue no-endpoint and unreachable-endpoint behavior

sylr:
Thanks @gfyrag — addressed all three in `b5b83bb`.

**1. Unreachable/slow collector still hit the 5s flush == 5s exec timeout.**
Kept the endpoint-translation approach (we want CLI-side traces when a collector is configured) but decoupled the two budgets: `ledgerExecTimeout` is now **20s**, strictly larger than ledgerctl's 5s OTEL flush budget plus a normal RPC round-trip. Your reproduction (`127.0.0.1:1` → 5.27s) now completes well under the ceiling instead of surfacing as `context deadline exceeded`. The prologue's disable-when-no-endpoint branch keeps the common no-collector case instant.

**2. Test not hermetic.**
`run` now executes the shell with an explicit, never-nil environment (`cmd.Env = append([]string{}, env...)`), so a stray `OTEL_*` on the dev/CI host can't leak into the no-endpoint case. Added grpc/http mode cases and a "mode without endpoint still disables" case.

**3. OTLP mode not translated.**
The prologue now also maps go-libs `OTEL_TRACES_EXPORTER_OTLP_MODE` (grpc|http) → standard `OTEL_EXPORTER_OTLP_PROTOCOL`, so an HTTP collector isn't hit with the default gRPC exporter.

`go build` / `go vet` / `gofmt` / `go test ./internal/controller` all clean. Live re-run against the staging pod was deferred (its EKS API was unreachable from my side at push time); the reachable path was verified live earlier (~0.8s end-to-end) and the unreachable path is now bounded by the 20s ceiling.

gfyrag (changes requested):
Re-reviewed current head `b5b83bbc818f4f6f82d352d5c38db694009c33b4`. The previous endpoint/protocol/test issues look addressed, but I still would not approve this head yet.

1. **P2: the scale-down `ledgerctl` execs are still not covered by `ledgerExecTimeout`.**

   `ledgerExecTimeout` is now applied to Ledger reconciliation execs, for example `LedgerReconciler.Reconcile` creates an `execCtx` before `ledgerctlExec` (`misc/operator/internal/controller/ledger_crd_controller.go:138-142`). But the Cluster scale-down path still passes the outer reconcile context straight into `raftScaleDown` (`misc/operator/internal/controller/reconcile_statefulset.go:114`), and `raftScaleDown` passes that same context to both `podExec` calls for `cluster transfer-leader` and `cluster remove-node` (`misc/operator/internal/controller/raft_scaledown.go:88-90`, `:176`).

   Those are still operator-owned in-pod `ledgerctl` invocations, so a blocked SPDY exec or a `ledgerctl` process stuck behind an OTEL flush can still hold the whole Cluster reconcile instead of being bounded by the new per-exec timeout. Please wrap each scale-down `podExec` call in its own `context.WithTimeout(ctx, ledgerExecTimeout)` as well.

2. **P2: `monitoring.traces.enabled: false` is ignored by the in-pod exec prologue when an endpoint is present.**

   The operator injects the CRD setting as `OTEL_TRACES=false` (`misc/operator/internal/controller/envvars.go:382-392`). `ledgerctl`, however, does not read that go-libs flag; it only disables tracing on `OTEL_SDK_DISABLED=true`, and otherwise builds export from standard `OTEL_TRACES_EXPORTER` / `OTEL_EXPORTER_OTLP*` variables (`cmd/ledgerctl/cmdutil/otel.go:65-67`, `:136-167`).

   The new prologue only checks whether `OTEL_TRACES_EXPORTER_OTLP_ENDPOINT` is set (`misc/operator/internal/controller/raft_scaledown.go:260-269`). With traces disabled but a stale/configured endpoint still present in the pod environment, the prologue exports the standard endpoint/protocol and leaves the SDK enabled, so the exec can still pay the 5s flush cost even though the Cluster explicitly disabled traces. Please make the prologue honor `OTEL_TRACES=false` by setting `OTEL_SDK_DISABLED=true` before translating endpoint/protocol.

sylr:
Thanks @gfyrag — both P2s addressed in `4a2919c`.

**1. Scale-down execs now bounded by `ledgerExecTimeout`.**
Added a `podExecWithTimeout` helper (wraps `podExec` in `context.WithTimeout(ctx, ledgerExecTimeout)`) and routed both scale-down invocations through it — `cluster transfer-leader` in `raftScaleDown` and `cluster remove-node` in `removeNode`. So a stuck SPDY exec or an OTEL-flush-blocked `ledgerctl` can no longer hold the whole Cluster reconcile; each in-pod exec is bounded per-call, matching the Ledger reconcile path. Left a directive comment that any new operator-owned in-pod exec must go through this helper.

**2. Prologue now honors `OTEL_TRACES=false`.**
The leading guard is now `if [ "${OTEL_TRACES:-}" = "false" ] || [ -z "${OTEL_TRACES_EXPORTER_OTLP_ENDPOINT:-}" ]; then export OTEL_SDK_DISABLED=true; else …`. With traces disabled at the Cluster level, a stale/configured endpoint no longer keeps the SDK enabled, so the exec pays no flush cost. Added test cases for `OTEL_TRACES=false` (disabled despite endpoint) and `OTEL_TRACES=true` (still translates).

`go build` / `go vet` / `gofmt` / `go test ./internal/controller` all clean. As before, the staging cluster's EKS API was unreachable from my side this round, so the live scale-down path wasn't re-exercised in-pod; the prologue change is covered by unit tests and the timeout change is a straightforward context wrap.

gfyrag (approved):

PR #1554, opened 2026-07-09
fix(backup): rebuild derived state fully on restore (RebuildDelta) and clear the stale cache zone

Azorlogh:
## Problem

Incremental restore (`RebuildDelta` + `FinalizeRestore`) left several pieces of derived state stale or missing, and the integrity checker did not verify most of the projections involved. Anything present only in the export delta (not the checkpoint) was affected:

- **Metadata schema** (`LedgerInfo.MetadataSchema`): field types declared post-checkpoint were dropped — `ReplayLedgerLog` treated the declarations as "no state to track", and `CreateLedger` replay dropped `initial_schema`. Surfaced under the `singleton_driver_model` oracle.
- **LedgerInfo at large**: `CreateLedger` replay rebuilt only name/mode — `Id`, `AccountTypes`, `DefaultEnforcementMode` were lost; post-checkpoint `AddAccountType`/`RemoveAccountType`, `DeleteLedger`, `PromoteLedger`, and `UpdatedDefaultEnforcementMode` logs were not replayed (a post-checkpoint deletion resurrected the ledger). The rebuilt `LedgerInfo` also went only to the Global zone, leaving the `SubAttrLedger` attribute the FSM preload reads stale.
- **Stale FSM cache zone**: the checkpoint's `ZoneCache` rows predate the replayed delta, yet survived the restore. The restored node loaded them and served checkpoint-era values as `CacheHit`s — and `MirrorPreload`'s existing-entry-wins seeding meant even a fresh Pebble reload could not displace them. Reproduced as a clobbered drain (double-spend-class stale read).
- **Boundaries, counters, reference index**: never rebuilt. Writes to a delta-created ledger failed `ErrLedgerNotFound`, ledger stats stayed at checkpoint values, a reference used in the delta was wrongly accepted again, mirror fill-gap advances were lost (the skipped ids live only on the order), and ledger metadata written post-checkpoint was dropped.

## Changes

1. **Schema reconstruction** — `SetMetadataFieldType` / `RemovedMetadataFieldType` route through new `replay.Writer` methods; the backup writer folds them onto an in-memory `LedgerInfo` (seeded from the checkpoint, extended as `CreateLedger` logs replay) and re-saves it.
2. **Full `LedgerInfo` reconstruction** — `ToLedgerInfo()` carries every creation-time field; `AddAccountType` / `RemoveAccountType` / `SetDefaultEnforcementMode` replay through the writer; `DeleteLedger` reproduces the live apply state (tombstone, boundary-row drop, pending-cleanup marker — the data purge stays deferred to a covering purge range, exactly like a live node); `PromoteLedger` ends mirror mode; `saveLedgerInfo` dual-writes the `SubAttrLedger` attribute + the Global zone.
3. **`PrepareForBackup` clears the cache zone** — same rationale as the existing bloom-block drop: the rows predate the replayed delta. The restored node boots with a cold cache and re-seeds from the rebuilt attribute zone on first touch, covering every cached attribute type for both restore entry points (gRPC `FinalizeRestore` and `ledgerctl store bootstrap`).
4. **Boundaries + counters + reference index** — boundaries seed from the checkpoint (or `{1,1}` on a delta `CreateLedger`) and advance per replayed transaction; `applyAuditOrderEffects` decodes the delta's chain-bound `AuditItem.serialized_order` bytes to fold in **mirror fill-gap advances** and **`NumscriptExecutionCount`** (both live on the order, not the log stream); `VolumeCount` / `MetadataCount` / `ReferenceCount` derive from the committed attribute state; `Writer.SetTransactionReference` rebuilds the reference→txID uniqueness index; `SavedLedgerMetadata` / `DeletedLedgerMetadata` replay into the ledger-metadata attribute.
5. **Checker coverage** (folds #1555 into this PR) — `compareSchema`, `compareAccountTypes`, `compareLedgerPresence` (`MISSING_LEDGER` / `UNAUDITED_LEDGER`), with all archived-recovery state seeded from the baseline checkpoint, never the live store. New in this PR: `compareReferences` (both directions + txID identity; deleted / cleanup-pending ledgers skipped) and `compareBoundaries` (replay+audit-derived id fields and counters; net key counters verified against a recount of the stored rows their own passes verify entry-by-entry; `EphemeralEvictedCount`/`TransientUsedCount` excluded — informational, cf. BuildStatus). The audit-order decode is shared between `RebuildDelta` and the checker (`replay.DecodeOrderEffects`).
6. **Prefix-collision fix the new pass caught** — `LedgerKey.Bytes()` is the raw unpadded name while ledger-scoped attribute keys embed the 64-byte padded name, so ledger-prefix scans also matched every ledger whose name extends it. All such scans now use `domain.LedgerScopedPrefix`.

## Verification

- e2e `Restore` suite extended: schema survival, apply-path assertions (post-checkpoint drain, balances, account-type chart), delta-created ledger writes, ledger stats, reference-conflict rejection — 28 specs.
- New e2e `Restore stale cache` suite: deterministic repro of the stale-cache read (fails without change 3).
- e2e `CheckStore` suite (folded + extended): 25 specs against real-server stores — this is what surfaced the prefix-collision bug.
- Unit: rebuild / prepare / check suites incl. fill-gap, numscript-count, ledger-metadata, delete/promote/enforcement replay, tampered-boundary and missing/unaudited-reference detection; pre-commit clean.

## Follow-up (not in this PR)

- `LedgerInfo.DefaultEnforcementMode` and the ledger-metadata attribute (`SubAttrLedgerMetadata`) have no compare pass yet (documented in invariant 8's known-gaps note; both are rebuilt on restore).
- `LedgerBoundaries.EphemeralEvictedCount` / `TransientUsedCount` are carried across restore and excluded from `compareBoundaries` (informational).
- `fix/rebuilddelta-restore-cache` and #1555 are superseded by this PR.

## Notes

- Targets `release/v3.0`.
- `tests/e2e/cluster/restore_test.go` / `restore_stale_cache_test.go` are `//go:build e2e && s3`; the CheckStore suite is `//go:build e2e`.


internal/application/check/checker.go
```diff
@@ -549,6 +565,10 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 
 	c.compareIndexes(snap, expectedIndexes, indexReplayActivity, deletedInReplay, hasArchivedChapters, pendingCleanupLedgers, callback)
 
+	if err := c.compareSchema(ctx, snap, expectedSchemas, callback); err != nil {
```
NumaryBot, line 639: 🟠 **[major] Seed schema expectations across archived chapters**

When `hasArchivedChapters` is true, the replay starts after `archiveEndSeq`, so `expectedSchemas` only contains schema declarations from non-archived logs. Calling `compareSchema` with that partial map will report false `SCHEMA_MISMATCH` events for any live ledger whose schema was declared before the archive boundary (or compare only the post-archive delta), unlike the volume/metadata passes that merge against a baseline.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1554#issuecomment-4927262965

internal/application/check/replay_store.go
```diff
@@ -319,6 +319,17 @@ func (s *replayStore) DeleteTxMetadata(canonicalKey []byte, metaKey string) erro
 	return s.db.Merge(key, buf, pebble.NoSync)
 }
 
+// SetMetadataFieldType / RemoveMetadataFieldType are no-ops here: the schema
+// lives on LedgerInfo, not in this attribute merge store. Verifying the schema
+// projection in the checker (the invariant-8 gap) is left as a follow-up.
+func (s *replayStore) SetMetadataFieldType(string, commonpb.TargetType, string, commonpb.MetadataType) error {
```
NumaryBot, line 348: 🟠 **[major] Checker discards schema mutations during log replay, leaving integrity verification incomplete**

During `Check()` log replay, `SetMetadataFieldType` and `RemovedMetadataFieldType` callbacks flow through `replay_store.go` but the checker implementation silently discards them. As a result, `expectedSchemas` is never populated from replay, so any divergence between the stored `LedgerInfo.MetadataSchema` and the schema reconstructed from logs is never detected — the checker will report success even when the projection is corrupt or incomplete.

**Suggestion:** Implement the schema-mutation callbacks in the checker's replay store to accumulate expected schema state, then compare it against the persisted `LedgerInfo.MetadataSchema` at the end of the pass.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1554#issuecomment-4927262965

internal/application/check/replay_store.go
Azorlogh: This is already left as a follow-up: EN-1497

internal/application/check/checker.go
Azorlogh: Addressed in follow-up PR: https://github.com/formancehq/ledger/pull/1555

internal/application/check/replay_store.go
Azorlogh: duplicate, Addressed in follow-up PR: https://github.com/formancehq/ledger/pull/1555

gfyrag:
Review note for current head `251cbc384519b32fac52b704fcac96d7020622e6`.

I found one restore-side gap that is separate from the existing checker comments:

**`CreateLedger` replay reconstructs only a partial `LedgerInfo`.**

In `internal/infra/backup/rebuild.go:165`, the rebuild path creates a fresh `LedgerInfo` from the `CreatedLedgerLog`, but only copies `Name`, `Id`, `CreatedAt`, `Mode`, and `MetadataSchema`. The creation log also carries fields that are part of the stored ledger projection: `MirrorSource`, `AccountTypes`, and `DefaultEnforcementMode` (`CreatedLedgerLog.ToLedgerInfo()` already copies them, although it currently lacks `Id`).

That means a ledger created after the checkpoint can be restored with different `LedgerInfo` than the audited create event. In particular, initial account types are also not seeded into `rawLedgerTypes` / `ledgerAccountTypes`, so later replay that depends on creation-time account type enforcement or purge simulation can run with the wrong compiled type set.

I would reconstruct the full ledger info from the log, including `Id`, and seed the account-type replay maps from `CreatedLedgerLog.AccountTypes` before continuing with subsequent ledger logs.

A smaller invariant point in the same area: `SetMetadataFieldType` returns nil when the ledger info is missing (`internal/infra/backup/rebuild.go:456`). The comment says a schema op cannot arrive for a ledger that was never created, so if that case is truly impossible for the replay stream, it should return a loud invariant error rather than silently dropping the mutation.

internal/infra/backup/rebuild.go
```diff
@@ -438,6 +452,84 @@ func seedLedgerContext(
...
+		info.MetadataSchema.TransactionFields[key] = field
+	case commonpb.TargetType_TARGET_TYPE_LEDGER:
+		if info.MetadataSchema.LedgerFields == nil {
+			info.MetadataSchema.LedgerFields = make(map[string]*commonpb.MetadataFieldSchema)
+		}
+
+		info.MetadataSchema.LedgerFields[key] = field
+	}
+
+	if err := state.SaveLedger(w.batch, info); err != nil {
```
NumaryBot: 🟠 **[major] Rebuilt metadata schema not persisted to SubAttrLedger projection**

When restoring a schema declaration that occurred after the checkpoint, only the Global `LedgerInfo` row is updated. The FSM preload path and the index builder read ledger state from the `SubAttrLedger` attribute projection (`attrs.Ledger.Get`). After restore, `GetMetadataSchemaStatus` may show the rebuilt fields correctly while the hot path still sees the checkpoint-era schema. The next ledger-schema mutation after restore will clone that stale attribute value and silently overwrite the rebuilt fields. The removal path has the same issue.

**Suggestion:** After updating the Global `LedgerInfo` row, also write the updated `LedgerInfo` (including the rebuilt `MetadataSchema`) to the ledger attribute projection (`SubAttrLedger`), so that both reads and future mutations operate from the same consistent state.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1554#issuecomment-4927262965

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1554#issuecomment-4927262965

Azorlogh:
Thanks — all three addressed, plus a related zone gap they surfaced.

**1. Partial `LedgerInfo` reconstruction.** The `CreateLedger` replay now rebuilds the full info via `CreatedLedgerLog.ToLedgerInfo()` (so `MirrorSource`, `AccountTypes`, and `DefaultEnforcementMode` are carried), and I extended `ToLedgerInfo()` to include `Id` as you noted — its two existing HTTP/CLI callers were returning `Id=0`, so that's a fix for them too. (`eb60c5750`)

**2. Account types not seeded into the replay maps.** Initial `account_types` on the creation log now seed `rawLedgerTypes`/`ledgerAccountTypes` (`eb60c5750`). While fixing this I found the sibling gap: post-checkpoint `AddAccountType`/`RemoveAccountType` updated only the purge-sim maps, never `LedgerInfo`, so a chart change after the checkpoint was lost on restore — those now route through `replay.Writer` callbacks that fold onto `LedgerInfo` and re-persist (`2501866ee`).

**3. Silent `return nil` on missing `LedgerInfo`.** `Set`/`RemoveMetadataFieldType` (and the new account-type callbacks) now return a loud `invariant:` error — every live ledger is seeded from the checkpoint or its `CreateLedger` log, so a missing entry is a corrupt stream, not a runtime condition. `RemoveMetadataFieldType` on a ledger with no schema stays a benign no-op. (`eb60c5750`)

**Related (from the automated review):** rebuild was writing `LedgerInfo` only to the Global `SubGlobLedgerInfo` zone, not the `SubAttrLedger` attribute projection the FSM hot path / index builder / preload read. A `saveLedgerInfo` helper now dual-writes both zones wherever the rebuild persists a ledger, matching the normal FSM apply. (`2501866ee`)

Each fix has unit coverage (including negative checks that they fail without the fix).

gfyrag:
Re-reviewed current head `2501866eebb3b3c2542896bf323220154843eeb0`.

Two issues remain from this pass:

1. High: incremental restore still drops post-checkpoint `LedgerInfo` mutations.

`internal/infra/backup/rebuild.go` still ignores `DeleteLedger` and `PromoteLedger`, and `internal/domain/replay/replay.go` still treats `UpdatedDefaultEnforcementMode` as having no state to track. Those log types normally mutate `LedgerInfo`: delete sets `DeletedAt`, promote changes `Mode` and clears `MirrorSource`, and default enforcement changes `DefaultEnforcementMode`. If those mutations are only present in an incremental export, restore keeps the checkpoint-era ledger state: deleted ledgers can reappear, promoted mirror ledgers can resume as mirrors, and enforcement can revert to its old mode.

2. Medium: the checker still does not verify the rebuilt `LedgerInfo` projections.

The new checker replay hooks for metadata schema/account types remain no-ops, and the account-type comment references `compareAccountTypes`, but I do not see such a checker pass in this head. That leaves metadata schema, account types, default enforcement, mode, mirror source, and deleted state outside invariant-8 projection verification, even though this PR now reconstructs and persists some of them during restore.

What looks fixed since the prior review:

- `CreateLedger` reconstruction now includes the full `LedgerInfo` fields I previously called out.
- Rebuilt `LedgerInfo` is now written to both `SubAttrLedger` and global state.
- Restore-side impossible schema ops now fail loudly for unknown ledgers. The checker-side verification gap above remains.

Checks run:

- `git diff --check release/v3.0...HEAD`
- `go test ./internal/infra/backup -run 'TestRebuildDelta|TestAttributeReplayWriter' -count=1`
- `go test ./internal/application/check -run 'Test.*(AccountType|MetadataSchema|Replay|CheckStore|Checker)' -count=1`

Azorlogh:
Pushed four commits extending this PR's scope:

- **`test(e2e)` apply-path assertions** (`76c42da`) — the restore suite now exercises the *apply path* (not just queries) against post-checkpoint state: eve's delta drain, and the account-type chart.
- **Stale cache zone on restore** (`8e55b64`, `e03477c`) — while verifying the above I found the restored store keeps the checkpoint-era `ZoneCache` rows, while `RebuildDelta` rewrites only the attribute zone. The restored FSM serves those rows as `CacheHit`s, and `MirrorPreload`'s existing-entry-wins seeding prevents even a fresh Pebble reload from displacing them — so an apply against a delta-modified key computes off checkpoint-era values (double-spend-class; deterministic repro in the new `Restore stale cache` e2e suite). Fix: `PrepareForBackup` now clears the cache zone alongside the bloom blocks it already drops for exactly the same staleness reason; the restored node boots with a cold cache and re-seeds from the rebuilt attribute zone.
- **Boundaries, counters and the reference index** (`18688a9`) — ported from `fix/rebuilddelta-restore-cache`, adapted to attribute-zone-only writes (the 0xFF mirror that branch added is unnecessary once the cache zone is cleared, so that branch is now superseded by this PR). Restores writes to delta-created ledgers (`ErrLedgerNotFound` before), ledger stats, and reference idempotency for delta references.

PR description is updated with the full scope. 28 restore-related e2e specs pass.

gfyrag (changes requested):
Re-reviewed current head `18688a970f647a4429a9503ed5c33db81fa859b3`. I would not approve this version yet.

1. **High: `RebuildDelta` does not reconstruct `NextTransactionId` for mirror `FillGap` logs.**

   The live apply path advances the boundary past every skipped v2 transaction id in `processMirrorFillGap` (`internal/domain/processing/processor_mirror.go:118-128`). The restore replay path delegates apply logs to `replay.ReplayLedgerLog(...)` (`internal/infra/backup/rebuild.go:153-167`), but `ReplayLedgerLog` treats `LedgerLogPayload_FillGap` as `// No state to track` (`internal/domain/replay/replay.go:209-210`). Since `FilledGapLog` only stores the original v2 id, the skipped transaction ids are not available to the restore replay.

   That means a backup checkpoint followed by mirrored gap-fills can restore with `Boundary.NextTransactionId` lower than the live cluster's value, and the rebuilt counters/boundaries will not match the state produced by the FSM. The delta replay needs enough data to advance the boundary exactly the same way as live apply, or another deterministic source for the skipped ids.

2. **High: ledger metadata logs are still not replayed.**

   `SaveLedgerMetadata` and `DeleteLedgerMetadata` emit `SavedLedgerMetadata` / `DeletedLedgerMetadata` log payloads (`internal/domain/processing/processor_ledger_metadata.go:11-44`, `:47+`). Incremental backups only bring the cold log/audit stream, and `RebuildDelta` replays each apply log through `ReplayLedgerLog`; however there is no `SavedLedgerMetadata` or `DeletedLedgerMetadata` case in either `internal/infra/backup/rebuild.go` or `internal/domain/replay/replay.go`.

   As a result, ledger metadata written after the checkpoint is dropped during restore, and metadata deleted after the checkpoint can be resurrected from the checkpoint snapshot. This is the same category of issue as the other derived-state rebuild gaps this PR is fixing.

3. **Major: this PR now rebuilds additional persisted projections without checker coverage.**

   `RebuildDelta` now rebuilds `Boundary` counters from the restored attribute state (`internal/infra/backup/rebuild.go:723-733`) and rebuilds transaction references through the restore writer, but the checker side explicitly leaves references unverified (`internal/application/check/replay_store.go:266-270`). The updated invariant says every persisted projection must be verified by `internal/application/check/checker.go`; documenting `Boundary` / `Reference` as known gaps under that invariant does not satisfy it.

   If the intended design is to exempt these projections, the invariant should be changed explicitly. Otherwise this PR should either add checker coverage for the projections it rebuilds, or avoid expanding the persisted projection surface in this change.

Already-covered but still relevant for this head: `DeleteLedger`, `PromoteLedger`, and `UpdatedDefaultEnforcementMode` are still no-ops in restore replay, and the `LedgerInfo` schema/account-type checker gap is being handled in #1555 / EN-1497 rather than here.

Azorlogh:
@gfyrag All three points addressed as of `2d0f2bef4`:

1. **FillGap boundaries** (`fd27ac29c`) — the skipped ids indeed exist nowhere in the log stream, but they are on the order, and `AuditItem.serialized_order` preserves the order chain-bound and exported. `applyAuditOrderEffects` decodes the delta's audit items and folds the fill-gap advances into the boundary accumulation. The same decode rebuilds `NumscriptExecutionCount` (same shape: content source lives on the order, not the `CreatedTransaction` log). Items with `log_sequence == 0` (failed / idempotent orders) contribute nothing, and admission's postings-XOR-script rule makes script detection exact.

2. **Ledger metadata** (`fd27ac29c`) — `SavedLedgerMetadata` / `DeletedLedgerMetadata` now replay into the ledger-metadata attribute. Your footnote items came along: `DeleteLedger` reproduces the live apply state exactly (LedgerInfo tombstone, boundary-row drop, pending-cleanup marker — the data purge stays deferred to a covering purge range, which the restored cluster picks up through `ReadPendingLedgerCleanups` at boot), `PromoteLedger` ends mirror mode, and `UpdatedDefaultEnforcementMode` folds onto the rebuilt LedgerInfo.

3. **Checker coverage** (`2d0f2bef4`) — rather than documenting the gap, #1555 is folded into this PR and extended: `compareReferences` (both directions + txID identity; deleted / cleanup-pending ledgers skipped since their rows legitimately linger until a covering purge) and `compareBoundaries` (`NextTransactionId`/`NextLogId`/`PostingCount`/`RevertCount`/`NumscriptExecutionCount` re-derived from the replayed logs + the same chain-bound audit-order effects, baseline-seeded under archiving; the net key counters are verified against a recount of the stored rows their own passes verify entry-by-entry, so the chain is rows↔audit, counter↔rows). `EphemeralEvictedCount`/`TransientUsedCount` stay excluded as informational (cf. BuildStatus) — documented in invariant 8 rather than silently skipped, along with the two remaining LedgerInfo-level gaps (`DefaultEnforcementMode`, ledger-metadata attribute — both rebuilt on restore, neither compared yet).

Worth noting: `compareBoundaries` caught a real bug on its first run against the e2e stores — ledger-prefix scans used the unpadded `LedgerKey` bytes and counted rows of every ledger whose name extends another's (`check-revert` swallowed `check-revert-then-meta`); the ported `countNetAttributes` had it too. Fixed via `domain.LedgerScopedPrefix` with a regression test.

PR description updated with the full scope; #1555 closed as folded.

gfyrag (approved):
Re-reviewed the current head — my three prior CHANGES_REQUESTED blockers are all genuinely fixed with matching tests and a shared restore↔checker derivation (replay.DecodeOrderEffects):
1. FillGap NextTransactionId now folded from the chain-bound serialized_order in both the restore path and the checker.
2. SavedLedgerMetadata / DeletedLedgerMetadata replayed on restore.
3. Rebuilt-but-unverified projections now covered by compareBoundaries/compareReferences (+ compareSchema/compareAccountTypes/compareLedgerPresence), per invariant #8.

Blast radius is clean (only FSM-path change is an export-visibility rename; all new logic is in the offline restore/checker paths; OpenWriteSession stays within the backup forbidigo exclusion). Build + backup/check/replay tests pass. The two remaining rebuilt-on-restore projections without a compare pass (DefaultEnforcementMode, SubAttrLedgerMetadata) are pre-existing and now explicitly documented as known gaps — reasonable follow-up, not a blocker. LGTM.

internal/application/check/checker.go
```diff
@@ -4609,6 +4640,351 @@ func verifySealingHash(p *commonpb.Chapter, callback func(*servicepb.CheckStoreE
...
+	// its DeleteLedger log sits past the verified range, so the two sides can
+	// legitimately disagree on it. Not tampering, in either direction.
+
+	// audit says live → the store must hold a live LedgerInfo.
+	for name := range knownLedgers {
+		if _, ok := live[name]; ok {
+			continue
+		}
+
+		if _, ok := pendingCleanup[name]; ok {
```
NumaryBot, line 4864: 🟠 **[major] Pending-cleanup row can suppress MISSING_LEDGER for live ledgers**

When the new ledger-presence check finds a ledger in `knownLedgers` (i.e. still considered live) but missing its LedgerInfo projection, the code skips the MISSING_LEDGER event if a pending-cleanup row exists for that ledger. However, pending-cleanup is itself an untrusted store projection and is only legitimate after a DeleteLedger event, which would already remove the ledger from `knownLedgers`. A stale or injected cleanup row for a ledger that is still live can therefore silently suppress the very mismatch the check is intended to detect.

**Suggestion:** Remove the pending-cleanup suppression for ledgers that are still present in `knownLedgers`. A valid DeleteLedger flow removes the ledger from `knownLedgers` before any cleanup row is written, so the two conditions are mutually exclusive for correct state. If the ledger is in `knownLedgers` and its LedgerInfo projection is missing, that is always a MISSING_LEDGER regardless of any cleanup rows.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1554#issuecomment-4927262965

PR #1557, opened 2026-07-10
feat(worker): add AWS IAM flags to worker command for RDS IAM auth

### LINKED ISSUE #1556: RDS IAM database auth for the primary ledger store (not just the mirror source)?
We run Formance Ledger (CC) on AWS ECS Fargate against Aurora PostgreSQL. The ledger currently connects with a static password via `POSTGRES_URI`, sourced from AWS Secrets Manager.

We saw [#1434](https://github.com/formancehq/ledger/pull/1434) add SigV4 token minting - but scoped to the mirror source connection and driven through the Kubernetes operator CRD (`--mirror-aws-iam-region`).

Questions on your plans:
1. Any intent to extend RDS IAM auth to the primary datastore used by serve/worker (e.g. a `--postgres-aws-iam-region-style` option), so the main store can connect without a static password?
2. Rough target release (v3.x) and timeline?
3. Would it be exposed on the binary/CLI directly, usable outside the operator? We deploy the raw image on ECS, not via the operator.
4. Any recommended interim pattern for password-less main-store DB auth today, short of running the operator?

tennyenoch:
## What

The `worker` command was missing `iam.AddFlags()` which provides the
`--aws-region`, `--aws-access-key-id`, `--aws-secret-access-key`,
`--aws-session-token`, `--aws-profile` and `--aws-role-arn` flags.

## Why

Without these flags, users running the standalone worker binary (e.g.
on AWS ECS Fargate without the Kubernetes operator) could not configure
AWS RDS IAM authentication for the primary PostgreSQL store via CLI flags.

The primary store already supports IAM auth via the `--postgres-aws-enable-iam`
flag from `connect.AddFlags`, but the AWS credential/region flags were absent
on the worker command.

The `serve` command already calls both `connect.AddFlags` and `iam.AddFlags`;
this commit brings the `worker` command to parity.

## Testing

Added unit tests in `cmd/worker_test.go` verifying that all AWS IAM flags
(`--postgres-aws-enable-iam`, `--aws-region`, `--aws-access-key-id`,
`--aws-secret-access-key`, `--aws-session-token`, `--aws-role-arn`) are
registered on both the `worker` and `serve` commands (12 tests, all passing).

Closes #1556


gfyrag:
Re-reviewed current head `3fcc7d81c211fe583631c839b422c19ef7e81295`.

Findings:

1. `--aws-role-arn` is exposed on `ledger worker`, but the Postgres/RDS IAM path does not consume it.

`connect.ConnectionOptionsFromFlags` loads AWS config through `iam.LoadOptionFromFlags`, and in the current go-libs version that helper reads region/access key/secret/session/profile, not role ARN. A worker started with `--postgres-aws-enable-iam --aws-role-arn ...` will not assume that role; it will use the default/static/profile credential chain instead. This predates the PR for `serve`, but this PR newly exposes the no-op flag on `worker`.

2. The new regression tests do not assert `--aws-profile`.

The PR description says `--aws-profile` is part of the worker IAM flag parity fix, and `iam.AddFlags` registers it, but the new tests do not assert it for either worker or serve. Dropping that flag would still pass the new tests.

3. `git diff --check` fails on an extra blank line at EOF in `cmd/worker_test.go`.

Checks run:

- `go test ./cmd -run 'Test(Worker|Serve)CommandHasAWSIAMFlags' -count=1 -v`
- `go test ./cmd -count=1`
- `GOROOT= go build ./...`
- `go test ./... -run '^$'`
- `go test` on the go-libs IAM/connect packages
- `git diff --check 0695abc...HEAD` failed only on the blank EOF line above.

tennyenoch:
@gfyrag - Fixed the review comments. Please review.

NumaryBot (commented):

The PR wires AWS IAM flags onto the worker Cobra command to mirror the existing serve command pattern, and adds tests verifying flag registration and defaults. The overall approach is sound and consistent with the codebase. However, one major issue was identified: the `--aws-role-arn` flag is registered and exposed to operators but is never consumed by the underlying `iam.LoadOptionFromFlags` helper, so passing it has no effect — this is a silent no-op that can mislead users configuring RDS IAM role assumption. Two lesser issues were also noted: the new tests do not cover `--aws-profile` registration (meaning that flag could be silently removed without breaking tests), and there is a trailing blank line in the test file that fails whitespace checks.

### Findings outside the diff

🟠 **[major] `--aws-role-arn` flag is registered but never consumed for RDS IAM auth** — `cmd/worker.go`

`connect.ConnectionOptionsFromFlags` uses `iam.LoadOptionFromFlags` to build AWS credentials, and the current go-libs version of that helper reads region, access key, secret, session token, and profile — but not role ARN. A worker started with `--postgres-aws-enable-iam --aws-role-arn <arn>` will silently ignore the ARN and fall back to the default credential chain instead of assuming the intended role. The flag is therefore a no-op that misleads operators.

**Suggestion:** Either (a) update `iam.LoadOptionFromFlags` (and the corresponding go-libs version) to support role assumption via the ARN, or (b) remove the `--aws-role-arn` flag from the worker command until the underlying implementation can honour it, and document the limitation.



🟡 **[minor] New tests do not assert `--aws-profile` flag registration** — `cmd/worker_test.go`

`iam.AddFlags` registers `--aws-profile` as part of the IAM flag set, and the PR description states that full IAM flag parity with `serve` is the goal. However, the new regression tests do not assert the presence of `--aws-profile` for either the worker or serve commands, meaning the flag could be silently dropped without causing test failures.

**Suggestion:** Add an assertion for `--aws-profile` (and any other flags registered by `iam.AddFlags`) alongside the existing flag checks in the worker and serve test cases.



⚪ **[nit] Extra blank line at EOF causes `git diff --check` failure** — `cmd/worker_test.go`

There is a trailing blank line at the end of `cmd/worker_test.go` that causes `git diff --check` to report a whitespace error.

**Suggestion:** Remove the extra blank line at the end of the file.



Summary: https://github.com/formancehq/ledger/pull/1557#issuecomment-4932067371

cmd/aws.go
```diff
@@ -0,0 +1,142 @@
...
+	cfg, err := awsconfig.LoadDefaultConfig(ctx, iam.LoadOptionFromFlags(flags))
+	if err != nil {
+		return nil, fmt.Errorf("aws: loading base aws config: %w", err)
+	}
+
+	// Wrap the base credentials with an STS AssumeRole provider.
+	// stscreds.AssumeRoleProvider caches the temporary credentials and refreshes
+	// them automatically before they expire, so callers do not need to manage
+	// the STS session lifetime explicitly.
+	stsClient := sts.NewFromConfig(cfg)
```
NumaryBot: 🟠 **[major] AssumeRoleProvider not cached — STS called on every DB connection**

When both `--postgres-aws-enable-iam` and `--aws-role-arn` are set, the raw `stscreds.AssumeRoleProvider` is stored directly as the credentials provider. This provider does not cache credentials internally, so every new physical PostgreSQL connection triggers an STS `AssumeRole` call before building the RDS IAM auth token. Under connection churn or startup bursts this adds latency and can hit STS throttling limits.

**Suggestion:** Wrap the provider with `aws.NewCredentialsCache` before storing it, e.g.:
```go
creds := aws.NewCredentialsCache(stscreds.NewAssumeRoleProvider(stsClient, roleARN))
```
This ensures credentials are reused until they expire and STS is only called to refresh them.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1557#issuecomment-4932067371

cmd/aws.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil, fmt.Errorf("aws: loading base aws config: %w", err)
+	}
+
+	// Wrap the base credentials with an STS AssumeRole provider, then cache
+	// the resulting temporary credentials with aws.NewCredentialsCache.
+	// Without the cache, stscreds.AssumeRoleProvider would call STS on every
+	// new physical DB connection, adding latency and risk of STS throttling
+	// under connection churn. The cache reuses credentials until they are
+	// about to expire and only then calls STS to refresh them.
+	stsClient := sts.NewFromConfig(cfg)
```
NumaryBot: 🟠 **[major] AssumeRoleProvider not wrapped with credentials cache — STS called on every DB connection**
<sub>reported by NumaryBot, coderabbitai[bot]</sub>

When both `--postgres-aws-enable-iam` and `--aws-role-arn` are set, the raw `stscreds.AssumeRoleProvider` is stored directly as the credentials provider. This provider does not cache credentials internally, so every new physical PostgreSQL connection triggers an STS `AssumeRole` API call before building the RDS IAM auth token. Under connection churn or startup bursts this adds latency and risks hitting STS throttling limits. The provider is also not safe to share across goroutines without synchronization.

**Suggestion:** Wrap the provider with `aws.NewCredentialsCache` before storing it:
```go
creds := aws.NewCredentialsCache(stscreds.NewAssumeRoleProvider(stsClient, roleARN))
```
This ensures credentials are reused until they expire and STS is only called when a refresh is actually needed.

cmd/aws.go
```diff
@@ -0,0 +1,145 @@
+package cmd
+
+import (
+	"context"
+	"database/sql/driver"
+	"fmt"
+	"net/url"
+
```
NumaryBot: 🟠 **[major] New AWS SDK imports still listed as indirect in go.mod — will break CI Dirty check**

The newly introduced direct AWS SDK imports (e.g., `stscreds`, `sts`) are only listed as `// indirect` dependencies in `go.mod`. The CI 'Dirty' job runs `go mod tidy`, which promotes them to direct requirements, causing a diff and failing the job.

**Suggestion:** Run `go mod tidy` locally and commit the updated `go.mod` (and `go.sum` if changed) so that the new direct imports are recorded correctly before merging.

cmd/aws.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/aws/aws-sdk-go-v2/aws"
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
+	"github.com/aws/aws-sdk-go-v2/feature/rds/auth"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"github.com/formancehq/go-libs/v5/pkg/cloud/aws/iam"
```
NumaryBot: 🟠 **[major] Import block ordering not compliant with gci config — will break CI Dirty check**

The import block in the new `cmd/aws.go` does not follow the repository's configured `gci` import ordering (first-party `github.com/formancehq/...` imports must be separated from other third-party imports). The CI formatter will reorder the imports, leaving the tree dirty and failing the Dirty job.

**Suggestion:** Run the project's formatter (e.g., `just pre-commit` or `gci write cmd/aws.go`) locally and commit the result so the import order matches the configured style.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1557#issuecomment-4932067371

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1557#issuecomment-4932067371

cmd/aws.go
```diff
@@ -0,0 +1,146 @@
...
+		return nil, fmt.Errorf("aws: loading base aws config: %w", err)
+	}
+
+	// Wrap the base credentials with an STS AssumeRole provider, then cache
+	// the resulting temporary credentials with aws.NewCredentialsCache.
+	// Without the cache, stscreds.AssumeRoleProvider would call STS on every
+	// new physical DB connection, adding latency and risk of STS throttling
+	// under connection churn. The cache reuses credentials until they are
+	// about to expire and only then calls STS to refresh them.
+	stsClient := sts.NewFromConfig(cfg)
```
NumaryBot: 🟠 **[major] AssumeRoleProvider not wrapped with credentials cache — STS called on every DB connection**
<sub>reported by NumaryBot, coderabbitai[bot]</sub>

When both `--postgres-aws-enable-iam` and `--aws-role-arn` are set, the raw `stscreds.AssumeRoleProvider` is stored directly as the credentials provider. This provider does not cache credentials internally, so every new physical PostgreSQL connection triggers an STS `AssumeRole` API call before building the RDS IAM auth token. Under connection churn or startup bursts this adds latency and risks STS throttling. The provider is also not safe to share across goroutines without synchronization.

**Suggestion:** Wrap the provider with `aws.NewCredentialsCache` before storing it:
```go
creds := aws.NewCredentialsCache(stscreds.NewAssumeRoleProvider(stsClient, roleARN))
```
This ensures credentials are reused until expiry and STS is only called when a refresh is actually needed.

cmd/aws.go
```diff
@@ -0,0 +1,146 @@
+package cmd
+
+import (
+	"context"
+	"database/sql/driver"
+	"fmt"
+	"net/url"
+
```
NumaryBot: 🟠 **[major] New AWS SDK imports still listed as indirect in go.mod — will break CI Dirty check**

The newly introduced direct AWS SDK imports (e.g., `stscreds`, `sts`) are only recorded as `// indirect` dependencies in `go.mod`. The CI 'Dirty' job runs `go mod tidy`, which promotes them to direct requirements, producing a diff and failing the job.

**Suggestion:** Run `go mod tidy` locally and commit the updated `go.mod` (and `go.sum` if changed) so that the new direct imports are recorded correctly before merging.

cmd/aws.go
```diff
@@ -0,0 +1,146 @@
...
+	authedURL := u.URL
+	authedURL.User = url.UserPassword(u.URL.User.Username(), authToken)
+	authedDSN := authedURL.String()
+
+	pgxCfg, err := pgx.ParseConfig(authedDSN)
+	if err != nil {
+		return nil, fmt.Errorf("aws: parsing dsn with iam auth token: %w", err)
+	}
+
+	return stdlib.GetConnector(*pgxCfg).Connect(ctx)
```
NumaryBot: 🟠 **[major] Assumed-role connector path bypasses go-libs pgx connector safeguards**

When `--postgres-aws-enable-iam --aws-role-arn` is used, the direct `stdlib.GetConnector` path skips the go-libs connector setup that all other Postgres auth paths use, including the read-write `ValidateConnect`, read-only session reset, and tracing hooks. Against an HA endpoint or a connection returned as read-only, the pool can retain connections that other auth modes would reject, causing later write failures.

**Suggestion:** Apply the same connector setup (ValidateConnect, read-only session reset, tracing hooks) that the existing connect path uses when constructing the assumed-role connector, rather than calling `stdlib.GetConnector` directly.

cmd/aws.go
```diff
@@ -0,0 +1,146 @@
...
+	connMaxLifetime, _ := flags.GetDuration(connect.PostgresConnMaxLifetimeFlag)
+	maxOpenConns, _ := flags.GetInt(connect.PostgresMaxOpenConnsFlag)
+
+	return &connect.ConnectionOptions{
+		DatabaseSourceName: postgresURI,
+		MaxIdleConns:       maxIdleConns,
+		ConnMaxIdleTime:    connMaxIdleTime,
+		ConnMaxLifetime:    connMaxLifetime,
+		MaxOpenConns:       maxOpenConns,
+		Connector: func(dsn string) (driver.Connector, error) {
```
NumaryBot: 🟠 **[major] Assumed-role connector ignores DSN argument — breaks schema overrides**

The connector closure always returns the prebuilt connector whose DSN was captured from the original `--postgres-uri`, silently ignoring the `dsn` argument passed in at call time. `OpenDBWithSchema` mutates `ConnectionOptions.DatabaseSourceName` to inject `search_path` before calling `options.Connector(options.DatabaseSourceName)`. For the assumed-role path this override is dropped, so tables are opened/migrated in the default schema instead of the configured one.

**Suggestion:** Rebuild the connector inside the closure using the `dsn` argument so that schema overrides (and any other DSN mutations applied by callers) are honoured, e.g.:
```go
return func(dsn string) (driver.Connector, error) {
    cfg, err := pgx.ParseConfig(dsn)
    // ... apply token refresh and return connector built from cfg
}
```

cmd/aws.go
```diff
@@ -0,0 +1,146 @@
...
+			return connector, nil
+		},
+	}, nil
+}
+
```
NumaryBot: ⚪ **[nit] Trailing blank line at EOF causes whitespace/formatter check failure**

There is an extra blank line at EOF in `cmd/aws.go`. The CI Dirty workflow runs formatting/pre-commit and then fails on any resulting git diff, so this whitespace issue will keep CI red.

**Suggestion:** Remove the trailing blank line at the end of `cmd/aws.go`.

cmd/aws.go
```diff
@@ -0,0 +1,242 @@
...
+		return nil, fmt.Errorf("aws: loading base aws config: %w", err)
+	}
+
+	// Wrap the base credentials with an STS AssumeRole provider, then cache
+	// the resulting temporary credentials with aws.NewCredentialsCache.
+	// Without the cache, stscreds.AssumeRoleProvider would call STS on every
+	// new physical DB connection, adding latency and the risk of throttling
+	// under connection churn. The cache reuses credentials until they are
+	// about to expire and only then calls STS to refresh them.
+	stsClient := sts.NewFromConfig(cfg)
```
NumaryBot: 🟠 **[major] AssumeRoleProvider not wrapped with credentials cache — STS called on every DB connection**
<sub>reported by NumaryBot, coderabbitai[bot]</sub>

The raw `stscreds.AssumeRoleProvider` is stored directly as the credentials provider. This provider does not cache credentials internally, so every new physical PostgreSQL connection triggers a live STS `AssumeRole` API call before building the RDS IAM auth token. Under connection churn or startup bursts this adds significant latency and risks hitting STS throttling limits. The provider is also not safe to share across goroutines without synchronization.

**Suggestion:** Wrap the provider with `aws.NewCredentialsCache` before storing it:
```go
creds := aws.NewCredentialsCache(stscreds.NewAssumeRoleProvider(stsClient, roleARN))
```
This ensures credentials are reused until expiry and STS is only called when a refresh is actually needed.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1557#issuecomment-4932067371

cmd/aws.go
```diff
@@ -0,0 +1,242 @@
+package cmd
+
+import (
+	"context"
+	"database/sql/driver"
+	"errors"
+	"fmt"
+	"net/url"
+	"time"
+
```
NumaryBot: 🟠 **[major] New AWS SDK imports still listed as indirect in go.mod — will break CI Dirty check**

The newly introduced direct AWS SDK imports (e.g., `stscreds`, `sts`) are only recorded as `// indirect` dependencies in `go.mod`. The CI 'Dirty' job runs `go mod tidy`, which promotes them to direct requirements, producing a diff and failing the job.

**Suggestion:** Run `go mod tidy` locally and commit the updated `go.mod` (and `go.sum` if changed) so that the new direct imports are recorded correctly before merging.

cmd/aws.go
```diff
@@ -0,0 +1,242 @@
...
+	span := trace.SpanFromContext(ctx)
+	defer span.End()
+	if data.Err != nil {
+		span.RecordError(data.Err)
+	}
+}
+
+var (
+	_ pgx.ConnectTracer = roleIAMPgxTracer{}
+	_ pgx.QueryTracer   = roleIAMPgxTracer{}
```
NumaryBot: 🟠 **[major] Assumed-role connector path bypasses go-libs pgx connector safeguards**

When `--postgres-aws-enable-iam --aws-role-arn` is used, the direct `stdlib.GetConnector` path skips the go-libs connector setup that all other Postgres auth paths use, including the read-write `ValidateConnect`, read-only session reset, and tracing hooks. Against an HA endpoint or a connection returned as read-only, the pool can retain connections that other auth modes would reject, causing later write failures.

**Suggestion:** Apply the same connector setup (ValidateConnect, read-only session reset, tracing hooks) that the existing connect path uses when constructing the assumed-role connector, rather than calling `stdlib.GetConnector` directly.

cmd/aws.go
```diff
@@ -0,0 +1,242 @@
...
+	return &connect.ConnectionOptions{
+		DatabaseSourceName: postgresURI,
+		MaxIdleConns:       maxIdleConns,
+		ConnMaxIdleTime:    connMaxIdleTime,
+		ConnMaxLifetime:    connMaxLifetime,
+		MaxOpenConns:       maxOpenConns,
+		// Use the dsn argument (not the captured postgresURI) so that
+		// OpenDBWithSchema's search_path injection and any other DSN
+		// mutation applied by callers is honoured correctly.
+		Connector: func(dsn string) (driver.Connector, error) {
```
NumaryBot: 🟠 **[major] Assumed-role connector ignores DSN argument — breaks schema overrides**

The connector closure always returns the prebuilt connector whose DSN was captured from the original `--postgres-uri`, silently ignoring the `dsn` argument passed in at call time. `OpenDBWithSchema` mutates `ConnectionOptions.DatabaseSourceName` to inject `search_path` before calling `options.Connector(options.DatabaseSourceName)`. For the assumed-role path this override is dropped, so tables are opened/migrated in the default schema instead of the configured one.

**Suggestion:** Rebuild the connector inside the closure using the `dsn` argument so that schema overrides (and any other DSN mutations applied by callers) are honoured:
```go
return func(dsn string) (driver.Connector, error) {
    cfg, err := pgx.ParseConfig(dsn)
    // ... apply token refresh and return connector built from cfg
}
```

cmd/aws.go
```diff
@@ -0,0 +1,242 @@
...
+		c.creds,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("aws: building rds iam auth token for assumed role: %w", err)
+	}
+
+	// Inject the auth token as the PostgreSQL password.
+	authedURL := u.URL
+	authedURL.User = url.UserPassword(u.URL.User.Username(), authToken)
+	authedDSN := authedURL.String()
```
NumaryBot: ⚪ **[nit] Trailing blank line at EOF causes whitespace/formatter check failure**

There is an extra blank line at the end of `cmd/aws.go`. The CI Dirty workflow runs formatting/pre-commit checks and then fails on any resulting git diff, so this whitespace issue will keep CI red.

**Suggestion:** Remove the trailing blank line at the end of `cmd/aws.go`.

tennyenoch:
@coderabbitai review

gfyrag (changes requested):
Thanks for this! The 2-line worker fix for #1556 is spot on. My main concern is about **where the AssumeRole logic should live**, plus a couple of concrete bugs in the reimplemented connector.

## Architecture — the AssumeRole machinery belongs in go-libs, not in `cmd/`

`cmd/aws.go` copy-pastes go-libs internals: `roleIAMPgxTracer` ≈ `connect.pgxTracer`, `validateReadWrite` ≈ `validateConnectTargetSessionAttrsReadWrite`, `resetReadOnly` ≈ `resetReadOnlySession`, and `roleIAMConnector` ≈ `connect.iamConnector`. These are near-verbatim duplicates of unexported functions. They **will drift** from go-libs over time — and they already have (see the missing deadline below).

The role-ARN support should be added **on the go-libs side** (`iam.LoadOptionFromFlags` / `connect.GetAWSIAMAuthConnector`), where the connector it mirrors actually lives. That way every consumer benefits and there's a single implementation to maintain.

We're fine with a single ledger PR that **bumps go-libs + keeps the 2-line worker fix**, but the connector/AssumeRole logic itself must move to go-libs in all cases — it shouldn't be reimplemented here.

## Scope vs. description

The PR is described as "add AWS IAM flags to the worker command", but it actually introduces a full STS AssumeRole DB connector and swaps `connect.ConnectionOptionsFromFlags` for a local wrapper in **both** `serve` and `worker`. Worth making that explicit (title/changelog), because it changes auth behavior — see next point.

## Concrete issues

**1. Silent behavior change for existing `serve` deployments.** `serve` already registers `--aws-role-arn` today (via `iam.AddFlags`), but go-libs ignores it — `iam.LoadOptionFromFlags` never reads `AWSRoleArnFlag`. After this PR, that same flag now triggers a real STS AssumeRole. Any existing `serve` operator who has `--aws-role-arn` set (it was a no-op until now) will see DB auth switch from caller-identity to assumed-role, breaking connections if the role isn't assumable or lacks `rds-db:connect`. Needs to be called out in the changelog at minimum.

**2. `resetReadOnly` has no deadline (robustness bug).** go-libs wraps its `show transaction_read_only` probes in `withPgConnDeadline` (5s). The copied `resetReadOnly` has no deadline, yet it runs on every connection reuse via `OptionResetSession`. On a half-open connection this can block indefinitely and stall the goroutine returning the conn to the pool. Note `validateReadWrite` did keep the deadline — so the two paths already diverged. (This is exactly the drift argument for moving it to go-libs.)

**3. `ValidateConnect` is set unconditionally.** go-libs only sets it when nil (`if config.ValidateConnect == nil`), preserving a `target_session_attrs` coming from the DSN. The copy overwrites it. Low impact, but another behavioral divergence.

## Verified OK

No duplicate flag registration (no pflag panic), all 6 `ConnectionOptions` fields repopulated, RDS token regenerated per `Connect`, `CredentialsCache` avoids per-connection STS calls, `search_path` DSN correctly propagated, `go.mod` direct/indirect changes consistent.

## Suggestion

Either (a) split: land just the 2-line worker flag fix to close #1556, and move AssumeRole to a go-libs PR; or (b) single PR that bumps go-libs (with AssumeRole implemented there) + the 2-line fix. In both cases the connector logic lives in go-libs, and points 1–3 get addressed there.

cmd/worker.go
```diff
@@ -132,6 +133,7 @@ func NewWorkerCommand() *cobra.Command {
 	connect.AddFlags(cmd.Flags())
 	metrics.AddFlags(cmd.Flags())
 	traces.AddFlags(cmd.Flags())
+	iam.AddFlags(cmd.Flags())
```
NumaryBot, line 153: 🟠 **[major] --aws-role-arn flag is exposed but silently ignored in the Postgres IAM connection path**

The worker command registers --aws-role-arn (and --postgres-aws-enable-iam), but connect.ConnectionOptionsFromFlags loads AWS config via iam.LoadOptionFromFlags, which reads region/static keys/session/profile but not the role ARN. A worker started with --postgres-aws-enable-iam --aws-role-arn will not assume the requested role; it falls back to the default/static credential chain. The flag is effectively a no-op, which will silently mislead operators.

**Suggestion:** Either make the connection path consume AWSRoleArnFlag (e.g. by wrapping the loaded credentials with stscreds.NewAssumeRoleProvider + aws.NewCredentialsCache when the flag is non-empty), or remove the flag from the worker command until the underlying library supports it, and document the limitation clearly.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1557#issuecomment-4932067371

cmd/worker.go
```diff
@@ -132,6 +133,17 @@ func NewWorkerCommand() *cobra.Command {
...
 	metrics.AddFlags(cmd.Flags())
 	traces.AddFlags(cmd.Flags())
+	// iam.AddFlags registers --aws-region, --aws-access-key-id,
+	// --aws-secret-access-key, --aws-session-token, --aws-profile, and
+	// --aws-role-arn. All flags except --aws-role-arn are consumed by
+	// iam.LoadOptionFromFlags inside connect.ConnectionOptionsFromFlags.
+	// --aws-role-arn is not yet read by iam.LoadOptionFromFlags in go-libs;
+	// passing it has no effect on the credential chain. Full role-assumption
+	// support is tracked for a future go-libs update.
+	iam.AddFlags(cmd.Flags())
```
NumaryBot, line 153: 🟠 **[major] --aws-role-arn flag is exposed but silently ignored in the Postgres IAM connection path**
<sub>reported by codex, gfyrag, NumaryBot</sub>

The worker command registers --aws-role-arn (and --postgres-aws-enable-iam), but connect.ConnectionOptionsFromFlags loads AWS config via iam.LoadOptionFromFlags, which reads region/static keys/session/profile but not the role ARN. A worker started with --postgres-aws-enable-iam --aws-role-arn will not assume the requested role; it falls back to the default/static credential chain. Additionally, MarkHidden (if used) only removes the flag from help output — Cobra still parses and accepts it — so the no-op behaviour is invisible to operators and will silently mislead them into believing role assumption is active when it is not.

**Suggestion:** Either (a) make the connection path consume AWSRoleArnFlag by wrapping the loaded credentials with stscreds.NewAssumeRoleProvider + aws.NewCredentialsCache when the flag is non-empty, or (b) do not register the flag on the worker command at all until the underlying library supports it, or (c) add an explicit pre-run validation that returns an error when --aws-role-arn is set alongside --postgres-aws-enable-iam, with a clear message explaining that role assumption is not yet implemented for the worker path.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1557#issuecomment-4932067371

cmd/worker.go
```diff
@@ -132,6 +150,10 @@ func NewWorkerCommand() *cobra.Command {
 	connect.AddFlags(cmd.Flags())
 	metrics.AddFlags(cmd.Flags())
 	traces.AddFlags(cmd.Flags())
+	iam.AddFlags(cmd.Flags())
```
NumaryBot, line 153: 🟠 **[major] --aws-role-arn flag is registered but silently ignored in the Postgres IAM connection path**
<sub>reported by codex, gfyrag, NumaryBot</sub>

The worker command registers --aws-role-arn (alongside --postgres-aws-enable-iam), but connect.ConnectionOptionsFromFlags loads AWS config via iam.LoadOptionFromFlags, which reads region/static keys/session/profile but not role ARN. A worker started with --postgres-aws-enable-iam --aws-role-arn will not assume the requested role; it silently falls back to the default/static credential chain. MarkHidden (if used) only hides the flag from help output — Cobra still parses and accepts it — so the no-op behaviour is invisible to operators.

**Suggestion:** Either (a) make the connection path consume AWSRoleArnFlag by wrapping the loaded credentials with stscreds.NewAssumeRoleProvider + aws.NewCredentialsCache when the flag is non-empty, (b) do not register --aws-role-arn on the worker command at all until the underlying library supports it and document the limitation clearly, or (c) add an explicit pre-run validation that returns a clear error when --aws-role-arn is set alongside --postgres-aws-enable-iam so operators are not silently misled.

cmd/worker.go
```diff
@@ -93,6 +94,23 @@ func NewWorkerCommand() *cobra.Command {
 		Use:          "worker",
 		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, _ []string) error {
+			// Fail fast if the operator combines --postgres-aws-enable-iam with
+			// --aws-role-arn. iam.LoadOptionFromFlags does not yet consume
+			// AWSRoleArnFlag, so role assumption would silently not occur and
+			// the worker would fall back to the default credential chain instead
+			// of the intended role. Return an explicit error rather than
+			// misleading the operator.
+			if iamEnabled, _ := cmd.Flags().GetBool(connect.PostgresAWSEnableIAMFlag); iamEnabled {
```
NumaryBot, line 103: 🟠 **[major] Registering --aws-role-arn breaks IRSA/web-identity environments via env-var binding**

service.Execute populates every registered flag from its uppercase environment variable equivalent, so adding iam.AddFlags (which registers --aws-role-arn) causes AWS_ROLE_ARN to be bound to that flag. In IRSA or web-identity environments where AWS_ROLE_ARN is legitimately set for the AWS SDK's standard credential chain alongside POSTGRES_AWS_ENABLE_IAM=true, any validation check that rejects or errors on --aws-role-arn will abort startup even though the operator never intended to pass this flag explicitly.

**Suggestion:** Avoid registering --aws-role-arn on the worker command unless the connection path actually consumes it. If the flag must be registered, ensure that the presence of AWS_ROLE_ARN via environment variable does not cause an error or unexpected behaviour for operators using standard SDK credential flows.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1557#issuecomment-4932067371

PR #1559, opened 2026-07-10
fix(EN-1502): gate filtered audit reads on audit-index progress

gfyrag:
## What & why

Fixes [EN-1502](https://formance-team.atlassian.net/browse/EN-1502) (follow-up to the EN-1241 audit-list contract, PR #1548).

`ListAuditEntries` filtered live reads resolve through the **audit** secondary index (`internal/application/auditindexer`, cursor `ReadAuditProgress`), but `options.read.min_log_sequence` only waited on the **log** index (`readStore.WaitForSequence` / `LastIndexedSequence`). A filtered audit read issued right after a write could therefore miss the just-written entry until the async audit indexer caught up — with no knob to force consistency, diverging from `ListTransactions` / `ListAccounts` where `min_log_sequence` gates the very index those reads consult.

## Approach (conservative, no proto/API change)

`min_log_sequence` is a **log** sequence; `ReadAuditProgress` is an **audit** sequence. They diverge (audit advances on every proposal, including failures that emit no log), so comparing `audit_progress >= min_log_sequence` is **not** a correct read-your-writes guarantee. Instead, for a live *filtered* read:

1. wait for the log index to reach `min_log_sequence` (existing `waitMinLogSequence`);
2. read the current live audit head from the main store (`query.ReadLastAuditSequence` on a direct read handle);
3. wait for the audit-index cursor to reach that observed head (`WaitForAuditSequence`);
4. only then compile + read the audit filter.

This can slightly over-wait, which is acceptable for v3.0 and avoids inventing a `min_audit_sequence` API.

## Changes

- **`internal/storage/readstore`**: `LastIndexedAuditSequence()` + `WaitForAuditSequence(ctx, minSeq)`, reusing the same `progressMu`/`progressCond` mechanism, with the correct `WaitForCheckpoint`-style missed-wakeup / cancellation locking (cancellation broadcast holds `progressMu`).
- **`internal/application/auditindexer`**: call `readStore.NotifyProgress()` after each committed audit-index batch so audit waiters wake immediately (lifecycle path, not FSM/hot path).
- **`internal/adapter/grpc` (`BucketServiceServerImpl`)**: `waitFilteredAuditConsistency`; wired into the live `ListAuditEntries` branch for `opts.GetFilter() != nil`. Unfiltered reads keep the direct Cold/Audit fast path unchanged. Errors from the audit-head read / waits bubble as gRPC codes (Unavailable/Internal) or the context error — never swallowed (CLAUDE.md invariant #7).

**Out of scope (documented):** checkpoint-scoped filtered audit reads (frozen snapshots) belong at the checkpoint-creation boundary — the existing caveat comment in `ListAuditEntries` points there.

## Guardrails

- No proto/API change (keeps `min_log_sequence` semantics).
- No FSM changes; no new Pebble reads on the hot path.
- No mockgen interface changed (methods added on concrete `*readstore.Store` / `*auditindexer.Indexer`).

## Tests

- `readstore.WaitForAuditSequence`: fast path (equal/ahead/zero), block-until-progress + wakeup after `NotifyProgress`, context cancellation returns promptly; `LastIndexedAuditSequence` independent of the log cursor.
- `auditindexer`: `ProcessOnce` wakes audit waiters after committing the cursor.
- gRPC: live filtered read waits until audit progress catches up to the live head; **regression** — audit head below `min_log_sequence` (intervening failed proposal, no log) does NOT block; unfiltered read does not wait on audit progress; context cancel while waiting returns promptly.

Table/parallel tests, `require.Eventually`-style channel waits (no `time.Sleep`), no hand-rolled mocks. Race detector clean.

## Stacking

Stacked on #1548 (`feat/en-1241-audit-list-contract`) since it depends on that PR's audit-list code. Will auto-retarget to `release/v3.0` when #1548 merges.

[EN-1502]: https://formance-team.atlassian.net/browse/EN-1502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/adapter/grpc/server_bucket.go
+++ b/internal/adapter/grpc/server_bucket.go
@@ -569,6 +569,55 @@ func (impl *BucketServiceServerImpl) waitMinLogSequence(ctx context.Context, min
 	return impl.readStore.WaitForSequence(ctx, minLogSequence)
 }
 
+// waitFilteredAuditConsistency gives a live, filtered ListAuditEntries read a
+// read-your-writes guarantee against the async audit secondary index.
+//
+// min_log_sequence is a LOG sequence, but a filtered audit read resolves through
+// the audit index whose cursor (ReadAuditProgress) is an AUDIT sequence. The two
+// spaces diverge — the audit sequence advances on every proposal, including
+// failures that emit no log — so waiting for audit_progress >= min_log_sequence
+// would be incorrect. Instead, conservatively:
+//
+//  1. wait for the log index to reach min_log_sequence (waitMinLogSequence);
+//  2. read the current live audit head from the main store;
+//  3. wait for the audit index cursor to reach that observed head.
+//
+// This can over-wait (the head observed in step 2 may include audit entries
+// produced after min_log_sequence), which is acceptable for v3.0 and avoids a
+// new min_audit_sequence API. A minLogSeq of 0 means "no consistency bound
+// requested", so this is a no-op.
+//
+// Only the LIVE filtered path calls this. Unfiltered reads scan the Cold/Audit
+// zone directly and are always current, so they keep the direct fast path.
+// Checkpoint reads are frozen snapshots and are handled (and documented) at the
+// checkpoint-creation boundary, out of scope here.
+func (impl *BucketServiceServerImpl) waitFilteredAuditConsistency(ctx context.Context, minLogSeq uint64) error {
+	if minLogSeq == 0 {
+		return nil
+	}
+
+	if err := impl.waitMinLogSequence(ctx, minLogSeq); err != nil {
+		return err
+	}
+
+	handle, err := impl.store.NewDirectReadHandle()
+	if err != nil {
+		return status.Errorf(codes.Unavailable, "opening read handle for audit head: %v", err)
+	}
+
+	auditHead, err := query.ReadLastAuditSequence(handle)
+	// Close before waiting: the handle holds dbMu.RLock for its lifetime, and the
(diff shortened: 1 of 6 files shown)


internal/adapter/grpc/server_bucket.go
```diff
@@ -569,6 +569,55 @@ func (impl *BucketServiceServerImpl) waitMinLogSequence(ctx context.Context, min
...
+	// Close before waiting: the handle holds dbMu.RLock for its lifetime, and the
+	// wait below can block for a while. We only need the head value, not the
+	// handle, past this point.
+	_ = handle.Close()
+
+	if err != nil {
+		return status.Errorf(codes.Internal, "reading live audit head: %v", err)
+	}
+
+	return impl.readStore.WaitForAuditSequence(ctx, auditHead)
```
NumaryBot, line 618: 🟠 **[major] Avoid waiting forever when audit indexing is disabled**

When the server is started with the documented `--disable-audit-index` flag, the audit indexer never advances `ReadAuditProgress` or calls `NotifyProgress`. In that configuration, a live filtered `ListAuditEntries` request with `min_log_sequence` after a recent write will wait here for the cursor to reach the live audit head, but that cursor can never advance, so the RPC hangs until the client cancels instead of returning a clear unavailable/unsupported error.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1559#issuecomment-4933635037

Issue #341, opened 2022-09-21
Negative values for account balance


agallan:
**Is your feature request related to a problem? Please describe.**
About the fact that accounts can't have negative values, is it possible to overcome this in some way ?
(In the formance documentation: Accounts in Formance Ledger cannot go negative! (Except for the special @world account).)

jdupas22:
Hey Agallan ! Yes it's possible to have account with negative values, there is a mapping system with some rules you can use to allow this, and soon you'll be able to do apply some rules 'per transaction' if you want to allow overdraft on some accounts ! I'll get back to you soon with the documentation about this :)

edwardmp:
@jdupas22 we are interested in this too, do you have an example Numscript that shows how to allow overdrafting on a transaction level? Could not find it in the docs.

jdupas22:
Hey Edward ! Sorry for the delay ! Yes of course ! 

```
send [USD/2 500] (
  source = @foo allowing overdraft up to [USD/2 100]
  destination = @world
)
```

PR #1436, opened 2026-06-29
feat(EN-1356): per-order skippable_reasons (continue-on-failure)

gfyrag:
## Summary

- Adds a per-`LedgerApplyRequest.skippable_reasons` whitelist so an atomic batch can keep going when a single order trips an idempotency-style business failure — instead of failing the whole proposal. Wire opt-in surfaces at the LedgerApplyRequest envelope (gRPC) and per-entry on the bulk endpoint (JSON); the unitary POST endpoint intentionally does NOT expose it.
- Per-action whitelist is declared in `.proto` via a custom `(allowed_skippable_reasons)` field option on `LedgerAction`'s oneof cases. A new protoc plugin (`tools/protoc-gen-skippable`) generates the Go dispatch (`SkippableReasonsForLedgerAction`) admission consults — the policy stays co-located with the wire declaration and never drifts.
- Current annotations:
  - `create_transaction` → `TRANSACTION_REFERENCE_CONFLICT`
  - `revert_transaction` → `TRANSACTION_ALREADY_REVERTED`
  - `delete_metadata` → `METADATA_NOT_FOUND`
  - `add_account_type` → `ACCOUNT_TYPE_ALREADY_EXISTS`
  - `remove_account_type` → `ACCOUNT_TYPE_NOT_FOUND`
- The FSM wraps the order's scope in an `orderOverlayScope` (buffers writes) inside a `skipSafeScope` decorator (compile-time exhaustive: any new Scope method must be explicitly classified as buffered / read-passthrough / trap). On a whitelisted failure the overlay is dropped without `Commit()`; on success the staged writes flush. Non-buffered mutations (signing keys, chapters, maintenance mode, numscript library, query-checkpoint state) trap with `assert.Unreachable` + `panic()` — the rollback contract is structural, not discipline-based.
- Checker: `verifySkippedOrder` reverifies every `OrderSkipped` projection against the audit-bound `Order.skippable_reasons` whitelist + reason-specific correlator (reference, transaction id, metadata target/key, account-type name) — closing the tampering vector on the not-hash-bound `LedgerLog`. Emits `CHECK_STORE_ERROR_TYPE_INVALID_SKIP` on divergence. A fail-loud `default:` branch forces the checker to be extended whenever a new reason is annotated in `.proto`.
- Admission rejects `UNSPECIFIED` / out-of-whitelist reasons up front (`ErrInvalidSkippableReason` → validation 4xx). Non-empty `skippable_reasons` on an action with no annotation is rejected too — no silent drop.

[...]

PR #1561, opened 2026-07-10
refactor(EN-1504): single source of truth for per-target QueryFilter validation

gfyrag:
## What & why

Jira: [EN-1504](https://formance-team.atlassian.net/browse/EN-1504)

The rules for "which QueryFilter condition is valid on which query target" were scattered across two layers with overlapping-but-different, drift-prone subsets — a gap that already let transaction builtin-uint fields go un-gated until #1547. This PR introduces a **single declarative source of truth** consumed by both layers.

## The table (single source of truth)

`internal/proto/commonpb/query_target_validity.go`, colocated with the `QueryFilter`/`QueryTarget` proto definitions:

- `ConditionKind` — enum at the granularity target-validity is decided (combinators collapse to one `ConditionKindCombinator`; metadata field conditions collapse to `ConditionKindField`).
- `ConditionKindOf(*QueryFilter) ConditionKind` — maps a filter node to its kind; nil / unmapped arm → `ConditionKindUnknown` (never valid).
- `targetConditionValidity` — the authoritative `QueryTarget → ConditionKind → bool` table. Every pair is declared explicitly (no implicit `false`).
- `ConditionValidForTarget(target, kind)` and `TargetHumanName(target)`.

Declarations reproduce the previously scattered guards **exactly**:

| Condition kind | Valid targets (before → after) |
|---|---|
| combinator (`$and`/`$or`/`$not`) | all (no guard existed) |
| metadata (`$match`/`$gt`/`$exists` on `metadata[…]`) | all (no guard; per-target index/schema check unchanged) |
| address | all (no guard existed) |
| reference | all (no guard existed) |
| reverted | TRANSACTIONS only |
| accountHasAsset | ACCOUNTS only |
| builtin uint (`id`/`timestamp`/`insertedAt`/`revertedAt`) | TRANSACTIONS only |
| log field (`date`) | LOGS only |
| logId | LOGS only |
| ledger | LOGS only |

## How the two layers consume it

- **compile layer** (`internal/query/compile.go`): a new `rejectInvalidCondition` runs at the dispatch site in `compile()`, failing with a uniform message *"condition X is not valid on target Y"*. Replaces the per-arm dispatch guards (`logBuiltinUint`/`logId`/`ledger`) and the per-compiler guards in `compileRevertedCondition`, `compileAccountHasAssetCondition`, `compileBuiltinUintCondition`. `targetHumanName` now delegates to `commonpb.TargetHumanName` (DRY).
- **REST decode layer** (`internal/adapter/http/prepared_query_filter.go`): `rejectLogOnlyConditions` is replaced by `rejectConditionsInvalidForAllTargets`, which rejects a condition only when it is invalid on **every** REST-eligible target — the set derived from `preparedQueryTargets`. For today's `{ACCOUNTS, TRANSACTIONS}` set this reproduces the old blanket rejection of `logId`/`date`/`ledger` exactly.

## Completeness test (closes the bug class)

`query_target_validity_test.go` asserts:
- every `ConditionKind` has a name and an explicit entry for **every** target,
- every `QueryFilter` oneof arm maps to a real (non-Unknown) kind,
- nil/empty filters are Unknown and invalid on all targets.

So a new condition arm or a new target **cannot** be added without declaring per-target validity — the class of gap that motivated the ticket, not just the instance.

## Behavior preservation

Pure consolidation: the set of accepted/rejected queries is unchanged. The only behavioral change is that error messages are now uniform (invariant #7 respected — no condition is ever silently dropped; unknown kinds fail loudly). Affected test expectations updated to the new uniform messages.

## Checks

- `GOROOT= go build ./...` ✅
- `go test ./internal/query/... ./internal/adapter/http/... ./internal/proto/commonpb/...` ✅
- `go vet`, `gofmt`, `golangci-lint` on touched packages ✅

## ⚠️ Coordination with EN-1503 (parallel)

EN-1503 (allow LOGS as a REST prepared-query target) overlaps on `prepared_query_filter.go` / the validation logic. Its "make `rejectLogOnlyConditions` target-aware" is **subsumed** by this table: once LOGS is added to `preparedQueryTargets`, `rejectConditionsInvalidForAllTargets` relaxes automatically. This PR intentionally does **not** implement the LOGS-prepared-query feature. Merge order should be **EN-1504 → EN-1503**; expect a conflict on `prepared_query_filter.go` / `compile.go` to be resolved in EN-1503's rebase.

[EN-1504]: https://formance-team.atlassian.net/browse/EN-1504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

## Update — scope grown during review (single source of truth extended)

The original description above covered the compile + REST decode layers. Review surfaced adjacent gaps in the same "one declarative source of truth for target validity" story; they are now closed:

- **Audit is a first-class target.** `QUERY_TARGET_AUDIT` was added to the enum and `QueryFilter_Audit` is annotated valid only for that target, so audit-condition validity lives in the same generated table (`common_queryfilter_validity.pb.go`) as accounts/transactions/logs — no undocumented exception. `CompileAuditFilter` gates on the same table (`and`/`or` valid, `not` not); dispatch stays internal. The proto validity table is now generated from per-arm `allowed_query_targets` annotations via `tools/protoc-gen-queryfilter-validity`, with a build-time error for any unannotated arm (or the explicit `valid_on_no_query_target` marker for the "valid nowhere" case) — the completeness gate is descriptor-driven, so a new arm/target cannot be added without declaring validity.

- **Prepared queries are always executable.** `domain.IsPreparedQueryExecutableTarget` gates prepared-query writes to targets the executor can actually hydrate (ACCOUNTS/TRANSACTIONS today; LOGS lands with #1563). Enforced on **all** write paths — HTTP create, admission, FSM create, and FSM update (the update path reads the stored target from cache, the only layer that sees it) — returning `ErrPreparedQueryTargetUnsupported`. This prevents persisting a query that would fail or return unusable results at execute time.

- **Write-time filter validation is depth-bounded.** `ValidateFilterForTarget` caps recursion at `domain.MaxFilterDepth` (100) — the same bound `query.Compile` enforces at execute time, now the single source of truth in `internal/domain` (`internal/query` re-exports it). A deep-but-valid tree can no longer be persisted only to fail every execution, and the deep-write stack-exhaustion path (#341) stays closed.

Merge-order note (unchanged): **EN-1504 → EN-1503 (#1563)**. #1563 makes LOGS an executable prepared-query target (adds `QUERY_TARGET_LOGS` to `IsPreparedQueryExecutableTarget`), which is why the CLI keeps advertising `--target logs`: the surface and the runtime gate become consistent once #1563 lands directly after this PR.


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1561#issuecomment-4934850795

internal/proto/commonpb/query_target_validity.go
```diff
@@ -0,0 +1,214 @@
...
+// ConditionKindOf maps a QueryFilter node to its ConditionKind. A nil filter or
+// an unmapped arm returns ConditionKindUnknown, which always fails validation
+// (invariant: a condition the validator does not understand is rejected, never
+// silently admitted).
+func ConditionKindOf(f *QueryFilter) ConditionKind {
+	if f == nil {
+		return ConditionKindUnknown
+	}
+
+	switch f.GetFilter().(type) {
```
NumaryBot: 🟠 **[major] Cover the audit filter arm explicitly**

`QueryFilter` already has a `QueryFilter_Audit` oneof arm, but it falls through to `ConditionKindUnknown` here and the new `TestConditionKindOfCoversEveryArm` omits it. That means the completeness gate can pass while an existing filter arm is still unmapped, so drift for audit filters is not caught by the new single-source validation table; please add an explicit audit kind/validity or an explicit documented exclusion in the coverage test.

gfyrag (commented):
Design comment on the QueryFilter validity source of truth. This does not need to be solved by compatibility shims; I think the PR should move the applicability declaration closer to the protobuf surface, like #1436 does for skippable reasons.

internal/proto/commonpb/query_target_validity.go
```diff
@@ -0,0 +1,214 @@
...
+//     on every target (subject to the compiler's own index/schema checks).
+//   - reverted, builtin uint (id/timestamp/insertedAt/revertedAt): TRANSACTIONS
+//     only.
+//   - accountHasAsset: ACCOUNTS only.
+//   - log field (date), logId, ledger: LOGS only.
+//
+// Every (target, kind) pair MUST be present — a missing entry is a bug caught by
+// the completeness test, not a silent "false". The map value is the explicit
+// verdict for that pair.
+var targetConditionValidity = map[QueryTarget]map[ConditionKind]bool{
```
gfyrag: I think this table is the wrong source of truth for the long term. #1436 solved a similar problem by annotating the protobuf `oneof` cases with applicability metadata (`allowed_skippable_reasons`) and generating the Go table from those annotations. QueryFilter target validity would benefit from the same pattern: annotate each `QueryFilter.filter` oneof arm in `common.proto` with the `QueryTarget`s it supports, then generate the `ConditionValidForTarget`/arm-dispatch table.

That would avoid maintaining `ConditionKind`, `allConditionKinds`, `allQueryTargets`, `ConditionKindOf`, and this matrix by hand, and it would also make the completeness gate real: the proto descriptor can enumerate query targets and oneof arms instead of relying on manually maintained test lists. The current manual approach is exactly the kind of drift this PR is trying to remove.

Concretely, I would expect something like a `repeated QueryTarget allowed_query_targets = ...` field option on each `QueryFilter` oneof case, plus a small `protoc-gen-queryfilter-validity` sibling to `protoc-gen-skippable`. While doing that, please also revisit `address`/`reference` on `LOGS`: unless the compiler translates transaction ids to log ids, those should not be marked valid for logs.

internal/proto/commonpb/query_target_validity.go
```diff
@@ -0,0 +1,214 @@
...
+	case *QueryFilter_LogId:
+		return ConditionKindLogID
+	case *QueryFilter_Ledger:
+		return ConditionKindLedger
+	default:
+		return ConditionKindUnknown
+	}
+}
+
+// allQueryTargets lists every QueryTarget. The completeness test iterates it to
```
gfyrag: Follow-up/correction to my earlier design comment: I no longer think `AuditCondition` should stay outside the target-validity model. The tickets around audit filtering (`EN-1241` / `EN-1305`) explicitly moved `ListAuditEntries` toward the shared `ListOptions.filter` contract, `EN-1339` made that index-backed, and `EN-1481` is about exposing those same audit reads over HTTP. So if `QueryTarget` means “target of a `QueryFilter`”, audit should be first-class here.

Concretely, the proto-driven validity approach should probably add `QUERY_TARGET_AUDIT`, annotate `QueryFilter_Audit` as valid only for audit, and mark structural nodes according to what the audit compiler really accepts (`And`/`Or` yes, `Not` no unless `CompileAuditFilter` grows support). The implementation can still dispatch to a separate `CompileAuditFilter` internally; the important part is that target→condition validity is declared in the same source-of-truth as accounts/transactions/logs instead of being an undocumented exception.

gfyrag (commented):
Follow-up to my previous design comment: after checking the Jira context (`EN-1241`, `EN-1305`, `EN-1339`, `EN-1481`, `EN-1459`), I want to correct the audit part. Audit should be treated as a first-class query target in this model.

internal/proto/commonpb/query_target_validity.go
gfyrag: Addressed in 324c90e5a. The audit arm is now a first-class `ConditionKindAudit` mapped in the generated `ConditionKindOf` and present in `allConditionKinds`, and the completeness test is descriptor-driven (`TestConditionKindsCoverEveryOneofArm` enumerates every `QueryFilter.filter` oneof arm from the proto descriptor and asserts each maps to a distinct, non-Unknown kind) — so an unmapped arm now fails the gate instead of silently falling through to Unknown.

gfyrag:
Reworked to make the proto the source of truth, per the design review (324c90e5a).

**Annotation + generator (like #1436's `protoc-gen-skippable`)**
- New field-option extension `common.allowed_query_targets` (extends `google.protobuf.FieldOptions`) annotates each arm of the `QueryFilter.filter` oneof in `common.proto` with the `QueryTarget`s it is valid on.
- New protogen plugin `tools/protoc-gen-queryfilter-validity` (sibling to `protoc-gen-dethash`/`protoc-gen-reader`) reads those annotations from the descriptor and emits `common_queryfilter_validity.pb.go`: the `ConditionKind` enum, `ConditionKindOf`, the `ConditionValidForTarget` table, `TargetHumanName`, and the enumerations. Wired into `just generate-proto`; the hand-written table is deleted.
- The completeness gate is now real: the test enumerates the oneof arms and the `QueryTarget` enum **from the proto descriptor** (no hand lists) and asserts every arm maps to a distinct kind and every (target, kind) pair is decided.

**Audit as a first-class target** (your follow-up)
- Added `QUERY_TARGET_AUDIT` to the enum; `QueryFilter_Audit` is annotated valid only for audit. Combinators are marked per what the compiler really accepts: `and`/`or` valid on audit, `not` **not** valid on audit (matches `CompileAuditFilter`). `compileAuditNode` now gates on the same generated table, so the audit column is load-bearing rather than decorative; dispatch still goes to `CompileAuditFilter` internally. Existing audit tests (incl. `RejectsNonAuditCondition`) still pass.

**FINDING 2 (`address`/`reference` on LOGS)**
- Fixed: `address` is valid on ACCOUNTS+TRANSACTIONS only, `reference` on TRANSACTIONS only. Neither has an account→log translation, so both are no longer (silently) accepted on LOGS.

`just generate-proto` is idempotent (no drift); build + `internal/query` / `internal/adapter/http` / `internal/proto/commonpb` tests green; gofmt + golangci-lint clean on touched packages and the new plugin. EN-1503 coupling unchanged (merge EN-1504 → EN-1503).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1561#issuecomment-4934850795

misc/proto/common.proto
```diff
@@ -1434,6 +1515,14 @@ enum QueryTarget {
...
   QUERY_TARGET_TRANSACTIONS = 1;
   QUERY_TARGET_LOGS = 2;
+  // Audit entries. Dispatched internally to query.CompileAuditFilter (not
+  // query.Compile), but modeled here as a first-class QueryFilter target so
+  // audit condition validity lives in the same source of truth as the other
+  // targets (EN-1241 / EN-1305 / EN-1339 / EN-1481 move ListAuditEntries onto
+  // the shared ListOptions.filter contract). CompileAuditFilter accepts only
+  // audit[...] leaves combined with and/or — not, and every non-audit
+  // condition, are rejected — which the per-arm annotations below reflect.
+  QUERY_TARGET_AUDIT = 3;
```
NumaryBot, line 1542: 🟠 **[major] QUERY_TARGET_AUDIT exposed as a valid PreparedQuery target without full prepared-query support**

Adding `QUERY_TARGET_AUDIT` to the shared `QueryTarget` enum makes it a valid `PreparedQuery.Target` value for gRPC callers, but the prepared-query creation path only validates the name. Execution still goes through `query.Compile`/`executeList` rather than `CompileAuditFilter`, and `queryTargetToJSON` only covers ACCOUNTS/TRANSACTIONS/LOGS. A client can therefore persist an AUDIT prepared query that fails at execution time or returns unusable results.

**Suggestion:** Either explicitly reject `QUERY_TARGET_AUDIT` in prepared-query creation validation, or wire it through all prepared-query paths (compilation, execution, and JSON marshalling) before exposing it in the enum.

tools/protoc-gen-queryfilter-validity/main.go
```diff
@@ -0,0 +1,477 @@
...
+		return nil, fmt.Errorf("marshaling options for %q: %w", field.Desc.FullName(), err)
+	}
+
+	resolved := dynamicpb.NewMessage(opts.ProtoReflect().Descriptor())
+	if err := (proto.UnmarshalOptions{Resolver: resolver}).Unmarshal(raw, resolved); err != nil {
+		return nil, fmt.Errorf("re-parsing options for %q: %w", field.Desc.FullName(), err)
+	}
+
+	typeDesc := extType.TypeDescriptor()
+	if !resolved.Has(typeDesc) {
```
flemzord, line 270: [P2] An unannotated oneof arm silently becomes invalid on every target here, so the claimed completeness gate does not catch a forgotten applicability declaration. I verified this by adding an unannotated arm: `just generate-proto` and the `commonpb` tests still pass because generation adds a distinct kind plus a full-false row, while the expected-matrix test does not require every kind to appear. Please return an error when this extension is absent (or require an explicit marker if “valid nowhere” is intentional).

flemzord (commented):
One actionable issue remains in the generator completeness gate.

gfyrag:
Review pass on `324c90e5ad3a478e352b6c7ce4cc9c8a9cd13b15`.

Findings:

1. Major: `QUERY_TARGET_AUDIT` is now accepted as a persisted `PreparedQuery.target` through gRPC/FSM, but prepared-query execution/listing does not support it. The enum includes AUDIT, admission only validates the query name, and the FSM persists the query unchanged. Execution still calls `query.Compile` rather than `CompileAuditFilter`, list hydration only handles ACCOUNTS/TRANSACTIONS, REST listing cannot marshal AUDIT, and audit filters currently fail JSON marshal. Either reject AUDIT for prepared queries until the execution/listing/marshalling path is implemented, or make AUDIT a fully supported prepared-query target end to end.

2. Major: REST prepared-query create/update validation is still not target-aware. Create decodes and validates the filter before parsing the selected target, and update has no target context. `decodePreparedQueryFilter` rejects only conditions invalid for all REST targets, so REST can still create an ACCOUNTS prepared query with transaction-only filters, or update an existing ACCOUNTS query to one. Please validate against the create target and the stored target on update, using the generated matrix rather than the ACCOUNTS/TRANSACTIONS union.

Outdated compared with earlier review: the audit-arm/manual-list coverage gaps are fixed by generated `ConditionKindAudit` and descriptor enumeration, and the generated matrix now marks address/reference invalid for LOGS.

Validation run: `go test ./internal/proto/commonpb ./internal/query ./internal/adapter/http ./internal/application/admission ./internal/domain/processing` passed.

tools/protoc-gen-queryfilter-validity/main.go
gfyrag: Good catch — you're right, this defeated the anti-drift gate. An unannotated arm became a distinct `ConditionKind` + an all-false row, and `just generate-proto` + the commonpb tests passed, so a forgotten `allowed_query_targets` slipped through silently.

Fixed in c6349c1a0: annotation presence is now a **build-time requirement**, not inferred from absence.

- Added a second field-option extension `bool valid_on_no_query_target = 50124`. Every `QueryFilter.filter` arm must declare **either** a non-empty `allowed_query_targets` **or** `valid_on_no_query_target = true`. (An empty repeated `allowed_query_targets` is indistinguishable from absent on the wire, so the "valid nowhere" intent needs its own explicit boolean marker rather than being inferred.)
- The generator now returns an error (non-zero exit, actionable message) when an arm has neither annotation — so `just generate-proto` **fails** instead of emitting a silent all-false row — and also when an arm declares both (contradictory).

Manually verified all three paths: unannotated arm → `just generate-proto` fails with `QueryFilter oneof arm "…" has no target-validity annotation: add one or more [(common.allowed_query_targets) = QUERY_TARGET_...], or [(common.valid_on_no_query_target) = true] …`; adding a target (or the explicit none-marker) → succeeds; declaring both → fails as contradictory. The currently-annotated arms are unchanged (the 13×4 table is byte-for-byte identical).

misc/proto/common.proto
```diff
@@ -1434,6 +1528,14 @@ enum QueryTarget {
...
```
NumaryBot, line 1542: 🔴 **[blocker] QUERY_TARGET_AUDIT accepted for prepared queries but execution/listing/marshalling paths are not wired**

Adding QUERY_TARGET_AUDIT to the shared QueryTarget enum makes it a valid PreparedQuery.Target for gRPC callers. However, prepared-query admission only validates the query name and the FSM persists the query unchanged. Execution still goes through query.Compile/executeList rather than CompileAuditFilter, list hydration only handles ACCOUNTS/TRANSACTIONS/LOGS, queryTargetToJSON does not cover AUDIT, and audit filters currently fail JSON marshalling. A client can therefore persist an AUDIT prepared query that silently fails or returns unusable results at execution or listing time.

**Suggestion:** Either explicitly reject QUERY_TARGET_AUDIT in prepared-query creation validation (returning an appropriate error) until all prepared-query paths — compilation (CompileAuditFilter dispatch), execution, listing hydration, and JSON marshalling — are implemented for audit; or wire it through all those paths end to end before exposing the enum value.

internal/adapter/http/prepared_query_filter.go
```diff
@@ -60,42 +60,76 @@ func decodePreparedQueryFilter(raw json.RawMessage) (*commonpb.QueryFilter, erro
...
-	// ledger) can never be satisfied here — they would compile a log iterator
-	// and silently return unrelated/empty results. Reject them up front.
-	if err := rejectLogOnlyConditions(filter); err != nil {
+	// (see parsePreparedQueryTarget), so a condition that is invalid on BOTH of
+	// those targets (logId, date, ledger — all log-only) can never be satisfied
+	// here: it would compile a log iterator and silently return unrelated/empty
+	// results. Reject such conditions up front, deriving the verdict from the
+	// same commonpb per-target validity table the compile layer uses, so the two
+	// layers cannot drift.
+	if err := rejectConditionsInvalidForAllTargets(filter, restPreparedQueryTargets); err != nil {
```
NumaryBot: 🟠 **[major] REST filter validation uses union of all targets instead of the selected target**
<sub>reported by gfyrag, codex</sub>

decodePreparedQueryFilter validates a filter against the union of all REST-supported targets. This means a create request for an ACCOUNTS prepared query can store transaction-only conditions (e.g. reference, reverted) because they are valid on TRANSACTIONS. Similarly, an update request has no access to the stored query's target, so target-specific conditions are never rejected. The query is accepted and persisted, then fails or behaves incorrectly at execution time.

**Suggestion:** For create, parse and validate the requested target first, then validate the filter against that specific target using ConditionValidForTarget. For update, fetch or carry the stored query's target and validate the incoming filter against it before accepting the update.

tools/protoc-gen-queryfilter-validity/main.go
```diff
@@ -0,0 +1,516 @@
...
+		return nil, false, fmt.Errorf("re-parsing options for %q: %w", field.Desc.FullName(), err)
+	}
+
+	noneDesc := noneType.TypeDescriptor()
+	if resolved.Has(noneDesc) {
+		validOnNone = resolved.Get(noneDesc).Bool()
+	}
+
+	typeDesc := extType.TypeDescriptor()
+	if !resolved.Has(typeDesc) {
```
NumaryBot, line 270: 🟠 **[major] Unannotated oneof arm silently becomes invalid-everywhere, bypassing the completeness gate**

When a QueryFilter oneof arm lacks the allowed_query_targets field option, the generator emits a distinct ConditionKind for it but fills its validity row with all-false. The expected-matrix test does not require every kind to appear on at least one target, so the completeness gate passes without any error. A developer who forgets to annotate a new arm gets no feedback and the arm is silently rejected on all targets at runtime.

**Suggestion:** Return a hard generation error (or at minimum a test assertion) when a oneof arm is present in the descriptor but has no allowed_query_targets annotation. If 'valid nowhere' is genuinely intentional for a specific arm, require an explicit marker (e.g. a boolean extension flag) so the absence of targets is always a deliberate, documented choice rather than an accidental omission.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1561#issuecomment-4934850795

tools/protoc-gen-queryfilter-validity/main.go
```diff
@@ -0,0 +1,516 @@
...
+//   - a ConditionKind enum (one constant per oneof arm, plus a zero Unknown),
+//   - ConditionKindOf(*QueryFilter) mapping a filter node to its kind,
+//   - the ConditionValidForTarget(target, kind) table,
+//   - TargetHumanName(target),
+//   - and the enumerations (allConditionKinds, allQueryTargets) the
+//     completeness test iterates.
+//
+// Both the compile layer (internal/query) and the REST decode layer
+// (internal/adapter/http) consume the generated table, so they cannot drift.
+// An arm with no annotation maps to "valid on no target" — the fail-safe
```
flemzord: [P3] This header still describes the pre-fix contract: an unannotated arm no longer maps to “valid on no target”; `collectArms` now rejects it unless it declares allowed targets or the explicit none marker. Keeping the old wording can lead maintainers to rely on behavior the generator no longer permits. Please update this paragraph to match the new build-time requirement.

flemzord (commented):
The functional fix is correct; one documentation inconsistency remains.

tools/protoc-gen-queryfilter-validity/main.go
gfyrag: Fixed in d1daae7. Rewrote the header paragraph to describe the current build-time contract: every oneof arm MUST declare exactly one of `common.allowed_query_targets` or `common.valid_on_no_query_target = true`; an arm with neither (or both) fails generation with an error naming the offending arm — no "missing annotation → valid on no target" fallback anymore.

misc/proto/common.proto
NumaryBot, line 1542: 🔴 **[blocker] QUERY_TARGET_AUDIT accepted for prepared queries without execution/listing/marshalling support**

Adding QUERY_TARGET_AUDIT to the shared QueryTarget enum makes it a valid PreparedQuery.Target for gRPC callers. However, prepared-query admission only validates the query name and the FSM persists the query unchanged. Execution still routes through query.Compile/executeList rather than CompileAuditFilter, list hydration only handles ACCOUNTS/TRANSACTIONS/LOGS, and queryTargetToJSON does not cover AUDIT. A client can therefore persist an AUDIT prepared query that silently fails or returns unusable results at execution or listing time.

**Suggestion:** Either explicitly reject QUERY_TARGET_AUDIT in prepared-query creation validation (returning an appropriate error) until all prepared-query paths — compilation (CompileAuditFilter dispatch), execution, listing hydration, and JSON marshalling — are fully implemented for audit; or wire it through all those paths end to end before exposing the enum value.

internal/adapter/http/prepared_query_filter.go
NumaryBot: 🟠 **[major] REST filter validation uses the union of all targets instead of the specific query target**

decodePreparedQueryFilter validates a filter against the union of all REST-supported targets rather than the target of the specific query being created or updated. This means a create request for an ACCOUNTS prepared query can store transaction-only conditions (e.g. reference, reverted) because they are valid on TRANSACTIONS. An update request has no access to the stored query's target, so target-specific conditions are never rejected. The query is accepted and persisted, then fails or behaves incorrectly at execution time.

**Suggestion:** For create, parse and validate the requested target first, then pass that specific target to ConditionValidForTarget. For update, fetch or carry the stored query's target and validate the incoming filter against it before accepting the update.

tools/protoc-gen-queryfilter-validity/main.go
```diff
@@ -0,0 +1,521 @@
...
+//   - a ConditionKind enum (one constant per oneof arm, plus a zero Unknown),
+//   - ConditionKindOf(*QueryFilter) mapping a filter node to its kind,
+//   - the ConditionValidForTarget(target, kind) table,
+//   - TargetHumanName(target),
+//   - and the enumerations (allConditionKinds, allQueryTargets) the
+//     completeness test iterates.
+//
+// Both the compile layer (internal/query) and the REST decode layer
+// (internal/adapter/http) consume the generated table, so they cannot drift.
+//
```
NumaryBot, line 19: ⚪ **[nit] Package-level comment still describes the pre-fix unannotated-arm contract**

The file header still describes the old behavior where an unannotated oneof arm mapped to 'valid on no target' silently. Since c6349c1a0 the generator now returns a hard build-time error for unannotated arms and requires an explicit valid_on_no_query_target marker for the intentional 'valid nowhere' case. Keeping the old wording can mislead maintainers into relying on behavior the generator no longer permits.

**Suggestion:** Update the header comment to accurately describe the new build-time requirement: every QueryFilter oneof arm must declare either a non-empty allowed_query_targets or valid_on_no_query_target = true; omitting both is a generator error.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1561#issuecomment-4934850795

gfyrag:
Review pass on `d1daae721ae40c2f190f9ce18f25d45397888d2c` after the rebase.

Still actionable:

1. Blocker: `QUERY_TARGET_AUDIT` can still be persisted as a prepared-query target without a working prepared-query path. Admission only validates the name, the FSM persists the target unchanged, execution still goes through `query.Compile` rather than audit compilation, list hydration only handles accounts/transactions, and REST JSON target mapping omits AUDIT. Either reject AUDIT for prepared queries at admission/FSM validation, or implement audit prepared queries end to end.

2. Major: REST prepared-query filter validation is still target-unaware. Create decodes the filter before parsing the selected target, and `decodePreparedQueryFilter` validates against the ACCOUNTS/TRANSACTIONS union, so transaction-only filters can still be accepted for ACCOUNTS prepared queries. Update has the same issue because it has no stored-target context. Validate create filters against the parsed target and update filters against the existing query target.

3. Minor: the `AuditCondition` proto comment still says there is no QueryTarget dispatch / no `QUERY_TARGET_AUDIT` enum needed, but this PR now adds and uses `QUERY_TARGET_AUDIT`. Please update the comment and regenerate.

Now fixed/stale: the generated validity header and unannotated-arm generator checks are addressed; proto-annotation validity and first-class AUDIT target modeling are implemented.

misc/proto/common.proto
NumaryBot, line 1542: 🔴 **[blocker] QUERY_TARGET_AUDIT accepted for prepared queries without full execution/listing/marshalling support**

Adding QUERY_TARGET_AUDIT to the shared QueryTarget enum makes it a valid PreparedQuery.Target for gRPC callers. However, prepared-query admission only validates the query name and the FSM persists the query unchanged. Execution still routes through query.Compile/executeList rather than CompileAuditFilter, list hydration only handles ACCOUNTS/TRANSACTIONS/LOGS, and queryTargetToJSON does not cover AUDIT. A client can therefore persist an AUDIT prepared query that silently fails or returns unusable results at execution or listing time.

**Suggestion:** Either explicitly reject QUERY_TARGET_AUDIT in prepared-query creation validation (returning an appropriate error) until all prepared-query paths — compilation (CompileAuditFilter dispatch), execution, listing hydration, and JSON marshalling — are fully implemented for audit; or wire it through all those paths end to end before exposing the enum value.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1561#issuecomment-4934850795

misc/proto/common.proto
NumaryBot, line 1542: 🔴 **[blocker] QUERY_TARGET_AUDIT accepted for prepared queries without full execution/listing/marshalling support**

Adding QUERY_TARGET_AUDIT to the shared QueryTarget enum makes it a valid PreparedQuery.Target for gRPC callers. However, prepared-query admission only validates the query name and the FSM persists the target unchanged. Execution still routes through query.Compile/executeList rather than CompileAuditFilter, list hydration only handles ACCOUNTS/TRANSACTIONS/LOGS, and queryTargetToJSON does not cover AUDIT. A client can therefore persist an AUDIT prepared query that silently fails or returns unusable results at execution or listing time.

**Suggestion:** Either explicitly reject QUERY_TARGET_AUDIT in prepared-query creation validation (returning an appropriate error) until all prepared-query paths — compilation (CompileAuditFilter dispatch), execution, listing hydration, and JSON marshalling — are fully implemented for audit; or wire it through all those paths end to end before exposing the enum value.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1561#issuecomment-4934850795

internal/domain/validation.go
```diff
@@ -161,6 +162,72 @@ func ValidatePreparedQueryName(name string) Describable {
...
+// enforce filter presence separately.
+func ValidateFilterForTarget(f *commonpb.QueryFilter, target commonpb.QueryTarget) Describable {
+	if f == nil {
+		return nil
+	}
+
+	switch v := f.GetFilter().(type) {
+	case *commonpb.QueryFilter_And:
+		for _, sub := range v.And.GetFilters() {
+			if err := ValidateFilterForTarget(sub, target); err != nil {
```
flemzord: [P1] Keep this write-time recursion under the compiler's depth limit. This walk runs in admission and again in the FSM before `query.Compile` can apply `MaxFilterDepth = 100`, so a gRPC client can submit a deeper otherwise-valid `and`/`or`/`not` tree: it is accepted and persisted here, then every execution fails with `ErrFilterTooDeep`; sufficiently deep input also reopens #341's stack-exhaustion path on writes. Please thread a depth counter capped at the same shared limit and add create/update regressions so stored filters remain executable and validation stays bounded.

internal/domain/processing/processor_prepared_query.go
```diff
@@ -93,6 +105,16 @@ func processUpdatePreparedQuery(ledger string, order *raftcmdpb.UpdatePreparedQu
 	}
 
 	updated := existing.Mutate()
+
+	// An update carries only the new filter; the target is fixed at creation and
+	// read here from the stored query (the cache, per invariant #3). Validate the
+	// new filter against that stored target so an update cannot smuggle in a
+	// condition invalid for the query's target — the case admission cannot catch
+	// because the update request has no target field.
+	if err := domain.ValidateFilterForTarget(order.GetFilter(), updated.GetTarget()); err != nil {
```
flemzord, line 125: [P2] Reject legacy non-executable targets on update as well. Before this commit, CLI/gRPC creation explicitly allowed `LOGS`, so such stored queries can already exist; this update path only checks condition validity, meaning a valid log filter is written back even though `ExecutePreparedQuery` still has no LOGS hydration. Please apply `IsPreparedQueryExecutableTarget(updated.GetTarget())` before persisting the update (delete can remain allowed), with a regression covering a pre-existing LOGS query.

flemzord (commented):
Follow-up review of the updated head. These are new, validated findings not covered by the existing threads.

misc/proto/common.proto
flemzord, line 1542: [P3] Please update the `AuditCondition` documentation above to match this new first-class target. It still says that audit has no `QueryTarget` dispatch and that no `QUERY_TARGET_AUDIT` enum value is needed, which is now the opposite of the contract introduced here and can mislead future changes to bypass the generated validity table.

flemzord (commented):
One additional documentation inconsistency remains after the functional re-review.

misc/proto/common.proto
flemzord: Revalidated on `c5dacf7ab`: this blocker is fixed. Both admission (`validateOrderPreparedQuery`) and the FSM create path (`processCreatePreparedQuery`) now reject any target outside ACCOUNTS/TRANSACTIONS via `IsPreparedQueryExecutableTarget`, with coverage for AUDIT.

gfyrag:
Re-reviewed current head `c5dacf7abc04548afa10974fe31ac1d1c28baf02`.

Remaining findings:

1. P1: bound write-time filter validation recursion.

`ValidateFilterApplicableToTarget` recurses through `$and`/`$or`/`$not` without a depth cap. This runs during prepared-query create/update admission/FSM before `query.Compile` can enforce `MaxFilterDepth = 100`, so a deep valid tree can be persisted and then fail on every execution, and a very deep tree can reopen the stack-exhaustion write path. Please use the same depth limit or an iterative traversal, with create/update regressions.

2. P2: reject legacy non-executable targets on update.

Update validates the new filter against the stored target but never checks whether the stored target itself is executable. Existing `LOGS` prepared queries can predate the new create guard, so this path can still update and persist a query that #1561 cannot execute. Add `IsPreparedQueryExecutableTarget(updated.GetTarget())` before persisting updates. This is partly superseded if #1563 lands on top, because #1563 makes `LOGS` executable.

3. P3: update stale `AuditCondition` proto documentation and regenerate.

The proto comment still says audit has no `QueryTarget` dispatch and no `QUERY_TARGET_AUDIT` enum is needed. This PR now adds `QUERY_TARGET_AUDIT` and uses it in the generated validity table. The stale text also appears in generated `common.pb.go`.

What looks resolved since the prior pass:

- AUDIT can no longer be newly persisted as a prepared-query target; create is blocked at admission and FSM.
- REST create validation is now target-aware.
- `QueryFilter_Audit` is first-class in `ConditionKindAudit`.
- Applicable-filter annotations/generator coverage are in place.
- Generator header comments are fixed.

Note on #1563 overlap: #1563 is stacked ahead and makes LOGS executable, so the legacy-LOGS update point may change if #1563 lands immediately after. The recursion issue carries over.

Checks run:

- `GOROOT= go test ./internal/domain ./internal/domain/processing ./internal/application/admission ./internal/adapter/http ./internal/proto/commonpb ./internal/query -run 'Test.*Prepared|Test.*QueryFilter|TestCondition|TestValidateFilter|TestCompileAudit|TestCompile|Test.*Reverted|Test.*Builtin' -count=1`
- `GOROOT= go test . -count=1` in `tools/protoc-gen-queryfilter-validity`
- `GOROOT= go test -tags e2e ./tests/e2e/business -run '^$' -count=0`
- `GOROOT= go build ./...`

internal/domain/validation.go
gfyrag: Fixed in ba6d706ca. `ValidateFilterForTarget` now threads a depth counter capped at the shared `MaxFilterDepth` (100) — the exact bound `query.Compile`/`CompileAuditFilter` enforce at execute time. I moved the constant **and** the `ErrFilterTooDeep` sentinel into `internal/domain` as the single source of truth (`internal/query` re-exports both, so its existing references read unchanged and there is no import cycle). Because the cap lives inside `ValidateFilterForTarget` itself, all four write-time call sites inherit it: HTTP create, admission create, FSM create, and FSM update. A deeper-but-valid tree is now rejected at write time with `ErrFilterTooDeep` instead of being persisted (only to fail every execution) or overflowing the Go stack on the write path (#341). Regression added: `TestValidateFilterForTarget_DepthCap` — a chain at exactly the limit passes, one level deeper is rejected with `ErrFilterTooDeep`.

internal/domain/processing/processor_prepared_query.go
gfyrag: Fixed in ba6d706ca. `processUpdatePreparedQuery` now calls `IsPreparedQueryExecutableTarget(updated.GetTarget())` on the **stored** target before condition validity, and rejects with `ErrPreparedQueryTargetUnsupported` when it is non-executable. So a pre-existing LOGS (or AUDIT) query — creation used to allow LOGS before this PR — can no longer have its filter rewritten into storage while `ExecutePreparedQuery` still has no LOGS hydration. Delete stays allowed. Regression added: `TestProcessUpdatePreparedQuery_RejectsNonExecutableStoredTarget` uses a stored LOGS query with a `logId` filter (valid *for* LOGS, so it isolates the executability gate from condition validity) and asserts the update is rejected with `ErrPreparedQueryTargetUnsupported` and `Put` is never called.

misc/proto/common.proto
gfyrag: Fixed in ba6d706ca. Rewrote the `AuditCondition` doc block: it now states audit is a first-class `QUERY_TARGET_AUDIT` target whose condition validity lives in the same generated source-of-truth table as the other targets, and that dispatch remains internal (`ListAuditEntries` routes `QUERY_TARGET_AUDIT` to `CompileAuditFilter` rather than `query.Compile`). Removed the stale "no QueryTarget dispatch for audit … no QUERY_TARGET_AUDIT enum value is needed" wording. Regenerated proto so the generated Go comment matches.

misc/proto/common.proto
```diff
@@ -1434,6 +1532,14 @@ enum QueryTarget {
...
```
NumaryBot, line 1542: 🔴 **[blocker] QUERY_TARGET_AUDIT accepted for prepared queries without full execution/listing/marshalling support**

Adding QUERY_TARGET_AUDIT to the shared QueryTarget enum makes it a valid PreparedQuery.Target value for gRPC callers. However, prepared-query admission only validates the query name, the FSM persists the target unchanged, execution still routes through query.Compile/executeList rather than CompileAuditFilter, list hydration only handles ACCOUNTS/TRANSACTIONS/LOGS, and queryTargetToJSON does not cover AUDIT. A client can therefore persist an AUDIT prepared query that silently fails or returns unusable results at execution or listing time.

**Suggestion:** Either explicitly reject QUERY_TARGET_AUDIT in prepared-query creation validation (returning an appropriate error) until all prepared-query paths — compilation (CompileAuditFilter dispatch), execution, listing hydration, and JSON marshalling — are fully implemented for audit; or wire it through all those paths end to end before exposing the enum value.

internal/domain/validation.go
```diff
@@ -161,6 +187,85 @@ func ValidatePreparedQueryName(name string) Describable {
...
+// target JSON mapping — so a prepared query stored on those targets would fail
+// later at execute/marshal time. Enforced at write time (admission + FSM) across
+// gRPC and HTTP so a persisted prepared query is always executable. LOGS becomes
+// executable with EN-1503.
+func IsPreparedQueryExecutableTarget(target commonpb.QueryTarget) bool {
+	switch target {
+	case commonpb.QueryTarget_QUERY_TARGET_ACCOUNTS,
+		commonpb.QueryTarget_QUERY_TARGET_TRANSACTIONS:
+		return true
+	default:
```
NumaryBot, line 204: 🟠 **[major] LOGS prepared queries silently rejected while CLI/docs still advertise LOGS as a valid target**

The new executable-target gate returns false for LOGS, so requests from the existing `ledgerctl queries create --target logs` path are rejected at admission/FSM even though the CLI help, completions, parser, and documentation still advertise `logs` as a valid prepared-query target. Users can choose an option that always fails with no clear explanation.

**Suggestion:** Either remove LOGS from all client/config surfaces (CLI help, completion, parser, docs) until EN-1503 makes it a fully executable prepared-query target, or defer rejecting LOGS at this gate until that work lands. The user-facing surface and the runtime gate must be consistent.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1561#issuecomment-4934850795

misc/proto/common.proto
gfyrag: Rebutting — this is a loop of an already-fixed finding. QUERY_TARGET_AUDIT (and any target outside ACCOUNTS/TRANSACTIONS) can **not** be persisted as a prepared query: `domain.IsPreparedQueryExecutableTarget` returns false for AUDIT and LOGS, and it is enforced on **both** write paths — admission (`validateOrderPreparedQuery`) and the FSM create path (`processCreatePreparedQuery`, processor_prepared_query.go:49) — returning `ErrPreparedQueryTargetUnsupported`. flemzord already re-validated exactly this on c5dacf7ab ("Both admission and the FSM create path now reject any target outside ACCOUNTS/TRANSACTIONS via IsPreparedQueryExecutableTarget, with coverage for AUDIT"). ba6d706ca additionally closed the update path (processor_prepared_query.go:118) so a legacy non-executable target cannot be rewritten either. So the "client can persist an AUDIT prepared query" premise no longer holds — there is nothing left to wire because the value is rejected at write time by design until the execution path exists. Test coverage: `TestIsPreparedQueryExecutableTarget` (AUDIT/LOGS → false) and the FSM reject tests. Resolving.

internal/domain/validation.go
gfyrag: Rebutting — deferred to #1563 by design, not a defect in this PR. The executable-target gate deliberately rejects LOGS because #1561 does **not** implement LOGS prepared-query execution (no `query.EnrichLogs` hydration yet); relaxing the gate would let a non-executable LOGS query be persisted, which is the exact bug the update-path P2 thread flagged. The CLI surface (`queries create --target logs`, completion, help) staying is intentional: #1563 (feat/EN-1503) is the stacked follow-up that makes LOGS executable — it adds `QUERY_TARGET_QUERY_TARGET_LOGS` to `IsPreparedQueryExecutableTarget` (wired via `query.EnrichLogs`), at which point CLI surface and runtime gate are consistent again. This EN-1504 → EN-1503 merge order and overlap is called out explicitly in the PR description. Removing `logs` from the CLI here would just be reverted by #1563 one merge later. NumaryBot's alternative ("defer rejecting LOGS at this gate") is explicitly rejected: the gate must stay closed until the execution path lands. Resolving as intended/deferred.

gfyrag:
## Status vs current head `ba6d706ca` — the automated "Changes requested" summary is stale

Every finding in the latest NumaryBot review is either **already fixed** in a cited commit or **deferred to #1563 by design**. All inline threads are resolved; nothing actionable remains on this PR. Point-by-point:

**🔴 `QUERY_TARGET_AUDIT` persistable without execution support — FIXED (c5dacf7ab).**
AUDIT (and every target outside ACCOUNTS/TRANSACTIONS) is rejected at write time by `domain.IsPreparedQueryExecutableTarget`, enforced on both write paths: admission (`validateOrderPreparedQuery`) and the FSM create path (`processCreatePreparedQuery`), returning `ErrPreparedQueryTargetUnsupported`. `ba6d706ca` additionally closed the **update** path so a legacy non-executable target cannot be rewritten. A client can no longer persist an AUDIT prepared query, so there is nothing to "wire end to end" — the value is rejected until the execution path exists. @flemzord already re-validated this exact point on `c5dacf7ab`.

**🟠 REST filter validation used union of all targets — FIXED (c5dacf7ab).**
`decodePreparedQueryFilter` / the create handler now validate against the *specific* requested target via `ConditionValidForTarget`, and the FSM update path validates the incoming filter against the **stored** target (`processUpdatePreparedQuery`, processor_prepared_query.go:125) — the case REST admission cannot see. Covered by `TestProcessUpdatePreparedQuery_RejectsFilterInvalidForStoredTarget`.

**🟠 `ValidateFilterForTarget` recurses without a depth cap — FIXED (ba6d706ca).**
The walk now threads a depth counter capped at `domain.MaxFilterDepth` (100) — the same bound `query.Compile` enforces at execute time, moved into `internal/domain` as the single source of truth (`internal/query` re-exports it). Applies at all four write sites (HTTP create, admission create, FSM create, FSM update). Regression: `TestValidateFilterForTarget_DepthCap` (at-limit passes, over-limit → `ErrFilterTooDeep`).

**🟡 Stale `AuditCondition` proto comment — FIXED (ba6d706ca).**
Rewritten to describe audit as a first-class `QUERY_TARGET_AUDIT` target (dispatch stays internal to `CompileAuditFilter`); regenerated so `common.pb.go` matches.

**🟠 LOGS rejected while CLI advertises it — DEFERRED to #1563 by design (not a defect here).**
The gate deliberately rejects LOGS because #1561 does not implement LOGS prepared-query execution; relaxing it would let a non-executable query be persisted (the exact bug the update-path fix closes). #1563 (feat/EN-1503) makes LOGS executable (adds `QUERY_TARGET_LOGS` to `IsPreparedQueryExecutableTarget`, wired via `query.EnrichLogs`), restoring CLI/runtime consistency. This EN-1504 → EN-1503 merge order is documented in the PR description. Removing `logs` from the CLI here would be reverted by #1563 one merge later.

Verified at head: `GOROOT= go build ./...`, `GOROOT= go test ./internal/domain/... ./internal/domain/processing/... ./internal/query/...`, `golangci-lint` on touched packages — all green.

flemzord (approved):
Re-reviewed the current head after the latest fixes. The write-time depth bound now matches execution, legacy non-executable targets are rejected on update, and the audit-target documentation is current. All substantive checks pass and no actionable findings remain.

PR #1531, opened 2026-07-08
feat(EN-1472): expose BucketService reads + index management over HTTP

gfyrag:
## Summary

Aligns the HTTP surface with the gRPC `BucketService` for the read-side
business flows and for the full index lifecycle. REST-only clients no
longer need to fall back to gRPC for `ListTransactions`, per-log lookup,
chapters, signing keys, event sinks, or index management.

Ticket: [EN-1472](https://formance-team.atlassian.net/browse/EN-1472).
Follow-up: [EN-1481](https://formance-team.atlassian.net/browse/EN-1481)
(audit index wiring + HTTP expo, see "Audit follow-up" below).

### Business reads (BucketService parity)

| Method | Path | Scope | gRPC counterpart |
|---|---|---|---|
| GET | `/v3/{ledger}/transactions` | `ledger:TransactionRead` | `ListTransactions` |
| GET | `/v3/_/logs/{sequence}` | `ledger:OpsRead` | `GetLog` |
| GET | `/v3/_/chapters` | `ledger:OpsRead` | `ListChapters` |
| GET | `/v3/_/chapter-schedule` | `ledger:OpsRead` | `GetChapterSchedule` |
| GET | `/v3/_/events-sinks` | `ledger:OpsRead` | `GetEventsSinks` |
| GET | `/v3/_/signing-keys` | `ledger:OpsRead` | `ListSigningKeys` |

Bucket-wide / non-ledger reads live under the reserved `_` path segment
(`/v3/_/…`) so they can never shadow a `/v3/{ledger}/…` route — the ledger
name `_` is reserved. Responses serialize the protobuf message in
protobuf-JSON camelCase, wrapped in the `{data:…}` envelope.

### Index management (ledger-scoped)

| Method | Path | Scope | Behaviour |
|---|---|---|---|
| GET | `/v3/{ledger}/indexes` | `ledger:LedgerRead` | list per-ledger |
| POST | `/v3/{ledger}/indexes` | `ledger:LedgerWrite` | create (201, body: `{"data":{"id":"<canonical>"}}`) |
| GET | `/v3/{ledger}/indexes/{canonicalId}` | `ledger:LedgerRead` | get single Index entry |
| GET | `/v3/{ledger}/indexes/{canonicalId}/status` | `ledger:LedgerRead` | IndexEntry (build status + cursor + version) |
| GET | `/v3/{ledger}/indexes/{canonicalId}/inspect` | `ledger:LedgerRead` | InspectIndex (metadata only) |
[...]

PR #1562, opened 2026-07-10
feat(EN-1481): expose audit reads over HTTP

gfyrag:
## Summary

Exposes audit reads over the REST API, completing [EN-1481](https://formance-team.atlassian.net/browse/EN-1481). The audit index wiring already shipped in #1548 / EN-1241 (now on `release/v3.0`); this PR is the HTTP exposure only. Follows the HTTP router/handler conventions established in draft #1531 (which intentionally deferred audit HTTP to this ticket), based on `release/v3.0`.

## New HTTP endpoints

| Method | Path | Scope | gRPC counterpart |
|---|---|---|---|
| GET | `/v3/_/audit-entries` | `ledger:AuditRead` | `BucketService.ListAuditEntries` |
| GET | `/v3/_/audit-entries/{sequence}` | `ledger:AuditRead` | `BucketService.GetAuditEntry` |

Audit reads are **bucket-wide, not ledger-scoped** — a single proposal can touch several ledgers — so the routes live under the reserved system segment `/v3/_/…` rather than under `/v3/{ledgerName}/…`. The two namespaces are disjoint by construction: the ledger name `_` is reserved at admission (`admission.ErrLedgerNameReservedPrefix`, enforced in `validateOrderLedgerName` on **every** ledger-scoped order), so a ledger named `_` can never be created or written to and can never shadow the system routes. (This replaces the earlier top-level `/v3/audit-entries` + per-word `audit-entries` reserved-name approach, which was dropped: a static top-level segment shadowed same-named ledgers, and a per-word guard did not scale to new system routes.) Ledger scope is expressed as a filter condition.

**List** query params: `pageSize` (default 100, cap 1000), `after` (opaque uint64 sequence cursor, exclusive), `reverse`, `filter`. It exposes the same audit data and filter grammar as `ctrl.Controller.ListAuditEntries`; like every other release/v3.0 list endpoint it is `after`-cursor driven (no server next-token). **Get** returns a single entry by sequence with per-order `items` populated; missing sequence → 404.

## Auth

`internalauth.ScopeAuditRead` (`"ledger:AuditRead"`) — confirmed in `internal/adapter/auth/scopes.go`, identical to the gRPC audit reads. Enforced via `RequireScope` middleware in `handler.go`, matching sibling routes.

## Filter reuse (deviation from the ticket, after challenging it)

The ticket said to reuse the QueryFilter **JSON** codec (`internal/proto/commonpb/query_filter.go`). That codec **deliberately rejects audit conditions** (`"audit conditions are not supported over REST JSON (gRPC-only)"`) because audit field names (`ledger`, `timestamp`, …) collide with the transaction/log conditions the field-dispatched JSON DSL already claims — it fails loudly by design (invariant #7).

The canonical audit-filter surface is instead the shared **`filterexpr` text DSL** (`internal/pkg/filterexpr`), which fully supports `audit[field] OP value` → `commonpb.QueryFilter` and is exactly what `ledgerctl audit list --filter` uses. This PR reuses `filterexpr.Parse` for the HTTP `filter` param, so **no second audit filter syntax is introduced** — the ticket's actual intent. Examples: `audit[outcome] == failure`, `audit[ledger] == main`, `audit[order_type] in (create_transaction, revert_transaction)`, `audit[seq] between 1000 and 2000`.

## camelCase JSON

`auditpb.AuditEntry` has no generated JSON marshaler, so default sonic encoding would leak snake_case struct tags (`proposal_id`, `order_count`, …). Added a hand-written `MarshalJSON` (`internal/proto/auditpb/audit.pb.json.go`) mirroring `commonpb/common.pb.json.go`: camelCase keys, RFC3339 timestamps via the commonpb marshaler, hex-encoded hash/serialized-order, enum reason as its string name, and protojson for sub-messages lacking a Go marshaler (`CallerSnapshot`/`Idempotency`/`SignedApplyBatch`), with a typed-nil guard so unset sub-messages are omitted rather than rendered as `{}`.

## Invariants respected (read-side only)

- No FSM apply-path / cache / preload changes; no Pebble writes.
- Audit log remains the source of truth; the HTTP path consumes the **same** controller/query path — no new persisted projection.
- Consistency: this endpoint always performs a **live, best-effort** read. Unlike the gRPC surface it does **not** expose the read-consistency options (`checkpointId` pinned reads, `minLogSequence` audit-index catch-up wait); a filtered live read resolves through the async audit index and may transiently omit very recent unindexed entries. Documented as such in OpenAPI + the handler; clients needing a pinned/consistency-bounded audit read use gRPC.
- No silent skips (invariant #7): malformed filter / after / sequence → 400, missing entry → 404. Audit DTO marshaling failures are buffered before the response header is written (`writeOKChecked`), so a submessage marshal error yields a clean 500, never a truncated 200. Numeric audit fields whose zero value is meaningful (`orderIndex`, `logSequence`, `minLogSequence`/`maxLogSequence`, plus entry identity/counter scalars) are serialized unconditionally (no `omitempty`).

## Deviation: e2e

No e2e/API test added. The entire `tests/e2e` suite is gRPC-only — there is no REST/HTTP client harness and no comparable REST audit surface to mirror (the ticket made this test conditional on such a surface existing). Handler-level tests drive the real chi router in a full-route integration test, providing equivalent coverage.

## Test plan

- [x] `GOROOT= go test ./internal/adapter/... ./internal/proto/auditpb/` — green
- [x] `GOROOT= gofmt -l` clean on touched files
- [x] `golangci-lint run` on touched packages — 0 issues
- New tests: list (success/empty; empty serializes as `[]`), pagination cursor, reverse, ledger filter, outcome filter, invalid page-size/after/filter, get-by-sequence (success/not-found/invalid/missing), full-router integration (verifies the `/v3/_/…` system segment is disjoint from `/{ledgerName}`), `_` reserved-name admission (rejected on create and on write), audit DTO zero-value preservation and marshal-failure→500, and AuditEntry/AuditFailure JSON camelCase codec.

## Docs

`openapi.yml` (2 operations + `AuditEntry`/`AuditItem`/`AuditSuccess`/`AuditFailure` schemas), `docs/technical/architecture/subsystems/api/http-api.md` (Audit section), `docs/technical/contributing/api-comparison.md` (audit read rows).

Refs [EN-1481](https://formance-team.atlassian.net/browse/EN-1481).

[EN-1481]: https://formance-team.atlassian.net/browse/EN-1481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EN-1481]: https://formance-team.atlassian.net/browse/EN-1481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/adapter/http/handlers_list_audit_entries.go
```diff
@@ -0,0 +1,94 @@
...
+// parseAuditFilter parses the optional `filter` query parameter into a
+// QueryFilter using the shared filterexpr grammar. An absent filter yields a nil
+// filter (unfiltered read). A malformed filter is a 400.
+func parseAuditFilter(w http.ResponseWriter, r *http.Request) (*commonpb.QueryFilter, bool) {
+	expr := r.URL.Query().Get("filter")
+	if expr == "" {
+		return nil, true
+	}
+
+	filter, err := filterexpr.Parse(expr)
```
NumaryBot, line 99: 🟠 **[major] Reject semantically invalid audit filters before forwarding**

For filters that parse but the audit compiler rejects, such as `audit[outcome] == maybe`, `not audit[outcome] == failure`, or `ledger == main`, this returns a `codes.InvalidArgument` status from `ListAuditEntries`; `handleError` does not translate gRPC status errors, so the HTTP response becomes a 500 instead of the documented 400. Please validate that the parsed filter is audit-supported here or map that invalid-argument error before falling through.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1562#issuecomment-4934913701

internal/adapter/http/handler.go
```diff
@@ -115,6 +116,15 @@ func NewHandler(logger logging.Logger, backend Backend, authCfg internalauth.Aut
 				r.Get("/{ledgerName}/transactions/{transactionId}", server.handleGetTransaction)
 			})
 
+			// Audit read scope. Audit reads are cluster/bucket-wide (a proposal can
+			// touch several ledgers), so these routes are NOT under {ledgerName};
+			// ledger scope is expressed via the `filter` query parameter. The static
+			// /audit-entries segment is matched before the /{ledgerName} wildcard.
+			r.With(requireAuditRead).Group(func(r chi.Router) {
+				r.Get("/audit-entries", server.handleListAuditEntries)
```
flemzord: 🟠 **[major] Preserve access to ledgers named `audit-entries`**

This static prefix shadows the existing `/{ledgerName}` namespace even though `audit-entries` is a valid ledger name (`ValidateLedgerName` allows it and no reserved-name rule exists). `GET /v3/audit-entries` stops retrieving that ledger, and reads such as `/v3/audit-entries/accounts` bind `accounts` as `{sequence}` instead of reaching the ledger handler. Please move the audit routes outside the ledger-name namespace, or reserve/migrate this name and add compatibility coverage.

internal/adapter/http/handlers_list_audit_entries.go
```diff
@@ -0,0 +1,94 @@
...
+
+		return
+	}
+
+	entries, ok := drainCursor(w, r, cursor)
+	if !ok {
+		return
+	}
+
+	writeOK(w, entries)
```
flemzord: 🟡 **[minor] Keep empty list responses as arrays**

When the cursor is empty, `drainCursor` returns a nil slice (`var items []T`), so this writes `{"data":null}` even though the new OpenAPI schema promises an array. Generated clients may fail when iterating a documented list response. Please normalize empty results to a non-nil slice (ideally in `drainCursor`) and assert `data: []` in the empty-response test.

internal/proto/auditpb/audit.pb.json.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil
+	}
+
+	if v := reflect.ValueOf(msg); v.Kind() == reflect.Pointer && v.IsNil() {
+		return nil
+	}
+
+	b, err := protojson.Marshal(msg)
+	if err != nil {
+		return nil
```
flemzord: 🟠 **[major] Surface audit submessage serialization failures**

If `protojson.Marshal` fails (for example on invalid UTF-8), returning nil makes `AuditEntry.MarshalJSON` succeed while silently omitting `callerSnapshot`, `idempotency`, or `signature`. Those fields are chain-bound audit evidence; an encoding defect must not look like a valid incomplete record, and the repository invariant requires impossible branches to fail loudly. Please return and propagate this error from `MarshalJSON` instead of dropping the field.

flemzord (commented):
Found three actionable issues in the HTTP audit exposure.

gfyrag:
Review pass on `6cd8973edd12971e0dc8dca0dcdcba62ebe3ca52`.

Findings:

1. Major: `/v3/audit-entries` shadows a valid ledger name. The new static audit routes live inside the same `/v3/{ledgerName}` namespace, and `audit-entries` is not reserved. For a ledger named `audit-entries`, `GET /v3/audit-entries/accounts` binds to the audit `{sequence}` route and returns audit behavior/400 instead of listing accounts. Please move the audit routes out of the ledger namespace, or explicitly reserve/migrate that ledger name with compatibility tests.

2. Major: HTTP list lacks gRPC read-consistency parity. The handler parses `pageSize`, `after`, `filter`, and `reverse`, then calls the controller directly. gRPC `ListAuditEntries` honors `ReadOptions` such as `checkpointId` and `minLogSequence`, including the filtered-audit-index wait path. The new HTTP docs claim the consistency guarantee matches gRPC, but there are no query params or implementation for those semantics. Please add the same read options through the same path, or document/reject the unsupported options explicitly.

3. Major: semantically invalid audit filters become HTTP 500. `parseAuditFilter` checks syntax only; filters such as `audit[outcome] == maybe`, `not audit[outcome] == failure`, or `ledger == main` parse and are rejected later by `CompileAuditFilter`. `handleError` does not map gRPC `InvalidArgument` status errors, so this falls through to 500. Please map `codes.InvalidArgument` to 400 or validate before forwarding.

4. Major: audit JSON silently drops submessage marshal errors. `protoFieldJSON` returns nil on `protojson.Marshal` failure, so `callerSnapshot`, `idempotency`, or `signature` can disappear from an otherwise successful `AuditEntry.MarshalJSON`. Please return the error and propagate it from `MarshalJSON`.

5. Minor: empty list responses encode `data:null` instead of an empty array, while OpenAPI declares `data` as an array. Please initialize the drained slice to empty and assert the response shape in tests.

The unresolved inline review threads I checked are current and actionable; the NumaryBot top-level changes-requested comment duplicates the invalid-filter thread.

Validation run: `GOROOT= go test ./internal/adapter/http ./internal/proto/auditpb` and `git diff --check release/v3.0...HEAD` passed.

internal/adapter/http/handler.go
gfyrag: Fixed in 1f13a27d4. Investigated the established pattern first: the top-level static routes added in #1531 (`/v3/logs/{sequence}`, `/v3/chapters`, `/v3/signing-keys`, `/v3/events-sinks`, `/v3/indexes`) have the identical collision and rely purely on chi routing precedence (static segment matched before `/{ledgerName}` — I confirmed empirically: `/v3/audit-entries/accounts` binds `accounts` as `{sequence}`). There is no reserved-name mechanism anywhere in the repo, and `ValidateLedgerName` (invariants v0.11.0) allows `audit-entries`.

Rather than special-case the router, I reserve the name at ledger-creation **admission** (`validateOrderReservedLedgerName`, create-only) so a shadowed ledger can never be created — this is pre-Raft validation, not the FSM apply path, so it respects this ticket's read-side-only constraint. `reservedLedgerNames` + `ErrLedgerNameReserved` live in the admission package; `handler.go` and the openapi `createLedger` description cross-reference it. Scoped to just `audit-entries` (this PR's route) to avoid conflicting with #1531, which can add its own words on merge. Covered by `TestValidateOrder_LedgerName` (reserved create rejected; Apply on the name not spuriously rejected).

internal/adapter/http/handlers_list_audit_entries.go
gfyrag: Fixed in 1f13a27d4. Normalized in the shared `drainCursor` (`items := []T{}` instead of `var items []T`), so all four list handlers (audit, logs, accounts, all-ledgers) now emit `{"data":[]}` for an empty result — the other three had the same latent `{"data":null}` bug. Added `require.JSONEq(t, "{\"data\":[]}", ...)` to the audit empty-list test. Verified the full `internal/...` suite stays green.

internal/proto/auditpb/audit.pb.json.go
gfyrag: Fixed in 1f13a27d4. `protoFieldJSON` now returns `(json.RawValue, error)` and `AuditEntry.MarshalJSON` propagates it with context (`audit entry %d: marshaling callerSnapshot: %w`), so a protojson failure fails the whole marshal instead of emitting a valid-looking record with the chain-bound field dropped (invariant #7). The typed-nil guard is retained so genuinely-absent submessages are omitted (not rendered as `{}`). Added `TestAuditEntry_MarshalJSON_PropagatesSubmessageError`, which forces the failure via an invalid-UTF8 `callerSnapshot.scopes` value.

internal/adapter/http/handlers_list_audit_entries.go
gfyrag: Fixed in 1f13a27d4. Confirmed: `query.CompileAuditFilter` (shared with the gRPC path) rejects filters that parse but are not audit-supported (`not audit[...]`, a non-audit condition, unknown fields) with `codes.InvalidArgument`, and `handleError` did not translate gRPC status errors → 500. Added an `InvalidArgument` → HTTP 400 branch in `handleError` (only `InvalidArgument` is mapped; other codes keep the 500 fallthrough deliberately, since no other HTTP handler is expected to surface gRPC statuses). Added `TestHandleListAuditEntries_UnsupportedFilterMapsTo400`. Note the parse-time errors (malformed expression) were already 400 via `parseAuditFilter`; this covers the compile-time rejections.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1562#issuecomment-4934913701

internal/adapter/http/handler.go
```diff
@@ -115,6 +116,20 @@ func NewHandler(logger logging.Logger, backend Backend, authCfg internalauth.Aut
...
+			// touch several ledgers), so these routes are NOT under {ledgerName};
+			// ledger scope is expressed via the `filter` query parameter. The static
+			// /audit-entries segment is matched by chi before the /{ledgerName}
+			// wildcard, so it would shadow a ledger literally named "audit-entries".
+			// That name is therefore reserved at ledger-creation admission
+			// (admission.reservedLedgerNames / ErrLedgerNameReserved) so a shadowed
+			// ledger can never be created — keep the two in sync when adding
+			// top-level static routes here.
+			r.With(requireAuditRead).Group(func(r chi.Router) {
+				r.Get("/audit-entries", server.handleListAuditEntries)
```
NumaryBot: 🟠 **[major] Audit routes shadow ledgers named `audit-entries`**
<sub>reported by flemzord, gfyrag, codex</sub>

The new static `/v3/audit-entries` prefix lives inside the same `/{ledgerName}` namespace. `audit-entries` is a valid ledger name (no reserved-name rule prevents it). On upgraded clusters that already have such a ledger, `GET /v3/audit-entries` and `GET /v3/audit-entries/accounts` will match the new audit routes instead of the ledger routes, breaking existing clients. Reserving the name only at create-time does not protect pre-existing ledgers.

**Suggestion:** Move the audit routes outside the ledger-name namespace (e.g. `/v3/_audit/entries`), or add an explicit migration/compatibility path and cover the edge case with tests.

internal/adapter/http/handlers_list_audit_entries.go
NumaryBot, line 99: 🟠 **[major] Semantically invalid audit filters produce HTTP 500 instead of 400**
<sub>reported by NumaryBot, gfyrag, flemzord</sub>

The filter parsing step (`parseAuditFilter`) checks syntax only. When the audit compiler rejects a syntactically valid but semantically invalid filter (e.g. `audit[outcome] == maybe`, `ledger == main`), `ListAuditEntries` returns a `codes.InvalidArgument` gRPC status. Because `handleError` does not translate gRPC status codes, this propagates as an HTTP 500 instead of the documented 400. This was also flagged by the automated NumaryBot comment on this thread.

**Suggestion:** Validate that the parsed filter is audit-supported before forwarding to the controller, or map `codes.InvalidArgument` to HTTP 400 in `handleError`.

internal/proto/auditpb/audit.pb.json.go
```diff
@@ -0,0 +1,154 @@
...
+// field is simply absent.
+//
+// A protojson failure is PROPAGATED, never swallowed. These fields
+// (callerSnapshot, idempotency, signature) are chain-bound audit evidence; a
+// serialization defect must surface as a failed response, not a valid-looking
+// record with the field silently dropped (invariant #7: never silently skip a
+// "should not happen" branch).
+func protoFieldJSON(msg proto.Message) (json.RawValue, error) {
+	if msg == nil {
+		return nil, nil
```
NumaryBot, line 43: 🟠 **[major] Audit submessage serialization errors are silently swallowed**

If `protojson.Marshal` fails (e.g. on invalid UTF-8 in a submessage), returning `nil` makes `AuditEntry.MarshalJSON` succeed while silently omitting `callerSnapshot`, `idempotency`, or `signature`. These are chain-bound audit evidence fields; a silent encoding defect produces a structurally valid but incomplete audit record, violating the integrity guarantees of the audit log.

**Suggestion:** Return and propagate the error from `MarshalJSON` instead of ignoring it, so callers receive an error rather than a silently truncated record.

internal/adapter/http/handlers_list_audit_entries.go
NumaryBot: 🟡 **[minor] Empty list responses return `null` instead of `[]`**

When the cursor is empty, `drainCursor` returns a nil slice, causing the JSON response to serialize as `{"data":null}` rather than `{"data":[]}`. The new OpenAPI schema promises an array, and generated clients may fail when iterating what they expect to be a list.

**Suggestion:** Normalize empty results to a non-nil empty slice (ideally in `drainCursor` or at the handler level) and assert `data: []` in the empty-response test.

gfyrag:
Review pass on `1f13a27d459c6ad9ff8df8d15cd40243d41032a5` after the rebase.

Still actionable:

1. Major: `/v3/audit-entries` still breaks pre-existing ledgers named `audit-entries`. The fix reserves the name only for future `CreateLedger` admission, but upgraded clusters may already have such a ledger. The router still registers static audit routes before `/{ledgerName}`, so existing REST reads such as `/v3/audit-entries/accounts` remain shadowed. Please move the audit route outside the ledger namespace or add an explicit migration/compatibility path for existing ledgers.

2. Major: HTTP audit list still lacks gRPC read-option parity while docs claim parity. HTTP parses only `pageSize`, `after`, `filter`, and `reverse`, then calls `ListAuditEntries` directly. gRPC honors `checkpoint_id` and `min_log_sequence`, including checkpoint reads and filtered audit-index consistency waits. OpenAPI/docs still say the HTTP consistency guarantee matches gRPC. Please add `checkpointId` / `minLogSequence` support or explicitly document/reject the unsupported semantics.

Fixed compared with my previous pass: invalid audit filters now map to 400, empty lists return non-nil arrays, and audit submessage marshal errors are propagated with coverage.

Validation run: `git diff --check 6cd8973e..1f13a27d459c6ad9ff8df8d15cd40243d41032a5`, `GOROOT= go test ./internal/adapter/http ./internal/application/admission ./internal/proto/auditpb`, and `GOROOT= go build ./...` passed.

internal/adapter/http/handler.go
NumaryBot: 🟠 **[major] Audit routes shadow pre-existing ledgers named `audit-entries`**
<sub>reported by flemzord, gfyrag, codex</sub>

Reserving the name `audit-entries` only at ledger-creation admission does not protect upgraded clusters that already have a ledger with that name. The router still registers the static `/v3/audit-entries` routes before `/{ledgerName}`, so `GET /v3/audit-entries` and `GET /v3/audit-entries/accounts` will match the audit routes instead of the ledger routes for any pre-existing ledger, permanently breaking existing clients on upgrade. The create-time guard prevents future collisions but does nothing for the existing ones.

**Suggestion:** Move the audit routes outside the ledger-name namespace entirely (e.g. `/v3/_audit/entries`), or add an explicit migration/compatibility path with tests that cover the pre-existing ledger scenario.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1562#issuecomment-4934913701

internal/proto/auditpb/audit.pb.json.go
```diff
@@ -0,0 +1,154 @@
...
+		Failure:     x.GetFailure(),
+		OrderCount:  x.GetOrderCount(),
+		Items:       x.GetItems(),
+		Ledgers:     x.GetLedgers(),
+		HashVersion: x.GetHashVersion(),
+	}
+
+	var err error
+
+	if aux.CallerSnapshot, err = protoFieldJSON(x.GetCallerSnapshot()); err != nil {
```
NumaryBot, line 98: 🟠 **[major] HTTP 200 already committed when audit submessage MarshalJSON fails**

Now that `MarshalJSON` correctly returns an error when `protojson.Marshal` fails on a submessage (e.g. invalid UTF-8 in `callerSnapshot`, `idempotency`, or `signature`), the error is surfaced too late: the HTTP response writer has already sent the 200 status header before encoding begins. A marshaling failure therefore produces a 200 response with an incomplete or truncated body, making a failed audit read indistinguishable from a successful one to clients. This defeats the integrity guarantee that the serialization fix was intended to provide.

**Suggestion:** Marshal all audit entries into a buffer before writing any headers, and route marshaling failures through the normal error handler so that a failure produces a proper 5xx response instead of a 200 with a corrupt body.

internal/adapter/http/handler.go
```diff
@@ -115,6 +116,18 @@ func NewHandler(logger logging.Logger, backend Backend, authCfg internalauth.Aut
...
+			// touch several ledgers), so these routes are NOT under {ledgerName};
+			// ledger scope is expressed via the `filter` query parameter. They live
+			// under the reserved `/v3/_/…` system segment so they cannot collide
+			// with `/v3/{ledgerName}` routes: the ledger name `_` is reserved at
+			// admission (admission.ErrLedgerNameReservedPrefix), keeping the system
+			// namespace `/v3/_/…` and the ledger namespace disjoint by construction.
+			r.With(requireAuditRead).Group(func(r chi.Router) {
+				r.Get("/_/audit-entries", server.handleListAuditEntries)
+				r.Get("/_/audit-entries/{sequence}", server.handleGetAuditEntry)
+			})
```
NumaryBot, line 133: 🟠 **[major] Audit routes shadow pre-existing ledgers named `audit-entries` on upgrade**
<sub>reported by flemzord, gfyrag, codex</sub>

Reserving the name `audit-entries` only at ledger-creation admission does not protect clusters that already have a ledger with that name before this upgrade. The router still registers the static `/v3/audit-entries` routes before `/{ledgerName}`, so on upgraded deployments `GET /v3/audit-entries` and `GET /v3/audit-entries/accounts` will permanently match the audit routes instead of the ledger routes, breaking existing clients. The create-time guard prevents future collisions but cannot remediate pre-existing ones.

**Suggestion:** Move the audit routes outside the ledger-name namespace entirely (e.g. `/v3/_audit/entries`), or add an explicit migration/compatibility path with tests covering the pre-existing ledger upgrade scenario.

internal/proto/auditpb/audit.pb.json.go
```diff
@@ -0,0 +1,154 @@
...
```
NumaryBot, line 98: 🟠 **[major] HTTP 200 already committed when audit submessage `MarshalJSON` fails**

Now that `MarshalJSON` correctly propagates errors from `protojson.Marshal`, the error is surfaced too late: the HTTP response writer has already sent the 200 status header before encoding begins. A marshaling failure on any audit submessage (`callerSnapshot`, `idempotency`, `signature`) therefore produces a 200 response with an incomplete or truncated body. A client cannot distinguish this from a successful response, defeating the integrity guarantee the serialization fix was intended to provide.

**Suggestion:** Marshal all audit entries into an in-memory buffer before writing any response headers. Route marshaling failures through the normal error handler so a failure produces a proper 5xx response with no partial body written.

internal/proto/auditpb/audit.pb.json.go
```diff
@@ -0,0 +1,154 @@
...
+		aux.Hash = hex.EncodeToString(h)
+	}
+
+	return json.Marshal(aux)
+}
+
+// MarshalJSON implements json.Marshaler for AuditItem.
+func (x *AuditItem) MarshalJSON() ([]byte, error) {
+	type Aux struct {
+		OrderIndex      uint32 `json:"orderIndex,omitempty"`
```
NumaryBot: 🟠 **[major] `omitempty` on numeric audit item fields drops meaningful zero values**

For the first item in every audit entry `orderIndex` is `0`, and for failed proposals or idempotent replays `logSequence` is explicitly documented as `0`. The `omitempty` JSON tags on these numeric fields cause them to be omitted from the HTTP response when they are zero. This makes it impossible for clients to determine the first order's position or interpret the no-log sentinel, stripping required per-order evidence from the audit record.

**Suggestion:** Remove `omitempty` from `orderIndex`, `logSequence`, and any other numeric audit item fields that carry meaning when zero, so they are always serialized.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1562#issuecomment-4934913701

internal/adapter/http/handler.go
```diff
@@ -115,6 +116,18 @@ func NewHandler(logger logging.Logger, backend Backend, authCfg internalauth.Aut
 				r.Get("/{ledgerName}/transactions/{transactionId}", server.handleGetTransaction)
 			})
 
+			// Audit read scope. Audit reads are cluster/bucket-wide (a proposal can
+			// touch several ledgers), so these routes are NOT under {ledgerName};
+			// ledger scope is expressed via the `filter` query parameter. They live
+			// under the reserved `/v3/_/…` system segment so they cannot collide
+			// with `/v3/{ledgerName}` routes: the ledger name `_` is reserved at
+			// admission (admission.ErrLedgerNameReservedPrefix), keeping the system
```
flemzord, line 128: 🟠 **[major] The `_` system namespace is not actually reserved**

This comment references `admission.ErrLedgerNameReservedPrefix`, but that sentinel does not exist on this head. `domain.ValidateLedgerName("_")` succeeds because `_` is part of the allowed identifier charset, and this commit removes the only reserved-name admission check and its tests. A ledger named `_` can therefore still be created, so `/v3/_/...` is not disjoint from the ledger namespace as claimed and future system routes can shadow that ledger's endpoints.

Please enforce the `_` reservation at the canonical ledger-name validation/admission boundary (with HTTP/gRPC/bulk coverage), or choose a system route that does not rely on this missing invariant. Also account for already-persisted `_` ledgers when defining the compatibility behavior.

flemzord (commented):
Found one new actionable issue in the namespace move.

gfyrag:
Re-reviewed current head `790b861e56c91b08da693f05dc2a0ea006042a24`.

Remaining findings:

1. Major: `/v3/_` is documented as a reserved system namespace, but admission no longer enforces that.

The router/OpenAPI claim ledger `_` is rejected, while admission only calls `ValidateLedgerName`; `github.com/formancehq/invariants v0.11.0` allows `_`. Either enforce the `_` reservation with coverage, including existing `_` ledger compatibility, or remove the claim and choose a route that does not rely on it.

2. Major: HTTP audit list still lacks gRPC read consistency parity while docs claim parity.

HTTP forwards only `pageSize`, `after`, `filter`, and `reverse`; gRPC handles `checkpointId` and `minLogSequence`. OpenAPI and HTTP API docs still say the consistency guarantee matches gRPC.

3. Major: audit response marshaling failures still commit HTTP 200.

The response writer writes status before `json.MarshalWrite`; if `AuditEntry.MarshalJSON` fails, the handler can only append an error body to an already-200 response. Buffer/marshal before `WriteHeader` for audit responses, or make the write path return errors before committing headers.

4. Major: `omitempty` still drops meaningful zero audit item fields.

`orderIndex:0` and `logSequence:0` are omitted, but both are valid audit evidence. `buildAuditItems` explicitly creates zero `LogSequence` items for failure/no-log cases.

What looks resolved since the prior pass:

- Old `/v3/audit-entries` route shadowing for pre-existing `audit-entries` ledgers.
- Semantically invalid audit filters returning 400.
- Empty list responses as `data: []`.
- Codec-level propagation of audit submessage marshal errors.

Checks run:

- `git diff --check e0ad00216..790b861e`
- `GOROOT= go test ./internal/adapter/http ./internal/application/admission ./internal/proto/auditpb`
- `GOROOT= go test ./internal/adapter/grpc -run 'TestListAuditEntries|TestValidateListOptions|Test.*Checkpoint|Test.*MinLog'`
- `GOROOT= go build ./...`
- `nix develop --command bash -c "just pre-commit"` failed during `go generate` with Go toolchain mismatch (`go1.26.3` vs `go1.26.4`); generated proto files were restored and the final worktree is clean.

flemzord:
The PR description is now stale relative to `790b861e5`. Its endpoint table and routing explanation still advertise top-level `GET /v3/audit-entries[/{sequence}]` and the old `audit-entries` reserved-name approach, while the current head exposes `GET /v3/_/audit-entries[/{sequence}]` and removed that per-word admission guard. Please update the public contract and rationale in the description so reviewers and downstream users do not implement against routes that no longer exist.

internal/adapter/http/handler.go
gfyrag: Fixed in 009aac319. You were right — the sentinel was referenced but did not exist on this branch, so `_` was not actually reserved (invariants v0.11.0 allows `_` in the ledger-name charset). The reservation lived only in the still-open EN-1472 PR (#1531), which this PR was implicitly depending on for merge order — a real hazard.

This PR now carries the reservation on its own tree, since it is the PR that introduces the `/v3/_/` routes that need it:
- `ErrLedgerNameReservedPrefix` sentinel added to `internal/application/admission/errors.go`.
- The `_` check is enforced in `validateOrderLedgerName` — the single admission validation gate for gRPC/HTTP/bulk — on **every** ledger-scoped order, not just CreateLedger. So `_` can neither be created nor written to.

On the pre-existing-`_`-ledger-on-upgrade concern: because the guard rejects all ledger-scoped orders (not just create), a `_` ledger persisted before this reservation becomes permanently non-writable and its routes stay shadowed. We accept that edge rather than build a migration — `_` was never a supported/system name and a `_` ledger in the wild is exceptionally unlikely. Covered by `TestValidateOrder_LedgerName` (`_` rejected on CreateLedger and on Apply; `_internal` still allowed).

Coordination: #1531 (EN-1472) carries the identical guard for its bucket-wide routes. Whichever lands first owns the code; the other resolves the trivial conflict.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1562#issuecomment-4934913701

internal/application/admission/validate_order.go
```diff
@@ -51,7 +51,21 @@ func validateOrderLedgerName(order *raftcmdpb.Order) domain.Describable {
...
+		return err
+	}
+
+	// Reserve the ledger name "_" for the system / non-ledger HTTP routes, which
+	// all live under /v3/_/… so they never shadow a real ledger (see
+	// ErrLedgerNameReservedPrefix and internal/adapter/http/handler.go). Applied
+	// to every ledger-scoped order, not just CreateLedger: a "_" ledger can
+	// never legitimately exist, so rejecting it everywhere is safe and keeps the
+	// rule in one place.
+	if ls.GetLedger() == "_" {
```
NumaryBot, line 69: 🔴 **[blocker] Admission check rejects all writes to pre-existing `_` ledgers on upgrade**
<sub>reported by codex, flemzord</sub>

The new admission guard rejects every ledger-scoped order for any ledger named `_`, not just future ledger creation. On upgraded clusters where a `_` ledger already exists (previously allowed by `domain.ValidateLedgerName`), all writes — transactions, metadata updates, etc. — are permanently broken after upgrade. This is a silent, unrecoverable breaking change for existing deployments.

**Suggestion:** Scope the `_` reservation to ledger-creation admission only (i.e. `CreateLedger` requests), mirroring how the earlier `audit-entries` guard was scoped. Do not reject writes to existing ledgers. Provide a migration/compatibility path and tests for the pre-existing `_` ledger scenario.

internal/adapter/http/handler.go
NumaryBot, line 128: 🟠 **[major] `_` system namespace is not actually enforced — ledger named `_` can still be created**

The comment and routing rationale claim `/v3/_/...` is a reserved system namespace, but `domain.ValidateLedgerName("_")` succeeds and the admission check that was supposed to guard this no longer exists on the current head. A `_` ledger can therefore still be created, making the audit routes non-disjoint from the ledger namespace and leaving future system routes vulnerable to collisions.

**Suggestion:** Enforce the `_` reservation at the canonical ledger-name validation or admission boundary (covering HTTP, gRPC, and bulk paths). Add tests for the HTTP 400/gRPC InvalidArgument response when attempting to create a `_` ledger. Also specify the compatibility behavior for already-persisted `_` ledgers.

internal/proto/auditpb/audit.pb.json.go
NumaryBot, line 98: 🟠 **[major] HTTP 200 already committed when audit submessage `MarshalJSON` fails**
<sub>reported by gfyrag, flemzord</sub>

The HTTP response writer sends the 200 status header before encoding begins. Even though `MarshalJSON` now propagates errors from `protojson.Marshal`, a marshaling failure on any audit submessage (`callerSnapshot`, `idempotency`, `signature`) results in a 200 response with an incomplete or truncated body. Clients cannot distinguish this from a successful response, defeating the integrity guarantee the serialization fix was intended to provide.

**Suggestion:** Marshal all audit entries into an in-memory buffer before writing any response headers. Route marshaling failures through the normal error handler so a failure produces a proper 5xx response with no partial body written.

internal/proto/auditpb/audit.pb.json.go
NumaryBot: 🟠 **[major] `omitempty` on numeric audit item fields silently drops meaningful zero values**
<sub>reported by gfyrag, flemzord</sub>

For the first item in every audit entry, `orderIndex` is `0`. For failed proposals or idempotent replays, `logSequence` is explicitly documented as the no-log sentinel `0`. The `omitempty` JSON tags on these numeric fields cause them to be omitted from HTTP responses when zero, making it impossible for clients to determine the first order's position or interpret the no-log sentinel.

**Suggestion:** Remove `omitempty` from `orderIndex`, `logSequence`, and any other numeric audit item fields that carry meaning when zero, so they are always serialized in HTTP responses.

internal/adapter/http/handler.go
gfyrag: Addressed in 009aac319 — this is the same finding as the flemzord thread on this line, which is now resolved. The `_` reservation is enforced at the canonical admission boundary (`validateOrderLedgerName`, the single validation gate for gRPC/HTTP/bulk) via the new `ErrLedgerNameReservedPrefix` sentinel, on every ledger-scoped order (not just CreateLedger). So `domain.ValidateLedgerName("_")` still succeeds at the charset layer, but admission now rejects `_` before any order enters Raft — create and write alike. `/v3/_/…` is disjoint from `/v3/{ledgerName}` by construction on this branch alone.

Compatibility for already-persisted `_` ledgers: the all-orders guard makes such a ledger permanently non-writable and its routes shadowed. Accepted fail-closed edge (no migration) — `_` was never a supported name and a `_` ledger in the wild is exceptionally unlikely. Covered by `TestValidateOrder_LedgerName` (`_` rejected on CreateLedger and Apply; `_internal` allowed).

internal/adapter/http/handler.go
gfyrag: Addressed — the audit routes were moved out of the `/{ledgerName}` namespace entirely (commit 790b861e5), so this collision no longer exists. They now live under `/v3/_/audit-entries[/{sequence}]`, and the ledger name `_` is reserved at admission (`ErrLedgerNameReservedPrefix` in `validateOrderLedgerName`, rejecting every ledger-scoped order for `_`, added in 009aac319). So `/v3/_/…` is disjoint from `/v3/{ledgerName}/…` by construction — a ledger named `audit-entries` is fully reachable again and no longer shadowed. Resolving as superseded.

internal/adapter/http/handlers_list_audit_entries.go
gfyrag: Addressed at head — `drainCursor` now initializes `items := []T{}` (not `var items []T`), so an empty audit list serializes as `{"data":[]}`, not `{"data":null}`. Asserted by `TestHandleListAuditEntries_Empty` (`require.JSONEq(t, "{\"data\":[]}", ...)`). The same non-nil init fixed the latent null-data bug in the other list handlers that share `drainCursor`. Resolving.

internal/proto/auditpb/audit.pb.json.go
gfyrag: Addressed at head — `protoFieldJSON` returns `(json.RawValue, error)` and `AuditEntry.MarshalJSON` propagates it with context (`audit entry %d: marshaling callerSnapshot: %w`), so a protojson failure on a chain-bound submessage fails the whole marshal instead of emitting a valid-looking record with the field silently dropped (invariant #7). The typed-nil guard is retained so genuinely-absent submessages are omitted (not rendered as `{}`). Covered by `TestAuditEntry_MarshalJSON_PropagatesSubmessageError`.

Follow-up (6b6409232): a propagated marshal error was still surfacing too late at the HTTP layer (200 header already sent) — that is now fixed too via `writeOKChecked`, which buffers before writing any header. Resolving this thread as addressed.

internal/application/admission/validate_order.go
gfyrag: This is a deliberate, reviewed design decision, not an oversight — the all-orders scope is intentional.

The `_` segment backs the entire `/v3/_/…` system route namespace; it was never a supported ledger name, and the reservation exists precisely so that namespace stays disjoint from `/v3/{ledgerName}`. We considered scoping the guard to `CreateLedger` only (which would leave a pre-existing `_` ledger writable/deletable via gRPC, HTTP-shadowed only) versus rejecting all ledger-scoped orders (full fail-closed freeze). The owner (gfyrag) chose the strict all-orders freeze:

- A `_` ledger in the wild is exceptionally unlikely — it would require a client to have deliberately created a ledger with the single-underscore name before this reservation existed.
- Fail-closed is the safer posture for a system-namespace reservation: we would rather a hypothetical pre-existing `_` ledger be inert than allow writes to a name the system now claims.
- No migration is built, by design — the accepted edge is that such a ledger becomes non-writable and its routes stay shadowed.

This is coordinated with the identical guard in #1531 (EN-1472). Resolving as an intentional design choice; the create-only alternative was explicitly evaluated and declined.

internal/proto/auditpb/audit.pb.json.go
gfyrag: Fixed in 6b6409232. You're right — propagating the marshal error from `MarshalJSON` was necessary but not sufficient, because `writeJSONResponse` writes the 200 header before streaming the body, so a protojson failure on a chain-bound submessage still produced a truncated 200.

Added `writeOKChecked` (response.go): it marshals the whole `BaseResponse` to a buffer FIRST, and only writes the 200 header + body on success; a marshal error is routed through `handleError` before any header is sent, producing a clean 500 (`INTERNAL_ERROR` + correlation ID, raw error logged server-side) with no partial body. Both audit handlers (`handleListAuditEntries`, `handleGetAuditEntry`) now use it.

Scoped to the audit surface rather than changing the shared `writeJSONResponse` for all ~56 endpoints: audit is the only endpoint whose `MarshalJSON` can genuinely fail (it renders chain-bound submessages via protojson); the other list/get handlers marshal plain structs that cannot fail, so they keep the streaming path and avoid the extra buffer allocation. Covered by `TestHandleListAuditEntries_MarshalFailureIsClean500` (invalid-UTF8 callerSnapshot -> 500, error envelope, no `data` payload). Flagged the shared-helper latent case (its fallback appends error JSON into an already-sent body) to the team separately as a possible follow-up.

internal/proto/auditpb/audit.pb.json.go
gfyrag: Fixed in 6b6409232. Confirmed against the proto: `order_index` is zero-based (the first item of every entry is 0) and `log_sequence` is documented as the 0 sentinel for an idempotent replay or failed proposal — both meaningful at zero, both dropped by `omitempty`.

Removed `omitempty` from `orderIndex` and `logSequence`, and — same class, so fixed together to avoid a half-fix — from `AuditSuccess.minLogSequence`/`maxLogSequence` (the proto documents min==max==0 for an all-idempotent success) and from the always-present scalar identity/counter fields on `AuditEntry` (`sequence`, `proposalId`, `orderCount`, `hashVersion`; e.g. the first audit entry legitimately has sequence 0). `omitempty` is kept only for genuinely-absent fields: the success/failure oneof, timestamp, the optional submessages (callerSnapshot/idempotency/signature), and the empty-on-list slices (items/ledgers) + hash/serializedOrder.

Covered by `TestAuditEntry_MarshalJSON_PreservesMeaningfulZeros` (every zero-valued numeric field present in the output). Left the openapi schema as-is (the fields are already documented as present properties); did not add them to `required` to avoid a broader schema-contract change beyond the finding.

internal/adapter/http/handlers_list_audit_entries.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	filter, ok := parseAuditFilter(w, r)
+	if !ok {
+		return
+	}
+
+	reverse := queryParamBool(r, "reverse")
+
+	cursor, err := s.backend.ListAuditEntries(r.Context(), pageSize, afterSequence, filter, reverse)
```
NumaryBot, line 72: 🟠 **[major] HTTP audit list lacks gRPC read-consistency parity while docs claim parity**

The HTTP handler parses only `pageSize`, `after`, `filter`, and `reverse`, then calls the controller directly. The gRPC `ListAuditEntries` path honours `checkpointId` and `minLogSequence`, including checkpoint reads and filtered-audit-index consistency waits. The OpenAPI spec and docs still assert that the HTTP consistency guarantee matches gRPC, which is false and will lead clients to observe stale or missing audit entries in scenarios where those options are required.

**Suggestion:** Add `checkpointId` and `minLogSequence` query parameters and wire them through to the controller the same way the gRPC path does, or explicitly reject/document these options as unsupported and remove the parity claim from the OpenAPI/docs.

internal/proto/auditpb/audit.pb.json.go
```diff
@@ -0,0 +1,173 @@
...
+		Failure:     x.GetFailure(),
+		OrderCount:  x.GetOrderCount(),
+		Items:       x.GetItems(),
+		Ledgers:     x.GetLedgers(),
+		HashVersion: x.GetHashVersion(),
+	}
+
+	var err error
+
+	if aux.CallerSnapshot, err = protoFieldJSON(x.GetCallerSnapshot()); err != nil {
```
NumaryBot, line 98: 🟠 **[major] HTTP 200 already committed when audit submessage MarshalJSON fails**

The HTTP response writer sends the 200 status header before JSON encoding begins. Although `MarshalJSON` now correctly propagates `protojson.Marshal` errors for submessages (`callerSnapshot`, `idempotency`, `signature`), a failure occurs after the status is committed, producing a 200 response with an incomplete or truncated body. Clients cannot distinguish this from a successful response, defeating the integrity guarantee the serialization fix was intended to provide.

**Suggestion:** Marshal all audit entries into an in-memory buffer before writing any response headers. Route marshaling failures through the normal error handler so a failure produces a proper 5xx response with no partial body written.

internal/proto/auditpb/audit.pb.json.go
```diff
@@ -0,0 +1,173 @@
...
+		return nil, fmt.Errorf("audit entry %d: marshaling idempotency: %w", x.GetSequence(), err)
+	}
+
+	if aux.Signature, err = protoFieldJSON(x.GetSignature()); err != nil {
+		return nil, fmt.Errorf("audit entry %d: marshaling signature: %w", x.GetSequence(), err)
+	}
+
+	if h := x.GetHash(); len(h) > 0 {
+		aux.Hash = hex.EncodeToString(h)
+	}
```
NumaryBot, line 112: 🟠 **[major] `omitempty` on numeric audit item fields silently drops meaningful zero values**

`orderIndex` is `0` for the first item in every audit entry, and `logSequence` is explicitly documented as `0` for failed proposals or idempotent replays. The `omitempty` JSON tags on these numeric fields cause them to be omitted from the HTTP response when their value is zero, making it impossible for clients to determine the first order's position or interpret the no-log sentinel, stripping required per-order evidence from the audit record.

**Suggestion:** Remove `omitempty` from `orderIndex`, `logSequence`, and any other numeric audit item fields that carry meaning when zero so they are always serialized.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1562#issuecomment-4934913701

internal/adapter/http/handlers_list_audit_entries.go
gfyrag: Legitimate — fixed in 82a8cface (by documenting the gap, not by adding the options). You're right that gRPC honors `checkpointId` (pinned checkpoint read) and `minLogSequence` (audit-index catch-up wait) and this HTTP endpoint honored neither, while the wording implied parity.

Building checkpoint-read + consistency-wait plumbing into the HTTP surface is out of scope for a read-exposure ticket (it's a heavier feature, and the checkpoint path even carries a known audit-indexer-lag caveat on the gRPC side). So I took the alternative your suggestion explicitly allows — remove the parity claim and document the endpoint's real behavior:
- Handler comment (handlers_list_audit_entries.go): now states the endpoint always does a live, best-effort read, does NOT expose checkpointId/minLogSequence, and that a filtered live read may transiently omit very recent unindexed entries; a pinned/consistency-bounded read must use gRPC. Left a pointer to the gRPC entry points (openCheckpointStores / minLogSequence gating) for whoever adds them later.
- openapi.yml: the description no longer says "the consistency guarantee is the same as gRPC"; it now spells out the live/best-effort behavior and the missing options.
- PR description updated to match.

If you'd rather we actually expose checkpointId/minLogSequence over HTTP in this PR, that's a scope call — flag it and I'll raise it with the owner. Resolving on the documented-gap fix.

internal/proto/auditpb/audit.pb.json.go
gfyrag: Already fixed at head (6b6409232) — this looks like a re-scan of the diff that didn't pick up the fix. `writeJSONResponse` streaming was indeed the problem; the audit handlers no longer use it. They now call `writeOKChecked` (internal/adapter/http/response.go), which marshals the whole BaseResponse to a buffer BEFORE writing any header and routes a marshal failure through `handleError` — a clean 500 (INTERNAL_ERROR + correlation ID), never a truncated 200. Both handleListAuditEntries and handleGetAuditEntry use it. Covered by TestHandleListAuditEntries_MarshalFailureIsClean500 (invalid-UTF8 callerSnapshot -> 500, error envelope, no `data`). Resolving as already addressed.

internal/proto/auditpb/audit.pb.json.go
gfyrag: Already fixed at head (6b6409232) — re-scan duplicate. omitempty was removed from orderIndex and logSequence (and, same class, from AuditSuccess minLogSequence/maxLogSequence and the always-present entry scalars sequence/proposalId/orderCount/hashVersion), so meaningful zeros are always serialized. omitempty is kept only for genuinely-absent fields (oneof, timestamp, optional submessages, empty slices). Covered by TestAuditEntry_MarshalJSON_PreservesMeaningfulZeros. Resolving as already addressed.

gfyrag:
Fixed — updated the PR description to match head (82a8cface). The endpoint table and routing rationale now show `GET /v3/_/audit-entries[/{sequence}]` under the reserved `/v3/_/…` system segment, and explain the current mechanism: the ledger name `_` is reserved at admission (`ErrLedgerNameReservedPrefix`, enforced on every ledger-scoped order) so it can never shadow the system routes. The stale top-level `/v3/audit-entries` + per-word `audit-entries` reserved-name approach is now described as the dropped earlier design. Also refreshed the consistency note (live/best-effort, no checkpointId/minLogSequence) and the test list. Thanks for catching it.

docs/technical/architecture/subsystems/api/http-api.md
```diff
@@ -347,6 +347,44 @@ DELETE /v3/{ledgerName}/accounts/{address}/metadata/{key}
...
+Query parameters:
+- `pageSize`: max entries (default 100, capped at 1000)
+- `after`: audit sequence to start after (exclusive, opaque cursor)
+- `reverse`: `true` iterates newest-first
+- `filter`: a filter expression restricted to `audit[...]` conditions, using the
+  same grammar as `ledgerctl audit list --filter` (e.g.
+  `audit[outcome] == failure`, `audit[ledger] == main`,
+  `audit[order_type] in (create_transaction, revert_transaction)`). This is the
+  shared `filterexpr` DSL — **not** the JSON `QueryFilter` DSL used by prepared
+  queries, which cannot represent audit conditions. Filtered reads are served by
```
NumaryBot, line 373: 🟠 **[major] HTTP audit endpoint promises gRPC-level read-consistency it cannot deliver**
<sub>reported by codex, gfyrag, NumaryBot</sub>

The HTTP audit list handler forwards only `pageSize`, `after`, `filter`, and `reverse` to `ListAuditEntries`. It does not expose gRPC's `checkpointId` or `minLogSequence` consistency gates, so the async audit index can transiently omit recent entries. The documentation states the consistency guarantee matches gRPC, which misleads callers into relying on guarantees the endpoint does not provide.

**Suggestion:** Either add `checkpointId` and `minLogSequence` query parameters and wire them through to the controller (matching gRPC behaviour), or update the documentation to clearly state this is a best-effort, eventually-consistent read and that callers requiring strong consistency must use the gRPC surface.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1562#issuecomment-4934913701

internal/application/admission/validate_order.go
flemzord: Revalidated on `82a8cfa`: this remains a compatibility blocker. Before this PR, `domain.ValidateLedgerName("_")` and admission accepted the name, so an upgraded store can legitimately contain that ledger. The new guard runs for every `LedgerScopedOrder`, which makes such a ledger unable to transact, update metadata, promote, or even follow a normal deletion/recovery path. Calling the case intentional or unlikely does not make the upgrade non-breaking; the new Apply test currently pins the data-stranding behavior instead of proving compatibility. Please reserve `_` for future `CreateLedger` requests only, or provide an explicit migration and recovery path for already-persisted `_` ledgers before freezing all operations.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1562#issuecomment-4934913701

internal/application/admission/validate_order.go
NumaryBot, line 69: 🟠 **[major] Existing '_' ledgers permanently broken after upgrade**

The new `ErrLedgerNameReservedPrefix` admission check rejects every ledger-scoped order for any ledger named `_`. Because `domain.ValidateLedgerName` previously allowed this name, clusters that already have a `_` ledger will have all writes and deletes permanently rejected after upgrading. The audit routes only occupy `/v3/_/audit-entries[...]`, so this is a pure compatibility break for pre-existing ledgers with that name. The `_` reservation enforcement mechanism has also been flagged as potentially un-enforced in the current head (invariants v0.11.0 allows `_`), and there is no migration or compatibility path.

**Suggestion:** Add a migration/compatibility path for pre-existing `_` ledgers (e.g. reject only at ledger creation, not at every operation), or verify and document that the `_` name has never been valid so no such ledgers exist in the wild. Also add integration coverage for the upgrade scenario.

docs/technical/architecture/subsystems/api/http-api.md
NumaryBot, line 373: 🟠 **[major] HTTP audit API docs falsely claim gRPC read-consistency parity**
<sub>reported by codex, gfyrag, NumaryBot</sub>

The HTTP handler only forwards `pageSize`, `after`, `filter`, and `reverse` to `ListAuditEntries`. The gRPC path additionally honors `checkpointId` and `minLogSequence` for checkpoint reads and filtered-audit-index consistency waits. The documentation (and previously the OpenAPI spec) still states that the HTTP API provides the same consistency guarantees as gRPC. Clients following this documentation may rely on an audit-index catch-up guarantee that the HTTP endpoint silently does not provide, with no error or indication of the missing guarantee.

**Suggestion:** Update the documentation (and OpenAPI spec if applicable) to clearly state that HTTP audit reads are live/best-effort and that `checkpointId`/`minLogSequence` are not supported. Alternatively, implement the missing read options in the HTTP handler.

flemzord (approved):
Re-reviewed current head d28f5fcba6c66121c70349d864695d17ad468097. No High or Critical findings remain under the requested threshold. The pre-existing ledger named exactly `_` compatibility break is real, but classified Medium because it is limited to that narrow legacy-name case; the remaining consistency/documentation gaps are also non-blocking.

PR #1563, opened 2026-07-10
feat(EN-1503): allow LOGS as a prepared-query target over REST

gfyrag:
## Summary

Makes **LOGS** a usable prepared-query target over REST, at parity with the direct `ListLogs` endpoint. Closes [EN-1503](https://formance-team.atlassian.net/browse/EN-1503).

## Why LOGS returned an empty cursor before

`query.Execute` → `executeList` only populated `PreparedQueryCursor` for the `ACCOUNTS`/`TRANSACTIONS` targets, and `PreparedQueryCursor` had **no log result field**. A LOGS prepared query therefore compiled and paginated the log iterator correctly but then dropped every entity because the `switch target` had no LOGS branch — an execution-wiring gap, not a fundamental limit (LOGS is a first-class target in `compile.go` / `list_entities.go`, and direct `ListLogs` works). Because of that gap, REST rejected `LOGS` at ingestion and blanket-rejected the log-only conditions `logId`/`date`/`ledger`.

## How it's wired now (steps 1–2, the core value)

- **Proto**: add `PreparedQueryCursor.log_data` (`repeated Log`) + a `logData` field on the hand-written camelCase JSON marshaller; regenerate `commonpb`.
- **Execution**: wire the LOGS branch in `query.Execute` via a new `query.EnrichLogs` that reuses the direct `ListLogs` machinery — `ReadLedgerLogsCompiled` resolves the compiled per-ledger logIDs to global sequences through the log read-index and reads the log payloads from Pebble. The `switch target` now **fails loudly** on an unhandled target (invariant #7) instead of returning an empty cursor.
- **Target enablement**: `"LOGS"` added to `preparedQueryTargets` / `parsePreparedQueryTarget`.

## Target-aware validation (step 3)

Per-target condition validity is driven by a **single generated source of truth** shared with EN-1504: the `allowed_query_targets` annotations on the `QueryFilter.filter` oneof arms in `common.proto` generate `commonpb.targetConditionValidity`. Both the write path (`domain.ValidateFilterForTarget`, run at admission **and** in the FSM) and the execute path (`query.Compile` → `rejectInvalidCondition`) consult this same table, so the two layers cannot drift.

For the LOGS target specifically:

- **Valid**: `logId`, `date` (`logBuiltinUint`), `ledger`, plus the structural combinators (`$and`/`$or`/`$not`).
- **Rejected**: `address`, `reference`, `builtin_uint`, `accountHasAsset`, `reverted` (no account/transaction→log translation), **and `metadata`** — there is no populated log-metadata index (the index builder never fills the `NamespaceLog` metadata keyspace `compileFieldCondition` scans), so advertising metadata filters for LOGS would return empty/error at execute time. Building a real log-metadata index is a separate feature (out of scope, tracked separately).
- **`ledger` condition** is evaluated as an exact match against the executing ledger: naming the executing ledger returns the full log universe; naming any other ledger returns an empty (unsatisfiable) result rather than silently returning all of the executing ledger's logs.

The **update** handler carries only the new filter (the target is immutable and not read by the handler); the FSM's `processUpdatePreparedQuery` reads the stored query and runs `ValidateFilterForTarget(filter, storedTarget)` before persisting, so an update can no longer smuggle in a condition invalid for the query's target.

**Filter recursion is depth-bounded on both paths.** `domain.MaxFilterDepth` (100) is the single source of truth, re-exported by `query.Compile`. `ValidateFilterForTarget` enforces it at write time (admission + FSM) with the **same node-for-node counting** as `Compile` (checked on entry of every node, combinators and leaf), so a filter accepted at write time is guaranteed to compile — and a hostile deeply-nested gRPC filter is rejected before it can overflow the stack (a fatal DoS that `Compile`'s guard alone could not prevent, since the write walk runs first).

## Tests

- Unit: target-aware `decodePreparedQueryFilter` / `ValidateFilterForTarget` (accepts log-only for LOGS, rejects `address`/`reference`/`metadata` for LOGS, rejects log-only for ACCOUNTS/TRANSACTIONS), `parsePreparedQueryTarget` incl. LOGS, create-handler accepts LOGS + preserves the target/filter.
- Depth guard: over-limit `$and`/`$or`/`$not` rejected at write time; `TestCompile_DepthBoundMatchesValidator` pins the write- and execute-time guards to the same boundary so future drift fails loudly.
- Ledger condition: `TestCompile_LedgerConditionOtherLedgerIsEmpty` / `_MissingValue`.
- Validity matrix: the human-readable contract test (`common_queryfilter_validity_test.go`) pins the full (target × condition) matrix, including `metadata`=false on LOGS.
- e2e (gRPC): LOGS execute with a `ledger` filter (asserts non-empty `logData`, ascending order) and a `logId`-range filter.
- e2e (REST): create → execute a LOGS prepared query, asserting `cursor.logData` is populated (the pre-fix empty-cursor regression); both body reads assert their `io.ReadAll` error.

## Docs

`openapi.yml` (target enums + QueryFilter log-only-field note), architecture doc, and the `actions.LogIdGreaterThanFilter` helper (parameter documented as the ledger-local `LedgerLog.Id`, not the global sequence).

## ⚠️ Coordination with EN-1504

Built on top of **EN-1504** (harmonize per-target `QueryFilter` validation into a single generated source of truth). The LOGS execution wiring + target enablement here are **not** in EN-1504's scope; the target-aware validation reuses EN-1504's generated validity table.

## Verification

- `GOROOT= go build ./...` ✅
- `go test ./internal/adapter/http/... ./internal/domain/... ./internal/query/... ./internal/proto/commonpb/... ./pkg/actions/...` ✅
- `gofmt` clean, `golangci-lint` clean on touched packages
- e2e compiles under the `e2e` tag

## Invariants

Read-side only — no FSM/cache/preload changes, no new Pebble writes; the audit/log source-of-truth is untouched. No silent skips (invariant #7): the enrichment switch and both filter-depth guards hard-fail rather than degrade silently.

[EN-1503]: https://formance-team.atlassian.net/browse/EN-1503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/adapter/http/handlers_update_prepared_query.go
```diff
@@ -33,7 +34,14 @@ func (s *Server) handleUpdatePreparedQuery(w http.ResponseWriter, r *http.Reques
...
 
-	filter, err := decodePreparedQueryFilter(body.Filter)
+	// The update request carries only the new filter, not the target: the target
+	// is immutable and lives on the stored prepared query, which this handler
+	// does not read. We therefore cannot apply the target-aware log-only reject
+	// here (doing so would wrongly reject valid LOGS-target updates). Log-only
+	// conditions on a non-LOGS target are still caught at execute time by the
+	// compile layer, which rejects logId/date/ledger conditions unless the
+	// target is LOGS. Pass QUERY_TARGET_LOGS to skip the early reject.
+	filter, err := decodePreparedQueryFilter(body.Filter, commonpb.QueryTarget_QUERY_TARGET_LOGS)
```
NumaryBot: 🟠 **[major] Preserve target validation on prepared-query updates**

When updating an existing ACCOUNTS or TRANSACTIONS prepared query with a `logId`, `date`, or `ledger` filter over REST, this always passes `QUERY_TARGET_LOGS`, so the decoder accepts and persists a filter that is invalid for the stored target. That makes the update return 204 and only fail later at execute time, unlike create and the documented target-aware validation; validate against the existing query target before applying the update.

tests/e2e/business/prepared_queries_test.go
```diff
@@ -704,6 +704,112 @@ var _ = Describe("PreparedQueries", Ordered, func() {
...
+		It("Should honor a logId range filter", func() {
+			// Find the ledger's log sequences to build a bounded logId range.
+			all, err := sharedClient.ExecutePreparedQuery(sharedCtx, &servicepb.ExecutePreparedQueryRequest{
+				Ledger:    ledgerName,
+				QueryName: "all-logs",
+				Mode:      commonpb.QueryMode_QUERY_MODE_LIST,
+			})
+			Expect(err).To(Succeed())
+			Expect(all.GetCursor().LogData).To(HaveLen(3))
+			firstSeq := all.GetCursor().LogData[0].GetSequence()
```
NumaryBot: 🟠 **[major] Use ledger-local log ID for logId filter**

This uses `Log.GetSequence()` as the lower bound for a `logId` filter, but `LogIdCondition` filters the ledger-local `Apply.Log.Id`, not the global audit/log sequence. As soon as the global sequence diverges from the ledger-local ids (for example after ledger creation or other ledgers' activity), `logId > firstSeq` skips too many rows and this e2e test will return fewer than the expected two logs.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1563#issuecomment-4934926395

internal/adapter/http/handlers_update_prepared_query.go
NumaryBot: 🟠 **[major] Update path forces LOGS target, bypassing target-aware validation**

When updating an existing ACCOUNTS or TRANSACTIONS prepared query via REST, the update handler always passes QUERY_TARGET_LOGS to the decoder. This causes the decoder to accept and persist filters (e.g. logId, date, ledger) that are invalid for the stored target. The update returns 204 but only fails at execute time, inconsistently with the documented target-aware validation on create.

**Suggestion:** Fetch the existing query's target before decoding the update payload and validate the incoming filters against the stored target, not a hardcoded LOGS target.

tests/e2e/business/prepared_queries_test.go
```diff
@@ -704,6 +710,117 @@ var _ = Describe("PreparedQueries", Ordered, func() {
...
+			for i, l := range result.GetCursor().LogData {
+				seqs[i] = l.GetSequence()
+			}
+			Expect(sort.SliceIsSorted(seqs, func(i, j int) bool { return seqs[i] < seqs[j] })).To(BeTrue())
+		})
+
+		It("Should honor a logId range filter", func() {
+			// The logId filter operates on the per-ledger LedgerLog.Id (1-based,
+			// scoped to this ledger), NOT the global Log.sequence — the log index
+			// is keyed [0x09][ledger][ledgerLogID]. Read the per-ledger ids from
```
NumaryBot, line 777: 🟠 **[major] logId filter uses global sequence instead of ledger-local log ID**

The e2e test uses Log.GetSequence() as the lower bound for a logId filter, but LogIdCondition filters on the ledger-local Apply.Log.Id, not the global audit/log sequence. Once the global sequence diverges from ledger-local IDs (e.g. after ledger creation or activity on other ledgers), the filter skips too many rows and fewer than the expected two logs are returned, causing a flaky or systematically failing test.

**Suggestion:** Replace Log.GetSequence() with the ledger-local log ID (Apply.Log.Id or equivalent) when constructing the logId filter bound in the e2e test.

internal/adapter/http/prepared_query_filter.go
```diff
@@ -59,28 +57,31 @@ func decodePreparedQueryFilter(raw json.RawMessage) (*commonpb.QueryFilter, erro
...
-	// (see parsePreparedQueryTarget), so log-only conditions (logId, date,
-	// ledger) can never be satisfied here — they would compile a log iterator
-	// and silently return unrelated/empty results. Reject them up front.
-	if err := rejectLogOnlyConditions(filter); err != nil {
-		return nil, err
+	// Log-only conditions (logId, date, ledger) are only meaningful when the
+	// prepared query targets LOGS. For ACCOUNTS/TRANSACTIONS they would compile
+	// against a non-log iterator and silently return unrelated/empty results, so
+	// reject them up front for those targets.
+	if target != commonpb.QueryTarget_QUERY_TARGET_LOGS {
```
NumaryBot: 🟠 **[major] Non-log filters accepted for LOGS queries produce semantically incorrect results**

When a LOGS prepared query is created with transaction/account filters such as reference or address, the branch allows those filters to persist. The existing compile paths for those filters yield transaction IDs or account keys; the LOGS enrichment layer then treats those bytes as ledger-local log IDs. This can cause the query to return unrelated logs (e.g. the query-creation log) instead of the intended transaction or account log.

**Suggestion:** Either reject filters that do not produce log IDs when the query target is LOGS, or add a dedicated LOGS-specific translation/compilation path that correctly maps each filter type to log IDs.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1563#issuecomment-4934926395

internal/query/executor.go
```diff
@@ -218,6 +220,19 @@ func executeList(
...
 
 		cursor.TransactionData = txns
+	case commonpb.QueryTarget_QUERY_TARGET_LOGS:
+		logs, err := EnrichLogs(reader, indexReader, ledgerName, entities)
+		if err != nil {
+			return nil, err
+		}
+
+		cursor.LogData = logs
+	default:
```
flemzord, line 230: [P2] This fallback is bypassed when the unsupported target produces no entities: `executeList` returns the empty cursor before reaching this switch. That is reachable because prepared-query admission validates only the name and `compileUniverse` returns an empty iterator for unknown enum values, so a gRPC-created/stored query can still silently succeed instead of surfacing the invariant violation. Please validate the target before the empty-page shortcut (or reject it in `Compile`) and add a regression test.

flemzord (commented):
One additional actionable issue found; existing open threads were not duplicated.

gfyrag (commented):
J'ai repris les commentaires NumaryBot sur le head courant `b90caf41a2b818358bd4f5e8e8459d8fde9e422c`.

Validation:
- Le finding update REST hardcodée en `QUERY_TARGET_LOGS` est toujours valide: `handleUpdatePreparedQuery` décode avec `LOGS` sans lire la target stockée, donc une query `ACCOUNTS`/`TRANSACTIONS` peut persister un filtre log-only et échouer seulement à l'exécution.
- Le finding LOGS + filtres non-log est toujours valide: `address`/`reference` peuvent compiler vers des IDs transaction/account, puis `ExecutePreparedQuery` hydrate ces bytes via `EnrichLogs` comme s'il s'agissait de log IDs.
- Le finding e2e `logId` semble corrigé au head courant: le test lit désormais `firstLogID := ledgerLogID(all.GetCursor().LogData[0])` et le commentaire précise bien que `logId` est le `LedgerLog.Id` local, pas `Log.sequence`.

Tests ciblés lancés localement:

```bash
go test ./internal/adapter/http ./internal/query -run 'Test.*Prepared|Test.*Log|Test.*Query|TestCompile' -count=1
```

Ils passent, mais ne couvrent pas les deux problèmes encore actifs.

internal/query/executor.go
gfyrag: Confirmed and fixed in c835b50b5 — good catch. The `executeList` default was indeed unreachable via the empty-page shortcut: `len(entities)==0` returns `emptyListResponse` before the target switch, and `compileUniverse`'s default arm returned an empty iterator for an unknown enum, so a gRPC-stored prepared query with an unsupported target succeeded silently with an empty page.

Fixed at the earliest choke point instead of the switch: `Compile` now rejects an unsupported target up front via an explicit allow-list (`isSupportedTarget` = ACCOUNTS/TRANSACTIONS/LOGS), returning an `ErrFilterCompilation` ("unsupported query target …"). This covers every prepared-query and list path (all route through `Compile`). `compileUniverse`'s default arm is hardened from a silent empty iterator to the same loud error (now unreachable-by-design), and `executeList`'s default is kept as defense in depth.

Regression tests added in `internal/query/compile_test.go`:
- `TestCompile_RejectsUnsupportedTarget` — nil (universe) and field-filter paths both fail loudly rather than returning an empty success.
- `TestIsSupportedTarget` — pins the allow-list so a newly added `QueryTarget` enum value is rejected until its iteration/enrichment paths are wired.

internal/adapter/http/handlers_update_prepared_query.go
NumaryBot: 🟠 **[major] Update handler hardcodes LOGS target, bypassing target-aware validation**

When updating an existing ACCOUNTS or TRANSACTIONS prepared query via REST, the update handler always passes QUERY_TARGET_LOGS to the decoder. This causes the decoder to accept and persist filters (e.g. logId, date, ledger) that are invalid for the stored target. The update returns 204 but only fails at execute time, inconsistently with the target-aware validation enforced on create.

**Suggestion:** Fetch the existing query's target before decoding the update payload and validate the incoming filters against the stored target rather than hardcoding QUERY_TARGET_LOGS.

internal/adapter/http/prepared_query_filter.go
NumaryBot: 🟠 **[major] Non-log filters accepted for LOGS queries produce semantically incorrect results**

When a LOGS prepared query is created with transaction/account filters such as reference or address, the branch skips only the log-only reject, allowing those filters to persist. The existing compile paths for those filters yield transaction IDs or account keys; the LOGS enrichment layer then treats those bytes as ledger-local log IDs. This can cause the query to return unrelated logs instead of the intended ones, or fail with an index-inconsistency error.

**Suggestion:** Either reject filters that do not produce log IDs when the query target is LOGS, or add a dedicated LOGS-specific translation/compilation path that correctly maps each filter type to log IDs.

tests/e2e/business/prepared_queries_test.go
NumaryBot, line 777: 🟠 **[major] logId filter uses global sequence instead of ledger-local log ID**

The e2e test uses Log.GetSequence() as the lower bound for a logId filter, but LogIdCondition filters on the ledger-local Apply.Log.Id, not the global audit/log sequence. Once the global sequence diverges from ledger-local IDs (e.g. after ledger creation or activity on other ledgers), the filter skips too many rows and fewer than the expected two logs are returned, making the test flaky or systematically failing.

**Suggestion:** Replace Log.GetSequence() with the ledger-local log ID (Apply.Log.Id or equivalent) when constructing the logId filter bound in the e2e test.

internal/query/executor.go
NumaryBot, line 230: 🟠 **[major] Unsupported target fallback is bypassed by empty-page shortcut**

The error path for an unsupported query target is never reached when executeList returns an empty cursor early (before the switch). A gRPC-created/stored query with an unknown target enum can therefore silently succeed (returning an empty page) instead of surfacing the invariant violation, because compileUniverse returns an empty iterator for unknown enum values and admission only validates the name.

**Suggestion:** Validate the target before the empty-page shortcut (or reject unknown targets in Compile) and add a regression test for this path.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1563#issuecomment-4934926395

gfyrag:
Review pass on `c835b50b552cc44eb58cdfffdb2c8e91ee1a92b3`.

Findings:

1. Major: REST update still decodes every incoming prepared-query filter as `QUERY_TARGET_LOGS`. Updating an existing ACCOUNTS or TRANSACTIONS prepared query with `logId`, `date`, or `ledger` returns 204 and persists an invalid filter, then fails only at execution. The update handler needs to validate against the stored prepared-query target before applying the update.

2. Major: LOGS still accepts non-log filters such as `reference` and `address`. The compiler routes `reference` through transaction IDs, and `address` falls through to the transaction translation path for non-ACCOUNTS targets; `EnrichLogs` then treats those bytes as ledger-local log IDs. Please reject account/transaction-only filters for LOGS, or add a real LOGS-specific translation from those predicates to log IDs.

Outdated after the rebase: the unsupported-target/fallback concern is fixed by rejecting unsupported targets in `Compile` before pagination, with regression coverage. The `logId` e2e comment is also stale: the test now reads the ledger-local ID with `ledgerLogID(...)`. The NumaryBot summary remains partially valid for the two points above.

Validation run: `go test ./internal/adapter/http ./internal/query -run 'Test.*Prepared|Test.*Log|Test.*Query|TestCompile' -count=1` and `git diff --check origin/release/v3.0...HEAD` passed. Current CI was still pending when checked.

internal/adapter/http/handlers_update_prepared_query.go
NumaryBot: 🟠 **[major] Update handler hardcodes LOGS target, bypassing target-aware validation**
<sub>reported by NumaryBot, codex, gfyrag</sub>

When updating an existing ACCOUNTS or TRANSACTIONS prepared query via REST, the update handler always passes QUERY_TARGET_LOGS to the decoder. This causes the decoder to accept and persist filters (e.g. logId, date, ledger) that are invalid for the stored target. The update returns 204 but only fails at execute time, which is inconsistent with the target-aware validation enforced on create.

**Suggestion:** Fetch the existing query's target before decoding the update payload and validate the incoming filters against the stored target rather than hardcoding QUERY_TARGET_LOGS.

internal/adapter/http/prepared_query_filter.go
NumaryBot: 🟠 **[major] Non-log filters accepted for LOGS queries produce semantically incorrect results**
<sub>reported by NumaryBot, codex, gfyrag</sub>

When a LOGS prepared query is created with transaction/account filters such as reference or address, those filters are not rejected. The existing compile paths for those filters yield transaction IDs or account keys; the LOGS enrichment layer then treats those bytes as ledger-local log IDs. This can cause the query to return unrelated logs instead of the intended ones, or fail with an index-inconsistency error. Additionally, a ledger filter is silently treated as a no-op (the compiler returns the universe for the executing ledger), so a filter like {"$match":{"ledger":"other"}} returns all logs for the executing ledger rather than no rows.

**Suggestion:** Either reject filters that do not produce log IDs when the query target is LOGS (including ledger/reference/address filters), or add a dedicated LOGS-specific translation/compilation path that correctly maps each filter type to log IDs. For the ledger filter specifically, evaluate the condition against the request's ledger name before accepting it.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1563#issuecomment-4934926395

tests/e2e/business/prepared_query_rest_shape_test.go
```diff
@@ -108,6 +111,69 @@ var _ = Describe("PreparedQuery REST shape (EN-1465)", Ordered, func() {
...
+
+		createReq, err := http.NewRequestWithContext(sharedCtx, http.MethodPost, restURL("/prepared-queries"), bytes.NewBufferString(createBody))
+		Expect(err).To(Succeed())
+		createReq.Header.Set("Content-Type", "application/json")
+
+		createResp, err := http.DefaultClient.Do(createReq)
+		Expect(err).To(Succeed())
+		defer func() { _ = createResp.Body.Close() }()
+
+		createRaw, _ := io.ReadAll(createResp.Body)
```
flemzord: [P2] This ignores the `io.ReadAll` error, and the execute path repeats the same pattern below. A truncated/failed body read can then be reported only as a misleading status or JSON assertion failure. Please assert each read error explicitly before using the body, in line with the repository rule that errors must never be ignored.

flemzord (commented):
Follow-up review of the updated head. These are new, validated findings not covered by the existing review threads.

gfyrag:
Re-reviewed current head `0b7eed5b58961d7e4f70ad7f3452a296217fb1c3`.

The two major issues from the previous pass still look open:

1. REST update still decodes filters as `QUERY_TARGET_LOGS`, so ACCOUNTS/TRANSACTIONS prepared-query updates can persist log-only filters and only fail later at execute time.

2. LOGS still accepts non-log filters because the LOGS validation path only skips `rejectLogOnlyConditions`; `Compile` still routes address/reference/metadata paths before `EnrichLogs` hydrates bytes as log IDs. The ledger filter also still compiles as a no-op.

New minor finding:

- `pkg/actions/filters.go` documents `LogIdGreaterThanFilter` as taking a “sequence” and names the parameter `sequence`, but this filter operates on the per-ledger `LedgerLog.Id`, not the global audit/log `sequence`. The PR fixed the e2e to read `LedgerLog.Id`, so the exported helper should use `logID` / “ledger-local log ID” wording to avoid pushing callers back into the old bug once global sequence and ledger-local log IDs diverge.

What looks fixed since the previous pass:

- Unsupported-target empty-page bypass now fails up front.
- The e2e bound now uses ledger-local `LedgerLog.Id`.
- LOGS cursor hydration is wired through `EnrichLogs`.

Checks run:

- `git diff --check e47c3fc016f50d74d9d6d09a03b0eb64fe3cb11d...HEAD`
- `GOROOT= go test ./internal/adapter/http ./internal/query ./pkg/actions -run 'Test.*Prepared|Test.*Log|Test.*Query|TestCompile|TestSchemaFieldsForTarget|TestCompileBuiltin' -count=1`
- `go test ./internal/adapter/http ./internal/query -run 'Test.*Prepared|Test.*Log|Test.*Query|TestCompile' -count=1`
- `GOROOT= go test -tags e2e ./tests/e2e/business -run '^$' -count=0`

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1563#issuecomment-4934926395

internal/adapter/http/handlers_update_prepared_query.go
```diff
@@ -33,6 +33,13 @@ func (s *Server) handleUpdatePreparedQuery(w http.ResponseWriter, r *http.Reques
...
 
+	// The update request carries only the new filter, not the target: the target
+	// is immutable and lives on the stored prepared query, which this handler
+	// does not read. Per-target filter validity is therefore enforced against the
+	// stored target by the FSM (processUpdatePreparedQuery →
+	// domain.ValidateFilterForTarget), the only layer that sees it — so an update
+	// cannot smuggle in a condition invalid for the query's target (EN-1504).
+	// This handler stays purely structural.
 	filter, err := decodePreparedQueryFilter(body.Filter)
 	if err != nil {
```
NumaryBot, line 44: 🟠 **[major] Update handler hardcodes LOGS target, bypassing target-aware validation**
<sub>reported by NumaryBot, codex, gfyrag</sub>

When updating an existing ACCOUNTS or TRANSACTIONS prepared query via REST, the update handler always passes QUERY_TARGET_LOGS to the filter decoder. This allows the decoder to accept and persist filters (e.g. logId, date, ledger) that are invalid for the stored target. The update returns 204 but only fails later at execute time, which is inconsistent with the target-aware validation enforced on create.

**Suggestion:** Fetch the existing query's target before decoding the update payload, then validate the incoming filters against the stored target rather than hardcoding QUERY_TARGET_LOGS.

internal/adapter/http/prepared_query_filter.go
```diff
@@ -59,43 +57,12 @@ func decodePreparedQueryFilter(raw json.RawMessage) (*commonpb.QueryFilter, erro
...
-	// and silently return unrelated/empty results. Reject them up front.
-	if err := rejectLogOnlyConditions(filter); err != nil {
-		return nil, err
-	}
-
+	// Per-target validity (a condition must be valid on the query's specific
+	// target — logId/date/ledger are valid on LOGS but not on
+	// ACCOUNTS/TRANSACTIONS) is enforced by domain.ValidateFilterForTarget at the
+	// admission + FSM layers, which know the concrete target for both create
+	// (from the request) and update (from the stored query). The create handler
```
NumaryBot, line 64: 🟠 **[major] Non-log filters accepted for LOGS queries produce semantically incorrect results**
<sub>reported by NumaryBot, codex, gfyrag</sub>

When a LOGS prepared query is created with transaction/account filters (e.g. reference, address), those filters are not rejected. The existing compile paths for those filters yield transaction IDs or account keys; the LOGS enrichment layer then treats those bytes as ledger-local log IDs, causing the query to return unrelated logs or fail with an index-inconsistency error. Additionally, a ledger filter in a LOGS query is silently treated as a no-op, returning all logs for the executing ledger instead of filtering by the named ledger.

**Suggestion:** Reject filters that do not produce log IDs when the query target is LOGS (including ledger/reference/address/metadata filters), or add a dedicated LOGS-specific translation path that correctly maps each filter type to log IDs. For the ledger filter specifically, evaluate the condition against the request's ledger name before accepting it.

misc/proto/common.proto
```diff
@@ -1248,21 +1249,114 @@ message UpdatedDefaultEnforcementModeLog {
...
-    LedgerCondition ledger = 8;
-    LogIdCondition log_id = 9;
-    LogBuiltinUintCondition log_builtin_uint = 10;
-    AccountHasAssetCondition account_has_asset = 11;
-    RevertedCondition reverted = 12;
-    AuditCondition audit = 13;
+    // Metadata field conditions ($match/$gt/$exists on metadata[<key>]) are
+    // valid on every target; per-target index/schema availability is enforced
+    // separately by the compiler.
+    FieldCondition field = 1 [
```
NumaryBot, line 1292: 🟠 **[major] Metadata filters accepted for LOGS but no real log-metadata index exists**

The proto annotation allows metadata conditions to be persisted for LOGS prepared queries on create/admission, but the execution path lacks a real log-metadata index. compileFieldCondition scans a NamespaceLog metadata keyspace that is never populated by the index builder for logs. These queries will either fail with index/schema errors or silently return no results at execution time.

**Suggestion:** Reject metadata filters for LOGS queries until a proper log-metadata index and execution path are implemented.

tests/e2e/business/prepared_query_rest_shape_test.go
NumaryBot: ⚪ **[nit] io.ReadAll errors silently ignored in e2e test body reads**

The test ignores errors from io.ReadAll for both the create and execute response bodies. A truncated or failed read will then surface only as a misleading status or JSON assertion failure rather than a clear read error, making failures harder to diagnose.

**Suggestion:** Assert each io.ReadAll error explicitly before using the body, consistent with the repository convention that errors must never be ignored.

internal/domain/validation.go
```diff
@@ -161,6 +162,73 @@ func ValidatePreparedQueryName(name string) Describable {
...
+// enforce filter presence separately.
+func ValidateFilterForTarget(f *commonpb.QueryFilter, target commonpb.QueryTarget) Describable {
+	if f == nil {
+		return nil
+	}
+
+	switch v := f.GetFilter().(type) {
+	case *commonpb.QueryFilter_And:
+		for _, sub := range v.And.GetFilters() {
+			if err := ValidateFilterForTarget(sub, target); err != nil {
```
flemzord: [P1] Bound this write-time traversal before recursing. `CreatePreparedQuery` and `UpdatePreparedQuery` now call this on an untrusted gRPC filter in admission/the FSM, before `query.Compile` can apply its `MaxFilterDepth` guard. A sufficiently deep `$and`/`$or`/`$not` tree can therefore exhaust the process stack—the exact fatal DoS that `Compile` caps at 100 levels. Please enforce the same limit here (or make the walk iterative) and add an over-limit regression test for both create and update.

flemzord (commented):
Found one new blocking issue in the updated target-validation path.

internal/adapter/http/handlers_update_prepared_query.go
flemzord: Rechecked on `d1f2a4c`: this is fixed. The update handler now performs structural decoding only, and `processUpdatePreparedQuery` reads the stored query from the cache and calls `ValidateFilterForTarget(order.GetFilter(), updated.GetTarget())` before mutating or persisting it. An invalid update therefore fails in the FSM rather than returning 204.

internal/adapter/http/prepared_query_filter.go
flemzord: Rechecked on `d1f2a4c`: the `reference`/`address` part is fixed—the generated validity matrix rejects both for LOGS, and create/FSM/Compile all consult it. The ledger-condition part remains actionable: `LedgerCondition` is documented as an exact match, but `compile()` still returns the current ledger's universe without resolving/comparing the condition, so a filter naming a different ledger is silently ignored. The metadata case is tracked separately below.

gfyrag:
Re-reviewed current head `d1f2a4ce3435bcb6ad11e9e9741c9d2f3afb22ed`.

Remaining findings:

1. P1: bound `ValidateFilterForTarget` recursion.

`ValidateFilterForTarget` recursively walks `$and`/`$or`/`$not` without the compiler’s `MaxFilterDepth` guard. It runs before `Compile` on create and inside the FSM on update, so a deeply nested gRPC filter can still stack-overflow the process. Please add the same depth limit or make the walk iterative, with create/update regressions.

2. Major: LOGS metadata filters are still admitted, but there is no coherent log-metadata execution path.

The proto marks metadata valid on LOGS, `SchemaFieldsForTarget(LOGS)` falls back to account schema, `compileFieldCondition` builds a ledger metadata index and scans the `l:` log namespace, while the index builder only writes account/transaction metadata indexes. Please reject metadata for LOGS until a real log metadata index exists.

3. Major: `LedgerCondition` is documented as an exact-match log filter, but compile ignores the condition and returns the current ledger universe.

A LOGS prepared query scoped to ledger A with `{ "ledger": "ledger-b" }` returns ledger A logs. Either compare the condition against `ctx.ledgerName` or reject the field.

4. Minor: the REST e2e still ignores `io.ReadAll` errors in create/execute response reads.

5. Minor: `LogIdGreaterThanFilter` still uses “sequence” wording for a ledger-local log ID.

6. Minor docs: prepared-query docs still describe the old decode/compile-backstop model for update validation, but current code validates at domain/FSM write time.

What looks fixed since the previous pass:

- REST update no longer decodes as LOGS; FSM validates against the stored target.
- Unsupported-target empty-page bypass is fixed; `Compile` rejects unsupported targets up front.
- The e2e `logId` bound now uses ledger-local `LedgerLog.Id`.
- LOGS cursor hydration is wired through `PreparedQueryCursor.logData` and `EnrichLogs`.
- LOGS now rejects `reference`/`address`; `ledger` and metadata remain open as above.

Checks run:

- `git diff --check 0b7eed5b58961d7e4f70ad7f3452a296217fb1c3..HEAD`
- `git diff --check e47c3fc016f50d74d9d6d09a03b0eb64fe3cb11d...HEAD`
- Focused `GOROOT= go test` over domain/query/http/admission/processing/actions
- `go test ./...` in `tools/protoc-gen-queryfilter-validity`
- `GOROOT= go test -tags e2e ./tests/e2e/business -run '^$' -count=0`
- `GOROOT= go build ./...`

internal/adapter/http/prepared_query_filter.go
gfyrag: Fixed in `048799594` (the ledger-condition part; reference/address were already handled by the validity matrix as you noted).

`compile()`'s `QueryFilter_Ledger` arm no longer returns the universe unconditionally. It now resolves the `LedgerCondition` value via `resolveString` and compares it to `ctx.ledgerName` (the ledger the query executes against):

- names the executing ledger → full log universe (as before);
- names any other ledger → empty (unsatisfiable) `SliceIterator`, not a silent "all logs".

`LedgerCondition` is a `StringCondition` (equality-only: hardcoded/param), and compilation is always scoped to a single ledger, so exact-match is the complete semantics — no cross-ledger scan is possible or needed. A missing condition value now fails loudly rather than matching everything.

Regression: `TestCompile_LedgerConditionOtherLedgerIsEmpty` (different ledger → no rows) and `TestCompile_LedgerConditionMissingValue` (nil value → error) in `internal/query/compile_test.go`.

internal/domain/validation.go
gfyrag: Fixed in `048799594`. `ValidateFilterForTarget` now recurses through an internal depth-bounded core (`validateFilterForTarget(f, target, depth)`) that caps the `$and`/`$or`/`$not` walk at `MaxFilterDepth`, matching `query.Compile`'s guard. To keep the two guards from drifting, `MaxFilterDepth` now lives in `internal/domain` and `query.compile` re-exports it (`const MaxFilterDepth = domain.MaxFilterDepth`) — a filter accepted at write time can no longer be rejected only at execute time, or vice versa.

Both `CreatePreparedQuery` and `UpdatePreparedQuery` route their untrusted gRPC filter through this same function (admission + FSM), so the one guard covers both paths, before any downstream recursion.

Regression: `TestValidateFilterForTarget_RejectsDeeplyNestedFilter` in `internal/domain/prepared_query_filter_test.go` — asserts an over-limit `$and`/`$or`/`$not` chain is rejected with a "nesting depth" error, and that a tree exactly at the limit is still accepted (off-by-one check).

internal/adapter/http/handlers_update_prepared_query.go
gfyrag: Thanks for re-checking — agreed, this one is fixed as of `d1f2a4c` (structural decode in the handler; `processUpdatePreparedQuery` reads the stored query and runs `ValidateFilterForTarget(order.GetFilter(), updated.GetTarget())` in the FSM before persisting). Resolving.

tests/e2e/business/prepared_query_rest_shape_test.go
gfyrag: Fixed in `048799594`. Both body reads now assert the `io.ReadAll` error before using the body:

- create path: `createRaw, err := io.ReadAll(createResp.Body); Expect(err).To(Succeed())`
- execute path (inside the `Eventually`): `execRaw, execErr := io.ReadAll(execResp.Body); g.Expect(execErr).To(Succeed())`

A truncated/failed read now surfaces as an explicit read-error assertion rather than a misleading status/JSON failure.

gfyrag: Fixed in `048799594` (same change as the [P2] thread on this line): both the create and execute `io.ReadAll` calls now assert their error explicitly before the body is used.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1563#issuecomment-4934926395

misc/proto/common.proto
NumaryBot, line 1292: 🟠 **[major] Metadata filters accepted for LOGS but no log-metadata index exists**
<sub>reported by NumaryBot, codex, flemzord</sub>

The proto annotation allows metadata conditions to be persisted for LOGS prepared queries on create/admission. However, the execution path has no real log-metadata index: SchemaFieldsForTarget(LOGS) falls back to account fields, targetTypeForQueryTarget(LOGS) requests a ledger metadata index, and the index builder only writes account/transaction metadata indexes. These accepted queries therefore fail later with index/schema errors or silently return empty results instead of filtering logs.

**Suggestion:** Reject metadata filters for LOGS queries at the admission/validation layer until a proper log-metadata index and execution path are implemented.

internal/domain/validation.go
```diff
@@ -161,6 +173,105 @@ func ValidatePreparedQueryName(name string) Describable {
...
+// own guard, so a hostile deeply-nested filter is rejected at write time
+// (admission + FSM) before it can overflow the stack here — the exact fatal
+// DoS Compile caps downstream (invariant #7: fail loudly, never recurse
+// unbounded on untrusted input).
+func validateFilterForTarget(f *commonpb.QueryFilter, target commonpb.QueryTarget, depth int) Describable {
+	if f == nil {
+		return nil
+	}
+
+	switch v := f.GetFilter().(type) {
```
NumaryBot, line 248: 🟠 **[major] ValidateFilterForTarget depth limit is off-by-one relative to query.Compile**

ValidateFilterForTarget checks the depth limit only at combinator nodes, but query.Compile checks the limit before every node including leaf nodes. For a filter with exactly MaxFilterDepth nested combinators around a leaf, write-time validation accepts the prepared query while Compile rejects it at execution time. This allows persisting prepared queries that immediately fail to execute.

**Suggestion:** Apply the same node-counting rule here as the compiler — check depth at every node, not only at combinator nodes, so the write-time guard is consistent with the execution-time guard.

internal/domain/validation.go
```diff
@@ -161,6 +173,105 @@ func ValidatePreparedQueryName(name string) Describable {
...
+}
+
+// ValidateFilterForTarget walks a QueryFilter tree and returns the first leaf
+// condition that is not valid on `target`, per the generated per-target
+// validity table (commonpb.ConditionValidForTarget) — the same gate
+// query.compile applies per-condition at execute time. Combinators
+// ($and/$or/$not) are structural and always valid; the walk recurses into their
+// children. Used at prepared-query write time so a stored filter is always
+// executable against its target: a condition invalid on the target can never be
+// satisfied and would otherwise silently widen/empty results at execute time
```
NumaryBot, line 217: 🟠 **[major] ValidateFilterForTarget unbounded recursion enables stack-overflow DoS**

ValidateFilterForTarget recursively walks $and/$or/$not combinators without the MaxFilterDepth guard that query.Compile enforces (capped at 100 levels). It is called on untrusted gRPC filters during CreatePreparedQuery and UpdatePreparedQuery admission before Compile runs. A sufficiently deeply nested filter tree can exhaust the process stack before the compiler's guard is ever reached, causing a fatal DoS.

**Suggestion:** Enforce the same depth limit as query.Compile in ValidateFilterForTarget (or make the traversal iterative), and add over-limit regression tests for both create and update paths.

tests/e2e/business/prepared_query_rest_shape_test.go
```diff
@@ -108,6 +111,71 @@ var _ = Describe("PreparedQuery REST shape (EN-1465)", Ordered, func() {
...
+
+		createReq, err := http.NewRequestWithContext(sharedCtx, http.MethodPost, restURL("/prepared-queries"), bytes.NewBufferString(createBody))
+		Expect(err).To(Succeed())
+		createReq.Header.Set("Content-Type", "application/json")
+
+		createResp, err := http.DefaultClient.Do(createReq)
+		Expect(err).To(Succeed())
+		defer func() { _ = createResp.Body.Close() }()
+
+		createRaw, err := io.ReadAll(createResp.Body)
```
NumaryBot, line 142: ⚪ **[nit] io.ReadAll errors silently ignored in e2e test body reads**
<sub>reported by flemzord, NumaryBot</sub>

The test ignores errors from io.ReadAll for both the create and execute response bodies. A truncated or failed read will surface only as a misleading status or JSON assertion failure rather than a clear read error, making failures harder to diagnose.

**Suggestion:** Assert each io.ReadAll error explicitly before using the body, consistent with the repository convention that errors must never be ignored.

misc/proto/common.proto
gfyrag: Fixed in `e1eda8f1b` — took option (a): align the contract with reality rather than advertise an unsupported capability.

You are right that there is no populated log-metadata index — the index builder never fills the `NamespaceLog` metadata keyspace that `compileFieldCondition` scans, so a metadata filter on LOGS returns empty/errors at execute time. Dropped `QUERY_TARGET_LOGS` from the `field` (metadata) arm in `misc/proto/common.proto` and regenerated. The generated validity matrix now has `ConditionKindField = false` for LOGS, so both `domain.ValidateFilterForTarget` (admission + FSM) and `query.Compile` reject a metadata filter on LOGS up front with a `"…not valid on logs queries"` error, instead of silently returning empty at execute time.

Building a real log-metadata index is a separate feature, out of scope for EN-1503 — left to the existing tracking. This is consistent with the analogous #1436 decision not to advertise a capability that has no execution path behind it.

Tests: updated the human-readable validity-matrix contract test (`common_queryfilter_validity_test.go`, `ConditionKindField` LOGS → false) and added `$match` / `$exists` metadata-on-LOGS rejection cases to `TestValidateFilterForTarget`.

internal/domain/validation.go
gfyrag: Good catch — legitimate, fixed in `55e89af21`. You are right: `query.compile` checks `depth >= MaxFilterDepth` on entry of *every* node (leaves included), while the validator only checked at combinator nodes, so a tree with exactly `MaxFilterDepth` combinators wrapping a leaf was accepted at write time but rejected by Compile at execute time.

Moved the depth check to the top of every node in `validateFilterForTarget` so both count identically: N combinators wrapping a leaf enters the leaf at depth N → `MaxFilterDepth-1` is the deepest both accept, `MaxFilterDepth` is rejected by both. Added `TestCompile_DepthBoundMatchesValidator` (query pkg) as a cross-guard parity test that pins `compile()` and `domain.ValidateFilterForTarget` to the same boundary, plus corrected the boundary assertions in `TestValidateFilterForTarget_RejectsDeeplyNestedFilter`.

internal/domain/validation.go
gfyrag: Already fixed in `048799594` (and the boundary hardened in `55e89af21`): `ValidateFilterForTarget` now enforces `MaxFilterDepth` on the write-time walk (admission + FSM), before any downstream recursion, so a deeply-nested filter can no longer overflow the stack. `MaxFilterDepth` lives in `internal/domain` and is re-exported by `query.compile` (single source of truth). Over-limit regression tests cover $and/$or/$not and both create + update route through this same function. Resolving as duplicate of the now-fixed thread.

misc/proto/common.proto
gfyrag: Already fixed in `e1eda8f1b`: dropped `QUERY_TARGET_LOGS` from the `field` (metadata) arm in `misc/proto/common.proto` and regenerated, so the validity matrix now has `ConditionKindField = false` for LOGS. Both `domain.ValidateFilterForTarget` (admission + FSM) and `query.Compile` reject a metadata filter on LOGS up front, rather than failing/empty at execute time. A real log-metadata index is a separate feature, tracked separately. Resolving.

tests/e2e/business/prepared_query_rest_shape_test.go
gfyrag: Already fixed in `048799594`: both the create and execute `io.ReadAll` calls assert their error explicitly before the body is used. Resolving.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1563#issuecomment-4934926395

internal/domain/validation.go
```diff
@@ -161,6 +173,100 @@ func ValidatePreparedQueryName(name string) Describable {
...
+// of every node before dispatching). Matching it node-for-node keeps the two
+// guards in lockstep: a filter accepted at write time is guaranteed to compile
+// at execute time, and vice versa (a shallower write-time bound would let an
+// unexecutable prepared query be persisted; a deeper one would overflow the
+// stack here before Compile's guard is ever reached — the exact fatal DoS,
+// invariant #7).
+func validateFilterForTarget(f *commonpb.QueryFilter, target commonpb.QueryTarget, depth int) Describable {
+	if f == nil {
+		return nil
+	}
```
NumaryBot, line 242: 🟠 **[major] Filter depth limit in ValidateFilterForTarget inconsistent with Compile's counting semantics**

ValidateFilterForTarget recursively walks $and/$or/$not nodes without the same depth guard that query.Compile enforces via MaxFilterDepth. Compile counts every recursive call (including the leaf), while the validator only checks combinator nodes. A filter at exactly MaxFilterDepth nesting can therefore pass admission/FSM validation but be rejected at execution time — the write path persists a prepared query that can never successfully execute. Additionally, there is no recursion depth guard at all on the validator path, meaning a deeply nested gRPC filter could exhaust the stack before Compile's guard is even reached.

**Suggestion:** Add the same depth-counting semantics and limit to ValidateFilterForTarget as are used in Compile (MaxFilterDepth), counting every recursive node not just combinators. Consider making the walk iterative, and add create/update regression tests for the boundary depth.

pkg/actions/filters.go
```diff
@@ -105,6 +105,21 @@ func LedgerFilter(ledger string) *commonpb.QueryFilter {
 	}
 }
 
+// LogIdGreaterThanFilter creates a LOGS filter matching entries whose logId is
```
NumaryBot, line 108: 🟡 **[minor] LogIdGreaterThanFilter helper misleadingly named/documented as 'sequence'**

The exported helper names and documents its argument as 'sequence', but the filter operates on the per-ledger LedgerLog.Id, not the global audit/log sequence. Callers following this doc can pass Log.GetSequence() and get incorrect results once global sequences diverge from ledger-local log IDs. The PR already fixed the e2e test to use the ledger-local ID, but the exported API remains a trap for future callers.

**Suggestion:** Rename the parameter to logID (or ledgerLocalLogID) and update the doc comment to explicitly state this is the per-ledger LedgerLog.Id, not the global sequence.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1563#issuecomment-4934926395

pkg/actions/filters.go
NumaryBot, line 108: 🟡 **[minor] LogIdGreaterThanFilter helper uses misleading 'sequence' naming instead of ledger-local log ID**

The exported helper `LogIdGreaterThanFilter` names its parameter `sequence` and is documented in terms of a global sequence, but the underlying `LogIdCondition` filters on the per-ledger `LedgerLog.Id`, not the global audit/log sequence. Callers who follow the parameter name and pass `Log.GetSequence()` will get wrong results once the global sequence diverges from ledger-local IDs (e.g. after ledger creation or activity on other ledgers). This was the root cause of a previously flaky e2e test (now fixed by using `LedgerLog.Id`), but the exported API surface still pushes callers back toward the old bug.

**Suggestion:** Rename the parameter from `sequence` to `logID` (or `ledgerLocalLogID`) and update the associated comment to clearly state this operates on the ledger-local `LedgerLog.Id`, not the global sequence. Additionally update any generated or exported documentation that references 'sequence' in this context.

internal/domain/validation.go
gfyrag: Already fixed in `55e89af21` — this scanned the pre-fix code. `validateFilterForTarget` now checks `depth >= MaxFilterDepth` at the **top of every node** (combinators AND leaf), the exact node-for-node counting `query.compile` uses, so the two guards agree at the boundary: N combinators wrapping a leaf enters the leaf at depth N → `MaxFilterDepth-1` accepted by both, `MaxFilterDepth` rejected by both. `MaxFilterDepth` lives in `internal/domain` and is re-exported by `query.compile` (single source of truth). Added `TestCompile_DepthBoundMatchesValidator` (query pkg) as a cross-guard parity regression, plus corrected boundary assertions in `TestValidateFilterForTarget_RejectsDeeplyNestedFilter`. Resolving.

pkg/actions/filters.go
gfyrag: Fixed in `1cc67338e`. Renamed the parameter `sequence` → `ledgerLocalLogID` and rewrote the doc to state explicitly that `LogIdGreaterThanFilter` filters the per-ledger `LedgerLog.Id`, not the global `Log.GetSequence()` — and that passing the global sequence skips too many rows once the two diverge (the trap that made the earlier e2e test flaky). Behaviour unchanged; the sole caller passes positionally.

gfyrag: Duplicate of the finding on this line — fixed in `1cc67338e`: parameter renamed to `ledgerLocalLogID` and the doc now spells out that this is the ledger-local `LedgerLog.Id`, not the global audit/log sequence. Resolving.

gfyrag:
Status of the latest NumaryBot review pass against current head `1cc67338e`. All inline threads are resolved; recapping the non-inline verdict:

**Fixed**
- **`pkg/actions/filters.go` — `LogIdGreaterThanFilter` misleading `sequence` naming** (minor): fixed in `1cc67338e`. Parameter renamed `sequence` → `ledgerLocalLogID` and the doc now states it filters the per-ledger `LedgerLog.Id`, not the global `Log.GetSequence()`, so the exported API no longer pushes callers toward the divergence bug that made the earlier e2e test flaky.
- **`common.proto` — metadata filters accepted for LOGS with no log-metadata index** (major): fixed in `e1eda8f1b`. Dropped `QUERY_TARGET_LOGS` from the `field` (metadata) arm; the generated validity matrix now has `ConditionKindField = false` for LOGS, so both `domain.ValidateFilterForTarget` and `query.Compile` reject metadata-on-LOGS up front instead of failing/empty at execute time.

**Rebutted (already fixed before the bot re-scanned)**
- **`internal/domain/validation.go` — depth-guard off-by-one / unbounded recursion** (major): fixed in `048799594` (bound added) and `55e89af21` (counting aligned node-for-node with `query.compile`). The re-post scanned the pre-fix code. Both guards now agree at the boundary (`MaxFilterDepth-1` accepted, `MaxFilterDepth` rejected), pinned by the cross-guard parity test `TestCompile_DepthBoundMatchesValidator`.

**Commit map**
- `048799594` — ledger-condition exact match + bound `ValidateFilterForTarget` recursion + e2e `io.ReadAll` error asserts
- `e1eda8f1b` — reject metadata filters on LOGS (validity matrix)
- `55e89af21` — align write-time depth guard with `Compile` node-for-node
- `1cc67338e` — clarify `LogIdGreaterThanFilter` operates on the ledger-local log ID

Verification: `go build ./...` clean; `go test ./internal/adapter/http/... ./internal/domain/... ./internal/query/... ./internal/proto/commonpb/... ./pkg/actions/...` green; e2e compiles under the `e2e` tag (running it needs Docker). The PR description has been updated to match the current implementation (single generated validity matrix, metadata rejected on LOGS, exact-match ledger condition, depth-guard parity).

NumaryBot (commented):

All previously identified blocker and major issues have been resolved in the current head. The REST update handler no longer hardcodes QUERY_TARGET_LOGS — it fetches the stored query, validates the incoming filter against the stored target via ValidateFilterForTarget, and only persists on success. Non-log filters (reference, address) are rejected for LOGS via the validity matrix. The LedgerCondition compile path now compares against the executing ledger name and returns an empty iterator for non-matching ledgers. The unsupported-target silent empty-page bypass is fixed in Compile with an explicit allow-list. The e2e test now uses the ledger-local log ID. io.ReadAll errors are asserted in tests. The previously noted naming/documentation hazard in LogIdGreaterThanFilter (using "sequence" wording for a per-ledger ID) is a minor documentation nit with no correctness impact on the current implementation. No remaining blocker or major findings were identified by any reviewer.

### Findings outside the diff

⚪ **[nit] LogIdGreaterThanFilter parameter named 'sequence' but operates on ledger-local log ID** — `pkg/actions/filters.go`

The exported LogIdGreaterThanFilter helper uses the parameter name 'sequence' and documentation language implying the global audit/log sequence, while the filter actually operates on the per-ledger LedgerLog.Id. This naming mismatch could mislead future callers into passing a global sequence value, reintroducing the bug that the e2e fix corrected once global sequence and ledger-local log IDs diverge.

**Suggestion:** Rename the parameter to 'logID' (or similar) and update any documentation to clarify this is a per-ledger log ID, not the global audit sequence.



Summary: https://github.com/formancehq/ledger/pull/1563#issuecomment-4934926395

gfyrag:
Rechecked current head `1cc67338e293cd26c1936f0a8d52c344a71f7b1a`. The functional LOGS prepared-query issues look addressed, but `docs/technical/architecture/subsystems/read-path/prepared-queries.md` still describes the update path with the old contract:

> The update path carries only the new filter ... so it defers the log-only check to the compile backstop.

That is stale now. `processUpdatePreparedQuery` reads the stored immutable target and validates the new filter against it before persisting, so an invalid LOGS-only condition on an ACCOUNTS/TRANSACTIONS stored query is rejected at update time, not deferred until execution. Please update that paragraph so the architecture docs match the current write-time validation behavior.

internal/adapter/http/handlers_create_prepared_query.go
```diff
@@ -57,6 +57,15 @@ func (s *Server) handleCreatePreparedQuery(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
+	// Reject conditions invalid on this query's specific target early (precise
```
NumaryBot: ⚪ **[nit] Duplicated ValidateFilterForTarget call on create path**

The REST create handler calls ValidateFilterForTarget twice in sequence for the same filter and target. The second call is redundant — it doubles validation work for large/deep filters and leaves two copies of the same error path to maintain.

**Suggestion:** Remove the second (redundant) ValidateFilterForTarget call, keeping only one validation block before persisting the prepared query.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1563#issuecomment-4934926395

internal/adapter/http/handlers_create_prepared_query.go
gfyrag: Fixed in `fd04968cd`. You're right — the create handler had two identical back-to-back `ValidateFilterForTarget(filter, target)` blocks (a merge-resolution artifact from reconciling with the #1561 QueryFilter foundation). Removed the redundant second block; one validation before persisting remains. Build green.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1563#issuecomment-4934926395

flemzord (approved):
Re-reviewed current head 5df668d36d4c88fcba0cf1411f9fbe2ad7b38ac9. No High or Critical findings remain. The previously reported stale architecture-doc wording is non-blocking under the requested severity threshold; the functional LOGS prepared-query path and current fixes are coherent.

PR #1564, opened 2026-07-11
fix(EN-1525): fail closed when a WAL exists without its creation marker

gfyrag:
## Summary

Fixes a destructive fail-open branch in `internal/storage/wal/wal_default.go` (EN-1520 finding). On the creation-marker-missing startup branch, a WAL with populated consensus state hit `assert.Unreachable` and then **fell through to `os.RemoveAll`**. `assert` is a no-op in production, so valid consensus state (votes, unsnapshotted entries, cluster membership) could be deleted — risking double-vote/split-brain and acknowledged-write loss. A verification error was likewise coerced to "not populated" and removed.

## Fix

On the marker-missing branch, `walDirPopulated` inspects the WAL read-only and cleanup proceeds only when the directory is **proven empty**; anything populated or unverifiable returns a **contextual startup error before any destructive cleanup**. `assert.Unreachable` is kept as the Antithesis signal but is now **paired with a mandatory return**, so a production / no-op-assert build cannot fall through.

`walDirPopulated` does not go through `wal.OpenForRead` + `ReadAll`: that read-mode path silently drops two signals a marker-missing WAL can carry (verified against etcd v3.7.0):

- `ReadAll` returns only metadata / HardState / entries and **drops snapshot records entirely**, so a WAL that recorded `CreateSnapshot(0, initialConfState)` but crashed before any HardState/entry would read as empty and be deleted, discarding the persisted cluster membership.
- `ReadAll` (and `wal.Verify`) **silently accept `io.ErrUnexpectedEOF`** in read mode and return the records decoded so far, so a segment torn mid-write past its first record reads as state-free instead of surfacing the torn tail.

Instead it runs a single raw `wal.NewDecoder` pass over the sorted `.wal` segments (public etcd API, opened read-only, no lock) and classifies the directory as **populated** on any of:

- a non-empty `HardState` (`StateType` record) — losing it risks double-vote → split brain;
- at least one log entry (`EntryType` record) — losing an unsnapshotted entry is acknowledged-write loss;
- a snapshot record carrying a `ConfState` with voters (`SnapshotType` record) — the persisted cluster membership, even at index 0.

Any terminal decode error other than a clean `io.EOF` (torn tail `io.ErrUnexpectedEOF`, CRC mismatch, unknown record type, open failure) is returned so the caller **fails closed**. A genuinely fresh, empty bootstrap WAL decodes cleanly to `io.EOF` with no state and is reported as not populated, preserving the empty-remnant recovery path.

## Tests (filesystem)

Exercised under the default no-op-assert build, proving no fall-through:

- empty bootstrap remnant → recreated + marker rewritten (existing recovery behaviour preserved)
- populated HardState → fails closed, files preserved
- entries-only (empty HardState) → fails closed, files preserved
- malformed WAL → fails closed, files preserved
- **torn tail on a state-free-prefix WAL → fails closed, files preserved** (injects a genuine torn record; fails against the old `ReadAll`-based logic)
- **snapshot-ConfState-only WAL → fails closed, files preserved** (initial `CreateSnapshot(0, cs)`; fails against the old `ReadAll`-based logic)

`GOROOT= go build ./...`, full `internal/storage/wal` tests, gofmt and `golangci-lint run internal/storage/wal/...` (0 issues) all pass.

Refs EN-1520, EN-1525.


internal/storage/wal/wal_default.go
```diff
@@ -73,6 +73,34 @@ type DefaultWAL struct {
 }
 
 // New creates a new DefaultWAL instance.
+// walDirPopulated reports whether the etcd WAL directory at walDir holds any
```
NumaryBot, line 75: 🟠 **[major] Restore New's doc comment**

This helper was inserted between `// New creates a new DefaultWAL instance.` and `func New`, so `New` no longer has an immediately preceding doc comment. With the repo's enabled `revive` exported rule, `golangci-lint` will flag `New` as missing a comment; move the `New` comment back above `func New` and give `walDirPopulated` its own comment.

internal/storage/wal/wal_default.go
```diff
@@ -73,6 +73,34 @@ type DefaultWAL struct {
...
+		return false, fmt.Errorf("opening WAL for verification: %w", err)
+	}
+	defer func() { _ = w.Close() }()
+
+	_, hardState, entries, err := w.ReadAll()
+	if err != nil {
+		return false, fmt.Errorf("reading WAL for verification: %w", err)
+	}
+
+	return !raft.IsEmptyHardState(hardState) || len(entries) > 0, nil
```
NumaryBot: 🟠 **[major] Treat WAL snapshot records as populated**

When the creation marker is lost after `CreateSnapshot(0, initialConfState)` but before any HardState or log entries are written, the WAL still contains the snapshot record that carries the persisted ConfState. This predicate ignores snapshot records, so that WAL is classified as empty and the caller falls through to `os.RemoveAll`, discarding the record needed to recover the initial ConfState; include WAL snapshot records in the populated/ambiguous check before allowing cleanup.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1564#issuecomment-4949194895

internal/storage/wal/wal_default.go
flemzord: Confirmed on `64a84e9c`. `CreateSnapshot` persists a `walpb.Snapshot` carrying `ConfState`, while `ReadAll()` only uses snapshot records for start-position matching and returns metadata, HardState and entries — not the snapshot itself. With the empty start snapshot, an index/term 0 initial snapshot matches cleanly and this helper returns false, so the WAL record and its otherwise orphaned snap file are discarded. This needs a snapshot-record check plus a marker-missing initial-ConfState regression test before the directory can be declared empty.

internal/storage/wal/wal_default.go
flemzord: Confirmed as a documentation/style issue on the current head: the `New` sentence is now part of `walDirPopulated`'s comment block and `func New` has no doc comment. The Default workflow is green, so this is not currently a CI blocker as phrased, but the comments should still be split so each function documents its own contract.

internal/storage/wal/wal_default.go
```diff
@@ -73,6 +73,34 @@ type DefaultWAL struct {
...
+// empty WAL decodes cleanly to an empty HardState with no entries and is
+// reported as not populated.
+func walDirPopulated(lg *zap.Logger, walDir string) (bool, error) {
+	w, err := wal.OpenForRead(lg, walDir, &walpb.Snapshot{})
+	if err != nil {
+		return false, fmt.Errorf("opening WAL for verification: %w", err)
+	}
+	defer func() { _ = w.Close() }()
+
+	_, hardState, entries, err := w.ReadAll()
```
flemzord: [P1] `ReadAll` does not actually surface every torn-tail read error in this mode. etcd's read-only path explicitly accepts `io.ErrUnexpectedEOF` and returns the records decoded so far, so a segment truncated before its HardState/entry record can come back with empty state and be classified safe here, falling through to `os.RemoveAll`. The malformed test overwrites records with `0xFF`, which exercises a decoder/CRC error rather than this accepted truncated-tail case. Please validate the complete segment/tail before cleanup and add a truncated populated-WAL regression test.

flemzord (commented):
One additional destructive fail-open path remains in the read-only verification.

internal/storage/wal/wal_default.go
gfyrag: Fixed in 956680d5. Split the comments: `func New` now carries its own `// New creates a new DefaultWAL instance.` doc line, and `walDirPopulated` has its own comment block documenting its contract.

internal/storage/wal/wal_default.go
gfyrag: Legitimate — confirmed and fixed in 956680d5. `wal.ReadAll` returns only metadata/HardState/entries and drops snapshot records entirely, so a WAL that recorded `CreateSnapshot(0, initialConfState)` but crashed before any HardState/entry read as empty and would have been deleted.

`walDirPopulated` no longer goes through `ReadAll`. It now runs a single raw `wal.NewDecoder` pass over the sorted `.wal` segments and inspects `SnapshotType` records directly: a snapshot record carrying a `ConfState` with voters is classified as populated (even at index 0). Added `TestNew_MarkerMissing_SnapshotConfStateOnly_FailsClosed`, which fails against the old `ReadAll`-based logic and passes now.

internal/storage/wal/wal_default.go
gfyrag: Legitimate — confirmed and fixed in 956680d5. I verified against etcd v3.7.0: in read mode (`OpenForRead`) `ReadAll` hits the `case nil:` (nil tail) branch, which explicitly accepts both `io.EOF` and `io.ErrUnexpectedEOF` and returns the records decoded so far with `err == nil`. A fresh bootstrap WAL with a torn record appended reads back as empty HardState + 0 entries — so the old logic would have deleted it. (`wal.Verify` has the same tolerant contract, so it was not a fix either.)

`walDirPopulated` no longer relies on `ReadAll`. It drains a raw `wal.NewDecoder` over the segments and treats any terminal decode error other than a clean `io.EOF` (torn tail `io.ErrUnexpectedEOF`, CRC mismatch, unknown record type) as ambiguous, returning an error so the caller fails closed.

Re: the malformed-`0xFF` test — you are right that it exercises a CRC/decode path, not the accepted truncated-tail case. Added a dedicated `TestNew_MarkerMissing_TornTail_FailsClosed` that starts from a fresh (state-free-prefix) bootstrap WAL and injects a genuine torn record (a frame header claiming a payload larger than the bytes on disk), which makes the decoder return `io.ErrUnexpectedEOF`. It fails against the old logic and passes now.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1564#issuecomment-4949194895

gfyrag:
Addressed the remaining non-inline review items in 956680d5:

**@flemzord — "One additional destructive fail-open path remains in the read-only verification":** confirmed and fixed. The fail-open path was the read-only `ReadAll` verification silently accepting a torn tail (`io.ErrUnexpectedEOF`) and returning empty state, so a torn WAL fell through to `os.RemoveAll`. Verified against etcd v3.7.0 that both `ReadAll` (read-mode `case nil` branch) and `wal.Verify` accept `io.ErrUnexpectedEOF` — neither surfaces it. `walDirPopulated` no longer uses that path: it drains a raw `wal.NewDecoder` and fails closed on any terminal error other than a clean `io.EOF`. The three inline threads (doc comment, snapshot-record, torn-tail) are all fixed and resolved, and the PR description is updated to describe the raw-decoder implementation.

**NumaryBot:** its latest automated review on head 956680d5 is "✅ Approve — no findings"; the two prior inline findings (snapshot-record + torn-tail handling) are the ones fixed by the raw-decoder rewrite above, so no findings remain outstanding.

CI green, coverage +0.13%.

gfyrag:
Rechecked current head `956680d518b777971155142afda235dec1b57d9b`. The snapshot-only and torn-tail cases are fixed, but the raw decoder path still needs the same fail-closed treatment for segment identity/continuity.

`walDirPopulated` now does `Glob("*.wal")`, lexical sort, and feeds the files directly to `wal.NewDecoder`. That avoids `ReadAll`'s tolerant tail behavior, but it also bypasses etcd WAL's normal segment-name and sequence checks. If a middle segment is missing, or a file has an invalid/out-of-sequence WAL name, that is ambiguous/corrupt state. Today the helper can still decode the remaining files cleanly; if those remaining records only contain metadata/CRC/empty snapshot state, `populated` stays false and `New` can fall through to deleting the directory. A missing segment should fail closed even when the surviving segments themselves decode cleanly.

Please validate the etcd segment names (`%016x-%016x.wal`) and sequence continuity before decoding, or reuse the etcd helper that enforces it while preserving the raw-decoder torn-tail behavior you need here. A regression with segments `0000000000000000-0000000000000000.wal` and `0000000000000002-...wal` (or an invalid segment basename) should return an error, not `populated=false`.

Docs also need the EN-1525 update: `docs/technical/architecture/subsystems/storage/storage.md` still says the marker-missing branch deletes any existing WAL directory and recreates it. It should describe the new verification behavior: delete only the proven empty-bootstrap remnant, and fail closed for populated or ambiguous WAL state.

internal/storage/wal/wal_default.go
```diff
@@ -72,6 +72,103 @@ type DefaultWAL struct {
...
+// clean io.EOF is treated as fully read; a torn tail (io.ErrUnexpectedEOF) or
+// any other error (CRC mismatch, unknown record type, open failure) is returned
+// so the caller fails closed. A genuinely fresh, empty WAL decodes cleanly (its
+// metadata + CRC + empty-snapshot records terminate on io.EOF) with no state and
+// is reported as not populated.
+func walDirPopulated(walDir string) (bool, error) {
+	// etcd names segments %016x-%016x.wal, so a lexical sort is chronological.
+	// A single concatenated decoder over all segments mirrors how etcd's own
+	// ReadAll / ValidSnapshotEntries stream the directory.
+	names, err := filepath.Glob(filepath.Join(walDir, "*.wal"))
```
NumaryBot, line 110: 🟠 **[major] Missing WAL segment name and sequence-continuity validation allows fall-through to RemoveAll**

The `walDirPopulated` helper uses `Glob("*.wal")` plus a lexical sort to discover segments, then feeds them directly to a raw `wal.NewDecoder`. This bypasses etcd's normal checks for the `%016x-%016x.wal` filename format and segment-sequence continuity. If a middle segment is absent, a filename is malformed, or segments are out of sequence, the remaining files may still decode cleanly without surfacing any State, Entry, or voter-Snapshot record. In that case `populated` stays false and the caller falls through to `os.RemoveAll`, deleting a WAL directory that is actually ambiguous or corrupt — exactly the fail-open behaviour the PR is intended to prevent.

**Suggestion:** Before decoding, validate each discovered filename against the `%016x-%016x.wal` pattern and verify that the sequence numbers form a contiguous, monotonically increasing series. Any gap or malformed name should cause `walDirPopulated` to return an error so the caller fails closed. Alternatively, reuse etcd's segment-discovery helper that enforces these invariants while keeping the strict raw-decoder torn-tail behaviour already implemented.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1564#issuecomment-4949194895

internal/storage/wal/wal_default.go
flemzord: Revalidated on current head `9a23d822`. Explicit segment-name/sequence validation is still worthwhile, but I no longer consider this High: for normally produced segments, removing or reordering a segment breaks the rolling CRC at the next non-CRC record, and any surviving HardState/Entry/voter-Snapshot is classified as populated, so the destructive fall-through is prevented. The residual gap is malformed or contrived state-free segment sets, which I am treating as non-blocking hardening under the current High/Critical-only approval threshold. Please still add explicit name/start-sequence/continuity validation and a focused regression.

flemzord (approved):
Reviewed current head `9a23d822`. I found no remaining High or Critical blocker. The open segment identity/continuity concern remains worthwhile hardening, but it is non-blocking under the current approval threshold because normally produced missing or reordered segment chains fail the rolling CRC/state scan closed.

PR #1566, opened 2026-07-11
fix(EN-1527): strict cache/Bloom recovery decoding, fail closed on corruption

gfyrag:
## Summary

Recovery accepted malformed persisted cache/Bloom projection rows and could publish a partially-restored Bloom filter as **ready** (EN-1520 finding). A Bloom false negative suppresses a required Pebble preload (invariant #6); a mis-decoded cache row diverges the in-memory cache (invariant #1). These are strict-decoding bugs — no new verification subsystem is added.

## Fix — `internal/infra/state/cache_snapshotter.go`

- Cache-meta and per-generation-meta reads treat **only** `pebble.ErrNotFound` as absent; every other Get / closer error is propagated (was: any error → silently default the generation, restoring the wrong gen mapping).
- Cache-entry rows require the **exact** `[gen][type][16B U128]` key shape — a short key (was skipped → dropped a live entry) or a longer key (was truncated → trailing bytes accepted) now fails recovery.
- **Whole-namespace validation** — `validateCacheNamespace` runs one linear pass over the entire `[ZoneCache]` namespace *before* `restoreGeneration` and fails closed on any row that is not the global meta key, a per-generation meta row, or a `[gen][registered-slot][U128]` entry (generation byte must be one of the two live gens; slot code must be registered — set derived from `s.slots`, no second hardcoded list). This closes the gap where a row with a drifted generation/attribute-type byte fell outside every per-slot iterator and was silently ignored.
- `parseLeanValue` returns a contextual **error instead of panicking** on a short header / tombstone-with-payload / unknown flag, with a single accurate doc comment.
- `hasPersistedBloomBlocks` returns `(bool, error)`; a probe read/iterator failure is no longer converted into first-boot (which would republish a ready filter from an unreadable DB and mask corruption).
- `restoreBloomFilters` **classifies the Bloom namespace before choosing the restore path**: genuine corruption fails closed, config drift takes the full async rebuild, a clean namespace with blocks takes the synchronous restore (EN-1410 ordering preserved).

## Fix — `internal/infra/bloom/bloom.go`

- `Filter.RestoreFromStore` requires the exact key length, an **in-range block index**, and a value of **exactly one block**; any malformed row fails the restore (was: silently skipped, then the filter published ready).
- **Namespace classification** — `FilterSet.ClassifyPersistedNamespace` returns `(configDrift, err)` over the whole `[SubGlobBloom]` range:
  - an attribute-type byte that is **not a known bloom type**, or a row too short to carry one → **corruption**, fail closed;
  - a **known-but-disabled** attribute type → **config drift** (a node restarted with CLI bloom settings that disable a previously-enabled attribute; the persisted blocks are stale but valid) → tolerated, caller takes the rebuild path. This avoids bricking a node after an attribute-type change: aborting recovery here would prevent the node from ever reaching `proposeClusterConfigIfNeeded` to purge/rebuild the blocks.
  - Known-code set derived from `bloomTypes(nil)`.

## Tests

- cache wrong-length key, unknown attribute-type byte, out-of-range generation byte → restore fails;
- malformed lean values (short / tombstone-with-payload / unknown-flag) assert an **error, not a panic**;
- bloom classification: unknown attr-type + short row → corruption (error); known-but-disabled attr → config drift (`configDrift=true`, no error); enabled rows → clean;
- state-level `RestoreToleratesConfigDrift` — a config-B restart over config-A blocks boots via the async populate path rather than aborting;
- a malformed persisted bloom block → restore fails and `IsReady()` stays **false**.

## Acceptance criteria

- No malformed cache/Bloom row is a silent no-op. ✅
- Whole-namespace validation catches rows outside known prefixes, not just within them. ✅
- Genuine corruption fails closed; legitimate config drift boots via rebuild (node stays bootable after attribute-type changes). ✅
- Non-`ErrNotFound` storage errors are propagated. ✅
- Bloom readiness remains false on every restore failure. ✅
- `GOROOT= go build ./...`, full state + bloom tests, gofmt, golangci-lint all pass. ✅

Refs EN-1520.


internal/infra/state/cache_snapshotter.go
```diff
@@ -575,7 +612,16 @@ func (s *CacheSnapshotter) restoreGeneration(reader dal.PebbleReader, genByte by
...
+	hasBlocks, err := s.hasPersistedBloomBlocks(store)
+	if err != nil {
+		// A probe failure is NOT first-boot: silently falling back to a full
+		// attribute scan would republish a "ready" filter from a database whose
+		// persisted blocks we could not even read, masking corruption. Fail
+		// recovery so readiness stays false (EN-1527).
+		return fmt.Errorf("probing persisted bloom blocks: %w", err)
+	}
+
+	if hasBlocks {
```
NumaryBot: 🔴 **[blocker] Reject malformed Bloom prefix rows before ready**

When the Bloom range contains a malformed key that is not visited by any enabled per-attribute iterator (for example `[ZoneGlobal, SubGlobBloom]`, `[ZoneGlobal, SubGlobBloom, 0x00]`, or a disabled/unknown attr code), `hasPersistedBloomBlocks` returns true here and boot takes the persisted-block restore path. `FilterSet.RestoreFromStore` then skips that row, `runBloomTaskBody` replays only cache generations and marks the filter ready, so stores with older attributes outside cache can still publish an incomplete Bloom and reintroduce false negatives instead of failing closed or doing a full scan.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1566#issuecomment-4949244008

internal/infra/state/cache_snapshotter.go
```diff
@@ -529,9 +560,15 @@ func (s *CacheSnapshotter) restoreGeneration(reader dal.PebbleReader, genByte by
...
 			key := iter.Key()
-			// Key format: [0xFF][gen][type][16-byte U128]
-			if len(key) < 3+16 {
-				continue
+			// Key format: [0xFF][gen][type][16-byte U128]. Require the exact
+			// shape — a short key would previously be skipped (dropping a live
+			// cache entry) and a longer key silently truncated to 16 bytes.
+			// Either is a corrupt row; fail closed rather than restore partial
+			// cache state (EN-1527).
+			if len(key) != 3+16 {
```
flemzord, line 577: 🔴 **[P1] Validate the whole cache generation namespace, not only known slot prefixes**

This exact-length check only runs inside the per-slot iterators. If corruption changes a persisted row's generation or attribute-type byte to a value outside those known prefixes, recovery silently ignores the row. On the persisted-Bloom restart path, that can drop a cache entry written since the last Bloom flush from `replayBloomFromCache`, then publish a ready filter with a false negative—the same invariant this PR is intended to close. Please scan/classify every `[ZoneCache][gen]` row once (allowing only generation metadata and registered slot codes), fail on unknown/malformed rows, and add a regression test for an unknown generation/type.

internal/infra/state/cache_snapshotter.go
```diff
@@ -29,10 +30,15 @@ import (
 // row is produced by writeCacheRaw, so any other shape means a corrupted store
 // or a forward-incompatible binary, and silently treating an unknown flag as
 // live would let either case resurrect deleted keys.
-func parseLeanValue(value []byte) (tag uint64, deleted bool, valueBytes []byte) {
+// parseLeanValue decodes a persisted 0xFF lean cache value. It returns a
```
flemzord: 🟡 **[P3] Remove the stale panic contract above this decoder**

The preceding comment still says `parseLeanValue` panics on malformed input, while this new contract explicitly returns errors. Keeping both descriptions makes recovery behavior ambiguous to future callers and tests. Please replace the two blocks with one accurate doc comment for the error-returning implementation.

flemzord (commented):
Two additional findings from the full recovery-path review are attached inline. The existing malformed-Bloom-prefix blocker also remains valid.

gfyrag:
Re-reviewed current head `06eed28f981d5442933b9f3a106aa0932251e5f3`.

The two P1 strict-recovery issues are still current, but they are already covered by the existing NumaryBot/flemzord review threads, so I am not duplicating them here.

One additional low-severity cleanup: a few comments still describe the old panic-based or non-strict contract.

- `cache_snapshotter.go` still documents `parseLeanValue` as panicking immediately before the new error-returning contract.
- Related stale/misplaced test comments remain around the cache snapshotter strict-recovery tests.

The code behavior is clear, but those comments make the strict recovery contract ambiguous and should be aligned with the new fail-closed behavior.

Checks run:

- `git diff --check 29742f6...06eed28`
- `GOROOT= go test ./internal/infra/state ./internal/infra/bloom`
- `GOROOT= go build ./...`
- `gh pr checks 1566 --repo formancehq/ledger`

internal/infra/state/cache_snapshotter.go
gfyrag: Fixed in ffe850fcb. You are right: `hasPersistedBloomBlocks` scans the whole `[ZoneGlobal][SubGlobBloom]` range and returns true for *any* row, but `FilterSet.RestoreFromStore` only iterated the enabled per-attribute sub-ranges — so a block persisted for a disabled/unknown attr code (or a row too short to carry an attrCode) was silently skipped while boot still took the persisted-block path and published a ready filter with a false negative.

Added `FilterSet.validatePersistedNamespace`, run before the per-filter restore: it classifies every row under `[SubGlobBloom]` and fails closed on any attrCode that maps to no enabled filter, or any row shorter than `[zone][sub][attrCode]`. Regression tests: `TestFilterSet_EN1527_RestoreRejectsOrphanAttrCode`, `TestFilterSet_EN1527_RestoreRejectsShortBloomRow`, and a positive `...RestoreAcceptsEnabledAttrRows`.

internal/infra/state/cache_snapshotter.go
gfyrag: Fixed in ffe850fcb. Correct — the exact-length check only fired inside the per-slot iterators, so a row whose generation or attribute-type byte drifted outside the known `[gen][slotCode]` ranges was never visited and silently ignored on restore, dropping a cache entry written since the last Bloom flush.

Added `validateCacheNamespace`, a single classification pass over the whole `[ZoneCache]` namespace run *before* `restoreGeneration`. It fails closed on anything that is not one of the three shapes the persist path emits:
- `[ZoneCache][SubCacheMeta]` (global meta),
- `[ZoneCache][gen][SubCacheGenMeta]` (per-gen meta),
- `[ZoneCache][gen][slotCode][U128]` (cache entry),

where `gen` must be one of the two live generation bytes and `slotCode` a registered slot. The valid slot set is derived from `s.slots` so it stays in lock-step with persist (no second hardcoded list). Regression tests: `TestCacheSnapshotter_EN1527_RestoreRejectsUnknownAttributeType` and `...RestoreRejectsUnknownGenerationByte`.

internal/infra/state/cache_snapshotter.go
gfyrag: Fixed in ffe850fcb. Collapsed the two doc blocks into a single accurate contract for the error-returning implementation — the stale `parseLeanValue panics on malformed input` wording is gone.

internal/infra/state/cache_snapshotter.go
```diff
@@ -575,7 +698,16 @@ func (s *CacheSnapshotter) restoreGeneration(reader dal.PebbleReader, genByte by
...
+	hasBlocks, err := s.hasPersistedBloomBlocks(store)
+	if err != nil {
+		// A probe failure is NOT first-boot: silently falling back to a full
+		// attribute scan would republish a "ready" filter from a database whose
+		// persisted blocks we could not even read, masking corruption. Fail
+		// recovery so readiness stays false (EN-1527).
+		return fmt.Errorf("probing persisted bloom blocks: %w", err)
+	}
+
+	if hasBlocks {
```
NumaryBot: 🔴 **[blocker] Malformed/unknown Bloom prefix rows trigger persisted-block restore path without being decoded**

When the Bloom range contains a malformed key not visited by any enabled per-attribute iterator (e.g. unknown attribute code, truncated key, or disabled attribute byte), `hasPersistedBloomBlocks` returns true and boot takes the persisted-block restore path. `FilterSet.RestoreFromStore` then silently skips that row, `runBloomTaskBody` replays only cache generations and marks the filter ready. This can publish an incomplete Bloom with false negatives instead of failing closed, directly undermining the fail-closed recovery goal of this PR.

**Suggestion:** Before returning true from `hasPersistedBloomBlocks`, classify every row in the Bloom namespace. Any row whose key does not fully parse against a known/enabled attribute code and expected length should either abort recovery (fail-closed) or force a full-scan rebuild path. Do not let an unclassified row silently select the persisted-block restore branch.

internal/infra/state/cache_snapshotter.go
```diff
@@ -529,9 +569,15 @@ func (s *CacheSnapshotter) restoreGeneration(reader dal.PebbleReader, genByte by
...
 			key := iter.Key()
-			// Key format: [0xFF][gen][type][16-byte U128]
-			if len(key) < 3+16 {
-				continue
+			// Key format: [0xFF][gen][type][16-byte U128]. Require the exact
+			// shape — a short key would previously be skipped (dropping a live
+			// cache entry) and a longer key silently truncated to 16 bytes.
+			// Either is a corrupt row; fail closed rather than restore partial
+			// cache state (EN-1527).
+			if len(key) != 3+16 {
```
NumaryBot, line 577: 🔴 **[blocker] Exact-length check only runs inside per-slot iterators, leaving unknown/corrupt generation rows undetected**

The per-slot iterators only validate keys they already expect. If corruption or a format change causes a `[ZoneCache][gen]` row to have an attribute-type byte or generation value outside the known set, recovery silently skips the row. On the persisted-Bloom restart path this can drop a cache entry written since the last Bloom flush from `replayBloomFromCache`, then publish a ready filter with a false negative — exactly the invariant this PR is meant to close.

**Suggestion:** Add a single linear pass over the entire `[ZoneCache][gen]` namespace that classifies every row. Any row with an unknown or malformed generation/type byte should cause recovery to fail closed (return an error) rather than be silently skipped. Add a regression test covering an unknown generation/type byte.

internal/infra/bloom/bloom.go
```diff
@@ -530,6 +549,56 @@ func (fs *FilterSet) RestoreFromStore(ctx context.Context, store dal.PebbleReade
...
+
+		key := it.Key()
+		// Key format: [ZoneGlobal][SubGlobBloom][attrCode][blockIndex BE 8].
+		// The attrCode lives at offset 2; a shorter row cannot belong to any
+		// filter's sub-range.
+		if len(key) < 3 {
+			return fmt.Errorf("persisted bloom row %x is too short to carry an attribute-type byte — store corrupted", key)
+		}
+
+		if _, ok := enabled[key[2]]; !ok {
```
NumaryBot: 🔴 **[blocker] Config-drift rows treated as corruption, preventing boot and config-update proposal**

When a node restarts with CLI Bloom settings that disable an attribute type that was enabled in the previously persisted cluster config, `FilterSet` is built from the new CLI config before recovery. Stored Bloom blocks still reflect the old config until `proposeClusterConfigIfNeeded` can apply a purge/rebuild. The strict namespace validation treats those old-but-valid rows as corruption and aborts `RestoreCacheFromStore`, so the node can never boot far enough to propose the config update.

**Suggestion:** Distinguish malformed/unknown rows (true corruption) from rows belonging to a previously-valid persisted config (config drift). On config drift, force the full-rebuild path instead of aborting. This may require reading the persisted cluster config before constructing `FilterSet` for recovery, so that the validator knows which attribute codes were legitimately written.

internal/infra/state/cache_snapshotter.go
```diff
@@ -22,17 +23,20 @@ import (
...
-// live would let either case resurrect deleted keys.
-func parseLeanValue(value []byte) (tag uint64, deleted bool, valueBytes []byte) {
+// parseLeanValue decodes a persisted cache lean value from the ZoneCache
+// namespace. Lean format: [8-byte tag LE][1-byte flag][raw value bytes]. It
+// returns a contextual error — never panics or silently accepts — on any
+// malformed row (short buffer, unknown flag byte, or a tombstone carrying
+// trailing payload), so recovery fails closed rather than restoring partial or
+// attacker-forged cache state (EN-1527). Every row is produced by
+// writeCacheRaw, so any other shape means a corrupted store or a
+// forward-incompatible binary, and silently accepting it would let a one-byte
```
NumaryBot, line 33: 🟡 **[minor] Stale doc comment still describes panic-based contract for `parseLeanValue`**

The comment immediately preceding `parseLeanValue` still documents it as panicking on malformed input, while the new implementation explicitly returns an error. Having both descriptions in place makes the strict recovery contract ambiguous to future callers and reviewers.

**Suggestion:** Replace the stale panic-contract comment with an accurate doc comment describing the error-returning, fail-closed behaviour.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1566#issuecomment-4949244008

internal/infra/state/cache_snapshotter.go
gfyrag: Addressed in 5e55ae79c. `restoreBloomFilters` now classifies the entire `[SubGlobBloom]` namespace *before* choosing the restore path (via `FilterSet.ClassifyPersistedNamespace`): an unknown attribute-type byte or a row too short to carry one fails closed; a known-but-disabled attr is config drift and routes to the full async rebuild. So a malformed/unknown row can no longer select the persisted-block restore branch and publish a ready-but-incomplete filter. Regression: `TestFilterSet_EN1527_ClassifyRejectsUnknownAttrCode`, `...ClassifyRejectsShortBloomRow`.

internal/infra/state/cache_snapshotter.go
gfyrag: Addressed in ffe850fcb. `validateCacheNamespace` runs a single linear pass over the whole `[ZoneCache]` namespace before `restoreGeneration` and fails closed on any row that is not one of the three shapes the persist path emits (global meta, per-gen meta, or `[gen][registered-slot][U128]`), where `gen` must be one of the two live generation bytes and the slot code must be registered (set derived from `s.slots`). A row with an unknown attribute-type byte or an out-of-range generation byte now aborts recovery instead of being silently skipped. Regression: `TestCacheSnapshotter_EN1527_RestoreRejectsUnknownAttributeType` and `...RestoreRejectsUnknownGenerationByte`.

internal/infra/bloom/bloom.go
gfyrag: Good catch — legitimate, and my first pass (ffe850fcb) did introduce exactly this regression by treating any orphan row as corruption. Fixed in 5e55ae79c by distinguishing the two cases:

- **corruption** (attribute-type byte that is not a known bloom type, or a row too short to carry one) -> fail closed;
- **config drift** (a KNOWN attribute type that the current CLI config disables) -> tolerated, and `restoreBloomFilters` takes the full async rebuild path (`StartAsyncBloomPopulate`) instead of trusting the stale persisted blocks — without aborting, so the node boots and can reach `proposeClusterConfigIfNeeded` to purge/rebuild.

The known-code set comes from `bloomTypes(nil)` so it tracks the type registry with no second hardcoded list. Regression: `TestFilterSet_EN1527_ClassifyReportsConfigDrift` (unit) and `TestCacheSnapshotter_EN1527_RestoreToleratesConfigDrift` (boots via async populate over config-A blocks after config B disables Metadata).

internal/infra/state/cache_snapshotter.go
gfyrag: Already addressed in ffe850fcb — the doc comment above `parseLeanValue` now describes only the error-returning, fail-closed contract; the stale panic wording is gone. This finding was raised against an earlier snapshot.

gfyrag:
All findings from the automated "Changes requested" review are now addressed at head `5e55ae79c`; every inline thread has been replied to and resolved.

**Blocker 1 — malformed/unknown Bloom rows select the restore path** → fixed in `ffe850fcb` + `5e55ae79c`. `restoreBloomFilters` classifies the whole `[SubGlobBloom]` namespace via `FilterSet.ClassifyPersistedNamespace` **before** choosing a path. Unknown attribute-type byte or a too-short row fails closed; a malformed/unknown row can no longer reach `runBloomTaskBody` and publish a ready-but-incomplete filter.

**Blocker 2 — unknown/corrupt `[ZoneCache][gen]` rows silently skipped** → fixed in `ffe850fcb`. `validateCacheNamespace` runs one linear pass over the entire cache namespace before `restoreGeneration` and fails closed on any unknown generation/attribute-type byte or malformed shape (valid slot set derived from `s.slots`).

**Blocker 3 — strict validator cannot distinguish corruption from config drift, node unbootable after an attribute-type change** → fixed in `5e55ae79c`. This regression was introduced by my blocker-1 pass and is now corrected: corruption (unknown bloom attr type / unparseable row) fails closed, while a **known-but-disabled** attribute is treated as config drift and routes to the full async rebuild (`StartAsyncBloomPopulate`) instead of aborting — so the node boots and can reach `proposeClusterConfigIfNeeded` to purge/rebuild the stale blocks. Covered by `TestCacheSnapshotter_EN1527_RestoreToleratesConfigDrift`.

**Doc nit — stale `parseLeanValue` panic contract** → fixed in `ffe850fcb`; the comment now describes only the error-returning, fail-closed behaviour.

`GOROOT= go build ./...`, `go test ./internal/infra/state/... ./internal/infra/bloom/...`, and `golangci-lint run` on both packages all pass. Requesting a re-review to clear the "Changes requested" state.

internal/infra/bloom/bloom.go
```diff
@@ -530,6 +537,103 @@ func (fs *FilterSet) RestoreFromStore(ctx context.Context, store dal.PebbleReade
...
+		key := it.Key()
+		// Key format: [ZoneGlobal][SubGlobBloom][attrCode][blockIndex BE 8].
+		// The attrCode lives at offset 2; a shorter row cannot belong to any
+		// filter's sub-range and is unparseable — treat as corruption.
+		if len(key) < 3 {
+			return false, fmt.Errorf("persisted bloom row %x is too short to carry an attribute-type byte — store corrupted", key)
+		}
+
+		attrCode := key[2]
+		if _, ok := enabled[attrCode]; ok {
```
NumaryBot, line 617: 🟠 **[major] Config-drift fallback masks corrupt rows for enabled attr codes**

When the Bloom namespace contains both a known-but-disabled attr row (config drift) and a malformed row for an enabled attr (e.g. wrong-length value), the `continue` on the drift row causes the enabled row to go un-decoded during classification. `configDrift` then triggers the async full-scan rebuild path, silently bypassing `RestoreFromStore` and masking the corrupt enabled block instead of failing closed. This directly undermines the fail-closed invariant the PR is meant to enforce.

**Suggestion:** Separate the config-drift classification from the per-row shape/value validation. Validate every row's structure (key length, value length, encoding) regardless of whether its attr code is currently enabled or disabled. Only after confirming no row is malformed should a config-drift rebuild fallback be allowed. Rows with truly unknown/corrupt structure should still cause recovery to return an error.

internal/infra/state/cache_snapshotter.go
NumaryBot, line 33: 🟡 **[minor] Stale doc comment still describes panic-based contract for `parseLeanValue`**
<sub>reported by flemzord, gfyrag, NumaryBot</sub>

The comment immediately preceding `parseLeanValue` still documents it as panicking on malformed input, while the new implementation explicitly returns an error. Prior threads indicate this was addressed in ffe850fc, but at the current head (06eed28f) the stale wording is reportedly still present, making the strict recovery contract ambiguous to future callers.

**Suggestion:** Replace the stale panic-contract comment with an accurate doc comment describing the error-returning, fail-closed behaviour.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1566#issuecomment-4949244008

gfyrag:
Rechecked current head `5e55ae79c080d3eef1f799d13e852e208eb688d8`. The namespace validation/config-drift fixes cover the cases where a `FilterSet` exists, but the all-disabled Bloom transition still leaves stale persisted blocks behind.

`machine_technical_updates.go` only purges `[ZoneGlobal][SubGlobBloom]` when `fsm.BloomFilters != nil && !BloomConfigEqual(...)`. When the effective config disables every Bloom type, `NewFilterSet` returns nil, so an apply that persists the disabled Bloom config can skip the purge entirely. The old Bloom blocks remain on disk even though no Bloom updates are recorded while Bloom is disabled.

If Bloom is later re-enabled for the same attr codes, recovery can see those old blocks as valid persisted blocks and publish a ready filter from them plus cache replay, but the blocks miss all attributes written during the disabled interval. That reintroduces the false-negative path this PR is meant to fail closed on: `resolve.go` may trust `IsReady()` and skip Pebble reads for keys whose writes happened while Bloom was disabled.

The purge of persisted Bloom blocks should not be gated on the in-memory `FilterSet` being non-nil. A config change that disables all Bloom filters still needs to delete the persisted Bloom namespace, and re-enable should create/rebuild a fresh FilterSet rather than reuse stale blocks. Please add a regression for enabled -> all-disabled -> writes while disabled -> re-enabled, asserting stale persisted blocks cannot be restored as ready.

internal/infra/bloom/bloom.go
NumaryBot, line 617: 🟠 **[major] Config-drift fallback masks corrupt rows for enabled attr codes**

When the Bloom namespace contains both a known-but-disabled attr row (config drift) and a malformed row for an enabled attr (e.g. wrong-length value), the `continue` on the drift row causes the enabled row to go unvalidated during classification. The `configDrift` flag then triggers the async full-scan rebuild path, silently bypassing `RestoreFromStore` and masking the corrupt enabled block instead of failing closed. This directly undermines the fail-closed invariant the PR is meant to enforce.

**Suggestion:** Separate the config-drift classification from the per-row shape/value validation. Validate every row's structure (key length, value length, encoding) regardless of whether its attr code is currently enabled or disabled. Only after confirming no row is structurally malformed should the config-drift rebuild fallback be allowed. Rows with truly unknown or corrupt structure should still cause recovery to return an error.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1566#issuecomment-4949244008

flemzord (approved):
No High or Critical findings remain on this head. The remaining Bloom namespace strictness concern takes the full attribute-rebuild path while the filter is not ready, so it does not expose an incomplete filter and is non-blocking under the current severity policy.

PR #1464, opened 2026-07-01
feat(EN-1334): usagebuilder subsystem + event-counter cutover (EN-1420)

gfyrag:
## Summary

- **EN-1334** — new `usagebuilder` subsystem materialises per-Numscript-template invocation counters (count + lastUsed) from the FSM audit chain into a dedicated peer secondary Pebble store (`usagestore`). Exposed via `GET /v3/{ledger}/numscripts/{name}/usage`.
- **EN-1420 / EN-1422** — the per-ledger stat counters migrate out of `LedgerBoundaries` into `usagestore`. This is the **full cutover**: the event counters (posting, revert, numscript_execution, reference) AND the attribute-derived counters (Volume, EphemeralEvicted, TransientUsed) now live in `usagestore`; `MetadataCount` is **dropped** for now (no sound foundation — see below). The FSM increment sites are removed and the proto fields deleted + reserved. `LedgerStats` sources every migrated counter from `usagestore`.

## Architecture notes

- The usagebuilder tails the **audit chain** (AuditEntry + AuditItem), not the log stream, because the log's `CreatedTransaction` payload doesn't carry `NumscriptReference` — the raw order in `AuditItem.SerializedOrder` does. When a posting count is needed (revert / script-backed create) the specific log is fetched via `ReadLogBySequence` — a single point read per item on hot Pebble cache.
- Volume-count derivation rides on three disjoint per-log annotation lists on `LedgerLog` — `NewKeptVolumes` (+1), `PurgedVolumes` (−1), `EphemeralVolumes` (0, tracked for `EphemeralEvictedCount`) — split this way (vs. the earlier 2-list encoding) to keep ephemeral-heavy workloads from paying 2× bytes on the log payload (EN-1422).
- Cursor semantics: last consumed audit sequence, persisted in usagestore at `[0xFE][0x01]`. LogCommitted notification wakes the loop; the audit chain advances in the same batch as the log. If the primary store is rolled back beneath the cursor, the builder self-heals: it wipes the projection and replays from sequence 0 (`usagestore.Store.Reset()` on boot).
- Physical separation: `usagestore` opens its own Pebble instance at `<data-dir>/usage/` (peer to `readstore`). Rebuild = drop the directory (or `Reset()`); the builder replays from cursor=0.
- Notifications wiring uses **constructor injection** (per `feedback_constructor_injection` in project memory), and the existing `signal.FanOut` is extended to a 4th target.

## Cutover semantics (EN-1420 / EN-1422)

[...]

PR #1568, opened 2026-07-12
fix(indexbuilder): detect primary-store rollback via restore generation

gfyrag:
## Stacked on #1464

**This PR is stacked on #1464** (`feat/EN-1334-usagebuilder`) — it needs `dal.Store.RestoreGeneration()` and the `restoreGeneration` bump in `RestoreCheckpoint`, which exist only on that branch. Base is set to `feat/EN-1334-usagebuilder` to keep this diff limited to the indexbuilder change. **Retarget to `release/v3.0` after #1464 merges.**

## The gap

The read-side index builder (`internal/application/indexbuilder`) had **no primary-store rollback detector at all** — unlike the usage builder and audit indexer, which #1464 gave the restore-generation watermark. It was missing **both**:

- the boot-time `cursorAheadOfHead` guard, and
- the runtime generation watermark.

After a `RestoreCheckpoint` rolled the primary log head back below the persisted read-index cursor, the builder resumed its incremental catch-up from the stale cursor and **skipped every surviving (lower-seq) log forever** — the rolled-back-then-regrown gap was never re-indexed. This is a confirmed, currently-open primary-store-rollback catch-up gap.

## The fix (parity with auditindexer / usagebuilder)

**Boot guard** — `bootInit` now wipes+rebuilds the read index when:
- the persisted cursor overtakes the log head (`cursorAheadOfHead` — the direct rollback signature), or
- the head leads the cursor by more than `rebuildThreshold` (the boot-only gap net for a rollback whose restored gap re-grew before the head was sampled).

**Runtime watermark** — a `restoreGen atomic.Uint64` seeded at boot and re-checked at the top of every steady-state tick (`maybeResetOnRestore`). A `dal.Store.RestoreGeneration()` change is the **authoritative** rollback signal: it fires on any `RestoreCheckpoint` regardless of where the head landed, so it catches the **catch-up race** that erases the position signal (restore below cursor → head re-grows past the old cursor before the next tick re-samples → `cursorAheadOfHead` is inert).

## Heavier reset path (boot-equivalent re-init)

A read-index rebuild is **not** a bare `cursor = 0`. The reset routes through a boot-equivalent re-init:

- `readstore.ResetIndexes()` wipes the **builder-owned** keyspace only — ledger-scoped index rows (prefixes `0x01..0x0D`), the log/applied-proposal progress cursors, backfill cursors, and per-index version state. It deliberately **leaves the auditindexer-owned audit index (`0xFE05`/`0xFE06`) untouched** — that worker has its own rollback detector.
- `resetAndReinit` re-syncs `restoreGen`, then re-runs `initIndexConfig` against the now-empty store, so index config, the per-replica version cache, and backfill/rewrite tasks are all re-derived and backfills replay from scratch.

Blast radius stayed contained: the boot init + reset machinery `initIndexConfig` already provided was reusable as-is; no wider refactor was needed. `NewBuilder` gained one `rebuildThreshold uint64` param, wired from `cfg.AuditIndexConfig.RebuildThreshold` (same knob the sibling workers use); the offline CLI rebuild passes `0`.

## Invariants

Poller/side-store path — the read index lives in the readstore (separate Pebble instance), so no FSM hot-path Pebble reads are involved. Reset is deterministic; the reset error paths log-and-continue in the loop (matching the existing catch-up error handling), and the reset helper returns errors loudly.

## Tests

`internal/application/indexbuilder/builder_restore_test.go`:
- `TestMaybeResetOnRestore_CatchUpRace` — the exact race: cursor advanced to 3, `RestoreCheckpoint` rolls head back, head re-grows to 5 (**past** the old cursor). Proves `shouldRebuildOnBoot`/`cursorAheadOfHead` stay inert while the generation gate forces a reset+rebuild, and that a second tick does not re-reset (generation resync). Mirrors `usagebuilder.TestTickResetsOnRestoreGenerationChange`.
- `TestBootInit_ResetsOnPrimaryStoreRollback` — real `RestoreCheckpoint` with head below cursor; boot wipes and returns cursor 0.
- `TestResetIndexes_WipesBuilderKeyspaceButNotAuditIndex` — pins the reset boundary (audit index + cursor survive).
- `TestShouldRebuildOnBoot` — boot guard decision table.

## Verification
- `GOROOT= go build ./...` clean (also full-feature build with all tags).
- `GOROOT= go test ./internal/application/indexbuilder/... ./internal/storage/readstore/...` green.
- `go vet` + `golangci-lint run --fix` on the changed packages: 0 issues.

### DIFF
--- a/internal/application/indexbuilder/builder.go
+++ b/internal/application/indexbuilder/builder.go
@@ -46,6 +46,25 @@ type Builder struct {
 	metricsRegistration     metric.Registration // tailworker gauge triplet
 	logsIndexedRegistration metric.Registration // builder-specific logs_indexed_total gauge
 
+	// rebuildThreshold forces a boot-time reset+rebuild when the log head leads
+	// the persisted cursor by more than this many entries (0 disables the
+	// gap-based net). Boot-only, mirroring auditindexer.shouldRebuildOnBoot and
+	// usagebuilder.gapExceedsThreshold: it is the secondary catch of a rollback
+	// whose restored gap re-grew past the cursor before pebbleLast was sampled,
+	// so the direct cursorAheadOfHead signature was never observable.
+	rebuildThreshold uint64
+
+	// restoreGen is the primary-store restore generation this builder last
+	// processed under (see dal.Store.RestoreGeneration). Seeded on boot, re-read
+	// every steady-state tick: a change means the primary store was rolled back
+	// beneath the cursor at runtime, so a full reset+rebuild is forced regardless
+	// of where the log head landed relative to the cursor — the position-based
+	// cursorAheadOfHead / rebuildThreshold signals can be erased by a catch-up
+	// race (rollback below cursor, head re-grows past the old cursor before the
+	// next tick re-samples). Mirrors auditindexer.Indexer.restoreGen and
+	// usagebuilder.Builder.restoreGen.
+	restoreGen atomic.Uint64
+
 	// Per-ledger index configuration cache.
 	indexConfig map[string]*ledgerIndexConfig
 
@@ -324,6 +343,10 @@ func (b *Builder) coerceForLedger(ledger string, target commonpb.TargetType, key
 // NewBuilder creates a new index builder.
 // batchSize controls how many log entries are buffered per Pebble batch commit.
 // Use 0 for the default (DefaultBatchSize).
+// rebuildThreshold forces a boot-time reset+rebuild when the log head leads the
+// persisted cursor by more than this many entries (0 disables the gap net); it
+// is the boot-only secondary net for a primary-store rollback (see the
+// rebuildThreshold field comment).
 func NewBuilder(
 	pebbleStore *dal.Store,
 	readStore *readstore.Store,
@@ -331,23 +354,25 @@ func NewBuilder(
 	logger logging.Logger,
 	meter metric.Meter,
 	batchSize int,
+	rebuildThreshold uint64,
 ) *Builder {
 	if batchSize <= 0 {
 		batchSize = DefaultBatchSize
 	}
 
 	return &Builder{
-		pebbleStore:    pebbleStore,
-		readStore:      readStore,
-		attrs:          attrs,
-		logger:         logger.WithFields(map[string]any{"cmp": "index-builder"}),
-		meter:          meter,
-		batchSize:      batchSize,
-		backfillBudget: 50 * time.Millisecond,
-		indexConfig:    make(map[string]*ledgerIndexConfig),
-		kb:             dal.NewKeyBuilder(),
-		wb:             readstore.NewWriteBatch(),
-		accounts:       make(map[string]struct{}, 64),
+		pebbleStore:      pebbleStore,
+		readStore:        readStore,
+		attrs:            attrs,
+		logger:           logger.WithFields(map[string]any{"cmp": "index-builder"}),
+		meter:            meter,
+		batchSize:        batchSize,
+		rebuildThreshold: rebuildThreshold,
+		backfillBudget:   50 * time.Millisecond,
+		indexConfig:      make(map[string]*ledgerIndexConfig),
+		kb:               dal.NewKeyBuilder(),
+		wb:               readstore.NewWriteBatch(),
+		accounts:         make(map[string]struct{}, 64),
 	}
 }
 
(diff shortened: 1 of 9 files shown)


internal/application/indexbuilder/builder.go
```diff
@@ -620,5 +655,105 @@ func (b *Builder) bootInit(ctx context.Context) (cursor uint64, pebbleLast uint6
...
+// re-derives all in-memory init state (index config, per-replica version cache,
+// backfill/rewrite tasks) from the now-empty store, so a rebuild replays from
+// log sequence 0. It re-syncs restoreGen up-front so a RestoreCheckpoint that
+// lands during the subsequent replay bumps past this value and is re-detected
+// on the next tick (rather than being silently swallowed). Returns the reset
+// cursor (always 0). Mirrors usagebuilder.resetProjection, adapted for the
+// heavier read-index reset: a bare cursor=0 is not enough — the index rows,
+// backfill cursors, and version state must be wiped and re-seeded too.
+func (b *Builder) resetAndReinit(ctx context.Context) (uint64, error) {
+	b.restoreGen.Store(b.pebbleStore.RestoreGeneration())
```
NumaryBot, line 715: 🟠 **[major] Defer restore generation sync until reset succeeds**

If `ResetIndexes()` or the subsequent `initIndexConfig()` fails after a runtime `RestoreCheckpoint`, this stores the new generation before returning the error, so later ticks see `gen == restoreGen` and never retry the reset. In that scenario the loop logs the error and continues from the stale in-memory cursor, which can permanently skip the surviving lower-sequence logs the rollback detector was meant to rebuild.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1568#issuecomment-4952642382

internal/application/indexbuilder/builder.go
NumaryBot, line 715: 🟠 **[major] Restore generation updated before reset succeeds, silently disarming rollback detector**

After a runtime RestoreCheckpoint is detected, the new restore generation is stored before ResetIndexes() or initIndexConfig() completes. If either of those calls returns an error, the loop will log the error and continue from a stale in-memory cursor. On the next tick, gen == restoreGen, so the rollback is never retried. This permanently skips lower-sequence logs that the detector was designed to rebuild.

**Suggestion:** Only persist/update the restore generation after the full reset pipeline (ResetIndexes + initIndexConfig) completes successfully. Move the generation update to after the error check so a failed reset is retried on the next tick.

internal/application/indexbuilder/builder.go
```diff
@@ -620,5 +655,105 @@ func (b *Builder) bootInit(ctx context.Context) (cursor uint64, pebbleLast uint6
...
 
+	// Primary-store rollback guard (parity with auditindexer.shouldRebuildOnBoot
+	// and usagebuilder.rewindOnRollback boot path). When the persisted cursor
+	// overtakes the log head — the direct signature of a restore/rollback to an
+	// earlier head while the read index was retained — or the head leads the
+	// cursor by more than rebuildThreshold, the retained index rows reflect logs
+	// that no longer exist. An incremental catch-up from the stale cursor would
+	// scan past every surviving (lower-seq) log forever, so wipe the read index
+	// and rebuild from sequence 0.
+	if b.shouldRebuildOnBoot(cursor, pebbleLast) {
```
NumaryBot, line 666: 🟠 **[major] Boot rollback guard destroys read index when primary head sample fails**

If ReadLastSequence returns an error, pebbleLast remains at its zero value. The subsequent guard then incorrectly interprets any non-zero persisted cursor as being ahead of head (cursorAheadOfHead), triggering a full wipe of the read index even though no actual rollback occurred. A transient or corrupt iterator error is now sufficient to destroy the index.

**Suggestion:** Gate the rebuild/wipe check on a successful head read. If ReadLastSequence returns an error, either return the error immediately or skip the rollback guard entirely for that iteration, rather than proceeding with a zero pebbleLast value.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1568#issuecomment-4952642382

gfyrag:
Closing: this PR (and the restore-generation watermark it extends from #1464) defends against a rollback that cannot occur. The FSM applies only committed Raft entries (`rd.CommittedEntries` → applier → `applyProposal` → `writeAuditEntry`; node.go:1404-1411), so the audit chain and log are append-only and monotone by Raft leader-completeness — once written, an entry survives every leadership change. `RestoreCheckpoint` only ever advances a node that is behind (`IsStoreUpToDate = LastAppliedIndex >= SnapshotIndex`). A projection cursor therefore cannot legitimately end up ahead of the audit/log head after a restore, so the catch-up race is unreachable. Reverting the watermark on #1464 too.

PR #1569, opened 2026-07-12
refactor(EN-1334): drop impossible primary-store-rollback detection from auditindexer

gfyrag:
## Summary

Removes dead primary-store-rollback detection from the audit indexer. This mirrors the identical removal just made on the usagebuilder (PR #1464, commit b09f82457).

## Why the rollback guard is unreachable

The audit chain is **append-only and monotone**:

- `writeAuditEntry` runs **only** inside `Machine.applyProposal` (`internal/infra/state/machine.go`).
- The applier feeds `applyProposal` **exclusively** from committed Raft entries: `node.applier.Submit(rd.CommittedEntries, …)` (`internal/infra/node/node.go:1411`), and etcd/raft only emits `MsgStorageApply` when `CommittedEntries > 0` (`node.go:1404-1407`).
- By **Raft leader-completeness**, a committed entry is present on every future leader — once an audit entry is written it can never be un-written or superseded.
- `RestoreCheckpoint` only ever **advances** a node that is behind: sync runs solely when `IsStoreUpToDate` is false, i.e. `LastAppliedIndex < SnapshotIndex` (`internal/infra/state/synchronizer.go:148`), and the leader's checkpoint is at an index ≥ ours.

Therefore the persisted audit-index cursor can **never legitimately sit ahead of the audit head**, and the "cursor overtook head after a rollback" scenario is unreachable.

## What was removed

- `cursorAheadOfHead(cursor, last)` and both its call sites: the boot-time branch in `shouldRebuildOnBoot` and the per-tick re-check + `Rebuild` call in `processTick`.
- The `RebuildThreshold` gap branch in `shouldRebuildOnBoot` (`i.cfg.RebuildThreshold > 0 && last > cursor && last-cursor > i.cfg.RebuildThreshold`). It was framed as a "rollback net", but a rebuild-from-0 re-indexes `[0,cursor]` on top of the `[cursor,last]` range that incremental catch-up already covers — it is **strictly more work** than catch-up, not a perf optimization. With rollback impossible its only purpose is void.
- The now-inert `RebuildThreshold` config field (`AuditIndexConfig.RebuildThreshold`), its `--audit-index-rebuild-threshold` CLI flag (`cmd/server/server.go`), and the `docs/ops/cli.md` entry. auditindexer was its only consumer, and the surface is self-contained (one flag, one config field, one wiring site, one docs line), so it was removed rather than left inert.
- Obsolete tests: `TestBootRebuildOnStaleCursor`, `TestProcessTickRebuildsWhenCursorAheadOfHead`, and the `RebuildThreshold`/cursor-ahead cases in `TestShouldRebuildOnBoot`.

## What was kept

- The legitimate **fresh-build** case in `shouldRebuildOnBoot`: `cursor == 0 && last > 0` (empty cursor, entries exist → build the index). This is **not** rollback detection.
- `TestShouldRebuildOnBoot` retains coverage of that case, and now asserts a lagging cursor is resumed incrementally (never rebuilt from 0).
- `processTick` still refreshes the audit-head gauge and calls `ProcessOnce`.

## Related

- Companion usagebuilder removal: PR #1464, commit `b09f82457`.
- PR #1568 (the indexbuilder watermark) and the #1464 restore-generation watermark were reverted for the same monotonicity reason.

## Verification

- `GOROOT= go test ./internal/application/auditindexer/... -count=1` — green
- `go vet` + `golangci-lint run --fix` on the changed packages — 0 issues

### DIFF
--- a/cmd/server/server.go
+++ b/cmd/server/server.go
@@ -228,7 +228,6 @@ func NewRunCommand() *cobra.Command {
 
 	// Audit index configuration
 	runCmd.Flags().Int("audit-index-batch-size", 0, "Audit entries per Pebble batch commit (0 = default 1000)")
-	runCmd.Flags().Uint64("audit-index-rebuild-threshold", 0, "Drop+rebuild the audit index on boot when the cursor is this far behind (0 = never)")
 	runCmd.Flags().Bool("disable-audit-index", false, "Disable the audit secondary index worker")
 
 	// Query profiling
@@ -663,12 +662,10 @@ func LoadConfig(ctx context.Context, cmd *cobra.Command) (*bootstrap.Config, err
 
 	// Audit index configuration
 	auditBatchSize, _ := cmd.Flags().GetInt("audit-index-batch-size")
-	auditRebuildThreshold, _ := cmd.Flags().GetUint64("audit-index-rebuild-threshold")
 	auditDisabled, _ := cmd.Flags().GetBool("disable-audit-index")
 	cfg.AuditIndexConfig = bootstrap.AuditIndexConfig{
-		BatchSize:        auditBatchSize,
-		RebuildThreshold: auditRebuildThreshold,
-		Disabled:         auditDisabled,
+		BatchSize: auditBatchSize,
+		Disabled:  auditDisabled,
 	}
 
 	// Query profiling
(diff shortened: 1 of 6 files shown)


flemzord (approved):
Reviewed the full diff and the current discussion on dd79edb2628bd6edd09a9ce5b1a9df6b91c210a2. I found no High or Critical issues; the rollback guard and threshold removal are consistent with the follower-sync monotonicity contract, and the configuration/docs cleanup is complete.

PR #1570, opened 2026-07-13
fix(EN-1509): preserve typed metadata in HTTP revert-transaction requests

gfyrag:
## Summary

Ref EN-1509 (Medium bug — "Preserve typed metadata values in HTTP revert-transaction requests").

`RevertTransactionPayload.metadata` is a `map<string, common.MetadataValue>` (typed), but `internal/adapter/http/handlers_revert_transaction.go` decoded the JSON `metadata` field as `map[string]any`, kept **only** string values in a `map[string]string`, and converted via `commonpb.MetadataFromGoMap`. Typed (numeric / boolean) values a client sent were **silently dropped** — a client got a successful revert with metadata missing (data-loss + parity bug).

## Fix

- Decode `metadata` through the shared typed-metadata path `commonpb.MetadataFromAnyMap` — the same decoder the set-metadata HTTP handlers use (`parseMetadataBody`) and consistent with create-transaction's typed proto-JSON codec. No new decoder was forked.
- All supported `MetadataValue` kinds (string, integer, boolean) are mapped losslessly to `RevertTransactionPayload.Metadata`.
- Unsupported values (objects, arrays, non-integer floats) are now **rejected with `400 INVALID_REQUEST`** instead of being silently dropped.

## Reused decoder

`internal/proto/commonpb/metadata.go` — `MetadataFromAnyMap` (built on `MetadataValueFromAny`), the exact path behind `parseMetadataBody` used by `handleSaveTransactionMetadata` / `handleSaveAccountMetadata`.

## Files

- `internal/adapter/http/handlers_revert_transaction.go` — typed decode + 400 on invalid.
- `internal/adapter/http/handlers_revert_transaction_test.go` — tests: string, numeric (uint + negative int), boolean, date-as-string (all asserted losslessly on the forwarded payload); invalid (nested object) → 400 INVALID_REQUEST.
- `docs/technical/contributing/api-comparison.md` — note typed-metadata support + 400 behavior on revert.

## OpenAPI

`openapi.yml` already documented `RevertTransactionRequestBody.metadata` as `additionalProperties: $ref MetadataValue` (typed). The spec was correct; the handler had diverged from it — this restores parity. No schema change needed.

## Verification

- `GOROOT= go build ./...` — OK
- `GOROOT= go test ./internal/adapter/http/... -count=1` — OK
- `go generate ./... && go mod tidy && golangci-lint run --fix` — tree clean (only the 3 intended files changed)

### DIFF
--- a/internal/adapter/http/handlers_revert_transaction.go
+++ b/internal/adapter/http/handlers_revert_transaction.go
@@ -41,15 +41,19 @@ func (s *Server) handleRevertTransaction(w http.ResponseWriter, r *http.Request)
 	// Extract optional fields from request body
 	if reqBody != nil {
 		if metadata, ok := reqBody["metadata"].(map[string]any); ok {
-			metadataMap := make(map[string]string)
+			// Decode metadata through the shared typed-metadata path so numeric,
+			// boolean and other non-string values survive losslessly (parity with
+			// create-transaction / set-metadata). Invalid values (objects, arrays,
+			// non-integer floats) are rejected with 400 INVALID_REQUEST instead of
+			// being silently dropped (EN-1509).
+			ms, err := commonpb.MetadataFromAnyMap(metadata)
+			if err != nil {
+				writeBadRequest(w, "INVALID_REQUEST", fmt.Errorf("invalid metadata: %w", err))
 
-			for k, v := range metadata {
-				if strVal, ok := v.(string); ok {
-					metadataMap[k] = strVal
-				}
+				return
 			}
 
-			payload.Metadata = commonpb.MetadataFromGoMap(metadataMap)
+			payload.Metadata = ms
 		}
 
 		if force, ok := reqBody["force"].(bool); ok {
--- a/docs/technical/contributing/api-comparison.md
+++ b/docs/technical/contributing/api-comparison.md
@@ -159,7 +159,7 @@ See [Numscript Guide](./numscript.md) for complete documentation.
 - ✅ Standard revert
 - ✅ `force` option (ignore insufficient balances)
 - ✅ `atEffectiveDate` option (use original transaction timestamp)
-- ✅ Revert metadata
+- ✅ Revert metadata (typed values — string, integer, boolean — preserved losslessly; unsupported values rejected with `400 INVALID_REQUEST`)
 - ✅ Verification that transaction is not already reverted
 
 **Navigable revert relationship.** The revert link is a first-class part of the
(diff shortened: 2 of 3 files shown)


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1570#issuecomment-4955758877

internal/adapter/http/handlers_revert_transaction.go
```diff
@@ -41,15 +41,19 @@ func (s *Server) handleRevertTransaction(w http.ResponseWriter, r *http.Request)
 	// Extract optional fields from request body
 	if reqBody != nil {
 		if metadata, ok := reqBody["metadata"].(map[string]any); ok {
```
NumaryBot, line 43: 🟠 **[major] Reject non-object metadata instead of ignoring it**

When the request contains `metadata` but it is not a JSON object (for example `{"metadata":[]}`, `null`, or a string), this type assertion fails and the handler skips the new typed-metadata validation entirely, then still applies the revert without metadata. That preserves the same silent data-loss behavior for an invalid metadata shape even though the API schema requires an object and the new behavior is meant to return `400 INVALID_REQUEST` for unsupported metadata.

internal/adapter/http/handlers_revert_transaction.go
```diff
@@ -41,15 +41,19 @@ func (s *Server) handleRevertTransaction(w http.ResponseWriter, r *http.Request)
...
-			for k, v := range metadata {
-				if strVal, ok := v.(string); ok {
-					metadataMap[k] = strVal
-				}
+			// Decode metadata through the shared typed-metadata path so numeric,
+			// boolean and other non-string values survive losslessly (parity with
+			// create-transaction / set-metadata). Invalid values (objects, arrays,
+			// non-integer floats) are rejected with 400 INVALID_REQUEST instead of
+			// being silently dropped (EN-1509).
+			ms, err := commonpb.MetadataFromAnyMap(metadata)
```
flemzord, line 49: **Medium:** This can silently corrupt valid `int64` metadata above JSON's exact `float64` range. `reqBody` is decoded into `map[string]any`, so `9007199254740993` is rounded before `MetadataFromAnyMap` sees it, even though the OpenAPI schema permits `int64`; the reverted transaction can therefore store a different value while returning success. Consider decoding metadata with an exact-number path (`UseNumber`/raw typed decoding) and add boundary tests above `2^53` and near the signed limits.

flemzord (approved):
No High or Critical issue found. Approving under the requested severity threshold; the inline Medium finding and the existing non-object-shape finding remain actionable but non-blocking.

PR #1571, opened 2026-07-13
fix(EN-1510): return transaction receipt from HTTP GetTransaction

gfyrag:
### Problem
gRPC \`GetTransaction\` returns \`GetTransactionResponse{transaction, receipt}\`, but the HTTP handler (\`internal/adapter/http/handlers_get_transaction.go\`) did \`transaction, _, err := s.backend.GetTransaction(...)\` — discarding the receipt. This is a transport-parity gap for a security/verifiability feature.

### Fix
- \`GET /v3/{ledgerName}/transactions/{transactionId}\` now returns \`data: { transaction, receipt }\` (clean envelope; v3 is alpha).
- The receipt is **reused verbatim** from the backend read path — never recomputed in the HTTP handler — so checkpoint/live consistency stays owned by the controller.
- The \`receipt\` field is **always present**; an empty string means the transaction legitimately has none (e.g. a reversal, or a serving node that cannot sign). The transaction is rendered via its existing camelCase \`MarshalJSON\`, keeping the body consistent with the gRPC-gateway shape.

### Response shape
\`\`\`json
{ "data": { "transaction": { ... }, "receipt": "eyJhbGci..." } }
\`\`\`

### Changes
- \`internal/adapter/http/handlers_get_transaction.go\` — surface receipt in a \`{transaction, receipt}\` envelope.
- \`openapi.yml\` — new \`GetTransactionResponse.data\` shape + receipt semantics.
- \`docs/technical/contributing/api-comparison.md\` — receipt on GetTransaction now available over HTTP too.
- \`internal/adapter/http/handlers_get_transaction_test.go\` — receipt-present, receipt-empty (empty-string + nil), plus existing backend-error/not-found coverage.

### Validation
- \`GOROOT= go build ./...\` — clean
- \`GOROOT= go test ./internal/adapter/http/... -count=1\` — pass
- \`go generate ./... && go mod tidy && golangci-lint run --fix\` — tree clean (no drift)


internal/adapter/http/handlers_get_transaction.go
```diff
@@ -37,5 +53,16 @@ func (s *Server) handleGetTransaction(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	writeOK(w, transaction)
+	// receipt is non-nil on the success path (the gRPC client always returns a
```
NumaryBot, line 56: 🔴 **[blocker] Sign local HTTP transaction reads before returning**

When the routed controller serves a local linearizable/stale read, it calls `DefaultController.GetTransaction`, which intentionally returns a nil receipt because the gRPC server signs local reads afterward. This nil guard turns that common local-success path into `"receipt":""`, so HTTP still does not return a JWT on the leader/healthy local node even when receipt signing is configured; only reads forwarded through the gRPC client get a populated receipt.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1571#issuecomment-4955757565

flemzord (changes requested):
High blocker remains on this head: the normal local `RoutedController -> DefaultController.GetTransaction` path still returns a nil receipt, and receipt generation still happens only in the gRPC server layer. The HTTP handler therefore emits `"receipt":""` on the healthy local/leader path even when signing is configured. Existing inline finding (not duplicated): https://github.com/formancehq/ledger/pull/1571#discussion_r3568983616

gfyrag:
@flemzord addressed in f53e5019b — the blocker (receipt only produced in the gRPC layer, HTTP local path emitting `"receipt":""`) is fixed.

Local receipt signing moved into the **shared controller layer**: `DefaultController.GetTransaction` now signs from a `*receipt.Signer` injected via constructor (nil on non-signing nodes). Both transports route through it — gRPC (`impl.ctrl` on the live path) and HTTP (`RoutedController → DefaultController`) — so the HTTP local/leader path now returns a populated, verifiable receipt.

Invariants preserved:
- **Freshness barrier**: the tx is read via `GetTransactionFrom` (handle closed), then a *new* read handle is opened and the receipt signed from it — snapshot strictly after the tx read, same ordering as the old gRPC code.
- **Forwarded relay**: when RoutedController forwards, the leader's authoritative token is relayed as-is; the adapter no longer re-signs.
- **Checkpoint path**: still signs from the checkpoint's fixed store, via `ComputeTransactionReceipt(ctx, mainStore, ...)` on the controller.

Tests: receipt-computation tests migrated into the ctrl package (local-sign / nil-signer / reversal-empty / reader-not-live-store / error-propagation) + a new e2e `get_transaction_receipt_http_test.go` asserting a non-empty verifiable `data.receipt` over HTTP. Controller interface signature unchanged (no mock drift). Could you re-review?

flemzord (approved):

PR #1572, opened 2026-07-13
fix(EN-1533): stop mutating initial account types before CreateLedger audit capture

gfyrag:
## Summary

`processCreateLedger` assigned the map key into each order-owned `AccountType` message (`at.Name = name`) **before** the accepted order was captured for the audit chain. As a result the audit chain recorded a *normalized* account type rather than the accepted order. With an idempotency key on the batch, the pre-processing frozen outcome hash (over the accepted order) could then disagree with the checker's hash re-derived from the audited order, surfacing a false idempotency mismatch. This is an aliasing bug: the accepted order, the mutable `LedgerInfo`, and the generated log all shared the same messages.

## Fix

- Build a **new canonical account-type map** from proto clones (`CloneVT`) of the order-owned messages, assigning the map key to each **clone's** `Name`. The order-owned messages under `CreateLedgerOrder.AccountTypes` are never mutated.
- Use the canonical map for `LedgerInfo.AccountTypes` (map-key-authoritative — unchanged semantics).
- Give the created-ledger log its **own independent cloned snapshot**, sharing no mutable map or message with the store (which `processAddAccountType` mutates in place) or the order. The clones carry the same map-key canonical `Name`, so `CreatedLedgerLog.ToLedgerInfo()` still reconstructs an identical `LedgerInfo`.

## Preserved semantics / no regressions

- The EN-1521 deterministic sorted-key validation (first invalid pattern reported identically on every replica) is preserved.
- No new key/name validation contract or business rule: a nil map value has `GetPattern() == ""` and is rejected by `ValidatePattern` (`"pattern must not be empty"`) before the clone, so every cloned entry is guaranteed non-nil. No contract decision required.

## Tests

- **Processing** (`processor_ledger_test.go`): with blank and mismatched embedded names, snapshot the deterministic order bytes and assert the order bytes + embedded names are unchanged after processing, while `LedgerInfo` and the created log use the map-key canonical name (on independent messages). EN-1521 determinism test kept green.
- **FSM/audit** (`create_ledger_account_type_aliasing_test.go`): apply a keyed `CreateLedger` through the real machine; assert the audited `SerializedOrder` preserves the submitted embedded names verbatim while the derived log is map-key canonical; assert an identical keyed replay surfaces **no** idempotency conflict. Verified the test fails against the old mutating behavior.

## Checks

- `go test ./internal/domain/processing/... ./internal/infra/state/... ./internal/application/check/... -count=1` green
- `go generate ./...` + `go mod tidy` produce no changes; `golangci-lint run --fix` 0 issues; tree clean

### DIFF
--- a/internal/domain/processing/processor_ledger.go
+++ b/internal/domain/processing/processor_ledger.go
@@ -26,17 +26,34 @@ func processCreateLedger(ledger string, order *raftcmdpb.CreateLedgerOrder, ctx 
 		return nil, &domain.ErrLedgerAlreadyExists{Name: ledger}
 	}
 
-	// Validate initial account types if provided. Iterate names in sorted order
-	// so the first invalid pattern reported (chain-bound ErrInvalidPattern →
-	// AuditFailure) is identical on every replica — a raw map range could
-	// surface a different offending pattern per node (invariant #2). See
-	// EN-1521.
+	// Validate initial account types if provided, then build a canonical map
+	// for derived state. Iterate names in sorted order so the first invalid
+	// pattern reported (chain-bound ErrInvalidPattern → AuditFailure) is
+	// identical on every replica — a raw map range could surface a different
+	// offending pattern per node (invariant #2). See EN-1521.
+	//
+	// The canonical map holds CLONES of the order-owned messages with each
+	// clone's Name set to the authoritative map key. We MUST NOT mutate the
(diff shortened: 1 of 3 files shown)


flemzord (approved):
No High/Critical findings. The clone-based canonicalization preserves accepted order bytes and isolates derived state/log messages; the regression tests cover audit capture and idempotent replay.

PR #1573, opened 2026-07-13
fix(EN-1524): preflight technical-update envelopes before mutation

gfyrag:
### Problem
`applyTechnicalUpdates` validated scopes and kinds **sequentially while dispatching**, so a malformed later `TechnicalUpdate` — or an unsupported mixture of orders + technical updates — could be discovered only **after** an earlier handler had already mutated state (cache reset via `applyClusterConfig`, backup-job map via `applyBackupOrder`, WriteSet queue via `applyMirrorSyncUpdate`).

### Fix
A mutation-free **preflight** over the whole envelope runs before the first handler dispatch (`preflightTechnicalUpdates`, `internal/infra/state/machine_technical_updates.go`). It performs three classes of check, each mapped to the surfacing the FSM already uses:

1. **Coverage bitset well-formedness** — validated against the same execution plan the handlers' scopes will use, via a shared `validateCoverageBits` extracted from `applyPlans` (`scope.go`) so both paths use identical rules with **no slot mutation**. Malformed → `*domain.ErrInvalidExecutionPlan` (business rejection, recognised by `planInvariantDescribable`).
2. **Producer-shape enforcement** — at most one `TechnicalUpdate` per proposal; non-MirrorSync updates are technical-only (zero orders); MirrorSync may coexist with its mirror-ingest order batch and remains the sole update. Violations → `*domain.ErrInvalidExecutionPlan`.
3. **nil / unknown kind** — stays **FSM-fatal**, identical to the dispatch loop's `default` arm and the EN-1323 rolling-upgrade divergence contract (invariant #7).

Handlers run only after the preflight succeeds. MirrorSync's WriteSet queue and the existing handler order are preserved. Coordinated with the merged PR #1558: complete execution-plan validation runs before any preload mutation; this adds the symmetric guarantee for the technical-update phase.

### Non-goals (per ticket)
No staging/cloning of cluster config / backup jobs / idempotency / sink state; no rollback / copy-on-write / prepared-effect framework; no broadening of TechnicalUpdate business semantics. Post-mutation I/O failure handling is EN-1302.

### Where
- `internal/infra/state/machine_technical_updates.go` — `preflightTechnicalUpdates`, called first in `applyTechnicalUpdates`.
- `internal/infra/state/scope.go` — extracted `validateCoverageBits` (shared by `applyPlans`, DRY).

### Tests
- `machine_technical_updates_test.go` (new): producer-shape acceptance for every builder (cluster config, sink cursor, idempotency eviction, full/incremental backup, MirrorSync data + error); forged multi-update / order-coexistence / nil-kind / out-of-range-coverage rejection; **FSM-level no-mutation-before-reject** for ClusterConfig (threshold + epoch untouched, `LastClusterConfig` nil) and backup Start (job not registered in `BackupJobsState`).
- `scope_test.go`: the historical two-TU `CoverageMissShortCircuits` test is retargeted to the now-canonical single-TU shape (`CoverageMissPropagates`) — the short-circuit-between-handlers concern is subsumed by the one-per-proposal cap.

### Verification
- `GOROOT= go test ./internal/infra/state/... ./internal/application/... -count=1` — pass
- `go generate ./...` + `go mod tidy` — tree clean
- `golangci-lint run --fix` — 0 issues

### DIFF
--- a/internal/infra/state/machine_technical_updates.go
+++ b/internal/infra/state/machine_technical_updates.go
@@ -32,6 +32,18 @@ import (
 // those and converts them into ApplyResult.Error so the proposer learns
 // about the rejection without an FSM-level abort.
 func (fsm *Machine) applyTechnicalUpdates(scopeFactory processing.ScopeFactory, batch *dal.WriteSession, raftIndex uint64, proposal *raftcmdpb.Proposal) error {
+	// Preflight the WHOLE technical-update envelope before dispatching a
+	// single handler. applyTechnicalUpdates below mutates state as it
+	// walks the slice (applyClusterConfig resets the cache, applyBackupOrder
+	// touches BackupJobsState, applyMirrorSyncUpdate queues into the
+	// WriteSet, …) — so a malformed later update, or an unsupported mixture
+	// of orders and technical updates, must be caught FIRST. Otherwise an
+	// earlier handler's direct mutation lands before the later update is
+	// rejected. See EN-1524.
+	if err := preflightTechnicalUpdates(proposal); err != nil {
+		return err
+	}
+
 	for i, tu := range proposal.GetTechnicalUpdates() {
 		scope, scopeErr := scopeFactory.NewScope(tu.GetCoverageBits())
 		if scopeErr != nil {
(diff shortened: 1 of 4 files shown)


internal/infra/state/machine_technical_updates.go
```diff
@@ -94,6 +106,100 @@ func (fsm *Machine) applyTechnicalUpdates(scopeFactory processing.ScopeFactory,
...
+	// plan without touching any coverage slot (the mutation applyPlans
+	// would do). NewScope re-runs the mutating variant per handler; here we
+	// only need the reject-before-mutation guarantee.
+	if err := validateCoverageBits(plans, tu.GetCoverageBits(), nil); err != nil {
+		return fmt.Errorf("technical_updates[0]: %w", err)
+	}
+
+	// Kind + order coexistence. MirrorSync is the only kind allowed to ride
+	// alongside an order batch; every other kind is technical-only.
+	switch tu.GetKind().(type) {
```
NumaryBot, line 180: 🟠 **[major] Reject MirrorSync with non-ingest orders**

When a `MirrorSync` technical update is paired with orders, this case accepts the proposal without checking that every order is actually a mirror-ingest order for the same mirror batch. A forged or buggy producer can therefore bundle a `MirrorSync` cursor update with unrelated normal orders; if those orders succeed, `WriteSet.Merge` drains the pending mirror sync and advances the mirror cursor without ingesting the source logs that cursor represents. The preflight should enforce the documented MirrorSync + mirror-ingest-only shape rather than allowing arbitrary orders here.

internal/infra/state/machine_technical_updates.go
```diff
@@ -94,6 +106,100 @@ func (fsm *Machine) applyTechnicalUpdates(scopeFactory processing.ScopeFactory,
...
+	var plans []*raftcmdpb.AttributeCoverage
+	if plan := proposal.GetExecutionPlan(); plan != nil {
+		plans = plan.GetAttributes()
+	}
+
+	// Shape: at most one technical update per proposal. Every current
+	// producer emits exactly one; more than one means a producer bug or a
+	// forged proposal, and the two-shape contract below (which reasons
+	// about "the" technical update) only holds for a single update.
+	if len(updates) > 1 {
```
NumaryBot, line 162: 🟠 **[major] Preserve fatal handling for unknown updates in multi-TU envelopes**

If a proposal contains more than one technical update and any later update has a nil/unknown kind, this early multi-update check returns `ErrInvalidExecutionPlan`, which `applyProposal` converts into a business rejection. That bypasses the FSM-fatal default-arm behavior explicitly kept for nil/unknown kinds (the rolling-upgrade divergence signal), so a multi-update envelope containing an unknown oneof is no longer surfaced loudly. Scan for unsupported kinds before converting the envelope to a business rejection for shape.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1573#issuecomment-4955814547

gfyrag:
Fermeture sans merge : le problème EN-1524 n'existe pas sur `release/v3.0` aujourd'hui.

Au plus un `TechnicalUpdate` par proposition chez tous les producteurs actuels (cluster config, sink, éviction idempotence, backup, mirror), donc pas de séquence « muter puis rejeter » — il n'y a pas de *later update*. Et depuis #1558, la validation complète du plan d'exécution (+ couverture par TU) précède déjà toute mutation dans `applyProposal`. Le seul handler qui mute en mémoire directement (`applyClusterConfig`) est toujours une proposition technical-only isolée.

Le preflight ajouté ici n'aurait été qu'une défense en profondeur contre une forme de proposition qu'aucun producteur ne génère (ce que les non-goals du ticket demandent de ne pas sur-concevoir). Détail complet posté sur EN-1524. À rouvrir si un futur producteur introduit un multi-TU mutant dans une même proposition.

PR #1574, opened 2026-07-13
feat(EN-1507): expose full account-type model through HTTP create/add APIs

gfyrag:
## Summary

Brings the HTTP `create-ledger` and `add-account-type` endpoints to parity with gRPC (ref **EN-1507**). gRPC already supported richer creation than HTTP exposed; this closes the gap. No new semantics — pure HTTP-layer parity.

### Write side (the ticket)
- **`POST /v3/{ledgerName}`** now accepts `initialSchema` (metadata field type declarations) and `accountTypes` (full model), mapped to `servicepb.CreateLedgerRequest.InitialSchema` / `.AccountTypes`. Also documents the already-wired `defaultEnforcementMode`.
- **`POST /v3/{ledgerName}/account-types`** now accepts the full `AccountType` shape: `name`, `pattern`, `persistence` (NORMAL/EPHEMERAL/TRANSIENT), and `segmentTypes` (regex/uuid/uint64/bytes constraints) → `commonpb.AccountType`.

### DRY / reuse
- New shared helpers in `internal/proto/commonpb/account_type_json.go` (colocated with the proto types, mirroring `metadata_type_parse.go`): `ParsePersistence` / `PersistenceToString` and `SegmentType <-> JSON`. Both HTTP handlers use them; the `ledgerctl account-types` CLI now delegates its persistence parse/format to these instead of its own copies.
- `create-ledger` and `add-account-type` share a single `accountTypeBody` DTO.
- Metadata field type parsing reuses the existing `commonpb.ParseTargetType` / `ParseMetadataType`.

### Read side (coherence)
- `list`/`get` account-types now emit `persistence` + `segmentTypes` for round-trip coherence with the new write shape.

### Docs
- `openapi.yml`: new `SegmentType` and `MetadataFieldTypeCommand` schemas; extended `CreateLedgerRequest`, `AddAccountTypeRequest`, `AccountTypeInfo`, and `LedgerInfo.account_types` with fields + examples.
- `docs/technical/contributing/api-comparison.md` updated (create-ledger + account-types notes).

## Tests
- HTTP handler tests for **both** endpoints: full-model pass-through, non-default persistence (TRANSIENT/EPHEMERAL), regex + uuid + uint64 segment-type constraints, and invalid-input 400s.
- Round-trip unit tests for the `commonpb` persistence + segment-type helpers.

## Verification
- `GOROOT= go test ./internal/adapter/http/... ./internal/proto/commonpb/... ./cmd/ledgerctl/...` green
- `go generate ./...` + `go mod tidy` → no diff; `golangci-lint run --fix` → 0 issues
- `openapi.yml` parses as valid YAML

## Nothing left gRPC-only
Both `initial_schema`/`account_types` (create) and `persistence`/`segment_types` (add) are now reachable over HTTP. No field intentionally withheld.

### DIFF
--- a/internal/adapter/http/handlers_create_ledger.go
+++ b/internal/adapter/http/handlers_create_ledger.go
@@ -18,6 +18,38 @@ type createLedgerBody struct {
 	Mode                   string            `json:"mode,omitempty"`
 	MirrorSource           *mirrorSourceBody `json:"mirrorSource,omitempty"`
 	DefaultEnforcementMode string            `json:"defaultEnforcementMode,omitempty"`
+	// InitialSchema declares metadata field types to seed at creation time.
+	InitialSchema []metadataFieldTypeBody `json:"initialSchema,omitempty"`
+	// AccountTypes declares the initial account types (name -> full model).
+	AccountTypes map[string]accountTypeBody `json:"accountTypes,omitempty"`
+}
+
+// metadataFieldTypeBody is the camelCase JSON representation of a
+// SetMetadataFieldTypeCommand used to seed a ledger's initial metadata schema.
+type metadataFieldTypeBody struct {
+	TargetType string `json:"targetType"` // account | transaction | ledger
+	Key        string `json:"key"`
+	Type       string `json:"type"` // string | int64 | bool | uint64 | ... | datetime
+}
+
+// toProto converts the metadata field type body to its proto command, reusing
+// the shared commonpb enum parsers.
+func (b metadataFieldTypeBody) toProto() (*commonpb.SetMetadataFieldTypeCommand, error) {
+	targetType, err := commonpb.ParseTargetType(b.TargetType)
+	if err != nil {
+		return nil, err
+	}
+
+	metadataType, err := commonpb.ParseMetadataType(b.Type)
+	if err != nil {
+		return nil, err
+	}
+
+	return &commonpb.SetMetadataFieldTypeCommand{
+		TargetType: targetType,
+		Key:        b.Key,
+		Type:       metadataType,
+	}, nil
 }
 
 // mirrorSourceBody holds the mirror source configuration.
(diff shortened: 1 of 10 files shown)


internal/adapter/http/handlers_create_ledger.go
```diff
@@ -87,6 +119,31 @@ func (s *Server) handleCreateLedger(w http.ResponseWriter, r *http.Request) {
...
+			createReq.AccountTypes = make(map[string]*commonpb.AccountType, len(body.AccountTypes))
+			for name, at := range body.AccountTypes {
+				converted, err := at.toProto()
+				if err != nil {
+					writeBadRequest(w, "INVALID_REQUEST", fmt.Errorf("accountTypes[%q]: %w", name, err))
+
+					return
+				}
+
+				createReq.AccountTypes[name] = converted
```
NumaryBot, line 144: 🟠 **[major] Validate initial account type segment constraints**

When `accountTypes` are supplied to create-ledger, this path only converts the JSON and passes it through; the create apply path currently validates the pattern but does not validate `segmentTypes` against the pattern variables or compile regex constraints. A request like `{"accountTypes":{"x":{"pattern":"users:{id}","segmentTypes":{"missing":{"type":"uuid"}}}}}` can therefore succeed and persist an account type that `CompileTypes` later silently skips, so enforcement for that type is lost. Please validate the segment constraints for initial account types before they are accepted.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1574#issuecomment-4955828335

flemzord (approved):
Re-checked the current head and existing review threads. I found no High or Critical issue. The missing initial `segmentTypes` validation can persist a configuration that is later skipped by compilation, and the LedgerInfo/OpenAPI response-shape drift can break generated clients, but both are Medium correctness/contract issues rather than High/Critical blockers under the requested threshold.

PR #1575, opened 2026-07-13
feat(EN-1511): dual-format filter input on all filtered endpoints

gfyrag:
## What & why

Ref **EN-1511**. Filter input was inconsistent per endpoint: prepared queries (REST) consumed the structured `QueryFilter` JSON DSL, while `ledgerctl --filter` and audit HTTP consumed the textual `filterexpr` DSL. Callers had to know which one each endpoint wanted.

Now **every filtered surface accepts BOTH representations interchangeably**, both parsing into the same `*commonpb.QueryFilter` and flowing through the same per-target validity gate (EN-1504 `domain.ValidateFilterForTarget`):

- list transactions / accounts / logs (`?filter=`)
- audit HTTP (`GET /v3/_/audit-entries?filter=`)
- prepared-query create / update (`filter` body field)
- `ledgerctl --filter`

## Shared decoder

One helper, `filterexpr.DecodeDualFormat` (`internal/pkg/filterexpr/decode.go`), detects the form from the first non-whitespace byte and applies the per-target gate:

- `{` → structured v2 JSON `QueryFilter` DSL
- `"` → JSON-quoted textual `filterexpr` (body-field form)
- else → raw textual `filterexpr` (query-string form)

HTTP list handlers consume it via `parseListFilter` (`internal/adapter/http/list_filter.go`) for the generic `?filter=` param and AND-combine it with each endpoint's convenience params (`reference`, `prefix`, `startDate`/`endDate`) through `combineFilters` (which also replaces the duplicated ad-hoc AND-combine logic the list handlers each carried). `DecodeDualFormatStructuralOnly` is the target-less variant for the prepared-query update path (the FSM applies the gate against the immutable stored target).

## Decisions

- **Audit arm → kept text-only, documented.** Audit conditions (`audit[...]`) have no structured JSON representation — their field names (`ledger`, `timestamp`, …) collide with the transaction/log conditions the JSON DSL already claims (EN-1241), and the codec fails loud rather than round-trip to the wrong condition. Lower-risk than inventing a disambiguating JSON wrapper (which would be a third syntax). The dual-format decoder still accepts both forms as input on the audit endpoint; textual is simply the only one that can carry an audit condition.
- **Query-param vs body form (consistent across endpoints).** Query-string endpoints (`?filter=`): textual passed verbatim, structured passed by URL-encoding the JSON object. JSON-body endpoints (prepared queries): a JSON object is structured, a JSON string is textual.
- read-side only; no new persisted projection; camelCase JSON; EN-1504 per-target validity stays the single gate.

## Tests

- `internal/pkg/filterexpr/decode_test.go` — text→QueryFilter, JSON→QueryFilter, **equivalence** of the two for the same logical filter (`proto.Equal`), JSON-quoted-text form, empty/no-op inputs, malformed rejection in each form, per-target rejection in each form, audit text-only, structural-only skips the gate.
- `internal/adapter/http/handlers_list_transactions_test.go` — endpoint-level acceptance: the same filter via `?filter=` in both forms reaches the backend as the same `QueryFilter`; both forms rejected with 400 for a target-invalid condition.

## Docs

`openapi.yml` (shared `ListFilter` query param on the list endpoints + `PreparedQueryFilterInput` `oneOf` on prepared-query request bodies), `api-comparison.md` (new "Filter input formats" section), `cli.md`, `query-pipeline.md`.

## Verification

- `GOROOT= go test ./internal/adapter/http/... ./internal/pkg/filterexpr/... ./internal/proto/commonpb/... ./internal/query/... -count=1` green
- `go generate ./... && go mod tidy` → tree clean (no proto touched); `golangci-lint run --fix` → 0 issues

### DIFF
--- a/docs/ops/cli.md
+++ b/docs/ops/cli.md
@@ -97,7 +97,7 @@ scripting against the CLI predictable across resources.
 | Pagination | `--cursor` | string | Opaque cursor returned by the previous page. Parsed per-resource into the appropriate type (string address, uint64 sequence, …). |
 | Pagination | `--reverse` | bool | Reverse iteration order. Available on commands whose server endpoint supports it. |
 | Pagination | `--all` | bool | Fetch every page at once (no interactive pagination). Available on `accounts` and `transactions`. |
-| Filter | `--filter` | string | Boolean filter expression (see `filterexpr` grammar). |
+| Filter | `--filter` | string | Boolean filter expression — textual `filterexpr` grammar OR the structured JSON `QueryFilter` DSL (dual-format, EN-1511). |
 | Filter | `--prefix` | string | Account-address prefix shortcut. Only on `accounts list` (server-side `HardcodedPrefix` optimization). |
 | Consistency | `--checkpoint-id` | uint64 | Read from a named query checkpoint instead of the live store. |
 | Consistency | `--min-log-sequence` | uint64 | Require the server to have applied at least this log sequence before reading. `FailedPrecondition` if not. |
@@ -971,6 +971,14 @@ ledgerctl accounts list --ledger my-ledger --json
 ##### Filter Expression Syntax
 
 The `--filter` flag accepts a human-readable boolean expression that maps to the underlying `QueryFilter` model.
+
+> **Two interchangeable forms (EN-1511).** `--filter` also accepts the structured
+> JSON `QueryFilter` DSL, e.g. `--filter '{"$match":{"metadata[category]":"premium"}}'`.
+> The textual grammar below and the JSON DSL parse to the same filter and are
+> validated against the same per-target rules, so either form works everywhere
+> `--filter` is accepted (the same dual-format contract every server-side
+> filtered endpoint honors). The textual form is the recommended, more concise
+> one for CLI use.
 
 **Grammar:**
 
--- a/cmd/ledgerctl/cmdutil/flags.go
+++ b/cmd/ledgerctl/cmdutil/flags.go
@@ -242,10 +242,13 @@ type FilterOptions struct {
 	PrefixHelp string
 }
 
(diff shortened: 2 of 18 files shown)


docs/ops/cli.md
```diff
@@ -972,6 +972,14 @@ ledgerctl accounts list --ledger my-ledger --json
 
 The `--filter` flag accepts a human-readable boolean expression that maps to the underlying `QueryFilter` model.
 
+> **Two interchangeable forms (EN-1511).** `--filter` also accepts the structured
+> JSON `QueryFilter` DSL, e.g. `--filter '{"$match":{"metadata[category]":"premium"}}'`.
+> The textual grammar below and the JSON DSL parse to the same filter and are
```
NumaryBot, line 977: 🟠 **[major] Scope the CLI JSON-filter claim to implemented commands**

This says the JSON DSL works everywhere `--filter` is accepted, but some CLI commands with their own `--filter` still call `filterexpr.Parse` directly, e.g. `ledgerctl queries create/update --filter`, so `--filter '{"$match":{"metadata[k]":"v"}}'` fails there instead of using the new dual-format decoder. Either route those commands through `DecodeDualFormat...` as well or narrow the documentation to the list/aggregate commands that use `cmdutil.BuildQueryFilter`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1575#issuecomment-4955869018

flemzord (approved):
Re-reviewed the full current head and the existing prior-head finding. I found no High or Critical issue. The remaining CLI parity gap is Medium: `queries create/update --filter` still use the text-only parser while the docs promise both forms everywhere, causing documented JSON input to fail but not bypassing server validation or corrupting persisted/query state. No duplicate inline comment added.

PR #1577, opened 2026-07-13
fix(EN-1542): reject pre-epoch RFC3339 bounds in HTTP log queries

gfyrag:
## Problem

`GET /v3/{ledgerName}/logs` parsed `startDate`/`endDate` as RFC3339 then cast `time.Time.UnixMicro()` **directly** to `uint64`. A valid RFC3339 date before the Unix epoch (e.g. `1960-01-01T00:00:00Z`) has a negative `UnixMicro()` that wraps into a huge positive `uint64`, silently running the wrong query (a bad `startDate` excludes everything; a bad `endDate` includes everything).

The transaction-list endpoint already rejected this via `parseFilterDateMicros`; the logs endpoint carried a duplicate, weaker parser without the pre-epoch guard.

This is a transport validation bug only — no FSM / storage / QueryFilter semantics change.

## Fix

- Moved the existing `parseFilterDateMicros` (RFC3339 → non-negative unsigned micros, 400 on malformed **and** pre-epoch, per-param error message) into the shared HTTP helper file `internal/adapter/http/list_filter.go`.
- Both `handleListTransactions` and `handleListLedgerLogs` now call the single shared parser (DRY). Removed the logs handler's duplicate inline parsing.
- Validation runs **before** the backend/controller is invoked.
- Range semantics preserved: start inclusive, end exclusive; epoch and post-epoch values map to exact microsecond bounds.

## Tests

Added table-driven tests for the logs endpoint (`TestHandleListLedgerLogs_DateBounds`):
- pre-epoch `startDate` → 400 (names `startDate`)
- pre-epoch `endDate` → 400 (names `endDate`)
- malformed RFC3339 `startDate`/`endDate` → 400
- epoch `startDate` → exact inclusive bound (`micros == 0`)
- post-epoch `startDate` → exact inclusive bound
- post-epoch `endDate` → exact exclusive bound
- full range → start-inclusive / end-exclusive, exact micros on both bounds

Controller-not-called on validation error is asserted with the generated `MockBackend` using `EXPECT().ListLogs(...).Times(0)`; the success cases use `.Times(1)` and capture the `QueryFilter` to assert the exact `LOG_BUILTIN_INDEX_DATE` bounds.

Transaction handler tests re-run as regression coverage after sharing the helper.

## Docs

- `openapi.yml`: clarified that the log `startDate`/`endDate` params must be on or after the Unix epoch (earlier dates → 400).

Closes EN-1542

### DIFF
--- a/internal/adapter/http/handlers_list_ledger_logs_test.go
+++ b/internal/adapter/http/handlers_list_ledger_logs_test.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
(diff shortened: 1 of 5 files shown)


internal/adapter/http/handlers_list_ledger_logs_test.go
```diff
@@ -151,6 +152,179 @@ func TestHandleListLedgerLogs_WithDateFilters(t *testing.T) {
...
+				require.Equal(t, *tc.bound.max, uc.GetMax())
+				require.Equal(t, tc.bound.maxExclusive, uc.GetMaxExclusive())
+			} else {
+				require.Nil(t, uc.Max)
+			}
+		})
+	}
+}
+
+func ptrUint64(v uint64) *uint64 { return &v }
```
flemzord: [P3] This leaves the required `Dirty` check failing because `just pre-commit` rewrites this helper and all of its call sites to the modern `new(expr)` form. Please run `just pre-commit` and commit the generated cleanup so the branch is clean under the repository's required pre-commit workflow.

flemzord (approved):

PR #1578, opened 2026-07-13
fix(EN-1540): remove redundant account prefix and transaction reference HTTP filter aliases

gfyrag:
## What

Removes two redundant HTTP convenience query params now that the shared dual-format `filter` parameter (EN-1511, #1575) is the canonical query-selection surface. This is an alpha v3 API, so no deprecated aliases / compat shims are retained.

- `GET /v3/{ledgerName}/accounts?prefix=…` — removed. Canonical replacement: `filter=address ^= "users:"` (textual) or the structured `{"$match":{"address":"users:"}}` (trailing `:` = prefix match), URL-encoded.
- `GET /v3/{ledgerName}/transactions?reference=…` — removed. Canonical replacement: `filter={"$match":{"reference":"ref-1"}}` (structured, URL-encoded) or textual `reference == "ref-1"`.

## Why

Both aliases were pure sugar over conditions the canonical `filter` already expresses identically (`prefix=` built an `AddressMatch_HardcodedPrefix`; `reference=` built a `ReferenceCondition`). Keeping them is redundant API surface.

## Kept unchanged

`pageSize`, `after`, `reverse`, and the transactions `startDate`/`endDate` RFC3339 range (the timestamp coercion the DSL does not replace). The date range still AND-combines with the canonical `filter`. Both textual and structured forms of `filter` remain accepted.

## Handler / OpenAPI / doc changes

- `internal/adapter/http/handlers_list_accounts.go` — drop `prefix` parse + `prefixFilter` build; `filter` is now taken directly from `parseListFilter`.
- `internal/adapter/http/handlers_list_transactions.go` — drop the `reference` parse/`ReferenceCondition` block; date range + generic `filter` still AND-combine via `combineFilters`. Updated the handler doc comment.
- `openapi.yml` — removed the accounts `prefix` param and the transactions `reference` param; updated both endpoint descriptions, the shared `ListFilter` note, and re-inlined the `/volumes` `prefix` description (it kept its own `prefix`; the note that cross-referenced the now-removed accounts param would otherwise dangle). `/volumes?prefix=` is out of scope and untouched behaviorally.
- `docs/technical/contributing/api-comparison.md` — updated the dual-format section and the transactions read-feature row to show the canonical replacements with URL-encoding for the structured form.

## Tests

- `TestHandleListAccounts_PrefixFilterCanonicalReplacement` — `filter=address ^= "users:"` reaches the backend as `AddressMatch_HardcodedPrefix{"users:"}` (identical to what the old alias produced); the removed `prefix=` alias alone yields a nil (unfiltered) filter.
- `TestHandleListTransactions_ReferenceFilterAndDateRange` — structured reference `filter` AND-combines with `startDate`/`endDate` into a 2-element `$and` (date range + `ReferenceCondition{Hardcoded:"ref-1"}`); the removed `reference=` alias alone yields a nil filter.
- Removed the now-superseded `TestHandleListAccounts_WithPrefix` (asserted old alias behavior); replaced by the above.
- Existing dual-format / date / pagination / ordering tests unchanged and green.

`GOROOT= go build ./...` and `go test ./internal/adapter/http/...` pass; `go generate ./... && go mod tidy` produce no drift; golangci-lint reports 0 issues.

Closes EN-1540

### DIFF
--- a/docs/technical/contributing/api-comparison.md
+++ b/docs/technical/contributing/api-comparison.md
@@ -399,8 +399,15 @@ non-whitespace byte:
 - **Query-string endpoints** (`?filter=`): the value is textual `filterexpr`
   passed verbatim (URL-encoded). To pass the structured form, URL-encode the JSON
   object as the value (`?filter=%7B%22%24match%22%3A…%7D`). The generic `filter`
-  parameter is AND-combined with the endpoint's convenience params (`reference`,
-  `prefix`, `startDate`/`endDate`).
+  parameter is AND-combined with the endpoint's remaining convenience params (the
+  transactions `startDate`/`endDate` timestamp range). It has no dedicated
+  address-prefix or reference aliases: an account address prefix is the textual
+  `filter=address ^= "users:"` (or structured
+  `?filter=%7B%22%24match%22%3A%7B%22address%22%3A%22users%3A%22%7D%7D`, i.e.
+  `{"$match":{"address":"users:"}}` with the trailing `:` marking a prefix
+  match), and a transaction reference is the structured
+  `?filter=%7B%22%24match%22%3A%7B%22reference%22%3A%22ref-1%22%7D%7D`
+  (`{"$match":{"reference":"ref-1"}}`) or the textual `filter=reference == "ref-1"`.
 - **JSON-body endpoints** (prepared-query create/update `filter` field): a JSON
   object is the structured form; a JSON string (`"filter": "metadata[k] == v"`)
   is the textual form.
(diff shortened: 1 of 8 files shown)


docs/technical/contributing/api-comparison.md
```diff
@@ -399,8 +399,15 @@ non-whitespace byte:
 - **Query-string endpoints** (`?filter=`): the value is textual `filterexpr`
   passed verbatim (URL-encoded). To pass the structured form, URL-encode the JSON
   object as the value (`?filter=%7B%22%24match%22%3A…%7D`). The generic `filter`
-  parameter is AND-combined with the endpoint's convenience params (`reference`,
-  `prefix`, `startDate`/`endDate`).
+  parameter is AND-combined with the endpoint's remaining convenience params (the
+  transactions `startDate`/`endDate` timestamp range). It has no dedicated
```
flemzord, line 403: [P3] Keep the architecture guide in sync

This leaves `docs/technical/architecture/subsystems/read-path/query-pipeline.md:65-68` saying list handlers still combine the `reference` and `prefix` convenience parameters, so the canonical read-path documentation advertises parameters this PR removes. Please update that paragraph to say only the transaction date range remains.

flemzord (approved):
No High or Critical blockers found. The documentation consistency note below is non-blocking.

PR #1560, opened 2026-07-10
feat(EN-1406): resolve Numscript deps via upstream ResolveDependencies + stale-inputs check

gfyrag:
## What

Implements [EN-1406](https://formance-team.atlassian.net/browse/EN-1406): replace the emulation-based Numscript dependency discovery with the upstream library's `ParseResult.ResolveDependencies(ctx, vars, store)` API, and add a stale-inputs detection mechanism.

## ⚠️ Dependency pin caveat (blocking for merge)

The numscript dependency is pinned to the **unmerged** PR-branch pseudo-version:

```
github.com/formancehq/numscript v0.0.25-0.20260710091422-92efd3e3a5ac
```

This is the head of upstream PR [formancehq/numscript#169](https://github.com/formancehq/numscript/pull/169) (`feat/resolve-dependencies-simplified`). **This PR must be repinned to a tagged numscript release before it is merged.** Kept as **draft** for that reason.

## Changes

- **discover.go** (new): parse (shared `NumscriptCache`) + `ResolveDependencies`; map resolved account/metadata reads & writes to Ledger volume/metadata preload keys. `AccountsReads`→volume preload, `AccountsWrites`→volume preload (source+dest mutation), `MetaReads`→metadata preload, `MetaWrites`→metadata preload (previous-value capture, #186). Ledger volumes/metadata have no color/scope dimension.
- **store.go** (new): `ValueSource`-backed `numscript.Store` (force mode returns `MaxForceBalance`) + `RecordingStore` recording every read value and producing a deterministic BLAKE3 `Hash()` over sorted records. Absent metadata is recorded distinctly from present. **Color/scope-qualified balance/metadata reads are rejected** with `domain.ErrColoredBalanceUnsupported` — serving one underlying uncolored volume as a full per-color view would enable a double-spend (an unbounded colored source, which reads no balance, is still allowed).
- **value.go** (new): raw-string extraction for `set_account_meta` / `set_tx_meta` values — upstream `Value.String()` now quotes strings, so scalars are unwrapped from the tagged-JSON form to preserve the stored-value contract.
- **emulate.go / emulate_test.go**: deleted. `ErrMetaNotSupported`, `ErrNonDeterministicScript`: removed.
[...]

PR #1579, opened 2026-07-13
docs(EN-1545): RFC — unified hot-path apply simulation

gfyrag:
## Summary

Investigation spike (EN-1545). Deliverable is a single design RFC — **no production code changes** (docs-only; `go build ./...` clean).

Adds `docs/drafts/apply-simulation-rfc.md`: a **BLOCKED investigation** that neutrally compares bounded options for reducing the divergence between admission's planning pass and the FSM apply path (admission's hand-derived shadow + the redundant double numscript evaluation).

## Status: BLOCKED — no single recommendation asserted

Per the authoritative EN-1545 decision, this is a no-go for the universal simulator; the RFC keeps the ticket **blocked** and presents a neutral comparison rather than a "GO".

- **Blocked on**: EN-1406 / #1560, EN-1532, EN-1288 (see the RFC §0 dependencies table; the post-EN-1406 machinery — `ValueSource`, `ResolveDependencies`, `InputsResolutionHash` — does not exist on the current baseline yet).
- **Options compared** (§5): (A) current design + targeted options, (B) shared typed non-reading helpers, (C) transcript/sidecar with an explicit preflight manifest, (D) universal `simulate(order, StateReader, WriteSink)` interpreter — evaluated as the **no-go control**.
- **Direction pending re-validation**: A ≥ B > C ≫ D; no-go is an explicitly valid outcome. Go/no-go gates are in §8.

## Key design corrections in this RFC

- **Digest ≠ preflight manifest** (§4.1): `InputsResolutionHash` is a digest, not key identities; recomputing it re-runs resolution (issuing the reads) before any hash compare, so "check hash before any read" is not achievable with a digest — a true preflight needs an explicit key manifest (Option C). The earlier hash-before-read claim is removed.
- **Atomic per-order sink boundary** (§3.1a): a failed order must contribute NO effects (buffer-and-commit-per-order or rollback), not "swallow and keep the delta".
- **Typed metadata at the sink** (§3.1b): the sink carries `*commonpb.MetadataValue`, not flattened strings.
- Correctness lever unchanged: the coverage gate (invariant #9) turns any preload-narrowing mistake into a loud reject, not a silent desync.

## Notes

- Linked from `docs/drafts/README.md` (entry updated to match this framing).

Relates EN-1545


docs/drafts/apply-simulation-rfc.md
```diff
@@ -0,0 +1,272 @@
...
+**Invariant #9 holds because the FSM backend's reads are the coverage-gated `Scope`.** Each `StateReader` method on the FSM side goes through `Scope.GetX(...)`, so the per-order `coverage_bits` admit it. A read the interpreter makes that admission never declared hits `*state.ErrCoverageMiss` — see §4.
+
+### 3.2 `WriteSink` — the write half
+
+Today the two write halves are `batchEffects` (admission) and `WriteSet` (FSM). Generalize:
+
+```go
+// Illustrative — the write half of the interpreter.
+type WriteSink interface {
+    AddBalanceDelta(key domain.VolumeKey, delta *big.Int)
```
NumaryBot: 🟠 **[major] Expose postings in the write sink contract**

For the FSM-mode implementation described here, `AddBalanceDelta(key, delta)` does not carry enough information to preserve current posting semantics: `applyPosting` needs the source, destination, asset, amount, and side so it can increment source `Output`, destination `Input`, and perform the source balance check. If implementers follow this interface, the FSM backend can only see a net delta per volume key, which conflicts with the RFC's own requirement to keep materialized volumes; the contract should expose postings (or equivalent source/destination operations) instead.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1579#issuecomment-4957927378

gfyrag:
Planning gate: EN-1545 is not unblocked by the latest merge wave. The Jira ticket is now a Ledger v3.1 RFC and remains blocked until EN-1406 / PR #1560, EN-1532, and EN-1288 / PR #1529 have settled.

This PR also follows the superseded universal `simulate(order, StateReader, WriteSink)` framing and reaches a GO recommendation, whereas the reviewed ticket plan requires a neutral comparison of bounded alternatives and explicitly treats that universal simulator as a control/no-go candidate unless new evidence disproves the identified gaps.

Please do not merge this PR in its current form. Re-run the analysis against the post-dependency code, target the current EN-1545 contract, and compare the conservative model, typed planning/effect helpers, transcript/sidecar option, and universal simulator before making a recommendation. No production behavior belongs in this RFC.

docs/drafts/apply-simulation-rfc.md
gfyrag: Addressed in 00750d458. The WriteSink balance primitive is now posting-shaped, not delta-shaped: `ApplyPosting(p Posting, force bool)` (Posting = source/destination/asset/amount) replaces `AddBalanceDelta(key, delta)`. This preserves exactly what `applyPosting` needs — which side moved (source Output++ vs destination Input++), the source/destination pairing for the balance check, and `force` threaded so both backends apply the same policy. The net (input−output) delta is now demoted to a *projection admission derives* for its intra-batch overlay, never the sink primitive. The §5 deletion-mapping table + §3.2/§6/§7/§8 were realigned; no `AddBalanceDelta` reference remains, and materialized-volume semantics are preserved on the FSM side.

docs/drafts/apply-simulation-rfc.md
```diff
@@ -0,0 +1,289 @@
...
+- **Scope:** Collapse admission's hand-derived shadow of the FSM apply path into a single order interpreter, `simulate(order, StateReader, WriteSink)`, run in two modes (admission dry-run, FSM apply) over an abstract state interface.
+- **Depends on:** **EN-1406 / #1560 must land first** (`feat/en-1406-numscript-resolve-deps`). This RFC targets the post-EN-1406 shape: the `numscript.ValueSource` abstraction, the `SafeResolveDependencies` / `SafeRun` panic-recovery wrappers, the apply-time triage block in `processor_transaction_numscript.go`, and the `batchEffects` intra-batch overlay in admission. None of those exist on `release/v3.0` yet.
+
+> **Note on the Go API:** the interfaces and types below are **illustrative** and **MAY be refined** during implementation. They exist to make the contract concrete and to pin the correctness argument.
+
+---
+
+## 0. Recommendation (read this first)
+
+**Go — but phased and narrowed.** The unified model is viable and worth doing, with two important qualifications that the investigation surfaced:
```
flemzord: 🔴 **[high] Rebase the RFC on the authoritative EN-1545 decision**

The current EN-1545 contract explicitly supersedes the universal `simulate(order, StateReader, WriteSink)` proposal, keeps the ticket blocked until #1560 / EN-1532 / EN-1288 settle, and requires a neutral comparison of the conservative model, typed helpers, a transcript/sidecar, and the universal simulator as the control/no-go option. This unconditional GO recommendation instead selects that superseded design and starts P1 immediately, so merging it would record the opposite architecture decision before the required baseline and proof work exist. Please rerun the RFC against the settled dependencies and include the required order matrix, experiments, benchmarks, and neutral alternative comparison before recommending a direction.

docs/drafts/apply-simulation-rfc.md
```diff
@@ -0,0 +1,289 @@
...
+
+### 4.1 Preload = the simulated path, not all branches
+
+`ResolveDependencies` today explores **all** oneof branches of a script to build the dependency set. The unified model narrows this: it preloads only the branch the single simulation pass actually takes.
+
+This is **narrower**, and it is correct **iff the `InputsHash` pins branch selection**. The argument, confirmed by the EN-1406 independent review:
+
+1. `Run`'s reads are a **subset** of `ResolveDependencies`' reported reads — a single execution touches one branch; branch selection is a pure function of the resolved inputs.
+2. Admission binds those inputs into `InputsHash` via the `RecordingStore` (including metadata *absence*, via the NUL-prefixed absent sentinel — a key that gains a value between admit and apply hashes differently).
+3. At apply, the FSM re-resolves against the coverage-gated `Scope` and compares hashes **before** taking the divergent path (§4.3). If any input shifted, the branch could differ, and the hash mismatch rejects as stale *before* the interpreter can read a key the narrower preload never loaded.
```
flemzord: 🔴 **[high] Do not treat the digest as a preflight manifest**

This ordering cannot be guaranteed by the proposed single pass: `InputsResolutionHash` contains only a digest, not the identities of the keys needed to recompute it, and upstream `Run` is not a conservative dependency-analysis API. Re-running the actual path can therefore request a newly-derived key before the comparison happens; the coverage gate can only report that late read, not make the preceding hash check possible. This is exactly why EN-1545 excludes replacing conservative resolution unless a typed key manifest/transcript with preflight is proven. Please retain conservative coverage or specify and test such a manifest; the hash alone does not make single-path preload safe.

docs/drafts/apply-simulation-rfc.md
```diff
@@ -0,0 +1,289 @@
...
+
+// Posting is the balance-mutation unit the interpreter emits (already the shape
+// of a numscript-produced posting and of RevertTransaction's reversed postings).
+type Posting struct {
+    Source, Destination, Asset string
+    Amount                     *big.Int
+}
+```
+
+- Admission's `WriteSink` = the `batchEffects` overlay. `ApplyPosting` **projects** the posting to a net (input−output) delta per volume key (source `−amount`, destination `+amount`) so a later same-batch `balance()` sees the right number — exactly what `batchEffects.addBalanceDelta` does today, but now driven by the posting rather than being the interpreter's primitive. `SetMetadata` / `DeleteMetadata` accumulate as today. All writes are discarded after the batch. Admission's `ApplyPosting` swallows the balance-check failure (best-effort, §4.2); it still records the delta so downstream orders resolve correctly.
```
flemzord: 🔴 **[high] Define an atomic per-order sink boundary**

This says admission swallows a posting failure and keeps its delta, while §8 says a failed order contributes no effects. With a streaming `WriteSink`, a later posting, metadata conversion, overflow, or validation failure can leave partial effects in the admission overlay even though the FSM rejects and rolls back the order; subsequent orders then resolve and hash against state that will never exist. Please specify commit/abort semantics (for example, a per-order buffered sink flushed only after successful authoritative-equivalent simulation) and prove rollback for partial postings, metadata writes, and predecessor failures.

docs/drafts/apply-simulation-rfc.md
```diff
@@ -0,0 +1,289 @@
...
+    // Balance mutation is posting-shaped: the FSM needs source/destination,
+    // asset, amount and side to update materialized volumes (source Output++,
+    // destination Input++) and run the source balance check. Admission's
+    // implementation projects this to a net (input−output) delta per volume key
+    // for its intra-batch overlay; the FSM's implementation materializes the two
+    // volume sides. force is threaded so each backend applies the balance-check
+    // policy identically (§8, force question).
+    ApplyPosting(p Posting, force bool) error
+
+    SetMetadata(key domain.MetadataKey, value string)
```
flemzord: 🟠 **[medium] Preserve typed metadata at the sink boundary**

A `string` value cannot reproduce the current FSM writes: `processAddMetadata` and caller account metadata store `*commonpb.MetadataValue` verbatim, including int, bool, uint, datetime, and null variants. Flattening those values here would change persisted/API-visible metadata and typed index behavior. Please keep the typed value in the FSM-facing operation, or split the Numscript string write from typed caller/AddMetadata writes.

flemzord (changes requested):
Requesting changes for the validated high-severity design blockers below. The current RFC contradicts the authoritative EN-1545 decision, and its single-path preload proof and admission rollback model are not safe as specified. The typed metadata contract is also incomplete.

docs/drafts/apply-simulation-rfc.md
flemzord: Rechecked on `41cf877b`: the merge commit only incorporates `release/v3.0` and leaves this RFC byte-for-byte unchanged. The planning gate is still unmet: #1560 remains open, #1529 remains open with changes requested, and the RFC still selects the superseded universal simulator instead of performing the required neutral comparison and proof work. This remains blocking.

docs/drafts/apply-simulation-rfc.md
flemzord: Rechecked on `41cf877b`: this section is unchanged. The merge does not add a typed key manifest/transcript or a preflight phase; `InputsHash` remains only a digest, so the proposed ordering still cannot compare the hash before a newly derived branch read. This remains blocking.

docs/drafts/apply-simulation-rfc.md
flemzord: Rechecked on `41cf877b`: this section is unchanged and still exposes a streaming sink with no per-order commit/abort boundary. A later failure can therefore leave partial admission effects while the FSM rolls the order back, contaminating subsequent same-batch resolution. This remains blocking.

flemzord (changes requested):
Re-reviewed the exact current head `41cf877b`. The only new commit merges `release/v3.0`; both RFC files are byte-for-byte unchanged and none of the three existing high-severity design blockers has been addressed. Requesting changes on this head; no new distinct findings were found.

docs/drafts/apply-simulation-rfc.md
gfyrag: Addressed in 178bf8c2b. Rebased on the authoritative EN-1545 decision (read the Jira ticket + work plan): retitled to "Safe reuse between admission planning and FSM apply", status BLOCKED investigation, added §0 blocked-on table (EN-1406/#1560, EN-1532, EN-1288) + a baseline note that the post-EN-1406 machinery (ValueSource/ResolveDependencies/InputsResolutionHash) does not exist yet. The universal simulator is now Option D (no-go control) inside a neutral §5 comparison of A (current+opts) / B (typed non-reading helpers) / C (transcript+preflight) / D; recommendation "no-go is valid; A≥B>C≫D pending re-validation." No "Go — proceed" framing remains.

docs/drafts/apply-simulation-rfc.md
gfyrag: Addressed in 178bf8c2b (new §4.1). InputsResolutionHash is a digest, not key identities; recomputing it requires re-running resolution, which issues the reads (incl. a possibly-undeclared newly-derived key) BEFORE any hash compare — so "check hash before any read" is unachievable with a digest, and dependency-set ≠ hash-input-set, and Run is not a conservative-analysis API. Reframed the digest as a post-re-resolution shift detector; a true preflight needs an explicit key manifest (= Option C). Removed the old "hash-check short-circuits before any divergent read" claim.

docs/drafts/apply-simulation-rfc.md
gfyrag: Addressed in 178bf8c2b (new §3.1a). Documented that a streaming sink can leave partial effects when a later posting/metadata-key/overflow/validation failure hits (processCreateTransaction validates numscript metadata keys AFTER postings apply). Specified an atomic per-order boundary — buffer-and-commit-per-order (preferred; matches the FSM orderOverlayScope) or rollback-on-failure — and corrected the contradiction: admission may decline to reject but must contribute NO effects, not "swallow and keep the delta."

docs/drafts/apply-simulation-rfc.md
gfyrag: Addressed in 178bf8c2b (new §3.1b). Verified processAddMetadata stores *commonpb.MetadataValue verbatim and SaveMetadataOrder.GetMetadata() is map[string]*commonpb.MetadataValue (6 variants). Changed the sink to SetMetadata(key domain.MetadataKey, value *commonpb.MetadataValue). (The numscript ValueSource.Metadata READ method stays string — that is the library read interface, correct as-is.)

docs/drafts/apply-simulation-rfc.md
flemzord: 🔴 **[high] Align the checked-in index and PR entry points with the revised decision**

Thanks — the RFC itself now addresses this thread’s architecture concerns. However, `docs/drafts/README.md` still describes this document as proposing the universal `simulate(...)` interpreter and carrying a go/no-go recommendation, while the PR description still says **Recommendation: GO** and repeats the disproved hash-before-read claim. Those entry points now state the opposite of the revised RFC and the authoritative EN-1545 decision. Because this PR’s deliverable is the architecture documentation, please update the checked-in README summary (and the PR description) to the blocked, neutral comparison / universal-no-go framing.

flemzord (changes requested):
Re-reviewed the exact current head `178bf8c2`. The revised RFC substantively clears the previous planning/spec, digest/preflight, atomic-sink, and typed-metadata findings. One high-severity documentation inconsistency remains: the checked-in README and PR entry points still advertise the superseded universal-simulator / GO decision, which is the opposite of the RFC now being delivered. Requesting changes only for that stale framing.

docs/drafts/apply-simulation-rfc.md
gfyrag: Fixed in 16f9f79ff + PR description updated. `docs/drafts/README.md` now frames this as a BLOCKED neutral comparison (options A/B/C/D, universal simulator = no-go control), and the PR description no longer says "Recommendation: GO" or repeats the hash-before-read claim. Both entry points now match the revised RFC and the authoritative EN-1545 decision.

flemzord (approved):
Re-reviewed the stable current head `16f9f79ff`. The final documentation inconsistency is fixed: the checked-in drafts index and PR description now match the RFC’s blocked, neutral comparison and universal-simulator no-go framing. The previous technical blockers remain cleared, and no new high- or critical-severity findings were found.

paul-nicolas (commented):
## Multi-model review (Claude + Codex) — PR #1579

Docs-only investigation spike, so the review focused on the factual accuracy of every code claim, internal consistency, and the accuracy of the BLOCKED status. Reviewed against the actual PR base `release/v3.0`.

**Verdict: the RFC is well-grounded — no Critical/Major/Medium findings.** Every concrete claim was verified against `origin/release/v3.0`:

- `DiscoverNumscriptDependencies` in `internal/domain/processing/numscript/emulate.go`, a single `SafeRun` with infinite balances for discovery — confirmed.
- EN-1406 machinery genuinely absent on the base: no `ResolveDependencies` / `SafeResolveDependencies` / `InputsResolutionHash` / `ErrStaleInputsResolution` / `ValueSource`. The "prospective / BLOCKED on EN-1406" framing is therefore accurate.
- `bulkOverlay` (`admission/overlay.go`), verbatim typed `*commonpb.MetadataValue` storage + the `AccountMetadata()` accessor signature, exactly six `MetadataValue` oneof variants (string/int/uint/datetime/bool/null), and `orderOverlayScope` (`overlay_scope.go`) — all confirmed.
- The `README.md` entry matches the RFC framing (BLOCKED on EN-1406/EN-1532/EN-1288).

**Caveat for anyone re-checking:** a checkout of the EN-1406 feature branch (`feat/en-1406-numscript-resolve-deps`) makes every "prospective" symbol appear to already exist. They do not exist on the PR base `release/v3.0`. Verify the RFC's "does not exist today" claims against the base branch, not a feature-branch worktree — otherwise the whole framing looks (incorrectly) stale.

**Severity counts:** Critical 0 · Major 0 · Medium 0 · Minor 1 · Nit 0. Nothing met the inline-posting threshold.

### Minor (not posted inline)
- `docs/drafts/apply-simulation-rfc.md:113` — the parenthetical attributes the post-postings metadata-key validation to `processCreateTransaction` directly, but on `release/v3.0` that function delegates to `numscriptPostingProducer` (`processor_transaction.go:64-68`), which runs `SafeRun` and `ValidateMetadataKey` in `processor_transaction_numscript.go`. The substantive claim (Numscript-produced metadata keys are validated *after* postings are applied) and the file cross-reference are both correct — only the function name is imprecise. Optional: name the producer.

### Considered and rejected (verified invalid)
- Two invariant-label nits (lines 26 and 52) raised by Codex do not hold: the RFC's citations match `CLAUDE.md` — invariant #6 defines the cache-miss → silent-no-op behavior, and #7 is the never-silently-skip / node-desync rule, which is exactly how the RFC uses them.

_Models: Claude (this repo's perspective) + Codex (external). No Gemini._

docs/drafts/apply-simulation-rfc.md
```diff
@@ -0,0 +1,271 @@
...
+2. **No Pebble capability in the FSM / hot path (#3).** No FSM-facing interface may expose `dal.Store`, `PebbleGetter`, parent `KeyStore` iteration, or a generic escape hatch. Enforced by construction and by a compile-time/wiring proof.
+3. **Every FSM cache read is coverage-gated (#9).** All reads go through `Scope.GetX(...)`; a read admission never declared must hit the gate.
+4. **Coverage misses remain loud (#6/#7).** A read the preload did not cover surfaces as a loud invariant failure (`ErrCoverageMiss` / equivalent), never a silent no-op, and is never reclassified as retryable stale.
+5. **Admission planning is advisory, not authoritative.** Admission cannot become the business authority; the FSM remains responsible for balance/overflow checks, account types, references, metadata behavior, atomic failure and rollback.
+6. **Atomic per-order and per-batch semantics.** A failed order contributes **no** effects (§4). A first-order failure stops and rolls back the atomic proposal exactly as today.
```
NumaryBot, line 66: 🟠 **[major] Skippable-failure side effects missing from the failed-order safety contract**

The RFC states that a failed order contributes no effects, but this does not hold for orders with `skippable_reasons`. In the current FSM, `ProcessOrders` discards the failed order's business overlay yet still emits an `OrderSkippedLog`, advances the ledger log boundary via `assignSkipLogIDAndDate`, and consumes a global sequence number. A future simulator or transcript implementation that follows the RFC's blanket guarantee will model batched skippable failures with the wrong audit and log-ID side effects, producing divergence from the authoritative FSM behaviour the RFC is meant to capture.

**Suggestion:** Qualify the safety property to cover only business/planning effects (postings, metadata, balance deltas), and add a separate clause specifying that skippable-failure side effects — log-ID assignment, sequence consumption, and `OrderSkippedLog` emission — must also be reproduced faithfully by any compliant simulator or transcript.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1579#issuecomment-4957927378

PR #1580, opened 2026-07-13
feat(EN-1544): RFC3339 date coercion + textual date field in the QueryFilter DSL

gfyrag:
## What

Closes EN-1544.

The v3.0 HTTP cleanup wants the shared `filter` param to be the single query surface. The logs/transactions `startDate`/`endDate` convenience params were kept (EN-1540/EN-1543) only because the DSL could not yet replace their RFC3339 ergonomics — it expressed date ranges only as raw uint64 microseconds. This closes that gap on both serializations and both targets.

### Shared coercion (DRY)

`commonpb.CoerceDatetimeMicros(s) (uint64, error)` (in `internal/proto/commonpb/metadata_convert.go`, next to the existing `ParseDatetimeMicros`) is the **one** RFC3339-or-raw-micros coercion, rejecting pre-epoch RFC3339 values with `ErrDatetimeBeforeEpoch`. It is reused by:

- the textual audit `audit[timestamp]` datetime field (previously had its own inline RFC3339 + pre-epoch logic — now deleted);
- the structured JSON DSL date/timestamp bounds;
- the new textual `date`/`timestamp` DSL field;
- the HTTP `startDate`/`endDate` transport helper (`parseFilterDateMicros`), which keeps its RFC3339-only contract (rejects raw integers) but delegates the pre-epoch rejection to the shared helper.

### Structured JSON DSL (`query_filter.go`)

Date-semantic uint bounds — `timestamp`/`insertedAt`/`revertedAt` (transactions) and `date` (logs) — now accept an RFC3339 string **in addition to** raw micros (`boundUintCoercer` dispatches on a `dateSemanticFields` set). `id`/`logId` and metadata uint ranges stay raw-only, so `{"$gte":{"id":"2023-..."}}` still fails as before.

### Textual DSL (`internal/pkg/filterexpr/parser.go`)

New `DateCond` for the top-level `date OP value` / `timestamp OP value` grammar, producing `LogBuiltinUint(DATE)` and `BuiltinUint(TIMESTAMP)` respectively. The operator-to-bound assembly is factored into a shared `uintConditionFromOp` reused by the audit uint/datetime path. `date`/`timestamp` were added to the lexer keyword rule and to the `Value.Kw` allow-list so they remain usable as bare metadata values.

### Per-target validity + pre-epoch

Validity is **not** re-implemented here: the fields only select the proto arm, and the existing per-target gate (`domain.ValidateFilterForTarget` -> `commonpb.ConditionValidForTarget`, EN-1504) routes `date` -> logs and `timestamp` -> transactions, rejecting the wrong target identically for both serializations. Pre-epoch RFC3339 is rejected deterministically at DSL decode time in both forms, consistent with EN-1542's transport-level rejection.

No FSM/storage semantics change — the wire types (`BuiltinUintCondition`, `LogBuiltinUintCondition`) already existed; this is parse/coerce + grammar only.

## Tests

- `internal/proto/commonpb/query_filter_date_test.go`: structured RFC3339 decodes `proto.Equal` to raw micros (log date, tx timestamp/insertedAt/revertedAt closed range); numeric-literal micros still parse; pre-epoch rejected each field; non-date field (`id`) still rejects RFC3339; `CoerceDatetimeMicros` unit coverage.
- `internal/pkg/filterexpr/date_test.go`: textual `date`/`timestamp` RFC3339 == raw; closed range; between; pre-epoch rejection; garbage/`$param` rejection; keywords still usable as bare values.
- `internal/pkg/filterexpr/decode_test.go`: dual-format equivalence rows for date/timestamp; per-target acceptance/rejection matrix (both forms); pre-epoch rejection (both forms). Existing audit datetime tests unchanged and green.

## Gates

- `GOROOT= go build ./...`: green
- `GOROOT= go test ./internal/pkg/filterexpr/... ./internal/proto/commonpb/... ./internal/query/... ./internal/domain/... ./internal/adapter/http/...`: green
- `go generate ./... && go mod tidy`: no drift beyond intended edits
- `golangci-lint run --fix` on changed packages: 0 issues

## Blast radius

`buildRange`/`uintConditionFromBounds` are internal to `query_filter.go` (reached only via `QueryFilter.UnmarshalJSON`); the change is additive (raw micros still parse). The textual grammar addition is additive. `ParseDatetimeMicros` is untouched. Risk: LOW — no FSM/storage/hot-path code touched.

### DIFF
--- a/internal/pkg/filterexpr/format.go
+++ b/internal/pkg/filterexpr/format.go
@@ -45,11 +45,11 @@ func formatFilter(f *commonpb.QueryFilter) (string, int) {
 	case *commonpb.QueryFilter_AccountHasAsset:
 		return formatAccountHasAsset(v.AccountHasAsset), precLeaf
 	case *commonpb.QueryFilter_Audit:
-		return formatAuditCondition(v.Audit), precLeaf
+		return formatAuditCondition(v.Audit)
 	case *commonpb.QueryFilter_BuiltinUint:
-		return formatBuiltinUintCondition(v.BuiltinUint), precLeaf
+		return formatBuiltinUintCondition(v.BuiltinUint)
 	case *commonpb.QueryFilter_LogBuiltinUint:
-		return formatLogBuiltinUintCondition(v.LogBuiltinUint), precLeaf
+		return formatLogBuiltinUintCondition(v.LogBuiltinUint)
 	default:
 		return "<unknown filter>", precLeaf
 	}
@@ -61,9 +61,9 @@ func formatFilter(f *commonpb.QueryFilter) (string, int) {
 // it is the only one we emit — advertising id/insertedAt/revertedAt here would
 // produce strings Parse cannot re-read, breaking the config export/apply
 // round-trip. Its bounds render as quoted RFC3339 (EN-1544).
-func formatBuiltinUintCondition(bc *commonpb.BuiltinUintCondition) string {
+func formatBuiltinUintCondition(bc *commonpb.BuiltinUintCondition) (string, int) {
 	if bc.GetField() != commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_TIMESTAMP {
-		return "<unknown builtin field>"
+		return "<unknown builtin field>", precLeaf
 	}
 
 	return formatDateUintCondition("timestamp", bc.GetCond())
@@ -71,12 +71,12 @@ func formatBuiltinUintCondition(bc *commonpb.BuiltinUintCondition) string {
 
 // formatLogBuiltinUintCondition renders a log builtin uint range. The only field
 // the textual grammar reads back is `date` (quoted RFC3339 output).
-func formatLogBuiltinUintCondition(lc *commonpb.LogBuiltinUintCondition) string {
+func formatLogBuiltinUintCondition(lc *commonpb.LogBuiltinUintCondition) (string, int) {
 	switch lc.GetField() {
 	case commonpb.LogBuiltinIndex_LOG_BUILTIN_INDEX_DATE:
 		return formatDateUintCondition("date", lc.GetCond())
 	default:
-		return "<unknown log field>"
+		return "<unknown log field>", precLeaf
 	}
 }
 
@@ -91,34 +91,37 @@ func formatLogBuiltinUintCondition(lc *commonpb.LogBuiltinUintCondition) string 
 // back into the same single condition (see foldDateRangeAnd), so the exclusivity
 // survives the round-trip instead of being silently widened. This is the
 // top-level counterpart of formatAuditUintCondition (prefixed with `audit[...]`).
-func formatDateUintCondition(field string, uc *commonpb.UintCondition) string {
+func formatDateUintCondition(field string, uc *commonpb.UintCondition) (string, int) {
 	render := func(v uint64) string {
 		return strconv.Quote(time.UnixMicro(int64(v)).UTC().Format(time.RFC3339Nano))
 	}
 
 	if uc.Min != nil && uc.Max != nil && uc.GetMin() == uc.GetMax() && !uc.GetMinExclusive() && !uc.GetMaxExclusive() {
-		return fmt.Sprintf("%s == %s", field, render(uc.GetMin()))
+		return fmt.Sprintf("%s == %s", field, render(uc.GetMin())), precLeaf
 	}
 
 	if uc.Min != nil && uc.Max != nil {
 		if !uc.GetMinExclusive() && !uc.GetMaxExclusive() {
-			return fmt.Sprintf("%s between %s and %s", field, render(uc.GetMin()), render(uc.GetMax()))
+			return fmt.Sprintf("%s between %s and %s", field, render(uc.GetMin()), render(uc.GetMax())), precLeaf
 		}
 
+		// An exclusive bound folds into two comparison clauses joined by
+		// `and`; report precAnd so a wrapping `not` parenthesizes the pair
+		// (otherwise `not a and b` mis-associates as `(not a) and b`).
 		return fmt.Sprintf("%s %s %s and %s %s %s",
 			field, lowerOp(uc.GetMinExclusive()), render(uc.GetMin()),
-			field, upperOp(uc.GetMaxExclusive()), render(uc.GetMax()))
+			field, upperOp(uc.GetMaxExclusive()), render(uc.GetMax())), precAnd
 	}
 
 	if uc.Min != nil {
-		return fmt.Sprintf("%s %s %s", field, lowerOp(uc.GetMinExclusive()), render(uc.GetMin()))
+		return fmt.Sprintf("%s %s %s", field, lowerOp(uc.GetMinExclusive()), render(uc.GetMin())), precLeaf
 	}
 
 	if uc.Max != nil {
-		return fmt.Sprintf("%s %s %s", field, upperOp(uc.GetMaxExclusive()), render(uc.GetMax()))
+		return fmt.Sprintf("%s %s %s", field, upperOp(uc.GetMaxExclusive()), render(uc.GetMax())), precLeaf
 	}
 
-	return field + " <uint?>"
+	return field + " <uint?>", precLeaf
 }
 
 // lowerOp / upperOp map a bound's exclusivity to its DSL comparison operator.
@@ -152,19 +155,21 @@ var auditFieldNames = map[commonpb.AuditField]string{
 
 // formatAuditCondition renders an AuditCondition back into `audit[field] OP
 // value`, inverse of the parser's AuditCond production.
-func formatAuditCondition(ac *commonpb.AuditCondition) string {
+func formatAuditCondition(ac *commonpb.AuditCondition) (string, int) {
 	key, ok := auditFieldNames[ac.GetField()]
 	if !ok {
-		return "audit[<unknown>]"
+		return "audit[<unknown>]", precLeaf
 	}
 
 	switch cond := ac.GetCondition().(type) {
 	case *commonpb.AuditCondition_StringCond:
-		return fmt.Sprintf("audit[%s] == %s", key, formatStringCondValue(cond.StringCond))
+		return fmt.Sprintf("audit[%s] == %s", key, formatStringCondValue(cond.StringCond)), precLeaf
 	case *commonpb.AuditCondition_UintCond:
-		return formatAuditUintCondition(key, ac.GetField(), cond.UintCond)
+		// Audit ranges always render as `between`/single-bound (never an
+		// `and`-join), so they are always leaf-precedence.
+		return formatAuditUintCondition(key, ac.GetField(), cond.UintCond), precLeaf
 	default:
-		return fmt.Sprintf("audit[%s] <unknown>", key)
+		return fmt.Sprintf("audit[%s] <unknown>", key), precLeaf
 	}
 }
 
--- a/internal/pkg/filterexpr/format.go
+++ b/internal/pkg/filterexpr/format.go
@@ -55,61 +55,45 @@ func formatFilter(f *commonpb.QueryFilter) (string, int) {
 	}
 }
 
-// txBuiltinFieldNames is the reverse of the parser/codec txBuiltinFieldToJSON:
-// tx builtin enum -> DSL field name. Only the date-semantic fields plus id are
-// range-expressible in the DSL.
-var txBuiltinFieldNames = map[commonpb.TransactionBuiltinIndex]string{
-	commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_ID:          "id",
-	commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_TIMESTAMP:   "timestamp",
-	commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_INSERTED_AT: "insertedAt",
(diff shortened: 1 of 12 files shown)


internal/pkg/filterexpr/parser.go
```diff
@@ -49,7 +49,7 @@ var filterLexer = lexer.MustSimple([]lexer.SimpleRule{
 	{Name: "Dollar", Pattern: `\$`},
 	{Name: "String", Pattern: `"[^"]*"|'[^']*'`},
 	{Name: "Comma", Pattern: `,`},
-	{Name: "Keyword", Pattern: `\b(and|or|not|in|between|metadata|address|source|destination|ledger|audit|exists|true|false)\b`},
+	{Name: "Keyword", Pattern: `\b(and|or|not|in|between|metadata|address|source|destination|ledger|audit|date|timestamp|exists|true|false)\b`},
```
NumaryBot: 🟠 **[major] Avoid splitting date-prefixed identifiers**

When an existing textual filter uses an unquoted identifier that starts with `date` or `timestamp` followed by allowed identifier punctuation, such as `metadata[date-range] == x` or `metadata[k] == timestamp:created`, this keyword rule now tokenizes only the prefix because `\b` matches before `-`, `:`, `.`, and `/`. The parser then sees extra tokens and rejects filters that were valid before; the keyword boundary needs to exclude the full Ident continuation set or otherwise only reserve exact identifiers.

internal/pkg/filterexpr/parser.go
```diff
@@ -415,6 +432,63 @@ func (a *AuditCond) stringToProto(field commonpb.AuditField) (*commonpb.QueryFil
...
+	}
+
+	uc, err := uintConditionFromOp(d.Op, parse)
+	if err != nil {
+		return nil, fmt.Errorf("%s field: %w", d.Field, err)
+	}
+
+	switch d.Field {
+	case "date":
+		return &commonpb.QueryFilter{Filter: &commonpb.QueryFilter_LogBuiltinUint{
```
NumaryBot, line 589: 🟡 **[minor] Format the new date filter oneofs**

When a textual `date` filter is parsed and later displayed (for example, prepared-query listings call `filterexpr.Format`), it now produces a `QueryFilter_LogBuiltinUint`, but `Format` has no case for `LogBuiltinUint` or `BuiltinUint` and falls through to `<unknown filter>`. Please add formatting for these new parser outputs so saved date/timestamp filters round-trip instead of being shown as unknown.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1580#issuecomment-4958143344

internal/pkg/filterexpr/parser.go
gfyrag: Fixed in 0d42f2027 (reproduced first: `metadata[date-range] == "x"`, `metadata[timestamp-created]`, `metadata[k] == timestamp:created`, `date-2023` all failed on the branch). Root cause: adding `date`/`timestamp` to the `Keyword` lexer rule with `\b` — Go's `\b` treats `-`/`:`/`.`/`/` as non-word, so it matched right after the prefix and tokenized only `date`/`timestamp`. Fix: removed `date`/`timestamp` from the Keyword rule entirely and match the `'date'`/`'timestamp'` grammar literals against the plain `Ident` token (participle matches literals by value) — so they act as date fields at condition position while remaining ordinary identifiers everywhere else; `audit[timestamp]` still parses. No compat shim. Regression test `TestParseDate_DoesNotMistokenizeIdentifierPrefix` covers `-`/`.`/`/`/`:` prefixed keys + bare values. (Note: the base branch has the same pre-existing mis-tokenization for `ledger`/`address` prefixes — left out of scope here; happy to open a separate follow-up.)

internal/pkg/filterexpr/parser.go
gfyrag: Fixed in 0d42f2027. Added `formatBuiltinUintCondition` (timestamp/insertedAt/revertedAt → quoted RFC3339, id → raw int) and `formatLogBuiltinUintCondition` (date → quoted RFC3339) to `formatFilter`, mirroring the audit datetime renderer, so textual date/timestamp filters no longer render as `<unknown filter>`. `TestFormatDate_RoundTrip` asserts Format→reparse `proto.Equal` (incl. raw-micros normalizing to canonical quoted RFC3339).

internal/pkg/filterexpr/format.go
```diff
@@ -46,11 +46,101 @@ func formatFilter(f *commonpb.QueryFilter) (string, int) {
...
+			return strconv.Quote(time.UnixMicro(int64(v)).UTC().Format(time.RFC3339Nano))
+		}
+	}
+
+	if uc.Min != nil && uc.Max != nil && uc.GetMin() == uc.GetMax() && !uc.GetMinExclusive() && !uc.GetMaxExclusive() {
+		return fmt.Sprintf("%s == %s", field, render(uc.GetMin()))
+	}
+
+	if uc.Min != nil && uc.Max != nil {
+		return fmt.Sprintf("%s between %s and %s", field, render(uc.GetMin()), render(uc.GetMax()))
```
flemzord: 🔴 **[high] Preserve exclusive bounds when formatting closed ranges**

A structured filter such as an `$and` of `$gt` and `$lt` on `timestamp` folds into one `UintCondition` with both exclusive flags, but this branch always emits `timestamp between ... and ...`; `between` parses both bounds as inclusive. `ledgerctl ledgers configuration export` calls `Format`, and config apply parses that string again, so reapplying an exported prepared query silently widens it and can change LIST or aggregate results at the boundaries. Emit the original comparison clauses whenever either exclusivity flag is set, and cover an exclusive structured range through Format -> Parse (or the full config export/import path).

internal/pkg/filterexpr/format.go
```diff
@@ -46,11 +46,101 @@ func formatFilter(f *commonpb.QueryFilter) (string, int) {
...
 		return "<unknown filter>", precLeaf
 	}
 }
 
+// txBuiltinFieldNames is the reverse of the parser/codec txBuiltinFieldToJSON:
+// tx builtin enum -> DSL field name. Only the date-semantic fields plus id are
+// range-expressible in the DSL.
+var txBuiltinFieldNames = map[commonpb.TransactionBuiltinIndex]string{
+	commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_ID:          "id",
+	commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_TIMESTAMP:   "timestamp",
```
flemzord: 🟡 **[medium] Only emit fields the textual parser can read back**

This map makes `Format` produce `id`, `insertedAt`, and `revertedAt` expressions, but `DateCond` only admits `date` and `timestamp`; `Parse(Format(filter))` therefore fails for those three arms. This is observable through ledger configuration export/apply for prepared queries created with the structured JSON DSL. Either extend the textual grammar and round-trip tests for the mapped fields, or scope this formatter to the two EN-1544 fields instead of advertising an inverse that is not parseable.

flemzord (changes requested):
One blocking formatter round-trip issue remains. The two earlier lexer/unknown-filter findings are fixed on this head, and the shared coercion matches EN-1544.

internal/pkg/filterexpr/format.go
gfyrag: Fixed in a8033eca4. Root cause: `Format` always emitted `between` for a two-bound range, and `between` reparses inclusive on both ends — so a JSON `$and` of `$gt`/`$lt` (folded by `commonpb.foldRangeAnd` into one UintCondition with exclusive flags) silently widened at the boundaries on Format→Parse. Fix, two halves for `proto.Equal` round-trip: (1) Format emits `between` only when BOTH bounds are inclusive; if either is exclusive it emits `field > lo and field < hi` (lowerOp/upperOp helpers). (2) Added the textual counterpart `foldDateRangeAnd` in `AndExpr.toProto` (mirrors `commonpb.foldRangeAnd`) so those two same-field bounds re-fold into one condition preserving each side's exclusivity. `TestFormatDate_PreservesExclusiveClosedRange` covers both-exclusive / upper-only / lower-only / both-inclusive. audit[timestamp] closed ranges are unaffected (audit is textual-only, not folded).

internal/pkg/filterexpr/format.go
gfyrag: Fixed in a8033eca4. The tx-builtin field map advertised `id`/`insertedAt`/`revertedAt`, but `DateCond` only reads back `date`/`timestamp`, so `Parse(Format(f))` failed for those arms. Dropped the maps: `formatBuiltinUintCondition` now emits only `timestamp` (tx) and `formatLogBuiltinUintCondition` only `date` (log) — exactly the set the grammar admits — and any other builtin field renders as `<unknown builtin field>` rather than a non-reparseable string. `TestFormatDate_OnlyEmitsParseableFields` asserts id/insertedAt/revertedAt → unknown, and timestamp+date round-trip for every single-bound op.

internal/pkg/filterexpr/format.go
```diff
@@ -46,11 +46,98 @@ func formatFilter(f *commonpb.QueryFilter) (string, int) {
 		return formatAccountHasAsset(v.AccountHasAsset), precLeaf
 	case *commonpb.QueryFilter_Audit:
 		return formatAuditCondition(v.Audit), precLeaf
+	case *commonpb.QueryFilter_BuiltinUint:
+		return formatBuiltinUintCondition(v.BuiltinUint), precLeaf
```
flemzord: 🔴 **[high] Preserve precedence when a range expands to `and`**

`formatDateUintCondition` now expands an exclusive two-sided range into two clauses joined by `and`, but this case still reports `precLeaf` (the log arm below does the same). Under `not`, `Format` therefore emits `not timestamp > lo and timestamp < hi`; `Parse` reads that as `(not lower) and upper`, rather than `not (lower and upper)`, changing the prepared-query meaning during export/re-apply. Return `precAnd` for the expanded shape (or parenthesize it) and add a nested `not` Format → Parse regression.

flemzord (changes requested):
The earlier top-level exclusive-range widening is fixed. One blocking nested-format round-trip issue remains; the current E2E failure is in `get_ledger_snapshot_test.go` and is unrelated to this PR date-filter path.

internal/pkg/filterexpr/format.go
gfyrag: Fixed in 91349ef66. `formatDateUintCondition` now returns its precedence: `precAnd` for the exclusive and-join form, `precLeaf` otherwise, threaded up through `formatBuiltinUintCondition`/`formatLogBuiltinUintCondition`. `formatWithPrec` then parenthesizes under `not`, so an exclusive range renders as `not (timestamp > lo and timestamp < hi)` and round-trips. Audit ranges always render as `between`/single-bound (never an and-join), so they stay `precLeaf`. Regression added: `TestFormatDate_NotWrapsExclusiveRange` (both-exclusive, plus the inclusive `between` case that must NOT gain parens).

internal/pkg/filterexpr/format.go
```diff
@@ -45,12 +45,102 @@ func formatFilter(f *commonpb.QueryFilter) (string, int) {
...
+// A closed range only renders as `between` when BOTH bounds are inclusive, since
+// `between` parses back inclusive on both ends. If either bound is exclusive (the
+// shape a JSON `$and` of `$gt`/`$lt` folds into) it renders as two comparison
+// clauses joined by `and` — `field > lo and field < hi` — which the parser folds
+// back into the same single condition (see foldDateRangeAnd), so the exclusivity
+// survives the round-trip instead of being silently widened. This is the
+// top-level counterpart of formatAuditUintCondition (prefixed with `audit[...]`).
+func formatDateUintCondition(field string, uc *commonpb.UintCondition) (string, int) {
+	render := func(v uint64) string {
+		return strconv.Quote(time.UnixMicro(int64(v)).UTC().Format(time.RFC3339Nano))
```
flemzord: 🟡 **[medium] Preserve raw uint64 bounds outside RFC3339 range**

This can make configuration export/apply fail for a valid raw-microsecond filter when the bound is outside the RFC3339 year range (or above `math.MaxInt64`). The decoder explicitly accepts the full uint64 raw form, but converting through `int64(v)` can wrap and `time.Format` can emit a timestamp that `CoerceDatetimeMicros` cannot parse back. Consider falling back to `strconv.FormatUint(v, 10)` whenever the RFC3339 rendering does not round-trip, and cover a bound above year 9999 / `math.MaxUint64`.

flemzord (approved):
The blocking nested `not` round-trip issue is fixed on this head and the targeted filterexpr suite passes. I found one non-blocking uint64/RFC3339 edge case inline; there are no remaining High/Critical findings.

internal/pkg/filterexpr/format.go
gfyrag: Confirmed and fixed in dd3c54d81. You're right — `CoerceDatetimeMicros` accepts the full uint64 raw range, so the formatter must round-trip anything it accepts. Extracted a shared `renderDatetimeBound` helper (used by both `formatDateUintCondition` and `formatAuditUintCondition`) that emits quoted RFC3339 only when it round-trips through the actual decoder (`v <= MaxInt64 && CoerceDatetimeMicros(s) == v`), else falls back to `strconv.FormatUint(v, 10)`. Regression `TestFormatDate_RawMicrosOutsideRFC3339Range` covers a year > 9999 bound (≤ MaxInt64, RFC3339 5-digit-year) and `math.MaxUint64` (int64 wrap), on both the transaction and `audit[timestamp]` arms.

PR #1582, opened 2026-07-13
fix(EN-1547): tighten filter DSL identifiers — special-char keys/values must be quoted

gfyrag:
## What

Closes EN-1547.

The filterexpr textual DSL let a bare identifier contain `-`, `:`, `.`, `/` (the `Ident` lexer pattern was `[a-zA-Z_][a-zA-Z0-9_:.\-/]*`). That was the sole reason the keyword `\b` boundary could match inside a punctuated identifier and reject valid filters like `metadata[ledger-x]` / `source.id` / `address:foo`.

**Design (no compat shim, pre-GA — coherence over stability):** a bare identifier is now **plain-alphanumeric only**, and any key or value containing a special character **must be quoted**. This removes the mis-tokenization at the root — a keyword can no longer be the prefix of a punctuated bare identifier because no such identifier exists.

## Changes

- **Tighten `Ident`** to `[a-zA-Z_][a-zA-Z0-9_]*` (drop `:`, `.`, `-`, `/`). Because punctuated bare identifiers no longer exist, the noun/field keywords stay **exactly** as on `release/v3.0` — no keyword demotion, no `Value.Kw` change. The grammar is `release/v3.0` + three surgical changes.
- **Add an `AssetRef` token** (`[a-zA-Z][a-zA-Z0-9]*/[0-9]+`, matched before `Ident`) so the base/precision asset reference `has asset USD/2` stays bare — it is a first-class structured literal, not a free-form string. `AssetCond` now reads `@(AssetRef | Ident)`.
- **Quoted keys/values are reachable in every position.** The metadata-key slot already accepts `String` (`metadata["x-request-id"]`); the value position accepts `String` (`== "date-2023"`, `reference == "..."`, `address ^= "users:"`, `ledger == "..."`, `audit[ledger] == "my-ledger"`).
- **Format** (`needsQuoting`): quote any key/value that is not a plain bare `Ident` (bareIdent regex) or that collides with a structural operator, so round-trip holds now that more values require quoting. Metadata keys are quoted too. This also fixes the pre-existing `in`-as-value round-trip gap.

## Repro → fix (on release/v3.0)

Failed on base, now parse **when quoted** (and are intentionally rejected bare): `metadata["ledger-x"]`, `metadata["source.id"]`, `address ^= "users:"`, `audit[caller_subject] == "svc:payments"`, `metadata["x-request-id"]`, … And `has asset USD/2` stays bare via the AssetRef token.

## Blast radius (breaking, accepted — no compat)

Bare punctuated filter identifiers are now rejected. In-repo surface relying on them, fixed to quoted:
- **1 CLI help example** — `cmd/ledgerctl/audit/list.go` (`audit[ledger] == "my-ledger"`).
- **2 docs examples** — `docs/ops/cli.md` (`audit[ledger] == "my-ledger"`, `address ^= "users:"`).
- **5 test assertions** — `parser_test.go` (`address ^= "users:"`, `has:wallet` → now assert quoted + bare-rejected), `audit_test.go` (`caller_subject == "svc:payments"`), `format_test.go` (address/source punctuated Format expectations → quoted).

No non-test / library callers relied on bare punctuated forms (`--filter` / prepared-query filters are user-supplied strings; no hardcoded punctuated DSL in tests/e2e/scenarios).

## Tests

`internal/pkg/filterexpr/quoting_test.go`:
- Punctuated metadata **key** parses only quoted, rejected bare (× `-`/`:`/`.`/`/`).
- Punctuated **value** parses only quoted, rejected bare — metadata / address exact / address prefix / ledger (× each special char).
- Punctuated **audit** value quoted-only.
- `AssetRef` (`USD/2`) still bare; bare base still bare.
- Plain-alphanumeric keys/values still bare (no needless quoting); noun words still usable as bare values.
- Format round-trips special-char keys/values by quoting (`proto.Equal`), incl. reserved operators as string values.

Docs (`docs/ops/cli.md`) grammar section documents the plain-alphanumeric bare-Ident rule and the quoting requirement.

## Gates

- `GOROOT= go build ./...`: green
- `GOROOT= go test ./internal/pkg/filterexpr/... ./internal/proto/commonpb/... ./internal/adapter/http/...` (+ ledgerctl): green
- `go generate ./... && go mod tidy`: no drift
- `golangci-lint run --fix`: 0 issues


internal/pkg/filterexpr/parser.go
```diff
@@ -49,7 +66,7 @@ var filterLexer = lexer.MustSimple([]lexer.SimpleRule{
 	{Name: "Dollar", Pattern: `\$`},
 	{Name: "String", Pattern: `"[^"]*"|'[^']*'`},
 	{Name: "Comma", Pattern: `,`},
-	{Name: "Keyword", Pattern: `\b(and|or|not|in|between|metadata|address|source|destination|ledger|audit|exists|true|false)\b`},
+	{Name: "Keyword", Pattern: `\b(and|or|not|in|between)\b`},
```
flemzord: [P2] Preserve structural-keyword prefixes as identifiers

This keeps the same `\\b` boundary failure for valid identifiers that start with a structural word: `metadata[and-x] == v`, `metadata[or:flag] == v`, and bare values such as `between/value` still fail at the punctuation. It also violates the EN-1547 round-trip criterion: parsing `metadata[k] == "and-x"` succeeds, but `Format` emits `metadata[k] == and-x`, which then fails to parse because `needsQuoting` only reserves the exact word. I reproduced all five operator prefixes against this head. Please make keyword recognition use a true DSL-token boundary (or otherwise protect these prefixes) and add key, value, and `Format -> Parse` regression cases.

internal/pkg/filterexpr/parser.go
```diff
@@ -30,9 +30,26 @@ const MaxParseDepth = 200
 // indicators exceed MaxParseDepth.
 var ErrFilterTooDeep = fmt.Errorf("filter expression nesting exceeds maximum depth (%d)", MaxParseDepth)
 
-// Custom lexer: Keywords are matched before Ident so that reserved words
-// (and, or, not, between, metadata, address, source, destination, exists,
-// true, false) cannot be consumed as bare values.
+// Custom lexer: only the STRUCTURAL operators (and, or, not, in, between) are
+// lexer keywords, matched before Ident so they always terminate expressions and
+// cannot be swallowed as bare values. Every "noun"/field word (metadata, address,
+// source, destination, ledger, audit, exists) and the boolean literals (true,
```
flemzord: [P3] Update the documented reserved-word contract

This changes which words are reserved, but `docs/technical/architecture/subsystems/read-path/prepared-queries.md` still lists `metadata`, `address`, `source`, `destination`, `exists`, `true`, and `false` as reserved. The repository documentation-maintenance rule requires behavior changes to keep the architecture/CLI contract current. Please update that reserved-word paragraph (and any equivalent grammar wording) with the new structural-only set.

internal/pkg/filterexpr/format.go
```diff
@@ -438,22 +442,30 @@ func quoteIfNeeded(s string) string {
...
-	"and": true, "or": true, "not": true, "between": true,
-	"metadata": true, "address": true, "source": true, "destination": true,
-	"exists": true, "true": true, "false": true, "has": true, "asset": true,
+// reservedKeywords is exactly the set of lexer keywords (the structural
+// operators). A bare value equal to one of these would re-lex as that operator
+// token and fail to reparse as a value, so Format must quote it. The noun words
+// (metadata/address/…/audit) and the boolean literals are NOT here: they lex as
+// Ident and round-trip bare — quoting them would be spurious. Kept in lockstep
+// with the lexer `Keyword` rule in parser.go.
+var reservedKeywords = map[string]bool{
```
flemzord: [P3] Group the adjacent package variables

The repository convention requires related variable declarations to use one `var (...)` block. Please group `reservedKeywords` with `bareIdent` below; the same applies to the adjacent `demotedKeywords` and `identContinuations` declarations in the new test file.

flemzord (approved):
No High/Critical blocker found, so I am approving under the current review policy. I left one validated P2 for structural-keyword prefixes, plus two P3 standards/documentation follow-ups.

gfyrag:
Re-reviewed current head `3dc5b5a70812a8f4b4c75dae051138470461cb10`. I cannot submit a formal `REQUEST_CHANGES` from this GitHub identity because it is the PR author, but I would treat the following as blocking before merge.

1. **P1: `Format` still drops `ledger` filters to `<unknown filter>`.**

   The PR updates the parser/formatter contract and explicitly exercises quoted `ledger == "..."` parsing in the new quoting matrix, but `formatFilter` still has no `case *commonpb.QueryFilter_Ledger` (`internal/pkg/filterexpr/format.go:35-50`). `LedgerCond` does parse into `QueryFilter_Ledger` (`internal/pkg/filterexpr/parser.go:562-579`), so any parse/format round-trip involving `ledger == main`, `ledger == "my-ledger"`, or `ledger == $ledger` is lost as `<unknown filter>`.

   Please add a ledger formatter using the existing `formatStringCondValue` path and cover both plain and quoted ledger names in the round-trip tests. This is especially relevant here because the PR's stated contract is that quoting/formatting now preserves values requiring quotes.

2. **P2: some user-facing examples are invalid under the new quoted-only rule.**

   The new docs correctly say values containing `:` must be quoted, but several examples still use bare `address ^= users:`. These now fail to parse:

   - `cmd/ledgerctl/accounts/list.go:37`
   - `docs/ops/cli.md:1080`
   - `docs/ops/cli.md:1095`
   - `docs/ops/cli.md:1104`
   - `docs/ops/cli.md:1113`

   Please update them to `address ^= "users:"` so the CLI help and docs remain executable examples.

3. **P2: `quoteIfNeeded` can emit unparsable strings containing `"`.**

   `quoteIfNeeded` always wraps with double quotes (`internal/pkg/filterexpr/format.go:433-436`), but the lexer string pattern does not support escapes (`internal/pkg/filterexpr/parser.go:61`) and `unquote` only strips the delimiters (`internal/pkg/filterexpr/parser.go:910-915`). A value parsed from a single-quoted string such as `metadata[k] == 'a"b'` is representable today, but `Format` emits `metadata[k] == "a"b"`, which no longer reparses. The same applies to metadata keys.

   Either choose a quote delimiter that preserves the value when possible, or add explicit escaping support to the grammar/formatter. Please add a round-trip test for quote-containing keys/values so this stays pinned.

I also noticed the already-open P3 about the stale reserved-word paragraph in `docs/technical/architecture/subsystems/read-path/prepared-queries.md`; that should be fixed with the docs update, but the three points above are the blockers from my pass.

internal/pkg/filterexpr/parser.go
gfyrag: This is moot on the current head — the quote-strings redesign (3dc5b5a70, after your review) resolves all five cases. `needsQuoting = !bareIdent.Match(s) || reservedOperators[s]` with `bareIdent = ^[a-zA-Z_][a-zA-Z0-9_]*$`, so: `metadata["and-x"]` parses as a key (bare `metadata[and-x]` is rejected — the accepted quote-strings design, same as any punctuated identifier); `metadata[k] == "and-x"` parses as a value; Format quotes `and-x`/`or:flag` → reparse `proto.Equal`; and a structural-word-EXACT value `== "and"` is emitted quoted and round-trips. Locked in a5f694c60 with `TestQuoting_StructuralKeywordPrefixes` covering (a)-(d) for each of and/or/not/in/between × each special char + the exact-value + bare-rejected cases.

internal/pkg/filterexpr/parser.go
gfyrag: Fixed in a5f694c60. `prepared-queries.md` reserved-word paragraph now lists only the structural operators (`and`, `or`, `not`, `in`, `between`) and notes field/noun words + `true`/`false` are ordinary identifiers, bare Ident is plain-alphanumeric, and special-char strings must be quoted (with the `has asset BASE/PRECISION` exception). Grepped docs — no other equivalent reserved-word wording (the cli.md grammar note landed in the redesign commit).

internal/pkg/filterexpr/format.go
gfyrag: Fixed in a5f694c60 — `bareIdent` + `reservedOperators` grouped into one `var (...)` block in format.go. (The two test-file package vars aren't adjacent — each is colocated with the test that uses it ~185 lines apart — so no grouping there, per the adjacent-declarations convention.)

flemzord (approved):
Re-reviewed current head `a5f694c60fe41555ab98d22d1fe5229a320187f8` after the two follow-up commits. The previous structural-keyword-prefix P2 is resolved by the quoted-only redesign and its new matrix, and the reserved-word documentation plus adjacent variable grouping P3s are fixed. Targeted filter/parser/HTTP/CLI tests and the full GitHub check suite are green. I found no new High/Critical regression. The remaining formatter/example concerns already documented in the author’s top-level review are non-blocking under the current policy, so I am not duplicating them inline.

PR #1584, opened 2026-07-13
feat(EN-1549): harmonize audit filter DSL — bare target-aware fields

gfyrag:
## What & why

The filter DSL was asymmetric: audit fields used a namespaced bracket syntax (`audit[outcome] == failure`, `audit[ledger] == main`, `audit[seq] between 1000 and 2000`, …) while transaction/log fields were bare (`timestamp`, `date`, `ledger`, `has asset`, `metadata[key]`). EN-1549 harmonizes this by **dropping the `audit[...]` prefix** and exposing audit fields as **bare, target-aware** names:

| Before | After |
|---|---|
| `audit[outcome] == failure` | `outcome == failure` |
| `audit[ledger] == main` | `ledger == main` |
| `audit[seq] between 1000 and 2000` | `seq between 1000 and 2000` |
| `audit[timestamp] >= "2023-11-14T22:13:20Z"` | `timestamp >= "2023-11-14T22:13:20Z"` |
| `audit[order_type] in (create_transaction, revert_transaction)` | `order_type in (...)` |
| `audit[proposal_id] == 42` / `audit[log_seq] == 500` / `audit[caller_subject] == "svc:payments"` | `proposal_id == 42` / `log_seq == 500` / `caller_subject == "svc:payments"` |

## Target-aware resolution (the crux)

The DSL is context-free, but a few bare fields are ambiguous across targets:
- bare `timestamp` collides with the transaction `timestamp` (BuiltinUint arm) and the audit timestamp field;
- bare `ledger` collides with the top-level `LedgerCondition` and the audit ledger field.

Resolution is threaded through **`Parse(input string, target commonpb.QueryTarget)`** and dispatched at `toProto` time:
- **`QUERY_TARGET_AUDIT`** → all 8 audit fields resolve to the `AuditCondition` arm.
- **transactions / logs / accounts** → `timestamp` → transaction builtin range, `date` → log range, `ledger` → `LedgerCondition`; audit-only field names (`outcome`, `seq`, …) are rejected with `unknown field` (they carry no meaning off the audit target).

The existing per-target validity gate (`domain.ValidateFilterForTarget`, EN-1504/EN-1503) runs downstream unchanged and still rejects e.g. `date` on transactions.

Implementation: the grammar's separate `AuditCond` / `DateCond` / `LedgerCond` productions are unified into a single generic **`FieldCond`** (`FIELD OP value`) placed last in the `Condition` alternation (so the specialized `asset`/`metadata`/`address` productions keep winning their lead tokens). The `audit` lexer keyword is removed (it is now an ordinary identifier). The shared entry points already carry the target:
- `DecodeDualFormat(raw, target)` threads it straight through;
- `DecodeDualFormatStructuralOnly(raw)` (prepared-query update, which never sees the target — it lives on the stored query and the FSM re-validates) resolves with a fixed non-audit target. This is sound because prepared queries are only ever ACCOUNTS/TRANSACTIONS/LOGS (audit is gRPC-only) and every non-audit target resolves the bare intrinsic fields identically.

## Formatter

`Format` emits bare audit field names (inverse of `Parse` on the audit target). `Parse(Format(f))` round-trips for every audit filter shape — an audit filter is always re-parsed with `QUERY_TARGET_AUDIT`, which is what disambiguates the shared `timestamp`/`ledger` renderings.

## Breaking change (no compat)

This is a **breaking DSL change** and intentionally ships **without any backward-compatibility shim** — the old `audit[...]` syntax no longer parses. Pre-GA, coherence over stability.

## Out of scope

Unifying the `date` (log) vs `timestamp` (transaction) naming is explicitly **NOT** part of EN-1549 — those remain distinct field names. `metadata[key]` bracket syntax is untouched (a genuine key namespace, not a field prefix).

## Coverage
- **Parser/format/decode**: rewrote the filterexpr audit tests to the bare form; added target-aware resolution tests (bare `timestamp`/`ledger` resolve to different arms per target, audit-only fields rejected off-audit), round-trip tests, and fuzz seeds for the bare grammar (fuzzed across TRANSACTIONS + AUDIT targets, no crashes).
- **Query/adapter**: updated `audit_filter*` error messages/comments to bare fields; updated HTTP audit handler tests.
- **E2E**: updated `filterexpr.Parse` call sites (now target-parameterized) and audit test comments.
- **Docs**: `docs/ops/cli.md`, `docs/technical/contributing/api-comparison.md`, `docs/technical/architecture/subsystems/read-path/query-pipeline.md`, `docs/technical/architecture/subsystems/api/http-api.md`, `openapi.yml`, and the `.proto` comments (+ synced generated `.pb.go` comments) — every `audit[...]` example replaced with the bare form.

## Verification
- `GOROOT= go test ./internal/pkg/filterexpr/ ./internal/query/... ./internal/adapter/http/... ./internal/proto/commonpb/...` — pass
- `go test -tags e2e ./tests/e2e/...` compiles clean
- fuzz `FuzzFilterExprParse` — no crashes

### DIFF
--- a/cmd/ledgerctl/queries/update.go
+++ b/cmd/ledgerctl/queries/update.go
@@ -1,6 +1,7 @@
 package queries
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/pterm/pterm"
@@ -8,6 +9,7 @@ import (
 
 	"github.com/formancehq/ledger/v3/cmd/ledgerctl/cmdutil"
 	"github.com/formancehq/ledger/v3/internal/pkg/filterexpr"
+	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
 	"github.com/formancehq/ledger/v3/internal/proto/servicepb"
 )
 
@@ -51,15 +53,29 @@ func runUpdate(cmd *cobra.Command, args []string) error {
 
 	filterExpr, _ := cmd.Flags().GetString("filter")
 
+	// An update replaces the stored filter, so an empty --filter would decode to a
+	// nil filter and silently erase the prepared query's filter on the server
+	// (ValidateFilterForTarget accepts nil). Require a non-empty filter, matching
+	// the HTTP update handler's guard (handlers_update_prepared_query.go).
+	if filterExpr == "" {
+		return errors.New("--filter is required")
+	}
+
 	// The update carries only the new filter, not the target — the target is
 	// immutable and lives on the stored prepared query (the FSM re-validates the
 	// filter against it). DecodeDualFormatStructuralOnly is the shared "target not
 	// known here" entry point: it resolves bare fields with a non-audit target
 	// (prepared queries are never audit) and defers the per-target validity gate
 	// to the server (EN-1549).
-	filter, err := filterexpr.DecodeDualFormatStructuralOnly([]byte(filterExpr))
+	filter, err := filterexpr.DecodeDualFormatStructuralOnly([]byte(filterExpr), commonpb.QueryTarget_QUERY_TARGET_TRANSACTIONS)
 	if err != nil {
 		return fmt.Errorf("invalid filter expression: %w", err)
+	}
+
+	// A structurally-valid but conditionless filter (e.g. `""` quoted, whitespace)
+	// also decodes to nil; reject it for the same reason.
+	if filter == nil {
+		return errors.New("--filter must contain at least one condition")
 	}
 
 	ctx, cancel := cmdutil.GetContext(cmd)
--- a/cmd/ledgerctl/queries/update.go
+++ b/cmd/ledgerctl/queries/update.go
@@ -51,7 +51,13 @@ func runUpdate(cmd *cobra.Command, args []string) error {
 
 	filterExpr, _ := cmd.Flags().GetString("filter")
 
-	filter, err := filterexpr.Parse(filterExpr)
+	// The update carries only the new filter, not the target — the target is
+	// immutable and lives on the stored prepared query (the FSM re-validates the
+	// filter against it). DecodeDualFormatStructuralOnly is the shared "target not
+	// known here" entry point: it resolves bare fields with a non-audit target
+	// (prepared queries are never audit) and defers the per-target validity gate
+	// to the server (EN-1549).
+	filter, err := filterexpr.DecodeDualFormatStructuralOnly([]byte(filterExpr))
 	if err != nil {
 		return fmt.Errorf("invalid filter expression: %w", err)
 	}
--- a/docs/ops/cli.md
+++ b/docs/ops/cli.md
@@ -1000,11 +1000,16 @@ has_asset_cond := "has" "asset" ASSET_BASE ["/" PRECISION]
 A bare `KEY` or `VALUE` is a plain-alphanumeric identifier (`[a-zA-Z_][a-zA-Z0-9_]*`).
 Any key or value that contains a special character — `-`, `:`, `.`, `/`, spaces, etc.
 — **must be quoted** (e.g. `metadata["x-request-id"] == "user:42"`,
-`address ^= "users:"`, `audit[ledger] == "my-ledger"`). The one exception is the
+`address ^= "users:"`, `ledger == "my-ledger"`). The one exception is the
 asset reference in `has asset BASE/PRECISION`, whose `/` is part of the grammar
 (e.g. `has asset USD/2`).
 
-`audit[...]` conditions are only valid on `audit list` (see [audit list](#audit-list) for the field/operator matrix). They are rejected on other list endpoints.
+Bare audit fields (`outcome`, `ledger`, `seq`, `proposal_id`, `timestamp`,
+`log_seq`, `caller_subject`, `order_type`) are only valid on `audit list` (see
+[audit list](#audit-list) for the field/operator matrix). They are written
+without any prefix and resolved against the audit query target (EN-1549 — this
+replaced the old `audit[...]` namespaced syntax, a breaking change with no
+backward compatibility). They are rejected on other list endpoints.
 
 **Conditions:**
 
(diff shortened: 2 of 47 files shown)


internal/pkg/filterexpr/decode.go
```diff
@@ -71,12 +77,22 @@ func DecodeDualFormat(raw []byte, target commonpb.QueryTarget) (*commonpb.QueryF
...
+// Bare-field resolution still needs a target (EN-1549), and here it is fixed to a
+// non-audit target. This is sound because prepared queries are only ever
+// ACCOUNTS/TRANSACTIONS/LOGS (audit is gRPC-only, never a prepared-query target —
+// see http.preparedQueryTargets), and every non-audit target resolves the bare
+// intrinsic fields identically (`timestamp` → transaction range, `date` → log
+// range, `ledger` → LedgerCondition). Whether the resolved field is actually
+// valid on the query's specific stored target is still enforced downstream by the
+// FSM's ValidateFilterForTarget — this path only picks the proto arm, not the
+// validity.
 func DecodeDualFormatStructuralOnly(raw []byte) (*commonpb.QueryFilter, error) {
```
NumaryBot: 🟠 **[major] Decode audit CLI filters with the audit target**

Because `cmdutil.BuildQueryFilter` is still used by `ledgerctl audit list`, this hard-coded non-audit target makes the newly documented bare audit filters fail before they reach the server: `outcome == failure` returns `unknown field`, while shared names like `ledger == main` are encoded as non-audit conditions and then rejected by the audit compiler. The audit list command needs a target-aware decode path instead of this structural-only helper.

cmd/ledgerctl/queries/update.go
```diff
@@ -51,7 +51,13 @@ func runUpdate(cmd *cobra.Command, args []string) error {
...
 	filterExpr, _ := cmd.Flags().GetString("filter")
 
-	filter, err := filterexpr.Parse(filterExpr)
+	// The update carries only the new filter, not the target — the target is
+	// immutable and lives on the stored prepared query (the FSM re-validates the
+	// filter against it). DecodeDualFormatStructuralOnly is the shared "target not
+	// known here" entry point: it resolves bare fields with a non-audit target
+	// (prepared queries are never audit) and defers the per-target validity gate
+	// to the server (EN-1549).
+	filter, err := filterexpr.DecodeDualFormatStructuralOnly([]byte(filterExpr))
```
NumaryBot: 🟠 **[major] Reject empty query updates before decoding**

When `ledgerctl queries update` is run without `--filter` or with an empty value, `DecodeDualFormatStructuralOnly` returns `(nil, nil)`, so the CLI now sends an update with a nil filter; the FSM treats nil as valid and overwrites the stored prepared query filter. Previously `Parse("")` errored here, and the HTTP update path still rejects nil filters, so this can accidentally clear a query instead of reporting a missing filter.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1584#issuecomment-4960609664

cmd/ledgerctl/queries/update.go
flemzord: 🔴 **[high] Still present on `1a36343c`**: an empty `--filter` decodes to nil; the gRPC endpoint accepts it, `ValidateFilterForTarget(nil, ...)` succeeds, and `processUpdatePreparedQuery` assigns that nil filter. This remains a destructive CLI regression rather than a validation error.

internal/pkg/filterexpr/decode.go
flemzord: 🔴 **[high] Still present on `1a36343c`**: `ledgerctl audit list` still reaches `BuildQueryFilter`, which calls this structural-only decoder with the transactions target. The primary EN-1549 CLI path therefore cannot parse its documented audit-only fields and mis-encodes shared fields.

flemzord (changes requested):
Requesting changes for the two unresolved High-severity functional regressions documented inline. On the current head, the advertised audit CLI filter path still cannot decode bare audit fields with the audit target, and an empty prepared-query CLI update can still erase the stored filter. The remaining lint failure is also visible in CI, but is not part of the blocking severity decision.

internal/pkg/filterexpr/decode.go
gfyrag: Fixed in 0fe1ba46b. `BuildQueryFilter` and `DecodeDualFormatStructuralOnly` now take the endpoint query target; each list command passes its own — `audit list` → `QUERY_TARGET_AUDIT`. So the CLI audit path decodes bare audit fields correctly (`outcome == failure`, and bare `ledger`/`timestamp` resolve to the audit arms), instead of the hardcoded TRANSACTIONS target. Regression tests added in `flags_test` (BuildQueryFilter bare-audit resolution) and `decode_test` (audit-target resolution). Server-side per-target validity gate unchanged.

cmd/ledgerctl/queries/update.go
gfyrag: Fixed in 0fe1ba46b, dual-layer. An empty/nil `--filter` on `queries update` is now rejected: the CLI/HTTP update path returns a validation error, and `processUpdatePreparedQuery` rejects a nil filter with the new `ErrPreparedQueryFilterRequired` sentinel (the audit/replay-determinism layer) — so a stored prepared query can never lose its definition by being overwritten with nil. Regression test in `processor_prepared_query_test`; HTTP missing/empty-filter rejection already covered.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1584#issuecomment-4960609664

docs/ops/cli.md
```diff
@@ -1000,11 +1000,16 @@ has_asset_cond := "has" "asset" ASSET_BASE ["/" PRECISION]
...
 Any key or value that contains a special character — `-`, `:`, `.`, `/`, spaces, etc.
 — **must be quoted** (e.g. `metadata["x-request-id"] == "user:42"`,
-`address ^= "users:"`, `audit[ledger] == "my-ledger"`). The one exception is the
+`address ^= "users:"`, `ledger == "my-ledger"`). The one exception is the
 asset reference in `has asset BASE/PRECISION`, whose `/` is part of the grammar
 (e.g. `has asset USD/2`).
 
-`audit[...]` conditions are only valid on `audit list` (see [audit list](#audit-list) for the field/operator matrix). They are rejected on other list endpoints.
+Bare audit fields (`outcome`, `ledger`, `seq`, `proposal_id`, `timestamp`,
+`log_seq`, `caller_subject`, `order_type`) are only valid on `audit list` (see
```
flemzord, line 1008: 🟠 **[medium] Keep the published grammar aligned with the breaking syntax change**\n\nThe grammar immediately above still defines `condition := ... | audit_cond` and `audit_cond := "audit" "[" AUDIT_FIELD "]" ...`, but the parser now rejects that syntax while this paragraph says the prefix was removed. Please replace that production with the new bare/generic field production so users do not copy a filter that always fails.

docs/technical/architecture/subsystems/api/http-api.md
```diff
@@ -365,12 +365,18 @@ Query parameters:
 - `pageSize`: max entries (default 100, capped at 1000)
 - `after`: audit sequence to start after (exclusive, opaque cursor)
 - `reverse`: `true` iterates newest-first
-- `filter`: a filter expression restricted to `audit[...]` conditions, using the
-  same grammar as `ledgerctl audit list --filter` (e.g.
-  `audit[outcome] == failure`, `audit[ledger] == main`,
-  `audit[order_type] in (create_transaction, revert_transaction)`). This is the
-  shared `filterexpr` DSL — **not** the JSON `QueryFilter` DSL used by prepared
-  queries, which cannot represent audit conditions. Filtered reads are served by
+- `filter`: a filter expression restricted to bare audit fields (`outcome`,
```
flemzord, line 368: 🟠 **[medium] Update the encoded HTTP example**\n\nThe request example immediately above still decodes to `audit[outcome] == failure`, which this change intentionally rejects. Please encode the documented bare form instead (`filter=outcome%20%3D%3D%20failure`).

flemzord (approved):
The two previously blocking functional regressions are fixed on this head: the audit CLI now decodes with QUERY_TARGET_AUDIT, and nil prepared-query updates are rejected at both the transport and FSM layers. The remaining comments are documentation-only Medium findings and are non-blocking under the requested High/Critical threshold.

PR #1587, opened 2026-07-14
feat(EN-1550): idempotent mirror ingest on v2LogId

gfyrag:
## What

Makes mirror ingest **idempotent on `v2LogId`** at the FSM apply path, so replaying an already-applied source log is a deterministic no-op. Closes the "tampered/rolled-back `MirrorCursor` → worker re-emits already-applied logs → `processMirrorIngest` force-adds postings twice → doubled balances" concern raised on #1581.

## How — the guard

- New per-ledger field `LedgerBoundaries.last_mirror_v2_log_id` (field 11, `fixed64`): the highest source `v2LogId` already applied to this mirror ledger.
- In `processMirrorIngest` (`internal/domain/processing/processor_mirror.go`), **before any mutation** (no ledger re-touch, no cache write), an entry whose `v2LogId <= last_mirror_v2_log_id` returns a deterministic no-op `(nil, nil)`.
- On a forward ingest, `last_mirror_v2_log_id` is advanced to the entry's `v2LogId` — set once on the wrapping `MirrorLogEntry`, so it covers every inner kind (CreatedTransaction, SavedMetadata, DeletedMetadata, RevertedTransaction, FillGap).
- `ProcessOrders` (`internal/domain/processing/processor.go`) now treats a `(nil, nil)` processor return as a **no-log outcome**: it consumes no sequence id, appends no degenerate `Log{Payload:nil}`, records no `CreatedLog`, and does not absorb into the sink. The per-order slot stays nil and is skipped by `WriteSet.Merge` exactly like an idempotency-replay `ReferenceSequence`.

## Invariant safety (FSM hot path)

- **Deterministic**: the guard is a pure function of applied per-ledger state (`last_mirror_v2_log_id`) and the order (`v2LogId`). No time, no randomness, no map iteration.
- **No new Pebble read / no new preload or coverage key**: the field lives inside `LedgerBoundaries`, already preloaded (`SubAttrBoundary`) and coverage-gated on every mirror-ingest order (`Worker.extractMirrorNeeds`) and already mutated on this path.
- **Expected soft-skip, not fail-loud**: replay is an *expected* stale-ingest case, so `return nil, nil` is correct per invariant #7's expected-vs-impossible distinction (not `assert.Unreachable`, which is reserved for impossible-by-design branches).
- **1-based / monotonic**: v2 log ids are 1-based and strictly increasing per source (`adapter/v2/translator.go: TranslateBatch`), so `<= 0` never wrongly skips the first ingest (default 0 → first real id >= 1 applies).

## Checker

`last_mirror_v2_log_id` needs **no** new tamper-vector checker pass: it is derivable from the audited `MirrorIngest` orders (the max applied `v2LogId` — the same quantity the still-owed `MirrorCursor` equality re-derives) and only *gates future application*. It is not independently business-authoritative, so it rides on the (pending) mirror-cursor coverage. Documented in `docs/technical/architecture/audit-vs-technical-state.md` (Mirror Cursor section).

## Out of scope

The **ahead** direction (a cursor beyond the true source head) is worker-side source-head recovery and is explicitly out of scope here — only the **behind**/replay direction is closed.

## Tests

`internal/domain/processing`:
- Replaying the same `v2LogId` (== high-water mark) is a no-op: no volume writes (balances not doubled), no ledger re-touch, no boundary Put, `(nil, nil)` returned.
- A lower/older `v2LogId` after a higher one is skipped.
- A forward ingest advances `last_mirror_v2_log_id`.
- Batch-level (`ProcessOrders`): a replay emits no log slot, consumes no sequence id, and never reaches the sink.

Verified: `go build ./...`, `go test ./internal/domain/processing/... ./internal/infra/state/... ./internal/proto/... -count=1` all green; `golangci-lint` clean on the touched package.

Relates to #1581.


misc/proto/raft_cmd.proto
```diff
@@ -706,6 +706,10 @@ message LedgerBoundaries {
   fixed64 transient_used_count = 8;
   fixed64 revert_count = 9;
   fixed64 numscript_execution_count = 10;
+  // Highest source (v2) log ID already applied to this mirror ledger. Gates
+  // idempotent replay: a MirrorIngest whose v2LogId is <= this value is a
+  // deterministic no-op on the FSM apply path (see processMirrorIngest).
+  fixed64 last_mirror_v2_log_id = 11;
```
NumaryBot, line 712: 🔴 **[blocker] Verify last_mirror_v2_log_id against audit**

This adds a new Pebble-persisted projection that directly gates future FSM application, but the patch does not add any checker pass that re-derives it from the audited MirrorIngest orders. If this value is tampered lower together with a rolled-back MirrorCursor, already-applied v2 logs can be admitted again; if it is tampered higher, legitimate future logs are skipped. Please add checker coverage that compares this field to the audited max v2LogId before trusting it.

internal/domain/processing/processor_mirror.go
```diff
@@ -39,15 +35,48 @@ func processMirrorIngest(ledger string, order *raftcmdpb.MirrorIngestOrder, ctx
...
+	// the correct outcome per invariant #7's expected-vs-impossible distinction;
+	// failing loud (assert.Unreachable) is reserved for branches reachable only
+	// when an invariant is already broken, which replay is not. ProcessOrders
+	// treats the (nil, nil) return as a no-log outcome: no sequence id, no audit
+	// log, no sink absorb.
+	//
+	// First-ingest case: LastMirrorV2LogId defaults to 0, so the first real
+	// v2LogId (>= 1) is strictly greater and applies normally.
+	v2LogID := entry.GetV2LogId()
+	if v2LogID != 0 && v2LogID <= boundaries.GetLastMirrorV2LogId() {
```
NumaryBot: 🟠 **[major] Backfill the mirror high-water mark on upgrade**

For mirror ledgers that already ingested source logs before this field exists, decoded boundaries have `LastMirrorV2LogId == 0`. If their MirrorCursor is rolled back before any newer ingest advances this field, old source logs have `v2LogId > 0` and pass this guard as fresh, so postings are applied a second time. The high-water mark needs to be initialized from existing audited mirror history or the trusted cursor during migration/recovery.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1587#issuecomment-4967538696

flemzord (changes requested):
Two High-severity blockers remain on head `7ea7abd56694dc7bb883fa27716b5f7573801e78`:

- The new persisted high-water mark gates future FSM application but is not checker-verified against the audit chain, violating the repository projection invariant: https://github.com/formancehq/ledger/pull/1587#discussion_r3577708043
- Existing mirror ledgers decode the new field as zero and receive no migration/backfill, so a rolled-back cursor can still replay and double-apply historical postings until a newer ingest advances it: https://github.com/formancehq/ledger/pull/1587#discussion_r3577708052

I am not duplicating the existing inline comments. Targeted processing tests pass under the Nix toolchain; these are correctness and upgrade-safety blockers.

internal/domain/processing/processor_mirror.go
gfyrag: Decision: we are **not** handling existing-cluster backward-compat / migration at this stage (pre-GA — no compat shims). So no upgrade backfill of the high-water mark in this PR.

Rationale: the change is a **strict improvement, never a regression**. New ingests are fully protected immediately, and an already-mirrored ledger self-heals on its next ingest (which sets `LastMirrorV2LogId`). During the upgrade→first-ingest window a rolled-back cursor behaves exactly as it does *today without this PR* (double-apply on replay) — i.e. no worse than status quo, and only via cursor tampering. Backfilling from audited history for pre-existing clusters is deliberately out of scope now; if we later decide to support in-place migration of existing mirror clusters, it can be a dedicated follow-up.

gfyrag: Added a checker bound for the field in 7f67d017e (same PR): new `compareMirrorV2LogID` pass verifies `stored last_mirror_v2_log_id <= max(audited MirrorIngest.v2_log_id)` per mirror ledger, emitting `CHECK_STORE_ERROR_TYPE_MIRROR_V2LOGID_AHEAD` when `stored > max`. The audited max is `max(archived-chapter floor, live)` so archived-ingest ledgers do not false-positive.

This detects the genuinely dangerous **ahead** direction (field claims to have consumed a v2 log absent from the audit → future silent skips). The **behind** direction (stored < max, incl. legacy `0`) is intentionally not flagged — that is the accepted no-backfill/self-healing state per the compat decision above, and the FSM idempotency guard already makes a behind replay a no-op for anything the field does cover. So: replay double-apply is closed functionally for new ingests, the ahead-corruption is now checker-detected, and existing-cluster backfill remains deliberately out of scope pre-GA.

internal/domain/processing/processor_mirror.go
```diff
@@ -93,6 +122,14 @@ func processMirrorIngest(ledger string, order *raftcmdpb.MirrorIngestOrder, ctx
 	// Assign per-ledger log ID and advance boundaries
 	nextLogID := boundaries.GetNextLogId()
 	boundaries.NextLogId = nextLogID + 1
+	// Advance the idempotent-replay high-water mark. Set once here regardless of
+	// the inner ingest kind, because v2LogId lives on the wrapping MirrorLogEntry
+	// and applies to every kind (CreatedTransaction, SavedMetadata,
+	// DeletedMetadata, RevertedTransaction, FillGap). Guarded by v2LogID != 0 so
+	// an entry that somehow carries no source id never rewinds the mark.
+	if v2LogID != 0 {
```
NumaryBot, line 150: 🔴 **[blocker] High-water mark can skip unapplied logs when IDs are non-contiguous**

If the mirror cursor is ever ahead of the true applied prefix (e.g. cursor advanced to 9 while only logs 1–5 were applied), the worker fetches and applies log 10, recording `LastMirrorV2LogId = 10`. When the cursor is later corrected back to 5, logs 6–9 satisfy the `<= LastMirrorV2LogId` replay guard and are silently skipped even though they were never applied. This causes permanent, undetectable under-ingestion in any cursor-ahead or out-of-order scenario.

**Suggestion:** Only advance `LastMirrorV2LogId` when the incoming v2 log ID is exactly `LastMirrorV2LogId + 1` (strictly contiguous). If a gap is detected, reject the batch or fill the missing IDs before updating the high-water mark, so that no legitimate log can fall beneath the stored maximum without having been applied.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1587#issuecomment-4967538696

gfyrag:
I reviewed the current head (`7f67d017e625784f9ed98299a5d1fa84513bb445`) in detail. I would not merge this as-is.

GitHub will not let me submit a formal “request changes” review on my own PR identity, so I am posting the blocking review as a PR comment.

Findings:

1. **Blocker: `LastMirrorV2LogId` is advanced as a max, not as a contiguous prefix.** I confirmed NumaryBot's finding. `processMirrorIngest` accepts any `v2LogId > LastMirrorV2LogId` and then writes that value back to boundaries. If `MirrorCursor` is ahead of the actually applied prefix, the worker fetches from that cursor and `TranslateBatch` only gap-fills from `cursor+1`, not from the FSM boundary. Applying log `10` after only `1..5` were actually applied records `LastMirrorV2LogId=10`; after the cursor is repaired, `6..9` are now `<= LastMirrorV2LogId` and are silently skipped. The marker must represent a contiguous applied prefix: reject `v2LogId != last+1`, or emit/apply explicit `FillGap` orders for every missing id before advancing it.

2. **Blocker: the checker accepts `stored < auditedMax`, which leaves the rollback/double-apply case open.** `compareMirrorV2LogID` only reports `stored > auditedMax`, and `TestCompareMirrorV2LogID_AtOrBelowNotFlagged` explicitly blesses `stored=3, auditedMax=7`. But the FSM skips only `v2LogId <= stored`; if both `MirrorCursor` and `LastMirrorV2LogId` are rolled back, already-audited logs above the stored value are treated as fresh and force-applied again. Since we are not doing compatibility/backfill handling here, the checker should verify the projection as the audit-derived value, not just a one-sided upper bound. Otherwise invariant #8 is not actually satisfied for this persisted projection.

3. **Major: the archive baseline floor used by `compareMirrorV2LogID` is currently ineffective.** The new docs/comment say `foldBaselineBoundaries` seeds `maxMirrorV2LogID` from the archived baseline, but `CreateBaselineSnapshot` only writes Volume, Metadata, and Transaction attributes into the baseline DB. It does not write Boundary rows. With archived mirror ingests and a baseline present, `foldBaselineBoundaries` will not see the archived `last_mirror_v2_log_id`, so the live audit max is undercounted and a legitimate stored high-water mark can be reported as `MIRROR_V2LOGID_AHEAD`. Either include boundaries in the baseline snapshot or change this checker pass so archived history does not produce false positives.

I also checked the generated protobuf files. I did not find an issue there: `raft_cmd_dethash.pb.go` delegates `LedgerBoundaries` deterministic marshaling to `MarshalVT()`, and the new field is present in the vtproto output.

Verification run locally:

```sh
GOROOT= go test ./internal/domain/processing ./internal/application/check -run 'Mirror|V2Log|ProcessOrders' -count=1
```

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1587#issuecomment-4967538696

internal/domain/processing/processor_mirror.go
```diff
@@ -93,6 +132,14 @@ func processMirrorIngest(ledger string, order *raftcmdpb.MirrorIngestOrder, ctx
...
```
NumaryBot, line 150: 🔴 **[blocker] High-water mark advances as a maximum, not a contiguous prefix — unapplied logs silently skipped**

The `LastMirrorV2LogId` marker is updated whenever a v2 log ID exceeds the stored value (max semantics), not when the incoming ID is exactly `stored + 1` (prefix semantics). If the mirror cursor is ever ahead of the actually-applied prefix — e.g. cursor at 9 while only logs 1–5 were applied — log 10 is processed and `LastMirrorV2LogId` is set to 10. When the cursor is later rolled back to 5, logs 6–9 satisfy the `<= LastMirrorV2LogId` replay guard and are permanently, silently skipped even though they were never applied. This produces undetectable under-ingestion.

**Suggestion:** Only advance `LastMirrorV2LogId` when the incoming v2 log ID is exactly `LastMirrorV2LogId + 1`. If a gap is detected, either reject the batch with an explicit error or fill the gap (emit FillGap orders for every missing ID) before advancing the marker, so that no legitimate log can fall beneath the stored maximum without having been applied.

internal/application/check/checker.go
```diff
@@ -897,6 +899,70 @@ func (c *Checker) compareIndexes(
...
+func (c *Checker) compareMirrorV2LogID(reader dal.PebbleReader, chainBound *chainBoundState, callback func(*servicepb.CheckStoreEvent)) {
+	iter, err := c.attrs.Boundary.NewStreamingIter(reader, nil)
+	if err != nil {
+		callback(errorEvent(servicepb.CheckStoreErrorType_CHECK_STORE_ERROR_TYPE_MIRROR_V2LOGID_MISMATCH,
+			fmt.Sprintf("failed to create live boundaries iterator: %v", err), 0, "", "", ""))
+
+		return
+	}
+
+	for iter.Next() {
```
NumaryBot, line 948: 🟠 **[major] Checker misses missing/deleted mirror boundary rows — projection deletion goes undetected**

The `compareMirrorV2LogID` checker pass only iterates over existing boundary entries. If a mirror ledger's `Boundary` row is deleted or absent, the pass never compares it: `chainBound.maxMirrorV2LogID` may carry an audited max for the ledger, but no error is emitted. This means the `LastMirrorV2LogId` projection can disappear from storage without any checker detection, leaving the idempotency guard silently non-functional.

**Suggestion:** Track the set of ledger IDs seen in existing boundary rows, then cross-reference it against the set of audited mirror ledgers. For any audited mirror ledger with no live boundary row (excluding legitimately cleaned-up ledgers), emit a `CHECK_STORE_ERROR_TYPE_MIRROR_V2LOGID_AHEAD` (or a dedicated error type) to surface the missing projection.

internal/domain/processing/processor_mirror.go
gfyrag: Fixed in d360d6fef. The guard is now a **contiguous applied prefix**, not a `<= max` skip: `v2LogId <= last` → idempotent no-op; `v2LogId == last+1` → apply+advance; `v2LogId > last+1` → **loud reject** (`ErrMirrorV2LogIDGap`, a deterministic Describable → BusinessError + failure audit, no state mutated). So the "cursor ahead of applied prefix, then corrected → silent skip of the gap" scenario can no longer happen — a gap wedges the ledger loudly (worker backs off, cursor advances neither on disk nor in memory) instead of silently under-ingesting. Regression `TestMirrorIngest_GapRejected` asserts zero Boundaries/Ledgers/Volumes writes + exact error fields.

gfyrag:
Update — all three review findings are now addressed (head \`d360d6fef\`), following the product decision that we do **not** support backward-compat with existing mirror clusters (no backfill; a pre-field cluster re-mirrors):

1. **Contiguity (was the blocker):** `LastMirrorV2LogId` is now a true contiguous applied prefix. `processMirrorIngest` is 3-way — `<= last` idempotent no-op, `== last+1` apply+advance, `> last+1` **loud reject** (`ErrMirrorV2LogIDGap`, deterministic Describable → BusinessError + failure audit, no state mutated; cursor discarded so the ledger wedges loudly with worker backoff). No more silent under-ingestion on a corrected-back cursor.
2. **Checker equality:** `compareMirrorV2LogID` now verifies `stored == max(audited v2LogId)` and flags **both** directions (`CHECK_STORE_ERROR_TYPE_MIRROR_V2LOGID_MISMATCH`) — full invariant-#8 verification, no `<=` leniency.
3. **Baseline floor (confirmed bug):** `writeBaselineAttributes` now includes `Boundary`, so the archived `last_mirror_v2_log_id` is folded into the audited max — no false positive for archived-ingest ledgers (real end-to-end test added).

Independently re-reviewed at `d360d6fef`: gap path mutates nothing and is deterministic across replicas; both corruption directions covered (behind→no-op, ahead→loud reject); proto enums appended sequentially (wire-compat, dethash unaffected); build + processing/check/attributes/backup/proto tests green. Ready for another look, @flemzord.

internal/domain/processing/processor_mirror.go
```diff
@@ -39,15 +35,58 @@ func processMirrorIngest(ledger string, order *raftcmdpb.MirrorIngestOrder, ctx
...
+	//     every node, so determinism holds.
+	//
+	// First-ingest case: LastMirrorV2LogId defaults to 0, so expected = 1 and the
+	// first real v2LogId (>= 1) applies. The v2LogID != 0 guard keeps a malformed
+	// entry with no source id from being treated as the "expected" first slot.
+	v2LogID := entry.GetV2LogId()
+	last := boundaries.GetLastMirrorV2LogId()
+
+	switch {
+	case v2LogID != 0 && v2LogID <= last:
```
NumaryBot: 🟠 **[major] Zero v2LogId bypasses idempotency guard and high-water mark update**

If a mirror ingest order arrives with `v2LogId == 0`, neither the stale-replay guard (`v2LogId <= last`) nor the gap-reject guard (`v2LogId > last+1`) fires for the initial state where `LastMirrorV2LogId == 0`, and the trailing `if v2LogID != 0` block skips updating the high-water mark. The same zero-id order can therefore be replayed indefinitely, double-applying postings or metadata on every replay with no state mutation to prevent it. Since v2 source log IDs are 1-based, a zero value is always malformed and should be rejected loudly before any FSM mutation occurs.

**Suggestion:** Add an explicit guard at the top of the mirror ingest handler: if `v2LogId == 0`, return a deterministic `Describable`/`BusinessError` (e.g. `ErrMirrorV2LogIDInvalid`) immediately, before any state is read or written. This keeps the zero case symmetrical with the gap-reject path added in d360d6fef.

internal/application/check/checker.go
NumaryBot, line 948: 🟡 **[minor] Checker does not detect deleted or absent mirror boundary rows**

The `compareMirrorV2LogID` pass only iterates over existing boundary entries. If a mirror ledger's `Boundary` row is deleted or absent, the pass never compares the audited max against a stored value, so the disappearance of the `LastMirrorV2LogId` projection goes completely undetected. This was raised in the prior PR discussion (Thread 5) by an automated reviewer; no human resolution or dismissal was recorded for it.

**Suggestion:** Collect the set of ledger IDs seen in existing boundary rows, then cross-reference against the set of ledger IDs with audited mirror-ingest history. For any audited mirror ledger that has no live boundary row, emit an appropriate check error (e.g. a dedicated `CHECK_STORE_ERROR_TYPE_MIRROR_BOUNDARY_MISSING`) to surface the missing projection.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1587#issuecomment-4967538696

internal/domain/processing/processor_mirror.go
flemzord: Confirmed on current head `dd8bff930ebcb026639ac00f856541e8fe09fe12`: this remains a blocker. `v2LogId == 0` bypasses both guards (`<= last` and gap) and then skips the high-water mark update, so the same malformed mirror ingest can be replayed and applied repeatedly with no persisted marker to stop the next replay. Since source v2 log ids are 1-based, please reject zero explicitly before any FSM mutation, with a deterministic Describable/BusinessError and a regression test for the zero-id replay path.

flemzord (changes requested):
Current head `dd8bff930ebcb026639ac00f856541e8fe09fe12` still has one High/Critical-blocking issue: malformed mirror ingest orders with `v2LogId == 0` bypass the replay/gap guards and do not advance `LastMirrorV2LogId`, so they can be replayed and applied repeatedly. I replied on the existing inline thread instead of opening a duplicate: https://github.com/formancehq/ledger/pull/1587#discussion_r3578806223

The previous contiguity, checker equality, and baseline-boundary blockers look addressed. The remaining missing-boundary checker point is minor and not an approval blocker under the current High/Critical threshold. Targeted tests pass under Nix: `GOROOT= go test ./internal/domain/processing ./internal/application/check ./internal/infra/attributes -run 'Mirror|V2Log|Baseline|ProcessOrders' -count=1`.

internal/domain/processing/processor_mirror.go
gfyrag: Fixed in f09640bb9. `processMirrorIngest` now rejects `v2LogId == 0` **before any mutation** (before the ledger re-touch and the no-op/gap/apply switch) with a dedicated deterministic `*domain.ErrMirrorV2LogIDInvalid` (→ `ERROR_REASON_MIRROR_V2_LOG_ID_INVALID`, KindInternal) — mutates nothing, so it can never apply or advance the mark. The now-redundant `!= 0` guards on the `<= last` / `> last+1` branches are dropped. Regression `TestMirrorIngest_ZeroV2LogIdRejected` asserts a zero-id ingest is rejected with no Boundaries/Ledgers/Volumes Put, and that an identical second zero-id order is rejected the same way (the replay path you flagged) — proving the reject is stateless/deterministic. Thanks for catching it.

internal/domain/processing/processor_mirror.go
```diff
@@ -39,15 +35,68 @@ func processMirrorIngest(ledger string, order *raftcmdpb.MirrorIngestOrder, ctx
...
+	//   - v2LogID <= last  → already applied. EXPECTED replay (a tampered/rolled-
+	//     back MirrorCursor makes the worker re-emit applied logs — flemzord,
+	//     #1581). Idempotent no-op: return (nil, nil), which ProcessOrders treats
+	//     as a no-log outcome (no sequence id, no audit log, no sink absorb). Soft
+	//     skip is correct per invariant #7's expected-vs-impossible distinction.
+	//
+	//   - v2LogID == expected → the next contiguous log. Apply and advance below.
+	//
+	//   - v2LogID > expected → a GAP: the cursor/high-water mark is ahead of the
+	//     applied prefix. Impossible in normal operation (contiguous ingest), so
```
NumaryBot, line 75: 🟠 **[major] v2LogId == 0 bypasses both idempotency guards and high-water mark update**

When a mirror ingest order arrives with `v2LogId == 0` and `LastMirrorV2LogId == 0` (initial state), neither the stale-replay guard (`v2LogId <= last`) nor the gap-reject guard (`v2LogId > last+1`) fires, and the trailing `if v2LogID != 0` block skips updating the high-water mark entirely. The same zero-id order can therefore be replayed and applied repeatedly with no persisted marker to prevent future replays. Since v2 source log IDs are 1-based, a zero value is always malformed.

**Suggestion:** Add an explicit guard at the very top of the mirror ingest handler: if `v2LogId == 0`, return a deterministic `Describable`/`BusinessError` (e.g. `ErrMirrorV2LogIDInvalid`) immediately, before any state is read or written. Add a regression test asserting zero-id replay is rejected and no state is mutated.

internal/domain/processing/processor.go
```diff
@@ -253,6 +253,24 @@ func (p *RequestProcessor) ProcessOrders(orders []*raftcmdpb.Order, scopeFactory
...
+		// entirely — do NOT consume a sequence id, absorb into the sink, or
+		// append a degenerate Log{Payload:nil}. The nil slot mirrors the
+		// idempotency-replay ReferenceSequence path and is skipped identically
+		// by WriteSet.Merge (GetCreatedLog()==nil) and checkCloseChapter.
+		//
+		// Any overlay staged by such an order is dropped without Commit(): a
+		// no-log outcome makes no persistent mutation, so there is nothing to
+		// flush. (Mirror ingest declares no skippable_reasons, so no overlay is
+		// created here in practice.)
+		if payload == nil {
```
NumaryBot, line 270: 🟠 **[major] Nil per-order slots not preserved in idempotency replays for mixed proposals**

When a proposal with an idempotency key contains a replayed mirror ingest (which produces no log) alongside at least one fresh order, the `continue` at this line leaves a nil per-order slot, but the success idempotency record stores only `FirstLogSequence` + `LogCount` from `CreatedLogs`. On a retry, the response is reconstructed as a dense slice of length `LogCount`, which no longer matches the original order positions. All-no-log successes are not recorded at all, making the idempotency guarantee incorrect for these cases.

**Suggestion:** Persist and replay nil slots explicitly, or restructure the stored outcome so it can faithfully reconstruct the original per-order result shape including no-log positions.

internal/application/check/checker.go
NumaryBot, line 948: 🟡 **[minor] Checker does not detect deleted or absent mirror boundary rows**

The `compareMirrorV2LogID` checker pass only iterates over existing boundary entries. If a mirror ledger's `Boundary` row is deleted or absent, the pass never compares the audited max against a stored value, so the disappearance of the `LastMirrorV2LogId` projection is silently undetected. This leaves the idempotency guard non-functional without any observability.

**Suggestion:** Track the set of ledger IDs seen in existing boundary rows, then cross-reference against the set of ledger IDs with audited mirror-ingest history. For any audited mirror ledger that has no live boundary row, emit an appropriate check error (e.g. a dedicated `CHECK_STORE_ERROR_TYPE_MIRROR_BOUNDARY_MISSING`) to surface the missing projection.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1587#issuecomment-4967538696

internal/domain/processing/processor_mirror.go
gfyrag: Already fixed in f09640bb9 (this comment describes the pre-fix code). `processMirrorIngest` now rejects `v2LogId == 0` **upfront**, before the no-op/gap/apply switch and before any mutation, returning `*domain.ErrMirrorV2LogIDInvalid` — the trailing `if v2LogID != 0` block referenced here was removed. Covered by `TestMirrorIngest_ZeroV2LogIdRejected` (incl. the zero-id replay path).

internal/domain/processing/processor.go
gfyrag: Not reachable, so no action. The no-log/nil-slot path only occurs for a **mirror-ingest replay**, and mirror proposals are worker-generated with `SystemCallerSnapshot(ComponentMirror)`, contain only MirrorIngest orders + the MirrorSync TU, and carry **no idempotency key** (`worker.go` NewCommand; no IdempotencyKey set). Client idempotency-keyed proposals go through admission and never contain a MirrorIngest. So "a replayed mirror ingest (no-log) alongside a fresh order under an idempotency key" is not a constructible proposal shape — the idempotency success-record path is never taken for a proposal containing a no-log mirror order.

internal/application/check/checker.go
gfyrag: Good catch — addressing it. `compareMirrorV2LogID` will drive the comparison from the union of stored-boundary ledgers and audit-derived ledgers (`auditedMax > 0`), so a mirror ledger with audited ingests but an absent/deleted Boundary row is compared as stored=0 vs auditedMax>0 → flagged (not skipped). Fix + test incoming; will link the SHA here.

gfyrag: Fixed in 3f5af678f. `compareMirrorV2LogID` now drives from the **union** of stored-boundary ledgers and audit-derived ledgers (`auditedMax > 0`): a mirror ledger with audited ingests but no stored Boundary row is compared as stored=0 vs auditedMax>0 → `MIRROR_V2LOGID_MISMATCH` with a distinct "no stored boundary row (last_mirror_v2_log_id lost)" detail (sorted-name iteration keeps event order deterministic). Non-mirror ledger with no row stays 0==0 (clean). Tests: `AbsentRowFlagged`, `AbsentRowNonMirrorNotFlagged`, plus the existing present-row cases unchanged.

internal/application/check/checker.go
```diff
@@ -897,6 +899,114 @@ func (c *Checker) compareIndexes(
...
+	// Compare over the union of stored rows and audited-mirror ledgers, in
+	// sorted order so events are emitted deterministically. A ledger present in
+	// only one set defaults to 0 on the other side (absent row → stored 0;
+	// non-mirror ledger → audited max 0).
+	ledgers := make(map[string]struct{}, len(stored)+len(chainBound.maxMirrorV2LogID))
+	for name := range stored {
+		ledgers[name] = struct{}{}
+	}
+
+	for name := range chainBound.maxMirrorV2LogID {
```
NumaryBot, line 986: 🟠 **[major] Checker false-positives on legitimately deleted mirror ledgers**

The `compareMirrorV2LogID` pass (per the fix incoming for Thread 10) drives comparison from the union of stored-boundary ledgers and all ledger IDs with audited mirror-ingest history. When a mirror ledger is later deleted, `WriteSet.Absorb` intentionally removes its `Boundary` row, but the audited max for that ledger is still non-zero. The pass therefore treats the absent row as `stored=0` vs `auditedMax>0` and emits a `MIRROR_V2LOGID_MISMATCH` error for a valid, legitimately deleted ledger — a functional false positive that will cause `Check()` to report corruption on healthy stores.

**Suggestion:** Before flagging an absent boundary row, exclude ledger IDs that have a `DeleteLedger` log in the audited history (or otherwise mark them as cleaned up). Only emit the missing-boundary error for mirror ledgers that are audited as active (i.e. no delete log observed during replay).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1587#issuecomment-4967538696

internal/application/check/checker.go
gfyrag: Fixed in 30330d3fd. `compareMirrorV2LogID` now takes the checker's existing `deletedInReplay` set (the same one `compareIndexes` uses, populated from `DeleteLedger` during replay) and, on the **absent-row branch only**, skips a ledger that is in `deletedInReplay` — `WriteSet.Absorb` legitimately removes the Boundary row on delete, so a missing row for an audited-deleted ledger is expected, not corruption. Scoping preserved: an active (non-deleted) ledger with ingests and no row is still flagged; present-row equality (ahead/behind/corrupt-to-zero) is unchanged for all ledgers, including a pending-cleanup ledger whose row is still present. Archived-delete edge checked too (fully-archived delete → auditedMax 0 → out of scope). Test `DeletedMirrorLedgerNotFlagged` added; `AbsentRowFlagged` (active) kept. Thanks — good catch.

internal/application/check/checker.go
```diff
@@ -897,6 +899,130 @@ func (c *Checker) compareIndexes(
...
```
NumaryBot, line 986: 🟠 **[major] Checker false-positives on legitimately deleted mirror ledgers**

The `compareMirrorV2LogID` pass drives comparison from the union of stored-boundary ledgers and all ledger IDs with audited mirror-ingest history. When a mirror ledger is deleted, `WriteSet.Absorb` intentionally removes its `Boundary` row, but the audited max for that ledger remains non-zero. The pass therefore compares stored=0 vs auditedMax>0 and emits `MIRROR_V2LOGID_MISMATCH` for a valid, legitimately deleted ledger, causing `Check()` to report corruption on a healthy store.

**Suggestion:** Before emitting a missing-boundary error for an absent row, cross-reference against the audited history to determine whether a `DeleteLedger` log exists for that ledger ID. Only flag the absent row if the ledger has no `DeleteLedger` entry in the audit trail (i.e. it is audited as active). Ledger IDs with a confirmed delete log should be excluded from the absent-boundary check entirely.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1587#issuecomment-4967538696

flemzord (approved):
Approved on current head `30330d3fdb8d67437c681ad1734f09af42106926`.

Re-reviewed the follow-up fixes: the zero `v2LogId` path now rejects before mutation, the mirror high-water mark is checked as audit equality, absent boundary rows are detected for active mirror ledgers, and legitimately deleted mirror ledgers are excluded from the absent-row branch via `deletedInReplay`. I resolved the remaining duplicate thread after confirming `TestCompareMirrorV2LogID_DeletedMirrorLedgerNotFlagged` covers that case.

Targeted verification passed under Nix: `GOROOT= go test ./internal/domain/processing ./internal/application/check ./internal/infra/attributes -run 'Mirror|V2Log|Baseline|ProcessOrders|DeletedMirror|AbsentRow|ZeroV2LogId' -count=1`.

PR #1590, opened 2026-07-14
refactor(mirror): rename source types to ledgerV2Http / ledgerV2Database

sylr:
## What

Renames the two mirror source-type discriminators from `http` / `postgres` to **`ledgerV2Http`** / **`ledgerV2Database`**. The `mirrorSource` config stays a **single object** with one sub-property selecting the source type.

> Supersedes the earlier multi-source (`mirrorSources` list) exploration on this branch — we decided **not** to implement several sources per mirror ledger for now. The history was rewritten to a single focused rename commit.

## Why

`http` / `postgres` describe the transport, not what they connect to. `ledgerV2Http` / `ledgerV2Database` convey that both current sources read from a **Ledger v2** system — via its HTTP API, or directly from its PostgreSQL database.

## Scope

- **Proto** (`common.proto`): oneof field names `http`/`postgres` → `ledger_v2_http`/`ledger_v2_database`. **Tag numbers (2/3) are preserved**, so the change is wire-compatible with existing persisted data — only the field *names* (JSON keys + Go accessors) change.
- **HTTP API**: `mirrorSource.type` values → `ledgerV2Http` / `ledgerV2Database` (default `ledgerV2Http`).
- **ledgerctl**: `--mirror-source-type` values + completions + help.
- **Operator CRD**: `spec.mirrorSource.ledgerV2Http` / `.ledgerV2Database` sub-fields, plus a new CEL rule enforcing exactly one source type. Regenerated deepcopy + CRD YAML (config + helm, in sync).
- **Docs + openapi** updated.

The underlying message/struct type names (`HttpMirrorSourceConfig`, `PostgresMirrorSource`, …) are intentionally **unchanged** — they are internal identifiers that never appear in JSON/YAML.

## Compatibility

Breaking rename of the API/CRD discriminator on pre-GA `v1alpha1`; existing mirror `Ledger` CRs / API bodies using `http`/`postgres` must switch to `ledgerV2Http`/`ledgerV2Database`. Persisted proto data is unaffected (tag numbers unchanged).

## Verification

- Generated proto diff is minimal and canonical (commonpb only, no toolchain churn).
- Build green (main + operator); touched unit/integration suites pass: admission, HTTP handler, FSM create/promote, ledgerctl flags, operator controller; e2e mirror suite compiles.
- Report-only `golangci-lint` on the touched packages: **0 issues**.

Note: the sandbox's proto/golangci toolchain drifted mid-session (a `--fix` tag-reorder + a golangci-lint linter-set change); those artifacts were kept out of this diff, which contains only the rename.


openapi.yml
```diff
@@ -4198,10 +4198,10 @@ components:
         type:
           type: string
           enum:
-            - http
-            - postgres
-          default: http
-          description: Source type (defaults to "http" if omitted)
+            - ledgerV2Http
```
flemzord, line 4367: [P2] Split the request and response mirror-source schemas

This schema is also referenced by `LedgerInfo`, but that response is not encoded through `mirrorSourceBody`. `LedgerInfo.MarshalJSON` calls `protoFieldJSON`, so the renamed proto oneof serializes as `{\"ledgerV2Http\":{...}}` (or `ledgerV2Database`), not the flat `{\"type\":\"ledgerV2Http\",...}` declared here. Generated clients therefore still cannot model mirror-ledger responses from this contract. Please give `LedgerInfo` a oneof-shaped response schema, or normalize response serialization to the flat discriminator.

flemzord (changes requested):
Reviewed at `ca1468e6`. The rename is otherwise consistently propagated through proto, HTTP input, CLI, operator CRDs, docs, and generated artifacts, and all targeted tests plus CI pass. One OpenAPI response-contract gap remains; details are inline.

openapi.yml
sylr: Addressed in the latest push (rebased onto release/v3.0).

Added a dedicated oneof-shaped response schema `MirrorSourceConfigResponse` (with nested `HttpMirrorSourceConfig` / `PostgresMirrorSourceConfig` / `OAuth2ClientCredentials` / `PostgresAwsIamAuth`) and pointed `LedgerInfo` at it, so the response contract now matches the protojson output (`{"ledgerV2Http":{…}}` / `{"ledgerV2Database":{…}}`) instead of the flat request shape. The flat `MirrorSourceConfig` is kept for the request body (`CreateLedgerRequest`), which is decoded through `mirrorSourceBody`. Also fixed the field key on `LedgerInfo` (`mirror_source` → `mirrorSource`) to match the actual camelCase output.

openapi.yml
```diff
@@ -4296,10 +4364,10 @@ components:
...
```
NumaryBot, line 4367: 🟠 **[major] Mirror-source schema shape mismatch between mirror request and LedgerInfo response**

The shared mirror-source schema uses a flat discriminator shape (`{"type":"ledgerV2Http",...}`), but `LedgerInfo.MarshalJSON` serializes the proto oneof as a nested-key shape (`{"ledgerV2Http":{...}}`). Generated clients that consume `LedgerInfo` responses cannot correctly model the data from this contract, and the request and response representations are conflated into a single schema.

**Suggestion:** Split the mirror-source schema into separate request and response variants. Give the `LedgerInfo` response a oneof-shaped schema that matches the actual proto serialization (`{"ledgerV2Http":{...}}`), and keep the flat discriminator only for the request body, or normalize response serialization so it matches the OpenAPI contract.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1590#issuecomment-4970907351

flemzord (approved):
Reviewed current head `9116120a`. The previous OpenAPI response-shape blocker is fixed: LedgerInfo now points to the oneof-shaped MirrorSourceConfigResponse while the create request keeps the flat discriminator schema. The remaining NumaryBot status repeats that resolved finding, and the pre-stable wire/API rename is not a runtime blocker. Main CI is green.

gfyrag (changes requested):
The actual naming is ok.

PR #1591, opened 2026-07-14
fix(replay): carry typed metadata values through backup restore

Azorlogh:
Follow-up to #1554, found by the model test's restore cycle (`account read outside model`: model held `b:false`, server returned `s:false` after a restore).

## Bug

`replay.Writer.SetMetadata` took the value as a plain **string**, and both account-metadata replay sites (transaction-embedded and standalone SaveMetadata) funneled the log's typed `MetadataValue` through `MetadataValueToString`. The FSM stores the order's typed value verbatim, so every bool/int/uint/datetime account-metadata value inside an exported delta came back from a restore as its string rendering:

- live row: `1800` → `boolValue:false`
- rebuilt row: `0a0566616c7365` → `stringValue:"false"`

Reads served by the restored store then return the wrong value type. Transaction and ledger metadata were unaffected (typed maps/values throughout).

## Fix

- `SetMetadata` now takes `*commonpb.MetadataValue`; the rebuild stores the typed value directly and the checker's scratch store keeps the marshaled proto after the flag byte.
- **Checker strengthening**: `compareMetadata` compares typed values (`EqualVT`) instead of string renderings. It previously stringified both sides before comparing, normalizing away exactly the information the rebuild destroyed — the whole class was invisible to `Check()`. Events now render values with their type: `"false" (bool)` vs `"false" (string)`.

## Behavioral note

Stores restored with the old rebuild hold string-rendered account-metadata rows and will now correctly fail the check with `METADATA_MISMATCH` — worth knowing for existing restored environments.

## Tests

- New e2e `tests/e2e/cluster/restore_metadata_type_test.go`: writes `flag=true` (bool) and `count=42` (int64) through both command shapes, asserts the live node serves them typed, backup → restore, asserts the types survive. Fails without the fix commit, passes with it.
- Unit-level bool round-trip on the checker's replay store.

### DIFF
--- a/internal/application/check/checker.go
+++ b/internal/application/check/checker.go
@@ -1293,9 +1293,39 @@ func (c *Checker) compareVolumes(ctx context.Context, reader dal.PebbleReader, b
 	return errorCount
 }
 
+// metaValueDisplay renders a metadata value with its type for check events, so
+// a type-only divergence (bool false vs string "false") is visible in the
+// message.
+func metaValueDisplay(v *commonpb.MetadataValue) string {
+	var kind string
+
+	switch v.GetType().(type) {
+	case *commonpb.MetadataValue_StringValue:
+		kind = "string"
+	case *commonpb.MetadataValue_IntValue:
+		kind = "int"
+	case *commonpb.MetadataValue_UintValue:
+		kind = "uint"
+	case *commonpb.MetadataValue_BoolValue:
+		kind = "bool"
+	case *commonpb.MetadataValue_DatetimeValue:
+		kind = "datetime"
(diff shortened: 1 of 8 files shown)


gfyrag (approved):
Reviewed current head `bdc867797bf7619c640ee2a612fedfe32f3466fd` in detail.

The restore/replay/checker path now preserves typed `MetadataValue` protos instead of stringifying them: replay passes the original typed values for both transaction-embedded account metadata and standalone `SavedMetadata`, the restore writer stores the proto directly, and the checker compares typed metadata with `EqualVT`. The new e2e covers both affected account-metadata log shapes across backup restore.

No blocking findings.

PR #1592, opened 2026-07-14
fix(backup): rebuild the reversion bitsets on restore

Azorlogh:
Found by the model test's restore cycle, immediately after a backup→restore: a revert the model knew was already applied got re-attempted by the server and rejected with `insufficient funds` instead of `TRANSACTION_ALREADY_REVERTED`. Same mechanism as an earlier unexplained finding where the incidental rejection was `account does not match any account type pattern`.

## Bug

The FSM's already-reverted gate reads the **per-ledger reversion bitset** (`ZonePerLedger`/`SubPLReversions`) — not the tx rows' `RevertedByTransaction` markers. That bitset is persisted only by the FSM hot path and loaded verbatim on boot (`query.ReadReversions`). The restore rebuild replayed `RevertedTransaction` logs into the tx-row markers but never into the bitset, so a restored node forgot **every revert inside the exported delta**:

- On a solvent account, a second revert of the same transaction passes the gate and **succeeds — a double refund**.
- On an insolvent one, it surfaces as a wrong rejection (insufficient funds / account-type mismatch), which is how the model caught it.

Evidence from the failing run: the restored store's log stream held 30 reverts while the `0301` key range held **zero rows**; the re-attempted revert targeted a transaction whose own row said `revertedByTransaction: 41`.

## Fix

The replay writer seeds the bitsets from the checkpoint rows, folds every replayed revert into them (`SetRevertedBy`), resets a ledger's fold on `DeleteLedger`, and flushes touched ledgers' nonzero words in the FSM's own layout (`state.SaveReversionWord`, exported for this).

## Checker strengthening

The persisted bitset was an unverified projection (invariant 8): `checkReversionInvariants` only validates the log stream's internal consistency. The new `compareReversions` pass compares stored bitsets against the audit-derived reverted set (baseline tx-row markers + replayed reverts, complete across archived chapters), exact equality both ways — a lost bit re-admits a double revert, an unaudited bit silently blocks a legitimate one. Cleanup-pending ledgers are skipped (rows linger until the covering purge).

## Tests

- New e2e `tests/e2e/cluster/restore_reversion_test.go`: reverts a funded transaction pre-backup, keeps the account solvent, restores, and asserts a second revert is still rejected with "already reverted" and the account volumes are unchanged. Without the fix commit, the double revert **succeeds**.
- Unit tests for `compareReversions` (match / missing bit / unaudited bit / cleanup-pending skip).

## Behavioral note

Stores restored with the old rebuild are missing delta reverts from their bitsets and will now correctly fail `Check()` with `REVERTED_MISMATCH` — and genuinely accept double reverts until re-restored with the fix, so worth flagging to operators of existing restored environments.


internal/infra/backup/rebuild.go
```diff
@@ -800,6 +830,12 @@ func (w *attributeReplayWriter) deleteLedger(name string, deletedAt *commonpb.Ti
 
 	delete(w.boundaries, name)
 
+	// The stored reversion rows linger until the covering purge (parity with
```
NumaryBot: 🟠 **[major] DeleteLedger replay leaves stale reversion rows in Pebble**

When replaying a DeleteLedger event for a ledger that had checkpoint-time reversion words, the code only removes the bitset from the in-memory map of the replay writer. The existing ZonePerLedger/SubPLReversions rows are not deleted from Pebble. The live FSM path deletes those rows immediately on ledger deletion, so a restored node at the same applied index will load stale Registry.Reversions for the deleted ledger. The new checker skips non-live ledgers, so the divergence is never repaired until a later covering purge runs, leaving the restored projection inconsistent with the live state.

**Suggestion:** When handling DeleteLedger during replay, explicitly delete the corresponding ZonePerLedger/SubPLReversions rows from Pebble (mirroring what the live FSM path does), in addition to clearing the in-memory bitset map entry.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1592#issuecomment-4974958342

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1592#issuecomment-4974958342

gfyrag (changes requested):
I found one blocking issue in the new reversion projection checker.

1. `compareReversions` must not skip based on the persisted pending-cleanup marker, and it should also account for stored rows for non-live ledgers.

The checker derives `knownLedgers` from the audit stream, but then `compareReversions` skips a ledger whenever `pendingCleanupLedgers[name]` exists. That marker is itself a Pebble projection, so a stale or forged pending-cleanup row for an audit-live ledger suppresses both missing-bit and unaudited-bit errors. The current `TestCompareReversions_CleanupPendingSkipped` locks that behavior in: it marks `ledger-a` as live, writes an unaudited reversion bit, adds a pending cleanup marker, and expects no error.

This exception needs to be driven by audit-derived deletion state, not by the projection being checked. With this PR, the live FSM path deletes `SubPLReversions` rows immediately on `DeleteLedger`, and the rebuild path mirrors that, so stale reversion rows for deleted/non-live ledgers should not be treated as legitimate lingering cleanup data either. Otherwise a regression in the DeleteLedger replay cleanup, or a tampered pending-cleanup marker on a live ledger, can pass `Check()` cleanly while `Registry.Reversions` would be wrong on recovery.

Please make the checker compare the decoded stored reversion rows against the audit-derived live/deleted state in both directions, and adjust the cleanup test so a pending-cleanup marker alone cannot hide mismatches for an audit-live ledger.

2. Related hardening: the checker now uses `query.ReadReversions` as an integrity source, but that reader silently skips malformed keys/short values and does not return `iter.Error()`. For a checker pass, malformed rows under `ZonePerLedger/SubPLReversions` should be surfaced as an error/event instead of being dropped before comparison.

gfyrag:
Documentation follow-up: this PR adds a new restore/checker mechanism for `ZonePerLedger/SubPLReversions`, but the docs still describe the restore/checker surface without that projection pass. At minimum, please update the checker/restore documentation to mention the reversion-bitset projection validation, and correct the documented reversion key layout to the actual `[zone][sub][ledgerName padded 64B][wordIndex]` format. This is relevant here because the PR changes both rebuild behavior and invariant-#8 coverage for this projection.

gfyrag (approved):
Approved current head `88cd73b`. I rechecked the previous requested-change items: `compareReversions` is now driven by audit-derived state rather than pending-cleanup markers, stored rows for non-live ledgers are flagged, malformed reversion rows surface instead of narrowing the comparison silently, and restore rebuild now seeds/folds/flushes the reversion bitsets in the FSM layout.

Caveats: GitHub currently reports this PR as conflicting with `release/v3.0`, so it is not merge-ready until rebased/resolved. `codecov/patch` is also failing, but I did not find a remaining code blocker in this pass.

PR #1534, opened 2026-07-08
test(antithesis): expand model-driver coverage, extract oracle model library

Azorlogh:
## Summary

Extract the model checker's state model into a standalone `tests/oracle`
library, and substantially widen the `singleton_driver_model` Antithesis
driver's coverage of the transaction and metadata surface. Test-only — no
production code changes.

## What's here

**Oracle library** — the pure forward model (`GlobalState.Apply`) moves out of
the driver into `tests/oracle`, with its own unit tests and shared request
builders (`oracletest`). The live checker and the unit tests now drive the same
model.

**Coverage** — the driver and oracle now exercise:
- **Transactions**: multi-posting (intra-tx composition), non-world debits +
  balance floor (`INSUFFICIENT_FUNDS`), force, custom timestamps,
  `ExpandVolumes: false`, all revert modes (force / non-force / at-effective-date
  / receipt-carried), and the create rejection branches
  (`TRANSACTION_REFERENCE_CONFLICT`, empty → `VALIDATION`, `VOLUME_OVERFLOW`).
- **Metadata**: every value kind, `DATETIME` field types, revert-carried and
  account metadata, and deletes — all read back and verified.
- **Reads**: id-probed `GetTransaction`, metadata-schema read-back.

**Tooling** — `MODEL_DUMP_BATCHES` dumps every submitted bulk; `tests/oracle/cmd/replay`
replays them offline (committed or dispatch order) for deterministic divergence
diagnosis.

## Testing

`just test-model` (single-worker) and `just test-model-cluster` (3-node, rolling
restarts) pass with zero findings; oracle unit tests pass.

PR #1593, opened 2026-07-15
test(model): exercise RebuildDelta via a model-test restore cycle

Azorlogh:
Follow-up to #1534. Adds backup→restore coverage to the model test: the driver periodically quiesces itself, triggers a full restore, and resumes against the rebuilt store — so every continuous model check (committed bulks, failures, account/ledger/transaction/schema reads) validates the state `RebuildDelta` reconstructed, with no dedicated post-restore assertions needed. Wired both locally and on Antithesis k8s.

## How it works

- **Driver** (`restore.go`): a restore cycle fires every ~45s (180s on k8s) — `pauseAndDrain` gates the workers and waits for every in-flight bulk and read to resolve, then signals an external orchestrator through a file rendezvous and blocks until the restore completes, then resumes. The request file is taken by rename on both sides, so exactly one wins: the orchestrator claims it to start a cycle, the driver withdraws it on lease expiry — and if the claim already happened, the driver stays parked until the cycle answers rather than resuming mid-restore.
- **Local runner** (`run_model_test.sh --restore`, single-node): starts MinIO, takes a full backup of the empty store at boot so all activity lands in incremental deltas, then per cycle: incremental backup → SIGKILL the node → `ledgerctl store bootstrap` from the backup (RebuildDelta) → relaunch on the rebuilt store.
- **Antithesis k8s** (`restore-orchestrator.sh`, sidecar in the workload pod): services the same rendezvous with a full disaster-recovery pass through the operator — incremental backup at the quiesce point (exec `ledgerctl` in a ledger pod) → delete the Cluster CR and every PVC → re-create the Cluster in restore mode → `restore download` + `finalize` (RebuildDelta) → flip back to normal mode, where pod 0 self-bootstraps from the restored store and the other replicas join fresh via snapshot install. Fault-hardened: teardown is verified complete before the point of no return, the post-teardown restore retries until it lands (the driver's state exists only in the backup from there), download/finalize recognise lost-response retries as prior success, and an exports guard fails cycles whose delta would make RebuildDelta a no-op. The sidecar uses an image variant without `/opt/antithesis/test` so the Composer never schedules test commands into it.
- **Coverage signal**: a `Sometimes` property (`singleton_driver_model: restore cycle completed`) is registered statically, so a run with zero completed cycles reports it failing; the local runner additionally fails on zero cycles or any failed cycle.

## What it has already caught

Restore bugs fixed this week were found by exactly this cycle:
- #1554 — RebuildDelta lost the metadata schema/account types
- #1591 — RebuildDelta stringified typed account metadata (model held `b:false`, restored server returned `s:false`).
- #1592 — RebuildDelta lost the reversion bitsets, letting a double revert past the already-reverted gate after restore.

Locally, a 300s run with 5 restore cycles passes clean.

On Antithesis (runs `87dbab…-57-8`, `5e600bbb…-57-8`, 1h each), the restore property passes — and the cycle surfaced a further, still-open SUT finding: right after a restore, the freshly-joined follower's post-commit sentinel fires `aggregated volume imbalance` on its first applies and the node crash-loops (all hits on the snapshot-installed follower, never on the pod that booted from the restored store; a run whose cycles never completed shows no hits). Those runs' red `Unreachable` properties trace back to this, not to the harness.

### DIFF
--- a/tests/antithesis/run_model_test.sh
+++ b/tests/antithesis/run_model_test.sh
@@ -248,8 +248,11 @@ do_one_restore() {
 # service_restore_request handles one pending request, if any. Runs in the main
 # shell (not a subshell) so start_node's SERVER_PIDS update is visible to cleanup.
 service_restore_request() {
-	[ "$RESTORE" = 1 ] && [ -f "$RESTORE_REQ" ] || return 0
-	rm -f "$RESTORE_REQ"
+	[ "$RESTORE" = 1 ] || return 0
+	# Atomic claim, mirroring the driver's withdraw-by-rename on lease expiry:
+	# exactly one side gets the request.
+	mv "$RESTORE_REQ" "$RESTORE_REQ.claimed" 2>/dev/null || return 0
+	rm -f "$RESTORE_REQ.claimed"
 	log "restore: cycle requested"
 	if do_one_restore; then
 		RESTORE_CYCLES=$(( RESTORE_CYCLES + 1 ))
--- a/tests/antithesis/run_model_test.sh
+++ b/tests/antithesis/run_model_test.sh
@@ -21,7 +21,7 @@
 # the cluster failing to recover to N voters after a restart.
 #
 # Usage:
-#   ./run_model_test.sh [--nodes N | --cluster] [DURATION_SECONDS]
+#   ./run_model_test.sh [--nodes N | --cluster] [--restore] [DURATION_SECONDS]
 #   NODES=3 ./run_model_test.sh 300
 #
 # Environment:
@@ -30,6 +30,7 @@
 #   GRPC_BASE         base gRPC port; raft = +100, http = +200 per band (default: random)
 #   RESTART_INTERVAL  seconds to soak between restarts, N>1 only; 0 disables restarts (default: 15)
 #   RECOVER_TIMEOUT   seconds to wait for N-voter recovery after a restart, N>1 only (default: 90)
+#   RESTORE_INTERVAL  seconds between backup/restore cycles, --restore only (default: 45)
 #   DEAD_TIME             seconds a killed node stays down, N>1 only (default: 30)
 #   COMPACTION_MARGIN     raft log entries between snapshots; low forces snapshot recovery (default: 200)
 #   MAINTENANCE_INTERVAL  background WAL snapshot + checkpoint cadence (default: 10s)
@@ -51,7 +52,7 @@ while [ $# -gt 0 ]; do
 		--nodes)    NODES="$2"; shift 2 ;;
 		--nodes=*)  NODES="${1#*=}"; shift ;;
 		--restore)  RESTORE=1; shift ;;
-		-h|--help)  sed -n '2,40p' "$0"; exit 0 ;;
+		-h|--help)  sed -n '2,42p' "$0"; exit 0 ;;
 		*)          DURATION="$1"; shift ;;
 	esac
 done
@@ -72,9 +73,22 @@ case "$NODES" in
 esac
 [ "$NODES" -ge 1 ] || { echo "ERROR: --nodes must be >= 1" >&2; exit 2; }
 
-# Single-node iterates fast; a cluster needs longer to see restart cycles.
+# The driver fires the first restore within [interval, 1.5*interval] (jittered,
+# see runRestoreCycle), so a restore run's duration must cover at least one full
+# cycle or the run passes without exercising the restore path at all.
+RESTORE_INTERVAL="${RESTORE_INTERVAL:-45}"
+
+# Single-node iterates fast; a cluster needs longer to see restart cycles; a
+# restore run needs at least one backup/restore cycle (2*interval covers the
+# jittered first fire plus the cycle itself for any interval).
 if [ -z "$DURATION" ]; then
-	if [ "$NODES" -gt 1 ]; then DURATION=120; else DURATION=30; fi
+	if [ "$RESTORE" = 1 ]; then
+		DURATION=$(( RESTORE_INTERVAL * 2 + 30 ))
+	elif [ "$NODES" -gt 1 ]; then
+		DURATION=120
+	else
+		DURATION=30
+	fi
 fi
 
 RESTART_INTERVAL="${RESTART_INTERVAL:-15}"
@@ -125,9 +139,9 @@ LEDGERCTL_BIN="$WORKDIR/ledgerctl"
 DRIVER_PID=""
 RECOVERY_FAILED=0
 DRIVER_EXITED_EARLY=0
+RESTORE_CYCLES=0
 
 # --- Restore cycle (single-node, --restore) -------------------------------
-RESTORE_INTERVAL="${RESTORE_INTERVAL:-45}"
 RESTORE_REQ="$WORKDIR/restore.req"
 RESTORE_RESP="$WORKDIR/restore.resp"
 RESTORED_DATA="$WORKDIR/restored-data"
@@ -208,10 +222,16 @@ do_one_restore() {
 	fi
 
 	kill -9 "${SERVER_PIDS[0]}" 2>/dev/null; wait "${SERVER_PIDS[0]}" 2>/dev/null; SERVER_PIDS[0]=""
+
+	# The relaunched node appends to the same log file; only leadership lines
+	# past this offset count (see wait_leader).
+	local log_offset
+	log_offset=$(wc -c < "${SERVER_LOGS[0]}" 2>/dev/null || echo 0)
+
 	rm -rf "$RESTORED_DATA"
 	if ! "$LEDGERCTL_BIN" store bootstrap --data-dir "$RESTORED_DATA" "${S3_FLAGS[@]}" -y >>"$WORKDIR/restore.log" 2>&1; then
 		log "restore: bootstrap failed; relaunching on original data"
-		start_node 0 rejoin; wait_leader 0
+		start_node 0 rejoin; wait_leader 0 "$log_offset"
 		return 1
 	fi
 
@@ -220,7 +240,7 @@ do_one_restore() {
 	rm -rf "${NODE_DIRS[0]}/data" "${NODE_DIRS[0]}/wal"
 	mv "$RESTORED_DATA" "${NODE_DIRS[0]}/data"
 	start_node 0 bootstrap
-	wait_leader 0 || return 1
+	wait_leader 0 "$log_offset" || return 1
 	log "restore: node back up on restored (RebuildDelta) store"
 }
 
@@ -230,7 +250,12 @@ service_restore_request() {
 	[ "$RESTORE" = 1 ] && [ -f "$RESTORE_REQ" ] || return 0
 	rm -f "$RESTORE_REQ"
 	log "restore: cycle requested"
-	if do_one_restore; then printf 'ok\n' >"$RESTORE_RESP"; else printf 'err\n' >"$RESTORE_RESP"; fi
+	if do_one_restore; then
+		RESTORE_CYCLES=$(( RESTORE_CYCLES + 1 ))
+		printf 'ok\n' >"$RESTORE_RESP"
+	else
+		printf 'err\n' >"$RESTORE_RESP"
+	fi
 }
 
 if [ ! -d "$REPO" ]; then
@@ -269,11 +294,14 @@ start_node() {
 	SERVER_PIDS[$i]=$!
 }
 
-# Wait (bounded) for a node to log leadership; fail if it dies first.
+# Wait (bounded) for a node to log leadership; fail if it dies first. A caller
+# relaunching a node passes the log's current byte size: server logs are
+# appended across boots, so without the offset a leadership line from an
+# earlier boot satisfies the check instantly.
 wait_leader() {
-	local i="$1"
+	local i="$1" offset="${2:-0}"
 	for _ in $(seq 1 60); do
-		if grep -qE "Became leader|became leader at term" "${SERVER_LOGS[$i]}" 2>/dev/null; then return 0; fi
+		if tail -c +"$(( offset + 1 ))" "${SERVER_LOGS[$i]}" 2>/dev/null | grep -qE "Became leader|became leader at term"; then return 0; fi
 		if ! kill -0 "${SERVER_PIDS[$i]}" 2>/dev/null; then
 			echo "ERROR: node $(( i + 1 )) exited during startup:" >&2
 			tail -20 "${SERVER_LOGS[$i]}" >&2
@@ -490,6 +518,7 @@ echo
 echo "================= model test report ================="
 log "topology: $NODES node(s)"
 [ "$NODES" -gt 1 ] && log "restart cycles completed: $cycle"
+[ "$RESTORE" = 1 ] && log "restore cycles completed: $RESTORE_CYCLES"
 
 findings=0
 
@@ -559,6 +588,17 @@ if [ "$DRIVER_EXITED_EARLY" -ne 0 ] && [ "$findings" -eq 0 ]; then
 	echo "DRIVER EXITED EARLY: the model ran for less than the requested ${DURATION}s"
 	echo "  (no assertion or crash recorded -- likely a setup/connection error; driver log:)"
 	tail -20 "$DRIVER_LOG" 2>/dev/null
+	findings=$((findings + 1))
+fi
+
+# 6. --restore ran zero cycles: the restore path was never exercised, so a
+# green run says nothing about RebuildDelta -- that vacuous pass is itself a
+# failure. Skipped when a finding already stopped the run early (fail-fast
+# legitimately preempts the first cycle).
+if [ "$RESTORE" = 1 ] && [ "$RESTORE_CYCLES" -eq 0 ] && [ "$findings" -eq 0 ]; then
+	echo
+	echo "NO RESTORE CYCLES: --restore was requested but no backup/restore cycle completed"
+	echo "  (duration ${DURATION}s vs first cycle within ~$(( RESTORE_INTERVAL * 3 / 2 ))s; increase the duration)"
 	findings=$((findings + 1))
 fi
 
--- a/tests/antithesis/run_model_test.sh
+++ b/tests/antithesis/run_model_test.sh
@@ -88,6 +88,9 @@ RECOVER_TIMEOUT="${RECOVER_TIMEOUT:-90}"
 DEAD_TIME="${DEAD_TIME:-30}"
 COMPACTION_MARGIN="${COMPACTION_MARGIN:-200}"
 MAINTENANCE_INTERVAL="${MAINTENANCE_INTERVAL:-10s}"
+# WAL/data disk-usage guard (fraction). Default matches the server; raise it when
+# the host filesystem is already above 80% used (the guard blocks writes then).
+HEALTH_THRESHOLD="${HEALTH_THRESHOLD:-0.8}"
 CLUSTER_ID="model-test-cluster"
 # HMAC key for signed transaction receipts, so the driver can exercise the
 # receipt-carried revert path (admission verifies the receipt and reverses its
@@ -113,7 +116,7 @@ RAFT_BASE=$(( GRPC_BASE + 100 ))
 HTTP_BASE=$(( GRPC_BASE + 200 ))
 MINIO_PORT=$(( GRPC_BASE + 300 ))
 
-WORKDIR="$(mktemp -d /tmp/model-test.XXXXXX)"
+WORKDIR="$(mktemp -d "${TMPDIR:-/tmp}/model-test.XXXXXX")"
 DRIVER_LOG="$WORKDIR/driver.log"
 ASSERTIONS="$WORKDIR/assertions.json"
 SERVER_BIN="$WORKDIR/ledger-server"
@@ -259,6 +262,8 @@ start_node() {
 		--data-dir "${NODE_DIRS[$i]}/data" \
 		--raft-compaction-margin "$COMPACTION_MARGIN" \
 		--maintenance-interval "$MAINTENANCE_INTERVAL" \
+		--health-wal-threshold "$HEALTH_THRESHOLD" \
+		--health-data-threshold "$HEALTH_THRESHOLD" \
 		${flags[@]+"${flags[@]}"} \
 		>> "${SERVER_LOGS[$i]}" 2>&1 &
 	SERVER_PIDS[$i]=$!
--- a/tests/antithesis/run_model_test.sh
+++ b/tests/antithesis/run_model_test.sh
@@ -43,16 +43,29 @@ set -uo pipefail
 
 # --- Arguments ------------------------------------------------------------
 NODES="${NODES:-1}"
+RESTORE="${RESTORE:-0}"
 DURATION=""
 while [ $# -gt 0 ]; do
 	case "$1" in
 		--cluster)  NODES=3; shift ;;
 		--nodes)    NODES="$2"; shift 2 ;;
 		--nodes=*)  NODES="${1#*=}"; shift ;;
+		--restore)  RESTORE=1; shift ;;
 		-h|--help)  sed -n '2,40p' "$0"; exit 0 ;;
 		*)          DURATION="$1"; shift ;;
 	esac
 done
+
+case "$RESTORE" in ''|0|off|false|no) RESTORE=0 ;; *) RESTORE=1 ;; esac
+# The restore cycle periodically backs the node up, restores the backup into a
+# fresh store (exercising RebuildDelta), and relaunches the node on it while the
+# driver keeps running -- its ordinary checks validate the rebuilt state. Only
+# single-node is wired: a whole-store swap keeps the served state == RebuildDelta's
+# output (a cluster rejoin would snapshot-install over it).
+if [ "$RESTORE" = 1 ] && [ "$NODES" -ne 1 ]; then
+	echo "ERROR: --restore is only supported with a single node (got --nodes $NODES)" >&2
+	exit 2
+fi
 
 case "$NODES" in
 	''|*[!0-9]*) echo "ERROR: --nodes must be a positive integer (got '$NODES')" >&2; exit 2 ;;
@@ -98,6 +111,7 @@ REPO="${REPO:-$(cd "$SCRIPT_DIR/../.." && pwd)}"
 GRPC_BASE="${GRPC_BASE:-$(( 20000 + RANDOM % 10000 ))}"
 RAFT_BASE=$(( GRPC_BASE + 100 ))
 HTTP_BASE=$(( GRPC_BASE + 200 ))
+MINIO_PORT=$(( GRPC_BASE + 300 ))
 
 WORKDIR="$(mktemp -d /tmp/model-test.XXXXXX)"
 DRIVER_LOG="$WORKDIR/driver.log"
@@ -108,6 +122,20 @@ LEDGERCTL_BIN="$WORKDIR/ledgerctl"
 DRIVER_PID=""
 RECOVERY_FAILED=0
 DRIVER_EXITED_EARLY=0
+
+# --- Restore cycle (single-node, --restore) -------------------------------
+RESTORE_INTERVAL="${RESTORE_INTERVAL:-45}"
+RESTORE_REQ="$WORKDIR/restore.req"
+RESTORE_RESP="$WORKDIR/restore.resp"
+RESTORED_DATA="$WORKDIR/restored-data"
+MINIO_CONTAINER="model-minio-$$"
+S3_BUCKET="backups"
+S3_FLAGS=(
+	--driver s3 --s3-bucket "$S3_BUCKET"
+	--s3-endpoint "http://127.0.0.1:$MINIO_PORT" --s3-region us-east-1
+	--s3-access-key-id minioadmin --s3-secret-access-key minioadmin
+	--bucket-id "$CLUSTER_ID"
+)
 
 GRPC_PORTS=() RAFT_PORTS=() HTTP_PORTS=() NODE_DIRS=() SERVER_LOGS=() SERVER_PIDS=()
 for i in $(seq 0 $(( NODES - 1 ))); do
@@ -129,10 +157,78 @@ cleanup() {
 	sleep 1
 	[ -n "$DRIVER_PID" ] && kill -9 "$DRIVER_PID" 2>/dev/null
 	for pid in "${SERVER_PIDS[@]}"; do [ -n "$pid" ] && kill -9 "$pid" 2>/dev/null; done
+	[ "$RESTORE" = 1 ] && stop_minio
 	wait 2>/dev/null
 	if [ -z "${KEEP_WORKDIR:-}" ]; then rm -rf "$WORKDIR"; else log "work dir kept at $WORKDIR"; fi
 }
 trap cleanup EXIT INT TERM
+
+# --- MinIO (backup storage for --restore) ---------------------------------
+start_minio() {
+	docker rm -f "$MINIO_CONTAINER" >/dev/null 2>&1
+	docker run -d --rm --name "$MINIO_CONTAINER" -p "127.0.0.1:$MINIO_PORT:9000" \
+		-e MINIO_ROOT_USER=minioadmin -e MINIO_ROOT_PASSWORD=minioadmin \
+		minio/minio:RELEASE.2024-06-13T22-53-53Z server /data >/dev/null 2>&1 || return 1
+	# Retry until MinIO is reachable, then create the bucket (idempotent). The mc
+	# image's entrypoint is mc itself, so override it to run a shell.
+	for _ in $(seq 1 30); do
+		if docker run --rm --network host --entrypoint sh minio/mc:RELEASE.2024-06-12T14-34-03Z -c \
+			"mc alias set m http://127.0.0.1:$MINIO_PORT minioadmin minioadmin && mc mb --ignore-existing m/$S3_BUCKET" >/dev/null 2>&1; then
+			return 0
+		fi
+		sleep 1
+	done
+	return 1
+}
+stop_minio() { docker rm -f "$MINIO_CONTAINER" >/dev/null 2>&1; }
+
+# --- Restore orchestrator (single-node) -----------------------------------
+# One cycle: incremental-backup the (quiesced) node, kill it, bootstrap the
+# backup into a fresh store (RebuildDelta), then relaunch on it. The driver is
+# paused across this and resumes against the restored node.
+do_one_restore() {
+	local srv="127.0.0.1:${GRPC_PORTS[0]}" out
+	out="$("$LEDGERCTL_BIN" store incremental-backup --insecure --server "$srv" \
+		"${S3_FLAGS[@]}" --json --timeout 120s 2>&1)" \
+		|| { log "restore: incremental-backup failed: $(printf '%s' "$out" | tail -1)"; return 1; }
+
+	# RebuildDelta only runs when the manifest carries exports; a cycle that
+	# produced none would silently test nothing, so skip it loudly.
+	if command -v jq >/dev/null 2>&1; then
+		local segs logs
+		segs="$(printf '%s' "$out" | jq -r '.segmentsUploaded // 0' 2>/dev/null)"
+		logs="$(printf '%s' "$out" | jq -r '.logEntriesExported // 0' 2>/dev/null)"
+		if [ "${segs:-0}" = 0 ] && [ "${logs:-0}" = 0 ]; then
+			log "restore: incremental produced no exports (RebuildDelta would be a no-op); skipping"
+			return 1
+		fi
+	fi
+
+	kill -9 "${SERVER_PIDS[0]}" 2>/dev/null; wait "${SERVER_PIDS[0]}" 2>/dev/null; SERVER_PIDS[0]=""
+	rm -rf "$RESTORED_DATA"
+	if ! "$LEDGERCTL_BIN" store bootstrap --data-dir "$RESTORED_DATA" "${S3_FLAGS[@]}" -y >>"$WORKDIR/restore.log" 2>&1; then
+		log "restore: bootstrap failed; relaunching on original data"
+		start_node 0 rejoin; wait_leader 0
+		return 1
+	fi
+
+	# Swap the restored store in with a fresh WAL: on boot the RESTORED marker
+	# makes the node self-bootstrap as a single voter from the rebuilt data.
+	rm -rf "${NODE_DIRS[0]}/data" "${NODE_DIRS[0]}/wal"
+	mv "$RESTORED_DATA" "${NODE_DIRS[0]}/data"
+	start_node 0 bootstrap
+	wait_leader 0 || return 1
+	log "restore: node back up on restored (RebuildDelta) store"
+}
+
+# service_restore_request handles one pending request, if any. Runs in the main
+# shell (not a subshell) so start_node's SERVER_PIDS update is visible to cleanup.
+service_restore_request() {
+	[ "$RESTORE" = 1 ] && [ -f "$RESTORE_REQ" ] || return 0
+	rm -f "$RESTORE_REQ"
+	log "restore: cycle requested"
+	if do_one_restore; then printf 'ok\n' >"$RESTORE_RESP"; else printf 'err\n' >"$RESTORE_RESP"; fi
+}
 
 if [ ! -d "$REPO" ]; then
 	echo "ERROR: REPO not found at $REPO (set REPO=...)" >&2
@@ -237,9 +333,13 @@ check_fail_fast() {
 # ---------------------------------------------------------------------------
 # Build (server + driver; ledgerctl only when a cluster needs health checks).
 # ---------------------------------------------------------------------------
-build_cmd="go build -o '$SERVER_BIN' . && "
-if [ "$NODES" -gt 1 ]; then
-	build_cmd="${build_cmd}go build -o '$LEDGERCTL_BIN' ./cmd/ledgerctl && "
+# --restore drives S3 backups (to MinIO), so the server and ledgerctl need the
+# s3 build tag; the light default build stubs S3 out.
+server_tags=""
+[ "$RESTORE" = 1 ] && server_tags="-tags s3"
+build_cmd="go build $server_tags -o '$SERVER_BIN' . && "
+if [ "$NODES" -gt 1 ] || [ "$RESTORE" = 1 ]; then
+	build_cmd="${build_cmd}go build $server_tags -o '$LEDGERCTL_BIN' ./cmd/ledgerctl && "
 fi
 build_cmd="${build_cmd}cd tests/antithesis/workload && go build -o '$DRIVER_BIN' ./bin/cmds/model/singleton_driver_model"
 
@@ -286,6 +386,24 @@ else
 fi
 
 # ---------------------------------------------------------------------------
+# Restore setup (--restore): MinIO + one full backup as the checkpoint every
+# per-cycle incremental exports beyond (so bootstrap replays through RebuildDelta).
+# ---------------------------------------------------------------------------
+if [ "$RESTORE" = 1 ]; then
+	log "starting MinIO (:$MINIO_PORT) for backup storage..."
+	if ! start_minio; then echo "ERROR: MinIO did not start" >&2; exit 2; fi
+	log "taking initial full backup..."
+	if ! "$LEDGERCTL_BIN" store backup --insecure --server "127.0.0.1:${GRPC_PORTS[0]}" \
+		"${S3_FLAGS[@]}" --timeout 120s >>"$WORKDIR/restore.log" 2>&1; then
+		echo "ERROR: initial backup failed (see $WORKDIR/restore.log)" >&2
+		[ -n "${KEEP_WORKDIR:-}" ] || cat "$WORKDIR/restore.log" >&2
+		exit 2
+	fi
+	command -v jq >/dev/null 2>&1 || log "WARNING: jq not found; cannot verify each cycle produced exports"
+	log "restore cycle armed (interval ~${RESTORE_INTERVAL}s)"
+fi
+
+# ---------------------------------------------------------------------------
 # Run the driver against all node(s).
 # ---------------------------------------------------------------------------
 ADDR_LIST="127.0.0.1:${GRPC_PORTS[0]}"
@@ -295,12 +413,19 @@ log "running driver for ${DURATION}s against $ADDR_LIST ..."
 # MODEL_MAX_SECONDS makes the driver self-terminate even if this script never
 # gets to signal it (defence-in-depth against orphaned drivers). A small buffer
 # over DURATION lets the script-driven stop win in the normal case.
+RESTORE_REQ_ENV="" RESTORE_RESP_ENV="" RESTORE_INTERVAL_ENV=""
+if [ "$RESTORE" = 1 ]; then
+	RESTORE_REQ_ENV="$RESTORE_REQ"; RESTORE_RESP_ENV="$RESTORE_RESP"; RESTORE_INTERVAL_ENV="$RESTORE_INTERVAL"
+fi
 LEDGER_GRPC_ADDR="$ADDR_LIST" \
 ANTITHESIS_SDK_LOCAL_OUTPUT="$ASSERTIONS" \
 MODEL_DEBUG="${MODEL_DEBUG:-}" \
 MODEL_LEDGERS="${MODEL_LEDGERS:-}" \
 MODEL_WORKERS="${MODEL_WORKERS:-}" \
 MODEL_MAX_SECONDS="$(( DURATION + 15 ))" \
+MODEL_RESTORE_REQ="$RESTORE_REQ_ENV" \
+MODEL_RESTORE_RESP="$RESTORE_RESP_ENV" \
+MODEL_RESTORE_INTERVAL="$RESTORE_INTERVAL_ENV" \
 	"$DRIVER_BIN" > "$DRIVER_LOG" 2>&1 &
 DRIVER_PID=$!
 
@@ -341,6 +466,8 @@ while [ "$(date +%s)" -lt "$deadline" ]; do
 		next_restart=$(( $(date +%s) + RESTART_INTERVAL ))
 		continue
 	fi
+
+	service_restore_request
 	sleep 1
 done
 
--- a/tests/antithesis/run_model_test.sh
+++ b/tests/antithesis/run_model_test.sh
@@ -140,6 +140,7 @@ DRIVER_PID=""
 RECOVERY_FAILED=0
 DRIVER_EXITED_EARLY=0
 RESTORE_CYCLES=0
+RESTORE_FAILED_CYCLES=0
 
 # --- Restore cycle (single-node, --restore) -------------------------------
 RESTORE_REQ="$WORKDIR/restore.req"
@@ -254,6 +255,7 @@ service_restore_request() {
 		RESTORE_CYCLES=$(( RESTORE_CYCLES + 1 ))
 		printf 'ok\n' >"$RESTORE_RESP"
 	else
+		RESTORE_FAILED_CYCLES=$(( RESTORE_FAILED_CYCLES + 1 ))
 		printf 'err\n' >"$RESTORE_RESP"
 	fi
 }
@@ -518,7 +520,7 @@ echo
 echo "================= model test report ================="
 log "topology: $NODES node(s)"
 [ "$NODES" -gt 1 ] && log "restart cycles completed: $cycle"
-[ "$RESTORE" = 1 ] && log "restore cycles completed: $RESTORE_CYCLES"
+[ "$RESTORE" = 1 ] && log "restore cycles completed: $RESTORE_CYCLES ($RESTORE_FAILED_CYCLES failed)"
 
 findings=0
 
@@ -599,6 +601,15 @@ if [ "$RESTORE" = 1 ] && [ "$RESTORE_CYCLES" -eq 0 ] && [ "$findings" -eq 0 ]; t
 	echo
 	echo "NO RESTORE CYCLES: --restore was requested but no backup/restore cycle completed"
 	echo "  (duration ${DURATION}s vs first cycle within ~$(( RESTORE_INTERVAL * 3 / 2 ))s; increase the duration)"
+	findings=$((findings + 1))
+fi
+
+# 7. A restore cycle failed mid-run (backup RPC, no exports, bootstrap, or the
+# post-restore leader wait). The driver resumes so later cycles still run, but
+# a broken restore path is a defect regardless of how many cycles succeeded.
+if [ "$RESTORE" = 1 ] && [ "$RESTORE_FAILED_CYCLES" -gt 0 ]; then
+	echo
+	echo "RESTORE CYCLE FAILURES: $RESTORE_FAILED_CYCLES cycle(s) failed (see 'restore:' lines above)"
 	findings=$((findings + 1))
 fi
 
--- a/tests/antithesis/workload/bin/cmds/model/singleton_driver_model/restore.go
+++ b/tests/antithesis/workload/bin/cmds/model/singleton_driver_model/restore.go
@@ -1,0 +1,178 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/formancehq/ledger/v3/tests/antithesis/workload/internal"
+)
+
+// The restore cycle periodically hands the live ledger to an external
+// orchestrator that backs it up, restores the backup into a fresh store (running
+// the RebuildDelta replay), and brings a node back up on it. The driver keeps
+// running against the restored node afterward, so its ordinary read/commit
+// checks validate the rebuilt state — no separate comparison. The driver owns
+// only the timing and the quiescence; the environment-specific work lives behind
+// RestoreTrigger (a file rendezvous locally, the k8s client on Antithesis).
+
+// awaitResume blocks while a restore cycle has paused dispatch, returning false
+// if ctx ends. Safe to call without holding mu.
+func (c *Checker) awaitResume(ctx context.Context) bool {
+	for {
+		c.mu.Lock()
+		if !c.paused {
+			c.mu.Unlock()
+			return true
+		}
+		ch := c.resumeCh
+		c.mu.Unlock()
+
+		select {
+		case <-ch:
+		case <-ctx.Done():
+			return false
+		}
+	}
+}
+
+// pauseAndDrain stops new dispatch, then waits until every dispatched operation
+// (in-flight bulk, outstanding read, buffered success) has drained — at which
+// point modelState is the exact committed state the backup will capture. Returns
+// false if ctx ends first. Pair with resume.
+func (c *Checker) pauseAndDrain(ctx context.Context) bool {
+	c.mu.Lock()
+	if !c.paused {
+		c.paused = true
+		c.resumeCh = make(chan struct{})
+	}
+	c.mu.Unlock()
+
+	for {
+		c.mu.Lock()
+		_, empty := c.earliestOutstanding()
+		if empty {
+			// Nothing is in flight; flush the re-order buffer. handleObservation's
+			// failure/transient paths drop the ticket without calling tryDrain, so
+			// with no further op to re-trigger it a committed success can sit here.
+			c.tryDrain()
+		}
+		idle := empty && len(c.pending) == 0
+		c.mu.Unlock()
+		if idle {
+			return true
+		}
+
+		select {
+		case <-ctx.Done():
+			return false
+		case <-time.After(quiescePoll):
+		}
+	}
+}
+
+// resume releases workers parked in awaitResume.
+func (c *Checker) resume() {
+	c.mu.Lock()
+	if c.paused {
+		c.paused = false
+		close(c.resumeCh)
+	}
+	c.mu.Unlock()
+}
+
+// RestoreTrigger fires one backup+restore cycle and returns once the ledger is
+// serving again from restored state. Fire runs while the driver is quiesced.
+type RestoreTrigger interface {
+	Fire(ctx context.Context) error
+}
+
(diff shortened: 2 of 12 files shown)


tests/antithesis/run_model_test.sh
```diff
@@ -109,6 +126,20 @@ DRIVER_PID=""
 RECOVERY_FAILED=0
 DRIVER_EXITED_EARLY=0
 
+# --- Restore cycle (single-node, --restore) -------------------------------
+RESTORE_INTERVAL="${RESTORE_INTERVAL:-45}"
```
NumaryBot: 🟠 **[major] Restore mode may complete without exercising any backup/restore cycle**

When running with `--restore` and no explicit duration, the script still uses the single-node default timeout (30s). The first restore request is only scheduled after `RESTORE_INTERVAL` plus jitter (45–67s by default). This means the `--restore` code path can finish successfully without ever performing a backup/restore cycle, producing a false pass and defeating the purpose of the new coverage.

**Suggestion:** When `--restore` is active and no explicit duration is provided, default the run duration to at least `RESTORE_INTERVAL` + max-jitter + a safety margin (e.g. 120s) so that at least one full backup/restore cycle is guaranteed to execute before the script exits.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1593#issuecomment-4980540231

tests/antithesis/run_model_test.sh
```diff
@@ -129,11 +160,79 @@ cleanup() {
...
+		log "restore: bootstrap failed; relaunching on original data"
+		start_node 0 rejoin; wait_leader 0
+		return 1
+	fi
+
+	# Swap the restored store in with a fresh WAL: on boot the RESTORED marker
+	# makes the node self-bootstrap as a single voter from the rebuilt data.
+	rm -rf "${NODE_DIRS[0]}/data" "${NODE_DIRS[0]}/wal"
+	mv "$RESTORED_DATA" "${NODE_DIRS[0]}/data"
+	start_node 0 bootstrap
```
NumaryBot, line 243: 🟠 **[major] Stale log lines can make `wait_leader` return prematurely after a restore**

After a restore, `wait_leader` greps the same server log file that already contains the initial boot's `Became leader` line. The grep therefore returns immediately even if the restored node has failed to start or is not yet ready. The orchestrator writes `ok`, the driver resumes against a dead/not-ready server, and the restore cycle can silently pass without actually validating the restored store's liveness.

**Suggestion:** Before calling `wait_leader` post-restore, truncate or rotate the log file (or record the current byte offset and grep only new output) so that only lines emitted after the restore is initiated are considered as evidence of a successful leader election.

tests/antithesis/run_model_test.sh
```diff
@@ -98,8 +128,9 @@ REPO="${REPO:-$(cd "$SCRIPT_DIR/../.." && pwd)}"
 GRPC_BASE="${GRPC_BASE:-$(( 20000 + RANDOM % 10000 ))}"
 RAFT_BASE=$(( GRPC_BASE + 100 ))
 HTTP_BASE=$(( GRPC_BASE + 200 ))
```
NumaryBot, line 130: 🟠 **[major] Restore mode may complete without exercising any backup/restore cycle**

When running with `--restore` and no explicit duration, the script still uses the single-node default timeout (30s). The first restore request is only scheduled after `RESTORE_INTERVAL` plus jitter (45–67s by default). This means the `--restore` code path can finish successfully without ever performing a backup/restore cycle, producing a false pass and defeating the purpose of the new coverage.

**Suggestion:** When `--restore` is active and no explicit duration is provided, default the run duration to at least `RESTORE_INTERVAL` + max-jitter + a safety margin (e.g. 120s) so that at least one full backup/restore cycle is guaranteed to execute before the script exits.

tests/antithesis/run_model_test.sh
```diff
@@ -129,11 +174,90 @@ cleanup() {
...
+		log "restore: bootstrap failed; relaunching on original data"
+		start_node 0 rejoin; wait_leader 0 "$log_offset"
+		return 1
+	fi
+
+	# Swap the restored store in with a fresh WAL: on boot the RESTORED marker
+	# makes the node self-bootstrap as a single voter from the rebuilt data.
+	rm -rf "${NODE_DIRS[0]}/data" "${NODE_DIRS[0]}/wal"
+	mv "$RESTORED_DATA" "${NODE_DIRS[0]}/data"
+	start_node 0 bootstrap
```
NumaryBot, line 243: 🟠 **[major] Stale log lines can make `wait_leader` return prematurely after a restore**

After a restore, `wait_leader` greps the same server log file that already contains the initial boot's `Became leader` line. The grep therefore returns immediately even if the restored node has failed to start or is not yet ready. The orchestrator writes `ok`, the driver resumes against a dead/not-ready server, and the restore cycle can silently pass without actually validating the restored store's liveness.

**Suggestion:** Before calling `wait_leader` post-restore, truncate or rotate the log file (or record the current byte offset and grep only new output) so that only lines emitted after the restore is initiated are considered as evidence of a successful leader election.

tests/antithesis/run_model_test.sh
```diff
@@ -129,11 +174,90 @@ cleanup() {
...
+	log "restore: node back up on restored (RebuildDelta) store"
+}
+
+# service_restore_request handles one pending request, if any. Runs in the main
+# shell (not a subshell) so start_node's SERVER_PIDS update is visible to cleanup.
+service_restore_request() {
+	[ "$RESTORE" = 1 ] && [ -f "$RESTORE_REQ" ] || return 0
+	rm -f "$RESTORE_REQ"
+	log "restore: cycle requested"
+	if do_one_restore; then
```
NumaryBot, line 257: 🟠 **[major] Failed restore cycles not surfaced — restore mode can report PASS despite mid-run failures**

If a restore attempt fails after at least one prior successful cycle (e.g., a later `store bootstrap` regression or post-restore leader wait failure), `runRestoreCycle` logs the error and continues. The final report only fails when `RESTORE_CYCLES == 0`. This means `--restore` can report PASS even though one or more backup/restore cycles did not successfully restore the node or validate the rebuilt store.

**Suggestion:** Track any per-cycle error and propagate it to the final exit code, so that a single failing restore cycle causes the overall run to be reported as FAIL regardless of how many other cycles succeeded.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1593#issuecomment-4980540231

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1593#issuecomment-4980540231

flemzord (approved):
Reviewed at `32b47dd`. The restore cycle now exercises a non-empty incremental backup through store bootstrap/RebuildDelta, resumes the existing model checks against the rebuilt store, and fails the run when no cycle completes or any cycle fails. The shutdown, relaunch, and quiescence paths are coherent, all prior review threads are resolved, and the full CI suite is green. I found no remaining actionable issue.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,167 @@
...
+
+log() { echo "[restore-orchestrator] $*"; }
+
+k() { kubectl -n "$NS" "$@"; }
+
+# exec_ledgerctl POD ARGS... -- runs ledgerctl inside a ledger pod with the
+# pod-local server address and cluster credentials (the operator-e2e idiom).
+exec_ledgerctl() {
+	local pod="$1"; shift
+	k exec "$pod" -c ledger -- /bin/sh -c \
```
NumaryBot, line 60: 🟠 **[major] Unsupported `--timeout` flag passed to `ledgerctl restore download`**

The restore orchestrator invokes `ledgerctl restore download` with a `--timeout` flag that the command does not recognise (it uses `--rpc-timeout`/`--poll-interval` instead). Cobra will reject the unknown flag at runtime, causing every restore cycle to fail before `RebuildDelta` is ever exercised. This defeats the stated purpose of the PR.

**Suggestion:** Replace `--timeout` with the correct flags accepted by `ledgerctl restore download` (e.g. `--rpc-timeout` and/or `--poll-interval`). Verify the exact flag names against the CLI definition before merging.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1593#issuecomment-4980540231

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1593#issuecomment-4980540231

tests/antithesis/workload/restore-orchestrator.sh
NumaryBot, line 60: 🟠 **[major] Unsupported `--timeout` flag passed to `ledgerctl restore download`**

`ledgerctl restore download` does not accept a `--timeout` flag; the CLI uses `--rpc-timeout` and/or `--poll-interval` instead. Cobra rejects the unknown flag at runtime, causing every restore cycle to fail before `RebuildDelta` is ever reached. This defeats the core purpose of the PR.

**Suggestion:** Replace `--timeout` with the correct flags accepted by `ledgerctl restore download` (e.g. `--rpc-timeout` and/or `--poll-interval`). Verify the exact flag names against the CLI definition before merging.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,167 @@
...
+while true; do
+	if [ ! -f "$REQ" ]; then sleep 1; continue; fi
+	rm -f "$REQ" "$RESP"
+	log "restore cycle requested"
+	if do_one_restore; then
+		printf 'ok\n' > "$RESP"
+	else
+		# Converge back to normal mode so the driver never resumes against a
+		# cluster parked in restore mode; the driver logs the err and continues.
+		k apply -f "$NORMAL_SPEC" 2>/dev/null || true
```
NumaryBot: 🟠 **[major] Failed restore cycles do not restore full cluster readiness before signalling error**

When `do_one_restore` fails mid-way (e.g. during `restore download` or `finalize`), the error path reapplies `NORMAL_SPEC` but does not delete pod 0 or wait for the cluster to become ready, unlike the success path. The driver resumes against an absent or still-in-restore-mode cluster, causing cascading failures and preventing later restore cycles from providing meaningful coverage.

**Suggestion:** Mirror the success path's pod-delete-and-wait logic in the error recovery path so that the cluster is fully back in normal mode and ready before `err` is written to the pipe and the driver resumes.

tests/antithesis/workload/restore-orchestrator.sh
NumaryBot, line 60: 🟠 **[major] Unsupported `--timeout` flag passed to `ledgerctl restore download`**

The restore orchestrator appends `--timeout 120s` to `ledgerctl restore download`, but that command does not recognise `--timeout` (it uses `--rpc-timeout` and/or `--poll-interval`). Cobra rejects the unknown flag at runtime, causing every restore cycle to fail before `RebuildDelta` is ever exercised, which defeats the stated purpose of the PR.

**Suggestion:** Replace `--timeout` with the correct flags accepted by `ledgerctl restore download` (e.g. `--rpc-timeout` and/or `--poll-interval`). Verify the exact flag names against the CLI definition before merging.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1593#issuecomment-4980540231

tests/antithesis/workload/restore-orchestrator.sh
NumaryBot, line 60: 🔴 **[blocker] Unsupported `--timeout` flag passed to `ledgerctl restore download`**

`ledgerctl restore download` does not accept a `--timeout` flag; the CLI uses `--rpc-timeout` and/or `--poll-interval` instead. Cobra rejects the unknown flag at runtime, causing every restore cycle to fail before `RebuildDelta` is ever exercised. This defeats the stated purpose of the PR and was flagged in multiple prior review rounds without being resolved.

**Suggestion:** Replace `--timeout 120s` with the correct flags accepted by `ledgerctl restore download` (e.g. `--rpc-timeout` and/or `--poll-interval`). Verify the exact flag names against the CLI definition before merging.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,167 @@
...
+	wait_ready_replicas "$REPLICAS" || { log "cluster did not return to $REPLICAS ready replicas"; return 1; }
+	log "cluster back up on the restored (RebuildDelta) store"
+}
+
+log "waiting for $STS to reach $REPLICAS ready replicas"
+wait_ready_replicas "$REPLICAS" 1800 || log "WARNING: cluster not ready at startup; continuing"
+
+retry 10 capture_specs || { log "FATAL: cannot capture cluster spec"; exit 1; }
+
+log "taking initial full backup"
```
NumaryBot, line 238: 🟠 **[major] Failed initial full backup leaves all subsequent incremental cycles unable to run**

If the startup full backup exhausts its retries (e.g. because the cluster or MinIO is not yet ready), the sidecar only logs a warning and later cycles proceed to `store incremental-backup`. Incremental backups require a prior full backup to exist; without one every incremental attempt will fail, no restore cycle will succeed, and `RebuildDelta` will never be exercised despite the PR's intent.

**Suggestion:** Treat a fully-exhausted initial full backup as a fatal error (exit non-zero) rather than a warning, or implement a retry mechanism that defers the incremental cycle until a successful full backup exists.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1593#issuecomment-4980540231

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,187 @@
...
+k() { kubectl -n "$NS" "$@"; }
+
+# exec_ledgerctl POD ARGS... -- runs ledgerctl inside a ledger pod with the
+# pod-local server address and cluster credentials (the operator-e2e idiom).
+# Bounded by coreutils timeout rather than ledgerctl's --timeout: not every
+# subcommand defines that flag (restore download does not), and a uniform
+# client-side bound also covers a hung exec stream.
+exec_ledgerctl() {
+	local pod="$1"; shift
+	k exec "$pod" -c ledger -- /bin/sh -c \
```
NumaryBot, line 60: 🔴 **[blocker] Unsupported `--timeout` flag passed to `ledgerctl restore download`**

`ledgerctl restore download` does not accept a `--timeout` flag; the CLI uses `--rpc-timeout` and/or `--poll-interval` instead. Cobra rejects the unknown flag at runtime, causing every restore cycle to fail before `RebuildDelta` is ever exercised. This defeats the core purpose of the PR and has been raised across multiple prior review rounds without resolution.

**Suggestion:** Replace `--timeout 120s` with the correct flags accepted by `ledgerctl restore download` (e.g. `--rpc-timeout` and/or `--poll-interval`). Verify the exact flag names against the CLI definition before merging.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,187 @@
...
+	log "cluster back up on the restored (RebuildDelta) store"
+}
+
+log "waiting for $STS to reach $REPLICAS ready replicas"
+wait_ready_replicas "$REPLICAS" 1800 || log "WARNING: cluster not ready at startup; continuing"
+
+retry 10 capture_specs || { log "FATAL: cannot capture cluster spec"; exit 1; }
+
+FULL_BACKUP_OK=0
+log "taking initial full backup"
```
NumaryBot, line 238: 🟠 **[major] Exhausted initial full backup treated as warning, breaking all subsequent incremental cycles**

If the startup full backup exhausts all retries (e.g. due to a slow cluster or MinIO startup), the sidecar only logs a warning and later cycles proceed directly to `store incremental-backup`. Incremental backups require a prior full backup to exist; without one, every incremental attempt will fail, no restore cycle will succeed, and RebuildDelta will never be exercised.

**Suggestion:** Treat a fully-exhausted initial full backup as a fatal error (exit non-zero) rather than a warning, or implement a retry mechanism that defers incremental cycles until a successful full backup is confirmed to exist.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,187 @@
...
+		k delete pvc "wal-$POD_PREFIX-$i" "data-$POD_PREFIX-$i" "cold-cache-$POD_PREFIX-$i" \
+			--ignore-not-found --wait=true --timeout=120s || return 1
+	done
+}
+
+do_one_restore() {
+	# Incrementals build on the initial full backup; if that never succeeded
+	# (or a cycle-time check finds it missing), (re)take it here — the driver
+	# is quiesced, so a cycle-time full backup is as valid a base as any.
+	if [ "$FULL_BACKUP_OK" != 1 ]; then
```
NumaryBot, line 190: 🟠 **[major] Retry path can produce an empty incremental delta, reporting false RebuildDelta coverage**

When the startup full backup failed and is retried at the quiesce point, the script immediately runs `store incremental-backup` with no intervening writes. That incremental can be empty, yet the script still proceeds to restore and returns `ok`, recording restore coverage even though RebuildDelta's export replay was not actually exercised.

**Suggestion:** In this retry path, skip or fail the cycle until a subsequent incremental actually exports data, or verify the incremental result (e.g. check that the delta is non-empty) before signalling success.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1593#issuecomment-4980540231

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,205 @@
...
+	log "cluster back up on the restored (RebuildDelta) store"
+}
+
+log "waiting for $STS to reach $REPLICAS ready replicas"
+wait_ready_replicas "$REPLICAS" 1800 || log "WARNING: cluster not ready at startup; continuing"
+
+retry 10 capture_specs || { log "FATAL: cannot capture cluster spec"; exit 1; }
+
+FULL_BACKUP_OK=0
+log "taking initial full backup"
```
NumaryBot, line 238: 🟠 **[major] Exhausted initial full backup treated as warning, breaking all subsequent incremental cycles**

If the startup full backup exhausts all retries (e.g. due to a slow cluster or MinIO startup), the sidecar only logs a warning and later cycles proceed directly to `store incremental-backup`. Incremental backups require a prior full backup to exist; without one, every incremental attempt will fail, no restore cycle will succeed, and RebuildDelta will never be exercised.

**Suggestion:** Treat a fully-exhausted initial full backup as a fatal error (exit non-zero) rather than a warning, or implement a retry mechanism that defers incremental cycles until a successful full backup is confirmed to exist.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,205 @@
...
+		k delete pvc "wal-$POD_PREFIX-$i" "data-$POD_PREFIX-$i" "cold-cache-$POD_PREFIX-$i" \
+			--ignore-not-found --wait=true --timeout=120s || return 1
+	done
+}
+
+do_one_restore() {
+	# Incrementals build on the initial full backup; if that never succeeded
+	# (or a cycle-time check finds it missing), (re)take it here — the driver
+	# is quiesced, so a cycle-time full backup is as valid a base as any.
+	if [ "$FULL_BACKUP_OK" != 1 ]; then
```
NumaryBot, line 190: 🟠 **[major] Retry path can produce an empty incremental delta, reporting false RebuildDelta coverage**

When the startup full backup failed and is retried at the quiesce point, the script immediately runs `store incremental-backup` with no intervening writes. That incremental can be empty, yet the script still proceeds to restore and returns `ok`, recording restore coverage even though RebuildDelta's export replay was not actually exercised.

**Suggestion:** In the retry path, skip or fail the cycle until a subsequent incremental actually exports data, or verify the incremental result (e.g. check that the delta is non-empty) before signalling success.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,205 @@
...
+	rm -f "$REQ" "$RESP"
+	log "restore cycle requested"
+	if do_one_restore; then
+		printf 'ok\n' > "$RESP"
+	else
+		# Converge back to normal mode and wait for the cluster to serve
+		# before releasing the driver, so it never resumes against a cluster
+		# parked in restore mode or still forming; the driver logs the err
+		# and continues.
+		k apply -f "$NORMAL_SPEC" 2>/dev/null || true
```
NumaryBot: 🟠 **[major] Driver resumed after post-teardown restore failure against empty or partially-restored cluster**

If `do_one_restore` has already completed `teardown_cluster` (destroying the PVCs containing the driver's current state) but then fails before a successful `restore finalize`, the error path recreates the normal cluster and writes `err` to the pipe, allowing the driver to resume against an empty or only partially-restored cluster. This turns an orchestration failure into spurious model divergence rather than a clearly-flagged restore error.

**Suggestion:** Mirror the success path's pod-delete-and-wait logic in the error recovery path so the cluster is fully operational before releasing the driver, or keep retrying/blocking rather than releasing the driver when storage has already been destroyed.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1593#issuecomment-4980540231

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,229 @@
...
+k() { kubectl -n "$NS" "$@"; }
+
+# exec_ledgerctl POD ARGS... -- runs ledgerctl inside a ledger pod with the
+# pod-local server address and cluster credentials (the operator-e2e idiom).
+# Bounded by coreutils timeout rather than ledgerctl's --timeout: not every
+# subcommand defines that flag (restore download does not), and a uniform
+# client-side bound also covers a hung exec stream.
+exec_ledgerctl() {
+	local pod="$1"; shift
+	k exec "$pod" -c ledger -- /bin/sh -c \
```
NumaryBot, line 60: 🔴 **[blocker] Unsupported `--timeout` flag passed to `ledgerctl restore download`**

`ledgerctl restore download` does not accept a `--timeout` flag; the CLI uses `--rpc-timeout` and/or `--poll-interval` instead. Cobra rejects the unknown flag at runtime, causing every restore cycle to fail before `RebuildDelta` is ever exercised. This defeats the stated purpose of the PR and has been flagged across multiple prior review rounds without resolution.

**Suggestion:** Replace `--timeout 120s` with the correct flags accepted by `ledgerctl restore download` (e.g. `--rpc-timeout` and/or `--poll-interval`). Verify the exact flag names against the CLI definition before merging.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,229 @@
...
+		FULL_BACKUP_OK=1
+	fi
+
+	retry 3 backup_exec "store incremental-backup" || { log "incremental backup failed"; return 1; }
+
+	# RebuildDelta only runs when the manifest carries exports; a cycle whose
+	# incremental exported nothing (e.g. taken right after a cycle-time full
+	# backup) would restore without exercising the replay. Fail it rather than
+	# record hollow coverage — the local runner has the same guard.
+	local exports
```
NumaryBot, line 201: 🟠 **[major] Single-line grep fails to parse multi-line pretty-printed incremental-backup JSON**

`ledgerctl store incremental-backup --json` emits pretty-printed JSON via `json.MarshalIndent`, spanning multiple lines. The `grep -o '{.*}'` pattern only matches single-line JSON objects, so `exports` is always empty even after a successful non-empty incremental backup. In k8s restore runs this causes every cycle to return `err` before teardown/restore, meaning the operator restore/RebuildDelta path is never exercised.

**Suggestion:** Replace the single-line grep with a proper multi-line JSON parser (e.g. use `jq` to extract the relevant field: `ledgerctl store incremental-backup --json | jq -r '.exports'`). Validate the resulting value for non-emptiness before proceeding.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,229 @@
...
+	log "cluster back up on the restored (RebuildDelta) store"
+}
+
+log "waiting for $STS to reach $REPLICAS ready replicas"
+wait_ready_replicas "$REPLICAS" 1800 || log "WARNING: cluster not ready at startup; continuing"
+
+retry 10 capture_specs || { log "FATAL: cannot capture cluster spec"; exit 1; }
+
+FULL_BACKUP_OK=0
+log "taking initial full backup"
```
NumaryBot, line 238: 🟠 **[major] Exhausted initial full backup treated as warning, breaking all subsequent incremental cycles**

If the startup full backup exhausts all retries (e.g. due to a slow cluster or MinIO startup), the sidecar only logs a warning and later cycles proceed directly to `store incremental-backup`. Incremental backups require a prior full backup to exist; without one, every incremental attempt will fail, no restore cycle will succeed, and `RebuildDelta` will never be exercised.

**Suggestion:** Treat a fully-exhausted initial full backup as a fatal error (exit non-zero) rather than a warning, or implement a retry mechanism that defers incremental cycles until a successful full backup is confirmed to exist.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,229 @@
...
+# reports err and the driver carries on. From the teardown onward the driver's
+# committed state exists ONLY in the backup — releasing the driver against a
+# fresh empty cluster would make every model check diverge — so the restore is
+# retried until it lands, however long that takes (the run's duration bounds
+# it; the driver's lease expiry just parks it on transient retries meanwhile).
+do_one_restore() {
+	# Incrementals build on the initial full backup; if that never succeeded
+	# (or a cycle-time check finds it missing), (re)take it here — the driver
+	# is quiesced, so a cycle-time full backup is as valid a base as any.
+	if [ "$FULL_BACKUP_OK" != 1 ]; then
```
NumaryBot, line 190: 🟠 **[major] Retry path can produce an empty incremental delta, reporting false RebuildDelta coverage**

When the startup full backup failed and is retried at the quiesce point, the script immediately runs `store incremental-backup` with no intervening writes. That incremental can be empty, yet the script still proceeds to restore and returns `ok`, recording restore coverage even though `RebuildDelta`'s export replay was not actually exercised.

**Suggestion:** In the retry path, skip or fail the cycle until a subsequent incremental actually exports data, or verify the incremental result (e.g. check that the delta is non-empty) before signalling success.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,229 @@
...
+		k delete cluster "$CLUSTER" --wait=true --timeout=120s 2>/dev/null || true
+		local i
+		for i in $(seq 0 $(( REPLICAS - 1 ))); do
+			k delete pvc "wal-$POD_PREFIX-$i" "data-$POD_PREFIX-$i" "cold-cache-$POD_PREFIX-$i" \
+				--ignore-not-found --wait=true --timeout=120s 2>/dev/null || true
+		done
+		sleep 10
+	done
+	log "cluster back up on the restored (RebuildDelta) store"
+}
```
NumaryBot, line 230: 🟠 **[major] Driver resumed after post-teardown restore failure against empty or partially-restored cluster**

If `do_one_restore` has already completed `teardown_cluster` (destroying the PVCs containing the driver's current state) but then fails before a successful `restore finalize`, the error path recreates the normal cluster and writes `err` to the pipe, allowing the driver to resume against an empty or only partially-restored cluster. This turns an orchestration failure into spurious model divergence rather than a clearly-flagged restore error.

**Suggestion:** Mirror the success path's pod-delete-and-wait logic in the error recovery path so the cluster is fully operational before releasing the driver, or keep retrying/blocking rather than releasing the driver when storage has already been destroyed.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1593#issuecomment-4980540231

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,249 @@
...
+		FULL_BACKUP_OK=1
+	fi
+
+	retry 3 backup_exec "store incremental-backup" || { log "incremental backup failed"; return 1; }
+
+	# RebuildDelta only runs when the manifest carries exports; a cycle whose
+	# incremental exported nothing (e.g. taken right after a cycle-time full
+	# backup) would restore without exercising the replay. Fail it rather than
+	# record hollow coverage — the local runner has the same guard.
+	local exports
```
NumaryBot, line 201: 🔴 **[blocker] Single-line grep fails to parse multi-line pretty-printed incremental-backup JSON**

`ledgerctl store incremental-backup --json` emits pretty-printed JSON via `json.MarshalIndent`, spanning multiple lines. The `grep -o '{.*}'` pattern only matches single-line JSON objects, so `exports` is always empty even after a successful non-empty incremental backup. In k8s restore runs this causes every cycle to return `err` before teardown/restore, meaning the operator restore/RebuildDelta path is never exercised.

**Suggestion:** Replace the single-line grep with a proper multi-line JSON parser. For example, use `jq` to extract the relevant field: `ledgerctl store incremental-backup --json | jq -r '.exports'`. Validate the resulting value for non-emptiness before proceeding.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,249 @@
...
+	log "cluster back up on the restored (RebuildDelta) store"
+}
+
+log "waiting for $STS to reach $REPLICAS ready replicas"
+wait_ready_replicas "$REPLICAS" 1800 || log "WARNING: cluster not ready at startup; continuing"
+
+retry 10 capture_specs || { log "FATAL: cannot capture cluster spec"; exit 1; }
+
+FULL_BACKUP_OK=0
+log "taking initial full backup"
```
NumaryBot, line 238: 🟠 **[major] Exhausted initial full backup treated as warning, breaking all subsequent incremental cycles**

If the startup full backup exhausts all retries (e.g. due to a slow cluster or MinIO startup), the sidecar only logs a warning and later cycles proceed directly to `store incremental-backup`. Incremental backups require a prior full backup to exist; without one, every incremental attempt will fail, no restore cycle will succeed, and `RebuildDelta` will never be exercised.

**Suggestion:** Treat a fully-exhausted initial full backup as a fatal error (exit non-zero) rather than a warning, or implement a retry mechanism that defers incremental cycles until a successful full backup is confirmed to exist.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,249 @@
...
+# teardown a failure leaves the running cluster untouched, so the cycle just
+# reports err and the driver carries on. From the teardown onward the driver's
+# committed state exists ONLY in the backup — releasing the driver against a
+# fresh empty cluster would make every model check diverge — so the restore is
+# retried until it lands, however long that takes (the run's duration bounds
+# it; the driver's lease expiry just parks it on transient retries meanwhile).
+do_one_restore() {
+	# Incrementals build on the initial full backup; if that never succeeded
+	# (or a cycle-time check finds it missing), (re)take it here — the driver
+	# is quiesced, so a cycle-time full backup is as valid a base as any.
```
NumaryBot, line 189: 🟠 **[major] Retry path can produce an empty incremental delta, reporting false RebuildDelta coverage**

When the startup full backup failed and is retried at the quiesce point, the script immediately runs `store incremental-backup` with no intervening writes. That incremental can be empty, yet the script still proceeds to restore and returns `ok`, recording restore coverage even though `RebuildDelta`'s export replay was not actually exercised.

**Suggestion:** In the retry path, skip or fail the cycle until a subsequent incremental actually exports data, or verify the incremental result (e.g. check that the delta is non-empty) before signalling success.

NumaryBot (commented):
NumaryBot posted 7 new inline findings.

Summary: https://github.com/formancehq/ledger/pull/1593#issuecomment-4980540231

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,249 @@
...
+		k delete cluster "$CLUSTER" --wait=true --timeout=120s 2>/dev/null || true
+		local i
+		for i in $(seq 0 $(( REPLICAS - 1 ))); do
+			k delete pvc "wal-$POD_PREFIX-$i" "data-$POD_PREFIX-$i" "cold-cache-$POD_PREFIX-$i" \
+				--ignore-not-found --wait=true --timeout=120s 2>/dev/null || true
+		done
+		sleep 10
+	done
+	log "cluster back up on the restored (RebuildDelta) store"
+}
```
NumaryBot, line 230: 🟠 **[major] Driver resumed after post-teardown restore failure against empty or partially-restored cluster**

If `do_one_restore` has already completed `teardown_cluster` (destroying the PVCs containing the driver's current state) but then fails before a successful `restore finalize`, the error path recreates the normal cluster and writes `err` to the pipe, allowing the driver to resume against an empty or only partially-restored cluster. This turns an orchestration failure into spurious model divergence rather than a clearly-flagged restore error.

**Suggestion:** Mirror the success path's pod-delete-and-wait logic in the error recovery path so the cluster is fully operational before releasing the driver, or keep retrying/blocking rather than releasing the driver when storage has already been destroyed.

tests/antithesis/run_model_test.sh
```diff
@@ -129,11 +175,91 @@ cleanup() {
...
+		log "restore: bootstrap failed; relaunching on original data"
+		start_node 0 rejoin; wait_leader 0 "$log_offset"
+		return 1
+	fi
+
+	# Swap the restored store in with a fresh WAL: on boot the RESTORED marker
+	# makes the node self-bootstrap as a single voter from the rebuilt data.
+	rm -rf "${NODE_DIRS[0]}/data" "${NODE_DIRS[0]}/wal"
+	mv "$RESTORED_DATA" "${NODE_DIRS[0]}/data"
+	start_node 0 bootstrap
```
NumaryBot, line 243: 🟠 **[major] Stale log lines can make `wait_leader` return prematurely after a restore**

After a restore, `wait_leader` greps the same server log file that already contains the initial boot's `Became leader` line. The grep therefore returns immediately even if the restored node has failed to start or is not yet ready. The orchestrator writes `ok`, the driver resumes against a dead/not-ready server, and the restore cycle can silently pass without actually validating the restored store's liveness.

**Suggestion:** Before calling `wait_leader` post-restore, truncate or rotate the log file (or record the current byte offset and grep only new output) so that only lines emitted after the restore is initiated are considered as evidence of a successful leader election.

tests/antithesis/run_model_test.sh
NumaryBot, line 130: 🟠 **[major] Restore mode may complete without exercising any backup/restore cycle**

When running with `--restore` and no explicit duration, the script still uses the single-node default timeout (30s). The first restore request is only scheduled after `RESTORE_INTERVAL` plus jitter (45–67s by default). The `--restore` code path can finish successfully without ever performing a backup/restore cycle, producing a false pass.

**Suggestion:** When `--restore` is active and no explicit duration is provided, default the run duration to at least `RESTORE_INTERVAL` + max-jitter + a safety margin (e.g. 120s) so that at least one full backup/restore cycle is guaranteed to execute before the script exits.

tests/antithesis/run_model_test.sh
```diff
@@ -129,11 +175,91 @@ cleanup() {
...
+	log "restore: node back up on restored (RebuildDelta) store"
+}
+
+# service_restore_request handles one pending request, if any. Runs in the main
+# shell (not a subshell) so start_node's SERVER_PIDS update is visible to cleanup.
+service_restore_request() {
+	[ "$RESTORE" = 1 ] && [ -f "$RESTORE_REQ" ] || return 0
+	rm -f "$RESTORE_REQ"
+	log "restore: cycle requested"
+	if do_one_restore; then
```
NumaryBot, line 257: 🟠 **[major] Failed restore cycles not surfaced — restore mode can report PASS despite mid-run failures**

If a restore attempt fails after at least one prior successful cycle, `runRestoreCycle` logs the error and continues. The final report only fails when `RESTORE_CYCLES == 0`. This means `--restore` can report PASS even though one or more backup/restore cycles did not successfully restore the node or validate the rebuilt store.

**Suggestion:** Track any per-cycle error and propagate it to the final exit code, so that a single failing restore cycle causes the overall run to be reported as FAIL regardless of how many other cycles succeeded.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,249 @@
...
+k() { kubectl -n "$NS" "$@"; }
+
+# exec_ledgerctl POD ARGS... -- runs ledgerctl inside a ledger pod with the
+# pod-local server address and cluster credentials (the operator-e2e idiom).
+# Bounded by coreutils timeout rather than ledgerctl's --timeout: not every
+# subcommand defines that flag (restore download does not), and a uniform
+# client-side bound also covers a hung exec stream.
+exec_ledgerctl() {
+	local pod="$1"; shift
+	k exec "$pod" -c ledger -- /bin/sh -c \
```
NumaryBot, line 60: 🔴 **[blocker] Unsupported `--timeout` flag passed to `ledgerctl restore download`**

`ledgerctl restore download` does not accept a `--timeout` flag; the CLI uses `--rpc-timeout` and/or `--poll-interval` instead. Cobra rejects the unknown flag at runtime, causing every restore cycle to fail before `RebuildDelta` is ever exercised. This defeats the stated purpose of the PR and has been raised across multiple prior review rounds without resolution.

**Suggestion:** Replace `--timeout 120s` with the correct flags accepted by `ledgerctl restore download` (e.g. `--rpc-timeout` and/or `--poll-interval`). Verify the exact flag names against the CLI definition before merging.

tests/antithesis/workload/restore-orchestrator.sh
NumaryBot, line 201: 🟠 **[major] Single-line grep fails to parse multi-line pretty-printed incremental-backup JSON**

`ledgerctl store incremental-backup --json` emits pretty-printed JSON via `json.MarshalIndent`, spanning multiple lines. The `grep -o '{.*}'` pattern only matches single-line JSON objects, so `exports` is always empty even after a successful non-empty incremental backup. In Kubernetes restore runs this causes every cycle to return an error before teardown/restore, meaning the operator restore/RebuildDelta path is never exercised.

**Suggestion:** Replace the single-line grep with a proper multi-line JSON parser, e.g. pipe through `jq`: `ledgerctl store incremental-backup --json | jq -r '.exports'`. Validate the resulting value for non-emptiness before proceeding.

tests/antithesis/workload/restore-orchestrator.sh
NumaryBot, line 238: 🟠 **[major] Exhausted initial full backup treated as warning, breaking all subsequent incremental cycles**

If the startup full backup exhausts all retries (e.g. due to a slow cluster or MinIO startup), the sidecar only logs a warning and later cycles proceed directly to `store incremental-backup`. Incremental backups require a prior full backup to exist; without one, every incremental attempt will fail, no restore cycle will succeed, and `RebuildDelta` will never be exercised.

**Suggestion:** Treat a fully-exhausted initial full backup as a fatal error (exit non-zero) rather than a warning, or implement a retry mechanism that defers incremental cycles until a successful full backup is confirmed to exist.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,249 @@
...
+# reports err and the driver carries on. From the teardown onward the driver's
+# committed state exists ONLY in the backup — releasing the driver against a
+# fresh empty cluster would make every model check diverge — so the restore is
+# retried until it lands, however long that takes (the run's duration bounds
+# it; the driver's lease expiry just parks it on transient retries meanwhile).
+do_one_restore() {
+	# Incrementals build on the initial full backup; if that never succeeded
+	# (or a cycle-time check finds it missing), (re)take it here — the driver
+	# is quiesced, so a cycle-time full backup is as valid a base as any.
+	if [ "$FULL_BACKUP_OK" != 1 ]; then
```
NumaryBot, line 190: 🟠 **[major] Retry path can produce an empty incremental delta, reporting false RebuildDelta coverage**

When the startup full backup failed and is retried at the quiesce point, the script immediately runs `store incremental-backup` with no intervening writes. That incremental can be empty, yet the script still proceeds to restore and returns success, recording restore coverage even though `RebuildDelta`'s export replay was not actually exercised.

**Suggestion:** In the retry path, skip or fail the cycle until a subsequent incremental actually exports data, or verify the incremental result (e.g. check that the delta is non-empty) before signalling success.

tests/antithesis/workload/restore-orchestrator.sh
NumaryBot, line 230: 🟠 **[major] Driver resumed after post-teardown restore failure against empty or partially-restored cluster**

If `do_one_restore` has already completed `teardown_cluster` (destroying the PVCs containing the driver's current state) but then fails before a successful `restore finalize`, the error path recreates the normal cluster and writes `err` to the pipe, allowing the driver to resume against an empty or only partially-restored cluster. This turns an orchestration failure into spurious model divergence rather than a clearly-flagged restore error.

**Suggestion:** Mirror the success path's pod-delete-and-wait logic in the error recovery path so the cluster is fully operational before releasing the driver, or keep retrying/blocking rather than releasing the driver when storage has already been destroyed.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,249 @@
...
+	case "$exports" in
+	""|0|null)
+		log "incremental produced no exports (RebuildDelta would be a no-op); failing cycle"
+		return 1
+		;;
+	esac
+
+	# Point of no return: the quiesce-point backup above is verified, so the
+	# data survives whatever happens to the cluster from here.
+	teardown_cluster || log "teardown failed; continuing into restore recovery"
```
NumaryBot: 🟠 **[major] Restore proceeds against stale pods/volumes if `teardown_cluster` fails partially**

If `teardown_cluster` encounters a transient kubectl/API error and exits before the Cluster or PVCs are actually removed, the script still proceeds into `restore_from_backup` against whatever old pods or volumes remain. This can make the restore-mode apply run on stale disks rather than a fresh store, so the cycle can report success without forcing the operator through the intended disaster-recovery/RebuildDelta path.

**Suggestion:** Add verification (e.g. wait until the Cluster CR and associated PVCs are fully gone) before proceeding to `restore_from_backup`, and treat a failed or incomplete teardown as a fatal cycle error.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,249 @@
...
+	fi
+
+	retry 3 backup_exec "store incremental-backup" || { log "incremental backup failed"; return 1; }
+
+	# RebuildDelta only runs when the manifest carries exports; a cycle whose
+	# incremental exported nothing (e.g. taken right after a cycle-time full
+	# backup) would restore without exercising the replay. Fail it rather than
+	# record hollow coverage — the local runner has the same guard.
+	local exports
+	exports=$(jq -r '(.segmentsUploaded // 0) + (.logEntriesExported // 0)' "$BACKUP_OUT" 2>/dev/null)
```
NumaryBot, line 202: 🟠 **[major] Non-empty incremental cycles with only audit entries are incorrectly rejected**

The export guard treats a cycle with `auditEntriesExported > 0` but no log entries or uploaded segments as empty. Windows containing only audited failures or idempotency outcomes are valid restore data that `RebuildDelta`/checker coverage should validate; rejecting them as empty causes real restore cycles to be skipped.

**Suggestion:** Include `auditEntriesExported` in the non-empty check so that cycles with only audit entries are not incorrectly discarded.

tests/antithesis/run_model_test.sh
NumaryBot, line 130: 🟠 **[major] Restore mode can complete without exercising any backup/restore cycle**

When running with `--restore` and no explicit duration, the script uses the single-node default timeout (30s). The first restore request is only scheduled after `RESTORE_INTERVAL` plus jitter (45–67s by default). The `--restore` code path can therefore finish successfully without ever performing a backup/restore cycle, producing a false pass.

**Suggestion:** When `--restore` is active and no explicit duration is provided, default the run duration to at least `RESTORE_INTERVAL` + max-jitter + a safety margin (e.g. 120s) so that at least one full backup/restore cycle is guaranteed to execute before the script exits.

tests/antithesis/run_model_test.sh
NumaryBot, line 243: 🟠 **[major] Stale log lines can make `wait_leader` return prematurely after a restore**

After a restore, `wait_leader` greps the same server log file that already contains the initial boot's `Became leader` line. The grep therefore returns immediately even if the restored node has failed to start or is not yet ready. The orchestrator writes `ok`, the driver resumes against a dead/not-ready server, and the restore cycle can silently pass without actually validating the restored store's liveness.

**Suggestion:** Before calling `wait_leader` post-restore, truncate or rotate the log file (or record the current byte offset and grep only new output) so that only lines emitted after the restore is initiated are considered as evidence of a successful leader election.

tests/antithesis/run_model_test.sh
NumaryBot, line 257: 🟠 **[major] Failed restore cycles not surfaced — restore mode can report PASS despite mid-run failures**

If a restore attempt fails after at least one prior successful cycle, `runRestoreCycle` logs the error and continues. The final report only fails when `RESTORE_CYCLES == 0`. This means `--restore` can report PASS even though one or more backup/restore cycles did not successfully restore the node or validate the rebuilt store.

**Suggestion:** Track any per-cycle error and propagate it to the final exit code so that a single failing restore cycle causes the overall run to be reported as FAIL regardless of how many other cycles succeeded.

NumaryBot (commented):
NumaryBot posted 10 new inline findings.

Summary: https://github.com/formancehq/ledger/pull/1593#issuecomment-4980540231

tests/antithesis/workload/restore-orchestrator.sh
NumaryBot, line 60: 🔴 **[blocker] Unsupported `--timeout` flag passed to `ledgerctl restore download`**

`ledgerctl restore download` does not accept a `--timeout` flag; the CLI uses `--rpc-timeout` and/or `--poll-interval` instead. Cobra rejects the unknown flag at runtime, causing every restore cycle to fail before `RebuildDelta` is ever exercised. This defeats the stated purpose of the PR and was raised across multiple prior review rounds without resolution.

**Suggestion:** Replace `--timeout 120s` with the correct flags accepted by `ledgerctl restore download` (e.g. `--rpc-timeout` and/or `--poll-interval`). Verify the exact flag names against the CLI definition before merging.

tests/antithesis/workload/restore-orchestrator.sh
NumaryBot, line 201: 🔴 **[blocker] Single-line grep fails to parse multi-line pretty-printed incremental-backup JSON**

`ledgerctl store incremental-backup --json` emits pretty-printed JSON via `json.MarshalIndent`, spanning multiple lines. The `grep -o '{.*}'` pattern only matches single-line JSON objects, so the `exports` field is always empty even after a successful non-empty incremental backup. In Kubernetes restore runs this causes every cycle to return `err` before teardown/restore, meaning the operator restore/RebuildDelta path is never exercised.

**Suggestion:** Replace the single-line grep with a proper multi-line JSON parser. For example, use `jq` to extract the relevant field: `ledgerctl store incremental-backup --json | jq -r '.exports'`. Validate the resulting value for non-emptiness before proceeding.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1593#issuecomment-4980540231

tests/antithesis/workload/restore-orchestrator.sh
NumaryBot, line 190: 🟠 **[major] Retry path can produce an empty incremental delta, reporting false RebuildDelta coverage**

When the startup full backup failed and is retried at the quiesce point, the script immediately runs `store incremental-backup` with no intervening writes. That incremental can be empty, yet the script still proceeds to restore and returns `ok`, recording restore coverage even though `RebuildDelta`'s export replay was not actually exercised.

**Suggestion:** In the retry path, skip or fail the cycle until a subsequent incremental actually exports data, or verify the incremental result (e.g. check that the delta is non-empty) before signalling success.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/restore.go
```diff
@@ -0,0 +1,191 @@
...
+			if line := strings.TrimSpace(string(data)); line != "" && line != "ok" {
+				return fmt.Errorf("orchestrator reported: %s", line)
+			}
+			return nil
+		}
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-timeout.C:
```
NumaryBot, line 167: 🟠 **[major] Stale restore request file left behind after timeout allows out-of-quiescence restore**

If the orchestrator does not consume the restore request before `MODEL_RESTORE_TIMEOUT` elapses, `Fire` returns and `runRestoreCycle` resumes the workers, but the request file is left behind. A restarted or late sidecar can later pick up that stale request and run backup/restore while the driver is no longer quiesced. The backup may then be cut while new bulks are committing, causing the restored store to fall behind the model state.

**Suggestion:** Before resuming workers (i.e. before or immediately after the timeout branch), explicitly delete or invalidate the stale request file so that a late-arriving sidecar cannot act on it.

flemzord: Confirmed on current head `e01cdee3`, with a broader state-safety impact than just the stale file. On timeout or context cancellation, `Fire` returns and `runRestoreCycle` resumes the workers, but the sidecar may already have consumed the request, deleted the PVCs, and entered its unbounded restore retry loop. Removing the request file cannot cancel that in-flight cycle. The driver must remain quiesced until the orchestrator returns a definitive restored/aborted handshake; otherwise writes resume against an absent or partially restored cluster.

tests/antithesis/workload/restore-orchestrator.sh
flemzord: This remains valid on `e01cdee3`. `teardown_cluster || log ...` still proceeds directly into `restore_from_backup`; if deletion stopped after the CR or only some PVCs, the restore can reuse stale storage and report successful RebuildDelta coverage without proving recovery onto a clean store. Please require and verify complete teardown before crossing into restore.

flemzord (changes requested):
Current-head re-review on `e01cdee3`: this PR has materially changed since my approval at `32b47dd`, so that approval is no longer valid. Two state-safety blockers remain in the existing threads: the driver can resume after its restore lease expires while the sidecar is already past teardown and still retrying recovery, and an incomplete teardown is allowed to continue into restore against potentially stale PVCs. In addition, the Antithesis success property is only registered after a completed cycle, so a zero-cycle run provides no failing coverage signal. I replied on the existing threads rather than duplicating findings. Repository CI is green; NumaryBot remains red.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,252 @@
...
+k() { kubectl -n "$NS" "$@"; }
+
+# exec_ledgerctl POD ARGS... -- runs ledgerctl inside a ledger pod with the
+# pod-local server address and cluster credentials (the operator-e2e idiom).
+# Bounded by coreutils timeout rather than ledgerctl's --timeout: not every
+# subcommand defines that flag (restore download does not), and a uniform
+# client-side bound also covers a hung exec stream.
+exec_ledgerctl() {
+	local pod="$1"; shift
+	k exec "$pod" -c ledger -- /bin/sh -c \
```
NumaryBot, line 60: 🔴 **[blocker] Unsupported `--timeout` flag passed to `ledgerctl restore download`**

`ledgerctl restore download` does not recognise a `--timeout` flag; the CLI uses `--rpc-timeout` and/or `--poll-interval` instead. Cobra rejects the unknown flag at runtime, causing every restore cycle to fail before `RebuildDelta` is ever exercised. This defeats the stated purpose of the PR and was raised across multiple review rounds without resolution.

**Suggestion:** Replace `--timeout 120s` with the correct flags accepted by `ledgerctl restore download` (e.g. `--rpc-timeout` and/or `--poll-interval`). Verify the exact flag names against the CLI definition before merging.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,252 @@
...
+		FULL_BACKUP_OK=1
+	fi
+
+	retry 3 backup_exec "store incremental-backup" || { log "incremental backup failed"; return 1; }
+
+	# RebuildDelta only runs when the manifest carries exports; a cycle whose
+	# incremental exported nothing (e.g. taken right after a cycle-time full
+	# backup) would restore without exercising the replay. Fail it rather than
+	# record hollow coverage — the local runner has the same guard.
+	local exports
```
NumaryBot, line 201: 🔴 **[blocker] Single-line `grep -o '{.*}'` fails to parse multi-line pretty-printed incremental-backup JSON**

`ledgerctl store incremental-backup --json` emits pretty-printed JSON via `json.MarshalIndent`, spanning multiple lines. The `grep -o '{.*}'` pattern only matches single-line JSON objects, so the `exports` field is always empty even after a successful non-empty incremental backup. In Kubernetes restore runs this causes every cycle to return `err` before teardown/restore, meaning the operator restore/RebuildDelta path is never exercised.

**Suggestion:** Replace the single-line grep with a proper multi-line JSON parser. For example: `ledgerctl store incremental-backup --json | jq -r '.exports'`. Validate the resulting value for non-emptiness before proceeding.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,252 @@
...
+	log "cluster back up on the restored (RebuildDelta) store"
+}
+
+log "waiting for $STS to reach $REPLICAS ready replicas"
+wait_ready_replicas "$REPLICAS" 1800 || log "WARNING: cluster not ready at startup; continuing"
+
+retry 10 capture_specs || { log "FATAL: cannot capture cluster spec"; exit 1; }
+
+FULL_BACKUP_OK=0
+log "taking initial full backup"
```
NumaryBot, line 238: 🟠 **[major] Exhausted initial full backup treated as warning, silently breaking all subsequent incremental cycles**

If the startup full backup exhausts all retries (e.g. due to a slow cluster or MinIO startup), the sidecar only logs a warning and proceeds. Subsequent incremental backups require a prior full backup; without one, every incremental attempt fails, no restore cycle succeeds, and `RebuildDelta` is never exercised.

**Suggestion:** Treat a fully-exhausted initial full backup as a fatal error (exit non-zero) rather than a warning, or implement a retry mechanism that defers incremental cycles until a successful full backup is confirmed to exist.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,252 @@
...
+# reports err and the driver carries on. From the teardown onward the driver's
+# committed state exists ONLY in the backup — releasing the driver against a
+# fresh empty cluster would make every model check diverge — so the restore is
+# retried until it lands, however long that takes (the run's duration bounds
+# it; the driver's lease expiry just parks it on transient retries meanwhile).
+do_one_restore() {
+	# Incrementals build on the initial full backup; if that never succeeded
+	# (or a cycle-time check finds it missing), (re)take it here — the driver
+	# is quiesced, so a cycle-time full backup is as valid a base as any.
+	if [ "$FULL_BACKUP_OK" != 1 ]; then
```
NumaryBot, line 190: 🟠 **[major] Retry path can produce an empty incremental delta, reporting false `RebuildDelta` coverage**

When the startup full backup failed and is retried at the quiesce point, the script immediately runs `store incremental-backup` with no intervening writes. That incremental can be empty, yet the script still proceeds to restore and returns `ok`, recording restore coverage even though `RebuildDelta`'s export replay was not actually exercised.

**Suggestion:** In the retry path, skip or fail the cycle until a subsequent incremental actually exports data, or verify the incremental result (e.g. check that the delta is non-empty) before signalling success.

tests/antithesis/workload/restore-orchestrator.sh
```diff
@@ -0,0 +1,252 @@
...
+		k delete cluster "$CLUSTER" --wait=true --timeout=120s 2>/dev/null || true
+		local i
+		for i in $(seq 0 $(( REPLICAS - 1 ))); do
+			k delete pvc "wal-$POD_PREFIX-$i" "data-$POD_PREFIX-$i" "cold-cache-$POD_PREFIX-$i" \
+				--ignore-not-found --wait=true --timeout=120s 2>/dev/null || true
+		done
+		sleep 10
+	done
+	log "cluster back up on the restored (RebuildDelta) store"
+}
```
NumaryBot, line 230: 🟠 **[major] Driver resumed against empty/partially-restored cluster after post-teardown restore failure**

If `do_one_restore` has already completed `teardown_cluster` (destroying the PVCs containing the driver's current state) but then fails before a successful `restore finalize`, the error path recreates the normal cluster and writes `err` to the pipe, allowing the driver to resume against an empty or only partially-restored cluster. This turns an orchestration failure into spurious model divergence rather than a clearly-flagged restore error.

**Suggestion:** Mirror the success path's pod-delete-and-wait logic in the error recovery path so the cluster is fully operational before releasing the driver, or keep retrying/blocking rather than releasing the driver when storage has already been destroyed.

tests/antithesis/run_model_test.sh
```diff
@@ -129,11 +175,94 @@ cleanup() {
...
+# service_restore_request handles one pending request, if any. Runs in the main
+# shell (not a subshell) so start_node's SERVER_PIDS update is visible to cleanup.
+service_restore_request() {
+	[ "$RESTORE" = 1 ] || return 0
+	# Atomic claim, mirroring the driver's withdraw-by-rename on lease expiry:
+	# exactly one side gets the request.
+	mv "$RESTORE_REQ" "$RESTORE_REQ.claimed" 2>/dev/null || return 0
+	rm -f "$RESTORE_REQ.claimed"
+	log "restore: cycle requested"
+	if do_one_restore; then
```
NumaryBot, line 257: 🟠 **[major] Failed restore cycles not propagated to final exit code — restore mode can report PASS despite mid-run failures**

If a restore attempt fails after at least one prior successful cycle, `runRestoreCycle` logs the error and continues. The final report only fails when `RESTORE_CYCLES == 0`. This means `--restore` can report PASS even though one or more backup/restore cycles did not successfully restore the node or validate the rebuilt store.

**Suggestion:** Track any per-cycle error and propagate it to the final exit code, so that a single failing restore cycle causes the overall run to be reported as FAIL regardless of how many other cycles succeeded.

tests/antithesis/run_model_test.sh
```diff
@@ -129,11 +175,94 @@ cleanup() {
...
+		log "restore: bootstrap failed; relaunching on original data"
+		start_node 0 rejoin; wait_leader 0 "$log_offset"
+		return 1
+	fi
+
+	# Swap the restored store in with a fresh WAL: on boot the RESTORED marker
+	# makes the node self-bootstrap as a single voter from the rebuilt data.
+	rm -rf "${NODE_DIRS[0]}/data" "${NODE_DIRS[0]}/wal"
+	mv "$RESTORED_DATA" "${NODE_DIRS[0]}/data"
+	start_node 0 bootstrap
```
NumaryBot, line 243: 🟠 **[major] Stale log lines can cause `wait_leader` to return prematurely after a restore**

After a restore, `wait_leader` greps the same server log file that already contains the initial boot's `Became leader` line. The grep therefore returns immediately even if the restored node has failed to start or is not yet ready. The orchestrator writes `ok`, the driver resumes against a potentially dead or not-ready server, and the restore cycle can silently pass without validating the restored store's liveness.

**Suggestion:** Before calling `wait_leader` post-restore, truncate or rotate the log file (or record the current byte offset and grep only new output) so that only lines emitted after the restore is initiated are considered evidence of a successful leader election.

NumaryBot (commented):
NumaryBot posted 7 new inline findings.

Summary: https://github.com/formancehq/ledger/pull/1593#issuecomment-4980540231

NumaryBot (commented):

Resolved 7 stale NumaryBot review threads (7 fixed, 0 outdated).

Summary: https://github.com/formancehq/ledger/pull/1593#issuecomment-4980540231

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/restore.go
Azorlogh: Fixed in 1a5ff58: the request file is now taken by rename on both sides, so exactly one wins. On lease expiry the driver either withdraws it — proving no cycle ever started, so resuming is safe — or, if the orchestrator already claimed it, stays parked until the cycle answers. It never resumes against an in-flight cycle, including one past the teardown.

tests/antithesis/workload/restore-orchestrator.sh
Azorlogh: Fixed in bf2d0bd: teardown_cluster is now idempotent, verifies every PVC is actually gone before reporting success, and both call sites retry it until it does — the restore is only entered on a verified-complete teardown.

Azorlogh:
@flemzord on the third point from your review (zero-cycle runs giving no failing signal): the instrumentor registers every assertion call site statically, so the Sometimes property exists in the report even when never executed. Empirically, run `5d2b32277a29c92fec628349635d223f-57-8` completed zero restore cycles and reports `failing 0 — singleton_driver_model: restore cycle completed`; the local runner additionally fails explicitly on zero cycles (check #6 in run_model_test.sh). A zero-cycle run is red, not silent.

The two thread findings are addressed at head (1a5ff58, bf2d0bd) — re-requesting your review.

flemzord (approved):
Approved current head `bf2d0bd`. The restore-cycle blockers from my previous review are fixed: the file rendezvous now uses an atomic claim/withdraw so an expired unclaimed request cannot later start a restore after the driver resumes, and the orchestrator now enters restore only after teardown has verifiably removed the old cluster storage. Required checks are green.

flemzord (approved):
Approved current head `bf2d0bd`. The two restore-state blockers from my previous review look addressed: request ownership is now atomic claim/withdraw, so the driver does not resume while an orchestrator cycle is already in flight, and teardown is retried/verified through CR, pod, and PVC removal before restore mode starts. CI is green.

PR #1595, opened 2026-07-16
feat(ledgerctl): add --tls-server-name to verify by name while dialing by IP

gfyrag:
## Problem

On a TLS-enabled cluster, `ledgerctl` is unusable when run **from inside a pod** (e.g. `kubectl exec`). TLS verification matches the `--server` host against the server certificate's SANs, and the client never sets `ServerName` — so it verifies against whatever host is in `--server`. The default `localhost:8888` fails because the operator-issued cert covers only the in-cluster DNS names (`<pod>.<cluster>-headless.<ns>.svc.cluster.local`, `<cluster>.<ns>.svc.cluster.local`) and deliberately **never** `localhost`/`127.0.0.1`.

## Fix

Add a persistent `--tls-server-name` flag that overrides the verification hostname independently of the dial address — the `curl --resolve` / `openssl -servername` equivalent. This lets a client dial by IP/loopback while still validating an FQDN-only certificate.

```sh
ledgerctl --server 127.0.0.1:8888 \
  --tls-server-name "ledger-<cluster>.<ns>.svc.cluster.local" \
  --tls-ca-cert "$TLS_CA_CERT_FILE" ledgers list
```

The inter-node transport already does exactly this (`connection_pool.go` sets `cfg.ServerName`), so this brings `ledgerctl` in line with the existing pattern.

## Details

- Persistent flag `--tls-server-name`, env `LEDGERCTL_TLS_SERVER_NAME`, profile field `tlsServerName` (full CLI > env > profile > default precedence).
- Sets `tls.Config.ServerName` in `GetClientTransportCredentials`.
- Rejected together with `--insecure` (conflicting intent), matching the existing `--tls-ca-cert` guard — also catches a stray `LEDGERCTL_INSECURE` leaking in.
- Wired into `profile create`, `auth login` (not dropped on re-login), and `profile show`.
- Docs: `docs/ops/cli.md` global-flags table, `profile create` table, and a new subsection covering the in-pod use case (dial-FQDN vs dial-by-IP options).

## Not included

No operator change: the operator already dials the in-cluster FQDN (`podSelfServerAddr`, `backupServerAddr`), so it has no mismatch and doesn't need the flag.

## Tests

- `insecure` + `tls-server-name` rejected; TLS-with-server-name succeeds.
- Profile resolution of `tlsServerName` through the shared resolver.
- `go build ./...`, `go vet`, and all `./cmd/ledgerctl/...` tests green.

### DIFF
--- a/cmd/ledgerctl/auth/login.go
+++ b/cmd/ledgerctl/auth/login.go
@@ -187,6 +187,14 @@ func syncProfile(cmd *cobra.Command, server string) error {
 		signingKeyID, _ = cmd.Flags().GetString("key-id")
 	}
 
+	// Reject contradictory TLS flags before bootstrapping the profile — this
+	// path persists without opening a connection, so GetClientTransportCredentials'
+	// guard never runs; without this check we'd store a profile every subsequent
+	// command immediately rejects. --insecure may arrive via LEDGERCTL_INSECURE.
+	if err := cmdutil.ValidateTLSFlags(insecure, tlsCaCert, tlsServerName); err != nil {
+		return err
+	}
+
 	cfg.Profiles[profileName] = cmdutil.Profile{
 		Server:            server,
 		Insecure:          insecure,
--- a/cmd/ledgerctl/auth/login.go
+++ b/cmd/ledgerctl/auth/login.go
@@ -172,6 +172,7 @@ func syncProfile(cmd *cobra.Command, server string) error {
 
 	insecure, _ := cmd.Flags().GetBool("insecure")
 	tlsCaCert, _ := cmd.Flags().GetString("tls-ca-cert")
+	tlsServerName, _ := cmd.Flags().GetString("tls-server-name")
 	signingKey, _ := cmd.Flags().GetString("signing-key")
 	responseVerifyKey, _ := cmd.Flags().GetString("response-verify-key")
 
@@ -190,6 +191,7 @@ func syncProfile(cmd *cobra.Command, server string) error {
 		Server:            server,
 		Insecure:          insecure,
 		TLSCaCert:         tlsCaCert,
+		TLSServerName:     tlsServerName,
 		SigningKey:        signingKey,
 		SigningKeyID:      signingKeyID,
 		ResponseVerifyKey: responseVerifyKey,
--- a/cmd/ledgerctl/profile/create.go
+++ b/cmd/ledgerctl/profile/create.go
@@ -26,6 +26,7 @@ Use --use to activate it immediately even when other profiles already exist.`,
 	cmd.Flags().String("server", "", "gRPC server address (required)")
 	cmd.Flags().Bool("insecure", false, "Use insecure connection (no TLS)")
 	cmd.Flags().String("tls-ca-cert", "", "Path to CA certificate file (PEM)")
+	cmd.Flags().String("tls-server-name", "", "Hostname to verify against the server certificate SANs, overriding the --server host")
 	cmd.Flags().String("signing-key", "", "Path to Ed25519 private key file for request signing")
 	cmd.Flags().String("signing-key-id", "", "Key ID for request signatures")
 	cmd.Flags().String("response-verify-key", "", "Path to Ed25519 seed file for verifying server response signatures")
@@ -52,6 +53,7 @@ func runCreate(cmd *cobra.Command, args []string) error {
 	server, _ := cmd.Flags().GetString("server")
 	insecure, _ := cmd.Flags().GetBool("insecure")
 	tlsCaCert, _ := cmd.Flags().GetString("tls-ca-cert")
+	tlsServerName, _ := cmd.Flags().GetString("tls-server-name")
 	signingKey, _ := cmd.Flags().GetString("signing-key")
 	signingKeyID, _ := cmd.Flags().GetString("signing-key-id")
 	responseVerifyKey, _ := cmd.Flags().GetString("response-verify-key")
@@ -61,6 +63,7 @@ func runCreate(cmd *cobra.Command, args []string) error {
 		Server:            server,
 		Insecure:          insecure,
 		TLSCaCert:         tlsCaCert,
+		TLSServerName:     tlsServerName,
 		SigningKey:        signingKey,
 		SigningKeyID:      signingKeyID,
 		ResponseVerifyKey: responseVerifyKey,
--- a/cmd/ledgerctl/profile/create.go
+++ b/cmd/ledgerctl/profile/create.go
@@ -59,6 +59,12 @@ func runCreate(cmd *cobra.Command, args []string) error {
 	responseVerifyKey, _ := cmd.Flags().GetString("response-verify-key")
 	useFlag, _ := cmd.Flags().GetBool("use")
 
+	// Reject contradictory TLS flags before persisting, so we never store a
+	// profile that every subsequent TLS-aware command would immediately reject.
+	if err := cmdutil.ValidateTLSFlags(insecure, tlsCaCert, tlsServerName); err != nil {
+		return err
+	}
+
 	p := cmdutil.Profile{
 		Server:            server,
 		Insecure:          insecure,
--- a/cmd/ledgerctl/cmdutil/client.go
+++ b/cmd/ledgerctl/cmdutil/client.go
@@ -51,12 +51,8 @@ func GetClientTransportCredentials(cmd *cobra.Command) (credentials.TransportCre
 	caCertPath, _ := cmd.Flags().GetString("tls-ca-cert")
 	serverName, _ := cmd.Flags().GetString("tls-server-name")
 
-	if insecureMode && caCertPath != "" {
-		return nil, errors.New("--insecure and --tls-ca-cert are mutually exclusive (--insecure may also come from the LEDGERCTL_INSECURE env var; unset it to use TLS)")
-	}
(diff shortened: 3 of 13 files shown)


cmd/ledgerctl/profile/create.go
```diff
@@ -26,6 +26,7 @@ Use --use to activate it immediately even when other profiles already exist.`,
 	cmd.Flags().String("server", "", "gRPC server address (required)")
 	cmd.Flags().Bool("insecure", false, "Use insecure connection (no TLS)")
 	cmd.Flags().String("tls-ca-cert", "", "Path to CA certificate file (PEM)")
+	cmd.Flags().String("tls-server-name", "", "Hostname to verify against the server certificate SANs, overriding the --server host")
```
NumaryBot, line 29: 🟠 **[major] Treat tls-server-name as a ledgerctl-owned flag**

Because this new local flag is not added to `ledgerctlOwnedFlagNames`, `bindSubcommandEnv` treats it as an ordinary subcommand flag and binds the bare `TLS_SERVER_NAME` environment variable for `profile create`. In environments where that variable is set, creating a profile will silently persist `tlsServerName` even though the documented/resolved env name is `LEDGERCTL_TLS_SERVER_NAME`, which can make the new profile verify TLS against the wrong name.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1595#issuecomment-4990410682

gfyrag:
Addressed findings from an independent review pass (commit f71f4117d):

- **[Medium] restore ignored TLS flags** — `getRestoreClient` built credentials by hand, reading only `--server`/`--insecure` and silently dropping `--tls-ca-cert` and the new `--tls-server-name`. Now routed through the shared `GetClientTransportCredentials`, so restore honors the same persistent flags and mutual-exclusion guards as every other command (this also fixes a pre-existing `--tls-ca-cert` gap). The restore server is a plausible in-pod dial target, so `--tls-server-name` is exactly relevant there.
- **[Medium] feature line untested** — extracted `buildClientTLSConfig` and the test now asserts `ServerName` is actually applied (the previous `NotNil` check couldn't catch a deleted assignment; `creds.Info().ServerName` is deprecated so the config is asserted directly).
- **[Low] `profile list`** — now surfaces a `TLS SERVER NAME` column, matching how `--tls-ca-cert` is shown.

`go build ./...`, `go vet`, gofmt, and all `./cmd/ledgerctl/...` tests green.

cmd/ledgerctl/profile/create.go
NumaryBot, line 29: 🟠 **[major] tls-server-name not added to ledgerctlOwnedFlagNames**

The new tls-server-name flag is not registered in ledgerctlOwnedFlagNames, so bindSubcommandEnv binds it to the bare TLS_SERVER_NAME environment variable instead of the intended LEDGERCTL_TLS_SERVER_NAME. In environments where TLS_SERVER_NAME is set, profile creation will silently persist the wrong TLS server name, leading to misconfigured profiles.

**Suggestion:** Add "tls-server-name" (or its canonical key) to the ledgerctlOwnedFlagNames list so that bindSubcommandEnv correctly maps it to the LEDGERCTL_TLS_SERVER_NAME environment variable.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1595#issuecomment-4990410682

cmd/ledgerctl/profile/create.go
gfyrag: Fixed in 602fe6e7e — added `tls-server-name` to `ledgerctlOwnedFlagNames` (alongside `tls-ca-cert`) so `bindSubcommandEnv` never binds the bare `TLS_SERVER_NAME`; only `LEDGERCTL_TLS_SERVER_NAME` resolves it. Covered by a new case in TestSubcommandLocalFlagsIgnoreBareEnv.

gfyrag: Fixed in 602fe6e7e — added `tls-server-name` to `ledgerctlOwnedFlagNames` (alongside `tls-ca-cert`) so `bindSubcommandEnv` never binds the bare `TLS_SERVER_NAME`; only `LEDGERCTL_TLS_SERVER_NAME` resolves it. Covered by a new case in TestSubcommandLocalFlagsIgnoreBareEnv.

cmd/ledgerctl/profile/create.go
```diff
@@ -61,6 +63,7 @@ func runCreate(cmd *cobra.Command, args []string) error {
 		Server:            server,
 		Insecure:          insecure,
 		TLSCaCert:         tlsCaCert,
+		TLSServerName:     tlsServerName,
```
NumaryBot, line 72: 🟠 **[major] No mutual-exclusion guard between --insecure and --tls-server-name at profile creation time**

When `profile create --insecure --tls-server-name <name>` is used, both fields are persisted to the profile without validating their incompatibility. The mutual-exclusion guard only runs inside `GetClientTransportCredentials`, which is called when a connection is opened, not when a profile is saved. The result is a profile that every subsequent TLS-aware command immediately rejects, making the profile permanently unusable.

**Suggestion:** Add the same mutual-exclusion check (insecure + tls-server-name conflict) to the profile creation path before saving, so the flag combination is rejected at `profile create` time rather than discovered later at connection time.

cmd/ledgerctl/auth/login.go
```diff
@@ -190,6 +191,7 @@ func syncProfile(cmd *cobra.Command, server string) error {
 		Server:            server,
 		Insecure:          insecure,
 		TLSCaCert:         tlsCaCert,
+		TLSServerName:     tlsServerName,
```
NumaryBot, line 202: 🟠 **[major] auth login bootstrap path saves unusable profile when --insecure and --tls-server-name are combined**

When `auth login --profile <new>` bootstraps a profile, if `--tls-server-name` is combined with `--insecure` (including via the `LEDGERCTL_INSECURE` environment variable), this code path saves the conflicting configuration without opening a client connection — so `GetClientTransportCredentials` is never invoked and its guard never fires. Every subsequent command that loads the profile will fail immediately with the mutual-exclusion error.

**Suggestion:** Extract the insecure/tls-server-name conflict check into a shared validation helper and call it from both the profile-create and auth-login bootstrap paths before any profile is persisted.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1595#issuecomment-4990410682

cmd/ledgerctl/profile/create.go
gfyrag: Fixed in 50a0a9cf4 — extracted `cmdutil.ValidateTLSFlags` (single source of truth) and call it in `profile create` before persisting, so `--insecure` + `--tls-server-name` (and `--tls-ca-cert`) is rejected at create time, not only at connection time. Covered by TestValidateTLSFlags.

cmd/ledgerctl/auth/login.go
gfyrag: Fixed in 50a0a9cf4 — the auth login bootstrap path now calls the shared `cmdutil.ValidateTLSFlags` before persisting the profile, so the conflict (including `--insecure` via `LEDGERCTL_INSECURE`) is rejected before any profile is written. Same helper is used by profile create and the connection path.

cmd/ledgerctl/profile/create.go
NumaryBot, line 29: 🟡 **[minor] LEDGERCTL_TLS_SERVER_NAME not applied when creating a profile via environment variable**

When `profile create` is invoked without an explicit `--tls-server-name` flag but with `LEDGERCTL_TLS_SERVER_NAME` set, the owned-flag exclusion in `bindSubcommandEnv` means the env var is not read for the local flag, and `ResolveConnectionFlags` is not called for profile subcommands. The profile is therefore saved without `tlsServerName`, so subsequent commands that rely on the persisted value fail unless the user repeats `--tls-server-name` explicitly every time.

**Suggestion:** Explicitly read `LEDGERCTL_TLS_SERVER_NAME` (and similarly `LEDGERCTL_TLS_CA_CERT`) in the profile-create path before persisting the profile, either by calling a shared resolver or by directly checking the prefixed env var and populating the flag default.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1595#issuecomment-4990410682

cmd/ledgerctl/profile/create.go
gfyrag: Working as intended — declining this one. `ResolveConnectionFlags` deliberately skips profile-management commands (see resolve.go: *"they define local flags with the same names and must not be contaminated by the active profile or environment variables"*), so `profile create` never inherits any `LEDGERCTL_*` value — not just `LEDGERCTL_TLS_SERVER_NAME` but also `LEDGERCTL_SERVER`, `LEDGERCTL_INSECURE`, `LEDGERCTL_TLS_CA_CERT`, etc. `--tls-server-name` is fully consistent with every other connection flag here. Auto-baking an ambient env var into a newly-created profile would break that contract and risk silent cross-contamination between profiles (an env set for cluster A leaking into a profile meant for cluster B). Users who want env-driven values use them at command time (where `ResolveConnectionFlags` does apply them), not at profile-create time. No change.

paul-nicolas (approved):

PR #1597, opened 2026-07-16
feat(numscript): execute transactions on the bytecode VM

gfyrag:
## Context

Integrates the numscript **compiler + register VM** ([formancehq/numscript#167](https://github.com/formancehq/numscript/pull/167), branch `feat/exp/vm`) into the ledger and routes the **live create-transaction path through the VM** instead of the tree-walking interpreter, so we can benchmark VM vs interpreter under the perf harness.

> ⚠️ Built against a **draft** numscript branch (pseudo-version pinned by SHA). This is a benchmarking branch, not intended to merge as-is.

## What changed

- **Dependency bump** — `github.com/formancehq/numscript` → `feat/exp/vm` (`v0.0.25-…-bf5bbb96`), which exposes `Compile` / `NewVm` / `ExecVm` / `VMStore`. The bump is a **breaking API change** on the interpreter side too, so the existing path had to be adapted.
- **Interpreter path adapted to the new slice-based `Store` API** — `BalanceQuery` / `Balances` / `MetadataQuery` / `AccountsMetadata` are now row slices carrying `color`/`scope`; `ExecutionResult.AccountsMetadata` is a typed `SetAccountMetadataRow` list. Both the store adapter and the dependency-discovery store were ported.
- **New VM path**
  - `numscriptVMStoreAdapter` implements `vm.Store` (lazy per-key `GetBalance` / `GetMetadata` over the coverage-gated `Scope`).
  - `numscriptVMPostingProducer` compiles → encodes vars → `NewVm` → `ExecVm`.
  - `NumscriptCache.GetOrCompile` memoizes the compiled bytecode (fair steady-state comparison, mirroring the memoized parse).
  - `SafeExecVM` wraps `ExecVm` with recover + error mapping (mirror of `SafeRun`).
- **DRY** — `applyNumscriptResult` (postings→volumes) and `numscriptBalance` / `numscriptAccountMetadata` are shared by both producers.
- **Verbatim metadata fix** — the numscript branch changed `Value.String()` to quote strings; `numscriptMetaValueToString` unwraps via numscript's tagged-JSON form so stored metadata stays verbatim.
- **Live wiring** — `processor_transaction.go` now selects the VM producer for all numscript. The interpreter producer is kept (parity test / comparison).

## Known limitation

`ExecVm` does **not yet map** `set_tx_meta` / `set_account_meta` output (numscript TODO). Since the VM is now on the live path, numscript that sets metadata produces **none**. Plain sends (e.g. the `world-to-bank` perf script) are unaffected. Consequently 5 metadata-asserting unit tests are `t.Skip`-ed with an explicit reason. Other caveats: VM missing-funds maps to `ErrNumscriptRuntime` (not `ErrInsufficientFunds`); `GetBalance` `color` is ignored (no ledger equivalent).

## Verification

- `go build ./...`, `go vet ./...` — clean
- `golangci-lint run` on touched packages — 0 issues
- `go test ./internal/domain/processing/...` — green (incl. VM sanity + VM/interpreter parity on an allotment split)


internal/domain/processing/processor_transaction.go
```diff
@@ -55,11 +55,13 @@ func processCreateTransaction(ledger string, order *raftcmdpb.CreateTransactionO
...
 
-	// Select the appropriate posting producer
+	// Select the appropriate posting producer. Numscript scripts run on the
+	// bytecode VM (numscriptVMPostingProducer); the tree-walking interpreter
+	// producer (numscriptPostingProducer) is kept for comparison/benchmarks.
 	var producer postingProducer
 	isNumscript := script != nil && script.GetPlain() != ""
 	if isNumscript {
-		producer = &numscriptPostingProducer{cache: ctx.NumscriptCache, ledgerName: ledger, assetCache: ctx.AssetCache}
+		producer = &numscriptVMPostingProducer{cache: ctx.NumscriptCache, ledgerName: ledger}
```
NumaryBot, line 72: 🔴 **[blocker] Preserve numscript metadata on the live path**

When a live script contains `set_tx_meta` or `set_account_meta`, this now routes it through `numscriptVMPostingProducer`, but the VM result currently leaves `ExecutionResult.Metadata`/`AccountsMetadata` empty. The transaction still commits with its postings, so metadata writes and their key/value validation are silently skipped instead of being persisted or rejected as before.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1597#issuecomment-4991019391

internal/domain/processing/numscript/errors.go
```diff
@@ -82,3 +82,32 @@ func SafeRun(parsed numscriptlib.ParseResult, ctx context.Context, vars numscrip
...
+	store S,
+) (result numscriptlib.ExecutionResult, err domain.Describable) {
+	defer func() {
+		if r := recover(); r != nil {
+			result = numscriptlib.ExecutionResult{}
+			err = &domain.ErrNumscriptRuntime{Detail: fmt.Sprintf("numscript panic: %v", r)}
+		}
+	}()
+
+	result, execErr := numscriptlib.ExecVm(ctx, machine, vars, store)
```
NumaryBot, line 109: 🟠 **[major] Preserve insufficient-funds errors from the VM**

When a VM-executed script overdrafts without `force` or enough overdraft allowance, `ExecVm` returns the VM missing-funds error type, but this conversion path only recognizes the interpreter's `numscriptlib.MissingFundsErr`. Those user-facing failures now become `NUMSCRIPT_RUNTIME`/internal errors instead of the existing `INSUFFICIENT_FUNDS`/failed-precondition response.

internal/domain/processing/processor_transaction.go
NumaryBot, line 72: 🔴 **[blocker] Numscript metadata silently lost on the live VM path**

When a live script contains `set_tx_meta` or `set_account_meta`, execution is now routed through `numscriptVMPostingProducer`, but the VM result leaves `ExecutionResult.Metadata` and `AccountsMetadata` empty. The transaction still commits with its postings, so metadata writes and their key/value validation are silently skipped rather than being persisted or rejected.

**Suggestion:** Populate `ExecutionResult.Metadata` and `AccountsMetadata` from the VM execution result before returning, mirroring what the interpreter path does. Add test coverage that asserts metadata produced by `set_tx_meta`/`set_account_meta` is present in the committed transaction.

internal/domain/processing/numscript/errors.go
NumaryBot, line 109: 🟠 **[major] Insufficient-funds VM errors not mapped to the established domain error**

When a VM-executed script overdrafts without `force` or sufficient overdraft allowance, `ExecVm` returns the VM's missing-funds error type, but the conversion logic only recognises `numscriptlib.MissingFundsErr` (the interpreter type). VM overdraft failures therefore surface as `NUMSCRIPT_RUNTIME` / internal errors instead of the expected `INSUFFICIENT_FUNDS` / failed-precondition response, breaking existing client contracts.

**Suggestion:** Extend the error-mapping switch/if-chain to also handle the VM-specific missing-funds error type, converting it to the same `INSUFFICIENT_FUNDS` domain error that the interpreter path produces.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1597#issuecomment-4991019391

internal/domain/processing/processor_transaction.go
NumaryBot, line 72: 🔴 **[blocker] Numscript metadata silently lost on the live VM path**

When a live script contains `set_tx_meta` or `set_account_meta`, execution is now routed through `numscriptVMPostingProducer`, but the VM result leaves `ExecutionResult.Metadata` and `AccountsMetadata` empty. The transaction still commits with its postings, so metadata writes and their key/value validation are silently skipped rather than being persisted or rejected. This is a silent data-loss regression.

**Suggestion:** Populate `ExecutionResult.Metadata` and `AccountsMetadata` from the VM execution result before returning, mirroring what the interpreter path does. Add test coverage asserting that metadata produced by `set_tx_meta`/`set_account_meta` is present in the committed transaction.

internal/domain/processing/numscript/errors.go
NumaryBot, line 109: 🟠 **[major] VM insufficient-funds errors not mapped to the established domain error**

When a VM-executed script overdrafts without `force` or sufficient overdraft allowance, `ExecVm` returns the VM's missing-funds error type, but the error-conversion logic only recognises `numscriptlib.MissingFundsErr` (the interpreter type). VM overdraft failures therefore surface as `NUMSCRIPT_RUNTIME`/internal errors instead of the expected `INSUFFICIENT_FUNDS`/failed-precondition response, breaking existing client-facing contracts.

**Suggestion:** Extend the error-mapping switch/if-chain to also handle the VM-specific missing-funds error type, converting it to the same `INSUFFICIENT_FUNDS` domain error that the interpreter path produces.

internal/domain/processing/numscript/cache.go
```diff
@@ -48,23 +61,19 @@ func NewNumscriptCache(maxSize int) *NumscriptCache {
...
-
-	h.Sum(result[:0])
-
-	return result
+// HashScript computes the xxh3 128-bit hash of the script content, used as the
+// cache key. xxh3 is a fast non-cryptographic hash; the cache key needs no
+// cryptographic collision resistance (the audit hash chain lives elsewhere),
+// and 128 bits makes accidental collisions negligible. Lock-free and
+// allocation-free.
+func HashScript(script string) [16]byte {
```
NumaryBot, line 75: 🟠 **[major] Non-cryptographic hash used as sole cache key risks script collision**

Both `GetOrParse` and `GetOrCompileVM` use an xxh3 hash as the sole cache lookup key. If two distinct scripts produce a hash collision, the later script reuses the earlier cached AST/VM program. Whether the collision is hit depends on cache state (warm vs. cold/evicted), making transaction semantics non-deterministic across nodes or restarts. xxh3 is a fast, non-cryptographic hash and is not collision-resistant for adversarial or even coincidental inputs.

**Suggestion:** Either use a collision-resistant key (e.g., the full script text as the map key, or a SHA-256 hash), or verify the cached entry's original script text on every cache hit before returning the cached program.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1597#issuecomment-4991019391

internal/domain/processing/processor_transaction.go
```diff
@@ -61,11 +61,13 @@ func processCreateTransaction(ledger string, order *raftcmdpb.CreateTransactionO
...
```
NumaryBot, line 72: 🔴 **[blocker] Numscript metadata silently dropped on the live VM path**

When a live script contains `set_tx_meta` or `set_account_meta`, execution is now routed through `numscriptVMPostingProducer`, but the VM result does not populate `ExecutionResult.Metadata` or `AccountsMetadata`. The transaction can still commit its postings, so metadata writes and their key/value validation are silently skipped rather than being persisted or rejected — a silent data-loss regression compared to the previous interpreter path.

**Suggestion:** Populate `ExecutionResult.Metadata` and `AccountsMetadata` from the VM execution result before returning, mirroring what the interpreter path does. Add test coverage asserting that metadata produced by `set_tx_meta`/`set_account_meta` is present in the committed transaction.

internal/domain/processing/numscript/cache.go
NumaryBot, line 75: 🟠 **[major] Non-cryptographic xxh3 hash used as sole cache key risks script collision**

Both `GetOrParse` and `GetOrCompileVM` use an xxh3 hash as the sole cache lookup key. If two distinct scripts produce a hash collision, the later script reuses the earlier cached AST/VM program. Whether the collision is hit depends on cache state (warm vs. cold/evicted), making transaction semantics non-deterministic across nodes or restarts. xxh3 is a fast, non-cryptographic hash and is not collision-resistant.

**Suggestion:** Use a collision-resistant key — either the full script text as the map key, or a SHA-256/SHA-512 hash — or verify the cached entry's original script text on every cache hit before returning the cached program.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1597#issuecomment-4991019391

internal/domain/processing/processor_transaction.go
paul-nicolas, line 72: [Critical] Live VM path silently drops `set_tx_meta` / `set_account_meta` — data loss with no error

This line routes **every** numscript create-transaction through `numscriptVMPostingProducer`. Its `produce` calls `numscript.SafeExecVM` → `numscriptlib.ExecVm`, and in the pinned library `ExecVm` returns **only postings**:

```go
// numscript.go (v0.0.25-…-bf5bbb96)
// TODO map VM tx/account metadata (stringified) onto the typed contract
return ExecutionResult{Postings: res.Postings}, nil
```

`vm.Exec` actually computes `txMeta`/`accountsMeta` internally, but the wrapper discards them, so `result.Metadata` and `result.AccountsMetadata` are always empty. `applyNumscriptResult` then guards on `len(...) > 0` and skips the metadata blocks. Net effect: a script calling `set_tx_meta`/`set_account_meta` commits a **successful** transaction with the metadata silently missing — no error, no signal. This is squarely the situation CLAUDE.md invariant #7 forbids on the apply path, and it is a data-integrity regression versus the interpreter.

Fix: do not put the VM on the live path until `ExecVm` maps metadata; or make the VM producer fail loudly (`ErrNumscriptRuntime`) when the compiled program contains metadata opcodes, so such scripts are rejected rather than silently truncated. Re-enable the skipped metadata tests before this goes live.

internal/domain/processing/numscript/errors.go
```diff
@@ -82,3 +82,32 @@ func SafeRun(parsed numscriptlib.ParseResult, ctx context.Context, vars numscrip
...
+	err = convertNumscriptError(execErr)
```
paul-nicolas, line 110: [Major] Insufficient funds now surfaces as an internal runtime error instead of a business precondition error

`convertNumscriptError` only matches `numscriptlib.MissingFundsErr` (the interpreter's type, `interpreter.MissingFundsErr`). On the VM path, missing funds is raised as `internal/vm.MissingFundsError` / `NegativeBalanceError` — distinct types that are **not exported at the library root**, so `errors.As` misses and the error falls through to `&domain.ErrNumscriptRuntime{}` (KindInternal).

Result on the now-live path: a routine insufficient-funds outcome maps to an internal error kind instead of `domain.ErrInsufficientFunds` (INSUFFICIENT_FUNDS → KindPrecondition). Clients can no longer distinguish "not enough money" from "script/engine bug" — a contract regression for one of the most common ledger outcomes. It is also untested: the only non-force numscript funds tests exercise the interpreter/raw-posting paths.

Fix: detect the VM missing-funds error in `convertNumscriptError` and map it to `ErrInsufficientFunds` (push numscript to export the VM error, or match on its structured fields), and add a non-force numscript insufficient-funds test asserting `INSUFFICIENT_FUNDS`.

internal/domain/processing/numscript/cache.go
```diff
@@ -131,6 +140,75 @@ func (c *NumscriptCache) GetOrParse(script string) (numscriptlib.ParseResult, do
...
+
+	var (
+		compileErr domain.Describable
+		machine    *numscriptlib.Vm
+	)
+
+	if err != nil {
+		compileErr = &domain.ErrNumscriptParse{Details: err.Error()}
+	} else {
+		machine = numscriptlib.NewVm(prog)
```
paul-nicolas, line 189: [Major] Reused VM is execution-history-dependent state on the deterministic FSM path; determinism is unproven (invariant #2)

`GetOrCompileVM` builds one `*numscriptlib.Vm` per script and reuses it across every transaction. The `Vm` holds mutable register banks (`stringsRegs/intsRegs/portionsRegs/monetariesRegs [256]…`). On each run `vm.Exec` calls `vm.runstate.Reset(...)` which clears only the `RunState` (balances/sources/postings/currentAsset) — the **register banks are never cleared**.

So a freshly-booted node (fresh `NewVm`, zeroed regs) and a long-running node (reused VM carrying the previous transaction's register contents) run the same command against different starting register state. Output is identical only if numscript's codegen guarantees write-before-read for every register on every control-flow path — an unenforced, undocumented property of an experimental (`feat/exp/vm`) library. If it ever fails, two nodes diverge for the same applied index → cache/FSM divergence (invariants #1/#2). The burden of proving determinism is on this change, and there is no test that fresh-VM and reused-VM produce identical results.

Fix: until the library documents/guarantees register hygiene, build a fresh `NewVm(prog)` per apply (keep the compiled program + encoder memoized — those are immutable). Alternatively, add a differential determinism test (see the test-file comment).

internal/domain/processing/processor_transaction_numscript_vm_test.go
```diff
@@ -0,0 +1,135 @@
...
+	require.Equal(t, "users:alice", result.Postings[0].GetDestination())
+	require.Equal(t, "USD/2", result.Postings[0].GetAsset())
+	require.Equal(t, int64(10), result.Postings[0].GetAmount().ToBigInt().Int64())
+}
+
+// TestNumscriptVMProducer_ReusesVMAcrossCalls guards the cached-machine
+// optimization: two transactions run through the same producer/cache reuse the
+// memoized *Vm, and numscript's runstate.Reset must clear the prior run's
+// postings so the second result is not polluted by the first.
+func TestNumscriptVMProducer_ReusesVMAcrossCalls(t *testing.T) {
```
paul-nicolas, line 50: [Medium] `ReusesVMAcrossCalls` does not actually guard the determinism claim it documents

The test runs the **same** script twice and asserts `Len==1` + amount. That catches gross posting accumulation (a failed reset giving `Len==2`), but it never (a) compares a reused-VM result against a fresh-VM result, nor (b) runs a **different** script/vars through the memoized `*Vm`. So it cannot detect stale register-bank leakage between differing runs — which is exactly the determinism risk flagged on `cache.go:189`. `TestNumscriptVMProducer_MatchesInterpreter` also allocates a fresh cache per `run()`, so it doesn't exercise reuse either.

Fix: add a case that runs script A then a different script B (different postings/vars) through the same producer+cache and asserts B is independent of A; ideally also assert the reused-VM output equals a fresh-VM run for identical inputs.

internal/domain/processing/processor_transaction_test.go
```diff
@@ -709,6 +709,7 @@ func TestProcessCreateTransaction_Numscript_SendToMultipleDestinations(t *testin
 
 func TestProcessCreateTransaction_Numscript_SetTxMeta(t *testing.T) {
 	t.Parallel()
+	t.Skip("live numscript path now runs on the bytecode VM, which does not yet map set_tx_meta/set_account_meta output (numscript feat/exp/vm ExecVm TODO)")
```
paul-nicolas: [Medium] Five skipped tests leave the live-path metadata behavior — including the #322 security guard — untested

These `t.Skip`s (712, 781, 828, 909, 998) remove the only coverage of numscript metadata on the create path. Two of them (`RejectsEmptyMetadataKey`, `RejectsNullByteMetadataValue`) guard the #322 fix (empty/NUL keys corrupting read-index entries). Because the VM now drops metadata entirely, `domain.ValidateMetadataKey`/`ValidateMetadataString` inside `applyNumscriptResult` are unreachable on the live path, and the only producer that still reaches them (`numscriptPostingProducer`) is exercised solely by `MatchesInterpreter`, whose script sets no metadata — so the #322 guard is now dead **and** untested. The metadata-drop behavior change itself is asserted by no running test (it lives only in the skip strings).

Fix: add a running test pinning the current "metadata dropped" behavior (so the regression is intentional and visible), and keep a test that still drives the validation guards via the interpreter producer so #322 isn't silently un-covered.

paul-nicolas (commented):
## Multi-model PR review (Claude + Codex)

Both reviewers independently converged on a metadata-integrity regression and an error-contract regression introduced by routing the **live** create-transaction path through the numscript bytecode VM. Findings were verified against the current PR head (`1d34505`) and the pinned numscript library source.

**Severity counts (posted inline):** 1 Critical, 2 Major, 2 Medium.

### Verification outcome
5 Critical/Major/Medium findings entered adversarial verification against the actual code; all 5 confirmed `valid`. One Codex finding was **rejected**: a claimed "VM producer drops the stale-inputs re-resolution guard (`inputsResolutionHash`)" — that guard does not exist on this PR branch (it belongs to a different, unrelated branch that briefly contaminated the review checkout), so `numscriptVMPostingProducer` drops nothing. Dropped, not posted.

### Note
The PR description itself states this is a **benchmarking branch, not intended to merge as-is** (VM on a draft `feat/exp/vm` dependency). The findings below are what must be resolved before any of this reaches a release path; several are already acknowledged as known limitations in the description.

- `internal/domain/processing/processor_transaction_numscript.go:237` — [Minor] `numscriptMetaValueToString` is correct but now dead on the live path (VM emits no metadata), and silently falls back to the quoted form if numscript's tagged-JSON shape ever changes; consider a guard/test if it stays.
- `internal/domain/processing/processor_transaction_numscript_vm.go:26` — [Nit] `GetBalance` color arg is ignored; fine for single-color volumes, but worth an assert/comment if numscript ever relies on color.
- `go.mod:12` — [Nit] numscript pinned to a draft-branch pseudo-version (`bf5bbb96`) on the live path; acknowledged in the PR description.

internal/domain/processing/processor_transaction.go
```diff
@@ -61,12 +61,18 @@ func processCreateTransaction(ledger string, order *raftcmdpb.CreateTransactionO
...
+	// (numscriptVMPostingProducer) or the tree-walking interpreter
+	// (numscriptPostingProducer).
 	var producer postingProducer
 	isNumscript := script != nil && script.GetPlain() != ""
-	if isNumscript {
-		producer = &numscriptPostingProducer{cache: ctx.NumscriptCache, ledgerName: ledger, assetCache: ctx.AssetCache}
-	} else {
+	switch {
+	case isNumscript && ctx.NumscriptUseVM:
+		producer = &numscriptVMPostingProducer{cache: ctx.NumscriptCache, ledgerName: ledger}
```
NumaryBot, line 72: 🔴 **[blocker] Metadata from set_tx_meta / set_account_meta silently dropped on VM path**
<sub>reported by NumaryBot, codex, paul-nicolas</sub>

When a live script contains `set_tx_meta` or `set_account_meta`, execution is routed through `numscriptVMPostingProducer`, but the underlying `ExecVm` wrapper discards the metadata fields computed by `vm.Exec` (returning only postings). `ExecutionResult.Metadata` and `AccountsMetadata` are therefore always empty, so `applyNumscriptResult` skips the metadata-persist and validation blocks entirely. The transaction commits successfully with its postings but with no metadata — a silent data-loss regression. The `ValidateMetadataKey` / `ValidateMetadataString` guards added for issue #322 are also unreachable on this path.

**Suggestion:** Populate `ExecutionResult.Metadata` and `AccountsMetadata` from the VM execution result before returning, mirroring what the interpreter path does. Until `ExecVm` propagates metadata, either keep the VM off the live path or make the producer return an explicit error when the script contains metadata-setting instructions. Add test coverage asserting that metadata produced by `set_tx_meta`/`set_account_meta` is present in the committed transaction, and restore or adapt the currently-skipped metadata-validation tests.

internal/domain/processing/processor_transaction.go
```diff
@@ -61,12 +61,18 @@ func processCreateTransaction(ledger string, order *raftcmdpb.CreateTransactionO
...
+	// Select the appropriate posting producer. Numscript scripts run on the
+	// engine chosen by cluster config (ctx.NumscriptUseVM): the bytecode VM
+	// (numscriptVMPostingProducer) or the tree-walking interpreter
+	// (numscriptPostingProducer).
 	var producer postingProducer
 	isNumscript := script != nil && script.GetPlain() != ""
-	if isNumscript {
-		producer = &numscriptPostingProducer{cache: ctx.NumscriptCache, ledgerName: ledger, assetCache: ctx.AssetCache}
-	} else {
+	switch {
```
NumaryBot, line 70: 🔴 **[blocker] Engine selection is node-local and can violate FSM determinism in a mixed cluster**

The `numscript-engine` flag is evaluated locally on each node to decide whether to execute a Raft-applied transaction through the interpreter or the VM. This flag is not part of the replicated Raft order and is not validated or persisted with startup configuration. In a partially-rolled-out or misconfigured cluster, different nodes apply the same Raft index through different engines. Since the PR itself documents that the VM and interpreter already diverge for metadata-setting scripts, this can produce different resulting state for the same log entry, silently violating FSM determinism.

**Suggestion:** Validate and replicate the engine selection as part of the committed Raft configuration, or gate the feature flag so it is rejected at startup unless all nodes agree on the engine. At minimum, fail fast (with a logged fatal or startup error) if the flag differs from what the FSM state records as having been used previously.

internal/domain/processing/numscript/errors.go
NumaryBot, line 109: 🟠 **[major] VM insufficient-funds errors not mapped to the INSUFFICIENT_FUNDS domain error**
<sub>reported by NumaryBot, paul-nicolas</sub>

The `convertNumscriptError` function only matches `numscriptlib.MissingFundsErr` (the interpreter's type). On the VM path, missing-funds conditions are raised as distinct VM-internal types that are not exported at the library root, so `errors.As` misses them and they fall through to `ErrNumscriptRuntime` (KindInternal). A routine 'not enough balance' outcome now surfaces as an internal/runtime error instead of the established `INSUFFICIENT_FUNDS` / KindPrecondition domain error, breaking existing client contracts and making it impossible to distinguish a balance shortfall from a genuine engine bug.

**Suggestion:** Extend the error-mapping logic to also detect the VM-specific missing-funds error type and convert it to the same `ErrInsufficientFunds` domain error the interpreter path produces. Push the numscript library to export the VM error type, or match on its structured fields. Add a test that drives a non-force numscript script to an insufficient-funds outcome via the VM path and asserts the `INSUFFICIENT_FUNDS` error code.

internal/domain/processing/numscript/cache.go
NumaryBot, line 75: 🟠 **[major] Non-cryptographic xxh3 hash used as sole cache key risks script collision**
<sub>reported by NumaryBot, paul-nicolas</sub>

Both `GetOrParse` and `GetOrCompileVM` use an xxh3 hash as the sole cache lookup key. xxh3 is optimised for speed and is not collision-resistant. A hash collision between two distinct scripts causes the second to silently execute using the first script's compiled AST or VM program. Whether a collision is hit depends on cache warmth, making transaction semantics non-deterministic across nodes and restarts.

**Suggestion:** Use a collision-resistant key — either the full script text as the map key, or a SHA-256/SHA-512 hash — or verify the cached entry's original script text on every cache hit before returning the cached program.

internal/domain/processing/numscript/cache.go
NumaryBot, line 189: 🟠 **[major] Reused VM instance carries uncleared register state across transactions**

`GetOrCompileVM` builds a single `*numscriptlib.Vm` per script and reuses it across every transaction. `vm.Exec` calls `vm.runstate.Reset(...)` which clears only the `RunState`; the underlying register banks (`stringsRegs`, `intsRegs`, `portionsRegs`, `monetariesRegs`) are never cleared. A freshly-booted node (zeroed registers) and a long-running node (registers contain previous-transaction state) therefore execute the same instruction stream against different starting register state. Output is identical only if numscript's codegen guarantees write-before-read for every register on every control-flow path — an unenforced, undocumented property of the experimental VM library. If that property fails, two nodes diverge for the same applied Raft index.

**Suggestion:** Until the library explicitly documents and guarantees that register banks are safe to reuse across invocations, allocate a fresh `*Vm` per transaction (or call the appropriate reset API if one exists). Alternatively, add a test that runs a fresh VM and a reused VM with identical inputs and asserts identical outputs across varied scripts.

internal/domain/processing/processor_transaction_numscript_vm_test.go
NumaryBot, line 50: 🟡 **[minor] ReusesVMAcrossCalls test does not guard against stale register-bank leakage**

The test runs the same script twice and asserts posting count and amount, which catches gross posting accumulation but cannot detect stale register state leaking between different scripts. It never compares a reused-VM result against a fresh-VM result, nor runs a different script through the same cached `*Vm`. The determinism claim the test documents therefore remains unverified.

**Suggestion:** Add a case that runs script A then a structurally different script B through the same producer and cache, and asserts B's result is independent of A. Ideally also assert that the reused-VM output equals a fresh-VM output for identical inputs.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1597#issuecomment-4991019391

internal/domain/processing/processor_transaction.go
NumaryBot, line 72: 🔴 **[blocker] set_tx_meta / set_account_meta metadata silently dropped on VM path**
<sub>reported by NumaryBot, paul-nicolas, codex</sub>

Every live Numscript transaction is now routed through `numscriptVMPostingProducer`. The underlying `ExecVm` wrapper returns only postings — `ExecutionResult.Metadata` and `AccountsMetadata` are always empty. `applyNumscriptResult` guards on `len(...) > 0` and skips all metadata persistence and validation blocks entirely. A script calling `set_tx_meta`/`set_account_meta` therefore commits successfully with its postings but with no metadata, a silent data-loss regression. The `ValidateMetadataKey`/`ValidateMetadataString` guards added for issue #322 are also unreachable on this path.

**Suggestion:** Populate `ExecutionResult.Metadata` and `AccountsMetadata` from the VM execution result before returning, mirroring what the interpreter path does. Until `ExecVm` propagates metadata, either keep the VM off the live path or make the producer return an explicit error when the script contains metadata-setting instructions. Add test coverage asserting that metadata from `set_tx_meta`/`set_account_meta` is present in the committed transaction.

internal/domain/processing/processor_transaction.go
NumaryBot, line 70: 🔴 **[blocker] Engine selection is node-local and can violate FSM determinism in a mixed cluster**
<sub>reported by NumaryBot, paul-nicolas</sub>

The `numscript-engine` flag is evaluated independently on each node and is not part of the replicated Raft order. In a partially-rolled-out or misconfigured cluster, different nodes may apply the same Raft index through different engines. Since the VM and interpreter already diverge on metadata-setting scripts, this can produce different ledger state on different nodes for the same committed log entry, silently violating FSM determinism.

**Suggestion:** Validate and replicate the engine selection as part of committed Raft configuration, or gate the feature flag so it is rejected at startup unless all nodes agree on the engine. At minimum, fail fast with a logged fatal or startup error if the flag differs from what the FSM state records as having been used previously.

internal/domain/processing/numscript/errors.go
NumaryBot, line 109: 🟠 **[major] VM insufficient-funds errors not mapped to the INSUFFICIENT_FUNDS domain error**
<sub>reported by NumaryBot, paul-nicolas</sub>

`convertNumscriptError` only matches `numscriptlib.MissingFundsErr` (the interpreter type). On the VM path, missing-funds conditions are raised as distinct VM-internal types not exported at the library root, so `errors.As` misses them and they fall through to `ErrNumscriptRuntime` (KindInternal). A routine 'not enough balance' outcome now surfaces as an internal/runtime error instead of the established `INSUFFICIENT_FUNDS`/KindPrecondition domain error, breaking existing client contracts and making it impossible to distinguish a balance shortfall from a genuine engine bug.

**Suggestion:** Extend the error-mapping logic to detect the VM-specific missing-funds error type and convert it to the same `ErrInsufficientFunds` domain error the interpreter path produces. Push the numscript library to export the VM error type, or match on its structured fields. Add a test that drives a non-force numscript script to an insufficient-funds outcome via the VM path and asserts the `INSUFFICIENT_FUNDS` error code.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1597#issuecomment-4991019391

internal/domain/processing/numscript/cache.go
NumaryBot, line 189: 🟠 **[major] Reused VM instance carries uncleared register state across transactions**
<sub>reported by NumaryBot, paul-nicolas</sub>

`GetOrCompileVM` builds a single `*numscriptlib.Vm` per script and reuses it across every transaction. `vm.Exec` calls `vm.runstate.Reset(...)` which clears only the `RunState`; the underlying register banks (`stringsRegs`, `intsRegs`, `portionsRegs`, `monetariesRegs`) are never cleared. A freshly-booted node (zeroed registers) and a long-running node (registers contain previous-transaction state) therefore execute the same instruction stream against different starting register state. Output is identical only if numscript's codegen guarantees write-before-read for every register — an unenforced, undocumented property of an experimental library. If that guarantee fails, two nodes diverge for the same applied Raft index.

**Suggestion:** Until the library explicitly documents and guarantees that register banks are safe to reuse across invocations, allocate a fresh `*Vm` per transaction (or call the appropriate reset API if one exists). Add a test that runs a fresh VM and a reused VM with identical inputs and asserts identical outputs across varied scripts.

cmd/server/server.go
```diff
@@ -532,6 +533,10 @@ func LoadConfig(ctx context.Context, cmd *cobra.Command) (*bootstrap.Config, err
 	cfg.SpoolSegmentMaxBytes = bytesize.Get(cmd, "spool-segment-max-bytes").Int64()
 	cfg.BackupMaxSegmentBytes = bytesize.Get(cmd, "backup-max-segment-bytes").Int64()
 	cfg.NumscriptCacheSize = getInt("numscript-cache-size", 1024)
+	// Numscript execution engine. Only "vm" enables the bytecode VM; any other
+	// value (including the default "interpreter") uses the tree-walking
+	// interpreter. Must match on every node (FSM determinism).
+	cfg.NumscriptUseVM = getString("numscript-engine", "interpreter") == "vm"
```
NumaryBot, line 539: 🟠 **[major] Invalid numscript-engine values silently fall back to interpreter**

When `--numscript-engine` / `NUMSCRIPT_ENGINE` is set to an unrecognised value (e.g. `VM`, `vmm`), the code silently falls back to the interpreter instead of failing startup. Because the selected engine affects the FSM apply path, a misconfigured node can diverge from correctly-configured peers, creating an avoidable mixed-engine cluster.

**Suggestion:** Validate the flag value as exactly `interpreter` or `vm` at startup and return a configuration error otherwise, preventing silent misconfiguration.

internal/domain/processing/processor_transaction_numscript_vm_test.go
NumaryBot, line 50: 🟡 **[minor] `ReusesVMAcrossCalls` does not guard the claimed determinism property**

The test runs the same script twice and asserts `Len==1` plus the amount. This catches gross posting accumulation but does not compare a reused-VM result against a fresh-VM result, nor does it run a different script through the memoized `*Vm`. It therefore cannot detect stale register-bank leakage between differing runs, which is the actual determinism risk.

**Suggestion:** Add a case that runs script A then a different script B through the same producer+cache and asserts B is independent of A. Ideally also assert that the reused-VM output equals a fresh-VM run for identical inputs.

flemzord (changes requested):
Reviewed current head `9985a80b`. The interpreter-default switch narrows the earlier metadata-loss wording, but the existing NumaryBot threads on this same head still identify active correctness blockers when VM mode is enabled: engine selection remains node-local on the FSM apply path, VM execution drops metadata, the cache trusts an unchecked xxh3-only key, and a mutable VM instance is reused without clearing its register banks. These can respectively diverge nodes, lose data, execute the wrong script after a collision, or make apply history-dependent. Main CI is also not green (`Tests-Model` and NumaryBot review fail). I am not duplicating the existing inline findings.

flemzord (changes requested):
Revalidated live head `9985a80b`. I am not repeating the existing VM metadata, engine-selection, cache-key, register-reuse, or error-mapping threads. Two additional correctness/runtime blockers remain on this head:

1. The new interpreter adapter aliases Numscript's `(account, asset, color, scope)` slots onto one ledger `(account, asset)` balance: `GetBalances` returns the same full balance for every queried color/scope, then `applyNumscriptResult` drops `Posting.Color`, `SourceScope`, `DestinationScope`, and account-metadata `Scope`. A colored/scoped script can therefore validate the same funds in multiple independent slots and commit aggregate debits or metadata collapsed onto the unscoped key. Rebase/merge resolution must preserve color-keyed volumes and the deterministic rejection of unsupported scopes.

2. The pinned numscript `ec9561f` has an incomplete speculative-`oneof` rollback on the default interpreter path. `SourceWithScaling` calls `ForcePosting` before a leading branch is known to satisfy the amount, while `Snapshot`/`Restore` only marks and repays queued pulls. If that branch falls short, its balance mutations and conversion postings remain before the next branch runs, producing phantom postings/wrong balances. This is also identified upstream on numscript #167 and must be fixed in the pinned dependency (or avoided) before this can be safe.

Main CI is still not green (`Tests-Model`), independently of these blockers.

flemzord (changes requested):
Revalidated live head `9985a80b`. One additional correctness/runtime blocker is not covered by the existing threads: admission dependency discovery still uses the interpreter, while `numscriptVMPostingProducer` only compiles on the FSM apply path. The pinned compiler returns `FeatureNotImplemented` for source/destination `oneof`, scaling, and colors, so scripts that are valid to the existing interpreter and pass admission/preload can fail only after their Raft proposal when `numscript-engine=vm`. Please validate/compile with the selected engine before proposal, or deterministically reject unsupported constructs at admission, and cover those three feature families. The previously reported metadata loss, engine-selection, cache/state, color/scope flattening, rollback, and error-mapping blockers remain unchanged and are not repeated here.

flemzord (changes requested):
Revalidated unchanged head `9985a80b` against the live `release/v3.0` base. One new correctness/runtime blocker has appeared as the base advanced, distinct from the older stale-input claim that was rejected before this protection existed: the base now carries admission-derived `OrderTechnical.inputs_resolution_hash` into `numscriptPostingProducer` and re-resolves it on the coverage-gated FSM path before execution. This head constructs both the interpreter and VM producers without that hash or an equivalent check. During rebase/conflict resolution, VM mode must preserve the same stale-input binding; otherwise mutable balance/metadata inputs can change between admission and apply, invalidating the preloaded dependency set and producing wrong execution or a late coverage failure after Raft proposal. Please pass and enforce the current base guard for every selectable engine, with a VM regression. I am not repeating the existing color/scope, metadata-loss, engine-selection, cache-state, rollback, compilation-timing, or error-mapping blockers.

PR #1602, opened 2026-07-17
fix(processing): stop Numscript set_tx_meta from mutating the accepted order

Azorlogh:
## Context

On `release/v3.0`, the FSM freezes a keyed proposal's idempotency hash **before** `ProcessOrders` (`HashProposal`) but serializes each order into `AuditItem.SerializedOrder` **after** processing.

`processCreateTransaction` aliased the caller-owned order metadata map:

```go
finalMetadata := order.GetMetadata()          // aliases order.Metadata
for key, value := range result.TransactionMetadata {
    if _, exists := finalMetadata[key]; !exists {
        finalMetadata[key] = value            // mutates CreateTransactionOrder.Metadata in place
    }
}
```

When the accepted order carried a non-nil metadata map and the script emitted a distinct `set_tx_meta` key, FSM execution mutated the accepted `CreateTransactionOrder` before audit capture.

## Impact

- The audit chain binds post-execution metadata rather than the accepted business order.
- With an idempotency key, the frozen pre-mutation hash differs from the checker hash re-derived from the audited order → false `CHECK_STORE_ERROR_TYPE_IDEMPOTENCY_MISMATCH`.
- Without a key, the chain stays internally valid but proves the wrong order bytes.

## Fix

Build the effective transaction metadata in a map independent of `order.Metadata`. Output is unchanged: caller metadata still wins collisions, script metadata still fills gaps, nil/empty behavior is preserved, values verbatim.

## Invariant

Documented in `AGENTS.md` (invariant #10): once a request becomes a `raftcmdpb.Order`, neither admission nor FSM processing may mutate its business payload before audit capture — including indirect mutation through maps, slices, byte slices and nested messages. `OrderTechnical` fields (`coverage_bits`, `inputs_resolution_hash`) are outside the rule.

## Tests (both fail before, pass after)

- **processing**: snapshots deterministic order bytes across a `set_tx_meta` run; asserts the order is not mutated and caller-over-script precedence holds.
- **check**: drives a keyed Numscript transaction through the pipeline; asserts the audited order carries only submitted metadata while the log carries the merged metadata, and runs the checker asserting no idempotency mismatch.

## Coordination

EN-1288 also changes Numscript execution in `processor_transaction.go`. This is the small aliasing fix to land first, then rebase EN-1288.


paul-nicolas (approved):

flemzord (approved):
Reviewed current head `655cec19`. The merge now uses a fresh map whenever Numscript contributes transaction metadata, so the accepted order remains immutable while caller metadata still wins collisions. The focused audit/idempotency regression coverage and main CI are green; no remaining correctness blocker found.

PR #1603, opened 2026-07-17
feat(EN-1546): post-commit volumes on every transaction

Azorlogh:
## EN-1546 — Post-commit volumes are part of every transaction

Post-commit volumes become unconditional data on `common.Transaction` instead of an opt-in response field gated by `expandVolumes`. This removes a presentation flag from audit/idempotency input (identical business operations no longer differ because of it) and lets unitary/list/prepared-query reads expose the same snapshot as create/revert.

Built on the colored `(asset, color)` volume shape (EN-1519), which is already on `release/v3.0`.

### Data model
- Add `post_commit_volumes` to `common.Transaction`; remove it from the `CreatedTransaction` / `RevertedTransaction` log wrappers.
- Remove `expandVolumes` from public create/revert payloads, Raft orders, bucket payloads, CLI flags (`--expand-volumes`), the `WithExpandVolumes` helper, and docs. No compat shim.

### FSM
- Computes the snapshot **unconditionally**, right after applying postings and **before** the proposal-level ephemeral purge, on the normal, Numscript, force, bulk, and **mirror** create/revert paths.
- An original transaction keeps its creation-time snapshot after being reverted; the compensating transaction carries its own post-revert snapshot. Never recomputed from current balances.

### Integrity (invariant #8)
- New checker pass `compareTransactionPostCommitVolumes`: verifies every stored `Transaction.post_commit_volumes` against the replayed pre-purge volume state at that transaction's log sequence — run inline in the replay loop before the buffered purge flushes, so it sees exactly the state the FSM captured.
- Missing, extra, duplicate, or modified rows emit `CHECK_STORE_ERROR_TYPE_VOLUME_MISMATCH`.

### Reads
- Get, list, prepared-query, gRPC, and JSON return the immutable historical snapshot from the creation/reversion log. Archive/purge follows transaction availability — no current-state fallback.

### Tests & docs
- Checker unit tests for missing/extra/modified/duplicate/colored/revert rows; e2e specs asserting the snapshot is returned verbatim by get/list and that reverted originals keep their creation-time snapshot; mirror + scenario coverage; antithesis workload + model migrated off `expandVolumes`.
- OpenAPI, api-comparison, CLI docs, sales docs, scenario READMEs updated.

### Verification
- `go build ./...`, `golangci-lint` (0 issues), `gofmt`, `go mod tidy` (no-op)
- Full main-module unit suite; full e2e business suite (550 specs); gaming_wallet + marketplace scenarios; antithesis module vet — all green.


internal/application/check/checker.go
```diff
@@ -1748,6 +1763,127 @@ func (c *Checker) compareTransactions(ctx context.Context, reader dal.PebbleRead
...
+	for _, posting := range tx.GetPostings() {
+		for _, account := range []string{posting.GetSource(), posting.GetDestination()} {
+			k := tupleKey{account: account, asset: posting.GetAsset(), color: posting.GetColor()}
+			if _, seen := expected[k]; seen {
+				continue
+			}
+
+			expected[k] = struct{}{}
+
+			pair, err := replay.GetVolume(domain.NewVolumeKey(ledgerName, account, posting.GetAsset(), posting.GetColor()).Bytes())
```
NumaryBot: 🟠 **[major] Include archived baseline volumes in PCV check**

When the checker runs after chapters have been archived, the replay store only contains post-archive deltas while pre-archive balances live in `baselineDB` and are merged later by `compareVolumes`. This new comparison reads only `replay.GetVolume(...)`, so the first post-archive transaction touching an existing `(account, asset, color)` will compare its stored post-commit snapshot against just the delta (or zero) and emit a false `VOLUME_MISMATCH`. Please seed or consult the baseline for this check as the volume comparison does.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1603#issuecomment-5003293957

internal/application/check/checker.go
```diff
@@ -1748,6 +1773,174 @@ func (c *Checker) compareTransactions(ctx context.Context, reader dal.PebbleRead
...
+	replay *replayStore,
+	baselineVolumes map[string]*raftcmdpb.VolumePair,
+	callback func(*servicepb.CheckStoreEvent),
+) error {
+	var tx *commonpb.Transaction
+
+	switch p := data.GetPayload().(type) {
+	case *commonpb.LedgerLogPayload_CreatedTransaction:
+		tx = p.CreatedTransaction.GetTransaction()
+	case *commonpb.LedgerLogPayload_RevertedTransaction:
```
NumaryBot, line 2064: 🟠 **[major] False VOLUME_MISMATCH on archived stores due to missing baseline in PCV check**

When chapters have been archived, the replay store only contains post-archive deltas while pre-archive balances live in the baseline DB and are merged later during volume comparison. The new post-commit volume check reads only from the replay store (e.g. `replay.GetVolume(...)`), so the first post-archive transaction touching an existing (account, asset, color) compares its stored post-commit snapshot against just a delta (or zero), producing a false VOLUME_MISMATCH on otherwise valid ledgers.

**Suggestion:** Seed the check with, or consult, the baseline volumes — the same way the existing `compareVolumes` helper does — before comparing against the replay store state.

misc/proto/common.proto
```diff
@@ -83,6 +83,10 @@ message Transaction {
   fixed64 reverted_by_transaction = 10;
   // Id of the original transaction this transaction reverts. 0 = not a revert.
   fixed64 reverts_transaction = 11;
+  // Post-commit volumes for every (account, asset, color) tuple touched by
+  // this transaction, captured at the transaction's sequence. Immutable
+  // historical snapshot: never recomputed from current balances.
+  PostCommitVolumes post_commit_volumes = 12;
```
NumaryBot, line 89: 🟠 **[major] No migration/backfill for existing transactions lacking post-commit volumes**

Transactions written before this change do not carry `Transaction.post_commit_volumes` (field 12). After upgrading, reads assemble those old transactions directly from stored logs without the new unconditional field. The new checker now expects `tx.GetPostCommitVolumes()` for every touched posting, so any store with pre-existing data will report VOLUME_MISMATCH during integrity checks, even though the data is perfectly valid.

**Suggestion:** Add a schema migration or backfill pass that reconstructs and persists `post_commit_volumes` for all existing transactions, or add an explicit schema-incompatibility guard that rejects/re-indexes old transaction logs before the new checker path is exercised.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1603#issuecomment-5003293957

misc/proto/raft_cmd.proto
```diff
@@ -376,8 +376,7 @@ message CreateTransactionOrder {
   map<string, common.MetadataValue> metadata = 5;
   map<string, common.MetadataMap> account_metadata = 6;
   bool force = 7;  // Skip balance checks when true
-  bool expand_volumes = 8;  // Include post-commit volumes in response
-  NumscriptReference numscript_reference = 9;
+  NumscriptReference numscript_reference = 8;
```
NumaryBot, line 379: 🟠 **[major] Proto field renumbering breaks deserialization of existing AuditItem records**

When a store already contains persisted `AuditItem.serialized_order` bytes, `numscript_reference` was encoded as field tag 9. Renumbering it to tag 8 (by removing `expand_volumes` which occupied tag 8) causes the new `UnmarshalVT` to silently drop that field, so audit/usage rebuild paths such as `DecodeOrderEffects` and `usagebuilder` treat those historical transactions as non-scripted and miss numscript/template counters. Tag 8 should be reserved (tombstoned) and `numscript_reference` must remain at tag 9.

**Suggestion:** Reserve tag 8 with a `reserved 8;` declaration (or a deprecated placeholder field) and keep `numscript_reference` at its original tag 9 to preserve wire compatibility with already-serialized bytes.

internal/application/check/checker.go
```diff
@@ -1965,6 +1990,174 @@ func (c *Checker) compareTransactions(ctx context.Context, reader dal.PebbleRead
...
+			// Expected = baseline (pre-archive state carried in the checkpoint) +
+			// replayed delta. After archiving the replay store holds only
+			// post-archive deltas, so the baseline supplies the volume accumulated
+			// before the boundary; without archiving baselineVolumes is empty and a
+			// missing replay entry is the zero volume, matching the FSM's
+			// readVolumeOrZero when it built the snapshot.
+			wantInput := big.NewInt(0)
+			wantOutput := big.NewInt(0)
+
+			if base := baselineVolumes[string(keyBytes)]; base != nil {
```
NumaryBot: 🟠 **[major] New PCV checker does not consult baseline volumes, causing false VOLUME_MISMATCH on archived stores**

When chapters have been archived, the replay store only contains post-archive deltas while pre-archive balances live in the baseline DB and are merged later by the existing `compareVolumes` helper. The new post-commit volume check reads only from the replay store (e.g. `replay.GetVolume(...)`), so the first post-archive transaction touching an existing (account, asset, color) compares its stored post-commit snapshot against just a delta (or zero), producing a false VOLUME_MISMATCH on otherwise valid ledgers. This issue was independently identified across multiple reviews.

**Suggestion:** Seed the new PCV check with, or consult, the baseline volumes the same way the existing `compareVolumes` helper does before comparing against the replay store state.

NumaryBot: 🟠 **[major] Baseline tombstones not tracked — stale archived volumes cause false VOLUME_MISMATCH for churned accounts**

When an archived baseline volume is purged after the archive boundary (e.g. an ephemeral account reaches zero) and the same (account, asset, color) is later re-touched, the FSM snapshot starts from the recreated/current value. However, the checker still adds the stale baseline row because there is no tombstone state for baseline-backed deletions. Valid post-commit snapshots therefore compare against pre-purge balances and emit false VOLUME_MISMATCH on archived stores with churned ephemeral/transient volumes.

**Suggestion:** Before adding archived baseline volumes, check whether the account/asset has been explicitly zeroed or deleted after the archive boundary. Maintain tombstone state for baseline-backed deletions so they are excluded from the baseline seed.

misc/proto/common.proto
NumaryBot, line 89: 🟠 **[major] No migration/backfill for pre-existing transactions missing post_commit_volumes**

Transactions written before this change do not carry the `Transaction.post_commit_volumes` field. After upgrading, the checker expects `tx.GetPostCommitVolumes()` for every touched posting. Any store with pre-existing data will report VOLUME_MISMATCH during integrity checks even though the data is perfectly valid, making the new checker unusable on existing deployments without a migration.

**Suggestion:** Add a schema migration or backfill pass that reconstructs and persists `post_commit_volumes` for all existing transactions, or add an explicit schema-version guard that skips the new checker path for transaction logs written before the field was introduced.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1603#issuecomment-5003293957

misc/proto/raft_cmd.proto
NumaryBot, line 379: 🔴 **[blocker] Proto field renumbering breaks deserialization of existing AuditItem/CreateTransactionPayload records**

Removing `expand_volumes` (tag 8) and shifting `numscript_reference`/`script_reference` from tag 9 to tag 8 breaks wire compatibility with all bytes already persisted in Raft logs and any gRPC clients built from the previous proto. The new deserializer silently drops the old field, causing historical audit/numscript metadata to be lost or misinterpreted, and scripted transaction requests to arrive without their script reference.

**Suggestion:** Reserve tag 8 with a `reserved 8;` declaration (or a deprecated placeholder field) and keep `numscript_reference`/`script_reference` at its original tag 9 to preserve wire compatibility with already-serialized bytes.

internal/application/check/checker.go
```diff
@@ -1965,6 +1990,190 @@ func (c *Checker) compareTransactions(ctx context.Context, reader dal.PebbleRead
...
+
+			expected[k] = struct{}{}
+
+			keyBytes := domain.NewVolumeKey(ledgerName, account, posting.GetAsset(), posting.GetColor()).Bytes()
+
+			pair, err := replay.GetVolume(keyBytes)
+			if err != nil {
+				return fmt.Errorf("reading replay volume for post-commit check tx %d: %w", tx.GetId(), err)
+			}
+
```
NumaryBot, line 2116: 🟠 **[major] New PCV checker does not consult baseline volumes — false VOLUME_MISMATCH on archived stores**

When chapters have been archived, the replay store only contains post-archive deltas while pre-archive balances live in the baseline DB. The new post-commit volume check reads only from the replay store (e.g. `replay.GetVolume(...)`), so the first post-archive transaction touching an existing (account, asset, color) compares its stored post-commit snapshot against just a delta (or zero), producing false VOLUME_MISMATCH errors on otherwise valid ledgers.

**Suggestion:** Seed the new PCV check with, or consult, the baseline volumes the same way the existing `compareVolumes` helper does before comparing against the replay store state.

internal/application/check/checker.go
```diff
@@ -1965,6 +1990,190 @@ func (c *Checker) compareTransactions(ctx context.Context, reader dal.PebbleRead
...
+			// readVolumeOrZero when it built the snapshot.
+			//
+			// Skip the baseline when the tuple has already been ephemeral-purged
+			// during replay: the FSM deleted that volume and the re-touch restarts
+			// from zero, so its post-purge snapshot is the delta alone. Adding the
+			// stale pre-purge baseline would over-count and false-positive.
+			wantInput := big.NewInt(0)
+			wantOutput := big.NewInt(0)
+
+			if base := baselineVolumes[string(keyBytes)]; base != nil &&
```
NumaryBot, line 2131: 🟠 **[major] Stale baseline rows for churned accounts cause false VOLUME_MISMATCH**

When an archived baseline volume is purged after the archive boundary (e.g. an ephemeral account reaches zero) and the same (account, asset, color) is later re-touched, the checker still adds the stale baseline row. `SimulateEphemeralPurge` reads only the replay delta without adding `baselineVolumes` when deciding the tuple reached zero, so the tombstone state for baseline-backed deletions is never tracked. Valid post-commit snapshots therefore compare against pre-purge balances and emit false VOLUME_MISMATCH.

**Suggestion:** Before adding archived baseline volumes, check whether the account/asset has been explicitly zeroed or deleted after the archive boundary. Maintain tombstone state for baseline-backed deletions so they are excluded from the baseline seed.

misc/proto/common.proto
NumaryBot, line 89: 🟠 **[major] No migration/backfill for pre-existing transactions missing post_commit_volumes**

Transactions written before this change do not carry `Transaction.post_commit_volumes` (field 12). After upgrading, the checker expects `tx.GetPostCommitVolumes()` for every touched posting, so any store with pre-existing data will report VOLUME_MISMATCH during integrity checks even though the data is perfectly valid. This makes the new checker unusable on existing deployments without a migration.

**Suggestion:** Add a schema migration or backfill pass that reconstructs and persists `post_commit_volumes` for all existing transactions, or add an explicit schema-version guard that skips the new checker path for transaction logs written before the field was introduced.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1603#issuecomment-5003293957

misc/proto/raft_cmd.proto
NumaryBot, line 379: 🔴 **[blocker] Proto field renumbering breaks deserialization of existing AuditItem/CreateTransactionPayload records**

Removing `expand_volumes` (originally tag 8) and allowing `numscript_reference`/`script_reference` to shift from tag 9 to tag 8 breaks wire compatibility with all bytes already persisted in Raft logs and with any gRPC clients built from the previous proto schema. The new deserializer silently drops the old tag-9 encoded field, causing historical audit and numscript metadata to be lost or misinterpreted, and scripted transaction requests to arrive without their script reference.

**Suggestion:** Reserve tag 8 with a `reserved 8;` declaration (or a deprecated placeholder field) and keep `numscript_reference`/`script_reference` at their original tag 9 to preserve wire compatibility with already-serialized bytes.

internal/application/check/checker.go
NumaryBot, line 2116: 🟠 **[major] PCV checker does not consult baseline volumes — false VOLUME_MISMATCH on archived stores**

When chapters have been archived, the replay store only contains post-archive deltas while pre-archive balances live in the baseline DB and are merged later by the existing `compareVolumes` helper. The new post-commit volume check reads only from the replay store (e.g. `replay.GetVolume(...)`), so the first post-archive transaction touching an existing (account, asset, color) compares its stored post-commit snapshot against just a delta (or zero), producing false VOLUME_MISMATCH errors on otherwise valid ledgers.

**Suggestion:** Seed the new PCV check with, or consult, the baseline volumes the same way the existing `compareVolumes` helper does before comparing against the replay store state.

internal/application/check/checker.go
NumaryBot, line 2131: 🟠 **[major] Stale baseline rows for churned accounts cause false VOLUME_MISMATCH**

When an archived baseline volume is purged after the archive boundary (e.g. an ephemeral account reaches zero) and the same (account, asset, color) is later re-touched, the checker still adds the stale baseline row. `SimulateEphemeralPurge` reads only the replay delta without merging `baselineVolumes` when deciding whether a tuple reached zero, so tombstone state for baseline-backed deletions is never tracked. Valid post-commit snapshots therefore compare against pre-purge balances and emit false VOLUME_MISMATCH.

**Suggestion:** Before adding archived baseline volumes, check whether the account/asset has been explicitly zeroed or deleted after the archive boundary. Maintain tombstone state for baseline-backed deletions so they are excluded from the baseline seed.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1603#issuecomment-5003293957

flemzord (approved):
Reviewed current head `e8b65e71`. The archived-baseline and post-purge churn cases are now handled in the PCV checker: replay uses baseline-aware purge boundaries and excludes stale baseline tuples after purge. The remaining NumaryBot complaint is pre-stable proto/history compatibility, which is non-blocking here; Codecov patch is tooling-only, while all main build/test/Schemathesis jobs pass. No remaining runtime blocker found.

PR #1605, opened 2026-07-17
fix(EN-1553): require TLS when --auth-enabled

paul-nicolas:
## Summary

Enabling external-client authentication (`--auth-enabled`, OIDC JWT / Ed25519) over a plaintext transport sends bearer tokens in the clear — an on-path attacker can intercept and replay them. The `--cluster-secret` (inter-node) credential already requires TLS at startup; `--auth-enabled` (external client) did not. This makes the protection symmetric.

`validateAuthConfig()` now rejects the `--auth-enabled` + `--tls-mode=disabled` combination at boot, with a message styled after the existing `--cluster-secret` rule.

## Decision

**Hard error, no opt-out** — mirroring `--cluster-secret` exactly (which has no escape hatch either). Same predicate: only `TLSModeDisabled` is rejected; `optional` and `required` both pass. TLS must be terminated on the ledger process itself, even behind an ingress/mesh that also terminates TLS upstream.

## Changes

- `internal/bootstrap/config.go` — TLS check in `validateAuthConfig()`.
- `internal/bootstrap/config_test.go` — added a `tlsMode` dimension: TLS-disabled rejection, `optional`/`required` acceptance.
- `docs/ops/authentication.md`, `docs/ops/cli.md`, `docs/ops/deployment.md` — documented the invariant; made runnable examples TLS-correct.

## Testing

- `just pre-commit` — 0 lint issues, no generated-file drift.
- `just test` (full `-race` suite) — green.

## Ticket

https://formance-team.atlassian.net/browse/EN-1553

### DIFF
--- a/docs/ops/authentication.md
+++ b/docs/ops/authentication.md
@@ -9,16 +9,23 @@ ledger run \
   --node-id 1 \
   --cluster-id prod-ledger \
   --bootstrap \
+  --tls-mode required \
+  --tls-cert-file /etc/ledger/tls.crt \
+  --tls-key-file /etc/ledger/tls.key \
   --auth-enabled \
   --auth-issuer https://auth.example.com \
   --auth-service ledger
 ```
 
+> `--auth-enabled` requires TLS (`--tls-mode` set to `optional` or `required`);
+> the server refuses to start with `--tls-mode=disabled` so bearer tokens are
+> never sent in plaintext. See [Configuration Invariants](#configuration-invariants).
+
 ## CLI Flags
 
 | Flag | Type | Default | Description |
 |------|------|---------|-------------|
-| `--auth-enabled` | bool | `false` | Enable JWT authentication and scope-based authorization |
+| `--auth-enabled` | bool | `false` | Enable JWT authentication and scope-based authorization. Requires TLS (`--tls-mode` `optional` or `required`) — rejected with `--tls-mode=disabled` |
 | `--auth-issuer` | string | `""` | OIDC issuer URL (used for discovery and token validation) |
 | `--auth-service` | string | `""` | Service name prefix for scopes (e.g., `ledger` for `ledger:read`) |
 
@@ -59,6 +66,13 @@ The server enforces the following rules at startup:
 - `--auth-enabled` requires at least one of `--auth-issuer` (OIDC) or
   `--auth-ed25519-keys` (Ed25519 key file). The server refuses to start
   without a credential source.
+- `--auth-enabled` requires TLS (`--tls-mode` set to `optional` or `required`).
+  Over a plaintext transport the bearer JWT/Ed25519 tokens would be sent in the
+  clear and could be intercepted and replayed by an on-path attacker, so the
+  combination `--tls-mode=disabled --auth-enabled` is rejected at startup. This
+  mirrors the `--cluster-secret` rule, which likewise requires TLS. There is no
+  opt-out; terminate TLS on the ledger process itself even when an ingress or
+  service mesh also terminates TLS upstream.
 - Setting auth-related flags (`--auth-issuer`, `--auth-ed25519-keys`,
   `--auth-scope-mapping-file`) without `--auth-enabled` is rejected to
   prevent operators from believing authentication is active when it is not.
@@ -82,6 +96,7 @@ The writes-only configuration is:
 
 ```bash
 ledger run --auth-enabled --auth-issuer https://auth.example.com \
+  --tls-mode required --tls-cert-file /etc/ledger/tls.crt --tls-key-file /etc/ledger/tls.key \
   --auth-anonymous-scopes "*:read"
 ```
 
--- a/docs/ops/authentication.md
+++ b/docs/ops/authentication.md
@@ -17,15 +17,15 @@ ledger run \
   --auth-service ledger
 ```
 
-> `--auth-enabled` requires TLS (`--tls-mode` set to `optional` or `required`);
-> the server refuses to start with `--tls-mode=disabled` so bearer tokens are
-> never sent in plaintext. See [Configuration Invariants](#configuration-invariants).
+> `--auth-enabled` requires `--tls-mode=required`; the server refuses to start
+> with `--tls-mode=disabled` or `--tls-mode=optional` so bearer tokens are never
+> exposed to plaintext. See [Configuration Invariants](#configuration-invariants).
 
 ## CLI Flags
 
 | Flag | Type | Default | Description |
 |------|------|---------|-------------|
-| `--auth-enabled` | bool | `false` | Enable JWT authentication and scope-based authorization. Requires TLS (`--tls-mode` `optional` or `required`) — rejected with `--tls-mode=disabled` |
+| `--auth-enabled` | bool | `false` | Enable JWT authentication and scope-based authorization. Requires `--tls-mode=required` — rejected with `--tls-mode=disabled` or `--tls-mode=optional` |
 | `--auth-issuer` | string | `""` | OIDC issuer URL (used for discovery and token validation) |
 | `--auth-service` | string | `""` | Service name prefix for scopes (e.g., `ledger` for `ledger:read`) |
 
@@ -66,11 +66,15 @@ The server enforces the following rules at startup:
 - `--auth-enabled` requires at least one of `--auth-issuer` (OIDC) or
   `--auth-ed25519-keys` (Ed25519 key file). The server refuses to start
   without a credential source.
-- `--auth-enabled` requires TLS (`--tls-mode` set to `optional` or `required`).
-  Over a plaintext transport the bearer JWT/Ed25519 tokens would be sent in the
-  clear and could be intercepted and replayed by an on-path attacker, so the
-  combination `--tls-mode=disabled --auth-enabled` is rejected at startup. This
-  mirrors the `--cluster-secret` rule, which likewise requires TLS. There is no
+- `--auth-enabled` requires `--tls-mode=required`. Over a plaintext transport
+  the bearer JWT/Ed25519 tokens would be sent in the clear and could be
+  intercepted and replayed by an on-path attacker, so both
+  `--tls-mode=disabled` and `--tls-mode=optional` are rejected at startup —
+  `optional` runs a dual listener that still accepts plaintext client
+  connections, so it is not sufficient. This is intentionally stricter than the
+  `--cluster-secret` rule (which permits `optional`): the operator drives
+  zero-downtime inter-node TLS migration through the transitional `optional`
+  mode, but the external service API has no such requirement. There is no
   opt-out; terminate TLS on the ledger process itself even when an ingress or
   service mesh also terminates TLS upstream.
 - Setting auth-related flags (`--auth-issuer`, `--auth-ed25519-keys`,
--- a/docs/ops/authentication.md
+++ b/docs/ops/authentication.md
@@ -19,7 +19,10 @@ ledger run \
 
 > `--auth-enabled` requires `--tls-mode=required`; the server refuses to start
 > with `--tls-mode=disabled` or `--tls-mode=optional` so bearer tokens are never
-> exposed to plaintext. See [Configuration Invariants](#configuration-invariants).
+> exposed to plaintext **on the gRPC service transport**. The HTTP REST-compat
+> listener is not covered by `--tls-mode` and must be TLS-terminated separately
+> (HTTPS at the ingress/proxy) when authentication is enabled.
+> See [Configuration Invariants](#configuration-invariants).
 
 ## CLI Flags
 
@@ -76,7 +79,10 @@ The server enforces the following rules at startup:
   zero-downtime inter-node TLS migration through the transitional `optional`
   mode, but the external service API has no such requirement. There is no
   opt-out; terminate TLS on the ledger process itself even when an ingress or
-  service mesh also terminates TLS upstream.
+  service mesh also terminates TLS upstream. Note this guard covers the
+  **gRPC service transport** only — `--tls-mode` does not govern the HTTP
+  REST-compat listener, which remains plaintext and must be protected by
+  separate HTTPS termination (ingress/proxy) when authentication is enabled.
 - Setting auth-related flags (`--auth-issuer`, `--auth-ed25519-keys`,
   `--auth-scope-mapping-file`) without `--auth-enabled` is rejected to
   prevent operators from believing authentication is active when it is not.
--- a/docs/ops/cli.md
+++ b/docs/ops/cli.md
@@ -4449,6 +4449,7 @@ one node's secret drifted from the rest of the cluster.
 
 ```bash
 ledger run --cluster-secret "my-cluster-secret" --auth-enabled \
+  --tls-mode required \
   --tls-cert-file /etc/ledger/tls.crt --tls-key-file /etc/ledger/tls.key \
   [other flags...]
 ```
--- a/docs/ops/cli.md
+++ b/docs/ops/cli.md
@@ -4638,7 +4638,7 @@ Enable JWT/OIDC authentication with scope-based authorization. See [Authenticati
 
 | Flag | Type | Default | Description |
 |------|------|---------|-------------|
-| `--auth-enabled` | bool | `false` | Enable JWT authentication and scope-based authorization. Requires TLS (`--tls-mode` `optional` or `required`) — rejected with `--tls-mode=disabled` |
+| `--auth-enabled` | bool | `false` | Enable JWT authentication and scope-based authorization. Requires `--tls-mode=required` — rejected with `--tls-mode=disabled` or `--tls-mode=optional` (`optional` still accepts plaintext client connections) |
 | `--auth-issuer` | string | `""` | OIDC issuer URL (used for discovery and token validation) |
 | `--auth-service` | string | `""` | Service name prefix for scopes (e.g., `ledger` for `ledger:read`) |
 | `--auth-read-key-set-max-retries` | int | `10` | Maximum retries when fetching the JWKS key set |
--- a/docs/ops/cli.md
+++ b/docs/ops/cli.md
@@ -4515,6 +4515,7 @@ ledger run --auth-scope-mapping-file /etc/ledger/scope-mapping.json [other flags
 
 ```bash
 ledger run --auth-enabled --auth-issuer https://auth.example.com \
+  --tls-mode required --tls-cert-file /etc/ledger/tls.crt --tls-key-file /etc/ledger/tls.key \
   --auth-anonymous-scopes "*:read"
 ```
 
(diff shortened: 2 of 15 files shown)


internal/bootstrap/config.go
```diff
@@ -336,6 +336,13 @@ func (c Config) validateAuthConfig() error {
 			return errors.New("--auth-enabled requires either --auth-issuer (OIDC) or --auth-ed25519-keys (Ed25519)")
 		}
 
+		// Reject auth without TLS — bearer JWT/Ed25519 tokens would travel in
+		// plaintext over the external gRPC API. Mirrors the --cluster-secret rule
+		// in Validate() so both bearer credentials are protected symmetrically.
+		if c.TLSConfig.Mode == TLSModeDisabled {
```
NumaryBot: 🔴 **[blocker] HTTP API listener remains plaintext despite TLS enforcement for gRPC**

When `--auth-enabled` is combined with `--tls-mode=required`, the new validation guards the gRPC path but the HTTP listener is still started without TLS. Since `HTTPAuthMiddleware` accepts bearer tokens over that plaintext HTTP port, the validation gives a false sense of security—credentials can still be intercepted on the unprotected HTTP endpoint.

**Suggestion:** Either wire TLS configuration into the HTTP listener when auth is enabled (e.g. pass a TLS config to `httpserver.NewHook`), disable the plaintext HTTP listener entirely when `--auth-enabled` is set, or extend the validation to fail hard when an unprotected HTTP listener is also active.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1605#issuecomment-5019606962

internal/bootstrap/config.go
paul-nicolas: Thanks — this is a real observation, but EN-1553 is explicitly scoped to authenticated service gRPC (per the ticket: "this ticket is explicitly scoped to authenticated service gRPC"). The gRPC API terminates TLS on-process and that is exactly what this guard protects. The plaintext HTTP REST-compat listener is a separate, pre-existing surface and out of scope for this change. Resolving here.

paul-nicolas:
Fixed in c79b3f877. The auth E2E fixtures (`auth_test.go`, `auth_writes_only_test.go`, and the three in `ed25519_auth_test.go`) now provision a throwaway CA + server cert, start with `--tls-mode=required` + cert/key, and dial gRPC over TLS via the existing `newTLSGRPCClient` helper. This satisfies the new guard and exercises the auth-over-TLS path rather than plaintext.

internal/bootstrap/config.go
NumaryBot: 🔴 **[blocker] HTTP listener remains plaintext while TLS is only enforced on gRPC**

When `--auth-enabled` is combined with `--tls-mode=required`, the new validation guards only the gRPC transport. The HTTP listener is still started without TLS, and the HTTP auth middleware continues to accept bearer tokens over that plaintext connection. This gives a false sense of security — credentials can still be intercepted on the unprotected HTTP endpoint.

**Suggestion:** Wire TLS configuration into the HTTP listener when auth is enabled (e.g. pass a TLS config to `httpserver.NewHook`), disable the plaintext HTTP listener entirely when `--auth-enabled` is set, or extend the hard validation to also fail when an unprotected HTTP listener is active alongside `--auth-enabled`.

NumaryBot: 🔴 **[blocker] TLS mode `optional` still permits plaintext authenticated connections**

When `--auth-enabled` is used with `--tls-mode=optional`, the new check passes. However, `TLSModeOptional` still permits plaintext gRPC connections via `AllowsPlaintext()` and the dual listener, so a client can send bearer `Authorization` metadata over cleartext. The interception/replay risk that the validation aims to prevent is still fully present in this documented mode.

**Suggestion:** Extend the validation to also reject `TLSModeOptional` when `--auth-enabled` is set, requiring `TLSModeRequired` (or equivalent) as the only acceptable TLS mode when authentication is enabled.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1605#issuecomment-5019606962

internal/bootstrap/config.go
paul-nicolas: EN-1553 is explicitly scoped to authenticated service gRPC (per the ticket). The gRPC API terminates TLS on-process, which is what this guard protects. The plaintext HTTP REST-compat listener is a separate, pre-existing surface and out of scope for this change. Resolving.

paul-nicolas: This is intentional and consistent with the existing `--cluster-secret` guard just above (config.go:271), which likewise rejects only `TLSModeDisabled`. `optional` is deliberately permitted because it is the mode that enables zero-downtime TLS migration (docs/ops/tls-migration.md) — plaintext and TLS clients coexist while certs roll out — and this PR intentionally adds a passing test for `auth + optional`. Forcing `required` would break that migration path for authenticated clusters and diverge from the cluster-secret precedent. The guard's purpose is to reject the clearly-unsafe `disabled` + auth combination; hardening `optional` itself would be a separate, broader decision (covering cluster-secret too), out of scope here. Resolving.

flemzord (changes requested):
Reviewed current head `c79b3f87`. One runtime security blocker remains, already identified by NumaryBot: `TLSModeOptional` passes validation while its dual listener still accepts plaintext authenticated gRPC, contradicting the new guarantee that bearer credentials cannot travel in cleartext. The migration rationale does not invalidate that behavior. Auth must require `required` mode, or the change must otherwise prevent authenticated plaintext connections; I am not duplicating the existing inline thread.

internal/bootstrap/config.go
paul-nicolas: Update: after @flemzord's review we've reversed this. As of f79904298, `--auth-enabled` now requires `--tls-mode=required` — `optional` is rejected because its dual listener still accepts plaintext client connections, which would let bearer tokens travel in cleartext. This intentionally diverges from the `--cluster-secret` guard (which keeps permitting `optional`): the operator drives zero-downtime inter-node TLS migration through the transitional `optional` mode, but the external service API has no such requirement.

paul-nicolas:
@flemzord addressed in f79904298. `--auth-enabled` now requires `--tls-mode=required`; both `disabled` and `optional` are rejected, so authenticated plaintext connections can no longer occur.

Verified this does not break the operator's TLS migration: `computeTargetTLSMode` (reconcile_tls.go) bootstraps a fresh cluster directly to the desired mode and only routes through `optional` on a `disabled↔required` toggle — an auth-enabled cluster can never be at `disabled` (already rejected by this guard), so it never transits `optional`. A fresh auth+TLS cluster boots straight to `required`.

This intentionally diverges from the `--cluster-secret` guard, which keeps permitting `optional`: inter-node migration genuinely needs the transitional dual-listener, but the external service API does not. Docs (authentication / deployment / tls-migration / cli) updated to match. Re-requesting your review.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1605#issuecomment-5019606962

docs/ops/cli.md
```diff
@@ -4636,7 +4638,7 @@ Enable JWT/OIDC authentication with scope-based authorization. See [Authenticati
 
 | Flag | Type | Default | Description |
 |------|------|---------|-------------|
-| `--auth-enabled` | bool | `false` | Enable JWT authentication and scope-based authorization |
+| `--auth-enabled` | bool | `false` | Enable JWT authentication and scope-based authorization. Requires TLS (`--tls-mode` `optional` or `required`) — rejected with `--tls-mode=disabled` |
```
NumaryBot: 🟡 **[minor] CLI reference docs still indicate auth works with --tls-mode=optional**

According to the reviewer, the flag table at this line states that `--auth-enabled` is compatible with `--tls-mode=optional`, but the prior PR discussion clarifies that `optional` is intentionally permitted by the validation (only `TLSModeDisabled` is rejected). This means the concern is partially addressed by design intent, but the documentation may still be misleading to operators who expect the full set of modes to work with auth without caveats. If the docs do not clearly communicate that plaintext connections remain possible under `optional` mode, users may assume a stronger security guarantee than actually exists.

**Suggestion:** Add a note to the `--auth-enabled` flag documentation clarifying that `--tls-mode=optional` permits the server to start but still allows plaintext gRPC connections; recommend `--tls-mode=required` for production deployments where credentials must be protected in transit.

paul-nicolas: Good catch — fixed in 57439f149. The Server Authentication Flags table now states `--auth-enabled` requires `--tls-mode=required` and is rejected with `disabled` or `optional`, matching the guard and the other docs (authentication / deployment / tls-migration).

docs/ops/authentication.md
```diff
@@ -9,16 +9,23 @@ ledger run \
...
   --bootstrap \
+  --tls-mode required \
+  --tls-cert-file /etc/ledger/tls.crt \
+  --tls-key-file /etc/ledger/tls.key \
   --auth-enabled \
   --auth-issuer https://auth.example.com \
   --auth-service ledger
 ```
 
+> `--auth-enabled` requires `--tls-mode=required`; the server refuses to start
```
NumaryBot, line 20: 🟡 **[minor] TLS guarantee in docs not scoped to gRPC transport**

The updated authentication documentation states that `--auth-enabled` + `--tls-mode=required` ensures bearer tokens are protected in transit, but this guarantee applies only to the in-process gRPC listener. The HTTP REST-compat listener remains plaintext and will still accept bearer `Authorization` headers over `http://`. Operators reading the docs may incorrectly believe all credential traffic is TLS-protected by this flag combination.

**Suggestion:** Qualify the TLS guarantee explicitly to the gRPC/service transport (e.g., 'protects the gRPC service transport'). Add a note that the HTTP listener must be separately protected via HTTPS termination at the ingress/proxy layer when used with authentication.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1605#issuecomment-5019606962

docs/ops/authentication.md
paul-nicolas: Fixed in 119a40371. Both the note and the Configuration Invariants bullet now scope the guarantee explicitly to the gRPC service transport, and note that the HTTP REST-compat listener is not governed by `--tls-mode` and must be TLS-terminated separately (HTTPS at the ingress/proxy) when auth is enabled. This matches the PR's scope — the plaintext HTTP surface is pre-existing and out of scope for the code change, but the docs shouldn't overclaim.

internal/bootstrap/config.go
```diff
@@ -336,6 +336,17 @@ func (c Config) validateAuthConfig() error {
...
 		}
 
+		// Reject auth without full TLS — bearer JWT/Ed25519 tokens would be
+		// exposed to plaintext interception otherwise. Unlike the --cluster-secret
+		// guard in Validate() (which permits `optional` because the operator drives
+		// zero-downtime inter-node TLS migration through that transitional
+		// dual-listener), the external service API demands `required`: `optional`
+		// still accepts plaintext client connections, so a client could send
+		// Authorization metadata in cleartext.
+		if c.TLSConfig.Mode != TLSModeRequired {
```
NumaryBot, line 346: 🟠 **[major] Auth + TLS migration deadlock: operator may emit AUTH_ENABLED=true while TLS_MODE=optional during rollout**

When a user updates a CR to enable both TLS and auth on an existing plaintext cluster in a single operation, `computeTargetTLSMode` targets `optional` as the transitional mode for zero-downtime rolling TLS rollout, while `buildEnvVars` simultaneously emits `AUTH_ENABLED=true`. The new guard then rejects every pod with `TLS_MODE=optional + AUTH_ENABLED=true`, preventing rollout convergence and leaving the cluster stuck until the user manually splits the change into two separate CR updates.

**Suggestion:** Coordinate this guard with the operator rollout. Options include: (1) have the operator's `buildEnvVars` defer emitting `AUTH_ENABLED=true` until `computeTargetTLSMode` has fully converged to `required`; (2) document clearly in the operator's CR spec that enabling auth and TLS in the same update is unsupported — users must first reach `required`, then enable auth; or (3) add an operator-level validation webhook that rejects simultaneous auth+TLS enablement on a currently-disabled-TLS cluster.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1605#issuecomment-5019606962

internal/bootstrap/config.go
paul-nicolas: Real bug — thank you, this is a genuine regression the strict guard introduced. Fixed in 567cad7bb (option 1: coordinate the operator).

`buildEnvVars` now defers the ENTIRE auth env block (AUTH_ENABLED, AUTH_ISSUER(S), AUTH_SERVICE, AUTH_CHECK_SCOPES, AUTH_READ_KEY_SET_MAX_RETRIES, AUTH_SCOPE_MAPPING, AUTH_ANONYMOUS_SCOPES) and AUTH_ED25519_KEYS until `targetTLSMode == required`. So when a user enables TLS+auth together on an existing plaintext cluster, the transitional `optional` phase runs auth-off (no pod is ever booted into auth+optional → no crashloop → the rollout converges), and the next reconcile — now at `required` — enables auth on top. Fresh clusters bootstrap straight to `required` (no transitional `optional`), so a clean provision is unaffected. The API is briefly unauthenticated during the `optional` window, which is no worse than the plaintext+unauth state the cluster is migrating away from.

Gating the whole block (not just AUTH_ENABLED) is deliberate: emitting AUTH_ISSUER without AUTH_ENABLED would trip the server's "auth flags set without --auth-enabled" rejection, and AUTH_ED25519_KEYS auto-enables auth. Operator build/unit tests/lint pass.

misc/operator/internal/controller/envvars.go
```diff
@@ -280,8 +280,23 @@ func buildEnvVars(ledger *ledgerv1alpha1.Cluster, targetTLSMode string, credenti
...
+	// that phase would boot every pod into a config the server refuses, so the
+	// rolling update never becomes Ready, never converges, and the migration
+	// deadlocks at `optional`. We therefore hold ALL auth env off until the
+	// target mode reaches `required`; a subsequent reconcile then enables auth
+	// on top of the converged TLS. The API is briefly unauthenticated during
+	// the `optional` window — no worse than the plaintext state the cluster is
+	// migrating away from. When TLS is `disabled`/`optional` no auth env is
+	// emitted at all (auth defaults off), which also avoids the server's
+	// "auth flags set without --auth-enabled" rejection.
+	if targetTLSMode == tlsModeRequired && spec.Auth != nil {
```
NumaryBot, line 306: 🟠 **[major] Operator silently drops auth env vars when TLS is disabled, allowing unauthenticated pods**

When a CR has `spec.auth.enabled=true` but `spec.tls.enabled` is false (or TLS is toggled back toward disabled), the operator's `buildEnvVars` omits all authentication environment variables entirely. This means pods boot with authentication silently disabled even though the CR explicitly requests it. The new server-side guard (rejecting `--tls-mode` other than `required` when `--auth-enabled`) cannot fire because `AUTH_ENABLED` is never emitted, bypassing the security invariant the server guard is intended to enforce.

**Suggestion:** The operator should either (1) surface a validation error / set a Degraded status condition and refuse to reconcile when `spec.auth.enabled=true` and TLS is not converged to `required`, rather than silently omitting auth vars; or (2) if deferring auth emission is intentional for the TLS-enable migration path, document this clearly, gate it strictly to the transition window (TLS mode converging toward `required`), and surface a warning condition on the CR so the operator is not in a silently-unauthenticated state.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1605#issuecomment-5019606962

flemzord (changes requested):
Reviewed current head `567cad7b`. The previous optional-mode blocker is fixed: the server now requires `tls-mode=required`, with aligned tests and docs. One correctness/security blocker remains, already identified by NumaryBot: the operator suppresses requested auth whenever the target TLS mode is not `required`, including a steady `spec.auth.enabled=true` + TLS-disabled configuration or a transition toward disabled, so pods can run unauthenticated instead of rejecting or holding the invalid spec. Please restrict any auth deferral to the transition toward `required` and surface invalid configurations rather than silently disabling auth. Main CI is green; I am not duplicating the existing inline thread.

flemzord (dismissed):
Re-checked current head `567cad7b`. The previous authenticated-plaintext blocker is fixed: server validation now requires `TLSModeRequired`. One correctness/runtime security blocker remains, already captured by NumaryBot on this head: when the CR requests auth while TLS is disabled, `buildEnvVars` omits the entire auth block, so the pods start unauthenticated instead of rejecting or surfacing the invalid configuration. Deferring auth is reasonable only during a transition toward `required`; the stable disabled state must not silently contradict `spec.auth.enabled=true`. I am not duplicating the existing inline finding. Main CI is green; the NumaryBot failure corresponds to this active blocker.

flemzord (changes requested):
Current head `567cad7b` still has the existing NumaryBot correctness/security blocker: outside the transition toward TLS `required`, the operator must not silently drop requested auth and start unauthenticated pods. Main CI is green; no additional finding is introduced here.

misc/operator/internal/controller/envvars.go
paul-nicolas: Fixed in 6018bc502. `validateClusterConfig` (cluster_controller.go) now rejects `auth.enabled=true` with `tls.enabled=false|nil`, surfaced as a `ConfigValid=False / ValidationFailed` status condition with `Phase=Error` and no requeue — so a CR requesting auth without TLS is refused loudly instead of reconciling into silently-unauthenticated pods.

The `buildEnvVars` deferral this thread is anchored on is now provably safe: the only reachable non-`required` auth case is the transient `optional` window of a `disabled`->`required` TLS migration, where `tls.enabled` is already true (the invalid steady state is rejected up front). I added a cross-reference in the `buildEnvVars` comment to make that explicit, documented the coordination in `docs/ops/tls-migration.md`, and added table test `TestValidateClusterConfig_AuthRequiresTLS` covering the accept/reject cases. Operator build/tests/lint green.

paul-nicolas:
Addressed the remaining operator blocker in 6018bc502.

The operator now **rejects** `auth.enabled=true` with `tls.enabled=false|nil` in `validateClusterConfig` (ValidationFailed condition + `Phase=Error`, no reconcile) rather than silently omitting the auth env block. Auth deferral in `buildEnvVars` is thereby restricted to the legitimate `disabled`->`required` TLS transition window (where `tls.enabled` is already true), so pods can no longer boot unauthenticated behind an `auth.enabled=true` spec. Added `TestValidateClusterConfig_AuthRequiresTLS`, a cross-reference comment at the deferral site, and an operator-coordination note in `docs/ops/tls-migration.md`. Operator build/tests/lint green.

@flemzord re-requesting your review.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1605#issuecomment-5019606962

flemzord (approved):
Revalidated current head `6018bc50`. The prior operator blocker is fixed: `validateClusterConfig` now rejects an explicit auth-enabled/TLS-disabled steady state, while auth env deferral is limited to the legitimate transition toward TLS `required`. This prevents silently unauthenticated pods for the reported configuration and avoids the rollout deadlock. Main CI, NumaryBot, and CodeRabbit are green; I found no remaining correctness/runtime blocker.

Azorlogh (approved):

PR #1606, opened 2026-07-20
fix(EN-1554): make Schemathesis CI blocking and fix API conformity failures

paul-nicolas:
## Summary

The `Tests-Schemathesis` CI job ran under `continue-on-error: true`, so real API-conformity failures were reported green and no report was uploaded. This makes the job blocking and fixes the three failure classes it was masking. **All changes stay outside the FSM/Raft hot path.**

Jira: https://formance-team.atlassian.net/browse/EN-1554

## Changes

- **CI (`.github/workflows/main.yml`)** — remove `continue-on-error: true` so a non-zero `just test-schemathesis` fails the job.
  - *Note:* "fails the overall **required** workflow" also requires branch-protection to list `Tests-Schemathesis` as a required check (repo setting, not code). This PR is the code-side half.
- **`tests/schemathesis/run.sh`** — tee the run to `/tmp/schemathesis-report.txt` (`pipefail` preserves the exit code) and copy the server log to `/tmp/schemathesis-server.txt` before cleanup, so output + server diagnostics upload even on failure (both match the existing artifact glob).
- **`tests/schemathesis/test_api.py`** — set `with_security_parameters=False`. The harness runs auth-disabled, but `openapi.yml` declares a global `BearerAuth`; without this, Schemathesis synthesized `Authorization` headers whose fuzzed values Go's `net/http` rejected as plain-text 400s before routing (transport artifacts, not application responses).
- **`internal/application/ctrl/controller_default.go` + `internal/query/executor.go`** — map the internal `domain.ErrNotFound` sentinel to the typed `domain.ErrLedgerNotFound` (HTTP 404 `LEDGER_NOT_FOUND`) on `GetNumscript`, `ListNumscripts`, `ListPreparedQueries`, and `query.Execute` (`ExecutePreparedQuery`), instead of leaking it as a sanitized 500.
- **`openapi.yml`** — document the valid `LEDGER_DELETED` 409 on `setMetadataFieldType`, `removeMetadataFieldType`, `setDefaultEnforcementMode`, `removeAccountType`.

## Testing

- `MAX_EXAMPLES=10 just test-schemathesis` → **75/75 endpoints pass, 0 failures**, with all three checks enabled (`not_a_server_error`, `status_code_conformance`, `response_schema_conformance`); both diagnostic artifacts written.
- New table-driven unit tests assert all four read paths return `*domain.ErrLedgerNotFound` for missing and soft-deleted ledgers.
- `just pre-commit` (0 lint issues) and `just test` (full suite) both pass.

## Acceptance criteria

- [x] Non-zero `just test-schemathesis` fails `Tests-Schemathesis` (required-workflow gating is a branch-protection setting).
- [x] No `Authorization` header generated in the auth-disabled harness.
- [x] Missing/deleted ledgers → HTTP 404 `LEDGER_NOT_FOUND` on the four read paths.
- [x] The four `LEDGER_DELETED` responses documented as HTTP 409.
- [x] `MAX_EXAMPLES=10 just test-schemathesis` passes with existing checks enabled.
- [x] Output + server diagnostics uploaded even on failure.
- [x] No failure solved by accepting 500s, documenting parser-level 400s, or broad ignore lists.


openapi.yml
```diff
@@ -1475,6 +1475,8 @@ paths:
           $ref: '#/components/responses/Forbidden'
         '404':
           $ref: '#/components/responses/NotFound'
+        '409':
+          $ref: '#/components/responses/Conflict'
```
paul-nicolas, line 1484: [Medium] `dropIndex` can emit an undocumented 409 that the now-blocking gate is exposed to

This PR correctly adds `409 Conflict` to the four write endpoints whose apply paths return `ErrLedgerDeleted` on a soft-deleted ledger. But `dropIndex` (`DELETE /v3/{ledgerName}/indexes/{canonicalId}`) routes through the **same** `processApply` orchestrator (`internal/domain/processing/processor_apply.go`, the `DropIndex` case), which returns `&domain.ErrLedgerDeleted{}` → `KindConflict` → HTTP 409 *before dispatch* when the ledger is soft-deleted. Yet its response block declares only `204/400/401/403/404/500` — no 409 — whereas its sibling `createIndex` already documents one.

`MAX_EXAMPLES=10` didn't trip it, but now that the job is blocking, any fuzz sequence that soft-deletes the shared `test-ledger` and then calls `dropIndex` on it produces an undeclared 409 and `status_code_conformance` turns CI red on an endpoint this PR didn't touch. That directly undermines the goal of a *reliable* blocking gate.

Fix (one line, same as the four added here):

```yaml
        '409':
          $ref: '#/components/responses/Conflict'
```

on `dropIndex`. Worth a quick pass over the other `processApply`-routed writes for the same latent gap.

internal/application/ctrl/controller_default.go
```diff
@@ -1837,6 +1837,10 @@ func (ctrl *DefaultController) ListPreparedQueries(ctx context.Context, ledger s
 
 	ledgerInfo, err := query.GetLedgerByName(ctx, handle, ledger)
 	if err != nil {
+		if errors.Is(err, domain.ErrNotFound) {
+			return nil, &domain.ErrLedgerNotFound{Name: ledger}
```
paul-nicolas, line 1841: [Medium] Two different errorCodes now returned for the same "ledger not found" condition

Mapping to `*domain.ErrLedgerNotFound` here (and at :1886, :1947, and `internal/query/executor.go:65`) is correct for the 404 the PR targets, but it renders errorCode `LEDGER_NOT_FOUND` (`Reason()` → `error_handler.go:88`). The other 16 ledger-existence guards in this same file map `domain.ErrNotFound` → `commonpb.NewNotFoundError(...)`, which renders errorCode `NOT_FOUND` (`error_handler.go:80`). Both are HTTP 404 and both are schema-conformant (errorCode is a free-form string — no schemathesis impact), so this is not a bug, but clients now see two different errorCodes for the identical outcome across sibling read endpoints.

The typed `ErrLedgerNotFound` is the documented-preferred conversion target (`internal/domain/errors.go:13-16`), so the cleaner resolution is probably a follow-up migrating the 16 legacy sites to the typed error rather than changing this PR — but the divergence is worth a deliberate decision rather than landing by accident.

paul-nicolas (commented):
## Multi-model review — PR #1606 (Claude + Codex)

Reviewed the diff with two independent reviewers (Claude in-repo + Codex external), then verified every surviving finding against the actual code at the PR head.

**Verdict: the PR is sound and does what it claims.** I independently confirmed the load-bearing claims: `set -euo pipefail` is set (`run.sh:11`), so the added `| tee` genuinely propagates the non-zero exit and the gate actually blocks; the server log path (`$TMPDIR/server.log`) and CI upload glob (`/tmp/schemathesis-*.txt`, `if: always()`) are correct; the four openapi endpoints (`setMetadataFieldType`, `removeMetadataFieldType`, `setDefaultEnforcementMode`, `removeAccountType`) truly emit `409 LEDGER_DELETED` via `processApply → ErrLedgerDeleted → KindConflict`; the `ErrNotFound → ErrLedgerNotFound` mapping is precise (404) and the four read endpoints already document 404; and `GetLedgerByName` returns `ErrNotFound` for both missing and soft-deleted, so the new tests are valid.

**Findings entering verification: 6 Critical/Major/Medium candidates → 2 confirmed Medium (posted inline), 4 dropped/downgraded.**

Dropped/downgraded:
- *(Codex, major → rejected)* "the new 409 is wrong for these endpoints" — false: Codex conflated the read path (numscript/queries → 404) with the four write endpoints that got the 409; the write endpoints do emit 409 through `processApply`.
- *(Codex, major → Minor)* schemathesis `GenerationConfig` reliability — the pin `>=3.25.0,<4.0` resolves to 3.39.16 today, where `with_security_parameters` works (verified empirically); only a reproducibility nit.

Models: Claude (in-repo, 2 agents split by area) + Codex (external). No unresolved conflicts — the one Codex/Claude conflict (409 correctness) was resolved against the code in Claude's favor.

### Posted inline (2 × Medium)
- `openapi.yml:1479` — `dropIndex` can emit an undocumented 409 that the now-blocking gate is exposed to
- `internal/application/ctrl/controller_default.go:1841` — two errorCodes (`LEDGER_NOT_FOUND` vs `NOT_FOUND`) for the same condition

- `tests/schemathesis/test_api.py:179` — Minor: requirements pin `schemathesis>=3.25.0,<4.0` allows versions <3.36 where `with_security_parameters` was unreliable; it resolves to 3.39.16 today so it works, but consider raising the floor to `>=3.36` for reproducible CI.
- `internal/application/ctrl/controller_default_ledger_not_found_test.go:20` — Minor: the new tests assert the internal typed error; consider adding HTTP-level coverage asserting the actual 404 + reason for the four read endpoints (handler tests like `TestHandleExecutePreparedQuery_NotFound` already cover part of this).
- `internal/application/ctrl/controller_default.go` (`GetTemplateUsage` doc comment, ~1901) — Minor: comment claims the "same contract as GetNumscript"; now stale since `GetNumscript` emits `LEDGER_NOT_FOUND` while `GetTemplateUsage` still emits `NOT_FOUND`.
- `internal/query/executor.go:65` — Nit: the mapped branch returns the typed error raw while the fallback (:68) wraps with `fmt.Errorf`; the controller returns both raw. Harmless asymmetry.
- `tests/schemathesis/run.sh:296` — Nit: the trap runs async `kill "$SERVER_PID"` immediately before `cp`; the server's final buffered output may not be flushed, so the preserved `server.log` can be slightly truncated. Add `wait "$SERVER_PID" 2>/dev/null` before the `cp`.
- `tests/schemathesis/run.sh:296` — Nit: under `set -u`, `kill "$SERVER_PID"` errors if the trap fires before the server starts (e.g. `go build` fails). Use `"${SERVER_PID:-}"`. Pre-existing, but the whole trap line was rewritten here.

openapi.yml
```diff
@@ -1475,6 +1480,8 @@ paths:
...
```
NumaryBot, line 1484: 🟠 **[major] dropIndex endpoint missing documented 409 response**

The `dropIndex` (`DELETE /v3/{ledgerName}/indexes/{canonicalId}`) endpoint routes through the same `processApply` orchestrator as the four write endpoints fixed in this PR. When the ledger is soft-deleted, it returns `&domain.ErrLedgerDeleted{}` → `KindConflict` → HTTP 409, but the endpoint's response block only declares 204/400/401/403/404/500 — no 409. Its sibling `createIndex` already documents a 409. Now that the Schemathesis gate is blocking, any fuzz sequence that soft-deletes the shared `test-ledger` and then calls `dropIndex` on it will produce an undeclared 409, causing `status_code_conformance` failures in CI — directly undermining the goal of a reliable blocking gate.

**Suggestion:** Add a `'409': $ref: '#/components/responses/Conflict'` entry to the `dropIndex` response block, matching the pattern applied to the four other write endpoints in this PR. Also consider auditing any other `processApply`-routed endpoints for the same omission.

tests/schemathesis/requirements.txt
```diff
@@ -1 +1,8 @@
-schemathesis>=3.25.0,<4.0
+# Pinned exactly so the gate is reproducible: `derandomize=True` (see
+# test_api.py) makes Hypothesis generate a fixed input set, but only for a
+# given Hypothesis version — a different resolved version would generate
+# different inputs and could surface a failure locally != CI. Pin both so a
+# local run reproduces CI exactly. Bump deliberately (and re-run the suite) to
+# widen fuzzing coverage.
+schemathesis==3.39.16
```
NumaryBot, line 7: 🟡 **[minor] Pinned requirements not enforced when a pre-existing virtualenv exists**

In `run.sh`, the setup step only checks whether `import schemathesis` succeeds before skipping installation. If a developer or CI runner already has a `.venv` with the old floating versions, the new exact pins in `requirements.txt` are never applied. The suite then runs with the old resolved versions of Schemathesis/Hypothesis, defeating the reproducibility guarantee the pins were added to provide.

**Suggestion:** Change the venv check in `run.sh` to compare the installed package versions against the pinned requirements (e.g., use `pip install -r requirements.txt --quiet` unconditionally or check `pip freeze` against the requirements file), so the pinned environment is always enforced regardless of workspace state.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1606#issuecomment-5022238990

openapi.yml
paul-nicolas: Fixed in 43089756c — added `'409': $ref: '#/components/responses/Conflict'` to `dropIndex`.

I ran the audit you suggested over the other `processApply`-routed writes: `ErrLedgerDeleted` has exactly two producers (`processor_apply.go` for all apply orders, and `processCreateLedger`, which already documents 409), and cross-referencing every write op in `openapi.yml`, `dropIndex` was the only remaining gap. The other no-409 writes are not on this surface: `promoteLedger` / `updatePreparedQuery` / `deletePreparedQuery` have no `ErrLedgerDeleted` producer, `executePreparedQuery` maps missing/soft-deleted → 404, and `bulkOperations` reports sub-errors per-element inside a 200 envelope (top-level status is only 200 or 503).

openapi.yml
paul-nicolas: Fixed in 43089756c — 409 added to `dropIndex`, matching the four other write endpoints and `createIndex`.

I also audited the remaining `processApply`-routed writes: `dropIndex` was the only endpoint missing 409. `createLedger` already documents it; `promoteLedger`/`updatePreparedQuery`/`deletePreparedQuery` don't emit `ErrLedgerDeleted`; `executePreparedQuery` maps to 404; `bulkOperations` reports per-element inside a 200 envelope.

internal/application/ctrl/controller_default.go
paul-nicolas: Intentional — kept the typed `ErrLedgerNotFound` here: `internal/domain/errors.go` documents it as the preferred conversion target and the new `controller_default_ledger_not_found_test.go` locks it in. Both errorCodes are HTTP 404 with no conformance impact, so this PR deliberately starts the migration toward the typed error rather than reverting these four sites.

Migrating the ~16 legacy `NewNotFoundError("ledger %s not found", ...)` sites to the typed error is tracked as a follow-up: **EN-1568**. Resolving here.

tests/schemathesis/requirements.txt
paul-nicolas: Fixed in 35c0e257a — `run.sh` now installs the pinned requirements unconditionally (dropped the `import schemathesis` guard), so a pre-existing `.venv` with a different resolved version can no longer bypass the exact pins. `pip install` is a fast no-op when the pins are already satisfied, so the reproducibility guarantee now holds regardless of workspace state.

paul-nicolas:
Addressed all three points:
- `dropIndex` now documents 409 (43089756c) — audited every processApply-routed write, it was the only gap.
- Pinned schemathesis deps are now installed unconditionally (35c0e257a), so a reused `.venv` can't bypass the pins.
- The ledger-not-found errorCode divergence is intentional (`ErrLedgerNotFound` is the documented-preferred target, locked by a new test; both codes are HTTP 404). The legacy-site migration is tracked in EN-1568.

openapi.yml
NumaryBot, line 1484: 🟠 **[major] dropIndex endpoint missing documented 409 response**
<sub>reported by paul-nicolas, NumaryBot</sub>

The `dropIndex` (`DELETE /v3/{ledgerName}/indexes/{canonicalId}`) endpoint routes through the same `processApply` orchestrator as the four write endpoints fixed in this PR. When the ledger is soft-deleted, it returns `&domain.ErrLedgerDeleted{}` → `KindConflict` → HTTP 409, but the endpoint's response block only declares 204/400/401/403/404/500. Its sibling `createIndex` already documents a 409. With the Schemathesis gate now blocking, any fuzz sequence that soft-deletes the shared `test-ledger` and then calls `dropIndex` on it will produce an undeclared 409, causing `status_code_conformance` failures in CI and directly undermining the goal of a reliable blocking gate.

**Suggestion:** Add `'409': $ref: '#/components/responses/Conflict'` to the `dropIndex` response block, matching the pattern applied to the four other write endpoints in this PR. Also audit any other `processApply`-routed endpoints for the same omission.

tests/schemathesis/run.sh
```diff
@@ -17,9 +21,13 @@ HTTP_PORT=${HTTP_PORT:-9099}
...
 RAFT_PORT=${RAFT_PORT:-7779}
 MAX_EXAMPLES=${MAX_EXAMPLES:-50}
+SCHEMATHESIS_WORKERS=${SCHEMATHESIS_WORKERS:-1}
 
 TMPDIR=$(mktemp -d)
-trap 'kill "$SERVER_PID" 2>/dev/null; rm -rf "$TMPDIR"' EXIT
+# On exit, preserve the server log as an uploadable diagnostic BEFORE removing
+# TMPDIR, so a failing run still ships server-side context. The filename matches
+# the CI artifact glob (/tmp/schemathesis-*.txt).
+trap 'kill "${SERVER_PID:-}" 2>/dev/null; wait "${SERVER_PID:-}" 2>/dev/null; cp "$TMPDIR/server.log" /tmp/schemathesis-server.txt 2>/dev/null || true; rm -rf "$TMPDIR"' EXIT
```
NumaryBot: 🟠 **[major] EXIT trap may abort before copying server log, and can override test exit status**

With `set -e` still active inside the `EXIT` trap, a non-zero `kill` or `wait` (e.g., when the server has already exited or crashed, or when `wait` reports the SIGTERM status) stops the trap before the `cp` command runs. This means the server log artifact — most valuable precisely when the server has crashed — may never be copied. It can also override the intended exit status of the test command, corrupting CI pass/fail signalling.

**Suggestion:** Guard the cleanup commands so the log copy and tmpdir removal always run regardless of earlier failures in the trap, for example by prefixing individual commands with `|| true` or by temporarily disabling `set -e` within the trap body.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1606#issuecomment-5022238990

tests/schemathesis/requirements.txt
NumaryBot, line 7: 🟡 **[minor] Pinned requirements not enforced when a pre-existing virtualenv exists**

In `run.sh`, the setup step only checks whether `import schemathesis` succeeds before skipping installation. If a developer or CI runner already has a `.venv` with old floating versions, the new exact pins in `requirements.txt` are never applied. The suite then runs with the old resolved versions of Schemathesis/Hypothesis, defeating the reproducibility guarantee the pins were intended to provide.

**Suggestion:** Change the venv check in `run.sh` to unconditionally run `pip install -r requirements.txt --quiet`, or compare the installed package versions against the pinned requirements (e.g., via `pip freeze`), so the pinned environment is always enforced regardless of workspace state.

internal/application/ctrl/controller_default.go
NumaryBot, line 1841: ⚪ **[nit] Two different errorCodes returned for the same 'ledger not found' condition**

The new mappings to `*domain.ErrLedgerNotFound` render errorCode `LEDGER_NOT_FOUND`, while the 16 existing ledger-existence guards in the same file map `domain.ErrNotFound` → `NOT_FOUND`. Both are HTTP 404 and schema-conformant (no Schemathesis impact), but clients now see two different errorCodes for the identical outcome across sibling endpoints. This is a pre-existing inconsistency that this PR widens rather than introduces.

**Suggestion:** As a follow-up, migrate the 16 legacy sites to the typed `*domain.ErrLedgerNotFound` error so all ledger-not-found conditions render a consistent `LEDGER_NOT_FOUND` errorCode. No change needed in this PR to unblock merge.

tests/schemathesis/run.sh
paul-nicolas: Good catch — this is a real regression from the `wait` I added, fixed in a2dd35e6b. Under the active `set -e`, `wait` returns the server's SIGTERM status (143) and aborted the trap before `cp`, losing the server.log and corrupting the exit status. I reproduced it (trap aborts at exit 143) and confirmed the fix: every trap command is now guarded with `|| true`, so cleanup always completes and the real exit status is preserved.

openapi.yml
paul-nicolas: Already addressed — 43089756c added `'409': $ref: '#/components/responses/Conflict'` to `dropIndex` (present on the reviewed head). Audited every processApply-routed write; dropIndex was the only gap.

tests/schemathesis/requirements.txt
paul-nicolas: Already addressed in 35c0e257a — `run.sh` now installs the pinned requirements unconditionally (the `import schemathesis` guard is gone), so a reused `.venv` can't run unpinned versions.

internal/application/ctrl/controller_default.go
paul-nicolas: Intentional and acknowledged — the typed `ErrLedgerNotFound` is the documented-preferred target (locked by a new test); both codes are HTTP 404 with no conformance impact. Migrating the ~16 legacy `NewNotFoundError` sites is tracked as a follow-up in EN-1568.

NumaryBot (commented):

Resolved 4 stale NumaryBot review threads (3 fixed, 1 outdated).

Summary: https://github.com/formancehq/ledger/pull/1606#issuecomment-5022238990

flemzord (approved):
Reviewed current head `a2dd35e6`. The missing 409 contract, pinned-environment bypass, and EXIT-trap status/log-loss issue are fixed; the latest commit correctly guards every cleanup command under `set -e`. NumaryBot reports no findings and the complete main workflow, including E2E, model, Schemathesis, and coverage, is green. No remaining runtime blocker found.

PR #1607, opened 2026-07-20
fix(backup): rebuild idempotency keys on restore

Azorlogh:
## Problem

`RebuildDelta` (`internal/infra/backup/rebuild.go`) reconstructs every derived
projection from the exported audit delta — ledgers, txns, volumes, metadata,
references, boundaries, reversions (#1592) — but **not the idempotency keys**.

A backup is a full Pebble checkpoint (raw SSTs) plus incremental audit exports.
The raw checkpoint drags the idempotency zone (`0x05`) along incidentally, so
keys frozen **before** the last full checkpoint survive a restore. But keys
frozen **after** it live only in the audit, and nothing re-materialized them —
so on restore a node silently forgot every idempotency key in the delta window:

- a retried key **re-executed** instead of replaying (a duplicate), and
- a reused key with a **different** body **executed** instead of returning
  `IDEMPOTENCY_KEY_CONFLICT`.

This is the same shape as the #1592 reversion-bitset loss. The main-store
integrity checker did not catch it: `compareIdempotencyOutcomes` is
one-directional (it flags stored-but-unaudited / tampered entries, never
audited-but-missing ones).

Surfaced by the model-driver restore cycle (two findings, both post-restore: a
`removeType` reused-key bulk executed to `NotFound` and a `delMeta` reused-key
bulk committed, where the model expected `IDEMPOTENCY_KEY_CONFLICT`).

## Fix

`RebuildDelta` now iterates the delta-range audit entries and, for each keyed
proposal, re-derives its frozen outcome (committed-log range or definitive
failure) and persists it via `state.SaveIdempotencyKey` (main key + time index).

- The `(auditEntry, items) → IdempotencyKeyValue` derivation is extracted to
  `state.IdempotencyValueFromAudit`, **shared** with the checker's
  `expectedIdempotencyOutcome` so the two can't drift (the proposal hash is
  recomputed via `processing.HashOrders`, byte-identical to what was frozen).
- Expired entries are re-materialized like a live node's: the read-path TTL gate
  ignores them and the eviction scheduler reclaims them — no wall-clock read is
  introduced into the deterministic restore path.
- `saveIdempotencyKey` is exported as `state.SaveIdempotencyKey` (callable from
  the backup package); all callers updated.

## Testing

`tests/e2e/cluster/restore_idempotency_test.go` (`e2e && s3`), modeled on
`restore_reversion_test.go`: empty full checkpoint → keyed commits land in the
incremental delta → restore into a fresh store → assert both directions still
hold post-restore (same key+body replays with no duplicate; same key + different
body → `AlreadyExists`), pinned by the account volume.

Verified the test **catches the bug**:

- fix enabled → `5 Passed | 0 Failed`
- fix disabled → `4 Passed | 1 Failed`, failing exactly at the Phase-3 assertion
  `restored: reusing a key with a different body must conflict`.

Lint clean on the changed packages; unit tests green; main module builds.

## Out of scope

Making `compareIdempotencyOutcomes` bidirectional (so `Check()` itself flags an
audited-but-missing key) would close the checker's blind spot, but needs to be
TTL-window-aware to avoid flagging legitimately-expired keys — left as a
follow-up. The e2e test is the regression guard here.

### DIFF
--- a/internal/infra/backup/rebuild.go
+++ b/internal/infra/backup/rebuild.go
@@ -448,6 +448,15 @@ func RebuildDelta(
 		return fmt.Errorf("flushing rebuilt reversion bitsets: %w", err)
 	}
 
+	// Re-materialize idempotency keys frozen in the delta range (pre-checkpoint
+	// keys ride along in the raw checkpoint SSTs; post-checkpoint ones live only
+	// in the audit). Reads the delta audit entries seeded by ApplyExports.
+	if err := writer.rebuildIdempotency(ctx, readHandle, fromAuditSeq); err != nil {
+		_ = writer.batch.Cancel()
+
+		return fmt.Errorf("rebuilding idempotency keys: %w", err)
+	}
+
 	if err := writer.batch.Commit(); err != nil {
 		return fmt.Errorf("committing boundaries batch: %w", err)
 	}
@@ -798,6 +807,60 @@ func (w *attributeReplayWriter) applyAuditOrderEffects(reader dal.PebbleReader, 
 	}
 
 	return iter.Error()
+}
+
+// rebuildIdempotency re-materializes the idempotency-key projection for the
+// delta range. The full checkpoint carries pre-checkpoint keys as raw SSTs, but
+// keys frozen after it live only in the audit — so RebuildDelta re-derives each
+// keyed proposal's frozen outcome (committed-log range or definitive failure)
+// and persists it, letting a restored node dedup a retried key instead of
+// re-executing it. A key produced no frozen outcome (all-replay success or a
+// non-freezable failure) is skipped, matching the apply path. Expired entries
+// are re-materialized too: the read-path TTL gate ignores them and the leader's
+// eviction scheduler reclaims them, exactly as on a live node.
+func (w *attributeReplayWriter) rebuildIdempotency(ctx context.Context, reader dal.PebbleReader, fromAuditSeq uint64) error {
+	var after *uint64
+	if fromAuditSeq > 0 {
+		after = &fromAuditSeq
+	}
+
+	auditCursor, err := query.ReadAuditEntries(ctx, reader, after)
+	if err != nil {
+		return fmt.Errorf("reading audit entries for idempotency rebuild: %w", err)
+	}
+
+	defer func() { _ = auditCursor.Close() }()
+
+	for {
+		entry, err := auditCursor.Next()
+		if errors.Is(err, io.EOF) {
+			break
+		}
+		if err != nil {
+			return fmt.Errorf("reading audit entry for idempotency rebuild: %w", err)
+		}
+
+		key := entry.GetIdempotency().GetKey()
+		if key == "" {
+			continue
+		}
+
+		items, err := query.ReadAuditItems(ctx, reader, entry.GetSequence())
+		if err != nil {
+			return fmt.Errorf("reading audit items for idempotency rebuild (seq %d): %w", entry.GetSequence(), err)
+		}
+
+		value, ok := state.IdempotencyValueFromAudit(entry, items)
+		if !ok {
+			continue
+		}
+
+		if err := state.SaveIdempotencyKey(w.batch, key, value); err != nil {
+			return fmt.Errorf("persisting rebuilt idempotency key (seq %d): %w", entry.GetSequence(), err)
+		}
+	}
+
+	return nil
 }
 
 // deleteLedger reproduces the live DeleteLedger apply state: the LedgerInfo
--- a/internal/infra/backup/rebuild.go
+++ b/internal/infra/backup/rebuild.go
@@ -831,6 +831,10 @@ func (w *attributeReplayWriter) rebuildIdempotency(ctx context.Context, reader d
 
 	defer func() { _ = auditCursor.Close() }()
 
+	// Keys already given an outcome in this pass; a later failure entry (e.g. a
+	// same-key/different-body conflict) must not overwrite one — see below.
+	seen := make(map[string]struct{})
+
 	for {
 		entry, err := auditCursor.Next()
 		if errors.Is(err, io.EOF) {
@@ -855,9 +859,42 @@ func (w *attributeReplayWriter) rebuildIdempotency(ctx context.Context, reader d
 			continue
 		}
 
+		// A failure must not overwrite a key that already holds an outcome, just
+		// as the live path's recordIdempotencyFailure never overwrites a real
+		// result ("a conflict holds the real outcome"): a reused key with a
+		// different body is audited as a freezable (AlreadyExists) failure, so
+		// without this guard it would clobber the original success/failure. The
+		// prior outcome is either an earlier delta entry (seen) or a checkpoint
+		// SST (reader — its snapshot predates this pass's writes). A SUCCESS
+		// always writes: it only reaches the FSM when the gate found no live
+		// outcome, so any prior had expired — a legitimate re-freeze.
+		//
+		// Expiry is not re-evaluated here (unlike the live guard's !IsExpired):
+		// an outcome expiring within the exported delta would let the live path
+		// record a later failure this keeps as the earlier outcome — benign and
+		// unreachable at the default 24h TTL; matching it would need the TTL
+		// threaded through RebuildDelta.
+		if value.GetFailure() != nil {
+			_, dup := seen[key]
+			if !dup {
+				existing, err := state.LoadIdempotencyKey(reader, key)
+				if err != nil {
+					return fmt.Errorf("checking existing idempotency outcome for rebuild (seq %d): %w", entry.GetSequence(), err)
+				}
+
+				dup = existing != nil
+			}
+
+			if dup {
+				continue
+			}
+		}
+
 		if err := state.SaveIdempotencyKey(w.batch, key, value); err != nil {
 			return fmt.Errorf("persisting rebuilt idempotency key (seq %d): %w", entry.GetSequence(), err)
 		}
+
+		seen[key] = struct{}{}
 	}
 
 	return nil
--- a/internal/infra/backup/rebuild.go
+++ b/internal/infra/backup/rebuild.go
@@ -831,9 +831,26 @@ func (w *attributeReplayWriter) rebuildIdempotency(ctx context.Context, reader d
 
 	defer func() { _ = auditCursor.Close() }()
 
-	// Keys already given an outcome in this pass; a later failure entry (e.g. a
-	// same-key/different-body conflict) must not overwrite one — see below.
-	seen := make(map[string]struct{})
+	// The overwrite guard below mirrors the live recordIdempotencyFailure, which
+	// only protects a NON-expired outcome, so read the persisted TTL and apply
+	// the same expiry test. An absent config (exports-only restore, no
+	// checkpoint) yields ttl 0 = never expire — the conservative "never overwrite
+	// a prior" behaviour. (Threading TTL config changes through Raft so expiry is
+	// fully deterministic across nodes is future work; here it is the persisted
+	// value the checkpoint carried.)
+	cfg, err := query.ReadPersistedConfig(reader)
+	if err != nil {
+		return fmt.Errorf("reading persisted config for idempotency TTL: %w", err)
+	}
+
+	var ttlMicros uint64
+	if cfg != nil {
+		ttlMicros = cfg.GetIdempotencyTtlSeconds() * 1_000_000
+	}
+
+	// seen maps a key to the created_at of the outcome currently rebuilt for it,
+	// so a later failure can tell whether that prior is still live.
+	seen := make(map[string]uint64)
 
 	for {
 		entry, err := auditCursor.Next()
@@ -859,33 +876,31 @@ func (w *attributeReplayWriter) rebuildIdempotency(ctx context.Context, reader d
 			continue
 		}
 
-		// A failure must not overwrite a key that already holds an outcome, just
-		// as the live path's recordIdempotencyFailure never overwrites a real
-		// result ("a conflict holds the real outcome"): a reused key with a
-		// different body is audited as a freezable (AlreadyExists) failure, so
-		// without this guard it would clobber the original success/failure. The
-		// prior outcome is either an earlier delta entry (seen) or a checkpoint
-		// SST (reader — its snapshot predates this pass's writes). A SUCCESS
-		// always writes: it only reaches the FSM when the gate found no live
-		// outcome, so any prior had expired — a legitimate re-freeze.
-		//
-		// Expiry is not re-evaluated here (unlike the live guard's !IsExpired):
-		// an outcome expiring within the exported delta would let the live path
-		// record a later failure this keeps as the earlier outcome — benign and
-		// unreachable at the default 24h TTL; matching it would need the TTL
-		// threaded through RebuildDelta.
+		// A failure must not overwrite a key that already holds a LIVE outcome,
+		// mirroring the live recordIdempotencyFailure ("a conflict holds the real
+		// outcome"): a reused key with a different body is audited as a freezable
+		// (AlreadyExists) failure, so without this guard it would clobber the
+		// original success/failure. The prior outcome is an earlier delta entry
+		// (seen) or a checkpoint SST (reader — its snapshot predates this pass's
+		// writes). A SUCCESS always writes: it only reaches the FSM when the gate
+		// found no live outcome, so any prior had expired — a legitimate re-freeze.
 		if value.GetFailure() != nil {
-			_, dup := seen[key]
-			if !dup {
+			priorTs, have := seen[key]
+			if !have {
 				existing, err := state.LoadIdempotencyKey(reader, key)
 				if err != nil {
 					return fmt.Errorf("checking existing idempotency outcome for rebuild (seq %d): %w", entry.GetSequence(), err)
 				}
 
-				dup = existing != nil
+				if existing != nil {
+					priorTs, have = existing.GetCreatedAt(), true
+					seen[key] = priorTs
+				}
 			}
 
-			if dup {
+			// Skip only while the prior is still live at this entry's time —
+			// exactly the FSM's !IsExpired guard; an expired prior is overwritten.
+			if have && !state.IdempotencyExpired(priorTs, value.GetCreatedAt(), ttlMicros) {
 				continue
 			}
 		}
@@ -894,7 +909,7 @@ func (w *attributeReplayWriter) rebuildIdempotency(ctx context.Context, reader d
 			return fmt.Errorf("persisting rebuilt idempotency key (seq %d): %w", entry.GetSequence(), err)
 		}
 
-		seen[key] = struct{}{}
+		seen[key] = value.GetCreatedAt()
 	}
 
 	return nil
--- a/internal/infra/backup/rebuild.go
+++ b/internal/infra/backup/rebuild.go
@@ -852,6 +852,8 @@ func (w *attributeReplayWriter) rebuildIdempotency(ctx context.Context, reader d
 	// so a later failure can tell whether that prior is still live.
 	seen := make(map[string]uint64)
 
+	var written uint64
+
 	for {
 		entry, err := auditCursor.Next()
 		if errors.Is(err, io.EOF) {
@@ -910,6 +912,19 @@ func (w *attributeReplayWriter) rebuildIdempotency(ctx context.Context, reader d
 		}
 
 		seen[key] = value.GetCreatedAt()
+
+		// Flush in bounded chunks so restore memory does not grow with the whole
+		// delta, mirroring the log-replay pass above. The read snapshot is
+		// unaffected (committed keys stay invisible to it), and prior-outcome
+		// tracking lives in `seen`, so the guard above is unchanged across a flush.
+		written++
+		if written%5000 == 0 {
+			if err := w.batch.Commit(); err != nil {
+				return fmt.Errorf("committing idempotency batch at audit %d: %w", entry.GetSequence(), err)
+			}
+
+			w.batch = w.store.OpenWriteSession()
+		}
 	}
 
 	return nil
--- a/internal/infra/backup/rebuild.go
+++ b/internal/infra/backup/rebuild.go
@@ -812,12 +812,18 @@ func (w *attributeReplayWriter) applyAuditOrderEffects(reader dal.PebbleReader, 
 // rebuildIdempotency re-materializes the idempotency-key projection for the
 // delta range. The full checkpoint carries pre-checkpoint keys as raw SSTs, but
 // keys frozen after it live only in the audit — so RebuildDelta re-derives each
-// keyed proposal's frozen outcome (committed-log range or definitive failure)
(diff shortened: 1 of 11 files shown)


internal/infra/backup/rebuild.go
```diff
@@ -800,6 +809,60 @@ func (w *attributeReplayWriter) applyAuditOrderEffects(reader dal.PebbleReader,
...
+		if err != nil {
+			return fmt.Errorf("reading audit items for idempotency rebuild (seq %d): %w", entry.GetSequence(), err)
+		}
+
+		value, ok := state.IdempotencyValueFromAudit(entry, items)
+		if !ok {
+			continue
+		}
+
+		if err := state.SaveIdempotencyKey(w.batch, key, value); err != nil {
```
NumaryBot, line 866: 🔴 **[blocker] Do not overwrite live idempotency outcomes with conflicts**

When the audit delta contains a keyed request followed by a same-key/different-body retry, the live path writes an audit failure for the conflict but `recordIdempotencyFailure` deliberately does not overwrite the existing live outcome. This rebuild loop saves every freezable failure unconditionally, so a restore overwrites the original success/failure (from either the checkpoint SSTs or an earlier delta audit entry) with the later conflict entry; after restore, retrying the original body conflicts instead of replaying.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1607#issuecomment-5021332689

internal/infra/backup/rebuild.go
```diff
@@ -800,6 +809,97 @@ func (w *attributeReplayWriter) applyAuditOrderEffects(reader dal.PebbleReader,
...
+
+				dup = existing != nil
+			}
+
+			if dup {
+				continue
+			}
+		}
+
+		if err := state.SaveIdempotencyKey(w.batch, key, value); err != nil {
```
NumaryBot, line 866: 🔴 **[blocker] Rebuild loop unconditionally overwrites live idempotency outcomes with conflict entries**

When the audit delta contains a keyed request followed by a same-key/different-body retry, the live FSM records an audit failure for the conflict but deliberately does not overwrite the existing live outcome (via `recordIdempotencyFailure`'s non-overwrite guard). The rebuild loop, however, saves every freezable failure unconditionally. This means a restore overwrites the original success/failure (from checkpoint SSTs or an earlier delta audit entry) with the later conflict entry. After restore, retrying the original body will conflict instead of replaying the original frozen outcome, corrupting idempotency state.

**Suggestion:** Mirror the live `recordIdempotencyFailure` guard in the rebuild loop: before saving a freezable failure, check whether a non-expired outcome already exists for that key (either from the checkpoint or from an earlier delta entry), and skip the write if one does — exactly as the live FSM does.

internal/infra/backup/rebuild.go
```diff
@@ -800,6 +809,97 @@ func (w *attributeReplayWriter) applyAuditOrderEffects(reader dal.PebbleReader,
...
+		// SST (reader — its snapshot predates this pass's writes). A SUCCESS
+		// always writes: it only reaches the FSM when the gate found no live
+		// outcome, so any prior had expired — a legitimate re-freeze.
+		//
+		// Expiry is not re-evaluated here (unlike the live guard's !IsExpired):
+		// an outcome expiring within the exported delta would let the live path
+		// record a later failure this keeps as the earlier outcome — benign and
+		// unreachable at the default 24h TTL; matching it would need the TTL
+		// threaded through RebuildDelta.
+		if value.GetFailure() != nil {
```
NumaryBot: 🟠 **[major] Rebuild skips later freezable failures when earlier outcome has expired**

The rebuild deduplication logic treats any prior `seen` entry or checkpoint value as a duplicate and skips it, without checking whether that earlier outcome has expired. The live `recordIdempotencyFailure` path only skips existing values if they are non-expired. When a delta spans the TTL window (e.g. a short or custom TTL), the rebuild retains the old expired value and does not materialize the newer freezable failure. After restore, the TTL gate ignores the stale entry and the key can re-execute instead of replaying the frozen failure.

**Suggestion:** When determining whether to skip a later freezable failure, apply the same expiry check that the live FSM uses: only skip the write if the existing outcome is present AND has not yet expired at the time of rebuild.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1607#issuecomment-5021332689

internal/infra/backup/rebuild.go
```diff
@@ -800,6 +809,112 @@ func (w *attributeReplayWriter) applyAuditOrderEffects(reader dal.PebbleReader,
...
+			}
+
+			// Skip only while the prior is still live at this entry's time —
+			// exactly the FSM's !IsExpired guard; an expired prior is overwritten.
+			if have && !state.IdempotencyExpired(priorTs, value.GetCreatedAt(), ttlMicros) {
+				continue
+			}
+		}
+
+		if err := state.SaveIdempotencyKey(w.batch, key, value); err != nil {
```
NumaryBot, line 866: 🔴 **[blocker] Rebuild loop unconditionally overwrites live idempotency outcomes with conflict entries**

When the audit delta contains a keyed request followed by a same-key/different-body retry, the live FSM records an audit failure for the conflict but deliberately does not overwrite the existing live outcome (via `recordIdempotencyFailure`'s non-overwrite guard). The rebuild loop saves every freezable failure unconditionally, so a restore overwrites the original success/failure (from checkpoint SSTs or an earlier delta audit entry) with the later conflict entry. After restore, retrying the original body will conflict instead of replaying the original frozen outcome, corrupting idempotency state.

**Suggestion:** Mirror the live `recordIdempotencyFailure` guard in the rebuild loop: before saving a freezable failure, check whether a non-expired outcome already exists for that key (either from the checkpoint or from an earlier delta entry), and skip the write if one does — exactly as the live FSM does.

internal/infra/backup/rebuild.go
```diff
@@ -800,6 +809,112 @@ func (w *attributeReplayWriter) applyAuditOrderEffects(reader dal.PebbleReader,
...
+
+		// A failure must not overwrite a key that already holds a LIVE outcome,
+		// mirroring the live recordIdempotencyFailure ("a conflict holds the real
+		// outcome"): a reused key with a different body is audited as a freezable
+		// (AlreadyExists) failure, so without this guard it would clobber the
+		// original success/failure. The prior outcome is an earlier delta entry
+		// (seen) or a checkpoint SST (reader — its snapshot predates this pass's
+		// writes). A SUCCESS always writes: it only reaches the FSM when the gate
+		// found no live outcome, so any prior had expired — a legitimate re-freeze.
+		if value.GetFailure() != nil {
```
NumaryBot: 🟠 **[major] Rebuild skips later freezable failures when earlier outcome has expired (missing TTL expiry check)**

The rebuild deduplication logic treats any prior `seen` entry or checkpoint value as a duplicate and skips it, without checking whether that earlier outcome has expired. The live `recordIdempotencyFailure` path only skips existing values if they are non-expired. When a delta spans the TTL window (e.g. a short or custom TTL), the rebuild retains the old expired value and does not materialize the newer freezable failure. After restore, the TTL gate ignores the stale entry and the key can re-execute instead of replaying the frozen failure.

**Suggestion:** When determining whether to skip a later freezable failure, apply the same expiry check that the live FSM uses: only skip the write if the existing outcome is present AND has not yet expired at the time of rebuild.

internal/infra/backup/rebuild.go
NumaryBot, line 866: 🟠 **[major] Unbounded batch growth in idempotency rebuild WriteSession**

The idempotency rebuild loop accumulates every rebuilt idempotency main key and time-index write in the same `WriteSession` that is only committed after the entire audit scan. The log-replay pass above commits every 5000 records specifically to avoid unbounded batch growth, but this new pass carries no equivalent flush boundary. For large incremental backups with many keyed proposals, restore memory usage grows proportionally to the whole delta size and can OOM or fail on large backups.

**Suggestion:** Apply the same periodic flush strategy used in the log-replay pass (e.g., commit every 5000 idempotency entries) to bound memory consumption during rebuild.

internal/infra/backup/rebuild.go
```diff
@@ -800,6 +809,112 @@ func (w *attributeReplayWriter) applyAuditOrderEffects(reader dal.PebbleReader,
...
+	// checkpoint) yields ttl 0 = never expire — the conservative "never overwrite
+	// a prior" behaviour. (Threading TTL config changes through Raft so expiry is
+	// fully deterministic across nodes is future work; here it is the persisted
+	// value the checkpoint carried.)
+	cfg, err := query.ReadPersistedConfig(reader)
+	if err != nil {
+		return fmt.Errorf("reading persisted config for idempotency TTL: %w", err)
+	}
+
+	var ttlMicros uint64
```
NumaryBot: 🟠 **[major] Zero-value IdempotencyTtlSeconds treated as never-expire during rebuild**

When the persisted config has `IdempotencyTtlSeconds == 0` (e.g. a legacy row that will be backfilled on next boot, or because the runtime `--idempotency-ttl` was sub-second and got truncated by `Seconds()`), the rebuild treats the TTL as never-expiring. In a delta spanning the actual runtime TTL window, a later freezable failure can be skipped because the prior outcome appears still-live. After restore, the runtime TTL gate ignores that stale prior entry and the key can execute again instead of replaying the frozen failure.

**Suggestion:** Resolve the effective TTL from the runtime configuration rather than relying solely on the persisted `IdempotencyTtlSeconds` field; treat a zero value as 'unknown / use runtime default' rather than 'never expire', and apply the same expiry logic as the live FSM.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1607#issuecomment-5021332689

flemzord (changes requested):
Reviewed current head `9b92a306`. The live-outcome overwrite guard and expiry-aware replacement are now present, so I am not carrying those older blockers forward. Two current NumaryBot findings remain runtime blockers: the rebuild accumulates the entire idempotency projection in one WriteSession (unbounded restore memory/OOM risk), and a zero/absent persisted TTL is treated as never-expire even when the source runtime TTL can be finite (notably exports-only/legacy or sub-second truncation), which can rebuild the wrong frozen outcome. Please bound commits and supply an unambiguous effective TTL. I am not duplicating the existing inline threads.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1607#issuecomment-5021332689

internal/infra/backup/rebuild.go
```diff
@@ -800,6 +809,127 @@ func (w *attributeReplayWriter) applyAuditOrderEffects(reader dal.PebbleReader,
...
+			}
+
+			// Skip only while the prior is still live at this entry's time —
+			// exactly the FSM's !IsExpired guard; an expired prior is overwritten.
+			if have && !state.IdempotencyExpired(priorTs, value.GetCreatedAt(), ttlMicros) {
+				continue
+			}
+		}
+
+		if err := state.SaveIdempotencyKey(w.batch, key, value); err != nil {
```
NumaryBot, line 866: 🔴 **[blocker] Rebuild loop unconditionally overwrites live idempotency outcomes with conflict entries**

When the audit delta contains a keyed request followed by a same-key/different-body retry, the live FSM records an audit failure for the conflict but deliberately does not overwrite the existing live outcome (via recordIdempotencyFailure's non-overwrite guard). The rebuild loop saves every freezable failure unconditionally, so a restore overwrites the original success/failure (from checkpoint SSTs or an earlier delta audit entry) with the later conflict entry. After restore, retrying the original body will conflict instead of replaying the original frozen outcome, corrupting idempotency state.

**Suggestion:** Mirror the live recordIdempotencyFailure guard in the rebuild loop: before saving a freezable failure, check whether a non-expired outcome already exists for that key (either from the checkpoint or from an earlier delta entry), and skip the write if one does — exactly as the live FSM does.

internal/infra/backup/rebuild.go
```diff
@@ -800,6 +809,127 @@ func (w *attributeReplayWriter) applyAuditOrderEffects(reader dal.PebbleReader,
...
+
+		// A failure must not overwrite a key that already holds a LIVE outcome,
+		// mirroring the live recordIdempotencyFailure ("a conflict holds the real
+		// outcome"): a reused key with a different body is audited as a freezable
+		// (AlreadyExists) failure, so without this guard it would clobber the
+		// original success/failure. The prior outcome is an earlier delta entry
+		// (seen) or a checkpoint SST (reader — its snapshot predates this pass's
+		// writes). A SUCCESS always writes: it only reaches the FSM when the gate
+		// found no live outcome, so any prior had expired — a legitimate re-freeze.
+		if value.GetFailure() != nil {
```
NumaryBot: 🟠 **[major] Rebuild skips later freezable failures when earlier outcome has expired (missing TTL expiry check)**

The rebuild deduplication logic treats any prior 'seen' entry or checkpoint value as a duplicate and skips it, without checking whether that earlier outcome has expired. The live recordIdempotencyFailure path only skips existing values if they are non-expired. When a delta spans the TTL window (e.g. a short or custom TTL), the rebuild retains the old expired value and does not materialize the newer freezable failure. After restore, the TTL gate ignores the stale entry and the key can re-execute instead of replaying the frozen failure.

**Suggestion:** When determining whether to skip a later freezable failure, apply the same expiry check that the live FSM uses: only skip the write if the existing outcome is present AND has not yet expired at the time of rebuild.

internal/infra/backup/rebuild.go
NumaryBot, line 866: 🟠 **[major] Unbounded batch growth in idempotency rebuild WriteSession**

The idempotency rebuild loop accumulates every rebuilt idempotency main key and time-index write in the same WriteSession that is only committed after the entire audit scan. The log-replay pass above commits every 5000 records specifically to avoid unbounded batch growth, but this new pass carries no equivalent flush boundary. For large incremental backups with many keyed proposals, restore memory usage grows proportionally to the whole delta size and can OOM or fail on large backups.

**Suggestion:** Apply the same periodic flush strategy used in the log-replay pass (e.g., commit every 5000 idempotency entries) to bound memory consumption during rebuild.

internal/infra/backup/rebuild.go
```diff
@@ -800,6 +809,127 @@ func (w *attributeReplayWriter) applyAuditOrderEffects(reader dal.PebbleReader,
...
+	// checkpoint) yields ttl 0 = never expire — the conservative "never overwrite
+	// a prior" behaviour. (Threading TTL config changes through Raft so expiry is
+	// fully deterministic across nodes is future work; here it is the persisted
+	// value the checkpoint carried.)
+	cfg, err := query.ReadPersistedConfig(reader)
+	if err != nil {
+		return fmt.Errorf("reading persisted config for idempotency TTL: %w", err)
+	}
+
+	var ttlMicros uint64
```
NumaryBot: 🟠 **[major] Zero-value IdempotencyTtlSeconds silently treated as never-expiring during rebuild**

When the persisted config has IdempotencyTtlSeconds == 0 (e.g. a legacy row that will be backfilled on next boot, or because the runtime --idempotency-ttl was sub-second and got truncated by Seconds()), the rebuild treats the TTL as never-expiring. In a delta spanning the actual runtime TTL window, a later freezable failure can be skipped because the prior outcome appears still-live. After restore, the runtime TTL gate ignores that stale prior entry and the key can execute again instead of replaying the frozen failure.

**Suggestion:** Resolve the effective TTL from the runtime configuration rather than relying solely on the persisted IdempotencyTtlSeconds field; treat a zero value as 'unknown / use runtime default' rather than 'never expire', and apply the same expiry logic as the live FSM.

flemzord (changes requested):
Reviewed current head `dcb10485`. The new 5,000-write flush bounds restore memory, and the current code also contains the non-overwrite guard plus its expiry check, so I am not carrying those three repeated bot findings forward. One correctness blocker remains: `IdempotencyTtlSeconds == 0` is ambiguous between a true never-expire configuration, missing/legacy persisted config, and a sub-second runtime TTL truncated by `Seconds()`. Rebuild currently treats all three as never-expire, so it can retain an old outcome and skip the later frozen failure; after startup applies the finite runtime TTL, the retained outcome is expired and the key can execute again. Please provide an unambiguous effective TTL for rebuild. Main CI is still completing on this head.

Azorlogh:
@flemzord 

On the zero/absent TTL: the divergent cases are legacy-only. idempotency-ttl is persisted on first boot and backfilled, and a backup only comes from a booted cluster, so any current checkpoint carries the real value — a bare 0 is then either an intentional never-expire (handled, matching the FSM) or a pre-persistence legacy row.
Sub-second isn't a realistic idempotency TTL (it'd evict before any retry).
A general fix is a raft-deterministic TTL config, which is out of scope here and left as follow-up (the rebuild documents the 0 = never-expire assumption and points at it). Proposing to defer.

flemzord:
Thanks. This is not legacy-only under the current accepted configuration: `--idempotency-ttl` accepts, for example, `500ms`; the live machine derives expiry with `cfg.IdempotencyTTL.Microseconds()`, while persisted config stores `uint64(cfg.IdempotencyTTL.Seconds())`, which truncates that finite TTL to `0`. A checkpoint from a normally booted cluster therefore still loses the distinction, and rebuild treats the key as never-expiring, potentially retaining an expired earlier outcome and skipping the later frozen failure. Operational rarity is not enough for restore correctness. Rejecting positive TTLs below one second, persisting an unambiguous unit/presence, or supplying the effective runtime TTL to rebuild would close it; until then I am keeping this blocker.

internal/infra/backup/rebuild.go
```diff
@@ -800,6 +809,80 @@ func (w *attributeReplayWriter) applyAuditOrderEffects(reader dal.PebbleReader,
...
+		if err != nil {
+			return fmt.Errorf("reading audit items for idempotency rebuild (seq %d): %w", entry.GetSequence(), err)
+		}
+
+		value, ok := state.IdempotencyValueFromAudit(entry, items)
+		if !ok {
+			continue
+		}
+
+		if err := state.SaveIdempotencyKey(w.batch, key, value); err != nil {
```
NumaryBot, line 866: 🔴 **[blocker] Rebuild loop unconditionally overwrites live idempotency outcomes with conflict entries**

When the audit delta contains a keyed request followed by a same-key/different-body retry, the live FSM records an audit failure for the conflict but deliberately does not overwrite the existing live outcome (via `recordIdempotencyFailure`'s non-overwrite guard). The rebuild loop saves every freezable failure unconditionally, so a restore overwrites the original success/failure (from checkpoint SSTs or an earlier delta audit entry) with the later conflict entry. After restore, retrying the original body will conflict instead of replaying the original frozen outcome, corrupting idempotency state.

**Suggestion:** Mirror the live `recordIdempotencyFailure` guard in the rebuild loop: before saving a freezable failure, check whether a non-expired outcome already exists for that key (either from the checkpoint or from an earlier delta entry), and skip the write if one does — exactly as the live FSM does.

internal/infra/backup/rebuild.go
```diff
@@ -800,6 +809,80 @@ func (w *attributeReplayWriter) applyAuditOrderEffects(reader dal.PebbleReader,
...
+
+	defer func() { _ = auditCursor.Close() }()
+
+	var written uint64
+
+	for {
+		entry, err := auditCursor.Next()
+		if errors.Is(err, io.EOF) {
+			break
+		}
```
NumaryBot, line 846: 🔴 **[blocker] Zero-value IdempotencyTtlSeconds silently treated as never-expiring during rebuild**

The runtime `--idempotency-ttl` flag accepts sub-second durations (e.g. `500ms`). The live machine derives expiry via `cfg.IdempotencyTTL.Microseconds()`, but persisted config stores `uint64(cfg.IdempotencyTTL.Seconds())`, which truncates any sub-second TTL to 0. During rebuild, a zero value is treated as never-expiring. In a delta spanning the actual runtime TTL window, a later freezable failure can be skipped because the prior outcome appears still-live. After restore, the runtime TTL gate ignores that stale prior entry and the key can re-execute instead of replaying the frozen failure. This is not a legacy-only concern; it affects any normally-booted cluster using a sub-second idempotency TTL.

**Suggestion:** Resolve the effective TTL from the runtime configuration rather than relying solely on the persisted `IdempotencyTtlSeconds` field; treat a zero value as 'unknown / use runtime default' rather than 'never expire'. Alternatively, reject sub-second TTL values at configuration time, or persist an unambiguous unit/presence marker to close the truncation gap.

internal/infra/backup/rebuild.go
```diff
@@ -800,6 +809,80 @@ func (w *attributeReplayWriter) applyAuditOrderEffects(reader dal.PebbleReader,
...
+
+		// Flush in bounded chunks so restore memory does not grow with the whole
+		// delta, mirroring the log-replay pass above. The read snapshot is
+		// unaffected — committed keys stay invisible to it.
+		written++
+		if written%5000 == 0 {
+			if err := w.batch.Commit(); err != nil {
+				return fmt.Errorf("committing idempotency batch at audit %d: %w", entry.GetSequence(), err)
+			}
```
NumaryBot, line 877: 🟠 **[major] Deduplication check does not account for TTL expiry, dropping valid later freezable failures**

The rebuild deduplication logic treats any prior `seen` entry or checkpoint value as a duplicate and skips it, without checking whether that earlier outcome has expired. The live `recordIdempotencyFailure` path only skips existing values if they are non-expired. When a delta spans the TTL window (e.g. a short or custom TTL), the rebuild retains the old expired value and does not materialize the newer freezable failure. After restore, the TTL gate ignores the stale entry and the key can re-execute instead of replaying the frozen failure.

**Suggestion:** When determining whether to skip a later freezable failure, apply the same expiry check that the live FSM uses: only skip the write if the existing outcome is present AND has not yet expired at the time of rebuild.

internal/infra/backup/rebuild.go
NumaryBot, line 866: 🟠 **[major] Unbounded batch growth in idempotency rebuild WriteSession**

The idempotency rebuild loop accumulates every rebuilt idempotency main key and time-index write in the same `WriteSession` that is only committed after the entire audit scan. The log-replay pass above commits every 5000 records specifically to avoid unbounded batch growth, but this new pass carries no equivalent flush boundary. For large incremental backups with many keyed proposals, restore memory usage grows proportionally to the entire delta size and can OOM or fail.

**Suggestion:** Apply the same periodic flush strategy used in the log-replay pass (e.g., commit every 5000 idempotency entries) to bound memory consumption during rebuild.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1607#issuecomment-5021332689

flemzord (approved):
Reviewed current head `d98936da`. The previous runtime blockers are resolved by the current audit-driven fold: `IDEMPOTENCY_KEY_CONFLICT` entries are skipped from the chain-bound audit reason, genuine non-conflict outcomes remain last-write-wins, and the 5,000-entry flush still bounds restore memory. The current NumaryBot findings describe the removed TTL/overwrite-guard implementation and are not active on this head. Main CI is green. The remaining Codecov patch threshold failure (75.28%) is tooling-only here; project coverage and the repository Coverage job pass.

PR #1609, opened 2026-07-21
fix(EN-1564): avoid backfill for indexes declared on an empty ledger

paul-nicolas:
## Summary

Avoids the unnecessary historical backfill (and the readiness gap) for an index declared in the **same atomic apply batch** as its `CreateLedger`, before any ledger-local data exists — the Bitcoin-ingestion pattern.

Jira: https://formance-team.atlassian.net/browse/EN-1564

## Problem

When a ledger is created empty and its indexes are declared atomically, each `CreateIndex` was still treated as an index added to a historical ledger: the indexbuilder seeded `IndexVersionState{Current:0, Pending:1}`, scheduled a backfill from `cursor=0`, and gated indexed reads with `ErrIndexBuilding` until the backfill caught the moving global cursor. For a ledger born empty there is nothing to backfill, so this is pure wasted work plus an availability gap on the hot path.

## Approach

The FSM tracks, per proposal, ledgers created in the batch that have not yet emitted any indexable data log (transient, non-persisted, deterministic). When `CreateIndex` applies for such a ledger it stamps a new `CreatedIndexLog.initial` flag. The read-side indexbuilder promotes an `initial` index straight to live (`IndexVersionState{Current:1, Pending:0}`) and schedules **no** backfill; the live indexing path maintains it from ledger birth. Later indexes (added to a ledger with data, or created in a separate batch) are unchanged — they still backfill and stay gated until `current_version > 0`.

The classification is conservative: only same-atomic-batch-before-any-data qualifies. A false positive would silently skip indexing real history, so anything else keeps the existing backfill behavior.

## Changes

- `misc/proto/common.proto`: add `bool initial = 5` to `CreatedIndexLog` (regenerated).
- `internal/domain/processing`: per-proposal born-empty tracking on `Context`; stamp `initial` in `processCreateIndex`.
- `internal/application/indexbuilder`: `handleCreatedIndexLog` promotes initial indexes to live with no backfill.
- Tests: processing unit tests (born-empty + stamping), indexbuilder unit tests (initial vs non-initial), and an e2e scenario (atomic create+index → live immediately, immediate ingestion resolves with no `ErrIndexBuilding`).
- Docs: initial-vs-later index semantics.

## Invariants

- **#2 determinism:** born-empty tracking is a pure function of the deterministically-ordered batch.
- **#8 checker:** no new persisted projection; `compareIndexes` keys on `IndexID` identity and excludes `BuildStatus`, so the flag needs no compare pass.
- **#9 coverage gate:** no new cache-attribute reads on the hot path.
- **#10 order immutability:** only the FSM-produced log is stamped; the accepted `Order` is untouched.
- **Boot safety:** an initial index persists `Current:1`, and `loadIndexRegistry` skips backfill scheduling when `current != 0`.

## Testing

- `just pre-commit`: 0 lint issues (both modules).
- `just test` (root module unit tests): pass.
- New unit tests pass; e2e compiles under the `e2e` tag (runs in CI).

## Acceptance criteria

- [x] Atomic create + index creates no backfill task.
- [x] `GetIndexStatus` reports `current_version > 0` and no backfill cursor for initial indexes.
- [x] Immediately-ingested transactions are indexed by the live path.
- [x] A later index on a ledger with data still backfills and stays gated until complete.
- [x] Covered by tests (unit + e2e).
- [x] Docs clarify initial vs later index semantics.

### DIFF
--- a/internal/application/indexbuilder/index_config.go
+++ b/internal/application/indexbuilder/index_config.go
@@ -402,10 +402,11 @@ func (b *Builder) getOrCreateLedgerConfig(ledger string) *ledgerIndexConfig {
 // declared on a born-empty ledger), there is no local history to replay — the
 // index is promoted straight to live (current=1) and NO backfill is scheduled.
 //
-// Idempotency: when the same CreateIndex is replayed against an index that
-// is already cached as READY (the processor short-circuited a duplicate
-// create on an already-built index), we skip the backfill scheduling so the
-// builder does not redo work that has already completed.
+// Idempotency: when the same CreateIndex is replayed (or re-submitted) against
+// an index this replica has already promoted to live, we skip the reset and
+// backfill scheduling so the builder does not redo work that has already
+// completed — and, more importantly, does not knock a live index back into
+// ErrIndexBuilding.
 func (b *Builder) handleCreatedIndexLog(ledgerName string, log *commonpb.CreatedIndexLog) {
 	id := log.GetId()
 	if id == nil {
@@ -414,8 +415,17 @@ func (b *Builder) handleCreatedIndexLog(ledgerName string, log *commonpb.Created
 
 	cfg := b.getOrCreateLedgerConfig(ledgerName)
 
-	if existing := cfg.byCanonical[indexes.Canonical(id)]; existing != nil &&
-		existing.GetBuildStatus() == commonpb.IndexBuildStatus_INDEX_BUILD_STATUS_READY {
+	// Post-EN-1323 the per-replica readiness signal is
+	// IndexVersionState.CurrentVersion, not the (now purely informational)
+	// registry BuildStatus — nothing ever flips BuildStatus back to READY, so a
+	// BuildStatus-based short-circuit here is dead. If this replica has already
+	// promoted the index to live (current != 0), a repeated CreatedIndexLog —
+	// a duplicate CreateIndex re-emitted by the processor, or an apply replay —
+	// must be a no-op: re-seeding {current:0, pending:1} and rescheduling a
+	// backfill would flip an already-live index back to ErrIndexBuilding. This
+	// mirrors the loadIndexRegistry boot guard and covers both the EN-1564
+	// initial fast path and the normal post-backfill live state.
+	if current, _ := b.versionFor(ledgerName, indexes.Canonical(id)); current != 0 {
 		return
 	}
 
--- a/internal/application/indexbuilder/index_config.go
+++ b/internal/application/indexbuilder/index_config.go
@@ -398,6 +398,10 @@ func (b *Builder) getOrCreateLedgerConfig(ledger string) *ledgerIndexConfig {
 // The index starts in BUILDING state — it is NOT marked as ready here.
 // A backfill task is created to replay historical logs for the new index.
 //
+// Initial fast path (EN-1564): when the log carries the initial flag (index
+// declared on a born-empty ledger), there is no local history to replay — the
+// index is promoted straight to live (current=1) and NO backfill is scheduled.
+//
 // Idempotency: when the same CreateIndex is replayed against an index that
 // is already cached as READY (the processor short-circuited a duplicate
 // create on an already-built index), we skip the backfill scheduling so the
@@ -419,6 +423,33 @@ func (b *Builder) handleCreatedIndexLog(ledgerName string, log *commonpb.Created
 		Id:                     id,
 		BuildStatus:            commonpb.IndexBuildStatus_INDEX_BUILD_STATUS_BUILDING,
 		ForwardEncodingVersion: 1,
+	}
+
+	// EN-1564: an index declared on a born-empty ledger has no local history to
+	// replay. Promote it straight to live (current=1) and skip the backfill;
+	// the live indexing path maintains it from ledger birth. Persist so a reboot
+	// sees current!=0 and loadIndexRegistry skips scheduling a backfill.
+	if log.GetInitial() {
+		state := readstore.IndexVersionState{
+			CurrentVersion: 1,
+			PendingVersion: 0,
+		}
+
+		if b.wb != nil && b.readStore != nil {
+			if batch := b.wb.Batch(); batch != nil {
+				if err := b.readStore.WriteIndexVersionState(batch, ledgerName, indexes.Canonical(id), state); err != nil {
+					b.logger.WithFields(map[string]any{
+						"ledger": ledgerName,
+						"index":  indexes.Canonical(id),
+						"error":  err,
+					}).Errorf("Persisting IndexVersionState on initial CreateIndex")
+				}
+			}
+		}
+
+		b.putVersionState(ledgerName, indexes.Canonical(id), state)
+
+		return
 	}
 
 	// First time this replica sees the index: target v=1 via the
--- a/docs/technical/architecture/subsystems/indexer/indexes.md
+++ b/docs/technical/architecture/subsystems/indexer/indexes.md
@@ -84,6 +84,15 @@ stateDiagram-v2
     Steady --> [*]: DropIndex
 ```
 
+### Initial indexes vs. later indexes
+
+An index gets a fast path when it is declared in the **same atomic apply batch** as the `CreateLedger` that creates its ledger, before any indexable data log for that ledger. The FSM classifies this per-proposal — a ledger is treated as "born empty" until it emits its first indexable data log — and stamps the result on a new `CreatedIndexLog.initial` boolean.
+
+- **Initial index** (`CreatedIndexLog.initial == true`): there is no history to replay, so the indexbuilder promotes the index straight to live on applying the log — it seeds `IndexVersionState{CurrentVersion: 1, PendingVersion: 0}` and schedules **no** historical backfill. `GetIndexStatus` immediately reports `current_version > 0` and carries no backfill cursor.
+- **Later index** (`CreatedIndexLog.initial == false`): the unchanged path described above. This covers an index added to a ledger that already holds data, **and** an index created in a separate apply batch even if the ledger is still empty. It is seeded `IndexVersionState{CurrentVersion: 0, PendingVersion: 1}`, backfilled from cursor `0`, and gated by `current_version == 0` (queries get `ErrIndexBuilding`) until the backfill completes and the atomic switch flips the served version.
+
+The classification is deliberately conservative: only the same-atomic-batch-before-any-data case qualifies as initial. A separate-batch index on a still-empty ledger backfills exactly as before — safe (it replays an empty history and completes immediately), just not routed through the zero-cost promotion.
+
 ## Statistics (computed on demand)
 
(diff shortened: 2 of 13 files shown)


flemzord (approved):
Reviewed current head `b038066f`. The born-empty tracker is scoped to one `ProcessOrders` call, so it cannot leak across proposals; the initial flag is derived deterministically from the ordered batch, and the read-side version state is persisted in the same batch before later live logs are indexed. I found no correctness/runtime blocker. The protobuf addition is pre-stable and non-blocking, and main CI is green.

NumaryBot (commented):

The PR correctly avoids triggering a backfill for indexes declared on an empty ledger, but it leaves an exploitable gap: the builder cache entry for the index remains in BUILDING state even though no follow-up task will ever move it to READY. If the same index is submitted a second time before the service restarts, the duplicate is accepted, the READY short-circuit is bypassed, version state is reset, a full historical backfill is scheduled, and the index becomes temporarily unavailable via ErrIndexBuilding — exactly the condition the PR is trying to prevent. The cache entry must be advanced to READY as part of the initial fast-path to close this window.

### Findings outside the diff

🟠 **[major] Initial index not marked READY in builder cache, enabling accidental backfill on duplicate CreateIndex** — `internal/application/indexbuilder/index_config.go:424`

When a CreateIndex is handled via the `initial` fast-path (empty ledger), the builder cache entry is left in BUILDING state. No backfill task or readiness transition is ever scheduled to advance it. If the same index is submitted again before a restart, `processCreateIndex` accepts the duplicate, `handleCreatedIndexLog` misses the READY short-circuit, resets version state to `{current:0, pending:1}`, schedules a full historical backfill, and reintroduces `ErrIndexBuilding` for an index that was already live.

**Suggestion:** After taking the initial fast-path, immediately transition the builder cache entry from BUILDING to READY (same as the normal post-backfill promotion). This prevents any subsequent duplicate CreateIndex from scheduling an unnecessary backfill and eliminates the transient `ErrIndexBuilding` window.



Summary: https://github.com/formancehq/ledger/pull/1609#issuecomment-5032502475

flemzord (changes requested):
Re-checked current head `b038066f` after the new NumaryBot finding. The initial fast path persists version state `{current:1,pending:0}` but leaves the builder cache entry (and FSM registry entry) in `BUILDING`, with no readiness transition scheduled. A later duplicate `CreateIndex` therefore bypasses the READY short-circuit, emits a non-initial log, and can schedule a historical backfill, reintroducing `ErrIndexBuilding` for an already-live index. This is an active correctness/runtime blocker. I am not duplicating the existing inline finding. Main CI is green; the NumaryBot status failure corresponds to this blocker.

flemzord (changes requested):
Revalidated live head `b038066f`. The existing NumaryBot duplicate-create/BUILDING blocker remains active and is not repeated here. One distinct correctness/invariant blocker also remains: the initial fast path newly persists and trusts readstore `IndexVersionState{current:1,pending:0}` to skip backfill and serve the index, but the checker explicitly verifies only SubAttrIndex identity and does not rederive either readstore contents or version state from the audit-bound `CreatedIndexLog.initial`. Corruption can therefore make an empty/stale index appear live without any check-store mismatch. Invariant #8 requires this persisted projection/lifecycle path to have a checker comparison. Main CI is green apart from NumaryBot, whose failure corresponds to the other active blocker.

flemzord (changes requested):
Revalidated live head `b038066f`. One additional correctness/runtime blocker is not covered by the existing duplicate-create or checker threads: a same-batch `CreateLedger` → `CreateIndex(log-date)` marks the index `initial`, but the read-side builder snapshots `cfg` before `indexPayload` handles the `CreatedIndexLog`. The date-index write therefore skips the CreateIndex ledger log itself. The `initial` path then sets `CurrentVersion=1` and schedules no backfill, so the index is announced live while a date-filtered log query permanently omits that log (the normal backfill path would have indexed it). Please make the initial fast path account for the activation log/current batch, and add a regression covering this sequence. Main CI is green; the existing NumaryBot failure corresponds to the separate BUILDING-state blocker.

paul-nicolas:
Thanks for the careful reviews. Addressing the three distinct concerns raised across the review verdicts and the NumaryBot finding:

**Concern A — duplicate `CreateIndex` reschedules a backfill (NumaryBot / CHANGES_REQUESTED): confirmed and fixed in 1df4c344c.**
The root cause is deeper than the initial fast-path. Since EN-1323 nothing ever sets `BuildStatus = READY` — it is now purely informational, and readiness is the per-replica `IndexVersionState.CurrentVersion`. So the `BuildStatus == READY` short-circuit in `handleCreatedIndexLog` was dead code, and `completeBackfill` leaves `cfg.byCanonical` at BUILDING even on the normal path. A second `CreatedIndexLog` for an already-live index therefore re-seeded `{current:0, pending:1}` and rescheduled a full backfill — flipping a live index back into `ErrIndexBuilding` on both the EN-1564 initial path *and* (pre-existing) the normal post-backfill path.
Rather than resurrect the obsolete READY signal, the fix replaces the dead guard with a per-replica readiness guard on `IndexVersionState.CurrentVersion != 0` (mirroring the `loadIndexRegistry` boot guard), making a repeated `CreatedIndexLog` a no-op once the index is live. Regression test `TestHandleCreatedIndexLog_DuplicateAfterLive_IsIdempotent` added.

**Concern B — checker/invariant #8 does not verify the persisted `IndexVersionState`: respectfully, this is out of main-store checker scope by construction.**
`IndexVersionState` lives in the readstore's own separate Pebble DB (`readstore/store.go`), which `Check()` never opens — `checker.go:98` opens only the primary `*dal.Store`, and the `check` package doesn't import `readstore`. That is exactly the peer-secondary-store carve-out in invariant #8 refinement (a): the readstore (and usagestore) are out of main-store checker scope by construction, with corruption/tamper handled by the readstore detect/drop/rebuild tracked under **EN-1323** — a per-replica rebuild-health concern, not an invariant-#8 main-store concern.
This PR introduces no new checker obligation: the non-initial path already persisted this same out-of-scope key (`{0,1}`) before EN-1564; the audit-bound `CreatedIndexLog.Initial` bit stays hash-chained; and the primary `SubAttrIndex` registry entry is unchanged by the flag and still fully verified both-ways by `compareIndexes`. Extending the main-store checker to walk readstore version state would contradict the invariant #8 peer-store boundary — the right home for hardening this is EN-1323.

**Concern C — initial index skips its own activation log for date-filtered log queries: the premise does not hold.**
The normal backfill also never indexes a `CreateIndex` config log — `isDataLog` (`backfill.go:1400-1424`) excludes it and the backfill loop skips every `!isDataLog` (`backfill.go:1307-1313`). So the initial fast path is behavior-identical to the backfill it replaces: both write zero `lldt` entries for the CreateIndex log. Additionally, `initial=true` is only stamped for born-empty ledgers (`processor_index.go`), which by definition contain no indexable data, so the elided backfill would have indexed nothing. The CreateIndex log itself stays discoverable via the unconditional per-ledger `llog` index for unfiltered `ListLogs`.
Separately (out of scope for this PR): there is a pre-existing live-vs-backfill asymmetry — the live date-index write at `process_logs.go:187` lacks the `isDataLog` guard that `backfill.go:1308` has, so a config log arriving while the date index is already active gets an `lldt` entry live but would be dropped on a backfill rebuild (over-indexing, opposite direction to this concern). Worth its own ticket.

flemzord (changes requested):
Revalidated live head `b038066f` after the latest author response. I withdraw my checker/version-state and activation-log concerns: the live repository instructions explicitly place the peer `readstore` outside the main-store checker scope, and `CreateIndex` config logs are excluded from the normal backfill as well as this fast path. One correctness/runtime blocker remains on the actual PR head: duplicate `CreateIndex` can still reset a live index to building and schedule a backfill. Commit `1df4c344` appears to address it, but it is not the PR head yet, so I cannot validate it as part of this review. Main CI is green; the NumaryBot failure corresponds to this remaining blocker.

internal/application/indexbuilder/index_config.go
```diff
@@ -410,8 +415,17 @@ func (b *Builder) handleCreatedIndexLog(ledgerName string, log *commonpb.Created
...
+	// IndexVersionState.CurrentVersion, not the (now purely informational)
+	// registry BuildStatus — nothing ever flips BuildStatus back to READY, so a
+	// BuildStatus-based short-circuit here is dead. If this replica has already
+	// promoted the index to live (current != 0), a repeated CreatedIndexLog —
+	// a duplicate CreateIndex re-emitted by the processor, or an apply replay —
+	// must be a no-op: re-seeding {current:0, pending:1} and rescheduling a
+	// backfill would flip an already-live index back to ErrIndexBuilding. This
+	// mirrors the loadIndexRegistry boot guard and covers both the EN-1564
+	// initial fast path and the normal post-backfill live state.
+	if current, _ := b.versionFor(ledgerName, indexes.Canonical(id)); current != 0 {
```
NumaryBot, line 428: 🟡 **[minor] Stale in-memory version state may suppress legitimate index creation after ledger name reuse**

When a ledger is deleted and its name is reused, the DeleteLedger path clears persisted IndexVersionState but does not evict b.indexVersions in memory. The new `current != 0` guard introduced to prevent accidental backfill on duplicate CreateIndex can then misidentify the recreated ledger's CreateIndex as an already-live duplicate, silently return without scheduling a backfill or seeding fresh version state, and leave the new index permanently behind ErrIndexBuilding until the service restarts.

**Suggestion:** On the DeleteLedger path, evict the corresponding entries from b.indexVersions (and any other in-memory caches keyed by ledger name/index) so that a subsequent CreateIndex for a reused ledger name is always treated as genuinely new rather than a duplicate.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1609#issuecomment-5032502475

internal/application/indexbuilder/index_config.go
paul-nicolas: Good catch, fixed in b3f3175d9. `DeleteLedgerIndexes` already wipes the persisted `IndexVersionState` (SubInternalIndexVersion is a ledger-scoped prefix), but the live apply path left the in-memory `b.indexVersions` mirror intact, so a same-name recreate in the same process would read a stale CurrentVersion and be skipped by the new guard. Added `dropLedgerVersionState`, called from the live `processLogs` DeleteLedger branch — deliberately not from the shared `markLedgerDeletedInBatch`, since the backfill replay path calls that for a historical delete of the task ledger and must keep the in-progress version state so `completeBackfill` can promote the recreated generation. Regression tests added.

flemzord:
Revalidated live head `b3f3175d`. The current-version guard keeps an already-live index live on duplicate `CreateIndex`, and the new live `DeleteLedger` eviction closes the same-name reuse gap without disturbing backfill replay state; the regressions cover both paths. Build, unit, E2E, model, scenarios, schemathesis, Dirty, and NumaryBot are green (Coverage is still queued). I found no remaining correctness/runtime blocker. GitHub approval is intentionally disabled by this cron configuration.

Azorlogh (approved):

flemzord (approved):

PR #1610, opened 2026-07-21
chore(nix): bump Go to 1.26.5 via stable 26.05 go_1_26

sylr:
## What

- Ran `nix flake update` — refreshed the `nixpkgs`, `nixpkgs-unstable`, and `nur` pins in `flake.lock`.
- Moved the Go compiler from `nixpkgs-unstable` (`go`) to the stable 26.05 channel (`go_1_26`) in `flake.nix`.
- Regenerated `*.pb.go` for the protoc bump (`v7.34.1 → v7.34.2`) pulled in by the flake update (version-comment-only change, no wire/API impact).

## Why

The flake describes itself as a "Go 1.26 development environment" but sourced the compiler from `nixpkgs-unstable`, which currently tops out at **1.26.4**. The 26.05 stable channel already ships `go_1_26` **1.26.5**.

## Verification

```
$ nix develop --command go version
go version go1.26.5 darwin/arm64
```

---
The Go-tools-to-`go.mod` migration was split out into #1611.

PR #1611, opened 2026-07-21
chore(tools): source Go codegen/lint tools from go.mod tool directives

sylr:
## What

Move the Go codegen/lint toolchain out of the Nix dev shell into `tool` directives in `go.mod`, invoked via `go tool <name>`, so each tool's version is pinned alongside the code it generates instead of tracking a floating nixpkgs channel.

- **Root `go.mod`:** mockgen, protoc-gen-go/-grpc/-vtproto, golangci-lint, gomarkdoc, ginkgo
- **Operator `go.mod`:** controller-gen, setup-envtest, golangci-lint

## Invocation changes
- `go:generate` directives call `go tool mockgen`
- `just lint` uses `go tool golangci-lint`
- `just generate-proto` builds protoc-gen-go/-grpc/-vtproto from the pinned tool deps into `build/` and passes them via `--plugin` (protoc needs plugin executables by path; matches the existing local-plugin pattern)
- `controller-gen` (root + operator justfiles) uses `go tool controller-gen`
- flake shellHook runs `go tool setup-envtest` from `misc/operator`

**Still in Nix** (not `go install`-able, or out of scope): `protoc` (protobuf_34), jsonnet/jb, vhs, k6, trufflehog, grpcurl, go-tools/gotools/goperf.

## Notable side effects (validated)
- **vtproto** now generates with **v0.6.1** (already required by go.mod) instead of nixpkgs' 0.6.0 — fixes a latent generator/runtime-lib skew. Regenerated `*_vtproto.pb.go` verified via `just fuzz-check` (marshal/unmarshal seed round-trip).
- **operator CRDs** restamped controller-gen `v0.21.0 → v0.17.3` (the series matching `controller-runtime v0.20.4`); **CRD content is byte-identical**.
- controller-gen pinned to v0.17.3 rather than `@latest`: a `tool` directive shares the module graph, and v0.21.0 pulls k8s.io v0.34+ which conflicts with the operator's pinned controller-runtime v0.20.4.

## Verification
- `just pre-commit` (rebuilt shell, no Nix tools): 0 lint issues in both modules; generate/proto/controller-gen all run via `go tool`.
- `just fuzz-check`: all proto marshal/unmarshal seed corpora pass.

---
Split out of #1610 (which now carries only the Go 1.26.5 bump).


flake.nix
```diff
@@ -41,10 +41,13 @@
     {
       devShells = forEachSupportedSystem ({ pkgs, pkgs-unstable, system }:
         let
+          # Go-based codegen/lint tools (mockgen, protoc-gen-go/-grpc/-vtproto,
```
NumaryBot, line 44: 🔴 **[blocker] Finish migrating every mockgen directive**

In a fresh `nix develop` shell, `just pre-commit` still runs `just generate`, which executes `go generate ./...`; many remaining directives still call bare `mockgen` (for example `internal/application/ctrl/controller_default.go:87`, `internal/storage/spool/spool.go:35`, and `internal/infra/node/transport.go:29`). Since this patch removes `mockgen` from the shell, generation fails with `mockgen: command not found`; update the remaining directives to `go tool mockgen` or keep the binary available.

flake.nix
```diff
@@ -41,10 +41,13 @@
...
+          # golangci-lint, gomarkdoc, ginkgo, controller-gen, setup-envtest) are
```
NumaryBot, line 45: 🟠 **[major] Keep the pre-commit lint hook runnable**

When developers use the installed pre-commit hook from `.pre-commit-config.yaml`, it still invokes `golangci-lint fmt` with `language: system`. Removing `golangci-lint` from the dev shell means that hook now fails in the intended Nix environment unless the user has a global binary; the hook should invoke `go tool golangci-lint fmt` (or the shell should continue to provide it).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1611#issuecomment-5032275021

flemzord (approved):
Reviewed current head `c67ea892`. The current diff has migrated the remaining `go:generate mockgen` directives to `go tool mockgen`; a live scan of this SHA found no bare directives, so the earlier NumaryBot blocker no longer applies. Main CI and Codecov are green, and I found no correctness/runtime blocker. The existing `.pre-commit-config.yaml` hook still invoking bare `golangci-lint` is a tooling issue already reported by NumaryBot; it should be fixed, but it is non-blocking under this review policy. The PR remains a draft.

PR #1438, opened 2026-06-29
feat!: color of money — segregated balances per (account, asset, color)

gfyrag:
## Summary

Introduces **strict balance segregation by color** as a first-class
dimension of the ledger. A balance lives at `(account, asset, color)`;
postings carry an explicit `color` field. The empty color is the
*uncolored* bucket and is itself fully segregated from every colored
bucket.

Companion PR upstream: [formancehq/numscript#139](https://github.com/formancehq/numscript/pull/139) — `Posting.Color` exposed in numscript output, `BalanceQuery` refactored to carry color as a separate dimension instead of an asset suffix. The ledger pins that branch via a local `replace` in `go.mod` for now; bump to a tagged release once #139 is merged.

## Invariants

1. **Color is immutable.** Once funds carry a color, no operation
   changes it. Re-coloring is a deliberate composition (clearing
   account + mint from `@world`), not a primitive.
2. **Empty color is a bucket.** `Color == ""` participates in
   segregation just like any colored bucket. A draw under color `X`
   cannot dip into uncolored funds, and vice versa.
3. **`^[A-Z]*$`** color charset, enforced both upstream (numscript) and
   at the ledger admission boundary. The empty string remains valid.
4. **Double-entry per bucket.** For every `(asset, color)`, the sum of
   all account balances is zero. Each bucket is its own conservation
   universe; the `scenariotest` invariant check is updated accordingly.

## Wire shapes

- `Posting.color` exposed everywhere postings appear.
- `Account.volumes` becomes `repeated AccountVolume{asset, color, volumes}`
  sorted by `(asset, color)` ascending — deterministic JSON.
- `PostCommitVolumes` inner shape becomes a sorted list of
  `VolumeEntry{asset, color, volumes}`.
- `AggregatedVolume.color` is set on every entry; the default response
  is one entry per `(asset, color)` tuple.

For consumers that don't care about color, two opt-in flags let the
server collapse colors into a single per-asset entry:

- `GET /{ledger}/accounts/{address}?collapseColors=true`
- `GET /{ledger}/volumes?collapseColors=true` (or
  `AggregateVolumesRequest.collapse_colors = true`)

## Storage layout

`domain.VolumeKey` carries `Color string`. Canonical bytes:

\`\`\`
[ledgerID BE 4B] [account] \x00 [color] \x00 [asset_base] [precision 1B]
\`\`\`

Color sits between account and asset so prefix scans behave naturally:
\`[ledgerID][account]\x00\` still returns every `(color, asset)` for the
[...]

PR #1613, opened 2026-07-21
feat(EN-1541)!: make QueryFilter the sole HTTP selector for volume aggregation

Azorlogh:
## Context

[EN-1541](https://formance-team.atlassian.net/browse/EN-1541)

`GET /v3/{ledgerName}/volumes` exposed only `prefix`, while the controller and gRPC request already accept a general Accounts-target `QueryFilter`. This exposes that existing capability through the HTTP adapter and removes `prefix`, aligning volumes with the other filtered endpoints. No new aggregation or QueryFilter semantics are introduced.

Rebased on the post-#1438 handler; `collapseColors` behavior and tests are preserved.

## Changes

- **`handlers_aggregate_volumes.go`** — drop `prefix` and its hand-built `AddressMatch`; parse the shared dual-format `filter` via `parseListFilter(w, r, QUERY_TARGET_ACCOUNTS)` and forward it unchanged to `AggregateVolumes`. Mirrors `handleListAccounts`.
- `useMaxPrecision` / `groupByPrefixes` / `collapseColors` are untouched and stay independent of filtering.
- **`openapi.yml`** — `prefix` replaced with `$ref: ListFilter`.
- **`api-comparison.md`**, **`query-pipeline.md`** — document the generic `filter` as the sole account selector.

## Client impact (breaking)

- `?prefix=users:` → `?filter=address ^= "users:"` (or structured `{"$match":{"address":"users:"}}`, URL-encoded).
- Hard removal, no alias: a caller still sending `?prefix=` gets no error — the param is ignored and the endpoint aggregates over all accounts.
- New: metadata / `account has asset` / boolean `and`/`or`/`not` selection; invalid-for-target and malformed filters return `400`.

## Tests

Extended HTTP handler tests: dual-format equivalence, address + metadata reaching the controller, target-invalid + malformed → `400` without invoking the backend, coexistence with precision/grouping/color-collapse, and absence of legacy `prefix` interpretation.

`GOROOT= go build ./...`, `go vet`, and `go test ./internal/adapter/http/` all pass.

[EN-1541]: https://formance-team.atlassian.net/browse/EN-1541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/technical/architecture/subsystems/read-path/query-pipeline.md
+++ b/docs/technical/architecture/subsystems/read-path/query-pipeline.md
@@ -67,7 +67,10 @@ query parameter and AND-combine the result with the transactions endpoint's
 remaining convenience param — the `startDate`/`endDate` range — through
 `combineFilters` (the `reference` and `prefix` aliases were removed in EN-1540:
 account prefix is now `address ^= "..."`, transaction reference is structured
-`{"$match":{"reference":"..."}}`); prepared-query
+`{"$match":{"reference":"..."}}`); the aggregate-volumes endpoint
+(`GET /{ledger}/volumes`) also parses `filter` through `parseListFilter` for the
+Accounts target as its sole account selector, its `prefix` parameter removed the
+same way (EN-1541); prepared-query
 create/update decode the JSON `filter` body field through the same helper; and
 `ledgerctl --filter` routes through it in `cmdutil.BuildQueryFilter`. Audit
 fields are written **bare** (`outcome`, `ledger`, `seq`, `proposal_id`,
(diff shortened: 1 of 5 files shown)


flemzord (approved):
Reviewed current head `8b4c6cc8`. The handler now uses the shared dual-format filter parser with the Accounts target and forwards the compiled filter without changing precision, grouping, or color-collapse behavior. The hard removal of the legacy `prefix` query parameter is documented across OpenAPI and architecture docs; as a pre-stable API compatibility change it is non-blocking here. Main CI, NumaryBot, CodeRabbit, and coverage are green, and I found no correctness/runtime blocker.

paul-nicolas (approved):

PR #1614, opened 2026-07-21
fix(revert): keep original postings out of the audit-bound order

Azorlogh:
## Problem

Admission enriched `RevertTransactionOrder` with `OriginalPostings` read from the target transaction's state at propose time. That order is serialized into `AuditItem.SerializedOrder` and bound into the audit hash chain — so the chain proved an *enriched* order rather than the request the caller submitted: data the caller never provided, and whose presence/absence depended on which node admitted the request and what it could read. This violates accepted-order immutability and contributes to audit/idempotency hash pollution.

## Fix

- `RevertTransactionOrder` now carries only caller intent: `transaction_id`, `force`, `at_effective_date`, `metadata`.
- The FSM sources the original postings from the coverage-gated `TransactionState` it already reads (preloaded via `addTransactionTargetNeeds`) — no order field, no new read, invariant #3 (no Pebble reads on the hot path) untouched.
- Admission still resolves the postings at order-build time (signed receipt or `Transaction` attribute) to declare the volume-preload coverage the reversed postings touch (invariant #9), but holds them in an admission-local overlay sidecar rather than on the wire order.
- A missing `TransactionState` for an *allocated* id is now a loud inconsistency (invariant #7): chapter archival does not evict `TransactionState`, and a genuinely non-existent id is already rejected by the `txID >= NextTransactionId` boundary check.
- Revert color validation dropped — reversed postings inherit colors already validated at create time.

## Acceptance criteria

- [x] Audit-serialized revert order contains only caller-provided intent.
- [x] `OriginalPostings` no longer lives on the accepted/audit-bound order.
- [x] FSM execution stays deterministic, uses only covered/preloaded data, no Pebble reads.
- [x] Idempotency for the same revert request is unaffected by admission-time enrichment.
- [x] Tests cover successful revert, missing original transaction, and the receipt/store path.
- [x] Docs describe where state-derived revert execution data lives.

## Testing

- Unit suites green: `admission`, `processing`, `check`, `state`, `usagebuilder`, `adapter/http`, `adapter/v2`, `raftcmdpb`.
- New FSM test for the allocated-but-missing-state inconsistency path; updated success / at-effective-date / not-found / receipt / intra-batch-fold tests.
- `just pre-commit` clean — 0 lint issues, no codegen/`go mod tidy` drift.

## Docs

New "Accepted Order Enrichment" section in `docs/technical/architecture/audit-vs-technical-state.md`.

### DIFF
--- a/docs/technical/architecture/audit-vs-technical-state.md
+++ b/docs/technical/architecture/audit-vs-technical-state.md
@@ -78,6 +78,38 @@ adding a variant, classify it with the table above and answer:
 If any answer is yes, the variant needs either an audited order or an explicit
 checker pass that verifies the stored value against audit-bound data.
 
+## Accepted Order Enrichment
+
+An accepted order is audit-bound: its serialized bytes feed the hash chain
+(`AuditItem.SerializedOrder`, via `state.BuildPerItemPayload`). The order must
+therefore carry only **caller intent** — the fields the caller submitted.
+Admission must not enrich it with state-derived execution data, even when that
+data is deterministic given the request. Doing so binds admission-time state into
+the audit chain and makes the serialized order depend on which node admitted it
+and what that node could read.
+
+State-derived execution inputs the FSM needs at apply time flow through a
+different channel: the **coverage-gated preloaded state**. Admission declares the
+`preload.Needs` for every key the apply path reads (invariant #6), and the FSM
+reads the value through `Scope.GetX(...)` under the coverage gate (invariant #9).
+Because the apply path cannot read Pebble (invariant #3), the value comes from the
+cache the preload seeded — never from the order.
+
+**Revert is the canonical example.** `RevertTransactionOrder` carries only
+`transaction_id`, `force`, `at_effective_date`, `metadata`, and `expand_volumes`.
+The original postings the FSM reverses are **not** on the order: admission
+preloads the target `TransactionState` (`addTransactionTargetNeeds`) and
+`processRevertTransaction` reads `origState.GetPostings()` through the coverage
+gate. Admission still resolves the postings at order-build time — from the signed
+receipt or the `Transaction` attribute — to declare the volume-preload coverage
+the reversed postings touch (invariant #9), but it holds them in an
+admission-local sidecar (`bulkOverlay.revertOriginalPostings`), never on the wire
+order. Chapter archival does not evict `TransactionState` (`executePurge` removes
+only the cold log/audit ranges), so a revert of a pre-archival transaction reads
+its postings from the surviving projection; a missing state for an *allocated* id
+is an inconsistency surfaced loudly (invariant #7), not a routine not-found —
+that case is caught earlier by the `txID >= NextTransactionId` boundary check.
+
 ## Mirror Cursor
 
 Mirror ingestion is the highest-risk boundary because the mirror cursor is an
--- a/internal/application/admission/admission.go
+++ b/internal/application/admission/admission.go
@@ -515,7 +515,7 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) ([]*
 	}
 
 	// Step 1: Extract preload needs from orders (excludes script-dependent needs)
-	needs, perOrder, err := a.extractPreloadNeeds(ctx, orders)
+	needs, perOrder, err := a.extractPreloadNeeds(ctx, orders, overlay)
 	if err != nil {
 		return nil, err
 	}
@@ -1053,7 +1053,7 @@ func addTransactionTargetNeeds(p *plan.Coverage, ledgerName string, txID uint64)
 // extractLedgerScopedNeeds populates the preload Coverage for a ledger-scoped
 // order. The ledger lives once on the wrapper; every payload variant reads it
 // from there instead of carrying its own field.
-func extractLedgerScopedNeeds(p *plan.Coverage, ls *raftcmdpb.LedgerScopedOrder) {
+func extractLedgerScopedNeeds(p *plan.Coverage, ls *raftcmdpb.LedgerScopedOrder, overlay *bulkOverlay) {
 	ledgerName := ls.GetLedger()
 	ledgerKey := domain.LedgerKey{Name: ledgerName}
 
@@ -1222,7 +1222,13 @@ func extractLedgerScopedNeeds(p *plan.Coverage, ls *raftcmdpb.LedgerScopedOrder)
 		case *raftcmdpb.LedgerApplyOrder_RevertTransaction:
 			addTransactionTargetNeeds(p, ledgerName, applyData.RevertTransaction.GetTransactionId())
 
-			for _, posting := range applyData.RevertTransaction.GetOriginalPostings() {
+			// The reversed postings touch the same accounts as the original;
+			// admission resolved them at order-build time into the overlay
+			// sidecar (the order itself carries only caller intent).
+			originalPostings := overlay.revertOriginalPostingsFor(
+				domain.TransactionKey{LedgerName: ledgerName, ID: applyData.RevertTransaction.GetTransactionId()},
+			)
+			for _, posting := range originalPostings {
 				addVolumeNeed(p, ledgerName, posting.GetDestination(), posting.GetAsset(), posting.GetColor())
 				addVolumeNeed(p, ledgerName, posting.GetSource(), posting.GetAsset(), posting.GetColor())
 			}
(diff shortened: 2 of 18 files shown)


internal/application/admission/admission.go
```diff
@@ -1222,7 +1222,13 @@ func extractLedgerScopedNeeds(p *plan.Coverage, ls *raftcmdpb.LedgerScopedOrder)
 		case *raftcmdpb.LedgerApplyOrder_RevertTransaction:
 			addTransactionTargetNeeds(p, ledgerName, applyData.RevertTransaction.GetTransactionId())
 
-			for _, posting := range applyData.RevertTransaction.GetOriginalPostings() {
+			// The reversed postings touch the same accounts as the original;
+			// admission resolved them at order-build time into the overlay
+			// sidecar (the order itself carries only caller intent).
+			originalPostings := overlay.revertOriginalPostingsFor(
```
NumaryBot, line 1228: 🟠 **[major] Add volume coverage for state-sourced reverts**

When a batch creates transaction N and later reverts N (or any admission-side Transaction.Get miss for a tx that is visible by apply time), the overlay has no postings, so this loop declares no volume keys for the revert. The new FSM then reads postings from TransactionState and calls applyPosting/buildPostCommitVolumes, whose volume reads are per-order coverage-gated; without those keys in the revert order's coverage bits the request fails with a coverage miss/internal storage error instead of the intended business outcome.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1614#issuecomment-5035006146

flemzord (changes requested):
Reviewed current head `d470da2c`. I confirm the active NumaryBot correctness blocker: when admission cannot resolve postings but the transaction becomes visible by apply time (for example create + revert in one batch), the revert order declares no volume coverage and the FSM coverage-gated volume reads fail instead of applying the valid revert.

A second correctness/runtime blocker remains for upgrades: on the base branch, `OriginalPostings` is explicitly the migration bridge for pre-EN-1242 `TransactionState` values that do not carry `Postings`. This patch removes that bridge and then treats an empty stored posting set as `ErrTransactionStateInconsistent`, making those existing transactions impossible to revert after upgrade. This is not a pre-stable protobuf reservation concern; it affects persisted data at runtime. Preserve a deterministic fallback or migration path before removing the order-carried postings. Main CI is still pending, but these blockers are independently sufficient.

internal/application/admission/admission.go
Azorlogh: Not reachable for reverts — leaving as-is.

Revert targets are by explicit `transaction_id` only (`resolveRevertTarget` rejects id 0; there's no reference-based or batch-local revert), so a batch can't create tx N and revert N in one request — the caller can't know N's apply-assigned id. The generic "admission miss for a tx visible at apply" would need the create to apply *between* the revert's propose and apply, which is impossible for revert-by-id: the caller only learns the id after the create applied and responded, and admission is leader-only gated by `waitLeaderReady`.

For every reachable input the outcome is unchanged from pre-PR:
- existing tx → admission reads its postings, declares the same volume keys, and the FSM reads the identical postings from `TransactionState` → covered;
- non-existent tx → the `txID >= NextTransactionId` boundary check returns `ErrTransactionNotFound` before any `applyPosting` (previously the empty-`OriginalPostings` check did the same).

The coverage miss you describe is only reachable if admission fails to read an allocated, already-applied tx — an admission↔FSM cache desync (an invariant violation) where the pre-PR code also failed, returning a misleading `NotFound`. There's no correct fix at admission: coverage can't be declared for postings admission can't read, and when it can read them the coverage is already correct.

flemzord (changes requested):
Revalidated the author reply against current head `d470da2c` and the live base. I withdraw the preload scenario: reverts require an explicit non-zero transaction ID, so a caller cannot target an apply-assigned ID from an earlier create in the same batch.

One correctness/runtime upgrade blocker remains. On the base branch, admission explicitly uses `OriginalPostings` as the migration bridge for pre-EN-1242 `TransactionState` values whose `Postings` field is empty. This patch removes that bridge, then reads only `origState.GetPostings()` and returns `ErrTransactionStateInconsistent` when it is empty. Existing pre-migration transactions therefore become impossible to revert after upgrade. Please preserve a deterministic fallback or provide a migration path before removing the order-carried postings. Main CI is green; the remaining NumaryBot status is tied to the withdrawn scenario and is not itself blocking.

Azorlogh:
> Existing pre-migration transactions therefore become impossible to revert after upgrade

@flemzord software is not stable, project rules reject compatiblity shims

docs/technical/architecture/audit-vs-technical-state.md
```diff
@@ -78,6 +78,38 @@ adding a variant, classify it with the table above and answer:
 If any answer is yes, the variant needs either an audited order or an explicit
 checker pass that verifies the stored value against audit-bound data.
 
+## Accepted Order Enrichment
```
paul-nicolas, line 81: [Medium] docs/technical/architecture/subsystems/chapters/receipts.md:26,78 — stale: documents the removed field and the FSM consuming receipt postings

receipts.md still states RevertTransactionOrder "carries an optional original_postings field," that "the FSM reads the postings from Pebble," and that the FSM "consumes them [the receipt's Postings] at revert time." All three are false after this PR: the field is gone, and the FSM now reverts from the coverage-gated TransactionState — the receipt is used only by admission to declare volume coverage. CLAUDE.md's Documentation Maintenance rule requires this be updated in the same change. Fix: rewrite the receipts.md "The revert path" section to describe TransactionState-sourced postings and the receipt's new admission-only role. (Codex also named counters.md here — I checked it and it's fine; it never claims original_postings, only that revert counting reads the order, which still holds.)

NumaryBot, line 81: 🟡 **[minor] Stale documentation: receipts.md still describes the removed original_postings field and FSM consuming receipt postings**

After this PR, the RevertTransactionOrder no longer carries an `original_postings` field and the FSM sources postings from TransactionState rather than the receipt. However, the related receipts.md documentation still states that the receipt 'carries an optional original_postings field,' that 'the FSM reads the postings from Pebble,' and that 'the FSM consumes them [the receipt's Postings] at revert time.' CLAUDE.md's Documentation Maintenance rule requires documentation be updated in the same change.

**Suggestion:** Rewrite the receipts.md 'The revert path' section to describe TransactionState-sourced postings and the receipt's new admission-only role (declaring volume coverage). Remove all references to original_postings being carried in the receipt or consumed by the FSM.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1614#issuecomment-5035006146

flemzord (changes requested):
Revalidated live head `8d7d0c70`. The new commit fixes the stale receipt documentation, and I continue to withdraw the unreachable same-batch preload scenario. One correctness/runtime blocker remains: the base explicitly uses order-carried `OriginalPostings` as the migration bridge for pre-EN-1242 `TransactionState` records whose `Postings` is empty; this PR removes that bridge and makes the FSM return `ErrTransactionStateInconsistent` for those allocated transactions. That is not a reserved-field or pre-stable wire objection: persisted transactions become impossible to revert after upgrade. Preserve a deterministic fallback or provide a migration before removing the bridge. The remaining E2E check is still running; all completed main checks are green.

Azorlogh:
> persisted transactions become impossible to revert after upgrade

@flemzord As stated [in the deployment docs](https://github.com/formancehq/ledger/blob/d9bd01821e9b1cfea9de0ab33e5c940acdfb9738/docs/ops/deployment.md?plain=1#L649), there are no persisted transactions to revert after an upgrade because everything gets wiped.

> The service is not yet stable, so this break is accepted instead of adding migration code. **The only supported upgrade path is a full data wipe**:

docs/technical/architecture/audit-vs-technical-state.md
Azorlogh: Thanks, fixed in https://github.com/formancehq/ledger/pull/1614/commits/8d7d0c7085302a5d107dc33c3c22e0f7bc4b0b57

flemzord (approved):
Revalidated live head `8d7d0c70` after the author clarified the supported upgrade contract. I withdraw my remaining blocker: the live deployment documentation explicitly requires a full data wipe for the supported pre-stable upgrade path, so pre-existing `TransactionState` compatibility is non-blocking under this review policy. The same-batch preload scenario was already invalidated; the current code and docs have no remaining correctness/runtime blocker. Main CI, NumaryBot, and CodeRabbit are green.

paul-nicolas (approved):

PR #1616, opened 2026-07-21
fix(EN-1558): keep technical execution data out of audit hashes

paul-nicolas:
## Summary

The audit hash chain bound more than accepted user intent: `marshalOrdersForAudit` serialized the **whole** `raftcmdpb.Order` — including its `OrderTechnical` sub-message (`coverage_bits`, `inputs_resolution_hash`, `preload_unavailable`) — into `AuditItem.serialized_order`, so admission-derived execution-planning metadata became audit-hash-bound. Meanwhile the idempotency hash (`hashOrder`) already excluded `OrderTechnical`, creating an asymmetry between the two notions of "the order."

This PR collapses the two notions into a single source of truth: a shared `processing.MarshalOrderBusinessIntent(order, buf) []byte` that returns the order's deterministic bytes with `OrderTechnical` nilled. Both `hashOrder` (idempotency) and `marshalOrdersForAudit` (audit) now use it, so `AuditItem.serialized_order` and the audit hash chain bind **business intent only**.

## Key properties

- **Byte-exact verification preserved** — the hash is still computed over the exact stored `serialized_order` bytes; verification never re-marshals an `Order`.
- **Idempotency alignment holds by construction** — `recomputeProposalHash` unmarshals the stored bytes and re-hashes via `HashOrders` (which also nils `Technical`), so it stays byte-identical to the frozen idempotency hash.
- **No migration** — audit entries persisted before this change still embed `OrderTechnical` in their stored bytes and continue to verify verbatim; the verifier hashes stored bytes as-is. No hash-version bump.

## Changes

- `refactor`: extract shared `processing.MarshalOrderBusinessIntent`; `hashOrder` now uses it (behavior-preserving).
- `fix`: `marshalOrdersForAudit` stores business-intent bytes (strips `OrderTechnical`).
- `test`: pin that `coverage_bits` does not change the audit per-item payload; prove idempotency/audit-recompute alignment with `coverage_bits` populated; prove legacy (pre-change) bytes still verify.
- `docs`: `AuditItem.serialized_order` proto comment, `CLAUDE.md` invariant #10, and `docs/technical/architecture/audit-vs-technical-state.md`.

## Acceptance criteria (EN-1558)

- [x] Technical execution fields do not participate in the hash of business intent.
- [x] Idempotency hashing and audit-derived recomputation stay aligned.
- [x] Tests prove changing `coverage_bits` does not change the audit business-intent hash.
- [x] The invariant is documented in the relevant architecture/invariant docs.

## Verification

- Affected suites green: `internal/infra/state`, `internal/domain/processing`, and all five `serialized_order` consumers (`check`, `usagebuilder`, `auditindexer`, `backup`).
- `just pre-commit`: `go generate` + `go mod tidy` clean, `golangci-lint` 0 issues.

## Scope note

Payload-embedded state-derived data (e.g. `RevertTransactionOrder.original_postings`) is intentionally **out of scope** here and left for a separate follow-up, as agreed during design.

Jira: https://formance-team.atlassian.net/browse/EN-1558


flemzord (approved):
Reviewed current head `ee6bc974`. Audit serialization and the idempotency hash now share the same deterministic business-intent projection with OrderTechnical excluded; legacy serialized orders remain verified verbatim, and downstream audit consumers only require business fields. Main CI and Codecov are green. CodeRabbit skipped the draft and NumaryBot has not reported a finding. I found no correctness/runtime blocker.

internal/domain/processing/processor.go
```diff
@@ -356,33 +356,50 @@ func HashOrders(orders []*raftcmdpb.Order) []byte {
...
-// one shot means a new technical field can never silently break idempotency.
-func hashOrder(order *raftcmdpb.Order, buf []byte) (hash []byte, grownBuf []byte) {
-	// Temporarily nil the technical sub-message, marshal, then restore it.
-	// Avoids a full CloneVT of the order.
+// out is marshalled into buf[:0] (grown as needed) and returned; reuse it as buf
+// on the next call to amortize allocations. Pass nil to allocate a fresh slice
+// the caller can retain (the audit path stores one slice per order). Single-
+// threaded apply only: it transiently nils order.Technical, marshals, then
+// restores, leaving the live order byte-identical — mirroring the historical
+// hashOrder behaviour and safe because the FSM apply path is the sole caller.
```
Azorlogh, line 384: Non-blocking: "the FSM apply path is the sole caller" isn't accurate — `HashOrders` reaches this too, from `recomputeProposalHash` (checker + backup restore), off the hot path. It's still safe there since those are freshly-unmarshalled local orders, but the thread-safety rationale should be "caller holds the order exclusively for the marshal window," not "single caller."

Azorlogh (approved):
LGTM. Two non-blocking doc nits below (both comment-only, no code bugs).

**`Order.technical` proto comment** (`misc/proto/raft_cmd.proto:35`, mirrored in `raft_cmd.pb.go`): stale after this change — still says only `hashOrder` excludes the sub-message from the idempotency hash. It's now excluded from the audit serialization too, via the shared `MarshalOrderBusinessIntent`. Worth a mention here since you updated the equivalent comment in `audit.proto`. (That file isn't in this diff, hence the note here rather than inline.)

PR #1617, opened 2026-07-21
feat(operator): support multiple DNSEndpoints per Cluster

sylr:
## What

Changes the Cluster CRD field `dnsEndpoint` (a single `*DNSEndpointSpec`) into `dnsEndpoints` (a slice), so distinct ExternalDNS endpoints — e.g. a **public** and a **private** one — can each carry their own annotations.

```yaml
dnsEndpoints:
  - name: private
    enabled: true
    annotations:
      service.beta.kubernetes.io/aws-load-balancer-internal: "true"
    endpoints: [...]
  - name: public
    enabled: true
    annotations: { ... }   # different annotations
    endpoints: [...]
```

## How

- **API** (`cluster_types.go`): `DNSEndpoint *DNSEndpointSpec` → `DNSEndpoints []DNSEndpointSpec`, annotated `+listType=map` / `+listMapKey=name` (K8s-enforced name uniqueness). New required `name` field, validated as a DNS-1123 label (`MinLength=1`, `MaxLength=63`, pattern) since it becomes the suffix of the generated object's `metadata.name`.
- **Naming** (`names.go`): new `dnsEndpointName(crName, epName)` → `ledger-<cr>-<name>` (e.g. `ledger-foo-public`).
- **Reconcile** (`reconcile_dnsendpoint.go`): creates one `DNSEndpoint` object per enabled entry that has endpoints. New `pruneDNSEndpoints` lists objects by the **non-overridable** `managed-by` label and deletes only those this Cluster **controls** (owner reference) and no longer desires — never on a label match alone (the `instance` label is user-overridable via `spec.additionalLabels`).
- `hash.go` unchanged in intent: `dnsEndpoints` stays excluded from the pod-template spec hash (does not roll the StatefulSet).
- Regenerated `zz_generated.deepcopy.go` + CRD (`config/crd/bases` and synced `helm/crds`).

## Review

Reviewed by Codex; addressed the owner-reference prune safety, the DNS-1123 name validation, and removed a redundant pre-`Get`.

## ⚠️ Breaking change / migration

This is a **hard field rename** with no back-compat shim. On upgrade, an existing Cluster still carrying the old `spec.dnsEndpoint` deserializes with an empty slice, so its previously-created `ledger-<cr>` DNSEndpoint object is **pruned**. Existing CRs must be migrated to `dnsEndpoints` (and the object is renamed to `ledger-<cr>-<name>`). Say the word if a deprecation shim is preferred instead.

## Verification

- `just generate` ✅
- `go build ./...` ✅
- `go test ./internal/controller/ ./api/...` ✅

## Not tested

No dedicated `reconcile_dnsendpoint` tests yet (multi-entry create, annotation retention, disable/rename prune, owner-reference deletion guard).

### DIFF
--- a/misc/operator/api/v1alpha1/cluster_types.go
+++ b/misc/operator/api/v1alpha1/cluster_types.go
@@ -312,9 +312,14 @@ type ClusterSpec struct {
 	// +optional
 	NetworkPolicy *NetworkPolicySpec `json:"networkPolicy,omitempty"`
 
-	// DNSEndpoint configuration for ExternalDNS.
+	// DNSEndpoints configures one or more ExternalDNS DNSEndpoint resources.
+	// Each entry is reconciled into its own DNSEndpoint object, so distinct
+	// endpoints (e.g. a public and a private one) can carry different
+	// annotations. Entries are keyed by their unique name.
 	// +optional
-	DNSEndpoint *DNSEndpointSpec `json:"dnsEndpoint,omitempty"`
+	// +listType=map
+	// +listMapKey=name
+	DNSEndpoints []DNSEndpointSpec `json:"dnsEndpoints,omitempty"`
 
 	// LivenessProbe overrides the default liveness probe for the ledger container.
 	// +optional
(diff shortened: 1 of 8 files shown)


flemzord (approved):
Reviewed current head `062f15de`. The multi-endpoint reconciliation keeps names unique, applies each desired object, and prunes only DNSEndpoints controlled by this Cluster; the generated CRD/deepcopy updates match the API change. The documented v1alpha1 field rename is pre-stable compatibility, not a runtime blocker. Main CI is green.

PR #1620, opened 2026-07-22
fix(backup): drop hardcoded "backups/" segment from S3 keys

sylr:
## What

Removes the redundant hardcoded `backups/` path segment from backup S3 object keys.

Since `bucketID` already namespaces every object under the destination S3 bucket, the extra `backups/` middle segment added nothing. Keys are now flat under the `bucketID` prefix:

| Object | Before | After |
|---|---|---|
| Manifest | `<bucketID>/backups/manifest.json` | `<bucketID>/manifest.json` |
| Checkpoints | `<bucketID>/backups/data/…` | `<bucketID>/data/…` |
| Exports | `<bucketID>/backups/exports/…` | `<bucketID>/exports/…` |

Also routes `bootstrap.go` through `backup.ManifestKey()` instead of re-deriving the manifest key inline (DRY).

## Files

- `internal/infra/backup/segment.go` — 6 key builders (`ExportLogSegmentKey`, `ExportAuditSegmentKey`, `ExportAuditItemSegmentKey`, `ExportAppliedProposalSegmentKey`, `CheckpointPrefix`, `ExportPrefix`)
- `internal/infra/backup/manifest.go` — `ManifestKey`
- `cmd/ledgerctl/store/bootstrap.go` — use `backup.ManifestKey()` helper
- `internal/infra/backup/manifest_test.go` — updated round-trip literal to new layout

## ⚠️ Breaking layout change

Backups already written under the old `.../backups/...` keys are **not** found by this code. Fine for greenfield; needs a migration story if pre-existing backups must be read.

## Not touched

Cold storage shares the same `bucketID` root but keeps its own `chapters/` segment (`<bucketID>/chapters/<chapterID>/archive.sst`). Left as-is — no collision since the sub-prefixes differ. Can follow up for symmetry if desired.

## Test plan

- [x] `go build ./internal/infra/backup/... ./cmd/ledgerctl/...`
- [x] `go test ./internal/infra/backup/...` (green)
- [ ] Reading a backup written by a pre-change client (intentionally not covered — breaking change)

### DIFF
--- a/internal/adapter/grpc/server_restore_download.go
+++ b/internal/adapter/grpc/server_restore_download.go
@@ -216,7 +216,7 @@ func (s *RestoreServiceServerImpl) prepareDownload(
 		bucketID = s.clusterID
 	}
 
-	manifestKey := bucketID + "/backups/manifest.json"
+	manifestKey := backup.ManifestKey(bucketID)
 
 	manifestReader, err := storage.GetFile(ctx, manifestKey)
 	if err != nil {
--- a/internal/infra/backup/manifest.go
+++ b/internal/infra/backup/manifest.go
@@ -187,7 +187,7 @@ func WriteManifest(ctx context.Context, storage Storage, key string, manifest *M
 
 // ManifestKey returns the S3 key for the manifest file.
 func ManifestKey(bucketID string) string {
-	return bucketID + "/backups/manifest.json"
+	return bucketID + "/manifest.json"
 }
 
 // ReadManifestOrEmpty reads a manifest from storage, returning an empty
--- a/cmd/ledgerctl/store/bootstrap.go
+++ b/cmd/ledgerctl/store/bootstrap.go
@@ -90,7 +90,7 @@ func runBootstrap(cmd *cobra.Command, _ []string) error {
 		bucketID = "default"
 	}
 
-	manifestKey := bucketID + "/backups/manifest.json"
+	manifestKey := backup.ManifestKey(bucketID)
 
 	// Read manifest
 	spinner, _ := pterm.DefaultSpinner.Start("Reading backup manifest...")
--- a/cmd/ledgerctl/store/bootstrap.go
+++ b/cmd/ledgerctl/store/bootstrap.go
@@ -90,7 +90,7 @@ func runBootstrap(cmd *cobra.Command, _ []string) error {
 		bucketID = "default"
 	}
 
-	manifestKey := bucketID + "/backups/manifest.json"
+	manifestKey := backup.ManifestKey(bucketID)
 
 	// Read manifest
 	spinner := cmdutil.StartSpinner("Reading backup manifest...")
--- a/internal/infra/backup/manifest_test.go
+++ b/internal/infra/backup/manifest_test.go
@@ -43,14 +43,14 @@ func TestReadManifest_CurrentFormatDecodes(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	storage := NewMockStorage(ctrl)
 
-	current := `{"checkpoint":{"timestamp":"t","lastAppliedIndex":1,"lastLogSequence":1,"lastAuditSequence":1,"files":{"000001.sst":{"size":123,"key":"b/backups/data/000001.sst.abc"}}},"exports":null}`
+	current := `{"checkpoint":{"timestamp":"t","lastAppliedIndex":1,"lastLogSequence":1,"lastAuditSequence":1,"files":{"000001.sst":{"size":123,"key":"b/data/000001.sst.abc"}}},"exports":null}`
 	storage.EXPECT().GetFile(gomock.Any(), "k").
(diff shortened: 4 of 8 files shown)


internal/infra/backup/manifest.go
```diff
@@ -187,7 +187,7 @@ func WriteManifest(ctx context.Context, storage Storage, key string, manifest *M
 
 // ManifestKey returns the S3 key for the manifest file.
 func ManifestKey(bucketID string) string {
-	return bucketID + "/backups/manifest.json"
+	return bucketID + "/manifest.json"
```
NumaryBot, line 190: 🔴 **[blocker] Use ManifestKey in restore downloads**

When backups are written with this new manifest path, `RestoreServiceServerImpl.prepareDownload` still requests `bucketID + "/backups/manifest.json"` in `internal/adapter/grpc/server_restore_download.go:219`. A backup created by the updated manager is therefore unavailable to the gRPC restore download path unless a legacy manifest is also present, so that path should use `backup.ManifestKey(bucketID)` too.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1620#issuecomment-5043189892

paul-nicolas (dismissed):

flemzord (changes requested):
Reviewed current head `d08a95d7`. The existing NumaryBot restore-path finding is an active correctness/runtime blocker: backup writers now store the manifest at `<bucket>/manifest.json`, while `prepareDownload` still reads `<bucket>/backups/manifest.json`, so newly created backups cannot be downloaded through the gRPC restore path. I am not duplicating the inline finding. Main CI is otherwise green.

sylr:
Fixed in 50af69d3. `prepareDownload` in the gRPC restore service (`internal/adapter/grpc/server_restore_download.go:219`) still hand-rolled the old `<bucketID>/backups/manifest.json` key — good catch, that was the one writer/reader mismatch. It now routes through `backup.ManifestKey()` like every other call site.

Swept the rest of the tree: `grep -rn "/backups" --include="*.go"` (non-test) is now empty, so all manifest-key producers/consumers share the helper. `go build` + `go test ./internal/adapter/grpc/... ./internal/infra/backup/...` are green.

Note the breaking-layout caveat from the PR description still stands: backups written under the old `.../backups/...` keys are not readable by this code.

internal/infra/backup/manifest.go
NumaryBot, line 190: 🟠 **[major] E2E backup/restore tests still use the old 'backups/' key layout**

Since ManifestKey now returns '<bucketID>/manifest.json' (dropping the hardcoded 'backups/' segment), the S3 and Azure E2E test helpers still fetch 'test-cluster/backups/manifest.json' and check paths under '.../backups/data/'. Running those suites with '-tags e2e,s3' or '-tags e2e,azure' will produce 404s and test failures because the fixtures no longer match the new key layout.

**Suggestion:** Update the E2E backup/restore test helpers (S3 and Azure variants) to use the backup key helpers (e.g. ManifestKey) instead of hardcoded paths, so the test fixtures exercise the new layout consistently.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1620#issuecomment-5043189892

flemzord (dismissed):
Revalidated live head `50af69d3`. The prior runtime blocker is fixed: the gRPC restore-download path now uses the shared `backup.ManifestKey` and matches the new writer layout. The remaining NumaryBot finding is confined to stale S3/Azure E2E helper constants, so it is CI/tooling rather than a runtime blocker; main CI is green. The explicitly documented old-layout break is pre-stable compatibility and is non-blocking under the review criteria.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1620#issuecomment-5043189892

internal/adapter/grpc/server_restore_download.go
```diff
@@ -216,7 +216,7 @@ func (s *RestoreServiceServerImpl) prepareDownload(
 		bucketID = s.clusterID
 	}
 
-	manifestKey := bucketID + "/backups/manifest.json"
+	manifestKey := backup.ManifestKey(bucketID)
```
gfyrag, line 219: [P2][blocking] Keep the restore lifecycle test aligned with the new manifest key

The current merge candidate times out in `TestRestoreDownloadRemainsAsynchronousAndSucceeds`: the test added on `release/v3.0` still detects the manifest request with `strings.HasSuffix(key, "/backups/manifest.json")`. This line now requests `<bucketID>/manifest.json`, so the mock never releases the manifest, the unit-test job hits its 20-minute timeout, and the required Coverage job then fails because `internal.out` is missing. Please update/rebase the branch and change that lifecycle fixture to recognize the new manifest key (preferably by comparing with `backup.ManifestKey("test-cluster")`), then rerun the required checks.

PR #1622, opened 2026-07-22
fix(EN-1557): classify numscript admission failures by state provenance

paul-nicolas:
## Summary

Admission classifies Numscript dependency-resolution *preparation* failures by **state provenance** to decide whether to terminate or forward to the FSM — replacing the previous behavior where every unrecognised failure collapsed to `KindInternal` and, under an idempotency key, was forwarded as retryable `PRELOAD_UNAVAILABLE`.

**The bug:** a *deterministic* failure on an inline script or an exact immutable version (one that can never succeed) with an idempotency key was forwarded as `PRELOAD_UNAVAILABLE`. The FSM found no frozen outcome and rejected with the retryable reason; the client retried; admission failed identically and forwarded again — an **unbounded retry loop**.

**The fix:** classify from two signals Ledger already owns — never from error strings or Numscript internals:
1. **Selector mutability** — `latest` may resolve to a different, previously-saved version, so it stays forwardable under a key; inline scripts and exact versions are deterministic.
2. **Read-attempt provenance** — whether resolution attempted a mutable balance/metadata lookup before failing (including a lookup that itself errored). Exposed via `RecordingStore.MutableReadAttempted()` (set before delegating to the inner store) and carried out of discovery through the typed `DependencyResolutionError`. The pre-existing `ReadNothing()` was insufficient — it reflects only successfully-recorded values.

Deterministic inline/exact no-read failures now terminate (surfacing the real cause); `latest` no-read failures and state-dependent failures stay forwardable under a key. Recovered panics and context cancellation/deadline are surfaced loudly and never softened (invariant #7). No accepted-order business mutation (only `OrderTechnical.PreloadUnavailable`, invariant #10-exempt); no FSM coverage-gate change; no new persisted projection.

EN-1563 (a proposed public Numscript resolver-error taxonomy) is **not required** and cancelled.

## Changes
- `RecordingStore` gains a `MutableReadAttempted()` marker set before any inner balance/metadata lookup.
- New `DependencyResolutionError{Cause, MutableReadAttempted}` (`Unwrap`-transparent); `DiscoverNumscriptDependencies` wraps its resolve-error return with it.
- `forwardOrFail` extracted into a directly-testable `Admission.classifyResolutionFailure`, rewritten with ordered guards (panic → cancellation → freezable → provenance).
- Stale "public classification required" comment replaced; provenance-based classification documented under `docs/technical/architecture/subsystems/admission/idempotency.md`.

## Testing
- `just pre-commit` (0 lint issues) and `just test` (full unit suite) pass via nix.
- New table-driven tests: `RecordingStore` read-attempt (incl. failed-lookup gap), `DependencyResolutionError`, discovery provenance wrapping, and the full admission decision table (inline/exact/latest × read/no-read × key/no-key, panic both-key, cancellation-with-key, freezable). Regression test pins that an inline deterministic no-read failure with a key terminates and is NOT forwarded.

## Ticket
https://formance-team.atlassian.net/browse/EN-1557


flemzord (approved):
Reviewed live head `93a7b40e`. The provenance wrapper records failed as well as successful mutable-read attempts, and admission keeps panic/cancellation and deterministic inline/exact failures terminal while forwarding only state-dependent or mutable-selector failures when idempotent replay may matter. The mutation stays confined to OrderTechnical, tests cover the decision table, and main CI plus both review bots are green. I found no correctness/runtime blocker.

flemzord (changes requested):
Revalidated live head `93a7b40e`. One correctness/runtime blocker remains in the provenance rule: `MutableReadAttempted` records that any mutable read happened before failure, not that the read caused the failure. In the pinned Numscript, `ResolveDependencies` first binds all variable origins, then walks statements; a successful `balance()`/`meta()` origin can set this flag before a later `SourceWithScaling` deterministically returns `ErrScalingNotSupported`. That error maps to `ErrNumscriptRuntime`, so an inline/exact script with an idempotency key is forwarded as `PRELOAD_UNAVAILABLE` even though no state change can make it succeed—the retry loop EN-1557 aims to stop remains. Please distinguish causal state-dependent failures from deterministic post-read failures (or expose/use an upstream taxonomy) and add a regression for read-success followed by unsupported scaling. Main CI and both review bots are green.

paul-nicolas:
@flemzord good catch — confirmed. `ResolveDependencies` binds var origins before walking statements, so a successful `balance()`/`meta()` origin sets `MutableReadAttempted` before `SourceWithScaling` returns the deterministic `ErrScalingNotSupported`, which mapped to `ErrNumscriptRuntime` (KindInternal, non-freezable) and got forwarded under a key — the loop this PR is meant to stop.

Fixed in d6fbed3df: `ErrScalingNotSupported` is a *public* numscript sentinel (`numscriptlib.ErrScalingNotSupported`) and unconditionally state-independent, so `convertNumscriptError` now maps it to a freezable validation sentinel (`ErrNumscriptScalingUnsupported`, same pattern as the existing `ErrScopedBalanceUnsupported`). The existing freezable branch in `classifyResolutionFailure` then terminates it *before* the provenance branch is consulted, so `MutableReadAttempted` no longer decides scaling — no admission logic change.

Regressions added (the read-success-then-scaling case you asked for):
- `TestConvertNumscriptError_Scaling` — the sentinel converts to a freezable validation error, not `ErrNumscriptRuntime`.
- `TestDiscover_ReadThenScalingIsFreezable` — real resolution path: a successful `balance()` origin read followed by `… with scaling through …` yields a `*DependencyResolutionError` with `MutableReadAttempted == true` whose carried cause is the freezable scaling sentinel.
- `TestForwardOrFail_ProvenanceClassification/inline read-then-scaling with key terminates` — end-to-end through admission: with an idempotency key it now terminates (`*domain.BusinessError`, `PreloadUnavailable == false`) instead of forwarding.

On the broader "read-caused vs deterministic-post-read" split: the genuinely state-dependent case (`MetadataNotFound`) and the other deterministic post-read types (`TypeError`, `UnboundFunctionErr`, …) are all `internal/interpreter`-only and surface identically as `ErrNumscriptRuntime`, so they can't be separated from admission without the upstream taxonomy (the EN-1563 path). Scaling is the one deterministic case the library exposes as a public sentinel, so it's the one we can close cleanly now; the rest stay conservatively forwarded, unchanged from before this PR.

internal/application/admission/admission.go
```diff
@@ -1443,61 +1443,146 @@ func (a *Admission) extractPreloadNeeds(ctx context.Context, orders []*raftcmdpb
...
+	}
+
+	// Provenance-based classification (EN-1557). A DependencyResolutionError
+	// carries whether resolution attempted a mutable balance/metadata read
+	// before failing, letting us tell a state-dependent failure from a
+	// deterministic one that the conservative KindInternal mapping can no
+	// longer distinguish on its own.
+	var dre *numscript.DependencyResolutionError
+	if errors.As(cause, &dre) {
+		if dre.MutableReadAttempted {
```
NumaryBot, line 1556: 🔴 **[blocker] Deterministic failures after a prior state read still cause indefinite retry loops**

When an inline or exact-version script performs any successful balance/meta origin read and then fails later with a deterministic resolver error (e.g. a mistyped or missing variable during `send` statement walking), `MutableReadAttempted` is set to true by the earlier read. The existing no-read guard therefore does not apply, and the failure is forwarded under an idempotency key. Because the admitted order can never produce a frozen outcome, the FSM returns a non-frozen `PRELOAD_UNAVAILABLE` state and the retry loop runs indefinitely — the exact failure mode this PR is intended to stop.

**Suggestion:** Deterministic sentinel errors (e.g. those mapping to `ErrNumscriptRuntime` for known-bad resolver paths) should be classified as freezable validation failures unconditionally, regardless of whether `MutableReadAttempted` is true. Only genuinely state-dependent failures — those that could resolve differently with fresh state — should consult the `MutableReadAttempted` flag before deciding whether to forward.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1622#issuecomment-5044113526

flemzord (changes requested):
Revalidated live head `e1df77b7`. The read-then-scaling case is fixed, but the existing NumaryBot thread on this same head remains a correctness/runtime blocker: `MutableReadAttempted` proves only that a mutable read preceded the failure, not that it caused it. Other deterministic statement errors still collapse to `ErrNumscriptRuntime` and, after a successful balance/meta read, are forwarded under an idempotency key as non-frozen `PRELOAD_UNAVAILABLE` indefinitely. I am not duplicating the inline finding. The separate `Dirty`/`Tests` typecheck failures are CI/tooling, not the basis of this request for changes.

internal/application/admission/admission.go
paul-nicolas: Acknowledged, and the mechanism is real: `MutableReadAttempted` records that an origin `balance()`/`meta()` read happened before the failure (origins are bound before statements are walked), not that it *caused* it — so a deterministic post-read error that collapses to `ErrNumscriptRuntime` (KindInternal) is still forwarded under an idempotency key.

This residual is a documented, deliberate tradeoff, not an oversight:

- It is **safe** — the FSM returns the correct non-frozen `PRELOAD_UNAVAILABLE`, no state divergence — and a **strict improvement** over pre-PR behavior: before this change every KindInternal-under-key failure forwarded (including no-read deterministic ones); now only the read-then-deterministic-error subset does.
- The one deterministic case the numscript library exposes as a public sentinel (`ErrScalingNotSupported`) is already terminated *before* the provenance branch (`d6fbed3df`).
- Splitting the remaining `internal/interpreter`-only errors (`TypeError`, `UnboundFunctionErr`, …) would require the upstream resolver-error taxonomy (EN-1563), which was evaluated and **cancelled**; the design deliberately forbids classifying on error strings or numscript internals. This is documented in `docs/technical/architecture/subsystems/admission/idempotency.md` and `internal/domain/processing/numscript/errors.go`.

A hard bound on non-frozen `PRELOAD_UNAVAILABLE` retries would be a separate FSM-level change on its own ticket, not part of EN-1557.

Separately, the `Dirty`/`Tests` failures were a stale-merge issue, not head: base `release/v3.0` (#1614) added an `overlay *bulkOverlay` param to `extractPreloadNeeds` that the branch's new test call site lacked. Fixed by merging base and threading the overlay (`cea2bc79c`); local pre-commit and unit tests are green.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1622#issuecomment-5044113526

internal/domain/processing/numscript/store.go
```diff
@@ -165,6 +167,8 @@ func NewRecordingStore(inner numscriptlib.Store) *RecordingStore {
 }
 
 func (s *RecordingStore) GetBalances(ctx context.Context, query numscriptlib.BalanceQuery) (numscriptlib.Balances, error) {
```
NumaryBot, line 172: 🟠 **[major] Forced balance queries incorrectly set MutableReadAttempted, causing indefinite retry loops**
<sub>reported by codex, NumaryBot, paul-nicolas</sub>

When `force=true`, `NewStore.GetBalances` returns `MaxForceBalance` without consulting the underlying `ValueSource`, so no mutable balance state is actually read. Despite this, the code sets `MutableReadAttempted = true`. This means a forced script with a `balance()` origin followed by a deterministic, state-independent resolver error (e.g. `ErrScalingNotSupported` or a mistyped variable) appears state-dependent to `classifyResolutionFailure`. The failure is forwarded under an idempotency key, the FSM never produces a frozen outcome, and the retry loop runs indefinitely — the exact failure mode this PR targets.

**Suggestion:** Only set `MutableReadAttempted = true` when an actual mutable state lookup is performed (i.e., `force=false`). Guard the flag assignment so that forced balance queries, which bypass the real `ValueSource`, do not trigger the provenance branch in `classifyResolutionFailure`. Deterministic sentinel errors should be classified as freezable validation failures unconditionally, independent of the `MutableReadAttempted` flag.

flemzord (changes requested):
Revalidated current head b96f9c59 after the release/v3.0 merge and lint-only follow-up. The existing NumaryBot thread at internal/application/admission/admission.go:1543 remains an active correctness/runtime blocker: MutableReadAttempted still proves only that a mutable read preceded the failure, so deterministic post-read resolver errors that remain KindInternal are forwarded as non-frozen PRELOAD_UNAVAILABLE indefinitely under an idempotency key. The new commits do not change that classifier. I am not duplicating the existing inline finding; pending CI is not the basis of this review.

internal/domain/processing/numscript/store.go
paul-nicolas: Fixed in `658fa5de5`. Confirmed the bug: `RecordingStore.GetBalances` set `readAttempted` unconditionally, but under force the inner `Store` short-circuits to `MaxForceBalance` without consulting the `ValueSource` (`store.go:82-83`), so a forced balance query reads no mutable state — marking it made a forced `balance()` origin followed by a deterministic error look state-dependent and forwarded it under a key.

The `force` flag is now threaded into `RecordingStore`, and the balance read-attempt is gated on `!force`. Metadata reads consult real state even under force (`GetAccountsMetadata` always hits the source), so they keep marking unconditionally. `Hash()`/`ReadNothing()` are unaffected (they key off recorded values, not the flag), so the FSM stale-input check is unchanged.

Regressions added in `store_test.go`: `forced balance read does not count` (flag stays false) and `forced metadata read still counts`. Doc updated in `docs/technical/architecture/subsystems/admission/idempotency.md`.

Note this is the state-independent-read half of the finding; the second suggestion (classify deterministic sentinels freezable unconditionally) is the separate residual tracked in the sibling thread — scaling is already terminated pre-provenance, and the remaining internal-only-typed errors need the cancelled EN-1563 taxonomy.

internal/domain/processing/numscript/store.go
```diff
@@ -152,19 +152,36 @@ type RecordingStore struct {
...
+func NewRecordingStore(inner numscriptlib.Store, force bool) *RecordingStore {
 	return &RecordingStore{
 		inner:           inner,
+		force:           force,
 		balanceRecords:  map[string]string{},
 		metadataRecords: map[string]string{},
 	}
 }
 
 func (s *RecordingStore) GetBalances(ctx context.Context, query numscriptlib.BalanceQuery) (numscriptlib.Balances, error) {
```
NumaryBot, line 172: 🟠 **[major] Forced balance queries incorrectly set MutableReadAttempted, causing indefinite retry loops**
<sub>reported by codex, NumaryBot, paul-nicolas</sub>

When `force=true`, `NewStore.GetBalances` returns `MaxForceBalance` without consulting the underlying `ValueSource`, so no mutable balance state is actually read. Despite this, `MutableReadAttempted` is set to `true`. This means a forced script with a `balance()` origin followed by a deterministic, state-independent resolver error appears state-dependent to `classifyResolutionFailure`. The failure is forwarded under an idempotency key, the FSM never produces a frozen outcome, and the retry loop runs indefinitely — the exact failure mode this PR targets.

**Suggestion:** Only set `MutableReadAttempted = true` when an actual mutable state lookup is performed (i.e., `force=false`). Guard the flag assignment so that forced balance queries, which bypass the real `ValueSource`, do not trigger the provenance branch in `classifyResolutionFailure`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1622#issuecomment-5044113526

internal/application/admission/admission.go
```diff
@@ -1449,61 +1449,146 @@ func (a *Admission) extractPreloadNeeds(ctx context.Context, orders []*raftcmdpb
...
+	// (ErrNumscriptScalingUnsupported), a missing ledger, and so on — so there
+	// is no frozen outcome to replay and re-running always fails identically.
+	// Surface the real error rather than forwarding it as a retryable
+	// preload-unavailable (which would spin forever, since no retry can
+	// succeed). IsFreezableFailure captures exactly this "definitive &
+	// deterministic" class (validation, parse, not-found, already-exists,
+	// conflict, precondition). Scaling reaches this branch even when a prior
+	// successful balance()/meta() read set MutableReadAttempted, because the
+	// freezable check runs before the provenance branch (EN-1557).
+	var d domain.Describable
```
NumaryBot, line 1531: 🟠 **[major] Freezable branch bypasses latest-ref idempotency replay path**

When an order uses a `numscriptReference` with `version: latest` and an idempotency key, the new freezable shortcut introduced for `ErrScalingNotSupported` runs before the `refIsLatest` forwarding logic. A retry whose original attempt succeeded on an older `latest` version will now be terminated immediately if a newer `latest` version hits unsupported scaling, preventing the FSM from replaying the frozen outcome. The freezable classification should not short-circuit the mutable-script-content replay path.

**Suggestion:** Ensure the freezable branch only terminates the request after the `refIsLatest`/idempotency replay check has been consulted, or add an explicit guard that skips the terminal freezable classification when the script reference is `latest` and an idempotency key is present.

internal/domain/processing/numscript/store.go
paul-nicolas: Already fixed in `658fa5de5` (pushed just before this re-review — the diff hunk above shows the fix in place). `RecordingStore` now carries the `force` flag and `GetBalances` gates the read-attempt on `!force`, so a forced balance query no longer sets `MutableReadAttempted`. Regressions `forced balance read does not count` / `forced metadata read still counts` in `store_test.go` pin it. Resolving as addressed.

internal/application/admission/admission.go
paul-nicolas: Good catch — fixed in `5ebc6b1d2`. The freezable branch was terminating before the `refIsLatest` forward branch, which is only sound for immutable content (inline / exact version). For a `latest` reference under an idempotency key the resolved content can differ from the version that froze an outcome on the original attempt (the idempotency hash binds the reference `name`+`latest`, not the resolved content), so terminating denied the FSM its replay.

The freezable branch now skips terminal classification when `refIsLatest && hasIdempotencyKey`, letting those fall through to the forward/replay path; inline/exact scripts and keyless `latest` keep the precise terminal error (so the read-then-scaling fix for inline/exact is untouched). Added `classifyResolutionFailure` regressions: `freezable latest ref with key forwards for replay` and `freezable latest ref without key still terminates`.

docs/technical/architecture/subsystems/admission/idempotency.md
```diff
@@ -63,6 +63,44 @@ type IdempotencyKeyValue struct {
...
+
+### Decision flow
+
+```mermaid
+flowchart TD
+    FAIL[Dependency discovery fails] --> PANIC{Recovered<br/>Numscript panic?}
+    PANIC -->|Yes| LOUD1[Surface loudly<br/>never forwarded]
+    PANIC -->|No| CANCEL{Context canceled /<br/>deadline exceeded?}
+    CANCEL -->|Yes| LOUD2[Surface loudly<br/>never forwarded]
+    CANCEL -->|No| FREEZE{Freezable deterministic<br/>rejection?<br/>parse / validation / not-found / …}
```
NumaryBot, line 83: 🟡 **[minor] Flowchart incorrectly shows freezable errors always terminate before latest-ref replay check**

The flowchart depicts every freezable rejection as terminating before the selector/key handling step. However, the implementation intentionally falls through to the refIsLatest/idempotency replay logic when the script reference is 'latest' and an idempotency key is present, so that the FSM can replay a previously frozen outcome. This contradiction could lead a future maintainer to reintroduce the latest-ref idempotency regression that was explicitly fixed in this PR.

**Suggestion:** Add a decision node to the flowchart after the freezable check: 'Is ref=latest AND idempotency key present?' → yes → proceed to replay path; no → terminate with freezable rejection. Alternatively, add a prose note clarifying the exception.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1622#issuecomment-5044113526

docs/technical/architecture/subsystems/admission/idempotency.md
paul-nicolas: Fixed in `c686a2ae1`. The flowchart now routes a freezable rejection through a `Selector is 'latest'/empty AND idempotency key present?` decision: yes → forward for replay, no → terminate. The surrounding prose is updated to match (the "likewise terminal" line and the scaling paragraph now carry the latest-ref carve-out).

docs/technical/architecture/subsystems/admission/idempotency.md
```diff
@@ -63,6 +63,46 @@ type IdempotencyKeyValue struct {
...
```
NumaryBot, line 83: 🟡 **[minor] Flowchart incorrectly shows freezable errors always terminate before latest-ref replay check**

The flowchart depicts every freezable rejection as terminating before the selector/key handling step. However, the implementation intentionally falls through to the refIsLatest/idempotency replay logic when the script reference is 'latest' and an idempotency key is present. This contradiction could mislead a future maintainer into re-introducing the latest-ref idempotency regression that was explicitly fixed in this PR.

**Suggestion:** Add a decision node to the flowchart after the freezable check: 'Is ref=latest AND idempotency key present?' → yes → proceed to replay path; no → terminate with freezable rejection. Alternatively, add a prose note clarifying the exception.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1622#issuecomment-5044113526

docs/technical/architecture/subsystems/admission/idempotency.md
paul-nicolas: Already addressed in `c686a2ae1` (present on the current head): the flowchart now has a `Selector is 'latest'/empty AND idempotency key present?` decision after the freezable check — yes → forward for replay, no → terminate — and the prose carries the same carve-out. This re-review ran against the prior head before that commit. Resolving as addressed.

flemzord (changes requested):
Revalidated current head c686a2ae. The three new commits fix forced-balance provenance, latest-ref replay, and the flowchart, but the existing NumaryBot thread at internal/application/admission/admission.go:1556 remains a correctness/runtime blocker: MutableReadAttempted still records only that a mutable read preceded the failure, so deterministic post-read KindInternal failures for inline/exact scripts with an idempotency key are forwarded as non-frozen PRELOAD_UNAVAILABLE indefinitely. I am not duplicating the inline finding. CI is green.

paul-nicolas:
@flemzord — you are right on the mechanism, and I want to be precise about what remains and why I am asking to unblock here.

**The residual is real and exactly as you describe.** `MutableReadAttempted` records that a mutable `balance()`/`meta()` origin read *preceded* the failure, not that it *caused* it (origins are bound before statements are walked). So for an inline/exact script with an idempotency key, a deterministic post-read error that stays `ErrNumscriptRuntime` (KindInternal) — `TypeError`, `UnboundFunctionErr`, etc. — is still forwarded as non-frozen `PRELOAD_UNAVAILABLE`.

**Why I consider it shippable rather than a merge blocker:**
- It is **safe**: the FSM returns the correct non-frozen `PRELOAD_UNAVAILABLE`, no state divergence.
- It is a **strict improvement** over pre-PR behavior: before this change *every* KindInternal-under-key failure forwarded (including the no-read deterministic ones); now only the read-then-deterministic-error subset does.
- The one deterministic case the numscript library exposes as a **public sentinel** (`ErrScalingNotSupported`) is already terminated before the provenance branch (`d6fbed3df`).

**Why the clean fix is not in this PR:** separating the remaining `internal/interpreter`-only errors from genuinely state-dependent ones (e.g. `MetadataNotFound`) requires the upstream resolver-error taxonomy (EN-1563), which was evaluated and cancelled — and the design deliberately forbids classifying on error strings or importing numscript internals. A hard bound on non-frozen `PRELOAD_UNAVAILABLE` retries is an FSM-level change that stands on its own ticket, independent of the provenance classification this PR introduces.

The tradeoff is documented in `docs/technical/architecture/subsystems/admission/idempotency.md` and `internal/domain/processing/numscript/errors.go`. If you are aligned that the residual is a follow-up rather than a regression this PR introduces, I would like to unblock and merge; happy to file the FSM-retry-bound ticket as the tracking item.

Azorlogh (approved):

flemzord (approved):

PR #1623, opened 2026-07-22
fix(query): read archived transactions from cold storage in GetTransaction

Azorlogh:
## Problem

After a chapter is archived and its logs are purged from hot storage,
`GetTransaction` returned `NotFound` for any transaction in that chapter — even
though the log is safe in cold storage and `GetLog` can still read it.

`GetTransaction` rebuilds a transaction from its **creation log**
(`assembleTransactionFromState`): the attribute-zone `TransactionState` survives
purge, but it carries no `reference` and no `post_commit_volumes` (added by
EN-1546), so those fields come only from the log. The read used hot-only
`query.ReadLogBySequence`, so once the log was purged the whole read failed.

This was a latent inconsistency, not intended behavior:
- `GetLog` already falls back to cold storage (`ReadLogBySequenceWithCold`).
- `TransactionState.postings` is retained specifically to avoid needing the log.
- `ComputeTransactionReceipt`'s own comment asserts *"the transaction is still
  readable; it just has no receipt"* after archival — but the code returned
  NotFound from the transaction read **before** that branch was ever reached, so
  the branch was dead.

## Fix

`assembleTransactionFromState` and `query.FindTransactionCreationLog` now use
`ReadLogBySequenceWithCold`, so an archived transaction is assembled from the
cold-stored log.

- `buildTransaction` (the single assembly choke point — serves both
  `GetTransaction` and `ListTransactions`) passes `ctrl.coldReader`.
- The **receipt** shares the fallback too: an archived transaction keeps its
  receipt. The only log-derived input to the signature is the creation log's
  chapter id, which is immutable, so the token is byte-identical to the one
  issued while the transaction was hot. (This is why the receipt lookup also
  moved to the cold-aware read rather than staying hot-only.)
- Degrades to hot-only when cold storage is disabled (nil cold reader), so
  checkpoint / restore-mode reads are unaffected.
- `FindTransactionCreationLog` / `ComputeTransactionReceipt` reader params
  widened `dal.PebbleGetter` → `dal.PebbleReader` (the cold chapter lookup
  iterates); the checkpoint receipt path now passes a read handle instead of the
  raw `*dal.Store`.

## Testing

`tests/e2e/business/get_transaction_cold_test.go` (`e2e`, filesystem cold
storage, modeled on `audit_purge_test.go`'s `archiveChapterFull`): a referenced
transaction whose chapter is archived+purged is still returned whole (reference,
postings, post-commit volumes) with its receipt.

Verified it **catches the bug**:
- fix reverted → `FAIL` at "archived transaction must still be readable from
  cold storage" (`NotFound: transaction 1 not found`)
- fix in place → `1 Passed`

Changed-package unit tests (query, ctrl, state, grpc) green; lint clean.

## Notes

Surfaced while planning a chapter-archival model test — this is the server-side
read regression that planning uncovered, split out as its own fix.

### DIFF
--- a/internal/adapter/grpc/server_bucket.go
+++ b/internal/adapter/grpc/server_bucket.go
@@ -349,8 +349,16 @@ func (impl *BucketServiceServerImpl) GetTransaction(ctx context.Context, req *se
 		if impl.receiptSigner != nil {
 			// Checkpoint read: sign from the checkpoint's fixed store so the
 			// receipt's ledger + creation-log reads stay self-consistent with the
-			// transaction, which was read from that same store.
-			receiptToken, cErr := impl.localCtrl.ComputeTransactionReceipt(ctx, mainStore, req.GetLedger(), req.GetTransactionId(), tx)
+			// transaction, which was read from that same store. A read handle (not
+			// the raw store) so the creation-log lookup's cold-storage fallback can
+			// scan chapters.
+			receiptHandle, hErr := mainStore.NewReadHandle()
+			if hErr != nil {
+				return nil, fmt.Errorf("opening checkpoint receipt read handle: %w", hErr)
+			}
+
+			receiptToken, cErr := impl.localCtrl.ComputeTransactionReceipt(ctx, receiptHandle, req.GetLedger(), req.GetTransactionId(), tx)
+			_ = receiptHandle.Close()
 			if cErr != nil {
 				return nil, fmt.Errorf("computing transaction receipt: %w", cErr)
 			}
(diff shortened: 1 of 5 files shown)


flemzord (approved):
Reviewed live head `41fd51fa`. Transaction assembly and receipt lookup now share the existing hot-first, cold-storage-aware log reader, preserve hot-only behavior when cold storage is disabled, and keep checkpoint reads on their fixed store handle. The archive-and-purge E2E test covers the previously lost reference, post-commit volumes, and receipt. Main CI and NumaryBot are green; I found no correctness/runtime blocker.

paul-nicolas (approved):

PR #1631, opened 2026-07-23
fix: harden S3 multipart uploads against CompleteMultipartUpload InvalidPart

sylr:
## Problem

Full S3 backups of a large ledger fail with `CompleteMultipartUpload … InvalidPart: One or more of the specified parts could not be found`. Investigation against the customer's dev bucket showed the bucket/KMS/S3 are healthy for multipart from a clean path — the failure is specific to the client's multipart execution, on files large enough to go multipart.

## Change

- **32 MiB multipart part size** on both the backup (`internal/infra/backup/s3.go`) and cold-storage (`internal/infra/coldstorage/s3.go`) uploaders (SDK default is 5 MiB). Cuts part count ~6.4× for multi-GB objects (smaller failure surface) and lifts the 10,000-part single-object ceiling from ~48.8 GiB to ~312 GiB.
- **`RequestChecksumCalculation=WhenRequired`** on the shared S3 client (`coldstorage.NewS3Client`, reused by the backup client). Recent `aws-sdk-go-v2` defaults to `WhenSupported`, which attaches a trailing CRC via `aws-chunked` encoding on every part — the leading suspect for the `InvalidPart` at Complete, and the older binary that predates the default uploaded fine.

Object integrity is preserved end-to-end without the SDK checksums: checkpoint keys are content-addressed by sha256, and cold-storage archives carry a verified sha256 metadata checksum.

## Verification

- `go build -tags s3 ./...` and `go vet` clean.
- Multipart + trailing-CRC upload verified working against the exact bucket from a clean network path (bucket/KMS are not the problem).
- ⚠️ s3 integration tests (MinIO testcontainer) require Docker → CI.
- ⚠️ Confidence on the checksum lever is medium (hypothesis, not reproduced from a clean path); the part-size change is unambiguously beneficial.

## Notes

- Operational: with 32 MiB parts, a non-seekable upload can buffer ~160 MiB (concurrency × part size), up from ~25 MiB.

PR #1632, opened 2026-07-23
fix(backup): skip re-hashing unchanged files and retry transient uploads

sylr:
## Problem

A full backup of a large store (~2 TiB) (1) re-hashed **every** checkpoint file on **every** run — a ~30–70 min silent, non-cancellable prelude before any upload — and (2) failed the whole run on any single transient upload error (no application-level retry around `PutFile`).

Independent of #1631 (disjoint files); both are needed to get a 2 TiB backup through.

## Change (`internal/infra/backup/`)

**Skip re-hashing unchanged files.** `RunBackup` now reads the previous manifest and lists existing objects *before* enumerating files, and reuses a prior file's content-addressed key instead of re-hashing when ALL hold:
- immutable-by-name (`.sst`/`.blob`) — metadata files (MANIFEST/OPTIONS/CURRENT/marker) are always re-hashed;
- same name + size + **mtime** as the previous manifest — mtime is the store-lineage discriminator (`RestoreCheckpoint` writes fresh files → fresh mtime, so a restored colliding file is correctly re-hashed);
- **the reused key is still present remotely** — otherwise a missing object would be re-uploaded under a stale content hash. This guard is load-bearing.

`CheckpointFile` gains `ModTimeUnixNano` (backward compatible: old manifests → zero → one full re-hash after upgrade, then steady state). Hashing is now context-aware (walk + `hashFile`) so a cancelled backup exits the hashing phase promptly instead of holding the destination lock.

**Bounded upload retry** (`retry.go`). `uploadFile`, the export-segment path, and the manifest write retry with bounded exponential backoff (5 attempts, 500 ms→30 s). A fresh body per attempt via a factory (re-`os.Open` the file / re-seek the Pebble iterator; the segment path joins its writer goroutine before re-seeking since iterators aren't concurrency-safe). All errors retried except context cancellation (checked via `ctx.Err()` and the sentinels).

## Verification

- 8 unit tests (reuse-skip, re-hash on missing-remote / size-change / **mtime-change**, retry succeeds/exhausts/no-retry-on-cancel, **streamed-segment retry round-trip**) + all existing backup tests pass.
- `go build ./...` / `-tags s3 ./...`, `go vet`, `golangci-lint` clean.
- ⚠️ No live multi-TiB run yet; s3 integration tests need Docker (CI).

## Residual risk (accepted)

Reusing a prior hash means the backup no longer *incidentally* re-hashes each `.sst`/`.blob` every run, so it wouldn't catch out-of-band corruption preserving name+size+mtime. Pebble's immutable-file contract precludes this in normal operation; object integrity is covered by content-addressing, the audit hash chain, and the checker (invariant #8). Documented on `reusePriorKey`.

### DIFF
--- a/internal/infra/backup/manager.go
+++ b/internal/infra/backup/manager.go
@@ -1,6 +1,7 @@
 package backup
 
 import (
+	"bytes"
 	"context"
 	"crypto/sha256"
 	"encoding/binary"
@@ -10,6 +11,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/cockroachdb/pebble/v2"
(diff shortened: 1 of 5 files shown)


flemzord:
Reviewed live head `7e1d6207`. The immutable-file reuse is guarded by file type, size, mtime, and remote-key presence; upload retries recreate file/manifest bodies and re-seek streamed export segments before replay. Main build/test checks and NumaryBot are green. The Codecov patch-threshold failure is CI/tooling rather than a demonstrated runtime blocker. I found no correctness/runtime blocker on this head; GitHub approval is intentionally disabled by this cron configuration.

flemzord:
Revalidated live head `56e8de30`. The new commit only adds coverage for retry, hashing-error, and cancellation branches; the production behavior previously reviewed is unchanged. Hash reuse remains guarded by immutable file type, size/mtime, and remote-key presence, and retry bodies are recreated or re-seeked before replay. CI is green and I found no correctness/runtime blocker. GitHub approval is intentionally disabled for this cron task.

Azorlogh (approved):

PR #1633, opened 2026-07-23
fix(query,admission): read archived logs from cold storage across replay / ListLogs / prepared-query paths

Azorlogh:
## Problem

`GetLog` reads a log with a cold-storage fallback (`ReadLogBySequenceWithCold`), so it keeps working after a chapter is archived and its logs are purged from hot storage. Several sibling read paths used the hot-only `ReadLogBySequence` and did **not**, so they broke once a log was archived — even though the log is safe in cold storage and the read-side index still references it:

- **Idempotent replay** — a committed idempotency outcome references its committed log by sequence; resolving that reference for the response was hot-only, so replaying a key whose chapter was archived returned an **empty log (sequence 0)** instead of the committed outcome (an exactly-once violation). The key itself is not purged, so the replay was still served — only the log resolution regressed.
- **`ListLogs` / prepared-query `LOGS`** — the read-side log index still lists an archived log's sequence, but the `ledgerLogCursor` fetched each log hot-only, so a page including an archived log failed the whole call with `log with sequence N not found in Pebble` (surfaced to the client as an `Unknown` gRPC error).

This is the same class as #1623 (which fixed `GetTransaction` and its receipt lookup). This PR sweeps the remaining reachable sites.

## Fix

Both paths now resolve logs via `ReadLogBySequenceWithCold`, degrading to hot-only when cold storage is disabled (nil reader) — and archival can't happen without cold storage, so no referenced/listed log is ever purged in that case.

- **admission** (`fceddbc5b`): the idempotent-replay resolution loop opens a read handle lazily (replay path only — the cold chapter lookup iterates, which the raw store can't) and falls back to cold. `ColdReader` is wired into `Admission` as an **optional** fx dependency (`optional:"true"`, matching the controller provider) via a new `WithColdReader` option.
- **query** (`61c7e8dc4`): the `ledgerLogCursor` falls back to cold; the reader is threaded through `ReadLedgerLogsCompiled` / `EnrichLogs` / `executeList` / `Execute`, supplied from `ctrl.coldReader` at both callers (`ListLogs` and `ExecutePreparedQuery`).

## Not affected (reviewed)

`usagebuilder`'s `readLog` (the third `ReadLogBySequence` grep hit) is a forward-only fold over the **audit** stream, and archival purges audit entries **with** their logs — so a present audit entry's log is also present. A log-read fallback there would address nothing; left unchanged.

## Testing

Two e2e tests (filesystem cold storage, `//go:build e2e`), each verified to **fail without the fix**:

- `tests/e2e/business/idempotency_replay_cold_test.go` — a keyed transaction whose chapter is archived replays the original committed log (not sequence 0). Without the fix: replay returns sequence 0.
- `tests/e2e/business/listlogs_cold_test.go` — after archiving a chapter of logs, `ListLogs` still returns them all. Without the fix: `log with sequence N not found in Pebble`.

Changed-package unit tests (query, ctrl, admission, bootstrap) green; lint clean; full module builds.

## Notes

Surfaced while building chapter-archival coverage for the model test (the model driver replaying idempotency keys across archival hit the admission bug; a probe confirmed the `ListLogs` one). No existing Jira ticket covers this class — closest neighbours are EN-1341 (checker-side archived-idempotency verification) and EN-1463 (chapter-close checkpoint queries), neither of which is this read-fallback defect.

### DIFF
--- a/internal/query/log.go
+++ b/internal/query/log.go
@@ -65,9 +65,10 @@ func ReadLogBySequence(ctx context.Context, reader dal.PebbleGetter, sequence ui
 // ledgerLogCursor iterates over pre-fetched global sequences and fetches full
 // Log entries from Pebble on demand. It holds no long-lived resources.
 type ledgerLogCursor struct {
-	pebble dal.PebbleGetter
-	seqs   []uint64
-	pos    int
+	pebble     dal.PebbleReader
+	coldReader *coldstorage.ColdReader
+	seqs       []uint64
+	pos        int
 }
 
 func (c *ledgerLogCursor) Next() (*commonpb.Log, error) {
@@ -78,7 +79,10 @@ func (c *ledgerLogCursor) Next() (*commonpb.Log, error) {
 	seq := c.seqs[c.pos]
 	c.pos++
 
-	log, err := ReadLogBySequence(context.Background(), c.pebble, seq)
+	// Cold-storage fallback: a listed sequence may belong to a chapter that has
+	// been archived and purged from hot storage, so fall back to cold (like
+	// GetLog) instead of failing the whole listing.
+	log, err := ReadLogBySequenceWithCold(context.Background(), c.pebble, c.coldReader, seq)
 	if err != nil {
 		return nil, err
 	}
@@ -102,7 +106,8 @@ func (c *ledgerLogCursor) Close() error { return nil }
 // truncated query result is worse than a clear error because the caller
 // cannot tell it apart from a legitimate empty result.
 func ReadLedgerLogsCompiled(
-	pebbleReader dal.PebbleGetter,
+	pebbleReader dal.PebbleReader,
+	coldReader *coldstorage.ColdReader,
 	indexReader dal.PebbleGetter,
 	ledgerName string,
 	logIDs [][]byte,
@@ -153,7 +158,7 @@ func ReadLedgerLogsCompiled(
 		_ = closer.Close()
 	}
 
-	return &ledgerLogCursor{pebble: pebbleReader, seqs: seqs}, nil
+	return &ledgerLogCursor{pebble: pebbleReader, coldReader: coldReader, seqs: seqs}, nil
 }
 
 // ReadLogsSinceRaw returns a raw Pebble iterator for logs after the given
--- a/internal/query/log.go
+++ b/internal/query/log.go
@@ -65,6 +65,10 @@ func ReadLogBySequence(ctx context.Context, reader dal.PebbleGetter, sequence ui
 // ledgerLogCursor iterates over pre-fetched global sequences and fetches full
 // Log entries from Pebble on demand. It holds no long-lived resources.
 type ledgerLogCursor struct {
+	// ctx is the request context, carried because Cursor.Next is ctx-less: the
+	// cold-storage fallback below does slow S3 fetch/ingest that must honor
+	// request cancellation and deadlines.
+	ctx        context.Context
 	pebble     dal.PebbleReader
 	coldReader *coldstorage.ColdReader
 	seqs       []uint64
@@ -82,7 +86,7 @@ func (c *ledgerLogCursor) Next() (*commonpb.Log, error) {
 	// Cold-storage fallback: a listed sequence may belong to a chapter that has
 	// been archived and purged from hot storage, so fall back to cold (like
 	// GetLog) instead of failing the whole listing.
-	log, err := ReadLogBySequenceWithCold(context.Background(), c.pebble, c.coldReader, seq)
+	log, err := ReadLogBySequenceWithCold(c.ctx, c.pebble, c.coldReader, seq)
 	if err != nil {
 		return nil, err
 	}
@@ -106,6 +110,7 @@ func (c *ledgerLogCursor) Close() error { return nil }
 // truncated query result is worse than a clear error because the caller
 // cannot tell it apart from a legitimate empty result.
 func ReadLedgerLogsCompiled(
+	ctx context.Context,
 	pebbleReader dal.PebbleReader,
 	coldReader *coldstorage.ColdReader,
 	indexReader dal.PebbleGetter,
@@ -158,7 +163,7 @@ func ReadLedgerLogsCompiled(
 		_ = closer.Close()
 	}
 
-	return &ledgerLogCursor{pebble: pebbleReader, coldReader: coldReader, seqs: seqs}, nil
+	return &ledgerLogCursor{ctx: ctx, pebble: pebbleReader, coldReader: coldReader, seqs: seqs}, nil
 }
 
 // ReadLogsSinceRaw returns a raw Pebble iterator for logs after the given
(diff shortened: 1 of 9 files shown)


internal/query/log.go
```diff
@@ -78,7 +79,10 @@ func (c *ledgerLogCursor) Next() (*commonpb.Log, error) {
 	seq := c.seqs[c.pos]
 	c.pos++
 
-	log, err := ReadLogBySequence(context.Background(), c.pebble, seq)
+	// Cold-storage fallback: a listed sequence may belong to a chapter that has
+	// been archived and purged from hot storage, so fall back to cold (like
+	// GetLog) instead of failing the whole listing.
+	log, err := ReadLogBySequenceWithCold(context.Background(), c.pebble, c.coldReader, seq)
```
NumaryBot: 🟠 **[major] Propagate request context to cold log reads**

When a `ListLogs` or prepared-query `LOGS` page includes an archived log, this now reaches `ColdReader.GetReader`, but the cursor always passes `context.Background()`. With S3 cold storage, client cancellation or gRPC deadlines are ignored while the archive is fetched/ingested, so a canceled request can keep running until the storage operation times out; carry the request context into the cursor instead of constructing a background context here.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1633#issuecomment-5058048961

flemzord (changes requested):
Revalidated live head `61c7e8dc`. The cold fallback fixes archived-log resolution, but the existing NumaryBot finding on this same head remains a runtime blocker: `ledgerLogCursor.Next` calls `ReadLogBySequenceWithCold` with `context.Background()`, so ListLogs and prepared-query LOGS requests no longer propagate cancellation or deadlines into potentially long S3 fetch/ingest work. Please carry the request context into the cursor. The separate Codecov patch failure is CI/tooling and is not the basis of this request for changes; I am not duplicating the existing inline finding.

internal/query/log.go
Azorlogh: Fixed in 6144a65b8. The cursor now carries the request context into the cold read (`context.Background()` → request `ctx`, threaded through `ReadLedgerLogsCompiled`/`EnrichLogs` from the request-scoped callers), so client cancellation / gRPC deadlines propagate to the S3 fetch+ingest.

internal/query/log.go
```diff
@@ -78,7 +83,10 @@ func (c *ledgerLogCursor) Next() (*commonpb.Log, error) {
 	seq := c.seqs[c.pos]
 	c.pos++
```
NumaryBot, line 85: 🟠 **[major] Request context not propagated to cold storage reads**

When a ListLogs or prepared-query LOGS page includes an archived log, the cold storage fetch is performed with context.Background() instead of the incoming request context. This means client cancellations and gRPC/HTTP deadlines are ignored while the archive is being fetched from S3 (or any cold backend), keeping the operation running until the storage layer times out on its own.

**Suggestion:** Thread the request context into the cursor callback so that ColdReader.GetReader (and any downstream I/O) receives the real request context, allowing prompt cancellation when the caller disconnects or a deadline expires.

tests/e2e/business/listlogs_cold_test.go
```diff
@@ -0,0 +1,85 @@
...
+				if re == io.EOF {
+					break
+				}
+				if re != nil {
+					listErr = re
+					return true
+				}
+				count++
+			}
+			return true
```
NumaryBot: 🟡 **[minor] Eventually predicate always returns true, defeating the retry loop**

The Eventually callback unconditionally returns true regardless of whether ListLogs succeeded or returned the expected count. This means the assertion exits after the very first attempt, so transient read-index lag can cause the subsequent count == 3 check to fail even though the service would return correct results a moment later.

**Suggestion:** Change the predicate to return true only when both listErr == nil and count == 3 (or move the assertion inside the Eventually block), so the helper actually waits for the async indexer to catch up before declaring success.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1633#issuecomment-5058048961

flemzord:
Revalidated live head `6144a65b`. The request-context blocker from my earlier changes-requested review is fixed: cold log reads now receive the caller context. The remaining `Eventually`/Codecov items are CI/tooling, not demonstrated runtime blockers. I found no remaining correctness/runtime blocker. GitHub approval is intentionally disabled by this cron configuration.

internal/query/log.go
Azorlogh: stale — this re-raises the context finding from https://github.com/formancehq/ledger/pull/1633#discussion_r3637851009, already fixed in 6144a65b8 (the commit this review ran against): `log.go` passes the request `ctx` into the cold read, not `context.Background()`.

tests/e2e/business/listlogs_cold_test.go
Azorlogh: Fixed in d78653920 — the assertions now run inside `Eventually(func(g Gomega){…})`, so it retries until all three archived logs are listable instead of passing on the first attempt.

NumaryBot (commented):
## 💬 Comments — automated review

The cold-storage fallback is correctly wired through the replay, ListLogs, and prepared-query paths. The previously flagged context propagation blocker (context.Background() in cold log reads) was fixed in commit 6144a65b and confirmed resolved by a human reviewer. One minor issue remains: the e2e test's Eventually predicate unconditionally returns true, so the retry loop provides no actual resilience against transient read-index lag — the subsequent count assertion can still fail on a slow CI run before the async indexer catches up.

### Findings outside the diff

🟡 **[minor] Eventually predicate always returns true, defeating the retry loop** — `tests/e2e/business/listlogs_cold_test.go:79`

The Eventually callback unconditionally returns true regardless of whether ListLogs succeeded or returned the expected count. This means the assertion exits after the very first attempt, so transient read-index lag can cause the subsequent count == 3 check to fail even though the service would return correct results moments later.

**Suggestion:** Move the assertion inside the Eventually block and return true only when both listErr == nil and the returned count equals 3 (or the full expected result set). This lets the helper actually wait for the async indexer to catch up before declaring success.



Summary: https://github.com/formancehq/ledger/pull/1633#issuecomment-5058048961

Azorlogh:
stale — the `[minor]` "Eventually predicate always returns true" re-posts https://github.com/formancehq/ledger/pull/1633#discussion_r3638057769, already fixed in d78653920 (the commit this review ran against): the predicate is now `Eventually(func(g Gomega){ … g.Expect(count).To(Equal(3)) }).Should(Succeed())` — assertions run inside, so it retries the async indexer instead of returning true unconditionally.

flemzord:
Revalidated live head `d7865392`. The cold-storage fallback still uses the caller request context, and the latest test-only commit now makes the `Eventually` assertion wait for the indexed result as intended. The remaining `codecov/patch` failure is CI/tooling rather than a demonstrated runtime blocker. I found no correctness/runtime blocker on this head. GitHub approval is intentionally disabled for this cron task.

paul-nicolas (approved):

flemzord (approved):

PR #1634, opened 2026-07-23
fix(operator): apply persistence volume size increases online

sylr:
## Problem

Editing `spec.persistence.<vol>.size` on a `Cluster` (e.g. `data: 2Ti → 3Ti`) did nothing to the disks. The StatefulSet reconcile only detected volumeClaimTemplate **name** changes (`volumeClaimTemplatesChanged` compares the name set), and the normal update path deliberately skips the immutable VCT. The size bump *did* change the pod-template spec-hash — so pods rolled — while the volumes stayed at the old size. The STS `volumeClaimTemplates` kept reporting the old size.

## Change

- **New `reconcilePVCExpansion`** (`internal/controller/reconcile_pvc_expansion.go`): patches each existing PVC's `spec.resources.requests.storage` upward, triggering CSI online expansion. Grow-only.
- **Wired into `reconcileStatefulSet`**: after the expansion pass, delete-recreate the StatefulSet with `DeletePropagationOrphan` (pods/PVCs retained and re-adopted, no restart) whenever VCT names changed **or** a size grew, so future scale-out ordinals inherit the new template size.
- **README**: new "Volume Resizing" section documenting the online-grow flow, the `allowVolumeExpansion: true` prerequisite, and grow-only semantics.

### Behavior notes
- **Grow-only.** A smaller desired size emits a `VolumeShrinkRejected` warning event and leaves the PVC untouched (K8s cannot shrink a PVC). Each successful expansion emits `VolumeExpanded`.
- **Shrink is decided from the template delta, not per-PVC size** — a PVC manually expanded past the new template size is left alone rather than falsely warned every reconcile. Warns once per volume, not per PVC.
- **Skipped during scale-down** so the recreate's early return does not bypass `deleteScaledDownPVCs` and orphan the removed ordinals' PVCs (also closes the same latent issue in the pre-existing name-change branch). The template change is picked up on the next reconcile once the scale-down settles.

## Why not operator-owned PVCs?
Keeping `volumeClaimTemplates` preserves per-ordinal RWO identity for the Raft nodes for free. Owning PVCs directly means either shared storage (broken) or reimplementing StatefulSet (unacceptable risk). The gap was only the missing grow path — same pattern as Prometheus Operator / CloudNativePG.

## Review
Reviewed by Codex (gpt-5.5): confirmed the `resource.Quantity.Cmp` approach, the `requests.storage` patch API, and orphan re-adoption. It found two Medium issues — false shrink warnings for over-expanded PVCs, and the resize+scale-down cleanup bypass — both fixed in this branch and covered by tests.

## Tests
- 6 unit tests in `reconcile_pvc_expansion_test.go` (grow-all + recreate signal, no-op, shrink rejected, missing ordinal skipped, over-expanded left alone with no event, shrink-warns-once-per-volume).
- Full `internal/controller` package passes; `golangci-lint` 0 issues; `go build ./...` clean.

## Not covered (needs e2e follow-up)
The full delete-recreate + pod re-adoption cycle and the resize+scale-down deferral are not unit-testable with a fake clientset (`raftScaleDown` execs into pods).

### DIFF
--- a/misc/operator/internal/controller/reconcile_statefulset.go
+++ b/misc/operator/internal/controller/reconcile_statefulset.go
@@ -181,6 +181,11 @@ func (r *ClusterReconciler) reconcileStatefulSet(ctx context.Context, ledger *le
 			}
 			// Returning here requeues via the owned-StatefulSet delete event — the next
 			// reconciliation creates the new StatefulSet and the orphaned pods are adopted.
+			// This branch only DELETES; it never builds the replacement template. The
+			// new StatefulSet is created on the next reconcile, which re-runs
+			// reconcilePVCExpansion above and re-clamps `desired` against the live PVCs
+			// (which survive the orphan delete) — so a rejected shrink cannot reach the
+			// recreated template even though this delete carries the grown volume.
 			return ctrl.Result{}, nil
 		}
 	}
--- a/misc/operator/internal/controller/reconcile_statefulset.go
+++ b/misc/operator/internal/controller/reconcile_statefulset.go
@@ -126,36 +126,47 @@ func (r *ClusterReconciler) reconcileStatefulSet(ctx context.Context, ledger *le
 
 	desired := buildStatefulSetSpec(ledger, specHash, credentials, targetTLSMode)
 
-	// Check if VolumeClaimTemplates changed on an existing StatefulSet.
-	// VCTs are immutable, so any change that must land in the template requires a
-	// delete-recreate with orphan propagation:
-	//   - a name-set change (a volume switching between PVC and hostPath mode), and
-	//   - a size increase, so future scale-out ordinals are born at the new size.
-	// A size increase additionally requires patching the *existing* PVCs directly:
-	// recreating the StatefulSet re-adopts them without resizing, so the live
-	// disks only grow via reconcilePVCExpansion's CSI online expansion.
-	//
-	// This is deliberately skipped while a scale-down is in progress: the
-	// delete-recreate returns early, which would bypass the deleteScaledDownPVCs
-	// cleanup below and orphan the removed ordinals' PVCs. Two disruptive
-	// StatefulSet operations must not interleave in one pass — the scale-down
-	// completes this reconcile (including PVC cleanup) and the template change is
-	// picked up on the next one, once scalingDown is false.
 	existing := &appsv1.StatefulSet{}
-	if err := r.Get(ctx, types.NamespacedName{Name: name, Namespace: ledger.Namespace}, existing); err == nil && !scalingDown {
-		namesChanged := volumeClaimTemplatesChanged(existing.Spec.VolumeClaimTemplates, desired.VolumeClaimTemplates)
+	stsFound := r.Get(ctx, types.NamespacedName{Name: name, Namespace: ledger.Namespace}, existing) == nil
 
-		templateGrew := false
-		if r.Clientset != nil {
-			grew, err := reconcilePVCExpansion(ctx, r.Clientset, r.Recorder, ledger,
-				ledger.Namespace, name, desiredReplicas,
-				existing.Spec.VolumeClaimTemplates, desired.VolumeClaimTemplates)
-			if err != nil {
-				return ctrl.Result{}, fmt.Errorf("expanding PVCs for volume size change: %w", err)
-			}
-			templateGrew = grew
+	var existingVCTs []corev1.PersistentVolumeClaim
+	if stsFound {
+		existingVCTs = existing.Spec.VolumeClaimTemplates
+	}
+
+	// Reconcile PVC sizes: clamp the desired templates up to the largest live PVC
+	// (grow-only — a spec shrink is rejected and the current size kept), grow any
+	// lagging PVCs via CSI online expansion, and learn whether the template grew.
+	//
+	// This runs on EVERY path, not only when the StatefulSet exists: the create
+	// path right after a recreate (below) rebuilds the template straight from the
+	// spec, so the clamp must re-apply there too, or a rejected shrink would leak
+	// back into the rebuilt template. Because the clamp floors against the live
+	// PVCs (which survive the orphan delete), it is stable across the recreate.
+	templateGrew := false
+	if r.Clientset != nil {
+		grew, err := reconcilePVCExpansion(ctx, r.Clientset, r.Recorder, ledger,
+			ledger.Namespace, name, desiredReplicas, existingVCTs, desired.VolumeClaimTemplates)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("reconciling PVC sizes for volume change: %w", err)
 		}
+		templateGrew = grew
+	}
 
+	// VolumeClaimTemplates are immutable, so a change that must land in the
+	// template requires a delete-recreate with orphan propagation:
+	//   - a name-set change (a volume switching between PVC and hostPath mode), or
+	//   - a size increase, so future scale-out ordinals are born at the new size.
+	// The pods and PVCs are retained and re-adopted by the recreated StatefulSet.
+	//
+	// Deliberately skipped while a scale-down is in progress: the delete-recreate
+	// returns early, which would bypass the deleteScaledDownPVCs cleanup below and
+	// orphan the removed ordinals' PVCs. Two disruptive StatefulSet operations
+	// must not interleave in one pass — the scale-down completes this reconcile
+	// (including PVC cleanup), and the template change is picked up on the next
+	// one, once scalingDown is false.
+	if stsFound && !scalingDown {
+		namesChanged := volumeClaimTemplatesChanged(existing.Spec.VolumeClaimTemplates, desired.VolumeClaimTemplates)
 		if namesChanged || templateGrew {
 			reason := "VolumeClaimTemplates changed"
 			if templateGrew {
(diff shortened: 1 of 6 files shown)


misc/operator/internal/controller/reconcile_statefulset.go
```diff
@@ -127,11 +127,41 @@ func (r *ClusterReconciler) reconcileStatefulSet(ctx context.Context, ledger *le
...
+			grew, err := reconcilePVCExpansion(ctx, r.Clientset, r.Recorder, ledger,
+				ledger.Namespace, name, desiredReplicas,
+				existing.Spec.VolumeClaimTemplates, desired.VolumeClaimTemplates)
+			if err != nil {
+				return ctrl.Result{}, fmt.Errorf("expanding PVCs for volume size change: %w", err)
+			}
+			templateGrew = grew
+		}
+
+		if namesChanged || templateGrew {
```
NumaryBot, line 170: 🟠 **[major] Preserve rejected shrink sizes when recreating**

When one volume grows and another volume is shrunk in the same Cluster update, `reconcilePVCExpansion` warns and skips the shrunk PVC, but `templateGrew` still makes this branch delete/recreate the StatefulSet with the full `desired.VolumeClaimTemplates`. The recreated template will therefore carry the rejected smaller size for that other volume, so future scale-out ordinals can be created with the shrunken PVC size even though shrink was supposed to be grow-only and left untouched.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1634#issuecomment-5058166235

flemzord (changes requested):
Reviewed live head `f064a124`. The existing NumaryBot finding on this same head is a correctness/runtime blocker: if one volume grows while another is requested smaller, the PVC expansion pass rejects the shrink but the grow still triggers StatefulSet recreation from the unmodified desired templates. The recreated template therefore persists the rejected smaller size for future ordinals. Please sanitize rejected shrink templates back to their existing sizes (and add the mixed grow/shrink regression) before recreating. I am not duplicating the inline finding. E2E CI is still running and is not the basis of this request for changes.

sylr:
Fixed in 08227de6e.

**Root cause:** the grow forced a StatefulSet delete-recreate that rebuilt every `volumeClaimTemplate` from spec, and my shrink handling didn't survive the recreate — on the post-delete reconcile the old template is gone, so the rejected smaller size leaked back into the rebuilt template.

**Fix:** floor each desired template size up to the **largest live PVC** for that volume (grow-only), applied on **every** reconcile path — including the create path right after the orphan delete. The floor is sourced from the live PVCs (which survive the orphan delete), not the StatefulSet template, so it's stable across the recreate. A spec size below the live disk is rejected with a single `VolumeShrinkRejected` event per volume and the template is clamped back to the current size — so a rejected shrink can no longer reach a rebuilt template or future scale-out ordinals.

The clamp + grow are now a single PVC pass run unconditionally (previously only inside the STS-found branch), so the create path is clamped too.

**Regression tests added** (`reconcile_pvc_expansion_test.go`):
- `MixedGrowShrinkDoesNotLeakShrink` — data 2Ti→3Ti grows while wal 5Gi→4Gi is clamped back to 5Gi; asserts the desired `wal` template stays 5Gi.
- `ClampSurvivesRecreatePath` — models the post-orphan-delete reconcile (`existingTemplates == nil`); the live PVC still floors the rebuilt template.

Full `internal/controller` package passes; `golangci-lint` 0 issues.

Thanks @flemzord for the catch.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1634#issuecomment-5058166235

misc/operator/internal/controller/reconcile_statefulset.go
```diff
@@ -126,12 +126,53 @@ func (r *ClusterReconciler) reconcileStatefulSet(ctx context.Context, ledger *le
...
+	//
+	// Deliberately skipped while a scale-down is in progress: the delete-recreate
+	// returns early, which would bypass the deleteScaledDownPVCs cleanup below and
+	// orphan the removed ordinals' PVCs. Two disruptive StatefulSet operations
+	// must not interleave in one pass — the scale-down completes this reconcile
+	// (including PVC cleanup), and the template change is picked up on the next
+	// one, once scalingDown is false.
+	if stsFound && !scalingDown {
+		namesChanged := volumeClaimTemplatesChanged(existing.Spec.VolumeClaimTemplates, desired.VolumeClaimTemplates)
+		if namesChanged || templateGrew {
```
NumaryBot, line 170: 🟠 **[major] Rejected shrink size persisted into recreated StatefulSet template**

When one volume grows and another is shrunk in the same Cluster update, `reconcilePVCExpansion` warns and skips the shrunk PVC, but `templateGrew` still causes the StatefulSet to be deleted and recreated using the full `desired.VolumeClaimTemplates`. The recreated template then carries the smaller (rejected) size for the shrunk volume, so future scale-out ordinals will be created with that shrunken PVC size even though the grow-only semantics should have left that volume's size untouched.

**Suggestion:** Before passing `desired.VolumeClaimTemplates` to the recreated StatefulSet, replace any shrunk volume's size with the current (larger) value from the existing StatefulSet's VolumeClaimTemplates, so only the genuinely grown volumes are updated in the new template.

NumaryBot, line 170: 🟠 **[major] Resize-only changes alter pod-template hash and trigger a pod rollout**

When `templateGrew` is true, the recreate path builds a replacement StatefulSet whose spec hash still includes `spec.persistence.*.size`. Because `computeSpecHash` and `buildPodTemplate` embed this hash into the pod template annotation, the new StatefulSet gets a different ControllerRevision from the adopted pods. With partition 0, Kubernetes will then roll the adopted pods, making the expansion path not actually online/no-restart as intended.

**Suggestion:** Exclude PVC size fields from the pod-template hash computation (or use a separate hash for VolumeClaimTemplates), so a pure volume-size increase does not change the pod-template hash and the adopted pods are not rolled.

sylr:
Second review pass addressed in 97c445092.

**Finding B — resize alters pod-template hash → pod rollout (valid, fixed).**
`computeSpecHash` included `persistence.*.size`, so a resize bumped the pod-template annotation; on the expansion recreate the new StatefulSet derived a different ControllerRevision from the adopted pods and (Partition 0) rolled them — so the "no-restart" expansion actually restarted every pod. Fixed by excluding the PVC-only `VolumeSpec` fields (`size`, `storageClass`, `accessMode`, `volumeAttributesClassName`) from the hash — they map to the VolumeClaimTemplates/PVCs and never appear in `buildPodTemplate`, same rationale as the already-excluded `DeletionProtection`. `HostPath` is deliberately kept (PVC↔hostPath does change the pod template). Tests added in `hash_test.go` (size + PVC-only fields excluded; hostPath still included).

**Finding A — rejected shrink persisted into recreated template (already handled; false positive on this head).**
The `templateGrew` branch only **deletes** the StatefulSet with orphan propagation and returns — it does not build the replacement template. The new StatefulSet is created on the *next* reconcile, which re-runs `reconcilePVCExpansion` and re-clamps `desired` against the **live PVCs** (which survive the orphan delete), so the shrunk volume is floored back to its current size before the template is ever rebuilt. That's exactly what the `ClampSurvivesRecreatePath` test covers (`existingTemplates == nil`, live PVC still floors the rebuilt template), and `MixedGrowShrinkDoesNotLeakShrink` asserts the `wal` template stays 5Gi. The suggested "sanitize against the existing StatefulSet VCTs" wouldn't work here — on the post-delete reconcile the old StatefulSet is gone, which is why the floor is sourced from PVCs instead. Added a comment at the recreate site to make the deferred-recreate flow explicit.

All CI green on the prior head; `internal/controller` package + lint clean locally.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1634#issuecomment-5058166235

flemzord:
Revalidated live head `97c44509`. The correctness blocker from my earlier changes-requested review is fixed on this head, and NumaryBot reports no remaining finding; CI is green. I found no remaining correctness/runtime blocker. GitHub approval is intentionally disabled by this cron configuration.

Azorlogh:
**Non-expandable StorageClass turns a size bump into a wedged reconcile.** The `PersistentVolumeClaimResize` admission plugin (default-enabled) rejects the `requests.storage` patch with 403 when the SC lacks `allowVolumeExpansion: true`; `reconcilePVCExpansion` returns that error and `reconcileStatefulSet` aborts before the `CreateOrUpdate` — blocking image bumps, replica changes, scale-down PVC cleanup, and volume-protection labels on every requeue until the spec size is reverted. No event is emitted on this path. Suggest catching the forbidden error, emitting a warning event, and skipping that volume so the rest of the pass proceeds.

flemzord (changes requested):
Revalidated live head `97c44509` after the new Azorlogh comment. I confirm that existing comment as a correctness/runtime blocker: a rejected PVC expansion patch returns before the normal StatefulSet update and the remaining reconciliation passes, so an unsupported StorageClass can leave unrelated desired changes permanently blocked until the size request is reverted. I am not duplicating the existing details. Please ensure an unsupported expansion cannot wedge the rest of reconciliation. CI is green and is not the basis of this review.

sylr:
superseded by #1674

PR #1635, opened 2026-07-23
fix(EN-1597): honor absolute SeekGE/SeekLE reposition-after-exhaustion in read-store iterators

Azorlogh:
## Problem

A `ListTransactions`/`ListAccounts` filter composing `NOT` (especially nested `NOT`) with other conditions could return entities it must exclude. Reduced case: `and(not(not(reverted=false)), reverted=true)` — a logical contradiction — returned a transaction instead of nothing.

## Root cause

`SeekGE`/`SeekLE` are contractually *absolute* repositioning ("position at the first entity ≥/≤ target"), but the read-store iterators latched on exhaustion: once exhausted they early-returned `false` and refused to reposition, even though their bodies already recompute position from `target`.

A `NOT` drives a child forward to exhaustion while emitting an early row; an `AND`/`OR` merge then seeks the tree back to a smaller key; the latch turns that backward seek into a no-op, so the `NOT` stops excluding and an excluded entity leaks through `universe \ child`.

## Changes

1. **`NotIterator` + `BitsetIterator`** — re-seek the child unconditionally / reposition from target instead of latching (the confirmed root cause).
2. **Uniform pass** — clear the exhausted/done latch in every leaf and merge iterator so `SeekGE`/`SeekLE` reposition after exhaustion: `PebbleTxIterator`, `PebbleAccountIterator`, `PebbleTxRangeIterator`, `PebbleReverseTxIterator`, `PebbleReverseAccountIterator`, `PrefixIterator`, `ReversePrefixIterator`, `AndIterator`, `OrIterator`, `NotIterator`, `ReverseOrIterator`. `RangeIterator` is the deliberate exception: a raw range scan surfaces rows in `(value, entity)` order across index-value buckets, so an entity-space `SeekGE` is unimplementable without draining the range. Every construction site already materializes it into a sorted `SliceIterator`; its `SeekGE` now fails the query with an invariant error instead of mis-seeking plausibly.
3. **`AndIterator.SeekGE`** — re-seek **all** children, not just `children[0]`; a stale non-first child could otherwise become the `converge` candidate and skip valid intersections below it (reported by NumaryBot).
4. **`AddressTxIterator` restructured into a conforming absolute repositioner** — the old implementation destructively consumed a materialized slice (`pendingTxns[idx+1:]`) behind an exhaustion latch, so `SeekGE` on it dropped rows once the `AndIterator` change re-seeked children unconditionally. It now materializes the account→tx union once into a stable slice and seeks with a `sort.Search` cursor — repeatable, backward-seekable, well-defined after exhaustion.
5. **`seekFloor`/`seekCeil` exhaustion-proof cache** (`iterator_floor.go`) — removing the latches made composite parents re-seek exhausted children once per merge step (O(rows) Pebble seeks). A *cleanly* failed `SeekGE(t)` proves no entity ≥ t exists in the iterator's snapshot-fixed view, so the leaf records the bound and answers later covered seeks in O(1). Preconditions: the Pebble view is fixed at iterator creation (proofs never expire), and an I/O-failed seek proves nothing — `fail()` takes the iterator's storage error and drops the proof when non-nil.
6. **Contract made normative** — `EntityIterator.SeekGE` / `ReverseIterator.SeekLE` doc blocks now state the absolute-reposition guarantees (computed from target alone, idempotent/non-consuming, valid after exhaustion, failed seek leaves the iterator re-seekable), with a dedicated docs page: `docs/technical/architecture/subsystems/read-path/iterator-seek-contract.md`. `query-pipeline.md`/`prepared-queries.md` updated — `AndIterator.SeekGE` now force-seeks every child, so the old "monotonic-skip on seek" description no longer holds.

## Tests

- e2e regressions: `reverted_filter_contradiction_test.go` and `filter_nested_not_reposition_test.go` (tx-id-range + account-address leaves) — fail pre-fix, pass after. The nested-NOT suite guards against vacuous passes with a `BeforeAll` baseline (aborts the container if the seeded rows are not visible); those leaves read the FSM-synchronous attributes zone.
- e2e coverage: two tx-address contradictions in the nested-NOT suite (`ListTransactionsFiltered` + `AddressPrefixFilter` behind a READY tx-address index) — the only e2e path that compiles an `AddressTxIterator`. These pin reachability, not the pre-fix bug: with the `NotIterator`/`AndIterator` fixes in place, `converge` only seeks strictly-behind children, so no single-page query re-seeks the same target twice and the old destructive consume has no e2e repro (verified: the specs pass against the pre-fix `AddressTxIterator`). The conformance itself is pinned at unit level.
- readstore unit suites: `iterator_and_seek_test.go` (AndIterator re-seeks all children), `iterator_address_test.go` (AddressTxIterator absolute-seek conformance), `iterator_floor_test.go` (floor/ceil algebra incl. the I/O-error carve-out, forward + reverse leaf repositioning through the cache, `PebbleReverseTxIterator` exhaust-then-seek-back branches), `iterator_leaf_test.go` (RangeIterator forward drain + loud invariant error on `SeekGE`).
- `internal/storage/readstore` + `internal/query` unit suites green.

Found by the model-based antithesis driver (`singleton_driver_model`); confirmed on a single node and on a 3-node cluster with rolling restarts.

### DIFF
--- a/internal/storage/readstore/iterator_address.go
+++ b/internal/storage/readstore/iterator_address.go
@@ -3,6 +3,7 @@ package readstore
 import (
 	"bytes"
 	"encoding/binary"
+	"sort"
 
 	"github.com/cockroachdb/pebble/v2"
 
@@ -14,17 +15,23 @@ import (
 //  1. Scanning the existence index for matching account addresses
 //  2. For each matching account, scanning the account→tx mapping
 //  3. Merge-unioning all transaction ID sets into a single sorted output
+//
+// The union is materialized in full on first use and kept for the iterator's
+// lifetime; Next and SeekGE are cursor moves over the stable sorted slice, so
+// SeekGE is a true absolute reposition — seekable backwards, repeatable, and
+// well-defined after exhaustion — as the EntityIterator contract requires.
 type AddressTxIterator struct {
-	reader      dal.PebbleReader
-	kb          *dal.KeyBuilder
-	ledgerName  string
-	prefix      byte           // which account→tx prefix to scan
-	addrIter    EntityIterator // iterates over matching account addresses
-	current     []byte         // current txID (8 bytes)
-	exhausted   bool
-	err         error    // first I/O error from materialize / addrIter
-	pendingTxns [][]byte // merge-union buffer for deduplication
-	txSeen      map[uint64]struct{}
+	reader     dal.PebbleReader
+	kb         *dal.KeyBuilder
+	ledgerName string
+	prefix     byte           // which account→tx prefix to scan
+	addrIter   EntityIterator // iterates over matching account addresses
+	current    []byte         // current txID (8 bytes)
+	err        error          // first I/O error from materialize / addrIter
+
+	materialized bool
+	txns         [][]byte // all matching txIDs, sorted and deduplicated
+	pos          int      // index into txns of the entry the next Next() yields
 }
 
 // NewAddressTxIterator creates an iterator that, for each address matching
@@ -43,39 +50,20 @@ func NewAddressTxIterator(
 		ledgerName: ledgerName,
 		prefix:     prefix,
 		addrIter:   addrIter,
-		txSeen:     make(map[uint64]struct{}),
 	}
 }
 
 func (it *AddressTxIterator) Next() bool {
-	if it.exhausted {
+	if !it.ensureMaterialized() {
 		return false
 	}
 
-	// Try to get next tx from pending buffer
-	if len(it.pendingTxns) > 0 {
-		it.current = it.pendingTxns[0]
-		it.pendingTxns = it.pendingTxns[1:]
-
-		return true
-	}
-
-	// Materialize all txIDs for remaining accounts and merge-union
-	if err := it.materialize(); err != nil {
-		it.err = err
-		it.exhausted = true
-
+	if it.pos >= len(it.txns) {
 		return false
 	}
 
-	if len(it.pendingTxns) == 0 {
-		it.exhausted = true
-
-		return false
-	}
-
-	it.current = it.pendingTxns[0]
-	it.pendingTxns = it.pendingTxns[1:]
+	it.current = it.txns[it.pos]
+	it.pos++
 
 	return true
 }
@@ -85,44 +73,21 @@ func (it *AddressTxIterator) Current() []byte {
 }
 
 func (it *AddressTxIterator) SeekGE(target []byte) bool {
-	if it.exhausted {
+	if !it.ensureMaterialized() {
 		return false
 	}
 
-	// Re-materialize with filter
-	if len(it.pendingTxns) == 0 {
-		if err := it.materialize(); err != nil {
-			it.err = err
-			it.exhausted = true
-
-			return false
-		}
-
-		if len(it.pendingTxns) == 0 {
-			it.exhausted = true
-
-			return false
-		}
-	}
-
-	// Binary search in sorted pendingTxns
-	idx := 0
-	for idx < len(it.pendingTxns) {
-		if bytes.Compare(it.pendingTxns[idx], target) >= 0 {
-			break
-		}
-
-		idx++
-	}
-
-	if idx >= len(it.pendingTxns) {
-		it.exhausted = true
+	idx := sort.Search(len(it.txns), func(i int) bool {
+		return bytes.Compare(it.txns[i], target) >= 0
+	})
+	if idx >= len(it.txns) {
+		it.pos = len(it.txns)
 
 		return false
 	}
 
-	it.current = it.pendingTxns[idx]
-	it.pendingTxns = it.pendingTxns[idx+1:]
+	it.current = it.txns[idx]
+	it.pos = idx + 1
 
 	return true
 }
@@ -139,11 +104,34 @@ func (it *AddressTxIterator) Close() {
 	it.addrIter.Close()
 }
 
+// ensureMaterialized runs the one-time materialization, latching any I/O error
+// (an error is permanent — positioning calls after it always return false).
+func (it *AddressTxIterator) ensureMaterialized() bool {
+	if it.err != nil {
+		return false
+	}
+
+	if it.materialized {
+		return true
+	}
+
+	it.materialized = true
+	if err := it.materialize(); err != nil {
+		it.err = err
+
+		return false
+	}
+
+	return true
+}
+
 // materialize collects all transaction IDs from all matching accounts,
 // deduplicates, and sorts them. Surfaces I/O errors from the underlying
 // Pebble iterators and from the addrIter through addrIter.Err()
 // (checked by the caller via it.Err()).
 func (it *AddressTxIterator) materialize() error {
+	txSeen := make(map[uint64]struct{})
+
 	for it.addrIter.Next() {
 		account := string(it.addrIter.Current())
 		prefix := AccountTxPrefix(it.kb, it.prefix, it.ledgerName, account)
@@ -167,15 +155,15 @@ func (it *AddressTxIterator) materialize() error {
 			txIDBytes := k[len(k)-8:]
 			txID := binary.BigEndian.Uint64(txIDBytes)
 
-			if _, seen := it.txSeen[txID]; seen {
+			if _, seen := txSeen[txID]; seen {
 				continue
 			}
 
-			it.txSeen[txID] = struct{}{}
+			txSeen[txID] = struct{}{}
 
 			txCopy := make([]byte, 8)
 			copy(txCopy, txIDBytes)
-			it.pendingTxns = insertSorted(it.pendingTxns, txCopy)
+			it.txns = insertSorted(it.txns, txCopy)
 		}
 
 		iterErr := iter.Error()
--- a/internal/storage/readstore/iterator_and.go
+++ b/internal/storage/readstore/iterator_and.go
@@ -47,9 +47,13 @@ func (it *AndIterator) Current() []byte {
 }
 
 func (it *AndIterator) SeekGE(target []byte) bool {
-	if it.exhausted || len(it.children) == 0 {
+	if len(it.children) == 0 {
 		return false
 	}
+
+	// Absolute reposition: clear the exhausted latch so a re-seek after
+	// exhaustion re-establishes the intersection (converge re-seeks children).
+	it.exhausted = false
 
 	// Seek all children to target
 	if !it.children[0].SeekGE(target) {
--- a/internal/storage/readstore/iterator_and.go
+++ b/internal/storage/readstore/iterator_and.go
@@ -51,15 +51,19 @@ func (it *AndIterator) SeekGE(target []byte) bool {
 		return false
 	}
 
-	// Absolute reposition: clear the exhausted latch so a re-seek after
-	// exhaustion re-establishes the intersection (converge re-seeks children).
+	// Absolute reposition: seek EVERY child to target, not just children[0].
+	// converge uses each child's Current() as a candidate, so a child left at a
+	// stale position past target would become the candidate and skip valid
+	// intersections below it. Clearing exhausted lets a re-seek after exhaustion
+	// re-establish the intersection.
 	it.exhausted = false
 
-	// Seek all children to target
-	if !it.children[0].SeekGE(target) {
-		it.exhausted = true
+	for i := range it.children {
+		if !it.children[i].SeekGE(target) {
+			it.exhausted = true
 
-		return false
+			return false
+		}
 	}
 
 	return it.converge()
--- a/internal/storage/readstore/iterator_and.go
+++ b/internal/storage/readstore/iterator_and.go
@@ -51,10 +51,10 @@ func (it *AndIterator) SeekGE(target []byte) bool {
 		return false
 	}
 
-	// Absolute reposition: seek EVERY child to target, not just children[0].
-	// converge uses each child's Current() as a candidate, so a child left at a
-	// stale position past target would become the candidate and skip valid
-	// intersections below it. Clearing exhausted lets a re-seek after exhaustion
+	// Absolute reposition: seek EVERY child to target. converge uses each
+	// child's Current() as a candidate, so a child left at a stale position
+	// past target would become the candidate and skip valid intersections
+	// below it. Clearing exhausted lets a re-seek after exhaustion
 	// re-establish the intersection.
 	it.exhausted = false
 
--- a/internal/storage/readstore/iterator_floor.go
+++ b/internal/storage/readstore/iterator_floor.go
@@ -3,12 +3,15 @@ package readstore
 import "bytes"
 
 // seekFloor caches an exhaustion proof for a forward iterator: a failed
-// SeekGE(t) proves no entity >= t exists in the iterator's view. The view is
-// fixed at pebble.Iterator creation, so the proof never goes stale and the
-// floor is never cleared. Composite iterators re-seek exhausted children once
-// per merge step (AND converge, OR findMin, the NOT child catch-up); the floor
-// turns every such re-seek at or above the proven bound into an O(1)
-// comparison instead of a Pebble seek.
+// SeekGE(t) proves no entity >= t exists in the iterator's view. The proof is
+// valid only for a CLEAN exhaustion of a snapshot-fixed view: the Pebble view
+// is fixed at pebble.Iterator creation (so a proof never goes stale and the
+// floor is never cleared), and a seek that failed because of an I/O error
+// proves nothing about the view's contents (fail drops it — see #320 for the
+// error class). Composite iterators re-seek exhausted children once per merge
+// step (AND converge, OR findMin, the NOT child catch-up); the floor turns
+// every such re-seek at or above the proven bound into an O(1) comparison
+// instead of a Pebble seek.
 type seekFloor struct {
 	bound []byte
 	set   bool
@@ -19,17 +22,24 @@ func (f *seekFloor) covers(target []byte) bool {
 	return f.set && bytes.Compare(target, f.bound) >= 0
 }
 
-// fail records a failed SeekGE(target): no entity >= target exists. The bound
-// is copied — target may alias a sibling iterator's Pebble key buffer. Callers
-// only reach a real (uncovered) seek with target below the current bound, so
+// fail records a failed SeekGE(target): no entity >= target exists. err is the
+// iterator's storage error: a non-nil err means the seek failed for I/O
+// reasons, which proves nothing, so no bound is recorded. The bound is copied
+// — target may alias a sibling iterator's Pebble key buffer. Callers only
+// reach a real (uncovered) seek with target below the current bound, so
 // overwriting always tightens it.
-func (f *seekFloor) fail(target []byte) {
+func (f *seekFloor) fail(target []byte, err error) {
+	if err != nil {
+		return
+	}
+
 	f.bound = append(f.bound[:0], target...)
 	f.set = true
 }
 
 // seekCeil is the reverse-iteration analog of seekFloor: a failed SeekLE(t)
-// proves no entity <= t exists.
+// proves no entity <= t exists. Same clean-exhaustion / snapshot-fixed-view
+// precondition.
 type seekCeil struct {
 	bound []byte
 	set   bool
@@ -40,8 +50,13 @@ func (c *seekCeil) covers(target []byte) bool {
 	return c.set && bytes.Compare(target, c.bound) <= 0
 }
 
-// fail records a failed SeekLE(target): no entity <= target exists.
-func (c *seekCeil) fail(target []byte) {
+// fail records a failed SeekLE(target): no entity <= target exists. A non-nil
+// err drops the proof, like seekFloor.fail.
+func (c *seekCeil) fail(target []byte, err error) {
+	if err != nil {
+		return
+	}
+
 	c.bound = append(c.bound[:0], target...)
 	c.set = true
 }
--- a/internal/storage/readstore/iterator_floor.go
+++ b/internal/storage/readstore/iterator_floor.go
@@ -1,0 +1,47 @@
+package readstore
+
+import "bytes"
+
+// seekFloor caches an exhaustion proof for a forward iterator: a failed
+// SeekGE(t) proves no entity >= t exists in the iterator's view. The view is
+// fixed at pebble.Iterator creation, so the proof never goes stale and the
+// floor is never cleared. Composite iterators re-seek exhausted children once
+// per merge step (AND converge, OR findMin, the NOT child catch-up); the floor
+// turns every such re-seek at or above the proven bound into an O(1)
+// comparison instead of a Pebble seek.
+type seekFloor struct {
+	bound []byte
+	set   bool
+}
+
+// covers reports whether SeekGE(target) is proven empty by a prior failure.
+func (f *seekFloor) covers(target []byte) bool {
+	return f.set && bytes.Compare(target, f.bound) >= 0
+}
+
+// fail records a failed SeekGE(target): no entity >= target exists. The bound
+// is copied — target may alias a sibling iterator's Pebble key buffer. Callers
+// only reach a real (uncovered) seek with target below the current bound, so
+// overwriting always tightens it.
+func (f *seekFloor) fail(target []byte) {
+	f.bound = append(f.bound[:0], target...)
+	f.set = true
+}
+
+// seekCeil is the reverse-iteration analog of seekFloor: a failed SeekLE(t)
+// proves no entity <= t exists.
+type seekCeil struct {
+	bound []byte
+	set   bool
+}
+
+// covers reports whether SeekLE(target) is proven empty by a prior failure.
+func (c *seekCeil) covers(target []byte) bool {
+	return c.set && bytes.Compare(target, c.bound) <= 0
+}
+
+// fail records a failed SeekLE(target): no entity <= target exists.
+func (c *seekCeil) fail(target []byte) {
+	c.bound = append(c.bound[:0], target...)
+	c.set = true
+}
--- a/internal/storage/readstore/iterator_floor.go
+++ b/internal/storage/readstore/iterator_floor.go
@@ -8,10 +8,11 @@ import "bytes"
 // is fixed at pebble.Iterator creation (so a proof never goes stale and the
 // floor is never cleared), and a seek that failed because of an I/O error
 // proves nothing about the view's contents (fail drops it — see #320 for the
-// error class). Composite iterators re-seek exhausted children once per merge
-// step (AND converge, OR findMin, the NOT child catch-up); the floor turns
-// every such re-seek at or above the proven bound into an O(1) comparison
-// instead of a Pebble seek.
+// error class). Exhausted children still get re-seeked: AndIterator.converge
+// seeks per merge step, and the composite SeekGE/SeekLE child loops (AND, OR,
+// NOT, ReverseOr) seek unconditionally — once per parent merge step when
+// nested. The floor turns each such re-seek at or above the proven bound into
+// an O(1) comparison instead of a Pebble seek.
 type seekFloor struct {
 	bound []byte
 	set   bool
--- a/internal/storage/readstore/iterator_not.go
+++ b/internal/storage/readstore/iterator_not.go
@@ -71,9 +71,9 @@ func (it *NotIterator) Current() []byte {
 }
 
 func (it *NotIterator) SeekGE(target []byte) bool {
-	if it.exhausted {
-		return false
-	}
+	// Absolute reposition: clear the exhausted latch so a re-seek after
+	// exhaustion still finds entities (the universe is re-seeked below).
+	it.exhausted = false
 
 	it.started = true
 
--- a/internal/storage/readstore/iterator_not.go
+++ b/internal/storage/readstore/iterator_not.go
@@ -75,19 +75,21 @@ func (it *NotIterator) SeekGE(target []byte) bool {
 		return false
 	}
 
-	if !it.started {
-		it.started = true
-		if it.child.SeekGE(target) {
-			it.childVal = it.child.Current()
-		} else {
-			it.childDone = true
-		}
-	} else if !it.childDone {
-		if it.child.SeekGE(target) {
-			it.childVal = it.child.Current()
-		} else {
-			it.childDone = true
-		}
+	it.started = true
+
+	// Re-position the child on every SeekGE, including after it has reported
+	// done. Forward Next() iteration can consume the child past a later,
+	// smaller seek target (nested NOTs reach this: the outer NOT advances the
+	// inner one to emit its first row, exhausting a finite child such as the
+	// reversion bitset). A latched childDone would then leave the child unable
+	// to report that the entity at target is excluded, leaking it into the
+	// difference. SeekGE is absolute repositioning (>= target), so re-seeking
+	// is always well-defined.
+	if it.child.SeekGE(target) {
+		it.childVal = it.child.Current()
+		it.childDone = false
+	} else {
+		it.childDone = true
 	}
 
 	if !it.universe.SeekGE(target) {
--- a/internal/storage/readstore/iterator_or.go
+++ b/internal/storage/readstore/iterator_or.go
@@ -51,9 +51,13 @@ func (it *OrIterator) Current() []byte {
 }
 
 func (it *OrIterator) SeekGE(target []byte) bool {
-	if it.exhausted || len(it.children) == 0 {
+	if len(it.children) == 0 {
 		return false
 	}
+
+	// Absolute reposition: clear the exhausted latch so a re-seek after
+	// exhaustion re-establishes the union (all children are re-seeked below).
+	it.exhausted = false
 
 	for i := range it.children {
 		it.valid[i] = it.children[i].SeekGE(target)
--- a/internal/storage/readstore/iterator_pebble.go
+++ b/internal/storage/readstore/iterator_pebble.go
@@ -188,9 +188,9 @@ func (it *PebbleAccountIterator) Current() []byte {
 }
 
 func (it *PebbleAccountIterator) SeekGE(target []byte) bool {
-	if it.exhausted {
-		return false
-	}
+	// Absolute reposition: clear the exhausted latch so a re-seek after
+	// exhaustion still finds entities (the body re-seeks from target).
+	it.exhausted = false
 
 	it.started = true
 
@@ -355,9 +355,9 @@ func (it *PebbleReverseAccountIterator) Current() []byte {
 }
 
 func (it *PebbleReverseAccountIterator) SeekLE(target []byte) bool {
-	if it.exhausted {
-		return false
-	}
+	// Absolute reposition: clear the exhausted latch so a re-seek after
+	// exhaustion still finds entities (the body re-seeks from target).
+	it.exhausted = false
 
 	it.started = true
 
@@ -519,9 +519,9 @@ func (it *PebbleTxIterator) Current() []byte {
 }
 
 func (it *PebbleTxIterator) SeekGE(target []byte) bool {
-	if it.exhausted {
-		return false
-	}
+	// Absolute reposition: clear the exhausted latch so a re-seek after
+	// exhaustion still finds entities (the body re-seeks from target).
+	it.exhausted = false
 
 	it.started = true
 
@@ -656,9 +656,9 @@ func (it *PebbleReverseTxIterator) Current() []byte {
 }
 
 func (it *PebbleReverseTxIterator) SeekLE(target []byte) bool {
-	if it.exhausted {
-		return false
-	}
+	// Absolute reposition: clear the exhausted latch so a re-seek after
+	// exhaustion still finds entities (the body re-seeks from target).
+	it.exhausted = false
 
 	it.started = true
 
@@ -843,9 +843,9 @@ func (it *PebbleTxRangeIterator) Next() bool {
 func (it *PebbleTxRangeIterator) Current() []byte { return it.current }
 
 func (it *PebbleTxRangeIterator) SeekGE(target []byte) bool {
-	if it.exhausted {
-		return false
-	}
+	// Absolute reposition: clear the exhausted latch so a re-seek after
+	// exhaustion still finds entities (the body re-seeks from target).
+	it.exhausted = false
 
 	it.started = true
 
--- a/internal/storage/readstore/iterator_pebble.go
+++ b/internal/storage/readstore/iterator_pebble.go
@@ -58,6 +58,7 @@ type PebbleAccountIterator struct {
 	current   []byte
 	started   bool
 	exhausted bool
+	floor     seekFloor
 }
 
 // newSingleTypeAccountIterator creates a forward account iterator for one attribute type.
@@ -188,6 +189,13 @@ func (it *PebbleAccountIterator) Current() []byte {
 }
 
 func (it *PebbleAccountIterator) SeekGE(target []byte) bool {
+	// A prior failed seek at or below target proves this one empty too.
+	if it.floor.covers(target) {
+		it.exhausted = true
+
+		return false
+	}
+
 	// Absolute reposition: clear the exhausted latch so a re-seek after
 	// exhaustion still finds entities (the body re-seeks from target).
 	it.exhausted = false
@@ -200,6 +208,7 @@ func (it *PebbleAccountIterator) SeekGE(target []byte) bool {
 
 	if !it.iter.SeekGE(seekKey) {
 		it.exhausted = true
+		it.floor.fail(target)
 
 		return false
 	}
@@ -218,6 +227,7 @@ func (it *PebbleAccountIterator) SeekGE(target []byte) bool {
 	}
 
 	it.exhausted = true
+	it.floor.fail(target)
 
 	return false
 }
@@ -250,6 +260,7 @@ type PebbleReverseAccountIterator struct {
 	current   []byte
 	started   bool
 	exhausted bool
+	ceil      seekCeil
 }
 
 // newSingleTypeReverseAccountIterator creates a reverse account iterator for one attribute type.
@@ -355,6 +366,13 @@ func (it *PebbleReverseAccountIterator) Current() []byte {
 }
 
 func (it *PebbleReverseAccountIterator) SeekLE(target []byte) bool {
+	// A prior failed seek at or above target proves this one empty too.
+	if it.ceil.covers(target) {
+		it.exhausted = true
+
+		return false
+	}
+
 	// Absolute reposition: clear the exhausted latch so a re-seek after
 	// exhaustion still finds entities (the body re-seeks from target).
 	it.exhausted = false
@@ -382,12 +400,14 @@ func (it *PebbleReverseAccountIterator) SeekLE(target []byte) bool {
 		// Past target, step back
 		if !it.iter.Prev() {
 			it.exhausted = true
+			it.ceil.fail(target)
 
 			return false
 		}
 	} else if !it.iter.Last() {
 		// Past end, go to last
 		it.exhausted = true
+		it.ceil.fail(target)
 
 		return false
 	}
@@ -406,6 +426,7 @@ func (it *PebbleReverseAccountIterator) SeekLE(target []byte) bool {
 	}
 
 	it.exhausted = true
+	it.ceil.fail(target)
 
 	return false
 }
@@ -442,6 +463,7 @@ type PebbleTxIterator struct {
 	current   []byte
 	started   bool
 	exhausted bool
+	floor     seekFloor
 }
 
 // NewPebbleTxIterator creates an iterator over all transactions in a ledger.
@@ -519,6 +541,13 @@ func (it *PebbleTxIterator) Current() []byte {
 }
 
 func (it *PebbleTxIterator) SeekGE(target []byte) bool {
+	// A prior failed seek at or below target proves this one empty too.
+	if it.floor.covers(target) {
+		it.exhausted = true
+
+		return false
+	}
+
 	// Absolute reposition: clear the exhausted latch so a re-seek after
 	// exhaustion still finds entities (the body re-seeks from target).
 	it.exhausted = false
@@ -531,6 +560,7 @@ func (it *PebbleTxIterator) SeekGE(target []byte) bool {
 
 	if !it.iter.SeekGE(seekKey) {
 		it.exhausted = true
+		it.floor.fail(target)
 
 		return false
 	}
@@ -543,6 +573,7 @@ func (it *PebbleTxIterator) SeekGE(target []byte) bool {
 	}
 
 	it.exhausted = true
+	it.floor.fail(target)
 
 	return false
 }
@@ -578,6 +609,7 @@ type PebbleReverseTxIterator struct {
 	current   []byte
 	started   bool
 	exhausted bool
+	ceil      seekCeil
 }
 
 // NewPebbleReverseTxIterator creates a reverse transaction iterator.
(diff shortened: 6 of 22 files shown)


flemzord:
Reviewed live head `e234888c`. The bitset and NOT child now re-seek after exhaustion, which restores the intended result for the nested-NOT/reverted contradiction covered by the new E2E regression. Functional CI and NumaryBot are green; the remaining `codecov/patch` failure is CI/tooling rather than a demonstrated runtime blocker. I found no correctness/runtime blocker on this head. GitHub approval is intentionally disabled for this cron task.

internal/storage/readstore/iterator_and.go
```diff
@@ -47,10 +47,14 @@ func (it *AndIterator) Current() []byte {
 }
 
 func (it *AndIterator) SeekGE(target []byte) bool {
-	if it.exhausted || len(it.children) == 0 {
+	if len(it.children) == 0 {
 		return false
 	}
 
+	// Absolute reposition: clear the exhausted latch so a re-seek after
```
NumaryBot: 🟠 **[major] AndIterator.SeekGE does not re-seek all children on absolute reposition**

When SeekGE is called on an AndIterator after a previous pass has advanced or exhausted a non-first child, only children[0] is repositioned. The remaining children retain stale Current() values. If any stale value exceeds the new target, converge() uses that stale value as the next candidate and can skip valid intersections. For example, with two child sets [1,2,3,10] and [1,2,3], after exhaustion at 10, a SeekGE(1) may return 3 instead of 1, violating absolute SeekGE semantics for any nested boolean filter containing an AND subtree.

**Suggestion:** In SeekGE, re-seek every child (not just children[0]) to the requested target before calling converge. Iterate over all children and call child.SeekGE(target) on each, then reset exhausted=false and proceed with convergence.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1635#issuecomment-5060665074

internal/storage/readstore/iterator_and.go
Azorlogh: Good catch — confirmed and fixed in bd8fd0eaf6733b70ceabce51302e71b263eb0291. `AndIterator.SeekGE` now re-seeks **every** child to the target (not just `children[0]`), so a stale non-first child can't become the `converge` candidate and skip intersections below it. Mirrors `OrIterator`; added a readstore regression test that fails on the prior behavior.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1635#issuecomment-5060665074

flemzord:
Revalidated live head `bd8fd0ea`. The earlier `AndIterator.SeekGE` blocker is fixed: every child is re-seeked before convergence, and the affected scoped iterators reset their exhaustion latches; focused regressions cover repositioning after exhaustion. `AddressTxIterator` remains explicitly tracked separately in EN-1597 and is outside this index-free fix scope. Functional CI and both review bots are green; the remaining `codecov/patch` failure is CI/tooling rather than a demonstrated runtime blocker. I found no correctness/runtime blocker on this head. GitHub approval is intentionally disabled for this cron task.

internal/storage/readstore/iterator_not.go
```diff
@@ -71,23 +71,25 @@ func (it *NotIterator) Current() []byte {
...
+
+	// Re-position the child on every SeekGE, including after it has reported
+	// done. Forward Next() iteration can consume the child past a later,
+	// smaller seek target (nested NOTs reach this: the outer NOT advances the
+	// inner one to emit its first row, exhausting a finite child such as the
+	// reversion bitset). A latched childDone would then leave the child unable
+	// to report that the entity at target is excluded, leaking it into the
+	// difference. SeekGE is absolute repositioning (>= target), so re-seeking
+	// is always well-defined.
+	if it.child.SeekGE(target) {
```
paul-nicolas, line 88: [Major] The unconditional child re-seek lands on `AddressTxIterator`'s surviving latch — the excluded rows still leak

*Re: `internal/storage/readstore/iterator_address.go:87-128`, which is outside this diff.*

This line is the right fix, but it is a no-op against the one iterator the sweep skipped. `AddressTxIterator.SeekGE` still opens with the verbatim `if it.exhausted { return false }` (`iterator_address.go:88`), and it is worse than latched — it is destructive: `:124-125` returns `pendingTxns[idx]` then reslices to `[idx+1:]`, and `txSeen` (`:170-174`) blocks re-materialization. It is not an absolute repositioner at all.

"Index-gated" is not "unreachable". `compileNot` (`internal/query/compile.go:387`) compiles any sub-filter as the NOT child; `compileAddressPrefix`/`compileAddressExact` build this iterator whenever `ctx.target == TRANSACTIONS` (`compile.go:1052`, `:1094`); the sole gate is `requireIndexReady` (`:1002`), which passes once the tx-address index is READY. So `not(address^"acc:")` on TRANSACTIONS builds `NotIterator{universe: PebbleTxIterator, child: AddressTxIterator}` and reaches this line. Modelled exactly: `NOT(universe={a,b,c}, child={a,b,c})` after the child is consumed returns `[a b c]` instead of `[]` — every excluded row leaks. That is the EN-1597 symptom, still live.

No test in this PR reaches it: both address specs use `ListAccountsFiltered`, and on the ACCOUNTS target `compileAddressPrefix` returns at `compile.go:1043-1045` before ever constructing an `AddressTxIterator`. The unfixed code and the coverage hole are the same hole.

**Fix:** replace the consumed `pendingTxns` slice with a stable `txns [][]byte` plus a `pos int` cursor; `SeekGE` clears `exhausted`, materializes once, `sort.Search`es the full slice and sets `pos` — never reslices. This mirrors the already-conforming `query.SliceIterator.SeekGE` (`compile.go:2023-2037`). Then add the `ListTransactionsFiltered` + `AddressPrefixFilter` nested-NOT spec (fixture needs the tx-address index READY); it fails today.

internal/storage/readstore/iterator_and.go
```diff
@@ -47,15 +47,23 @@ func (it *AndIterator) Current() []byte {
...
-		it.exhausted = true
+	// Absolute reposition: seek EVERY child to target, not just children[0].
+	// converge uses each child's Current() as a candidate, so a child left at a
+	// stale position past target would become the candidate and skip valid
+	// intersections below it. Clearing exhausted lets a re-seek after exhaustion
+	// re-establish the intersection.
+	it.exhausted = false
 
-		return false
+	for i := range it.children {
```
paul-nicolas, line 61: [Major] Seeking every child unconditionally silently drops a row against a destructively-consuming child

This one is introduced by the PR rather than merely left unfixed. The old code seeked only `children[0]`; `converge()` then seeked child *i* only when strictly behind (`:99`) — never backwards, and never onto its own position. This loop seeks every child to `target` unconditionally, **including children already sitting at `target`**.

Against `AddressTxIterator` that is destructive: `iterator_address.go:124-125` returns `pendingTxns[idx]` and then reslices to `[idx+1:]`, so `SeekGE(T)` on a child already positioned at `T` returns the row *after* `T`. Reproduced with a destructive fake: `SeekGE("a")` → `"a"`, then `SeekGE("a")` again → `"b"`. A valid intersection is dropped with no error and nothing in `Err()`.

Reachable via nested AND (`compileAnd` does not flatten `and(a, and(b,c))`, so the outer `converge` seeks the inner AND once per merge step), via `NotIterator.SeekGE` → `child.SeekGE` where the child is an AND containing an address filter, and via OR.

**Fix:** fix `AddressTxIterator` rather than guarding this loop — the loop is correct for a conforming iterator, and a guard would paper over the non-conformance. But the two changes must land together, because this one widens the blast radius of the iterator the PR chose not to fix. If `AddressTxIterator` is genuinely out of scope for this PR, this loop should not ship before it.

internal/storage/readstore/iterator_or.go
```diff
@@ -51,10 +51,14 @@ func (it *OrIterator) Current() []byte {
...
-	if it.exhausted || len(it.children) == 0 {
+	if len(it.children) == 0 {
 		return false
 	}
 
+	// Absolute reposition: clear the exhausted latch so a re-seek after
+	// exhaustion re-establishes the union (all children are re-seeked below).
+	it.exhausted = false
+
 	for i := range it.children {
```
paul-nicolas, line 62: [Medium] Dead children are re-seeked once per parent merge step, not once per query

This loop has no `valid[i]` guard, so children already known exhausted are re-seeked on every call. Pre-PR the child's own latch made that a bool check; now a dead `PrefixIterator` allocates a seekKey and runs a real `SeekPrefixGE` plus a `Valid()`/`Next()` scan every time (`iterator_leaf.go:92-127`).

The cost is not amortized over the query: `OrIterator.SeekGE` is reached from `AndIterator.converge` at `iterator_and.go:101` and `:116`, inside the per-row merge loop. For `and(or(a,b,c), d)` where `a` and `b` exhaust early, that is on the order of `2 x rows` new Pebble seeks plus allocations. Same shape in `ReverseOrIterator.SeekLE` (`iterator_reverse_or.go:74-76`) and in the now-unconditional `NotIterator` child re-seek (`iterator_not.go:88`) when the child is a Pebble iterator rather than an in-memory bitset.

Neither `internal/storage/readstore/` nor `internal/query/` has a single benchmark, so this is currently unmeasured.

**Fix:** keep the fast path without reintroducing the bug — record an exhaustion floor instead of a bare bool (`floor = target` when a seek exhausts, `floor = current` when `Next()` exhausts), then short-circuit while `bytes.Compare(target, floor) >= 0`. Sound ("nothing at or above the floor"), restores O(1) for the forward-monotonic case, and still repositions on the backward seek this PR exists to fix. Applies uniformly in place of the unconditional `exhausted = false`.

tests/e2e/business/filter_nested_not_reposition_test.go
```diff
@@ -0,0 +1,68 @@
...
+	. "github.com/onsi/gomega"
+)
+
+// Nested-NOT filters over index-free leaves (tx-id range, account address).
+// A NOT drives its child forward — exhausting a finite leaf — before the AND
+// merge seeks the tree back to a smaller key; the leaf must reposition on that
+// backward SeekGE rather than latch on exhaustion (EN-1597). Each filter here
+// is a contradiction (`and(not(F), F)`, spelled with an extra NOT to force the
+// exhaust-then-seek-back path) and must return nothing.
+var _ = Describe("Nested-NOT filter reposition", Ordered, ContinueOnFailure, func() {
```
paul-nicolas: [Medium] No readiness wait — all four specs can pass vacuously on a cold index

Every spec in this file ends in a single-shot `BeEmpty()` (`:45`, `:52`, `:59`, `:66`) and nothing asserts the 8 transactions from `BeforeAll` (`:33-38`) are queryable at all. `accounts_test.go:167-176` runs the same `ListAccountsFiltered` + `AddressPrefixFilter` shape and wraps it in `Eventually(...).Within(5 * time.Second).ProbeEvery(200 * time.Millisecond)`, with the comment *"The Pebble read index is populated asynchronously by the index builder, so we need to wait for it to catch up after writing data."* This file has no `Eventually` and no readiness wait.

So "correctly empty" is indistinguishable from "data not visible yet". That also weakens the "fail pre-fix, pass after" claim in the description — a pre-fix run could pass by accident on a cold index. Worth confirming whether these particular leaves (attributes-zone scans via `ctx.pebbleReader`) are subject to the same lag as the read-index path the sibling test guards against; either way the inconsistency with established practice in the suite is worth closing.

The sibling file added in this same PR gets this right — `reverted_filter_contradiction_test.go:31` asserts `ConsistOf(revertedIDs)` before testing the contradictions.

**Fix:** add a leading spec asserting the bare `TxIDRangeFilter(3,5)` returns exactly ids 3-5 and `AddressPrefixFilter("acc:")` returns the 8 `acc:N` accounts, wrapped in `Eventually`. Per the `iterator_not.go` comment, also add the `ListTransactionsFiltered` + `AddressPrefixFilter` variant — that is the one combination that reaches `AddressTxIterator`.

paul-nicolas (commented):
Multi-model review (Claude + Codex, 4 independent review passes + adversarial verification). **2 Major, 2 Medium** posted inline; 9 Minor/Nit listed below.

The sweep over the 13 touched iterators is **correct and complete for what it touches** — independently confirmed: every failing return path in every modified seek re-sets `exhausted`/`done` (the four unset paths are the `len(children)==0` guards at `iterator_and.go:51`, `iterator_or.go:54`, `iterator_reverse_or.go:67` and the nil-bitset guard at `iterator_bitset.go:71`, all shadowed by identical guards in `Next()`); no seek half-resets state; no consumer used the latch as a loop terminator (`paginate.go:15,53` are the only external call sites and both issue a single straight-line seek). `TestAndIterator_SeekGERepositionsAllChildren` is a genuine regression test — run at merge-base `e430164390` it fails at `:32`, and with only the latch cleared it fails at `:33` with `expected "a", actual "c"`, so it discriminates both halves of the AndIterator change.

The blocker is the one iterator left out. `AddressTxIterator` is reachable, and the AndIterator change turns its non-conformance into a silent dropped row — the two need to land together.

Two small factual notes on the description: the diff removes **13** latch guards, not 12 (the list of 12 omits `BitsetIterator`, named separately in point 1); and "Tracked in EN-1597" is self-referential with this PR's own ticket, so the remaining `AddressTxIterator` work loses its ticket on merge.

- `internal/storage/readstore/iterator_bitset_test.go:70` — the root-cause path IS covered by the e2e contradiction spec (`reverted_filter_contradiction_test.go:70`, and `Tests-E2E` runs unguarded on every PR per `main.yml:80`), so this is hygiene not a hole: one line, `require.True(t, past.SeekGE(be8(1)))`, would pin it. The four `SeekLE` sites (`iterator_pebble.go:357,658`; `iterator_reverse.go:90`; `iterator_reverse_or.go:65`) have no unit regression at all.
- `internal/storage/readstore/iterator_pebble.go:28` and `paginate.go:40` — the absolute-reposition contract is stated at 13 implementation sites (8 verbatim copies + 5 near-variants) but on neither interface; `ReverseIterator.SeekLE` has no doc comment at all. No runtime exposure in this PR; the risk is the next implementer reading only the interface.
- `internal/storage/readstore/iterator_not.go:92` — a child I/O error is indistinguishable from clean exhaustion, so `childDone=true` and NOT stops excluding. `Err()` surfaces it via `PaginateForward` so it is not silent end-to-end, but the diff rewrites this exact branch and could harden it.
- `internal/storage/readstore/iterator_bitset.go:71` — the nil-bitset return is the only `return false` in the file that does not set `done`; harmless, but the sole break in an otherwise uniform postcondition.
- `internal/storage/readstore/iterator_bitset.go:80` — `Current()` returns a stale txID after a failed seek. Latent: every consumer guards on the false return.
- `internal/storage/readstore/` and `internal/query/` contain zero benchmarks, so the seek-complexity change is unmeasured. `IteratorStats.SeekCalls` (`tracked_iterator.go:44`) is a cheap deterministic proxy if wall-clock is too noisy for CI.
- `tests/e2e/business/filter_nested_not_reposition_test.go:22` — `ContinueOnFailure` appears here and nowhere else in `tests/e2e/business` (56 files use plain `Ordered`), including its own sibling from this PR.
- `internal/storage/readstore/iterator_and_seek_test.go:1` — AndIterator tests now span two packages (`readstore_test` vs `readstore`, needed for the unexported helper). Justified, but fragments discoverability.
- `docs/technical/architecture/subsystems/read-path/query-pipeline.md:154` — the "Iterator algebra" section names the `EntityIterator` interface and every changed operator, and `:164` attributes AndIterator seek behavior to a commit this PR rewrites. Weak on precedent: comparable readstore fixes (`8845ad587`, `2f53d9747`) shipped without touching it.

flemzord (changes requested):
Revalidated current head `bd8fd0ea`. The two existing `paul-nicolas` Major threads identify one active correctness/runtime blocker: `AndIterator.SeekGE` now re-seeks every child unconditionally, but `AddressTxIterator.SeekGE` is still destructive and latched after exhaustion. Nested transaction filters containing an address iterator can therefore either leak excluded rows or skip a valid intersection. Please make `AddressTxIterator` an absolute, non-consuming repositioner (with a transaction-address nested-filter regression), or avoid widening the all-child re-seek until that iterator conforms. I am not duplicating the inline details. Functional CI and review bots are green; this request is based only on the runtime result corruption.

internal/storage/readstore/iterator_not.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/a6c148f7580d19cf1064a28a05199712913f9ac0 — `AddressTxIterator` is now a conforming absolute repositioner (materialized union + cursor; the latch is gone). Red-on-old regression test covers the NOT leak.

internal/storage/readstore/iterator_or.go
Azorlogh: Done in https://github.com/formancehq/ledger/commit/a6c148f7580d19cf1064a28a05199712913f9ac0 — exhaustion floor (`seekFloor`/`seekCeil`) added to all 8 Pebble-backed leaves; a failed absolute seek is a permanent vacancy proof for the iterator's fixed view, so dead re-seeks are one compare.

tests/e2e/business/filter_nested_not_reposition_test.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/a6c148f7580d19cf1064a28a05199712913f9ac0 — leading Eventually baseline spec (3 txs / 8 accounts visible) now gates the contradiction specs.

internal/storage/readstore/iterator_and.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/a6c148f7580d19cf1064a28a05199712913f9ac0 — the destructive consume in `AddressTxIterator.SeekGE` is gone (cursor over a stable slice), so the AND loop is now safe; regression test drives converge over an AddressTxIterator child.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1635#issuecomment-5060665074

internal/storage/readstore/iterator_not.go
NumaryBot, line 88: 🟠 **[major] NotIterator child re-seek is a no-op / destructive against AddressTxIterator**

The unconditional child.SeekGE call in NotIterator.SeekGE reaches AddressTxIterator, which still has an exhausted latch (iterator_address.go:88) and a destructive consume pattern (pendingTxns resliced at :124-125, txSeen blocks re-materialization). The NOT child is reached from compileNot whenever the sub-filter is an address predicate on a TRANSACTIONS target and the tx-address index is READY, so this is reachable in production. After the universe iterator is re-seeked correctly, the excluded child is silently wrong, causing previously-excluded rows to leak back into results.

**Suggestion:** Fix AddressTxIterator.SeekGE to be a true absolute repositioner: drop the exhausted latch guard, reset pendingTxns and txSeen, and re-materialize from the new target. This is the root fix needed by both NotIterator and AndIterator callers.

internal/storage/readstore/iterator_and.go
NumaryBot, line 61: 🟠 **[major] Unconditional re-seek of every child silently drops rows against destructively-consuming children**

The new loop seeks every child to `target` unconditionally, including children already positioned at `target`. Against AddressTxIterator this is destructive: SeekGE(T) on a child already at T returns the row after T (pendingTxns is resliced at iterator_address.go:124-125), so a valid intersection is silently dropped with nothing in Err(). This is reachable via nested AND, via NotIterator.SeekGE → child.SeekGE where the child is an AND containing an address filter, and via OR.

**Suggestion:** Fix AddressTxIterator.SeekGE to be idempotent / non-destructive on its current position (true absolute repositioner). With a conforming iterator the unconditional loop is correct; without it, guarding the loop is only a partial mitigation.

internal/storage/readstore/iterator_or.go
NumaryBot, line 62: 🟡 **[minor] Dead children re-seeked on every parent merge step, not once per query**

The SeekGE loop has no guard for already-exhausted children, so a child that exhausted early is re-seeked on every call to OrIterator.SeekGE. Because this is invoked from AndIterator.converge inside the per-row merge loop, for `and(or(a,b,c), d)` with early-exhausting a and b, this generates O(rows) extra Pebble seeks and allocations. The same pattern exists in ReverseOrIterator.SeekLE and the now-unconditional NotIterator child re-seek. There are no benchmarks in the readstore or query packages, so the cost is currently unmeasured.

**Suggestion:** Track an exhaustion floor per child (e.g., `floor = target` when a seek exhausts, `floor = current` when it lands) and skip re-seeking children whose floor already covers the new target. This avoids reintroducing the original bug while eliminating redundant seeks.

tests/e2e/business/filter_nested_not_reposition_test.go
```diff
@@ -0,0 +1,85 @@
...
```
NumaryBot: 🟡 **[minor] No index-readiness wait — specs can pass vacuously on a cold index**

Every spec ends with a single-shot BeEmpty() assertion (lines ~45, 52, 59, 66) but there is no Eventually/readiness wait to confirm the 8 seeded transactions from BeforeAll are actually queryable before the check runs. A cold or lagging index would make the result empty for the wrong reason, so a pre-fix run could pass by accident. The sibling file added in this PR (reverted_filter_contra...) does include such a wait, and accounts_test.go:167-176 documents the pattern with an explicit comment about the async Pebble read-index builder.

**Suggestion:** Add an Eventually poll (matching the pattern in accounts_test.go:167-176) that first confirms the seeded transactions are visible before asserting the filtered result is empty. This ensures the test distinguishes 'correctly empty' from 'data not yet indexed'.

flemzord:
Revalidated live head `a6c148f7`. The prior correctness blocker is fixed: `AddressTxIterator` now provides absolute, repeatable repositioning over a stable materialized slice, and the AND/NOT regressions plus readiness guard cover the affected paths. Functional CI is green; the remaining Codecov patch failure is CI/tooling, and the current NumaryBot findings describe the pre-fix implementation. I found no remaining correctness/runtime blocker. GitHub approval is intentionally disabled by this cron configuration.

internal/storage/readstore/iterator_not.go
Azorlogh: Stale — pre-a6c148f75 code; raised and fixed in https://github.com/formancehq/ledger/pull/1635#discussion_r3664092700.

internal/storage/readstore/iterator_and.go
Azorlogh: Stale — pre-a6c148f75 code; raised and fixed in https://github.com/formancehq/ledger/pull/1635#discussion_r3664092704.

internal/storage/readstore/iterator_or.go
Azorlogh: Stale — pre-a6c148f75 code; raised and fixed in https://github.com/formancehq/ledger/pull/1635#discussion_r3664092707.

tests/e2e/business/filter_nested_not_reposition_test.go
Azorlogh: Stale — pre-a6c148f75 code; raised and fixed in https://github.com/formancehq/ledger/pull/1635#discussion_r3664092710.

flemzord (approved):

flemzord (dismissed):

internal/storage/readstore/iterator_pebble.go
```diff
@@ -58,6 +58,7 @@ type PebbleAccountIterator struct {
 	current   []byte
 	started   bool
 	exhausted bool
+	floor     seekFloor
```
paul-nicolas, line 75: [Medium] The absolute-reposition contract is never documented on the interface

Re: `EntityIterator.SeekGE` at **lines 28-30** of this file (outside the diff, so anchored here).

This PR's entire premise is that `SeekGE`/`SeekLE` are *absolute* repositioning -- computed from `target`, idempotent at the same target, valid after exhaustion or after a prior failed seek, and non-consuming. You conformed 12+ implementations plus `AddressTxIterator` to that. But the interface documents only the positional result:

```go
// SeekGE positions the iterator at the first entity >= target.
// Returns false if no such entity exists OR on I/O error -- see Err().
SeekGE(target []byte) bool
```

Nothing there forbids latching, requires a backward or repeated seek to work, or says a seek must not consume `Current()`. `iterator_address.go:22` appeals to "as the `EntityIterator` contract requires" -- a contract with no normative source. EN-1597 exists precisely because this was unwritten, so the next implementer has nothing to conform to and the latch class of bug can trivially reappear.

**Fix:** extend the `SeekGE` doc block with the three guarantees (absolute positioning from `target`; idempotent at the same target; well-defined after exhaustion or a prior failed seek), and state that a failed seek leaves the iterator un-positioned but still re-seekable. Mirror the same text onto `ReverseIterator.SeekLE` (`paginate.go:37-42`), which currently has no doc comment at all.

internal/storage/readstore/iterator_pebble.go
```diff
@@ -355,10 +366,17 @@ func (it *PebbleReverseAccountIterator) Current() []byte {
 }
 
 func (it *PebbleReverseAccountIterator) SeekLE(target []byte) bool {
```
paul-nicolas, line 382: [Medium] The entire reverse/`SeekLE` + `seekCeil` half of this change ships with zero test coverage

`PebbleReverseAccountIterator.SeekLE` (here), `PebbleReverseTxIterator.SeekLE` (`:690`), and `ReversePrefixIterator.SeekLE` (`iterator_reverse.go:92`) all had their exhaustion latch removed and gained a `seekCeil`; `ReverseOrIterator.SeekLE` lost its `done` latch. **No test in the repo constructs any of them** -- a repo-wide grep over `*_test.go` for `PebbleReverseTxIterator`, `PebbleReverseAccountIterator` and `ReversePrefixIterator` returns nothing, and the only two test files mentioning `SeekLE` use hand-rolled fakes whose `SeekLE` just delegates to `Next`.

`iterator_floor_test.go:30 TestSeekCeil` exercises the bare struct only, while the forward analogue *did* get a real integration test (`TestPrefixIterator_SeekFloorKeepsRepositioning`). So three distinct `ceil.fail` sites in `PebbleReverseTxIterator.SeekLE` alone (`:714`, `:721`, `:741`) are never executed by any test.

This matters more than a coverage count: the reverse contract is strictly harder than the forward one (seek-then-step-back with three failure branches), `PaginateReverse` is a live production read path, and a separate audit found that `seekCeil.covers()` can never even fire in production -- reverse pagination seeks once on a fresh iterator and then only calls `Next()`. So the new reverse caching is both unexercised by tests and unreachable in practice.

**Fix:** add a reverse mirror of `TestPrefixIterator_SeekFloorKeepsRepositioning` for `ReversePrefixIterator`, plus one exhaust-then-seek-back test for `PebbleReverseTxIterator` covering the `Prev()`-fails and `Last()`-fails branches. If `seekCeil` genuinely cannot fire, consider dropping it rather than shipping untested dead caching.

internal/storage/readstore/iterator_floor.go
```diff
@@ -0,0 +1,47 @@
...
+// covers reports whether SeekGE(target) is proven empty by a prior failure.
+func (f *seekFloor) covers(target []byte) bool {
+	return f.set && bytes.Compare(target, f.bound) >= 0
+}
+
+// fail records a failed SeekGE(target): no entity >= target exists. The bound
+// is copied — target may alias a sibling iterator's Pebble key buffer. Callers
+// only reach a real (uncovered) seek with target below the current bound, so
+// overwriting always tightens it.
+func (f *seekFloor) fail(target []byte) {
```
paul-nicolas: [Medium] All 19 `fail()` sites treat an I/O-induced false return as a proof of emptiness

The mechanism's contract says a failed seek *proves* no entity exists past the bound, and the bound is then never cleared. But `EntityIterator.SeekGE` is explicitly documented to return false **"if no such entity exists OR on I/O error"** (`iterator_pebble.go:29`). Every `floor.fail(target)` / `ceil.fail(target)` call site conflates the two: I checked all 19 and **none** guards on `it.iter.Error()`. A transient Pebble fault (block checksum, blob read error -- exactly the class #320 was about) is therefore recorded as a permanent "proven empty" bound.

Today this does not produce a silent wrong result: Pebble's `Error()` is sticky and `paginate.go:30` returns `iter.Err()` unconditionally, so the query fails loudly instead of returning short rows. But that means the safety of this cache rests entirely on an **unstated second-order property** of two other components, while the comment right above asserts an invariant the code does not establish. One future `floor` consumer that does not check `Err()` turns this into silent wrong results -- and under a `NOT`, a leaf that wrongly reports empty stops excluding, which is exactly the EN-1597 failure mode arriving through a different door.

**Fix:** guard every site, e.g.

```go
if it.iter.Error() == nil {
	it.floor.fail(target)
}
```

and amend the comment to state the precondition explicitly: the proof is valid only for a clean exhaustion of a snapshot-fixed view.

internal/storage/readstore/iterator_floor.go
```diff
@@ -0,0 +1,47 @@
+package readstore
+
+import "bytes"
+
+// seekFloor caches an exhaustion proof for a forward iterator: a failed
```
paul-nicolas, line 5: [Medium] Documentation drift: a new mechanism with no page, and an existing page this PR falsifies

This PR touches **no** documentation, and two things now need it.

**1. `seekFloor`/`seekCeil` is a new technical mechanism with a non-obvious invariant** -- the Pebble view is fixed at iterator creation, therefore a failed-seek proof never expires and the bound is never cleared. That precondition lives only in this source comment, so it is silently violated the day someone hands one of these iterators a non-snapshot view. `CLAUDE.md` (Documentation Maintenance) requires a dedicated page under `docs/technical/architecture/` linked from the corresponding `README.md`; `docs/technical/architecture/subsystems/read-path/README.md` lists five documents and none covers iterator seek semantics.

**2. `docs/technical/architecture/subsystems/read-path/query-pipeline.md:164` is now stale.** It advertises:

> Recent work (commit `7662d2bae`) added monotonic-skip and probe-based optimisations to the `AndIterator` to short-circuit when one child runs ahead of the others.

That is the *old* `SeekGE` behaviour -- seek `children[0]` only and let `converge` leapfrog off an ahead child. `iterator_and.go:54-67` now force-seeks every child to `target`, deliberately discarding an ahead child's position, so the leapfrog survives only inside `converge`'s `cmp > 0` branch. Anyone reading that page to reason about query cost now reasons from false semantics; `prepared-queries.md` repeats the claim.

**Fix:** document the absolute-seek contract and the floor/ceil snapshot precondition (either a new `iterator-seek-contract.md` linked from the read-path README, or a section in `query-pipeline.md`'s "Iterator algebra"), and correct lines 154-164 to say that `SeekGE` repositions every child.

tests/e2e/business/filter_nested_not_reposition_test.go
paul-nicolas: [Medium] `ContinueOnFailure` neuters the anti-vacuity guard, and the comment justifying it is factually wrong

Two separate problems here.

**1. The guard does not guard.** The baseline spec exists explicitly to "Guard against vacuous passes", but `ContinueOnFailure` on the `Ordered` container makes Ginkgo run the four contradiction specs *even when that baseline fails*. Those four assert only `BeEmpty()`, so against an empty view they all pass and the guard protects nothing -- the suite would stay green on the very regression it is meant to catch. This is also the only `ContinueOnFailure` in the whole `tests/` tree.

**2. The stated rationale is wrong.** The comment says:

> The Pebble read index is populated asynchronously by the index builder, so wait for both leaf shapes to return the seeded rows...

Neither leaf touches the async read index. `TxIDRangeFilter` compiles via `compileTxIDCondition`, whose own comment reads *"No index is required -- the Pebble cold zone is always present and sorted by txID"* (`internal/query/compile.go:1198-1199`), and it builds `NewPebbleTxRangeIterator(ctx.pebbleReader, ...)`. `AddressPrefixFilter` on the ACCOUNTS target uses the always-on main-store existence index. Both read `pebbleReader` -- the FSM-written main store, synchronous with the `Apply` that already returned.

**Fix:** move the `Eventually` into `BeforeAll` (a `BeforeAll` failure *does* abort the container's specs, which is the behaviour you want), drop `ContinueOnFailure`, and correct the comment to state that both leaves are main-store / FSM-synchronous.

Worth noting separately: `TxIDRangeFilter(3, 5)` carries both bounds and compiles to `PebbleTxRangeIterator`, so this file does not exercise `RangeIterator` at all -- and `RangeIterator` has no unit test anywhere in the repo.

internal/storage/readstore/iterator_address.go
```diff
@@ -85,44 +73,21 @@ func (it *AddressTxIterator) Current() []byte {
 }
 
 func (it *AddressTxIterator) SeekGE(target []byte) bool {
```
paul-nicolas, line 75: [Medium] The PR description is stale, and omits the largest new abstraction in the diff

The body states:

> **Excluded:** `AddressTxIterator.SeekGE` destructively consumes a materialized slice, so it needs restructuring rather than a latch reset, and it is index-gated (unreachable on the index-free paths). Tracked in EN-1597.

But the head commit is `a6c148f75 fix(EN-1597): make AddressTxIterator a conforming absolute repositioner`, and the diff rewrites exactly this method with a `pos` cursor, `sort.Search` and `ensureMaterialized`, plus 155 lines of new tests. A reviewer reading the description would skip the file.

Two further inaccuracies:

- The three-item "Changes" list and the "Tests" list never mention `seekFloor`/`seekCeil`, `iterator_floor.go`, `iterator_floor_test.go` or `iterator_address_test.go`. The new exhaustion-proof cache is the largest new abstraction in the PR and is invisible from the description -- reviewers are told to expect a latch-clearing pass only.
- The "Uniform pass" list names `RangeIterator` among the iterators made able to reposition after exhaustion. Its `SeekGE` still cannot reposition at all when `len(lowerBound) < entityOffset` (three compile paths build that shape), and the method is unreachable in production anyway because every construction site drains it via `materializeIterator`. Claiming it as fixed is misleading.

**Fix:** convert the "Excluded" paragraph into a `Changes` item describing the `AddressTxIterator` restructuring; add an item for `seekFloor`/`seekCeil` stating its snapshot-immutability precondition; list the three new unit-test files under "Tests"; and qualify the `RangeIterator` claim (or drop it and track the real defect separately).

paul-nicolas (commented):

**6 Medium, 8 Minor/Nit, 0 Critical, 0 Major.** 14 findings entered adversarial verification; 6 confirmed at Medium, 7 downgraded, 1 rejected. Only Medium findings are posted inline.

**The core fix is sound.** I verified the latch-clearing empirically by running probes against the PR head: backward seek after exhaustion, repeated seeks at the same target, and termination all behave correctly for `OrIterator`, `ReverseOrIterator`, `NotIterator`, nested `NOT`, and `AndIterator` over real `PrefixIterator` leaves. `started` is correctly set on every modified seek path. `internal/storage/readstore` + `internal/query` unit suites pass, and `go vet` is clean including `-tags e2e`. The `seekFloor` soundness audit found **no** counterexample on `SeekPrefixGE`/Comparer-Split, multi-prefix iterators, `fail()` bound-tightening, `target`/`f.bound` aliasing, or variable-length targets -- that claim holds. Everything below concerns the edges of the change, not its substance.

### Notable downgrades, for the record

- `converge`'s missing `cmp < 0` re-check after the forward re-seek **is** a textual silent false-positive (reproduced: `And({b,c},{a})` emits `[b c]`), but it is **dead code**: every real iterator except `BitsetIterator` verifies the `>= target` postcondition before returning true, and converge always passes an exact 8-byte entity ID. Filed as a hardening nit, not a bug.
- The `RangeIterator.SeekGE` prefix-less seek key (real, reproduced: forward pass yields rows yet `SeekGE` fails and poisons the floor) is **pre-existing, byte-identical to the merge base, behaviour-neutral vs base, and unreachable** -- all four `NewRangeIterator` sites hand the iterator to `materializeIterator`, which drains it with `Next()` and closes it. Minor, and not a defect this PR introduced.
- One reviewer proposed optimising `AndIterator.SeekGE` by skipping children already `>= target`. **That proposal is unsound and must not be applied** -- verified empirically: on this PR's own fixture it returns `"c"` instead of `"a"`, reinstating the pre-PR bug, because a child can be stale-*ahead* (positioned at `"c"` after a forward pass while still holding `"a"`). It would fail your own `TestAndIterator_SeekGERepositionsAllChildren`. The seek amplification is real but is a documented correctness-for-performance tradeoff.

- `internal/storage/readstore/iterator_leaf.go:259` -- `RangeIterator.SeekGE` builds a prefix-less seek key when `len(lowerBound) < entityOffset`; three compile paths produce that shape (`compile.go:1330-1332`, ~`:700`, ~`:869`; `:1441` unaffected). Pre-existing and unreachable via `materializeIterator`; these indexes are value-ordered so entity-ordered `SeekGE` is meaningless there. Latent trap for a future author who wires one into a tree without materialising.
- `internal/storage/readstore/iterator_bitset.go:83` (Minor) -- `SeekGE` silently ignores `len(target) < 8` (and truncates `> 8`), repositioning to the first set bit below target. Client-reachable: `decodeCursor` returns `cursorData.After` with no width check anywhere. Impact is bounded -- a bare `reverted=true` (or an `OR` containing the bitset) ignores the cursor for one page and self-corrects on the next; `reverted=false` (NOT over bitset) is unaffected because `NotIterator` anchors output to `universe.Current()`. Pre-existing (`0713a7669`), but this PR is the SeekGE-conformance change, so it is the one iterator left non-conforming.
- `internal/storage/readstore/iterator_and.go:107` (Nit) -- add a defensive `cmp < 0` failure after the re-seek to pin the postcondition the merge now relies on.
- `internal/storage/readstore/iterator_pebble.go:706` (Minor) -- `incrementUint64Bytes` wraps at `MaxUint64`, so `SeekLE(MaxUint64)` returns false and caches a ceiling of `0xff..ff` that `covers()` satisfies for every target. Found independently by two reviewers; the ceiling is never read in production.
- `internal/storage/readstore/paginate.go:40` (Minor) -- `ReverseIterator.SeekLE` has no doc comment at all, yet 4 implementations changed semantics.
- `internal/storage/readstore/iterator_address.go:166` (Minor) -- `insertSorted` is O(n^2); `sort` is now imported for `sort.Search`, so append-then-`sort.Slice` is a one-liner.
- `internal/storage/readstore/iterator_not.go:88` (Minor) -- NOT collapses a child I/O failure into "child exhausted" and keeps scanning. Mitigated by `Err()` propagation through `PaginateForward`, so cleanup debt rather than a wrong result.
- `internal/storage/readstore/iterator_and.go:61` (Nit) -- seek amplification: `SeekGE` costs k seeks and forces backward re-descents where an ahead child previously cost 0; `seekFloor` cannot mitigate backward seeks (`covers()` requires `target >= bound`). Consciously documented tradeoff.

_Reviewers: Claude (3 parallel, split by area) + Codex, with a separate adversarial verification pass per finding. Findings that could not be substantiated against the real code were dropped rather than posted._

internal/storage/readstore/iterator_pebble.go
Azorlogh: Added the ReversePrefixIterator mirror and PebbleReverseTxIterator Prev()/Last()-fails branch tests: https://github.com/formancehq/ledger/commit/224785e4a0b4038ff33882e8badbdbad64df8b58. Kept seekCeil — now tested, and symmetric with the load-bearing forward floor.

internal/storage/readstore/iterator_floor.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/224785e4a0b4038ff33882e8badbdbad64df8b58 — fail() now takes the iterator's storage error (compiler-enforced at all 19 sites) and drops the proof when non-nil; the clean-exhaustion precondition is in the seekFloor doc.

internal/storage/readstore/iterator_pebble.go
Azorlogh: Documented on both interfaces, plus a dedicated page (iterator-seek-contract.md): https://github.com/formancehq/ledger/commit/224785e4a0b4038ff33882e8badbdbad64df8b58

internal/storage/readstore/iterator_floor.go
Azorlogh: Added iterator-seek-contract.md (linked from the read-path README) and corrected query-pipeline.md / prepared-queries.md: https://github.com/formancehq/ledger/commit/224785e4a0b4038ff33882e8badbdbad64df8b58

tests/e2e/business/filter_nested_not_reposition_test.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/224785e4a0b4038ff33882e8badbdbad64df8b58 — guard moved into BeforeAll, ContinueOnFailure dropped, comment corrected (those leaves are FSM-synchronous). The tx-address nested-NOT specs and a first RangeIterator unit test landed in https://github.com/formancehq/ledger/commit/2ee79d7886cedeeda11e4d42627946372640867e.

internal/storage/readstore/iterator_address.go
Azorlogh: Description updated. Note RangeIterator.SeekGE now fails loudly (invariant error) instead of carrying the caveat — it's unimplementable on the raw (value, entity)-ordered scan and every site materializes: https://github.com/formancehq/ledger/commit/2ee79d7886cedeeda11e4d42627946372640867e

internal/storage/readstore/iterator_pebble.go
```diff
@@ -58,6 +71,7 @@ type PebbleAccountIterator struct {
 	current   []byte
 	started   bool
 	exhausted bool
+	floor     seekFloor
```
NumaryBot, line 75: 🟠 **[major] SeekGE/SeekLE absolute-reposition contract is undocumented on the interface**

The EntityIterator interface documents SeekGE/SeekLE only as positional operations. It does not state that implementations must be absolute repositioners (idempotent at the same target, valid after exhaustion, non-consuming, supporting backward seeks). This PR conformed 12+ implementations and AddressTxIterator to that contract, but the contract itself has no normative source. EN-1597 exists precisely because this was unwritten — the latch class of bug can trivially reappear in any new implementation.

**Suggestion:** Extend the SeekGE and SeekLE doc blocks on the EntityIterator interface with the three guarantees: (1) absolute positioning from target, not relative to current position; (2) idempotency — calling SeekGE(T) twice must leave the iterator at T both times; (3) non-consuming — the seek must not advance past the found element. Reference these guarantees from iterator_address.go and other implementations.

internal/storage/readstore/iterator_pebble.go
```diff
@@ -355,10 +379,17 @@ func (it *PebbleReverseAccountIterator) Current() []byte {
 }
 
 func (it *PebbleReverseAccountIterator) SeekLE(target []byte) bool {
```
NumaryBot, line 382: 🟠 **[major] Reverse SeekLE path ships with zero test coverage**

PebbleReverseAccountIterator.SeekLE, PebbleReverseTxIterator.SeekLE, and ReversePrefixIterator.SeekLE all had their exhaustion latch removed and gained a seekCeil. ReverseOrIterator.SeekLE lost its done latch. No test in the repository exercises any of these paths — a repo-wide search for the concrete reverse iterator types in test files returns nothing, and the only two test files mentioning SeekLE use hand-rolled fakes. Three distinct ceil.fail sites in PebbleReverseTxIterator.SeekLE alone are never executed. The reverse contract is strictly harder than the forward one (seek-then-step-back with multiple failure branches), and PaginateReverse is a live production read path.

**Suggestion:** Add integration tests for PebbleReverseAccountIterator.SeekLE, PebbleReverseTxIterator.SeekLE, and ReversePrefixIterator.SeekLE covering: normal reposition, reposition after exhaustion, backward reposition, and the error/empty-result branches. Mirror the pattern of TestPrefixIterator_SeekFloorKeepsRepositioning added for the forward path.

internal/storage/readstore/iterator_floor.go
```diff
@@ -0,0 +1,62 @@
...
+	set   bool
+}
+
+// covers reports whether SeekGE(target) is proven empty by a prior failure.
+func (f *seekFloor) covers(target []byte) bool {
+	return f.set && bytes.Compare(target, f.bound) >= 0
+}
+
+// fail records a failed SeekGE(target): no entity >= target exists. err is the
+// iterator's storage error: a non-nil err means the seek failed for I/O
```
NumaryBot, line 27: 🟡 **[minor] All fail() sites conflate I/O error with proven emptiness, creating latent silent-wrong-result risk**

floor.fail(target) and ceil.fail(target) permanently cache a seek failure as proof that no entity exists past the bound. EntityIterator.SeekGE returns false for both 'no such entity' and 'I/O error'. None of the 19 call sites check iter.Error() before recording the permanent bound. Today this does not produce silent wrong results because Pebble errors are sticky and paginate.go returns Err() unconditionally — but correctness rests on an unstated second-order property of two other components. A future consumer that does not check Err(), or a NOT iterator over a leaf that wrongly reports empty due to a transient fault, would produce silent short results.

**Suggestion:** Guard each fail() call site: only record the permanent bound if iter.Err() == nil. If there is an error, skip caching and let the error propagate normally. Add a comment explaining this invariant explicitly.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1635#issuecomment-5060665074

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1635#issuecomment-5060665074

internal/storage/readstore/iterator_pebble.go
Azorlogh: Stale — scan predates 224785e4a, which documented the contract on both interfaces plus a dedicated page (iterator-seek-contract.md): https://github.com/formancehq/ledger/commit/224785e4a0b4038ff33882e8badbdbad64df8b58

internal/storage/readstore/iterator_pebble.go
Azorlogh: Mostly stale — ReversePrefixIterator + PebbleReverseTxIterator branch tests landed in https://github.com/formancehq/ledger/commit/224785e4a0b4038ff33882e8badbdbad64df8b58. The two remaining mirrors (PebbleReverseAccountIterator, ReverseOrIterator) are in https://github.com/formancehq/ledger/commit/a335778d6.

internal/storage/readstore/iterator_floor.go
Azorlogh: Stale — scan predates 224785e4a: fail() now takes the iterator's storage error (all 19 sites, compiler-enforced) and drops the proof when non-nil: https://github.com/formancehq/ledger/commit/224785e4a0b4038ff33882e8badbdbad64df8b58

internal/storage/readstore/iterator_pebble.go
```diff
@@ -672,12 +724,14 @@ func (it *PebbleReverseTxIterator) SeekLE(target []byte) bool {
 	if it.iter.SeekGE(seekKey) {
 		if !it.iter.Prev() {
 			it.exhausted = true
+			it.ceil.fail(target, it.iter.Error())
```
NumaryBot: 🟠 **[major] MaxUint64 reverse seek incorrectly cached as proof of emptiness due to overflow**

When SeekLE is called with an all-0xff transaction cursor, incrementUint64Bytes(target) wraps to zero. The resulting SeekGE lands on the first key, and the subsequent Prev() fails (no predecessor), causing ceil.fail(target, nil) to record MaxUint64 as a permanent proof of emptiness. Every later SeekLE on the same iterator is then short-circuited by ceil.covers and returns empty, even though rows at or below the original target exist. This is a correctness regression introduced by the new seekCeil mechanism on the reverse path.

**Suggestion:** Detect the overflow case before calling SeekGE(incremented): if incrementUint64Bytes overflows (i.e. the result is all zeros or a wrap indicator), fall through to the Last()-based path rather than the SeekGE probe, and do not call ceil.fail for that case. This treats an all-0xff target as a past-end marker consistent with how Last() is used elsewhere in the method.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1635#issuecomment-5060665074

internal/storage/readstore/iterator_pebble.go
Azorlogh: Confirmed and fixed in https://github.com/formancehq/ledger/commit/eb4614fde2ca3172cde3556becfecd1d67f9c8c1 — an all-0xff target now positions at the end of the range directly instead of probing with the wrapped key; regression test included (red pre-fix).

paul-nicolas (approved):
LGTM

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1635#issuecomment-5060665074

PR #1636, opened 2026-07-23
fix(check): seed replay with baseline transaction states under archiving

Azorlogh:
## Problem

Under archiving, `store check` false-positives with `TRANSACTION_UPDATE_MISMATCH` on any transaction that was created before the archive boundary and later touched (metadata set/delete, or reverted).

`compareTransactions` builds the expected state as **"replay overrides baseline"**:

```go
var expected *commonpb.TransactionState
if rs, ok := replayTx[key]; ok {
    expected = rs                 // ← replay wins outright
} else if bs, ok := baselineTx[key]; ok {
    expected = bs
}
```

That's correct only when the replay is **complete** — a non-archived run that replays every log from seq 1. Under archiving the archived chapters' logs are purged, so the replay walks only the post-boundary delta. For a transaction created before the boundary (its full state in the baseline checkpoint) and merely *updated* after it, the replay finalizes to a **partial** state (e.g. just the one metadata key), and overriding the baseline throws away the create/postings/earlier metadata.

The stored `TransactionState` is correct — it lives in the attribute zone, which archival does **not** purge, and is updated in place — so `expected` (partial) ≠ live (full) → false mismatch. It trips on any archived transaction later reverted or given metadata, which is normal.

`compareVolumes` handles the identical situation correctly (`expected = baseline + delta`); transactions used override and were never made archive-safe.

## Fix

Seed the replay with each baseline transaction state (as a `txOpFinalized` merge base) **before** the log-replay pass, so post-archive delta operands merge on top of the full pre-archive state and `replayTx` becomes complete. Override then yields the correct result, mirroring the volume path's baseline+delta.

A finalized-level merge inside `compareTransactions` can't be correct: a post-archive metadata **delete** is just an absent key in the replay, so unioning with the baseline would wrongly re-add it. Seeding the base *before* the operands lets the merger apply the delete correctly.

`compareTransactions` is the **sole** reader of the replay's transaction prefix (the other hits are log payloads), so this is isolated — reversion and post-commit-volume checks are unaffected. The full `internal/application/check` suite passes.

## Testing

Adds a checker unit test (`checker_archived_tx_test.go`): a transaction created pre-boundary (baseline) with a post-boundary metadata set is

- flagged as tampered by an **unseeded** replay (reproduces the bug), and
- **clean** once the baseline is seeded (verifies the fix).

Full check-package suite green; module builds; lint clean.

## Notes

Surfaced by the archival model test's teardown `CheckStore` (a `--restore --archive` run reported 5 `TRANSACTION_UPDATE_MISMATCH`). Pre-existing checker gap, independent of the cold-read read-path fixes (#1623, #1633).


flemzord:
Reviewed live head `237c7fd1`. Seeding each baseline `TransactionState` as the replay merge base before post-archive operations preserves the full pre-archive state while still applying metadata deletes/sets and reversion updates in order; the regression reproduces the prior false mismatch and validates the seeded path. Functional CI and NumaryBot are green; the remaining `codecov/patch` failure is CI/tooling rather than a demonstrated runtime blocker. I found no correctness/runtime blocker on this head. GitHub approval is intentionally disabled for this cron task.

flemzord (changes requested):
Revalidated live head `237c7fd1`; this supersedes my earlier neutral comment. One correctness/runtime blocker remains in the new baseline seeding path: it seeds every archived `TransactionState` by canonical `(ledger, txID)` key, but replaying `DeleteLedger` does not clear those seeded transaction entries. If the ledger is then recreated under the same name, transaction IDs restart at 1, so the new `txOpCreate` merges onto the old generation state. The create op overwrites some fields but does not reset absent metadata, reversion fields, or a nil timestamp, allowing old-generation data to survive and causing a false `TRANSACTION_UPDATE_MISMATCH` against the correct live state. Please clear the replay transaction prefix for that ledger on `DeleteLedger` (or make create replace the state safely) and add an archive → delete → same-name recreate → reused txID regression. Functional CI is green; the separate Codecov patch failure is not the basis of this review.

flemzord (approved):

flemzord (changes requested):
Revalidated live head `237c7fd1`. One correctness/runtime blocker remains: baseline transaction seeds survive a replayed `DeleteLedger`, so recreating the same ledger and reusing a transaction ID merges the new `txOpCreate` onto the old generation. Because create does not clear absent metadata, reversion fields, or a nil timestamp, the checker can retain old-generation state and emit a false `TRANSACTION_UPDATE_MISMATCH`. Please clear the replay transaction range for the deleted ledger (or make create fully replace the prior state) and cover archive → delete → same-name recreate → reused txID. The `codecov/patch` failure is CI/tooling and is not the basis of this review.

flemzord:
Correction after re-reading the full live lifecycle on `237c7fd1`: the same-name recreation scenario in my latest review is unreachable. `processCreateLedger` returns `ErrLedgerDeleted` for a tombstoned `LedgerInfo`, and `deleteLedgerData` deliberately retains that `LedgerInfo`, so the name cannot be recreated after cleanup. I therefore withdraw that blocker; I found no correctness/runtime blocker introduced by this head. The remaining `codecov/patch` failure is CI/tooling. GitHub approval is intentionally disabled by this cron configuration.

flemzord (changes requested):
Revalidated current head `633ac8f7`. One correctness/runtime blocker remains in the replay merger: `txMerger.Finish` ignores Pebble’s `includesBase` flag and always turns the operands it sees into `txOpFinalized`. Pebble may compact only a suffix of operands (`includesBase=false`) and requires the merge operator to be associative. For example, baseline metadata `{k0}` followed by a post-archive delete of `k0` can have that delete compacted alone into an empty finalized state. When the full chain is later merged, `state.UnmarshalVT` at `replay_store.go:604` does not clear fields/map entries absent from the partial finalized payload, so baseline `k0` survives and the checker again reports a compaction-dependent false `TRANSACTION_UPDATE_MISMATCH`.

Please preserve the ordered raw delta semantics when `includesBase == false` (for example with an op-batch encoding), or otherwise make the merger genuinely associative, and add a forced flush/compaction regression for baseline metadata followed by a post-archive delete. The separate `codecov/patch` failure is CI/tooling and is not the basis of this review.

flemzord (changes requested):
Revalidated live head 237c7fd1 after the 27 July force-push back from 633ac8f7. The existing replay-merger blocker from my review on 633ac8f7 remains active on this head: txMerger.Finish still ignores includesBase and emits a finalized state for partial Pebble merges, so delta compaction is non-associative and can resurrect omitted baseline fields, producing compaction-dependent false TRANSACTION_UPDATE_MISMATCH results. I am only confirming that the already-reported blocker survives the force-push. Please preserve raw delta semantics when includesBase is false, or otherwise make the merger associative, and retain a forced-compaction regression. The codecov/patch failure is CI/tooling and is not the basis of this review.

flemzord:
Revalidated live head `1d215bf1`. The new `txOpBatch` path fixes the prior compaction-associativity blocker: partial merges (`includesBase=false`) preserve the ordered transaction operations, and only a base-inclusive fold resolves them, so metadata deletes cannot be lost and resurrected from the seeded baseline. The focused merger test and the flush/compaction regression cover both levels. Functional CI and both review bots are green; the remaining `codecov/patch` failure is CI/tooling rather than a demonstrated runtime blocker. I found no correctness/runtime blocker on this head. GitHub approval is intentionally disabled by this cron configuration.

internal/application/check/replay_store.go
```diff
@@ -573,7 +606,9 @@ func (m *txMerger) Finish(_ bool) ([]byte, io.Closer, error) {
 			delete(state.GetMetadata(), metaKey)
 
 		case txOpFinalized:
-			// Re-ingesting a previously finalized state (from a prior compaction).
+			// A resolved snapshot: a seeded baseline, or the output of a base-inclusive
+			// fold. It is always the oldest operand for the key, so it lands on an empty
+			// accumulator and the overlay coincides with a replace.
 			if err := state.UnmarshalVT(op[1:]); err != nil {
```
paul-nicolas, line 634: [Major] `txOpFinalized` folds as a protobuf overlay, not a replace — the comment above asserts an invariant nothing enforces

`state.UnmarshalVT(op[1:])` into a **non-empty** `state` is a proto merge, not an assignment: repeated fields append, maps merge, zero scalars don't clear. The comment at 609-611 claims the operand "is always the oldest operand for the key, so it lands on an empty accumulator and the overlay coincides with a replace" — that is an assumption about call-site ordering, not a property of the code, and `SeedTransaction` now makes `txOpFinalized` an operand any caller can inject at any sequence number.

The assumption is **already false in this PR**: `checker_archived_tx_test.go` seeds *after* the delta, so the fold order is `[setMeta(k2), finalized(k0)]` and the test only passes because of the overlay. Verified consequences when the invariant is violated with a postings-carrying operand: folding two `txOpFinalized` operands through `Finish(true)` yields **2 postings from a single-posting snapshot**, and `[txOpDeleteMeta(k0), txOpFinalized(base with k0)]` **resurrects the deleted key**. Blast radius is a false `CHECK_STORE_ERROR_TYPE_TRANSACTION_UPDATE_MISMATCH` — the integrity checker reporting tampering on correct data.

Not critical, because it is unreachable through today's single call site (`accumulatorBase` emits exactly one entry per canonical key, and every `Finish(true)` result is written as kind `SET`, so two finalized operands cannot stack). But invariant #7 explicitly rejects exactly this shape: a load-bearing "should not happen" branch documented instead of enforced.

Fix — one line, and it makes the comment true by construction:

```go
case txOpFinalized:
    // A resolved snapshot supersedes every older operand, so reset before
    // ingesting rather than relying on the accumulator being empty: proto
    // UnmarshalVT into a non-empty message is an overlay (repeated fields
    // append, maps merge), so an out-of-order operand would silently
    // duplicate postings.
    state.Reset()
    if err := state.UnmarshalVT(op[1:]); err != nil {
```

**Must land together with the test reordering flagged in `checker_archived_tx_test.go`** — applied alone, `TestCompareTransactions_ArchivedThenUpdated` fails.

internal/application/check/checker_archived_tx_test.go
```diff
@@ -0,0 +1,119 @@
...
+	})
+	require.NoError(t, err)
+	require.NoError(t, batch.Commit())
+
+	checker := NewChecker(store, attrs, "test-cluster", nil, nil, logger)
+
+	// A replay that saw ONLY the post-archive k2 set — the create log was purged.
+	buildReplay := func() *replayStore {
+		r := newTestReplayStore(t)
+		require.NoError(t, r.SaveTxMetadata(txKey.Bytes(), metaK2))
```
paul-nicolas: [Major] The test builds the replay in non-production operand order, and that is what makes the `txOpFinalized` overlay bug invisible

`buildReplay()` calls `SaveTxMetadata` here at line 87; `seedReplayTransactionsFromBaseline` only runs afterwards at line 116. That inverts both `SeedTransaction`'s own documented precondition (`replay_store.go:282-283`: "Must be called before any delta operand for the key so the merger sees the base first") and the order `Check()` uses (`checker.go:392`, before the replay pass opens at 408). The seed arrives as the **newest** operand.

Two compounding problems:

1. **It actively blocks the hardening.** Adding `state.Reset()` to the `txOpFinalized` branch — the fix that makes the merger's own comment true — makes this test fail (`expected` loses `k2`: `expected … metadata:{k0} postings:{…}, got … metadata:{k0} metadata:{k2} …`). Verified.
2. **The ordering production actually uses is left untested**, so a regression in the seed-before-replay sequencing at `checker.go:392` would not be caught.

The delete case makes the stakes concrete: with `DeleteTxMetadata` before the seed, the seed resurrects the deleted key and the checker false-positives.

Fix — seed inside `buildReplay()`, before the delta, so operand order matches production:

```go
buildReplay := func(seed bool) *replayStore {
    r := newTestReplayStore(t)
    if seed {
        require.NoError(t, checker.seedReplayTransactionsFromBaseline(baselineDB, r))
    }
    require.NoError(t, r.SaveTxMetadata(txKey.Bytes(), metaK2))

    return r
}
```

then `runCompare(buildReplay(false))` for the existing `NotEmpty` sanity assertion and `runCompare(buildReplay(true))` for the `Empty` fix assertion. Confirmed: the seed-first shape passes both with and without the `state.Reset()` hardening.

internal/application/check/checker.go
```diff
@@ -5290,6 +5298,38 @@ func (c *Checker) seedTxTrackingFromBaseline(baseline *pebble.DB, ledgerKnownTxI
...
+func (c *Checker) seedReplayTransactionsFromBaseline(baseline *pebble.DB, replay *replayStore) error {
+	txIter, err := c.attrs.Transaction.NewStreamingIter(baseline, nil)
+	if err != nil {
+		return fmt.Errorf("creating baseline tx streaming iter for replay seeding: %w", err)
+	}
+
+	for txIter.Next() {
+		entry := txIter.Entry()
+		if err := replay.SeedTransaction(entry.CanonicalKey, entry.Value); err != nil {
+			return fmt.Errorf("seeding replay transaction from baseline: %w", err)
```
paul-nicolas: [Medium] The baseline iterator leaks on this error path

This mid-loop `return` never closes `txIter`, and `StreamingIter.Close()` is the only thing that releases the underlying `*pebble.Iterator` (`internal/infra/attributes/streaming_iter.go:157`; its own doc comment at :20 prescribes `defer iter.Close()`).

Reproduced downstream: with one outstanding iterator, `pebble.DB.Close()` on the read-only baseline returns `leaked iterators: current` — and `Check()` swallows exactly that at `checker.go:206` (`defer func() { _ = baselineDB.Close() }()`). So on a seed failure the baseline is never really released and its table-cache refs / file handles stay pinned, once per failed `Check()` on a long-lived `Checker`. Silent.

The sibling `seedTxTrackingFromBaseline` is not a precedent for this shape — it `continue`s on an unparsable key and has no in-loop return, so its close-then-`Err()` tail covers every path. This is the first early return under an open iterator.

Fix — hoist the error and close on every path:

```go
var seedErr error
for txIter.Next() {
    entry := txIter.Entry()
    if seedErr = replay.SeedTransaction(entry.CanonicalKey, entry.Value); seedErr != nil {
        break
    }
}
if err := txIter.Close(); err != nil { ... }
if seedErr != nil {
    return fmt.Errorf("seeding replay transaction from baseline: %w", seedErr)
}
```

Folding this function into `seedTxTrackingFromBaseline` (see the separate comment on line 5310) removes the leak by construction.

*Raised independently by Codex and two Claude reviewers.*

internal/application/check/replay_store.go
```diff
@@ -477,10 +498,22 @@ func (m *txMerger) MergeOlder(value []byte) error {
...
+func (m *txMerger) Finish(includesBase bool) ([]byte, io.Closer, error) {
+	ops := make([][]byte, 0, len(m.ops))
+	if err := flattenTxOps(m.ops, &ops); err != nil {
+		return nil, nil, err
+	}
+
+	// A partial merge may not include the base value, and a delta op — a metadata
+	// delete has no snapshot representation — cannot be resolved without it. Defer by
+	// re-emitting the ordered ops; a base-inclusive fold resolves them.
+	if !includesBase {
```
paul-nicolas, line 528: [Medium] The deferral silently depends on `'T'`-prefix keys never being deleted

`Finish(false) → txOpBatch` assumes `includesBase=false` whenever the base is absent. Pebble breaks that in one shape: on `MERGE` + `DEL`/`SINGLEDEL`, `internal/compact/iterator.go:894-914` rewrites the key kind to `SETWITHDEL` and returns **without calling `MergeOlder` at all**, after which `includesBase` computes as `true` (`:640-648`). The merger would then fold post-tombstone deltas into a snapshot with no base — reinstating exactly the dropped-delete bug this PR fixes. A `DeleteRange` covering the base is the same class (`:886-892`).

Unreachable today — `db.Delete` is issued only on the volume (`replay_store.go:143`) and metadata (`:190`) prefixes, never on `replayPrefixTransaction`. But `DeleteVolume` in this same file invites a `DeleteTransaction` / ledger-cascade sibling, and the regression would be silent: a checker pass that wrongly **succeeds**.

Fix — state the dependency in the comment at 507-509 so it is not rediscovered the hard way:

> No `Delete`/`DeleteRange` may ever be issued on `replayPrefixTransaction`: Pebble reports `includesBase=true` for MERGE+DEL (kind → SETWITHDEL) without supplying a base, which would defeat this deferral.

Better still, make it enforceable — route transaction-key writes through a helper, or add a test asserting no tombstone is written under that prefix.

paul-nicolas (commented):

The core fix is correct and should land. I verified it by experiment rather than by reading: reverting only the `if !includesBase { return encodeTxBatch(ops) }` block makes both new merger tests fail, so they genuinely pin the bug. The `Finish` change also fixes a **latent pre-existing** defect independent of the seeding feature — the old collapse was accidentally correct for `txOpCreate`/`txOpSetMeta`/`txOpRevertedBy` (the proto overlay coincides with their semantics) but always wrong for `txOpDeleteMeta`, which has no snapshot representation, and that is reachable on the non-archived path too since a flush may report `includesBase=false`. Unconditional deferral is the right call; narrowing it to "only when a delete is present" would trade this bug for a subtler one.

**Findings: 2 Major, 7 Medium posted inline. 2 Minor/Nit listed below.** 13 candidates entered adversarial verification; 5 were rejected.

**Merge gate: the two Majors are one coupled change** — `state.Reset()` at `replay_store.go:612` plus reordering the test at `checker_archived_tx_test.go:87` to seed before the delta. Apply either alone and the suite breaks. Everything else, including the archive-boundary guard, is a follow-up.

Verified green at `1d215bf1`: `go test ./internal/application/check/...` → ok (4.3s); `golangci-lint run internal/application/check/...` → 0 issues.

- `internal/application/check/replay_store.go:672` — `decodeTxBatch` accepts `n == 0`, appending a zero-length op that `flattenTxOps:638` then silently skips; measured that `encodeTxBatch([][]byte{{}, deleteOp})` decodes to 2 ops and flattens to 1, so an op vanishes. Symmetrically `uint32(len(op))` at `:697` truncates silently on a >4 GiB op. Both unreachable today, but silent-drop is the class invariant #7 exists to forbid — error on `n == 0`, and bounds-check the encoder.
- `internal/application/check/replay_store.go:517` — `if len(op) == 0 { continue }` in the fold loop is dead; `flattenTxOps` already dropped empties. Related pre-existing: `checker.go:1872` compares against literal `0x00` in a comment that names `txOpFinalized` — use the constant.

### Rejected in verification (not posted)

- *Seeding creates new false positives for deleted / cleanup-pending ledgers* — `checker.go:1925-1956` already unioned `baselineTx` into `allKeys` and already used it as the `expected` fallback, so every key the seed adds was already present. Confirmed empirically: identical single event with and without the seed. (The absence of a `pendingCleanup` skip in `compareTransactions`, which `compareReferences` and `compareNumscripts` have, is a real **pre-existing** gap deserving its own ticket.)
- *`TestReplayStoreDeleteSurvivesCompaction` passes vacuously* — it does not; reverting only the deferral makes it fail.
- *`txOpBatch` grows unboundedly / nesting compounds framing overhead* — `encodeTxBatch` only ever receives flattened ops, so batches cannot nest; repeated partial merges measured size-idempotent over six `Finish(false)` cycles. Real cost is write amplification only.
- *Returning a re-emitted operand from `Finish(false)` violates the Pebble contract* — `internal/compact/iterator.go:629-654` keeps the kind as `MERGE` exactly when `includesBase` is false, so an operand-shaped value is what is required.
- *A raw `txOpBatch` could reach the `valBytes[0] != 0x00` tag check and abort the pass* — all three Pebble read paths hardcode `Finish(true)` (`iterator.go:721, 937, 1137`), and `base/merger.go:51-56` documents `includesBase` as "always true during normal iteration".

### Coverage caveats

The documentation-and-tests dimension had no full independent pass (one reviewer returned only a late addendum), so those two findings rest on my own verification plus Codex. And the out-of-order-archiving window behind the last Medium is established by code reading — nobody reproduced it against a running archived cluster.

internal/application/check/checker.go
```diff
@@ -5290,6 +5298,38 @@ func (c *Checker) seedTxTrackingFromBaseline(baseline *pebble.DB, ledgerKnownTxI
...
+// before the log-replay pass so the seed is the merger's base operand.
+func (c *Checker) seedReplayTransactionsFromBaseline(baseline *pebble.DB, replay *replayStore) error {
+	txIter, err := c.attrs.Transaction.NewStreamingIter(baseline, nil)
+	if err != nil {
+		return fmt.Errorf("creating baseline tx streaming iter for replay seeding: %w", err)
+	}
+
+	for txIter.Next() {
+		entry := txIter.Entry()
+		if err := replay.SeedTransaction(entry.CanonicalKey, entry.Value); err != nil {
```
paul-nicolas: [Medium] Seeding every baseline transaction defeats the replay store's stated OOM-avoidance purpose

*(The concrete cost lands at `checker.go:1851`, outside this diff — commenting here because this line is the cause.)*

This merges one operand per baseline transaction — the entire pre-archive transaction set. `compareTransactions` then materialises the whole replay transaction set into a Go map at `checker.go:1851`. That map was `O(post-archive txs)`; it is now `O(all txs ever)`, **on top of** the pre-existing `baselineTx` map built at `checker.go:1900-1910`. So the pass holds three full copies where two sufficed, roughly doubling peak memory on a large archived cluster — in a code path whose sibling `replayStore` exists specifically to keep replay state "on disk instead of in memory… prevents OOM with large datasets" (`replay_store.go:47-50`). The temp DB also grows to the full transaction set.

A consequence worth deciding explicitly: since `replayTx ⊇ baselineTx` by construction now, the `} else if bs, ok := baselineTx[key]; ok {` fallback at `checker.go:1954` is unreachable within `Check()` — but **not** on the test path, since this PR's own test calls `compareTransactions` with an unseeded replay, so deleting it naively breaks the suite.

Fix — either seed lazily (only for canonical keys the replay actually touches, e.g. on first delta op per key), or stream the replay/baseline/live three-way compare instead of building `replayTx` in full. If the fallback is kept as defence-in-depth against a partial seed, say so at 1949-1956 so the next reader doesn't read the key-set equality as accidental.

*Raised independently by two Claude reviewers.*

internal/application/check/checker.go
```diff
@@ -5290,6 +5298,38 @@ func (c *Checker) seedTxTrackingFromBaseline(baseline *pebble.DB, ledgerKnownTxI
...
+// seedReplayTransactionsFromBaseline pre-loads the replay with each baseline
+// transaction state, so a post-archive delta (metadata set/delete, revert
+// marker) merges on top of the full pre-archive state instead of replaying as a
+// standalone partial. Under archiving the create logs are purged, so without
+// this the replay holds only the delta and compareTransactions — which takes the
+// replay state as authoritative over the baseline — would flag the (correct)
+// live state as tampered. Reads the baseline, never the live store; must run
+// before the log-replay pass so the seed is the merger's base operand.
+func (c *Checker) seedReplayTransactionsFromBaseline(baseline *pebble.DB, replay *replayStore) error {
+	txIter, err := c.attrs.Transaction.NewStreamingIter(baseline, nil)
```
paul-nicolas: [Medium] Fifth full scan of the baseline `Transaction` attribute, eight lines from an identical fourth

Under archiving, `c.attrs.Transaction.NewStreamingIter(baseline, nil)` — full attribute space, no prefix — now runs **five** times per `Check()`:

| Line | Caller |
|---|---|
| 3723 | `foldBaselineTxMetadata` |
| 3881 | `foldBaselineReverted` |
| 5270 | `seedTxTrackingFromBaseline` |
| **5310** | **this function** |
| 1900 | `compareTransactions`' own baseline loop |

On an archived dataset the baseline transaction set is the entire pre-archive history, so each redundant pass fully decodes every transaction state.

5270 and 5310 are called eight lines apart (384 and 392), take the same `nil` prefix, and each read only `entry.CanonicalKey` / `entry.Value` — there is no ordering constraint between `trackTxID` and `replay.SeedTransaction`. CLAUDE.md calls out DRY explicitly.

Fix — fold this body into `seedTxTrackingFromBaseline` (one loop doing both, renamed e.g. `seedFromBaselineTransactions`, taking the replay as a third param) and drop the call site at 392. This also eliminates the iterator-leak finding on line 5318 by construction, since the merged loop has one early-return path to guard.

internal/application/check/checker.go
```diff
@@ -384,6 +384,14 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 		if err := c.seedTxTrackingFromBaseline(baselineDB, ledgerKnownTxIDs, ledgerRevertedTxIDs); err != nil {
 			return err
 		}
+
+		// Seed the replay with the baseline transaction states so a post-archive
+		// metadata/revert delta merges onto the full pre-archive state; otherwise
+		// compareTransactions (replay overrides baseline) would see only the delta
+		// and flag the correct live state as tampered. Runs before the replay pass.
+		if err := c.seedReplayTransactionsFromBaseline(baselineDB, replay); err != nil {
```
paul-nicolas: [Medium] `compareTransactions` is now baseline-seeded and no documentation says so

Every sibling baseline-seeded pass is explicitly annotated; this one is not:

- **`docs/technical/architecture/subsystems/checker/checker.md:37`** still gives row 4's re-derivation source as "Replay of `CreatedTransaction` / `RevertedTransaction` orders". Compare row 10 ("over a baseline floor (`foldBaselineBoundaries`)"), row 11 ("Baseline tx-row markers + replayed `RevertedTransaction` logs") and the note at line 50 naming `foldBaselineNumscripts`. Under archiving the real expectation is now baseline snapshot + replayed delta, so the row is materially incomplete.
- **`CLAUDE.md` invariant #8**'s `compareTransactions` clause omits "baseline-seeded under archiving", while `compareSchema`, `compareAccountTypes`, `compareReferences` and `compareNumscripts` all state it.

Separately: the new `txOpBatch` deferral is a new non-obvious invariant of the replay merge operator living only in code comments. `checker.md:65` describes the `replayStore` merge operators in a single sentence and never mentions that a partial merge defers rather than collapses — which is precisely the kind of mechanism the repo's Documentation Maintenance rule asks to be written down.

Fix — amend `checker.md:37` and the CLAUDE.md #8 clause to name `seedReplayTransactionsFromBaseline`, and extend `checker.md`'s "Replay machinery" section with a short paragraph on why `Finish(false)` re-emits ordered ops.

*Raised independently by Codex and two Claude reviewers.*

internal/application/check/checker_archived_tx_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/formancehq/ledger/v3/internal/storage/dal"
+)
+
+// A transaction created before the archive boundary (full state in the baseline
+// checkpoint, create log purged) and updated afterward (a metadata set the
+// replay sees as a partial delta) must not be flagged as tampered. The stored
+// state is the correct full state (attribute zone, not purged); the checker's
+// expected must be baseline + delta. Using the partial replay alone drops the
+// create data and false-positives — the bug this pins.
+func TestCompareTransactions_ArchivedThenUpdated(t *testing.T) {
```
paul-nicolas: [Medium] The two symptoms that motivated this PR — delete and revert — are never exercised through `compareTransactions`

The PR description calls the metadata **delete** the hard case ("a finalized-level merge inside `compareTransactions` can't be correct"), and the originating report says it "trips on any archived transaction later **reverted** or given metadata". Yet this — the only test that drives `compareTransactions` with a seeded baseline — covers a metadata **set** only. Delete and revert are covered solely at merger level in `replay_store_test.go`, one layer below the pass the bug was actually filed against. The PR also asserts "reversion and post-commit-volume checks are unaffected" with no test behind it.

I wrote both probes in production seed-first order and **both pass**, so this is a coverage gap rather than a defect — but the regression that would catch a future re-break of either path *through the pass itself* does not exist.

Fix — two subtests alongside this one, table-driven over `(live state, post-archive delta)`:

- pre-boundary tx + post-archive `DeleteTxMetadata` → `compareTransactions` emits nothing;
- pre-boundary tx + post-archive `SetRevertedBy` → `compareTransactions` emits nothing.

Both are ~10 lines given the fixture this test already builds.

internal/application/check/checker.go
```diff
@@ -384,6 +384,14 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
...
+			return err
+		}
 	}
 
 	// Pre-load baseline volumes so compareTransactionPostCommitVolumes can add
```
paul-nicolas, line 401: [Medium] The replay does not actually start at `archiveEndSeq+1`, and this PR's seeding turns that into an integrity gap

*(The defect is at `checker.go:408-419` and `:427`, just below this hunk — commenting here as the nearest diff line.)*

The replay iterator's bounds are `LowerBound: []byte{dal.ZoneCold, dal.SubColdLog}` (`:409`) — it scans **from sequence 0**, with no sequence guard anywhere in the loop body (`:421-472`). `expectedSeq := archiveEndSeq + 1` at `:419` is only the gap-detection counter; its comment ("Start after archived sequences") states an assumption rather than enforcing it, and the comment at `:864` goes further and asserts "the replay (which starts at archiveEndSeq+1)", which the code does not do. The asymmetry is the tell: the audit pass **does** enforce the boundary (`:5796`, `item.GetLogSequence() <= archiveEndSeq { continue }`).

**Why this PR makes it matter.** `archiveEndSeq` is `max(close_sequence)` over ARCHIVED chapters (`:152-155`), so an un-archived chapter *below* that max leaves retained logs under the boundary. Before this PR a replayed pre-boundary `CreateTransaction` merely rebuilt state the baseline also held. Now the baseline row is pre-loaded under the same key, and `txOpCreate` has **partial**-field semantics — `replay_store.go:543` replaces `state.Metadata` only `if metaLen > 0`, `:559` replaces `state.Postings` only `if postingsLen > 0`, `:533` sets `Timestamp` only when present — so a retained create silently inherits seeded values. Three distinct symptoms from the one missing guard:

1. **Silent tamper masking (invariant #8).** Baseline row tampered with an extra key `k9`; retained create carries `metaLen == 0` → seeded `k9` survives into `expected` and matches the tampered live row → **not reported**. Pre-PR this same tamper *was* caught. This is the only silent direction and the only invariant-#8 miss.
2. **False `TRANSACTION_UPDATE_MISMATCH`** in the inverse shape (key created then deleted pre-boundary, create log retained, delete log purged).
3. **`SEQUENCE_GAP` storm.** `:448` does `expectedSeq = seq + 1` unconditionally, so a sub-boundary log drops the counter below the archive floor and the loop at `:441-446` then fabricates one "log sequence N is missing" event for every sequence in the legitimately purged range — a healthy archived cluster looking corrupt.

**Reachability** (verified): the archive/purge is atomic — `executePurge` runs in the same Pebble write batch as the ARCHIVED flip (`internal/infra/state/write_set.go:783-786`, `:933-941`) — so there is no archived-but-unpurged race. The window opens via **out-of-order archiving**, which nothing prevents: `processArchiveChapter` gates only on `CHAPTER_CLOSED` (`internal/domain/processing/processor_chapter.go:107`) with no requirement that older chapters be archived; `ArchiveChapter` is operator-driven by explicit chapter ID (`cmd/ledgerctl/chapters/archive.go:43`) and admission adds no ordering validation (`admission.go:1955-1966`); and `DispatchArchiveRequests` (`internal/infra/state/machine.go:1838-1857`) iterates `allChapters map[uint64]*commonpb.Chapter` (`chapter_tracker.go:17`), so reconcile dispatch order is **random**. No monotonic archive cursor exists anywhere.

**Fix** — the guard the audit pass already has, immediately after the sequence extraction at `:427`:

```go
// Archived sequences are covered by the baseline checkpoint (and by the audit
// hash chain); the audit-order pass applies the same boundary. A retained
// pre-boundary log must not fold onto a seeded transaction row: txOpCreate
// replaces Metadata/Postings only when non-empty, so it would inherit the
// seeded values.
if seq <= archiveEndSeq {
    continue
}
```

or equivalently raise the iterator `LowerBound` to `archiveEndSeq+1`. Both are no-ops on the non-archived path (`archiveEndSeq == 0`, sequences start at 1). Placement at `:427` is safe despite sitting above the ephemeral-purge flush loop at `:429-438`: `NewProposalBoundaryTracker(archiveEndSeq)` (`:5104`) stores it as `lastLogSequence` and `Accept` rejects `maxLogSequence <= lastLogSequence` (`internal/domain/replay/replay.go:341`), so every yielded boundary is strictly `> archiveEndSeq` and the flush loop is already unreachable for sub-boundary sequences. Note `state.Reset()` does **not** fix this one — the fold order here is `[finalized, create]`, snapshot already first.

Land it with a pin for symptom 1 (the silent one). **This is pre-existing and should not gate the PR** — a separate commit is fine, ideally soon.

flemzord (changes requested):
Revalidated live head `1d215bf1` after the new human review. One correctness/runtime blocker is active in the existing `paul-nicolas` archive-boundary thread: the log replay still scans retained logs from sequence 0, and out-of-order chapter archiving can leave logs at or below `archiveEndSeq`. With this PR’s new baseline transaction seeding, replaying such a pre-boundary create over the seeded snapshot can preserve seeded fields that the create omits, allowing a tampered projection to match the derived expectation instead of being reported. Please skip `seq <= archiveEndSeq` (as the audit-order pass already does), or enforce the equivalent iterator lower bound, and add a regression for the masking case. I am not duplicating the inline evidence; Codecov, documentation/test quality, and the unreachable finalized-overlay shape are not the basis of this request.

flemzord (changes requested):
Revalidated live head `2cf5178f` after the force-push. The archive-boundary skip is fixed, but the previously reported DeleteLedger / same-name recreation blocker remains in the new lazy-seeding design. `lazyTxSeedWriter` does not override `CreateTransaction`, so a post-archive create does not mark its reused transaction key as seeded; a later metadata/revert delta looks up the old-generation baseline and appends `txOpFinalized` after the new `txOpCreate`, resetting the expected state back to the old transaction. In the inverse ordering (old transaction lazily seeded before delete/recreate), `txOpCreate` itself does not reset the merge accumulator, so absent metadata/reversion fields can survive into the new generation. Because replayed DeleteLedger does not clear replay transaction keys, either sequence can produce a false or masking `TRANSACTION_UPDATE_MISMATCH`. Please make a create replace/reset transaction state and prevent later baseline seeding for that key, with archive → delete → same-name recreate → reused txID regressions covering both orderings. CI is green; this request is based only on the checker correctness blocker and does not duplicate the superseded inline threads.

flemzord:
Correction on live head `2cf5178f`: I withdraw the blocker in my immediately preceding review after completing the reachability check. The problematic merge ordering requires a same-name ledger recreation and reused transaction key, but `DeleteLedger` deliberately retains the `LedgerInfo` tombstone, `CreateLedger` rejects that tombstoned name with `ErrLedgerDeleted`, cleanup preserves the tombstone, and transaction IDs otherwise advance monotonically. The synthetic operand sequence is therefore not reachable from an authentic current FSM history and is not a correctness/runtime blocker for this PR. The archive-boundary masking issue is fixed on this head, functional CI and review bots are green, and the remaining Codecov patch failure is CI/tooling only. This head appears acceptable under the cron criteria; GitHub approval is intentionally disabled by configuration.

flemzord (changes requested):
Revalidated live head `2cf5178f`. The nil-state lazy-seeding concern already mentioned in the NumaryBot summary is active on this head: baseline `Attribute.Get` returns `(nil, nil)` for an absent transaction, while `seedIfNeeded` only skips `pebble.ErrNotFound`. A post-archive transaction followed by a metadata or reversion delta therefore appends an empty `txOpFinalized` after its `txOpCreate`; the merger reset erases the create sequence, timestamp, postings, and metadata before applying the delta, producing a false `TRANSACTION_UPDATE_MISMATCH` against the correct live projection. Please treat a nil state as no baseline and prevent later seeding for post-archive creates, with a focused regression. The separate Codecov patch failure is not the basis of this review.

internal/application/check/replay_store.go
```diff
@@ -573,7 +625,12 @@ func (m *txMerger) Finish(_ bool) ([]byte, io.Closer, error) {
...
 
 		case txOpFinalized:
-			// Re-ingesting a previously finalized state (from a prior compaction).
+			// A resolved snapshot supersedes every older operand, so reset before
+			// ingesting rather than relying on the accumulator being empty:
+			// UnmarshalVT into a non-empty message overlays (repeated fields append,
+			// maps merge, zero scalars don't clear), so an out-of-order finalized
+			// operand would otherwise duplicate postings or resurrect deleted keys.
+			state.Reset()
 			if err := state.UnmarshalVT(op[1:]); err != nil {
```
NumaryBot, line 634: 🔴 **[blocker] `txOpFinalized` uses proto merge (overlay) instead of replace, violating the stated invariant**

`state.UnmarshalVT(op[1:])` into a non-empty `state` is a protobuf merge, not an assignment: repeated fields append, maps merge, and zero scalars do not clear. The comment at lines 609–611 claims the operand 'is always the oldest operand for the key, so it lands on an empty accumulator and the overlay coincides with a replace' — but `SeedTransaction` now allows `txOpFinalized` to be injected at any sequence position. Concrete consequences: folding two `txOpFinalized` operands through `Finish(true)` yields double postings from a single-posting snapshot; `[txOpDeleteMeta(k0), txOpFinalized(base with k0)]` resurrects the deleted key. This can produce a false `CHECK_STORE_ERROR_TYPE_TRANSACTION_UPDATE_MISMATCH`.

**Suggestion:** Call `state.Reset()` before `state.UnmarshalVT(op[1:])` in the `txOpFinalized` branch, or assert/enforce that the accumulator is always empty when this branch is reached.

internal/application/check/checker_archived_tx_test.go
```diff
@@ -0,0 +1,220 @@
...
+			t.Cleanup(func() { _ = store.Close() })
+
+			txKey := domain.TransactionKey{LedgerName: "L1", ID: 1}
+
+			// Pre-archive state: created (log 5) with metadata k0. Captured in the baseline.
+			batch := store.OpenWriteSession()
+			_, err = attrs.Transaction.Set(batch, txKey.Bytes(), &commonpb.TransactionState{
+				CreatedByLog: 5,
+				Postings:     postings,
+				Metadata:     metaK0,
```
NumaryBot, line 87: 🟠 **[major] Test seeds baseline *after* delta operands, inverting production order and hiding the overlay bug**

`buildReplay()` calls `SaveTxMetadata` at line 87 and only runs `seedReplayTransactionsFromBaseline` at line 116. This is the reverse of the order `Check()` uses (seed at line 392, replay pass at line 408) and contradicts the documented precondition in `replay_store.go:282–283` ('Must be called before any delta operand'). As a result: (1) adding `state.Reset()` to fix the `txOpFinalized` overlay bug causes this test to fail, actively blocking the hardening; (2) the production seed-before-replay sequencing is left untested, so a regression there would not be caught.

**Suggestion:** Restructure the test so `seedReplayTransactionsFromBaseline` (or equivalent seeding) is called before any delta operands are written, mirroring the order used by `Check()` in production.

internal/application/check/checker_archived_tx_test.go
```diff
@@ -0,0 +1,220 @@
...
+	"github.com/formancehq/ledger/v3/internal/proto/servicepb"
+	"github.com/formancehq/ledger/v3/internal/storage/dal"
+)
+
+// A transaction created before the archive boundary (full state in the baseline
+// checkpoint, create log purged) and updated afterward — via a metadata set,
+// delete, or revert that the replay sees as a partial delta — must not be
+// flagged as tampered. The stored state is the correct full state (attribute
+// zone, not purged); the checker's expected is baseline + delta, produced by the
+// lazy-seed writer seeding the base on the delta's first touch. An unseeded
```
NumaryBot, line 27: 🟠 **[major] Missing `compareTransactions` coverage for delete and revert through the full check pass**

The PR description identifies metadata delete and post-archive revert as the motivating cases, but the only test exercising `compareTransactions` with a seeded baseline covers a metadata *set* only. Delete and revert are covered solely at the merger unit-test level in `replay_store_test.go`, one layer below the pass where the original bug was reported. A future regression in either path through `compareTransactions` itself would not be caught.

**Suggestion:** Add two subtests alongside the existing one: (1) pre-boundary transaction + post-archive `DeleteTxMetadata` → `compareTransactions` emits no diff; (2) pre-boundary transaction + post-archive `SetRevertedBy` → `compareTransactions` emits no diff. Use seed-first operand order.

internal/application/check/checker.go
```diff
@@ -386,6 +386,18 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
...
+	// (keeping the replay store O(touched)); untouched ones fall back to the
+	// baseline in compareTransactions.
+	var replayWriter domainreplay.Writer = replay
+	if hasArchivedChapters {
+		replayWriter = newLazyTxSeedWriter(replay, func(canonicalKey []byte) (*commonpb.TransactionState, error) {
+			return c.attrs.Transaction.Get(baselineDB, canonicalKey)
+		})
+	}
+
 	// Pre-load baseline volumes so compareTransactionPostCommitVolumes can add
```
NumaryBot, line 401: 🟠 **[major] Replay iterator starts from sequence 0, not `archiveEndSeq+1`, creating an integrity gap with seeded baseline**

The replay iterator uses `LowerBound: []byte{dal.ZoneCold, dal.SubColdLog}` (line 409), scanning from sequence 0 with no sequence guard in the loop body. `expectedSeq := archiveEndSeq + 1` at line 419 is only a gap-detection counter; its comment states an assumption rather than enforcing it. Before this PR, replaying a pre-boundary `CreateTransaction` merely rebuilt state the baseline also held. Now the baseline row is pre-seeded, so a replayed pre-boundary log can double-apply a delta on top of the seeded state and produce a false mismatch. The audit pass correctly enforces the boundary (line 5796: `item.GetLogSequence() <= archiveEndSeq { continue }`); the replay pass does not.

**Suggestion:** Add a sequence guard at the top of the replay loop body: `if item.GetLogSequence() <= archiveEndSeq { continue }`, mirroring the audit pass at line 5796.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1636#issuecomment-5061017844

internal/application/check/replay_store.go
Azorlogh: Stale — already fixed on this head: https://github.com/formancehq/ledger/commit/0302eefb4584067d482be97deca4854a74f144c7 adds `state.Reset()` before ingesting a finalized operand (the quoted comment no longer exists).

internal/application/check/checker_archived_tx_test.go
Azorlogh: Stale — fixed in https://github.com/formancehq/ledger/commit/0302eefb4584067d482be97deca4854a74f144c7 (seed-before-delta) and superseded by https://github.com/formancehq/ledger/commit/c57d5d5ce7f8a744317a53bff05369b6f452279e: the bulk `seedReplayTransactionsFromBaseline` referenced here no longer exists; the test drives deltas through the lazy-seed writer in production order.

internal/application/check/checker_archived_tx_test.go
Azorlogh: Stale — added in https://github.com/formancehq/ledger/commit/dda844e8bc503490fd25996be94a25c2ee886c0e: the test is table-driven over set/delete/revert through compareTransactions.

internal/application/check/checker.go
Azorlogh: Stale — fixed in https://github.com/formancehq/ledger/commit/7957b1ab5d8686ee8b2cc91b9f69b2a806cee224: the replay loop skips `seq <= archiveEndSeq`, mirroring the audit-order pass.

internal/application/check/replay_store.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/0302eefb4584067d482be97deca4854a74f144c7 — `state.Reset()` before ingesting a finalized operand.

internal/application/check/checker_archived_tx_test.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/0302eefb4584067d482be97deca4854a74f144c7 (seed-before-delta), then superseded by https://github.com/formancehq/ledger/commit/c57d5d5ce7f8a744317a53bff05369b6f452279e — the test now drives deltas through the lazy-seed writer in production order.

internal/application/check/checker.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/c57d5d5ce7f8a744317a53bff05369b6f452279e — the bulk-seed function (and its leaky loop) is removed entirely in favor of lazy seeding.

internal/application/check/replay_store.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/b5710005a46e91a130e6166a6d30f43500567e7d — deletes route through a `deleteKey` guard that errors on the transaction prefix, and the deferral comment names the dependency.

internal/application/check/checker.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/c57d5d5ce7f8a744317a53bff05369b6f452279e — seeding is lazy (first delta touch), so the replay holds O(touched) transactions and the baselineTx fallback stays load-bearing for untouched keys. Follow-up https://github.com/formancehq/ledger/commit/b3ba551df06274859a273e26856771456e77cbef treats a nil lookup as no baseline and marks keys seeded on create.

internal/application/check/checker.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/c57d5d5ce7f8a744317a53bff05369b6f452279e — the duplicate scan is gone with the bulk-seed function.

internal/application/check/checker.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/5f090924cc1f61b517306eb78b5d8d3c07b8f445 — checker.md row 4 + a replay-machinery paragraph on the txOpBatch deferral, and the invariant #8 clause.

internal/application/check/checker_archived_tx_test.go
Azorlogh: Added in https://github.com/formancehq/ledger/commit/dda844e8bc503490fd25996be94a25c2ee886c0e — table-driven set/delete/revert through compareTransactions, each with the unseeded-reproduces sanity check.

internal/application/check/checker.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/7957b1ab5d8686ee8b2cc91b9f69b2a806cee224 — the replay loop skips `seq <= archiveEndSeq`, mirroring collectAuditOrderBoundaryEffects.

flemzord:
Revalidated live head `b3ba551d`. The previous nil-baseline correctness blocker is fixed: absent baseline states are no longer emitted as finalized operands, post-archive creates mark their key seeded, and the focused create-then-delta path is covered through `compareTransactions`. The current code also resets finalized overlays and skips replay entries at or below the archive boundary; the recent bot findings refer to pre-fix shapes. Functional CI is green, with only Codecov patch coverage failing. I found no remaining correctness/runtime blocker. GitHub approval is intentionally disabled by this cron configuration.

paul-nicolas (approved):

flemzord (approved):

flemzord (approved):

PR #1638, opened 2026-07-24
fix(EN-1522): gate DeleteLedger cascade and fail loudly on post-handler inconsistencies

paul-nicolas:
## Summary

Three FSM apply-path fixes against `release/v3.0`, per [EN-1522](https://formance-team.atlassian.net/browse/EN-1522). A focused re-analysis confirmed the gaps and found one ticket sub-item obsolete (see Scope note).

### A — Command-envelope key identity
Handlers derived write/log/delete keys from a **mutable** loaded `LedgerInfo.name` projection instead of the accepted order's envelope ledger identity; a divergent projection could redirect a write to another ledger's keys.
- `processDeleteLedger`: `DeletedLedgerLog.Name` = envelope `ledger` (also fixes the `deletedLedgers` cleanup-queue key that Absorb reads).
- `applyMirrorSyncUpdate`: `MirrorSyncWrite.LedgerName` = `update.GetLedgerName()`; the `ErrNotFound` stale-update skip is preserved.
- `processAddLedgerMetadata` / `processDeleteLedgerMetadata`: `LedgerMetadataKey.LedgerName` = envelope `ledger`.

### B — Gate the Boundary cascade (invariant #9)
`DeleteLedger` declared Boundary coverage in admission but consumed it via the **raw** `WriteSet.Absorb` → `Derived.Boundaries.Delete(...)`, bypassing the `gatedScope` coverage gate.
- `processDeleteLedger` now deletes the Boundary through `s.Boundaries().Delete(LedgerKey{Name: ledger})` (gated, envelope key).
- `WriteSet.Absorb` keeps only the `deletedLedgers` cleanup signal; the raw delete is removed. Admission comment updated.

### C — Classify expected absence vs real error (invariant #7)
Four lookups collapsed every error into "entity absent" and soft-skipped. Now only `domain.ErrNotFound` is soft; every other error propagates:
- `partitionVolumes` (now returns `error`, threaded through `Merge`), `ValidateTransientVolumes` transient base read, `GetSinkConfig`, `NumscriptVersionExists`.

## Scope note
The ticket's **boundary counter-aggregation** sub-item is **obsolete** on `release/v3.0`: per-ledger counters moved to the `usagestore` peer store (EN-1420 / EN-1422) and `Merge` no longer mutates the Boundary for counters, so there is no staged Boundary to consume. Numscript-deletion/tombstone work remains excluded (EN-1288).

## Tests
Focused unit tests: key-identity divergence, gated boundary delete + malformed coverage (`*ErrCoverageMiss`), same-proposal Apply→DeleteLedger validity, expected-absence soft cases, and non-`ErrNotFound` error propagation for each C site.

## Verification
- `just pre-commit` — 0 lint issues
- `just test` — full unit suite green
- `go build ./...` clean

🔗 https://formance-team.atlassian.net/browse/EN-1522

[EN-1522]: https://formance-team.atlassian.net/browse/EN-1522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


flemzord:
Revalidated live head `5759c3dc`. The DeleteLedger boundary cascade now consumes the admission-declared coverage through the gated Scope, affected writes use the immutable command-envelope ledger identity, and non-`ErrNotFound` cache/storage faults are propagated instead of being treated as absence. Functional CI, Codecov, and the available bot checks are green; CodeRabbit skipped substantive review because this is a draft. I found no correctness/runtime blocker on this head. GitHub approval is intentionally disabled by this cron configuration.

Azorlogh (commented):
Two review-level notes (details on inline comments for the rest):

**Gap A survives in three handlers outside this diff** — same loaded-projection `info.GetName()` key derivation:
- `processor_index.go:28,37,57` — index registry keys off the projection name; `processCreateIndex` even sets `Ledger: ledger` (envelope) inside a row keyed by the projection name, so a divergence produces an internally inconsistent registry entry.
- `processor_metadata_schema.go:70,79,130,136` — same for the schema-change index cascade.
- `processor_mirror.go:495` — `PromotedLedgerLog.Name` comes from the projection, and `backup/rebuild.go` replays promotion by that log name, so a divergent projection corrupts backup rebuild.

If these are intentionally out of scope, worth a follow-up ticket.

**Rebase**: branch is 9 commits behind `release/v3.0` (#1645 touches `partitionVolumes` too). Merge is clean — I verified the merged tree builds and both packages' tests pass — but rebase before merging.

internal/infra/state/write_set_ephemeral_purge.go
```diff
@@ -59,12 +62,24 @@ func (b *WriteSet) partitionVolumes(
 		compiled, ok := ledgerTypes[update.Key.LedgerName]
 		if !ok {
 			info, err := b.getLedgerData(update.Key.LedgerName)
-			if err != nil {
+			if errors.Is(err, domain.ErrNotFound) {
+				// Expected soft outcome: the ledger carries no
+				// account-type info (never created, or the volume belongs
+				// to a ledger deleted earlier in this batch). Default
```
Azorlogh: The "deleted earlier in this batch" clause can't produce `ErrNotFound`: DeleteLedger soft-deletes (the row is Put back with `DeletedAt` set), so `getLedgerData` still returns it.

internal/infra/state/en1522_test.go
```diff
@@ -0,0 +1,199 @@
...
+		},
+	}
+}
+
+// injectTagCollision forces the next KeyStore.Get for canonical to return a
+// non-ErrNotFound *ErrCollisionDetected by storing an entry under the key's
+// U128 id but with a deliberately mismatched tag. This is the state-package
+// fault-injection seam for the invariant-#7 error-classification tests
+// (EN-1522 gap C): a genuine storage/cache fault, distinct from an absence.
+func injectTagCollision[K attributes.Key, V any](t *testing.T, ks *attributes.KeyStore[K, V], canonical []byte, data V) {
```
Azorlogh: Only ticket-named test file in the tree, and this helper's consumers live in `write_set_error_classification_test.go` — consider folding it there and behavior-naming the rest.

internal/infra/state/write_set.go
```diff
@@ -1355,7 +1374,17 @@ func (b *WriteSet) SetMaintenanceMode(enabled bool) {
...
 func (b *WriteSet) GetSinkConfig(name string) (commonpb.SinkConfigReader, error) {
 	cfg, err := b.Derived.SinkConfigs.Get(domain.SinkConfigKey{Name: name})
-	if err != nil || cfg == nil {
+	// A cache miss is the documented "no sink config" outcome; any other
+	// error is a real storage/cache fault and must surface (invariant #7).
+	if errors.Is(err, domain.ErrNotFound) {
+		return nil, nil
+	}
+
+	if err != nil {
```
Azorlogh: Could match `ResolveNumscriptContent`'s shape: `if err != nil || cfg == nil { return nil, err }`.

internal/infra/state/write_set_ephemeral_purge.go
paul-nicolas: Correct, and fixed in 45d336698 — comment only, the branches themselves stay.

Verified your reasoning end to end: `processDeleteLedger` sets `DeletedAt` and then `s.Ledgers().Put(domain.LedgerKey{Name: ledger}, l)` (processor_ledger.go:119-121), and nothing under `internal/infra/state/` or `internal/infra/attributes/` filters on `DeletedAt` — so `getLedgerData` returns the soft-deleted row with its account types intact and `partitionVolumes` classifies its volumes normally. The clause described an unreachable state.

The comment now names the causes that can actually produce `ErrNotFound` here: a never-created ledger, or a key absent from the Derived cache — `partitionVolumes` runs at Merge time, outside the per-order coverage gate, and `getLedgerData` reads `Derived.Ledgers` directly. Worth getting right rather than deleting, since invariant #7 makes this comment the thing a reader uses to decide whether to harden the branch or relax the rule.

internal/infra/state/en1522_test.go
paul-nicolas: Both points taken — done in 5decefafb. `en1522_test.go` is gone, split by subject with every assertion byte-identical:

- `injectTagCollision` → `write_set_error_classification_test.go`, beside its four (and only) callers. It had zero consumers in the file that declared it.
- `delete_ledger_cascade_test.go` → the three DeleteLedger boundary-cascade tests plus the `deleteLedgerOrder` helper they alone use.
- `machine_technical_updates_test.go` → the mirror-sync envelope-keying test, colocated with the `machine_technical_updates.go` it covers, which had no test file until now.

On the naming: confirmed it was the only ticket-named test file in the repo. The test *names* were already behaviour-named, so only file names changed. I kept the ticket references inside doc comments — useful provenance there, unlike in a filename.

internal/infra/state/write_set.go
paul-nicolas: Agreed — done in 284a970c9. The `ErrNotFound` early return stays; the two trailing guards collapse into `if err != nil || cfg == nil { return nil, err }`.

Behaviourally identical, as you'd expect: past the `ErrNotFound` return, `err` is non-nil only for real faults, so the `err` arm propagates and the `cfg == nil` arm returns `(nil, nil)`. Six lines to three, and it stops the file carrying two different shapes for the same three-outcome contract — `ResolveNumscriptContent` is 270 lines up with exactly this form. `TestGetSinkConfig_AbsentIsSoftNil` and `TestGetSinkConfig_StorageFaultPropagates` still pass unchanged.

paul-nicolas:
@Azorlogh both review-level notes actioned. Branch force-pushed to `5decefafb` (rebase rewrote history).

**Gap A remainder — fixed here rather than deferred** (`f464bf38c`), all seven sites: `processor_index.go:28,37,57`, `processor_metadata_schema.go:70,79,130,136`, `processor_mirror.go:495`.

What decided it against a follow-up ticket was your `PromotedLedgerLog.Name` catch. This PR was already moving `DeletedLedgerLog.Name` to the envelope while leaving its sibling on the projection — so merging as-is would have shipped an asymmetry the PR introduced itself. Fixing one and ticketing the other was the worst of the three options.

Confirmed the downstreams you flagged:
- `backup/rebuild.go:251` replays promotion by that log name and resolves `LedgerInfo` from it, hitting the `invariant: PromoteLedger for ledger %q with no LedgerInfo…` guard at line 957 on a miss — so a divergent name either aborts a backup rebuild or promotes a different ledger.
- `processCreateIndex` was self-contradictory exactly as you described: the row was **keyed** by `info.GetName()` while its `Ledger` field carried the envelope.
- Both metadata-schema handlers already `Put` the `LedgerInfo` under `domain.LedgerKey{Name: ledger}` (lines 58 and 120) while keying the index cascade off the projection — the two projections could split apart within a single handler.

Regression tests added to `processor_ledger_key_identity_test.go` covering all five handlers: each seeds a `LedgerInfo` under the envelope key with a divergent `Name` and asserts the envelope wins. I verified all five **fail** against the pre-fix code, so they pin the behaviour rather than just passing. Worth noting `processSetMetadataFieldType`, `processRemoveMetadataFieldType` and `processPromoteLedger` had no unit coverage at all before this.

**Rebase — done.** It was 11 behind by the time I picked this up, not 9 (`c3ddbad66` and `6a696bb1f` landed after your review). No conflicts, matching your merge-tree check. Since `#1645` touches `partitionVolumes` — the same function the comment fix lands in — I ran the gates rather than relying on a clean textual merge: `just pre-commit` reports 0 lint issues in both modules with no codegen drift, and the full `just test` suite is 79 packages ok / 0 failures.

flemzord (changes requested):
Revalidated live head `5decefaf`. One correctness/runtime blocker remains in the modified `partitionVolumes` path. The new comment explicitly says it runs during `Merge` outside the per-order coverage gate, and it still calls `b.getLedgerData` → `b.Derived.Ledgers.Get` directly. That bypasses invariant #9: if a ledger declaration/preload is ever missing, one node may hit a ledger row left in its parent cache and classify the same volume as EPHEMERAL/TRANSIENT, while another misses it and the new `ErrNotFound` branch silently keeps the volume. The same Raft proposal can then produce different cache/Pebble projections depending on node-local cache history. Please derive this classification through proposal-declared, coverage-gated ledger data (or otherwise carry already-gated account types into Merge) and add a regression that cannot pass on a raw parent-cache hit. CI is fully green and is not the basis of this review; I found no other remaining blocker and am not repeating the resolved Azorlogh threads.

Azorlogh:
One residual gap-A site, one layer down: the Merge flush writes the `SubGlobLedgerInfo` row via `SaveLedger(batch, update.New)` (`write_set.go:645`), and `SaveLedger` keys the row off `info.GetName()` (`batch.go:36`) — the projection's mutable field — while the cache row for the same update is keyed by `update.Key`. A divergent `LedgerInfo.name` splits the two, and `query/ledger.go` reads the global row. Fix is the same shape as the rest of gap A, just across the helper boundary: give `SaveLedger` a name parameter and pass `update.Key.Name` (rebuild.go passes `info.GetName()`, log-derived there). Fine as a follow-up.

flemzord (dismissed):

paul-nicolas:
Fixed in `0c3565d27`, with a correction to the framing.

**What changed.** `partitionVolumes` no longer calls `getLedgerData`. `ValidateTransientVolumes` — which already resolves account types through the gated `scope.Ledgers().Get` — now publishes them on the `WriteSet`, and `partitionVolumes` consumes that map at Merge time. A ledger with a dirty volume but no gated entry is now an `invariant:` hard fail rather than a silent parent-cache read. The regression (`TestPartitionVolumes_UngatedLedgerFailsLoudly`) seeds the ledger directly in the parent KeyStore with an EPHEMERAL type and an empty gated map; I verified it fails against the pre-fix code with "An error is expected but got nil" — pre-fix it matched EPHEMERAL off the parent hit and purged the volume, so it cannot pass on a raw parent-cache hit, which is what you asked for. It also removes a duplicate ledger load over the same key set.

**On the blocker classification — it does not hold.** `ValidateTransientVolumes` runs at `machine.go:1539`, before `Merge` at `machine.go:1568`, under the proposal-wide gated scope, over the same `Derived.Volumes.DirtyValues()` map that `Merge` drains into `volumeUpdates` (`DirtyValues()` returns `s.values`; `Merge()` ranges `s.values`). For every distinct ledger it does a gated `scope.Ledgers().Get` (`write_set.go:1177-1180`); an undeclared key returns `*ErrCoverageMiss`, which is not `ErrNotFound`, so it bypasses the `continue` at `:1181` and lands in `storageFaults` — rejecting the proposal before `partitionVolumes` runs. `CheckCoverage` is a pure function of the replicated coverage bits (`plan/resolve.go:123-125`: `CacheHit → coverage-only, value nil`), so that rejection is bit-identical on every node. No node-local input feeds the decision, hence no divergent projection.

The technical-only `Merge(batch, nil)` path (`machine.go:1255`) is safe for a different reason: all four volume writers are order handlers (`processor_posting.go:167,199`, `processor_transaction_numscript.go:201,237`), so a proposal with no orders reaches Merge with no volume updates at all. That is also what makes the new hard fail safe rather than a new crash path.

The read is not new either: `partitionVolumes` called `getLedgerData` on `release/v3.0` too, where the code was `if err != nil { kept; continue }` — a catch-all that silently kept the volume even on genuine storage faults. This PR's `ErrNotFound` split was strictly tighter than what it replaced.

What your comment does correctly identify is structural, and it was worth fixing: the safety rested entirely on `ValidateTransientVolumes` having pre-gated the same key set — an undocumented temporal coupling nothing enforced. Your parenthetical, carrying the already-gated account types into Merge, is what I implemented; it makes the coupling structural and loud instead of implicit.

Note the ledger-read error classification moved with the read: it now lives in `ValidateTransientVolumes`, covered by `TestValidateTransientVolumes_LedgerStorageFaultPropagates`, plus `TestValidateTransientVolumes_PublishesGatedTypes` pinning the carry-forward contract.

`just pre-commit`: 0 issues both modules, no codegen drift. `just test`: 79 packages ok, 0 failures.

paul-nicolas:
@Azorlogh fixed here rather than deferred, in `9d04db8f8`.

`SaveLedger` now takes an explicit name; `write_set.go` passes `string(update.CanonicalKey)` and `rebuild.go` keeps the payload name. Small correction on the call shape: `update.Key.Name` doesn't exist — the merge entry carries `CanonicalKey []byte` (`attributes/accumulator.go:11`) — but since `LedgerKey.AppendBytes` is `append(dst, lk.Name...)` (`domain/keys.go:346-353`), the canonical key *is* the name bytes, so your fix shape was right.

What moved this off the follow-up pile was checking whether anything would catch it. Nothing would: `query.ReadLedgers` is a `ScanZone` that discards the Pebble key (`query/ledger.go:25`), and `compareLedgerPresence` (`checker.go:5374`) and `compareSchema` (`checker.go:5190`) both re-derive the name from the payload — so a row under key `a` carrying `Name: "b"` registers as ledger `b` and passes both loops. No invariant-#8 backstop at this site. It also amplifies: the Merge flush refiles the wrong-keyed row on every subsequent write to that ledger, and `deleteLedgerData` never reclaims the global row (`batch.go:520-521`), so one bad row is permanent and `GetLedgerByName` — a point lookup (`query/ledger.go:35-43`) — serves stale data indefinitely. Given the PR already fixes seven gap-A sites, shipping the eighth as a ticket was the same asymmetry you talked me out of last round.

One note on your rationale: `rebuild.go` is self-consistent, but not quite for the reason given — the checkpoint seed at `rebuild.go:584` is payload-derived, not log-derived. It writes both rows under the same key either way, so it isn't a gap-A split; I passed `info.GetName()` there explicitly and commented why.

Regression `TestMergeKeysGlobalLedgerRowOffCanonicalKey` seeds a `LedgerInfo` whose `Name` diverges from its canonical key, flushes through `Merge`, and asserts the global row is reachable under the canonical key and absent under the divergent name. I verified it fails pre-fix — the ledger is unreachable under its own identity.

Also in this push, `0c3565d27` addresses the `partitionVolumes` thread (details in the comment above).

Gates: `just pre-commit` 0 issues both modules, no codegen drift; `just test` 79 packages ok, 0 failures.

Azorlogh (approved):

PR #1639, opened 2026-07-24
fix(http): enforce exact-one typed log response in unitary handlers (EN-1537)

Azorlogh:
## Context

The four unitary HTTP handlers — create-ledger, promote-ledger, create-transaction, revert-transaction — rely on a strict backend contract: one request is admitted as one order, processing allocates one result slot, so a successful unitary request returns exactly one log whose payload type is fixed by the request type.

They enforced only part of that contract:

- zero logs was flagged via `unreachable`, but **more than one** log was silently accepted (only `logs[0]` used);
- a **nil** sole log was not guarded;
- a wrong/nil **payload type** was returned as an ordinary `writeInternalServerError` 500 instead of surfacing the impossible backend response.

A backend regression could therefore return extra logs or a mismatched payload without tripping the Antithesis invariant signal.

## Change

New shared helper `internal/adapter/http/unitary_log.go`:

- `exactlyOneLog` — rejects any cardinality other than one, and a nil sole log;
- `unexpectedLogPayload` — flags a wrong outer/inner log payload type, recording the actual types plus request context.

Both fail loudly through `unreachable`, so `jsonRecoverer` still returns a sanitized JSON 500 to the client while `assert.Unreachable` fails the Antithesis run (CLAUDE.md invariant #7). All four handlers now use the helper.

## Guarantees

- Successful response bodies and status codes are unchanged.
- Ordinary backend/business error mappings are unchanged.
- No public HTTP / OpenAPI contract change.

## Tests

- Table-driven handler regressions for each operation: zero logs, two logs, one nil log, wrong payload, valid expected payload.
- Retains the end-to-end recoverer assertion proving an invariant panic becomes a sanitized JSON 500.

Ticket: EN-1537


flemzord:
Revalidated live head `c5a345d9`. The shared exact-one helper now rejects zero/multiple/nil logs, and each unitary handler fails loudly on an unexpected outer or inner payload while preserving the normal success path; the focused regressions cover those contract violations. Functional CI, Codecov, NumaryBot, and the available CodeRabbit status are green. I found no correctness/runtime blocker on this head. GitHub approval is intentionally disabled by this cron configuration.

paul-nicolas (commented):
Multi-model review (Claude + Codex). **6 findings posted inline: 2 Major, 4 Medium.** 11 Minor/Nit listed below rather than inline.

The production change is sound. Client-visible behaviour really is unchanged — `jsonRecoverer` and `writeInternalServerError` emit an identical `500 {"errorCode":"INTERNAL_ERROR",...}` with the same correlation-ID mechanism, and `exactlyOneLog` runs before any write, so there is no partial-response hazard. The findings are concentrated in the **tests**, which mutation testing shows are largely not load-bearing.

**Coverage arithmetic across the 18 new subtests:** 4 (`zero logs` x4) would pass on the old code too — carry-over from the deleted `_NoLogReturned` tests. 6 (`two logs` x4, `wrong payload type` x2) are genuine mutation-proven regressions. The remaining 8 fail against the true pre-change code but survive removal of the specific branch they target. One mechanical change — `require.Panics` to `require.PanicsWithValue` — fixes 8 of the 9 test findings.

**Two claims were investigated and rejected, noted so they are not re-raised:**
- A reviewer flagged `internal/application/admission/admission.go` returning a nil log via idempotency replay against a chapter-purged log, as merge-blocking. The chain is sound link-by-link but was **already fixed on the merge target**: `ee6f2fb07` (#1633) switched admission to `ReadLogBySequenceWithCold`. That commit is on `release/v3.0` but is not an ancestor of this head (merge base `b160cb476`). Stale-base artifact, not a defect — though a rebase is worth doing so CI validates against the cold-storage-aware path.
- A reviewer flagged the wrong-payload path as losing the old 500 observability. It does not: the client response is identical and the server side is *enriched* (`panic.value`/`panic.stack` span attributes, `span.RecordError`, a stack trace). The only genuine residue is the dropped details map, posted inline. One consequence worth knowing rather than fixing: panic-rate alerting will now fire on payload mismatches that were previously quiet 500s — which is the intent.

**Verified on this head:** `go build ./...` clean, `go vet ./internal/adapter/http/...` clean, `golangci-lint run ./internal/adapter/http/...` 0 issues, `go test ./internal/adapter/http/... -count=1` ok. `jsonRecoverer` is registered globally at `handler.go:65` before route registration, no alternate mux bypasses it, and `internal/adapter/v2` never wraps these handlers — the panic path is covered on every route.

---

- `internal/application/admission/admission.go` -- rebase onto `release/v3.0`; base predates `ee6f2fb07`
- `internal/adapter/http/handlers_bulk.go:179` -- `runBulkAtomic` has no `len(logs) == len(requests)` check (unreachable today, out of diff)
- `internal/adapter/http/handlers_bulk.go:207` -- `runBulkSequential` skips the len/nil/payload-type checks (unreachable today, out of diff)
- `internal/adapter/grpc/server_cluster.go:465` -- unguarded `logs[0]`, untyped error on the impossible fallthrough (behind `recoveryInterceptor`, out of diff). These three are best handled as one follow-up ticket; note the helpers are package-private to `internal/adapter/http`, so it is not a straight reuse
- `docs/technical/architecture/subsystems/api/http-api.md:56-61` -- enumerates two 500 paths; this PR moves "unexpected log payload" from path 2 to path 1, leaving the section correct but incomplete
- `internal/adapter/http/handlers_create_ledger_test.go:170` -- the description claims a "valid expected payload" row, but every table row is a violation asserting `require.Panics`. Happy paths exist as separate `_Success` tests per handler, so no real gap -- just an inaccurate description
- `internal/adapter/http/unitary_log.go:52` -- `logDetails` is a misleading name for a generic map merge (`log` means `*commonpb.Log` everywhere else in this package); prefer `mergeDetails`
- `internal/adapter/http/unitary_log.go:26` -- `log_count` is always `1` in the nil-log branch; pass `details` directly
- `internal/adapter/http/unitary_log.go:37` -- parameter named `log`; call sites use `logEntry`
- `internal/adapter/http/test_helpers_test.go:58` -- `.AnyTimes()` is looser than needed; `.Times(1)` passes and makes "Apply was reached" a checked property
- The four handlers now build the `details` map unconditionally on the success path (one extra allocation per 2xx on the hottest write endpoint); build it inside the failure branches instead

internal/adapter/http/handlers_create_transaction_test.go
```diff
@@ -56,27 +56,60 @@ func TestHandleCreateTransaction_Success(t *testing.T) {
...
+
+			srv := newTestServer(t, backendReturningLogs(t, tc.logs))
+
+			w := httptest.NewRecorder()
+			body := strings.NewReader(`{"script":{"plain":"send [USD 100] (\n  source = @world\n  destination = @users:001\n)"}}`)
+			r := newRequest(t, http.MethodPost, "/ledger1/transactions", body, map[string]string{
+				"ledgerName": "ledger1",
+			})
+
+			require.Panics(t, func() {
```
paul-nicolas: [Major] `require.Panics` passes on an accidental nil deref, so the payload-mismatch cases do not test the new guard

Mutation-proven. Replacing the `unreachable(...)` call at `unitary_log.go:47` with a no-op and re-running `go test ./internal/adapter/http/ -run LogContractViolations -v`:

```
--- FAIL: TestHandleCreateLedger_LogContractViolations/wrong_payload_type
--- FAIL: TestHandlePromoteLedger_LogContractViolations/wrong_payload_type
--- PASS: TestHandleCreateTransaction_LogContractViolations/wrong_outer_payload
--- PASS: TestHandleCreateTransaction_LogContractViolations/wrong_inner_payload
--- PASS: TestHandleRevertTransaction_LogContractViolations/wrong_outer_payload
--- PASS: TestHandleRevertTransaction_LogContractViolations/wrong_inner_payload
```

Root cause at `handlers_create_transaction.go:60-65`: when the type assertion fails, `created` is a nil `*commonpb.LedgerLogPayload_CreatedTransaction`. With the guard neutered, control reaches `writeCreated(w, created.CreatedTransaction)` -> runtime nil deref -> still a panic -> the test stays green for the wrong reason. It would pass straight through the very regression it guards.

Same applies to `handlers_revert_transaction_test.go:109` (fall-through at `handlers_revert_transaction.go:99-104`).

**Fix:** add a `wantPanic string` column to the table and use `require.PanicsWithValue(t, tc.wantPanic, ...)`. The panic value is the plain message string (`unreachable.go:15` does `panic(message)`) — verified with a probe test. This one change also resolves the nil-sole-log coverage gap flagged on `unitary_log.go:24`.

internal/adapter/http/unitary_log.go
```diff
@@ -0,0 +1,58 @@
...
+// invariant #7). operation names the endpoint and details carries the stable
+// request context (ledger, transaction id) threaded into the signal.
+func exactlyOneLog(operation string, logs []*commonpb.Log, details map[string]any) *commonpb.Log {
+	if len(logs) != 1 {
+		unreachable(operation+" apply did not return exactly one log", logDetails(details, map[string]any{
+			"log_count": len(logs),
+		}))
+	}
+
+	if logs[0] == nil {
```
paul-nicolas, line 32: [Major] This nil-sole-log guard is entirely uncovered — deleting it leaves the whole suite green

Mutation-proven. Deleting only the `if logs[0] == nil { unreachable(...) }` block (lines 24-28) and re-running:

```
--- PASS: TestHandleCreateLedger_LogContractViolations/nil_sole_log
--- PASS: TestHandleCreateTransaction_LogContractViolations/nil_sole_log
--- PASS: TestHandlePromoteLedger_LogContractViolations/nil_sole_log
--- PASS: TestHandleRevertTransaction_LogContractViolations/nil_sole_log
```

All generated getters are nil-receiver-safe (`internal/proto/commonpb/common.pb.go:3091` `GetPayload`, `:3202` `GetApply`), so a nil sole log simply falls through to the payload-type check and panics from `unexpectedLogPayload` instead of from this branch. The four `nil sole log` rows therefore never exercise the guard they were added for.

**Fix:** `require.PanicsWithValue(t, "<operation> apply returned a nil log", ...)` for the `nil sole log` row in all four tables, which pins each case to the intended branch.

internal/adapter/http/unitary_log.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// exactlyOneLog enforces the unitary-handler backend contract: one admitted
+// request yields exactly one non-nil log, whose payload type is fixed by the
+// request type. Zero logs, several logs, or a nil sole log are impossible
+// backend responses; each fails loudly through unreachable, so jsonRecoverer
+// surfaces a sanitized JSON 500 while Antithesis flags the run (CLAUDE.md
+// invariant #7). operation names the endpoint and details carries the stable
+// request context (ledger, transaction id) threaded into the signal.
+func exactlyOneLog(operation string, logs []*commonpb.Log, details map[string]any) *commonpb.Log {
```
paul-nicolas, line 25: [Medium] A fifth handler with the same contract was not migrated, and its fallthrough is a silent 204

Re: `internal/adapter/http/handlers_save_numscript.go:53-63` — flagged independently by both reviewers, and cross-referenced here since that file is not in the diff.

It keeps the old shape: `if len(logs) == 0 { unreachable(...) }` — no exactly-one check, no nil-log check — and when `logs[0].GetPayload().GetSavedNumscript()` is nil it falls through to `w.WriteHeader(http.StatusNoContent)` at `:63`. That is a silent 2xx false success, strictly worse than the 500 this PR just replaced in the other four handlers, and a direct violation of CLAUDE.md invariant #7.

The "maybe the 204 is a legitimate no-op" defence was checked and fails: `processor_numscript_library.go:78-84` returns a non-nil `SavedNumscript` on every non-error path, and an idempotent re-save of an existing version returns `ErrNumscriptVersionAlreadyExists` (409), never a 204. The handler's own test says as much at `handlers_save_numscript_test.go:72-74`.

**Caveat to carry with the fix:** `openapi.yml:1645` does declare a `'204'` for this operation, so removing the branch means removing that response from the spec too, or code and spec diverge.

Also worth correcting: the PR description says "All four handlers now use the helper" when there are five consumers of this contract.

internal/adapter/http/unitary_log.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	return logs[0]
+}
+
+// unexpectedLogPayload flags a sole log whose payload type is not the one the
+// request type implies. It records the actual outer payload type — and, for an
+// Apply log, the inner ledger-log payload type — alongside the request context
+// so the invariant signal pinpoints the mismatch.
+func unexpectedLogPayload(operation string, log *commonpb.Log, details map[string]any) {
```
paul-nicolas: [Medium] `unexpectedLogPayload` returns normally, and nothing in the toolchain enforces termination

Call sites do `if !ok { unexpectedLogPayload(...) }` and then use the value. That is correct today only because `unreachable` panics — the compiler cannot see it, and neither can the linters: `go build`, `staticcheck`, `nilnesserr` and `golangci-lint run ./internal/adapter/http/...` are all clean. staticcheck's SA5011 does not fire because the nil check is on `ok`, not on the pointer.

If `unreachable` is ever softened, the degradation is silent and inconsistent across the four call sites:
- `handlers_create_ledger.go:163-168` -> `createLedgerLog` nil -> `ToLedgerInfo()` is nil-safe (`internal/proto/commonpb/ledger_log.go:7-9`) -> **201 Created with `{"data":null}`**
- `handlers_promote_ledger.go:31-36` -> same shape -> **201 Created** with an empty name
- `handlers_create_transaction.go:60-65` and `handlers_revert_transaction.go:99-104` -> nil deref

Two handlers silently succeed, two crash. **Fix:** make terminality compiler-visible — either return the typed payload from the helper so no caller holds a value that only a panic keeps valid, or have `unreachable` return a value so call sites read `panic(unreachable(...))`.

internal/adapter/http/unitary_log.go
```diff
@@ -0,0 +1,58 @@
...
+	extra := map[string]any{
+		"sequence":           log.GetSequence(),
+		"outer_payload_type": fmt.Sprintf("%T", log.GetPayload().GetType()),
+	}
+
+	if apply := log.GetPayload().GetApply(); apply != nil {
+		extra["inner_payload_type"] = fmt.Sprintf("%T", apply.GetLog().GetData().GetPayload())
+	}
+
+	unreachable(operation+" apply returned an unexpected log payload type", logDetails(details, extra))
```
paul-nicolas: [Medium] The details map is discarded in production

Re: `internal/adapter/http/unreachable.go:13-16` (`assert.Unreachable(message, details); panic(message)`), cross-referenced here because this is where the details are built.

The Antithesis SDK is explicit that these calls are "no-ops with minimal performance overhead when called outside of the Antithesis environment" unless `ANTITHESIS_SDK_LOCAL_OUTPUT` is set (`antithesis-sdk-go@v0.7.0/assert/assert.go` package doc), and the `no_antithesis_sdk` build tag makes it a literal no-op. `panic(message)` then drops the map, and `jsonRecoverer` logs only the panic value — the bare message.

So `ledger`, `transaction_id`, `log_count`, `outer_payload_type` and `inner_payload_type` never reach an operator's logs. The doc comment on `unexpectedLogPayload` ("so the invariant signal pinpoints the mismatch") holds only under Antithesis, which is precisely the environment where you least need it — in production this is strictly less diagnosable than the `writeInternalServerError` path it replaced.

**Fix:** fold the details into the panic value, e.g. `panic(fmt.Sprintf("%s: %v", message, details))`. Safe, because `jsonRecoverer` sanitizes: the panic value goes to the server log and the OTel span, never to the client body.

internal/adapter/http/unitary_log.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// logDetails merges extra into a copy of base, leaving the caller's map
+// untouched so the same request context can seed several invariant checks.
+func logDetails(base, extra map[string]any) map[string]any {
```
paul-nicolas: [Medium] No `unitary_log_test.go` — the diagnostic details map is never asserted

The richer diagnostics are the stated point of this change, yet nothing tests them. `require.Panics` observes neither the message nor the details, and the map is handed only to `assert.Unreachable`, a no-op sink outside Antithesis — so `recover()` cannot see it either. A wrong `operation` string, a swapped key name, a `%T` applied to the wrong expression, or a silently missing `inner_payload_type` would all be invisible to the suite. The non-mutation contract of `logDetails` is likewise unasserted.

**Fix:** extract the map builder as a pure function and table-test it:

```go
func payloadMismatchDetails(log *commonpb.Log, details map[string]any) map[string]any {
	extra := map[string]any{
		"sequence":           log.GetSequence(),
		"outer_payload_type": fmt.Sprintf("%T", log.GetPayload().GetType()),
	}
	if apply := log.GetPayload().GetApply(); apply != nil {
		extra["inner_payload_type"] = fmt.Sprintf("%T", apply.GetLog().GetData().GetPayload())
	}
	return logDetails(details, extra)
}
```

Then assert the exact map per case (non-Apply outer, Apply outer, nil log) and that `base` is not mutated. Add a `TestExactlyOneLog` alongside, asserting `PanicsWithValue` for `nil` / `[]` / two-element input and a clean return for the single-log case.

flemzord (changes requested):
Revalidated current head `c5a345d9` after the newer human review. One correctness/runtime blocker remains in the existing `paul-nicolas` thread: `handlers_save_numscript.go` is a fifth consumer of the same exact-one Apply-log contract, but it still accepts multiple/nil/unexpected logs and can fall through to a false `204 No Content` success. Please migrate that handler to the shared exact-one/payload guard and align its OpenAPI response. I am not duplicating the inline evidence; the other new findings are test/diagnostic quality and are not the basis of this review. Functional CI remains green.

flemzord:
Correction on current head `c5a345d9`: after re-reading the live PR scope, I withdraw the blocker from my immediately preceding review. The PR explicitly targets four named handlers; `save-numscript` is a pre-existing fifth path with a `204` response declared in OpenAPI, so folding it into this change is follow-up scope rather than a regression of this head. The remaining newer findings concern test/diagnostic quality, not a demonstrated production correctness/runtime defect. This head therefore appears acceptable under the cron criteria; GitHub approval is intentionally disabled by configuration.

flemzord:
Revalidated live head `475a0e82`. The prior exact-one Apply-log concern is addressed on this head: `save-numscript` now uses the shared cardinality/type guard, impossible branches terminate explicitly with preserved diagnostics, and the focused tests distinguish the intended panic paths. Functional CI and the available review checks are green. I found no remaining correctness/runtime blocker. GitHub approval is intentionally disabled by this cron configuration.

flemzord (changes requested):
Revalidated live head `475a0e82`. One correctness/runtime false-success remains in the newly migrated save-numscript path: `handlers_save_numscript.go` checks that `SavedNumscript` is non-nil, but then passes `saved.GetInfo()` to `writeCreated` without requiring the nested `Info`. A typed `SavedNumscriptLog{Info:nil}` therefore returns `201 {"data":null}` instead of surfacing the impossible FSM/backend result. This nested-nil case is not covered by the existing cardinality, nil-log, or wrong-payload threads/tests. Please fail loudly on a nil `Info` and add the focused regression. CI is green and is not the basis of this review.

flemzord:
Revalidated live head `ea5fe354`. The new commit fixes the remaining nil-body false-success: save-numscript, create-transaction, and revert-transaction now fail loudly when a correctly typed log has no meaningful payload body, with focused regressions for each path. Functional CI, Codecov, and the available review checks are green. I found no remaining correctness/runtime blocker on this head; GitHub approval is intentionally disabled by this cron configuration.

internal/adapter/http/handlers_create_transaction_test.go
Azorlogh: Fixed in 475a0e82 — each violation row now asserts its branch-specific panic message via `requirePanicsContaining`, so an accidental nil-deref panics with a different value and fails the test instead of passing for the wrong reason.

internal/adapter/http/unitary_log.go
Azorlogh: Fixed in 475a0e82 — the `nil sole log` row now pins on `apply returned a nil log`. Deleting the guard makes a nil log fall through to the payload-type branch (a different message), so the row fails as intended.

internal/adapter/http/unitary_log.go
Azorlogh: Fixed in 475a0e82 — save-numscript migrated onto the shared exact-one/typed guard and the `204` removed from `openapi.yml`. Follow-up ea5fe3546 additionally fails loudly on a nil `Info` body (the nested-nil case).

internal/adapter/http/unitary_log.go
Azorlogh: Fixed in 475a0e82 — `unreachable` now returns the panic value and call sites read `panic(unreachable(...))` / `panic(unexpectedLogPayload(...))`, so termination is visible to the compiler at the call site.

internal/adapter/http/unitary_log.go
Azorlogh: Fixed in 475a0e82 — `unreachable` folds the details into a deterministic panic value, so they reach the `jsonRecoverer` server log and OTel span, not only Antithesis.

internal/adapter/http/unitary_log.go
Azorlogh: Fixed in 475a0e82 — added `unitary_log_test.go` covering `observedPayloadDetails` (exact map + base non-mutation), `exactlyOneLog`, and that the panic value carries the diagnostics.

Azorlogh:
Nil payload-body false-success (flemzord, head `475a0e82`) fixed in ea5fe3546: a correctly-typed sole log with a nil body — save-numscript (`Info`), and the same latent nested-nil in create-transaction (`CreatedTransaction`) and revert-transaction (`RevertedTransaction`) — now fails loudly via a new `emptyLogPayload` invariant signal instead of returning `201 {"data":null}`. A focused `empty payload body` regression was added to each handler's contract table.

The earlier save-numscript `204` blocker was resolved in 475a0e82 (migrated onto the shared guard + `204` dropped from `openapi.yml`). All of @paul-nicolas's inline findings are addressed in 475a0e82 (replies on each thread). Package `go build`, tests, and `golangci-lint` are green.

flemzord (dismissed):

paul-nicolas (approved):

PR #1642, opened 2026-07-27
fix(EN-1506): classify every Apply request variant explicitly

paul-nicolas:
## Summary

`RequiredScopeForRequest` left 13 protobuf oneof variants falling through to a catch-all `default: ScopeOpsWrite` — 10 in `Request.type` and 3 in `LedgerAction.data`. That default served two opposite roles at once, "reject the unknown" and "nobody decided", so neither runtime behavior nor any test could tell them apart.

Both classifiers now return `(Scope, bool)`. The exported wrapper collapses an undecided variant to `ScopeOpsWrite`, preserving the fail-closed contract for genuinely unknown input, while the flag lets a test prove every declared variant was actually classified.

### Scope corrections

Each anchored to the scope the operation's dedicated route or RPC already required:

| Variants | Was | Now | Anchor |
|---|---|---|---|
| `save_numscript` | OpsWrite | `LedgerWrite` | `handler.go:176` |
| `save_ledger_metadata`, `delete_ledger_metadata` | OpsWrite | `MetadataWrite` | `handler.go:193-194` |
| account types + enforcement mode (top-level **and** `LedgerAction`) | OpsWrite | `MetadataWrite` | `handler.go:197-199` |
| 4 × query checkpoint | OpsWrite | `ClusterWrite` | `server_cluster.go:442,481` |

The 23 pre-existing decisions are unchanged in value (verified by diffing decisions against the base commit).

### :warning: Breaking change

Query-checkpoint create / delete / schedule over gRPC `Apply` now require `ledger:ClusterWrite`.

Previously `ledger:OpsWrite` sufficed. Since `DefaultMapping` grants `OpsWrite` to any `ledger:write` token but `ClusterWrite` only to `ledger:admin`, a plain `ledger:write` caller could perform through `Apply` an operation the dedicated `ClusterService` RPCs restrict to admins — a privilege escalation, not merely a scope mismatch.

Callers driving query checkpoints through `Apply` must move to `ledger:admin` or add the granular `ledger:ClusterWrite`. Callers using the dedicated RPCs are unaffected.

### Regression guard

`request_scope_exhaustiveness_test.go` walks both oneof descriptors and fails when a variant has no explicit arm. This is the assertion a "covered field names" list cannot make: such a list cannot distinguish a deliberate `OpsWrite` from an accidental one.

Verified non-vacuous by falsification — removing the `save_numscript` arm makes the guard fail by name; restoring it makes it pass.

### Scope notes

- **HTTP `/bulk` is unaffected.** It shares the classifier but cannot reach any affected variant: `convertBulkElementToRequest` always wraps in `Request_Apply`, and `BulkElement.UnmarshalJSON` accepts only `CREATE_TRANSACTION`, `ADD_METADATA`, `REVERT_TRANSACTION`, `DELETE_METADATA`, rejecting anything else before the scope loop runs. No HTTP behavior changes, so no HTTP tests were added.
- No `DeleteNumscript` mapping — EN-1288 removed that operation.
- Consolidating route-group scopes with the classifier is left to EN-1417; this PR supplies the reviewed policy that refactor should consume rather than reinvent.

## Testing

Transport-level tests use **granular** scopes only — the aggregate `ledger:write` expands to both `OpsWrite` and `MetadataWrite` and would pass regardless of which the classifier picks, masking exactly this bug.

- `go test -race -count=1 ./internal/adapter/auth/... ./internal/adapter/grpc/...` — pass
- `go test -race ./... -timeout 20m` — 114 packages, 0 failures, 0 data races
- `just pre-commit` — 0 lint issues (both modules)

[EN-1506](https://formance-team.atlassian.net/browse/EN-1506)

[EN-1506]: https://formance-team.atlassian.net/browse/EN-1506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/technical/contributing/api-comparison.md
+++ b/docs/technical/contributing/api-comparison.md
@@ -709,12 +709,12 @@ Read endpoints comparison with the original ledger:
 | `GET /v3/{ledgerName}/numscripts/{name}?version=` | ✅ | ❌ | Get numscript (version selector, empty/latest = greatest semver) |
 | `GET /v3/{ledgerName}/numscripts/{name}/usage` | ✅ | ❌ | Get invocation count + last-used timestamp |
 | `GET /v3/{ledgerName}/numscripts/{name}/versions` | ✅ | ❌ | List version history |
-| `PUT /v3/{ledgerName}/numscripts/{name}` | ✅ | ❌ | Save an immutable version (explicit full semver) |
+| `PUT /v3/{ledgerName}/numscripts/{name}` | ✅ | ❌ | Save an immutable version (explicit full semver). Requires `ledger:LedgerWrite` on both the dedicated route and gRPC `Apply(SaveNumscript)` |
 | `GET /v3/{ledgerName}/account-types` | ✅ | ❌ | List account types |
 | `GET /v3/{ledgerName}/account-types/{typeName}` | ✅ | ❌ | Get account type |
-| `POST /v3/{ledgerName}/account-types` | ✅ | ❌ | Add account type |
-| `DELETE /v3/{ledgerName}/account-types/{typeName}` | ✅ | ❌ | Remove account type |
-| `PUT /v3/{ledgerName}/account-types/default-enforcement-mode` | ✅ | ❌ | Set default enforcement mode (STRICT/AUDIT) |
+| `POST /v3/{ledgerName}/account-types` | ✅ | ❌ | Add account type. Requires `ledger:MetadataWrite` on both the dedicated route and gRPC `Apply` |
+| `DELETE /v3/{ledgerName}/account-types/{typeName}` | ✅ | ❌ | Remove account type. Requires `ledger:MetadataWrite` on both the dedicated route and gRPC `Apply` |
+| `PUT /v3/{ledgerName}/account-types/default-enforcement-mode` | ✅ | ❌ | Set default enforcement mode (STRICT/AUDIT). Requires `ledger:MetadataWrite` on both the dedicated route and gRPC `Apply` |
 | `GET /v3/{ledgerName}/transactions` | ✅ | ❌ | List transactions: cursor pagination, `startDate`/`endDate` range, and the generic `filter` (reference selection via `filter={"$match":{"reference":"..."}}`) |
 | `GET /v3/_/logs/{sequence}` | ✅ | ❌ | Fetch a single system log by bucket-wide sequence. No ledger identity → requires `ledger` ops-read (granular `ledger:OpsRead`) |
 | `GET /v3/_/chapters` | ✅ | ❌ | Stream chapters (audit-chain segments) |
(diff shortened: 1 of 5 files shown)


flemzord:
Revalidated live head `bb1d9fbe`. The descriptor-driven classifiers explicitly cover both protobuf oneofs, preserve fail-closed handling for malformed or unknown input, and align the affected Apply variants with their dedicated route/RPC scopes; in particular, query-checkpoint writes now require `ClusterWrite` instead of the broader `OpsWrite` path available to `ledger:write`. Functional CI and Codecov are green; CodeRabbit skipped substantive review because this is a draft. I found no correctness/runtime blocker on this head. GitHub approval is intentionally disabled by this cron configuration.

Azorlogh:
`request_scope.go:10` — the fail-closed default returns `ScopeOpsWrite`, described as "the most restrictive write scope," but `DefaultMapping` puts `OpsWrite` in the `ledger:write` bundle. So an unrecognized or malformed variant is admitted to any `ledger:write` caller — a low bar, not a restrictive one. That's the same shape as the checkpoint escalation this PR fixes: an operation nobody classified ends up behind a scope nearly everyone with write access already holds.

The exhaustiveness test makes this nearly unreachable for declared variants, so it's not urgent — but worth making deliberate rather than inherited. A true fail-closed default would require an admin-only scope (e.g. `ClusterWrite`) or a never-granted sentinel; at minimum the "most restrictive" wording should be corrected.

Azorlogh (dismissed):

NumaryBot (commented):
## 💬 Comments — automated review

The classifier changes explicitly cover all known protobuf request and LedgerAction oneof variants, preserve fail-closed handling for unknown input, and correctly align Apply request scopes with their dedicated route/RPC contexts. Tests adequately exercise the intended authorization behavior and coverage remains stable. One noteworthy design concern was raised: the fail-closed default scope (`ScopeOpsWrite`) is described as "the most restrictive write scope," but it is actually part of the broadly-held `ledger:write` bundle, meaning an unrecognized variant would be silently admitted to any write-capable caller. The exhaustiveness test makes this nearly unreachable for currently-declared variants, so it is not a blocker, but the misleading comment and the low-bar default are worth addressing deliberately to avoid a future recurrence of the same class of escalation this PR was written to fix.

### Findings outside the diff

🟡 **[minor] Fail-closed default returns OpsWrite, not a truly restrictive scope** — `request_scope.go:10`

The fail-closed default for unrecognized or malformed variants returns `ScopeOpsWrite`, which is described in comments as "the most restrictive write scope," but `DefaultMapping` places `OpsWrite` in the `ledger:write` bundle. This means any unrecognized variant is admitted to any `ledger:write` caller — a relatively low bar. This is structurally the same pattern as the checkpoint-escalation bug this PR fixes: an unclassified operation ends up behind a scope nearly everyone with write access already holds. The exhaustiveness test makes this nearly unreachable for declared variants, but the wording and intent are misleading and could become a real issue if new variants are added before the test catches them.

**Suggestion:** Change the fail-closed default to require an admin-only scope (e.g. `ClusterWrite`) or a never-granted sentinel scope, so that truly unknown input is denied rather than silently admitted. At minimum, correct the "most restrictive" comment to accurately reflect that `OpsWrite` is a broadly-held write scope.



Summary: https://github.com/formancehq/ledger/pull/1642#issuecomment-5091958984

flemzord:
Revalidated live head `c08d123f`. The new commit is documentation-only relative to the previously reviewed classifier, and all functional/coverage checks are green. The current `ScopeOpsWrite` default is not a demonstrated runtime authorization bypass on this head: malformed or unknown request variants are rejected by admission as unsupported before any Raft proposal, while any newly declared variant known to this binary is caught by the descriptor-driven exhaustiveness test until explicitly classified. The wording that calls `OpsWrite` the most restrictive scope could be tightened, but that is hygiene rather than a correctness blocker. I found no remaining correctness/runtime blocker; GitHub approval is intentionally disabled by this cron configuration.

Azorlogh (approved):

PR #1643, opened 2026-07-27
refactor(EN-1513)!: remove persisted MirrorCursor, resume from the audited applied boundary

paul-nicolas:
## Summary

`LedgerBoundaries.last_mirror_v2_log_id` becomes the only durable authority for mirror ingestion position. The duplicated, checker-unverified `SubPLMirrorCursor` projection is removed.

The mirror kept two durable records of one fact. The verified one (`last_mirror_v2_log_id`, covered by `compareMirrorV2LogID`) governed nothing on the read side; the unverified one drove both the worker's fetch position and the public progress API — including a failure mode where a cursor advanced past the source head makes the mirror report `FOLLOWING` while silently skipping source logs.

That failure mode was not newly discovered: `audit-vs-technical-state.md`, `checker/README.md`, and `checker/audit-chain.md` each already described it as a **current open correctness gap** awaiting "worker/startup reconciliation". This is that reconciliation, delivered by removing the duplicate rather than by adding a protocol to reconcile it — the bad state becomes unrepresentable instead of merely detectable. All of those declarations are updated to closed.

## Changes

- Worker reads `LedgerBoundaries` **once, before its first fetch**, serving both the source position and `nextTxID`. Previously two separate reads, the second of which loaded the authoritative field and discarded it.
- The in-memory position is now explicitly a cache over the durable authority: advanced only after confirmed FSM application, and dropped on batch error so the next tick re-reads. Previously the load latched forever, so a wedged worker never reconsulted durable state until leadership changed.
- Removed `MirrorSyncUpdate.cursor` (tags realigned), `MirrorSyncWrite.Cursor`, the `WriteSet.Merge` cursor drain, `SetMirrorCursor`, `ReadMirrorCursor`, and the CLI dump case.
- `dal.SubPLMirrorCursor` deleted; sub-prefix `0x05` left reserved and documented so it is never reassigned.
- `ReadMirrorSyncProgress` derives `cursor`, `FOLLOWING`, and `remaining_logs` from the boundary. The public `common.MirrorSyncProgress.cursor` field is unchanged, with its meaning restated in the proto.
- Storage schema version bumped 2 → 3.

## On the schema bump

It is load-bearing, not cosmetic. `MirrorSyncUpdate` is not a persisted Pebble projection (`AppliedProposal` does not embed the proposal; the audit chain serializes orders only), so its only stale-bytes exposure is un-applied Raft WAL entries. The gate fails `*dal.Store` construction, so a rejected store means the node never boots and never replays such an entry — which is what makes realigning the tags safe.

It also retroactively gates the EN-1550 `last_mirror_v2_log_id` field itself: v2 predates that field, yet `compareMirrorV2LogID` requires it with no backfill leniency, so the branch was previously accepting stores the checker considers unsupported.

No migration or backfill code, per the alpha no-compatibility policy.

## Testing

New worker tests cover fresh-ledger start, resume from an applied boundary, restart-after-commit, and snapshot invalidation. A query test writes a row under the retired `0x05` prefix and asserts it cannot influence progress — it would fail against the old cursor-based code.

`just pre-commit` (0 lint issues), `GOROOT= go build ./...`, and `just test` (80 packages, 0 failures) all pass.

## Coordination

Open PR #1590 (`refactor(mirror): rename source types`) touches `misc/proto/common.proto`, `internal/application/mirror/manager.go`, and `docs/.../events-mirror/mirror.md` — three files this branch also edits. The conflicts are textual, not semantic; whichever lands second rebases.

## Ticket

https://formance-team.atlassian.net/browse/EN-1513


flemzord:
Revalidated live head `4854b738`. `LedgerBoundaries.last_mirror_v2_log_id` now consistently drives worker resume and public progress, the worker advances its in-memory position only after successful FSM application and invalidates it after errors, and the retired Pebble prefix remains reserved. The storage-schema gate covers the intentionally realigned internal Raft fields. Functional CI is green; the remaining `codecov/patch` failure is CI/tooling rather than a demonstrated runtime blocker. I found no correctness/runtime blocker on this head. GitHub approval is intentionally disabled by this cron configuration.

flemzord (changes requested):
Revalidated live head `4854b738`. One correctness/runtime blocker remains in the `MirrorSyncUpdate` wire realignment: removing `cursor = 2` also moves `error` 3→2, `clear_error` 4→3, and `source_log_count` 5→4. In a live mixed-version Raft cluster, a technical update emitted by either version is decoded with incompatible field meanings/wire types by the other (for example, new length-delimited `error` at tag 2 versus old fixed64 `cursor`), so the same applied entry can fail or produce different state across nodes. The storage-schema v3 check gates local store opening/replay, but it does not prevent an already-running old follower or a freshly joined different-version node from applying the incompatible entry. This violates the deterministic-FSM invariant and is a runtime issue, not merely pre-stable API compatibility. Please either preserve/decode the old wire layout while retiring the cursor, or enforce a cluster-wide protocol/version activation barrier that makes mixed layouts impossible.

paul-nicolas:
Fixed in 01f7d6b — you were right, and the fix is in rather than an argument against it.

`MirrorSyncUpdate` now reserves the retired tag instead of realigning:

```proto
reserved 2;
reserved "cursor";
common.MirrorSyncError error = 3;
bool clear_error = 4;
fixed64 source_log_count = 5;
```

The regenerated wire tags byte-match `release/v3.0` for every surviving field (`0x1a`, `0x20`, `0x29`, `0xa`); the only wire-level delta is the disappearance of the cursor's `0x11`. No Go call site changed — the accessors were already tag-independent.

Two things your review surfaced that I want to record, because they are the actual root cause:

**1. The schema-gate defence was scoped to the wrong exposure.** The EN-1513 design rationale justified realigning on the grounds that `MirrorSyncUpdate` "is not a persisted Pebble projection … so its bytes persist only as un-applied Raft WAL/spool entries", which the v3 bump makes unreachable. That reasoning is sound for *persisted* stale bytes and silent about *live* ones — exactly the gap you identified. I have corrected the comment on `CurrentStorageSchemaVersion` to say what the gate does and does not cover. The v3 bump itself is kept: it still gates the removed Pebble cursor row and the EN-1550 `last_mirror_v2_log_id` field, which v2 predates.

**2. `raft_cmd.proto` had already settled this, and the docs contradicted it.** The file applies `reserved` to every removed field on a Raft-carried message — `Order` (1, 4, 6), `Proposal` (5, 6, 8–13, with *"Keep 13 reserved forever"* on `caller`), `TechnicalUpdate` (8), `BackupOrder` (2) — and renumbers in exactly one place, `BackupState` field 8, explicitly justified as safe *"because nothing on the wire carried it during the unmerged PR."* So the project's real rule is **renumber only if nothing ever shipped; reserve otherwise**, and this change sat on the wrong side of it.

`AGENTS.md` and `docs/technical/contributing/protobuf.md` both stated *"realign field numbers sequentially"* with no caveat, which is what led here. e887c60 adds a "Removing a field" section with the actual decision rule and cites both branches of the existing precedent, so the next removal does not repeat this.

Also added 3975270: table-driven coverage for `prefetchResult.afterID`, the guard that replaced the old cursor field in the prefetch-validity check. Mutation-verified — dropping the `afterID` comparison fails the stale-position case.

Gates: `just pre-commit` clean (0 lint issues, no codegen drift), unit suite green, and the five packages this PR touches pass explicitly.

flemzord:
Revalidated live head `3975270b`. The prior mixed-version Raft blocker is fixed: `MirrorSyncUpdate.cursor` now reserves tag 2 and its name, while `error`, `clear_error`, and `source_log_count` retain their shipped tags 3/4/5. Old and new nodes therefore decode the surviving fields identically, with the retired cursor handled as an unknown/absent field rather than a conflicting wire type. The added prefetch-position regression also covers the replacement resume guard. All functional checks, Coverage, and NumaryBot are green. I found no remaining correctness/runtime blocker; GitHub approval is intentionally disabled by this cron configuration.

Azorlogh:
Non-blocking, noting it for the pre-release cleanup pass: v3 alpha clusters are wiped and recreated on every `release/v3.0` push rather than rolling-upgraded, so the mixed-version decode scenario that motivated `reserved 2` can't occur — and reserving alone wouldn't cover it anyway, since an old binary would still *apply* the cursor write a new binary ignores (the divergence just moves from decode to apply). Fine as-is for now given the thread with @flemzord; flagging so the cleanup pass revisits both the reservation and the new reserve-vs-renumber sections in AGENTS.md / protobuf.md before release.

Separately, the PR description predates 01f7d6b: it still says the tags were realigned and justifies the schema bump as "what makes realigning the tags safe" — worth updating to describe the reservation.

Azorlogh:
Non-blocking: the `store.go` 0x05 comment says existing stores may still carry orphan rows under this sub-prefix, but no such store can reach this binary — anything carrying 0x05 rows is schema v2 and the new gate refuses to open it, and nothing on v3 writes the sub-prefix. Same premise in the `TestReadMirrorSyncProgress_IgnoresOrphanCursorRow` comment (the test itself remains a fine pin that nothing reads 0x05).

Azorlogh (dismissed):
Reviewed the resume mechanism end-to-end: position advance gated on confirmed FSM application, snapshot dropped on every error exit, prefetch invalidation keyed on the resume position (correctly discards prefetches made stale by a late-applying proposal), boundary lifecycle at create/delete matching the doc claims, and key encodings consistent across writer and readers. Sound. Remaining notes are non-blocking — see comments.

flemzord (dismissed):

Azorlogh (dismissed):

Azorlogh (dismissed):

Azorlogh (approved):

PR #1644, opened 2026-07-28
feat(release): build ledgerctl for Windows

flemzord:
## What changed

- build `ledgerctl` for Windows on `amd64` and `arm64`
- package Windows artifacts as ZIP archives
- align the GoReleaser project and artifact names with `main` (`ledger_<os>-<arch>`)
- teach `ledgerctl upgrade` to select and extract Windows ZIP artifacts and `ledgerctl.exe`
- select stable releases by the running binary's semver major, excluding GitHub drafts and prereleases across paginated results
- preserve executable permissions when staging an upgrade
- replace the running Windows executable through a rollback-safe `.old` backup and document manual recovery after an interrupted swap
- keep separate Unix and Windows GoReleaser builds/archives so platform consistency checks remain enabled
- cross-compile both Windows architectures in the regular Build CI job
- document prebuilt installation and platform-specific self-upgrade behavior

## Why

Windows users currently have no prebuilt `ledgerctl` artifact in GitHub releases. Adding the binaries also requires the release and self-upgrade paths to handle the Windows archive format, executable name, replacement constraints, and v2/v3 release coexistence deliberately.

## Impact

Future v3 releases will include `ledger_windows-amd64.zip` and `ledger_windows-arm64.zip`. Linux and macOS keep their existing `tar.gz` archives. `ledger-server` remains unchanged. On Windows, the previous executable remains available as `ledgerctl.exe.old`; if a swap is interrupted, it can be renamed back to `ledgerctl.exe`.

The artifact rename from `ledger-v3_*` to `ledger_*` is intentional to align v3 with `main`. Existing v3 alpha asset consumers must update their filenames. `ledgerctl upgrade --channel stable` selects only final releases matching the running binary's major, preventing selection of the active v2 line. Until the first final v3 release exists, the command reports that `nightly` must be used.

## Validation

- `nix develop --command go test ./cmd/ledgerctl/upgrade -count=1`
- `GOLANGCI_LINT_CACHE=<isolated-cache> nix develop --command bash -c "just pre-commit"`
- cross-compiled `ledgerctl` for Windows `amd64` and `arm64`
- tested draft/prerelease rejection, semver-major matching, and release pagination
- verified the live GitHub release list currently has no final v3 release
- `nix develop --command goreleaser check`
- full `nix develop --command goreleaser release --snapshot --clean --skip=docker`
- verified all six archives and `ledger_windows-amd64/ledgerctl.exe`

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,7 +1,7 @@
 # yaml-language-server: $schema=https://goreleaser.com/static/schema-pro.json
 ---
 version: 2
-project_name: ledger-v3
+project_name: ledger
 
 # IMAGE_REPOSITORY controls where Docker images are pushed. Override the env
 # var to publish to a different GHCR location. Default targets the canonical
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -66,6 +66,7 @@ archives:
   ids:
   - ledger-server
   - ledgerctl
+  allow_different_binary_count: true
   formats:
   - tar.gz
   format_overrides:
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -46,6 +46,7 @@ builds:
   goos:
   - linux
   - darwin
+  - windows
   goarch:
   - amd64
   - arm64
@@ -67,6 +68,10 @@ archives:
   - ledgerctl
   formats:
   - tar.gz
+  format_overrides:
+  - goos: windows
+    formats:
+    - zip
   name_template: "{{ .ProjectName }}_{{ .Os }}-{{ .Arch }}"
   wrap_in_directory: true
 
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -46,6 +46,21 @@ builds:
   goos:
   - linux
   - darwin
+  goarch:
+  - amd64
+  - arm64
+
+- binary: ledgerctl
+  id: ledgerctl-windows
+  main: ./cmd/ledgerctl
+  ldflags:
+  - -X github.com/formancehq/ledger/v3/internal/pkg/version.Version={{ .Version }}
+  - -X github.com/formancehq/ledger/v3/internal/pkg/version.Commit={{ .ShortCommit }}
+  - -X github.com/formancehq/ledger/v3/internal/pkg/version.BuildDate={{ .Date }}
+  - -extldflags=-static
+  env:
+  - CGO_ENABLED=0
+  goos:
   - windows
   goarch:
   - amd64
@@ -62,17 +77,19 @@ release:
   - glob: openapi.yml
 
 archives:
-- id: "{{ .ProjectName }}"
+- id: ledger-unix
   ids:
   - ledger-server
   - ledgerctl
-  allow_different_binary_count: true
   formats:
   - tar.gz
-  format_overrides:
-  - goos: windows
-    formats:
-    - zip
+  name_template: "{{ .ProjectName }}_{{ .Os }}-{{ .Arch }}"
+  wrap_in_directory: true
+- id: ledger-windows
+  ids:
+  - ledgerctl-windows
+  formats:
+  - zip
   name_template: "{{ .ProjectName }}_{{ .Os }}-{{ .Arch }}"
   wrap_in_directory: true
 
--- a/cmd/ledgerctl/upgrade/download.go
+++ b/cmd/ledgerctl/upgrade/download.go
@@ -2,6 +2,7 @@ package upgrade
 
 import (
 	"archive/tar"
+	"archive/zip"
 	"bufio"
 	"compress/gzip"
 	"crypto/sha256"
@@ -9,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime"
 	"strings"
 
 	"github.com/pterm/pterm"
@@ -74,7 +76,7 @@ func downloadAndVerify(archiveAsset, checksumsAsset *assetInfo, spinner *pterm.S
 
 	defer func() { _ = archiveResp.Body.Close() }()
 
-	tmpArchive, err := os.CreateTemp("", "ledgerctl-upgrade-*.tar.gz")
+	tmpArchive, err := os.CreateTemp("", "ledgerctl-upgrade-*")
 	if err != nil {
 		return "", fmt.Errorf("creating temp file: %w", err)
 	}
@@ -119,19 +121,39 @@ func downloadAndVerify(archiveAsset, checksumsAsset *assetInfo, spinner *pterm.S
 		return "", fmt.Errorf("checksum verification failed: expected %s, got %s", expectedHash, actualHash)
 	}
 
-	// 4. Extract ledgerctl binary from tar.gz.
+	// 4. Extract the ledgerctl binary from the platform archive.
 	spinner.UpdateText("Extracting ledgerctl...")
 
 	if _, err := tmpArchive.Seek(0, io.SeekStart); err != nil {
 		return "", fmt.Errorf("seeking archive: %w", err)
 	}
 
-	return extractBinary(tmpArchive, "ledgerctl")
+	archiveInfo, err := tmpArchive.Stat()
+	if err != nil {
+		return "", fmt.Errorf("reading archive metadata: %w", err)
+	}
+
+	return extractBinary(
+		tmpArchive,
+		archiveInfo.Size(),
+		archiveAsset.Name,
+		executableName(runtime.GOOS),
+	)
 }
 
-// extractBinary extracts a named file from a tar.gz archive and writes it to a temp file.
-// Returns the path to the temp file.
-func extractBinary(archive io.Reader, name string) (string, error) {
+// extractBinary extracts a named file from a supported release archive and writes it to a temp file.
+func extractBinary(archive io.ReaderAt, archiveSize int64, archiveName, binaryName string) (string, error) {
+	switch {
+	case strings.HasSuffix(archiveName, ".tar.gz"):
+		return extractTarGzBinary(io.NewSectionReader(archive, 0, archiveSize), binaryName)
+	case strings.HasSuffix(archiveName, ".zip"):
+		return extractZipBinary(archive, archiveSize, binaryName)
+	default:
+		return "", fmt.Errorf("unsupported archive format %q", archiveName)
+	}
+}
+
+func extractTarGzBinary(archive io.Reader, binaryName string) (string, error) {
 	gz, err := gzip.NewReader(archive)
 	if err != nil {
 		return "", fmt.Errorf("opening gzip reader: %w", err)
@@ -150,28 +172,71 @@ func extractBinary(archive io.Reader, name string) (string, error) {
 			return "", fmt.Errorf("reading tar: %w", err)
 		}
 
-		if hdr.Name == name || strings.HasSuffix(hdr.Name, "/"+name) {
-			tmpBinary, err := os.CreateTemp("", "ledgerctl-new-*")
-			if err != nil {
-				return "", fmt.Errorf("creating temp binary: %w", err)
-			}
-
-			if _, err := io.Copy(tmpBinary, tr); err != nil {
-				_ = tmpBinary.Close()
-				_ = os.Remove(tmpBinary.Name())
-
-				return "", fmt.Errorf("extracting binary: %w", err)
-			}
-
-			if err := tmpBinary.Close(); err != nil {
-				_ = os.Remove(tmpBinary.Name())
-
-				return "", fmt.Errorf("closing temp binary: %w", err)
-			}
-
-			return tmpBinary.Name(), nil
+		if isBinaryEntry(hdr.Name, binaryName) {
+			return writeExtractedBinary(tr)
 		}
 	}
 
-	return "", fmt.Errorf("binary %q not found in archive", name)
+	return "", fmt.Errorf("binary %q not found in archive", binaryName)
+}
+
+func extractZipBinary(archive io.ReaderAt, archiveSize int64, binaryName string) (string, error) {
+	zr, err := zip.NewReader(archive, archiveSize)
+	if err != nil {
+		return "", fmt.Errorf("opening zip reader: %w", err)
+	}
+
+	for _, file := range zr.File {
+		if !isBinaryEntry(file.Name, binaryName) {
+			continue
+		}
+
+		binary, err := file.Open()
+		if err != nil {
+			return "", fmt.Errorf("opening binary in zip: %w", err)
+		}
+
+		path, extractErr := writeExtractedBinary(binary)
+		closeErr := binary.Close()
+
+		if extractErr != nil {
+			return "", extractErr
+		}
+
+		if closeErr != nil {
+			_ = os.Remove(path)
+
+			return "", fmt.Errorf("closing binary in zip: %w", closeErr)
+		}
+
+		return path, nil
+	}
+
+	return "", fmt.Errorf("binary %q not found in archive", binaryName)
+}
+
+func isBinaryEntry(entryName, binaryName string) bool {
+	return entryName == binaryName || strings.HasSuffix(entryName, "/"+binaryName)
+}
+
+func writeExtractedBinary(binary io.Reader) (string, error) {
+	tmpBinary, err := os.CreateTemp("", "ledgerctl-new-*")
+	if err != nil {
+		return "", fmt.Errorf("creating temp binary: %w", err)
+	}
+
+	if _, err := io.Copy(tmpBinary, binary); err != nil {
+		_ = tmpBinary.Close()
+		_ = os.Remove(tmpBinary.Name())
+
+		return "", fmt.Errorf("extracting binary: %w", err)
+	}
+
+	if err := tmpBinary.Close(); err != nil {
+		_ = os.Remove(tmpBinary.Name())
+
+		return "", fmt.Errorf("closing temp binary: %w", err)
+	}
+
+	return tmpBinary.Name(), nil
 }
--- a/cmd/ledgerctl/upgrade/github.go
+++ b/cmd/ledgerctl/upgrade/github.go
@@ -15,7 +15,7 @@ import (
 
 const (
 	githubRepo  = "formancehq/ledger"
-	projectName = "ledger"
+	projectName = "ledger-v3"
 )
 
 type releaseInfo struct {
@@ -173,21 +173,38 @@ func githubDownload(url string) (*http.Response, error) {
 	return resp, nil
 }
 
-// archiveAssetName returns the expected archive filename for the current OS/arch.
-func archiveAssetName() string {
-	return fmt.Sprintf("%s_%s-%s.tar.gz", projectName, runtime.GOOS, runtime.GOARCH)
+// archiveAssetName returns the expected archive filename for an OS/arch pair.
+func archiveAssetName(goos, goarch string) string {
+	extension := ".tar.gz"
+	if goos == "windows" {
+		extension = ".zip"
+	}
+
+	return fmt.Sprintf("%s_%s-%s%s", projectName, goos, goarch, extension)
+}
+
+func executableName(goos string) string {
+	if goos == "windows" {
+		return "ledgerctl.exe"
+	}
+
+	return "ledgerctl"
 }
 
 // findAsset finds the archive asset matching the current OS/arch in the release.
 func findAsset(release *releaseInfo) (*assetInfo, error) {
-	want := archiveAssetName()
+	return findAssetForPlatform(release, runtime.GOOS, runtime.GOARCH)
+}
+
+func findAssetForPlatform(release *releaseInfo, goos, goarch string) (*assetInfo, error) {
+	want := archiveAssetName(goos, goarch)
 	for i := range release.Assets {
 		if release.Assets[i].Name == want {
 			return &release.Assets[i], nil
 		}
 	}
 
-	return nil, fmt.Errorf("no binary available for %s/%s (expected asset %q)", runtime.GOOS, runtime.GOARCH, want)
+	return nil, fmt.Errorf("no binary available for %s/%s (expected asset %q)", goos, goarch, want)
 }
 
 // findChecksumsAsset finds the checksums.txt asset in the release.
--- a/cmd/ledgerctl/upgrade/github.go
+++ b/cmd/ledgerctl/upgrade/github.go
@@ -7,10 +7,11 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
-	"regexp"
 	"runtime"
 	"strings"
 	"sync"
+
+	"github.com/Masterminds/semver/v3"
 )
 
 const (
@@ -22,6 +23,8 @@ type releaseInfo struct {
 	TagName         string      `json:"tag_name"`
 	Name            string      `json:"name"`
 	TargetCommitish string      `json:"target_commitish"`
+	Draft           bool        `json:"draft"`
+	Prerelease      bool        `json:"prerelease"`
 	Assets          []assetInfo `json:"assets"`
 }
 
@@ -32,13 +35,14 @@ type assetInfo struct {
 
 // fetchRelease fetches the release info for the given channel.
 // For "nightly", it fetches the release tagged "nightly".
-// For "stable", it fetches the most recent release matching a semver tag.
-func fetchRelease(channel string) (*releaseInfo, error) {
+// For "stable", it fetches the most recent final release matching the running
+// binary's major version.
+func fetchRelease(channel, currentVersion string) (*releaseInfo, error) {
 	switch channel {
 	case "nightly":
 		return fetchNightlyRelease()
 	case "stable":
-		return fetchStableRelease()
+		return fetchStableRelease(currentVersion)
 	default:
 		return nil, fmt.Errorf("unknown channel %q; use \"nightly\" or \"stable\"", channel)
 	}
@@ -57,29 +61,65 @@ func fetchNightlyRelease() (*releaseInfo, error) {
 	return &release, nil
 }
 
-var semverTagRe = regexp.MustCompile(`^v3\.\d+\.\d+$`)
+func fetchStableRelease(currentVersion string) (*releaseInfo, error) {
+	url := fmt.Sprintf("https://api.github.com/repos/%s/releases", githubRepo)
 
-func fetchStableRelease() (*releaseInfo, error) {
-	url := fmt.Sprintf("https://api.github.com/repos/%s/releases?per_page=20", githubRepo)
-
-	var releases []releaseInfo
-
-	err := githubGet(url, &releases)
-	if err != nil {
-		return nil, err
-	}
-
-	return findStableRelease(releases)
+	return fetchStableReleaseFromURL(currentVersion, url)
 }
 
-func findStableRelease(releases []releaseInfo) (*releaseInfo, error) {
-	for i := range releases {
-		if semverTagRe.MatchString(releases[i].TagName) {
-			return &releases[i], nil
+func fetchStableReleaseFromURL(currentVersion, releasesURL string) (*releaseInfo, error) {
+	current, err := semver.NewVersion(currentVersion)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"cannot select a stable release for current version %q: %w; use --channel nightly",
+			currentVersion,
+			err,
+		)
+	}
+
+	const releasesPerPage = 100
+
+	for page := 1; ; page++ {
+		url := fmt.Sprintf("%s?per_page=%d&page=%d", releasesURL, releasesPerPage, page)
+
+		var releases []releaseInfo
+		if err := githubGet(url, &releases); err != nil {
+			return nil, err
+		}
+
+		if release := findStableRelease(releases, current.Major()); release != nil {
+			return release, nil
+		}
+
+		if len(releases) < releasesPerPage {
+			break
 		}
 	}
 
-	return nil, errors.New("no stable v3 release found; use --channel nightly")
+	return nil, fmt.Errorf(
+		"no final release found for major v%d; use --channel nightly",
+		current.Major(),
+	)
+}
+
+func findStableRelease(releases []releaseInfo, currentMajor uint64) *releaseInfo {
+	for i := range releases {
+		release := &releases[i]
+		if release.Draft || release.Prerelease {
+			continue
+		}
+
+		releaseVersion, err := semver.StrictNewVersion(strings.TrimPrefix(release.TagName, "v"))
+		if err != nil || releaseVersion.Prerelease() != "" {
+			continue
+		}
+
+		if releaseVersion.Major() == currentMajor {
+			return release
+		}
+	}
+
+	return nil
 }
 
 var (
--- a/cmd/ledgerctl/upgrade/github.go
+++ b/cmd/ledgerctl/upgrade/github.go
@@ -8,10 +8,13 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
+	"runtime/debug"
+	"strconv"
 	"strings"
 	"sync"
 
 	"github.com/Masterminds/semver/v3"
+	"golang.org/x/mod/module"
 )
 
 const (
@@ -63,12 +66,16 @@ func fetchNightlyRelease() (*releaseInfo, error) {
 
 func fetchStableRelease(currentVersion string) (*releaseInfo, error) {
 	url := fmt.Sprintf("https://api.github.com/repos/%s/releases", githubRepo)
+	modulePath := ""
+	if buildInfo, ok := debug.ReadBuildInfo(); ok {
+		modulePath = buildInfo.Main.Path
+	}
 
-	return fetchStableReleaseFromURL(currentVersion, url)
+	return fetchStableReleaseFromURL(currentVersion, modulePath, url)
 }
 
-func fetchStableReleaseFromURL(currentVersion, releasesURL string) (*releaseInfo, error) {
-	current, err := semver.NewVersion(currentVersion)
+func fetchStableReleaseFromURL(currentVersion, modulePath, releasesURL string) (*releaseInfo, error) {
+	currentMajor, err := versionMajor(currentVersion, modulePath)
 	if err != nil {
 		return nil, fmt.Errorf(
 			"cannot select a stable release for current version %q: %w; use --channel nightly",
@@ -87,7 +94,7 @@ func fetchStableReleaseFromURL(currentVersion, releasesURL string) (*releaseInfo
 			return nil, err
 		}
 
-		if release := findStableRelease(releases, current.Major()); release != nil {
+		if release := findStableRelease(releases, currentMajor); release != nil {
 			return release, nil
 		}
 
@@ -98,8 +105,31 @@ func fetchStableReleaseFromURL(currentVersion, releasesURL string) (*releaseInfo
 
 	return nil, fmt.Errorf(
 		"no final release found for major v%d; use --channel nightly",
-		current.Major(),
+		currentMajor,
 	)
+}
+
+func versionMajor(currentVersion, modulePath string) (uint64, error) {
+	current, versionErr := semver.NewVersion(currentVersion)
+	if versionErr == nil {
+		return current.Major(), nil
+	}
+
+	_, pathMajor, ok := module.SplitPathVersion(modulePath)
+	if !ok || pathMajor == "" {
+		return 0, fmt.Errorf(
+			"version is not semantic and module path %q has no major suffix: %w",
+			modulePath,
+			versionErr,
+		)
+	}
+
+	major, err := strconv.ParseUint(strings.TrimPrefix(pathMajor, "/v"), 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf("parsing module major %q: %w", pathMajor, err)
+	}
+
+	return major, nil
 }
 
 func findStableRelease(releases []releaseInfo, currentMajor uint64) *releaseInfo {
--- a/cmd/ledgerctl/upgrade/github.go
+++ b/cmd/ledgerctl/upgrade/github.go
@@ -15,7 +15,7 @@ import (
 
 const (
 	githubRepo  = "formancehq/ledger"
-	projectName = "ledger-v3"
+	projectName = "ledger"
 )
 
 type releaseInfo struct {
--- a/cmd/ledgerctl/upgrade/github.go
+++ b/cmd/ledgerctl/upgrade/github.go
@@ -57,7 +57,7 @@ func fetchNightlyRelease() (*releaseInfo, error) {
 	return &release, nil
 }
 
-var semverTagRe = regexp.MustCompile(`^v\d+\.\d+\.\d+`)
+var semverTagRe = regexp.MustCompile(`^v3\.\d+\.\d+$`)
 
 func fetchStableRelease() (*releaseInfo, error) {
 	url := fmt.Sprintf("https://api.github.com/repos/%s/releases?per_page=20", githubRepo)
@@ -69,13 +69,17 @@ func fetchStableRelease() (*releaseInfo, error) {
 		return nil, err
 	}
 
+	return findStableRelease(releases)
+}
+
+func findStableRelease(releases []releaseInfo) (*releaseInfo, error) {
 	for i := range releases {
 		if semverTagRe.MatchString(releases[i].TagName) {
 			return &releases[i], nil
 		}
 	}
 
-	return nil, errors.New("no stable release found; use --channel nightly")
+	return nil, errors.New("no stable v3 release found; use --channel nightly")
 }
 
 var (
--- a/cmd/ledgerctl/upgrade/replace.go
+++ b/cmd/ledgerctl/upgrade/replace.go
@@ -100,7 +100,14 @@ func replaceExecutable(currentPath, replacementPath, goos string) error {
 
 	if err := os.Rename(replacementPath, currentPath); err != nil {
 		if rollbackErr := os.Rename(backupPath, currentPath); rollbackErr != nil {
-			return fmt.Errorf("installing replacement: %w; restoring current executable: %w", err, rollbackErr)
+			return fmt.Errorf(
+				"installing replacement: %w; restoring current executable from backup %q: %w; recover manually by renaming %q to %q",
+				err,
+				backupPath,
+				rollbackErr,
+				backupPath,
+				currentPath,
+			)
 		}
 
 		return fmt.Errorf("installing replacement: %w", err)
--- a/cmd/ledgerctl/upgrade/replace.go
+++ b/cmd/ledgerctl/upgrade/replace.go
@@ -4,11 +4,11 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 )
 
-// replaceBinary atomically replaces the current binary with the new one at srcPath.
-// It resolves symlinks to find the actual binary path, preserves permissions,
-// and uses rename for atomicity (requires same filesystem).
+// replaceBinary replaces the current binary with the new one at srcPath.
+// It resolves symlinks to find the actual binary path and preserves permissions.
 func replaceBinary(srcPath string) (string, error) {
 	currentPath, err := os.Executable()
 	if err != nil {
@@ -67,12 +67,38 @@ func replaceBinary(srcPath string) (string, error) {
 		return "", fmt.Errorf("closing temp file: %w", err)
 	}
 
-	// Atomic rename.
-	if err := os.Rename(tmpPath, currentPath); err != nil {
+	if err := replaceExecutable(currentPath, tmpPath, runtime.GOOS); err != nil {
 		_ = os.Remove(tmpPath)
 
 		return "", fmt.Errorf("replacing binary: %w (you may need elevated permissions)", err)
 	}
 
 	return currentPath, nil
+}
+
+func replaceExecutable(currentPath, replacementPath, goos string) error {
+	if goos != "windows" {
+		return os.Rename(replacementPath, currentPath)
+	}
+
+	backupPath := currentPath + ".old"
+	if err := os.Remove(backupPath); err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("removing previous backup: %w", err)
+	}
+
+	// Windows cannot replace a running executable in place. Renaming it keeps the
+	// current process alive while freeing the original path for the new binary.
+	if err := os.Rename(currentPath, backupPath); err != nil {
+		return fmt.Errorf("backing up current executable: %w", err)
+	}
+
+	if err := os.Rename(replacementPath, currentPath); err != nil {
+		if rollbackErr := os.Rename(backupPath, currentPath); rollbackErr != nil {
+			return fmt.Errorf("installing replacement: %w; restoring current executable: %w", err, rollbackErr)
+		}
+
+		return fmt.Errorf("installing replacement: %w", err)
+	}
+
+	return nil
 }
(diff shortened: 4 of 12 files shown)


.goreleaser.yml
```diff
@@ -67,6 +68,10 @@ archives:
   - ledgerctl
   formats:
   - tar.gz
+  format_overrides:
+  - goos: windows
+    formats:
+    - zip
```
NumaryBot: 🟠 **[major] Windows self-upgrade incompatible with ZIP artifacts**

The release config now builds Windows ZIPs, but the existing `ledgerctl upgrade` command still constructs a `.tar.gz` asset name and extracts it using gzip/tar looking for `ledgerctl` (not `ledgerctl.exe`). This means Windows users who try to self-upgrade will fail because the upgrade path cannot locate or extract the newly released `.zip` artifact.

**Suggestion:** Update the upgrade/download logic to detect the Windows platform and select the `.zip` artifact, then extract `ledgerctl.exe` from the ZIP instead of relying on the gzip/tar path. Ensure the binary name check also accounts for the `.exe` suffix on Windows.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1644#issuecomment-5104179917

.goreleaser.yml
```diff
@@ -46,6 +46,7 @@ builds:
   goos:
   - linux
   - darwin
+  - windows
```
NumaryBot, line 64: 🔴 **[blocker] Mixed-binary archive config causes GoReleaser to reject Windows builds**

The archive configuration groups both `ledger-server` and `ledgerctl` binaries together. Now that `ledgerctl` has Windows builds but `ledger-server` does not, GoReleaser's binary-count consistency check will see two binaries on Linux/macOS and only one on Windows, causing the release to fail before any artifact is published.

**Suggestion:** Split the archive into a separate `ledgerctl`-only archive (or a separate `ledger-server` archive) so that each archive contains a consistent set of binaries per platform, or explicitly configure GoReleaser to allow differing binary counts for that archive.

.goreleaser.yml
NumaryBot: 🟠 **[major] Windows self-upgrade incompatible with ZIP artifacts**

The release now produces ZIP archives for Windows, but the `ledgerctl upgrade` command still constructs a `.tar.gz` asset name and uses gzip/tar extraction looking for a binary named `ledgerctl` (without `.exe`). Windows users who attempt a self-upgrade will fail to locate or extract the new ZIP artifact.

**Suggestion:** Update the upgrade/download logic to detect the Windows platform, select the `.zip` artifact instead of `.tar.gz`, and extract `ledgerctl.exe` from the ZIP. Ensure the binary name check also accounts for the `.exe` suffix on Windows.

cmd/ledgerctl/upgrade/download.go
```diff
@@ -119,19 +121,39 @@ func downloadAndVerify(archiveAsset, checksumsAsset *assetInfo, spinner *pterm.S
...
+	archiveInfo, err := tmpArchive.Stat()
+	if err != nil {
+		return "", fmt.Errorf("reading archive metadata: %w", err)
+	}
+
+	return extractBinary(
+		tmpArchive,
+		archiveInfo.Size(),
+		archiveAsset.Name,
+		executableName(runtime.GOOS),
```
NumaryBot, line 140: 🟠 **[major] Running-executable replacement fails on Windows**

Even if ZIP extraction is fixed, the final `os.Rename` in `replaceBinary` will fail on Windows with an 'access denied' error because Windows keeps the currently-running `.exe` file locked. The self-upgrade flow therefore cannot complete on Windows without a deferred or helper-based replacement strategy.

**Suggestion:** Implement a Windows-safe replacement path, such as renaming the current executable to a `.old` file and placing the new binary at the original path, or using a helper process / scheduled task to swap the binary after the current process exits.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1644#issuecomment-5104179917

.goreleaser.yml
```diff
@@ -65,8 +66,13 @@ archives:
   ids:
   - ledger-server
   - ledgerctl
+  allow_different_binary_count: true
   formats:
   - tar.gz
+  format_overrides:
+  - goos: windows
+    formats:
+    - zip
```
NumaryBot: 🟠 **[major] Windows self-upgrade incompatible with ZIP artifacts**

The release now produces ZIP archives for Windows, but the `ledgerctl upgrade` command still constructs a `.tar.gz` asset name and uses gzip/tar extraction looking for a binary named `ledgerctl` (without `.exe`). Windows users who attempt a self-upgrade will fail to locate or extract the new ZIP artifact.

**Suggestion:** Update the upgrade/download logic to detect the Windows platform, select the `.zip` artifact instead of `.tar.gz`, and extract `ledgerctl.exe` from the ZIP. Ensure the binary name check also accounts for the `.exe` suffix on Windows.

cmd/ledgerctl/upgrade/replace_test.go
```diff
@@ -0,0 +1,89 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			dir := t.TempDir()
+			currentPath := filepath.Join(dir, "ledgerctl")
+			replacementPath := filepath.Join(dir, "ledgerctl-new")
+
+			require.NoError(t, os.WriteFile(currentPath, []byte("old"), 0o755))
+			require.NoError(t, os.WriteFile(replacementPath, []byte("new"), 0o755))
+			require.NoError(t, replaceExecutable(currentPath, replacementPath, test.goos))
```
NumaryBot: 🟠 **[major] Unix rename test case fails on Windows CI runners**

The newly added upgrade test does not skip the `unix` case when running on Windows. That branch calls `replaceExecutable(..., "linux")`, which takes the non-Windows code path and calls `os.Rename(replacementPath, currentPath)` while the destination already exists. Windows `os.Rename` cannot overwrite an existing file, causing `go test ./cmd/ledgerctl/upgrade` to fail on Windows before validating the actual Windows upgrade path.

**Suggestion:** Add a `runtime.GOOS == "windows"` guard (or `t.Skip`) for the unix-rename sub-case when the host OS is Windows, or restructure the test to avoid pre-creating the destination file for the simulated non-Windows platform.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1644#issuecomment-5104179917

cmd/ledgerctl/upgrade/replace_test.go
```diff
@@ -0,0 +1,93 @@
...
+				t.Skip("Windows cannot exercise Unix overwrite semantics")
+			}
+
+			dir := t.TempDir()
+			currentPath := filepath.Join(dir, "ledgerctl")
+			replacementPath := filepath.Join(dir, "ledgerctl-new")
+
+			require.NoError(t, os.WriteFile(currentPath, []byte("old"), 0o755))
+			require.NoError(t, os.WriteFile(replacementPath, []byte("new"), 0o755))
+			require.NoError(t, replaceExecutable(currentPath, replacementPath, test.goos))
```
NumaryBot: 🟠 **[major] Unix rename test case fails on Windows CI runners**

The newly added upgrade test does not skip the `unix` case when running on Windows. That branch calls `replaceExecutable(..., "linux")`, which takes the non-Windows code path and calls `os.Rename(replacementPath, currentPath)` while the destination already exists. Windows `os.Rename` cannot overwrite an existing file, causing `go test ./cmd/ledgerctl/upgrade` to fail on Windows before validating the actual Windows upgrade path.

**Suggestion:** Add a `runtime.GOOS == "windows"` guard (or `t.Skip`) for the unix-rename sub-case when the host OS is Windows, or restructure the test to avoid pre-creating the destination file for the simulated non-Windows platform.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1644#issuecomment-5104179917

flemzord:
Revalidated live head `524b5118`. The current diff now handles GoReleaser’s platform-specific binary count, selects and extracts the Windows ZIP/`ledgerctl.exe`, performs the Windows backup/install/rollback swap, and isolates Unix replacement coverage behind a non-Windows build tag. The earlier NumaryBot discussions target superseded heads and no longer match the current source. All functional and coverage checks are green. I found no correctness/runtime blocker; GitHub approval is intentionally disabled by this cron configuration.

.goreleaser.yml
```diff
@@ -1,7 +1,7 @@
 # yaml-language-server: $schema=https://goreleaser.com/static/schema-pro.json
 ---
 version: 2
-project_name: ledger-v3
+project_name: ledger
```
NumaryBot, line 4: 🟠 **[major] Changing `project_name` renames all release artifacts, breaking existing consumers**

Changing `project_name` from `ledger-v3` to `ledger` renames every generated archive from `ledger-v3_<os>-<arch>` to `ledger_<os>-<arch>`. Any consumer (scripts, CI pipelines, `ledgerctl upgrade` logic) that previously targeted the v3 naming will fail to locate artifacts unless migration is coordinated explicitly.

**Suggestion:** Preserve `project_name: ledger-v3` to keep backward-compatible artifact names, or ensure all downstream consumers (including the self-upgrade download URL construction) are updated atomically alongside this change.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1644#issuecomment-5104179917

flemzord:
Revalidated live head `152db858`. The final commit keeps GoReleaser and `ledgerctl upgrade` on the same `ledger_<os>-<arch>` artifact names used by `main`. NumaryBot’s remaining concern is pre-stable release-artifact compatibility, with no demonstrated correctness/runtime regression under this cron policy. Functional CI and Codecov are green; only the NumaryBot review status remains failed. I found no correctness/runtime blocker on this head. GitHub approval is intentionally disabled by this cron configuration.

cmd/ledgerctl/upgrade/replace.go
```diff
@@ -4,11 +4,11 @@ import (
...
 	"os"
 	"path/filepath"
+	"runtime"
 )
 
-// replaceBinary atomically replaces the current binary with the new one at srcPath.
-// It resolves symlinks to find the actual binary path, preserves permissions,
-// and uses rename for atomicity (requires same filesystem).
+// replaceBinary replaces the current binary with the new one at srcPath.
+// It resolves symlinks to find the actual binary path and preserves permissions.
```
paul-nicolas, line 11: [Major] The upgraded binary installs as `0600` and is not executable — this doc comment's "preserves permissions" is not true

Re: `os.OpenFile` at line 50 (just outside the diff), anchored here because this is the comment asserting the behavior.

`replaceBinary` creates the staging file with `os.CreateTemp` (line 32), which yields mode `0600`. Line 50 then does:

```go
dst, err := os.OpenFile(tmpPath, os.O_WRONLY|os.O_TRUNC, info.Mode())
```

Without `O_CREATE`, the `perm` argument is **ignored for an existing file**. So `info.Mode()` is silently discarded and the file keeps `0600`. Verified empirically:

```
current mode:            -rwxr-xr-x
CreateTemp mode:         -rw-------
after OpenFile(perm=-rwxr-xr-x): -rw-------
INSTALLED mode:          -rw-------   executable-by-owner=false
```

A "successful" `ledgerctl upgrade` on Linux/macOS therefore leaves a **non-executable binary** — the next invocation is `permission denied`, and the user cannot even self-upgrade back out of it.

Why this belongs to this PR even though line 50 predates it: until now the bug was unreachable, because `findAsset` looked for `ledger_*` while v3 published `ledger-v3_*`, so upgrade always failed early at asset lookup. This PR aligns those names, which makes the install step reachable for the first time (`--channel nightly` will now resolve, download, verify and install end-to-end). It also rewrites this very comment to re-assert permission preservation. Shipping the Windows path on top of a install step that bricks the Unix binary is worth fixing in the same change.

Fix: `os.Chmod(tmpPath, info.Mode())` after the copy and before `replaceExecutable`, or drop `CreateTemp` for `os.OpenFile(uniqueName, os.O_WRONLY|os.O_CREATE|os.O_EXCL, info.Mode())`. Add a test asserting the installed file keeps the source mode's exec bit — that assertion is what would have caught this.

.goreleaser.yml
paul-nicolas, line 64: [Medium] New Windows build target with zero compilation coverage in CI — it first executes at tag push

Adding `windows` here creates a release target that nothing in CI ever builds. Every runner across both workflows is Linux:

```
release.yml:13   runs-on: namespace-profile-linux-amd64-4vcpu
main.yml:34,60,83,106,129,146,192,209,244,299   runs-on: namespace-profile-linux-amd64-4vcpu
```

and no job sets `GOOS` — the Build job (`main.yml:203`) is a plain `nix develop --command go build ./...` against the host platform. `grep -iE 'windows|GOOS'` over `.github/workflows/` and the `justfile` returns nothing.

So the first time `GOOS=windows` is compiled is inside `just release` in `release.yml`, i.e. **after** the tag is pushed. It compiles today (`CGO_ENABLED=0 GOOS=windows GOARCH=amd64 go build ./cmd/ledgerctl` and `arm64` both succeed locally, also with the real ldflags), but nothing guards it: any future Unix-only import reachable from `cmd/ledgerctl` — `syscall`, a `golang.org/x/sys/unix` dependency, an unguarded `os/signal` usage — merges green and breaks the release job at the worst possible moment.

Related: `replaceExecutable(..., "windows")` in `replace_test.go` runs on the Linux runner, so it exercises the backup/rollback ordering but never the constraint the branch exists for (a running `.exe` cannot be replaced in place).

Fix (cheapest useful step) — add to the `Build` job in `main.yml`:

```yaml
      - name: Cross-compile ledgerctl for Windows
        run: >
          nix develop --command bash -c "CGO_ENABLED=0 GOOS=windows GOARCH=amd64 go build -o /dev/null ./cmd/ledgerctl && CGO_ENABLED=0 GOOS=windows GOARCH=arm64 go build -o /dev/null ./cmd/ledgerctl"
```

A `windows-latest` job running `go test ./cmd/ledgerctl/upgrade/...` would additionally cover the real rename semantics, and would convert several currently-simulated assertions into genuine coverage.

.goreleaser.yml
paul-nicolas, line 4: [Medium] The rename makes v3 artifacts byte-identical to the still-active v2 line, removing the only major-version disambiguator

The rename is directionally right — `github.go:18` already hardcodes `projectName = "ledger"`, so v3's `ledger-v3_*` assets could never be found. But `formancehq/ledger` is still actively releasing the v2 line from the same repo with the same effective naming, and v2 got there first:

```
$ gh release view v2.4.12 --json assets   # created 2026-07-24, TWO DAYS AFTER v3.0.0-alpha.13
ledger_checksums.txt
ledger_darwin-amd64.tar.gz
ledger_linux-amd64.tar.gz
ledger_linux-arm64.tar.gz
```

Post-rename, a v3 release publishes `ledger_linux-amd64.tar.gz` — the same filename v2 publishes. `ledger-v3` was the only thing distinguishing them.

This compounds with a pre-existing channel bug. `fetchStableRelease` (`github.go:62`) takes the **first** tag matching `^v\d+\.\d+\.\d+` from a created-at-desc list, and that regex happily matches `v2.4.12`:

```
$ gh api "repos/formancehq/ledger/releases?per_page=6" -q '.[].tag_name'
v2.4.12          <- picked by --channel stable
v2.3.22
v3.0.0-alpha.13
```

So `ledgerctl upgrade --channel stable` from a v3 binary resolves to a **v2** release. Before this PR the archive name didn't match, giving a clear `no binary available for linux/amd64 (expected asset "ledger_linux-amd64.tar.gz")`. After it, the archive name *does* match, and only two accidents stop a cross-major install: `findChecksumsAsset` (`github.go:196`) requires an exact `== "checksums.txt"` while v2 ships `ledger_checksums.txt`, and v2's tarball has no `ledgerctl` at all (`LICENSE README.md ledger benchmarks`). It fails safe today — but on the incidental spelling of a filename in a repo you don't control the release config of.

To be clear on scope: the `fetchStableRelease` mis-resolution is pre-existing and untouched here, so this is not a regression. The point is that the PR removes one of the two independent defenses while leaving `--channel stable` broken against its own premise.

Fix — either keep the artifact names distinct (set `projectName = "ledger-v3"` in `github.go:18` rather than renaming the goreleaser project, or major-qualify `name_template`), or make the channel major-aware so `--channel stable` filters to `v3.*` and stops resolving to the v2 line. Worth deciding deliberately rather than inheriting whichever release publishes last.

paul-nicolas (commented):

Reviewed by Codex and two independent Claude passes (Go code / release+docs), then every Critical–Medium finding was put through an adversarial verifier instructed to refute it. The goreleaser side of this PR holds up well: `goreleaser check` passes on the pinned goreleaser-pro 2.17.0, `allow_different_binary_count` and `format_overrides[].formats` are valid v2 keys, a real snapshot run does emit `ledger_windows-{amd64,arm64}.zip` containing `ledger_windows-amd64/ledgerctl.exe`, and `GOOS=windows` cross-compiles cleanly for both arches. The rename is also a genuine fix: `github.go` already hardcoded `projectName = "ledger"` while v3 published `ledger-v3_*`, so asset lookup could never match.

**Verification outcome:** 13 Critical/Major/Medium findings entered verification → **5 survived** (1 Major, 4 Medium), 3 were downgraded to Minor/Nit, 5 were rejected outright.

Rejected, with reason:
- *"Rollback-failed branch never discloses where the old binary went"* — **invalid**. `os.Rename` returns `*os.LinkError`, which already formats as `rename <old> <new>: <err>`; the forced branch prints `…restoring current executable: rename /…/ledgerctl.exe.old /…/ledgerctl.exe: permission denied`. The path *is* disclosed.
- *"Crash window between the two renames needs a helper process / startup recovery"* — **downgraded to a docs note**. Two adjacent `MoveFileEx` calls with no I/O between them, `.old` survives a crash, and renaming the running image aside is the only viable Windows approach. Disproportionate for a short-lived CLI.
- *"Stale locked `.old` permanently blocks upgrades"* and *"`replaceBinary` destroys the last usable binary"* — **downgraded to Nit**. Go's Windows `os.Rename` is `MoveFileEx(MOVEFILE_REPLACE_EXISTING)`, so a locked `.old` fails the backup rename too — the abort branch only relocates the same error. And `.old` survives, so recovery is one rename plus a re-download.
- *"`require.EqualError` violates the errors.Is rule"* — **invalid**. That rule is scoped to control flow; string-matched assertions are house style here (196 `ErrorIs` / 15 `ErrorContains` / 2 `EqualError`), and there is no sentinel to match against.
- *"Zip directory entries can false-match the binary name"* — **invalid**. `archive/zip` writes directory names with a trailing `/`, so they never match.

Codex and Claude agreed independently on the missing Windows CI coverage and on the `extractBinary` arity nit; they disagreed on the Windows-rollback severity (Codex: major) — resolved to Nit/docs by the verifier evidence above.

- `cmd/ledgerctl/upgrade/download.go:222` — a non-regular tar entry (symlink/hardlink/dir without trailing slash) matching the binary name extracts as a **0-byte file with a nil error** and is installed under a green "upgraded" message. Verified with a probe test. Pre-existing for tar, newly extended to the zip path. A single zero-byte guard in `writeExtractedBinary` covers every variant.
- `cmd/ledgerctl/upgrade/download.go:228` — unbounded `io.Copy`; no cap on extracted size. `io.LimitReader` / `file.UncompressedSize64` check.
- `cmd/ledgerctl/upgrade/download.go:127` — the `Seek(0, io.SeekStart)` is now dead code: both branches read exclusively via `ReadAt` (`io.NewSectionReader`, `zip.NewReader`). Leftover from the old `io.Reader` signature.
- `cmd/ledgerctl/upgrade/download.go:145` — `extractBinary` takes 4 positional params; CLAUDE.md requires an input struct above 2. Three of the four are transposable strings/ints.
- `cmd/ledgerctl/upgrade/download_test.go:43` — fixtures build `ledger/ledgerctl.exe`, but `wrap_in_directory` names the directory from `name_template`: the real archive is `ledger_windows-amd64/ledgerctl.exe` + `README.md`. Verified against the published `ledger-v3_linux-amd64.tar.gz` and a live snapshot run. Both satisfy the suffix match, so the test passes for the wrong reason — tightening `isBinaryEntry` to an exact compare keeps the suite green while breaking every real release (confirmed by patching it locally).
- `cmd/ledgerctl/upgrade/download_test.go` — `binary %q not found in archive` is untested for **both** formats; gzip-open failure, `zip.NewReader` failure, the copy-failure cleanup and the multi-entry skip path are all uncovered (`writeExtractedBinary` 45.5%, `extractZipBinary` 61.1%).
- `cmd/ledgerctl/upgrade/replace.go:85` — the fixed `.old` name makes a locked backup a hard (if transient) failure; a timestamped backup path removes the failure mode entirely.
- `cmd/ledgerctl/upgrade/replace_test.go:19` — single-case table leaves `wantBackup` always true and the `else { require.ErrorIs(…, os.ErrNotExist) }` branch dead. Vestigial: `524b5118c` moved the `linux` case out to `replace_unix_test.go` but left the fields behind.
- `cmd/ledgerctl/upgrade/github_test.go:55` — `TestFindAssetForPlatform` covers only the found path; the not-found error (the signal if a platform's asset is ever dropped) is untested. `TestExecutableName` is three bare asserts rather than a table.
- `docs/ops/cli.md:4930` — the section intro still says the binary is replaced "in-place", which now contradicts the Windows bullet added 18 lines below and the deliberate removal of "atomically" from `replaceBinary`'s doc comment.
- `docs/ops/cli.md:4948` — the `.old` bullet omits both its lifetime (never cleaned up except by the *next* upgrade, so it persists indefinitely) and the interrupted-upgrade recovery step (rename `.old` back if `ledgerctl.exe` is missing).
- `docs/ops/cli.md:4947` — doesn't state that the Windows zip ships `ledgerctl.exe` **only**; Linux/macOS tarballs still carry `ledger-server` too.
- `.goreleaser.yml:65` — `id: "{{ .ProjectName }}"` is not templated for archive ids; `artifacts.json` shows the literal `{{ .ProjectName }}`.
- `.goreleaser.yml` — image labels claim `licenses=MIT` but the repo has no `LICENSE` file, so archives ship none (the default `files` glob finds only `README.md`).

.goreleaser.yml
```diff
@@ -65,8 +66,13 @@ archives:
   ids:
   - ledger-server
   - ledgerctl
+  allow_different_binary_count: true
```
paul-nicolas: [Medium] `allow_different_binary_count` disarms the binary-count guard for all six archives, not just Windows

This is a single `archives` entry covering every platform, so relaxing the count check to accommodate the Windows archive (which legitimately carries only `ledgerctl.exe`) also silences it for the four Linux/macOS archives that are supposed to contain **both** `ledger-server` and `ledgerctl`. If a future change drops `linux` from the `ledger-server` build's `goos`, or a build id is renamed and silently stops matching, the Linux tarball ships with one binary and the release still succeeds — exactly the mistake this guard exists to catch.

A split configuration keeps the guard armed and makes the Windows/Unix difference explicit rather than blanket-permitted: add a third build id `ledgerctl-windows` (`goos: [windows]`, same ldflags/env), revert the existing `ledgerctl` build to `[linux, darwin]`, and use two `archives` entries — `unix` (`ids: [ledger-server, ledgerctl]`, `tar.gz`) and `windows` (`ids: [ledgerctl-windows]`, `zip`) — with the same `name_template` and `wrap_in_directory`. That deletes this line and makes `format_overrides` redundant too.

Both configurations were run through `goreleaser release --snapshot --clean --skip=publish,docker,sign,sbom` on the `flake.nix`-pinned goreleaser-pro 2.17.0 and produce **identical output**: the same six archive names, `ledger_windows-arm64/ledgerctl.exe` inside the zip, both binaries in the tarballs — with the binary-count guard still active.

Not merge-blocking, and the current form is legitimate goreleaser usage. But it trades a permanent safety net for a local convenience, and the alternative costs about eight lines of config.

docs/ops/cli.md
```diff
@@ -4944,7 +4944,8 @@ ledgerctl upgrade [flags]
 **Behavior:**
 - Downloads the latest release from `formancehq/ledger` GitHub releases
 - Verifies the archive's SHA256 checksum against `checksums.txt`
-- Extracts the `ledgerctl` binary and atomically replaces the current binary
+- Extracts `ledgerctl` from the Linux/macOS tarball or `ledgerctl.exe` from the Windows ZIP archive, then replaces the current binary
+- On Windows, keeps the previous executable as `ledgerctl.exe.old` because the running executable cannot be replaced in place
```
paul-nicolas: [Medium] The PR ships Windows binaries but no documented way to obtain one

Re: the Installation section at lines 9–17 (outside the diff), raised here because this bullet is where the Windows behavior lands.

The whole premise of this PR is that "Windows users currently have no prebuilt `ledgerctl` artifact in GitHub releases". After it, the artifact exists — but the only documented way to get `ledgerctl` anywhere in `docs/` is still building from source:

```markdown
## Installation

```bash
# Build from source
just build-client

# Or directly with Go
go build -o build/ledgerctl ./cmd/ledgerctl
```
```

That is the least likely path for the audience this PR targets: `just` and a nix/Go toolchain on Windows is a much higher bar than downloading a zip. And `ledgerctl upgrade` cannot bootstrap it — it self-updates an *already installed* binary.

Per the repo's CLAUDE.md documentation-maintenance rule ("Update `docs/ops/cli.md` when modifying CLI commands, flags, or behavior"), the release-download path is the missing half of this change.

Fix — extend the Installation section with the prebuilt route, naming the actual artifacts this PR produces:

```markdown
### Download a prebuilt binary

Releases publish `ledger_<os>-<arch>` archives at
https://github.com/formancehq/ledger/releases:

- Linux/macOS: `ledger_linux-amd64.tar.gz`, `ledger_darwin-arm64.tar.gz`, … (contain `ledger-server` and `ledgerctl`)
- Windows: `ledger_windows-amd64.zip`, `ledger_windows-arm64.zip` (contain `ledgerctl.exe` only)

Extract the archive and put the binary on your `PATH`; from then on `ledgerctl upgrade` keeps it current.
```

That also resolves the Minor note about the Windows zip shipping `ledgerctl.exe` only.

.goreleaser.yml
flemzord: The rename is intentional for v3 alignment with `main`, and the migration is now documented in the PR. The upgrader, tests, release config, and installation docs all use `ledger_<os>-<arch>` atomically in this branch. I also restricted `--channel stable` to final `v3.x.y` tags so the shared v2/v3 repository cannot select a v2 release. A full GoReleaser snapshot validates the six renamed artifacts.

cmd/ledgerctl/upgrade/replace.go
flemzord: Fixed in `2832c95ca`. Staging now goes through `copyReplacement`, which copies the binary and explicitly applies the current executable mode with `os.Chmod` before the swap. `TestCopyReplacementPreservesMode` verifies both the contents and `0755` permissions.

.goreleaser.yml
flemzord: Fixed in `2832c95ca`. The regular Build job now cross-compiles `ledgerctl` with `CGO_ENABLED=0` for both `windows/amd64` and `windows/arm64`. The same commands and the full GoReleaser snapshot pass locally.

.goreleaser.yml
flemzord: Kept the `ledger` artifact name as an explicit alignment decision, but fixed the cross-major risk in `2832c95ca`: the stable selector now accepts final `v3.x.y` tags only (`^v3\\.\\d+\\.\\d+$`) and rejects both v2 releases and v3 prereleases in unit tests. The alpha artifact filename migration is documented in the PR.

.goreleaser.yml
flemzord: Fixed in `2832c95ca` using the suggested structural approach. Unix and Windows now have separate build IDs and archive entries; `allow_different_binary_count` and the format override are removed. The full snapshot still emits the same six artifacts while keeping GoReleaser's binary-count guard enabled.

docs/ops/cli.md
flemzord: Fixed in `2832c95ca`. The Installation section now documents the GitHub release download path, exact Unix and Windows artifact names and contents, PATH installation, and the transition to `ledgerctl upgrade` for subsequent updates.

cmd/ledgerctl/upgrade/replace.go
```diff
@@ -37,42 +37,74 @@ func replaceBinary(srcPath string) (string, error) {
...
+		return "", err
 	}
 
-	defer func() { _ = src.Close() }()
+	if err := replaceExecutable(currentPath, tmpPath, runtime.GOOS); err != nil {
+		_ = os.Remove(tmpPath)
 
-	dst, err := os.OpenFile(tmpPath, os.O_WRONLY|os.O_TRUNC, info.Mode())
+		return "", fmt.Errorf("replacing binary: %w (you may need elevated permissions)", err)
+	}
```
NumaryBot, line 50: 🔴 **[blocker] Upgraded binary left non-executable: OpenFile ignores permissions on existing file**

CreateTemp creates the staging file with mode 0600. The subsequent os.OpenFile call (O_WRONLY|O_TRUNC, info.Mode()) does not include O_CREATE, so the perm argument is silently ignored for an already-existing file. The installed binary therefore keeps mode 0600 (not executable by group/other, and not executable at all on paths that require it). After a successful 'ledgerctl upgrade', the next invocation fails with 'permission denied', and the user cannot self-upgrade back. This bug was previously unreachable because the asset name mismatch caused upgrade to always fail; this PR makes it reachable for the first time.

**Suggestion:** After writing the binary content, explicitly apply the original file's permissions with os.Chmod(tmpPath, info.Mode()) before moving it into place. Alternatively, remove the old temp file and recreate it with O_CREATE|O_WRONLY|O_TRUNC plus the correct mode so the perm argument takes effect.

.goreleaser.yml
```diff
@@ -50,6 +50,22 @@ builds:
...
+  - -extldflags=-static
+  env:
+  - CGO_ENABLED=0
+  goos:
+  - windows
+  goarch:
+  - amd64
+  - arm64
+
 nightly:
```
NumaryBot, line 69: 🟠 **[major] allow_different_binary_count disarms the binary-count guard for all archives, not just Windows**

The single archives entry covers all six platform targets. Setting allow_different_binary_count to accommodate Windows (which legitimately carries only ledgerctl.exe) also silences the guard for the four Linux/macOS archives that are supposed to contain both ledger-server and ledgerctl. A future change that drops a build id or renames a binary would allow a Linux tarball to ship with one binary and the release would still succeed without any warning.

**Suggestion:** Split into two archives entries: one for Unix platforms (ids: [ledger-server, ledgerctl], format: tar.gz) and one for Windows (ids: [ledgerctl-windows], format: zip). This keeps the binary-count guard armed for Unix archives and makes the Windows/Unix difference explicit rather than blanket-permitted.

.goreleaser.yml
NumaryBot, line 4: 🟠 **[major] project_name rename collides with v2 artifact names and breaks stable-channel upgrade**
<sub>reported by paul-nicolas, NumaryBot</sub>

Renaming project_name from ledger-v3 to ledger makes v3 artifacts byte-identical in naming to those published by the still-active v2 release line (e.g., ledger_linux-amd64.tar.gz). Compounding this, fetchStableRelease picks the first tag matching ^v\d+\.\d+\.\d+ from a created-at-desc list, which currently resolves to a v2 tag. A v3 user running 'ledgerctl upgrade --channel stable' would download a v2 binary, and with identical artifact filenames there is no way to distinguish them.

**Suggestion:** Coordinate the rename with a fix to the channel-selection logic (e.g., filter by semver major version or use a dedicated release channel marker), or revert to project_name: ledger-v3 until the channel bug is fixed. Ensure all downstream consumers are updated atomically.

.goreleaser.yml
```diff
@@ -50,6 +50,22 @@ builds:
...
+  main: ./cmd/ledgerctl
+  ldflags:
+  - -X github.com/formancehq/ledger/v3/internal/pkg/version.Version={{ .Version }}
+  - -X github.com/formancehq/ledger/v3/internal/pkg/version.Commit={{ .ShortCommit }}
+  - -X github.com/formancehq/ledger/v3/internal/pkg/version.BuildDate={{ .Date }}
+  - -extldflags=-static
+  env:
+  - CGO_ENABLED=0
+  goos:
+  - windows
```
NumaryBot, line 64: 🟡 **[minor] Windows build target has no pre-release CI compilation check**

All CI runners use Linux hosts and no job sets GOOS=windows. The first time GOOS=windows is compiled is inside the GoReleaser release job, which runs after the tag is pushed. Any future Unix-only import introduced into cmd/ledgerctl would merge green and break the release job at the worst possible moment.

**Suggestion:** Add a CI step (e.g., in main.yml) that runs GOOS=windows GOARCH=amd64 go build ./cmd/ledgerctl to catch cross-compilation regressions before tags are pushed.

docs/ops/cli.md
```diff
@@ -4944,12 +4955,13 @@ ledgerctl upgrade [flags]
...
```
NumaryBot: 🟡 **[minor] No documentation for downloading the new Windows prebuilt binary**

The PR adds Windows release artifacts but the only documented installation path remains building from source. Windows users who cannot use nix/just have no documented way to obtain the binary, which is the primary audience this PR targets.

**Suggestion:** Extend the Installation section in docs/ops/cli.md with instructions for downloading and installing the prebuilt ZIP artifact from GitHub Releases, including the actual artifact name pattern produced by this PR.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1644#issuecomment-5104179917

.goreleaser.yml
flemzord: Already addressed in `2832c95ca`: `semverTagRe` is now exactly `^v3\.\d+\.\d+$`, so stable cannot select v2 or v3 prereleases. Unit tests cover both exclusions. The `ledger` rename remains intentional and all consumers in this branch use it atomically.

cmd/ledgerctl/upgrade/replace.go
flemzord: Already fixed in `2832c95ca`: `copyReplacement` closes the staging file and explicitly calls `os.Chmod(dstPath, mode)` before `replaceExecutable`. `TestCopyReplacementPreservesMode` asserts the copied binary remains `0755`. This finding is based on the pre-fix implementation.

.goreleaser.yml
flemzord: Already fixed in `2832c95ca`: the regular Build job now cross-compiles `cmd/ledgerctl` for both `windows/amd64` and `windows/arm64` with `CGO_ENABLED=0`. The exact commands also pass locally.

.goreleaser.yml
flemzord: Already fixed in `2832c95ca`: `.goreleaser.yml` now has separate `ledger-unix` and `ledger-windows` archive entries and no `allow_different_binary_count`. A full GoReleaser snapshot produced all six expected archives with the guard enabled.

docs/ops/cli.md
flemzord: Already fixed in `2832c95ca`: the Installation section now documents GitHub Releases, the exact Windows ZIP names, `ledgerctl.exe`, PATH installation, and subsequent use of `ledgerctl upgrade`.

flemzord:
Revalidated live head `2832c95c`. The current commit preserves executable permissions during self-upgrade, separates Unix and Windows release archives without disabling GoReleaser’s binary-count guard, cross-compiles both Windows architectures in CI, and restricts the stable channel to final v3 releases. The remaining NumaryBot failure repeats findings already fixed on this same SHA; functional CI and Codecov are green. I found no correctness/runtime blocker. GitHub approval is intentionally disabled by this cron configuration.

cmd/ledgerctl/upgrade/github.go
```diff
@@ -57,7 +57,7 @@ func fetchNightlyRelease() (*releaseInfo, error) {
 	return &release, nil
 }
 
-var semverTagRe = regexp.MustCompile(`^v\d+\.\d+\.\d+`)
+var semverTagRe = regexp.MustCompile(`^v3\.\d+\.\d+$`)
```
paul-nicolas: [Medium] Stable channel now resolves nothing at all, and the major is pinned in a regex

This is the fix for the v2-selection problem from the last round, and the direction is right — but as written `--channel stable` fails for *every* user today. `formancehq/ledger` has no final `v3.x.y` release: only `v3.0.0-alpha.1` through `alpha.13`, alongside the still-active `v2.4.12` / `v2.3.22` line. So `findStableRelease` walks the page and always falls through to `no stable v3 release found`. Erroring beats silently downgrading a v3 CLI to v2, so keep the guard — the issue is how the intent is encoded.

- The major is hardcoded, so this line must be edited every major release, and a v3 CLI will never see v4 regardless of what ships.
- The property you actually want is "final release, not a prerelease". GitHub reports exactly that via the `prerelease` field — which `releaseInfo` (github.go:21-26) does not decode. The regex is standing in for data already on the wire.
- `per_page=20` (github.go:63) assumes the newest final v3 tag sits on page 1. With v2 patches still shipping from this repo and every v3 prerelease consuming a slot, a GA tag can drop off page 1, and stable would then falsely report that no stable release exists.

Fix: add `Prerelease bool \`json:"prerelease"\`` to `releaseInfo`, select the newest release whose tag parses as semver with the same major as the running binary and `prerelease == false`, and raise `per_page` or paginate. That keeps the v2-downgrade protection, drops the per-major regex edit, and removes the page-1 assumption.

Separately, since no eligible release exists until v3 GA: if a dead stable channel is acceptable for now, say so in the docs bullet — otherwise the first Windows user to try it gets a confusing error on the documented happy path.

cmd/ledgerctl/upgrade/replace.go
```diff
@@ -37,42 +37,74 @@ func replaceBinary(srcPath string) (string, error) {
...
 
-	return currentPath, nil
+	backupPath := currentPath + ".old"
+	if err := os.Remove(backupPath); err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("removing previous backup: %w", err)
+	}
+
+	// Windows cannot replace a running executable in place. Renaming it keeps the
+	// current process alive while freeing the original path for the new binary.
+	if err := os.Rename(currentPath, backupPath); err != nil {
```
paul-nicolas, line 97: [Medium] An interrupted Windows upgrade can leave no executable, and nothing tells the user how to recover

The `.old` rotation is the right answer to the locked-running-executable problem, so this is not a request to change the strategy. The gap is the residual window: two sequential renames, `currentPath` -> `.old` here, then `replacementPath` -> `currentPath` at line 101. The rollback correctly covers the second `os.Rename` *returning an error*, but nothing covers the process dying in between — Ctrl-C, kill, crash, power loss. In that window `ledgerctl.exe` does not exist and the only good binary is `ledgerctl.exe.old`: the CLI is gone from `PATH` and cannot be re-run to repair itself.

What makes it worth addressing is that the recovery path is undiscoverable. The error wrapped at replace.go:49 adds only `(you may need elevated permissions)`, and docs/ops/cli.md:4959 explains that `.old` is *left behind* without saying it is the artifact to restore. Note too that `replaceBinary` at line 47 removes the staged replacement on any `replaceExecutable` error, so after a failed rollback the new binary is discarded as well and `.old` really is the last copy standing.

Cheapest fix: put the recovery step where the user will see it — add "if an upgrade is interrupted, rename `ledgerctl.exe.old` back to `ledgerctl.exe`" to the Windows bullet in docs/ops/cli.md, and include the backup path in the error returned from this branch.

If you would rather close the window than document it, run the swap from a short-lived helper that executes after the parent exits and prunes the backup only once the new binary is in place. For a CLI self-upgrade, documenting the recovery is probably the better cost/benefit — but as written, a user who hits this has no way to find out what happened to their CLI.

paul-nicolas (commented):
Re-review of head `2832c95ca` (Claude + Codex). 2 Medium findings posted inline, both new on this head; 0 Critical, 0 Major. 6 Minor/Nit listed below rather than inline.

First, the previous round verified as genuinely fixed, not just claimed:
- `allow_different_binary_count` is gone; `ledger-unix` / `ledger-windows` are separate archive entries over separate build ids, so the binary-count guard is armed for all six archives (.goreleaser.yml:79-94).
- The exec-bit bug is fixed: `copyReplacement` closes the staging file and applies `os.Chmod(dstPath, mode)` before the swap (replace.go:74-80).
- Windows is now compiled pre-tag in the regular Build job (main.yml:204-208).
- ZIP + `.exe` handling lands correctly: `extractBinary` dispatches on the asset suffix and `executableName(runtime.GOOS)` supplies the in-archive name (download.go:135-144, github.go:190-196).
- The Windows-hostile test case is properly split — `replace_unix_test.go` carries `//go:build !windows` while the Windows-branch tests stay portable.
- The Installation section now documents the prebuilt download path and exact archive names (docs/ops/cli.md:11-19).

Worth stating plainly in the changelog: this PR does not merely rename artifacts, it repairs a pre-existing total breakage of `ledgerctl upgrade`. `projectName` at github.go:18 was already `"ledger"` while releases published `ledger-v3_*`, so `findAsset` could never match a v3 asset on any channel. The rename is what makes self-upgrade work at all.

Release-pipeline side verified clean: no `ledger-v3_` asset reference remains anywhere in the repo (the surviving `ledger-v3` hits are k8s labels, the operator justfile and the Python project name — unrelated to release assets); no workflow globs `dist/`; `release.yml` calls `just release` generically so the rename needs no workflow change; `build/` is gitignored and the Build job runs no dirty-tree check after the new cross-compile step; every build id is referenced by exactly one archive and the two `name_template`s cannot collide.

Two Codex findings were rejected on inspection rather than posted:
- "decompression bomb in tar/zip extraction" — the archive SHA256 is verified against the release `checksums.txt` *before* extraction (download.go:105-121), so contents are authenticated; anyone able to swap the asset controls the binary itself and has no need of a bomb. The tar path is also unchanged by this PR.
- "the artifact rename needs a compatibility window with duplicate legacy assets" — no in-repo consumer references the old names, the break is declared intentional, and the pre-PR state was strictly worse. Not worth the release-config complexity at alpha stage.

- cmd/ledgerctl/upgrade/download_test.go:43 -- test archives use `ledger/ledgerctl.exe`, not GoReleaser's real `wrap_in_directory` layout `ledger_windows-amd64/ledgerctl.exe`. It passes only because `isBinaryEntry` suffix-matches, so it would not catch a future prefix-sensitive change. Build the entry from `archiveAssetName(...)` so the fixture matches reality.
- cmd/ledgerctl/upgrade/github_test.go:55 -- `TestFindAssetForPlatform` covers only the happy path; the `findAssetForPlatform` error branch (release missing the Windows asset) is untested, which is the branch a botched release would actually hit.
- .goreleaser.yml:60 -- `-extldflags=-static` is inert on the Windows build with `CGO_ENABLED=0` (no external linker runs). Harmless and symmetric with the unix build, but it is dead configuration.
- .goreleaser.yml:53 -- the `ledgerctl-windows` build duplicates the `ledgerctl` block verbatim except `goos`. Two build ids are genuinely forced here (archives cannot filter by platform), but a YAML anchor for the shared `ldflags`/`env` would satisfy the DRY rule in CLAUDE.md.
- cmd/ledgerctl/upgrade/replace.go:63 -- `os.OpenFile(dstPath, os.O_WRONLY|os.O_TRUNC, 0)` passes a literal `0` perm that is silently ignored because the file already exists; the effective mode comes from the `os.Chmod` six lines down. This is the same shape as the bug that was just fixed here, so pass `mode` or add a comment to stop it reading as a regression.
- .github/workflows/main.yml:204 -- the cross-compile builds `./cmd/ledgerctl` only, which does not type-check `_test.go` files. `replace_test.go` exercises the Windows branch but its Windows-compilability is never verified; `GOOS=windows ... go vet ./cmd/ledgerctl/...` would cover that at the same cost.

cmd/ledgerctl/upgrade/replace.go
flemzord: Fixed in `22739959b` while keeping the `.old` rotation strategy. If installation and rollback both fail, the error now includes the exact backup and executable paths plus the manual rename recovery step. The Windows upgrade documentation also explains that, after an interrupted swap, `ledgerctl.exe.old` must be renamed back to `ledgerctl.exe`.

cmd/ledgerctl/upgrade/github.go
flemzord: Fixed in `22739959b`. Stable resolution now decodes GitHub's `draft` and `prerelease` fields, parses the running binary version with semver, and selects only a final release with the same major. The releases endpoint is paginated at 100 items per page until an eligible release is found or the list is exhausted, so it no longer relies on page 1. Tests cover draft/prerelease rejection and finding an eligible release on page 2. The CLI docs now state that, until the first final release for the current major exists, `stable` returns an error and `nightly` must be used.

cmd/ledgerctl/upgrade/github.go
```diff
@@ -57,25 +61,65 @@ func fetchNightlyRelease() (*releaseInfo, error) {
...
+func fetchStableRelease(currentVersion string) (*releaseInfo, error) {
+	url := fmt.Sprintf("https://api.github.com/repos/%s/releases", githubRepo)
 
-	var releases []releaseInfo
+	return fetchStableReleaseFromURL(currentVersion, url)
+}
 
-	err := githubGet(url, &releases)
+func fetchStableReleaseFromURL(currentVersion, releasesURL string) (*releaseInfo, error) {
+	current, err := semver.NewVersion(currentVersion)
```
NumaryBot: 🟠 **[major] Nightly builds cannot upgrade to stable channel**

When a user running a v3 nightly binary (where `currentVersion` is `nightly-<shortcommit>`) invokes `ledgerctl upgrade --channel stable`, the semver parse of the current version string fails before any releases are even queried, making it impossible for a nightly installation to ever migrate to a final stable release. Since nightly is the default upgrade channel and its version format has no semver major, the stable selector needs a v3 fallback path (or an alternative way to determine the product's major version) rather than hard-failing on all non-semver version strings.

**Suggestion:** Detect a non-semver `currentVersion` (e.g. one starting with `nightly-`) and treat it as a v3 build when evaluating the stable channel filter, rather than returning an error. For example, fall back to assuming major version 3 when the parse fails but the caller explicitly requested `--channel stable`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1644#issuecomment-5104179917

cmd/ledgerctl/upgrade/github.go
flemzord: Fixed in `83a2daf06` without hardcoding v3. Stable selection first derives the major from a semantic `currentVersion`; when the version is `nightly-*` or `dev`, it falls back to the major suffix embedded in the running Go module path (for this binary, `github.com/formancehq/ledger/v3`). The pagination test now starts from `nightly-deadbeef` and proves that it reaches the final v3 release on page 2, and `TestVersionMajor` covers semver, nightly fallback, and the missing-module-major error.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1644#issuecomment-5104179917

flemzord:
Revalidated live head `83a2daf0`. One correctness/runtime blocker remains in the stable selector: `findStableRelease` returns the first final release of the current major in GitHub API order, not the maximum semantic version, and `runUpgrade` installs any target whose display version is merely different from the current one. GitHub release order is not a semver ordering (this repository already publishes maintenance lines interleaved), so a later-created `v3.0.5` can precede `v3.1.0` and silently downgrade a `v3.1.0` binary through `ledgerctl upgrade --channel stable`. Please scan all eligible releases for the max semver and refuse a non-newer target unless downgrade is an explicit operation; add regressions for an unsorted/interleaved release list and a current version above the selected candidate. This is distinct from the resolved pagination/major-selection thread. The authenticated account owns this PR, so GitHub does not permit a `CHANGES_REQUESTED` review here.

paul-nicolas (approved):

PR #1645, opened 2026-07-28
fix(state): purge stranded pre-transient volume rows at zero balance

Azorlogh:
## Bug

`partitionVolumes` classified a TRANSIENT-matched volume as steady-state (never written, **never deleted**) whenever `Old`'s balance was zero. But a row persisted while the account was NORMAL can already sit at `input == output` — e.g. funded then **reverted**. The next transient touch then strands that Pebble row forever:

1. `world → staging:s 100` (no type yet → NORMAL, row persisted)
2. revert → `{in:100, out:100}`, zero balance, row legitimately kept
3. `staging:{id}` becomes TRANSIENT
4. wash batch on `staging:s` → `Old.IsDefined() && !isVolumeZeroBalance(Old)` is false → steady-state → no write, no delete

Effects:
- stale `SubAttrVolume` row → **ghost account** in ListAccounts/GetAccount (never purgeable again — every later touch reclassifies as steady-state);
- FSM cache holds `{0,0}` while Pebble holds `{100,100}` for the same applied index — resurfaces via cache rotation + preload Pebble fallback;
- diverges from every re-derivation of the store: the checker's `SimulateEphemeralPurge` (deletes at `input == output`, no prior-balance precondition), the backup rebuild, and the antithesis model oracle — so a restored store would silently differ from the live one, and `Check()` flags `VOLUME_MISMATCH`.

Found by the model-based antithesis driver's account-query validation: the server's ListAccounts window contained an account the oracle had purged (verified on disk with a read-only Pebble scan of the failing run's data dir).

## Fix

Route the transient branch on "`Old` has non-zero **limbs**" (`isVolumePreloadZero`, the same test `splitPurged` already uses) instead of non-zero **balance**. Draining behavior is unchanged (non-zero balance ⊂ non-zero limbs); the stranded case now lands in `purged` → Pebble row deleted, listed as a draining eviction in `LedgerLog.PurgedVolumes`. The post-purge `{0,0}` cache placeholder and the preloader's zero seed still classify as steady-state, so no extra writes for genuine steady-state transients.

`New` is guaranteed zero-balance in the newly-purged case: the transient end-zero gate only exempts pre-batch non-zero-balance accounts.

## Tests

- e2e (`transient_accounts_test.go`, new Context "When a zero-balance row persisted before the transient pattern matched"): fund → revert → mark transient → wash; asserts `GetAccount` has no volumes and the account no longer appears in `ListAccounts`. Red before the fix (returned the stranded `{100,100}` row), green after.
- unit (`TestPartitionVolumesTransient_PreExistingBalance`): new `staging:stranded` case — `Old {100,100}` → purged.
- Full business e2e suite green; `state`, `check`, `indexbuilder`, `usagebuilder` unit tests green.

### DIFF
--- a/docs/technical/architecture/subsystems/usage/counters.md
+++ b/docs/technical/architecture/subsystems/usage/counters.md
@@ -79,7 +79,7 @@ The disjoint encoding pays exactly `len(ephemeral)` per log instead of `2 × len
 1. `partitionVolumes(volumeUpdates)` (existing) yields `partResult.{kept, purged, transient}`.
 2. `splitPurged(partResult.purged)` (new, in `write_set_new_volumes.go`) partitions `purged` further into:
    - **ephemeral**: `!Old.IsDefined() || isVolumePreloadZero(Old.Value())` — the key had no prior state, was touched, immediately purged.
-   - **draining**: `Old.IsDefined() && !isVolumePreloadZero(Old.Value())` — had a prior non-zero balance, drained to zero, evicted.
+   - **draining**: `Old.IsDefined() && !isVolumePreloadZero(Old.Value())` — had a prior persisted value, at zero balance now, evicted.
 3. `makeNewKeptKeySet(partResult.kept)` (new) yields the subset of `kept` where `Old` was undefined / zero-placeholder — i.e. new persistent volumes.
 4. `buildTouchedByLog(volumes.Slots(), setX)` (new, generalised from the previous `buildPurgedByLog`) intersects the per-order touched-volume tracking with each of the three sets to produce the per-log annotation lists. Deduplication + deterministic sort by (account, asset) keeps the log payload byte-identical across nodes.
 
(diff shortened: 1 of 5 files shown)


flemzord:
Revalidated live head `1dbd0518`. The transient partition now distinguishes the all-zero preload/cache placeholder from a persisted zero-balance row with non-zero limbs, so the latter is routed through the normal purge path and its Pebble row, cache state, exclusion projection, and replay-derived expectation stay aligned. Functional CI, Codecov, NumaryBot, and CodeRabbit are green. I found no remaining correctness/runtime blocker. GitHub approval is intentionally disabled by this cron configuration.

paul-nicolas (dismissed):

flemzord:
Revalidated live head `d7fbb088`. The only change since the previously covered `1dbd0518` is the refactor from explicit Uint256 limb checks to `Uint256.IsZero`; that helper is nil-safe, so the all-zero preload/cache placeholder remains distinct from persisted zero-balance rows with non-zero limbs. The purge behavior and its focused regressions are unchanged, and all live checks are green. I found no remaining correctness/runtime blocker. GitHub approval is intentionally disabled by this cron configuration.

paul-nicolas (approved):

PR #1647, opened 2026-07-29
fix(EN-1379): surface coverage misses instead of wrapping them as storage faults

paul-nicolas:
## Summary

An FSM coverage-gate violation (`*state.ErrCoverageMiss` — an admission bug, reason `COVERAGE_MISS`) was wrapped in `domain.ErrStorageOperation` at every read site. `buildAuditFailure` (`internal/infra/state/audit.go:25-26`) reads `Reason()`/`Metadata()` off the **outermost** `Describable` and never unwraps, so the admission bug was permanently recorded in the **hash-chained, immutable** audit as `STORAGE_OPERATION_FAILED`, with the context stripped to `{operation: "..."}` — losing the very fields that identify *which* declaration was missing. `businessErrorToGRPCStatus` (`grpc/errors.go:86`) performs the same outermost-only read, so the client saw the relabelled reason too.

This adds two functions to `internal/domain`, routes all 39 `ErrStorageOperation` construction sites through the second, and locks the convention in with a `forbidigo` rule:

```go
func CoverageContractViolation(err error) Describable       // innermost COVERAGE_MISS / INVALID_EXECUTION_PLAN, else nil
func StoreFailure(operation string, err error) Describable  // propagates a violation verbatim, wraps everything else
```

The reason is now `COVERAGE_MISS` on both affected surfaces with the identifying metadata intact. The gRPC status code is unchanged — both reasons map to `KindInternal`. (See the upgrade note below: the status code is unchanged, but the audit **hash** is not.)

Pass-through rather than a new sentinel, per the 2026-07-17 planning note. No proto change, no `KindForReason` change, no new import edge (`internal/domain` is already imported by both `state` and `processing`).

## Not a surface: the frozen idempotency outcome

An earlier revision of this description claimed the wrap also froze the wrong reason into the idempotency outcome. That was wrong, in both directions — **a coverage miss is never frozen at all**:

- `recordIdempotencyFailure` returns early at `machine.go:1729`: `if !errors.As(bizErr, &d) || !domain.IsFreezableFailure(domain.Kind(d)) { return nil }`
- `IsFreezableFailure` (`internal/domain/errors.go:152-158`) admits only `KindValidation`, `KindNotFound`, `KindAlreadyExists`, `KindConflict`, `KindPrecondition`.
- `KindForReason` (`internal/domain/reason.go:124`) classifies `ERROR_REASON_COVERAGE_MISS` as `KindInternal`.

`ERROR_REASON_STORAGE_OPERATION_FAILED` sits in the same `KindInternal` arm, so nothing was frozen on this path before either. An internal-kind failure is deliberately left unfrozen so the request stays retryable — freezing would pin a server bug against the caller's key for the whole TTL. `coverage-gate.md` now states this positively rather than listing a surface that does not exist.

## Upgrade note: the reason is hash-bound

`buildAuditFailurePayload` (`internal/infra/state/audit_envelope.go:176-198`) folds the reason, the message and every sorted context key **and value** into the audit hash pre-image. Relabelling an FSM-emitted failure is therefore **not hash-neutral**: for the same Raft entry, a node on the old build hashes `{STORAGE_OPERATION_FAILED, "storage operation failed: loading ledger", {operation}}` while a node on the new build hashes `{COVERAGE_MISS, "preload coverage miss (…)", {attribute, canonicalHex, idHex, raftIndex}}`.

During a rolling upgrade both builds apply the same replicated log, so **a coverage miss landing inside the mixed-version window writes different hashes for the same index**, and `checker.verifyAuditHashChain` then fails on whichever node disagrees with the persisted chain. Nothing absorbs this: `HashVersion` comes from `HashGenerator.Algorithm()` and identifies the hash *algorithm* only, carrying no notion of failure-projection semantics.

Taken as a documented upgrade constraint rather than a code change, because:

- the trigger is itself an admission bug that should never fire, so the path is low-probability;
- only the read sites this PR **converts** are affected — the numscript re-resolution path already returned the bare miss, so only its message shifts;
- the hazard is **general** to any change to an FSM-emitted error's reason, message or metadata, not specific to this PR.

Making it structurally safe would mean carrying a failure-projection semantics version alongside `HashVersion` so old entries keep reproducing their original bytes. That is a deliberate design decision this PR does not take on. Recorded in `coverage-gate.md` under "Upgrade note: the reason is hash-bound".

## Why a Reason-string match rather than `errors.As`

`internal/infra/state` imports `internal/domain/processing`, so `processing` can never import `state` to name `*ErrCoverageMiss`. This is the same workaround PR #1560 introduced privately for the numscript re-resolution triage; that helper is deleted here in favour of the shared one — same `Unwrap` walk, one implementation.

The walk also descends into multi-error nodes (`errors.Join`, or `fmt.Errorf` with several `%w`), which `errors.Unwrap` cannot follow, visiting members in slice order so the result stays deterministic on the apply path. This is not covered by the `forbidigo` rule — a `Join` is a plain call, not a forbidden type reference — so handling it in the walk is what keeps the guard whole.

## Notes for review

- **`processor.go:347` is a substantive fix, not a cosmetic conversion.** `gatedAccessor.Delete` (`accessor.go:113-118`) also calls `CheckCoverage`, so a staged tombstone flushed by `orderOverlayScope.Commit()` produces a genuine coverage miss. That path was mislabelling too.
- **`ValidateTransientVolumes` is included.** Its `Describable` reaches `buildAuditFailure` directly (`machine.go:1543`), bypassing `planInvariantDescribable`. This is the post-handler path from the EN-1520 comment.
- **Technical-update handlers need no change** — they return `fmt.Errorf("...%w")` and `applyTechnicalUpdates` already extracts the violation via `planInvariantDescribable` (defined at `machine.go:1133`; the TU call site is `machine.go:1214`), which can use `errors.As` because `state` can name the type.
- **`ErrCoverageMiss.Metadata()` keys moved to camelCase** (`canonical_hex` → `canonicalHex`, etc.) — **a deliberate breaking change, not an incidental cleanup.** This map was *not* previously unreachable; two paths already surfaced the bare miss on `release/v3.0`:
  - numscript re-resolution returns the bare miss (`convertNumscriptError` flattens the chain via `errors.As`), and a `ProcessOrders` failure feeds `buildAuditFailure` — so the snake_case keys were **already landing in the hash-chained audit**;
  - `planInvariantDescribable` already returned the bare miss to the client through `BusinessError.Metadata()` and the gRPC `ErrorInfo`.

  So operator tooling, alerts or dashboards keyed on `canonical_hex` / `id_hex` / `raft_index` stop matching at this deploy, and audit entries for this failure class carry different key names either side of it. The change is kept because this was the **last snake_case `Describable` in the repo** (~90 metadata keys in `internal/domain/errors.go`, zero snake_case; the only multi-word one is `existingTransactionId`) and CLAUDE.md mandates camelCase on the wire — deferring it would pay the same forensic seam later. `Error()` is aligned to the same `raftIndex` spelling so one audit record does not carry both. The structured log keeps snake_case field names (a log, not a wire payload). Old audit entries keep their stored bytes and re-verify against them, and `compareIdempotencyOutcomes` does not diverge.

## Regression guard

`forbidigo` forbids bare `ErrStorageOperation` construction, following the existing `OpenWriteSession` precedent. Verified in both directions: clean on the converted tree, and it fires with the intended message on a deliberately-bad line.

Because `analyze-types` is on, the rule matches **any** reference resolving to the type — an `errors.As` target declaration, a field read off it, a type-switch case — not only a composite literal (a construction-only pattern is not expressible: forbidigo renders the type expression, never the brace). The message says so and points legitimate consumers at `//nolint:forbidigo` with a justification. Two scoped exclusions, both `source:`-pinned: `coverage.go` (the one legitimate construction) and `errors.go` (where the type's own method receivers resolve).

## Testing

Three layers: a hermetic table test on the discriminator (`internal/domain/coverage_test.go`), a binding proof against the real `*ErrCoverageMiss` (`internal/infra/state/scope_unit_test.go`, including two `errors.Join` cases), and handler-level tests via the generated `MockScope` covering Ledger, Boundaries and post-commit volumes. Each new test was verified to fail without its fix (`expected COVERAGE_MISS, actual STORAGE_OPERATION_FAILED`), and each conversion has an over-correction guard asserting a genuine IO fault still yields `STORAGE_OPERATION_FAILED`.

Added on review:

- `TestBuildAuditFailure/CoverageMiss` — drives an `*ErrCoverageMiss` through `buildAuditFailure` and asserts the reason, the message and the **exact** `Context` key set. This is the assertion that was missing: the key names were pinned on `Metadata()` directly but nowhere they are consumed, which is how a rename could ship green while every context key feeds the immutable hash pre-image. A sibling negative control pins what the wrapped shape would have recorded instead.
- `TestValidateTransientVolumesLedgerCoverageMissPropagates` — covers the ledger-load `StoreFailure` site, which had no coverage-miss test; the existing one only drives the `CheckCoverage` site.

`TestValidateTransientVolumesStorageFaultTakesPrecedence` was updated — it pinned the old wrapping. Its precedence assertion (fault beats business offender) is unchanged.

## Question for @gfyrag re: the EN-1520 comment

Two of that comment's three bullets appear already resolved on `release/v3.0`, and I did not act on them:

- `updateBoundaryCounters` no longer exists — the per-ledger usage counters moved to the `usagestore` peer store under EN-1334 / EN-1420, leaving only a stale reference in a comment at `write_set.go:211`.
- I found no counter-underflow clamp in `internal/infra/state` or `internal/domain/processing`.

The third — post-handler coverage errors — is addressed here via `ValidateTransientVolumes`. Please confirm nothing was missed before EN-1522 rebases on this.

## Ticket

https://formance-team.atlassian.net/browse/EN-1379

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -139,6 +139,15 @@ linters:
           violation (coverage miss / invalid execution plan) verbatim so the audit chain
           records COVERAGE_MISS, and wraps everything else as a storage fault. See
           docs/technical/architecture/subsystems/fsm/coverage-gate.md.
+
+          analyze-types is on, so this rule fires on ANY reference whose type resolves
+          to ErrStorageOperation outside domain.StoreFailure — not only a composite
+          literal. An errors.As target declaration, a field read off such a target and
+          a type-switch case are all flagged. A construction-only pattern is not
+          expressible here: forbidigo renders the type expression and never the brace.
+          If you are legitimately CONSUMING a storage fault (reading Operation for a
+          log line, distinguishing an IO fault from a business outcome), the rule has
+          no alternative to offer — add //nolint:forbidigo with a justification.
   exclusions:
     rules:
     # I2 exceptions for OpenWriteSession / NewWriteSessionFromDB.
@@ -183,9 +192,12 @@ linters:
       linters: [forbidigo]
     # coverage.go owns the single legitimate ErrStorageOperation construction:
     # domain.StoreFailure is the wrapper every other site must go through, and
-    # it is the one place allowed to build the fallback. (EN-1379)
+    # it is the one place allowed to build the fallback. Pinned by source to that
+    # one construction — symmetric with the errors.go rule below — so a second,
+    # unintended reference in this file is still caught. (EN-1379)
     - path: internal/domain/coverage\.go
       linters: [forbidigo]
+      source: 'ErrStorageOperation\{'
     # ErrStorageOperation's own methods read its fields; with analyze-types the
     # receiver resolves to the forbidden type, so e.Operation / e.Cause are
     # flagged inside the declaration itself. Pinned to those field reads via
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -126,6 +126,19 @@ linters:
           OpenWriteSession check. It is reserved for subsystems that manage their own
           Pebble instance (currently: internal/storage/readstore). New call sites must
           be added to the .golangci.yaml exclusions block with a justification.
+      # EN-1379: a bare ErrStorageOperation swallows the reason of an
+      # admission-contract violation. buildAuditFailure reads Reason()/Metadata()
+      # off the OUTERMOST Describable and never unwraps, so wrapping a
+      # *ErrCoverageMiss permanently records an admission bug in the
+      # hash-chained audit as STORAGE_OPERATION_FAILED with the identifying key
+      # stripped. domain.StoreFailure propagates the violation verbatim.
+      - pattern: '\bErrStorageOperation\b'
+        msg: |
+          Do not construct domain.ErrStorageOperation directly. Use
+          domain.StoreFailure(operation, err), which propagates an admission-contract
+          violation (coverage miss / invalid execution plan) verbatim so the audit chain
+          records COVERAGE_MISS, and wraps everything else as a storage fault. See
+          docs/technical/architecture/subsystems/fsm/coverage-gate.md.
   exclusions:
     rules:
     # I2 exceptions for OpenWriteSession / NewWriteSessionFromDB.
@@ -168,6 +181,19 @@ linters:
     # NewWriteSessionFromDB by the same rationale.
     - path: internal/storage/usagestore/
       linters: [forbidigo]
+    # coverage.go owns the single legitimate ErrStorageOperation construction:
+    # domain.StoreFailure is the wrapper every other site must go through, and
+    # it is the one place allowed to build the fallback. (EN-1379)
+    - path: internal/domain/coverage\.go
+      linters: [forbidigo]
+    # ErrStorageOperation's own methods read its fields; with analyze-types the
+    # receiver resolves to the forbidden type, so e.Operation / e.Cause are
+    # flagged inside the declaration itself. Pinned to those field reads via
+    # source so a bare &ErrStorageOperation{...} in errors.go is still caught.
+    # (EN-1379)
+    - path: internal/domain/errors\.go
+      linters: [forbidigo]
+      source: 'e\.(Operation|Cause)'
     # Tests use OpenWriteSession to seed data; the structural guarantee is on
     # production paths.
     - path: _test\.go$
--- a/docs/ops/deployment.md
+++ b/docs/ops/deployment.md
@@ -660,6 +660,21 @@ Trying to keep existing Pebble data and upgrade in-place is **unsafe**:
 
 Mixed-binary rolling upgrades are **not supported** across this change. Stop all nodes before deploying the new binary.
 
+### Upgrading across EN-1379 (coverage-miss reclassification)
+
+EN-1379 changes how the FSM labels one failure class. A preload coverage miss used to be flattened into `ERROR_REASON_STORAGE_OPERATION_FAILED`; it now surfaces under its own `ERROR_REASON_COVERAGE_MISS` with the identifying context (`attribute`, `canonicalHex`, `idHex`, `raftIndex`) intact.
+
+That relabelling is **not hash-neutral**. `buildAuditFailurePayload` folds the failure reason, its message and every sorted context key *and value* into the audit hash pre-image, so for one and the same Raft entry an old binary and a new binary compute different `AuditEntry.Hash` values. `HashVersion` identifies the hash *algorithm* only and carries no notion of failure-projection semantics, so nothing lets a node recognise an entry written under the old classification.
+
+Mixed-binary rolling upgrades are **not supported** across this change. Restart every node onto the new binary; a coverage miss applied while the cluster straddles the two builds diverges the audit chain at that index, and `ledgerctl check` then reports `HASH_MISMATCH` on whichever node disagrees with the persisted chain.
+
+Unlike the pre-#400 upgrade above, **no data wipe is required**:
+
+- No persisted key layout or value encoding changed, so `storage-schema-version` is unaffected and existing entries stay verifiable byte-for-byte.
+- The exposure is confined to entries written *inside* the mixed-binary window, and only to this one failure class. A coverage miss is itself an admission-contract violation that should never fire (invariant #7), so in a healthy cluster the window is empty.
+
+The hazard generalises: **any** change to an FSM-emitted error's reason, message or metadata is observable in the hash chain, and should carry an upgrade note like this one. Making it structurally safe would require a failure-projection semantics version carried alongside `HashVersion`, so old entries keep reproducing their original bytes — a deliberate design change that EN-1379 does not take on. See [coverage-gate.md](../technical/architecture/subsystems/fsm/coverage-gate.md#upgrade-note-the-reason-is-hash-bound) for the mechanism in detail.
+
 ### Audit hash keying — threat model
 
 The audit hash chain (`processing.HashGenerator`) is keyed by a value derived from the immutable `cluster-id`. This is **defense in depth against offline grinding from outside the cluster boundary**, not a tamper-evidence guarantee against an attacker with persisted-store access.
--- a/docs/ops/deployment.md
+++ b/docs/ops/deployment.md
@@ -660,20 +660,24 @@ Trying to keep existing Pebble data and upgrade in-place is **unsafe**:
 
 Mixed-binary rolling upgrades are **not supported** across this change. Stop all nodes before deploying the new binary.
 
-### Upgrading across EN-1379 (coverage-miss reclassification)
+### Upgrading across an FSM error-identity change
 
-EN-1379 changes how the FSM labels one failure class. A preload coverage miss used to be flattened into `ERROR_REASON_STORAGE_OPERATION_FAILED`; it now surfaces under its own `ERROR_REASON_COVERAGE_MISS` with the identifying context (`attribute`, `canonicalHex`, `idHex`, `raftIndex`) intact.
+`buildAuditFailurePayload` folds a failure's reason, its message and every sorted context key *and value* into the audit hash pre-image. So **any** change to how the FSM identifies an error — its reason, its message text, or its metadata keys and values — makes an old binary and a new binary compute different `AuditEntry.Hash` values for one and the same Raft entry. `HashVersion` identifies the hash *algorithm* only and carries no notion of failure-projection semantics, so nothing lets a node recognise an entry written under an older classification.
 
-That relabelling is **not hash-neutral**. `buildAuditFailurePayload` folds the failure reason, its message and every sorted context key *and value* into the audit hash pre-image, so for one and the same Raft entry an old binary and a new binary compute different `AuditEntry.Hash` values. `HashVersion` identifies the hash *algorithm* only and carries no notion of failure-projection semantics, so nothing lets a node recognise an entry written under the old classification.
+This is a standing rule rather than a per-change note. Whenever a release changes an FSM-emitted error's identity:
 
-Mixed-binary rolling upgrades are **not supported** across this change. Restart every node onto the new binary; a coverage miss applied while the cluster straddles the two builds diverges the audit chain at that index, and `ledgerctl check` then reports `HASH_MISMATCH` on whichever node disagrees with the persisted chain.
+- **Mixed-binary rolling upgrades are not supported across that change.** Restart every node onto the new binary. Note that the Kubernetes operator performs a *rolling* update by default (a pod-template spec-hash annotation drives it), so this constraint has to be applied deliberately — it is not the default behaviour. A failure of the affected class applied while the cluster straddles two builds diverges the audit chain at that index, and `ledgerctl check` then reports `HASH_MISMATCH` on whichever node disagrees with the persisted chain.
+- **No data wipe is required**, unlike the pre-#400 upgrade above. No persisted key layout or value encoding changes, `storage-schema-version` is unaffected, and existing entries stay verifiable byte-for-byte. The exposure is confined to entries written *inside* the mixed-binary window, and only to the affected failure class.
 
-Unlike the pre-#400 upgrade above, **no data wipe is required**:
+Changes in this release line that fall under the rule:
 
-- No persisted key layout or value encoding changed, so `storage-schema-version` is unaffected and existing entries stay verifiable byte-for-byte.
-- The exposure is confined to entries written *inside* the mixed-binary window, and only to this one failure class. A coverage miss is itself an admission-contract violation that should never fire (invariant #7), so in a healthy cluster the window is empty.
+| Change | What shifts in the hash pre-image | Exposure |
+|--------|-----------------------------------|----------|
+| EN-1558 (`e43016439`) | `marshalOrdersForAudit` hashes business-intent bytes instead of the whole order | every order |
+| EN-1536 (`8b24ee535`) | `buildAuditFailure` resolves the outermost `Describable` instead of `errors.As`-unwrapping, and takes `Message` from `d.Error()` — reason, message and context can all shift for a wrapped failure | any wrapped FSM failure |
+| EN-1379 | a preload coverage miss stops being flattened into `ERROR_REASON_STORAGE_OPERATION_FAILED` and surfaces as `ERROR_REASON_COVERAGE_MISS`, with `attribute`, `canonicalHex`, `idHex`, `raftIndex` intact | coverage misses only — themselves admission-contract violations that must never fire (invariant #7), so in a healthy cluster the window is empty |
 
-The hazard generalises: **any** change to an FSM-emitted error's reason, message or metadata is observable in the hash chain, and should carry an upgrade note like this one. Making it structurally safe would require a failure-projection semantics version carried alongside `HashVersion`, so old entries keep reproducing their original bytes — a deliberate design change that EN-1379 does not take on. See [coverage-gate.md](../technical/architecture/subsystems/fsm/coverage-gate.md#upgrade-note-the-reason-is-hash-bound) for the mechanism in detail.
+Making this structurally safe would require a failure-projection semantics version carried alongside `HashVersion`, so old entries keep reproducing their original bytes. That is a deliberate design change, tracked as EN-1661; until it lands, the rule above is the mitigation. See [coverage-gate.md](../technical/architecture/subsystems/fsm/coverage-gate.md#upgrade-note-the-reason-is-hash-bound) for the mechanism in detail.
 
 ### Audit hash keying — threat model
 
--- a/docs/technical/architecture/subsystems/fsm/coverage-gate.md
+++ b/docs/technical/architecture/subsystems/fsm/coverage-gate.md
@@ -83,11 +83,28 @@ A coverage miss is an admission bug, not an infrastructure fault, and it is labe
 |---------|-------|
 | Hash-chained `AuditFailure.Reason` | `ERROR_REASON_COVERAGE_MISS` |
 | `AuditFailure.Context` | `attribute`, `canonicalHex`, `idHex`, `raftIndex` |
-| Frozen idempotency outcome | the same reason and metadata |
+| Frozen idempotency outcome | **not applicable** — `KindInternal` failures are deliberately never frozen (`IsFreezableFailure`, `internal/domain/errors.go:152-158`), so a coverage miss leaves no idempotency record and the request stays retryable |
 | gRPC / HTTP `ErrorInfo.reason` | `COVERAGE_MISS` (status `INTERNAL` — both reasons are `KindInternal`) |
 | Node-local log + OTel counter | `scope.go:374-378` (see [Metrics](#metrics)) |
 
 The `Metadata()` keys are camelCase, matching every other `Describable` and the repo-wide wire convention. The structured log emitted alongside keeps snake_case field names — that is a log, not a wire payload.
+
+The idempotency row is worth stating explicitly because the non-freezing is a design point, not an oversight: `recordIdempotencyFailure` returns early (`machine.go:1729`) for any non-freezable kind, and `KindForReason` classifies `ERROR_REASON_COVERAGE_MISS` as `KindInternal` (`internal/domain/reason.go:124`). Freezing would pin a server bug against the caller's key for the whole TTL. This was equally true before EN-1379 — `ERROR_REASON_STORAGE_OPERATION_FAILED` sits in the same `KindInternal` arm — so the idempotency outcome is not one of the surfaces the fix repairs.
+
+### Upgrade note: the reason is hash-bound
+
+`buildAuditFailurePayload` (`internal/infra/state/audit_envelope.go:176-198`) folds the reason, the message and every sorted context key **and value** into the audit hash pre-image. Relabelling an FSM-emitted failure is therefore *not* hash-neutral: for one and the same Raft entry, a node on the old build hashes `{STORAGE_OPERATION_FAILED, "storage operation failed: loading ledger", {operation}}` while a node on the new build hashes `{COVERAGE_MISS, "preload coverage miss (…)", {attribute, canonicalHex, idHex, raftIndex}}`.
+
+During a rolling upgrade both builds apply the same replicated log, so a coverage miss landing inside the mixed-version window writes **different hashes for the same index** — invariant #2 across the version boundary, after which `checker.verifyAuditHashChain` fails on whichever node disagrees with the persisted chain. Nothing absorbs this: `HashVersion` comes from `HashGenerator.Algorithm()` (`machine.go:1427`) and identifies the hash *algorithm* only, carrying no notion of failure-projection semantics, so a node cannot recognise "this entry was produced under the old classification".
+
+Consequences to hold in mind:
+
+- Coverage misses are correctly and consistently classified only once **every** node runs the new build. A miss inside the window can diverge the chain.
+- The window is narrow in practice: the trigger is itself an admission bug that should never fire. This is a low-probability path, not a routine one.
+- Only the read sites EN-1379 **converted** are affected. The numscript re-resolution path already returned the bare miss, so its reason and context are unchanged; only its message shifts, and only because `Error()` was aligned to the camelCase `raftIndex` spelling.
+- **This hazard is general.** It attaches to any change to an FSM-emitted error's reason, message or metadata — not to EN-1379 specifically. Treat "relabelling an FSM error is observable in the hash chain" as the standing rule when reviewing such a change.
+
+Making this structurally safe would mean carrying a failure-projection semantics version alongside `HashVersion` so old entries keep reproducing their original bytes. That is a deliberate design decision, not something EN-1379 took on.
 
 This holds because FSM read sites build their failure through `domain.StoreFailure(operation, err)` instead of constructing a `domain.ErrStorageOperation` directly:
 
@@ -109,9 +126,9 @@ failure.Reason = domain.ReasonCode(d.Reason())
 maps.Copy(failure.GetContext(), d.Metadata())
 ```
 
-So wrapping a `*ErrCoverageMiss` would permanently record an admission bug as `STORAGE_OPERATION_FAILED`, with the context stripped to `{operation: "..."}` — in a chain that is immutable by construction. The same outermost-only read happens in `recordIdempotencyFailure` (`machine.go:1746-1748`), which freezes the outcome for the idempotency TTL, and in `businessErrorToGRPCStatus` (`internal/adapter/grpc/errors.go:86`).
+So wrapping a `*ErrCoverageMiss` would permanently record an admission bug as `STORAGE_OPERATION_FAILED`, with the context stripped to `{operation: "..."}` — in a chain that is immutable by construction. `businessErrorToGRPCStatus` (`internal/adapter/grpc/errors.go:86`) performs the same outermost-only read, so the client sees the relabelled reason too. (`recordIdempotencyFailure` reads the outermost `Describable` the same way, but never reaches a coverage miss at all — see the idempotency row above.)
 
-`CoverageContractViolation` matches on the stable domain `Reason()` string rather than the concrete type, because `internal/domain/processing` cannot import `internal/infra/state` — `state` imports `processing`, so a type-based `errors.As` would be an import cycle. It recognises both `COVERAGE_MISS` and `INVALID_EXECUTION_PLAN`, walking the `Unwrap` chain so a violation nested behind a numscript `QueryBalanceError` is still found. (It follows single-error `Unwrap` only; an `errors.Join` hides its members.)
+`CoverageContractViolation` matches on the stable domain `Reason()` string rather than the concrete type, because `internal/domain/processing` cannot import `internal/infra/state` — `state` imports `processing`, so a type-based `errors.As` would be an import cycle. It recognises both `COVERAGE_MISS` and `INVALID_EXECUTION_PLAN`, walking the `Unwrap` chain so a violation nested behind a numscript `QueryBalanceError` is still found. It also descends into multi-error nodes (`errors.Join`, or `fmt.Errorf` with several `%w`), which `errors.Unwrap` cannot follow, visiting members in slice order so the result stays deterministic as the apply path requires. The `forbidigo` rule below cannot see a future `Join` — it is a plain call, not a forbidden type reference — so handling it in the walk is what keeps the guard whole.
 
 A `forbidigo` rule in `.golangci.yaml` forbids bare `ErrStorageOperation` construction, so a new read site cannot silently reintroduce the flattening.
 
--- a/docs/technical/architecture/subsystems/fsm/coverage-gate.md
+++ b/docs/technical/architecture/subsystems/fsm/coverage-gate.md
@@ -75,6 +75,50 @@ The result is a value the apply path was never authorized to consult. Two failur
 
 Both are catastrophic for an FSM that is supposed to be a pure function of its declared inputs.
 
+## How a violation surfaces
+
+A coverage miss is an admission bug, not an infrastructure fault, and it is labelled that way on every surface it reaches:
+
+| Surface | Value |
+|---------|-------|
+| Hash-chained `AuditFailure.Reason` | `ERROR_REASON_COVERAGE_MISS` |
+| `AuditFailure.Context` | `attribute`, `canonicalHex`, `idHex`, `raftIndex` |
+| Frozen idempotency outcome | the same reason and metadata |
+| gRPC / HTTP `ErrorInfo.reason` | `COVERAGE_MISS` (status `INTERNAL` — both reasons are `KindInternal`) |
+| Node-local log + OTel counter | `scope.go:374-378` (see [Metrics](#metrics)) |
+
+The `Metadata()` keys are camelCase, matching every other `Describable` and the repo-wide wire convention. The structured log emitted alongside keeps snake_case field names — that is a log, not a wire payload.
+
+This holds because FSM read sites build their failure through `domain.StoreFailure(operation, err)` instead of constructing a `domain.ErrStorageOperation` directly:
+
+```go
+// internal/domain/coverage.go
+func StoreFailure(operation string, err error) Describable {
+	if violation := CoverageContractViolation(err); violation != nil {
+		return violation // propagated verbatim — reason and metadata intact
+	}
+
+	return &ErrStorageOperation{Operation: operation, Cause: err}
+}
+```
+
+The distinction is load-bearing. `buildAuditFailure` (`internal/infra/state/audit.go:19-29`) reads `Reason()` and `Metadata()` off the **outermost** `Describable` and never unwraps:
+
+```go
+failure.Reason = domain.ReasonCode(d.Reason())
+maps.Copy(failure.GetContext(), d.Metadata())
+```
+
+So wrapping a `*ErrCoverageMiss` would permanently record an admission bug as `STORAGE_OPERATION_FAILED`, with the context stripped to `{operation: "..."}` — in a chain that is immutable by construction. The same outermost-only read happens in `recordIdempotencyFailure` (`machine.go:1746-1748`), which freezes the outcome for the idempotency TTL, and in `businessErrorToGRPCStatus` (`internal/adapter/grpc/errors.go:86`).
+
+`CoverageContractViolation` matches on the stable domain `Reason()` string rather than the concrete type, because `internal/domain/processing` cannot import `internal/infra/state` — `state` imports `processing`, so a type-based `errors.As` would be an import cycle. It recognises both `COVERAGE_MISS` and `INVALID_EXECUTION_PLAN`, walking the `Unwrap` chain so a violation nested behind a numscript `QueryBalanceError` is still found. (It follows single-error `Unwrap` only; an `errors.Join` hides its members.)
+
+A `forbidigo` rule in `.golangci.yaml` forbids bare `ErrStorageOperation` construction, so a new read site cannot silently reintroduce the flattening.
+
+Technical-update handlers take a different route to the same place: they return plain `fmt.Errorf("...: %w", err)` and `applyTechnicalUpdates` extracts the violation via `planInvariantDescribable` (`machine.go:1133`), which uses `errors.As` because the `state` package *can* name the concrete type. Both routes end at the same reason code.
+
+Note that a violation is not only produced by reads. `gatedAccessor.Delete` (`internal/infra/state/accessor.go:113-118`) also calls `CheckCoverage`, so a staged tombstone flushed by `orderOverlayScope.Commit()` surfaces a coverage miss too — which is why `ProcessOrders` routes that error through `StoreFailure` rather than dropping it (`internal/domain/processing/processor.go:341-350`).
+
 ## The cascade-on-delete edge case
 
 Some operations naturally want to scan: "delete every metadata row attached to this account", "purge every volume belonging to a deleted ledger", etc. These are the cases where a naive implementation reaches for `Registry.X.KeyStore().M` — and where the rule has historically been challenged.
--- a/docs/technical/architecture/subsystems/fsm/coverage-gate.md
+++ b/docs/technical/architecture/subsystems/fsm/coverage-gate.md
@@ -106,7 +106,7 @@ Consequences to hold in mind:
 
 Making this structurally safe would mean carrying a failure-projection semantics version alongside `HashVersion` so old entries keep reproducing their original bytes. That is a deliberate design decision, not something EN-1379 took on.
 
-The operational consequence is declared where operators look for supported upgrade paths: [deployment.md — Upgrading across EN-1379](../../../../ops/deployment.md#upgrading-across-en-1379-coverage-miss-reclassification) states that mixed-binary rolling upgrades are not supported across this change, and that no data wipe is required because no persisted layout changed.
+The operational consequence is declared where operators look for supported upgrade paths: [deployment.md — Upgrading across an FSM error-identity change](../../../../ops/deployment.md#upgrading-across-an-fsm-error-identity-change) states the standing rule — mixed-binary rolling upgrades are not supported across any such change, and no data wipe is required because no persisted layout changes — and lists EN-1379 alongside the earlier changes it covers.
 
 This holds because FSM read sites build their failure through `domain.StoreFailure(operation, err)` instead of constructing a `domain.ErrStorageOperation` directly:
 
--- a/docs/technical/architecture/subsystems/fsm/coverage-gate.md
+++ b/docs/technical/architecture/subsystems/fsm/coverage-gate.md
@@ -106,6 +106,8 @@ Consequences to hold in mind:
 
 Making this structurally safe would mean carrying a failure-projection semantics version alongside `HashVersion` so old entries keep reproducing their original bytes. That is a deliberate design decision, not something EN-1379 took on.
 
+The operational consequence is declared where operators look for supported upgrade paths: [deployment.md — Upgrading across EN-1379](../../../../ops/deployment.md#upgrading-across-en-1379-coverage-miss-reclassification) states that mixed-binary rolling upgrades are not supported across this change, and that no data wipe is required because no persisted layout changed.
+
 This holds because FSM read sites build their failure through `domain.StoreFailure(operation, err)` instead of constructing a `domain.ErrStorageOperation` directly:
 
 ```go
--- a/internal/application/admission/admission.go
+++ b/internal/application/admission/admission.go
@@ -1122,8 +1122,9 @@ func wrapSystemScoped(order *raftcmdpb.Order, ss *raftcmdpb.SystemScopedOrder) {
 // `Scope.GetVolume` returns `domain.ErrNotFound` and callers treat it
 // as a fresh zero balance (see `processing.readVolumeOrZero`). A
 // `*state.ErrCoverageMiss` (admission contract violation — need never
-// declared) stays distinct and propagates loud through
-// `ErrStorageOperation{Cause: covErr}`.
+// declared) stays distinct and propagates verbatim through
+// `domain.StoreFailure`, so the audit chain records `COVERAGE_MISS`
+// rather than a storage fault (EN-1379).
 func addVolumeNeed(p *plan.Coverage, ledgerName, account, asset, color string) {
 	p.Add(dal.SubAttrVolume, domain.VolumeKey{
 		AccountKey: domain.AccountKey{LedgerName: ledgerName, Account: account},
--- a/internal/domain/coverage.go
+++ b/internal/domain/coverage.go
@@ -8,8 +8,11 @@ import "errors"
 //
 // Both mean the FSM apply path read a key the proposer never declared, or was
 // handed a plan that cannot be applied: "should not happen" bugs (invariant #7)
-// that must reach the audit chain, the frozen idempotency outcome and the client
-// under their own reason, never relabelled as a storage fault.
+// that must reach the audit chain and the client under their own reason, never
+// relabelled as a storage fault. They never reach a frozen idempotency outcome:
+// KindForReason maps both reasons to KindInternal, which IsFreezableFailure
+// excludes, so recordIdempotencyFailure returns early — see
+// docs/technical/architecture/subsystems/fsm/coverage-gate.md.
 //
 // It matches on the stable domain Reason string rather than the concrete type so
 // callers need not import internal/infra/state, which owns *ErrCoverageMiss and
--- a/internal/domain/coverage.go
+++ b/internal/domain/coverage.go
@@ -18,18 +18,33 @@ import "errors"
 // wrappers — ErrStorageOperation, or the numscript library's QueryBalanceError /
 // QueryMetadataError — implement Unwrap.
 //
-// Limitation: the walk follows single-error Unwrap only. An error joined with
-// errors.Join hides its members from this check.
+// The walk also descends into multi-error nodes (errors.Join, or fmt.Errorf with
+// several %w verbs), which errors.Unwrap cannot follow. Members are visited in
+// slice order, so the Describable returned for a given tree is deterministic —
+// a requirement on the FSM apply path (invariant #2). Without this the forbidigo
+// rule guarding StoreFailure would not help: a future Join is a plain call the
+// linter cannot see, and it would silently relabel the violation as a storage
+// fault in the immutable audit chain.
 func CoverageContractViolation(err error) Describable {
 	for e := err; e != nil; e = errors.Unwrap(e) {
-		describable, ok := e.(Describable) //nolint:errorlint // deliberate per-node check; the loop unwraps the chain itself
+		if describable, ok := e.(Describable); ok { //nolint:errorlint // deliberate per-node check; the loop walks the chain itself
+			switch describable.Reason() {
+			case ErrReasonCoverageMiss, ErrReasonInvalidExecutionPlan:
+				return describable
+			}
+		}
+
+		// A multi-error node reports no single cause, so errors.Unwrap returns
+		// nil and this iteration is the last: recurse before the loop ends.
+		joined, ok := e.(interface{ Unwrap() []error })
 		if !ok {
 			continue
 		}
 
-		switch describable.Reason() {
-		case ErrReasonCoverageMiss, ErrReasonInvalidExecutionPlan:
-			return describable
+		for _, member := range joined.Unwrap() {
+			if violation := CoverageContractViolation(member); violation != nil {
+				return violation
+			}
 		}
 	}
 
--- a/internal/domain/coverage.go
+++ b/internal/domain/coverage.go
@@ -1,0 +1,57 @@
+package domain
+
+import "errors"
+
+// CoverageContractViolation returns the admission-contract violation carried by
+// err — a coverage miss (ErrReasonCoverageMiss) or a structurally inconsistent
+// execution plan (ErrReasonInvalidExecutionPlan) — or nil when err is neither.
+//
+// Both mean the FSM apply path read a key the proposer never declared, or was
+// handed a plan that cannot be applied: "should not happen" bugs (invariant #7)
+// that must reach the audit chain, the frozen idempotency outcome and the client
+// under their own reason, never relabelled as a storage fault.
+//
+// It matches on the stable domain Reason string rather than the concrete type so
+// callers need not import internal/infra/state, which owns *ErrCoverageMiss and
+// itself imports internal/domain/processing (an import cycle in the other
+// direction). Every Describable in the chain is inspected because intermediate
+// wrappers — ErrStorageOperation, or the numscript library's QueryBalanceError /
+// QueryMetadataError — implement Unwrap.
+//
+// Limitation: the walk follows single-error Unwrap only. An error joined with
+// errors.Join hides its members from this check.
+func CoverageContractViolation(err error) Describable {
+	for e := err; e != nil; e = errors.Unwrap(e) {
+		describable, ok := e.(Describable) //nolint:errorlint // deliberate per-node check; the loop unwraps the chain itself
+		if !ok {
+			continue
+		}
+
+		switch describable.Reason() {
+		case ErrReasonCoverageMiss, ErrReasonInvalidExecutionPlan:
+			return describable
+		}
+	}
+
+	return nil
+}
+
+// StoreFailure returns the Describable for a failed store operation.
+//
+// An admission-contract violation is propagated verbatim so its reason and
+// metadata survive to the audit chain; anything else is wrapped as
+// ErrStorageOperation under the given operation label. Every FSM read site must
+// build its failure through this function rather than constructing an
+// ErrStorageOperation directly — a bare construction swallows the coverage
+// reason, which is enforced by a forbidigo rule in .golangci.yaml.
+//
+// operation is the short identifier ErrStorageOperation surfaces ("loading
+// ledger", "checking transaction reference"). err is expected to be non-nil;
+// calling with nil yields an ErrStorageOperation with a nil cause.
+func StoreFailure(operation string, err error) Describable {
+	if violation := CoverageContractViolation(err); violation != nil {
+		return violation
+	}
+
+	return &ErrStorageOperation{Operation: operation, Cause: err}
+}
--- a/internal/domain/coverage_test.go
+++ b/internal/domain/coverage_test.go
@@ -34,6 +34,18 @@ func TestCoverageContractViolation(t *testing.T) {
 		{name: "nil", err: nil, want: nil},
 		{name: "plain error", err: errors.New("disk on fire"), want: nil},
 		{name: "unrelated describable", err: other, want: nil},
+		{
+			// A genuine stale-inputs error is retryable, not an admission bug —
+			// the numscript re-resolution triage depends on the two staying apart.
+			name: "stale inputs resolution sentinel",
+			err:  ErrStaleInputsResolution,
+			want: nil,
+		},
+		{
+			name: "invalid execution plan sentinel type",
+			err:  &ErrInvalidExecutionPlan{Reason_: "undeclared key"},
+			want: &ErrInvalidExecutionPlan{Reason_: "undeclared key"},
+		},
 		{name: "bare coverage miss", err: miss, want: miss},
 		{name: "bare invalid plan", err: plan, want: plan},
 		{
--- a/internal/domain/coverage_test.go
+++ b/internal/domain/coverage_test.go
@@ -63,6 +63,50 @@ func TestCoverageContractViolation(t *testing.T) {
 			err:  &ErrStorageOperation{Operation: "loading ledger", Cause: errors.New("pebble: closed")},
 			want: nil,
 		},
+		// The multi-error cases below are the branch the forbidigo rule cannot
+		// guard: a future errors.Join is a plain call the linter cannot see, so
+		// only these rows stand between a regressed descent and a contract
+		// violation silently relabelled as a storage fault in the audit chain.
+		{
+			name: "coverage miss joined with an unrelated error",
+			err:  errors.Join(errors.New("pebble: closed"), miss),
+			want: miss,
+		},
+		{
+			name: "coverage miss joined behind a wrap",
+			err:  fmt.Errorf("commit: %w", errors.Join(miss, errors.New("flush failed"))),
+			want: miss,
+		},
+		{
+			name: "multi-%w wrap carrying a coverage miss",
+			err:  fmt.Errorf("%w: %w", errors.New("outer"), miss),
+			want: miss,
+		},
+		{
+			name: "join with no violation",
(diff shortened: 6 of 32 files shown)


flemzord:
Revalidated live head `6a0daa70`. The new `StoreFailure` path preserves coverage/plan-contract violations and their identifying metadata through the audit/idempotency/client surfaces, while genuine store failures remain classified as `STORAGE_OPERATION_FAILED`; the converted read/delete sites consistently use that path. Functional CI is green. `codecov/patch` is below threshold (40.68%, while project coverage increased), which is CI/tooling rather than a demonstrated runtime blocker. CodeRabbit skipped substantive review because the PR is a draft, and NumaryBot has not reported a finding. I found no correctness/runtime blocker on this head. GitHub approval is intentionally disabled by this cron configuration.

internal/infra/state/scope.go
```diff
@@ -44,12 +44,18 @@ func (e *ErrCoverageMiss) Error() string {
...
+// AuditFailure context verbatim. The structured log emitted by coverageMiss
+// keeps its own snake_case field names — that is a log, not a wire payload.
 func (e *ErrCoverageMiss) Metadata() map[string]string {
 	return map[string]string{
-		"attribute":     e.Attribute,
-		"canonical_hex": e.CanonicalHex,
-		"id_hex":        e.IDHex,
-		"raft_index":    strconv.FormatUint(e.RaftIndex, 10),
+		"attribute":    e.Attribute,
+		"canonicalHex": e.CanonicalHex,
```
paul-nicolas, line 75: [Major] The camelCase rename is an observable change on paths that already exposed this map — the "previously unreachable" justification does not hold

The PR description states this map "was previously unreachable — always swallowed by the wrap, whose `Metadata()` returns only `{operation}` — so nothing could have depended on it," and the new comment above repeats it. Two pre-existing paths on `release/v3.0` contradict that, and they differ in how far they reach — worth separating, because a reviewer could otherwise dismiss this as audit-only or client-only:

**1. Numscript re-resolution — reaches the hash-chained audit.** The `isCoverageContractViolation(resolveErr) -> return nil, resolveErr` branch pre-dates this PR (line 97 on `release/v3.0`), and `convertNumscriptError` (`numscript/errors.go:70-73`) flattens the chain to the bare `*ErrCoverageMiss` via `errors.As`. That propagates through `ProcessOrders`, and a `ProcessOrders` failure writes an audit entry at `machine.go:1489-1494` via `buildAuditFailure`, which copies `d.Metadata()` straight into `AuditFailure.Context` (`audit.go:26`). The pre-existing test `internal/infra/state/coverage_miss_numscript_test.go:68` — untouched by this PR — already asserts `domain.ErrReasonCoverageMiss` on exactly this path, confirming the bare error was surfacing. So the snake_case keys were already landing in the immutable chain.

**2. Technical-update / preload — reaches the client but not the audit.** `planInvariantDescribable` (`machine.go:1133`) has always returned the bare `*ErrCoverageMiss`, and `applyProposal` surfaces it as `&domain.BusinessError{Err: invariant}` (`machine.go:1184`, `machine.go:1214`). `BusinessError.Metadata()` delegates through (`internal/domain/errors.go:259`), so it reached the gRPC `ErrorInfo` and the frozen idempotency metadata (`machine.go:1746-1748`). It does *not* extend the audit chain — the comment at `machine.go:1216-1224` says so explicitly.

Why it matters: this is not a correctness or integrity break. Old audit entries keep their stored bytes and still re-verify, and I confirmed the checker does not diverge — a pre-upgrade idempotency outcome and the audit entry that wrote it are both snake_case, so `compareIdempotencyOutcomes` stays consistent. The cost is that a live, client- and audit-visible key rename ships as an incidental cleanup justified by a premise that is false: operator tooling, alerts or dashboards keyed on `canonical_hex` / `id_hex` / `raft_index` silently stop matching, and audit entries for the same failure class now carry different key names either side of the deploy — an avoidable forensic seam in a chain that is immutable by construction.

Fix: either keep the snake_case keys, or keep the rename and make it a deliberate, announced decision — drop the "previously unreachable" rationale from the description and this comment, and note the metadata rename as a breaking change in the changelog. If you keep it, please also align `Error()` at line 42, which still formats `raft_index=%d` and feeds `AuditFailure.Message`, so one audit record does not carry both spellings.

.golangci.yaml
```diff
@@ -126,6 +126,19 @@ linters:
           OpenWriteSession check. It is reserved for subsystems that manage their own
           Pebble instance (currently: internal/storage/readstore). New call sites must
           be added to the .golangci.yaml exclusions block with a justification.
+      # EN-1379: a bare ErrStorageOperation swallows the reason of an
+      # admission-contract violation. buildAuditFailure reads Reason()/Metadata()
+      # off the OUTERMOST Describable and never unwraps, so wrapping a
+      # *ErrCoverageMiss permanently records an admission bug in the
+      # hash-chained audit as STORAGE_OPERATION_FAILED with the identifying key
+      # stripped. domain.StoreFailure propagates the violation verbatim.
+      - pattern: '\bErrStorageOperation\b'
```
paul-nicolas, line 135: [Medium] The pattern forbids every reference to the type, not just construction, while the message says only "Do not construct"

`\bErrStorageOperation\b` with `analyze-types` matches any expression whose type resolves to `ErrStorageOperation`, not only composite-literal construction. Verified empirically with a throwaway file — four diagnostics, only the first of which is intended:

```
:11  return &domain.ErrStorageOperation{...}      <- intended
:16  var target *domain.ErrStorageOperation       <- errors.As target declaration
:18  return target.Operation                      <- "use of `target.Operation` forbidden"
:27  case *domain.ErrStorageOperation:            <- type switch
```

The last three are all *consumption*: inspecting a storage fault to read `Operation` for a log line, or distinguishing an IO fault from a business outcome in a type switch. None of them swallows a coverage reason, yet each is blocked by a message whose only advice — "Use domain.StoreFailure(operation, err)" — is meaningless for a consumer and offers no way forward.

Nothing is broken today: I grepped and every remaining non-construction reference in production is a comment, and `_test.go` is already excluded. The cost is a future misfire with actively misleading guidance. Note also that a construction-only pattern is not expressible here — forbidigo renders the type expression `domain.ErrStorageOperation` and never the `{`.

Fix: extend the `msg` to say that *any* reference outside `domain.StoreFailure` trips the rule and that legitimate `errors.As` / type-switch consumers should add `//nolint:forbidigo` with a justification. If you want the guard to be structural rather than advisory, the stronger option is to unexport the struct as `errStorageOperation` and export only `StoreFailure` — the compiler then enforces it and this rule plus both its exclusions can be deleted outright. That is a larger change (cross-package test assertions reference the exported type today), so it is worth weighing rather than doing here.

internal/infra/state/scope_unit_test.go
```diff
@@ -33,15 +33,76 @@ func TestErrCoverageMiss_Describable(t *testing.T) {
...
+			require.Equal(t, "0102", md["idHex"])
+			require.Equal(t, "42", md["raftIndex"])
+		})
+	}
+}
+
+// TestGenuineStoreFaultStillWraps guards the other direction: a real IO failure
+// must keep the storage label, so the EN-1379 branch cannot silently
+// reclassify infrastructure faults as admission bugs.
+func TestGenuineStoreFaultStillWraps(t *testing.T) {
```
paul-nicolas, line 99: [Medium] Nothing pins the `AuditFailure` reason or context keys, which is the assertion that would have caught the metadata rename

This test is the right idea, and binding the Reason-string contract against the real `*ErrCoverageMiss` is exactly the seam a string-matched discriminator needs. But the PR's thesis is about what lands on three specific surfaces, and none is asserted:

- **Hash-chained `AuditFailure.Reason` / `Context`**: no test. `audit_test.go:266` exercises `buildAuditFailure` generically; nothing pins the coverage-miss case or its key set.
- **Frozen idempotency outcome**: no test. The `IdempotencyFailure` fixtures under `internal/application/check/` are all `INSUFFICIENT_FUNDS`.
- **gRPC `ErrorInfo.reason`**: no test — `grep COVERAGE_MISS internal/adapter --include=*_test.go` is empty.

The helper-level coverage is genuinely good at both ends, and the surfaces do mechanically read `d.Reason()` / `d.Metadata()` off whatever `Describable` they are handed, so the composition is tight. The gap is specifically the **key names**: this test asserts them on `Metadata()` directly, but nothing asserts them where they are actually consumed. That is why the rename discussed on `scope.go` could be justified with a false premise and still show a fully green suite.

The audit surface is the one that is immutable once written and the one invariant #8 designates the single source of truth, so the case for pinning it is stronger there than anywhere else.

Fix: add one test that drives an `*ErrCoverageMiss` through `buildAuditFailure` and asserts `Reason == ERROR_REASON_COVERAGE_MISS` plus the exact `Context` key set. That single test pins the reason, the key names and the no-wrap contract together, and it is cheap — `ErrCoverageMiss` is directly constructible, as this test already shows.

internal/domain/processing/processor_volumes.go
```diff
@@ -17,14 +17,15 @@ import (
...
-// is impossible by design under a correct preload — is propagated as an
-// ErrStorageOperation so the order is rejected loudly (invariant #7) rather than
-// returned to the client as a silently truncated volume map (EN-1440). Two
-// nodes must not emit divergent PCV payloads for the same applied index, so a
-// non-NotFound store error is always surfaced. This mirrors applyPosting, which
-// reads the same source+destination keys.
+// is reported as a zero balance, while an admission-contract violation — notably
+// *state.ErrCoverageMiss (invariants #6/#9), impossible by design under a correct
+// preload — propagates verbatim so the audit chain records COVERAGE_MISS rather
+// than a storage fault (EN-1379). Any other error is wrapped as ErrStorageOperation.
```
paul-nicolas, line 23: [Medium] A sibling comment on the invariant-#6 preload site still documents the removed flattening (cross-file: `internal/application/admission/admission.go:1125-1126`)

Anchoring here because this hunk is one of the comments the PR *did* update correctly. The same sweep updated five of them — `processor_volumes.go:19-24`, `processor_posting.go:102-105`, `processor_prepared_query.go:12-15`, `scope_helpers.go:12-16`, `processor_transaction_numscript.go:81-83` — but missed one outside `internal/domain/processing/`.

`internal/application/admission/admission.go:1125-1126`, in `addVolumeNeed`, still reads:

> `*state.ErrCoverageMiss` (admission contract violation — need never declared) stays distinct and propagates loud through `ErrStorageOperation{Cause: covErr}`

After this PR that is precisely what no longer happens — the violation propagates verbatim and is never wrapped. The location makes it worse than an ordinary stale comment: `addVolumeNeed` is the volume-preload declaration site, and CLAUDE.md calls volume preload "load-bearing" for the `new_kept_volumes` / `ephemeral_volumes` split under invariant #6. A reader auditing that invariant would come away believing the flattening is still in place, which is the exact misconception this PR exists to remove. It is also the one comment most likely to be consulted when someone next reasons about coverage misses on the volume path.

CLAUDE.md's Documentation Maintenance rule asks for code comments to be updated when behavior changes, so this is in scope for the PR rather than a follow-up.

Fix: reword to something like "…stays distinct and propagates verbatim through `domain.StoreFailure`, so the audit chain records `COVERAGE_MISS` (EN-1379)."

paul-nicolas (commented):
Multi-model review (Claude + Codex). Verified against the PR head worktree: `GOROOT= go build ./...` clean; `go test ./internal/domain/... ./internal/infra/state/...` all pass; `golangci-lint run ./internal/domain/... ./internal/infra/state/...` reports 0 issues; and the new forbidigo rule confirmed firing on a deliberately-bad construction.

The core mechanism is sound and the conversion is complete and faithful. Things I checked and can confirm as correct, so they need no attention:

- **Conversion completeness.** The only bare `ErrStorageOperation` construction left in production code is the intended one at `internal/domain/coverage.go:56`. All 39 `Operation` labels are byte-identical pre/post, including the five `fmt.Sprintf` variants; no `Cause` was swapped or dropped; the shadowed-variable sites correctly retain `findErr` / `existsErr` / `curErr`. Every call stays inside its original `if err != nil` guard, so `StoreFailure`'s documented nil-cause branch is unreachable.
- **`processor.go:349` is not a behavior change beyond scope.** `ProcessOrders` is declared `(*OrdersResult, domain.Describable)` at `processor.go:209`, so the `overlay.Commit()` failure was already a recorded business failure pre-PR, never raw/fatal. For a non-coverage-miss cause, `StoreFailure` returns a byte-identical `ErrStorageOperation`. Same for both `write_set.go` sites.
- **Invariant #8 holds.** `KindForReason` already maps both new reasons (`internal/domain/reason.go:123-124`), so `compareIdempotencyOutcomes` can still re-derive the kind from the chain-bound reason without it being persisted. The reclassification is also status-code-neutral — both reasons sit in the same `KindInternal` arm.
- **`produce()` returning `resolveErr` verbatim is safe.** `convertNumscriptError` (`internal/domain/processing/numscript/errors.go:70-73`) already flattens the chain to the innermost `Describable` via `errors.As`, so `resolveErr` *is* the bare `*ErrCoverageMiss`. (This was my initial suspicion for a Major finding; it is refuted.)
- **No openapi.yml change needed** — it does not enumerate reason strings as an enum.
- **Doc line references all check out**: `audit.go:19-29`, `machine.go:1746-1748`, `grpc/errors.go:86`, `machine.go:1133`, `accessor.go:113-118`, `scope.go:374-378` each point at what the prose claims.

**1 Major, 3 Medium posted inline.** The Major one contradicts a load-bearing claim in the PR description; it is worth reading first.

Verification: 8 findings entered adversarial verification, 4 confirmed (posted), 3 downgraded to Minor after checking real-world reachability, 1 rejected (the `produce()` one above). One reviewer claim was itself partly wrong and corrected during verification — see the Major comment.

- `internal/domain/coverage.go:24` -- the walk follows single-error `Unwrap` only, so `errors.Join` hides a member violation. Documented as a known limitation, and currently unreachable: there is no `errors.Join` anywhere in production code under `internal/domain`, `internal/infra/state` or `internal/application`. That is why this is Minor and not Medium. Still cheap insurance, since the forbidigo rule cannot catch a future `Join`.
- `.golangci.yaml:184` -- the whole-file forbidigo exclusion for `coverage.go` is broader than the one construction it exists for, and asymmetric with the correctly `source:`-pinned rule for `errors.go` right below it.
- `.golangci.yaml:198` -- the pre-existing `_test.go$` forbidigo exclusion means a future test can re-pin the flattened shape with no lint signal. Production risk is nil, but this PR had to fix exactly such a test (`write_set_test.go:958`), so the regression path is demonstrated rather than hypothetical.
- `internal/domain/processing/processor_transaction_numscript.go:98` -- `domain.CoverageContractViolation(resolveErr) != nil` discards the extracted violation and returns `resolveErr`. Correct today only because of the flattening two packages away; `if v := domain.CoverageContractViolation(resolveErr); v != nil { return nil, v }` is robust regardless.
- `internal/infra/state/write_set.go:1181` -- the ledger-load `StoreFailure` site in `ValidateTransientVolumes` has no coverage-miss test; `TestValidateTransientVolumesStorageFaultTakesPrecedence` only drives the `CheckCoverage` site at 1205.
- `internal/infra/state/scope.go:42` -- `Error()` still formats `raft_index=%d` while `Metadata()` six lines below now emits `raftIndex`. `Error()` feeds the hash-chained `AuditFailure.Message`, so the two halves of the same audit record now disagree on the field name.
- `internal/infra/state/write_set.go:1169` -- `storageFault` / `storageFaults` now misdescribes its dominant occupant (`*ErrCoverageMiss`); the comment at 1220 is already half-renamed to "A storage/coverage fault".
- PR description cites `machine.go:1215` for `planInvariantDescribable`; the definition is `machine.go:1133` (as the doc correctly says) and the TU call site is `machine.go:1214`. Description-only.

Housekeeping, unrelated to the diff: the branch worktree has an untracked `internal/query/zz_forbidigo_probe.go` (the manual lint probe). It is not gitignored, so it could be committed by accident.

internal/domain/coverage.go
```diff
@@ -0,0 +1,57 @@
...
+// build its failure through this function rather than constructing an
+// ErrStorageOperation directly — a bare construction swallows the coverage
+// reason, which is enforced by a forbidigo rule in .golangci.yaml.
+//
+// operation is the short identifier ErrStorageOperation surfaces ("loading
+// ledger", "checking transaction reference"). err is expected to be non-nil;
+// calling with nil yields an ErrStorageOperation with a nil cause.
+func StoreFailure(operation string, err error) Describable {
+	if violation := CoverageContractViolation(err); violation != nil {
+		return violation
```
paul-nicolas, line 71: [Major] This changes hash-bound FSM output, so a coverage miss during a rolling upgrade diverges the audit chain across nodes

This line is where the FSM's observable output changes for a given Raft entry, and that output is cryptographically bound. `buildAuditFailurePayload` (`internal/infra/state/audit_envelope.go:179-196`) folds all three of the changed fields into the hash pre-image:

```go
buf = appendU32(buf, uint32(f.GetReason()))     // STORAGE_OPERATION_FAILED -> COVERAGE_MISS
buf = appendLenString(buf, f.GetMessage())      // "storage operation failed: <op>" -> ErrCoverageMiss.Error()
// ... then every sorted Context key and value
```

So for one and the same Raft entry that triggers a coverage miss, a pre-EN-1379 node computes an `AuditEntry.Hash` over `{STORAGE_OPERATION_FAILED, "storage operation failed: loading ledger", {operation: ...}}` while a post-EN-1379 node computes it over `{COVERAGE_MISS, "preload coverage miss (...)", {attribute, canonicalHex, idHex, raftIndex}}`. During a rolling upgrade both versions apply the same replicated log, so the two nodes write different hashes for the same index — invariant #2 (FSM determinism) broken across the version boundary, and `checker.verifyAuditHashChain` then fails on whichever node disagrees with the persisted chain.

There is no version gate that could absorb this: `HashVersion` is set from `fsm.State.HashGenerator.Algorithm()` (`machine.go:1427`), i.e. it identifies the hash *algorithm* only, and carries no notion of failure-projection semantics. Nothing lets a node recognise "this entry was produced under pre-EN-1379 classification."

On likelihood, being fair to the change: the trigger requires a coverage miss — itself an admission bug that should never fire — to land inside the upgrade window, so this is a low-probability path. I am also not claiming it is unique to this PR: the same hazard attaches to any change in an FSM-emitted error's reason, message or metadata, so the repo presumably already relies on an upgrade convention here. But this PR deliberately changes all three at once on a path whose entire purpose is to fire when something is already wrong, and the PR description treats the change as observability-only ("The gRPC status code is unchanged"), which understates it.

Fix: no code change is necessarily right here — the ask is that this be a conscious decision rather than an implicit one. Either (a) document an upgrade constraint (coverage misses are only correctly classified once all nodes are on the new version; a miss during a mixed-version window can diverge the chain), or (b) if the project wants this to be structurally safe, gate the failure projection on a semantics version carried alongside `HashVersion` so old entries keep reproducing their original bytes. Whichever you choose, it belongs in the PR description and the `coverage-gate.md` page, because a future reader will otherwise reasonably assume relabelling an FSM error is hash-neutral.

docs/technical/architecture/subsystems/fsm/coverage-gate.md
```diff
@@ -75,6 +75,50 @@ The result is a value the apply path was never authorized to consult. Two failur
...
 
+## How a violation surfaces
+
+A coverage miss is an admission bug, not an infrastructure fault, and it is labelled that way on every surface it reaches:
+
+| Surface | Value |
+|---------|-------|
+| Hash-chained `AuditFailure.Reason` | `ERROR_REASON_COVERAGE_MISS` |
+| `AuditFailure.Context` | `attribute`, `canonicalHex`, `idHex`, `raftIndex` |
+| Frozen idempotency outcome | the same reason and metadata |
```
paul-nicolas: [Medium] This surface does not exist — a coverage miss is never frozen into an idempotency outcome

This table row claims the frozen idempotency outcome carries "the same reason and metadata," and the PR description makes the matching claim ("The same outermost-only read also froze the wrong reason into the idempotency outcome (`machine.go:1746-1748`)"). Neither holds:

- `recordIdempotencyFailure` returns early at `machine.go:1729`: `if !errors.As(bizErr, &d) || !domain.IsFreezableFailure(domain.Kind(d)) { return nil }`
- `IsFreezableFailure` (`internal/domain/errors.go:151-158`) admits only `KindValidation`, `KindNotFound`, `KindAlreadyExists`, `KindConflict`, `KindPrecondition` — `KindInternal` returns `false`.
- `KindForReason` (`internal/domain/reason.go:119-132`) classifies `ERROR_REASON_COVERAGE_MISS` as `KindInternal`.

So the write at `machine.go:1744-1750` is never reached for a coverage miss. Note this is also true of the *old* behavior: `ERROR_REASON_STORAGE_OPERATION_FAILED` is in the same `KindInternal` arm, so nothing was ever frozen on this path either. The premise is not that the PR broke the surface — it is that the surface has never existed in either direction, so it should not be listed as one of the three the fix repairs.

This matters more as shipped documentation than as a description slip: the table is presented as the authoritative account of where a violation surfaces, and a reader would reasonably conclude that replaying a request under the same idempotency key returns `COVERAGE_MISS`. In fact an internal-kind failure is deliberately left unfrozen so the request can be retried, which is the opposite operational expectation.

Fix: drop this row from the table, and drop the corresponding sentence from the PR description. If it is worth stating positively, replace it with something like: "Frozen idempotency outcome — not applicable; `KindInternal` failures are deliberately not frozen (`IsFreezableFailure`), so a coverage miss leaves no idempotency record and the request stays retryable." That turns a wrong claim into a useful one, since the non-freezing is itself a deliberate design point a reader would want to know.

paul-nicolas (commented):
Follow-up to my earlier review — a late reviewer pass surfaced two findings that outrank everything in it, both now verified against the code. Also, one **correction to my own earlier comment**.

**Correction.** In my comment on `internal/infra/state/scope_unit_test.go` I asked for a test asserting that the frozen idempotency outcome carries `COVERAGE_MISS`. That request was wrong — as the second comment below establishes, a coverage miss is never frozen at all, so no such test should exist. The other two asks in that comment (the `AuditFailure` reason and the `Context` key set) stand as written; please disregard the idempotency one.

The two new findings both concern the PR's central change — the reason relabeling itself — rather than the metadata key rename my earlier Major was about.

For the record, one more claim I verified and cleared: `StoreFailure(op, nil)` returning a non-nil `Describable` is a theoretical footgun (it would write a phantom audit failure), but every one of the 39 call sites is inside an `if err != nil` guard, so the branch is unreachable. Not worth changing.

internal/infra/state/scope.go
paul-nicolas: Confirmed, and the premise critique is right — I checked both paths on `release/v3.0` before deciding. `git show release/v3.0:internal/domain/processing/processor_transaction_numscript.go` already carries the `isCoverageContractViolation(resolveErr) -> return nil, resolveErr` branch, `numscript/errors.go:68-72` flattens to the bare miss via `errors.As`, and `buildAuditFailure` does `maps.Copy(failure.GetContext(), d.Metadata())` with no unwrap. The snake_case keys were already in the chain. Same for `planInvariantDescribable` reaching the gRPC `ErrorInfo`. "Previously unreachable" was false and the rename is a real breaking change.

Took option 2 — kept camelCase, made it deliberate and announced — rather than reverting. The deciding factor: `internal/domain/errors.go` has ~90 metadata keys and zero snake_case ones, the only multi-word one being `existingTransactionId` (`errors.go:562`). A repo-wide grep for snake_case metadata keys under `internal/domain` + `internal/infra/state` returns only `scope.go:375-377`, which is the structured log (deliberately snake_case), plus string *values* in `internal/domain/indexes/id.go`. `*ErrCoverageMiss` was the last snake_case `Describable` on the wire, and reverting would leave the only outlier on the error that most needs to be greppable — while deferring the identical forensic seam to a later PR.

Changes in `633c073c3`:

- `scope.go:48-72` — the comment now names both pre-existing paths explicitly and states the rename as a deliberate breaking change with its cost (tooling keyed on `canonical_hex`/`id_hex`/`raft_index` stops matching; audit entries differ either side of the deploy). The false rationale is gone.
- `scope.go:42` — `Error()` aligned to `raftIndex=%d`, so no single identifier appears in two spellings inside one audit record. Left `kind=` / `id=` alone: those are prose labels, different words rather than different spellings of the same key.
- PR description rewritten on the same basis.

On the changelog: there is no `CHANGELOG` file in this repo (`ls CHANGELOG*` is empty), so the announcement surfaces are the PR description and `coverage-gate.md` — both now carry it.

.golangci.yaml
paul-nicolas: Agreed — the message advised only "use `domain.StoreFailure`", which is no advice at all for a consumer, and `analyze-types` makes the rule fire on any reference resolving to the type.

Fixed in `ca762843a`. The `msg` now states that the rule matches any such reference — an `errors.As` target declaration, a field read off it, a type-switch case — not only a composite literal, notes that a construction-only pattern is not expressible (forbidigo renders the type expression, never the brace), and points legitimate consumers at `//nolint:forbidigo` with a justification.

Also took the adjacent Minor from your summary body: the `internal/domain/coverage\.go` exclusion was a whole-file rule, asymmetric with the `source:`-pinned `errors\.go` rule below it. It is now pinned to `source: 'ErrStorageOperation\{'`, so a second unintended reference in that file is still caught.

Not doing the stronger option here, per your own recommendation — unexporting the struct as `errStorageOperation` would be compiler-enforced and let this rule plus both exclusions go away, but cross-package test assertions name the exported type today, so it is a separate change.

One Minor from the summary body I am deliberately leaving: the `_test.go$` forbidigo exclusion at the bottom. It is pre-existing and repo-wide — it also carries the `OpenWriteSession` test exemption — so narrowing it to per-rule scope is its own change rather than something to fold in here. The demonstrated regression path you noted (`write_set_test.go:958` had pinned the flattened shape) is real, and the new `TestBuildAuditFailure/CoverageMiss` at least fails loudly if the production shape regresses.

Deleted the untracked `internal/query/zz_forbidigo_probe.go` you flagged under housekeeping — it is not gitignored and would have broken the build if committed.

internal/infra/state/scope_unit_test.go
paul-nicolas: Right, and the diagnosis of *why* it matters is the useful part: the key names were pinned on `Metadata()` directly but nowhere they are consumed, which is exactly how the rename could ship with a green suite.

Added in `d5177aef1`:

- `TestBuildAuditFailure/CoverageMiss` — drives an `*ErrCoverageMiss` through `buildAuditFailure` and asserts the reason, the message, and the **exact** `Context` key set (a whole-map `require.Equal`, so an added key fails too, not just a renamed one). The rationale is in the test comment: `buildAuditFailurePayload` folds every sorted context key *and* value into the hash pre-image, so those names are bytes that are immutable once written.
- `TestBuildAuditFailure/CoverageMissWrappedStaysStorageFault` — negative control. `buildAuditFailure` genuinely does not unwrap, so the no-wrap contract rests entirely on read sites going through `StoreFailure`; this pins the `{operation: "loading volume"}`-only shape the chain would otherwise record forever, and asserts the supported path avoids it.

Also took the related Minor from your summary body: `TestValidateTransientVolumesLedgerCoverageMissPropagates` now covers the ledger-load `StoreFailure` site (`write_set.go:1181`), which had no coverage-miss test — the existing test only drove the `CheckCoverage` site at 1205.

Honouring your correction: no idempotency assertion added, since the surface does not exist.

Not adding a gRPC-adapter test. `businessErrorToGRPCStatus` is fully generic over `Describable` (`grpc/errors.go:86`), so pinning a reason string there tests the adapter's plumbing rather than anything this PR changes — and your own fix line asks for the `buildAuditFailure` test as the one whose case is strongest. Happy to add it if you disagree.

internal/domain/processing/processor_volumes.go
paul-nicolas: Confirmed at `internal/application/admission/admission.go:1123-1127`, and the point about the location is fair — `addVolumeNeed` is the volume-preload declaration site CLAUDE.md invariant #6 singles out, so it is the comment a reader auditing coverage misses on the volume path lands on first.

Fixed in `633c073c3` with your wording:

> A `*state.ErrCoverageMiss` (admission contract violation — need never declared) stays distinct and propagates verbatim through `domain.StoreFailure`, so the audit chain records `COVERAGE_MISS` rather than a storage fault (EN-1379).

I re-swept for others outside `internal/domain/processing/` and found no further stale references to the flattening.

internal/domain/coverage.go
paul-nicolas: Verified and accepted — `buildAuditFailurePayload` (`audit_envelope.go:176-198`) folds the reason, the message and every sorted context key *and* value into the pre-image, and `HashVersion` genuinely cannot absorb it since it comes from `HashGenerator.Algorithm()` and names the algorithm only. The framing that this should be a conscious decision rather than an implicit one is the right ask.

Took option (a), documentation, in `633c073c3`: a new **"Upgrade note: the reason is hash-bound"** subsection in `coverage-gate.md` plus the matching section in the PR description. It states the divergence mechanism with both concrete pre-images, that correct classification is guaranteed only once every node runs the new build, that `HashVersion` cannot recognise the old classification, and — taking your point that this is not unique to the PR — that **the hazard is general to any change to an FSM-emitted error's reason, message or metadata**, so it reads as a standing review rule rather than an EN-1379 footnote. It also records that option (b) would mean a failure-projection semantics version alongside `HashVersion`, and that this PR deliberately does not take that on.

One precision I added while writing it up: the divergence applies only to the read sites this PR *converts*. The numscript re-resolution path already returned the bare miss on `release/v3.0`, so its reason and context are unchanged — only its message shifts there, and only because of the `Error()` alignment discussed on `scope.go`.

Also dropped the "The gRPC status code is unchanged" line from carrying the whole story in the description; it is still true, but the audit hash is not unchanged and the description now says so directly.

docs/technical/architecture/subsystems/fsm/coverage-gate.md
paul-nicolas: Confirmed every link: the early return at `machine.go:1729`, `IsFreezableFailure` admitting only the five definitive kinds (`errors.go:152-158`), and `KindForReason` putting `ERROR_REASON_COVERAGE_MISS` in the `KindInternal` arm (`reason.go:124`) — alongside `STORAGE_OPERATION_FAILED`, so the surface never existed in either direction. Listing it as one of the three the fix repairs was wrong.

Fixed in `633c073c3`, taking your positive formulation rather than just deleting the row:

> | Frozen idempotency outcome | **not applicable** — `KindInternal` failures are deliberately never frozen (`IsFreezableFailure`), so a coverage miss leaves no idempotency record and the request stays retryable |

Plus a short paragraph under the table explaining that the non-freezing is a design point, not an oversight — freezing would pin a server bug against the caller's key for the whole TTL — and noting it was equally true before EN-1379.

Two related edits in the same commit, since both rested on the same false premise:

- Further down the page, "The same outermost-only read happens in `recordIdempotencyFailure` (`machine.go:1746-1748`), which freezes the outcome for the idempotency TTL" now reads that `recordIdempotencyFailure` reads the outermost `Describable` the same way but never reaches a coverage miss at all, cross-referencing the table row. `businessErrorToGRPCStatus` keeps its mention — that one really is a live surface.
- The PR description's matching claim is gone, replaced by a short "Not a surface: the frozen idempotency outcome" section stating the correction explicitly rather than silently dropping it.

paul-nicolas:
Thanks — agreed on the classification of `codecov/patch` as tooling rather than a runtime blocker, and the review pass that followed found nothing that changes that. Pushed `ca762843a` addressing the six inline findings; the two substantive ones were both about the reason relabeling rather than the mechanism you revalidated:

- the metadata rename was justified by a false "previously unreachable" premise (two paths already surfaced the bare miss on `release/v3.0`, one of them into the hash-chained audit) — kept as a deliberate, announced breaking change;
- the relabeling changes hash-bound FSM output, so a coverage miss inside a rolling-upgrade window can diverge the chain — now documented as an upgrade constraint in `coverage-gate.md`.

On patch coverage specifically: the two new tests (`TestBuildAuditFailure/CoverageMiss` and `TestValidateTransientVolumesLedgerCoverageMissPropagates`) close the gap the audit and ledger-load surfaces had, so the number should improve, though the bulk of the uncovered lines are the mechanical `StoreFailure` conversions across the 39 read sites.

internal/domain/coverage.go
```diff
@@ -0,0 +1,72 @@
...
```
NumaryBot, line 71: 🔴 **[blocker] Rolling-upgrade hash divergence: FSM output changes break audit chain determinism**

`buildAuditFailurePayload` folds the failure reason, message, and all context keys into the cryptographic hash pre-image. For the same Raft log entry that triggers a coverage miss, a pre-EN-1379 node hashes `{STORAGE_OPERATION_FAILED, "storage operation failed: <op>", {operation: ...}}` while a post-EN-1379 node hashes `{COVERAGE_MISS, "preload coverage miss (...)", {attribute, canonicalHex, idHex, raftIndex}}`. During a rolling upgrade both nodes apply the same replicated log, so they write different hashes for the same index — breaking the FSM determinism invariant (#2) across the version boundary.

**Suggestion:** Introduce a versioned migration gate or a compatibility shim so the FSM produces identical output regardless of which binary processes the entry, or gate the new behavior behind a feature flag / min-version barrier that prevents mixed-version clusters from diverging.

internal/infra/state/scope.go
```diff
@@ -39,17 +39,42 @@ type ErrCoverageMiss struct {
...
+// The structured log emitted by coverageMiss keeps its own snake_case field
+// names — that is a log, not a wire payload.
 func (e *ErrCoverageMiss) Metadata() map[string]string {
 	return map[string]string{
-		"attribute":     e.Attribute,
-		"canonical_hex": e.CanonicalHex,
-		"id_hex":        e.IDHex,
-		"raft_index":    strconv.FormatUint(e.RaftIndex, 10),
+		"attribute":    e.Attribute,
+		"canonicalHex": e.CanonicalHex,
```
NumaryBot, line 75: 🟠 **[major] Metadata key rename is observable on audit and gRPC surfaces for pre-existing paths**

The PR claims the metadata map was "previously unreachable" because it was always swallowed by the storage-fault wrap. At least two pre-existing paths contradict this: (1) the `isCoverageContractViolation` → `return nil, resolveErr` branch in Numscript re-resolution propagates a bare `*ErrCoverageMiss` through `ProcessOrders`, whose failure writes an audit entry that copies `d.Metadata()` into `AuditFailure.Context`; (2) a parallel gRPC path reads the same keys for `ErrorInfo`. Any consumer keyed on the old camelCase names will silently break.

**Suggestion:** Either keep the original key names, or provide an explicit migration note and bump any contract version that captures these keys, ensuring downstream consumers are updated in the same change.

.golangci.yaml
```diff
@@ -126,6 +126,28 @@ linters:
...
```
NumaryBot, line 135: 🟠 **[major] Forbid pattern blocks consumption of `ErrStorageOperation`, not just construction**

The `\bErrStorageOperation\b` pattern with `analyze-types` matches any expression whose type resolves to `ErrStorageOperation`, including `errors.As` target declarations, type-switch cases, and field reads on an already-received value. The lint message only advises "Use domain.StoreFailure(operation, err)", which is meaningless for consumers. This will block legitimate inspection of storage faults (e.g. reading `Operation` for a log line or distinguishing IO from business outcomes in a type switch).

**Suggestion:** Narrow the pattern to match only composite-literal construction (e.g. `ErrStorageOperation{`) or add a linter exemption comment at each legitimate consumption site, and update the message to reflect that only construction is prohibited.

internal/application/admission/admission.go
```diff
@@ -1122,8 +1122,9 @@ func wrapSystemScoped(order *raftcmdpb.Order, ss *raftcmdpb.SystemScopedOrder) {
 // `Scope.GetVolume` returns `domain.ErrNotFound` and callers treat it
 // as a fresh zero balance (see `processing.readVolumeOrZero`). A
 // `*state.ErrCoverageMiss` (admission contract violation — need never
-// declared) stays distinct and propagates loud through
-// `ErrStorageOperation{Cause: covErr}`.
+// declared) stays distinct and propagates verbatim through
```
NumaryBot, line 1125: 🟠 **[major] Stale comment on invariant-#6 volume-preload site still documents the removed wrapping**

Lines 1125–1126 of `addVolumeNeed` still read: "`*state.ErrCoverageMiss` (admission contract violation — need never declared) stays distinct and propagates loud through `ErrStorageOperation{Cause: covErr}`". After this PR the violation propagates verbatim and is never wrapped. Because CLAUDE.md marks volume preload as load-bearing for the `new_kept_volumes`/`ephemeral_volumes` split under invariant #6, a reader auditing that invariant will come away believing the flattening is still in place.

**Suggestion:** Update the comment to reflect the new behavior: the miss propagates as a bare `*ErrCoverageMiss` without being wrapped in `ErrStorageOperation`.

internal/infra/state/scope_unit_test.go
```diff
@@ -33,15 +33,80 @@ func TestErrCoverageMiss_Describable(t *testing.T) {
...
```
NumaryBot, line 99: 🟡 **[minor] Tests do not assert metadata key names on the surfaces the PR claims to fix**

The unit test binds the Reason-string contract correctly but does not assert what actually lands on the three critical surfaces: (1) `AuditFailure.Reason` / `Context` key names in the hash-chained audit, (2) the frozen idempotency outcome, (3) `ErrorInfo.reason` in the gRPC adapter. Because the surfaces mechanically read `d.Reason()` / `d.Metadata()`, the composition is tight — but the specific key names (which the metadata rename affects) are only validated at the helper level, not at the consumption points.

**Suggestion:** Add targeted tests for each surface: one that exercises `buildAuditFailure` with a `*ErrCoverageMiss` and asserts the exact Context key set; one for the gRPC error-info path; and confirm via the `IsFreezableFailure` guard that the idempotency path is intentionally unreachable (or document it).

docs/technical/architecture/subsystems/fsm/coverage-gate.md
```diff
@@ -75,6 +75,67 @@ The result is a value the apply path was never authorized to consult. Two failur
...
 
+## How a violation surfaces
+
+A coverage miss is an admission bug, not an infrastructure fault, and it is labelled that way on every surface it reaches:
+
+| Surface | Value |
+|---------|-------|
+| Hash-chained `AuditFailure.Reason` | `ERROR_REASON_COVERAGE_MISS` |
+| `AuditFailure.Context` | `attribute`, `canonicalHex`, `idHex`, `raftIndex` |
+| Frozen idempotency outcome | **not applicable** — `KindInternal` failures are deliberately never frozen (`IsFreezableFailure`, `internal/domain/errors.go:152-158`), so a coverage miss leaves no idempotency record and the request stays retryable |
```
NumaryBot, line 86: ⚪ **[nit] Documentation incorrectly claims coverage misses are frozen into idempotency outcomes**

`recordIdempotencyFailure` returns early when `IsFreezableFailure` returns false. `KindForReason` classifies `ERROR_REASON_COVERAGE_MISS` as `KindInternal`, and `IsFreezableFailure` admits only `KindValidation`, `KindNotFound`, `KindAlreadyExists`, `KindConflict`, `KindPrecondition`. The write at `machine.go:1744-1750` is never reached for a coverage miss — in either old or new behavior. The table row and PR description claim this surface was fixed, but it was never active.

**Suggestion:** Remove or correct the table row claiming the frozen idempotency outcome carries the coverage-miss reason/metadata, and clarify in the PR description that this surface was not actually affected.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1647#issuecomment-5117696542

internal/domain/coverage.go
paul-nicolas: Duplicate of the human review's Major on this line, which is already handled in `633c073c3` — but the suggestion differs from what was agreed, so recording the decision rather than silently resolving.

The mechanism is confirmed and accepted. What was taken is the documentation option: a new "Upgrade note: the reason is hash-bound" section in `coverage-gate.md` plus a matching section in the PR description, stating the divergence with both concrete pre-images, that `HashVersion` identifies the algorithm only and cannot absorb it, and that the hazard is general to any change to an FSM-emitted error's reason, message or metadata.

The versioned gate you suggest is the structurally safe answer and is explicitly recorded as *not* taken here: it means carrying a failure-projection semantics version alongside `HashVersion` so old entries keep reproducing their original bytes. That is a deliberate design change, out of scope for this PR. The trigger is an admission bug that should never fire, so the exposure is a narrow, documented window rather than a routine path.

One correction to the finding: only the read sites this PR converts are affected. The numscript re-resolution path already returned the bare miss on `release/v3.0`, so its reason and context are unchanged.

internal/infra/state/scope.go
paul-nicolas: Duplicate of the human review's Major on this line; handled in `633c073c3`.

The premise critique is right and was verified against `release/v3.0`. Took the second of your two options: the keys stay camelCase and the change is now explicit rather than justified by a false "previously unreachable" claim. `scope.go:48-72` names both pre-existing paths, states the rename as a deliberate breaking change and spells out the cost; the PR description carries the same. There is no `CHANGELOG` file in this repo, so those are the announcement surfaces. `Error()` was also aligned to `raftIndex` so one audit record does not carry both spellings.

Two corrections to the finding as written:

- The direction is inverted. The **old** keys were snake_case (`canonical_hex`, `id_hex`, `raft_index`); camelCase is what this PR moves *to*. Consumers keyed on the snake_case names are the ones that break.
- Keeping the original names was weighed and rejected: `internal/domain/errors.go` has ~90 metadata keys and zero snake_case ones, so `*ErrCoverageMiss` was the last outlier on the wire, and CLAUDE.md mandates camelCase there. Reverting would defer an identical seam to a later PR.

.golangci.yaml
paul-nicolas: Duplicate of the human review's Major on this line; handled in `ca762843a`.

The diagnosis is right and the message now reflects it: it states that `analyze-types` makes the rule fire on any reference resolving to the type — `errors.As` target declaration, field read, type-switch case — and directs legitimate consumers to `//nolint:forbidigo` with a justification.

The first half of the suggestion is not achievable, however: **narrowing the pattern to composite-literal construction is not expressible in forbidigo.** With `analyze-types` the linter renders the *type expression* (`domain.ErrStorageOperation`) and never the trailing brace, so a pattern like `ErrStorageOperation\{` never matches a construction. That was verified empirically before the message was rewritten. Hence the advisory route rather than a narrower pattern.

The structurally stronger option — unexport the struct as `errStorageOperation` and export only `StoreFailure`, letting the compiler enforce it and deleting this rule plus both exclusions — is the real fix, but cross-package test assertions name the exported type today, so it is a separate change.

(Note `source:` pinning *is* used for the two exclusion entries, where forbidigo matches against the source line rather than the rendered type expression. `coverage.go`'s exclusion was tightened to `source: 'ErrStorageOperation\{'` in the same commit.)

internal/application/admission/admission.go
paul-nicolas: Already fixed in `633c073c3`, before this review ran — the finding is stale. The comment on the pushed head (`ca762843a`) reads:

> A `*state.ErrCoverageMiss` (admission contract violation — need never declared) stays distinct and propagates verbatim through `domain.StoreFailure`, so the audit chain records `COVERAGE_MISS` rather than a storage fault (EN-1379).

Verifiable with `git show origin/EN-1379-surface-coverage-miss:internal/application/admission/admission.go | sed -n '1119,1128p'`. The point about invariant #6 was well taken and is why it was fixed rather than deferred.

internal/infra/state/scope_unit_test.go
paul-nicolas: Mostly addressed in `d5177aef1`, before this review ran.

- **Audit surface** — added. `TestBuildAuditFailure/CoverageMiss` drives an `*ErrCoverageMiss` through `buildAuditFailure` and asserts the reason, the message and the exact `Context` key set (whole-map `require.Equal`, so an added key fails too). A sibling negative control pins what the wrapped shape would record instead.
- **Idempotency** — your own suggestion is the right one and is what was done: it is not a surface at all. `KindForReason` classifies `COVERAGE_MISS` as `KindInternal` and `IsFreezableFailure` rejects that kind, so `recordIdempotencyFailure` returns early. `coverage-gate.md` now documents the deliberate non-freezing instead of listing a fixed surface. No test added, since there is no behaviour to pin.
- **gRPC** — deliberately not added. `businessErrorToGRPCStatus` (`grpc/errors.go:86`) is generic over `Describable`, so asserting a reason string there tests the adapter's plumbing rather than anything this PR changes.

Also added `TestValidateTransientVolumesLedgerCoverageMissPropagates`, covering the ledger-load `StoreFailure` site that had no coverage-miss test.

docs/technical/architecture/subsystems/fsm/coverage-gate.md
paul-nicolas: Already fixed in `633c073c3`, before this review ran — the finding is stale. The row on the pushed head reads:

> | Frozen idempotency outcome | **not applicable** — `KindInternal` failures are deliberately never frozen (`IsFreezableFailure`, `internal/domain/errors.go:152-158`), so a coverage miss leaves no idempotency record and the request stays retryable |

with a paragraph below explaining that the non-freezing is a design point (freezing would pin a server bug against the caller's key for the whole TTL) and that it was equally true before EN-1379. The stale "the same outermost-only read happens in `recordIdempotencyFailure`" sentence further down the page was corrected in the same commit, and the PR description now carries an explicit "Not a surface: the frozen idempotency outcome" correction rather than silently dropping the claim.

flemzord (changes requested):
Revalidated live head `ca762843`. One correctness/runtime blocker remains, already identified by the current NumaryBot thread: this change makes mixed-version nodes hash different `AuditFailure` payloads for the same Raft entry when a converted coverage miss occurs. Documenting the rolling-upgrade window does not restore FSM determinism; the failure reason, message, and context are all part of the audit hash pre-image. Please add a versioned compatibility or cluster-wide activation barrier so every node produces identical bytes before enabling the new projection. The `codecov/patch` and bot status failures are not the basis of this review.

paul-nicolas:
Revalidated the mechanism and it is not in dispute — `buildAuditFailurePayload` (`internal/infra/state/audit_envelope.go:176-198`) folds the reason, the message and every sorted context key *and value* into the hash pre-image, and `HashVersion` comes from `HashGenerator.Algorithm()` (`machine.go:1427`), so it names the algorithm only and nothing lets a node recognise an entry written under the old classification.

On the remedy I am pushing back, while acting on the operational half of your point.

**A cluster-wide activation mechanism does exist**, and it would be the right home: `ClusterConfig` is Raft-replicated and applied at one deterministic index (`applyClusterConfig`, `machine_technical_updates.go:102`), and it already carries `hash_algorithm` for exactly this class of concern. But **a single-release flag does not close the window.** Config propagation is leader-driven — the leader proposes when its CLI config differs from the persisted one (`docs/ops/cluster-operations.md:517-522`) — and an old-binary node does not understand a new proto field: forward-compatibility means it ignores it and keeps the old projection. Genuinely closing the window requires the standard two-phase rollout (release N ships the field defaulting to old behaviour; release N+1, or an operator flip once every node is upgraded, enables the new projection). That is a two-release design change, not a patch to this PR.

It would also be inconsistent with how this branch has already decided the same class of question, three times, for strictly larger breaks:

- `docs/ops/deployment.md:642-661` — the pre-#400 change moved the whole audit hash chain to a cluster-id-keyed hash, making **every** pre-upgrade entry unverifiable. Recorded decision: "the service is not yet stable, so this break is accepted instead of adding migration code" and "mixed-binary rolling upgrades are **not supported** across this change. Stop all nodes before deploying the new binary." No gate.
- `e43016439` (EN-1558, merged six days ago) changed `marshalOrdersForAudit` to hash business-intent bytes instead of the whole order. Same Raft entry, different `AuditEntry.Hash` on **every order** across a mixed-binary window — certain, not low-probability. It proved *past* entries still verify verbatim ("no migration or hash-version bump is required") but added no mixed-version gate and no `deployment.md` note.
- `8b24ee535` (EN-1536, merged two days ago) changed `buildAuditFailure` itself — from `errors.As`-unwrapping to the outermost `Describable`, and `Message` from `err.Error()` to `d.Error()`. For a wrapped failure that shifts the hashed reason, message and context: precisely the fields at issue here. No gate, no upgrade note.

`release/v3.0` is still pre-GA (`v3.0.0-alpha.13`, no `v3.0.0`). EN-1379's exposure is narrower than any of the three: it needs a coverage miss — an admission-contract violation that invariant #7 says must never fire — to land inside the window. Gating this one while those shipped bare would single out the only change that actually documented its hazard.

**Where you are right is the operational half**, and that is now fixed. Your words were "before enabling the new projection" — and nothing operationally stopped an operator from rolling this out node-by-node, because the constraint lived only in `coverage-gate.md`, an internal architecture page, rather than where supported upgrade paths are documented. Fixed in `7240f5aa1`:

- `docs/ops/deployment.md` gains an **"Upgrading across EN-1379 (coverage-miss reclassification)"** section as a sibling of the pre-#400 one, declaring mixed-binary rolling upgrades **not supported** across this change and noting that, unlike pre-#400, no data wipe is required — no persisted layout changed, `storage-schema-version` is unaffected, and the exposure is confined to entries written inside the window.
- `coverage-gate.md` cross-links to it, so the architecture page points at the operational procedure instead of being the constraint's only home.

The general mechanism is tracked in **[EN-1661](https://formance-team.atlassian.net/browse/EN-1661)**: a failure-projection semantics version alongside `HashVersion`, gated through `ClusterConfig`, with the two-phase rollout constraint written into the scope. Filed separately because the hazard attaches to *any* change to an FSM-emitted error's reason, message or metadata — so it should not be solved once for one error class, and the pre-GA justification for accepting these breaks expires at v3.0 GA.

Local gates on `7240f5aa1`: `just pre-commit` reports 0 issues and the full unit suite passes with no failures. Happy to build the gate in this PR instead if you would rather not defer it — flagging that it only becomes fully effective in a second release either way.


[EN-1661]: https://formance-team.atlassian.net/browse/EN-1661?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1647#issuecomment-5117696542

internal/domain/coverage.go
```diff
@@ -0,0 +1,72 @@
...
+
+import "errors"
+
+// CoverageContractViolation returns the admission-contract violation carried by
+// err — a coverage miss (ErrReasonCoverageMiss) or a structurally inconsistent
+// execution plan (ErrReasonInvalidExecutionPlan) — or nil when err is neither.
+//
+// Both mean the FSM apply path read a key the proposer never declared, or was
+// handed a plan that cannot be applied: "should not happen" bugs (invariant #7)
+// that must reach the audit chain, the frozen idempotency outcome and the client
```
NumaryBot: ⚪ **[nit] Misleading comment references non-existent idempotency surface for coverage misses**

The comment references an idempotency surface that does not exist for coverage misses. `KindForReason` maps `ERROR_REASON_COVERAGE_MISS` to `KindInternal`, and `IsFreezableFailure` only admits five definitive kinds, so `recordIdempotencyFailure` exits early before writing any frozen outcome. This contradicts the updated `coverage-gate.md` and can mislead future readers into thinking coverage misses are freezable.

**Suggestion:** Remove or correct the reference to the idempotency surface in the comment, noting explicitly that `KindInternal` failures (including coverage misses) are never frozen by `recordIdempotencyFailure`.

flemzord (changes requested):
Revalidated live head 7240f5aa. The new deployment documentation accurately describes the mixed-binary hazard, but it does not remove the correctness/runtime blocker: old and new nodes still hash different AuditFailure reason/message/context bytes for the same applied Raft entry when a coverage miss occurs. A stop-all upgrade procedure is an operational mitigation, not deterministic enforcement in the FSM or cluster configuration, and this is therefore not merely pre-stable wire compatibility. The existing determinism finding remains active. The codecov/patch failure is not the basis of this review.

internal/domain/coverage.go
paul-nicolas: Fixed in `20bf89316` — dropped the idempotency surface from the `CoverageContractViolation` doc comment.

Confirmed unreachable, exactly as you describe: `KindForReason` puts both `ERROR_REASON_COVERAGE_MISS` and `ERROR_REASON_INVALID_EXECUTION_PLAN` in the `KindInternal` arm (`internal/domain/reason.go:123-124`), `IsFreezableFailure` admits only the five definitive kinds (`internal/domain/errors.go:151-158`), and `recordIdempotencyFailure` returns early on that gate before touching the write session (`internal/infra/state/machine.go:1729`). The same gate guards the admission-side path at `internal/application/admission/admission.go:1571`.

The comment now names the audit chain and the client as the surfaces, states the non-freezing explicitly rather than leaving it to be re-derived, and cross-links `coverage-gate.md` — which had it right at lines 86 and 92, and was what the stale comment contradicted.

paul-nicolas:
Head is now `20bf89316` (doc-comment only — the last open inline finding; no behavioural change since `7240f5aa1`, so the determinism assessment carries over unchanged).

On the determinism finding: the mechanism is not disputed and I am not asking you to drop it. The reason, message and every sorted context key and value go into the `AuditFailure` hash pre-image, and `HashVersion` names the algorithm only, so nothing lets a node recognise an entry written under the old classification. A coverage miss landing inside a mixed-binary window does diverge the chain.

The disagreement is confined to the remedy, and the position stated in [the earlier comment](https://github.com/formancehq/ledger/pull/1647#issuecomment-5128444097) stands:

- **A `ClusterConfig` flag added in this PR would not close the window.** Config propagation is leader-driven, and an old binary does not understand a new proto field — forward compatibility means it ignores it and keeps the old projection. Closing the window genuinely requires the standard two-phase rollout: release N ships the field defaulting to the old behaviour, release N+1 (or an operator flip once every node is upgraded) enables the new projection. That is a two-release design change, not a patch to this PR — which is why it is tracked as [EN-1661](https://formance-team.atlassian.net/browse/EN-1661) with the two-phase constraint written into its scope.
- **Gating this change alone would be inconsistent** with how `release/v3.0` has already resolved the same question three times for strictly larger breaks: the pre-#400 cluster-id-keyed hash chain (every pre-upgrade entry unverifiable), `e43016439` / EN-1558 (`marshalOrdersForAudit` — a different `AuditEntry.Hash` on *every* order across a mixed-binary window, certain rather than conditional), and `8b24ee535` / EN-1536 (`buildAuditFailure` itself — the same reason, message and context fields at issue here). None added a mixed-version gate; two added no upgrade note at all. EN-1379 is narrower than all three: it needs a coverage miss, an admission-contract violation invariant #7 says must never fire, to land inside the window.
- **The branch is pre-GA** (`v3.0.0-alpha.13`, no `v3.0.0`). The general mechanism — a failure-projection semantics version alongside `HashVersion` — belongs at EN-1661 rather than being solved once for one error class, and the pre-GA justification for accepting these breaks expires at v3.0 GA.

The operational half of your point was valid and is fixed in `7240f5aa1`: `docs/ops/deployment.md` now carries an "Upgrading across EN-1379" section as a sibling of the pre-#400 one, declaring mixed-binary rolling upgrades **not supported** across this change, and `coverage-gate.md` cross-links to it so the constraint no longer lives only on an internal architecture page.

I am deferring the enforcement half deliberately, on the record, rather than treating it as resolved. If the call is that this PR should not merge without the gate despite the three precedents, that is a judgement for a human reviewer to make — happy to build it here on that basis, flagging again that it only becomes fully effective in a second release either way.

[EN-1661]: https://formance-team.atlassian.net/browse/EN-1661?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1647#issuecomment-5117696542

flemzord:
Revalidated live head `20bf8931` under this cron run’s explicit policy. The remaining mixed-version audit-hash concern is an upgrade-compatibility hazard for this pre-stable release; the documented stop-all upgrade keeps homogeneous clusters deterministic, and this task does not treat pre-stable compatibility alone as a blocker. I found no correctness/runtime blocker on the current homogeneous head. The `codecov/patch` failure is CI/tooling only. GitHub approval is intentionally disabled by the cron configuration.

internal/domain/coverage_test.go
```diff
@@ -0,0 +1,112 @@
...
+func (s *stubDescribable) Metadata() map[string]string { return map[string]string{"stub": s.reason} }
+
+func TestCoverageContractViolation(t *testing.T) {
+	t.Parallel()
+
+	miss := &stubDescribable{reason: ErrReasonCoverageMiss, msg: "coverage miss"}
+	plan := &stubDescribable{reason: ErrReasonInvalidExecutionPlan, msg: "bad plan"}
+	other := &stubDescribable{reason: ErrReasonNumscriptRuntime, msg: "boom"}
+
+	tests := []struct {
```
Azorlogh, line 29: This table has no `errors.Join`/multi-`%w` case, so the Join descent (`coverage.go:47-50`) is uncovered in its own package — which is what `codecov/patch` is flagging. The only Join coverage lives in `state`'s `scope_unit_test.go`, via `StoreFailure`. A couple of Join rows here would make the discriminator's own test self-contained and should clear codecov.

docs/ops/deployment.md
```diff
@@ -660,6 +660,21 @@ Trying to keep existing Pebble data and upgrade in-place is **unsafe**:
 
 Mixed-binary rolling upgrades are **not supported** across this change. Stop all nodes before deploying the new binary.
 
+### Upgrading across EN-1379 (coverage-miss reclassification)
```
Azorlogh: V3 is pre-release and wiped on every push, so there is no upgrade path for this section to constrain yet — and it sets a precedent of one ops-guide entry per FSM error-identity change. The durable rule (reason/message/metadata feed the hash pre-image) already lives in coverage-gate.md; worth deciding whether this per-change section survives the pre-release doc cleanup or the standing rule alone suffices.

flemzord (dismissed):

internal/domain/coverage_test.go
paul-nicolas: Fixed in `8954dd910` — the coverage gap is real and `CoverageContractViolation` goes from 72.7% to 100%.

Measured before the change, the three Join blocks were the only zeros in the function:

```
coverage.go:47.3,47.42   1 0     <- joined, ok := e.(interface{ Unwrap() []error })
coverage.go:47.42,48.72  1 0     <- for _, member := range joined.Unwrap()
coverage.go:48.72,50.5   1 0     <- recursive CoverageContractViolation(member)
```

And the locality claim holds: the only Join exercise anywhere was `internal/infra/state/scope_unit_test.go:74-75`, two packages away and routed through `StoreFailure`.

Beyond the percentage, two things made this worth doing:

- This descent is exactly the branch the forbidigo rule **cannot** guard — `coverage.go:26-30` says so itself. A future `errors.Join` is a plain call the linter cannot see, so if the descent regressed it would silently relabel a contract violation as a storage fault in the immutable audit chain.
- The doc comment claims members are visited in slice order, "a requirement on the FSM apply path (invariant #2)", and **nothing pinned it**. A map-based or otherwise unordered traversal would have satisfied every existing row.

Added seven rows: the descent itself, the Unwrap→Join transition (`fmt.Errorf("commit: %w", errors.Join(...))`), the `fmt.Errorf("%w: %w", ...)` shape you named, a negative case so the descent cannot false-positive, a violation nested in a joined subtree (so the recursion is covered, not just one level), and a reversed pair — `errors.Join(miss, plan)` → `miss` versus `errors.Join(plan, miss)` → `plan` — which is what actually pins the determinism. Plus a `TestStoreFailure` subtest asserting `StoreFailure` does not wrap a Join-carried violation, so the end-to-end contract is asserted in the package that owns it.

**One correction: this will not clear codecov.** `codecov/patch` is 47.69% with 34 missing lines over a 65-line patch (31/65 = 47.6923%). Those lines are dominated by the mechanical `StoreFailure` conversions across `internal/domain/processing/processor_*.go` — metadata 4, mirror 4, transaction_numscript 4, metadata_schema 3, transaction 3, events_sink 2, index 2, ledger_metadata 2, numscript_library 2, posting 2, plus 4 more files. Covering these 3 lines takes the patch to 34/65 ≈ 52%, still well under the 80% target, so the check stays red. That is the same tooling-versus-blocker call already recorded twice on this PR. The rows are worth adding on their own merits, not as a way to flip the check.

docs/ops/deployment.md
paul-nicolas: Decided and acted on in `5f08837be`: the section stays in the ops guide, but it is no longer a per-ticket entry. Splitting your comment, because one half is right and the other rests on a premise I do not think holds.

**The precedent half is right, and it is my own argument turned back on me.** I argued earlier on this PR that `e43016439` (EN-1558) and `8b24ee535` (EN-1536) are strictly larger hash-identity breaks that shipped with no upgrade note at all. A named `### Upgrading across EN-1379` therefore has `deployment.md` asserting a per-ticket discipline that two already-merged changes silently violate, and every future error-identity change inherits the expectation. That is a real problem with the framing.

**The premise half I do not think holds.** Nothing in the repo documents a per-push wipe — `deployment.md:649` documents a wipe for one specific change (pre-#400), and this section explicitly says none is needed. More decisively, the default deployment path is a *rolling* update: `misc/operator/internal/controller/hash.go:13` computes a pod-template spec hash precisely "to trigger rolling updates on spec changes", against a StatefulSet. So a mixed-binary window is what an operator-managed upgrade produces by default unless a doc declares it unsupported. There is an upgrade path for the section to constrain, and it is the default one. That is also why the constraint was moved out of `coverage-gate.md` in the first place — an internal architecture page is not where operators look for supported upgrade paths.

So rather than deleting it or keeping it as-is, I reframed it as the standing rule: **`### Upgrading across an FSM error-identity change`**. It now leads with the durable rule (reason, message and every sorted context key *and value* go into the `buildAuditFailurePayload` pre-image; `HashVersion` names the algorithm only), states the operational consequence once, and lists EN-1558, EN-1536 and EN-1379 in a table as *instances* rather than as sections. That retires the one-entry-per-change precedent instead of establishing it, and it has the guide correctly covering the two changes that shipped bare. I also made the operator rolling-update default explicit, since the stop-all constraint has to be applied deliberately against it.

The `coverage-gate.md` cross-link is retargeted at the new anchor. EN-1661 still tracks the structural fix — a failure-projection semantics version alongside `HashVersion` — so the rule remains a mitigation, not a solution.

If you would rather the whole section go away in the pre-release doc cleanup, that is a fine outcome and easy to do later; my argument is only that while the operator default is a rolling update, the constraint belongs somewhere an operator will read it.

Azorlogh (approved):

PR #1649, opened 2026-07-29
feat(EN-1458): detect reverse-map orphans for removed indexed metadata fields

paul-nicolas:
## Problem

The read-side reverse map (`0x03`) is a persisted projection that **no verification pass has ever read**. Rows for a metadata field that no longer has an index survive invisibly to `Check()`, violating invariant #8.

Unlike the forward index (`0x01`) and entity-exists (`0x02`) limbs, the reverse map **cannot be range-deleted by field** — its metadata key sits *after* a fixed-width 4-byte version block — so field removal must scan the namespace and point-delete row by row. A row that scan misses is permanent and, until now, undetectable.

## Solution

- **`compareReverseMapOrphans`** (`internal/application/check/reverse_map_orphans.go`) streams the whole `0x03` keyspace off one read-only readstore snapshot and reports rows whose `(ledger, target, metadata key)` is in **neither** the stored `SubAttrIndex` registry **nor** the audit-replayed `MetadataSchema`. Two further classes share the enum: rows for a ledger the audit does not list as live, and keys that do not decode.
- **`ParseReverseMapKey`** moved from `indexbuilder` into `readstore`, beside the key builders it mirrors, and is now the single exported chokepoint for the rmap offset math with six typed sentinels. All five call sites migrated; the private parser and both `extract*` helpers are gone.
- **`Checker`** gains an optional `*readstore.Store`.
- New `CHECK_STORE_ERROR_TYPE_REVERSE_MAP_ORPHAN = 20`.

## Design decisions

- **Aggregated per field**, never per row — one event per `(ledger, namespace, metadata key)` with a row count and a sample entity. A field dropped on a large ledger strands millions of rows; per-row events would bury every other finding and blow up memory. EN-1357 is not done, so the pass keeps memory at O(indexes + orphaned fields + ledgers).
- **Version deliberately not validated.** Current and pending versions legitimately coexist during a per-replica rewrite, and stale versions are already reclaimed at boot by `purgeOrphanVersions`.
- **Cursor skew is handled per verdict class, not by one gate.** Every oracle term is pinned at the replayed sequence, but the peer read index is folded asynchronously on a live node, so its cursor can sit on *either* side of that pin — and the two directions fail differently. Verdicts resting on **positive evidence** (a replayed `RemovedMetadataFieldType` for the field, a replayed `DeleteLedger` for the ledger) are correct at any skew and need only that the peer has folded up to the verified sequence. Verdicts resting on **absence** additionally require an exactly aligned cursor, because a field or ledger created *after* the pin is indistinguishable from a removed one in a view frozen before it. Malformed keys need no oracle and always run. `Check()` pins the peer snapshot next to the primary one to keep the aligned case reachable; the two are not atomic across two Pebble stores, which is why correctness rests on the split and not on the pin. The pass skips with an INFO log and never reports a clean result it did not establish.
- **Nil readstore is correct, not a gap.** Restore validation and CLI bootstrap validate a *staged main store* that has no peer readstore; the pass skips there. An **empty audit is not** a skip, though: the read index folds *from* the log stream, so a reverse-map row over a zero-log store is unaudited by definition, and `Check()`'s `lastSequence == 0` early return would otherwise hide exactly the tamper classes this pass reports.
- **Oracles never come from the data being verified.** The schema is the replayed one, never stored `LedgerInfo`; the live-ledger set is `knownLedgers`. Otherwise a tampered row could legitimise its own orphans.
- **Composition closes the tamper path.** This pass checks the rmap against the stored registry; `compareIndexes` independently checks that registry against the audit chain.

## Reviewer attention

**This amends invariant #8.** The peer-store carve-out previously read "out of main-store checker scope *by construction* since `Check()` never opens them". `Check()` now *does* open the peer readstore — read-only, one snapshot, for this single pass. `AGENTS.md` states the exception, why it is justified narrowly (`0x03` is the only limb with no other detector), that `usagestore` and readstore *contents* stay out, and that it must not be widened without the same argument. That is a deliberate architecture change and wants explicit sign-off.

**Two behaviour changes reach outside the reverse map**, both from unifying two ledger-name decoders that disagreed about the same fixed-width block:

1. `ParseBackfillKey` now rejects a NUL-corrupted ledger-name block instead of silently truncating it. Identical for all valid data.
2. `GetIndexStatus` / `GetIndex` can now return an error where they previously could not. `decodeBackfillProgress` used to `continue` past an unparsable key — a branch that was dead before this change and became reachable with (1). Propagating was chosen over skipping because the consumers populate `IndexEntry.Cursor`: dropping a corrupt entry makes the index falsely report cursor 0 / not-started, which is worse than an error. Reachable only under real disk corruption.

## Bugs found and fixed along the way

- **A NUL in an account address parsed successfully into a wrong-but-plausible `(entity, version, metaKey)` triple.** The terminator search stopped at the embedded NUL, so the real terminator and version bytes landed in the metadata key. Left unchecked this would have produced orphan findings against entities the key does not name. Now rejected, as an encoding check only — deliberately *not* via `invariants.Validate*`, so a future charset tightening cannot retroactively reclassify stored keys as corrupt.
- **A second, unreachable implementation of the schema rewrite.** `indexSetMetadataFieldType` could not run in production: its only caller is `indexLogEntry`, whose only caller gates on `isDataLog`, which never admits a `SetMetadataFieldType` log. The live path has always been `indexPayload` → `addSchemaRewriteTask` → `processSchemaRewrite`. It carried a latent `entityID` aliasing bug (a sub-slice of `iter.Key()` retained past many iterator moves) and a stale-batch trap, both of which an earlier revision of this PR fixed with 231 lines of new tests — all of it hardening dead code, and the tests passed only because they called `indexLogEntry` directly, bypassing the `isDataLog` gate. Deleted instead, so `processSchemaRewrite` is the single implementation. The live path was never at risk: it also routes through the cloning parser. `indexLogEntry`'s switch now documents why schema logs are absent.
- **A corruption signal routed into a silent skip.** The decoder unification made a previously-dead `continue` in `decodeBackfillProgress` reachable, converting a corrupt cursor into a silent drop — the exact invariant #7 pattern this ticket exists to eliminate. Now propagated.
- **Two silent-success paths in the rmap purge.** `purgeReverseMapForKey` never checked `iter.Error()`, so a mid-scan I/O failure was indistinguishable from exhaustion: the function returned `nil`, the batch committed, and the un-scanned rows were never deleted and never retried — manufacturing the exact permanent orphan this PR exists to detect. And `handleRemovedMetadataFieldType` returned `nil` on a nil write batch, skipping all three limbs at once (`0x01` and `0x02` have no detector of their own). Both now fail loudly; the nil-batch case mirrors what `bumpPendingVersion` already did for the same condition. `gcReverseMapVersion` got the same `iter.Error()` treatment.

## Known gap, deliberately not covered

`DropIndex` removes the registry entry but purges **no** readstore rows (`0x01`/`0x02`/`0x03`) and leaves the schema field declared, and `purgeOrphanVersions` cannot reclaim them because the version state is already gone. Tracked as **EN-1621**.

The schema term in the oracle is what keeps that residue out of this pass. A registry-only oracle would have been permanently red on any cluster that ever dropped a metadata index — `Check()` has no warning channel, so that includes its restore and bootstrap validation gates, and a checker that is red on a legitimate user action trains operators to ignore red.

**Consequence for EN-1621:** once `DropIndex` purges rows, a regression in that new purge would strand rows while the schema field is still declared, and this oracle would not catch it. The oracle must be revisited then. Recorded in the pass's doc comment, `AGENTS.md`, the indexer docs, and on EN-1621 itself.

## Testing

- `readstore`: parser round-trips (account, transaction, max-length ledger name), a 16-case rejection table asserting specific sentinels, an `EntityID`-independence pin, and `ParseBackfillKey`'s previously untested contract.
- `indexbuilder`: EN-1443 purge guards still green, plus a regression pinning that a missing write batch fails loudly instead of reporting a completed removal.
- `check`: 19 pass tests including aggregation, version-agnosticism, target-scoped identity over both oracle terms, the unknown-ledger and malformed classes, the lag gate, nil readstore, deterministic ordering, and both the `DropIndex`-residue-not-flagged and `RemovedMetadataFieldType`-residue-flagged scenarios. Plus an end-to-end test driving real orders through `Check()`.

Gates: `just pre-commit` 0 issues (both modules), `go build ./...` clean, full `just test` suite green.

## Docs

Invariant #8 in `AGENTS.md`; `audit-vs-technical-state.md`; the checker pass table, surface note, derivation diagram and *not verified* list; `indexer.md` / `indexes.md` handler tables and a checker-coverage section; the usage docs, which also asserted the carve-out.

Jira: https://formance-team.atlassian.net/browse/EN-1458
Follow-up: https://formance-team.atlassian.net/browse/EN-1621


flemzord (changes requested):
Reviewed live head `e39c2461`. Two correctness/integrity blockers remain in the new reverse-map pass:

1. `compareReverseMapOrphans` only skips when `indexedSequence < lastSequence`. `Check()` pins the primary snapshot near entry but opens the peer snapshot much later, so a busy indexer can present sequence N+1 while the audit/schema/index/live-ledger oracles are still pinned at N. Accepting that ahead cursor judges future rows against old oracles and can report a healthy newly-created ledger or metadata field as `REVERSE_MAP_ORPHAN`. Please require a sequence-aligned view (or skip whenever the peer cursor differs from the verified sequence) and cover the cursor-ahead race.

2. `Check()` returns early when `lastSequence == 0`, before this pass runs. A stale or tampered peer readstore containing malformed reverse-map keys or rows for unaudited ledgers therefore receives a clean result, despite those being explicit detection classes of this change. Please run the attached-peer validation for the audit-empty case as well and add a regression.

Functional CI is green. The `codecov/patch` failure is CI/tooling and is not the basis of this review.

internal/application/check/reverse_map_orphans.go
```diff
@@ -0,0 +1,470 @@
...
+		callback(errorEvent(
+			servicepb.CheckStoreErrorType_CHECK_STORE_ERROR_TYPE_REVERSE_MAP_ORPHAN,
+			fmt.Sprintf("reading the read-index progress cursor: %v", err),
+			0, "", "", "",
+		))
+
+		return
+	}
+
+	if indexedSequence < lastSequence {
```
paul-nicolas: [Major] The lag gate is one-sided: a read index *ahead* of the checker's snapshot produces false-positive orphans

The gate correctly handles the readstore lagging, but nothing bounds it from above, and the oracle it is compared against is frozen at a strictly earlier point in time.

`Check()` pins a main-store snapshot at `checker.go:109` and derives `lastSequence` from it at `checker.go:116`. Both oracle terms are therefore a view frozen at `lastSequence`: `collectIndexedFields` reads the registry from that same `snap` (passed as `reader`), and `replayedSchemas` comes from the replay that stopped there. But the readstore snapshot is taken later, at line 164, and line 178 only skips when `indexedSequence < lastSequence`.

`CheckStore` is a gRPC endpoint on a live node (`server_bucket.go:960-968`) that passes the live `impl.readStore`, and there is no `WaitForSequence` anywhere in the check package — writes keep flowing during `Check()`. So this interleaving is reachable on any busy cluster:

1. Checker snapshots at seq 100; `lastSequence = 100`.
2. Concurrently: `SetMetadataFieldType(foo)` at 101, `CreateIndex(foo)` at 102, metadata writes at 103.
3. The index builder folds to 103 and writes rmap rows for field `foo`.
4. Gate passes (`103 >= 100`).
5. The registry snapshot at 100 has no `foo` index; the replay to 100 has no `foo` schema field.
6. `orphan = !indexed && declared == nil` is **true** → an orphan event is emitted for a perfectly healthy cluster.

The same asymmetry hits the unknown-ledger class at line 238: a ledger created after `lastSequence` has rmap rows but is absent from `knownLedgers`, yielding a false "absent from the live ledger set" event.

This matters more than a typical false positive because of the argument this PR itself makes: `Check()` has no warning channel, it gates restore and bootstrap validation, and "a check that is permanently red on a legitimate operator action trains operators to ignore red." A check that goes red on ordinary concurrent index creation has the same effect.

No test covers the ahead direction. `TestCompareReverseMapOrphans_LagGate` (`checker_reverse_map_orphans_test.go:560`) exercises only `progress 9 < 10` and `progress 10 == 10`, and `TestCheck_ReverseMapOrphans_EndToEnd` sets `progress` to `1_000_000` with a comment stating the gate "never suppresses the pass" in that direction — so the assumption is baked in rather than tested.

Fix: make the orphan verdict independent of the two views' relative position instead of tightening the gate. Restrict the orphan class to fields the replay actually observed a `RemovedMetadataFieldType` for, rather than inferring it from schema absence. That is the condition the pass is really after ("the point-delete scan missed rows"), it excludes "this field does not exist *yet* in my frozen view," and it removes the dependency on the readstore and the main store being at the same sequence. If you prefer to keep the schema-absence oracle, the gate has to become `indexedSequence == lastSequence` — which on a busy cluster means the pass almost never runs, so it is the weaker option. Either way, add a test with `progress > lastSequence` and a field absent from both oracle terms.

internal/application/indexbuilder/process_logs.go
```diff
@@ -555,7 +554,36 @@ func (b *Builder) indexLogEntry(cfg *ledgerIndexConfig, log *commonpb.Log, propo
 		batch := b.readStore.NewBatch()
 		b.initBatch(batch) // re-init with a new batch after flush
 
-		return b.indexSetMetadataFieldType(cfg, b.kb, ledgerName, p.SetMetadataFieldType)
+		if err := b.indexSetMetadataFieldType(cfg, b.kb, ledgerName, p.SetMetadataFieldType); err != nil {
```
paul-nicolas: [Major] `indexSetMetadataFieldType` is unreachable in production — the aliasing fix, the batch-Cancel handling and 231 lines of new tests all harden dead code

The call chain is closed and I could not find a way in:

1. `indexSetMetadataFieldType` (defined line 1133) has exactly one non-test caller: this line.
2. The enclosing `indexLogEntry` (line 493) has exactly one non-test caller: `backfill.go:1321` (repo-wide grep, all packages).
3. That call is guarded at `backfill.go:1314` by `if !isDataLog(log) { ...; continue }`.
4. `isDataLog` (`backfill.go:1406-1430`) returns true only for `CreatedTransaction`, `RevertedTransaction`, `SavedMetadata`, `DeletedMetadata`, `OrderSkipped`; `SetMetadataFieldType` falls through to `default: return false`.
5. `isDataLog` switches on `applyLog.Apply.GetLog().GetData().GetPayload()` — the *identical* discriminator `indexLogEntry` switches on at line 538 — so there is no wrapping difference that could let it match another case.

So a `SetMetadataFieldType` log never reaches this branch. The live path is a separate implementation: `indexPayload` (line 328) case at line 344 defers to `addSchemaRewriteTask` (`backfill.go:252`), and the rewrite runs in `processSchemaRewrite` (`backfill.go:550`).

The new tests do not catch this because they call `b.indexLogEntry(...)` directly (`index_set_metadata_field_type_test.go:84`, `:166`, `:228`), bypassing the `isDataLog` gate that the only production caller applies. They pass while exercising a path production never takes.

Consequences for the PR description, which is worth correcting since it will become the commit message: the "latent aliasing bug in `indexSetMetadataFieldType`" and the "stale-batch trap" are defects in dead code, not production defects. The good news is that the aliasing hazard genuinely does not exist on the live path — `processSchemaRewrite` also routes through `parseScopedReverseMapKey` (`backfill.go:727`) and uses the cloned `EntityID` within a single iteration (`backfill.go:736-748`) — but that is by luck of a parallel implementation, not because of this change.

Also folded in here rather than posted separately: Codex flagged the missing `iter.Error()` check after the scan loop that collects `entries` (line 1183 onward). It is real but unreachable for the same reason, so it only matters if this function is wired up.

Fix: pick one. Either delete `indexSetMetadataFieldType` and its tests and keep `processSchemaRewrite` as the single schema-rewrite implementation, or wire it up deliberately (add `SetMetadataFieldType` to `isDataLog`, or call it from a path that reaches it) and add `iter.Error()` handling before relying on it. Two divergent implementations of the same rewrite, one of them dead and now the better-tested of the pair, is the outcome to avoid. Whichever you choose, `processSchemaRewrite` is the function that needs the coverage — it is what actually runs.

internal/application/indexbuilder/process_metadata_field_removal.go
```diff
@@ -165,8 +166,12 @@ func (b *Builder) purgeReverseMapForKey(kb *dal.KeyBuilder, ledgerName string, n
 	for iter.First(); iter.Valid(); iter.Next() {
```
paul-nicolas, line 175: [Medium] `purgeReverseMapForKey` never checks `iter.Error()` — a mid-scan failure strands rows and reports success

This loop (166-183) exits on `!iter.Valid()`, which is indistinguishable from normal exhaustion when the iterator fails mid-scan on an I/O error or a corrupt block. The function then proceeds to the overlay pass and returns `nil`, so `handleRemovedMetadataFieldType` (line 82) treats the field removal as complete and commits the batch. The un-scanned rmap rows are never deleted and never retried.

That is precisely the permanent-orphan condition this PR exists to detect, in the one function whose miss creates it. The new checker pass would report those rows forever, with the purge that should have removed them already reporting success — and, per the PR's own EN-1621 discussion, with no remediation path.

For calibration on severity: this is a pre-existing package-wide gap, not a regression here — no production site in `internal/application/indexbuilder` checks `iter.Error()` (0 of 3 `for iter.First()` loops). The new checker pass does check it (`reverse_map_orphans.go:281`), which is the right pattern. I am flagging this one site rather than the whole package because of what its silent failure produces.

Fix: after the loop and before the overlay pass, `if err := iter.Error(); err != nil { return fmt.Errorf("scanning rmap for purge: %w", err) }`. The same treatment is worth applying to the scan in `gc.go`.

internal/application/indexbuilder/process_metadata_field_removal.go
```diff
@@ -165,8 +166,12 @@ func (b *Builder) purgeReverseMapForKey(kb *dal.KeyBuilder, ledgerName string, n
...
 		k := iter.Key()
 
-		_, mk, _, parsed := parseReverseMapKey(k, rmapPrefix, ns)
-		if !parsed || mk != key {
+		rk, err := parseScopedReverseMapKey(k, ledgerName, ns)
+		if err != nil {
```
paul-nicolas, line 179: [Medium] Re: line 68 (just above this hunk) — `if batch == nil { return nil }` silently skips the entire three-limb purge

`handleRemovedMetadataFieldType` does `batch := b.wb.Batch()` at line 66 and returns `nil` on a nil batch at 68-70. That success return skips everything the function exists to do: the forward-index range delete (73), the entity-exists range delete (78), and the reverse-map point deletes (82).

A nil batch is impossible by design — every caller runs `initBatch` first — which makes this exactly the invariant-#7 shape the surrounding work is about: a should-not-happen branch that returns success instead of failing loudly. The cost of getting it wrong is asymmetric here, because the silent skip strands all three limbs at once, and `0x01`/`0x02` have no detector of their own. A refactor that drops an `initBatch` would produce a fully unindexed removal with no signal anywhere.

This is pre-existing code, outside the diff — I am raising it because it sits in the purge path this PR polices and because the PR already fixes a sibling silent-skip (the `continue` in `decodeBackfillProgress`) on exactly this reasoning. Worth closing in the same pass for consistency.

Fix: `return fmt.Errorf("invariant: no readstore write batch bound during RemovedMetadataFieldType for ledger %q field %q", ledgerName, key)`, and drop a regression test alongside the EN-1443 purge guards.

paul-nicolas (commented):
Multi-model review (Claude + Codex). 4 findings posted inline: 2 Major, 2 Medium. 3 Minor withheld (listed below). 6 Critical/Major/Medium candidates entered adversarial verification; 4 survived, 2 were rejected on inspection.

The design write-up is unusually thorough and the oracle reasoning (replayed schema, never stored `LedgerInfo`; conjunction with the registry; version deliberately ignored) holds up. The two Major findings are not about the oracle's logic — they are about its *inputs* being pinned at two different points in time, and about a large share of this PR's hardening landing on a branch production never executes.

Where the two models disagreed: Codex rated the missing `iter.Error()` sites Major and did not surface either Major below. I downgraded the `iter.Error()` findings to Medium after measuring the local convention — **no** production site in `internal/application/indexbuilder` checks `iter.Error()` (0 of 3 `for iter.First()` loops), so this is a pre-existing package-wide gap rather than a regression this PR introduces. It is still worth fixing in `purgeReverseMapForKey` specifically, because that is the one function whose miss manufactures the exact orphan this PR exists to detect. Codex's third `iter.Error()` finding (`process_logs.go:1183`) is folded into the dead-code finding rather than posted separately, since the enclosing function does not run.

Rejected after verification (not posted):
- `readstore/keys.go:132` — "ParseBackfillKey should validate kind-specific detail payloads": no consumer of `BackfillEntry` was shown to silently drop a corrupt kind, and payload validation does not belong in the envelope parser given the layering. Moved to the Minor list rather than dropped.
- `reverse_map_orphans.go:229` — "unbounded hex dump defeats the memory bound": the malformed bucket is keyed per-ledger, so the worst case is one sample per ledger, and an arbitrarily long key requires direct Pebble write access, at which point payload size is not the concern. Downgraded to Minor.

Verified locally at `e39c246`: `go vet` clean on the three touched packages, `go test ./internal/application/check/... -run ReverseMap` green.

- `internal/storage/readstore/keys.go:353` — the NUL guard on the decoded metadata key is incidental rather than structural. After a mis-split on an embedded NUL in the account address, `MetadataKey` becomes `[4 version bytes][real metaKey]`; the guard fires only because realistic versions (1, 2, 3…) contain zero bytes. A version ≥ `0x01010101` would evade it and yield a wrong-but-plausible triple — the exact failure mode the doc comment claims to close. Consider asserting the account address is NUL-free at the write site, which closes it by construction.
- `internal/storage/readstore/keys.go:132` — `ParseBackfillKey` validates only the shared envelope (name block + kind byte), not the kind-specific detail payload; a corrupt kind byte or empty detail still flows into `BackfillEntry`.
- `internal/application/check/reverse_map_orphans.go:229` — the malformed-key sample embeds the full key hex (`fmt.Sprintf("key %x: ...", key, err)`) in both the retained aggregate and the emitted event. Truncate to a fixed prefix plus the original length.

internal/application/check/reverse_map_orphans.go
paul-nicolas: Confirmed and fixed in d7a5550 (code in 969b968).

Verified the mechanics before acting: primary snapshot at `checker.go:109`, `lastSequence` at `:116`, pass invoked at the end of the replay, peer snapshot opened inside the pass, gate one-sided at `reverse_map_orphans.go:178`. `CheckStore` (`server_bucket.go:960`) constructs the checker with the live `impl.readStore` and there is no quiescing anywhere in the package, so the skew window is the whole runtime of the replay, not microseconds. The end-to-end test made it worse than you described: it set `progress = 1_000_000` and asserted orphans *are* reported, baking the ahead case in as expected-good.

Went with your first suggestion rather than the `==` gate, for the reason you gave — under load `lastSequence` advances constantly while the indexer lags, so strict alignment would trade the false positive for near-total loss of coverage. Verdicts are now split by the kind of evidence they rest on:

- **Positive evidence** — a replayed `RemovedMetadataFieldType` for the field, a replayed `DeleteLedger` for the ledger. Rests on a log the replay actually saw, so it holds at any skew; requires only that the peer has folded up to the verified sequence. This is EN-1458's real target.
- **Absence** — in neither the registry nor the replayed schema, or merely missing from `knownLedgers`. Infers removal from a pinned view, so it now requires exact alignment.
- **Malformed keys** — no oracle, always reported, lag gate included.

The removal set is a new replay map populated next to `removeExpectedSchemaField`; `deletedInReplay` already existed. It is append-only and the verdict consults the replayed schema first, so a field removed and later re-declared is not flagged.

Also pinned the peer snapshot next to the primary one in `Check()` instead of opening it at pass time, which keeps the aligned case reachable. Two snapshots across two Pebble stores are not atomic — that is precisely why correctness rests on the evidence split and not on the cursors matching, and the doc comment says so.

Tests: cursor-ahead cases for both the field and the ledger class, the positive-evidence twins that must still fire under the same ahead cursor, removed-then-redeclared, and a malformed key under a lagging cursor. The end-to-end test keeps its ahead cursor and now proves the positive-evidence path.

internal/application/indexbuilder/process_logs.go
paul-nicolas: Confirmed and deleted in e1f6135.

Re-verified every link independently: `indexSetMetadataFieldType` had exactly one non-test caller (`indexLogEntry`); `indexLogEntry` has exactly one non-test caller (`backfill.go:1321`); that call is gated at `:1314` on `isDataLog`, which admits only `CreatedTransaction` / `RevertedTransaction` / `SavedMetadata` / `DeletedMetadata` / `OrderSkipped` and switches on the same discriminator `indexLogEntry` does. One extra corroboration you did not mention: `indexPayload` also has a `RemovedMetadataFieldType` case that `indexLogEntry` lacks, so the two dispatchers are deliberately different and only the live one handles schema logs.

Chose deletion over wiring up. Backfill genuinely does not need the case: a retype landing mid-backfill is handled by `addSchemaRewriteTask` resetting the in-flight cursor so the replay re-encodes under the new declared type via the per-batch `schemaResolver` — not by replaying the schema log. Wiring it up would have meant keeping two implementations of the same rewrite with the dead one better tested, which is the outcome you flagged.

Removed: the `SetMetadataFieldType` case in `indexLogEntry` (including the batch Flush/Cancel/Reset handling), `indexSetMetadataFieldType` itself, and `index_set_metadata_field_type_test.go`. That also disposes of the `process_logs.go:1183` missing `iter.Error()` you folded in here. `indexLogEntry`'s switch now carries a comment stating why schema logs are absent and that `isDataLog` would have to change alongside adding one back, so this does not get helpfully re-added.

You are right that the PR description needed correcting — the "latent aliasing bug" and "stale-batch trap" bullets described defects in dead code and would have become the merge commit message. Updated. Also confirmed your read that the live path was never at risk: `processSchemaRewrite` routes through `parseScopedReverseMapKey` and uses the cloned `EntityID` within a single iteration.

internal/application/indexbuilder/process_metadata_field_removal.go
paul-nicolas: Fixed in bf1f745, with the `gc.go` scan given the same treatment as you suggested.

Confirmed the audit: of the iterator loops in non-test `internal/application/indexbuilder` code — `gc.go:101`, `backfill_postings.go:86`, `backfill.go:670`/`:678`/`:711`, `process_metadata_field_removal.go:166` and `process_logs.go:1183` — none checked `iter.Error()`, so your calibration as a pre-existing package-wide gap rather than a regression is right.

One thing worth recording, since it was the open risk on this fix: returning an error here does **not** wedge the builder. It propagates `purgeReverseMapForKey` → `handleRemovedMetadataFieldType` → `indexPayload` → `process_logs.go:201`, which does `_ = batch.Cancel(); return cursor, err`. The batch is discarded and the fold resumes from the persisted cursor, so a transient I/O error retries the log instead of stranding it — strictly better than the silent success, not a trade.

Scoped to the two loops you argued for rather than all six: these are the ones whose silent truncation manufactures the orphan the new pass reports. The remaining four are worth a follow-up rather than widening this PR — `process_logs.go:1183` disappeared with the dead function.

internal/application/indexbuilder/process_metadata_field_removal.go
paul-nicolas: Fixed in bf1f745.

Your case is stronger than the comment argues, and that decided it. There are exactly two `batch == nil` sites in the package, and the other one — `bumpPendingVersion` at `backfill.go:365-372` — already returns `errors.New("invariant: bumpPendingVersion called without an active write batch")` with a comment explicitly citing CLAUDE.md invariant #7. So this was not a style preference to weigh; the codebase had already settled it and `process_metadata_field_removal.go:68` was the outlier. Mirrored it, comment included, and added a regression test next to the EN-1443 purge guards asserting the error rather than the old silent success.

paul-nicolas:
@flemzord both blockers confirmed and fixed. Head is now `d7a5550`.

**1. Cursor-ahead race.** Confirmed exactly as you described: the primary snapshot is pinned at `checker.go:109` and `lastSequence` derived at `:116`, while the pass opened its own peer snapshot at the end of the replay and gated only on `indexedSequence < lastSequence`. `CheckStore` (`server_bucket.go:960`) passes the live `impl.readStore` and nothing quiesces, so the window was the full runtime of the replay.

I did not take the "skip whenever the peer cursor differs" option, and want to be explicit about why since it was your suggested remedy. Under load `lastSequence` advances continuously while the indexer lags, so requiring exact equality would mean the pass almost never runs — it would trade a false positive for near-total loss of the detection EN-1458 exists to add. Instead the verdicts are separated by the kind of evidence each rests on:

- **Positive evidence** — a replayed `RemovedMetadataFieldType` for the field, a replayed `DeleteLedger` for the ledger — rests on a log the replay actually saw, so it is correct at *any* skew. It requires only that the peer has folded up to the verified sequence.
- **Absence** — in neither the registry nor the replayed schema, or merely missing from `knownLedgers` — infers removal from a view pinned at `lastSequence`, so it now requires the **exactly aligned** view you asked for.
- **Malformed keys** need no oracle and are always reported.

That gives your sequence-aligned requirement for precisely the verdicts that need it, without disabling the pass on busy clusters. I also moved the peer snapshot pin next to the primary one in `Check()` so the aligned case stays reachable; the two cannot be taken atomically across two Pebble stores, which is why correctness rests on the split rather than on the pin.

Covered by new tests: cursor-ahead for both the field and the ledger class (healthy rows, must stay silent), the positive-evidence twins that must still fire under the same ahead cursor, and removed-then-redeclared. The end-to-end test previously set `progress = 1_000_000` and asserted orphans *are* found — it had baked the ahead case in as expected-good. It keeps the ahead cursor and now proves the positive-evidence path instead.

**2. Audit-empty case.** Confirmed: `Check()` returned at `checker.go:121-131` before any pass ran. The pass now runs in that branch with empty oracles — correct by construction, since the read index folds *from* the log stream, so a reverse-map row over a zero-log store is unaudited by definition. The early return itself stays, as every other pass needs the replay. Regressions added at both the fixture and `Check()` level, plus the negative twin (attached-but-empty readstore must stay clean — only the nil-readstore case was covered before).

**Unrelated but found in the same pass, and it changes what this PR claims:** `indexSetMetadataFieldType` was unreachable in production — `indexLogEntry`'s only caller gates on `isDataLog`, which never admits a schema log. The aliasing fix and stale-batch handling this branch added were therefore hardening dead code. Deleted it; `processSchemaRewrite` remains the single schema-rewrite implementation, and the PR description is corrected accordingly.

Gates: `just pre-commit` 0 issues on both modules, full `just test` suite green.

paul-nicolas:
Addendum: head has moved to `722d0d9a` since the comment above. The extra commit bounds the malformed-key hex dump in the two reverse-map findings to a 64-byte prefix plus the original length — a self-review nit from the same pass, no behaviour change to the oracle or the gate. `just pre-commit` still 0 issues on both modules.

flemzord (changes requested):
Revalidated live head `722d0d9a`. One correctness blocker remains in the new evidence split: the supposed positive evidence is not monotonic. `checker.go` keeps both `deletedInReplay` and `removedSchemaFields` append-only; a later `CreateLedger` restores `knownLedgers` but does not clear `deletedInReplay`, while `compareReverseMapOrphans` checks the deleted set before the live set. Therefore `DeleteLedger` followed by `CreateLedger` of the same name makes healthy reverse-map rows of the recreated ledger report as `REVERSE_MAP_ORPHAN` even with an aligned cursor. The same issue appears with an ahead peer: a ledger or field recreated/redeclared after the primary snapshot is invisible to the pinned oracle, while the earlier deletion/removal still triggers the positive branch. Please require alignment for these verdicts or make the evidence generation-aware, and add delete/recreate plus remove/redeclare-after-snapshot regressions. The remaining pending CI is not the basis of this review.

flemzord:
Correction to my review on `722d0d9a`: I withdraw the `DeleteLedger` then same-name `CreateLedger` example; the retained ledger tombstone makes that lifecycle unreachable. The blocker remains for the field case: `RemovedMetadataFieldType` is present in the replay through N, the field is re-declared after the primary snapshot, and an ahead readstore contains the new legitimate rows. Because `removedFields` is append-only while `replayedSchemas` is pinned at N, `case removed && folded` still reports those healthy rows as orphans. Please treat the requested change as scoped to re-declaration after the snapshot (requiring alignment or generation-aware evidence), not ledger-name recreation.

paul-nicolas:
@flemzord confirmed, and fixed in `0d89a40` — but not by either of the two remedies you offered, so let me be explicit about why.

Your narrowed case reproduces exactly as you describe. On `722d0d9a` the verdict was `case removed && folded` with `folded := indexedSequence >= lastSequence`: a `RemovedMetadataFieldType` at or before N leaves `removed` set forever, `expectedSchemas` is pinned at N so a re-declaration past the pin gives `declared == nil`, and an ahead peer holds the new legitimate rows. Healthy rows, reported as `REVERSE_MAP_ORPHAN`. Your withdrawal of the ledger-recreation half is also right, and I took the ordering hazard behind it seriously anyway — see below.

**Both of your options accept "the peer cursor can be ahead" as a premise. It isn't one — it was an artefact of the order in which `Check()` took its two snapshots.** The index builder folds *from* the primary log stream: `processLogs` iterates primary `Log` rows and writes `WriteProgress(batch, lastSeq)` for logs it has already read out of the primary store, in the same readstore batch. So `progress(t) <= maxLogSeq(t)` holds at every instant, and both are monotonically non-decreasing. On `722d0d9a` the primary snapshot was pinned first (`checker.go:109`) and the peer second (`:125`), so `indexedSequence` was sampled *later* than `lastSequence` and could exceed it. Reversing the two pins gives

```
indexedSequence = progress(t_peer) <= maxLogSeq(t_peer) <= maxLogSeq(t_primary) = lastSequence
```

so ahead becomes impossible by construction. One statement moved.

That is a strictly better trade than requiring alignment for the positive path, which is what your first option amounts to: it keeps the aligned case exactly as reachable as before instead of narrowing it, and it needs no post-snapshot log reads, which is what generation-aware evidence would have required (and those reads would sit outside this run's verified range — a suppression channel I did not want to add). The pass now needs no cross-store snapshot atomicity at all: an ordering that can only leave the peer *behind* is sufficient, because behind was already a skip.

With ahead unreachable, the whole positive/absence evidence split had nothing left to protect, so **this commit is a net deletion**: the `folded` tier, the two-tier verdict switch, the `deletedLedgers` scope term and ~40 lines of the doc block that justified them are gone, replaced by one `indexedSequence == lastSequence` gate. Malformed keys still report unconditionally.

**On the ledger class you withdrew.** You were right that the tombstone makes same-name recreation unreachable, but the ordering you flagged — deleted-set checked before live-set — meant the code's correctness depended on that external guarantee rather than on its own structure. `deletedLedgers` is subsumed by `!liveLedgers` anyway (`checker.go:518` removes the name on the same `DeleteLedger` that `:526` records it), so I dropped it and the verdict now rests on liveness alone. A recreated ledger is live again, so its rows are legitimate by construction. Regression: `TestCompareReverseMapOrphans_RecreatedLedgerStaysSilent`.

**Residual ahead, handled explicitly rather than assumed away.** A primary-store rollback beneath the cursor (`RestoreCheckpoint` on a follower restore) does lower `maxLogSeq` while the read index keeps its progress. Unlike `usagebuilder`, which has `resetIfRolledBack` → `usagestore.Reset()`, the index builder has **no** rollback reset — so in that state the read index genuinely holds rows for logs the restored primary never had. It is logged loudly and skipped, not reported: that divergence belongs to the missing reset, not to the purge path this pass audits, and reporting it would paint `Check()` red on a restore that never self-heals. **That missing reset looks like a real pre-existing gap** — it predates this branch and is independent of this pass — so I have deliberately left it out of this PR rather than widen the diff. Say the word and I will open a ticket for it.

**Regressions.** Your exact scenario is `TestCompareReverseMapOrphans_CursorAhead`'s new `redeclaredAfterPin` case (removed at ≤N, re-declared past the pin, ahead cursor, live rows → silent), paired with an aligned twin on the same removal evidence and rows that must still fire, so the silence is provably earned by the gate and not by the fixture. `TestCheck_ReverseMapOrphans_EndToEnd` is now a subtest pair driving both cursor positions through a real `Check()`; its aligned arm reads the store's own last sequence instead of the `progress = 1_000_000` it used to hard-code — that value had baked the ahead case in as expected-good. I verified both new assertions fail against the previous logic before keeping them.

Gates: `just pre-commit` 0 issues on both modules, no codegen drift, full `just test` green.

flemzord:
Revalidated live head 0d89a40e. The previous reverse-map false-positive blocker is fixed: the peer snapshot is pinned before the primary snapshot, orphan verdicts are now limited to an exactly aligned cursor, and a field present again in the replayed schema suppresses the earlier removal evidence. I found no remaining correctness/runtime blocker on this head. The codecov/patch failure is CI/tooling only; GitHub approval is intentionally disabled by this cron configuration.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1649#issuecomment-5131021632

internal/application/check/reverse_map_orphans.go
```diff
@@ -0,0 +1,598 @@
...
+				// removal set safe to keep append-only.
+				orphan = false
+			case aligned:
+				// The replay observed the removal, OR the field is in neither
+				// oracle (a removal whose log was archived away, or a field the
+				// audit never declared at all). The removal term is kept even
+				// though absence subsumes it on a healthy store: it still fires
+				// when a RemovedMetadataFieldType is on record but the registry
+				// entry lingers, which absence alone would miss.
+				_, indexed := indexedFields[indexes.KeyFor(parsed.Ledger, indexes.MetadataID(target, parsed.MetadataKey))]
```
NumaryBot, line 383: 🟠 **[major] Archived field-removal evidence can be masked by stale registry rows, suppressing orphan detection**

When a field removal falls in an archived chapter, `removedFields` no longer retains the positive evidence for that removal. A surviving (or tampered) `SubAttrIndex` row then causes `indexed == true`, suppressing the orphan verdict. This gap is not closed by `compareIndexes` because that function deliberately skips stored index entries that have no replay activity when archives are present. As a result, a stale registry row and rmap orphan rows can pass `Check()` together without triggering any finding.

**Suggestion:** For fields whose removal evidence is in archived chapters, derive the expected schema state from the archived snapshots rather than relying solely on the live replay's `removedFields`. Alternatively, cross-check the replayed schema against the presence of rmap rows independently of the `indexed` flag when archive gaps exist, to ensure orphan rows are not silently skipped.

flemzord (changes requested):
Revalidated live head `0d89a40e` after the new NumaryBot finding. I confirm that finding as a correctness/integrity blocker: when `RemovedMetadataFieldType` is behind the archive boundary, the replay has no `removedFields` evidence; a stale or tampered `SubAttrIndex` row then makes `indexed` true and suppresses the aligned reverse-map orphan verdict, while `compareIndexes` separately skips that same unmatched registry row because archives exist and there was no replay activity. Those two corrupted projections can therefore mask each other and let `Check()` return clean, contrary to invariant #8. I am not duplicating the inline details. Please derive sufficient archive-bound schema/index evidence, or otherwise ensure stale registry presence cannot legitimize orphaned reverse-map rows.

internal/application/check/reverse_map_orphans.go
paul-nicolas: Confirmed as a real masking channel, and fixed in `cf71f5f7` — but the mechanism is not the one described, so the fix is not in this pass.

**`removedFields` losing its entry under archiving is not the weak link.** `expectedSchemas` is already seeded from the archived baseline snapshot (`checker.go:436` → `seedExpectedFromBaseline`), and `RemovedMetadataFieldType` apply deletes the field from `LedgerInfo.MetadataSchema` (`processor_metadata_schema.go:110`). The baseline is a live-state snapshot taken at the boundary, so a pre-boundary removal leaves the field absent from the seeded schema and `declared == nil` is reached correctly. Your primary suggestion — derive the expected schema state from the archived snapshots — was already in place, and absence in the schema subsumes the lost removal evidence.

**The load-bearing term is the registry conjunction, and the defect is in `compareIndexes`.** `checker.go:1043` skipped any stored entry with no replay activity whenever archives existed, so the stale `SubAttrIndex` row went unreported and made `indexed == true` legitimise the rows. That is where the two corruptions masked each other, and your conclusion holds exactly as written.

**I did not take the alternative you offered** (cross-check the schema independently of the `indexed` flag). `validateIndexTarget` (`processor_index.go:72`) requires the metadata field to be declared before an index on it can be created, so the conjunction is redundant today — but dropping it would report *healthy* rmap rows as orphans if that guarantee ever broke. The inconsistency worth reporting is the registry entry, not the rows.

**Instead the tolerance is gone.** The baseline checkpoint already carries the registry — `attributes/baseline.go:94` copies the whole `ZoneAttributes` range verbatim, and `SubAttrIndex` is an attribute sub-prefix — so `expectedIndexes` is now seeded from it via `foldBaselineIndexes`. Expected becomes baseline + replay delta and an unmatched stored entry is a hard mismatch with or without archives, which is the same treatment every other projection in this checker already gets (schema, account types, references, numscripts, reversions, boundaries). Two consequences worth noting:

- It closes the same hole for **builtin** index kinds (tx / account / log builtin), which the narrower metadata-only cross-check would have left open.
- Only ledgers the baseline lists as live are seeded. Rows of a ledger deleted before the boundary await the deferred purge, and expecting them would turn that normal cleanup into a phantom "registry has no matching row" event — a false positive the naive seed would have introduced.

`indexReplayActivity` existed solely to drive the tolerance and is removed with it; `compareIndexes` now takes a `compareIndexesScope` input struct.

**Regressions.** `TestCompareReverseMapOrphans_StaleRegistryCannotMaskOrphans` pins both halves of the scenario: the registry term still suppresses the orphan verdict (unchanged, by design), and that same stale entry is now reported as `INDEX_MISMATCH` — so `Check()` is no longer clean on a store carrying both corruptions. `TestFoldBaselineIndexes_SeedSilencesArchivedEntry` covers the legitimate archive case plus the liveness filter. `TestCompareIndexes_ArchiveOrphanIgnored` asserted the old tolerance and is inverted into `TestCompareIndexes_UnseededEntryFlaggedUnderArchiving`; `TestCompareIndexes_PostArchiveDropFlagged` tested a replayActivity distinction that no longer exists and was folded into it (its assertion is still covered by `TestCompareIndexes_ExtraInStored`).

Gates: `just pre-commit` 0 issues on both modules, full `just test` suite green.

paul-nicolas:
@flemzord confirmed and fixed in `cf71f5f7`. Details are in the [thread reply](https://github.com/formancehq/ledger/pull/1649#discussion_r3682789251); the short version, against your specific requirement:

> Please derive sufficient archive-bound schema/index evidence, or otherwise ensure stale registry presence cannot legitimize orphaned reverse-map rows.

The **schema** half already existed — `expectedSchemas` is baseline-seeded (`checker.go:436`) and `RemovedMetadataFieldType` apply deletes the field from `LedgerInfo.MetadataSchema`, so an archived removal correctly yields `declared == nil`. NumaryBot's premise that `removedFields` is the archive-sensitive term is wrong; absence in the schema subsumes it.

The **index** half was genuinely missing, and that is the whole defect. `compareIndexes` skipped any stored entry with no replay activity whenever archives existed, so the stale `SubAttrIndex` row was never reported and `indexed == true` legitimised the rows. Your framing ("schema **or index** evidence") was the accurate one.

Fixed by deriving it rather than tolerating the gap: the baseline checkpoint already carries the registry (`attributes/baseline.go:94` copies the whole `ZoneAttributes` range verbatim), so `expectedIndexes` is seeded from it via a new `foldBaselineIndexes` and the archive-orphan tolerance is **deleted** rather than narrowed. Expected = baseline + replay delta, and an unmatched stored entry is a hard mismatch with or without archives — the same treatment schema, account types, references, numscripts, reversions and boundaries already get.

Two notes on the choice:

- I did **not** drop the reverse-map registry conjunction, which was NumaryBot's alternative. `validateIndexTarget` enforces indexed-implies-declared, so dropping it would report healthy rows if that guarantee ever broke. The thing to report is the registry entry, not the rows — so `compareReverseMapOrphans` is unchanged and the masking is closed one layer down.
- Deleting the tolerance rather than adding a metadata-only schema cross-check also closes the same hole for **builtin** index kinds, which the narrower fix would have left open.

One false positive the naive seed would have introduced, and the reason for the liveness filter: a ledger deleted before the boundary still has its index rows in the snapshot while the deferred purge is pending, so only ledgers the baseline lists as live are seeded — otherwise a normal cleanup would surface as a phantom missing-row event.

`TestCompareReverseMapOrphans_StaleRegistryCannotMaskOrphans` pins the exact scenario end to end: the registry term still suppresses the orphan verdict, and that same stale entry is now reported as `INDEX_MISMATCH`, so the two corrupted projections can no longer both pass. `just pre-commit` 0 issues on both modules, full `just test` green.

This does land in pre-existing `compareIndexes` code rather than in the new pass — deliberately, since that is where the defect is and the blocker only closes if the stale row itself becomes reportable.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1649#issuecomment-5131021632

flemzord:
Revalidated live head `cf71f5f7`. The new baseline-seeded expected index registry removes the archive-orphan tolerance that allowed a stale `SubAttrIndex` row to mask orphaned reverse-map rows, and the current regressions cover that case. I found no remaining correctness/runtime blocker. The `codecov/patch` failure is CI/tooling rather than a demonstrated runtime defect. GitHub approval is intentionally disabled by the cron configuration.

Azorlogh (dismissed):

flemzord (dismissed):

Azorlogh (approved):

PR #1651, opened 2026-07-29
fix(security): redact event sink read credentials

flemzord:
## What changed

- redact reusable event-sink credentials before gRPC and HTTP read responses are serialized
- deep-clone the read projection so runtime sink configuration remains unchanged
- handle ClickHouse userinfo and credential query parameters with URL-aware parsing
- redact NATS password/token userinfo, including comma-separated server lists, and HTTP endpoint userinfo
- make opaque and URL redaction idempotent across leader forwarding and CLI compatibility passes
- share the implementation with `ledgerctl` and document the field-level API contract
- add direct primitive, projection, transport, and CLI regression coverage

## Root cause

`GetEventsSinks` returned the controller's full `SinkConfig` under `ledger:OpsRead`. Because the default `ledger:read` mapping includes that scope, ordinary read callers could retrieve reusable credentials. CLI masking happened only after the server response.

The original redaction pass also relied on string slicing for DSNs and NATS URLs. It missed ClickHouse query-string passwords, could misparse `@` outside userinfo, did not cover HTTP endpoint userinfo, and was not a fixed point when routed or CLI paths redacted a response twice.

## Security behavior

- Kafka SASL passwords, HTTP HMAC secrets, Databricks PATs, and Databricks OAuth client secrets become `(set)` or `(none)`.
- ClickHouse userinfo passwords and credential-bearing DSN query values become `****`.
- NATS and HTTP URL userinfo passwords or tokens become `****`.
- Usernames, client IDs, hosts, paths, topics, and sink statuses remain visible.
- Runtime/controller-owned configuration is not mutated.

Historical copies in system logs and audit records are a distinct persistence/read boundary and are handled independently by draft PR [#1654](https://github.com/formancehq/ledger/pull/1654); this PR deliberately does not duplicate that projection.

## Validation

- `nix develop --command bash -c "GOROOT= go test ./internal/pkg/redact ./internal/adapter/eventsink ./internal/adapter/grpc ./internal/adapter/http ./cmd/ledgerctl/cmdutil ./cmd/ledgerctl/events"`
- GitNexus index refreshed on the PR head
- GitNexus `detect_changes(scope: compare, base_ref: release/v3.0)`: LOW aggregate, expected files, no affected execution flows

## Tracking

- [EN-1632](https://formance-team.atlassian.net/browse/EN-1632)
- Historical-read companion: [EN-1634](https://formance-team.atlassian.net/browse/EN-1634) / [#1654](https://github.com/formancehq/ledger/pull/1654)

[EN-1632]: https://formance-team.atlassian.net/browse/EN-1632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/adapter/eventsink/redact.go
+++ b/internal/adapter/eventsink/redact.go
@@ -62,6 +62,7 @@ func redactConfigInPlace(cfg *commonpb.SinkConfig) {
 		}
 	case *commonpb.SinkConfig_Http:
 		if sink.Http != nil {
+			sink.Http.Endpoint = redact.URLUserinfo(sink.Http.GetEndpoint())
 			sink.Http.Secret = redact.Secret(sink.Http.GetSecret())
 		}
 	case *commonpb.SinkConfig_Databricks:
@@ -77,31 +78,9 @@ func redactNATSURLs(value string) string {
 	servers := strings.Split(value, ",")
 	for index, server := range servers {
 		trimmed := strings.TrimSpace(server)
-		schemeEnd := strings.Index(trimmed, "://")
-		if schemeEnd == -1 {
-			continue
-		}
-
-		rest := trimmed[schemeEnd+3:]
-		userinfoEnd := strings.LastIndex(rest, "@")
-		if userinfoEnd == -1 {
-			continue
-		}
-
-		userinfo := rest[:userinfoEnd]
-		if userinfo == "" {
-			continue
-		}
-
-		if username, _, hasPassword := strings.Cut(userinfo, ":"); hasPassword {
-			userinfo = username + ":" + redact.SecretSet
-		} else {
-			userinfo = redact.SecretSet
-		}
-
 		leadingSpace := server[:len(server)-len(strings.TrimLeft(server, " \t"))]
 		trailingSpace := server[len(strings.TrimRight(server, " \t")):]
-		servers[index] = leadingSpace + trimmed[:schemeEnd+3] + userinfo + rest[userinfoEnd:] + trailingSpace
+		servers[index] = leadingSpace + redact.URLUserinfo(trimmed) + trailingSpace
 	}
 
 	return strings.Join(servers, ",")
--- a/internal/adapter/eventsink/redact.go
+++ b/internal/adapter/eventsink/redact.go
@@ -1,0 +1,77 @@
+// Package eventsink provides the read-side projection of event-sink
+// configuration. Runtime configuration retains the original credentials;
+// externally serialized read models expose only redaction markers.
+package eventsink
+
+import (
+	"google.golang.org/protobuf/proto"
+
+	"github.com/formancehq/ledger/v3/internal/pkg/redact"
+	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
+	"github.com/formancehq/ledger/v3/internal/proto/servicepb"
+)
+
+// RedactConfig returns a deep clone of cfg with every credential-bearing field
+// replaced by a non-reusable marker. The supplied runtime configuration is not
+// mutated.
+func RedactConfig(cfg *commonpb.SinkConfig) *commonpb.SinkConfig {
+	if cfg == nil {
+		return nil
+	}
+
+	cloned, _ := proto.Clone(cfg).(*commonpb.SinkConfig)
+	redactConfigInPlace(cloned)
+
+	return cloned
+}
+
+// RedactGetResponse returns a deep clone of resp whose sink configurations are
+// safe to return through read-scoped APIs. Status records are cloned unchanged.
+func RedactGetResponse(resp *servicepb.GetEventsSinksResponse) *servicepb.GetEventsSinksResponse {
+	if resp == nil {
+		return nil
+	}
+
+	cloned, _ := proto.Clone(resp).(*servicepb.GetEventsSinksResponse)
+	for _, sink := range cloned.GetSinks() {
+		redactConfigInPlace(sink)
+	}
+
+	return cloned
+}
+
+func redactConfigInPlace(cfg *commonpb.SinkConfig) {
+	if cfg == nil {
+		return
+	}
+
+	switch sink := cfg.GetType().(type) {
+	case *commonpb.SinkConfig_Clickhouse:
+		if sink.Clickhouse != nil {
+			sink.Clickhouse.Dsn = redact.DSN(sink.Clickhouse.GetDsn())
+		}
+	case *commonpb.SinkConfig_Kafka:
+		if sink.Kafka != nil {
+			sink.Kafka.SaslPassword = redact.Secret(sink.Kafka.GetSaslPassword())
+		}
+	case *commonpb.SinkConfig_Http:
+		if sink.Http != nil {
+			sink.Http.Secret = redact.Secret(sink.Http.GetSecret())
+		}
+	case *commonpb.SinkConfig_Databricks:
+		if sink.Databricks != nil {
+			redactDatabricksAuthInPlace(sink.Databricks)
+		}
+	}
+}
+
+func redactDatabricksAuthInPlace(cfg *commonpb.DatabricksSinkConfig) {
+	switch auth := cfg.GetAuth().(type) {
+	case *commonpb.DatabricksSinkConfig_Token:
+		auth.Token = redact.Secret(auth.Token)
+	case *commonpb.DatabricksSinkConfig_OauthM2M:
+		if auth.OauthM2M != nil {
+			auth.OauthM2M.ClientSecret = redact.Secret(auth.OauthM2M.GetClientSecret())
+		}
+	}
+}
--- a/internal/adapter/eventsink/redact.go
+++ b/internal/adapter/eventsink/redact.go
@@ -4,6 +4,8 @@
 package eventsink
 
 import (
+	"strings"
+
 	"google.golang.org/protobuf/proto"
 
 	"github.com/formancehq/ledger/v3/internal/pkg/redact"
@@ -54,6 +56,10 @@ func redactConfigInPlace(cfg *commonpb.SinkConfig) {
 		if sink.Kafka != nil {
 			sink.Kafka.SaslPassword = redact.Secret(sink.Kafka.GetSaslPassword())
 		}
+	case *commonpb.SinkConfig_Nats:
+		if sink.Nats != nil {
+			sink.Nats.Url = redactNATSURLs(sink.Nats.GetUrl())
+		}
 	case *commonpb.SinkConfig_Http:
 		if sink.Http != nil {
 			sink.Http.Secret = redact.Secret(sink.Http.GetSecret())
@@ -63,6 +69,42 @@ func redactConfigInPlace(cfg *commonpb.SinkConfig) {
 			redactDatabricksAuthInPlace(sink.Databricks)
 		}
 	}
+}
+
+// redactNATSURLs removes password and token userinfo from the comma-separated
+// server list accepted by the NATS client while preserving server addresses.
+func redactNATSURLs(value string) string {
+	servers := strings.Split(value, ",")
+	for index, server := range servers {
+		trimmed := strings.TrimSpace(server)
+		schemeEnd := strings.Index(trimmed, "://")
+		if schemeEnd == -1 {
+			continue
+		}
+
+		rest := trimmed[schemeEnd+3:]
+		userinfoEnd := strings.LastIndex(rest, "@")
+		if userinfoEnd == -1 {
+			continue
+		}
+
+		userinfo := rest[:userinfoEnd]
+		if userinfo == "" {
+			continue
+		}
+
+		if username, _, hasPassword := strings.Cut(userinfo, ":"); hasPassword {
+			userinfo = username + ":" + redact.SecretSet
+		} else {
+			userinfo = redact.SecretSet
+		}
+
+		leadingSpace := server[:len(server)-len(strings.TrimLeft(server, " \t"))]
+		trailingSpace := server[len(strings.TrimRight(server, " \t")):]
+		servers[index] = leadingSpace + trimmed[:schemeEnd+3] + userinfo + rest[userinfoEnd:] + trailingSpace
+	}
+
+	return strings.Join(servers, ",")
 }
 
 func redactDatabricksAuthInPlace(cfg *commonpb.DatabricksSinkConfig) {
--- a/internal/adapter/grpc/server_bucket.go
+++ b/internal/adapter/grpc/server_bucket.go
@@ -22,6 +22,7 @@ import (
 	logging "github.com/formancehq/go-libs/v5/pkg/observe/log"
 
 	internalauth "github.com/formancehq/ledger/v3/internal/adapter/auth"
+	"github.com/formancehq/ledger/v3/internal/adapter/eventsink"
 	"github.com/formancehq/ledger/v3/internal/application/check"
 	"github.com/formancehq/ledger/v3/internal/application/ctrl"
 	"github.com/formancehq/ledger/v3/internal/domain"
@@ -1149,10 +1150,10 @@ func (impl *BucketServiceServerImpl) GetEventsSinks(ctx context.Context, _ *serv
 		return nil, fmt.Errorf("loading events sinks: %w", err)
 	}
 
-	return &servicepb.GetEventsSinksResponse{
+	return eventsink.RedactGetResponse(&servicepb.GetEventsSinksResponse{
 		Sinks:        sinks,
 		SinkStatuses: statuses,
-	}, nil
+	}), nil
 }
 
 func (impl *BucketServiceServerImpl) GetChapterSchedule(ctx context.Context, _ *servicepb.GetChapterScheduleRequest) (*servicepb.GetChapterScheduleResponse, error) {
--- a/internal/pkg/redact/redact.go
+++ b/internal/pkg/redact/redact.go
@@ -1,0 +1,42 @@
+package redact
+
+import "strings"
+
+const (
+	SecretSet  = "(set)"
+	SecretNone = "(none)"
+)
+
+// Secret replaces an opaque secret with a marker that only exposes whether a
+// value is configured.
+func Secret(value string) string {
+	if value == "" {
+		return SecretNone
+	}
+
+	return SecretSet
+}
+
+// DSN replaces the password in a URL-shaped DSN with "****" while preserving
+// the username and host for operator diagnostics. Non-URL DSNs and DSNs without
+// a password are returned unchanged.
+func DSN(dsn string) string {
+	schemeEnd := strings.Index(dsn, "://")
+	if schemeEnd == -1 {
+		return dsn
+	}
+
+	rest := dsn[schemeEnd+3:]
+	lastAt := strings.LastIndex(rest, "@")
+	if lastAt == -1 {
+		return dsn
+	}
+
+	creds := rest[:lastAt]
+	user, _, ok := strings.Cut(creds, ":")
+	if !ok {
+		return dsn
+	}
+
+	return dsn[:schemeEnd+3] + user + ":****" + rest[lastAt:]
+}
--- a/internal/pkg/redact/redact.go
+++ b/internal/pkg/redact/redact.go
@@ -1,42 +1,117 @@
+// Package redact provides deterministic, idempotent projections for values
+// that may contain reusable credentials.
 package redact
 
-import "strings"
+import (
+	"net/url"
+	"strings"
+)
 
 const (
 	SecretSet  = "(set)"
 	SecretNone = "(none)"
+	urlMask    = "****"
 )
 
 // Secret replaces an opaque secret with a marker that only exposes whether a
 // value is configured.
 func Secret(value string) string {
-	if value == "" {
+	switch value {
+	case "":
 		return SecretNone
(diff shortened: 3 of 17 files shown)


flemzord:
Revalidated live head `5f1d9832`. One security/runtime blocker remains: `eventsink.redactConfigInPlace` covers ClickHouse, Kafka, HTTP, and Databricks, but has no `SinkConfig_Nats` case. `NatsSinkConfig.url` is passed directly to `nats.Connect` and may contain reusable user/password or token userinfo (`nats://user:password@host` or `nats://token@host`), so both HTTP and gRPC `GetEventsSinks` still return those credentials unchanged. Please redact NATS URLs as well and add coverage for password and token forms (including comma-separated server lists if supported). CI is green; this comment is based only on the credential disclosure. GitHub does not permit the authenticated author account to submit `CHANGES_REQUESTED` on its own PR.

paul-nicolas (commented):

6 findings posted inline: **1 Critical, 4 Major, 1 Medium**. 13 Minor/Nit withheld (listed below).

Every inline finding went through an independent adversarial verification pass (one verifier per finding, instructed to refute). 12 Critical/Major/Medium candidates entered that gate: 2 confirmed as filed, 3 corrected (severity or scope), 3 rejected outright, 4 merged into surviving findings.

**Rejected, so deliberately not posted:** the ignored `proto.Clone` type assertion (`proto.Clone` provably returns a type-identical clone, and `protojson.Marshal` of a typed-nil pointer yields `{}` — the claimed silent-nil-200 symptom does not exist); the `internal/adapter/eventsink` layering objection (`internal/adapter/json/` is already a non-transport helper and `go list -deps` shows no server machinery reaches the CLI); and `ListLogs` as a credential vector (it resolves a `LedgerInfo` and filters against the per-ledger log index, so a system-scoped sink log is unreachable there — only `GetLog` is).

**Codex/Claude conflicts and how they were resolved:** Codex filed the `HttpSinkConfig.Endpoint` leak as Major on its own; verification downgraded it to Minor standalone (userinfo is undocumented for that sink and the pre-PR CLI redactor left it verbatim too) but confirmed the *class* — so it is merged into the NATS finding as the URL-credential defect. Codex scoped the log-stream bypass to Critical across `Apply`/`GetLog`/`ListLogs`; verification cut `ListLogs`, established the `Apply` echo is benign (write-scope caller receiving back what it just sent), and settled on Major for `GetLog` + the audit surface, framed as incomplete coverage plus a false claim in the PR description rather than a newly introduced vulnerability.

The theme across the surviving findings: the redaction is correct for the shapes it anticipates, but the *contract* the PR advertises (openapi "credential-bearing fields are redacted", `events.md` "all reusable credentials") is broader than what the code delivers. Either narrow the wording or close the gaps.

- `internal/adapter/eventsink/redact.go:65` -- neither type switch has a `default:` arm, so a future sink variant fails open. Note the mechanism a reviewer would expect to catch this does not: adding `//exhaustive:enforce` still yields `0 issues`, because the `exhaustive` linter never inspects type switches. A protoreflect-driven test over `SinkConfig.type` / `DatabricksSinkConfig.auth` would.
- `tests/` -- no test anywhere asserts a read response omits a plaintext credential. The new handler-level tests cover the full serialization path minus the network hop; an e2e case provisioning one credentialed sink per type and grepping the raw gRPC/HTTP payloads would close it (and would have caught the double-redaction defect above).
- `internal/adapter/grpc/server_bucket_events_sinks_test.go:37` -- tautological assertion: the expected value is computed by `eventsink.RedactConfig`, the same function the handler calls. If redaction regressed, both sides move together. Assert `redact.SecretSet` literally.
- `cmd/ledgerctl/events/list.go:120,126` -- the default table output prints `s.Nats.GetUrl()` and `s.Http.GetEndpoint()` unredacted, while `docs/ops/cli.md:4792` now claims `ledgerctl` redacts credential-bearing fields. Only the `--output json|yaml` path goes through the redactor.
- `internal/pkg/redact/redact.go:36` -- `clickhouse://op:@host/db` (explicitly empty password) renders `op:****@host`, reporting a password that is not configured, and contradicting the function's own doc comment.
- `cmd/ledgerctl/events/redact.go:14-17` -- `secretSet`/`secretNone` re-declared over `redact.SecretSet`/`SecretNone`; the whole file is now four pass-through wrappers. Calling the shared helpers directly at the six call sites would delete it.
- `cmd/ledgerctl/events/redact_test.go:174` -- comment still references `redactSinkConfigInPlace`, deleted by this PR.
- `internal/pkg/redact/redact.go:1,5` -- no package doc and no doc comment on the exported const block. Confirmed *not* lint-enforced here (revive `package-comments` is disabled and the run is clean), so this is a convention gap only.
- `internal/bootstrap/config.go:216` -- `RedactedSecretPlaceholder = "***REDACTED***"` remains a second redaction vocabulary whose own comment claims to be the single source of truth for secret rendering. Three markers for one concept now; worth cross-referencing at minimum.
- `internal/application/events/manager.go:197` -- sink-creation failures log the wrapped `*url.Error`, whose `Error()` embeds the full DSN including the password. Pre-existing and outside this diff, but it is the same disclosure via stdout/support bundles, and the most likely trigger is an operator mistyping a DSN.
- `internal/application/events/emitter.go:428` -- `SinkError.Message` is verbatim driver output, Raft-replicated and returned un-redacted; `redact.go:29`'s "Status records are cloned unchanged" rests on an unenforced property of third-party error strings.
- `cmd/ledgerctl/store/dump.go:239` -- prints a plaintext `SinkConfig`. Offline-only (requires the data directory with the server stopped), so outside this endpoint's threat model, but it is a route into a pasted support bundle.
- `openapi.yml:2380` -- redaction is documented in prose only; the `sinks` items schema stays `type: object` with no properties, so generated clients and schemathesis see no contract for the marker values.

internal/pkg/redact/redact.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	creds := rest[:lastAt]
+	user, _, ok := strings.Cut(creds, ":")
+	if !ok {
+		return dsn
+	}
+
+	return dsn[:schemeEnd+3] + user + ":****" + rest[lastAt:]
+}
```
paul-nicolas, line 117: [Critical] `redact.DSN` only masks userinfo, so a ClickHouse password passed in the query string is served in plaintext by the endpoint this PR exists to secure

The function returns `dsn` untouched whenever there is no `@` after the scheme (line 31) or the userinfo has no `:` (line 37). But `clickhouse-go` accepts the credential as a query parameter — `clickhouse_options.go:322-324` maps `case "username"` / `case "password"` onto `o.Auth` — and `internal/application/events/sink_clickhouse.go:111` feeds the configured DSN straight into `clickhouse.ParseDSN`. Nothing constrains the DSN shape at write time: `processAddEventsSink` (`internal/domain/processing/processor_events_sink.go:12-27`) validates only batch size and name collision, and the CLI passes `--clickhouse-dsn` through unparsed. So that DSN is fully functional, and any `ledger:OpsRead` caller reads the password verbatim.

Executing this exact function body:

```
DSN("clickhouse://host:9000/ledger?username=op&password=supersecret") -> unchanged
DSN("clickhouse://op@host:9000/ledger?password=supersecret")          -> unchanged
```

The same string surgery has two further defects. `strings.LastIndex(rest, "@")` on line 30 does not know where the authority ends, so an `@` in the query is taken as the userinfo delimiter:

```
DSN("clickhouse://host:9000/db?password=p@ss") -> "clickhouse://host:****@ss"
```

which is not a valid DSN, drops the host the doc comment promises to preserve, and still exposes the password suffix. And `clickhouse://op:@host/db` (explicitly empty password) emits `op:****@host`, reporting a password that does not exist.

Fix: parse with `net/url` instead of slicing. Mask the userinfo via `url.UserPassword(u.User.Username(), "****")` only when a non-empty password is actually present, then rewrite the query with `url.Values`, masking every credential-ish key (`password`, `pass`, `token`, `access_token`, ...) and re-encoding. On a parse error return an opaque marker rather than the input — failing closed. Keep the current heuristic only as a fallback if you need it for non-URL DSNs. Table cases to add: `?password=`, `?username=&password=`, userinfo+query combined, `?password=p@ss`, and `op:@host`.

internal/pkg/redact/redact.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Secret replaces an opaque secret with a marker that only exposes whether a
+// value is configured.
+func Secret(value string) string {
+	if value == "" {
+		return SecretNone
+	}
+
+	return SecretSet
+}
```
paul-nicolas, line 27: [Major] `redact.Secret` is not a fixed point, and this PR creates a second redaction pass — a sink with no configured secret is reported to the caller as `(set)`

`Secret("(none)")` returns `(set)`, because `(none)` is non-empty. That would be harmless if redaction ran once, but it now runs twice on two reachable paths:

1. **CLI**, unconditionally: `cmd/ledgerctl/events/list.go:56` receives an already-redacted response, then re-redacts at line 61 (`--output json|yaml`) and per-field at 127/146/167/174 (table). `docs/ops/cli.md:4792` documents this as intentional for older-server compatibility.
2. **Server-side leader forwarding**: `internal/bootstrap/module.go:828-837` wires `RoutedController` as the only `ctrl.Controller` both adapters get; `controller_routed.go:391-397` routes the read through `readCtrl` (lines 59-101), which forwards to a remote node either when the caller asks for it (`x-consistency: leader` off incoming metadata, `grpc/consistency.go:80-90`; `ledgerctl --consistency leader`) or via `leader_fallback` on any non-leader whose `ReadIndexAndWait` returns `ErrNodeSyncing`/`ErrNotLeader` — no client cooperation needed. `client_bucket.go:490-496` returns `resp.GetSinks()` verbatim from a leader that already redacted at `server_bucket.go:1153`, and the receiving node redacts it again.

Net effect for an HTTP sink created with no HMAC secret: local path `"" -> "(none)"`, forwarded path `"" -> "(none)" -> "(set)"`. Same cluster, same sink, two different answers depending on routing. Secret-less sinks are ordinary, not a corner case — `add_sink.go:87` defaults `--http-secret` to `""`, lines 260-266 store it unvalidated, and line 385's `if httpSecret != ""` shows emptiness is expected; SASL-less Kafka is the common configuration and `SaslPassword` is redacted unconditionally.

Nothing leaks here — the harm is an inverted security signal. An operator auditing which webhook sinks have HMAC signing enabled sees `(set)` for sinks that have none, in the PR that is supposed to make that view trustworthy. `redact.DSN` is idempotent, so the defect is confined to `Secret`.

Fix: make `Secret` a fixed point — return the input unchanged when it already equals `SecretNone` or `SecretSet` — and assert `Secret(Secret(x)) == Secret(x)` in a test. Alternatively drop the now-redundant client-side pass, though the fixed point is worth having regardless.

internal/adapter/eventsink/redact.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+	case *commonpb.SinkConfig_Http:
+		if sink.Http != nil {
+			sink.Http.Secret = redact.Secret(sink.Http.GetSecret())
+		}
+	case *commonpb.SinkConfig_Databricks:
+		if sink.Databricks != nil {
+			redactDatabricksAuthInPlace(sink.Databricks)
+		}
+	}
```
paul-nicolas, line 72: [Major] URL-shaped credentials are handled inconsistently: the NATS sink is absent from this switch entirely, so `NatsSinkConfig.url` is returned raw

`SinkConfig.type` has five variants (`misc/proto/common.proto:570-576`) and this switch handles four — there is no `case *commonpb.SinkConfig_Nats`. The NATS URL is not merely an address: `internal/application/events/sink_nats.go:46` calls `nats.Connect(cfg.URL)`, and `nats.go`'s `connectProto` reads credentials straight off `nc.current.URL.User` — `nats://user:pass@host:4222` becomes user/password, `nats://token@host:4222` becomes the auth token. That is standard, documented NATS authentication, and nothing rejects it at write time (`processAddEventsSink` checks only batch size and name; `add_sink.go:224-233` checks only non-emptiness). So for any cluster using an authenticated NATS sink, this endpoint hands out a working credential with no redaction at all.

`cmd/ledgerctl/events/redact_test.go:160` asserts the URL survives unchanged (`TestRedactSinkConfig_NatsHasNoSecret`), so the omission is an explicit premise rather than an oversight — and the premise is wrong.

Same class, lower likelihood: `HttpSinkConfig.Endpoint` (line 57) is also returned verbatim, and Go turns userinfo in it into a live `Authorization: Basic` header on the first plain POST — `net/http/client.go:251-257`, inside `send()`, not redirect-only; confirmed against an httptest server replicating the sink's exact call shape (`sink_http.go:74,95`, stock `http.Client`, no custom Transport). Userinfo is undocumented for this sink (the documented credential is `--http-secret`), so on its own it would be a nit; taken with NATS it shows the gap is the URL-credential class rather than one field.

Fix: add a `redact.URLUserinfo` helper that parses with `net/url` and clears or masks `u.User` whenever it is non-nil, then route both `Nats.Url` and `Http.Endpoint` through it. Do **not** reuse `redact.DSN` as-is: the token form `nats://token@host` has no `:` in the userinfo and comes back verbatim, and its `LastIndex` heuristic mangles `https://example.com:443/hook@v2` — a plausible webhook shape. Then invert `TestRedactSinkConfig_NatsHasNoSecret`, and apply the same to the CLI table path (`list.go:120,126`), which prints both fields raw.

internal/adapter/grpc/server_bucket.go
```diff
@@ -1149,10 +1150,10 @@ func (impl *BucketServiceServerImpl) GetEventsSinks(ctx context.Context, _ *serv
 		return nil, fmt.Errorf("loading events sinks: %w", err)
 	}
 
-	return &servicepb.GetEventsSinksResponse{
+	return eventsink.RedactGetResponse(&servicepb.GetEventsSinksResponse{
 		Sinks:        sinks,
 		SinkStatuses: statuses,
-	}, nil
+	}), nil
```
paul-nicolas, line 1156: [Major] The same plaintext `SinkConfig` is still returned by `GetLog` at the identical `ledger:OpsRead` scope, so the disclosure described in the PR's root-cause section is not actually closed

Redacting here closes one surface. The config is persisted verbatim on another: `processAddEventsSink` (`internal/domain/processing/processor_events_sink.go:29-34`) puts `order.GetConfig()` unmodified into `AddedEventsSinkLog.Config`, `processor.go:349-360,576` assigns it a global sequence, and `state/batch.go:16-29` writes it at `ZoneCold/SubColdLog`. That it is stored unredacted is proven by `internal/infra/backup/rebuild.go:288-296`, which reconstructs live runtime credentials from that payload alone.

Reading it back is not privileged any further than reading this endpoint:

- `server_bucket.go:1058-1069` — `GetLog` gates on `internalauth.ScopeOpsRead` (the same scope as line 1142) and returns the raw `*commonpb.Log`; `internal/query/log.go:48-62` is a keyed get by global sequence with no ledger scoping and no payload-type restriction.
- `internal/adapter/http/handler.go:148` registers `/logs/{sequence}` in the *same* `requireOpsRead` group, four lines above `/events-sinks` at 151; `handlers_get_log.go:23-30` → `response.go:102` protojson-marshals the whole `Log`.
- `internal/adapter/auth/scopes.go:69-77` — `ledger:read` expands to both `ScopeOpsRead` and `ScopeAuditRead`, which is exactly the "ordinary read callers" class the PR description names.

So the credential is one call away at the same scope. `ListLogs` is *not* an additional vector (`controller_default.go:1618-1634` filters against the per-ledger log index and a system-scoped sink log has no `LedgerLog.Id`), and the `Apply` echo is benign — it returns the config to the write-scope caller that just supplied it. The audit surface is worse: `AuditItem.serialized_order` carries the accepted `AddEventsSinkOrder` bytes (`server_bucket.go:973-979` under `ScopeAuditRead`, serialized whole by `handlers_get_audit_entry.go:16-32`), and per `misc/proto/audit.proto:44-46,97` plus CLAUDE.md invariant 10 those bytes are hash-chain-bound — no redaction pass can ever be applied there without breaking `verifyAuditHashChain`.

To be clear on severity: none of this is introduced by this PR — `GetLog` and the audit surfaces predate it and are untouched by the diff. What is wrong *with this PR* is that it ships a claim it does not satisfy. Suggested resolution: apply `eventsink.RedactConfig` in the log-serialization path as well (a `RedactLog(*commonpb.Log)` helper in this package, called from the controller so gRPC and HTTP both inherit it), and file a follow-up for the un-redactable audit disclosure — keeping credentials out of the ordered payload entirely, e.g. the order references a credential held in a separate non-chained keyspace. Until both land, the PR description and the docs should not assert the hole is closed.

openapi.yml
```diff
@@ -2348,7 +2348,9 @@ paths:
       summary: List configured event sinks
       description: |
         Returns every event sink configured on the cluster. Requires
-        `ledger:OpsRead` scope.
+        `ledger:OpsRead` scope. Credential-bearing fields are redacted: opaque
+        secrets use `(set)` or `(none)`, and URL-shaped DSN passwords use
+        `****`. Read responses must not be reused as sink mutation payloads.
```
paul-nicolas: [Medium] The published contract overstates what the code does — "credential-bearing fields are redacted" is false for three field shapes

This text, `docs/technical/architecture/subsystems/events-mirror/events.md:163` ("replaces **all** reusable credentials with non-reusable markers") and `docs/ops/cli.md:4792` all assert blanket coverage. As implemented, three credential-bearing shapes come back verbatim:

- `NatsSinkConfig.url` — no case in the redaction switch at all (`internal/adapter/eventsink/redact.go:48`), and `nats.Connect` treats URL userinfo as standard auth.
- ClickHouse DSNs carrying the password as a query parameter — `redact.DSN` returns them unchanged, and the driver accepts them.
- `HttpSinkConfig.endpoint` — userinfo in it becomes a live `Authorization: Basic` header.

A public spec that promises redaction is load-bearing in a way ordinary prose is not: an integrator reads it and decides the response is safe to log, forward, or store. That makes the overclaim a defect in its own right, independent of the code gaps.

Fix: once the gaps are closed, state the contract field by field — which fields are masked, which are intentionally visible, and that a field documented as visible must not carry credentials. If you would rather land the code fixes separately, narrow this wording now to name the fields actually covered (Kafka SASL password, HTTP HMAC secret, Databricks PAT / OAuth client secret, userinfo passwords in URL-shaped DSNs) instead of claiming the general property.

internal/pkg/redact/redact.go
```diff
@@ -0,0 +1,42 @@
+package redact
+
+import "strings"
+
+const (
+	SecretSet  = "(set)"
+	SecretNone = "(none)"
+)
```
paul-nicolas, line 14: [Medium] This package has no test file, and the PR body claims it was tested

`GOROOT= go test ./internal/pkg/redact/...` reports `?   github.com/formancehq/ledger/v3/internal/pkg/redact  [no test files]`; the directory contains only `redact.go`. The PR's Validation section lists `go test ./internal/pkg/redact` among the commands run, which cannot have exercised anything.

This is the only place both redaction primitives live, and it is now on a server read path rather than being CLI cosmetics. The only indirect coverage is `cmd/ledgerctl/cmdutil/format_test.go:85-137`, seven happy-path DSN shapes reached through the delegation — and notably it asserts at lines 118-122 that a keyword DSN containing `password=secret` is returned verbatim as *expected* behaviour. None of the shapes that actually leak or corrupt are covered: `?password=`, `?secure&password=`, userinfo-without-colon plus `?password=`, `@` inside the query, multi-host authorities, `user:@host`, or the `Secret`/`DSN` idempotency properties.

Fix: add `internal/pkg/redact/redact_test.go`, table-driven with `t.Parallel()` per the repo's testing conventions, covering each shape above. Assert `assert.NotContains(got, secret)` rather than only string equality — an equality assertion written against current behaviour is what let the query-parameter leak look intentional in `format_test.go`. Include `Secret(Secret(x)) == Secret(x)` and `DSN(DSN(x)) == DSN(x)`.

flemzord:
Revalidated live head `19beab26`. The NATS fix still leaves one credential-bearing form unredacted: `redactNATSURLs` skips every server entry without `://`, but the pinned `nats.go` v1.49.0 `parseServerURL` explicitly prepends the connection scheme when it is absent. Consequently `user:password@host:4222` and `token@host:4222` (including such entries in a comma-separated list) are accepted by `nats.Connect` yet returned unchanged by `GetEventsSinks`. Please redact the scheme-less forms accepted by the client and cover password/token cases. I am not repeating the existing ClickHouse query-password, repeated-redaction, or historical-log threads. GitHub does not permit the authenticated author account to submit `CHANGES_REQUESTED` on its own PR.

flemzord:
Revalidated live head `657b40bf`. The previously reported scheme-less NATS credential blocker remains active. `redactNATSURLs` now delegates each server to `redact.URLUserinfo`, but `URLUserinfo` passes `user:password@host:4222` or `token@host:4222` directly to `url.Parse`; those forms have no parsed `User`, so the helper returns the plaintext value unchanged. The pinned NATS client accepts them by adding its default scheme, and the new tests only cover scheme-bearing URLs. Please normalize scheme-less NATS servers before parsing (or parse them with the NATS default scheme) and add password/token regressions. I am not repeating the other existing redaction threads. GitHub does not permit the authenticated author account to submit `CHANGES_REQUESTED` on its own PR.

PR #1652, opened 2026-07-29
fix(snapshot): prevent FetchFile path traversal

flemzord:
## Summary

- reject non-local snapshot file paths at the `FetchFile` RPC boundary
- open snapshot files through an `os.Root` confined to the prepared checkpoint
- cover parent traversal, absolute paths, symlink escape, and legitimate nested files
- document the snapshot filesystem-containment invariant

## Root cause

The previous guard compared the caller's path with `filepath.Rel(checkpoint, filepath.Join(checkpoint, path))`. A clean parent traversal such as `../secret` survives that round trip unchanged, so the guard accepted it. The final `os.Open(filepath.Join(...))` also followed a symlink outside the checkpoint.

The lexical guard now uses `filepath.IsLocal`, while the authoritative final open uses `os.Root` so a caller-controlled path or symlink cannot escape the session checkpoint.

## Security impact

This closes Codex Security finding `csf_28a6468cfb15e1235fc5480c` / occurrence `occ_fc7550ae99fcd8c15ad3b04a` (High, CWE-22/CWE-200).

Jira: [EN-1633](https://formance-team.atlassian.net/browse/EN-1633)

## Validation

- `GOROOT= go test ./internal/adapter/grpc -run '^(TestStreamOneFile_PathConfinement|TestStreamOneFile_Content|TestStreamOneFile_Empty|TestSnapshotService_FullRoundTrip)$' -count=1`
- `GOROOT= go test ./internal/adapter/grpc -count=1`
- `GOROOT= go build ./...`
- `git diff --check`
- GitNexus `detect_changes` against `release/v3.0`: low risk, no affected execution flows

The new regression test failed before the patch because parent traversal and symlink escape streamed the external file. It passes after the patch, and the nested-file control remains readable.

[EN-1633]: https://formance-team.atlassian.net/browse/EN-1633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/technical/architecture/subsystems/consensus/snapshot-file-transfer.md
+++ b/docs/technical/architecture/subsystems/consensus/snapshot-file-transfer.md
@@ -1,0 +1,37 @@
+# Snapshot File Transfer
+
+Follower synchronization uses the internal `SnapshotService` gRPC API to copy
+a temporary Pebble checkpoint from a donor node. `PrepareSnapshot` creates the
+checkpoint and returns a manifest. Concurrent `FetchFile` calls then stream the
+manifest's relative paths before `CloseSession` removes the checkpoint.
+
+## Filesystem containment
+
+`FetchFileRequest.path` crosses an internal network boundary and must never
+grant filesystem authority beyond the prepared checkpoint. The server applies
+two complementary controls:
+
+1. `filepath.IsLocal` rejects empty, absolute, reserved, and parent-escaping
+   paths before streaming starts.
+2. `streamOneFile` opens the final file through `os.Root`, rooted at the session
+   checkpoint. The rooted open prevents a path from escaping through a symbolic
+   link or through a concurrent rename between validation and open.
+
+The rooted open is the authoritative containment control. Lexical validation is
+kept at the RPC boundary so malformed paths receive `InvalidArgument` instead
+of being interpreted as missing files. Valid nested regular files remain
+supported, and symlinks are followed only when their target remains beneath the
+checkpoint root.
+
+These checks do not turn the checkpoint into a general sandbox: `os.Root` does
+not block bind mounts, device files, or filesystem boundaries. Snapshot
+checkpoints are created by the ledger process from Pebble data and must not be
+populated from caller-controlled filesystem trees.
+
+## Owning code
+
+| Responsibility | Location |
+| --- | --- |
+| Snapshot session and lexical path validation | `internal/adapter/grpc/server_snapshot.go` |
+| Rooted file open and chunk streaming | `internal/adapter/grpc/file_streaming.go` |
+| Snapshot protocol | `misc/proto/snapshot.proto` |
--- a/docs/technical/architecture/subsystems/consensus/snapshot-file-transfer.md
+++ b/docs/technical/architecture/subsystems/consensus/snapshot-file-transfer.md
@@ -11,8 +11,8 @@ manifest's relative paths before `CloseSession` removes the checkpoint.
 grant filesystem authority beyond the prepared checkpoint. The server applies
 two complementary controls:
 
-1. `filepath.IsLocal` rejects empty, absolute, reserved, and parent-escaping
-   paths before streaming starts.
+1. `filepath.IsLocal` rejects empty, absolute, and parent-escaping paths before
+   streaming starts. It also rejects reserved device names on Windows.
 2. `streamOneFile` opens the final file through `os.Root`, rooted at the session
    checkpoint. The rooted open prevents a path from escaping through a symbolic
    link or through a concurrent rename between validation and open.
@@ -20,8 +20,25 @@ two complementary controls:
 The rooted open is the authoritative containment control. Lexical validation is
 kept at the RPC boundary so malformed paths receive `InvalidArgument` instead
 of being interpreted as missing files. Valid nested regular files remain
-supported, and symlinks are followed only when their target remains beneath the
-checkpoint root.
+supported. A symlink is followed only when its target is expressed relatively
+and remains beneath the checkpoint root; an absolute symlink target is rejected
+even when it points inside the root. Pebble checkpoints contain hard links or
+copied files rather than symlinks, so this distinction does not arise in normal
+checkpoint data.
+
+## Follower-side receiving
+
+The manifest returned by a donor is network-supplied input. The follower
+validates every manifest path with `filepath.IsLocal` immediately after
+`PrepareSnapshot`, before scanning resume state or requesting a file. A second
+validation in the file fetcher protects direct callers.
+
+Resume scans, temporary-file creation, hashing, and the final rename all use an
+`os.Root` opened on the follower's staging directory. This is the mirror of the
+serving-side containment boundary: neither a manifest traversal nor a symlink
+already present in the staging directory can redirect a read or write outside
+that directory. Temporary files are removed on stream, hash, close, or rename
+failure.
 
 These checks do not turn the checkpoint into a general sandbox: `os.Root` does
 not block bind mounts, device files, or filesystem boundaries. Snapshot
@@ -34,4 +51,5 @@ populated from caller-controlled filesystem trees.
 | --- | --- |
 | Snapshot session and lexical path validation | `internal/adapter/grpc/server_snapshot.go` |
 | Rooted file open and chunk streaming | `internal/adapter/grpc/file_streaming.go` |
+| Manifest validation and rooted receive-side writes | `internal/application/ctrl/file_receiver.go`, `internal/application/ctrl/file_fetcher.go` |
 | Snapshot protocol | `misc/proto/snapshot.proto` |
--- a/internal/adapter/grpc/file_streaming_test.go
+++ b/internal/adapter/grpc/file_streaming_test.go
@@ -115,3 +115,56 @@ func TestStreamOneFile_Empty(t *testing.T) {
 	require.Len(t, chunks, 1)
 	require.True(t, chunks[0].GetEof())
 }
+
+func TestStreamOneFile_PathConfinement(t *testing.T) {
+	t.Parallel()
+
+	checkpointDir := t.TempDir()
+	outsideDir := t.TempDir()
+	outsidePath := filepath.Join(outsideDir, "secret.txt")
+	require.NoError(t, os.WriteFile(outsidePath, []byte("secret"), 0644))
+
+	nestedDir := filepath.Join(checkpointDir, "nested")
+	require.NoError(t, os.MkdirAll(nestedDir, 0755))
+	require.NoError(t, os.WriteFile(filepath.Join(nestedDir, "data.txt"), []byte("nested data"), 0644))
+
+	symlinkPath := filepath.Join(checkpointDir, "outside-link")
+	require.NoError(t, os.Symlink(outsidePath, symlinkPath))
+
+	traversalPath, err := filepath.Rel(checkpointDir, outsidePath)
+	require.NoError(t, err)
+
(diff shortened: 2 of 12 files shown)


internal/adapter/grpc/server_snapshot.go
```diff
@@ -138,10 +138,7 @@ func (s *SnapshotServiceServerImpl) FetchFile(req *snapshotpb.FetchFileRequest,
 
 	// Validate the path stays within the checkpoint directory.
 	relPath := req.GetPath()
-	fullPath := filepath.Join(session.checkpointPath, relPath)
-
-	resolved, err := filepath.Rel(session.checkpointPath, fullPath)
-	if err != nil || resolved != relPath {
+	if !filepath.IsLocal(relPath) {
```
paul-nicolas, line 141: [Major] The mirror-image traversal on the receiving side is left unfixed

Re: `internal/application/ctrl/file_fetcher.go:70` and `:120-121` — flagging it here because this line is the counterpart control in the diff, and because the new architecture doc this PR adds presents itself as *the* snapshot filesystem-containment boundary while covering only the serving half.

This PR makes the server refuse to *read* outside the checkpoint. The client that consumes the same protocol still *writes* wherever the donor tells it to. In `fileFetcher.fetchFileOnce`, `entry.GetPath()` comes straight out of the donor's `SnapshotManifest` — a remote, network-supplied value — with no validation anywhere on the path from `FetchSnapshot` → `fetchWithSession` → `fetchFile` → `fetchFileOnce`:

```go
tmpPath := filepath.Join(targetDir, entry.GetPath()+".tmp")   // :70
os.MkdirAll(filepath.Dir(tmpPath), 0755)                      // :71
tmpFile, err := os.Create(tmpPath)                            // :75
// ... stream body written to tmpFile ...
finalPath := filepath.Join(targetDir, entry.GetPath())        // :120
os.Rename(tmpPath, finalPath)                                 // :121
```

A manifest entry with path `../../../../etc/cron.d/evil` makes a syncing follower create directories and write attacker-controlled bytes outside `targetDir`. That is an arbitrary file **write** — strictly more severe than the arbitrary read this PR closes (CWE-22, same protocol, same trust boundary: whoever can serve `FetchFile` on the Raft plane can also be the donor, and the follower does not choose the manifest).

Two details that remove the obvious mitigations:
- The SHA256 check at `:116` runs *after* the `.tmp` file is already created and fully written, and a hash mismatch returns without unlinking it — so the traversed write lands even on the failure path.
- The attacker authors the manifest, so they control content *and* the expected hash. The hash matches and the `os.Rename` to the clean final path succeeds.

Fix: validate manifest paths with the same two controls used here. Reject any entry where `!filepath.IsLocal(entry.GetPath())` as soon as the manifest is received (fail the whole sync — a donor sending a non-local path is either broken or hostile), and do the writes through an `os.Root` rooted at `targetDir` (`root.Create` / `root.Rename` / `root.MkdirAll`) so a symlink already present under `targetDir` cannot redirect them either. `scanCompletedFiles` (`file_receiver.go:24`) reads the same unvalidated paths and should sit behind the same gate.

If you would rather keep this PR tight, that is reasonable — but it should be a tracked follow-up rather than left implicit, and the new doc should say the receive side is not yet covered instead of implying the boundary is closed.

docs/technical/architecture/subsystems/consensus/snapshot-file-transfer.md
```diff
@@ -0,0 +1,37 @@
...
+1. `filepath.IsLocal` rejects empty, absolute, reserved, and parent-escaping
+   paths before streaming starts.
+2. `streamOneFile` opens the final file through `os.Root`, rooted at the session
+   checkpoint. The rooted open prevents a path from escaping through a symbolic
+   link or through a concurrent rename between validation and open.
+
+The rooted open is the authoritative containment control. Lexical validation is
+kept at the RPC boundary so malformed paths receive `InvalidArgument` instead
+of being interpreted as missing files. Valid nested regular files remain
+supported, and symlinks are followed only when their target remains beneath the
```
paul-nicolas: [Medium] Two claims on this page are measurably wrong about the controls they describe

**1. Line 23-24: "symlinks are followed only when their target remains beneath the checkpoint root."** `os.Root` rejects a symlink whose target is an *absolute* path even when that target resolves inside the root. Measured on Go 1.26.5, both symlinks living directly under the root:

```
<root>/rel-inside-link -> "nested/data.txt"                 root.Open -> OK
<root>/abs-inside-link -> "<root>/nested/data.txt"          root.Open -> openat abs-inside-link: path escapes from parent
```

So the predicate is not "target beneath the root" but "target beneath the root *and* expressed relatively". As written, a reader reasoning about whether a checkpoint may contain symlinks would conclude the absolute-target case works, and it does not. This is harmless in production today — Pebble's `Checkpoint` uses `vfs.LinkOrCopy`, i.e. hard links or copies, never symlinks — but the sentence is a normative statement about a security control on the page that now owns this invariant.

Suggested: "symlinks are followed only when their target is expressed relatively and remains beneath the checkpoint root; a symlink with an absolute target is rejected even if that target is inside the root. Pebble checkpoints contain hard links, not symlinks, so this case does not arise in practice."

**2. Line 14-15: "`filepath.IsLocal` rejects empty, absolute, reserved, and parent-escaping paths."** "Reserved" is Windows-only. On Linux and darwin `filepath.IsLocal("NUL") == true` (measured). CI is Linux-only (`namespace-profile-linux-amd64-4vcpu` across `.github/workflows/`), so the claim holds on no platform this software actually runs on. Either drop "reserved" or qualify it as Windows-only.

Worth noting the empty-path claim *is* correct: `IsLocal("") == false`. Everything else I checked on this page is accurate — the concurrency/`CloseSession` ordering matches `snapshot_fetcher.go:239-268`, the `os.Root` non-sandbox caveat is fair, and `misc/proto/snapshot.proto` exists.

internal/adapter/grpc/file_streaming_test.go
```diff
@@ -115,3 +115,56 @@ func TestStreamOneFile_Empty(t *testing.T) {
...
+	require.NoError(t, err)
+
+	tests := []struct {
+		name    string
+		path    string
+		wantErr bool
+	}{
+		{name: "nested file", path: filepath.Join("nested", "data.txt")},
+		{name: "parent traversal", path: traversalPath, wantErr: true},
+		{name: "absolute path", path: outsidePath, wantErr: true},
```
paul-nicolas: [Medium] The "absolute path" case is a vacuous assertion — it passed before the fix too

This subtest passes `outsidePath` (an absolute path) to `streamOneFile` and asserts an error. But the pre-PR unconfined open would also have failed on it, because `filepath.Join` concatenates rather than absolutizes. Measured:

```
OLD open "parent traversal"  joined=/T/out.../secret.txt                -> SUCCESS   <- the real vuln
OLD open "absolute path"     joined=/T/cp.../T/out.../secret.txt        -> FAIL (no such file or directory)
```

So the case discriminates nothing about containment: it is green with `os.Root`, with plain `os.Open`, and with no guard at all. The two cases that *do* prove the fix are `parent traversal` and `symlink escape` — those genuinely flip from SUCCESS to error, which is what makes this test worth having.

Absolute paths are actually the job of the `filepath.IsLocal` guard at `server_snapshot.go:141`, and that guard is not reached from `streamOneFile` — this subtest tests the wrong layer for its own name.

Fix: move the absolute-path expectation to a direct test of `SnapshotServiceServerImpl.FetchFile` asserting `codes.InvalidArgument`, which also gives the lexical guard its first coverage. Independently, `require.Error(t, err)` is weak for the two cases that matter — consider asserting the containment cause so an unrelated failure (missing fixture, permissions) cannot masquerade as a pass. note that `errors.Is(err, fs.ErrInvalid)` will *not* work here — I measured that `root.Open`'s escape error satisfies neither `fs.ErrInvalid` nor `fs.ErrNotExist` — so `require.ErrorContains(t, err, "path escapes from parent")` is the practical assertion until the stdlib exports a sentinel for it.

paul-nicolas (commented):
Multi-model review (Claude + Codex). The core fix is correct: `os.Root` is the right authoritative control, and I verified empirically on Go 1.26.5 that the pre-fix `os.Open(filepath.Join(cp, "../out/secret.txt"))` **did** succeed while `root.Open` rejects it with `path escapes from parent`. The new test genuinely fails without the patch for the traversal and symlink cases. Pebble checkpoints use `vfs.LinkOrCopy` (hard links, not symlinks), so the rooted open introduces no functional regression for real checkpoints.

**Findings: 1 major, 2 medium posted inline. 5 minor/nit withheld.**

The major one is a scope gap rather than a defect in the diff: this PR hardens the *serving* side of the snapshot protocol and leaves the symmetric *receiving* side (donor-supplied manifest paths → `os.Create`/`os.Rename`) unvalidated. That is an arbitrary file **write**, strictly worse than the read this PR closes, on the same protocol and the same trust boundary.

**Conflicts resolved by measurement.** Codex rated "rooted-open failures now surface as generic `Unknown`" as major. I withheld it: I measured `*os.PathError.Path` pre vs post and the PR *narrows* path disclosure (absolute joined path → bare relative name); the `Unknown` code is identical pre- and post-PR, `Unknown` is not in `isRetryableError`'s allowlist (`snapshot_fetcher.go:60-65`) so there is no wasted-retry cost, and the missing `errorConversionStreamInterceptor` on `NewRaftServer` is a pre-existing gap on the cluster-secret-guarded plane, not something this diff introduces. I also downgraded my own "the `IsLocal` guard is untested" finding to minor after proving it: deleting the guard keeps the whole package green, but it replaces an equally-untested pre-existing lexical check and `os.Root` alone is sufficient to block every escape, so nothing leaks without it.

- internal/adapter/grpc/server_snapshot.go:141 -- `filepath.IsLocal` guard has zero test coverage; verified that deleting it leaves `go test ./internal/adapter/grpc` fully green. Low impact (defense-in-depth only), but one direct unit test on the real `FetchFile` asserting `codes.InvalidArgument` for `../x` would close it — the bufconn harness cannot, since `testSnapshotServer.FetchFile` (server_snapshot_test.go:80-91) is a duplicate handler without the guard.
- internal/adapter/grpc/file_streaming.go:76 -- a containment rejection reaches the caller as bare `codes.Unknown` (measured on bufconn), indistinguishable from an I/O fault. No concrete cost today; worth a deliberate code if this ever gets a second client.
- internal/adapter/grpc/file_streaming_test.go:145 -- no coverage for a symlink with a *relative* in-root target, i.e. the exact case the new doc says remains supported.
- internal/adapter/grpc/file_streaming_test.go:142 -- no case for a path that resolves to a directory (`"."` passes `IsLocal` and `root.Open`, then fails on `Read` with `is a directory`) or for a dangling symlink.
- internal/adapter/grpc/file_streaming_test.go:132 -- the symlink subtest will fail rather than skip on a platform without `os.Symlink`. CI is Linux-only (`namespace-profile-linux-amd64-4vcpu`), so this is theoretical; a `t.Skip` on `os.Symlink` error would make it portable.

flemzord:
Reviewed live head `84a6a38f`. The serving side now combines lexical validation with an `os.Root` open, and the later commits close the symmetric receive-side write path by validating the donor manifest and rooting resume scans, temporary-file creation, hashing, cleanup, and final rename in the staging directory. The traversal and symlink regressions cover both directions, and functional CI plus Codecov are green. I found no correctness/runtime blocker on this head; GitHub approval is intentionally disabled by this cron configuration.

PR #1653, opened 2026-07-29
fix(security): redact mirror credentials from ledger reads

flemzord:
## What changed

- add a shared LedgerInfo response projection that deep-clones before redaction
- remove OAuth client secrets from HTTP mirror configuration
- strip PostgreSQL URI passwords and password query parameters, failing closed for ambiguous non-URI DSNs
- apply the projection to gRPC and HTTP GetLedger/ListLedgers responses
- document the read-response contract in OpenAPI and mirror architecture docs
- add regression coverage for both mirror variants, both transports, list/get paths, and non-mutation

## Why

A caller with ordinary `ledger:LedgerRead` authority could retrieve reusable mirror database passwords or OAuth client secrets because `LedgerInfo` embedded the stored `MirrorSourceConfig` unchanged.

The response boundary now returns a secret-safe clone while internal mirror workers continue to receive the original credentials.

Jira: [EN-1635](https://formance-team.atlassian.net/browse/EN-1635)
Codex Security: `csf_77e237f70693cb54836d4c80` / `occ_5acf221f891c2da6179d9da6`

## Validation

- `GOROOT= go test ./internal/adapter/ledgerinfo ./internal/adapter/grpc ./internal/adapter/http`
- `GOROOT= go build ./...`
- GitNexus `detect_changes` against `origin/release/v3.0`: low risk, no affected execution flow

## Coordination

EN-1632 introduces generic DSN redaction for a separate config surface. This PR stays independently mergeable with an adapter-local LedgerInfo projection; after EN-1632 lands, the small DSN helper overlap can be consolidated during rebase without coupling the two security fixes.

[EN-1635]: https://formance-team.atlassian.net/browse/EN-1635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/adapter/ledgerinfo/redact.go
+++ b/internal/adapter/ledgerinfo/redact.go
@@ -1,0 +1,102 @@
+package ledgerinfo
+
+import (
+	"net/url"
+	"strings"
+
+	"github.com/formancehq/ledger/v3/internal/pkg/cursor"
+	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
+)
+
+// RedactSecrets returns a deep-cloned LedgerInfo safe for external read
+// responses. The stored value must retain the credentials used by mirror
+// workers, so redaction belongs at the adapter boundary and never mutates the
+// caller-owned message.
+func RedactSecrets(info *commonpb.LedgerInfo) *commonpb.LedgerInfo {
+	if info == nil {
+		return nil
+	}
+
+	redacted := info.CloneVT()
+	source := redacted.GetMirrorSource()
+
+	if source == nil {
+		return redacted
+	}
+
+	switch typed := source.GetType().(type) {
+	case *commonpb.MirrorSourceConfig_Http:
+		if credentials := typed.Http.GetOauth2ClientCredentials(); credentials != nil {
+			credentials.ClientSecret = ""
+		}
+	case *commonpb.MirrorSourceConfig_Postgres:
+		if typed.Postgres != nil {
+			typed.Postgres.Dsn = redactPostgresDSN(typed.Postgres.GetDsn())
+		}
+	}
+
+	return redacted
+}
+
+// NewRedactingCursor applies RedactSecrets lazily while preserving the wrapped
+// cursor's streaming and close semantics.
+func NewRedactingCursor(inner cursor.Cursor[*commonpb.LedgerInfo]) cursor.Cursor[*commonpb.LedgerInfo] {
+	return &redactingCursor{inner: inner}
+}
+
+type redactingCursor struct {
+	inner cursor.Cursor[*commonpb.LedgerInfo]
+}
+
+func (c *redactingCursor) Next() (*commonpb.LedgerInfo, error) {
+	info, err := c.inner.Next()
+	if err != nil {
+		return nil, err
+	}
+
+	return RedactSecrets(info), nil
+}
+
+func (c *redactingCursor) Close() error {
+	return c.inner.Close()
+}
+
+// redactPostgresDSN removes password-bearing URI components while retaining
+// non-secret connection metadata. Unknown/non-URI forms are blanked so a
+// legacy value cannot bypass the response boundary through parser ambiguity.
+func redactPostgresDSN(dsn string) string {
+	if dsn == "" {
+		return ""
+	}
+
+	parsed, err := url.Parse(dsn)
+	if err != nil || (parsed.Scheme != "postgres" && parsed.Scheme != "postgresql") || parsed.Host == "" {
+		return ""
+	}
+
+	changed := false
+
+	if parsed.User != nil {
+		if _, hasPassword := parsed.User.Password(); hasPassword {
+			parsed.User = url.User(parsed.User.Username())
+			changed = true
+		}
+	}
+
+	query := parsed.Query()
+
+	for key := range query {
+		if strings.Contains(strings.ToLower(key), "password") {
+			query.Del(key)
+			changed = true
+		}
+	}
+
+	if !changed {
+		return dsn
+	}
+
+	parsed.RawQuery = query.Encode()
+
+	return parsed.String()
+}
--- a/internal/adapter/ledgerinfo/redact.go
+++ b/internal/adapter/ledgerinfo/redact.go
@@ -8,6 +8,8 @@ import (
 	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
 )
 
+const redactedDSNPlaceholder = "[redacted]"
+
 // RedactSecrets returns a deep-cloned LedgerInfo safe for external read
 // responses. The stored value must retain the credentials used by mirror
 // workers, so redaction belongs at the adapter boundary and never mutates the
@@ -61,9 +63,21 @@ func (c *redactingCursor) Close() error {
 	return c.inner.Close()
 }
 
+// NextCursor preserves the optional routed-cursor trailer capability used by
+// the gRPC pagination layer. A local cursor has no upstream token.
+func (c *redactingCursor) NextCursor() string {
+	if provider, ok := c.inner.(interface{ NextCursor() string }); ok {
+		return provider.NextCursor()
+	}
+
+	return ""
+}
+
 // redactPostgresDSN removes password-bearing URI components while retaining
-// non-secret connection metadata. Unknown/non-URI forms are blanked so a
-// legacy value cannot bypass the response boundary through parser ambiguity.
+// non-secret connection metadata. Unknown/non-URI forms are replaced in full
+// so a legacy value cannot bypass the response boundary through parser
+// ambiguity. The returned value is a display projection, not a connectable
+// DSN; url.URL normalizes query ordering while removing secrets.
 func redactPostgresDSN(dsn string) string {
 	if dsn == "" {
 		return ""
@@ -71,7 +85,7 @@ func redactPostgresDSN(dsn string) string {
 
 	parsed, err := url.Parse(dsn)
 	if err != nil || (parsed.Scheme != "postgres" && parsed.Scheme != "postgresql") || parsed.Host == "" {
-		return ""
+		return redactedDSNPlaceholder
 	}
 
 	changed := false
--- a/openapi.yml
+++ b/openapi.yml
@@ -3192,7 +3192,8 @@ components:
             - MIRROR
           description: Ledger mode (NORMAL or MIRROR)
         mirror_source:
-          $ref: '#/components/schemas/MirrorSourceConfig'
+          allOf:
+            - $ref: '#/components/schemas/MirrorSourceConfig'
           description: Mirror source configuration with reusable credentials removed from read responses.
         mirror_sync_progress:
           $ref: '#/components/schemas/MirrorSyncProgress'
@@ -4378,7 +4379,8 @@ components:
       description: >
         Configuration for the mirror source (required when mode is MIRROR).
         Ledger read responses preserve non-secret connection metadata but omit
-        OAuth client secrets and PostgreSQL password material.
+        OAuth client secrets and PostgreSQL password material. DSNs whose form
+        cannot be safely parsed are replaced in full with `[redacted]`.
       required:
         - ledgerName
       properties:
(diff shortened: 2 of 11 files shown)


flemzord:
Revalidated live head `dd927004`. One correctness/runtime blocker remains in the new list projection: `ledgerinfo.redactingCursor` forwards only `Next` and `Close`, so it hides the optional `NextCursor()`/`grpc.UpstreamTrailer` capability of a routed `ListLedgers` cursor. On a follower, the leader can return a full page plus `x-next-cursor`; the wrapped cursor reaches EOF, `sendPagedToStream` can no longer recover that upstream trailer, and the client receives no continuation token, truncating pagination. Please forward `NextCursor()` through the redacting wrapper (as the existing log/audit projecting cursor does) and cover a routed full-page list. CI is green; this comment is based only on the pagination data-loss path. GitHub does not permit the authenticated author account to submit `CHANGES_REQUESTED` on its own PR.

paul-nicolas (commented):
Multi-model review (Claude + Codex). 3 findings posted inline: 1 Critical, 2 Medium. 3 Minor/Nit withheld from inline posting (listed below). Every posted finding was verified against the code at the PR head in a dedicated worktree.

The redaction mechanism itself is sound: `CloneVT()` before mutation is correct, `RedactSecrets(nil)` is nil-safe, the cursor wrapper preserves `Next`/`Close` semantics (`cursor.Cursor` has exactly those two methods), and the HTTP list handler's `ret[i] = ...` write is safe because `drainCursor` (internal/adapter/http/params.go:165) allocates a fresh slice per call. I probed `redactPostgresDSN` against 14 hostile DSN forms and found no URI-form password that survives it.

The problem is the boundary's scope, not its implementation — see the Critical comment.

Conflicts resolved:
- Codex rated the logs-API leak `major`; raised to **Critical** because internal/adapter/auth/scopes.go:70-77 expands the virtual `ledger:read` scope into **both** `ledger:LedgerRead` and `ledger:OpsRead`, so it is the *same principal with the same token* that the PR description names as the attacker.
- Codex claimed the DSN query-key matcher is too narrow (suggested also dropping `sslkey`, `token`, ...). **Dropped as invalid**: `sslpassword` is already caught by the `strings.Contains(key, "password")` substring test (verified by probe), `sslkey`/`sslcert` are filesystem paths rather than secret values, and `token` is not a libpq connection parameter.
- Codex rated the OpenAPI `$ref`/`description` issue `minor`; raised to **Medium** since openapi.yml:1 declares `openapi: 3.0.3` and that line is the only place this security contract is published to API consumers.

- docs/technical/contributing/api-comparison.md:329-343 -- doc drift: the mirror section documents what `GET /v3/{ledgerName}` returns for `mirrorSource` without mentioning that credentials are now stripped. CLAUDE.md requires updating this file when endpoint behavior changes.
- internal/adapter/ledgerinfo/redact_test.go:355 -- the new tests use nested `t.Run` subtests rather than the table-driven form CLAUDE.md mandates; also missing cases for a `MirrorSourceConfig` with a nil `Type` oneof, a nil `Oauth2ClientCredentials`, and cursor error propagation through `redactingCursor.Next`.
- internal/adapter/ledgerinfo/redact.go:99 -- `parsed.RawQuery = query.Encode()` sorts query keys alphabetically and rewrites a valueless `?sslmode` as `?sslmode=`, so the returned DSN is not byte-identical to the stored one whenever redaction fires. Display-only value, so harmless, but worth a comment so nobody later assumes the returned DSN is connectable.

internal/adapter/ledgerinfo/redact.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/formancehq/ledger/v3/internal/pkg/cursor"
+	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
+)
+
+// RedactSecrets returns a deep-cloned LedgerInfo safe for external read
+// responses. The stored value must retain the credentials used by mirror
+// workers, so redaction belongs at the adapter boundary and never mutates the
+// caller-owned message.
+func RedactSecrets(info *commonpb.LedgerInfo) *commonpb.LedgerInfo {
```
paul-nicolas, line 17: [Critical] The same credentials stay readable through the logs API, with the same token -- EN-1635 is not actually closed

**What's wrong.** This boundary covers `LedgerInfo`, but `LedgerInfo` is not the only outbound shape carrying `MirrorSourceConfig`. `CreatedLedgerLog` embeds it too:

- misc/proto/common.proto:653 -- `CreatedLedgerLog.mirror_source = 5`
- misc/proto/common.proto:338 -- `LogPayload.create_ledger = 1`

and that payload is returned verbatim on two untouched read paths:

- internal/adapter/grpc/server_bucket.go:1065-1077 -- `GetLog` returns `c.GetLog(ctx, req.GetSequence())` with no redaction
- internal/adapter/http/handlers_get_log.go:23-30 -- `GET /v3/_/logs/{sequence}` does `writeProtoOK(w, log)` on the raw proto

**Why it matters.** Both are gated on `ScopeOpsRead`, which looks like a different authority than `ScopeLedgersRead` -- but internal/adapter/auth/scopes.go:70-77 expands the virtual `ledger:read` scope into **both**:

```go
service + ":read": {
    ScopeLedgersRead,
    ScopeTransactionsRead,
    ScopeAccountsRead,
    ScopeAuditRead,
    ScopeOpsRead,      // <-- same token
    ScopeQueriesRead,
},
```

So the exact caller the PR description names -- "a caller with ordinary `ledger:LedgerRead` authority" -- holds a `ledger:read` token that also grants `ledger:OpsRead`, walks log sequences via `GET /v3/_/logs/{sequence}`, finds the `create_ledger` entry, and reads the plaintext OAuth client secret or the full DSN password. Same attacker, same credential, one endpoint over. The vulnerability is relocated, not fixed.

(`ListLogs` is not affected: server_bucket.go:1131-1143 only yields `Apply` payloads. `GetLog` is unfiltered and bucket-wide.)

**Fix.** Add a sibling redactor in this package for the log shape and apply it on both transports:

```go
// RedactLogSecrets returns a deep-cloned Log safe for external read responses.
func RedactLogSecrets(log *commonpb.Log) *commonpb.Log {
    if log == nil {
        return nil
    }
    redacted := log.CloneVT()
    if created := redacted.GetPayload().GetCreateLedger(); created != nil {
        created.MirrorSource = redactMirrorSource(created.GetMirrorSource())
    }
    return redacted
}
```

Factor the existing oneof switch (lines 27-36) into a shared `redactMirrorSource(*commonpb.MirrorSourceConfig)` so both entry points use one implementation, then wrap internal/adapter/grpc/server_bucket.go:1077 and internal/adapter/http/handlers_get_log.go:30. Add a regression test per transport, mirroring the ones added here.

Worth also confirming the audit-read surface (`ledger:AuditRead`) does not expose the same `CreateLedger` order bytes -- if it does, that is a third door on the same key.

Raised independently by both reviewers (Codex flagged it as `major`; raised to Critical here on the scope-expansion evidence above).

internal/adapter/ledgerinfo/redact.go
```diff
@@ -0,0 +1,102 @@
...
+// non-secret connection metadata. Unknown/non-URI forms are blanked so a
+// legacy value cannot bypass the response boundary through parser ambiguity.
+func redactPostgresDSN(dsn string) string {
+	if dsn == "" {
+		return ""
+	}
+
+	parsed, err := url.Parse(dsn)
+	if err != nil || (parsed.Scheme != "postgres" && parsed.Scheme != "postgresql") || parsed.Host == "" {
+		return ""
```
paul-nicolas: [Medium] Fail-closed blanking makes a configured DSN indistinguishable from an unconfigured one, and contradicts the contract this PR just published

**What's wrong.** Any DSN that is not a host-bearing `postgres://` URI is replaced with the empty string. Two legitimate forms hit this:

1. libpq keyword/value DSNs (`host=db.example user=mirror password=x`) -- the case the test at redact_test.go:406 covers.
2. Unix-socket URIs (`postgres:///ledger?host=/var/run/postgresql&password=x`), rejected by the `parsed.Host == ""` clause. Verified by probe: returns `""`.

Neither is rejected at admission. internal/application/admission/validate_order.go:374-377 returns early when `aws_iam_auth` is nil, so the URI-form requirement in `dsnEnforcesTLS` only applies to IAM-auth mirrors. A plain Postgres mirror created with a keyword/value DSN is admissible today and will read back as `""`.

**Why it matters.** Fail-closed is the right *security* default and I am not arguing against it. The problem is that the redacted value is ambiguous: an operator doing `GET /v3/myledger` sees `mirrorSource.postgres.dsn: ""` and cannot distinguish "credentials were stripped" from "no source is configured" or "the mirror is broken". This also makes the OpenAPI text added in this same PR (openapi.yml:4378-4381, "read responses preserve non-secret connection metadata") false for these inputs -- the response preserves nothing at all.

**Fix.** Keep fail-closed, but make it self-describing rather than empty:

```go
// redactedDSNPlaceholder marks a DSN blanked because its form could not be
// safely parsed, distinguishing it from an unset DSN.
const redactedDSNPlaceholder = "[redacted]"
```

and return that instead of `""` on the two fail-closed branches (keeping `""` only for a genuinely empty input at line 68). Then update the OpenAPI description to state that an unparseable DSN is replaced wholesale by that marker, and extend the redact_test.go:406 case to assert the marker rather than `assert.Empty`.

openapi.yml
```diff
@@ -3193,6 +3193,7 @@ components:
           description: Ledger mode (NORMAL or MIRROR)
         mirror_source:
           $ref: '#/components/schemas/MirrorSourceConfig'
+          description: Mirror source configuration with reusable credentials removed from read responses.
```
paul-nicolas, line 3197: [Medium] This `description` is silently ignored -- the security contract never reaches the generated docs

**What's wrong.** openapi.yml:1 declares `openapi: 3.0.3`. Under OpenAPI 3.0.x, `$ref` is a JSON Reference: any sibling key is discarded by conforming parsers and generators. Sibling keys alongside `$ref` only became legal in 3.1. So this line:

```yaml
mirror_source:
  $ref: '#/components/schemas/MirrorSourceConfig'
  description: Mirror source configuration with reusable credentials removed from read responses.
```

resolves to nothing more than the bare `$ref`. The redaction behavior is documented in the source file but published nowhere -- which defeats the point, since CLAUDE.md requires `openapi.yml` to carry the contract for HTTP consumers.

**Fix.** Wrap the reference so the description survives 3.0 resolution:

```yaml
mirror_source:
  allOf:
    - $ref: '#/components/schemas/MirrorSourceConfig'
  description: Mirror source configuration with reusable credentials removed from read responses.
```

The companion change at openapi.yml:4377-4381 is fine as-is -- that description sits directly on the `MirrorSourceConfig` schema, not next to a `$ref`, so it renders correctly. Worth a quick grep for other `$ref` + sibling-key pairs in this file while you are here.

Flagged by both reviewers.

flemzord:
Revalidated live head `669c3567`. The new commit fixes the routed pagination cursor, the explicit DSN redaction marker, and the OpenAPI `$ref` sibling issue, but the existing `paul-nicolas` Critical thread remains active: `GetLog`/`ListLogs` still return the raw `CreatedLedgerLog.mirror_source`, while the virtual `ledger:read` scope includes `OpsRead`, so the same caller can recover the credentials through the logs API. This remains a security/runtime blocker; I am not duplicating the inline evidence. GitHub does not allow the authenticated author account to submit `CHANGES_REQUESTED` on its own PR.

flemzord:
Revalidated live head 669c3567. One additional security/runtime blocker is not covered by the existing threads: the HTTP mirror redactor clears only oauth2_client_credentials.client_secret and leaves base_url unchanged. The mirror client parses that base URL directly into the request URL, so a supported value such as https://user:password@host carries reusable HTTP Basic credentials; GetLedger and ListLedgers therefore still return those credentials verbatim to read-scoped callers. Please redact URL userinfo in the HTTP mirror base URL and add focused GetLedger/ListLedgers coverage. I am not repeating the existing raw-log credential thread. GitHub does not permit the authenticated author account to submit CHANGES_REQUESTED on its own PR.

PR #1654, opened 2026-07-29
fix(security): redact historical credentials

flemzord:
## What changed

- project sink and mirror credentials out of historical audit and system-log read responses
- redact credentials inside serialized audit orders and signed batch payloads without mutating authoritative stored records
- preserve exact historical bytes and Ed25519 signatures when an audit entry contains no redacted credential
- document the historical boundary and conditional hash/signature verification semantics

## Root cause

Accepted `AddEventsSink` and `CreateLedger` orders contain reusable credentials. The audit chain and system logs persist those exact values, then `GetAuditEntry`, `ListAuditEntries`, `GetLog`, and `ListLogs` returned the stored records directly to read-scoped callers. The default `ledger:read` bundle includes both `AuditRead` and `OpsRead`.

The fix is intentionally placed at the shared `DefaultController` historical read boundary so HTTP, gRPC, local, routed, checkpoint, and cold-storage reads all receive deep-cloned secret-safe projections. FSM processing, accepted orders, audit hashes, idempotency hashes, checker inputs, and persisted bytes remain unchanged.

## Security details

Covered credentials include NATS URL userinfo/tokens, ClickHouse DSNs, Kafka SASL passwords, HTTP webhook HMAC secrets, Databricks PAT/OAuth secrets, PostgreSQL mirror DSNs, and mirror OAuth client secrets.

When a credential is replaced, the projected audit response is not the original hash/signature preimage. The signer key ID and redacted signed payload remain available, while the Ed25519 signature bytes are omitted to avoid implying that the projected payload verifies. Entries without credentials retain their exact serialized order, signed payload, and signature bytes.

Live event-sink configuration responses are handled separately by [PR #1651](https://github.com/formancehq/ledger/pull/1651), and live `LedgerInfo.mirror_source` responses by [PR #1653](https://github.com/formancehq/ledger/pull/1653). Both companion changes cover HTTP and gRPC, so this PR does not duplicate their adapter boundaries.

Jira: [EN-1634](https://formance-team.atlassian.net/browse/EN-1634)

## Validation

- `nix develop --command bash -c "GOROOT= go test ./internal/application/ctrl ./internal/adapter/grpc ./internal/adapter/http -count=1"`
- GitNexus `detect_changes(scope=compare, base_ref=release/v3.0)`: HIGH aggregate graph risk from six `ListLogs` storage-read traces; reviewed individually, with the change confined to the returned cursor projection after existing reads/indexing

Regression tests prove every sink/mirror credential variant is absent from projected audit orders, signed payloads, JSON output, and system logs, while authoritative controller/store objects remain unchanged. They also prove that entries without credentials preserve exact legacy order bytes, signed payload bytes, and signatures.

[EN-1634]: https://formance-team.atlassian.net/browse/EN-1634?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/technical/architecture/subsystems/api/secret-redaction.md
+++ b/docs/technical/architecture/subsystems/api/secret-redaction.md
@@ -1,0 +1,54 @@
+# Secret Redaction on Historical Reads
+
+The audit chain and system log store accepted business intent. Some accepted
+orders necessarily contain credentials used to configure event sinks or mirror
+sources. Those persisted records remain authoritative and unchanged: mutating
+them would invalidate the audit hash chain, idempotency evidence, or signed
+client payloads.
+
+Public read APIs therefore return a deep-cloned, secret-safe projection from
+the controller boundary. This boundary is shared by gRPC and the HTTP
+compatibility API, including local-leader reads, follower-routed reads,
+checkpoint reads, and cold-storage log reads.
+
+## Covered fields
+
+The projection replaces non-empty credentials with `***REDACTED***` in:
+
+- NATS sink URLs (the URL may contain user/password or token userinfo);
+- ClickHouse sink DSNs;
+- Kafka SASL passwords;
+- HTTP sink HMAC secrets;
+- Databricks PATs and OAuth client secrets;
+- PostgreSQL mirror DSNs;
+- HTTP mirror OAuth client secrets.
+
+The same transformation is applied inside:
+
+- `AddedEventsSinkLog.config` and `CreatedLedgerLog.mirror_source`;
+- `AuditItem.serialized_order` for `AddEventsSink` and `CreateLedger` orders;
+- the `SignedApplyBatch.payload` retained on an audit entry.
+
+Projection always operates on deep clones. Stored records and objects owned by
+the controller are never mutated.
+
+## Integrity semantics
+
+`AuditEntry.hash` continues to identify and protect the authoritative persisted
+entry. A redacted response is deliberately not the original hash preimage.
+Likewise, an Ed25519 signature covers the original unredacted `ApplyBatch`, not
+the projected payload. The read projection keeps the signer key ID and a
+parseable redacted payload, but omits the signature bytes so clients cannot
+mistake the projection for independently verifiable signed evidence.
+
+Integrity checking and rebuild paths read the authoritative store directly and
+are not projected. Operators who need offline verification of raw audit
+evidence must use a separately controlled administrative export rather than a
+read-scoped API.
+
+## Failure behavior
+
+If a stored order or signed batch cannot be decoded, the read fails loudly. It
+must never fall back to returning opaque bytes that might contain a secret.
+This fail-closed behavior also surfaces corrupted or schema-incompatible audit
+data instead of silently weakening redaction.
--- a/docs/technical/architecture/subsystems/api/secret-redaction.md
+++ b/docs/technical/architecture/subsystems/api/secret-redaction.md
@@ -6,10 +6,12 @@ sources. Those persisted records remain authoritative and unchanged: mutating
 them would invalidate the audit hash chain, idempotency evidence, or signed
 client payloads.
 
-Public read APIs therefore return a deep-cloned, secret-safe projection from
-the controller boundary. This boundary is shared by gRPC and the HTTP
-compatibility API, including local-leader reads, follower-routed reads,
-checkpoint reads, and cold-storage log reads.
+The historical audit and system-log APIs therefore return a deep-cloned,
+secret-safe projection from the controller boundary. The projected controller
+methods are `GetAuditEntry`, `ListAuditEntriesFrom`, `GetLog`, and `ListLogs`.
+This covers HTTP and gRPC reads served locally, checkpoint audit reads, and
+cold-storage log reads. For a follower-routed request, the serving node applies
+the projection before its gRPC response leaves that node.
 
 ## Covered fields
 
@@ -35,11 +37,13 @@ the controller are never mutated.
 ## Integrity semantics
 
 `AuditEntry.hash` continues to identify and protect the authoritative persisted
-entry. A redacted response is deliberately not the original hash preimage.
-Likewise, an Ed25519 signature covers the original unredacted `ApplyBatch`, not
-the projected payload. The read projection keeps the signer key ID and a
-parseable redacted payload, but omits the signature bytes so clients cannot
-mistake the projection for independently verifiable signed evidence.
+entry. When credentials are replaced, the response is deliberately not the
+original hash preimage. Likewise, when a signed `ApplyBatch` contains replaced
+credentials, the read projection keeps the signer key ID and a parseable
+redacted payload but omits the Ed25519 signature bytes so clients cannot mistake
+the projection for independently verifiable signed evidence. Entries without
+credentials retain their exact serialized order bytes, signed payload, and
+signature, so their evidence remains independently verifiable.
 
 Integrity checking and rebuild paths read the authoritative store directly and
 are not projected. Operators who need offline verification of raw audit
@@ -52,3 +56,13 @@ If a stored order or signed batch cannot be decoded, the read fails loudly. It
 must never fall back to returning opaque bytes that might contain a secret.
 This fail-closed behavior also surfaces corrupted or schema-incompatible audit
 data instead of silently weakening redaction.
+
+## Related live read boundaries
+
+This mechanism is intentionally limited to historical audit and system-log
+records. Live event-sink configuration reads are protected separately by
+[PR #1651](https://github.com/formancehq/ledger/pull/1651), and live
+`LedgerInfo.mirror_source` reads by
+[PR #1653](https://github.com/formancehq/ledger/pull/1653). Those adapter-level
+projections cover both HTTP and gRPC without coupling their live response
+models to the historical audit projection in this package.
--- a/internal/application/ctrl/secret_read_projection.go
+++ b/internal/application/ctrl/secret_read_projection.go
@@ -1,0 +1,209 @@
+package ctrl
+
+import (
+	"fmt"
+
+	"github.com/formancehq/ledger/v3/internal/domain/processing"
+	"github.com/formancehq/ledger/v3/internal/pkg/cursor"
+	"github.com/formancehq/ledger/v3/internal/proto/auditpb"
+	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
+	"github.com/formancehq/ledger/v3/internal/proto/raftcmdpb"
+	"github.com/formancehq/ledger/v3/internal/proto/servicepb"
+)
+
+const redactedSecret = "***REDACTED***"
+
+// projectingCursor applies a read-only projection to each item while preserving
+// the underlying cursor's lifecycle and routed pagination trailer.
+type projectingCursor[T any] struct {
+	inner   cursor.Cursor[T]
+	project func(T) (T, error)
+}
+
+func (c *projectingCursor[T]) Next() (T, error) {
+	item, err := c.inner.Next()
+	if err != nil {
+		var zero T
+
+		return zero, err
+	}
+
+	return c.project(item)
+}
+
+func (c *projectingCursor[T]) Close() error {
+	return c.inner.Close()
+}
+
+// NextCursor forwards routed gRPC pagination metadata when the wrapped cursor
+// provides it. Local cursors return an empty token.
+func (c *projectingCursor[T]) NextCursor() string {
+	if upstream, ok := c.inner.(interface{ NextCursor() string }); ok {
+		return upstream.NextCursor()
+	}
+
+	return ""
+}
+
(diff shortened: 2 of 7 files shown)


paul-nicolas (commented):

The projection itself is well built: `CloneVT` deep-copies, non-mutation is asserted, the error path is fail-closed, and the field coverage inside `SinkConfig` / `MirrorSourceConfig` is complete (I enumerated every sub-message in `misc/proto/common.proto` — no credential field is missed). Tests pass locally (`go test ./internal/application/ctrl/ -run 'Secret|Redact|Project'`, 7/7) and `go build ./...` is clean.

The problem is scope, not implementation. The PR states the root cause as "read-scoped callers can retrieve reusable credentials", but two live read endpoints still hand out the *same* credentials in plaintext to the *same* caller, in one call, without going near the audit chain. `ledger:read` expands to `{LedgersRead, TransactionsRead, AccountsRead, AuditRead, OpsRead, QueriesRead}` (`internal/adapter/auth/scopes.go:70-77`), and both endpoints below sit inside that bundle. As written, the historical hardening is bypassable, and `docs/.../secret-redaction.md` asserts coverage the code does not have.

**Severity counts:** 1 Critical, 2 Major, 2 Medium posted inline; 2 Minor withheld.

**Codex/Claude conflict resolved:** Codex flagged the `projectingCursor` wrapper as a Major routed-pagination regression. That is **wrong** and I dropped that framing — `cursor.NewClosingCursor` never implemented `NextCursor()` either, so nothing regressed, and routed cursors are produced in `internal/adapter/grpc/client_bucket.go:96-211` outside `DefaultController` entirely. The real (smaller) issue is that the forwarding is unreachable dead code and the test only passes because it substitutes its own cursor — posted as Medium instead.

- internal/application/ctrl/secret_read_projection.go:14 — `redactedSecret` duplicates `bootstrap.RedactedSecretPlaceholder` (internal/bootstrap/config.go:216) with the identical literal; share one constant (DRY).
- internal/application/ctrl/secret_read_projection_test.go — coverage is unit-level only. No e2e test proves a `ledger:read` caller cannot retrieve a credential end-to-end through gRPC/HTTP; such a test at the auth boundary would have caught the Critical below.

internal/application/ctrl/secret_read_projection.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	if createdLedger := payload.GetCreateLedger(); createdLedger != nil {
+		redactMirrorSourceSecrets(createdLedger.GetMirrorSource())
+	}
+
+	return projected, nil
+}
+
+func redactSinkConfigSecrets(config *commonpb.SinkConfig) {
```
paul-nicolas: [Critical] `GetEventsSinks` still returns every sink credential in plaintext to the same read scope

This helper is never applied to `GetEventsSinks`. `internal/application/ctrl/controller_default.go:2014` returns `query.ReadAllSinkConfigs(...)` verbatim — no clone, no redaction — and `internal/adapter/grpc/server_bucket.go:1141` gates it on `ScopeOpsRead` alone, which is exactly the scope `GetLog`/`ListLogs` require and is part of the default `ledger:read` bundle (`internal/adapter/auth/scopes.go:70-77`).

So a caller holding `ledger:read` — the very actor this PR is written to defend against — can call `GET /v3/_/events-sinks` (or the gRPC `GetEventsSinks`) and receive the NATS URL userinfo, ClickHouse DSN, Kafka SASL password, HTTP webhook HMAC secret, and Databricks PAT / OAuth client secret in cleartext. That is the live config, not a historical record, so it is strictly more useful to an attacker than the audit path this PR closes. The stated root cause is not fixed; it is only made one step less convenient.

`openapi.yml:2358-2360` still documents that endpoint with `ledger:OpsRead` and no redaction note, so the spec now contradicts the redaction story the rest of this diff tells.

Fix: in `GetEventsSinks`, project each config before returning — `cfg = cfg.CloneVT(); redactSinkConfigSecrets(cfg)` — so the same boundary covers it. Confirm the HTTP handler routes through the controller (it does, via the `Backend` interface) so one change covers both transports. If leaving it plaintext is a deliberate decision, it needs to be stated explicitly in the doc and the endpoint gated behind a write/admin scope rather than `OpsRead`.

internal/application/ctrl/secret_read_projection.go
```diff
@@ -0,0 +1,209 @@
...
+			}
+		case *commonpb.DatabricksSinkConfig_OauthM2M:
+			if auth.OauthM2M != nil && auth.OauthM2M.GetClientSecret() != "" {
+				auth.OauthM2M.ClientSecret = redactedSecret
+			}
+		}
+	}
+}
+
+func redactMirrorSourceSecrets(source *commonpb.MirrorSourceConfig) {
```
paul-nicolas: [Major] `LedgerInfo.mirror_source` is returned unredacted by `GetLedger` / `ListLedgers`

Same gap as the sink configs, for mirror credentials. `MirrorSourceConfig` is stored on `LedgerInfo` (`misc/proto/common.proto:1116`), and `DefaultController.ListLedgers` (`internal/application/ctrl/controller_default.go:166`) plus `GetLedger` return that message as-is. Both handlers gate on `ScopeLedgersRead` (`internal/adapter/grpc/server_bucket.go:675` and `:725`), which is also inside the default `ledger:read` bundle.

Result: the Postgres mirror DSN — documented in the proto itself as `postgres://user:pass@host:5432/...` — and the mirror OAuth2 `client_secret` are readable in plaintext by a read-scoped caller, for every MIRROR-mode ledger. Redacting these two fields out of `CreatedLedgerLog` while `GetLedger` serves them live means the log projection buys nothing against this caller.

Fix: apply `redactMirrorSourceSecrets` to a `CloneVT()` of the `LedgerInfo` in both `GetLedger` and the `ListLedgers` cursor (the same `newProjectingCursor` helper works for `*commonpb.LedgerInfo`). Note the mutation hazard: `ListLedgers` enriches infos in-flight, so clone before redacting or you will corrupt cached ledger state.

internal/application/ctrl/secret_read_projection.go
```diff
@@ -0,0 +1,209 @@
...
+
+	if signature := projected.GetSignature(); signature != nil && len(signature.GetPayload()) > 0 {
+		payload, err := projectSignedBatchPayloadForRead(signature.GetPayload())
+		if err != nil {
+			return nil, fmt.Errorf("projecting signed audit payload: %w", err)
+		}
+
+		// The original Ed25519 signature covers the unredacted payload. Keeping
+		// it beside projected bytes would falsely imply that those bytes verify.
+		signature.Signature = nil
```
paul-nicolas: [Major] Ed25519 signature is dropped from every signed audit entry, including entries with no credentials

`signature.Signature = nil` fires for any entry whose signed payload is non-empty, unconditionally. An audit entry for a plain `CreateTransaction` batch — no sink, no mirror, nothing redacted — loses its signature bytes purely because it passed through `GetAuditEntry` / `ListAuditEntries`.

The reasoning in the comment is sound for entries that *were* redacted: the signature covers the original bytes, so keeping it beside modified bytes would falsely imply verifiability. But the conclusion is drawn too widely. For the overwhelming majority of entries the projection is a no-op, the payload is byte-identical, and the signature would still verify — clients lose independently verifiable evidence for no security benefit. Worth noting the asymmetry: `projectSerializedOrderForRead` re-marshals through `MarshalOrderBusinessIntent`, the same function used at capture, with `MarshalDeterministicVT` — so unredacted orders round-trip byte-identical and the audit hash still recomputes. Only the signature is destroyed regardless.

Fix: have `projectSignedBatchPayloadForRead` report whether it actually changed a field (e.g. return `([]byte, bool, error)`, or compare against the input bytes). Only assign `signature.Payload` and clear `signature.Signature` when a redaction occurred; otherwise leave both untouched. The `openapi.yml:5489-5494` wording ("the signature bytes are omitted") should then be conditioned on redaction actually having happened.

docs/technical/architecture/subsystems/api/secret-redaction.md
```diff
@@ -0,0 +1,54 @@
+# Secret Redaction on Historical Reads
+
+The audit chain and system log store accepted business intent. Some accepted
+orders necessarily contain credentials used to configure event sinks or mirror
+sources. Those persisted records remain authoritative and unchanged: mutating
+them would invalidate the audit hash chain, idempotency evidence, or signed
+client payloads.
+
+Public read APIs therefore return a deep-cloned, secret-safe projection from
```
paul-nicolas: [Medium] The doc claims coverage the code does not have

"Public read APIs therefore return a deep-cloned, secret-safe projection from the controller boundary" is not true as of this diff. Four methods are projected — `GetLog`, `ListLogs`, `GetAuditEntry`, `ListAuditEntriesFrom`. `GetEventsSinks` returns plaintext `SinkConfig` and `GetLedger`/`ListLedgers` return plaintext `LedgerInfo.mirror_source` (see the two findings above), both reachable with the default `ledger:read` bundle. A reader — or a future author checking whether a new endpoint needs redaction — will take this page as a guarantee and be wrong.

The "follower-routed reads" claim on line 12 is also misleading. Routed reads never traverse this code: on a follower, `ListLogs`/`ListAuditEntries` are served by `BucketGrpcClient` (`internal/adapter/grpc/client_bucket.go:96-211`), which returns an `upstreamPeekCursor` directly. Routed responses are safe because the *serving* node projected them, which is a different mechanism than the one this sentence describes.

Fix: narrow the scope sentence to "historical audit and system-log read APIs", list the four projected methods explicitly, and add a short "Not covered" section naming `GetEventsSinks` and `GetLedger`/`ListLedgers` until they are handled. Restate the routed case as "the serving node projects before the response leaves it".

internal/application/ctrl/secret_read_projection.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func (c *projectingCursor[T]) Close() error {
+	return c.inner.Close()
+}
+
+// NextCursor forwards routed gRPC pagination metadata when the wrapped cursor
+// provides it. Local cursors return an empty token.
+func (c *projectingCursor[T]) NextCursor() string {
+	if upstream, ok := c.inner.(interface{ NextCursor() string }); ok {
```
paul-nicolas: [Medium] `NextCursor` forwarding is unreachable, and the test hides that

This type assertion can never succeed in production. Both call sites wrap `cursor.NewClosingCursor(...)` (`controller_default.go:1694` and `:1748`), and `ClosingCursor` does not implement `NextCursor() string` — the only implementor in the tree is `upstreamPeekCursor` (`internal/adapter/grpc/cursor.go:102`), which is built in the routed gRPC client, never inside `DefaultController`. So `NextCursor()` here always returns `""`.

To be explicit, since Codex read this as a pagination break: it is **not** a regression. Before this PR the same `ClosingCursor` was returned directly, equally without a trailer, and `sendPagedToStream` (`internal/adapter/grpc/stream_helper.go:87`) only ever sees a trailer-capable cursor on the routed path, which is untouched. Nothing that worked stopped working.

What is wrong is that the code and its comment describe a mechanism that cannot fire, and `TestProjectingCursorPreservesItemsErrorsAndPagination:233` asserts it works only because it injects its own `projectionTestCursor` with a `NextCursor` method in place of the production stack — a green test for a dead path.

Fix: either drop the forwarding and the comment, or invert the wrapping to `cursor.NewClosingCursor(newProjectingCursor(c, projectLogForRead), handle)` so a trailer-capable inner cursor would actually surface, and cover it with a test that goes through `NewClosingCursor` rather than around it.

flemzord:
Reviewed live head `9df80b85`. One security/runtime blocker is not covered by the existing threads: historical HTTP sink configs can still expose reusable Basic-auth credentials from `HttpSinkConfig.endpoint`. `redactSinkConfigSecrets` masks only `httpConfig.Secret`; unlike the live projection in companion PR #1651, it never redacts URL userinfo. Consequently `AddedEventsSinkLog`, `AuditItem.serialized_order`, and signed `ApplyBatch` projections can return an endpoint such as `https://user:password@host/...` unchanged through `OpsRead`/`AuditRead`. Please redact endpoint userinfo in this shared historical projection and cover the log, serialized-order, and signed-payload shapes. The prior unconditional signature removal is fixed on this head; I am not repeating the separate live-endpoint or cursor threads. GitHub does not permit the authenticated author account to submit `CHANGES_REQUESTED` on its own PR.

flemzord:
Revalidated live head 9df80b85. One additional security/runtime blocker is not covered by the existing threads: redactMirrorSourceSecrets clears only the HTTP mirror oauth2 client secret and leaves base_url unchanged. That base URL is parsed directly into mirror requests, so values such as https://user:password@host contain reusable HTTP Basic credentials. The historical CreatedLedgerLog, AuditItem serialized order, and signed ApplyBatch projections therefore still expose those credentials verbatim. Please redact HTTP mirror URL userinfo in the shared historical projection and cover all three response shapes. I am not repeating the separately reported HTTP sink endpoint or live-ledger findings. GitHub does not permit the authenticated author account to submit CHANGES_REQUESTED on its own PR.

PR #1655, opened 2026-07-30
fix(indexbuilder): clear LedgerLog volume-annotation lists on cursor reuse

Azorlogh:
## Summary

The index builder's live log cursor runs with `dal.WithReuse()` — one proto message, reset (`resetLogForReuse`) and re-unmarshalled per log. The reset truncates `Transaction.Postings` but was never extended for the `LedgerLog` volume-annotation lists (`purged_volumes` / `new_kept_volumes` / `ephemeral_volumes`). Proto3 unmarshal leaves a repeated field untouched when the wire carries no entries for it, so every log inherited — and accumulated — the eviction lists of every preceding log in the same cursor window.

`excludedForLog` consumes exactly those lists (keyed by (account, asset, color), no ledger dimension), so a log touching a cell that any earlier log in the window had purged got its posting-index writes silently skipped: the account→tx role mappings, the any-role mapping, and the has-asset row were never written. Nothing repairs them short of a full read-store rebuild.

Only the live path is affected: the posting backfill parses raw wire bytes into its own `parsedLog`, which truncates the lists correctly on every parse.

## Reproduction

Found by the model-based antithesis driver's address-window validation (singleton_driver_model, EN-1625): a force-drain re-touching a just-purged ephemeral cell in the next log lost its source-address row, while the FSM's on-disk `LedgerLog` lists and a boot-time read-store rebuild were both provably correct.

**Deterministic e2e regression** (`index_cursor_reuse_leak_test.go`): one cross-ledger bulk pairs an ephemeral wash on ledger A (log annotated `purged=[e:1 USD]`) with a plain funding of the same-named account on ledger B (clean log). The two logs are adjacent in one proposal, hence always read back-to-back through the reused cursor message — pre-fix, B's funding deterministically inherits A's purged entry and loses its destination-address row (test times out red); post-fix it passes.

## Fix

Truncate the three lists in `resetLedgerLog`, same idiom as `Postings[:0]`. Red→green at both levels: `TestResetLogForReuse_ClearsVolumeAnnotationLists` (unit) and the e2e above.

### DIFF
--- a/internal/application/indexbuilder/proto_reset.go
+++ b/internal/application/indexbuilder/proto_reset.go
@@ -68,6 +68,13 @@ func resetLogPayload(p *commonpb.LogPayload) {
 
 func resetLedgerLog(ll *commonpb.LedgerLog) {
 	ll.Id = 0
+	// The volume-annotation lists are absent from the wire on most logs, so a
+	// leftover would survive the next unmarshal and poison excludedForLog —
+	// the exclusion set that gates every posting-index write. Truncate to keep
+	// the backing arrays (same idiom as Transaction.Postings).
+	ll.PurgedVolumes = ll.GetPurgedVolumes()[:0]
+	ll.NewKeptVolumes = ll.GetNewKeptVolumes()[:0]
+	ll.EphemeralVolumes = ll.GetEphemeralVolumes()[:0]
 	// Preserve ll.Date (always present in apply logs).
 	// Preserve ll.Data (always present).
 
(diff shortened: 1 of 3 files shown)


paul-nicolas (approved):

flemzord:
Reviewed live head `b8941e41`. The reset now clears all three reused LedgerLog volume-annotation slices before unmarshalling, and the focused unit plus cross-ledger E2E regression cover the stale-exclusion/index-loss failure mode. Functional CI, Codecov, and both review checks are green; I found no correctness/runtime blocker. GitHub approval is intentionally disabled by this cron configuration.

PR #1656, opened 2026-07-31
test(antithesis): negative-space read probes for the model driver

Azorlogh:
## What

The model driver's reads only ever targeted state the model already holds, so every read RPC was structurally blind to server state the model lacks (a stranded row, a ghost account, a ledger the model never created). This adds negative-space probes to all four read RPCs, validated through the existing candidate-base-aware validators (absent = empty, so the server must answer NotFound/empty and anything served is a finding):

| Read RPC | negative-space axes |
|---|---|
| `GetAccount` | model-absent account (5%) + absent ledger (2%) |
| `GetTransaction` | absent ledger (2%) — unassigned-id slack already existed |
| `GetLedger` | absent ledger (2%) |
| `GetMetadataSchemaStatus` | absent ledger (3%) |

Rates draw from the Antithesis-controlled source (`percentChance`), so the fuzzer can steer them.

## Why it matters

The point probe is the only detector for the stranded-row bug class that hides from list validation (point and list reads take different server code paths). Validated end-to-end: against a server with the #1645 fix reverted, the model-absent GetAccount probe finds the ghost account (`serverIn == serverOut`, model holds nothing) and fail-fast stops the run — see `scratch/model_test_distributions.md` for the detection-time analysis.

## Verification

- Unit tests for the pickers (`reads_test.go`).
- 30-min single-node run: PASS, all five coverage reachables fired, zero unreachability hits.
- Pre-#1645 A/B: finding surfaced in 18.4 min, exact expected signature.

### DIFF
--- a/tests/antithesis/workload/bin/cmds/model/singleton_driver_model/reads.go
+++ b/tests/antithesis/workload/bin/cmds/model/singleton_driver_model/reads.go
@@ -18,12 +18,13 @@ import (
 	"github.com/formancehq/ledger/v3/tests/antithesis/workload/internal"
 )
 
-// runRead picks a known account, issues a linearizable GetAccount, and validates
-// the result — the picked asset's volumes and the account's whole metadata map —
-// against the model (see validateAccountRead).
+// runRead picks a read target — usually a known account, sometimes one the model
+// holds no state for — issues a linearizable GetAccount, and validates the result
+// (the picked asset's volumes and the account's whole metadata map) against the
+// model (see validateAccountRead).
 func runRead(ctx context.Context, client servicepb.BucketServiceClient, c *Checker) {
 	c.mu.Lock()
-	ledger, addr, asset, ok := pickCell(c.modelState)
+	ledger, addr, asset, absent, ok := pickReadTarget(c.modelState, c.ledgerNames)
 	if !ok {
 		c.mu.Unlock()
 		return
@@ -31,6 +32,14 @@ func runRead(ctx context.Context, client servicepb.BucketServiceClient, c *Check
 	readID := c.registerRead()
 	c.mu.Unlock()
 	defer c.finishRead(readID)
+
+	if absent {
+		// Coverage: GetAccount is probing the negative space pickCell can't reach —
+		// an account the model holds no state for. The server must report it empty;
+		// a returned cell or metadata no candidate base explains is how this path
+		// catches server state the model lacks (validateAccountRead).
+		assert.Reachable("singleton_driver_model: GetAccount probing a model-absent account", internal.Details{"ledger": ledger})
+	}
 
 	// Be explicit about consistency so the test still validates the
 	// property it cares about if the server-side default ever changes.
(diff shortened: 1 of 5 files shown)


tests/antithesis/workload/bin/cmds/model/singleton_driver_model/reads.go
```diff
@@ -18,12 +19,13 @@ import (
...
-// the result — the picked asset's volumes and the account's whole metadata map —
-// against the model (see validateAccountRead).
+// runRead picks a read target — usually a known account, sometimes one the model
+// holds no state for — issues a linearizable GetAccount, and validates the result
+// (the picked asset's volumes and the account's whole metadata map) against the
+// model (see validateAccountRead).
 func runRead(ctx context.Context, client servicepb.BucketServiceClient, c *Checker) {
 	c.mu.Lock()
-	ledger, addr, asset, ok := pickCell(c.modelState)
+	ledger, addr, asset, absentAccount, absentLedger, ok := pickReadTarget(c.modelState, c.ledgerNames)
```
NumaryBot, line 27: 🟠 **[major] Validate all volumes returned by absent account probes**

When the new GetAccount negative-space probe hits an absent/model-absent target, a buggy server can still return an account whose only ghost volume is under a different asset than the randomly selected `asset`. `accountAssetVolumes` then reports `found=false`, and `validateAccountRead` can match the empty model as long as metadata is empty, so the probe misses the stranded-row class it is meant to detect two-thirds of the time with the current asset set. The absent-account/absent-ledger path needs to treat any unexplained returned volume (not just the selected asset) as a finding.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1656#issuecomment-5142986165

NumaryBot (commented):

The PR adds negative-space read probes for the model driver, but the GetAccount absent-account/absent-ledger path contains a coverage gap. The probe validates only the randomly sampled asset when checking whether a returned account is a ghost. A buggy server can serve a ghost account whose volume sits under any other workload asset; because only the sampled asset is inspected, `accountAssetVolumes` reports `found=false`, the empty-model candidate matches, and the stranded-row class the probe is designed to catch is missed approximately two-thirds of the time. The fix is to reject any returned volume on the absent-account path rather than restricting the check to one asset.

### Findings outside the diff

🟠 **[major] Absent-account probe only validates the sampled asset, missing ghost volumes under other assets** — `tests/antithesis/workload/bin/cmds/model/singleton_driver_model/reads.go:81` <sub>(NumaryBot, codex)</sub>

When the negative-space GetAccount probe hits an absent account (or absent ledger), it only checks the randomly selected `asset` via `accountAssetVolumes`. A buggy server can return a ghost account whose volume lives under a different workload asset; in that case `accountAssetVolumes` reports `found=false`, the empty model candidate matches (as long as metadata is empty), and the stranded-row class the probe is meant to detect is silently accepted. With the current asset set this causes the probe to miss the defect roughly two-thirds of the time.

**Suggestion:** In the absent-account/absent-ledger branches, reject any response that contains *any* returned volume regardless of asset, or iterate over all assets and validate each one. The guard should be: if the returned account has any non-zero volume entries at all, immediately fail the probe rather than only checking the sampled asset.



Summary: https://github.com/formancehq/ledger/pull/1656#issuecomment-5142986165

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/reads.go
Azorlogh: Fixed in 4a787eaf9 — account reads now extract the full uncolored volume set and require set-equality against each candidate base, so a ghost row under any asset (and an omitted cell) is a finding; colored buckets / unparseable amounts are unexplainable outright.

flemzord:
Revalidated current head `4a787eaf`. The earlier negative-space coverage gap is fixed: account reads now compare the complete returned volume set against each candidate model state, and functional CI plus the current bot checks are green. I found no correctness/runtime blocker. GitHub approval is intentionally disabled by this cron configuration.

paul-nicolas (approved):

PR #1650, opened 2026-07-29
test(antithesis): model-driver idempotency replay + conflict coverage

Azorlogh:
Extends the forward-model linearizability driver (`singleton_driver_model`) with idempotency coverage:

- Keyed bulks: a configurable fraction of driver bulks carry an idempotency key.
- Replay: accepted keyed bulks are re-submitted and must return the frozen original result — validated read-style against the original sequences, so a replay that re-executes (instead of replaying) is caught.
- Conflicts: re-using a key with a different payload must fail with `IDEMPOTENCY_KEY_CONFLICT`.
- Oracle: models keyed replay for successful bulks; frozen outcomes are also checked across a restore cycle (guards the RebuildDelta idempotency restore, #1607).

Known follow-up (deliberately out of scope here): failure-freezing — replaying a keyed bulk whose original outcome was a business failure is not yet modeled.

PR #1657, opened 2026-07-31
perf(antithesis): immutable fingerprinted oracle state — ~19x model-driver throughput

Azorlogh:
> Stacked on #1656 (the probes PR) — only the last 5 commits are new here; GitHub retargets to `release/v3.0` once #1656 merges.

## What

The model checker's serialization search (`candidateBases`) paid two full-state walks per explored base: `Apply`'s deep clone of every model map, and a full canonical re-render (fmt + sorts + uint256 decimal) as the sha256 dedup key — 6.7s and 5.8GB allocated per worst-case search at steady-state scale, growing with run length.

- **`tests/oracle/pmap.go`**: persistent collections (benbjohnson/immutable B+trees) carrying an incrementally-maintained 128-bit multiset fingerprint — forking a state is a struct copy, a mutation is O(log n), the dedup identity is read in O(1). `clone()` and its enumerate-every-field footgun are deleted; the compiled chart is memoized per state. Sorted iteration makes generator picks deterministic (Antithesis-replayable) by construction.
- **search dedup** keys on `{fingerprint, pendingIdx, inflight-bitmask}` — no hashing in the search at all.
- **Lock-free generation**: workers snapshot the committed state under `c.mu` (O(1) struct copy — a published state is never mutated) and generate outside it; generator picks seek a random probe key into the sorted collections instead of collecting whole tables; receipts move behind a leaf mutex. Race test covers the sharing pattern.
- **Workers run unpaced** (the 100ms/iteration sleep is gone) — RPC round-trips and channel back-pressure are the only throttles.

## Numbers

| Metric (steady-state scale: 4 ledgers, 4k-tx logs) | Before | After |
|---|---|---|
| Full worst-case search | 6.74 s / 5.8 GB alloc | 9.4 ms / 18 MB (~700x) |
| One Apply (fork + fold) | 631 µs | 13 µs |
| Dedup key per node | 7.1 ms | 1.6 µs |
| Cluster bulk throughput (3-node, 8 workers) | ~58 bulks/s | ~1,074 bulks/s (~19x) |

Fingerprint soundness: multiset sum of sha256-derived 128-bit terms — order-independent, collision-infeasible for the model's non-adversarial inputs (documented in pmap.go, with algebra unit tests: path independence, fork isolation, commuting-bulk dedup, the empty-ledger guard).

## Verification

- Oracle behavior unchanged: all pre-existing oracle/driver unit tests pass unmodified; search enumerates the identical base set (463 bases) on the benchmark scenario.
- 30-min single-node run: PASS, ~1.19M bulks, flat rate start-to-finish (the old grows-with-run-length slowdown is gone).
- Pre-#1645 detector A/B at full speed: the probe finds the reverted bug in 18 min (~3.6 h of pre-change wall clock).
- `go test -race` on both modules.

## Known limits (measured, out of scope here)

- Single-node server saturates ~700–800 committed bulks/s at ~33% CPU regardless of storage (tmpfs A/B) — the sequential propose→apply pipeline paces it; driver improvements pay off once that ceiling rises (multi-node already does ~1,074/s).
- The search is exponential in the in-flight window: at `MODEL_WORKERS=64` the checker wedges (dispatch dies while the run still reports PASS). Keep ≤32 workers; a loud fold-window guard is a candidate follow-up.


go.mod
```diff
@@ -126,6 +126,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sso v1.30.15 // indirect
 	github.com/aws/aws-sdk-go-v2/service/ssooidc v1.35.19 // indirect
 	github.com/bahlo/generic-list-go v0.2.0 // indirect
+	github.com/benbjohnson/immutable v0.4.3 // indirect
```
NumaryBot: 🟠 **[major] Mark immutable as a direct dependency**

`tests/oracle/pmap.go` imports `github.com/benbjohnson/immutable` directly in the root module, but this entry is added in the indirect block. Running the repository's required `go mod tidy` will rewrite `go.mod` to make this a direct requirement, so the pre-commit/tidy check will fail until this dependency is moved to the direct require block.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1657#issuecomment-5143012294

Azorlogh:
Rebased on current release/v3.0 (the stale merge-base was showing already-merged commits). One substantive integration: #1650 added a frozen-idempotency table to the oracle that is deliberately part of the search's dedup identity — it's now a fingerprinted persistent map (`frozenOutcomeTerm` preserves the exact key→per-order-id identity the old hash section encoded), so keyed bulks don't reintroduce an O(#keys) copy per Apply. Verified on the rebased chain: race tests, same benchmark profile (8.8ms / 463 bases), and a 120s smoke run at full throughput with the replay/conflict reachables firing.

go.mod
NumaryBot: 🟠 **[major] Direct import listed as indirect dependency**

`tests/oracle/pmap.go` imports `github.com/benbjohnson/immutable` directly within the root module, but the `go.mod` entry sits in the indirect block. Running `go mod tidy` (which the repository's pre-commit checks require) will promote it to a direct dependency, causing the tidy check to fail until this is corrected.

**Suggestion:** Move the `github.com/benbjohnson/immutable` entry from the `// indirect` block to the direct `require` block in `go.mod`, then run `go mod tidy` to confirm the file is clean.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1657#issuecomment-5143012294

go.mod
NumaryBot: 🟠 **[major] Direct import listed as indirect dependency**

`tests/oracle/pmap.go` imports `github.com/benbjohnson/immutable` directly in the root module, but the `go.mod` entry sits in the `// indirect` block. The repository's mandatory `go mod tidy` / pre-commit check will promote it to a direct dependency, causing the tidy check to fail until corrected.

**Suggestion:** Move the `github.com/benbjohnson/immutable` entry from the `// indirect` block to the direct `require` block in `go.mod`, then run `go mod tidy` to confirm the file is clean.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1657#issuecomment-5143012294

go.mod
Azorlogh: Fixed in 0256a4dcd276b88a34ea34d39042f2defa18e785 — the dependency moved to the direct require block (amended into the commit that introduces pmap.go, so the tidy/Dirty check passes at every commit of the chain).

Azorlogh: stale — duplicate of https://github.com/formancehq/ledger/pull/1657#discussion_r3690512934, fixed there.

Azorlogh: stale — duplicate of https://github.com/formancehq/ledger/pull/1657#discussion_r3690512934, fixed there.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1657#issuecomment-5143012294

paul-nicolas (dismissed):

flemzord:
Revalidated live head `4a328d74`. The immutable collections preserve fork isolation, all state-bearing collections feed the fingerprint, the search key retains pending/in-flight continuation state, and the change is confined to the Antithesis/model oracle path. NumaryBot and the complete functional/coverage CI are green. I found no correctness/runtime blocker on this head. GitHub approval is intentionally disabled for this cron task.

paul-nicolas (approved):

PR #1659, opened 2026-07-31
test(antithesis): model-driver query coverage — filters, indexes, typed metadata (EN-1625)

Azorlogh:
## Query coverage (EN-1625)

Extends `singleton_driver_model` and `tests/oracle` to validate filtered, paginated `ListAccounts`, `ListTransactions`, and `ListLogs` as exact ordered windows against candidate oracle states. Coverage includes cursors/page sizes/reverse where supported, boolean filter composition, index create/drop/recreate, per-replica readiness, transaction builtins, address roles, and typed metadata on accounts and transactions.

- Readiness polls are locally served and bound to the sampled index creation frontier. Every replica must have folded that incarnation before promotion.
- The oracle records end-of-bulk account/asset and account/transaction memberships under the committed exclusion verdict, and hashes those memberships when deduplicating candidates.
- Server-assigned transaction/log dates are learned from commit responses and checked. Unknown dates in in-flight candidates use three-valued filter evaluation and exact optional-row truncation; known values require equality.
- Metadata filters use the public coercion rules. Undeclared index creation and kind-mismatched filters are probed on both entity targets. Indexed-key retypes retain their serving window until the per-replica two-phase `last_indexed_sequence` / `pending_version` proof closes it. Not-ready tolerance during a window is restricted to the index named by the rejection; an unrelated active sibling must still produce a finding.
- `INDEX_BUILDING` and `INDEX_NOT_FOUND` are validated before generic transport transients and require the documented gRPC code and reason.

The broader Jira ticket also lists aggregate reads; this PR preserves the existing filtered-list/index/typed-metadata scope and does not claim new `AggregateVolumes`/`GetLedgerStats` coverage.

## Server defect exposed by Tests-Model

The log-date backfill skipped schema, account-type, and index-lifecycle logs while unfiltered `ListLogs` included them. For example, `NOT(ledger=current AND logDate>3)` returned configuration logs even though every committed date exceeded 3. The driver correctly rejected that page.

The fix for it lives in #1960 (EN-1987), so this PR is test-only again: its server commit was reverted in 5213f1a5c. `Tests-Model` here stays red until #1960 lands.

## Harness scope

- Bearer `ANTITHESIS_API_KEY` / `ANTITHESIS_TENANT` launch authentication, `k8s-launch-minutes`, `k8s-run-minutes`, `check-run`, and the 72-hour `check-k8s-image-freshness` gate prevent soaks against stale images.
- `env -u SOURCE_DATE_EPOCH` applies to five Justfile build recipes (seven Docker build invocations) so reproducible-build timestamps do not invalidate freshness checks.
- Local model servers enable Pebble `invariants`; `MODEL_DUMP_BATCHES` supports offline replay. Kubernetes `MODEL_DEBUG=1` preserves per-bulk diagnostics; the workload ignore entry excludes its generated binary.
- `WRITES_BLOCKED_DISK_FULL` is a code-and-reason-checked transient for every workload, reflecting fault-injected disk pressure.

## Base and prior fixes

Rebased onto `release/v3.0` at `37ccf5238ed460e0df2b9844869bdb9c7031f532` (2026-09-09). Prior server fixes are already in the base, including the iterator fixes, index-registry restore, version-bound metadata types, degenerate ranges, and fixed-horizon read alignment. The removed `min_log_sequence` field and `READ_INDEX_NOT_CAUGHT_UP` transient are not reintroduced. There is no dependency on the closed, unmerged #1817.

Jira: [EN-1625](https://formance-team.atlassian.net/browse/EN-1625)

## Validation

- `bash scripts/agent-check`: PASS.
- `AI_REVIEW_BASE_SHA=37ccf5238ed460e0df2b9844869bdb9c7031f532 bash scripts/agent-check-pr`: PASS (generation/tidy/lint, invariants, root build, focused indexbuilder/oracle race suites).
- Pinned Nix workload module `go test -race ./...`: PASS.
- Pinned Nix `just test-model-cluster 180`: PASS, 3 nodes, 4 completed restart/recovery cycles, no findings.
- Deterministic server regressions: failed before the fix, passed after it. Mutation checks prove the retype attribution and both-target generator regressions fail when their respective fixes are removed.
- Final diff/whitespace review: PASS; no generated drift; working tree clean.

CI on `350b4b188`: [Default run 34372646436](https://github.com/formancehq/ledger/actions/runs/34372646436) completed successfully, including `Tests-Model`, both E2E suites, `Tests`, `Dirty`, workload tests, build, Schemathesis and Coverage. Repository invariants and NumaryBot review are also green. Release/image jobs were not applicable and were skipped by the existing workflow conditions. GitHub still reports `CHANGES_REQUESTED`; reviewer sign-off is the remaining merge gate.

[EN-1625]: https://formance-team.atlassian.net/browse/EN-1625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/antithesis/k8s/workload.yaml
+++ b/tests/antithesis/k8s/workload.yaml
@@ -99,6 +99,11 @@ spec:
           value: "ledger-ledger-0.ledger-ledger-headless.default.svc.cluster.local:8888,ledger-ledger-1.ledger-ledger-headless.default.svc.cluster.local:8888,ledger-ledger-2.ledger-ledger-headless.default.svc.cluster.local:8888,ledger-ledger-3.ledger-ledger-headless.default.svc.cluster.local:8888,ledger-ledger-4.ledger-ledger-headless.default.svc.cluster.local:8888,ledger-ledger-5.ledger-ledger-headless.default.svc.cluster.local:8888,ledger-ledger-6.ledger-ledger-headless.default.svc.cluster.local:8888"
         - name: EXPECTED_VOTERS
           value: "3"
+        # Per-bulk driver decisions (dispatch, outcome, model verdict) in the
+        # container log, so a divergence finding carries the driver's view of
+        # every observation in the moment logs.
+        - name: MODEL_DEBUG
+          value: "1"
         # Dumps every committed ApplyRequest (base64) for exact deterministic
         # replay through the oracle (see tests/oracle/cmd/replay). Empty
         # disables; substituted by generate-manifests.sh from the
--- a/tests/antithesis/workload/bin/cmds/model/singleton_driver_model/coverage.go
+++ b/tests/antithesis/workload/bin/cmds/model/singleton_driver_model/coverage.go
@@ -101,33 +101,47 @@ func emitCoverage(cond bool, msg string, details internal.Details, hit bool) {
 		hit, true, "sometimes", "Sometimes", msg)
 }
 
-// noteQueryCoverage records what one validated query outcome proves. served is
-// true only when the oracle accepted the outcome AND the server returned rows:
-// a refusal the model predicted shows the lifecycle gate works, not that the
-// index can answer. needed is the filter's index requirement, so a sonde can
-// only be satisfied by a query that genuinely depended on that index.
-//
-// Every sonde is evaluated on every call, not only the satisfied one — the
-// false evaluations are what let Antithesis bias toward a starved index. The
-// caller must not hold c.mu.
-func (c *Checker) noteQueryCoverage(ledger string, target commonpb.QueryTarget, filter *commonpb.QueryFilter, needed map[string]struct{}, served bool) {
+// noteQueryCoverage records what one validated query outcome proves. Every
+// sonde is evaluated on every call, not only the satisfied one — the false
+// evaluations are what let Antithesis bias toward a starved index. The caller
+// must not hold c.mu.
+func (c *Checker) noteQueryCoverage(ledger string, target commonpb.QueryTarget, filter *commonpb.QueryFilter, needed map[string]struct{}, verified bool, rows int) {
 	details := internal.Details{"ledger": ledger}
+
+	for msg, cond := range coverageHits(target, filter, needed, verified, rows,
+		c.retypeWindowOpenFor(ledger, needed)) {
+		emitCoverage(cond, msg, details, coverageHit)
+	}
+}
+
+// coverageHits is the per-sonde verdict for one query outcome: message to
+// whether this outcome satisfies it.
+//
+// A sonde needs three things at once. The oracle must have ACCEPTED the
+// outcome; the server must have returned at least one ROW, because a verified
+// empty page proves the index was consulted but never that it can produce a
+// matching record — and the generator emits deliberately unmatchable filters,
+// so empty pages are the common case; and the filter must have NEEDED that
+// index, so a page served through one index cannot vouch for another.
+func coverageHits(target commonpb.QueryTarget, filter *commonpb.QueryFilter, needed map[string]struct{}, verified bool, rows int, retypeOpen bool) map[string]bool {
+	served := verified && rows > 0
+	out := make(map[string]bool, len(coverageIndexes)+2)
 
 	for _, wi := range coverageIndexes {
 		_, want := needed[wi.canonical]
-		emitCoverage(served && want, coverageIndexMessage(wi.canonical), details, coverageHit)
+		out[coverageIndexMessage(wi.canonical)] = served && want
 	}
 
 	// Only the entity targets have metadata fields; a LOGS filter cannot need a
 	// metadata index, so folding it in would report against an entity sonde it
 	// can never satisfy.
 	if target != commonpb.QueryTarget_QUERY_TARGET_LOGS {
-		emitCoverage(served && filterNeedsMetadataIndex(filter),
-			coverageMetadataMessage(target), details, coverageHit)
+		out[coverageMetadataMessage(target)] = served && filterNeedsMetadataIndex(filter)
 	}
 
-	emitCoverage(served && c.retypeWindowOpenFor(ledger, needed),
-		coverageRetypeMessage, details, coverageHit)
+	out[coverageRetypeMessage] = served && retypeOpen
+
+	return out
 }
 
 // filterNeedsMetadataIndex reports whether any leaf is a metadata-field
--- a/tests/antithesis/workload/bin/cmds/model/singleton_driver_model/coverage.go
+++ b/tests/antithesis/workload/bin/cmds/model/singleton_driver_model/coverage.go
@@ -1,0 +1,165 @@
+package main
+
+import (
+	"github.com/antithesishq/antithesis-sdk-go/assert"
+
+	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
+
+	"github.com/formancehq/ledger/v3/tests/antithesis/workload/internal"
+)
+
+// Coverage sondes prove the query oracle actually predicted a served response
+// for each index it models, rather than merely compiling code that could have.
+// A run in which an index never served a page the oracle checked says nothing
+// about that index, so each sonde is a must-hit `Sometimes`: satisfied the
+// first time that index serves a verified page, and evaluated false on every
+// other query so Antithesis has a gradient to steer along. `Reachable` cannot
+// do this — its condition is hard-wired true, so it never produces the failing
+// evaluations the platform's guidance biases on.
+//
+// The names are data-driven (one per index), so the antithesis-go-instrumentor
+// cannot catalogue them: it only resolves literal message arguments. They are
+// registered at run time through assert.AssertRaw instead — the same not-hit
+// emission the generated catalog performs for literal assertions — and the
+// hit-time emissions reuse the identical message/ID so they land on the
+// registered property. This mirrors internal/block/block.go, and AssertRaw is
+// invisible to the scanner, so no anonymous catalog entries are produced.
+const (
+	coverageClass = "github.com/formancehq/ledger/v3/tests/antithesis/workload/bin/cmds/model/singleton_driver_model"
+	coverageFile  = "tests/antithesis/workload/bin/cmds/model/singleton_driver_model/coverage.go"
+
+	coverageHit    = true
+	coverageNotHit = false
+)
+
+// coveragePrefix marks every sonde below. run_model_test.sh keys its
+// "registered but never satisfied" gate on it, and TestCoverageMessages pins
+// that no sonde escapes the prefix.
+const coveragePrefix = "singleton_driver_model: coverage "
+
+// coverageIndexes is the static index set the sondes are keyed on, resolved
+// once: noteQueryCoverage runs on every query.
+var coverageIndexes = workloadIndexes()
+
+// coverageIndexMessage names the per-index sonde. The canonical IndexID is the
+// oracle's own index-map key, so the sonde and the lifecycle it proves are
+// keyed identically.
+func coverageIndexMessage(canonical string) string {
+	return coveragePrefix + "index " + canonical + " served a model-verified page"
+}
+
+// coverageMetadataMessage names the per-target metadata-index sonde. Metadata
+// indexes are one per declared field, a set that churns as the schema changes,
+// so they are covered by shape rather than by identity: an unbounded, unstable
+// family of names cannot be registered up front.
+func coverageMetadataMessage(target commonpb.QueryTarget) string {
+	return coveragePrefix + "a metadata-field index served a model-verified page on " + coverageTargetName(target)
+}
+
+// coverageTargetName names one of the two entity targets; metadata sondes exist
+// for no others.
+func coverageTargetName(target commonpb.QueryTarget) string {
+	if target == commonpb.QueryTarget_QUERY_TARGET_TRANSACTIONS {
+		return "transactions"
+	}
+
+	return "accounts"
+}
+
+// coverageRetypeMessage names the sonde for the retype exception: a page served
+// while one of its indexes still had an open retype window. That window is the
+// subtlest state the oracle models — the served version may still be bound to a
+// superseded declared type — and nothing else proves the exception was
+// exercised rather than merely compiled.
+const coverageRetypeMessage = coveragePrefix + "a query was served while a retype window was open"
+
+// coverageMessages is every sonde this driver registers, in registration order.
+func coverageMessages() []string {
+	out := make([]string, 0, len(coverageIndexes)+3)
+	for _, wi := range coverageIndexes {
+		out = append(out, coverageIndexMessage(wi.canonical))
+	}
+
+	return append(out,
+		coverageMetadataMessage(commonpb.QueryTarget_QUERY_TARGET_ACCOUNTS),
+		coverageMetadataMessage(commonpb.QueryTarget_QUERY_TARGET_TRANSACTIONS),
+		coverageRetypeMessage,
+	)
+}
+
+// registerCoverage declares every sonde before the run loop starts, so one that
+// is never evaluated is still visible as an unsatisfied property instead of
+// being absent from the output altogether.
+func registerCoverage() {
+	for _, msg := range coverageMessages() {
+		emitCoverage(false, msg, nil, coverageNotHit)
+	}
+}
+
+func emitCoverage(cond bool, msg string, details internal.Details, hit bool) {
+	assert.AssertRaw(cond, msg, details, coverageClass, "noteQueryCoverage", coverageFile, 0,
+		hit, true, "sometimes", "Sometimes", msg)
+}
+
+// noteQueryCoverage records what one validated query outcome proves. served is
+// true only when the oracle accepted the outcome AND the server returned rows:
+// a refusal the model predicted shows the lifecycle gate works, not that the
+// index can answer. needed is the filter's index requirement, so a sonde can
+// only be satisfied by a query that genuinely depended on that index.
+//
+// Every sonde is evaluated on every call, not only the satisfied one — the
+// false evaluations are what let Antithesis bias toward a starved index. The
+// caller must not hold c.mu.
+func (c *Checker) noteQueryCoverage(ledger string, target commonpb.QueryTarget, filter *commonpb.QueryFilter, needed map[string]struct{}, served bool) {
+	details := internal.Details{"ledger": ledger}
+
+	for _, wi := range coverageIndexes {
+		_, want := needed[wi.canonical]
+		emitCoverage(served && want, coverageIndexMessage(wi.canonical), details, coverageHit)
+	}
+
+	// Only the entity targets have metadata fields; a LOGS filter cannot need a
+	// metadata index, so folding it in would report against an entity sonde it
+	// can never satisfy.
+	if target != commonpb.QueryTarget_QUERY_TARGET_LOGS {
+		emitCoverage(served && filterNeedsMetadataIndex(filter),
+			coverageMetadataMessage(target), details, coverageHit)
+	}
+
+	emitCoverage(served && c.retypeWindowOpenFor(ledger, needed),
+		coverageRetypeMessage, details, coverageHit)
+}
+
+// filterNeedsMetadataIndex reports whether any leaf is a metadata-field
+// condition — the leaves neededIndexCanonicals maps to a per-(target, key)
+// metadata index. Read off the filter rather than sniffed out of the canonical
+// strings, so it stays exact as the IndexID encoding changes.
+func filterNeedsMetadataIndex(f *commonpb.QueryFilter) bool {
+	found := false
+
+	visitLeaves(f, func(leaf *commonpb.QueryFilter) {
+		if _, ok := leaf.GetFilter().(*commonpb.QueryFilter_Field); ok {
+			found = true
+		}
+	})
+
+	return found
+}
+
+// retypeWindowOpenFor reports whether any of the query's indexes still had an
+// open retype window on the committed state. The window is read there rather
+// than from the candidate base the match came from, which matchesModel does not
+// report; the committed state is the same source the finding diagnostics use.
+func (c *Checker) retypeWindowOpenFor(ledger string, needed map[string]struct{}) bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	ls := c.modelState.Ledger(ledger)
+	for canon := range needed {
+		if _, open := ls.RetypeWindow(canon); open {
+			return true
+		}
+	}
+
+	return false
+}
--- a/tests/antithesis/workload/bin/cmds/model/singleton_driver_model/indexes.go
+++ b/tests/antithesis/workload/bin/cmds/model/singleton_driver_model/indexes.go
@@ -1,0 +1,1206 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/antithesishq/antithesis-sdk-go/assert"
+	"github.com/antithesishq/antithesis-sdk-go/random"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/formancehq/ledger/v3/internal/domain/indexes"
+	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
+	"github.com/formancehq/ledger/v3/internal/proto/servicepb"
+	"github.com/formancehq/ledger/v3/tests/oracle"
+
+	"github.com/formancehq/ledger/v3/tests/antithesis/workload/internal"
+)
+
+// Index lifecycle. The query surface has index-backed conditions the compiler
+// serves only from created indexes: AccountHasAsset (account-by-asset builtin)
+// on accounts, and reference / date builtins on transactions. To exercise those
+// paths with validated results (not just the not-found rejection), the
+// generator emits CreateIndex / DropIndex bulks for each (workloadIndexes) and
+// the model tracks each index's lifecycle:
+//
+//   - absent — no index. A has-asset query must be rejected (FailedPrecondition).
+//   - ambiguous — created, but not yet confirmed READY on every replica. A query
+//     may return a validated window OR be rejected not-ready, both legal (a
+//     replica whose initial backfill hasn't flipped CurrentVersion>0 rejects it).
+//   - active — the readiness poller confirmed CurrentVersion>0 on every reachable
+//     replica. A query must return a validated window; a not-ready rejection is
+//     now a finding.
+//
+// CreateIndex adds the index ambiguous; the poller promotes it to active once
+// every replica reports it live, and demotes it back to ambiguous whenever a
+// replica reports it not-ready again (a restored node rebuilding its read-store).
+// DropIndex removes it instantly — the drop order sits in the committed prefix a
+// linearizable read issued after its response must observe.
+
+// assetIndexID is the account-by-asset builtin index the has-asset condition is
+// served from. assetIndexCanonical is its stable map key.
+func assetIndexID() *commonpb.IndexID {
+	return indexes.AccountBuiltinID(commonpb.AccountBuiltinIndex_ACCT_BUILTIN_INDEX_ASSET)
+}
+
+var assetIndexCanonical = indexes.Canonical(assetIndexID())
+
+// workloadTxBuiltins are the transaction builtin indexes the generator churns:
+// the index-backed leaves of the transactions filter grammar (reference, the
+// three date fields, and the three address-role account→tx mappings).
+var workloadTxBuiltins = []commonpb.TransactionBuiltinIndex{
+	commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_REFERENCE,
+	commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_TIMESTAMP,
+	commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_INSERTED_AT,
+	commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_REVERTED_AT,
+	commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_ADDRESS,
+	commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_SOURCE_ADDRESS,
+	commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_DESTINATION_ADDRESS,
+}
+
+// addressRoleBuiltin maps an address role to the tx builtin index serving it —
+// the model's copy of the compiler's txAddressIndexID.
+func addressRoleBuiltin(role commonpb.AddressRole) commonpb.TransactionBuiltinIndex {
+	switch role {
+	case commonpb.AddressRole_ADDRESS_ROLE_SOURCE:
+		return commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_SOURCE_ADDRESS
+	case commonpb.AddressRole_ADDRESS_ROLE_DESTINATION:
+		return commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_DESTINATION_ADDRESS
+	default:
+		return commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_ADDRESS
+	}
+}
+
+// txBuiltinCanonical returns the canonical IndexID string of a tx builtin — the
+// model's index-map key the filter classifier and the lifecycle validation
+// share.
+func txBuiltinCanonical(field commonpb.TransactionBuiltinIndex) string {
+	return indexes.Canonical(indexes.TxBuiltinID(field))
+}
+
+// workloadIndex is one index the generator churns: its wire ID and canonical key.
+type workloadIndex struct {
+	id        *commonpb.IndexID
+	canonical string
+}
+
+// workloadIndexes is every index the workload creates and drops: the
+// account-by-asset builtin, the tx builtins, and the log-date builtin the LOGS
+// date filters are served from.
+func workloadIndexes() []workloadIndex {
+	out := []workloadIndex{
+		{assetIndexID(), assetIndexCanonical},
+		{logDateIndexID(), logDateIndexCanonical},
+	}
+	for _, field := range workloadTxBuiltins {
+		out = append(out, workloadIndex{indexes.TxBuiltinID(field), txBuiltinCanonical(field)})
+	}
+
+	return out
+}
+
+// indexStateLabel renders an index's model lifecycle state for finding details.
+// indexStateLabelFull renders one canonical's model state, retype window
+// included, for finding diagnostics.
+func indexStateLabelFull(ls oracle.LedgerState, canon string) string {
+	exists, active := ls.IndexState(canon)
+	label := indexStateLabel(exists, active)
+	if oldTypes, open := ls.RetypeWindow(canon); open {
+		parts := make([]string, len(oldTypes))
+		for i, t := range oldTypes {
+			parts[i] = fmt.Sprintf("%d", t)
+		}
+		label += fmt.Sprintf("+window(old=%s)", strings.Join(parts, ","))
+	}
+
+	return label
+}
+
+func indexStateLabel(exists, active bool) string {
+	switch {
+	case !exists:
+		return "absent"
+	case active:
+		return "active"
+	default:
+		return "ambiguous"
+	}
+}
+
+// --- has-asset evaluation ------------------------------------------------
+
+// workloadAsset is a (base, precision) pair matching one of the workload's asset
+// strings (config.go assets), so a has-asset filter built from it can match the
+// accounts those postings touched.
+type workloadAsset struct {
+	base      string
+	precision uint32
+}
+
+// workloadAssets mirrors config.go assets {"USD/2","EUR/2","COIN"} split into
+// (base, precision) via the server's asset-string convention (no "/N" suffix
+// means precision 0).
+var workloadAssets = []workloadAsset{{"USD", 2}, {"EUR", 2}, {"COIN", 0}}
+
+// accountHasAsset reports whether addr is in the account-by-asset index for
+// (assetBase, precision): whether it has EVER touched that asset via a committed,
+// non-excluded posting. This is a monotonic history the oracle tracks (see
+// LedgerState.everAsset / recordAssetTouches), NOT the current volume set — an
+// account drained to zero and purged from the volume table still matches.
+func accountHasAsset(ls oracle.LedgerState, addr, assetBase string, precision uint32) bool {
+	return ls.HasEverAsset(addr, assetBase, precision)
+}
+
+// assetWindow is the model's prediction of a bare has-asset ListAccounts page:
+// the accounts that ever touched (base, precision), in address order (reversed
+// when reverse), with the exclusive cursor applied and capped at pageSize —
+// mirroring accountWindow's pagination over the ever-touched set instead of the
+// current-volume universe.
+func assetWindow(ls oracle.LedgerState, base string, precision uint32, cursor string, pageSize int, reverse bool) []string {
+	window := ls.EverAssetAccounts(base, precision)
+
+	if reverse {
+		reverseStrings(window)
+	}
+
+	if cursor != "" {
+		kept := window[:0]
+		for _, addr := range window {
+			if reverse && addr >= cursor || !reverse && addr <= cursor {
+				continue
+			}
+
+			kept = append(kept, addr)
+		}
+		window = kept
+	}
+
+	if len(window) > pageSize {
+		window = window[:pageSize]
+	}
+
+	return window
+}
+
+// hasAssetTarget extracts the (base, precision) of a bare AccountHasAsset filter.
+// genAccountAssetFilter only produces bare has-asset leaves, so this is total for
+// the asset-index path.
+func hasAssetTarget(f *commonpb.QueryFilter) (base string, precision uint32, ok bool) {
+	if ha, isHA := f.GetFilter().(*commonpb.QueryFilter_AccountHasAsset); isHA {
+		return ha.AccountHasAsset.GetAssetBase(), ha.AccountHasAsset.GetPrecision(), true
+	}
+
+	return "", 0, false
+}
+
+// --- asset-filter generation ---------------------------------------------
+
+// genAccountAssetFilter rolls a bare AccountHasAsset leaf on a random workload
+// asset — the account-by-asset lifecycle path. It is deliberately NOT composed
+// with other leaves: the has-asset index returns accounts that may have been
+// purged from the volume table ("ever touched"), while an index-free address leaf
+// scans only current accounts, so a boolean of the two matches the server's
+// iterator intersection/union — set semantics the model would have to reproduce
+// the read-store to predict. A bare leaf's result set is exactly
+// EverAssetAccounts(base, precision), which the model tracks directly.
+func genAccountAssetFilter() *commonpb.QueryFilter {
+	a := workloadAssets[int(random.RandomChoice([]uint8{0, 1, 2}))]
+
+	return filterHasAsset(a.base, a.precision)
+}
+
+// --- asset-index query validation ----------------------------------------
+
+// validateAssetAccountQuery validates a ListAccounts query whose only index need
+// is the account-by-asset builtin. The legal outcomes depend on that index's
+// lifecycle in each candidate base:
+//
+//   - a result set is legal iff some base holds the index (ambiguous or active)
+//     and its ordered window equals the streamed accounts position-for-position;
+//   - a not-ready rejection (ErrIndexNotFound as FailedPrecondition + reason
+//     INDEX_NOT_FOUND, or ErrIndexBuilding as Unavailable + reason
+//     INDEX_BUILDING) is legal iff some base does not hold the index active
+//     (absent or ambiguous).
+//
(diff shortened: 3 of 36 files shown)


tests/antithesis/workload/bin/cmds/model/singleton_driver_model/indexes.go
```diff
@@ -0,0 +1,714 @@
...
+	for ledger, canons := range c.trackedIndexes() {
+		// readyAll starts true and is cleared for any canonical a replica reports
+		// missing, not-yet-live, or that we could not read at all.
+		readyAll := make(map[string]bool, len(canons))
+		for _, canon := range canons {
+			readyAll[canon] = true
+		}
+
+		for _, pc := range conns {
+			resp, err := pc.Bucket.GetIndexStatus(ctx, &servicepb.GetIndexStatusRequest{Ledger: ledger})
```
NumaryBot: 🟠 **[major] Stale index-status snapshot may cause false antithesis failures after drop/recreate**

When an index is dropped and later recreated with the same canonical ID, the readiness poller can promote the new model entry using a stale GetIndexStatus response that still reports the old ready generation. GetIndexStatus is not pinned to the model's committed sequence frontier. A subsequent query is sent with MinLogSequence at the committed frontier, so the server may correctly reject the query while rebuilding, but the model has already marked the index active — producing a false antithesis failure.

**Suggestion:** Pin the GetIndexStatus call to the current committed sequence (c.committedSeq) before promoting an index entry, or re-validate the returned generation against the model frontier after the status response is received. If the returned generation does not match the expected post-recreate generation, keep the entry in a pending/building state rather than promoting it to active.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

tests/oracle/model.go
```diff
@@ -478,8 +543,21 @@ type txRecord struct {
 	revertedBy         uint64
 	revertedAt         *commonpb.Timestamp
 	revertsTransaction uint64
+	// indexedAddrs is the transaction's account→tx index membership: per
+	// posting-side account, which role rows (AddrIndexedSource /
+	// AddrIndexedDestination bits) the index builder writes for this
+	// transaction. Stamped at end of bulk (recordIndexedAddrs) — a posting side
+	// whose cell lands in the exclusion projection gets no row. The any-role
+	// index is the union of the two bits.
+	indexedAddrs map[string]uint8
```
NumaryBot, line 645: 🟠 **[major] Hash indexed address membership for candidates**

Because `matchTxAddress` now uses `txRecord.indexedAddrs`, this field needs to participate in `LedgerState.Hash`. `candidateBases` deduplicates states by that hash; when different in-flight orderings converge to the same volumes/tx log/schema but only one ordering excluded a posting side from the account→tx index, the candidate with the valid address-filter result can be dropped and the model may report a correct `ListTransactions` page as outside the model.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries.go
```diff
@@ -0,0 +1,1525 @@
...
+	switch {
+	case rec.RevertedAt() != nil:
+		raOK = serverTx.GetRevertedAt() != nil && rec.RevertedAt().GetData() == serverTx.GetRevertedAt().GetData()
+	case rec.Reverted():
+		raOK = true
+	default:
+		raOK = serverTx.GetRevertedAt() == nil
+	}
+
+	return rec.Id() == serverTx.GetId() &&
```
NumaryBot, line 642: 🟠 **[major] Check learned inserted_at in transaction rows**

After `learnTxStamps` fills `rec.InsertedAt()` for drained committed transactions, row equality still ignores `serverTx.GetInsertedAt()`. For committed `ListTransactions` and single transaction reads this lets a response with an incorrect `inserted_at` pass validation, even though inserted_at filters are evaluated from the learned model value.

flemzord:
Reviewed live head `b064d0dc`. The production iterator changes preserve absolute re-seek semantics and are covered by focused regressions; the functional build, unit, E2E, scenario, model, and Schemathesis checks are green. The current NumaryBot findings concern Antithesis/model-oracle fidelity, while Dirty and Codecov are CI/tooling. They remain actionable but do not demonstrate a correctness/runtime regression under this cron policy. I found no correctness/runtime blocker on the current head. GitHub approval is intentionally disabled by the cron configuration.

gfyrag (commented):
Independent review at head 7bb4aeb336e64282ca74180e6ecc6fabed175d74. Four blocking findings remain; this draft is also conflicting and has no functional CI on the current head.

[P1][blocking] Rebuild index registry state during incremental restore
Location: internal/domain/replay/replay.go (ReplayLedgerLog); demonstrated by tests/e2e/cluster/restore_index_registry_test.go
Evidence: CreateIndex and DropIndex are discarded during RebuildDelta. The added red test proves the restored registry is empty and later metadata removal cannot report/drop the index.
Impact: disaster recovery loses a business-visible primary projection and diverges from the read-side lifecycle. PR #1769 addresses this separately but does not fix this head.

[P2][blocking] Compare learned inserted_at in transaction row validation
Location: singleton_driver_model/queries.go (txRecordMatches)
Evidence: the oracle learns InsertedAt and uses it for filters, but the matcher never compares it with the server row.
Impact: wrong inserted_at responses pass the advertised equality oracle.

[P2][blocking] Bind index readiness to the current incarnation/frontier
Location: singleton_driver_model/indexes.go (reconcileIndexes)
Evidence: a stale status sampled before DropIndex/CreateIndex reuse can promote the new model entry without proving LastIndexedSequence reaches its frontier or identifying the incarnation.
Impact: legal drop/recreate executions can produce false Antithesis findings.

[P2][blocking] Synchronize the HTTP/API contract
Location: internal/adapter/http/error_handler.go, API docs, api-comparison, OpenAPI 503 response
Evidence: all KindUnavailable responses now receive Retry-After, while docs still describe no-leader-only behavior and omit READ_INDEX_NOT_CAUGHT_UP and CHAPTER_ARCHIVE_OUT_OF_ORDER.
Resolution: document the broad retry contract and both public reasons/status/metadata.

Targeted root/nested model tests and tagged E2E compilation passed; the new restore scenario is knowingly red.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/indexes.go
```diff
@@ -0,0 +1,1165 @@
...
+		// Per-node observations for retype-window closure: the node's fold
+		// cursor and each index's in-flight rewrite target. nodeOK marks the
+		// nodes actually sampled this tick — an unreachable node makes no
+		// progress, in either phase.
+		nodeOK := make([]bool, len(conns))
+		nodeLastIndexed := make([]uint64, len(conns))
+		nodePending := make([]map[string]uint32, len(conns))
+
+		for i, pc := range conns {
+			resp, err := pc.Bucket.GetIndexStatus(ctx, &servicepb.GetIndexStatusRequest{Ledger: ledger})
```
NumaryBot: 🔴 **[blocker] Stale index-status snapshot may cause false failures after recreate**

If an index is dropped and recreated with the same canonical ID while these RPCs are in flight, a response describing the old ready incarnation can promote the new model entry because neither the request nor response is tied to `c.committedSeq` or an incarnation. Queries pinned to the newer committed frontier may then correctly receive `INDEX_BUILDING`, which the prematurely active oracle reports as an Antithesis failure. Pin or otherwise validate the status against the post-recreate generation before promotion.

internal/application/indexbuilder/index_config.go
```diff
@@ -408,6 +408,30 @@ func (b *Builder) getOrCreateLedgerConfig(ledger string) *ledgerIndexConfig {
...
+// metadata indexes carry one; for every other kind — and for a key declared
+// after the index — the version is bound to no type and rows keep each
+// value's natural encoding.
+func (b *Builder) boundTypeAtCreation(ledgerName string, id *commonpb.IndexID) (commonpb.MetadataType, bool) {
+	meta, ok := id.GetKind().(*commonpb.IndexID_Metadata)
+	if !ok || meta.Metadata == nil || b.batchSchema == nil {
+		return 0, false
+	}
+
+	schema, err := b.batchSchema.For(ledgerName)
```
NumaryBot: 🔴 **[blocker] Bind creation versions to the schema at the log sequence**

When the indexer lags across both a metadata-index creation and a later retype, `batchSchema.For` reads the latest FSM snapshot rather than the schema in force at the `CreatedIndexLog`. This binds the first version to the later type, contrary to this helper's contract; the creation backfill can therefore encode historical rows using the wrong type and expose incomplete results before the subsequent rewrite promotes. Derive the binding from sequence-correct log state rather than the batch's final schema.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

gfyrag (changes requested):
Independent review at exact head `35c788e27c0fa9f4c3c49ff4887b4a6d65b15fed`.

[P2][blocking] Make per-node readiness reads locally attributable
Location: `tests/antithesis/workload/bin/cmds/model/singleton_driver_model/indexes.go:605-607`
Evidence: `GetIndexStatus` uses the normal linearizable context, so a follower request may be forwarded to the leader. `internal.WithStaleConsistency` already exists for locally served per-node observations.
Impact: a rebuilding follower can be credited with the leader's ready state, masking cross-replica divergence.
Resolution: call each status RPC with `internal.WithStaleConsistency(ctx)`.

[P2][blocking] Bind readiness results to the sampled index incarnation
Location: `singleton_driver_model/indexes.go:589-641`
Evidence: after polling without the checker lock, the result is applied by checking only that the same canonical ID still exists. Drop+recreate can reuse that canonical for a new building incarnation.
Impact: stale readiness can promote the wrong incarnation and manufacture or mask model findings.
Resolution: snapshot a lifecycle generation/create frontier and require it unchanged before applying the poll.

[P2][blocking] Compare learned `inserted_at` in transaction equality
Location: `singleton_driver_model/queries.go:601-643`
Evidence: `InsertedAt()` is learned and used by filters but omitted from `txRecordMatches`.
Impact: an incorrect server `inserted_at` value passes the advertised equality oracle.
Resolution: compare it once learned and add a mismatched-value regression.

The nested model unit package passes, but the current distributed `Tests-Model` job is red (`ListLogs` returns an internal error). The PR body also depends on #1817, which is now closed without merge; replace that dependency explicitly before merge.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/indexes.go
Azorlogh: Fixed in 1b651b975: each poll snapshots the index's create frontier (the committed sequence of its latest CreateIndex, recorded at bulk fold), a node's report counts only once its LastIndexedSequence passed that frontier, and the apply step discards the verdict when the frontier moved mid-poll. Regression TestReconcileIndexes_IncarnationGuard covers all three legs, verified red with either guard weakened.

Azorlogh:
Addressed the three blocking findings in 1b651b975 (rebased onto release/v3.0 first — all the server fixes this branch validated are now merged, body's scope note updated):

1. GetIndexStatus polls now run with WithStaleConsistency — the per-node question gets a locally-attributable answer instead of a syncing follower forwarding to the leader.
2. Readiness is bound to the sampled incarnation via a create frontier (committed CreateIndex sequence, recorded at fold): reports from indexers that haven't folded past the create don't count, and a verdict whose frontier moved mid-poll is discarded. Regression drives promote / pre-create-cursor / mid-poll-recreate.
3. txRecordMatches compares the learned inserted_at (nil stays server-dated); serverTxFromRec carries it; mismatch regression added.

The Tests-Model rerun on the rebased head is the check for the ListLogs internal error you saw — that class matches the now-merged #1809.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries.go
```diff
@@ -0,0 +1,1533 @@
...
+	if err == nil {
+		accounts, err = drainStream(stream)
+	}
+
+	// High-water at the read's completion: only bulks dispatched by now could be
+	// reflected in the page. Captured before validation so later dispatches
+	// aren't folded into this read's candidate states.
+	maxTicket := c.ticketSeq.Load()
+
+	if err != nil {
```
NumaryBot, line 105: 🟠 **[major] Validate index-building errors before skipping transients**

When an indexed query returns `INDEX_BUILDING`, gRPC maps it to `Unavailable`, so `internal.IsTransient(err)` returns true and both account and transaction paths exit before their indexed validators run. Consequently, even after the poller marks an index active, an incorrect `INDEX_BUILDING` response is silently accepted, defeating the lifecycle coverage this PR claims. Route index-related errors through the indexed validators before the generic transient check. This is required by the test-review contract to ensure the failure signal proves the tested invariant ([docs/technical/contributing/ai-review.md:85-89](../../../../../../../docs/technical/contributing/ai-review.md#L85-L89)).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries.go
Azorlogh: Fixed in c0f78da43: the account/tx transient bails peel off the INDEX_BUILDING reason (isIndexNotReady) before skipping, classifyIndexedQueryError buckets the Unavailable+INDEX_BUILDING shape as not-ready, and the asset validator's code guard accepts it. TestIndexBuildingErrorReachesValidators pins the trap (the reason sits inside IsTransient) and the classification; a plain Unavailable stays a skippable transient.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/indexes.go
```diff
@@ -0,0 +1,1200 @@
...
+	indexedErrCompilation
+)
+
+// classifyIndexedQueryError buckets err for indexedQueryOutcomeLegal; ok=false
+// means the code is not part of the indexed-query surface at all (a finding).
+func classifyIndexedQueryError(err error) (indexedErrKind, bool) {
+	switch {
+	case err == nil:
+		return indexedErrNone, true
+	case status.Code(err) == codes.FailedPrecondition:
```
NumaryBot: 🟠 **[major] Match index-not-found by reason**

Any `FailedPrecondition` from an indexed list request is currently accepted as an index-not-ready outcome, even when it carries an unrelated reason or no reason. This can hide server regressions by allowing the model to explain an unexpected precondition error as a missing index; require `INDEX_NOT_FOUND` here, as is already done for `INDEX_BUILDING`. This weakens the test failure signal described in [docs/technical/contributing/ai-review.md:85-89](../../../../../../../docs/technical/contributing/ai-review.md#L85-L89).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/indexes.go
Azorlogh: Fixed in bf5cf226e — and the same weakness existed at two more sites, all three now require the INDEX_NOT_FOUND reason: classifyIndexedQueryError, validateAssetAccountQuery's code guard, and the ListLogs date-leaf refusal. TestIndexNotFoundMatchedByReason pins that a reasonless or unrelated-reason FailedPrecondition stays a finding. (Safety checked: describableToGRPCStatus attaches Reason as ErrorInfo — the same plumbing the INDEX_BUILDING match already relies on.)

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

tests/antithesis/k8s/workload.yaml
```diff
@@ -99,12 +99,17 @@ spec:
...
+        # Per-bulk driver decisions (dispatch, outcome, model verdict) in the
+        # container log, so a divergence finding carries the driver's view of
+        # every observation in the moment logs.
+        - name: MODEL_DEBUG
+          value: "1"
         # Dumps every committed ApplyRequest (base64) for exact deterministic
         # replay through the oracle (see tests/oracle/cmd/replay). Empty
         # disables; substituted by generate-manifests.sh from the
         # MODEL_DUMP_BATCHES env at config-image build.
         - name: MODEL_DUMP_BATCHES
```
NumaryBot, line 111: 🟠 **[major] Preserve the batch-dump configuration placeholder**

Hard-coding this value to `1` makes every Kubernetes model run dump every protobuf request as base64, while `generate-manifests.sh` still substitutes only `__MODEL_DUMP_BATCHES__`. Thus the documented opt-in flag can no longer disable dumping, increasing log volume substantially during long Antithesis runs; retain the placeholder here.

Azorlogh: Fixed in ab7aefaa6 — the placeholder is restored; the hard-coded 1 was hunt-run residue. MODEL_DEBUG=1 above it stays: it is deliberate and documented in place (divergence findings need the driver's per-bulk view in the moment logs).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

gfyrag (changes requested):
[P2][blocking] Require the documented gRPC code for INDEX_BUILDING
Location: tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries.go:658-665 (isIndexNotReady)
Evidence: isIndexNotReady matches only the ErrorInfo reason. HasErrorReason is independent of the status code, so an InvalidArgument, FailedPrecondition, or any other gRPC status carrying INDEX_BUILDING is classified as the expected not-ready response. The adjacent isIndexNotFound helper correctly requires both FailedPrecondition and INDEX_NOT_FOUND, while the comment here specifies Unavailable plus INDEX_BUILDING.
Impact: the model oracle accepts an API error-code regression instead of emitting an Antithesis finding, weakening the exact failure signal this PR adds.
Resolution: require status.Code(err) == codes.Unavailable together with the INDEX_BUILDING reason, and add a wrong-code/same-reason regression.

The three blockers from the immediately preceding review are fixed: status polling is locally served, readiness is bound to the sampled creation frontier, and learned inserted_at is compared. The later validator-routing and INDEX_NOT_FOUND fixes are also sound. This is one genuinely new finding.

DECISION: REQUEST CHANGES
Head reviewed: ab7aefaa6c3e06d0e110a2dd1e43565fe766820e
Blocking findings: 1
Non-blocking findings: 0
Previous findings: 3 fixed / 0 still valid / 0 outdated
Validation considered: nested singleton model unit tests + current CI
Residual risk: MEDIUM

Azorlogh:
Fixed in e5c98f16f: isIndexNotReady now requires codes.Unavailable alongside the INDEX_BUILDING reason, matching isIndexNotFound's shape; TestIndexNotReadyRequiresDocumentedCode pins that the reason on a wrong code stays a finding.

gfyrag:
Review checkpoint for tomorrow: please rebase this PR onto the current `release/v3.0` (`d095a8b085553f6045573d0317003b605d420193`). The previous `INDEX_BUILDING` blocker is resolved at the current head: the matcher now requires `codes.Unavailable` and the wrong-code regression is present. CI is green and there is no known active technical finding, but the stale-base gate prevents a valid final approval until the rebase and review-loop rerun.

Azorlogh:
Rebased onto `release/v3.0` @ d095a8b08 (no overlap with EN-1945 — clean rebase, agent-check PASS). CI rerunning.

Azorlogh:
Addressed the Shipfox review at 0563c4779:

1. **Duplicate assertion names** — split per target: `kind-mismatched field query rejected on transactions` / `... on accounts`.
2. **Swallowed recheck errors** — `recheckLogIDs`/`recheckLogKinds` now return their RPC/stream errors and the finding details render `recheck failed: <err>` instead of a silently empty list, so a failed diagnostic can't corrupt triage of a real finding.
3. **Stale testing.md** — the model-checker section now covers the filtered/paginated list surface and the index lifecycle, the file-role table lists `queries.go`/`queries_logs.go`/`indexes.go`/`metadata_filters.go`, and the read-surface guidance points at the right pattern per read shape.
4. **[Spec] stale scope-cut paragraph** — replaced: indexed-key retypes ARE generated and their serving windows modeled (EN-1724's per-version bindings made the window well-defined); the description now says so.
5. **[Spec] undeclared scope** — the description now carries a "Harness / infrastructure scope" section declaring the launch-interface auth + image-freshness gate, SOURCE_DATE_EPOCH build change, invariants-tagged local model builds + MODEL_DUMP_BATCHES, and the fleet-wide WRITES_BLOCKED_DISK_FULL transient reclassification, each with rationale.

tests/antithesis/workload/internal/client.go
```diff
@@ -355,6 +355,16 @@ func IsReadIndexNotCaughtUp(err error) bool {
 	return HasErrorReason(err, "READ_INDEX_NOT_CAUGHT_UP")
 }
 
+// IsWritesBlockedDiskFull returns true for the write gate's disk-pressure
+// rejection (ResourceExhausted / WRITES_BLOCKED_DISK_FULL). The gate refuses
+// the write before consensus, so the bulk definitively did not commit and a
+// retry is sound; under fault injection disk pressure comes and goes, so the
+// workload treats it as transient. The clock-skew twin is Unavailable and
+// already lands in the transient set through IsUnavailable.
+func IsWritesBlockedDiskFull(err error) bool {
```
NumaryBot, line 367: 🟠 **[major] Require ResourceExhausted for disk-full retries**

If a server regression returns `WRITES_BLOCKED_DISK_FULL` with any code other than `ResourceExhausted`, this matcher still classifies it as transient and suppresses the unexpected response. The documented pre-consensus retry guarantee applies only to the `ResourceExhausted`/reason tuple, so check `status.Code(err)` as well as the reason.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

tests/antithesis/workload/internal/client.go
Azorlogh: Fixed in 670e87c07 — `IsWritesBlockedDiskFull` now requires `codes.ResourceExhausted` alongside the reason, matching the documented pre-consensus tuple; the reason under any other code surfaces as unexpected.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

Azorlogh:
Re shipfox's [review](https://github.com/formancehq/ledger/pull/1659#issuecomment-5583335969): items 1, 3 and 4 taken in 614ec6e99f — one `foldFilter` traversal owns And/Or/Not for all nine filter walks (each supplies only its combinator step and leaf judgement; per-walk semantics preserved exactly, including the pre-existing empty-And difference between accounts and transactions/logs), the three bound predicates collapse onto a generic `withinBounds` (`matchUintBounds` / `matchIntBounds`), and the shadowing `intBound`/`uintBound` closures become one `renderBound`.

Item 2 (a `queryValidator` sub-type) is declined: every new `Checker` method takes `c.mu` and reads `c.modelState` / `c.indexCreateSeq` / `c.retypeObs`, so a composed sub-type would need a back-pointer to the `Checker` it validates for — a pointer cycle that only relocates the methods. The package's existing per-concern split of `Checker` methods (validate/search/debug/replay/processor/restore) is the pattern this follows.

Azorlogh:
Re shipfox's [review of 614ec6e99](https://github.com/formancehq/ledger/pull/1659#issuecomment-5584721216): taken in d6c1d2d549197ec7ce0b457f9e8abe654cde95a1 — the README's error-classification section now lists the `ResourceExhausted / WRITES_BLOCKED_DISK_FULL` refusal in the `IsTransient` set, and the `IsTransient` doc comment in `internal/client.go`, which had the same stale list, is updated with it.

Azorlogh:
Re shipfox's [review of d6c1d2d54](https://github.com/formancehq/ledger/pull/1659#issuecomment-5585151313): taken in 584587e8bdd66544b4d0111423763196601e74eb — `handleInvalidTargetError` now emits one `Reachable` literal per target ("target-invalid query rejected on accounts" / "… on transactions"), matching the existing `kind-mismatched field query rejected on …` pair, so the two coverage signals are catalogued separately.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries_logs.go
```diff
@@ -0,0 +1,400 @@
...
+
+	if err != nil {
+		if internal.IsTransient(err) || isShutdownError(err) {
+			return
+		}
+
+		// The log date is served by an opt-in builtin index this driver never
+		// creates, so a filter reading it is expected to be refused. That
+		// refusal is the coverage: it exercises the index gate on LOGS.
+		if hasDateLeaf(filter) && isIndexNotFound(err) {
```
NumaryBot: 🟠 **[major] Require the absent log-date index rejection**

When `filter` has a date leaf, this driver never creates the required log-date index, so `INDEX_NOT_FOUND` is the only valid non-environmental outcome. Currently this branch merely accepts that error: an accurate result page continues into `validateLogQuery`, and an incorrect `Unavailable/INDEX_BUILDING` response is skipped as transient above, allowing a broken index gate to pass the soak.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

Azorlogh:
Standards 1 confirmed and fixed in 45b5e295d: `describeCandidateVerdicts`, `foldDiag` and `assetContentDiags` all documented or assumed `c.mu` held, and all five call sites run after the details path unlocks, so `candidateBases`' iteration of `c.inflight` could race a worker's `c.inflight[t] = bulk` — a fatal throw, precisely while rendering a real divergence. All three now take `c.mu` themselves, matching `matchesModel`, and their contracts say so. The driver suite passes under `-race`.

Standards 2 (duplicated window logic) and 3 (stale one-in-six probability comments), plus the Spec note on the recipe count, are fair and left for a follow-up: the shared pagination block and one generic required/optional matcher are a refactor across accounts, transactions and logs, and I would rather not fold that into this PR now that its coverage has surfaced a server bug. Filed as follow-up work.

For the record, this PR's log-date coverage is red against a real defect it found, not against the driver: creating the log-date index on a ledger with history leaves every pre-index schema log out of that index for good, because the backfill's `isDataLog` gate skips them while the live fold indexes every log. A date-range `ListLogs` then omits them and the negated filter returns exactly them. Tracked as EN-1987 with a deterministic e2e repro; that fix lands before this PR.

gfyrag:
Rebased the existing branch onto `release/v3.0` at `37ccf5238ed460e0df2b9844869bdb9c7031f532` and reconciled all 21 issue comments, 20 reviews, and 10 inline threads.

**Tests-Model root cause and fix.** The failing `NOT(ledger=current AND logDate>3)` page contained configuration/schema logs whose dates were omitted by `processBackfill`'s data-only gate. The same defect reproduced locally on the rebased head. The correction backfills every ledger log's identity/date without replaying configuration mutations; initial log-date indexes also backfill their existing configuration history and own creation log. Regression tests failed before the fix for the missing schema-log date and premature initial readiness, then passed with drop/recreate, foreign-ledger isolation, and subsequent live-log parity. The model oracle remains strict.

**Review feedback.**
- Historical blocking findings remain fixed: per-replica `WithStaleConsistency`, sampled creation-frontier guards, learned `inserted_at` equality, and exact `Unavailable` + `INDEX_BUILDING` classification. The index-registry restore and version-bound schema fixes are in the base. The removed read-frontier wire field is not reintroduced.
- Narrowed the retype-window exception to the rejected index named in `ErrorInfo.metadata.index`. Both entity targets reject errors attributed to an active sibling, an unrelated index, or missing attribution. Restoring the old wildcard exception makes the new regression fail on both targets.
- Undeclared-field CreateIndex probes now exercise both account and transaction targets. Restoring the account-only generator makes the coverage regression fail.
- Corrected the probability comments and reused the shared LOGS boolean-filter constructors. Updated the PR body with the five build recipes, remaining harness details, current base, and the server fix.
- The optional larger validator/pagination refactors are not included: the shared lifecycle verdict already applies to both targets, their regression coverage is retained, and no differing behavior was demonstrated. The existing per-concern Checker layout was previously justified in the discussion. Prior diagnostic locking and assertion-name fixes are preserved.

All ten inline threads were already resolved at takeover; none was blindly resolved or dismissed. This comment addresses the remaining top-level requests. The PR remains open for reviewer sign-off; no merge was performed.

**Commits:** `af44c449a` (server fix and regressions), `350b4b188` (review corrections).

- `bash scripts/agent-check`: PASS.
- `AI_REVIEW_BASE_SHA=37ccf5238ed460e0df2b9844869bdb9c7031f532 bash scripts/agent-check-pr`: PASS (generation/tidy/lint, invariants, root build, focused indexbuilder/oracle race suites).
- Pinned Nix workload module `go test -race ./...`: PASS.
- Pinned Nix `just test-model-cluster 180`: PASS, 3 nodes, 4 completed restart/recovery cycles, no findings.
- Deterministic server regressions: failed before the fix, passed after it. Mutation checks prove the retype attribution and both-target generator regressions fail when their respective fixes are removed.
- Final diff/whitespace review: PASS; no generated drift; working tree clean.


CI on `350b4b188`: [Default run 34372646436](https://github.com/formancehq/ledger/actions/runs/34372646436) completed successfully, including `Tests-Model`, both E2E suites, `Tests`, `Dirty`, workload tests, build, Schemathesis and Coverage. Repository invariants and NumaryBot review are also green. Release/image jobs were not applicable and were skipped by the existing workflow conditions. GitHub still reports `CHANGES_REQUESTED`; reviewer sign-off is the remaining merge gate.

Azorlogh:
Reverted the server fix in 5213f1a5c: it lives in #1960 (EN-1987), which also carries the e2e coverage. This PR is test-only again, and its `Tests-Model` job stays red until #1960 lands.

Your indexbuilder unit regression went with the revert — worth porting into #1960 if you want it alongside the e2e.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries_logs.go
```diff
@@ -0,0 +1,575 @@
...
+// ListLogs page draws from. The endpoint has no reverse mode
+// (ValidateListOptions rejects it) and paginates forward only, so rows are
+// ascending by id with the cursor as an exclusive lower bound — exactly how
+// the controller translates afterSequence into a LogId condition. Truncation is
+// logWindowMatches' job: optional rows may or may not consume page slots, so a
+// fixed prefix cut would be wrong.
+func logWindowRows(ls oracle.LedgerState, ledger string, filter *commonpb.QueryFilter, afterSeq uint64) []logWindowRow {
+	var rows []logWindowRow
+
+	for _, row := range ls.LogDates() {
```
NumaryBot: 🔴 **[blocker] Restore complete log-date backfilling**

When a log-date index is created after configuration/schema logs exist, this model requires those rows in filtered windows, but HEAD reverted the server fix and `processBackfill` still skips every `!isDataLog` entry before indexing its date. Consequently filters such as `NOT(logDate > n)` can deterministically disagree with `logWindowRows`, causing Tests-Model failures; restore the backfill fix or do not enable this coverage until the server satisfies it.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries_logs.go
Azorlogh: Intended: the fix is in #1960 (EN-1987), so this PR stays test-only and its `Tests-Model` job is red until that merges. The coverage is correct as written — it is what found the defect.

gfyrag:
Review challenge on current HEAD `5213f1a5`: I do not think the current evidence proves the query coverage that this PR claims.

1. The new paths are selected probabilistically (`main.go`: a 1/5 read choice, then one of eight reads, followed by further random filter/index choices), and the success signals are `assert.Reachable`. The repository's own Antithesis guidance says a `Reachable` without an upstream `Sometimes` is passive. The local model runner only fails assertions that were emitted false; it does not require the query/index/retype signals to have been hit. A green run can therefore exercise none of the critical combinations. Please add either a deterministic phase (`create → ready → query → retype → drop → recreate`) or explicit coverage probes/counters that the runner requires before reporting PASS.

2. The result oracle is not exact enough. `ListLogs` reduces each returned log to its local id, so correct ids with a wrong date, ledger, payload, global sequence or signature pass. `accountMatches` silently ignores any colored volume returned by the server and does not compare `first_usage`, `insertion_date` or `updated_at`. `txRecordMatches` does not compare `updated_at` or `post_commit_volumes`, and `postingsEqual` omits `Posting.color`. These are returned fields, so a server payload corruption can be accepted as model-conformant.

I would treat both as blockers for calling this a strict query oracle. The stronger shape is canonical full-row equality with only explicitly documented normalization, plus negative mutant tests proving that each omitted field is detected.

Azorlogh:
Both addressed. `fbba258ad`, `ce4939954`, `4a82870bf`.

**1. Coverage.** Took the probes option: twelve must-hit `Sometimes` sondes — one per churned index, one per entity target for metadata-field indexes, one for the retype window — satisfied only by a page the index was needed for *and* the oracle verified. The runner fails on any registered sonde never satisfied, reading the required set off the registrations so it cannot drift from the catalogue.

I dropped the `Reachable`s rather than pairing them: both are `mustHit`, so on the same predicate the `Sometimes` alone enforces it and gives guidance a gradient. That also meant teaching the runner real class semantics — a false `Sometimes` is not a finding, a never-satisfied one is; `STRICT_SOMETIMES` keeps the shared `IsTolerated` probes universal locally, pinned to their call sites by `check-repo-invariants`. `check_fail_fast` shares the same filter, or a sonde false by design would abort every run.

Measured before hardening: 300s, 3 nodes, 7 leader-kill cycles → 12/12.

**2. Exactness.** Logs now compare kind, canonical payload, date, global sequence and the three volume annotations. `post_commit_volumes` is compared on reads in both directions — the old `pcvMatches` was model⊆server, so a fabricated cell passed. `updated_at` shares `inserted_at`'s proposal date and is never bumped afterwards, so it is compared as that equality. `first_usage`, `insertion_date` and `Account.updated_at` are pinned absent rather than predicted: `assembleAccount` stamps none of them, so there is no value to hold the server to until it starts sending one. Mutation test per item.

Colour is deliberately still open — `Posting.color`, coloured account volumes and coloured pcv/annotation entries are all skipped. `oracle.VolumeKey` has no colour dimension, so that is a model change rather than a patch.

Two notes. The branch still carries the revert of the EN-1987 log-date fix (#1960, open), so `just test-model-cluster` fails fast on that real finding; restore it locally to exercise anything else. And `ai-pr-loop` returns `BASE_UPDATE_REQUIRED` at `404ac87a4` — a rebase is still needed for final approval, and #1974 (duplicate index creations) landing in that window looks like it may need a matching model prediction.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/coverage.go
```diff
@@ -0,0 +1,165 @@
...
+// index can answer. needed is the filter's index requirement, so a sonde can
+// only be satisfied by a query that genuinely depended on that index.
+//
+// Every sonde is evaluated on every call, not only the satisfied one — the
+// false evaluations are what let Antithesis bias toward a starved index. The
+// caller must not hold c.mu.
+func (c *Checker) noteQueryCoverage(ledger string, target commonpb.QueryTarget, filter *commonpb.QueryFilter, needed map[string]struct{}, served bool) {
+	details := internal.Details{"ledger": ledger}
+
+	for _, wi := range coverageIndexes {
```
NumaryBot, line 130: 🟠 **[major] Require a nonempty verified result before satisfying coverage**

The callers pass `matched && errKind == indexedErrNone` as `served`, even when the verified page contains zero rows. Consequently, an empty query can satisfy an index sonde, so a run may pass without ever verifying that the index returns an actual matching record; an implementation that incorrectly returns empty results can remain untested. Pass the result count as well and require at least one returned row before emitting a coverage hit.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/coverage.go
Azorlogh: Fixed in 603bcc781 — callers pass their row count and the verdict moved into `coverageHits`, so a sonde needs an accepted outcome, at least one row, and the filter to have needed that index. A 300s three-node run still satisfies all twelve, so the tightening costs no coverage.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries.go
```diff
@@ -0,0 +1,1550 @@
...
+
+	model := map[string]oracle.VolumePair{}
+	for k, vp := range ls.Volumes().All() {
+		if k.Address == addr {
+			model[k.Asset] = vp
+		}
+	}
+
+	server := map[string]struct{ in, out uint256.Int }{}
+	for _, av := range serverAcct.GetVolumes() {
```
NumaryBot, line 592: 🟠 **[major] Reject unexpected colored account volumes**

When the server returns a colored volume, this `continue` removes it before the exact comparison. Since the workload only creates uncolored postings, an erroneous extra colored bucket can therefore pass validation unchanged, preserving the oracle gap this field-comparison change intends to close.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries.go
```diff
@@ -0,0 +1,1550 @@
...
+}
+
+// pcvSnapshotMatches compares a served post-commit snapshot against the
+// model's cell for cell, in both directions: an absent cell and a fabricated
+// one are equally wrong. A coloured entry is a cell the model's volume key
+// cannot address, so the whole snapshot is left uncompared for that row.
+func pcvSnapshotMatches(model map[oracle.VolumeKey]oracle.VolumePair, server *commonpb.PostCommitVolumes) bool {
+	served := map[oracle.VolumeKey]struct{}{}
+
+	for account, byAssets := range server.GetVolumesByAccount() {
```
NumaryBot, line 704: 🟠 **[major] Reject colored post-commit volume entries**

If any transaction snapshot entry has a color, returning `true` accepts the entire snapshot without comparing any cells. Because this workload produces only uncolored postings, a server response with a fabricated colored entry—or with otherwise incorrect snapshot data alongside it—will be accepted instead of reported.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries_logs.go
```diff
@@ -0,0 +1,767 @@
...
+			j++
+		} else if row.required {
+			return false
+		}
+	}
+
+	return j == len(page)
+}
+
+// logRowMatches compares a served log against the model's record of it, field
```
NumaryBot, line 394: 🟠 **[major] Compare transaction payloads in log rows**

For created/reverted transaction logs, `row.payload` is deliberately empty, so this condition skips the payload comparison entirely. A server can return the correct log ID and kind but corrupted transaction details while `ListTransactions` remains correct, and the log page will still be accepted; validating a separate endpoint does not establish that this specific log payload is correct.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries.go
Azorlogh: Fixed in 817a88c8d — a coloured account volume now fails the match. No generated posting carries a colour, so such a bucket is data nothing in the run could have produced.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries.go
Azorlogh: Fixed in 817a88c8d — same rule for the post-commit snapshot: a coloured cell fails instead of disabling the comparison.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries_logs.go
Azorlogh: Fixed in 817a88c8d — each transaction log now names its transaction and the payload is compared against the record. Only the creation-frozen fields, since the record keeps moving afterwards (a revert sets reverted/reverted_by/reverted_at, a metadata save rewrites the map) while the log keeps what it was written with; comparing the live record produced a false finding on a 300s run before I narrowed it.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries.go
Azorlogh: stale — inserted_at is compared at queries.go:649 (`insOK`). Original: https://github.com/formancehq/ledger/pull/1659#discussion_r3691067965

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries.go
Azorlogh: stale — INDEX_BUILDING is peeled off before the transient bail at queries.go:107 and :219 (`!isIndexNotReady(err)`). Original: https://github.com/formancehq/ledger/pull/1659#discussion_r3906031087

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries.go
```diff
@@ -0,0 +1,1554 @@
...
+
+		var in, out uint256.Int
+		if err := in.SetFromDecimal(av.GetVolumes().GetInput()); err != nil {
+			return false
+		}
+		if err := out.SetFromDecimal(av.GetVolumes().GetOutput()); err != nil {
+			return false
+		}
+
+		server[av.GetAsset()] = struct{ in, out uint256.Int }{in, out}
```
NumaryBot, line 612: 🟠 **[major] Reject duplicate account volume rows**

When a response contains the same uncolored asset twice, assigning both entries into `server[av.GetAsset()]` collapses them, so duplicate rows can still pass if the final duplicate matches the model. This contradicts the claimed exact account comparison and can mask malformed `ListAccounts` output; reject an already-seen asset before inserting it.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries.go
```diff
@@ -0,0 +1,1554 @@
...
+func pcvSnapshotMatches(model map[oracle.VolumeKey]oracle.VolumePair, server *commonpb.PostCommitVolumes) bool {
+	served := map[oracle.VolumeKey]struct{}{}
+
+	for account, byAssets := range server.GetVolumesByAccount() {
+		for _, entry := range byAssets.GetVolumes() {
+			if entry.GetColor() != "" {
+				return false
+			}
+
+			served[oracle.VolumeKey{Address: account, Asset: entry.GetAsset()}] = struct{}{}
```
NumaryBot: 🟠 **[major] Reject duplicate post-commit volume cells**

When `VolumesByAccount` repeats an asset entry for one account, this set assignment collapses the duplicate, and the later length/value checks can accept the malformed snapshot. Since this helper is intended to compare cells in both directions, detect an existing `VolumeKey` before inserting it.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries.go
Azorlogh: Fixed in 9d8daba6b — a repeated asset now fails instead of collapsing into the map key. Narrower than stated though: a repeat alongside a missing cell was already caught by the length check, and a repeat carrying a different value was caught in `accountMatches` by the value compare. What slipped through was a bare repeat, and in the snapshot case the second copy of a cell (only the first is read).

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/queries.go
Azorlogh: Fixed in 9d8daba6b — a repeated asset now fails instead of collapsing into the map key. Narrower than stated though: a repeat alongside a missing cell was already caught by the length check, and a repeat carrying a different value was caught in `accountMatches` by the value compare. What slipped through was a bare repeat, and in the snapshot case the second copy of a cell (only the first is read).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1659#issuecomment-5143883984

gfyrag (approved):
Reviewed at exact head 9d8daba6b8a8b5a5cd1604a174897e2aaca081fd. No actionable correctness findings remain; CI and focused model/oracle validations are green.

PR #1660, opened 2026-07-31
fix(EN-1501): bound live query checkpoints, enforced softly at admission

Azorlogh:
## EN-1501: bound live query checkpoints

Query checkpoints (a `db.Checkpoint()` of the main store + a read-index snapshot) pin disk via hard-linked SSTs. Left unbounded, a scheduler or client loop grows disk and `ListQueryCheckpoints` payloads without end. This PR bounds the number of **live** checkpoints.

### Design: soft admission-time enforcement

Per @gfyrag's review, the limit is enforced at **admission**, not in the FSM:

- Before proposing a `CreateQueryCheckpoint`, the leader counts its live checkpoints (`query.ReadLiveQueryCheckpointIDs`, a proposer-side Pebble read) and rejects at the limit with `CHECKPOINT_LIMIT_REACHED` (gRPC `ResourceExhausted` / HTTP 429). `DeleteQueryCheckpoint` for a zero / non-live id is rejected with `CHECKPOINT_ID_REQUIRED` / `CHECKPOINT_NOT_FOUND`.
- The committed command carries no limit and the **FSM apply path is unconditional** — it applies identically on every node regardless of each node's configured limit, so determinism (invariant #2) holds. This is why the limit can be a plain startup flag rather than a Raft-committed setting.
- **Bounded overshoot, no reservation machinery**: concurrent in-flight creates can each pass the check before any commits, overshooting by up to the number of simultaneous creates. Accepted deliberately for a soft disk ceiling.
- **Rolling upgrade**: enforcement lives on the current leader. A pre-upgrade leader won't enforce; an upgraded one will. Since the check is admission-side and the committed command is unchanged, replicas never disagree on *applied* state.

### Configuration

Per-node startup config, not a runtime command:

- `--query-checkpoint-limit` flag (default 10)
- env `QUERY_CHECKPOINT_LIMIT`
- operator CRD field `queryCheckpointLimit` (`minimum: 1`)
- `Config.Validate()` rejects `0` at boot (would silently disable all creation)

### Kept

- **Checker**: `compareQueryCheckpoints` verifies the stored rows both ways against the `CreatedQueryCheckpoint`/`DeletedQueryCheckpoint` audit stream (id from the key, plus `max_sequence`/`created_at`/payload-id), baseline-seeded under archiving.
- **Audit rebuild** recreates the rows + the monotonic next-ID counter from the logs.
- **Recovery orphan sweep** reclaims row-less physical checkpoint directories (snapshot-install followers), re-sourced from Pebble.
- Scheduler quiet-skips once the cap is full.

### Removed vs. the earlier revision of this PR

The replicated `SetQueryCheckpointLimit` order/log/request, `GetQueryCheckpointLimit` RPC, `ledgerctl set-limit`/`get-limit`, the `SubGlobQueryCheckpointLimit` projection + its checker pass and `QUERY_CHECKPOINT_LIMIT_MISMATCH` error, and the FSM live-ID/limit state.

### Tests

Admission unit tests (limit boundary + delete existence incl. `id==0`), config-validation test for the zero rejection, checker both-ways coverage, rebuild replay, recovery orphan reclamation, and e2e cap behavior.

### DIFF
--- a/internal/application/admission/admission.go
+++ b/internal/application/admission/admission.go
@@ -69,6 +69,12 @@ type Admission struct {
 	authEnabled        bool
 	waitLeaderReady    func(context.Context) error
 
+	// queryCheckpointLimit is the operator-configured max live query checkpoints,
+	// enforced softly here (admission rejects a create at the limit) rather than
+	// in the FSM apply path — a committed CreateQueryCheckpoint always applies
+	// identically on every node regardless of this node's configured value.
+	queryCheckpointLimit uint64
+
 	// Metrics (noop when metricsEnabled is false)
 	metricsEnabled                 bool
 	commandDurationHistogram       metric.Int64Histogram
@@ -155,6 +161,17 @@ func WithAuthEnabled() func(*Admission) {
 	}
 }
 
+// DefaultQueryCheckpointLimit is the max live query checkpoints when the operator
+// configures none. Enforced softly at admission; not a Raft protocol invariant.
+const DefaultQueryCheckpointLimit uint64 = 10
+
+// WithQueryCheckpointLimit sets the operator-configured max live query checkpoints.
+func WithQueryCheckpointLimit(limit uint64) func(*Admission) {
+	return func(a *Admission) {
+		a.queryCheckpointLimit = limit
+	}
+}
+
 func NewAdmission(
 	store *dal.Store,
 	logger logging.Logger,
@@ -170,16 +187,17 @@ func NewAdmission(
 	opts ...func(*Admission),
 ) *Admission {
 	a := &Admission{
-		store:           store,
-		logger:          logger,
-		proposer:        proposer,
-		builder:         builder,
-		writeGate:       writeGate,
-		keyStore:        keyStore,
-		sharedState:     sharedState,
-		attrs:           attrs,
-		numscriptCache:  numscriptCache,
-		waitLeaderReady: waitLeaderReady,
+		store:                store,
+		logger:               logger,
+		proposer:             proposer,
+		builder:              builder,
+		writeGate:            writeGate,
+		keyStore:             keyStore,
+		sharedState:          sharedState,
+		attrs:                attrs,
+		numscriptCache:       numscriptCache,
+		waitLeaderReady:      waitLeaderReady,
+		queryCheckpointLimit: DefaultQueryCheckpointLimit,
 	}
 	for _, opt := range opts {
 		opt(a)
@@ -1981,6 +1999,72 @@ func (a *Admission) resolveScriptsAndEnrichNeeds(ctx context.Context, orders []*
 	return nil
 }
 
+// checkQueryCheckpointLimit rejects a create once the live checkpoint count has
+// reached the configured limit. Soft enforcement: it reads the leader's committed
+// count from Pebble and folds in same-bulk creates/deletes via the overlay, so a
+// batch that deletes a checkpoint then creates one is admitted atomically.
+// Concurrent creates across separate proposals may still overshoot until a
+// checkpoint is deleted. The FSM applies a committed create unconditionally.
+// Records the admitted create in the overlay so a later create in the same bulk
+// counts it.
+func (a *Admission) checkQueryCheckpointLimit(overlay *bulkOverlay) error {
+	handle, err := a.store.NewReadHandle()
+	if err != nil {
+		return fmt.Errorf("reading query checkpoints: %w", err)
+	}
+
+	defer func() { _ = handle.Close() }()
+
+	ids, err := query.ReadLiveQueryCheckpointIDs(handle)
+	if err != nil {
+		return fmt.Errorf("counting query checkpoints: %w", err)
+	}
+
+	live := len(ids) + overlay.checkpointNetLive()
+	if live >= int(a.queryCheckpointLimit) {
+		return &domain.ErrCheckpointLimitReached{Limit: a.queryCheckpointLimit}
+	}
+
+	overlay.recordCheckpointCreate()
+
+	return nil
+}
+
+// checkQueryCheckpointExists rejects a delete for a checkpoint id that is zero,
+// has no committed row, or was already deleted earlier in this bulk. Soft: it
+// reads committed state, so a create/delete in flight across proposals can race.
+// Records the delete in the overlay so it frees a limit slot for a same-bulk
+// create and a repeated delete in the same bulk is rejected.
+func (a *Admission) checkQueryCheckpointExists(id uint64, overlay *bulkOverlay) error {
+	if id == 0 {
+		return domain.ErrCheckpointIDRequired
+	}
+
+	if overlay.isCheckpointDeletedInBulk(id) {
+		return &domain.ErrCheckpointNotFound{CheckpointID: id}
+	}
+
+	handle, err := a.store.NewReadHandle()
+	if err != nil {
+		return fmt.Errorf("reading query checkpoint: %w", err)
+	}
+
+	defer func() { _ = handle.Close() }()
+
+	cp, err := query.ReadQueryCheckpoint(handle, id)
+	if err != nil {
+		return fmt.Errorf("reading query checkpoint %d: %w", id, err)
+	}
+
+	if cp == nil {
+		return &domain.ErrCheckpointNotFound{CheckpointID: id}
+	}
+
+	overlay.recordCheckpointDelete(id)
+
+	return nil
+}
+
 // requestToOrder converts a single Request into its ledger- or system-scoped
 // raftcmdpb.Order. batchSig is consulted only by the signing-key registration
 // path, to record the signing key as the new key's parent.
@@ -2246,12 +2330,20 @@ func (a *Admission) requestToOrder(ctx context.Context, req *servicepb.Request, 
 			reqType.SaveNumscript.GetContent(),
 		)
 	case *servicepb.Request_CreateQueryCheckpoint:
+		if err := a.checkQueryCheckpointLimit(overlay); err != nil {
+			return nil, err
+		}
+
 		wrapSystemScoped(order, &raftcmdpb.SystemScopedOrder{
 			Payload: &raftcmdpb.SystemScopedOrder_CreateQueryCheckpoint{
 				CreateQueryCheckpoint: &raftcmdpb.CreateQueryCheckpointOrder{},
 			},
 		})
 	case *servicepb.Request_DeleteQueryCheckpoint:
+		if err := a.checkQueryCheckpointExists(reqType.DeleteQueryCheckpoint.GetCheckpointId(), overlay); err != nil {
+			return nil, err
+		}
+
 		wrapSystemScoped(order, &raftcmdpb.SystemScopedOrder{
 			Payload: &raftcmdpb.SystemScopedOrder_DeleteQueryCheckpoint{
 				DeleteQueryCheckpoint: &raftcmdpb.DeleteQueryCheckpointOrder{
--- a/internal/application/check/checker.go
+++ b/internal/application/check/checker.go
@@ -396,6 +396,15 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 		// SubAttrNumscriptVersion (latest pointer = greatest stored semver).
 		expectedNumscriptContent = make(map[domain.NumscriptEntryKey]*commonpb.NumscriptInfo)
 		expectedNumscriptLatest  = make(map[domain.NumscriptVersionKey]string)
+
+		// derivedLiveCheckpoints is the audit-derived set of live query-checkpoints
+		// (cluster-global), keyed by id with the CreatedQueryCheckpoint log as the
+		// value so max_sequence / created_at can be verified. A create sets the
+		// entry, a later delete removes it. Seeded from the baseline under archiving
+		// (the create logs are purged with the chapter), then advanced by the
+		// replayed create/delete logs. compareQueryCheckpoints compares it against
+		// the stored rows both ways, contents included.
+		derivedLiveCheckpoints = make(map[uint64]*commonpb.CreatedQueryCheckpointLog)
 	)
 
 	// excluded is built incrementally as SimulateEphemeralPurge decides to
@@ -479,6 +488,13 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 		// does not make the checker expect a lower latest than the store holds,
 		// and archived immutable content stays verified.
 		if err := c.foldBaselineNumscripts(baselineDB, expectedNumscriptContent, expectedNumscriptLatest); err != nil {
+			return err
+		}
+
+		// Seed the live query-checkpoint set from the baseline: the pre-archive
+		// CreatedQueryCheckpoint logs are purged with the archived chapter, so
+		// replay alone would under-derive the set and false-flag live rows.
+		if err := c.foldBaselineQueryCheckpoints(baselineDB, derivedLiveCheckpoints); err != nil {
 			return err
 		}
 
@@ -640,6 +656,14 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 					if cur, ok := expectedNumscriptLatest[vk]; !ok || numscriptVersionGreater(info.GetVersion(), cur) {
 						expectedNumscriptLatest[vk] = info.GetVersion()
 					}
+				}
+			case *commonpb.LogPayload_CreatedQueryCheckpoint:
+				if cp := payload.CreatedQueryCheckpoint; cp != nil {
+					derivedLiveCheckpoints[cp.GetCheckpointId()] = cp
+				}
+			case *commonpb.LogPayload_DeletedQueryCheckpoint:
+				if cp := payload.DeletedQueryCheckpoint; cp != nil {
+					delete(derivedLiveCheckpoints, cp.GetCheckpointId())
 				}
 			case *commonpb.LogPayload_Apply:
 				if payload.Apply != nil {
@@ -898,6 +922,10 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 	}
 
 	c.compareNumscripts(snap, expectedNumscriptContent, expectedNumscriptLatest, deletedInReplay, pendingCleanupLedgers, callback)
+
+	if err := c.compareQueryCheckpoints(snap, derivedLiveCheckpoints, callback); err != nil {
+		return err
+	}
 
 	return nil
 }
@@ -1297,6 +1325,95 @@ func (c *Checker) compareMirrorV2LogID(reader dal.PebbleReader, chainBound *chai
 				0, name, "", ""))
 		}
 	}
+}
+
+// compareQueryCheckpoints verifies the live query-checkpoint rows stored in the
+// primary store (ZoneGlobal/SubGlobQueryCheckpoint) against `derived`, the live
+// set re-derived from the CreatedQueryCheckpoint / DeletedQueryCheckpoint logs
+// (baseline-seeded under archiving). It emits
+// CHECK_STORE_ERROR_TYPE_QUERY_CHECKPOINT_MISMATCH on:
+//   - a stored row absent from the audit set (never created, or later deleted) —
+//     a phantom or stale row;
+//   - an audit-live checkpoint with no stored row — a lost or dropped projection;
+//   - a row whose key id, max_sequence, or created_at diverges from the
+//     audit-derived value — content corruption.
+//
+// The audit-rebuild path recreates the rows (and their max_sequence / created_at)
+// from the logs (internal/infra/backup), so a missing row is corruption, never a
+// legitimate restore artifact. Stored rows are keyed by the Pebble key id, not
+// the payload (see query.ReadQueryCheckpointRows).
+func (c *Checker) compareQueryCheckpoints(reader dal.PebbleReader, derived map[uint64]*commonpb.CreatedQueryCheckpointLog, callback func(*servicepb.CheckStoreEvent)) error {
+	stored, err := query.ReadQueryCheckpointRows(reader)
+	if err != nil {
+		return fmt.Errorf("reading stored query checkpoints: %w", err)
+	}
+
+	all := make(map[uint64]struct{}, len(stored)+len(derived))
+	for id := range stored {
+		all[id] = struct{}{}
+	}
+
+	for id := range derived {
+		all[id] = struct{}{}
+	}
+
+	ids := make([]uint64, 0, len(all))
+	for id := range all {
+		ids = append(ids, id)
+	}
+
+	slices.Sort(ids)
+
+	emit := func(msg string) {
+		callback(errorEvent(servicepb.CheckStoreErrorType_CHECK_STORE_ERROR_TYPE_QUERY_CHECKPOINT_MISMATCH, msg, 0, "", "", ""))
+	}
+
+	for _, id := range ids {
+		row, inStored := stored[id]
+		want, inDerived := derived[id]
+
+		switch {
+		case inStored && !inDerived:
+			emit(fmt.Sprintf("stored query checkpoint %d is not justified by the audit chain (no CreatedQueryCheckpoint, or a later DeletedQueryCheckpoint)", id))
+		case inDerived && !inStored:
+			emit(fmt.Sprintf("audit chain has live query checkpoint %d but no stored row (lost or dropped projection)", id))
+		default:
+			if pid := row.GetCheckpointId(); pid != id {
+				emit(fmt.Sprintf("stored query checkpoint keyed %d carries a mismatched payload checkpoint_id %d (corrupted row)", id, pid))
+			}
+
+			if row.GetMaxSequence() != want.GetMaxSequence() {
+				emit(fmt.Sprintf("query checkpoint %d max_sequence %d does not match the audit chain %d", id, row.GetMaxSequence(), want.GetMaxSequence()))
+			}
+
+			if row.GetCreatedAt().GetData() != want.GetCreatedAt().GetData() {
+				emit(fmt.Sprintf("query checkpoint %d created_at %d does not match the audit chain %d", id, row.GetCreatedAt().GetData(), want.GetCreatedAt().GetData()))
+			}
+		}
+	}
+
+	return nil
+}
+
+// foldBaselineQueryCheckpoints seeds the audit-derived live-checkpoint set from
+// the boundary-time baseline, whose SubGlobQueryCheckpoint rows (with their
+// max_sequence / created_at) stand in for the pre-archive CreatedQueryCheckpoint
+// logs purged with the archived chapter.
+func (c *Checker) foldBaselineQueryCheckpoints(baselineDB *pebble.DB, into map[uint64]*commonpb.CreatedQueryCheckpointLog) error {
+	rows, err := query.ReadQueryCheckpointRows(baselineDB)
+	if err != nil {
+		return fmt.Errorf("folding baseline query checkpoints: %w", err)
+	}
+
+	for id, cp := range rows {
+		into[id] = &commonpb.CreatedQueryCheckpointLog{
+			CheckpointId: id,
+			MaxSequence:  cp.GetMaxSequence(),
+			CreatedAt:    cp.GetCreatedAt(),
+		}
+	}
+
+	return nil
 }
 
 // numscriptVersionGreater reports whether a is a strictly greater full semver
--- a/internal/domain/processing/processor_query_checkpoint.go
+++ b/internal/domain/processing/processor_query_checkpoint.go
@@ -6,8 +6,13 @@ import (
 	"github.com/formancehq/ledger/v3/internal/proto/raftcmdpb"
 )
 
+// processCreateQueryCheckpoint applies a CreateQueryCheckpoint order
+// unconditionally: the live-checkpoint limit is enforced softly at admission, so
+// a committed create always applies identically on every node regardless of that
+// node's configured limit (FSM determinism).
 func processCreateQueryCheckpoint(order *raftcmdpb.CreateQueryCheckpointOrder, ctx *Context) (*commonpb.LogPayload, domain.Describable) {
 	s := ctx.Scope
+
 	checkpointID := s.IncrementNextQueryCheckpointID()
 
 	cp := &raftcmdpb.QueryCheckpointState{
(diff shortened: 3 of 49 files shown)


internal/query/query_checkpoint.go
```diff
@@ -53,3 +53,21 @@ func ListQueryCheckpoints(reader dal.PebbleReader) ([]*raftcmdpb.QueryCheckpoint
...
+// enforce the checkpoint cap and reject deletes of non-live IDs without
+// scanning Pebble on the apply path.
+func ReadLiveQueryCheckpointIDs(reader dal.PebbleReader) (map[uint64]struct{}, error) {
+	checkpoints, err := ListQueryCheckpoints(reader)
+	if err != nil {
+		return nil, err
+	}
+
+	ids := make(map[uint64]struct{}, len(checkpoints))
+	for _, cp := range checkpoints {
```
NumaryBot: 🟠 **[major] Use storage keys when counting checkpoints**

When a `SubGlobQueryCheckpoint` row's key and embedded `checkpoint_id` differ, this builds the live set from the payload ID and collapses/renames the stored row. In that corrupted or manually repaired-store scenario, recovery under-counts the actual rows for the cap and `compareQueryCheckpoints` can miss a phantom key if the payload ID is audit-justified; the persisted projection should be keyed/validated by the Pebble key ID instead of only `cp.GetCheckpointId()`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1660#issuecomment-5144065855

flemzord (changes requested):
Revalidated current head `3163e7fc`. I confirm the existing NumaryBot finding as a correctness/integrity blocker: recovery, cap enforcement, and the new checker derive the live set from embedded checkpoint IDs without validating them against the persisted Pebble key IDs. A mismatched row can therefore under-count live rows and evade the projection check. I am not duplicating the inline evidence. The NumaryBot and Codecov failures are not the basis of this review.

flemzord:
One additional correctness/integrity blocker on current head `3163e7fc`: `compareQueryCheckpoints` explicitly checks only stored ⊆ audit-derived and treats an audit-live checkpoint with no stored row as legitimate. That lets deletion or loss of a checkpoint projection pass `Check()` silently, while recovery then seeds the FSM cap/existence set from that incomplete store and changes subsequent create/delete outcomes. This is not full verification of a persisted projection under invariant #8. Please compare both directions, or represent the restore-time invalidation in authoritative state so an arbitrary missing row cannot be indistinguishable from an intentional restore.

internal/query/query_checkpoint.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/pull/1660/commits/4fdc99064 — `ReadLiveQueryCheckpointIDs` now derives the id from the Pebble key instead of the payload `checkpoint_id`, so a row whose key and embedded id diverge can no longer under-count the cap at recovery or hide a phantom key from the checker. The checker pass is also bidirectional now.

Azorlogh:
@flemzord addressed in https://github.com/formancehq/ledger/pull/1660/commits/4fdc99064:

- Checkpoint IDs are now read from the **Pebble key**, not the payload `checkpoint_id` — recovery, cap enforcement, and the checker all go through `ReadLiveQueryCheckpointIDs`, which iterates keys. A key≠payload row can no longer under-count or evade the check.
- `compareQueryCheckpoints` now verifies **both directions**. To make the reverse direction sound rather than false-positive after a restore, the audit-rebuild path now recreates the checkpoint metadata rows from the `CreatedQueryCheckpoint`/`DeletedQueryCheckpoint` logs (`rebuild.go`) — like every other projection it rebuilds. The physical files still can't be reconstructed, so a rebuilt checkpoint reads `Unavailable` until deleted (the existing EN-1460 state). A missing row is therefore corruption, not a legitimate restore artifact.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1660#issuecomment-5144065855

internal/infra/backup/rebuild.go
```diff
@@ -393,6 +393,33 @@ func RebuildDelta(
...
 			}
 
+		case *commonpb.LogPayload_CreatedQueryCheckpoint:
+			// Rebuild the metadata row (id + max_sequence) from the log. The
+			// physical checkpoint files cannot be reconstructed from the audit,
+			// so a rebuilt checkpoint reads as Unavailable until an operator
+			// deletes it — the same registered-but-not-ready state EN-1460
+			// already defines. The row keeps the projection audit-consistent so
+			// the cap and compareQueryCheckpoints stay correct after a rebuild.
+			if cp := p.CreatedQueryCheckpoint; cp != nil {
```
NumaryBot, line 431: 🔴 **[blocker] Next checkpoint ID counter not persisted during audit rebuild**

When `RebuildDelta` replays logs containing query checkpoints, it restores the live checkpoint rows but does not restore `SubGlobNextQueryCheckpointID`. On the next recovery, `ReadNextQueryCheckpointID` defaults to 1, so subsequent creates can reissue already-used checkpoint IDs and silently overwrite restored live rows instead of allocating monotonically increasing IDs. This breaks the cap/existence invariant after any full audit rebuild or restore.

**Suggestion:** After replaying all log entries, compute `max(checkpoint_id) + 1` across all restored checkpoint rows and persist that value as `SubGlobNextQueryCheckpointID` so the counter is consistent with the restored state.

internal/query/query_checkpoint.go
```diff
@@ -53,3 +54,40 @@ func ListQueryCheckpoints(reader dal.PebbleReader) ([]*raftcmdpb.QueryCheckpoint
...
+// live in the store. Used at recovery to rehydrate FSMState so the FSM can
+// enforce the checkpoint cap and reject deletes of non-live IDs without
+// scanning Pebble on the apply path, and by the checker to compare the stored
+// rows against the audit chain.
+//
+// The ID is read from the Pebble KEY, never the payload's checkpoint_id: the
+// key is what CreateQueryCheckpoint / DeleteQueryCheckpoint address, so it is
+// the authoritative identity of the row. Trusting the payload would let a row
+// whose key and embedded id diverge (a corrupted or hand-repaired store)
+// collapse or rename in the derived set, under-counting the cap and hiding a
```
NumaryBot, line 68: 🟠 **[major] Live checkpoint set built from payload ID instead of storage key**

When counting or validating live checkpoints, the code uses `cp.GetCheckpointId()` from the Pebble row's payload rather than the key under which the row is stored. If a `SubGlobQueryCheckpoint` row's storage key and embedded `checkpoint_id` ever diverge (e.g. after a manual repair or a corrupted write), the live set is indexed by the wrong ID. This can cause under-counting toward the 10-checkpoint cap and allow `compareQueryCheckpoints` to miss phantom keys, since the audit validates by payload ID while the persisted projection is keyed differently.

**Suggestion:** Build the live-ID set from the Pebble storage key, not from the payload's `checkpoint_id` field. Validate that the two agree and treat any mismatch as a corrupt row.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1660#issuecomment-5144065855

internal/infra/backup/rebuild.go
```diff
@@ -393,6 +399,35 @@ func RebuildDelta(
...
+				if err := state.SaveQueryCheckpoint(batch, &raftcmdpb.QueryCheckpointState{
```
NumaryBot, line 432: 🟠 **[major] Checkpoint creation timestamp not persisted during audit rebuild**

When `RebuildDelta` restores checkpoints from `CreatedQueryCheckpoint` log entries, the reconstructed `QueryCheckpointState` omits `CreatedAt`. After a full rebuild, `ListQueryCheckpoints`/`GetQueryCheckpointInfo` will report the zero timestamp (epoch 1970) for those rows, making the restored projection non-equivalent to the original state.

**Suggestion:** Carry the proposal timestamp from the `CreatedQueryCheckpoint` log entry into the restored `QueryCheckpointState.CreatedAt` field so the rebuilt projection matches what was originally written.

internal/application/check/checker.go
```diff
@@ -1136,6 +1163,78 @@ func (c *Checker) compareMirrorV2LogID(reader dal.PebbleReader, chainBound *chai
...
+//   - a stored row absent from the audit set (never created, or later deleted)
+//     is a phantom or stale row;
+//   - an audit-live checkpoint with no stored row is a lost or dropped
+//     projection.
+//
+// The audit-rebuild path recreates the rows from the logs (internal/infra/backup),
+// so a missing row is never a legitimate restore artifact — it is corruption.
+// IDs come from the Pebble key (ReadLiveQueryCheckpointIDs), not the payload.
+func (c *Checker) compareQueryCheckpoints(reader dal.PebbleReader, derived map[uint64]struct{}, callback func(*servicepb.CheckStoreEvent)) error {
+	stored, err := query.ReadLiveQueryCheckpointIDs(reader)
```
NumaryBot: 🟡 **[minor] Checker validates checkpoint existence by key only, not row contents**

`compareQueryCheckpoints` confirms that the right checkpoint IDs are present (both directions, per the recent fix) but does not verify the stored proto contents — e.g. `max_sequence` or other metadata fields. A checkpoint row whose payload was silently corrupted or altered will pass `Check()` while `List`/`Get` expose the wrong data, so the new projection check does not fully verify the persisted checkpoint state.

**Suggestion:** After confirming ID presence in both directions, also read each stored checkpoint row and compare its key fields (e.g. `max_sequence`, `created_at`) against the values derived from the audit log to detect content-level corruption.

internal/domain/processing/processor_query_checkpoint.go
```diff
@@ -35,6 +44,10 @@ func processDeleteQueryCheckpoint(order *raftcmdpb.DeleteQueryCheckpointOrder, c
 		return nil, domain.ErrCheckpointIDRequired
 	}
 
+	if !ctx.Scope.QueryCheckpointExists(order.GetCheckpointId()) {
```
paul-nicolas: [Major] Existence-aware delete removes the only path that reclaimed stray checkpoint directories

Returning `CHECKPOINT_NOT_FOUND` before emitting `DeletedQueryCheckpointLog` also suppresses every downstream cleanup, because all of it is log-driven:

- `Machine.deleteQueryCheckpointFiles` runs post-commit from `pb.checkpointDeletes` (`internal/infra/state/machine.go:889-891`), which is populated from `r.QueryCheckpointDeleted` (`machine.go:789-793`) — a signal derived from the delete log.
- the read-index cleanup is gated the same way (`internal/application/indexbuilder/process_logs.go:159`).

I checked for any other reclamation path and there is none: `DeleteQueryCheckpointFiles` (`internal/storage/dal/store.go:955`) has exactly one caller, `queryCheckpointsDir` is referenced only by the create/delete/path helpers in `store.go`, and `cleanupOldCheckpoints` reaps `checkpointsDir` (Pebble backup checkpoints), not `query-checkpoints/`. `process_logs.go:252` states outright that there is no reconciler.

Reachable case: a follower that falls far enough behind to be caught up by a snapshot install receives the leader's Pebble state (row already absent) and never executes the per-entry post-commit hook, so its local `query-checkpoints/<id>/` survives. Before this PR an operator could reclaim it by re-issuing `delete <id>` — the log was emitted unconditionally and every node deleted its files. Now that call returns `NotFound` and does nothing, so the directory leaks permanently with no API or boot-time sweep to recover it. Each one is a full `db.Checkpoint()` whose hard-linked SSTs keep pinning disk as the live store compacts — the same unbounded-disk failure mode EN-1501 is closing, reintroduced through a different door.

Fix: keep the typed `NotFound` for the caller (it is the right contract and the live-count exactness argument holds), but do not let it be the only signal. Either emit a file-cleanup-only side effect on the not-found branch, or add a boot-time sweep that reconciles `query-checkpoints/*` against `LiveQueryCheckpointIDs` (which recovery already loads) and removes directories with no live row — the analogue of `purgeOrphanVersions` for the read index.

internal/application/check/checker.go
paul-nicolas: [Major] Checker pass verifies only checkpoint IDs, leaving the audit-backed `max_sequence` unverified

`compareQueryCheckpoints` reads `ReadLiveQueryCheckpointIDs`, which decodes IDs from the Pebble key and discards the payload. So the pass verifies *set membership* only. `QueryCheckpointState.max_sequence` is a persisted field of this projection, and it **is** audit-backed — `CreatedQueryCheckpointLog.max_sequence` (field 2, `misc/proto/common.proto:559`) carries it, and the rebuild path already reads it back (`rebuild.go:412`).

Why it matters: invariant #8 requires the checker to verify every projection it persists, and no other pass touches `SubGlobQueryCheckpoint` (grep: only `baseline.go`, `query_checkpoint.go`, `batch.go`, and this file). A store whose `max_sequence` is edited in place therefore passes `Check()` clean, and the tampered value is served to clients by `ListQueryCheckpoints` / `GetQueryCheckpointInfo` (`internal/adapter/grpc/server_cluster.go:473` and `:577`) and printed by `ledgerctl query-checkpoint list/info`. It does not change what the read index serves (the indexbuilder materializes from the log's value inline), so this is an integrity/reporting gap rather than a query-correctness one — but it is exactly the "projection the checker does not verify is a tampering vector" case, and the PR adds this pass precisely to close that gap.

Fix: make `derived` a `map[uint64]*commonpb.CreatedQueryCheckpointLog` (or a small struct carrying `maxSequence`) instead of `map[uint64]struct{}`, load the full stored rows with `query.ListQueryCheckpoints`, and compare `max_sequence` per ID, emitting `CHECK_STORE_ERROR_TYPE_QUERY_CHECKPOINT_MISMATCH` on divergence. The baseline seed can carry the stored `max_sequence` the same way it carries the ID today.

internal/infra/backup/rebuild.go
paul-nicolas, line 432: [Medium] Rebuild silently drops `created_at`, and the field has no audit backing at all

The rebuilt row is constructed with only `CheckpointId` and `MaxSequence`, so `QueryCheckpointState.created_at` (`misc/proto/raft_cmd.proto:217`) comes back as nil after any restore + `RebuildDelta`. That value is user-visible: it is returned by `ListQueryCheckpoints` / `GetQueryCheckpointInfo` (`internal/adapter/grpc/server_cluster.go:473`, `:577`) and rendered by `ledgerctl query-checkpoint list` / `info`, so every live checkpoint reports an empty creation time post-restore.

The root cause is that `CreatedQueryCheckpointLog` only carries `checkpoint_id` and `max_sequence` (`misc/proto/common.proto:557-560`) — the timestamp is never audited, so it is not merely dropped here, it is *unrecoverable* and permanently unverifiable by any checker pass. Combined with the finding on `compareQueryCheckpoints`, `created_at` is a persisted projection field with zero audit binding, which is the invariant-#8 gap in its strongest form.

Fix: add `common.Timestamp created_at = 3;` to `CreatedQueryCheckpointLog`, populate it in `processCreateQueryCheckpoint` from the same proposal date the row already uses, run `just generate-proto`, set it on the rebuilt row here, and extend the checker comparison to cover it. v3 is unreleased, so there is no wire-compat cost. Worth a rebuild regression test asserting `CreatedAt` survives — `rebuild_test.go` currently only asserts the ID set.

paul-nicolas (commented):

Reviewed at head `5e2601ed1` from a dedicated worktree pinned to that SHA. `GOROOT= go build ./...` is green and all changed packages pass (`internal/domain/processing`, `internal/infra/state`, `internal/application/check`, `internal/query`, `internal/infra/backup`, `internal/infra/attributes`, `internal/domain`).

**Findings: 3 posted inline — 2 Major, 1 Medium.** Both models converged on the same cluster of issues; each was then verified against the real code before posting, and everything below is mechanically confirmed rather than inferred.

The three findings share one root: the *set* of live checkpoints is now rigorously audited, but the *contents* of each checkpoint row and the *filesystem* behind it are not. `max_sequence` is audit-backed yet uncompared, `created_at` is not audited at all, and the new existence-aware delete silently drops the only cleanup path for orphaned checkpoint directories.

### What I checked and found sound

Worth stating explicitly, since these are the parts most likely to hide a determinism or consistency bug:

- **Live-set consistency across every reconstruction path.** `Recovery.RecoverState` is the single path that rebuilds `FSMState`, and it is called on both restart *and* follower sync (`recovery.go:49`), so the new `LiveQueryCheckpointIDs` rehydration is reached on every boot path. `FSMState` is not proto-serialized into the Raft snapshot — the snapshot is the Pebble checkpoint — so Pebble rehydration is the correct and only mechanism; there is no snapshot field left to update.
- **Concurrency.** The map is touched only in `fsmstate.go`, `write_set.go`, and tests — never by the query path, the scheduler, or a health path. No unsynchronised map read/write.
- **Overlay semantics.** The lazy copy makes rollback correct (`Reset` nils it, so an aborted proposal leaves `FSMState` untouched), and the cap check precedes `IncrementNextQueryCheckpointID`, so a rejected create consumes no ID. No Pebble read is added on the hot path (invariant #3), and the gate reads only `len`/lookups, so there is no map-iteration determinism hazard (invariant #2).
- **Rebuild counter restore.** `RebuildDelta` is incremental on top of a restored Pebble checkpoint that already carries `SubGlobNextQueryCheckpointID`, so the "delta has no create logs" case correctly leaves the counter alone, and when the delta does have creates, `max+1` is exactly right. No ID-reuse or row-overwrite window — this addresses the concern the head commit was written for.
- **No checker false positive on restore.** The bidirectional check is safe because restore carries the rows and `RebuildDelta` patches the delta; the `inDerived && !inStored` direction cannot fire on a legitimate restore.
- **Error wiring.** Both reasons are complete end to end — the string→enum conversion is generic (`ErrorReason_value["ERROR_REASON_"+…]`), so no per-reason registration is missing, and `KindForReason` handles both. Freezability is correct: `CHECKPOINT_LIMIT_REACHED` → `ResourceExhausted` is not freezable (`IsFreezableFailure`, `errors.go:151`), sidestepping the idempotency-freeze hazard as intended, while `CHECKPOINT_NOT_FOUND` → `NotFound` is freezable, which is right since IDs are monotonic and never reused.
- **Proto and generated code** are consistent and sequentially numbered; mocks for the two new `Scope` methods are properly regenerated.
- **Docs are accurate.** `query-checkpoints.md`, `cli.md`, `api-comparison.md`, and the invariant-#8 pass list in `AGENTS.md` all correctly describe the **bidirectional** pass. No REST/openapi surface is affected (gRPC ClusterService only), so that claim holds too.
- **E2E test is properly isolated** — a dedicated single node on its own ports (9224/8224), so the *global* cap cannot leak across specs, and it asserts the typed reasons rather than just "an error occurred". Good test.

### Note on the PR description

The description still says the checker pass is "One-directional (`stored ⊆ derived`): the reverse is intentionally not flagged because checkpoint rows are deliberately not rebuilt on restore." The code, the docs, and `AGENTS.md` are all bidirectional, and rows *are* rebuilt on restore — the last commit changed this. Worth updating the body so it does not contradict the merged design.

- `internal/infra/state/query_checkpoint_scheduler.go:131` — the `limitReached` state machine itself (log-once, stay-armed, reset-on-success) has no test; only the `isCheckpointLimitReached` classifier is covered. "Stays armed so creation resumes after a delete" is the behavioural claim in the PR body and the docs, and nothing asserts it at any level. Consider a small loop-level test driving `proposeFn` through limit → limit → success.
- `internal/infra/state/write_set.go:1745` — `QueryCheckpointCount`/`QueryCheckpointExists` call `ensureLiveQueryCheckpoints()`, so even a *rejected* create (cap full) or a read-only existence check allocates a full copy of the set and makes `Merge` swap in a content-identical map. Harmless, but the copy could be deferred to the first mutation by having the read paths fall back to `b.fsm.State.LiveQueryCheckpointIDs` when the overlay is nil.

gfyrag:
Blocking design concern: EN-1501 establishes why query-checkpoint cardinality must be bounded, but neither the ticket nor this PR provides a legitimate basis for the value **10**. The fact that `--pebble-max-checkpoints` defaults to 10 is not sufficient evidence: maintenance checkpoints and query checkpoints have different lifecycle, storage cost, and product semantics. There is no sizing estimate, retention requirement, workload analysis, or API/SLA constraint showing that 10 is the correct boundary. Hard-coding it makes an arbitrary number part of the v3 protocol contract.

This should not be “fixed” by simply reusing `--pebble-max-checkpoints` or introducing another node-local CLI/env flag in the FSM path. The existing flag controls local Pebble housekeeping; the query-checkpoint limit determines whether a replicated command succeeds or fails. During a rolling upgrade—or with any configuration drift—nodes could run with different values and apply the same Raft entry differently, which would violate FSM determinism and can desynchronize replicated state.

If the limit is configurable, it needs cluster-wide replicated semantics: the value must be committed through Raft, take effect at a precise applied index on every node, and have explicit compatibility/validation rules for rolling upgrades. If it remains a protocol constant, the chosen value needs a documented product/operational rationale rather than being copied from an unrelated default.

Please document both the derivation of the limit and the safe configuration/rollout model before merging.

internal/infra/state/batch.go
```diff
@@ -340,6 +346,19 @@ func storeNextQueryCheckpointID(b *dal.WriteSession, id uint64) error {
 	return nil
```
NumaryBot, line 355: 🔴 **[blocker] Verify the persisted checkpoint limit against the audit chain**

If `SubGlobQueryCheckpointLimit` is corrupted on one replica, recovery loads the altered value and that node can accept or reject later checkpoint creations differently from its peers, yet `Check()` never compares this new projection with the latest audited `SetQueryCheckpointLimitLog`. Add a checker pass deriving and validating the limit as required by [AGENTS.md:21](/data/worktrees/formancehq__ledger-pr-1660/AGENTS.md#L21).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1660#issuecomment-5144065855

Azorlogh:
Reworked to a cluster-wide, **Raft-committed configurable limit** (default 10), applied deterministically in the FSM at a precise applied index — not a node-local flag. `ledgerctl query-checkpoint set-limit/get-limit`; rationale + rollout documented. https://github.com/formancehq/ledger/pull/1660/commits/858e115c1

internal/domain/processing/processor_query_checkpoint.go
Azorlogh: Fixed: `RecoverState` now sweeps `query-checkpoints/*` against `LiveQueryCheckpointIDs` and reclaims row-less dirs (best-effort). https://github.com/formancehq/ledger/pull/1660/commits/858e115c1

internal/application/check/checker.go
Azorlogh: Fixed: `compareQueryCheckpoints` now verifies `max_sequence` (and `created_at`) per row, both directions. https://github.com/formancehq/ledger/pull/1660/commits/858e115c1

internal/infra/backup/rebuild.go
Azorlogh: Fixed: `created_at` added to `CreatedQueryCheckpointLog`, populated at create, restored on rebuild, verified in the checker. https://github.com/formancehq/ledger/pull/1660/commits/858e115c1

internal/infra/state/batch.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/pull/1660/commits/006551430 — added `compareQueryCheckpointLimit`: the stored limit is verified against the value re-derived from the latest `SetQueryCheckpointLimit` log (baseline-seeded under archiving), emitting `CHECK_STORE_ERROR_TYPE_QUERY_CHECKPOINT_LIMIT_MISMATCH` on divergence.

internal/application/check/checker.go
```diff
@@ -1136,6 +1187,117 @@ func (c *Checker) compareMirrorV2LogID(reader dal.PebbleReader, chainBound *chai
...
+	return nil
+}
+
+// compareQueryCheckpointLimit verifies the stored cluster-wide checkpoint limit
+// (SubGlobQueryCheckpointLimit) against `derived`, the value re-derived from the
+// latest SetQueryCheckpointLimit log (baseline-seeded under archiving; default
+// when never set). The limit gates a replicated command, so a tampered value
+// would make one replica accept/reject creations differently from its peers —
+// hence it is a verified projection (invariant #8), not merely rebuildable.
+func (c *Checker) compareQueryCheckpointLimit(reader dal.PebbleReader, derived uint64, callback func(*servicepb.CheckStoreEvent)) error {
```
NumaryBot: 🔴 **[blocker] Verify the persisted checkpoint limit against the audit chain**

When the audited limit was explicitly set to the default value `10`, deleting `SubGlobQueryCheckpointLimit` is invisible here because `ReadQueryCheckpointLimit` substitutes the same default for a missing key. The checker therefore accepts a lost persisted projection instead of verifying its presence; track whether the audit contains a set operation and distinguish an absent row accordingly, as required by [AGENTS.md:21](/data/worktrees/formancehq__ledger-pr-1660/AGENTS.md#L21).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1660#issuecomment-5144065855

gfyrag:
I think the limit should be enforced at admission time rather than by the FSM.

The important rolling-upgrade property is that once a `CreateQueryCheckpoint` command is committed, every node applies it identically, regardless of its local version or infrastructure configuration. During a mixed-version rollout, different leaders may make different admission decisions, but this only affects whether a command is proposed—not how a committed Raft entry is applied.

Proposed changes:

1. Remove the replicated `SetQueryCheckpointLimit` setting and its Raft order, API, `ledgerctl set-limit/get-limit` commands, persisted projection, and checker pass.
2. Keep `CreateQueryCheckpoint` unconditional in the FSM: no node-local or version-dependent limit check on the apply path.
3. Make the limit an infrastructure-owned startup configuration, with no runtime mutation through `ledgerctl`.
4. Before proposing `CreateQueryCheckpoint`, admission checks the leader's current live checkpoint count and rejects the request when the configured limit is reached.
5. Accept that concurrent admissions may exceed the configured limit until a checkpoint is deleted. The possible overshoot is tied to the number of concurrent in-flight creations. Avoid adding reservation or serialization machinery unless we decide that the limit must be strict.
6. Document the rolling-upgrade behavior explicitly: an old leader may not enforce the limit, while an upgraded leader does; committed commands remain deterministic on every replica.
7. Add admission tests for rejection at the limit and acceptance below it. Keep the existing checker coverage for checkpoint rows and IDs; only the replicated-limit verification should disappear.

This makes the limit an operational safeguard rather than a Raft protocol invariant, which is the intended boundary here.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1660#issuecomment-5144065855

Azorlogh:
Done — refactored to soft admission-time enforcement per this plan, and rebased onto the latest `release/v3.0` as a single commit: 83d7a128200630e1707a410227d47694beee3490

Mapping to your 7 points:

1. Removed the replicated `SetQueryCheckpointLimit` order/log/request, the `GetQueryCheckpointLimit` RPC, `ledgerctl set-limit`/`get-limit`, the persisted `SubGlobQueryCheckpointLimit` projection, and its `compareQueryCheckpointLimit` checker pass + `QUERY_CHECKPOINT_LIMIT_MISMATCH` error type.
2. `CreateQueryCheckpoint`/`DeleteQueryCheckpoint` apply unconditionally in the FSM — no count, no gate, no id check.
3. The limit is an infrastructure-owned startup config: `--query-checkpoint-limit` flag (default 10) / env `QUERY_CHECKPOINT_LIMIT` / operator CRD `queryCheckpointLimit`. No runtime mutation.
4. Admission counts the leader's live checkpoints (`query.ReadLiveQueryCheckpointIDs`) before proposing a create and rejects at the limit; delete existence/id validity is checked the same way.
5. Concurrent overshoot accepted, bounded by in-flight creates. No reservation/serialization added.
6. Rolling-upgrade behavior documented in `docs/technical/architecture/subsystems/read-path/query-checkpoints.md`: a pre-upgrade leader may not enforce, an upgraded one does; committed commands stay deterministic on every replica.
7. Added admission tests (rejection at the limit, acceptance below it, `id==0`/non-live delete). Kept `compareQueryCheckpoints` (rows + IDs, both ways); only the replicated-limit verification is gone.

One addition beyond the plan: `Config.Validate()` rejects `--query-checkpoint-limit 0` at boot, since 0 would silently reject every create.

internal/application/admission/admission.go
```diff
@@ -2246,12 +2315,20 @@ func (a *Admission) requestToOrder(ctx context.Context, req *servicepb.Request,
 			reqType.SaveNumscript.GetContent(),
 		)
 	case *servicepb.Request_CreateQueryCheckpoint:
```
NumaryBot, line 2332: 🟠 **[major] Account for earlier deletes when enforcing the checkpoint limit**

When a valid batch deletes a live checkpoint and then creates one, this check reads only committed Pebble state and ignores the batch overlay. At the configured limit, the create is rejected even though the preceding delete frees a slot atomically, forcing clients to split an otherwise valid batch. Track checkpoint additions/deletions within the bulk overlay when evaluating the limit.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1660#issuecomment-5144065855

internal/application/admission/admission.go
Azorlogh: Fixed in 0177b3da0b9421802a8e188d6d5d4d7712933805.

Admission now folds same-bulk effects via the bulk overlay: the limit check evaluates `committed + bulkCreates − bulkDeletes`, so a batch that deletes a live checkpoint then creates one is admitted atomically, and a second create in one bulk counts the first. The existence check also rejects a repeated delete of the same id within a bulk. Added unit coverage for both.

internal/application/admission/admission.go
```diff
@@ -2246,12 +2330,20 @@ func (a *Admission) requestToOrder(ctx context.Context, req *servicepb.Request,
...
```
NumaryBot, line 2332: 🔴 **[blocker] Preserve idempotent checkpoint retries through admission**

When a keyed checkpoint request is retried after succeeding, these state-dependent checks run before the FSM idempotency gate: a create retry at the cap returns `CHECKPOINT_LIMIT_REACHED`, while a delete retry returns `CHECKPOINT_NOT_FOUND`. Both should instead replay the original response as required by [idempotency.md:26-35](/data/worktrees/formancehq__ledger-pr-1660/docs/technical/architecture/subsystems/admission/idempotency.md#L26-L35); keyed requests must be forwarded so the FSM can resolve the frozen outcome.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1660#issuecomment-5144065855

Azorlogh:
Superseded by #1767 (part of the EN-1827 plan). The soft admission-time enforcement here breaks idempotent retries (a create that fills the limit, or a delete, returns a different outcome when retried with the same key); EN-1827 moves the cap + existence checks into the FSM apply path after the idempotency gate, using a Raft-replicated cluster policy for the limit. That work landed as:

- #1750 — replicated cluster policy (the limit's home)
- #1767 — query-checkpoint limit enforced in the FSM (stacked on #1750)

Closing this one as superseded.

PR #1662, opened 2026-08-01
feat: add asynchronous historical balances projection

flemzord:
## Status

> [!IMPORTANT]
> This remains an exploratory draft for architecture, correctness, and performance review. It does not enable the feature by default or authorize a production rollout.

- Base: `release/v3.0`
- Scope: historical monetary account balances only
- Activation: explicitly configured per ledger by API clients
- Authority: the audit chain; the projection is local and rebuildable

## Why

Ledger v3 lacks the monetary point-in-time balance capability available in v2. This PR restores that read use case without historizing transaction objects, metadata, schemas, account types, or indexes, and without adding Pebble reads or secondary-store writes to the Raft/FSM hot path.

The narrower name **historical balances** distinguishes this projection from the broader v2 PIT behavior.

## Contract

Historical reads support two temporalities:

- **effective**: fold monetary effects using their business timestamp;
- **insertion**: fold monetary effects using their FSM-assigned insertion timestamp.

Metadata filters still select accounts from the current read-store snapshot. The returned money is historical; metadata remains current-state-only.

Configuration is an audited, ledger-scoped client command. There is no system flag or server-side ledger allowlist. Replica-local flags only tune batch size, backfill yielding, and durability.

## Architecture

```text
client ConfigureHistoricalBalances
        |
        v
admission -> Raft -> deterministic FSM -> audited ledger log
                                            |
                                            v post-commit notification
                                  per-replica history builder
                                            |
                          verified hot/cold audit source replay
                                            |
                                  local Pebble peer store
                                            |
                         immutable segments + pinned manifests
                                            |
                                historical balance queries
```

The builder subscribes to the same post-commit notification fanout used by the index builder. A 200 ms ticker remains a correctness fallback and coalescing boundary, including proposals that emit no log.

The source can read existing cold audit chapters during rebuild. The projection itself has no cold tier, archive, remote GC, content checksum, semantic digest, or certifier.

## Storage model

Projection keys use the ledger **name**, consistently with business-facing read-side identities. Pebble prefix compression makes repeated ledger-name prefixes inexpensive, and ledger-name recreation is rejected by the lifecycle reducer.

Each immutable segment contains only account-level identities:

```text
temporality / ledger-name / account / asset-base / precision / color / timestamp
```

Values are cumulative input/output volume pairs. There is no ledger-level aggregation row; unfiltered queries aggregate account rows at read time. Logical compaction bounds the number of segments a view must merge.

See:

- [`historical-balances.md`](docs/technical/architecture/subsystems/read-path/historical-balances.md)
- [`historical-balances-performance.md`](docs/technical/architecture/subsystems/read-path/historical-balances-performance.md)

## Performance

Local read benchmarks show query age has a modest impact, while segment count is the dominant variable:

| Shape | Approximate latency |
|---|---:|
| one segment, unfiltered | 115–133 µs |
| one segment, 100-account filter | 166–222 µs |
| one segment in segment-scaling benchmark | 1.14 ms |
| four segments | 4.26 ms |
| sixteen segments | 19.61 ms |

Removing ledger-level aggregate rows roughly halves the logical temporal rows and eliminates duplicate write/compaction amplification. The trade-off is that an unfiltered ledger query scans and aggregates account rows.

## Integrity and lifecycle

- The FSM only records deterministic audited configuration; it never reads the history store.
- The local store fails closed when rebuilding, behind, source-missing, structurally corrupt, or unsupported.
- Structural verification and source watermarks trigger quarantine/rebuild where applicable.
- Semantic authority remains the audit chain. Projection-specific checksums are deliberately out of scope because the projection is rebuildable.
- Deleting a ledger removes it from the configured projection set during replay/rebuild.
- Reducer restoration rejects duplicate ledger-name incarnations and enabled-but-inactive ledgers.

## Validation

The branch was rebased on the latest `release/v3.0` and validated with:

```text
GOROOT= go test ./...
just test-full
GOROOT= go test -race -tags scenario -p 1 ./tests/scenarios/... -timeout 300s
cd tests/antithesis/workload && GOROOT= go test ./...
just test-e2e
nix develop --command bash -c "just pre-commit"
GOROOT= go build ./...
```

GitNexus reports the full diff as CRITICAL by blast radius because it changes API/protobuf contracts and touches 219 execution flows. The final review found no residual projection archive/checksum/retention code, no uncovered new FSM cache read, and no unexpected primary-store write path.

## Suggested review order

1. `docs/technical/architecture/subsystems/read-path/historical-balances.md`
2. `internal/domain/balancehistory/`
3. `internal/domain/processing/processor_historical_balances.go`
4. `internal/application/balancehistory/`
5. `internal/storage/balancehistorystore/`
6. `internal/application/ctrl/volume_view.go`
7. API, CLI, bootstrap, E2E, and Antithesis changes

## Review questions

- Is the account-only historical balance scope sufficiently explicit?
- Is audited per-ledger client configuration the right activation contract?
- Are the effective/insertion temporalities compatible with the required v2 monetary semantics?
- Is read-time ledger aggregation an acceptable trade-off for lower write and storage amplification?
- Are the fail-closed and rebuild behaviors sufficient for a rebuildable peer projection?


flemzord:
Revalidated live head `0b519d9a`. One correctness/integrity blocker remains in the startup readiness path for the persisted PIT projection. Balance history is deliberately outside the global checker (`internal/bootstrap/balance_history.go:27-30`), but on an ordinary restart `completeCaughtUpHistory` skips `Certify` unless a repair flag is already set (`internal/application/balancehistory/builder.go:725-731`), then `markReadyAfterReconciliation` opens the read gate (`:752-778`). The verifier starts concurrently and its immediate pass is only bounded; with the production defaults the first full source-to-projection semantic replay occurs after 96 fifteen-minute passes (`internal/application/balancehistory/verifier.go:24-35,276-279,333-336`). A semantically falsified but structurally/checksum-consistent peer projection can therefore serve incorrect PIT balances before the first full certification. Please require a complete semantic certification before the process-local gate first becomes ready after startup, or cover this projection from the authoritative audit in the global checker. The Codecov failures are CI/tooling and are not the basis for this blocker.

flemzord:
Revalidated live head `0b519d9a`. One additional correctness/API blocker is not covered by the existing threads: `parseHTTPPointInTime` distinguishes whether the `pit` key is present, but then collapses a single empty value back to the absent case (`internal/adapter/http/handlers_aggregate_volumes.go:169-200`). Consequently `GET /{ledger}/volumes?pit=` returns the current live balances with HTTP 200 instead of rejecting the malformed historical selector. This silently substitutes a different temporal view and violates the documented RFC3339/fail-closed PIT contract. Please distinguish absent from present-but-empty, return 400 for the latter, and add an HTTP regression for `?pit=`. The existing startup-certification blocker and Codecov failures are not repeated here. GitHub does not permit the authenticated author account to submit `CHANGES_REQUESTED` on its own PR.

Azorlogh:
**Cross-store selection tear in filtered PIT aggregates** (the EN-1748 skew, inherited by `SelectedAccounts`)

The PIT arm of `AggregateVolumes` carefully binds the historical view to the primary handle's head (`requiredLogSequence = max(read.MinLogSequence, ReadLastSequence(handle))`) — but the metadata **selection** then takes `ctrl.readStore.NewSnapshot()` with no alignment to that same handle. When the read-store fold lags, the response's own provenance claims coverage its selection doesn't have:

```
selection tear: view logWatermark=23441 covers member "member:16" (commitSeq=23051),
yet team=blue aggregate input=15 < 16 members (max observed fold lag 542 seqs)
```

One committed bulk funds `member:N` at a historical effective date **and** tags it `team=blue`; the immediately-following filtered PIT aggregate returns a view watermark ≥ that commit while omitting the member's money — the pinned manifest faithfully folds over a stale account set. Pinned money, torn selection.

Reproduced on this branch **unmodified** — the fold lag comes purely from load: 8 live indexes multiply per-log fold cost, 6 goroutines sustain applies, and the fold falls ~540 sequences behind the raft head. Fails in ~60s on my machine.

<details><summary>e2e repro — drop into <code>tests/e2e/business/</code>, run <code>go test -tags e2e ./tests/e2e/business/ -args --ginkgo.focus 'PIT filtered aggregate selection tear'</code></summary>

```go
//go:build e2e

package business

import (
	"fmt"
	"math/big"
	"sync"
	"time"

	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
	"github.com/formancehq/ledger/v3/internal/proto/servicepb"
	"github.com/formancehq/ledger/v3/pkg/actions"
	"github.com/formancehq/ledger/v3/pkg/testserver"
	"github.com/formancehq/ledger/v3/tests/e2e/testutil"
	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"
)

var _ = Describe("PIT filtered aggregate selection tear", Serial, func() {
	It("never returns a view watermark covering a commit its selection missed", func() {
		ctx, client, _ := testutil.SetupSingleNode(9761, 8761, testserver.WithBalanceHistoryEnabled())

		const ledgerName = "pit-selection-tear"
		effective := time.Date(2021, time.March, 1, 12, 0, 0, 0, time.UTC)

		_, err := client.Apply(ctx, servicepb.UnsignedApplyRequest("", actions.CreateLedgerWithSchemaAction(ledgerName, nil, []*commonpb.SetMetadataFieldTypeCommand{
			{TargetType: commonpb.TargetType_TARGET_TYPE_ACCOUNT, Key: "team", Type: commonpb.MetadataType_METADATA_TYPE_STRING},
			{TargetType: commonpb.TargetType_TARGET_TYPE_ACCOUNT, Key: "tier", Type: commonpb.MetadataType_METADATA_TYPE_STRING},
		})))
		Expect(err).To(Succeed())

		// Fold cost per log is multiplicative in the live index count.
		for _, req := range []*servicepb.Request{
			actions.CreateAccountMetadataIndexAction(ledgerName, "team"),
			actions.CreateAccountMetadataIndexAction(ledgerName, "tier"),
			actions.CreateAccountAssetIndexAction(ledgerName),
			actions.CreateBuiltinTxIndexAction(ledgerName, commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_TIMESTAMP),
			actions.CreateBuiltinTxIndexAction(ledgerName, commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_INSERTED_AT),
			actions.CreateBuiltinTxIndexAction(ledgerName, commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_ADDRESS),
			actions.CreateBuiltinTxIndexAction(ledgerName, commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_SOURCE_ADDRESS),
			actions.CreateBuiltinTxIndexAction(ledgerName, commonpb.TransactionBuiltinIndex_TX_BUILTIN_INDEX_DESTINATION_ADDRESS),
		} {
			_, err := client.Apply(ctx, servicepb.UnsignedApplyRequest("", req))
			Expect(err).To(Succeed())
		}
		Expect(actions.WaitForMetadataIndexReady(ctx, client, ledgerName, commonpb.TargetType_TARGET_TYPE_ACCOUNT, "team")).To(Succeed())

		// Sustained apply pressure: fold work per log far exceeds apply work,
		// so the read-store fold falls behind while raft keeps committing.
		stop := make(chan struct{})
		var pressure sync.WaitGroup
		for w := 0; w < 6; w++ {
			pressure.Add(1)
			go func(w int) {
				defer GinkgoRecover()
				defer pressure.Done()
				n := 0
				for {
					select {
					case <-stop:
						return
					default:
					}
					reqs := make([]*servicepb.Request, 0, 30)
					for j := 0; j < 30; j++ {
						n++
						a := fmt.Sprintf("load:%d:%d:a", w, n%128)
						b := fmt.Sprintf("load:%d:%d:b", w, n%128)
						reqs = append(reqs, actions.CreateTransactionAction(ledgerName, []*commonpb.Posting{
							actions.NewPosting("world", a, big.NewInt(1), "COIN"),
							actions.NewPosting("world", b, big.NewInt(1), "EUR"),
							actions.NewPosting("world", a, big.NewInt(1), "USD/2"),
						}, nil, map[string]*commonpb.MetadataMap{
							a: {Values: map[string]*commonpb.MetadataValue{"tier": commonpb.NewStringValue("gold")}},
							b: {Values: map[string]*commonpb.MetadataValue{"tier": commonpb.NewStringValue("silver")}},
						}))
					}
					if _, err := client.Apply(ctx, servicepb.UnsignedApplyRequest("", reqs...)); err != nil {
						return
					}
				}
			}(w)
		}
		defer func() { close(stop); pressure.Wait() }()

		deadline := time.Now().Add(120 * time.Second)
		members := int64(0)
		checked := 0
		maxLag := int64(0)

		for time.Now().Before(deadline) {
			member := fmt.Sprintf("member:%d", members+1)

			// One committed bulk: fund the member at the historical effective
			// date AND tag it team=blue — money and metadata share one commit.
			resp, err := client.Apply(ctx, servicepb.UnsignedApplyRequest("",
				actions.WithTimestamp(actions.CreateForceTransactionAction(ledgerName, []*commonpb.Posting{
					actions.NewPosting("world", member, big.NewInt(1), "USD"),
				}, nil), effective),
				actions.SaveAccountMetadataAction(ledgerName, member, map[string]string{"team": "blue"}),
			))
			Expect(err).To(Succeed())
			members++
			commitSeq := pointInTimeLastLogSequence(resp)

			if status, err := client.GetIndexStatus(ctx, &servicepb.GetIndexStatusRequest{Ledger: ledgerName}); err == nil {
				if lag := int64(commitSeq) - int64(status.GetLastIndexedSequence()); lag > maxLag {
					maxLag = lag
				}
			}

			// No client pin: the view binds itself to the primary head.
			result, view, err := pointInTimeAggregate(ctx, client, &servicepb.AggregateVolumesRequest{
				Ledger: ledgerName,
				Filter: actions.StringMetadataFilter("team", "blue"),
				PointInTime: pointInTimeSelector(
					effective.Add(time.Hour),
					servicepb.PointInTimeAxis_POINT_IN_TIME_AXIS_EFFECTIVE,
				),
			})
			if err != nil {
				continue // fail-closed rejections are the honest outcome
			}
			if view.GetLogWatermark() < commitSeq {
				continue // the view never claimed to cover this commit
			}
			checked++

			var input int64
			for _, v := range result.GetVolumes() {
				if v.GetAsset() == "USD" && v.GetColor() == "" {
					input = v.GetInput().ToBigInt().Int64()
				}
			}

			if input < members {
				Fail(fmt.Sprintf(
					"selection tear: view logWatermark=%d covers member %q (commitSeq=%d), yet team=blue aggregate input=%d < %d members (max observed fold lag %d seqs)",
					view.GetLogWatermark(), member, commitSeq, input, members, maxLag))
			}
		}

		Expect(checked).To(BeNumerically(">", 0), "no covering view produced — inconclusive")
	})
})
```

</details>

Not a blocker for the exploratory evaluation — the fix belongs to EN-1748 (align the selection snapshot's fold cursor with the handle), after which `M` gets a precise meaning and the view token can record the selection sequence alongside the watermarks, making filtered PIT responses fully reproducible.

gfyrag:
Thanks for the extensive exploration here. The core historical-balance projection is valuable, but I think the current PR combines too many independent mechanisms into the first iteration.

I would strongly reduce the initial scope to a local, per-account, rebuildable historical-balance projection.

## 1. Scope and naming

This is not an isomorphic restoration of v2 point-in-time semantics.

The v2 `pit` affected accounts, transactions, transaction metadata and reversion state, in addition to balances and volumes. This implementation only historizes monetary effects.

I suggest naming the subsystem and native API around **historical balances** or **balance history**, and reserving “PIT” for a future temporal view covering multiple ledger resources. The HTTP `pit` parameter may remain as a compatibility mapping for aggregate volumes.

## 2. Client-driven configuration

The business decision to enable historical balances should not be controlled by deployment-level flags.

In particular, `--balance-history-ledgers` only gates reads: the builder and store still process every ledger. It therefore does not provide real per-ledger rollout or resource control.

I would expose an audited, replicated client configuration such as:

```text
EnableHistoricalBalances(ledger)
DisableHistoricalBalances(ledger)
GetHistoricalBalancesStatus(ledger)
```

with a lifecycle such as:

```text
DISABLED → BUILDING → READY
                    ↘ ERROR
```

Enabling the projection during `CreateLedger` can avoid a backfill. Enabling it later should trigger a background replay and remain fail-closed until it reaches `READY`.

Only physical resource settings should remain operator-controlled: local directory, memory budget, compaction concurrency and similar implementation limits.

## 3. Use the ledger name

The documentation justifies the numeric ledger ID through same-name ledger recreation, but that lifecycle is not supported.

`processCreateLedger` returns `ErrLedgerDeleted` when a tombstoned ledger name is reused, and the E2E suite explicitly verifies that recreating a deleted ledger is rejected. Restore code also documents that recreation is impossible.

The projection therefore does not need an incarnation abstraction to distinguish several lifetimes of the same name.

Using the ledger name would:

- align with the index builder;
- remove `activeByName`, `seenIDs` and `IncarnationState`;
- remove name-to-ID resolution from the read path;
- remove the ledger ID from view tokens;
- avoid persisting growing incarnation state in every manifest.

A numeric ID should only be retained if benchmarks demonstrate a meaningful improvement. Key-length arithmetic alone is insufficient because Pebble applies prefix compression to adjacent keys.

## 4. Store only per-account history

The initial projection should use one monetary identity:

```text
ledgerName
+ temporality
+ account
+ asset base
+ precision
+ color
+ timestamp
```

The current ledger-level asset scope is a denormalized rollup used only to accelerate unfiltered aggregate queries.

It causes every monetary effect to be materialized across:

```text
2 temporalities × 2 aggregation levels
```

That permanently increases ingestion, storage and compaction costs.

For the initial implementation, keep only:

```text
2 temporalities × 1 per-account representation
```

Unfiltered queries can stream the per-account histories and aggregate them at read time. A ledger-wide rollup can be introduced later as a secondary index if measured query SLOs require it.

The storage model consequently no longer needs a `scope` or `aggregation level` discriminator.

## 5. Rename `axis` to `temporality`

`axis` is too abstract for what is simply the choice of timestamp used by the historical query.

The two values are:

```text
EFFECTIVE
    Transaction.timestamp

INSERTION
    Transaction.inserted_at
```

`Temporality` communicates this intent better. `TimeBasis` would also be precise, although `Temporality` fits the API well:

```yaml
temporality: EFFECTIVE
```

## 6. Clarify the immutable segment model

A balance-history “run” is an application-level immutable segment, not a Pebble SSTable.

This distinction should be explicit because the store has two independent compaction mechanisms:

```text
balance-history compaction
    merges logical historical segments

Pebble compaction
    reorganizes physical SSTables
```

Naming the application object `ImmutableHistorySegment` or `DeltaSegment` would avoid confusion with LSM terminology.

Backdated effects should continue to be appended to a new segment and incorporated into older time ranges during logical compaction. That is the central useful property of this design.

## 7. Remove projection-specific checksums for now

Integrity certification is a separate design topic and should not be bundled into the first historical-balance implementation.

For this iteration, remove the checksums and digests introduced specifically for balance history:

- `RunRef.Checksum`;
- `Manifest.Digest`;
- `LogicalDigest`;
- `SemanticDigest`;
- the verifier/certification scratch-store machinery associated with them.

This does not mean removing:

- Pebble’s physical integrity checks;
- the existing authoritative audit hash chain.

The projection is rebuildable. If Pebble reports corruption or the projection becomes structurally unusable, the recovery model should be:

```text
fail closed
→ drop local projection
→ rebuild from the authoritative audit
→ become READY
```

A full semantic certification mechanism can be designed separately with an explicit threat/failure model.

This also avoids the current expensive verification behavior, including full local-run scans during startup and repeated reads before compaction or tiering.

## 8. Remove projection cold tiering

The cold tier of the projection should also be deferred.

The audit and ledger logs are already the authoritative rebuild source, including archived chapters through the hot+cold source. Persisting balance-history segments in cold storage creates a second derived dataset without adding a new correctness guarantee.

It introduces a large subsystem:

- archive codec;
- content addressing;
- archive cache;
- hot/cold mixed reads;
- tier manager;
- leases;
- archive bindings;
- remote GC;
- additional flags, metrics and maintenance workers;
- potentially one cold copy per replica.

If the authoritative source is unavailable, the projection must fail closed even when derived archived segments exist. The projection archive therefore only optimizes disk usage and rebuild latency.

Start with a local Pebble projection and add cold tiering later if measurements demonstrate that local capacity or rebuild SLOs require it.

## 9. Prefer commit notification with coalescing

The index builder is not purely polling-based: it combines commit notifications with a fallback ticker.

Balance history currently polls every 200 ms. The stated need to limit immutable-segment creation is valid, but it does not require pure polling.

A better model would be:

```text
AuditCommitted notification
    → coalesced dirty state
    → rate-limited publication

fallback ticker
    → recovery from missed notifications
```

The notification should reflect audit progress rather than only `LogCommitted`, because failed or non-monetary proposals can advance audit coverage without producing a ledger log.

This preserves the publication-rate limit while avoiding repeated primary-store head reads when the system is idle.

## 10. Define deleted-ledger cleanup

A deleted ledger cannot be queried and cannot be recreated under the same name, but its balance-history effects are currently retained and recopied by compaction.

This leaves permanently stored data that is inaccessible through the API.

The initial design should define explicit behavior. With a per-ledger projection, the simplest policy is likely to drop the ledger’s local historical projection on `DeleteLedger`.

## Suggested initial architecture

```text
Authoritative audit, hot or archived
                │
                ▼
Per-ledger background builder
                │
                ▼
Local rebuildable Pebble projection
                │
                ▼
Per-account historical series
    ├── EFFECTIVE temporality
    └── INSERTION temporality
```

No initial:

```text
general PIT abstraction
ledger incarnation model
ledger-wide rollup
projection checksums/certification
projection cold storage
remote GC
archive cache
```

This substantially reduces implementation and operational complexity while preserving the core feature: append-first historical balances with correct support for backdated transactions and both effective and insertion temporalities.

The deferred mechanisms can then be introduced independently, based on measured requirements:

- ledger-wide rollup for unfiltered-query SLOs;
- semantic certification for a defined integrity model;
- cold tiering for capacity or rebuild-time constraints;
- broader PIT semantics for transactions and metadata.

PR #1663, opened 2026-08-01
docs(openapi): document bigint response header

### LINKED ISSUE #1320: Document Formance-Bigint-As-String header in OpenAPI spec
## Context

The v2 API supports a `Formance-Bigint-As-String` request header that causes all bigint values in responses to be serialized as strings instead of JSON numbers. This is useful for clients that cannot handle large numbers natively (e.g. JavaScript).

The header is implemented in `internal/api/v2/views.go` but is **not documented in the OpenAPI spec** (`openapi/v2.yaml`, `openapi/v3.yaml`), making it invisible to SDK generators and API consumers.

## Accepted values

`true`, `yes`, `y`, or `1` (case-insensitive).

## Proposal

1. Add a reusable parameter in `components/parameters`:

```yaml
bigintAsString:
  name: Formance-Bigint-As-String
  in: header
  description: |
    When set to `true`, all bigint values in the response will be serialized
    as strings instead of JSON numbers.
    Useful for clients that cannot handle large numbers natively (e.g. JavaScript).
  required: false
  schema:
    type: boolean
    default: false
```

2. Reference it (`$ref: "#/components/parameters/bigintAsString"`) on all affected endpoints:
   - Transactions: list, read, create, revert
   - Accounts: list, read
   - Volumes
   - Balances
   - Logs: list
   - Workers run (queries)

3. Apply to both v2 and v3 specs.

shivanshgupta365:
## Summary

- document the Formance-Bigint-As-String response header as a reusable OpenAPI parameter
- reference it from every v2/v3 operation whose response renderer supports bigint string serialization
- regenerate the combined OpenAPI specification and API reference

## Validation

- parsed openapi/v2.yaml, openapi/v3.yaml, and openapi.yaml as YAML
- verified exactly 10 affected operations in v2 and 9 in v3 reference the reusable parameter
- bundled both versioned specifications with Redocly
- git diff --check

Closes #1320


flemzord:
Thank you for your contribution @shivanshgupta365

flemzord (approved):

PR #1664, opened 2026-08-03
fix(dal): retry transient Pebble lock opens

### LINKED ISSUE #1522: Pebble Open fails with EAGAIN when a hard-killed pod restarts before file lock is released
## Symptom

Under chaotic restart flows (`singleton_driver_rolling_restart`, `singleton_driver_quorum_recovery`), the newly-spawned ledger container fails at boot with:

```
Error: could not build arguments for function "internal/bootstrap".Module.func48
(internal/bootstrap/module.go:818): failed to build *dal.Store: received non-nil
error from function "internal/bootstrap".Module.func3 (internal/bootstrap/module.go:239):
opening pebble database: resource temporarily unavailable
```

The container exits with code **1**, the StatefulSet respawns it, the same failure repeats.

## Evidence

Antithesis k8s run `9a21d38769e6e03ebeba8cabc7223612-57-4` (2h, chaos coverage forced via `INCLUDED_DRIVERS=main/first_default_ledger,main/singleton_driver_rolling_restart,main/singleton_driver_quorum_recovery,main/singleton_driver_chapter_close,main/parallel_driver_bulk`) reports **55 counterexamples** on the property `container: ledger, namespace: default , exit code: 1`.

First occurrence traced to node-3 at vtime `238.29515` (see log window `--begin-vtime 237 ... vtime 238.2985`):

```
238.22167  INFO   Memory estimate: pebbleCache=64MiB ... total=790MiB
238.23242  INFO   Starting application
238.27816  INFO   Raft cluster stopped successfully              ← already tearing down
238.29515  ERROR  ... opening pebble database: resource temporarily unavailable
238.29654  ERROR  Usage: ledger run [flags] ...                  ← fx bootstrap failed, cobra dumped help
                                                                    → os.Exit(1)
```

The window between "Starting application" and the Pebble error is ~63 ms — well within the kernel-side asynchronous FD-close latency after SIGKILL on the previous pod incarnation.

## Root cause hypothesis

1. `singleton_driver_rolling_restart` (or `_quorum_recovery`) calls `kubectl delete pod --grace-period=0` → SIGKILL.
2. The old process dies immediately, but the kernel releases the pod's held file descriptors — including Pebble's `LOCK` file on the mounted volume — **asynchronously**.
3. The StatefulSet immediately recreates the pod, kubelet starts the new container.
4. New process calls `dal.NewStore(cfg.DataDir, ...)` → `pebble.Open(...)` → `flock(LOCK, LOCK_EX|LOCK_NB)` → **EAGAIN** because the kernel still holds the reference from the killed process.
5. `dal.NewStore` returns the error unchanged, fx bootstrap fails, `os.Exit(1)`.
6. StatefulSet respawns → same race → loop until the kernel finishes cleanup (typically 100ms–seconds), then the next boot succeeds.

Reproduction is chaos-timing dependent, but observably deterministic under `INCLUDED_DRIVERS=…rolling_restart,quorum_recovery`.

## Not caused by PR #1498

The async-storage-writes refactor does not touch `dal.NewStore` or the Pebble open path. This bug is exposed by chaos coverage that the composer usually doesn't schedule, not by the PR. Confirmed by inspecting the async-storage code paths in the run's log window — no hangs, no invariant violations, no failed antithesis assertions traceable to the applier / raft plumbing.

## Fix options

**A. Retry with bounded backoff inside `dal.NewStore`** (~30 lines)

On `pebble.Open` returning an EAGAIN-shaped error (`errors.Is(err, syscall.EAGAIN)` or `strings.Contains(err.Error(), "resource temporarily unavailable")` — Pebble wraps the raw errno), retry N times with exponential backoff (e.g. `100ms, 200ms, 400ms, 800ms, cap 5s, total budget 30s`). Log the retry count. If the total budget is exceeded, return the last error unchanged.

- Pros: self-contained, no infra dependency, safe (only retries on EAGAIN — real errors still fail fast).
- Cons: masks a symptom rather than fixing the cause, adds boot latency in the pathological case.

**B. Init container that acquires `flock LOCK` before starting** (infra-side)

An init container attempts to `flock` the `LOCK` file on the data volume, blocks until acquired (or times out), then releases and exits. The main container then boots into `dal.NewStore` when the lock is guaranteed free.

- Pros: fixes the race at the lifecycle layer.
- Cons: couples with the operator, requires a new init container in the ledger StatefulSet spec.

**Recommendation**: **A** first as a quick unblock, **B** as a longer-term operator-side change if we want to eliminate the retry latency entirely.

## Repro

```bash
git checkout feat/antithesis-included-drivers
export DOCKER_DEFAULT_PLATFORM=linux/amd64
cd tests/antithesis
just tag=repro-pebble-eagain k8s-run 2 "Pebble EAGAIN repro on chaos" \
    "main/first_default_ledger,main/singleton_driver_rolling_restart,main/singleton_driver_quorum_recovery,main/parallel_driver_bulk"
```

Expect `container: ledger ... exit code: 1` counterexamples > 0 in the run report.

## References

- Failing code path: `internal/bootstrap/module.go:239` (the `dal.NewStore` call inside `Module.func3`).
- Consumer that fails to build: `internal/bootstrap/module.go:818` (`Module.func48`).
- Antithesis run: `9a21d38769e6e03ebeba8cabc7223612-57-4` (config `antithesis-config-ledger-v3-k8s:async-cov-chaos`, description "PR#1498 chaos coverage").

shivanshgupta365:
## Summary

- Retry Pebble startup only when the wrapped error is syscall.EAGAIN.
- Use ten capped exponential-backoff waits from 100ms to 5s, for a 26.3s total retry budget.
- Preserve immediate failure for permission, corruption, configuration, and every other non-lock error.
- Cover transient success, non-retryable failure, retry exhaustion, and real cross-process Pebble lock contention.

Fixes #1522

## Validation

- go test -race ./internal/storage/dal -count=1 -timeout 10m
- go vet ./internal/storage/dal
- go build ./...
- go mod tidy in the root and operator modules
- golangci-lint 2.12.2 on ./internal/storage/dal/... with all project build tags: 0 issues

### DIFF
--- a/internal/storage/dal/store.go
+++ b/internal/storage/dal/store.go
@@ -11,6 +11,7 @@ import (
 	"runtime"
 	"strconv"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/cockroachdb/pebble/v2"
@@ -28,6 +29,51 @@ import (
 // ErrStoreClosed is returned when a store operation is attempted after the
 // Pebble database has been closed. This prevents panics during shutdown races.
 var ErrStoreClosed = errors.New("store closed")
+
+const (
+	pebbleOpenMaxRetries        = 10
+	pebbleOpenInitialRetryDelay = 100 * time.Millisecond
+	pebbleOpenMaxRetryDelay     = 5 * time.Second
+)
+
+type pebbleOpenFunc func(string, *pebble.Options) (*pebble.DB, error)
+
+func isRetryablePebbleOpenError(err error) bool {
+	return errors.Is(err, syscall.EAGAIN)
+}
+
+// openPebbleWithRetry absorbs the short lock-handoff window after a hard-killed
+// process exits. Ten capped waits total 26.3 seconds; every error except EAGAIN
+// still fails on the first attempt.
+func openPebbleWithRetry(
+	path string,
+	opts *pebble.Options,
+	logger logging.Logger,
+	open pebbleOpenFunc,
+	sleep func(time.Duration),
+) (*pebble.DB, error) {
+	delay := pebbleOpenInitialRetryDelay
+
+	for retry := 0; ; retry++ {
+		db, err := open(path, opts)
+		if err == nil {
+			return db, nil
+		}
+
+		if !isRetryablePebbleOpenError(err) || retry >= pebbleOpenMaxRetries {
+			return nil, err
+		}
+
+		logger.WithFields(map[string]any{
+			"retry": retry + 1,
+			"delay": delay.String(),
+			"error": err,
+		}).Infof("Pebble database lock is temporarily unavailable; retrying open")
+		sleep(delay)
+
+		delay = min(delay*2, pebbleOpenMaxRetryDelay)
+	}
+}
 
 const (
 	liveDir = "live"
@@ -504,7 +550,7 @@ func NewStore(
 
 	openStart := time.Now()
 
-	db, err = pebble.Open(liveDir, opts)
+	db, err = openPebbleWithRetry(liveDir, opts, logger, pebble.Open, time.Sleep)
 	if err != nil {
 		return nil, fmt.Errorf("opening pebble database: %w", err)
 	}
--- a/internal/storage/dal/store_open_retry_test.go
+++ b/internal/storage/dal/store_open_retry_test.go
@@ -1,0 +1,137 @@
+package dal
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"syscall"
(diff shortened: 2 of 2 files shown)


flemzord:
Revalidated live head `86d568d2`. The startup retry is restricted to wrapped EAGAIN lock-contention errors, remains bounded, and preserves immediate failure for every other Pebble open error. The focused tests cover transient success, exhaustion, non-retryable errors, and real cross-process lock contention; the available CI and review checks are green. I found no correctness/runtime blocker on this head. GitHub approval is intentionally disabled for this cron task.

PR #1666, opened 2026-08-04
fix(EN-1526): verify audit-only histories and compare logs bijectively against the audit chain

paul-nicolas:
[EN-1526](https://formance-team.atlassian.net/browse/EN-1526)

## Problem

Two independent defects let the store checker report a healthy store while proving nothing about the `Log` projection. Invariant #8 lists `Log` as a projection that must be verified, but no pass verified it.

**(a) Audit authentication was gated on a projection.** `Check()` read `query.ReadLastSequence` — a read of `SubColdLog` — and returned success immediately when it was 0, *before any verification ran*. A history of only rejected proposals, a fully-archived history, or a store whose log rows had simply been deleted all reported clean with nothing checked at all.

**(b) The only log-set check was self-referential.** It walked the log stream comparing each sequence to `seq + 1` from the previous row. It could report a hole between two rows while proving nothing about either row, and could not detect tail truncation or an injected row at all.

## Change

`compareLogs` (new, `internal/application/check/log_bijection.go`) compares the persisted `SubColdLog` key set against the log-sequence topology the audit hash chain authenticates, **both directions**:

- authenticated but absent -> `SEQUENCE_GAP`
- stored inside an ARCHIVED chapter's purged range -> `LOG_PURGE_RESIDUE`
- stored but unauthenticated -> `LOG_UNAUDITED`

Nothing new needed hash-binding: the chain already binds `AuditSuccess.[min,max]_log_sequence` and `AuditItem.log_sequence`, and `query.ReadAuditItems` folds every row in an entry's prefix into the hash, so item mutation/deletion/injection already yields `HASH_MISMATCH`. The `[min,max]` ranges are contiguous and adjacent across successive successes, so a healthy store collapses to a single interval — memory is O(anomalies), not O(logs).

Supporting changes:

- **`verifyAuditStructure` + `scanOrphanAuditItems`** (`AUDIT_STRUCTURE_INVALID`) — per-entry internal consistency, plus `AuditItem` groups with no entry. The latter is the one structural gap the chain genuinely does not cover, since the chain walk iterates *entries*; it is bounded to audit sequences above the archived boundary because `executePurge` deliberately leaves `SubColdAuditItem` in place.
- **A third `CheckStoreEvent` oneof arm**, `CheckStoreUnverifiableRange`, for ranges the checker could not authenticate. Deliberately not on the error arm: "cannot prove" is not "diverges". It exists so the absence of a finding in an unauthenticated range is never mistaken for proof.
- **Asymmetric handling of that arm.** `store check` warns and leaves the exit code alone; `restore validate` **blocks**. `store check` is diagnostic on a cluster you already run; `restore validate` gates adopting foreign data. Two consumers were silently dropping the arm entirely (`pkg/actions`, and `server_restore.go` whose switch had no `default`) — the restore one on the highest-stakes path there is.
- `LOG_PURGE_RESIDUE` takes precedence over `LOG_UNAUDITED` when both apply. That case is guaranteed, not hypothetical: archiving purges a chapter's audit entries and log rows in the same Pebble batch, so a survivor in a purged range is necessarily also unauthenticated. Reporting it as an injection would misdiagnose a purge-lifecycle failure.

## Notes for review

- **The old `SEQUENCE_GAP` test was rewritten, not weakened.** `TestCheckerDetectsSequenceGap` asserted the deleted self-referential scan's output: a hole at sequence 2 in a fixture with *no audit entries*, where sequences 1 and 3 are themselves unauthenticated. It is now `TestCheckerDetectsUnauditedLogs`, asserting both stored rows report `LOG_UNAUDITED` — strictly stronger. `SEQUENCE_GAP` remains reachable via the reverse direction and is covered by the new integration tests.
- **`compareLogs` bounds its own enumeration** (`maxEnumeratedGapSequences`). The reverse pass derives missing ranges by interval arithmetic rather than walking each sequence, because the range's upper edge is a store-supplied `max_log_sequence`: chain-verified, but forgeable by anyone who can write Pebble and knows the cluster ID, since the chain is keyed by cluster ID rather than a secret. Unbounded, a crafted value made `Check()` hang instead of report — a worse outcome than reporting, on the tool meant to catch that adversary. The bound is a constant, deliberately *not* derived from `NextLogId` or the highest stored sequence: that would gate verification on a projection, which is the defect this PR removes.
- **`audit-chain.md` carried a false claim that this PR corrects.** Its companion-streams table said a Failure outcome writes **0** `AuditItem` rows. A failure writes **N**, one per submitted order, each with `LogSequence = 0` (`buildAuditItems` runs in the closure shared by both outcomes and is sized on the full order set; the failure path only passes `logs = nil`). Designing from that row produced a check that would have flagged every rejected proposal in a production cluster. The correction is load-bearing and is now also stated in invariant #8.

## Rebase note

Rebased onto `d02eaccca`. Conflicts with EN-1458 and #1636 were all additive; both sides kept. Two worth flagging:

- `checker.go`: EN-1458 had added a `compareReverseMapOrphans` call *inside* the early-return block this PR deletes. Deleting it is still correct — a zero-log store now reaches the full-scope call at the end of `Check()`, and that pass gates verdicts on `indexedSequence == lastSequence`, which holds since both are 0. Their intent is recorded in a comment so an early return is not reintroduced. Their tests pass.
- The rebase merged cleanly but did not *compile*: `NewChecker` gained a seventh `*readstore.Store` parameter on the base branch, incompatible with a test helper here in a way no textual merge could detect (`bc3ff848b`).

## Testing

`just pre-commit` (0 lint issues, both modules) and `just test` (exit 0) both green on the rebased tree.

**Coverage gaps, stated explicitly:** every test here is a unit or in-process integration test against a Pebble store. `cmd/ledgerctl/store` and `cmd/ledgerctl/restore` have no test files at all, so the CLI changes and the restore-validation *blocking* behaviour are verified by construction rather than by a test that stages a corrupt backup and observes the refusal. The `e2e`/`scenario` suites are tagged builds requiring Docker and were not run.

## Follow-ups (not in scope)

- `server_restore.go` maps `CheckStoreError` to `ValidateRestoreError{Message}` only, discarding `error_type` — restore validation cannot distinguish a `HASH_MISMATCH` from a `VOLUME_MISMATCH`. Pre-existing.
- Log *payload* verification (reconstructing canonical payloads by re-running business processing) is EN-1531. This PR verifies the sequence *set*, not payload bytes.

[EN-1526]: https://formance-team.atlassian.net/browse/EN-1526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/application/check/log_bijection.go
```diff
@@ -0,0 +1,721 @@
...
+		key := iter.Key()
+
+		if len(key) < 10 {
+			return fmt.Errorf("invariant: log key %x is shorter than the 10-byte prefix", key)
+		}
+
+		seq := binary.BigEndian.Uint64(key[2:10])
+		stored.add(seq, seq)
+
+		if unverifiable.contains(seq) {
```
NumaryBot: 🟠 **[major] Purge-residue check suppressed by early unverifiable-span skip**

When an archived chapter's surviving boundary audit entry is structurally malformed, its log span is added to `unverifiable` and the loop continues early. If a log row also survived inside that archived chapter's purged range, the early `continue` prevents the `LOG_PURGE_RESIDUE` error from ever being emitted. A purged range should admit no legitimate stored log regardless of whether the audit span is provable, so hiding this failure is a correctness bug in the integrity checker.

**Suggestion:** Move the purged-range residue check to run before the unverifiable-span `continue`. This ensures that any surviving log rows inside a purged range are always reported, independently of whether the spanning audit entry is verifiable.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1666#issuecomment-5176463488

internal/application/check/log_bijection.go
```diff
@@ -0,0 +1,721 @@
...
+			return fmt.Errorf("invariant: audit item key %x is shorter than the 10-byte prefix", key)
+		}
+
+		auditSeq := binary.BigEndian.Uint64(key[2:10])
+
+		if auditSeq <= v.archiveEndAuditSeq {
+			continue
+		}
+
+		if _, ok := v.sequences[auditSeq]; ok {
```
NumaryBot, line 500: 🟠 **[major] Unverified audit entries misclassified as orphans after a chain break**

If audit hash verification stops at sequence N (due to a hash mismatch or an un-hashable entry), `v.sequences` only contains entries that were verified before N. All AuditItem groups for entries at N and beyond are then reported as 'no audit entry exists' (`AUDIT_STRUCTURE_INVALID`), producing a flood of false-positive orphan errors on any store with one early chain break. The check conflates stored AuditEntry existence with hash-verified membership in the chain.

**Suggestion:** Separate the concept of 'AuditEntry exists in storage' from 'AuditEntry is hash-verified'. The orphan check should only fire when no AuditEntry exists in storage for the item, not when an entry exists but was simply not reached by the hash-verification walk. Entries that exist but are unverified should be reported under a distinct diagnostic (e.g., `AUDIT_UNVERIFIED`) rather than the orphan path.

internal/application/check/log_bijection.go
paul-nicolas: Fixed in 25e6126 — the purged-range check now runs first on the stored side, ahead of both the unverifiable and the authenticated skip.

Confirmed reachable, and it is the archive boundary itself: `executePurge` deletes logs up to `close_sequence` while the audit entry that produced them survives at `close_audit_sequence + 1`, so when that survivor is structurally malformed `verifyAuditStructure`'s `blank()` adds `[min_log_sequence, max_log_sequence]` — reaching into the purged range — to `v.unverifiable`. A residue row there reported nothing at all.

Your justification is also what the file already claimed: the doc comment at log_bijection.go:549 said the purged check "runs FIRST and wins over BOTH other stored-side outcomes", while the code tested `unverifiable` first. The precedence holds because `buildPurgedLogSet` derives the purged set from the `Chapter` rows, never from an audit entry, so audit verifiability is orthogonal to it. Doc comment, the stale "skipped in both directions" summary and the AGENTS.md invariant #8 wording are all updated to name all three losers.

`TestCompareLogsArchiveBoundaryAsymmetry` gains the purged-inside-unverifiable case plus its neighbour (unverifiable but not purged must still report nothing, so the reorder does not over-report). Both were verified to fail before the fix.

internal/application/check/log_bijection.go
paul-nicolas: Fixed in 25e6126 — the orphan criterion is now the physical absence of the `SubColdAudit` key, probed by a point `Get` (`auditEntryExists`), not absence from `v.sequences`.

Confirmed: `verifyAuditHashChain` returns early on all three break shapes — embedded `entry.items` (checker.go:2433), an un-rehashable header (checker.go:2463), a hash mismatch (checker.go:2498) — each deliberately returning the partial maps so `Check()` keeps surfacing other projections. `verification.sequences` is written only after the hash comparison passes (checker.go:2503), so on a break at N it holds `[boundary+1, N)` and the entries above are on disk yet unlisted. The message was wrong, not just noisy.

`v.sequences` survives as a no-I/O fast path only — membership proves existence, so a walk that ran to completion admits every sequence and the scan does zero extra reads.

For your "distinct diagnostic": rather than a new error type I used the `unverifiable_range` arm this PR introduced, with a new `CHECK_STORE_UNVERIFIABLE_REASON_UNVERIFIED_AUDIT_ENTRY`. It is off the error arm, so it cannot affect the exit code or the `valid` flag, which is right for "cannot prove" rather than "diverges". Emitted **once** with the count and `[lo,hi]` audit-sequence span — per-sequence emission would only relocate the flood. `range_start`/`range_end` are 0/0 since those entries were never parsed for their log spans; the CLI already renders that as `logs=undetermined`.

Added `TestScanOrphanAuditItems` (7 subtests) — the pass had no coverage at all, which is why this escaped. It pins existing-but-unverified, genuinely-absent, the collapsed tail, the mixed case, the fast path, the archive-boundary carve-out and the empty store; the first three fail before the fix.

One adjacent note, not fixed here: `verification.authenticated` is partial for the same reason, so after a break `compareLogs` reports every log above N as `LOG_UNAUDITED`. That claim is not strictly false (no *verified* entry authenticates them) but it is the same misdiagnosis shape. The coherent fix is for the break path to add `[breaking entry's min_log_sequence, MaxUint64]` to `verification.unverifiable` and declare it once. Kept out of this PR deliberately — it changes behaviour on a path neither finding covers and interacts with the precedence change in the sibling thread — and flagged for a follow-up ticket.

paul-nicolas:
Both findings accepted and fixed in 25e6126; details in the two inline threads, now resolved.

Common root cause: a derived set read as authoritative when it is only partial. `verifyAuditHashChain` aborts at the first chain break and returns its partially populated maps on purpose, so `Check()` can keep surfacing other projections — downstream passes must treat those maps as incomplete.

1. **Purge-residue precedence** — the purged-range test now runs first on the stored side, ahead of both the unverifiable and the authenticated skip. Both overlaps trace to the same surviving boundary entry, so testing either skip first re-opened the identical hole. The purged set is derived from the `Chapter` rows, never from an audit entry, so audit verifiability cannot weaken it. This also brings the code in line with the doc comment, which already claimed the purged check ran first.
2. **Orphan vs unverified** — the orphan criterion is now the physical absence of the `SubColdAudit` key. Entries that exist above a break are declared once as the new `CHECK_STORE_UNVERIFIABLE_REASON_UNVERIFIED_AUDIT_ENTRY` on the non-error `unverifiable_range` arm, rather than reported as orphans.

Tests: new `TestScanOrphanAuditItems` (the pass had no coverage, hence the escape) plus two subtests on `TestCompareLogsArchiveBoundaryAsymmetry`. Each regression was verified to fail before the fix. `AGENTS.md` invariant #8 updated for both rules.

**Follow-up, deliberately out of scope:** `verification.authenticated` is partial for the same reason, so after a break `compareLogs` reports every log above it as `LOG_UNAUDITED` — one `HASH_MISMATCH` becoming a flood of injection reports. Fix would be to add `[breaking entry's min_log_sequence, MaxUint64]` to `verification.unverifiable` at the break and declare it once. Left out because it changes behaviour on a path neither finding covers and interacts with the precedence change above.

internal/application/check/log_bijection.go
```diff
@@ -0,0 +1,825 @@
...
+				0, 0,
+			))
+
+			continue
+		}
+
+		purged.add(start, closeSeq)
+	}
+
+	return purged
```
NumaryBot, line 607: 🟠 **[major] Purge-residue check suppressed by early unverifiable-span skip**

When an archived chapter's surviving boundary audit entry is structurally malformed, its log span is added to `unverifiable` and the loop continues early. If a log row survived inside that archived chapter's purged range, the early `continue` prevents `LOG_PURGE_RESIDUE` from ever being emitted. A purged range should admit no legitimate stored log regardless of whether the audit span is provable, so hiding this failure is a correctness bug in the integrity checker.

**Suggestion:** Move the purged-range residue check to run before the unverifiable-span `continue`, so that any surviving log rows inside a purged range are always reported independently of whether the spanning audit entry is verifiable.

internal/application/check/log_bijection.go
```diff
@@ -0,0 +1,825 @@
...
+			return fmt.Errorf("invariant: audit item key %x is shorter than the 10-byte prefix", key)
+		}
+
+		auditSeq := binary.BigEndian.Uint64(key[2:10])
+
+		if auditSeq <= v.archiveEndAuditSeq {
+			continue
+		}
+
+		if _, ok := v.sequences[auditSeq]; ok {
```
NumaryBot, line 500: 🟠 **[major] Unverified audit entries misclassified as orphans after a chain break**

If audit hash verification stops at sequence N (due to a hash mismatch or an un-hashable entry), `v.sequences` only contains entries verified before N. All AuditItem groups for entries at N and beyond are then reported as 'no audit entry exists' (`AUDIT_STRUCTURE_INVALID`), producing a flood of false-positive orphan errors on any store with one early chain break. The check conflates stored AuditEntry existence with hash-verified membership in the chain.

**Suggestion:** Separate the concept of 'AuditEntry exists in storage' from 'AuditEntry is hash-verified'. The orphan check should only fire when no AuditEntry exists in storage for the item, not when an entry exists but was simply not reached by the hash-verification walk. Entries that exist but are unverified should be reported under a distinct diagnostic (e.g., `AUDIT_UNVERIFIED`) rather than the orphan path.

internal/application/check/log_bijection.go
```diff
@@ -0,0 +1,825 @@
...
+		key := iter.Key()
+
+		// Key: [ZoneCold(1)][SubColdAuditItem(1)][audit_seq BE 8][order_idx BE 4]
+		if len(key) < 10 {
+			return fmt.Errorf("invariant: audit item key %x is shorter than the 10-byte prefix", key)
+		}
+
+		auditSeq := binary.BigEndian.Uint64(key[2:10])
+
+		if auditSeq <= v.archiveEndAuditSeq {
```
NumaryBot, line 496: 🟠 **[major] Orphan-skip uses a single floor instead of the union of archived audit ranges**

The orphan-skip guard uses `archiveEndAuditSeq` (the maximum close audit sequence across all archived chapters) as a single floor value. When archived chapters are non-contiguous or complete out of order (e.g., chapter 3 is ARCHIVED while chapter 2 is still ARCHIVING), any AuditItem at a chapter-2 audit sequence whose AuditEntry was deleted is silently ignored, even though `executePurge` never legitimised that orphan. This can cause real orphan entries to go unreported.

**Suggestion:** Replace the single-floor guard with a check against the union of actual archived audit-sequence ranges, so that only AuditItems whose sequences genuinely fall within a completed archived chapter are skipped.

internal/application/check/log_bijection.go
```diff
@@ -0,0 +1,825 @@
...
+			report("log sequence %d is claimed by two audit items", logSeq)
+			blank()
+
+			return false
+		}
+
+		seenLogSeq[logSeq] = struct{}{}
+	}
+
+	v.authenticated.add(minLog, maxLog)
```
NumaryBot, line 401: 🟠 **[major] Success-range authentication does not verify that every slot is claimed by an AuditItem**

For a hash-verified success entry whose `[min, max]` range contains a sequence that no non-legacy AuditItem claims (e.g., items claim logs 10 and 12 while the success range covers 10–12), the whole interval is added to `v.authenticated`. A stored log at the unclaimed sequence is then treated by `compareLogs` as fully audited, even though no serialised position in the hash chain corresponds to that log. This allows a malformed audit entry to hide an unaudited row.

**Suggestion:** Before adding a range to `v.authenticated`, verify that the claimed in-range log sequences from AuditItems cover every slot in the success range. Any gap should be reported as an integrity violation rather than treated as authenticated.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1666#issuecomment-5176463488

internal/application/check/log_bijection.go
paul-nicolas: Already fixed in 25e6126, which is the commit this review round ran against — this is a re-post of finding `b013902ba9a701b4` from the previous round (thread now resolved) under a new finding hash.

The anchor has drifted: line 607 of the current file is inside `buildPurgedLogSet`, not the `compareLogs` stored-side loop. That loop now reads:

- log_bijection.go:717 — `if purged.contains(seq)` → `LOG_PURGE_RESIDUE`
- log_bijection.go:727 — `if unverifiable.contains(seq)` → skip
- log_bijection.go:731 — `if authenticated.contains(seq)` → skip

so the residue check runs first, ahead of both skips, exactly as suggested. `TestCompareLogsArchiveBoundaryAsymmetry/a_purged_row_is_residue_even_inside_an_unverifiable_span` pins it, and it fails if the order is restored.

internal/application/check/log_bijection.go
paul-nicolas: Already fixed in 25e6126, the commit this round ran against — a re-post of finding `5afa15ebc2587693` from the previous round (thread now resolved) under a new hash.

`scanOrphanAuditItems` no longer keys the orphan verdict on `v.sequences`. It now probes storage directly:

- log_bijection.go:501 — archive-boundary guard
- log_bijection.go:505 — `v.sequences` membership, now only a **no-I/O fast path** (membership proves existence; absence proves nothing)
- log_bijection.go:510 — `auditEntryExists(reader, auditSeq)`, a point `Get` on the `SubColdAudit` key
- exists → tallied, declared once as `CHECK_STORE_UNVERIFIABLE_REASON_UNVERIFIED_AUDIT_ENTRY` (the distinct diagnostic you asked for, on the non-error `unverifiable_range` arm so it cannot flip the exit code)
- absent → `AUDIT_STRUCTURE_INVALID`, unchanged

`TestScanOrphanAuditItems` covers all of it, including `a_whole_unverified_tail_collapses_into_one_declaration`; three subtests fail if the criterion reverts to `v.sequences`.

internal/application/check/log_bijection.go
paul-nicolas: Not a reachable state — the described entry cannot exist, so there is no gap to report.

A log sequence is allocated **only** at the moment a log is created, and the created log's sequence is written onto the item at that same order index:

- `internal/domain/processing/processor.go:356` (normal path) and `:295` (skip path): `nextSequenceID := orderScope.IncrementNextSequenceID()` immediately followed by constructing the `Log{Sequence: nextSequenceID}` and assigning `logs[i]`.
- `internal/infra/state/audit.go:57` `buildAuditItems`: `item.LogSequence = created.GetSequence()` for that same index `i`.
- `MinLogSequence`/`MaxLogSequence` are folded from exactly those allocations (processor.go:310-315, :376-381).
- `WriteSet.IncrementNextSequenceID` (write_set.go:1389) is a single global counter stepping by 1 — not per-ledger — so a multi-ledger proposal does not interleave two counters.

The orders that produce no log do **not** consume a sequence: the no-log outcome (idempotent mirror replay) deliberately skips the slot without incrementing, and processor.go:322-330 spells that out. So the created sequences in a proposal are strictly contiguous `min..max`, one per item, and every slot in `[min,max]` is claimed. Your example — items claiming 10 and 12 while the range covers 10-12 — would require a sequence to be consumed without a log being created, and no path does that.

Two further points on "this allows a malformed audit entry to hide an unaudited row": `min_log_sequence`, `max_log_sequence` and every `AuditItem.log_sequence` are all hash-bound (`state.BuildHashedHeaderPayload` / `state.BuildPerItemPayload`, and `query.ReadAuditItems` folds every row in the prefix), so a tamperer cannot manufacture that shape without breaking the chain — which `verifyAuditHashChain` reports first. That is also why `verifyAuditStructure`'s doc comment states these checks are not tamper detection: they exist only because `compareLogs` consumes the fields. And a legacy in-range REFERENCE item cannot displace a claim either — it would be a second claim on an already-claimed sequence and the existing duplicate check (log_bijection.go:388) blanks the entry.

Resolving as invalid. Happy to reopen with a concrete reachable path if you have one.

internal/application/check/log_bijection.go
paul-nicolas: Real concern, but the suggested fix applied to this guard alone would introduce false positives — so I want a decision before changing it. Leaving this thread open.

**Your premise checks out.** Multiple chapters can be ARCHIVING at once (`DispatchArchiveRequests`, machine.go:1832, dispatches every ARCHIVING chapter) and each completes when its own cold-storage upload confirms, so out-of-order completion is structurally reachable. `buildPurgedLogSet` already asserts exactly this for the log side and deliberately uses per-chapter ranges instead of a single `max(close_sequence)` floor (log_bijection.go:470-473). The audit side using a single floor is a genuine asymmetry.

**But the floor is not the root, and it is not arbitrary.** `archiveEndAuditSeq` is the *same value* as `afterAuditSeq`, which is the chain walk's starting point: `query.ReadAuditEntries(ctx, reader, afterAuditSeq)` (checker.go:2342, assigned to `archiveEndAuditSeq` at checker.go:2354). So every audit entry at or below `max(close_audit_sequence)` is *already* outside everything the checker verifies — the walk never reads it, and `v.sequences` never contains it. The orphan floor is consistent with that horizon by construction.

Consequence: narrowing only this guard to the union of archived ranges makes the scan examine chapter-2 sequences the walk never read. `v.sequences` lacks them, so `auditEntryExists` finds the still-live entries and every one becomes a `UNVERIFIED_AUDIT_ENTRY` declaration — a fresh flood on a *healthy* mid-archiving cluster, which is the same failure mode as the finding I just fixed in the sibling thread.

The coherent fix is both together: make the chain walk start from the union of archived audit ranges (verifying the live entries of a still-ARCHIVING chapter below the max) and narrow this guard to match. That is a change to `verifyAuditHashChain`'s scope, it predates this PR, and it needs its own test coverage for the concurrent-archiving topology.

@paul-nicolas — fold that into this PR, or split it into a follow-up ticket? I lean follow-up: it widens what the checker authenticates on a path this PR does not touch, and the half-fix is worse than the status quo.

PR #1667, opened 2026-08-04
fix(EN-1515): repair signing-projection restore, boot decode, and admission validation

paul-nicolas:
Signing-projection correctness for EN-1515, end to end: the three defects found while analysing the missing verification, the checker pass that closes the gap (#1669, merged into this branch), and the FSM determinism and termination bugs that pass then surfaced.

Nothing here needs an attacker to be reachable.

---

## 1. Revoked signing keys resurrect after a backup restore

`RebuildDelta` replayed `RegisterSigningKey` and `SetSigningConfig` but filed `RevokeSigningKey` under *"Log types with no persistent state to rebuild"* as an empty case. Revocation's only persistent representation is **row absence** — the stored value carries no revoked flag (`state/batch.go:130-131`) and `ReadSigningKeys` applies no filter — and `state.DeleteSigningKey` has exactly one caller repo-wide, the live FSM path (`write_set.go:652`). So the rebuild was structurally incapable of revoking.

Reachable because the export is layered: the **full checkpoint** is an unfiltered whole-DB Pebble checkpoint (`manager.go:68` → `dal/store.go:876` → `db.Checkpoint` at `:897`) carrying Global-zone rows verbatim, while **incremental exports** are `ZoneCold`-only (`manager.go:498, 515, 541, 562`). A revoke *after* the checkpoint therefore travels only as a log, and the empty case dropped it — so a key revoked *because it was compromised* came back as a trusted signer, re-listed to operators as valid.

No proto change: `RevokedSigningKeyLog` already carries `key_id` and `cascaded_key_ids` (`common.proto:383-386`), precisely because cascade children are resolved at apply time from the live keystore and unavailable during rebuild. Deletion is set-based, so the pre-EN-1521 non-canonical cascade ordering in older logs is irrelevant.

Affects both restore entry points: `server_restore_download.go:309` and `cmd/ledgerctl/store/bootstrap.go:206`.

## 2. A short signing-key row panicked the boot path

`query.ReadSigningKeys` sliced `value[:32]` with no length guard. Go bounds-checks the slice expression *before* `copy` runs, so any row under 32 bytes panicked — and this decode runs in `state.Recovery` (`recovery.go:139`), so one row crash-looped **every** replica, since the row is Raft-replicated. (`copy` alone would have truncated harmlessly; the slice expression was the bug.)

Now guarded, returning malformed rows the way `ReadReversions` already does (`recovery.go:122-132`): boot proceeds on the decodable keys and logs each skip loudly. Silently dropping one would stop authorizing legitimately signed requests, so it must be visible — but crash-looping the cluster is strictly worse. `ReadSigningConfig` needed no change; `dal.ReadBool` was already defensive.

## 3. No server-side public-key length validation

Admission copied `public_key` verbatim (`admission.go:2035`), the FSM validated only the key *ID* (`processor_signing.go:10-20`), and `SaveSigningKey` stored the bytes unchanged. The only 32-byte check was client-side in `ledgerctl` (`register_key.go:129-131`), so a raw gRPC client could create the row from (2).

Validated at **admission, not the FSM**, deliberately: admission runs pre-Raft, so a rejection never enters the log. FSM validation would change the accept/reject decision for an already-committed entry, diverging state across a mixed-binary rolling upgrade and requiring a "stop all nodes" deployment note (cf. `docs/ops/deployment.md:661`).

Note this repo's own fixture was registering `PublicKey: []byte("pk")` — a 2-byte key, i.e. exactly the malformed row from (2). Fixed the fixture to a real 32-byte value rather than weakening the validator.

## 4. Making boot survive (2) opened an unsigned-write bypass

A genuine regression this branch introduced, not a pre-existing property. `authorizeUnsignedBatch` gates the unsigned `RegisterSigningKey` bootstrap exception on `!keyStore.HasKeys()`, and with every persisted row corrupt the key store is empty — indistinguishable from a fresh cluster. Before (2), such a store crash-looped, so the path was unreachable.

Worse than a plain bypass: an attacker with store write access could always inject a well-formed row with their own key, but the new checker pass reports that as `SIGNING_KEY_MISMATCH`. Going through this exception instead means the registration takes the **normal, audited** path — the checker then sees a legitimately chain-bound key and flags only the corrupt row it replaced. The exception is the route that *launders* an injected key past the detection this ticket adds.

Recovery now records the skipped-row count (`MarkUndecodableRows`) and the gate consults `HasUndecodableRows()` alongside `HasKeys()`: the question it needs is "has this cluster ever had a signing key", which is not the question `HasKeys` answers. Fail-closed rather than failing boot — failing boot would reinstate a remote DoS that one corrupt Raft-replicated byte can trigger on every replica. The cluster boots, serves reads and unsigned writes per its config, and refuses signing management until an operator restores the rows. `Reset` clears the marker so a repaired store is not permanently locked out.

## 5. The checker pass itself (#1669, merged here as `bf12d09`)

`signingVerifier` closes the invariant-#8 gap: both signing projections were persisted, loaded straight into the runtime key store and the `requireSignatures` gate, and verified by nothing. Two new enum values, `SIGNING_KEY_MISMATCH` and `SIGNING_CONFIG_MISMATCH`, plus `SIGNING_VERIFICATION_INCOMPLETE` for partial coverage.

The design points that carry the weight:

- **Never seeded from the live projection or the baseline checkpoint.** `attributes.writeBaselineAttributes` copies the attribute zone verbatim, so seeding would verify old keys against a copy of themselves. Cold storage is the only audit-derived source for pre-boundary keys, folded oldest-chapter-first with no TTL cutoff.
- **Incomplete coverage suppresses the comparisons** rather than reporting against a prefix, which is unsound in both directions. Reporting a mismatch we cannot substantiate is worse than admitting the gap.
- **The cascade is re-derived, never read off `cascaded_key_ids`** — trusting the recorded list would let a tampered projection justify itself.
- **Both folds are bounded by the entry's fresh-log window**, not the archive boundary alone, so a legacy pre-`f9ee1e829` replay reference cannot re-apply a register after its revoke.
- **Runs on the empty-audit fast path**, because the projections are cluster-global and a zero-log store *proves* the audit registered no key.

## 6. The pass immediately surfaced an FSM determinism breach

`keystore.AddPublicKey` wrote `parents[keyID]` only for a non-empty parent, so re-registering a key as a root left the stale edge in memory — and that map is what `WriteSet.GetSigningKeyChildren` walks to build a revocation cascade. The live in-memory keystore was the only representation that kept the edge: the persisted row, recovery, the audit-driven rebuild and the checker replay all drop it.

So the cascade was **restart-dependent**. An un-restarted replica cascaded the reassigned key, a restarted one did not, and the same revoke order emitted different `cascadedKeyIds` and deleted different rows on different replicas. Invariants #1 and #2, pre-existing, and only visible once something re-derived the cascade independently.

Fixed on the keystore side rather than by aligning the replay: the replay already matches the persisted row and the documented `"" = root` semantics, and mirroring the live map would leave it agreeing with un-restarted replicas and disagreeing with restarted ones. `docs/ops/signing.md` no longer claims the parent relationship is immutable, which re-registration has never honoured.

## 7. And a cascade that never terminates

Surfaced by Azorlogh's review of this PR. `processRevokeSigningKey` walked the child relation with **no visited set**, and `RemoveSigningKey` runs only after the walk, so the graph never shrinks underneath it. Nothing validates acyclicity — registration is an upsert that shape-checks the two key IDs and never verifies the parent exists — so `register a` → `register b under a` → `register a under b` is accepted end to end, and one cascade revoke of either spins forever.

The hang is in the **Raft apply path**: it wedges every replica simultaneously, and the order being committed it replays on every restart. Three ordinary requests from any signing-management principal, permanent cluster kill.

Guarded with the same visited set `signingVerifier.descendantsOf` already carried. It also dedups `cascaded_key_ids`, since `GetSigningKeyChildren` reports every pending addition whose parent matches. Rejecting cycles at registration is deliberately **not** done here: it is a behaviour change needing its own design pass, and it could not rescue a store that already holds a cycle.

## 8. Two review fixes on the pass itself

- **A hash chain break truncated the signing fold silently.** `verifyAuditHashChain` returns early at three non-error exits so `Check()` can still report other projections; the live signing fold runs inside that loop, so the expectation became a prefix while `coldComplete` was already true — reporting every registration past the break as injected and every revocation as a lost row, on top of the `HASH_MISMATCH` that is the store's real problem. Each exit now marks the fold truncated and the suppression gate covers both causes, with the message naming which side went unread.
- **Each audit item's order was decoded twice per `Check()` run** — once by `collectExpectedSkippable`, once by the signing fold. The decoded orders are now returned and reused.

---

## 8. And a cascade that skipped a key the FSM keeps

The mirror image of (6), found by review on the finished pass. `GetSigningKeyChildren` builds its `pendingRemovals` filter over the **whole** `pendingSigningKeyUpdates` slice with no ordering awareness, so once a proposal removes a key that key is excluded from *every* cascade in the same proposal — including one evaluated after a later registration in that proposal put it back, and even when that registration pointed it straight at the key being cascade-revoked.

The replay leaned on absence from its expected key map to reproduce that filter, but absence is a point-in-time fact where the filter is a set over the proposal: the re-registration reinstates the row and the walk follows the fresh edge. So `revoke(X)` + `register(X under P)` + `cascade-revoke(P)` in ONE proposal — one multi-request `ApplyRequest` is enough, no tampering — had the FSM keep X (`Absorb` replays `[remove X, save X→P, remove P]` in slice order) while the replay cascaded X out and reported a false `SIGNING_KEY_MISMATCH` against a healthy store.

Fixed on the **checker** side, the opposite call from (6) and for a stated reason: `pendingSigningKeyUpdates` is proposal-local state rebuilt identically on every replica, so the FSM here is deterministic and replica-consistent — no integrity defect at the source, and aligning it would change the chain-hashed `cascaded_key_ids`. Two details in the fix carry weight beyond "consult a set": `proposalRevoked` is filled *after* the walk, mirroring `processRevokeSigningKey` walking the child relation before removing anything (fill it first and a revoke excludes its own targets from its own cascade), and it skips the candidate *entirely* rather than omitting it from the result, because the FSM filters the key out of what `GetSigningKeyChildren` returns and its BFS therefore never recurses through it — leaving the excluded key's own subtree alive too.

---

## Testing

Restore, boot and admission:
- `TestRebuildDelta_ReplaysRevokeSigningKey` — non-cascade, multi-level cascade, register-then-revoke
- `TestReadSigningKeysMalformedRows` — 0/1/31-byte values report instead of panicking; well-formed rows still decode
- `TestValidateOrderSigningKey` — 0/31/33 bytes rejected, 32 accepted, unrelated orders ignored
- `TestAuthorizeUnsignedBatch_BootstrapExceptionClosesOnUndecodableRows` — fresh cluster still bootstraps; undecodable rows, existing keys, and both together all reject
- `TestKeyStoreUndecodableRowsResetWithKeys` — a repaired store is not permanently locked out

Checker pass:
- Tamper matrix over both projections: injected, missing, re-parented and byte-altered rows, config flipped both ways, undecodable rows
- Cascade model vs the FSM: reparenting inside vs before the revoking proposal, re-registration as root, in-proposal edges superseded by a later registration, and no leak of a superseded edge across proposals
- Termination: parent cycles and self-parents
- Coverage: cold-storage folding, per-cause `SIGNING_VERIFICATION_INCOMPLETE` including a chain-break-truncated live fold, and the malformed-row carve-out that survives suppression
- `TestVerifyAuditHashChain_MarksSigningFoldTruncatedOnEveryBreak` — one case per early exit
- `TestCollectExpectedSkippable_ReturnsOrdersParallelToItems` — the positional decode contract

Cascade removal filter:
- `TestSigningVerifier_CascadeSkipsKeysTheProposalAlreadyRemoved` — the three re-registration targets (revoked parent, unrelated parent, root), the excluded key's surviving subtree, a key removed as a cascade descendant, the walk-then-fill ordering, and a cross-proposal leak guard

FSM:
- `TestProcessRevokeSigningKey_CascadeVisitsEachKeyOnce` — parent cycle, self-parent, three-key cycle, duplicate pending addition, and a key reachable through two branches; `cascaded_key_ids` asserted exactly, since it is hashed into the audit chain

Every fix is mutation-verified: with the fix reverted the new test fails with the expected signature, and reverting a single guard fails exactly the cases that guard covers and nothing else.

## Gates

`just pre-commit` (0 issues, both modules), `GOROOT= go build ./...`, and full `just test` all pass.

Ticket: https://formance-team.atlassian.net/browse/EN-1515


internal/infra/state/recovery.go
```diff
@@ -136,11 +136,21 @@ func (r *Recovery) RecoverState() error {
...
 		if err != nil {
 			return fmt.Errorf("loading signing keys: %w", err)
 		}
 
+		// Only SaveSigningKey produces these rows and it always writes a full
+		// public key, so a malformed one means corruption or tampering. Boot
+		// proceeds on the decodable keys — the checker reports malformed rows as
+		// integrity events — but never silently: a dropped key stops authorizing
+		// requests signed with it.
+		for _, m := range malformedSigningKeys {
```
NumaryBot, line 149: 🟠 **[major] Skipped malformed signing-key rows can expose unsigned bootstrap exception**

When recovery skips malformed signing-key rows and no decodable keys remain, keyStore.HasKeys() returns false — the same state as a truly empty, fresh store. If require_signatures is enabled, the admission layer's unsigned bootstrap exception (which allows the first key to be registered without a signature) becomes reachable after a restart, even though signing-key rows already exist in storage. An attacker or misconfigured client could exploit this window to register an unauthorized signing key.

**Suggestion:** Track whether any signing-key rows were encountered (malformed or not) separately from whether any were successfully decoded. If rows were encountered but none decoded successfully, treat the store as non-empty (e.g., return an error or set a 'corrupt' sentinel) rather than falling through to the bootstrap-exception path. At minimum, log a clear warning and do not allow the unsigned bootstrap path when corrupt rows are detected.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1667#issuecomment-5178417429

paul-nicolas:
## Fixed in 5115611 — and it is a genuine regression this branch introduced

Confirmed, and worth being explicit that this is not a pre-existing property: before this branch a short signing-key row **panicked** during `RecoverState`, so the path was unreachable — the failure mode was a crash-loop, not a bypass. Making boot survive the row is what opens it. `authorizeUnsignedBatch` (`admission.go:996`) gates the unsigned `RegisterSigningKey` exception on `!a.keyStore.HasKeys()`, and with every persisted row corrupt the key store is empty, which `HasKeys` cannot distinguish from a fresh cluster.

**The part that makes it worse than a plain bypass.** An attacker with write access to the store could always inject a well-formed row with their own public key — but the new checker pass reports that as `SIGNING_KEY_MISMATCH`, a stored key with no audited registration. Going through this exception instead means the registration takes the **normal path and is audited**: the checker then sees a legitimately chain-bound key and flags only the corrupt row it replaced. So the exception is not just an alternative route to the same outcome, it is the route that *launders* an injected key past the detection this ticket adds. That is what moves it from theoretical to worth fixing now.

**Fix.** `Recovery` records the number of rows it skipped on the key store (`MarkUndecodableRows`), and the gate consults `HasUndecodableRows()` alongside `HasKeys()`. The distinction the gate actually needs is "has this cluster ever had a signing key", which is a different question from "are there usable keys loaded" — `HasKeys` answers the second, and the new method supplies the first.

I took the fail-closed option over your alternative of erroring out of boot. Failing boot would reintroduce a crash-loop that one corrupt byte can trigger, i.e. trade an auth bypass for a remote DoS on every replica, since the row is Raft-replicated. Fail-closed instead: the cluster boots, serves reads and unsigned writes per its config, and refuses all signing management until an operator restores the rows — from a backup or by rebuilding the projection from the audit log.

The marker is cleared by `Reset`, so a repaired store is not permanently locked out; `TestKeyStoreUndecodableRowsResetWithKeys` pins that, because a stale count surviving a restore would turn a transient corruption into a permanent lockout.

Covered by `TestAuthorizeUnsignedBatch_BootstrapExceptionClosesOnUndecodableRows` (fresh cluster still bootstraps; undecodable rows, existing keys, and both together all reject) plus an assertion added to the boot-path test that recovery actually sets the marker. Mutation-verified: dropping `&& !a.keyStore.HasUndecodableRows()` fails the relevant subtest and nothing else. `docs/ops/signing.md` now documents that the bootstrap path stays shut in this state.

`just pre-commit` 0 issues on both modules, full `just test` exit 0 with 0 FAIL, and the `Tests-E2E` failure on the previous revision was the known non-hermetic `CheckStore` reverse-map spec — green on the re-run.

internal/infra/state/recovery.go
```diff
@@ -136,11 +136,31 @@ func (r *Recovery) RecoverState() error {
...
```
NumaryBot, line 149: 🟠 **[major] Skipped malformed signing-key rows expose unsigned bootstrap exception**

When recovery skips malformed signing-key rows and no decodable keys remain, keyStore.HasKeys() returns false — the same state as a genuinely empty, fresh deployment. If require_signatures is enabled, the admission layer's unsigned bootstrap exception (which allows the very first key to be registered without a signature) becomes reachable after any restart that encounters only corrupt rows. Because signing-key rows do exist in storage, this is materially different from a true first-boot scenario and should not be treated equivalently. An attacker or misconfigured client could exploit this window to register an unauthorized signing key.

**Suggestion:** Track whether any signing-key rows were encountered (malformed or not) separately from whether any were successfully decoded. If rows were encountered but none decoded successfully, treat the store as non-empty — for example, return an error, set a 'corrupt' sentinel flag, or refuse to enter the bootstrap-exception path. At minimum, emit a clear warning and ensure the unsigned bootstrap path is not reachable when corrupt rows are detected.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1667#issuecomment-5178417429

internal/application/check/signing.go
```diff
@@ -0,0 +1,555 @@
...
+		v.ensureProposalSnapshot()
+
+		register := payload.RegisterSigningKey
+
+		// Upsert, not insert: the FSM has no duplicate-ID rejection, so
+		// re-registering a key ID legitimately replaces its material and its
+		// parent link. The bytes are copied because the order may be reused or
+		// mutated after this returns — an aliased expectation would follow that
+		// mutation and end up comparing a row against itself.
+		v.keys[register.GetKeyId()] = signingKeyExpectation{
```
NumaryBot, line 204: 🟠 **[major] Cascade replay diverges from live FSM on root re-registration of a child key**

For a key that was first registered under a parent and later re-registered as a root, the replay in Check() clears the parent association. However, the live KeyStore.AddPublicKey path only overwrites parents[keyID] when the new parent is non-empty. Until recovery reloads the store, a subsequent cascade revoke of the old parent will still delete that key in the live store but not in the replayed checker state (or vice versa). This means Check() will report a spurious missing-signing-key integrity error for stores produced by the current FSM on the reachable sequence: register child → re-register as root → cascade revoke old parent.

**Suggestion:** Either update the live AddPublicKey upsert to explicitly clear the stale parent when the new registration has no parent (making live and replay consistent), or update the replay logic in Check() to match the live FSM's current behaviour. The two code paths must be kept in sync to avoid false-positive integrity errors.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1667#issuecomment-5178417429

internal/infra/state/recovery.go
paul-nicolas: Already closed at the commit this finding was raised against (`5115611`) — same finding as the resolved thread above, re-posted under a new hash.

The remediation is the one suggested, all three parts of it:

- `keystore.MarkUndecodableRows` / `HasUndecodableRows` (`internal/domain/crypto/keystore/keystore.go:90-111`) track "rows existed but none decoded" separately from "no rows" — the distinction `HasKeys` cannot make.
- `RecoverState` sets the counter right after logging each skipped row (`internal/infra/state/recovery.go:150-162`), and it is the only non-test caller of `query.ReadSigningKeys`, so there is no second boot path that repopulates the key store without setting it.
- The bootstrap exception is gated on `isRegisterSigningKeyRequest(req) && !a.keyStore.HasKeys() && !a.keyStore.HasUndecodableRows()` (`internal/application/admission/admission.go:1007`), so with any corrupt row the unsigned `RegisterSigningKey` is not excepted and falls through to `signing.ErrMissingSignature` — fail closed until an operator repairs the rows.

`Reset()` zeroes the counter (`keystore.go:112-123`) so the gate reopens once the corrupt rows are gone, rather than latching permanently.

Pinned by `TestAuthorizeUnsignedBatch_BootstrapExceptionClosesOnUndecodableRows` (both directions: fresh cluster may bootstrap unsigned, undecodable rows close the exception) and `TestKeyStoreUndecodableRowsResetWithKeys`.

internal/application/check/signing.go
paul-nicolas: Valid, and the consequence is worse than a checker false positive — thanks.

The divergence is real and the sequence is reachable: `AddPublicKey` (`internal/domain/crypto/keystore/keystore.go:41`) set `parents[keyID]` only for a non-empty parent, `WriteSet.GetSigningKeyChildren` (`internal/infra/state/write_set.go:1313`) cascades off exactly that map, and nothing rejects a duplicate key ID — `processRegisterSigningKey` (`processor_signing.go:9`) validates the ID shape only, and admission (`validate_order.go:448`) only the key length.

The part worth naming: the live in-memory keystore was the **only** representation that kept the stale edge. The persisted row clears it (`SaveSigningKey`, `batch.go:132`, writes a bare 32-byte value when the parent is empty), so does recovery (`Reset()` then a per-row `AddPublicKey`, `recovery.go:137`), so does the audit-driven rebuild (`backup/rebuild.go:257`), and so does the replay. That made the cascade **restart-dependent**: an un-restarted replica cascaded the reassigned key, a replica restarted since the re-registration did not, and the same revoke order then emitted different `RevokedSigningKeyLog.cascadedKeyIds` and deleted different rows per replica. That is invariants #1 and #2, not just a spurious `Check()` finding — the finding you describe is the downstream symptom. Pre-existing; the new pass exposed it.

So I took your first alternative, not the second. Mirroring the live keystore in the replay would freeze the divergence in place: the replay would then agree with un-restarted replicas and disagree with restarted ones, and it cannot do both. `"" = root` is already the documented semantics (`processor_signing.go:14`) and the persisted one, so the in-memory map was the side that was wrong — `AddPublicKey` now deletes the edge on an empty parent.

`descendantsOf` is unchanged, and its union with the per-proposal `proposalParents` snapshot stays necessary: within one proposal `GetSigningKeyChildren` still reads the committed keystore (the flush at `write_set.go:650` runs after the cascade is computed), so a key re-registered in the same proposal as the cascade revoke of its old parent is still cascaded from the old edge. Only the committed relation at proposal boundaries changed.

Fixed in 6d26a14 and pinned at three levels: `TestAddPublicKey_ReregistrationAsRootClearsParent` (the keystore upsert), `TestWriteSetSigningKeyChildrenAfterReregistrationAsRoot` (what a restarted replica computes), and a new `TestSigningVerifier_CascadeUnionsBothParentEdges` subtest for register-child → re-register-as-root → cascade-revoke-old-parent, which locks in that the fix landed on the keystore side rather than by weakening the replay. Mutation-verified: reverting the one-line change fails the first two and nothing else. `docs/ops/signing.md` no longer claims the parent is immutable, which re-registration has never honoured.

paul-nicolas:
## Both majors from the automated review are now closed

**First major** (unsigned bootstrap exception reachable on corrupt signing rows) — closed at `5115611`, detailed in the comment above.

**Second major** (cascade replay diverging from the live FSM on root re-registration) — closed at `6d26a14`. It was valid, and the report understated it: the spurious `SIGNING_KEY_MISMATCH` is a symptom, and the actual defect is an FSM determinism breach.

`keystore.AddPublicKey` wrote `parents[keyID]` only for a non-empty parent, so re-registering a key ID as a root left the old edge in memory — and that map is what `WriteSet.GetSigningKeyChildren` walks to build a revocation cascade. The live in-memory keystore was the only representation that kept the edge; the persisted row, recovery, the audit-driven rebuild and the checker replay all drop it. So the cascade was restart-dependent: an un-restarted replica cascaded the reassigned key, a restarted one did not, and the same revoke order emitted different `RevokedSigningKeyLog.cascadedKeyIds` and deleted different rows on different replicas. Invariants #1 and #2. Pre-existing — the pass added in #1669 is what surfaced it.

Fixed on the keystore side rather than by aligning the replay to the live map: the replay already matches the persisted row and the documented `"" = root` semantics, and mirroring the live map would leave it agreeing with un-restarted replicas and disagreeing with restarted ones, which it cannot do for both. `descendantsOf` is unchanged — its per-proposal `proposalParents` union is still required for the same-proposal case, where the cascade legitimately reads the pre-flush committed edge.

Pinned at the keystore, WriteSet and checker levels; reverting the one-line change fails the first two assertions and nothing else. `docs/ops/signing.md` no longer claims the parent relationship is immutable, which re-registration has never honoured — worth a follow-up ticket if immutability was the intent, since nothing enforces it.

Gates on `6d26a14`: `just pre-commit` 0 issues on both modules, full `just test` exit 0 with 0 FAIL across 79 packages, no codegen churn.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1667#issuecomment-5178417429

internal/domain/crypto/keystore/keystore.go
```diff
@@ -33,14 +38,33 @@ func (ks *KeyStore) GetPublicKey(keyID string) ed25519.PublicKey {
...
+// #1 and #2 (identical cache state per applied index, deterministic FSM).
 func (ks *KeyStore) AddPublicKey(keyID string, pubKey ed25519.PublicKey, parentKeyID string) {
 	ks.mu.Lock()
 	defer ks.mu.Unlock()
 
 	ks.publicKeys[keyID] = pubKey
-	if parentKeyID != "" {
-		ks.parents[keyID] = parentKeyID
+
+	if parentKeyID == "" {
```
NumaryBot, line 61: 🔴 **[blocker] FSM signing-parent behavior change diverges live store from checker replay and breaks rolling upgrades**

When a child key is re-registered as a root, the new code clears the parent edge in the live key store. However, the checker replay in Check() handles this differently, so the two paths are out of sync. More critically, during a rolling deployment, upgraded nodes will delete the parent edge at this point while old nodes keep it in memory. For the same applied Raft index, the in-memory key stores diverge. A later cascade revoke reads that parent map via GetSigningKeyChildren, producing different cascaded_key_ids and different deleted key rows for the same order — a split-brain condition that corrupts auditable state.

**Suggestion:** Keep the live FSM behavior backward-compatible during rolling upgrades: instead of changing FSM logic directly, introduce a new log-entry version or a migration step so that both old and new nodes process the same sequence identically. Additionally, reconcile the live AddPublicKey upsert and the Check() replay to use exactly the same parent-edge semantics (either both clear stale parents on root re-registration, or neither does).

internal/application/check/signing.go
```diff
@@ -0,0 +1,555 @@
...
+		// Required here too, not just on the register path: descendantsOf reads
+		// proposalParents, so without this a cascade in a proposal that registered
+		// nothing would walk the snapshot left behind by an EARLIER proposal and
+		// cascade from a parent link that has since been committed away.
+		v.ensureProposalSnapshot()
+
+		revoke := payload.RevokeSigningKey
+
+		revoked := []string{revoke.GetKeyId()}
+		if revoke.GetCascade() {
```
NumaryBot, line 226: 🟠 **[major] Checker cascade replay misses in-proposal parent edges, causing false-positive integrity errors**

For a single proposal that (1) registers a child under a parent, (2) re-registers the same child under a different parent or as a root, and (3) cascade-revokes the original parent, the live FSM still cascades the child because WriteSet.GetSigningKeyChildren scans every pending addition in the proposal. The checker replay overwrites the child's parentKeyID with only the latest registration and only preserves pre-proposal edges, so descendantsOf misses the earlier pending parent edge and reports a healthy store as missing or tampered. This is a reachable sequence via normal business operations.

**Suggestion:** In the checker replay, accumulate all in-proposal parent additions before performing cascade revocation — do not overwrite the parent edge with only the last registration. Mirror the same scan logic that WriteSet.GetSigningKeyChildren uses in the live FSM so that both code paths cascade identically.

paul-nicolas: Valid, confirmed by tracing both paths — and thanks, this is a genuinely different case from the cross-proposal one, not a restatement of it.

`GetSigningKeyChildren` (`internal/infra/state/write_set.go:1332-1339`) loops over the whole `pendingSigningKeyUpdates` slice, not the last entry per key, so a superseded addition still contributes its edge. Walking the sequence for `register(child→parent)`, `register(child→root)`, `revoke(parent, cascade)` in one proposal: at the revoke, pending holds both additions and no removals, `child` is not a committed child of `parent` (it was created in this proposal), the first addition matches `parentKeyID == "parent"` → `children == ["child"]`. The flush then applies both adds and both removes in order, so the store ends with `child` deleted.

The replay disagreed for exactly the reason you give: `applyOrder` overwrote `v.keys["child"].parentKeyID` with the second registration, and `proposalParents` had no entry for `child` because the snapshot predates its creation. Both terms of the union missed the edge, `child` stayed in the expected set, and the pass would report `SIGNING_KEY_MISMATCH` against a store that legitimately deleted it.

Fixed in `a7e5778` by adding a third edge source rather than by reshaping the first two. `proposalEdges` records every edge the current proposal asserted, keyed by child, including superseded ones; `descendantsOf` treats a candidate as a child of `current` when the running pointer, the pre-proposal snapshot, **or** `proposalEdges` says so. That is the union `GetSigningKeyChildren` actually computes. The pending-removal filter still comes for free, since the candidate loop ranges over `v.keys` and revoked keys are already gone from it.

`proposalEdges` is per-proposal state, cleared in `ensureProposalSnapshot` alongside `proposalParents` — an append-only record of every parent a key ever had would over-cascade, deleting keys that should survive once the reassignment is committed. Pinned in both directions: two subtests for the in-proposal case (superseded by a root re-registration and by a reparenting one, since the two reach the branch differently) and one asserting a superseded edge does **not** leak into a later proposal's cascade. Mutation-verified — dropping the `proposalEdges` term fails the first two and nothing else.

internal/domain/crypto/keystore/keystore.go
paul-nicolas: Not taking this one, on two independent grounds.

**The premise about the two paths is no longer accurate.** The finding says the live store and the checker replay handle root re-registration differently. As of the commit it is filed against they handle it identically: the replay assigns `parentKeyId` verbatim, so an empty parent clears the edge, and `AddPublicKey` now does the same. That is the second half of your own suggestion — "either both clear stale parents on root re-registration, or neither does" — already implemented, and it is what the change accomplished rather than something it broke. There *was* a live-vs-replay gap left, but a different one: the in-proposal superseded edge, reported separately as `c46d3ed15c5604e9` and fixed alongside this.

**The rolling-upgrade remediation is explicitly ruled out for this repo.** v3 has never been released: no deployment carries data across versions and all state is wiped on every push to `release/v3.0`. `AGENTS.md` states the policy directly — "No compatibility shims. Do not add migrations, version guards, or fallback paths to preserve old wire or storage formats" — formalized in 47da27c2 (#1668). A new log-entry version or a migration step to keep mixed-version replicas byte-identical is precisely what that forbids, and there is no persisted state for it to protect.

Worth naming the shape of the argument, too: it proves too much. Every FSM behaviour fix changes what the apply path does, so "do not change FSM logic directly, add a version gate instead" would forbid fixing determinism bugs at all. And the status quo it defends is not a stable baseline — the old behaviour was *itself* a divergence, just triggered by a restart rather than a deploy: the stale edge lives only in memory, and recovery reloads from parent-less rows, so an un-restarted and a restarted replica already computed different `cascadedKeyIds` for the same order. The fix removes a permanent divergence source; declining it to avoid a transient one during an upgrade of a cluster whose state is wiped anyway would be the worse trade.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1667#issuecomment-5178417429

Azorlogh (commented):
The three fixes check out — I verified the claims against the branch and ran the new tests (all pass). Two things at the PR level:

**Stale description**: #1669 was merged into this branch (bf12d0980), so this PR now also ships the full `signingVerifier` checker pass, the `HasUndecodableRows` bootstrap gate, the new enum values and the doc updates — the body and Testing section still describe the pre-stack PR. Worth updating so the review scope is visible.

**FSM cascade BFS (out of diff, but surfaced by this PR's own analysis)**: `processRevokeSigningKey` (`processor_signing.go:44-52`) walks children with no visited set, and `RemoveSigningKey` only runs after the walk. `TestSigningVerifier_CascadeTerminatesOnAParentCycle`'s comment establishes that parent cycles and self-parents are accepted end to end — so a single cascade revoke on such a key infinite-loops the apply path on every replica, and being raft-committed it replays on restart. The checker got a visited set; the FSM needs the same (plus dedup of `cascaded`), here or as its own ticket.

Inline: one cascade-model divergence (repro'd), a chain-break interaction, a perf nit, and three comment nits.

internal/application/check/signing.go
```diff
@@ -0,0 +1,590 @@
...
+// (minus those the same proposal removed) and the key of EVERY pending addition in
+// the proposal whose parentKeyID matches. So a key re-registered under a new parent
+// within the same proposal as a cascade revoke of its OLD parent is cascaded from
+// both — the FSM never consults the reassigned pointer to exclude it. Walking only
+// the running relation would drop that key from the cascade, leave it in the
+// expected set, and report a false SIGNING_KEY_MISMATCH against a store that
+// legitimately deleted it.
+//
+// Hence THREE edge sources below, and all three are load-bearing:
+//   - the running relation, for keys whose parent is unchanged in this proposal;
```
Azorlogh, line 270: This union misses one edge the FSM has: `GetSigningKeyChildren` matches *every* pending add edge (`write_set.go:1333`), including ones superseded within the proposal. Repro (confirmed against this branch): one proposal with `register X under P` → `register X under Q` → `revoke P cascade` — the FSM cascades X via the pending X→P edge, but here neither the snapshot (X uncommitted at proposal start) nor the current parent (Q) matches, so X stays expected → false `SIGNING_KEY_MISMATCH` against a healthy store. Accumulating every register edge seen in the current proposal and unioning it here would close it.

internal/application/check/checker.go
```diff
@@ -852,6 +889,10 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 		return err
 	}
 
+	if err := signing.compare(snap, callback); err != nil {
```
Azorlogh, line 892: When `verifyAuditHashChain` stops early on a HASH_MISMATCH, this compare still runs against a prefix expectation with `coldComplete == true` — emitting the same false-positive classes the INCOMPLETE suppression exists to prevent. Marking the verifier incomplete on those early-return paths would be cheap.

internal/application/check/checker.go
```diff
@@ -2524,6 +2571,57 @@ func (c *Checker) verifyAuditHashChain(
...
+				logSeq := item.GetLogSequence()
+				if logSeq == 0 || logSeq <= signing.archiveEndSeq {
+					continue
+				}
+
+				if logSeq < success.GetMinLogSequence() || logSeq > success.GetMaxLogSequence() {
+					continue
+				}
+
+				order := &raftcmdpb.Order{}
```
Azorlogh: `collectExpectedSkippable` already unmarshaled this item's order — threading the decoded orders through would avoid a second full decode of the live audit range per Check() run.

internal/application/admission/admission.go
```diff
@@ -993,7 +993,18 @@ func (a *Admission) resolveBatch(req *servicepb.ApplyRequest) (verifiedBatch, er
 func (a *Admission) authorizeUnsignedBatch(reqs []*servicepb.Request) error {
 	for _, req := range reqs {
 		if isSigningManagementRequest(req) {
-			if isRegisterSigningKeyRequest(req) && !a.keyStore.HasKeys() {
+			// HasUndecodableRows closes the bootstrap window on a store whose signing
+			// rows exist but did not decode. Such a cluster is NOT fresh: recovery
+			// skips those rows (rather than panicking, as it used to), so with every
```
Azorlogh: "rather than panicking, as it used to" is change-narration — commit-message material.

internal/application/admission/errors.go
```diff
@@ -51,4 +51,14 @@ var (
 	// and HTTP) rather than silently creating a permanently-unbuilt index.
 	// See indexes.Supported.
 	ErrIndexTargetUnsupported = domain.NewValidationSentinel("index target not supported (metadata: ACCOUNT/TRANSACTION; account builtin: ASSET; log builtin: DATE)")
+
+	// ErrSigningKeyInvalidLength rejects a RegisterSigningKey order whose
+	// public_key is not exactly an Ed25519 public key (32 bytes). Nothing on the
+	// server write path checked this before: admission copied the bytes verbatim,
```
Azorlogh: "Nothing on the server write path checked this before…" narrates history — belongs in the commit message.

internal/application/check/signing.go
```diff
@@ -0,0 +1,590 @@
...
+
+// signingVerifier re-derives the two signing projections — the SubGlobSigningKey
+// rows and the SubGlobSigningConfig require-signatures flag — from chain-bound
+// orders, and compares that expectation to what is persisted.
+//
+// Both projections are invariant-#8 projections with no pass before this one:
+// state.Recovery loads them straight into the runtime key store and the
+// requireSignatures gate, which admission consults to accept or reject every
+// signed write. A disk edit therefore changes who may write, and CheckStore
+// previously reported the store healthy.
```
Azorlogh: "and CheckStore previously reported the store healthy" is change-narration — drop.

internal/application/check/signing.go
paul-nicolas: This is fixed at HEAD — I think the draft predates `a7e5778` ("cascade from superseded in-proposal parent edges in the signing replay"), whose whole subject is this case.

`proposalEdges` (`signing.go:107`) accumulates every parent edge the current proposal asserted, keyed by child, superseded ones included: `applyOrder` **appends** on each non-root registration rather than overwriting, and `descendantsOf` unions it as the third edge source (`!slices.Contains(v.proposalEdges[candidate], current)`). It is cleared with `proposalParents` in `ensureProposalSnapshot`, so it stays per-proposal.

Your repro walks through as: `register X under P` → `proposalEdges[X] = [P]`; `register X under Q` → `[P, Q]` with running parent `Q`; `revoke P cascade` → X matches on `slices.Contains(proposalEdges[X], P)` and is cascaded. Pinned by `TestSigningVerifier_CascadeUnionsBothParentEdges/in-proposal_edge_superseded_by_a_reparenting_re-registration_still_cascades`, with a root-re-registration variant and a `a superseded edge does not leak across proposals` counterpart for the case where the cascade lands in a later proposal.

Your read of the FSM was right, and it is the same reasoning that commit is built on — the analysis just landed a couple of hours ahead of the fix.

internal/application/check/checker.go
paul-nicolas: Valid, fixed in `f49e923`. And it is worse than one path: `verifyAuditHashChain` has **three** non-error early exits, not just the HASH_MISMATCH one — an entry carrying embedded items in its persisted value (`checker.go:2480`), a header that cannot be re-hashed (`:2510`), and the hash mismatch itself (`:2545`). The signing fold runs inside that loop, so all three truncate it.

What made it a false-positive generator rather than just incomplete: the comparison is over whole key **sets** in both directions, so a prefix reports every registration past the break as injected *and* every revocation past it as a lost row — noise stacked on top of the one event that describes the store's actual problem. Worth contrasting with the walk's other outputs: `expectedSkippable` and the chain-bound maps are consulted per log sequence, so an absent entry simply yields no expectation and needs no equivalent handling. That asymmetry is why only this pass had to change.

Added `liveTruncated` + `markLiveTruncated()` called at all three exits, and the suppression gate is now `!coldComplete || liveTruncated`. The emitted message names which side went unread (archived / live / both) so an operator can tell a missing cold-storage backend from a broken chain without correlating events by hand. The malformed-row carve-out stays outside the suppression as before.

Covered per cause in `signing_test.go` plus `TestVerifyAuditHashChain_MarksSigningFoldTruncatedOnEveryBreak`, one case per exit. Mutation-verified both ways: narrowing the guard back fails the three live-truncation compare cases, and dropping any single `markLiveTruncated()` call fails exactly its own exit's case and nothing else. `AGENTS.md` invariant #8 and the checker subsystem doc now state that a new early exit must mark the fold truncated.

internal/application/check/checker.go
paul-nicolas: Fixed in `e9256b0`. `collectExpectedSkippable` now returns the decoded orders, parallel to `items` by index and nil where the bytes did not decode, and the signing fold indexes that slice.

Returned rather than taken as an out-parameter so the 15 existing test call sites keep compiling untouched — a Go call used as a statement may discard its result — which kept the change to the two functions that actually care.

Two details worth stating since they are the contract now: the element is recorded *before* `collectExpectedSkippable`'s own filters, because the failure-side / out-of-window / duplicate-sequence skips are its concerns and not the caller's; and the pairing has to be positional, since the signing fold applies its own archive-boundary and fresh-log-window filters and cannot re-derive which item an order came from otherwise.

The loud path is preserved: an in-window item whose bytes do not decode still raises the invariant-#7 error, now off the nil element rather than a second failed unmarshal. The soft skip stays correct for the whitelist, which simply gains no entry, and stays wrong for the signing fold, where a dropped order silently changes the expected key set. `TestCollectExpectedSkippable_ReturnsOrdersParallelToItems` pins the index alignment and the nil-on-undecodable behaviour.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1667#issuecomment-5178417429

internal/application/check/checker.go
NumaryBot, line 892: 🟡 **[minor] False-positive integrity errors emitted after early HASH_MISMATCH return in verifyAuditHashChain**

When `verifyAuditHashChain` returns early on a HASH_MISMATCH, the subsequent comparison still runs against a prefix expectation with `coldComplete == true`. This can emit the same false-positive integrity-error classes that the INCOMPLETE suppression was designed to prevent, because the verifier is not marked incomplete on those early-exit paths.

**Suggestion:** On any early-return path inside `verifyAuditHashChain` that stops processing due to a hash mismatch, mark the verifier as incomplete before returning. This is symmetric with how the INCOMPLETE suppression already works for normal truncated runs and would prevent the follow-on comparison from treating a partial result as a full one.

internal/application/check/checker.go
```diff
@@ -2523,7 +2603,65 @@ func (c *Checker) verifyAuditHashChain(
...
+			// The decoded orders come back so the signing fold below reuses them
+			// instead of unmarshalling the whole live audit range a second time.
+			// Parallel to `items` by index; nil where the bytes did not decode.
+			decoded := collectExpectedSkippable(items, success.GetMinLogSequence(), success.GetMaxLogSequence(), expectedSkippable, chainBound)
+
+			// Fold signing orders from this successful entry, over the SAME fresh-log
+			// window collectExpectedSkippable uses. AuditSuccess.{Min,Max}LogSequence
+			// are computed only over freshly-created logs, never over reference
+			// sequences, so `min <= logSeq <= max` admits exactly this entry's own
+			// logs and excludes a legacy pre-f9ee1e829 per-order REFERENCE item
```
NumaryBot, line 2615: ⚪ **[nit] Redundant full decode of live audit range per Check() run**

`collectExpectedSkippable` already unmarshals each item's order; threading the decoded orders through would avoid a second full decode of the live audit range on every Check() invocation.

**Suggestion:** Pass the already-decoded orders from `collectExpectedSkippable` through to downstream consumers rather than re-decoding them in the second pass.

internal/application/check/signing.go
```diff
@@ -0,0 +1,640 @@
...
+//   - proposalEdges, for an edge this proposal asserted and then SUPERSEDED with a
+//     later registration of the same key. That case is invisible to the other two:
+//     the replacement overwrote the running pointer, and a key first registered
+//     inside this proposal has no pre-proposal entry at all. "Pending addition"
+//     means every element of pendingSigningKeyUpdates, not just the last one per
+//     key, so register(child→parent) + register(child→root) + cascade-revoke(parent)
+//     in ONE proposal does delete child on the live path.
+//
+// Keys the proposal already removed are absent from v.keys, which reproduces
+// GetSigningKeyChildren's pending-removal filter for free — and it filters
```
Azorlogh: The "comes for free" claim breaks when a key is revoked and then re-registered in the SAME proposal: the FSM builds `pendingRemovals` as a set over the whole proposal (`write_set.go:1316-1330`), so once X has any removal in the proposal it is excluded from every cascade in it, even after re-registration — while re-registration puts X back into `v.keys`, so the walk cascades it via the current edge.

Repro (confirmed on both sides against this branch): one proposal with `revoke(X)` → `register(X under P)` → `cascade-revoke(P)`. FSM: X excluded from the cascade, `Absorb` applies `[remove X, save X→P, remove P]` in order → the store keeps X (and `RebuildDelta` agrees). Verifier: X cascaded out of the expected set → false `SIGNING_KEY_MISMATCH` ("no audited registration") against a healthy store. A per-proposal revoked-keys set (cleared with `proposalParents`/`proposalEdges`, consulted by `descendantsOf`) would reproduce the FSM's filter.

Azorlogh (commented):
Went through the six new commits — all verified against the branch, affected suites pass, and the keystore stale-parent-edge find is a genuinely good catch. One residual divergence in the cascade model, inline.

internal/application/check/signing.go
paul-nicolas: Valid, and reproduced on both sides before touching anything — thanks, this is a genuinely different case from the superseded-edge one, not a restatement of it. `a7e5778` fixed superseded *additions*; nothing there touches the *removal* filter, so the bot summary calling this covered was wrong.

Fixed in `d689c59` with the per-proposal set you describe.

**The mechanism, confirmed.** `pendingRemovals` is built by ranging over the whole `pendingSigningKeyUpdates` slice (`write_set.go:1381-1385`) and then applied twice — to the committed children (`:1389-1393`) and to the pending-additions union (`:1396-1402`) — so one `RemoveSigningKey(X)` anywhere in the proposal excludes X from every cascade in it. `processRevokeSigningKey` walks the child relation before calling `RemoveSigningKey` (`processor_signing.go:83-89`), so the walk for revoke *N* sees the removals of orders *1..N-1* and none of its own. On the replay side the register reinstates the row and `descendantsOf` matches it on the first edge source. Measured, one proposal each:

| sequence | store | expected set | |
|---|---|---|---|
| `revoke(X)` → `register(X→P)` → `cascade-revoke(P)` | keeps X at `parent=P` | X cascaded out | divergent |
| X committed under P, `revoke(X)` → `register(X→Q)` → `cascade-revoke(P)` | keeps X at `parent=Q` | X cascaded out | divergent |
| `register(child→P)` → `register(child→"")` → `cascade-revoke(P)` | deletes child | child cascaded out | agrees |

The second matters because it lands through `proposalParents` rather than the running pointer — the fix had to sit in the candidate filter, not in one edge term. The third is the `bbcc821d4` behaviour and is now pinned as a regression guard.

**Reachability is ordinary traffic, not tampering.** `authorizeUnsignedBatch` takes `batch.GetRequests()` (`admission.go:982, 993`), so one `ApplyRequest` becomes N orders in one proposal. A client rotating a key's material and revoking its old parent in a single batch produces a false `SIGNING_KEY_MISMATCH` against a healthy store.

**Fixed on the checker side, which is the opposite call from the keystore one** — worth stating since the two look similar. There the live in-memory keystore was the only representation holding the stale edge, so the cascade was restart-dependent: invariants #1/#2, an integrity defect that had to be fixed at the source. Here `pendingSigningKeyUpdates` is proposal-local state rebuilt identically on every replica from the committed order sequence, so the FSM is deterministic and replica-consistent — nothing to fix there, and aligning it would change `cascaded_key_ids`, which is hashed into the chain.

**Two details in the fix carry weight beyond "consult a set".** `proposalRevoked` is filled *after* `descendantsOf` returns, mirroring the FSM's walk-then-remove ordering — filling it first makes a revoke exclude its own targets from its own cascade, and there is a test for exactly that. And the check skips the candidate entirely instead of dropping it from the result: `GetSigningKeyChildren` filters the key out of what it *returns*, so the FSM's BFS never recurses through it and the excluded key's own subtree survives too. Omitting from the output alone would still cascade the grandchild.

Recorded every key each revoke removed, target and re-derived descendants alike, since `RemoveSigningKey` is called on all of them — so a key removed as a cascade descendant is excluded from a second cascade in the same proposal as well.

Pinned by `TestSigningVerifier_CascadeSkipsKeysTheProposalAlreadyRemoved`: the three re-registration targets (revoked parent / unrelated parent / root), the surviving subtree, the cascade-descendant case, the walk-then-fill ordering, and a cross-proposal guard that the set does not leak. Mutation-verified — disabling the `proposalRevoked` check fails exactly the five cases that depend on it and nothing else in the package; dropping the `clear()` fails only the leak guard.

`AGENTS.md` invariant #8 updated: the edge union was only half the cascade model, and the "comes for free" claim you flagged is retracted there and in the code comment.

**One thing I did not change.** The FSM's order-blind filter leaves X stored with `parent_key_id` pointing at the parent it just deleted — a dangling pointer. Deterministic and harmless to consensus, and the replay now agrees with it, so it is not a correctness issue for this pass; but it is odd enough to be worth a follow-up ticket rather than a silent accept. Say the word and I will file it.

Gates on `d689c59`: `just pre-commit` 0 issues on both modules, `GOROOT= go build ./...`, full `just test` exit 0 with 0 FAIL across 81 packages, no codegen churn.

docs/technical/architecture/subsystems/checker/checker.md
```diff
@@ -62,6 +63,55 @@ Notes:
...
+
+The set of *edges*, however, is load-bearing, and this is the one place the cascade must follow the FSM closely. `GetSigningKeyChildren` returns the **union** of two relations: the revoked key's committed children, minus any the same proposal removed, and the keys that same proposal registered under it. It never consults a reassigned parent pointer to exclude a key. A key re-registered under a new parent *in the same proposal* as a cascade revoke of its old parent is therefore still revoked, so the checker walks both edges: `beginProposal` snapshots the parent relation once per audit entry — one entry is one proposal, matching the FSM's single `WriteSet.Reset` per proposal — and `descendantsOf` treats a key as a child if *either* its snapshotted or its current parent matches. Keys the proposal already removed are absent from the expectation, which reproduces the pending-removal filter for free.
```
Azorlogh: Stale after the last two fixes — this still describes the two-edge model and re-asserts the claim `d689c59` retracted. It's now a **union of three** edge sources (running + `proposalParents` + `proposalEdges`, per the superseded-edge fix), and "reproduces the pending-removal filter for free" is exactly the reasoning `proposalRevoked` replaced — absence from the expectation does *not* reproduce it, since re-registration reinstates the row. AGENTS.md invariant #8 was updated for both; this paragraph (and "Both halves of that" on line 85) was missed.

Azorlogh (commented):
The `proposalRevoked` fix is correct — I traced it against `GetSigningKeyChildren` + `processRevokeSigningKey` and re-ran the repro (incl. a ghost-revoke variant); it matches the FSM. One stale doc left over, inline.

docs/technical/architecture/subsystems/checker/checker.md
paul-nicolas: Valid on both counts, and it is doc-only drift — the code at HEAD already does what you describe. Fixed in `27c626c2`.

`descendantsOf` (`internal/application/check/signing.go:319-323`) admits a candidate on three disjuncts, and the third is not redundant: a replacement registration overwrites the running pointer and a key first registered inside the proposal has no `proposalParents` entry, so `register(child→parent)` + `register(child→root)` + `cascade-revoke(parent)` in one proposal is invisible to the other two edges. The paragraph named only two.

The "for free" sentence is worse than incomplete — it is the reasoning `d689c59` retracted. `pendingRemovals` is a set over the whole `pendingSigningKeyUpdates` slice with no ordering awareness, while absence from the expectation is a point-in-time fact, so `revoke(X)` + `register(X under P)` + `cascade-revoke(P)` in one proposal reinstates the row and the walk follows the reinstated edge. That is what `proposalRevoked` (`signing.go:313-315`) exists for.

Line 85 is rewritten with it, including the two load-bearing details: `proposalRevoked` is filled *after* `descendantsOf` runs (`signing.go:234-245`), or a revoke would exclude its own targets from its own cascade; and it skips the candidate entirely rather than dropping it from the result, so the excluded key's own subtree survives too.

I also re-checked the rest of the section against the other two commits in the range: `3a8ae47c` (`AddPublicKey` clearing the parent edge on root re-registration) and `44fb5d86` (the FSM cascade cycle guard) are both covered in `docs/ops/signing.md`, and lines 81 and 87 still match HEAD. This paragraph was the only drift.

Azorlogh (approved):

PR #1669, opened 2026-08-04
feat(EN-1515): verify signing key/config projections against audited orders

paul-nicolas:
Closes the invariant-#8 gap for the two persisted signing projections.

**Stacked on #1667** — retarget to `release/v3.0` once that merges.

## The gap

`SubGlobSigningKey` and `SubGlobSigningConfig` are projections of hash-chained orders, but no checker pass read them. The cause was structural, not an oversight: the checker only ever decoded **ledger-scoped** orders — `grep SystemScoped checker.go` returned nothing, and `replay.DecodeOrderEffects` returns empty for anything not ledger-scoped. Signing orders are `SystemScopedOrder` variants, and every `chainBoundState` field is keyed by ledger, so there was no slot for cluster-global expectations.

Meanwhile `recovery.go:139/151` loads both projections straight into the runtime key store and the `requireSignatures` gate, which admission consults to accept or reject every write. A disk edit could add a trusted key, replace one's public key, re-parent it, or disable signature enforcement cluster-wide with a single byte — and `CheckStore` reported the store healthy.

## Approach

A `signingVerifier` in a new `internal/application/check/signing.go` (a sub-type rather than more methods on an already ~240KB `checker.go`, per convention #6), with one `applyOrder` fold serving two ranges:

- **Archived history** via `coldstorage.ColdReader`, **oldest chapter first** — signing state accumulates forward and has no TTL, so unlike `reDeriveArchivedIdempotency` there is no cutoff and every archived chapter is read.
- **The live range**, hooked into the existing `verifyAuditHashChain` walk, gated on `entry.GetSuccess() != nil` so failed proposals contribute nothing.

Then a both-directions comparison — because **row absence is the only representation of revocation**, a lost key and an injected key are equally failures.

### Three deliberate design points

**Expectations are never seeded from the baseline checkpoint.** `attributes.writeBaselineAttributes` copies the attribute zone verbatim from the live store, so seeding would verify old, never-touched keys against a copy of themselves — worthless for exactly the keys most likely to be tampered. Cold replay keeps the derivation audit-bound. This is the one structural difference from the baseline-seeded passes, and it is why unreadable archived history is *reported* rather than papered over.

**Cascade descendants are re-derived** from the expected parent map, never read from the log's `cascadedKeyIds` — re-deriving it is the point, and trusting the recorded list would let a tampered projection justify itself. Traversal order is deliberately *not* the FSM's: the comparison is over the final key *set*, whose reachability the parent relation fully determines.

**Findings are sorted before emission** (`slices.SortFunc` + `cmp.Or`, matching `compareReverseMapOrphans`). Both sides of the comparison range Go maps, so unsorted emission would produce a different event stream on every run.

## Error types

`SIGNING_KEY_MISMATCH` (21) — injected, missing, altered, re-parented, or undecodable rows. `SIGNING_CONFIG_MISMATCH` (22) — `requireSignatures` drift. `SIGNING_VERIFICATION_INCOMPLETE` (23) — archived history required but not fully replayed; the pass never presents a partial replay as clean, and **skips the key and config comparisons entirely** in that case rather than reporting them against a prefix-only expectation. Public-key bytes never appear in a message.

## Testing

Nine tests covering fold semantics, both-directions comparison, every tamper class (inject / delete / replace material / re-parent / flip the config byte / truncate a row below 32 bytes), archived coverage including a real multi-chapter cold read, and two regression guards for the design decisions above: cascade order-independence, and repeated-run emission determinism.

Each guard was mutation-tested — removing it makes its test fail (or, for the cycle guard, hang). Notably the cascade BFS needs a `visited` set: registration is an upsert and nothing validates that a parent exists or that the graph is acyclic, so `register("a", parent="b")` then `register("b", parent="a")` is accepted end to end and an unguarded walk loops forever inside `Check()`.

## Known cost

Signing has no TTL, so every `Check()` reads every archived chapter. Acceptable for now (chapter counts are small, signing orders rare) but worth measuring. It must **not** be optimised by caching the fold into a projection — that reintroduces the circularity that ruled out the baseline approach.

## Review fixes (post-open)

Four defects found in review, all of the same shape — reporting a mismatch the expectation cannot support, i.e. flagging a healthy store as tampered — plus one genuine detection gap. Each is mutation-verified: with the guard removed the matching test fails.

- **Empty-audit path** (`86dc80a`): `Check()` returns early at `lastSequence == 0`, before the compare phase, so an injected key on a zero-log store was reported clean. The signing projections are cluster-global, and every successful signing order writes a log, so a zero-log store *proves* no key was audited — the expectation is legitimately empty and any stored row is unaudited. `foldArchived` still runs there rather than assuming empty coverage, so a fully-archived store reports one INCOMPLETE instead of a spurious mismatch per legitimate key.
- **Fresh-log window, live fold** (`cf67985`): bounded by `success.Min/MaxLogSequence`, the discriminant `collectExpectedSkippable` already uses, not by the archive boundary alone. On upgraded stores a legacy pre-`f9ee1e829` replay reference points back at a log an earlier entry folded; re-applying it re-registers a key after its revoke. The orders are individually idempotent — the *ordering* is not, which is why idempotence is not an argument for dropping the window.
- **Fresh-log window, archived fold** (`92d1e4e`): the same fix on the cold side, which the first pass missed. `foldChapter` now walks archived audit *entries* and reads each entry's items, because only the entry carries the window and the success/failure outcome; that also replaces the `LogSequence == 0` failure proxy with a direct success check. This half matters more: pre-boundary signing state has no second audit-derived source, the baseline checkpoint being a copy of the projection under test.
- **Incomplete coverage** (`cf67985`): a partial archived fold leaves the expectation a prefix of the real history, unsound in *both* directions — a revoke in an unread chapter leaves its key expected, a register leaves its row unexpected. The key and config comparisons are now skipped in that case. The malformed-row class stays, needing no oracle. This replaced a test of mine that had pinned the false positive as correct.

Also in the stacked PR (#1667): the unsigned `RegisterSigningKey` bootstrap exception had to be closed on stores with undecodable signing rows, since making boot survive those rows is what makes an empty key store reachable without a fresh cluster.

## Sharp edge for later

`signingArchivedSST` now emits audit **entries** alongside items. An items-only archived fixture folds nothing at all, so a future cold-storage signing test that omits entries will silently observe an empty expectation. The helper documents it.


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1669#issuecomment-5178419486

internal/application/check/checker.go
```diff
@@ -852,6 +863,10 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 		return err
 	}
 
+	if err := signing.compare(snap, callback); err != nil {
```
NumaryBot, line 892: 🟠 **[major] Signing verification skipped for empty-log audit stores**

When the primary store has `lastSequence == 0`, `Check` returns early before reaching the signing projection comparison. This means a store with an injected `SubGlobSigningKey` row or a tampered `requireSignatures=true` config row is still reported as clean, because the signing verification is never executed. Since signing projections are cluster-global and can exist or be tampered with independently of ledger logs, the early-exit fast path bypasses the exact tamper class this PR is intended to detect.

**Suggestion:** Execute the signing verifier (with an empty expected state) before returning on the empty-audit fast path, so that cluster-global signing key/config projections are still verified even when there are no ledger log entries.

paul-nicolas:
## Fixed in 86dc80a — the finding was real, the suggested fix needed one correction

The gap is confirmed. `Check()` returns at the `lastSequence == 0` fast path before the compare phase where `signingVerifier.compare` lived, so an injected `SubGlobSigningKey` row on a zero-log store was reported clean. Since signing projections are cluster-global rather than per-ledger, that is exactly the tamper class this PR exists to detect, on exactly the store state a freshly bootstrapped cluster has.

**Why the expectation may safely be empty there** — worth stating explicitly, because it is what makes the fix sound rather than merely plausible: every successful signing order returns a `LogPayload`, and `processOrder` (`internal/domain/processing/processor.go:351`) assigns each returned payload a global sequence and writes it to `ZoneCold/SubColdLog`. So `lastSequence == 0` *proves* the audit registered no signing key. Every stored row is unaudited by construction, not merely unmatched.

**The one correction:** taking "run the verifier with an empty expected state" literally would introduce a false-positive path. A chapter purge deletes the archived log range (`WriteSet.executePurge`), so a fully-archived quiescent store can read `lastSequence == 0` while holding real signing history in cold storage — and comparing that store against a hardcoded empty expectation would emit `SIGNING_KEY_MISMATCH` for every legitimately registered key, painting a healthy cluster red.

So the fast path runs `foldArchived` + `compare` rather than `compare` alone, and the chapter read is hoisted above the fast path to feed it. That yields the correct outcome in both worlds: no archived chapters → empty expectation, stored rows flagged; archived chapters present → cold storage folded, or one `SIGNING_VERIFICATION_INCOMPLETE` if it cannot be read.

That state is in fact unreachable today — archiving emits its own logs above the range it purges, so at least one log always survives — but that is a property of the archive flow's log emission, not an invariant of this pass, and I would rather not have this pass's correctness depend on another subsystem's ordering staying put. Both the code comment and `docs/.../checker/checker.md` record the reasoning.

Covered by `TestCheck_SigningProjections_EmptyAuditWiring` (injected key, tampered require-signatures flag, and clean-store cases). Mutation-tested: with the three added statements removed the test fails; restored, it passes.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1669#issuecomment-5178419486

internal/application/check/checker.go
```diff
@@ -852,6 +889,10 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
...
```
NumaryBot, line 892: 🟠 **[major] Signing verification skipped for empty-log audit stores**

When the primary store has `lastSequence == 0`, `Check` returns early before reaching the signing projection comparison. Because signing projections (SubGlobSigningKey, requireSignatures config rows) are cluster-global and can be injected or tampered with independently of ledger logs, an attacker can manipulate these rows in an otherwise empty store and the check will still pass clean. The new tamper-detection logic this PR introduces is therefore entirely bypassed on empty-log stores.

**Suggestion:** Execute the signing verifier with an empty expected state before returning on the empty-audit fast path, or hoist the signing check above the early-exit guard, so cluster-global signing key/config projections are verified even when there are no ledger log entries.

internal/application/check/checker.go
```diff
@@ -2524,6 +2571,39 @@ func (c *Checker) verifyAuditHashChain(
...
+			// — but not all: archived register(K), archived revoke(K), then a live
+			// reference item pointing at register(K) would put K back in the
+			// expected set and report a false SIGNING_KEY_MISMATCH against a
+			// healthy store. Hence the skip rather than relying on idempotence.
+			//
+			// Not covered by a dedicated test: reaching it needs an upgraded store
+			// with archived chapters AND a reference item straddling the boundary.
+			// Keep the skip; do not "simplify" it away on the idempotence argument.
+			for _, item := range items {
+				if item.GetLogSequence() == 0 || item.GetLogSequence() <= signing.archiveEndSeq {
```
NumaryBot: 🟠 **[major] Legacy replay audit items can cause false SIGNING_KEY_MISMATCH**

On upgraded stores that still contain pre-f9ee1e829 per-order idempotency replay audit items, a replay item can reference any earlier live log, not only logs at or below the archive boundary. If a signing key is registered, later revoked, and a duplicate legacy proposal records a ReferenceSequence pointing to the original register log, this loop re-applies the register after the revoke and produces a false `SIGNING_KEY_MISMATCH` against a healthy store. The archive-boundary skip alone is insufficient; the same `[MinLogSequence, MaxLogSequence]` range guard used by `collectExpectedSkippable` is needed here.

**Suggestion:** Apply the same `[MinLogSequence, MaxLogSequence]` range guard used by `collectExpectedSkippable` when folding signing orders so that out-of-range legacy replay items are skipped before being applied to the expected signing state.

internal/application/check/signing.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+	// Oldest first: a revoke in a later chapter must see the keys the earlier
+	// chapters registered, or it would delete nothing and leave the key expected.
+	slices.SortFunc(archived, func(a, b *commonpb.Chapter) int {
+		return cmp.Compare(a.GetCloseAuditSequence(), b.GetCloseAuditSequence())
+	})
+
+	for _, ch := range archived {
+		coldPebble, err := coldReader.GetReader(ctx, ch.GetId())
```
NumaryBot, line 319: 🟠 **[major] Partial archive fold leads to false signing-key mismatch on coverage failure**

If an earlier archived chapter is read and folded successfully but a later chapter fails to load or decode, the verifier retains the keys/config already accumulated from the prefix and uses them as the expected state during comparison, while only appending an incomplete-coverage event. A healthy store where the unreadable later chapter would have revoked one of those earlier keys will therefore generate a false signing-key mismatch, incorrectly flagging a clean store as tampered.

**Suggestion:** When archived coverage is detected as incomplete after a partial fold, clear or suppress the partially-built expectations so that the pass reports only a coverage gap and does not treat the prefix replay as the authoritative full archived truth.

paul-nicolas:
## Both new findings fixed in cf67985

Both were real, and both were the same class of bug: reporting a mismatch against an expectation that cannot support it, flagging a healthy store as tampered.

### Legacy replay references (checker.go)

Correct, and the fix was already sitting in the adjacent code. `collectExpectedSkippable` bounds its fold by `success.Min/MaxLogSequence` and documents exactly why that is the right discriminant — those fields are computed only over freshly-created logs, never over reference sequences, so a legacy back-reference falls below `min`. My signing fold used the archive boundary alone, which does nothing when nothing is archived. The fold now applies the same window.

Worth recording *why* this was easy to get wrong, since my original comment reasoned about it and still landed short: register is an upsert and revoke is a delete, so each order is individually idempotent. What is not idempotent is the **ordering** — register(K) at log 5, revoke(K) at log 10, then a reference back to log 5 re-registers K after its revocation. I had used the idempotence argument to justify only the archive-boundary skip; it does not extend to arbitrary back-references.

Pinned by `TestVerifyAuditHashChain_SigningFoldIgnoresLegacyReplayReferences`, which also asserts the entry's *own* fresh item still folds, so the guard cannot pass by over-skipping. One fixture note: `persistAuditEntry` always hashes with a nil previous hash, so a multi-entry fixture built on it breaks the chain from the second entry — and a mismatching entry is not folded, which made the revoke silently never apply and the test pass for the wrong reason before I chained it properly.

### Partial archive fold (signing.go)

Also correct, and it went further than I had reasoned. I had handled this *within* a chapter — `foldChapter` failing downgrades the whole fold — but across chapters I returned early and left the accumulated prefix in place as the expectation. That prefix is unsound in **both** directions: a revoke in an unread chapter leaves its key expected (reported missing), and a register in an unread chapter leaves its row unexpected (reported injected).

Incomplete coverage now skips the key and config comparisons entirely and emits only `SIGNING_VERIFICATION_INCOMPLETE`. Suppressing detection for that run is the honest outcome — a mismatch we cannot substantiate is worse than an admitted gap, and it is the same principle as the empty-audit path folding cold storage rather than assuming an empty expectation. The malformed-row class stays outside the suppression: a row too short to decode is a fact about that row and needs no oracle.

This one **replaced a test of mine** that asserted the old behaviour — it pinned the false positive as correct. Removing that assertion is part of the fix, not collateral.

Both mutation-verified: with each guard removed the corresponding test fails, and restored, passes. `just pre-commit` 0 issues, full `just test` exit 0 with 0 FAIL.

### On the two re-posted empty-audit findings

The first two inline comments on this revision restate the empty-audit gap, which `86dc80a` fixed one commit earlier (covered by `TestCheck_SigningProjections_EmptyAuditWiring`, and the `CheckStore On an empty store` e2e spec passes on both commits). I read them as re-posts of the unresolved finding ID rather than fresh analysis. Flagging in case they are meant to indicate something the fix misses — happy to look again if so.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1669#issuecomment-5178419486

internal/application/check/signing.go
```diff
@@ -0,0 +1,478 @@
...
+		if err := item.UnmarshalVT(value); err != nil {
+			return fmt.Errorf("unmarshaling archived audit item %x: %w", iter.Key(), err)
+		}
+
+		// Failure-side items carry LogSequence=0 and changed no state, so they must
+		// not contribute to the expected set. Sound because buildAuditItems
+		// (state/audit.go:57-78) sets LogSequence only from a CreatedLog sequence,
+		// failure entries carry no logs, and NextSequenceID starts at 1
+		// (state/fsmstate.go:62) so 0 is never a real log sequence.
+		if item.GetLogSequence() == 0 {
```
NumaryBot: 🟠 **[major] Legacy replay audit items in archived chapters cause false SIGNING_KEY_MISMATCH**

When cold storage contains legacy pre-f9ee1e829 per-order idempotency replay audit items, the archived chapter fold applies every non-zero AuditItem.LogSequence without the [MinLogSequence, MaxLogSequence] window guard used in the live verifier. If a signing key was registered, later revoked, and a legacy replay item references the original register log, the archived fold re-applies the registration after the revocation, producing a false SIGNING_KEY_MISMATCH on a perfectly healthy store.

**Suggestion:** Apply the same [MinLogSequence, MaxLogSequence] range guard used by collectExpectedSkippable (and the live signing fold) when folding signing orders from archived chapters, so out-of-range legacy replay items are skipped before being applied to the expected signing state.

paul-nicolas:
## Archived fold window fixed in 92d1e4e

Correct, and I had left exactly half the job done: `86dc80a`/`cf67985` put the `[Min,Max]` window on the live fold and did not carry it to `foldChapter`, which scanned `SubColdAuditItem` rows directly and applied every non-zero `LogSequence`. Same bug, same mechanism, cold side.

This half is arguably the worse one. Pre-boundary signing state has no second audit-derived source — the baseline checkpoint is a verbatim copy of the projection under test, which is why this pass reads cold storage in the first place — so a key resurrected by an archived replay reference has nothing downstream to contradict it.

`foldChapter` now walks archived audit **entries** and reads each entry's items (`query.ReadAuditEntries` → `query.ReadAuditItems`, the shape `collectChapterIdempotency` already uses) rather than scanning item rows. That is not incidental: only the entry carries the fresh-log window, and only the entry states the outcome. Scanning items alone forces both decisions onto per-item proxies — and it had been using `LogSequence == 0` as a failure-side proxy, which the entry walk now replaces with a direct `GetSuccess() == nil` check.

Covered by `TestSigningVerifier_FoldArchivedIgnoresLegacyReplayReferences`, the cold twin of the live test, again asserting the entry's own fresh item still folds so the guard cannot pass by over-skipping. Mutation-verified.

One knock-on worth noting for review: `signingArchivedSST` had to start emitting audit **entries** alongside items. An items-only archived fixture now folds nothing at all, which is a sharp edge for anyone writing a cold-storage signing test later — the helper documents it.

### Status of the other findings on this revision

The three re-posted findings (empty-audit ×2, partial archive fold) were fixed in `86dc80a` and `cf67985`; `TestCheck_SigningProjections_EmptyAuditWiring` and the incomplete-coverage subtests cover them, and both empty-store e2e specs pass. That is three revisions in a row re-posting resolved finding IDs, so I am reading them as sticky rather than fresh — but I would rather say so out loud than quietly ignore a review comment. If any of them is pointing at something the fixes still miss, say which and I will look again.

`just pre-commit` 0 issues both modules, full `just test` exit 0, 0 FAIL, `application/check` observed green.

internal/application/check/signing.go
```diff
@@ -0,0 +1,490 @@
...
+	case *raftcmdpb.SystemScopedOrder_RevokeSigningKey:
+		revoke := payload.RevokeSigningKey
+
+		revoked := []string{revoke.GetKeyId()}
+		if revoke.GetCascade() {
+			// Re-derived from the expected parent map, never read off the log's
+			// cascaded_key_ids: re-deriving the cascade is the entire point of
+			// this pass, and trusting the recorded list would make a tampered
+			// projection able to justify itself.
+			revoked = append(revoked, v.descendantsOf(revoke.GetKeyId())...)
```
NumaryBot, line 178: 🟠 **[major] Cascade reparenting within a proposal diverges from FSM, producing false SIGNING_KEY_MISMATCH**

When a signed ApplyBatch proposal simultaneously re-registers an existing child signing key under a new parent and cascade-revokes the original parent, the checker's fold applies the child's new parent assignment to `v.keys` before computing `descendantsOf` for the cascade revocation. The FSM's `GetSigningKeyChildren` operates on committed children and only filters pending removals, so the child is still included in the cascade and deleted at merge. Because the checker has already updated the parent pointer, `descendantsOf` omits the child from the cascade set, leaving it in the expected state while the real store has deleted it — a healthy store is then reported as having a missing key.

**Suggestion:** Snapshot the current `v.keys` state before applying any intra-proposal parent reassignments, and use that snapshot (rather than the partially-updated state) when computing `descendantsOf` for cascade revocations within the same proposal fold, mirroring the FSM's committed-state semantics.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1669#issuecomment-5178419486

paul-nicolas:
## Cascade edge-set fixed in 052dc21 — good catch, and it invalidated a claim I had written down

Verified against `state.WriteSet.GetSigningKeyChildren` and the analysis holds. It returns the **union** of two relations: the revoked key's committed children (filtering only keys the same proposal *removed*) and the keys that same proposal *registered* under it. It never consults a reassigned parent pointer to exclude a key. So a key re-registered under a new parent in the same proposal as a cascade revoke of its old parent is still revoked, and `descendantsOf` — walking only the running pointer — dropped it, left it expected, and would report a false `SIGNING_KEY_MISMATCH` against a store that legitimately deleted it.

What made this easy to miss is a claim in my own comment: "the reachable *set* is fully determined by the parent relation." That is true of traversal *order*, which is what I was arguing about, and false of the *edge set* once the two relations diverge. The comment now separates the two explicitly, because the next reader is otherwise entitled to the same wrong conclusion.

**Fix**, close to your suggestion with one adjustment. `beginProposal` snapshots the pre-proposal parent relation once per audit entry — one entry is one proposal, matching the FSM's single `WriteSet.Reset` per proposal, which is what makes "committed" mean the same thing on both sides — and `descendantsOf` accepts a key whose *either* snapshotted or current parent matches. Keys the proposal already removed are absent from the expectation, which reproduces the pending-removal filter for free. The snapshot is built lazily (`ensureProposalSnapshot`) so the vast majority of audit entries, which carry no signing order, cost one boolean instead of a map copy.

**The adjustment matters:** the snapshot must also be refreshed on the **revoke** path, not just where a reassignment happens. A cascade in a proposal that registered nothing would otherwise walk the snapshot left behind by an earlier proposal and delete a key whose reparenting has since been committed — the mirror-image false positive, in the other direction. I only found that because I wrote the counterpart test ("reparent in an earlier proposal does *not* cascade") alongside the one for your scenario; it failed on my first attempt at the fix.

Both guards are mutation-verified independently: dropping the union edge fails only the same-proposal subtest, dropping the revoke-path refresh fails only the earlier-proposal subtest. `just pre-commit` 0 issues, full `just test` exit 0, 0 FAIL. `AGENTS.md` invariant #8 and the checker page both corrected.

### The three re-posted findings on this revision

Empty-audit (×2) and partial-archive-fold are fixed in `86dc80a` / `cf67985`; I verified against `git show HEAD:` rather than the revision label, and the covering tests are `TestCheck_SigningProjections_EmptyAuditWiring` and the incomplete-coverage subtests of `TestSigningVerifier_Compare`. Reading them as sticky finding IDs — this cascade finding was genuinely new in the same batch, so I am checking each one rather than dismissing the set.

internal/application/check/signing.go
```diff
@@ -0,0 +1,555 @@
...
+		visited     = map[string]struct{}{keyID: {}}
+	)
+
+	for len(queue) > 0 {
+		current := queue[0]
+		queue = queue[1:]
+
+		// Collected into a slice rather than deleted in place: mutating v.keys
+		// while ranging over it is what this loop must not do.
+		for candidate, expectation := range v.keys {
```
NumaryBot, line 225: 🟡 **[minor] Cascade reparenting within a proposal can diverge from FSM, producing false SIGNING_KEY_MISMATCH**

When a signed ApplyBatch proposal simultaneously re-registers an existing child signing key under a new parent and cascade-revokes the original parent, the checker's fold may apply the child's new parent pointer to `v.keys` before computing `descendantsOf` for the cascade revocation. The FSM's `GetSigningKeyChildren` operates on committed children and maintains a proposal-local `pendingRemovals` set filtering prior removals, so the child is still included in the cascade and deleted at merge. If the checker has already updated the parent pointer, `descendantsOf` omits the child from the cascade set, leaving it in the expected state while the real store has deleted it — a healthy store is then incorrectly reported as having a missing key. Similarly, within a single proposal where a child is first revoked then re-registered before the cascade fires, the checker and FSM may diverge on whether that child survives.

**Suggestion:** Snapshot the current `v.keys` state before applying any intra-proposal parent reassignments, and use that snapshot (rather than the partially-updated state) when computing `descendantsOf` for cascade revocations within the same proposal fold. Also track a proposal-local pending-removals set mirroring the FSM's semantics so that a key revoked earlier in the same proposal is not re-included in the cascade just because it was subsequently re-registered.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1669#issuecomment-5178419486

PR #1672, opened 2026-08-04
refactor(EN-1551): remove dead proto fields, drop all reserved, realign tags

paul-nicolas:
Cleanup pass sanctioned by the v3-unreleased policy: nothing persists across versions, so wire and storage formats are free to change.

## Dead code removed

Found with a type-aware `go/packages` analyzer — selectors, composite-literal keys and type references resolved through `go/types`, loaded with every build tag and with test files. 27 of 1344 fields had zero references:

- **`raft_cmd.proto`** — `GenerationSnapshot` and its 7 entry messages (`VolumeAttributeSnapshotEntry`, `MetadataAttributeEntry`, `LedgerAttributeEntry`, `BoundaryAttributeEntry`, `TransactionReferenceAttributeEntry`, `TransactionStateAttributeEntry`, `IdempotencyKeyAttributeEntry`): a self-contained island referenced only by each other. One was already marked `Deprecated`.
- **`common.proto`** — `IdempotencyEntry`, plus `Index.last_built_at` / `Index.last_error`, documented in `openapi.yml` but never populated.

## Reserved and renumbering

All **68 `reserved` statements** across 6 files deleted, together with the comments that existed only to justify them. Every "keep reserved forever" note justified itself by compatibility with *already-persisted* data — exactly what the policy waives.

**132 declarations renumbered** to `1..N`; the `ErrorReason` enum gap at 59 closed.

## The one real hazard this uncovered

`internal/infra/plan/predicted_index.go` hand-rolled the wire encoding of `Proposal.predicted_index` with a **hardcoded tag 7**, to avoid re-marshalling a multi-megabyte `Proposal` while holding the proposal lock. Renumbering moved that field to 5, so the append started writing a `fixed64` into field 7 — `idempotency`, a length-delimited message.

Build, `vet` and lint all stayed silent. The events integration tests caught it (`proto: wrong wireType = 1 for field Idempotency`). Fixed at the root: the number is now read from the descriptor at package init, so the encoding follows any future renumbering automatically.

Reviewers: this is the class of breakage worth a second look — grep for `protowireutil` / `protowire.AppendTag` before approving any future renumbering.

## Also removed

`TestDatabricksProtoFieldNumbers` pinned `DatabricksSinkConfig`'s wire numbers, justified by "silently corrupts already persisted/replicated sink configs after upgrade" — the exact compatibility concern the v3-unreleased policy waives, and directly contradictory to this change.

## Verification

- **Descriptor gate**: exactly the 9 intended messages removed, no field name lost or gained anywhere else, every message `1..N`. All enum value names preserved, values sequential from 0.
- Renumbering is **idempotent** — a second run reports 0 changes.
- `go vet` clean under each of `kafka,nats,clickhouse,databricks,s3,azure,pyroscope`, `e2e`, `scenario`, `antithesis`.
- Full unit suite green (80 packages).
- `just pre-commit` reports 0 issues for both the main module and the operator.
- Docs updated: `AGENTS.md` (reserved-policy note + invariant #8's "reserves tags 3-10"), `indexes.md`, `http-api.md`, `openapi.yml`.

## Not covered

E2E / scenario / antithesis suites were **not executed** — they need a live cluster and Docker, so they only got a compile + vet check. Since this rewrites the gRPC wire format and the audit-hash / idempotency input bytes, those are worth watching in CI before merge.

## Note on the analysis

Two analyzer bugs produced false positives mid-pass, both caught by checking candidates against real code rather than deleting blind: oneof arms used only as a type-switch case (`case *pkg.Msg_Arm:` — no field is ever selected on an empty payload message like `DropAction`), and `GoCamelCase` digit handling (`oauth_m2m` → `OauthM2M`, not `OauthM2m`). Worth knowing if this analysis is ever repeated.


internal/infra/plan/predicted_index.go
```diff
@@ -16,5 +34,5 @@ func AppendProposalPredictedIndex(data []byte, index uint64) []byte {
 		return data
 	}
 
-	return protowireutil.AppendFixed64(data, 7, index)
+	return protowireutil.AppendFixed64(data, predictedIndexField, index)
```
Azorlogh, line 46: No test asserts this append equals marshalling with `PredictedIndex` set — the descriptor init heals the field number, but a wire-type change (fixed64 → varint) would still corrupt silently. Worth a small equivalence test.

AGENTS.md
```diff
@@ -7,7 +7,7 @@ This document contains rules and conventions for AI agents working on this codeb
 **CRITICAL**: The Raft-based ledger v3 has never been released. No deployment carries data across versions, and all state is wiped on every push to `release/v3.0`. Because nothing persists across versions, there is no backward-compatibility burden — break wire and storage formats freely:
 
 - **No compatibility shims.** Do not add migrations, version guards, or fallback paths to preserve old wire or storage formats. Change the format and move on.
-- **Do not reserve proto fields.** When removing a field, delete it and realign the remaining field numbers sequentially (see [Protocol Buffers](#protocol-buffers)). Pre-existing `reserved` declarations may stay — a dedicated cleanup pass will remove them before release.
+- **Do not reserve proto fields.** When removing a field, delete it and realign the remaining field numbers sequentially (see [Protocol Buffers](#protocol-buffers)). The cleanup pass that removed every pre-existing `reserved` declaration has run (EN-1551): there are none left in `misc/proto/`, and none should be reintroduced.
```
Azorlogh, line 10: Worth adding here (or in protobuf.md): before renumbering, grep for `protowire.Consume*` too — `indexbuilder/protowire_postings.go` hardcodes field numbers on the read side, which the `protowireutil`/`AppendTag` grep misses.

Azorlogh (commented):
Verified the renumbering independently (descriptor cross-check protoc vs generated registry: 0 mismatches; every hand-rolled wire site audited; no numeric ErrorReason usage anywhere) — code looks sound. Two doc stragglers:

- `docs/technical/architecture/subsystems/storage/storage.md:214-230` still quotes the removed `GenerationSnapshot` as the cache snapshot format (the snapshotter persists per-key vtproto values via `protoSnapshotSlot`).
- `docs/technical/architecture/subsystems/admission/pipeline.md:93` and `docs/technical/architecture/subsystems/fsm/preload.md:76` still say predicted_index is appended as "proto field 7" — better to drop the number entirely now that the encoding is descriptor-driven.

internal/infra/plan/predicted_index.go
paul-nicolas: Fixed in 2dcfe52 — added `internal/infra/plan/predicted_index_test.go`.

Your threat model was exactly right, so the core assertion is equivalence against canonical marshalling: `Id` is field 1 and `predicted_index` is field 5, so vtproto's field-order output makes the appended bytes the tail of the canonical encoding, and the test asserts strict byte equality against `(&Proposal{Id: …, PredictedIndex: idx}).MarshalVT()` over a table of index values. A `fixed64 → varint` change breaks that immediately.

Three supporting assertions: the appended suffix is decoded and required to carry `protowire.Fixed64Type` under the descriptor's field number (so a regression names the wire-type invariant rather than only showing a byte diff); `fd.Kind()` is required to still be `Fixed64Kind`, pointing at the file that hardcodes it; and the documented `index == 0` no-op is covered.

Mutation-verified — swapping `AppendFixed64` for a varint append fails the equivalence test (tag `0x29` vs `0x28`).

AGENTS.md
paul-nicolas: Fixed in d83511d — documented in both places: a new "Renumbering: audit the hand-rolled wire sites first" section in `protobuf.md` (with both greps and a table contrasting the two sites), plus a pointer in the `AGENTS.md` Protocol Buffers section so it is discoverable from the file agents actually load.

Worth recording that this PR is a **near miss** on it. The fields renumbered here are `Log.receipt` (6→3), `Log.response_signature` (7→4) and the high `LogPayload` oneof members (17-28 → 16-26). `protowire_postings.go` reads only `Log` fields 1-2 and `LogPayload` fields 2-3 — all unchanged — so it stays correct by which fields happened to move, not by construction.

The doc calls out why the two sites differ: `predicted_index.go` derives its number from the descriptor so the number self-heals (only the wire type is hardcoded, now test-pinned), while `protowire_postings.go` hardcodes both and fails silently — `ConsumeFieldValue` skips the unrecognised field and it yields empty postings, no error, no failing build.

paul-nicolas:
Both doc stragglers fixed in 65dd6af, and the branch is rebased onto `release/v3.0` (picking up `c35ae99b0`, EN-1515).

**`storage.md:214-230`** — confirmed: `GenerationSnapshot` has zero hits left in the tree. Replaced the prose and the proto block with the actual mechanism: one Pebble row per cache entry, keyed `[ZoneCache][gen][cacheType][U128]`, value in the lean format `[8-byte tag LE][1-byte flag][vtproto value]` with per-type marshalling from `protoSnapshotSlot`. Also documented the tombstone rule (header only — the pre-delete payload is deliberately not marshalled, or `Del` would resurrect the entry on restore) and that `parseLeanValue` fails closed on malformed rows. The zone `0x02` reference was correct and stays (`ZoneCache`, `dal/store.go:205`).

**`predicted_index` field number** — you were right that it should go, and it had actually become wrong rather than merely redundant: this PR renumbers the field to **5**, and 7 is now `common.Idempotency`. Dropped the literal in both docs rather than updating it to 5, per your suggestion — the encoding reads the number from the descriptor, so restating it in prose just recreates the drift. Kept the `fixed64` detail, since that part *is* hardcoded and is now pinned by the test in 2dcfe52. One note: the occurrence is at `preload.md:193`, not :76 (line 76 is unrelated `Coverage` prose).

**Rebase detail worth flagging** — the only conflict was `AGENTS.md` invariant #8. Resolved by taking your EN-1515 text (the `signingVerifier.compare` section) and re-applying this PR's single edit in that paragraph: dropping the "`LedgerBoundaries` now reserves tags 3-10 for them and" clause. `bucket.proto` and `bucket.pb.go` auto-merged, but I regenerated all protos from scratch rather than trust a text-merge of generated files. EN-1515 only appended `CheckStoreErrorType` values 21-23, which this PR never renumbered, so there was no tag collision.

`just pre-commit` (0 lint issues, no codegen drift) and `just test` both green.

docs/technical/architecture/subsystems/storage/storage.md
```diff
@@ -211,23 +211,19 @@ The snapshot captures the complete in-memory FSM cache state:
...
+
+Tombstones are exactly the 9-byte header with no payload -- the pre-delete value is deliberately not marshalled, since `AttributeCache.Del` only flips the `Deleted` flag and keeps the payload in memory, which would otherwise resurrect the entry on restore. On the read path `parseLeanValue` fails closed on any malformed row (short buffer, unknown flag byte, or a tombstone carrying trailing bytes) rather than restoring partial or forged cache state (EN-1527).
+
+Reversions are **not** part of the cache snapshot. They are stored per-word in Pebble zone `0x03` (`ZonePerLedger` + `SubPLReversions`) and reconstructed from Pebble via `ReadReversions` on startup or snapshot restore.
 
 The snapshot contains the full attribute cache state (volumes, metadata, ledger info, boundaries, references, transaction state) serialized per generation, allowing fast in-memory restoration.
```
Azorlogh: Stale leftover: this closing sentence still says "serialized per generation", contradicting the no-per-generation-envelope rewrite above.

paul-nicolas: Fixed in 3ba6d91ec — dropped the sentence rather than rewording it.

Confirmed stale, and for the reason you name: "serialized per generation" is `GenerationSnapshot`-era wording, describing one proto blob per generation. Nothing serializes a generation as a unit any more — the snapshotter writes one row per cache entry and the generation is only a key component. It survived `65dd6af` because it sits after the reversions paragraph, outside the block that was replaced.

Deleted rather than reworded because the only other thing in it was the cache-type list, already stated one paragraph up where `protoSnapshotSlot`'s per-type marshalling is described — a reworded version would just restate it. The `Reversions are **not** part of the cache snapshot` paragraph now closes the section.

Checked the rest while I was in there: `GenerationSnapshot` has zero hits left in the tree, and this was the last "per generation" occurrence in the snapshot section. `storage.md:319` ("Rotated per generation", zone table) is not stale — generations genuinely do rotate — so it stays.

paul-nicolas:
Also closed the codecov gap flagged above, in 2f37ba5a5.

The two uncovered lines were the invariant panic in `predicted_index.go` — the guard for `predicted_index` having been dropped from `raft_cmd.proto`. It sat inside a package-level `var` initializer, so it was unreachable from any test: `fd` is nil only if the field is gone, and by then the package will not load.

Extracted the lookup into `lookupPredictedIndexField(md)`, taking the descriptor instead of hardcoding `Proposal`'s. Exercising the panic is now just a matter of passing a message that lacks the field — `CloseChapterOrder`, since it is empty and so cannot drift into carrying one. The panic message names the descriptor it was handed rather than asserting `raft.Proposal`.

`predicted_index.go` now has no uncovered block. Worth noting the guard was not merely uncovered but untested in the stronger sense: nothing verified it fired at all, or what it said when it did.

`just pre-commit` (0 lint issues, no codegen drift) and `just test` both green on 2f37ba5a5.

Azorlogh (approved):
Renumbering verified independently (descriptor cross-check protoc vs generated registry: 0 mismatches at 2f37ba5a; all hand-rolled wire sites audited; no numeric enum/field usage anywhere outside generated code). All review findings addressed. LGTM.

gfyrag (approved):

PR #1674, opened 2026-08-05
feat(operator): auto-expand Ledger v3 PVCs

flemzord:
## Summary

- add a strictly opt-in `autoExpansion` policy to WAL and data volume specs
- add a separate `VolumeExpansionReconciler` that measures each replica through `ledgerctl cluster disk-usage --json`, with no Prometheus server dependency
- converge replica PVC requests, honor resize/cooldown state, cap growth at `maximumSize`, and recover from partial patches without in-memory state
- expose per-volume collection validity and freshness, and reject failed, zero-capacity, missing-timestamp, or older-than-one-minute measurements
- expose Kubernetes events and Operator metrics, regenerate CRD/RBAC/chart artifacts, and document EBS CSI prerequisites and diagnostics

## Safety properties

- never shrinks a PVC or exceeds `maximumSize`
- never changes `Cluster.spec.persistence.*.size`, StatefulSet volume claim templates, or PVs
- rejects hostPath and cold-cache policies and validates expandable StorageClasses through an uncached API reader
- reads PVCs through the uncached API reader, matching the intentionally watch-free PVC RBAC contract
- preserves last successful disk bytes only for diagnostics after `Statfs` failure; invalid or stale values cannot trigger expansion
- evaluates WAL and data validity independently, so one failed volume does not discard a valid sample for the selected volume
- derives required health peers from the committed Raft configuration, not the gRPC connection pool
- prevents invalid, stale, missing-pool, or unreachable health samples from clearing an existing disk write gate
- does not treat an incomplete below-threshold measurement as healthy; a confirmed threshold crossing can still proceed while emitting a warning
- leaves existing installations unchanged unless `autoExpansion.enabled` is explicitly set

## Validation

- `bash scripts/agent-check` — pass on the final tree
- targeted health, node, bootstrap, disk collector, gRPC, CLI, and Operator tests — pass, including race detection where applicable
- canonical PR normalization, generation, root/operator lint, build, and product packages under `go test -race ./...` — pass
- the canonical local gate still encounters the unchanged fixed-200ms `scripts/internal/testenv/TestRunSynchronizedTimeoutTerminatesAndReportsEveryPeer` load flake; the test passes 10/10 in isolation under `-race`
- GitHub CI at `eda02a574` — all checks pass, including unit, Operator, model, scenario, Schemathesis, Antithesis workload, business/cluster E2E, coverage aggregation, and PR image build
- Codecov patch coverage — 93.71% (80% target); project coverage — 77.26% (+0.12% versus `release/v3.0`)
- NumaryBot — no findings; all review threads resolved
- Chainsaw `volume-auto-expansion-contract` scenario against Kind

The EKS staging canary remains an operational follow-up because no staging target was available in this workspace.

Jira: [EN-1749](https://formance-team.atlassian.net/browse/EN-1749)
Parent: [EN-867](https://formance-team.atlassian.net/browse/EN-867)

[EN-1749]: https://formance-team.atlassian.net/browse/EN-1749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


misc/operator/internal/controller/volume_expansion.go
```diff
@@ -0,0 +1,548 @@
...
+	}
+}
+
+func (r *VolumeExpansionReconciler) loadVolumePVCs(
+	ctx context.Context,
+	ledger *ledgerv1alpha1.Cluster,
+	volume string,
+) ([]*corev1.PersistentVolumeClaim, []volumePVCState, error) {
+	replicas := int32(3)
+	if ledger.Spec.Replicas != nil {
```
NumaryBot: 🟠 **[major] PVC expansion uses spec replica count instead of live StatefulSet replicas**

When the Cluster spec is updated to an invalid replica count (e.g., 2), ClusterReconciler rejects the spec and leaves the existing 3-replica StatefulSet running. However, this reconciler still enumerates PVCs using the rejected spec replica count. If disk usage crosses the expansion threshold in that state, only ordinals 0 and 1 are patched while the still-running ordinal 2 PVC remains at the old size. This breaks per-replica convergence and can cause ordinal 2 to hit the disk gate.

**Suggestion:** Gate expansion on the same spec validation used by ClusterReconciler, or better yet, enumerate PVCs from the actual live StatefulSet's current replica count (e.g., read the StatefulSet's `.status.replicas` or `.spec.replicas` directly) rather than trusting the potentially-rejected Cluster spec value.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1674#issuecomment-5196242775

misc/operator/internal/controller/volume_expansion.go
flemzord: Fixed in fe97aa00c. The reconciler now reads the live StatefulSet once and uses its spec.replicas for both PVC enumeration and pod measurements. The regression test deliberately sets Cluster.spec.replicas=2 while the live StatefulSet remains at 3 and verifies that all three PVCs are expanded. Operator pre-commit, unit tests, envtest, and lint pass.

misc/operator/internal/controller/volume_expansion.go
```diff
@@ -0,0 +1,556 @@
...
+		}
+		target := *resource.NewQuantity(decision.TargetBytes, resource.BinarySI)
+		logger.Info("requested PVC expansion", "target", target.String())
+		r.recordNormalf(ledger, "VolumeExpansionRequested", "%s PVCs expanding from %s to %s after reaching %.1f%% usage", definition.Name, formatBytesAsQuantity(decision.LargestRequestBytes), target.String(), decision.MaxUsageRatio*100)
+		volumeRequestedBytesMetric.WithLabelValues(ledger.Namespace, ledger.Name, definition.Name).Set(float64(decision.TargetBytes))
+		volumeExpansionsMetric.WithLabelValues(ledger.Namespace, ledger.Name, definition.Name, "requested").Inc()
+
+		return true, nil
+	case volumeExpansionDecisionNone:
+		logger.Info("volume expansion not required")
```
NumaryBot, line 318: 🟠 **[major] PVC enumeration driven by spec replica count instead of live StatefulSet replicas**

The volume expansion reconciler enumerates PVCs using the replica count from the Cluster spec. If the Cluster spec is updated to an invalid value and rejected by ClusterReconciler, the live StatefulSet continues running with the original (higher) replica count. Only the subset of PVCs matching the rejected spec count are expanded, leaving running replicas unprotected against disk exhaustion. Similarly, during a StatefulSet scale-up, desired ordinals exist in the spec before their PVCs are actually created; `loadVolumePVCs` fails on the first missing PVC and skips expansion for the entire group, so already-running replicas remain at the old size past the threshold.

**Suggestion:** Drive PVC enumeration from the actual live StatefulSet's current replica count (read `.status.replicas` or `.spec.replicas` directly from the StatefulSet object) rather than from the Cluster spec. This covers both the rejected-spec scenario and the scale-up window where not-yet-created PVCs would otherwise abort the loop for existing replicas.

misc/operator/api/v1alpha1/cluster_types.go
```diff
@@ -1130,10 +1130,17 @@ type VolumeSpec struct {
...
 	AccessMode string `json:"accessMode,omitempty"`
 
-	// Size of the volume.
+	// Size is the initial and minimum requested capacity of the volume. Automatic
+	// expansion updates the live PVCs only and never rewrites this field or the
+	// StatefulSet VolumeClaimTemplates.
 	// +optional
 	Size resource.Quantity `json:"size,omitempty"`
 
+	// AutoExpansion configures opt-in automatic growth for PVC-backed WAL and
```
NumaryBot, line 1141: 🟠 **[major] autoExpansion field included in pod-template hash, causing unnecessary pod rolls**

The new `autoExpansion` field is serialized by `computeSpecHash` because only `Persistence.DeletionProtection` is currently excluded. Any change to `autoExpansion` settings therefore mutates the StatefulSet pod-template annotation and triggers a rolling restart of all Ledger pods, even though this policy is consumed solely by the separate PVC-patching reconciler and has no effect on the pod template itself.

**Suggestion:** Exclude the `autoExpansion` field from `computeSpecHash` (the same way `DeletionProtection` is excluded) so that tuning the expansion policy does not cause unnecessary pod rolls.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1674#issuecomment-5196242775

misc/operator/internal/controller/volume_expansion.go
flemzord: The current head already reads StatefulSet.spec.replicas in Reconcile and passes that live count to both PVC enumeration and pod measurement; Cluster.spec.replicas is no longer used for either. The regression test sets the rejected Cluster spec to 2, keeps the live StatefulSet at 3, and verifies all three PVCs are patched. During scale-up, a missing expected PVC intentionally blocks the whole group and retries after one minute, matching the approved safety contract: the controller must not expand only a subset while the live group is incomplete.

misc/operator/api/v1alpha1/cluster_types.go
flemzord: Fixed in 8dba7b3a3. computeSpecHash now excludes AutoExpansion for WAL, data, and cold-cache alongside DeletionProtection. TestComputeSpecHash_ExcludesNonPodFields verifies that changing WAL/data expansion policies leaves the pod-template hash unchanged. Operator pre-commit, all unit tests, and lint pass.

misc/operator/internal/controller/volume_expansion.go
NumaryBot, line 318: 🟠 **[major] PVC enumeration driven by spec replica count instead of live StatefulSet replicas**

The volume expansion reconciler enumerates PVCs using the replica count from the Cluster spec rather than from the live StatefulSet. This creates two failure modes: (1) if the Cluster spec is updated to an invalid value and rejected by ClusterReconciler, the live StatefulSet keeps running at the original replica count while only the smaller spec-derived subset of PVCs is expanded, leaving running replicas unprotected against disk exhaustion; (2) during a StatefulSet scale-up, desired ordinals exist in the spec before their PVCs are physically created, so the first missing PVC causes `loadVolumePVCs` to abort the entire expansion loop and skips protection for already-running replicas. Although a prior comment claimed this was fixed in fe97aa00c, the inline thread at line 279 was re-raised and is still open.

**Suggestion:** Drive PVC enumeration from the actual live StatefulSet's current replica count by reading `.spec.replicas` (or `.status.replicas`) directly from the StatefulSet object rather than from the Cluster spec. Also handle the scale-up window by tolerating NotFound for higher ordinals while still expanding PVCs for replicas that already exist.

misc/operator/api/v1alpha1/cluster_types.go
NumaryBot, line 1141: 🟠 **[major] autoExpansion field included in pod-template hash, triggering unnecessary rolling restarts**

The new `autoExpansion` field is serialized by `computeSpecHash` because only `Persistence.DeletionProtection` is currently excluded from the hash. Any change to `autoExpansion` settings therefore mutates the StatefulSet pod-template annotation and triggers a rolling restart of all Ledger pods, even though this policy is consumed solely by the separate PVC-patching reconciler and has no effect on the pod template itself.

**Suggestion:** Exclude the `autoExpansion` field from `computeSpecHash` in the same way `DeletionProtection` is already excluded, so that tuning the expansion policy does not cause unnecessary pod rolls.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1674#issuecomment-5196242775

misc/operator/internal/controller/volume_expansion.go
flemzord: This is a duplicate of the resolved thread and does not match the current head. Reconcile fetches the live StatefulSet, derives liveReplicas from StatefulSet.spec.replicas, and passes that value into both loadVolumePVCs and collectMeasurements; Cluster.spec.replicas is not used by this controller for enumeration. The regression test proves the rejected-spec case with Cluster=2 and StatefulSet=3. Tolerating a missing expected PVC and expanding only a subset is intentionally rejected by the approved contract: an incomplete live group must remain unchanged and retry, preserving equal replica requests.

misc/operator/api/v1alpha1/cluster_types.go
flemzord: This is also a duplicate and does not match current head 8dba7b3a3. computeSpecHash explicitly sets Persistence.WAL.AutoExpansion, Persistence.Data.AutoExpansion, and Persistence.ColdCache.AutoExpansion to nil before serialization. TestComputeSpecHash_ExcludesNonPodFields sets WAL/data policies and asserts that the hash remains unchanged. The latest remote Dirty, Tests, Build, and E2E jobs all pass.

gfyrag:
### Finding majeur : `VolumeExpansionReconciler` peut patcher des PVC pour une spec `hostPath + autoExpansion`

`VolumeExpansionReconciler` ne réutilise pas la validation qui rejette `hostPath + autoExpansion`. `enabledVolumeExpansionDefinitions` ne vérifie que `AutoExpansion.Enabled`, alors que le rejet `hostPath` vit uniquement dans `validateVolumeSpec` côté `ClusterReconciler`.

Scénario problématique : un cluster existant en PVC reçoit une update invalide ajoutant `hostPath + autoExpansion`. `ClusterReconciler` parque bien la CR en erreur, mais l’ancien StatefulSet/PVC reste actif, et le reconciler d’expansion peut encore charger, mesurer puis patcher ces PVC. Ça contredit la propriété annoncée “rejects hostPath policies”.

Suggestion : filtrer les volumes `hostPath` dans `enabledVolumeExpansionDefinitions`, ou réutiliser la même validation avant toute mesure/patch dans `VolumeExpansionReconciler`. Il faudrait aussi ajouter un test de régression où des PVC existants restent inchangés et où `ReadDiskUsage` n’est pas appelé quand la spec combine `hostPath` et `autoExpansion`.

gfyrag:
### Note d’architecture : deux reconcilers sur le même CRD

Le fait d’avoir plusieurs reconcilers qui prennent `Cluster` comme ressource primaire n’est pas interdit en soi, mais c’est un pattern à manier avec prudence. Il peut être acceptable ici parce que les responsabilités semblent séparées : `ClusterReconciler` gère le workload, les ressources déclaratives et le statut du `Cluster`, tandis que `VolumeExpansionReconciler` ne devrait piloter que l’état opérationnel des PVC live.

La contrepartie, c’est que les frontières d’ownership doivent être très nettes et que les deux reconcilers doivent partager les mêmes garde-fous de contrat. Sinon ils peuvent interpréter différemment la même spec. C’est exactement le risque visible ici : `ClusterReconciler` considère `hostPath + autoExpansion` invalide, mais `VolumeExpansionReconciler` peut encore agir parce qu’il ne réutilise pas cette validation.

Je ne pense donc pas que le split en deux reconcilers soit forcément un problème à lui seul, mais il impose au minimum ces invariants :

- `ClusterReconciler` reste le seul writer de `Cluster.status` et du StatefulSet/template.
- `VolumeExpansionReconciler` reste le seul writer des requests PVC liées à l’auto-expansion.
- toute validation liée à `autoExpansion` est factorisée et réutilisée par les deux chemins avant toute action externe.
- un reconciler secondaire ne doit jamais agir sur une spec que le reconciler principal parque comme invalide.

Dans ce cadre, la correction du bug `hostPath + autoExpansion` devrait être vue comme une correction de contrat entre reconcilers, pas seulement comme un guard local.

misc/operator/internal/controller/volume_expansion.go
```diff
@@ -0,0 +1,576 @@
...
+	ledger *ledgerv1alpha1.Cluster,
+	volume string,
+	replicas int32,
+) ([]*corev1.PersistentVolumeClaim, []volumePVCState, error) {
+	pvcs := make([]*corev1.PersistentVolumeClaim, 0, replicas)
+	states := make([]volumePVCState, 0, replicas)
+	for ordinal := range replicas {
+		name := fmt.Sprintf("%s-%s-%d", volume, resourceName(ledger.Name), ordinal)
+		var pvc corev1.PersistentVolumeClaim
+		if err := r.Get(ctx, types.NamespacedName{Namespace: ledger.Namespace, Name: name}, &pvc); err != nil {
```
NumaryBot: 🔴 **[blocker] Grant watch permission for cached PVC reads**

`r.Get` uses the manager's cached client, which must establish a PVC informer using both list and watch. The generated RBAC grants PVC `get`, `list`, and `patch` but not `watch`, so in a real RBAC-enabled cluster these reads fail and auto-expansion never reaches measurement or patching. Add `watch` to the PVC RBAC marker and regenerated manifests, or use an uncached reader for PVCs.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1674#issuecomment-5196242775

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1674#issuecomment-5196242775

misc/operator/internal/controller/volume_expansion.go
flemzord: Fixed in d676cb679. `loadVolumePVCs` now requires and uses `APIReader` for every PVC read, so the watch-free PVC RBAC contract is valid. `TestLoadVolumePVCsUsesDirectAPIReader` proves the cached client may lack the PVC while the direct reader succeeds, and also checks the missing-reader invariant. The full Operator suite passes.

internal/infra/health/healthcheck.go
```diff
@@ -173,26 +180,39 @@ func (hc *HealthChecker) check(stop <-chan struct{}) {
...
+		walError:    localWAL.Error,
 		dataUsed:    localDataUsed,
 		dataTotal:   localDataTotal,
 		dataPercent: safePercent(localDataUsed, localDataTotal),
+		dataValid:   localDataValid,
+		dataAge:     sampleAge(now, localData.ObservedAt),
+		dataError:   localData.Error,
 	})
 
 	// Check peers dynamically from the service pool
```
NumaryBot: 🔴 **[blocker] Derive health samples from cluster membership**

When a committed member fails `ConnectionPool.AddPeer` (for example, TLS probing or dialing fails), it is absent from `PeerIDs()` and contributes no invalid sample. Consequently, a previously active disk write gate can be cleared using only the remaining nodes, despite the new requirement that an unreachable member must not clear it. Enumerate authoritative cluster membership and treat members missing from the service pool as invalid samples.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1674#issuecomment-5196242775

internal/infra/health/healthcheck.go
flemzord: Fixed in eda02a574. The health checker now enumerates Node.MemberIDs() from the atomic committed Raft ConfState (including voters, learners, outgoing voters, and learners-next), then uses the service pool only for connectivity. A committed member missing from the pool therefore contributes an invalid sample and cannot clear an existing disk gate. The regression test starts with healthy local usage, member IDs [7,8], an empty pool, and a blocked gate, and proves the gate remains blocked.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1674#issuecomment-5196242775

internal/infra/health/healthcheck.go
```diff
@@ -267,14 +356,35 @@ func (hc *HealthChecker) check(stop <-chan struct{}) {
...
-	if prev := hc.gate.Load(); prev != nil {
+	var (
+		prevDiskBlocked   bool
+		hasCurrentVerdict bool
+	)
+	if prev := hc.gate.Load(); prev != nil && prev.leadershipEpoch == leadershipEpoch {
 		prevDiskBlocked = prev.diskBlocked
+		hasCurrentVerdict = true
+	}
+	if !hasCurrentVerdict && !allSamplesValid(samples) {
```
NumaryBot: 🔴 **[blocker] Allow a healthy quorum to resume writes after failover**

When leadership changes while any committed member is unavailable, `hasCurrentVerdict` is false and that member's invalid sample makes this branch leave the gate nil indefinitely. Consequently, a healthy quorum cannot accept writes after an ordinary node failure until the unavailable member returns or membership is changed; the modified leadership-transfer test confirms this behavior. Invalid samples should prevent clearing an existing disk block, but should not synthesize a permanent block when the new leader has no evidence that disks crossed the threshold.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1674#issuecomment-5196242775

internal/infra/health/healthcheck.go
flemzord: Fixed in 3e3274a16. A new leader now stays fail-closed until its first cluster poll has fresh local WAL and data samples, then derives the initial verdict from every fresh sample available. An unavailable remote member no longer creates a synthetic permanent block, while any fresh remote sample above threshold still blocks writes. Within the leadership epoch, invalid or unreachable samples still cannot clear an observed disk block. Unit coverage now proves both the unavailable-peer failover path and the invalid-local fail-closed path, and the leadership-transfer e2e scenario passes.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1674#issuecomment-5196242775

flemzord:
Addressed the current review feedback in 5b9840070 and 3e3274a16:

- fenced disk-gate verdicts by leadership epoch, require fresh local evidence before reopening, and preserve quorum availability when a remote committed member is unavailable;
- made the envtest expansion trigger deterministic and added optimistic locking for PVC patches;
- reject future cooldown timestamps as decision evidence, keep the target annotation diagnostic, and expose/retire freshness-aware operator metrics;
- consolidated the volume and sample-validity contracts, removed unreachable branches and dead accessors, and added overflow-safe target arithmetic.

Validation is green on the current head: all GitHub checks pass, including unit/race, Operator, E2E Business, E2E Cluster, model, scenarios, Schemathesis, Antithesis workload, coverage, and PR image build. NumaryBot re-reviewed 3e3274a16 with no findings.

PR #1675, opened 2026-08-06
perf(admission): reuse static Numscript discovery in bulks

flemzord:
## Summary

- memoize repeated, variable-free Numscript discovery within one admission request after proving that both volume and metadata read sets are empty
- keep per-order coverage and intra-batch effect folding active for every reused discovery
- add a reproducible admission benchmark, document the safety boundary, and give asynchronous mirror E2E assertions enough reconciliation time

The transaction JSON-view optimization previously included in this PR has been removed to keep the change scoped to admission. A future adapter-owned REST DTO and isolated encoding benchmark are tracked by [EN-1752](https://formance-team.atlassian.net/browse/EN-1752).

## Performance

The focused 50-order admission benchmark improves from 1,106,562 ns/bulk to 288,750 ns/bulk (-73.9%), with 69.1% fewer allocated bytes and 80.5% fewer allocations.

The benchmark isolates `Admission.resolveScriptsAndEnrichNeeds`. No end-to-end throughput claim is retained in this PR because the earlier A/B result also included the removed JSON optimization.

## Impact by execution path

“No discovery reuse” below means the original per-order admission path remains active. Every accepted order is still preloaded and executed by the FSM.

| Execution path | Numscript discovery | Expected impact |
|---|---|---|
| Bulk with repeated, variable-free script; same ledger, text, and `force`; empty `ReadVolumes` / `ReadMetadata` | First occurrence runs fully; later exact matches reuse it | Largest gain |
| Bulk with unique scripts | Full path for every order | No direct effect |
| Bulk with external variables | Full path for every order | No direct effect; bindings never share discovery |
| Bulk with state reads such as `balance()`, `meta()`, or a bounded balance-checked source | Full path for every order against the evolving overlay | No direct effect; state-dependent scripts are never reused |
| Different ledger, script text, or `force` value | Separate first discovery for each key | Later exact matches can reuse |
| Unitary request | Full path; there is no later match | No direct effect |
| FSM, Raft, Pebble writes, audit, storage, or response encoding | Unchanged | No effect |

## Safety

- memoization is request-local and keyed by ledger, script text, and force mode
- variable-bearing or state-reading scripts keep the complete discovery path
- coverage and intra-batch effect folding still execute for every order
- no HTTP schema, JSON encoding, protobuf, audit, storage, or FSM behavior changes

Targeted regression coverage proves that:

- two reused static deposits contribute two effects, so a later `balance()` observes 200 rather than 100
- an identical bounded-source script is rediscovered against the evolving balance and cannot contribute a phantom second effect
- an identical metadata-reading script sees an intervening metadata update
- identical script text with different external variables discovers both resolved destinations

## Validation

- `GOROOT= go build ./...`
- targeted admission and HTTP regression tests
- `BenchmarkResolveScriptsWorldToBankBulk50`
- GitNexus impact and change detection: LOW risk, no affected indexed execution flow

The original branch validation also covered the root test suite, operator tests, E2E business/cluster suites, and scenario suites.

[EN-1752]: https://formance-team.atlassian.net/browse/EN-1752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/application/admission/admission.go
```diff
@@ -1848,31 +1856,59 @@ func (a *Admission) resolveScriptsAndEnrichNeeds(ctx context.Context, orders []*
...
+			staticCandidate = true
+			staticKey = staticNumscriptDiscoveryKey{
+				ledger: ledgerName,
+				script: scriptText,
+				force:  createTx.CreateTransaction.GetForce(),
+			}
+			discovered = staticDiscoveries[staticKey]
+		}
+
+		if discovered == nil {
```
paul-nicolas, line 1888: [Major] The memo-hit branch has zero test coverage

I instrumented this branch with a `println` and ran the whole `./internal/application/admission` suite: **0 executions**. The suite passes without ever taking the path this PR adds.

- `TestReusableNumscriptDiscovery` (`numscript_intrabatch_test.go:16`) tests only the predicate against hand-built `DiscoveryResult` structs; it never calls `resolveScriptsAndEnrichNeeds`.
- `BenchmarkResolveScriptsWorldToBankBulk50` does reach the branch, but asserts nothing about correctness.
- Existing batches that repeat a script (`numscript_skip_parity_test.go:265,278,289`) all use `dependentBalanceScript(...)`, which is balance-reading and therefore memo-ineligible by design — so they exercise the miss path only.

Why this deserves more than the usual missing-test note: the reuse path is exactly where a `plan.Coverage` key could be dropped, and CLAUDE.md invariant #6 says a volume-preload miss turns an FSM read into a **silent no-op** that no checker pass detects (`usagestore` is outside checker scope). The safety argument also leans on numscript's `ResolveDependenciesCoversRuntime`, which does exist (`numscript@v0.0.25.../internal/interpreter/resolve_dependencies_test.go:43`) but is an upstream table test over a handful of cases, not a property proof — and this repo pins no assertion of its own.

To be clear: I verified the implementation **is** correct (see the review summary), so this is a regression guard, not a bug report.

Concrete fix — a differential test. I wrote this one, ran it, and confirmed it hits this branch exactly twice and passes on this branch:

- Batch A: 3 identical static scripts (memo hits orders 1 and 2) followed by a `balance(@bank)` order.
- Batch B: the same, but each static script gets distinct trailing newlines, so the memo key differs and every order resolves fully.
- Assert the per-order `plan.Coverage` fingerprints are equal for each index, and that the trailing order's `InputsResolutionHash` matches — the latter is what proves effects are folded exactly once per order rather than zero or twice.

internal/adapter/http/handlers_bulk.go
```diff
@@ -284,7 +284,11 @@ func writeBulkResponse(w http.ResponseWriter, elements []*servicepb.BulkElement,
 			if payload := log.GetData(); payload != nil {
 				switch p := payload.GetPayload().(type) {
 				case *commonpb.LedgerLogPayload_CreatedTransaction:
-					data = p.CreatedTransaction.GetTransaction()
+					// Feed the transaction's plain JSON view to the outer streaming
+					// encoder. Passing the proto pointer would invoke MarshalJSON for
+					// every bulk element, recursively allocating intermediate buffers
+					// for postings and post-commit volumes.
+					data = p.CreatedTransaction.GetTransaction().JSONView()
```
paul-nicolas: [Medium] Bulk responses now HTML-escape strings and sort map keys

Switching from the proto pointer to the plain view moves encoding from the inner marshaler to the outer streaming encoder, and the two use different sonic configs. Probed directly:

```
sonic.Marshal      (old inner path) = {"zeta":"a<b&c>d","alpha":"x","mid":"y"}
ConfigStd.Encoder  (new outer path) = {"alpha":"x","mid":"y","zeta":"a\u003cb\u0026c\u003ed"}
```

Previously the transaction was pre-marshalled by `Transaction.MarshalJSON` -> `sonic.Marshal` (`ConfigDefault`: no HTML escaping, no key sort) and embedded raw. Now `writeJSONResponse` -> `json.MarshalWrite` -> `sonic.ConfigStd.NewEncoder` (`internal/adapter/json/json.go:30`) encodes the struct directly, so metadata values containing `<`, `>` or `&` come back escaped, and `metadata` / `postCommitVolumes` keys are now sorted.

Both forms decode to identical values, so conformant clients are unaffected, and deterministic key order is an improvement. Two things still make it worth a decision rather than a silent side effect:

1. The **unitary** transaction endpoint still goes through `MarshalJSON` -> `sonic.Marshal`, so the two endpoints now disagree on escaping for the same resource.
2. `bulk-performance.md` enumerates the distinctions the view preserves and omits this one, which suggests it was not deliberate.

Fix: either document the escaping/ordering change alongside the other preserved distinctions, or route the unitary path through `JSONView()` too so both endpoints agree.

paul-nicolas (commented):

**1 Major, 1 Medium, 4 Minor/Nit.** 5 findings entered adversarial verification: 1 confirmed, 4 corrected downward, 0 rejected. Reviewed at head `c22929ff3`; `GOROOT= go build ./...` passes and CI is green.

**The core optimization is correct.** Rather than take the doc's word for it, I verified the memoization safety argument end to end and it holds:

- **No aliasing corruption.** `mergeDiscovery` -> `addBalanceDelta` does `new(big.Int).Set(delta)` on first write and `existing.Add(existing, delta)` afterwards (`numscript_source.go:120-132`), so the memoized `big.Int`s are only ever read. Coverage building only iterates the maps and calls `key.Bytes()` fresh per order, so there is no `Coverage.Add` ownership violation.
- **The reuse gate is sound.** `admissionValueSource` exposes exactly two read methods (`numscript_source.go:164,196`), and `ReadVolumes`/`ReadMetadata` come from the resolver's declared dependency sets — strictly stronger than `InputsHash != nil`, since under `force` a bounded read is declared but not recorded (`store.go:82,181`). Empty declared reads plus `ResolveDependenciesCoversRuntime` means the effect execution (`SafeRun` with a non-recording store, `discover.go:159`) cannot observe the overlay.
- **No stale-`err` bug.** `err` is the loop-scoped variable from `predictOrderSkip` (`admission.go:1681`); nothing reads it after the discovery block, and a memo hit simply does not reassign it.
- **Error fall-through is pre-existing.** When `classifyResolutionFailure` returns `(false, nil)`, `discovered` is nil, `reusableNumscriptDiscovery(nil)` is false, and the `if discovered != nil` guard skips the rest — same as base.
- **Coverage and effect folding run per order on a hit** (`admission.go:1923-1964`); `predictOrderSkip`, `foldCallerAccountMetadata`, reference recording and numscript-reference preloads all stay outside the memo block.
- **Invariant #10 respected** — only `order.Technical` is written, which is excluded from both the idempotency hash and the audit business-intent projection.
- **JSON field-for-field equivalence** — `postingJSON` matches `Posting.MarshalJSON` exactly (`posting.go:15-28`); `postCommitVolumeJSON` matches `VolumeEntry.MarshalJSON` including `input`/`output` as strings and always-emitted `color` (`common.pb.json.go:224-237`); empty snapshots emit `{}` in both (`common.pb.json.go:200`); the `*uint64` -> `uint64 omitempty` ID change is equivalent; amounts stay unquoted decimals (`uint256.go:92`).

- `internal/proto/commonpb/transaction.go:210` — nil-slice divergence, **unreachable today.** Old `flat[addr] = va.GetVolumes()` emitted `null` for a nil slice, and a nil `*VolumeEntry`/`*Posting` element emitted `null`; the new value-struct views emit `[]` and a zero-valued object. But `processor_volumes.go:71-81` only ever creates an account key together with an appended non-nil entry, and proto decoding never yields nil elements in a repeated message field, so no production input reaches these shapes. Worth a note now that `JSONView` is exported. (Codex rated this major; downgraded on the reachability evidence.)
- `internal/application/admission/numscript_bench_test.go:22` — the `orders` slice is built once outside `b.Loop()` while `resolveScriptsAndEnrichNeeds` stamps `order.Technical` (`admission.go:1957-1960`), so from iteration 2 the 50 `&OrderTechnical{}` allocations are skipped. Everything else (overlay, effects, memo map, coverage) is rebuilt per iteration and the parse cache persists in both arms, so the reported -73.9% is not materially distorted — but resetting `order.Technical = nil` per iteration would make it exact. The benchmark also covers only the hit path, so a miss-path regression would be invisible.
- `tests/e2e/cluster/mirror_test.go:337` (+6 sites) — timeout doubling is scope creep. `defaultPollInterval = 5s` (`internal/application/mirror/worker.go:29`), so 15s allowed ~3 poll cycles and 30s allows ~6; defensible hardening on a loaded runner, but it lands unexplained in a perf PR and doubles worst-case failure latency. Codex read it as masking a regression — I don't think that is supported, since this change touches neither the mirror worker nor its poll loop. Suggest splitting it out, or adding a one-line comment naming the 5s poll interval as the reason.
- `internal/proto/commonpb/transaction.go:188` — `JSONView() any` boxes the value because `transactionJSON` is unexported; a concrete exported return type would avoid the interface and document the shape.
- `docs/technical/architecture/subsystems/admission/bulk-performance.md` — the "plain Go values / no nested custom marshalers" framing is only partly accurate: `postingJSON.Amount` is still `*Uint256`, whose `MarshalJSON` (`uint256.go:92`) is invoked per posting.
- `internal/application/admission/admission.go:1641` — `ctx` is unused (the linter flags it), but it was already unused at base `ab36444a1`. Pre-existing, not this PR.

gfyrag:
After looking at this beyond the ideal microbenchmark, I don't think the complexity/benefit trade-off is justified.

Parsing is already cached across requests by `NumscriptCache`. This PR only memoizes dependency discovery and exploratory execution within one bulk, and only when all of the following hold:

- same ledger, exact script text and `force` value;
- no variables;
- no balance reads;
- no metadata reads;
- multiple occurrences in the same request.

That seems like a fairly marginal production shape. Real Numscript workloads generally use variables and/or read state, while a fully static transfer such as the benchmark's `@world -> @bank` can be sent more simply as postings.

The ideal bulk-of-50 result is good for the admission subphase, especially for allocations, but it does not remove any work from Raft, the FSM, Pebble, or the audit path. At the throughput we actually need to serve (around 10k sustained TPS already covers nearly all customers), this is unlikely to produce a meaningful user-visible improvement.

In exchange, it introduces another semantic reuse rule into an already delicate area involving intra-bulk effects, preload and per-order coverage. Its correctness also relies on `ReadVolumes` and `ReadMetadata` remaining an exhaustive definition of state dependence as Numscript discovery evolves.

My recommendation would be not to merge this optimization unless production telemetry demonstrates that repeated, variable-free, state-independent scripts are common enough—and create enough CPU/GC pressure—to justify maintaining this extra path. Otherwise this looks more like optimizing a favorable benchmark than solving a likely customer bottleneck.

PR #1676, opened 2026-08-06
perf: improve global ledger transaction throughput

flemzord:
## Summary

This stacked PR broadens the optimization from the repeated-Numscript case in #1675 to work performed by most transaction paths.

It:

- keeps small preload coverage sets inline and promotes to a map only after 8 entries;
- enables the volume Bloom filter by default for 1,000,000 expected keys and avoids dirtying unchanged Bloom blocks;
- replaces the account-grouped post-commit volume snapshot with a flat deterministic protobuf and builds it while postings mutate volumes, removing a second coverage-gated read pass;
- decodes only the audit fields required by the usage projection for the common native transaction path, with a generated-protobuf fallback for uncommon orders;
- buffers successful bounded HTTP bulk responses, avoiding one flush/write per result.

The REST response shape remains account-grouped. The protobuf/storage shape changes intentionally because Ledger v3 is unreleased.

## Why

Profiles showed that the remaining cost was spread across per-order allocations, duplicate volume reads, protobuf decoding of unused fields, first-touch Pebble lookups, and HTTP response writes. These costs affect native transactions with hot or unique accounts, rather than relying on repeated Numscript text.

## Benchmarks

Fresh store per run, 3-run elapsed-time medians, 4 logical CPUs, 100 VUs, bulk size 50. Every run completed with 0 request errors and projection lag 0. “Gain” is throughput gain: `baseline elapsed / candidate elapsed - 1`.

| Path | Transactions | Baseline | Candidate | Throughput gain |
| --- | ---: | ---: | ---: | ---: |
| Native, hot/reused accounts | 400,000 | 10,600 ms | 8,457 ms | **+25.3%** |
| Native, unique accounts | 400,000 | 19,489 ms | 14,340 ms | **+35.9%** |
| Native, 5 postings with unique volumes | 100,000 | 15,973 ms | 11,858 ms | **+34.7%** |
| Distinct Numscript text on every transaction | 100,000 | 11,162 ms | 10,344 ms | **+7.9%** |

### Paths that improve

- Native transactions, whether accounts are reused or unique.
- Transactions with several postings/volume mutations.
- First access to missing/new volumes while the Bloom filter is correctly sized.
- The usage projection for native create-transaction audit entries.
- Successful HTTP bulk responses.

### Paths with limited or no direct effect

- A different Numscript on every transaction still has to be parsed and executed independently; the measured gain is only +7.9%.
- Reusing the same stored Numscript benefits primarily from the parent PR #1675, not from these global changes.
- The Bloom filter has no direct benefit once all touched volumes are already hot in cache, and a disabled or heavily undersized/saturated filter remains a correctness-safe fallback with little/no performance gain.
- gRPC and non-bulk HTTP endpoints do not use the new bulk response buffering.
- Uncommon/non-create usage orders deliberately fall back to the full generated protobuf decoder.
- Coverage sets and post-commit snapshots above the small inline thresholds promote to maps; they remain correct but do not receive the full allocation benefit.

Experiments with a shared Numscript parser cache, parallel FSM parsing, and exact Bloom occupancy tracking were rejected because they were neutral or regressive in end-to-end measurements.

## Correctness coverage

- Inline and promoted coverage representations, hash collisions, and coverage-bit gating.
- Inline and promoted post-commit accumulators, latest-write semantics, colors, deterministic ordering, actual coverage misses, native/Numscript/revert/mirror paths.
- Selective decoder parity against generated protobufs, malformed/non-canonical wire data, unknown fields, oneofs, scratch reuse, and fallback paths.
- Bloom duplicate additions and concurrent access under the race detector.
- REST JSON compatibility and bounded bulk marshal failures.

## Validation

- `GOROOT= go build ./...`
- `go test ./...`
- Full unit suite under the race detector (the `just test` phase of `just test-all`)
- `go test -race -tags e2e -p 1 ./tests/e2e/business/... ./tests/e2e/cluster/... -timeout 20m`
  - 555/555 business specs passed
  - 266/266 cluster specs passed
- `go test -race -tags scenario ./tests/scenarios/... -timeout 300s`
- Antithesis workload module compilation (build-only)
- Local model checker — `HEALTH_THRESHOLD=0.99 just test-model 60`: PASS, single node, no findings
- Local model checker — `HEALTH_THRESHOLD=0.99 just test-model-cluster 180`: PASS, 3 voters, 4 restart cycles (including leader failures), no findings
- Hosted Antithesis k8s campaign — 60 minutes, full 62-driver workload, tag `pr-1676-global-perf`: **COMPLETED / NOT GREEN**, run `baeb8d253938464c95c2554dfed8ab22-58-11`.
  - One exercised behavioral property produced 2 counterexamples: `posting count must be >= transaction count` (`transactionCount=252`, `postingCount=15`).
  - Diagnosis: this workload assertion predates the usage-counter cutover and is no longer valid at arbitrary points in time. `TransactionCount` is synchronous FSM boundary state, while `PostingCount` is an eventually-consistent `usagestore` projection that may lag by one builder tick. The counterexample occurred during a write burst and does not identify a Ledger state/FSM divergence.
  - Coverage limitations: `singleton_driver_scaling_structured` was never started and 9 commands did not complete at least once; the image-recency infrastructure check also failed because the newly built images reported an epoch creation timestamp.

The VM filesystem was already above the server's normal 80% write-block threshold. For the successful e2e/scenario reruns, the test harness threshold was temporarily set to 99%; the production code and branch were unchanged, and the validation worktree was restored clean afterward. The local model-checker scripts used their documented `HEALTH_THRESHOLD` input for the same host constraint.

## Stack

Base: `feat/ledger-bulk-performance` / #1675. Retarget this PR to `release/v3.0` after the parent merges.


internal/application/usagebuilder/protowire_order.go
```diff
@@ -0,0 +1,393 @@
...
+	)
+
+	for len(data) > 0 {
+		num, typ, n := protowire.ConsumeTag(data)
+		if n < 0 {
+			return nil, false, errors.New("protowire: invalid tag in Order")
+		}
+		data = data[n:]
+
+		if num == 1 {
```
NumaryBot: 🟠 **[major] Decode the actual Order field numbers**

Canonical native transaction orders encode `ledger_scoped` as field 1 and `technical` as field 3, but this scanner treats field 1 as technical and field 3 as a oneof member. Consequently `parseUsageCreateOrder` never matches the common native transaction shape and always falls back to full protobuf decoding, defeating the primary usage-builder optimization.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1676#issuecomment-5206019791

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1676#issuecomment-5206019791

PR #1678, opened 2026-08-07
fix(EN-1748): cross-store read consistency — aligned acquisition + versioned metadata event keys

Azorlogh:
> [!NOTE]
> Supersedes #1677. Commit 1 is the acquisition fix; commit 2 the append-only event-key layer; commit 4 wires that layer through the index builder and the query compile path, so the whole read path now serves exact membership at one pinned sequence. Both reproduction e2e tests are in the branch and green.

## The bug ([EN-1748](https://formance-team.atlassian.net/browse/EN-1748))

Filtered queries open one handle on the main store and one snapshot on the read store, with no guarantee the two observe the same applied index. Some filter leaves read the main store, index leaves read the read store, and enrichment reads the main store again — so a response can mix memberships that **no single committed state ever had**.

Concretely, both directions are reproducible in seconds under plain sustained load, no fault injection:

- **A query filtered on `tier=gold` returns an account whose own metadata reads `tier=silver`.** Selection runs on the read-index at its fold cursor; enrichment reads the main-store handle; a tier flip folded between the two makes the returned row contradict the very filter that selected it.
- The complement direction: `not(ts[_,_])` over a **READY** timestamp index returns committed transactions as "missing from the index" while the fold lags (caught repeatedly by the model driver in #1659; also inherited by the PIT draft's filtered selection, see [#1662 comment](https://github.com/formancehq/ledger/pull/1662#issuecomment-5205419265)).

Both live in `tests/e2e/business/` as red→green regression tests: `cross_store_snapshot_alignment_test.go` (lag direction — red on `release/v3.0` at probe 1, green with commit 1) and `cross_store_value_skew_test.go` (flip direction — red with commit 1 alone at probe 33, green with the event keys wired):

```
[FAILED] probe 33: tier=gold returned load:0:61:b with tier="silver"
         — index selection and enrichment disagree on one row
```

## The design

1. **Metadata** index keys carry the raft sequence as a suffix; mutations only ever **append**:

   ```
   [prefix][encodedValue][entity]\x00[raftSeq BE][op]      op ∈ {ADD, DEL}
   ```

   The exists index (eidx) gets the same treatment keyed on its nullFlag; events are emitted only when the flag transitions.

2. A transition away from a value appends a **DEL event in the old value's own range** — so "does E match V at sequence P?" is answered entirely inside (V, E)'s adjacent event group: latest event ≤ P decides. Ranges never consult each other; iteration stays one ordered scan.

   ```
   seq 10:  set  k=red  on E   →  [red][E][10][ADD]
   seq 25:  set  k=blue on E   →  [blue][E][25][ADD] + [red][E][25][DEL]
   seq 40:  unset k     on E   →  [blue][E][40][DEL]

   pin 15 → red ✓        pin 30 → blue ✓ (red's own range says DEL@25)
   ```

3. A query takes the **main-store handle first**, keeps its applied index as the pin, waits (bounded, condvar) for the read-store fold to reach it, then resolves metadata / exists leaves at that pin. Events newer than the pin are invisible; events older are complete — selection and enrichment can no longer disagree.

4. A background job reclaims superseded events **below the watermark of live read leases** (registered per query by `AlignedIndexSnapshot`, floored by the fold cursor): within a group, only the latest below-watermark event may survive, and only as an ADD. Verdicts at every admissible pin are provably unchanged — that's a test.

5. **Add-only limbs keep their plain keys — no seq suffix.** Rows in timestamp, inserted_at, address→tx (all roles), reference, and has-asset are never revoked, so fold-ahead overshoot can only *add* rows — the per-row keep against the pinned main handle (`MainHorizonKeep`) restores exactness with no format change. Metadata is different because a revocation also *removes* rows, and a missing row is nothing a post-filter can resurrect — hence the event history.

## How it is wired

- **Fold**: the builder declares each log's sequence (`SetEventSequence`) before dispatch; `ReplaceMetadataIndexV` / `DeleteMetadataEntryWithPreviousV` append stamped events instead of put/delete. A same-value re-set writes nothing (membership unchanged — also what makes a schema rewrite racing a live dual-write at the same sequence harmless). The rmap keeps its role and prompt deletion; the checker's reverse-map pass is untouched.
- **Schema-rewrite backfill** stamps its v_pending events with the FSM sequence it reads from; the atomic switch is already gated on the read-store cursor reaching that sequence, so every queryable pin at v_pending sees them.
- **Query**: `Compile` carries the pin; point conditions stream through the group-resolve iterator (absolute-seek contract intact), int/uint ranges group-resolve then materialize+sort exactly as before, exists conditions resolve the eidx flag ranges. Add-only leaves are untouched.
- **GC**: an incremental, budgeted zone walk in the builder tick (`GCEventZone`, resume cursor, group-boundary stops), gated by the lease watermark.
- **Introspection** (`InspectIndex` distinct/facets/summary) resolves live groups instead of counting raw keys.

## What's in the diff

| Commit | Content |
|---|---|
| `fix(EN-1748): align read-index snapshots…` | The acquisition foundation: ordering + bounded wait + frozen-checkpoint exemption + horizon trim for the add-only limbs. |
| `feat(readstore): versioned event keys…` | The storage layer: key layout (NUL entity terminator), append-only write shapes, group-resolve leaf, read leases, GC — each unit-tested. |
| `test(e2e): red-to-green load regression…` | Lag-direction reproduction (`not(ts[_,_])` under fold pressure). |
| `feat(EN-1748): resolve metadata and exists index leaves…` | Everything under "How it is wired", the doc updates, the unit-test conversions — plus a latent-panic fix from commit 1 (filtered `AggregateVolumes` wrapped the deliberately-nil ACCOUNTS horizon keep unguarded; exposed by the first full-suite run). |
| `test(e2e): flip-direction regression…` | `cross_store_value_skew_test.go` joins the suite green. Both load specs run on isolated servers (their backlog would tax every later spec that walks the shared store), and the e2e client now retries the retryable not-caught-up precondition like a production client. |

## Deliberately not here

- Per-limb `MainHorizonKeep` extension for the add-only limbs (`reverted_at` needs the reverted-bitset predicate; ACCOUNTS-target selection is currently untrimmed).
- A max-lease cap: a leaked or stuck lease must not pin GC retention forever — today leases are request-scoped so the exposure is bounded, but the cap plus a persisted GC floor (pin below floor → loud snapshot-too-old) should land before long-lived pinned readers (PIT) do.
- The `EventResolveIterator` seekFloor exhaustion cache (TODO in code) and the registry-generation retry closing the version-switch residual window (see `read-snapshot-consistency.md`).
- Bulk destruction (`DropIndex`) stays an immediate range-delete: snapshot isolation keeps in-flight readers whole, and a drop landing in the acquisition window is an honest compilation rejection, not wrong data.

## Open questions for reviewers

1. Group-resolve cost on every metadata leaf scan: group size is 1 in steady state, 2–3 transiently, and superseded events are GC'd behind the lease watermark. Fine, or do we want a benchmark before merge?
2. GC retention bound: cap policy for leases (fixed duration vs. proportional to fold lag)?
3. Interaction with #1662: the PIT filtered selection would ride the same pinned resolution; its view token could then record the selection sequence next to its watermarks.

[EN-1748]: https://formance-team.atlassian.net/browse/EN-1748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/application/ctrl/controller_default.go
```diff
@@ -965,17 +965,23 @@ func (ctrl *DefaultController) AggregateVolumes(ctx context.Context, ledgerName
...
 
 	indexStart := time.Now()
 
-	iter, err := query.Compile(snap, kb, filter, commonpb.QueryTarget_QUERY_TARGET_ACCOUNTS, ledgerInfo.GetName(), nil, schemaFields, ledgerInfo, query.NewPebbleIndexReader(ctrl.attrs.Index, handle), readstore.SnapshotVersionResolver(snap, ledgerInfo.GetName()), profile, handle)
+	compiled, err := query.Compile(snap, kb, filter, commonpb.QueryTarget_QUERY_TARGET_ACCOUNTS, ledgerInfo.GetName(), nil, schemaFields, ledgerInfo, query.NewPebbleIndexReader(ctrl.attrs.Index, handle), readstore.SnapshotVersionResolver(snap, ledgerInfo.GetName()), profile, handle)
 	if err != nil {
 		return nil, domain.WrapCompileError(err)
 	}
+
+	iter := readstore.NewFilterIterator(compiled,
```
NumaryBot: 🔴 **[blocker] Avoid wrapping account aggregation with a nil predicate**

For every filtered account aggregation, `MainHorizonKeep(QUERY_TARGET_ACCOUNTS, ...)` returns nil, but `NewFilterIterator` invokes its predicate as soon as the compiled iterator yields a row. Thus any matching filtered `AggregateVolumes` request panics instead of returning a result; guard the wrapper as the other call sites do.

internal/query/aligned_snapshot.go
```diff
@@ -0,0 +1,140 @@
...
+
+			return nil, 0, fmt.Errorf("reading index progress: %w", err)
+		}
+
+		if lastIndexed >= mainSeq {
+			return snap, mainSeq, nil
+		}
+
+		_ = snap.Close()
+
```
NumaryBot, line 130: 🟠 **[major] Ensure the alignment timeout can always wake the waiter**

When the read index remains behind, this relies on `WaitForSequence` honoring the two-second context deadline, but its cancellation goroutine broadcasts the condition variable without holding `progressMu`. Cancellation can therefore occur after the waiter checks `ctx.Err()` but before `Cond.Wait()`, losing the only wake-up and leaving the query blocked indefinitely when no further progress notification arrives.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1678#issuecomment-5215764400

NumaryBot (commented):

The prior account-aggregation panic is fixed, but the active lost-wakeup race remains and can defeat the alignment timeout.

### Findings outside the diff

🔴 **[blocker] Ensure the alignment timeout can always wake the waiter** — `internal/storage/readstore/store.go:762`

When the read index remains behind, cancellation can occur after the waiter checks `ctx.Err()` but before `Cond.Wait()`. Because this broadcast occurs without holding `progressMu`, that wake-up is lost and the supposedly two-second-bounded query can block indefinitely if no later progress notification arrives.




Summary: https://github.com/formancehq/ledger/pull/1678#issuecomment-5215764400

internal/query/aligned_snapshot.go
```diff
@@ -0,0 +1,153 @@
...
+			return nil, 0, nil, fmt.Errorf("reading index progress: %w", err)
+		}
+
+		if lastIndexed >= mainSeq {
+			return snap, mainSeq, lease.Release, nil
+		}
+
+		_ = snap.Close()
+
+		if waitErr := rs.WaitForSequence(ctx, mainSeq); waitErr != nil {
```
NumaryBot, line 131: 🔴 **[blocker] Ensure the alignment timeout can always wake the waiter**

When the read index remains behind, `WaitForSequence` may miss the deadline wake-up because its cancellation goroutine broadcasts without holding `progressMu`. Cancellation can occur after the waiter checks `ctx.Err()` but before `Cond.Wait()`, causing this supposedly two-second-bounded call to block indefinitely if no later progress notification arrives.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1678#issuecomment-5215764400

internal/query/aligned_snapshot.go
```diff
@@ -0,0 +1,153 @@
...
+	// still bounds the index-ahead direction. No GC runs against a frozen
+	// store, so there is no lease to hold.
+	if rs.Frozen() {
+		return rs.NewSnapshot(), mainSeq, func() {}, nil
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), alignWait)
+	defer cancel()
+
+	// The lease is registered BEFORE the snapshot: the GC computes its
```
NumaryBot: 🔴 **[blocker] Synchronize lease acquisition with event GC**

When GC computes a cursor-based watermark before this lease is acquired but commits its deletions after acquisition and before `NewSnapshot`, it can reclaim events required by `mainSeq`. This is possible because the pinned main handle may lag the now-current fold cursor; merely acquiring the lease before snapshotting does not protect against a GC pass that already sampled the lease registry, so a query can resolve incorrect historical membership.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1678#issuecomment-5215764400

internal/query/aligned_snapshot.go
```diff
@@ -0,0 +1,153 @@
...
+	// watermark from registered pins, so the order guarantees no reclamation
+	// between pinning and snapshotting.
+	lease := rs.Leases().Acquire(mainSeq)
```
paul-nicolas: [Major] The event GC can reclaim below a pin that is about to be used

The comment says the lease-before-snapshot order "guarantees no reclamation between pinning and snapshotting", but it addresses the wrong window. The pin *value* is fixed at line 48 (`ReadLastSequence`), while registration happens here at 69. The unguarded gap is 48 → 69.

In that gap `runEventGC` can tick with zero leases and set `watermark = live fold cursor` (`gc.go:246-249`). The cursor can exceed `mainSeq`, because `NewReadHandle()` is a point-in-time snapshot (`dal/reader.go:58-71`) so `mainSeq` is frozen while the fold advances.

Concretely, group `{ADD@50, DEL@70, ADD@90}` with `mainSeq=60` and `watermark=100`: GC drops `ADD@50` and `DEL@70` as superseded and keeps `ADD@90`. A pin at 60 then finds no event `<= 60` and resolves the entity as **not matching**, where the truth at 60 is matching. Silently missing rows, one direction only, no error raised — and undetectable, since readstore index contents sit outside checker scope (invariant #8 scope refinement (a)).

Alignment makes this *more* reachable, not less: when `cursor > mainSeq` the check at line 82 passes on the first iteration, so the dangerous state is the fast path. Snapshot isolation does not help either — the GC batch commits before `NewSnapshot()` at line 72, so the deletion is visible in the snapshot.

Note `gc.go:242-244` states the false premise outright: *"every future pin is at least the fold cursor."* Nothing enforces that.

Fix: give `LeaseRegistry` a monotone reclaim floor published atomically with the watermark it is about to use (a combined `BeginGC(cursor)` under the registry mutex), and make `Acquire` return `ok=false` for a pin below that floor. `AlignedIndexSnapshot` then re-opens a fresh handle and retries instead of serving a truncated result. Publishing the floor *after* the walk reopens the same race.

internal/application/indexbuilder/backfill.go
```diff
@@ -688,6 +691,11 @@ func (b *Builder) processSchemaRewrite(task *schemaRewriteTask, maxEntries int,
 	b.initBatch(batch)
 	committed := false
 
+	// The rewrite stamps events with the FSM state it reads from; the
+	// atomic switch is gated on the read-store cursor reaching that same
+	// sequence, so every queryable pin at v_pending sees these events.
+	b.wb.SetEventSequence(fsmSeq)
```
paul-nicolas, line 705: [Major] A promoted rewrite can resolve as EMPTY at a reachable pin

Every event this rewrite writes is stamped with one `fsmSeq`. The comment says the switch is "gated on the read-store cursor reaching that same sequence, so every queryable pin at v_pending sees these events" — but the gate (`lastSeq >= task.requiredIndexedSeq`, line 929) constrains the **fold cursor**, which says nothing about pins already held by in-flight readers.

`AlignedIndexSnapshot` returns as soon as `lastIndexed >= mainSeq` (`aligned_snapshot.go:82-83`), so a reader's pin can sit strictly below the serving snapshot's cursor. Interleaving: a query reads `mainSeq = 100`; before it takes its snapshot the FSM applies through 150, the fold catches up, and the switch promotes `v_pending` whose rows are all stamped `fsmSeq = 150`. The query resolves the new version at `pin = 100`, sees none of those `ADD@150` events, and returns **zero rows from a fully populated index**. Because the rewrite stamps every entity at the same sequence, an entity the live dual-write never touched has no other event in its group at all.

This is novel relative to the residual window already documented in `read-snapshot-consistency.md:153-162`, and it corrects that paragraph on both counts:
- the window is not "the microseconds between the two acquisitions" — the bounded wait sits exactly between them, so it is up to `alignWait` (2s), and widest during a backfill, i.e. precisely when a promotion is likely;
- the blast radius is not "may miss entries the rewrite carried over" — it is an empty result set for that field.

It is also new with EN-1748: pre-PR the forward index had no sequence dimension, so a `v_pending` row was visible regardless of pin.

Fix: bind the pin to the version rather than only to the fold cursor. Either consult `b.readStore.Leases().Watermark()` in the switch gate and defer promotion while an in-flight pin sits below `requiredIndexedSeq`, or persist the activation sequence alongside `IndexVersionState.CurrentVersion` and have `requireIndexReady` serve the previous version (or return the already-typed `ErrIndexBuilding`) when `pin < activationSeq`. Please correct the doc paragraph either way.

internal/query/executor.go
```diff
@@ -88,34 +88,35 @@ func Execute(
...
+
+	// Check min_log_sequence freshness against the handle: alignment makes
+	// the index snapshot at least as fresh, so the handle's sequence is the
+	// response's consistent state.
+	if req.GetMinLogSequence() > 0 && mainSeq < req.GetMinLogSequence() {
+		return nil, &ErrReadIndexNotCaughtUp{
+			Requested: req.GetMinLogSequence(),
+			Current:   mainSeq,
+		}
+	}
```
paul-nicolas, line 161: [Major] Read-index lag surfaces as HTTP 500 on every REST list endpoint

(Re: the `ErrReadIndexNotCaughtUp` type declared at `executor.go:27-34`, just outside the diff — commenting here at a return site.)

The type is a plain struct: no `domain.Describable`, no `commonpb.NotFoundError`, no gRPC status. The HTTP error handler matches exactly those shapes plus `codes.InvalidArgument`, so this falls through to `writeInternalServerError` — a **500 `INTERNAL_ERROR`** whose message is replaced by a correlation ID.

This PR is what makes it matter. The error is now reachable from every aligned path (`handleListTransactions`, `handleListAccounts`, `handleListLedgerLogs`, `handleAggregateVolumes`, `handleExecutePreparedQuery`) whenever the fold trails past `alignWait`, so a 2-second builder hiccup turns the REST read API into opaque 500s with no reason code and no `Retry-After`. Clients and SDKs read that as a server fault rather than a transient, retryable condition — and this PR's own e2e harness adds a client-side retry because the condition is routine enough to need one.

To be precise about scope: the **gRPC** surface is fine. It maps to `FailedPrecondition` and attaches `errdetails.ErrorInfo{Reason: "READ_INDEX_NOT_CAUGHT_UP"}`, which is well-typed and machine-readable. This is an HTTP-surface gap only.

Fix: implement `Describable` with `KindUnavailable` and reason `READ_INDEX_NOT_CAUGHT_UP`, following the existing `ErrIndexBuilding` / `ErrCheckpointNotReady` precedent. `kindToHTTPStatus` then yields 503, and a `Retry-After` branch alongside the existing `plan.ErrCacheHorizonExceeded` one gives clients the same contract that error already has.

tests/e2e/testutil/server.go
```diff
@@ -62,6 +62,8 @@ func NewGRPCClient(grpcPort int) (servicepb.BucketServiceClient, clusterpb.Clust
 func NewGRPCClientWithRetry(grpcPort int, withRetry bool) (servicepb.BucketServiceClient, clusterpb.ClusterServiceClient, *grpc.ClientConn, error) {
 	opts := []grpc.DialOption{
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithUnaryInterceptor(NotCaughtUpUnaryInterceptor()),
+		grpc.WithStreamInterceptor(NotCaughtUpStreamInterceptor()),
```
paul-nicolas: [Major] The not-caught-up interceptors are installed on every e2e client, not just the load specs

Both interceptors are added to `opts` unconditionally — above and outside the `if withRetry` block that exists for exactly this kind of "should this connection retry automatically" decision. `NewGRPCClient`/`NewGRPCClientWithRetry` is the sole client-construction path behind `SetupSingleNode` and `SetupMultiNodeCluster`, so every e2e spec in the suite silently gained "block up to 90s, polling every 200ms, on FailedPrecondition/read-index-not-caught-up".

Two consequences:

1. **It masks regressions.** A spec that hits this precondition used to fail fast with a clear error. Now a genuine fold-never-catches-up regression in an unrelated code path manifests as a mysterious 90-second hang, then the same error. That is a global behavior change to the shared test client with no opt-out.
2. **It weakens the two new specs.** Both drive sustained apply pressure specifically to keep the fold behind, then run a 25s probe loop that already handles the precondition correctly by `continue`-ing to the next probe. With a 90s internal retry per call, a single probe can outlast the entire test deadline, so "dozens of racing probes" can collapse to one or two — and `Expect(served).To(BeNumerically(">", 0))` can fail as "inconclusive" for timing reasons unrelated to the fix.

Fix: make this opt-in. Add a dial option (or a `MultiNodeOption`) that the two cross-store specs request explicitly, and leave the other specs on the fail-fast client. Their own loops do not need the interceptor.

tests/e2e/testutil/retry_not_caught_up.go
```diff
@@ -0,0 +1,142 @@
...
+			continue
+		}
+
+		if s.sendsDone {
+			if cerr := replacement.CloseSend(); cerr != nil {
+				continue
+			}
+		}
+
+		s.ClientStream = replacement
```
paul-nicolas: [Medium] Stream retry spins against the dead stream and leaks the replacement

On both `continue` paths here — replay failure at 130, and `CloseSend` failure at 135 — `s.ClientStream` is never reassigned to `replacement`. The next loop iteration therefore calls `RecvMsg` on the **original, already-terminated** stream, which keeps returning the same terminal error. Since that error is by construction one `isNotCaughtUp` accepts, the loop does not exit: it burns 200ms sleeps until the 90s deadline and finally returns the original error.

So a genuine replay/`CloseSend` failure is not retried at all. It degrades into a 90-second busy-wait that is indistinguishable from "the server never caught up", which is the one signal this helper exists to report accurately. The freshly created `replacement` is also dropped without being closed or cancelled, leaving a server-side stream alive for the rest of the test.

Fix: pick one. Either adopt the replacement (`s.ClientStream = replacement`) and re-attempt replay/close against the new stream, or treat a replay failure as terminal and `return err` immediately. Silently converting it into a deadline is the worst of the three.

tests/e2e/testutil/retry_not_caught_up.go
```diff
@@ -0,0 +1,142 @@
...
+	notCaughtUpMarker  = "read index has not caught up"
+	notCaughtUpBackoff = 200 * time.Millisecond
+	notCaughtUpWindow  = 90 * time.Second
+)
+
+func isNotCaughtUp(err error) bool {
+	s, ok := status.FromError(err)
+
+	return ok && s.Code() == codes.FailedPrecondition && strings.Contains(s.Message(), notCaughtUpMarker)
+}
```
paul-nicolas: [Medium] Retry classification substring-matches the error message instead of the structured detail

`isNotCaughtUp` gates on `strings.Contains(s.Message(), "read index has not caught up")`. That is the string-matching-for-control-flow pattern the repo conventions rule out (`errors.Is`/`errors.As`, no string matching), and it is avoidable here: the gRPC server already attaches a stable, structured detail for this exact case — `errdetails.ErrorInfo{Reason: "READ_INDEX_NOT_CAUGHT_UP", Domain: "ledger"}` — which this helper ignores in favour of the human-readable text produced by `ErrReadIndexNotCaughtUp.Error()`.

The failure mode is quiet: any future rewording of that message (clarity tweak, added context) stops the interceptor recognising the condition. The retry becomes dead code, the two load specs lose their safety net, and no test fails, because this *is* the safety net.

Fix: walk `status.Details()` and match `*errdetails.ErrorInfo` with `Reason == "READ_INDEX_NOT_CAUGHT_UP"`. That is both immune to message changes and more precise — it cannot collide with an unrelated `FailedPrecondition` whose message happens to contain the substring.

internal/storage/readstore/store.go
```diff
@@ -148,9 +166,10 @@ func OpenReadOnly(dirPath string, logger logging.Logger) (*Store, error) {
 	}
 
 	s := &Store{
-		db:     db,
-		logger: logger.WithFields(map[string]any{"cmp": "read-store-readonly"}),
-		dir:    dirPath,
+		db:       db,
+		logger:   logger.WithFields(map[string]any{"cmp": "read-store-readonly"}),
+		dir:      dirPath,
+		readOnly: true,
```
paul-nicolas, line 172: [Medium] `OpenReadOnly` leaves `leases` nil, so `Leases()` hands out a nil registry

`New` sets `leases: NewLeaseRegistry()`, but this struct literal omits the field. `Leases()` returns it unconditionally, and none of `Acquire`, `Release` or `Watermark` is nil-receiver-safe — each takes `r.mu.Lock()` on the receiver, so any of them panics.

The only thing preventing that today is the `if rs.Frozen()` early return in `AlignedIndexSnapshot`, which returns before reaching `rs.Leases().Acquire(mainSeq)`. That makes an ordering convention in one caller load-bearing for a nil-pointer panic in a request goroutine — and the path is live, not theoretical: checkpoint read indexes are opened through `OpenReadOnly` and handed to the controller, whose list paths all call `AlignedIndexSnapshot`. Reordering that guard, or any future caller needing a lease before the freeze check, panics.

Fix: add `leases: NewLeaseRegistry(),` to this literal. A frozen store never has anyone GC against it, so a live-but-unused registry costs nothing, and "a `*Store` always has a registry" becomes unconditional instead of a convention.

internal/query/aligned_snapshot.go
```diff
@@ -0,0 +1,153 @@
...
+	// meaningless: the pair is consistent by construction at the checkpoint's
+	// sequence (the builder materializes the read index as its fold crosses
+	// the CreatedQueryCheckpoint log). Serve it as-is; the horizon filter
+	// still bounds the index-ahead direction. No GC runs against a frozen
+	// store, so there is no lease to hold.
+	if rs.Frozen() {
+		return rs.NewSnapshot(), mainSeq, func() {}, nil
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), alignWait)
```
paul-nicolas: [Medium] The bounded wait discards the caller's context

`AlignedIndexSnapshot` takes no `ctx` and builds its own from `context.Background()`, so the 2s bound is detached from the request. Every caller has a request context in hand — `Execute`, `AggregateVolumes`, `ListLogs`, and `listEntities`' callers — and all of them drop it here.

Consequences: a client that disconnects or hits its own deadline still occupies a server goroutine **and a registered read lease** for the full `alignWait`; a server shutdown cannot drain these waits; and the lease is what bounds event-GC retention, so cancelled requests hold reclamation back for longer than they should. It also runs against the project rule that a blocking call honours the `ctx` it is given.

There is a free win attached: this is also why `TestAlignedIndexSnapshot_TimesOutNotCaughtUp` costs 2.22s on every run (measured — the rest of the package is ~1.1s). With a `ctx` parameter the test injects a 50ms deadline instead of waiting out the constant.

Fix: `AlignedIndexSnapshot(ctx context.Context, rs *readstore.Store, mainReader dal.PebbleReader)` with `ctx, cancel := context.WithTimeout(ctx, alignWait)`, threaded through the call sites in `executor.go`, `controller_default.go` and `list_entities.go`.

paul-nicolas (commented):
Multi-model review (Claude + Codex), then every finding was independently re-verified by an adversarial pass instructed to refute it.

**Posted inline: 8** — 4 Major, 4 Medium. **Withheld: 20** Minor/Nit (listed below).

Verification did real work in both directions: 27 candidates entered, **3 were rejected outright** and 16 downgraded. Notably it killed the only *Critical* the reviewers produced (a claimed backfill-vs-GC-watermark corruption — blocked by `completeBackfill`'s replay gate plus the `CurrentVersion == 0` query gate, and unreachable because pins are always the current `mainSeq`), and it killed a claimed `ListAuditEntriesFrom` alignment bug that turned out to be correct, documented design (the audit index has its own sequence space and is already gated by `waitFilteredAuditConsistency`). Two findings that had *two-reviewer agreement* also failed verification — concurrence tracked plausibility, not reachability.

Build is clean (`go build ./...`), and `readstore` / `query` / `indexbuilder` / `ctrl` unit tests all pass on this head.

The two most substantial findings are both silent-wrong-result classes: the lease/GC ordering gap in `AlignedIndexSnapshot`, and the one-`fsmSeq` rewrite stamp vs. a stale pin. The latter is novel beyond the residual window already documented in `read-snapshot-consistency.md` — it disproves that section's own bound ("the microseconds between the two acquisitions") and its blast radius ("may miss entries").

- `internal/storage/readstore/store.go:757` — `WaitForSequence` broadcasts outside `progressMu` (lost wakeup); rescued within ~100ms by `NotifyProgress`, unlike the locked sibling `WaitForCheckpoint`
- `internal/query/aligned_snapshot.go:88` — `waitErr` discarded, laundering a Pebble read fault into a freshness precondition
- `internal/storage/readstore/iterator_event_resolve.go:37` — `TODO(EN-1748)` says "before this leaf serves hot query paths", but this PR wires it into every metadata/exists leaf; the `seekFloor` memo every sibling leaf carries is still absent (perf only, ~8 lines)
- `internal/storage/readstore/event_gc.go:71` / `inspect.go:87` — malformed event keys silently skipped, while `iterator_event_resolve.go:106` hard-errors on the same key; the "malformed keys are the checker's business" comment names a detector that does not exist (invariant #8 scope refinement (a) covers only the reverse map `0x03`)
- `internal/storage/readstore/keys.go:169,414` — `MetadataIndexKey` / `EntityExistsKey` now have zero callers repo-wide; `MetadataIndexKeyV` / `EntityExistsKeyV` are test-only. A key they build is rejected by the resolver as malformed
- `docs/.../read-path/read-snapshot-consistency.md:135` — links `readstore-event-keys.md`, which does not exist; the new mechanism (event layout, pin rule, lease registry, GC watermark) has no dedicated page
- `docs/.../indexer/indexer.md:105-106,130` — still documents the pre-PR key shape with no `\x00 + seq + op` tail, and names the two write-dead builders
- `internal/query/aligned_snapshot.go:105` — the ACCOUNTS carve-out justifies the absent direction but not the index-ahead one; `has_asset` can admit an account committed after the handle (worth a comment + follow-up, not a fix here)
- `internal/query/compile.go:95` — 13 positional params; `pin == 0` silently empties every event-resolved leaf. An input struct would name it at every call site
- `internal/application/indexbuilder/process_logs.go:181` — `eventSeq` is batch-sticky and the guard only rejects zero, so it catches "never set" but not "set for a previous log"
- `internal/application/ctrl/controller_default.go:1696` — the one unguarded `MainHorizonKeep` wrap; unreachable today (LOGS always returns non-nil). Making `NewFilterIterator` treat a nil keep as admit-all would remove the class and four guards
- index builder has no rollback reset (unlike `usagebuilder`); with event keys a post-`RestoreCheckpoint` pin reads stale rather than merely lagging
- `internal/storage/readstore/write_batch.go:186` — the same-value early return is state-based, not sequence-based; the comment's "identical at every pin" should read "every *reachable* pin"
- `internal/application/indexbuilder/gc.go:245` — no backfill-cursor term in the watermark; correctness currently rests on two indirect gates rather than the stated precondition
- `internal/storage/readstore/event_gc.go:33` — foreign-zone resume cursor degrades silently (unreachable by construction; wants a comment)
- `internal/storage/readstore/event_gc.go:62,109` — `_ = batch.Delete(...)` with no justifying comment (the error is in fact unreachable on a non-indexed batch)
- `internal/application/indexbuilder/gc.go:264` — `return` on a zone error abandons the loop, so the second zone is never swept that tick
- `internal/query/aligned_snapshot.go:36` — `alignWait` is an untunable const; the timeout test burns 2.22s of every run for want of a seam
- `tests/e2e/business/*_test.go` — the two new specs duplicate their `BeforeAll` ledger/index bootstrap verbatim
- `tests/e2e/testutil/retry_not_caught_up.go:20` — `notCaughtUpBackoff` is a fixed 200ms poll interval, not backoff, and has no jitter

internal/query/aligned_snapshot.go
```diff
@@ -0,0 +1,176 @@
...
+	if rs.Frozen() {
+		return rs.NewSnapshot(), mainSeq, func() {}, nil
+	}
+
+	waitCtx, cancel := context.WithTimeout(ctx, alignWait)
+	defer cancel()
+
+	for {
+		lease, ok := rs.Leases().Acquire(mainSeq)
+		if !ok {
```
NumaryBot, line 95: 🟠 **[major] Avoid retrying against the same frozen main snapshot**

When GC advances the reclaim floor between opening the main handle and acquiring the lease, `mainReader` is a Pebble snapshot, so re-reading it always returns the same rejected sequence. This loop then spins continuously until the two-second deadline and returns a 503 instead of obtaining a fresh pin; the caller must reopen the main handle or fail immediately rather than retrying this immutable reader.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1678#issuecomment-5215764400

internal/application/ctrl/controller_default.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/7a80a72be — the wrap is guarded, since `MainHorizonKeep` returns nil for ACCOUNTS by design.

internal/query/aligned_snapshot.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/4b292bb7d — the cancellation broadcast now takes `progressMu`, like `WaitForCheckpoint` already did, so the 2s bound holds.

internal/query/aligned_snapshot.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/4b292bb7d (same fix as the sibling thread on this file).

internal/query/aligned_snapshot.go
Azorlogh: Confirmed and fixed in https://github.com/formancehq/ledger/commit/4b292bb7d. The gap was wider than lease-before-snapshot: the pin is fixed at `ReadLastSequence` and registered later, so a pass ticking in between swept at the fold cursor, which the pin can sit far below. `LeaseRegistry` now publishes a monotone reclaim floor under the same lock that admits leases (`BeginGC`), `Acquire` refuses pins below it, and the reader re-pins against a fresh handle. The false premise in `gc.go` ("every future pin is at least the fold cursor") is gone too.

internal/query/aligned_snapshot.go
Azorlogh: Good catch — correct, and my comment claiming a fresh sequence was wrong: the handle is a fixed snapshot by construction. Fixed in https://github.com/formancehq/ledger/commit/9eac1bcb1. Re-pinning in place is not sound either (the pin must keep matching the reader enrichment uses), so a reclaimed pin now fails immediately with `ErrReadPinReclaimed` (Unavailable / READ_PIN_RECLAIMED) and a retry opens a fresh handle.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1678#issuecomment-5215764400

internal/application/indexbuilder/backfill.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/4b292bb7d, via your second option: promotion records an `ActivationSequence` and `PinnedVersionResolver` withholds the version from pins beneath it, so the query returns `ErrIndexBuilding` instead of an empty page. Doc paragraph corrected on both counts. Worth noting this is the same failure I'd reproduced independently on release/v3.0 by a different route (mixed-type reads during the rewrite window) — your framing is the sharper one.

internal/query/executor.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/4b292bb7d. `ErrReadIndexNotCaughtUp` implements `Describable` (Unavailable / READ_INDEX_NOT_CAUGHT_UP, new proto reason), so it renders as 503. I put the `Retry-After` on the whole Unavailable class in the Describable branch rather than adding a third dedicated one, since the two existing branches already promise it.

internal/query/aligned_snapshot.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/4b292bb7d — you're right that my comment guarded the wrong window. `LeaseRegistry` now publishes a monotone reclaim floor under the same lock that admits leases (`BeginGC`), and `Acquire` refuses pins below it. I went with fail-fast rather than internal retry (https://github.com/formancehq/ledger/commit/9eac1bcb1): the handle is a fixed snapshot so re-reading cannot advance the pin, and the pin must keep matching the reader enrichment uses — reclaimed pins return `ErrReadPinReclaimed` and the retry opens a fresh handle. The false premise in `gc.go` is gone.

tests/e2e/testutil/server.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/4b292bb7d — agreed, installing them globally was a mistake. Opt-in now via `SetupSingleNodeRetryingNotCaughtUp` / `NotCaughtUpRetryDialOptions`; only the two cross-store specs use it, every other spec keeps the fail-fast client.

tests/e2e/testutil/retry_not_caught_up.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/4b292bb7d — a failed replay or `CloseSend` is terminal now (returns the original error, after closing the replacement) rather than spinning against the dead stream until the window closes.

tests/e2e/testutil/retry_not_caught_up.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/4b292bb7d — it walks `status.Details()` for `*errdetails.ErrorInfo` with reason READ_INDEX_NOT_CAUGHT_UP instead of matching the message text.

internal/storage/readstore/store.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/4b292bb7d — `OpenReadOnly` sets the registry, so "a `*Store` always has one" holds unconditionally instead of resting on the `Frozen()` guard's position.

internal/query/aligned_snapshot.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/4b292bb7d — `AlignedIndexSnapshot` takes the caller's context and derives the grace from it, so a cancelled request stops holding a goroutine and a lease (which bounds reclamation). A Pebble fault during the wait no longer launders into a freshness precondition either.

internal/storage/readstore/iterator_event_resolve.go
```diff
@@ -0,0 +1,213 @@
...
+				it.err = fmt.Errorf("malformed metadata event key %x", it.iter.Key())
+
+				return false
+			}
+
+			if !bytes.Equal(g, group) {
+				break
+			}
+
+			// Events are seq-ascending within the group, so this keeps the
```
NumaryBot, line 147: 🟠 **[major] Reject unknown metadata event operations**

If an event key contains an op other than `MetadataEventAdd` or `MetadataEventDel`—for example from corruption or a malformed writer—this silently interprets it as DEL and can omit matching entities without surfacing an error. Validate the op while parsing, consistent with the loud-failure requirement in [AGENTS.md:26](/data/worktrees/formancehq__ledger-pr-1678/AGENTS.md#L26).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1678#issuecomment-5215764400

internal/storage/readstore/iterator_event_resolve.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/76b014fb6 — correct, and the same hole was in the GC and inspect parsers, so all three now validate the op. The query path refuses the key like any malformed one (iteration stops with an error rather than reporting a membership it can't vouch for); the GC leaves the group whole, since reclaiming around an unreadable event could drop a live ADD.

internal/storage/readstore/inspect.go
```diff
@@ -57,6 +57,75 @@ type InspectFacetEntry struct {
...
+
+	for iter.First(); iter.Valid(); iter.Next() {
+		key := iter.Key()
+
+		tpos := len(key) - suffix
+		if tpos < prefixLen || key[tpos] != metadataEventTerminator {
+			continue
+		}
+
+		g := key[prefixLen:tpos]
```
NumaryBot, line 95: 🟠 **[major] Reject unknown metadata event operations**

If an introspection scan encounters an event with an unknown operation, this silently ignores it and derives facets, cardinality, or entity counts from the remaining events. A corrupt latest event can therefore produce plausible but incorrect index statistics instead of surfacing corruption; return an error as required by [AGENTS.md:26](/data/worktrees/formancehq__ledger-pr-1678/AGENTS.md#L26).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1678#issuecomment-5215764400

internal/storage/readstore/inspect.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/5e3f17cfa — agreed, and the skip was worse than statistics-only: `live` is assigned at the end of the loop body, so a group whose first event was skipped inherited the *previous* group's verdict and could be reported as a current member on the strength of another entity's ADD. The malformed-key check just above had the same silent `continue` and is fixed with it.

internal/storage/readstore/event_gc.go
```diff
@@ -0,0 +1,136 @@
...
+		key := iter.Key()
+		tpos := len(key) - suffix
+		if tpos < 1 || key[tpos] != metadataEventTerminator {
+			scanned++
+
+			continue // malformed keys are the checker's business, not GC's
+		}
+
+		op := key[tpos+9]
+		if !validEventOp(op) {
```
NumaryBot: 🟠 **[major] Preserve groups containing unknown event operations**

When a group contains an unknown operation, this `continue` does not actually leave the group whole: valid events before and after it still participate in `pendingBelow` processing and may be deleted. Thus GC mutates evidence and may discard a live ADD around an unreadable event, contrary to the loud-failure requirement in [AGENTS.md:27](/data/worktrees/formancehq__ledger-pr-1678/AGENTS.md#L27); mark the group unsafe and skip all reclamation for it.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1678#issuecomment-5215764400

internal/storage/readstore/event_gc.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/dbb17232e — right, the `continue` only spared the unreadable event itself; the group still lost its ADD to the supersede path, which deletes eagerly as each event is judged. Whether a group is safe to reclaim is only known once it has been read in full, so condemned events are now buffered and applied at the group boundary, or dropped if the group turned out to hold something unreadable.

internal/query/executor.go
```diff
@@ -88,33 +104,37 @@ func Execute(
...
+	handle, releaseHold, err := OpenAlignedHandle(rs, pebbleStore)
+	if err != nil {
+		return nil, fmt.Errorf("creating read handle: %w", err)
 	}
 
-	schema := SchemaFieldsForTarget(ledgerInfo.GetMetadataSchema(), pq.GetTarget())
+	defer releaseHold()
+	defer func() { _ = handle.Close() }()
+
+	indexSnap, mainSeq, releaseLease, err := AlignedIndexSnapshot(ctx, rs, handle)
```
NumaryBot: 🟠 **[major] Skip alignment for main-store-only prepared queries**

When a prepared ACCOUNTS or TRANSACTIONS query has no filter, `Compile` reads its universe entirely from the main-store handle, but this unconditional alignment still waits for the read-index fold. If the builder is lagging or stopped and the caller has no deadline, an otherwise independent query now blocks indefinitely; even with a deadline it fails unnecessarily. Mirror `listEntities` by bypassing alignment for these main-store-only shapes.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1678#issuecomment-5215764400

internal/query/executor.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/c40e1e49f — confirmed: with a nil filter, ACCOUNTS/TRANSACTIONS take their universe from `ctx.pebbleReader` (the mainstore handle) and enrich from it, so there is no index leaf to align against. The rule now lives in `query.AlignmentOwed` and the list path calls it too, so the two cannot drift. LOGS stays aligned even unfiltered — its universe is the readstore's per-ledger log limb. The horizon filter is skipped with it: only an aligned read can hold membership the handle hasn't caught up to.

Covered by `TestExecute_UnfilteredQueryDoesNotWaitForTheFold` (readstore pinned permanently behind: unfiltered accounts/transactions answer, filtered + LOGS still hit the deadline).

internal/query/executor.go
```diff
@@ -88,33 +105,58 @@ func Execute(
...
-	// Check min_log_sequence freshness
-	if req.GetMinLogSequence() > 0 {
-		lastIndexed, err := rs.LastIndexedSequence()
+	schema := SchemaFieldsForTarget(ledgerInfo.GetMetadataSchema(), pq.GetTarget())
+
+	// Always open a read handle — needed for filter compilation and entity
+	// enrichment. Opened BEFORE the index snapshot: alignment guarantees the
+	// snapshot's fold cursor covers everything the handle sees (EN-1748), and
+	// OpenAlignedHandle holds reclamation still across the two steps.
+	handle, releaseHold, err := OpenAlignedHandle(rs, pebbleStore)
```
NumaryBot: 🟠 **[major] Avoid reserving GC history for unaligned reads**

When `AlignmentOwed` is false, this still opens the handle through `OpenAlignedHandle`, whose reservation remains held until the request or cursor closes. Unfiltered account/transaction reads therefore pin the event-GC watermark despite never reading event history; a slow or leaked cursor can prevent reclamation indefinitely. Choose whether alignment is owed before opening the handle and use a normal read handle for unaligned reads; the analogous list-controller call sites have the same issue.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1678#issuecomment-5215764400

internal/storage/readstore/event_gc.go
```diff
@@ -0,0 +1,158 @@
...
+			// unreadable event was meant to supersede. Preserve the group and
+			// let the query path reject it loudly.
+			groupUnsafe = true
+
+			continue
+		}
+
+		seq := binary.BigEndian.Uint64(key[tpos+1 : tpos+9])
+
+		if seq >= watermark {
```
NumaryBot, line 151: 🔴 **[blocker] Preserve groups containing unknown event operations**

When a group has a valid event at or above the watermark followed by an unknown operation, this early `flush()` commits deletions and clears `groupUnsafe` before the unreadable event is encountered. GC can therefore still discard earlier evidence from a corrupt group instead of preserving the whole group; defer applying reclamation until the actual group boundary, as required by [AGENTS.md:26](/data/worktrees/formancehq__ledger-pr-1678/AGENTS.md#L26).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1678#issuecomment-5215764400

internal/storage/readstore/event_gc.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/a880cba2a — correct, and my own comment on that closure claimed the rule the code then broke ("applied per group, because whether it is safe is only known once the whole group has been read") while the above-watermark arm called it mid-scan. Judging and reclaiming are now separate: `settle` condemns the pending event, `closeGroup` applies what the group condemned and runs only at the group boundary. Covered by `TestGCEventZone_PreservesAGroupCorruptedAfterItsWatermarkEvent` (ADD@5, DEL@7, ADD@150, unknown@160 — red before the split).

internal/storage/readstore/event_gc.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+
+		reclaimable = reclaimable[:0]
+		groupUnsafe = false
+	}
+
+	for iter.First(); iter.Valid(); iter.Next() {
+		key := iter.Key()
+		tpos := len(key) - suffix
+		if tpos < 1 || key[tpos] != metadataEventTerminator {
```
NumaryBot, line 95: 🟠 **[major] Preserve malformed groups that sort before their valid events**

For a truncated or otherwise malformed key that sorts immediately before valid events sharing its identity, `groupUnsafe` is set here but then cleared by `closeGroup()` when the first valid identity is encountered. GC may consequently reclaim the valid events in that unreadable group, despite the stated whole-group preservation guarantee. Buffer malformed keys or conservatively protect the adjacent group; silently continuing also conflicts with [AGENTS.md:26](/data/worktrees/formancehq__ledger-pr-1678/AGENTS.md#L26).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1678#issuecomment-5215764400

internal/storage/readstore/event_gc.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/8b32fda96 — right, and the reason is that a truncated key is a *prefix* of the events it shares an identity with, so it sorts ahead of them: marking the accumulating group protected the group on the wrong side, and the mark was cleared at the very next key. Since a malformed key cannot be attributed to either neighbour by definition, both are now preserved — the mark carries across one group boundary. Covered by `TestGCEventZone_PreservesTheGroupAfterAMalformedKey` (red without the carry).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1678#issuecomment-5215764400

gfyrag (approved):
Functional review passed. The cross-store acquisition, pinned event resolution, lease/GC boundaries, horizon trimming, and targeted regression coverage are coherent; the targeted query/readstore/indexbuilder/check/controller suites also pass locally.

Concern retained: `codecov/patch` is still below the repository target (72.36% vs 80%). Please address or explicitly waive that gate before merge.

PR #1679, opened 2026-08-07
fix(EN-1750): rebuild the chapter registry on restore; refuse to confirm mismatched cold archives

Azorlogh:
Fixes [EN-1750](https://formance-team.atlassian.net/browse/EN-1750): a restored store loses the chapter registry, and the next archival cycle then permanently destroys data.

## Problem

`RebuildDelta` replayed the chapter lifecycle logs (`CloseChapter`/`SealChapter`/`ArchiveChapter`/`ConfirmArchiveChapter`) as "no persistent state to rebuild" — but the registry rows (`ZoneGlobal/SubGlobChapters`) and `SubGlobNextChapterID` are precisely what FSM recovery reads at boot. A restored node's registry was therefore frozen at checkpoint time: chapters archived inside the exported delta were forgotten, and the boot genesis path re-created "chapter 1, StartSequence 1".

The next `CloseChapter` closed that impostor spanning `[1, now]`. The Archiver's crash-recovery path found the **real** chapter 1's SST in cold storage under the colliding id, verified only that the object was intact (stored checksum vs stored bytes — self-consistency, not correspondence to the request), and proposed `ConfirmArchiveChapter` **without uploading**. The deterministic purge then deleted `[1, now]` from hot storage while cold held only the old chapters' ranges — everything in between existed **nowhere**. Downstream: archived reads NotFound (the impostor's range also shadows valid chapters in `findArchivedChapterForSequence`), idempotency replays resolving to sequence 0, and ~29k `IDEMPOTENCY_MISMATCH` checker errors.

Found by the archival+restore model test (`run_model_test.sh --archive --restore`).

## Fix (two independent defects)

**A — rebuild the chapter projection on restore** (`internal/infra/backup/rebuild.go`). Every chapter log carries the full post-transition `Chapter`, so the rebuild upserts each snapshot in sequence order and advances the next-chapter id past every chapter a close opens. One field needs recovery: the close log's snapshot is cloned *before* the live apply stamps `LastAuditHash` onto the tracker's row, so for a chapter closed but not yet sealed at backup time the rebuild reads that hash back from the stored audit entry at `CloseAuditSequence` — it is the seed the checker chains from across the chapter's eventual purge, and no later lifecycle log exists in the delta to carry it.

**B — never confirm an archive that doesn't match the request** (`internal/infra/state/archiver.go`). `verifyExistingArchive` now also decodes the existing SST's embedded `chapterMetadata` (deterministic by construction) and requires it to equal the `ArchiveRequest` exactly. A mismatch means the stored object is a different chapter incarnation: confirming would purge the request's ranges with only the old object's ranges archived, so the archiver fails loudly and never confirms — whatever produced the identity confusion. This makes the data-loss half of the bug unreachable even if a future defect confuses chapter identity again.

## Verification

- First commit is the e2e repro (`tests/e2e/cluster/restore_chapter_registry_test.go`), committed **failing** at "restore lost the archived chapter's registry entry". The fix commits flip it green, including byte-for-byte `SealingHash`/`StateHash`/`LastAuditHash` fidelity across the restore and fresh chapter ids after a post-restore close (pins `NextChapterID` against partial fixes). ~4s, tags `e2e,s3`.
- New unit tests: full-lifecycle registry rebuild + close-only delta with `LastAuditHash` recovery (`rebuild_chapters_test.go`); archiver refuses a range-mismatched archive (`TestArchiverAlreadyArchivedMismatchedRangesRefusesConfirm`). Two pre-existing crash-recovery tests now seed real SSTs built for their requests — decodable metadata is part of the contract fix B enforces.
- **The combined `--archive --restore` model run that found the bug now passes**: 3 restore cycles interleaved with 6 archival cycles, zero model findings, clean teardown `store check`.
- Full `./internal/...` unit suite green.

Docs: `docs/technical/architecture/subsystems/chapters/backup.md` gains a "Chapter identity across restore" section; the ops restore flow now lists the chapter registry among rebuilt state.

[EN-1750]: https://formance-team.atlassian.net/browse/EN-1750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/infra/backup/rebuild.go
```diff
@@ -415,12 +415,46 @@ func RebuildDelta(
 				}
 			}
 
-		// Log types with no persistent state to rebuild:
-		case *commonpb.LogPayload_RemovedEventsSink:
 		case *commonpb.LogPayload_CloseChapter:
+			if p.CloseChapter == nil {
```
NumaryBot: 🟠 **[major] Reject incomplete chapter lifecycle logs**

If a restored log has a nil lifecycle payload or chapter snapshot, these branches silently skip rebuilding the registry and can report a successful restore with missing chapter state, recreating the identity-collision risk this patch addresses. Chapter lifecycle logs contractually carry complete snapshots, so malformed entries should fail the rebuild loudly rather than continue; this is also required by [AGENTS.md:26](../../../AGENTS.md#L26).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1679#issuecomment-5215880930

internal/infra/backup/rebuild.go
Azorlogh: Fixed in ffac62039 — all four chapter lifecycle cases now fail the rebuild loudly on a missing snapshot, with a table-driven regression test.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1679#issuecomment-5215880930

internal/infra/state/archiver.go
```diff
@@ -247,29 +251,119 @@ func (a *Archiver) archive(stop <-chan struct{}, req ArchiveRequest) error {
...
-			chapterID, hex.EncodeToString(expected), hex.EncodeToString(actual))
+			req.ChapterID, hex.EncodeToString(expected), hex.EncodeToString(actual))
+	}
+
+	meta, err := a.readArchivedMetadata(ctx, req.ChapterID)
+	if err != nil {
+		return fmt.Errorf("reading archived metadata for chapter %d: %w", req.ChapterID, err)
+	}
+
+	if want := chapterMetadata(req); *meta != want {
```
gfyrag: **[P1] Bind an existing archive to the current history, not only to its ranges**

This equality can still accept a different chapter incarnation. After restoring an older valid backup against the same surviving cold-storage namespace, the new timeline can reuse the same chapter ID and produce the same log/audit counts with different operations. The old SST then has the exact same metadata tuple, and its checksum validates the old object against itself, so this returns success and `ConfirmArchiveChapter` purges the new hot history while cold storage contains the old one.

Please compare a content commitment such as the current chapter's `SealingHash` / `LastAuditHash`, and carry that identity through the confirmation command so the FSM validates it immediately before purge. Ideally use `chapterID + sealingHash` as the object key so distinct incarnations cannot collide.

gfyrag (changes requested):
The chapter-registry rebuild fixes the EN-1750 failure mode where a restored node forgets archived chapters and reuses their IDs. However, the cold-object guard still proves only that the existing object is intact and that its sequence ranges match the current request; it does not prove that the object contains the current chapter incarnation.

A normal point-in-time restore can create this without two clusters running concurrently:

1. Take backup T0 while chapter 1 is still open.
2. The source continues, closes chapter 1 at log 100 / audit 10, and archives history A.
3. The source goes down.
4. Restore T0 against the same surviving cold-storage namespace.
5. The restored cluster produces different history B with the same number of logs/audits and closes chapter 1 at the same boundaries.

The legacy object for history A passes its stored checksum and the metadata equality below, so the worker may confirm and purge history B while cold storage serves A.

Recommended hardening:

- Carry the chapter's `SealingHash` (and optionally `LastAuditHash` / `StateHash` for diagnostics) in `ArchiveRequest` and the SST metadata. `SealingHash = BLAKE3(chapterID || closeSequence || LastAuditHash || StateHash)` already commits to the incarnation.
- Include that identity in `ConfirmArchiveChapter`; the FSM must compare the command's hash with the cached ARCHIVING chapter before transitioning or purging. This is deterministic and requires no Pebble read.
- Prefer content-addressed cold keys such as `chapters/{chapterID}/{sealingHash}.sst`. Different timelines then cannot collide or block one another, and old objects do not need to be overwritten/deleted.
- For legacy ID-only objects, fail closed unless the archived audit boundary can be matched to the current chapter's `LastAuditHash`; keep a verified read fallback or lazily copy the object to the content-addressed key.
- Add a regression test with the same chapter ID and identical log/audit ranges but a different audit/sealing hash, plus an FSM test proving a stale confirmation cannot purge the current incarnation.

This keeps the current range comparison as a useful structural check, while making archive identity cryptographic end-to-end.

internal/domain/processing/processor_chapter.go
```diff
@@ -137,6 +139,22 @@ func processConfirmArchiveChapter(order *raftcmdpb.ConfirmArchiveChapterOrder, c
...
 
+	// The confirm trades hot history for a cold archive, so it must name the
+	// incarnation the archive was built for. Ranges alone do not identify one: a
+	// store restored from an older backup over a surviving cold-storage namespace
+	// can reuse a chapter id and reach the same log and audit counts over different
+	// operations, so confirming against the stale archive would purge history that
+	// archive does not contain. The sealing hash commits a chapter to its content,
+	// and comparing it here — in apply, against the chapter the FSM holds — puts the
+	// check in the same step as the purge it authorises.
+	if !bytes.Equal(order.GetSealingHash(), chapterReader.GetSealingHash()) {
```
NumaryBot, line 150: 🟠 **[major] Pass the sealing hash from the Antithesis confirm caller**

When the Antithesis chapter-close workload manually confirms an `ARCHIVING` chapter, it still sends only `chapter_id`, so `order.GetSealingHash()` is empty and this check always rejects every sealed chapter. Update `tests/antithesis/workload/bin/cmds/main/singleton_driver_chapter_close/main.go` to copy the listed chapter's sealing hash into the confirmation request.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1679#issuecomment-5215880930

internal/domain/processing/processor_chapter.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/89db728622f9e5bcbf7f0d3ba5a3a0eb8ac2dd51 — the driver now copies the listed chapter's sealing hash into the confirm.

internal/infra/state/archiver.go
Azorlogh: Done in https://github.com/formancehq/ledger/commit/bac24cd2c — the sealing hash now travels into the archive's `chapterMetadata` (compared alongside the ranges) and into `ConfirmArchiveChapter`, where the handler checks it against the chapter it holds before mutating anything, rejecting a mismatch with `CHAPTER_ARCHIVE_IDENTITY_MISMATCH`.

Left the object-key suggestion out of this PR as [EN-1825](https://formance-team.atlassian.net/browse/EN-1825): it upgrades the failure from a loud stall needing manual cold-storage cleanup to self-healing, but it spans the cold-storage drivers, the reader, the backup backfill and the checker, and it needs a decision on orphan retention first.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1679#issuecomment-5215880930

gfyrag (dismissed):
Functional review passed. The previous archive-incarnation blocker is fixed: the sealing hash now binds the stored metadata, confirmation command, and FSM-side pre-purge validation. Chapter-registry rebuild and targeted backup/state/processing tests also pass locally.

Concern retained: `codecov/patch` is still below the repository target (60.83% vs 80%). Please address or explicitly waive that gate before merge.

paul-nicolas (dismissed):

paul-nicolas (approved):

PR #1680, opened 2026-08-07
ci: add manual Antithesis workflow

flemzord:
## What changed

Adds a dedicated manual GitHub Actions workflow for Antithesis. The workflow accepts a source ref, report name, duration in minutes, and an optional driver filter. It builds and publishes the AMD64 Kubernetes images, launches the Antithesis run, then releases the runner without waiting for the report.

## Why

The current development host is ARM64, so building the required AMD64 images locally relies on emulation and is prohibitively slow. Keeping the dispatcher on the default branch also makes GitHub's **Run workflow** button available while allowing `release/v3.0` or a feature branch to be selected through `source_ref`.

## Setup required

Create a protected GitHub environment named `antithesis` with:

- secrets `ANTITHESIS_API_KEY` and `ANTITHESIS_REGISTRY_KEY_JSON`;
- variables `ANTITHESIS_REPOSITORY`, `ANTITHESIS_REPORT_RECIPIENT`, and `ANTITHESIS_TENANT`.

For the first validation, use `feat/fix-antithesis-unreached-asserts` as `source_ref`.

## Validation

- `actionlint` (custom Namespace runner label excluded from the built-in label catalogue)
- YAML parsing with `yq`
- `git diff --check`

The PR contains only `.github/workflows/antithesis.yml`.

### DIFF
--- a/.github/workflows/antithesis.yml
+++ b/.github/workflows/antithesis.yml
@@ -1,0 +1,172 @@
+name: Run Antithesis
+
+on:
+  workflow_dispatch:
+    inputs:
+      source_ref:
+        description: Branch, tag, or commit to build and test
+        required: true
+        type: string
+        default: release/v3.0
+      name:
+        description: Name shown in the Antithesis report
+        required: true
+        type: string
+        default: Ledger v3 manual validation
+      duration_minutes:
+        description: Test duration in minutes
+        required: true
+        type: number
+        default: 20
+      include:
+        description: Optional comma-separated driver paths
+        required: false
+        type: string
+        default: ''
+
+permissions:
+  contents: read
+
+concurrency:
+  group: antithesis-${{ inputs.source_ref }}
+  cancel-in-progress: false
+
+jobs:
+  build-publish-launch:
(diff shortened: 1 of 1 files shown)


.github/workflows/antithesis.yml
```diff
@@ -0,0 +1,172 @@
+name: Run Antithesis
+
+on:
+  workflow_dispatch:
+    inputs:
+      source_ref:
+        description: Branch, tag, or commit to build and test
+        required: true
+        type: string
+        default: release/v3.0
```
NumaryBot, line 10: 🔴 **[blocker] Choose a source ref that provides the required recipes**

The default dispatch fails before building: the current `release/v3.0` `tests/antithesis/Justfile` does not define `k8s-launch-minutes` (or the later `check-k8s-image-freshness` recipe), so the validation step always exits with an error when users accept this default. Point the default at a ref containing the required recipes or add them to `release/v3.0` first.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1680#issuecomment-5217505590

PR #1681, opened 2026-08-07
fix(antithesis): stabilize assertions and quorum recovery

flemzord:
## What changed

- clean up Antithesis assertions so short runs only require properties that the selected drivers can actually exercise
- make operational singleton drivers deterministic and bound their execution
- add a manual GitHub Actions workflow that builds the AMD64 images and starts a named Antithesis run with a configurable duration
- make quorum recovery wait for the StatefulSet to settle before scaling back up
- remove deleted or pending Raft voters during scale-down, including pods recreated by Kubernetes before the operator finishes reconciliation

## Root cause

Several assertions were registered globally as `Reachable` or unconditional `Sometimes` properties even when the corresponding driver was not selected. Short targeted runs therefore reported red properties with no examples.

For quorum recovery, the operator inferred Raft membership from the current Kubernetes pod state. A pod recreated as `Pending` could be mistaken for a node that had never joined Raft, even though its predecessor was still a committed voter. Its PVC was then deleted without removing that voter, causing stale-membership failures on recovery.

## Impact

Targeted Antithesis runs now give meaningful assertion results, can be launched directly from GitHub Actions, and recover cleanly after both non-leader pods are killed.

## Validation

- `just pre-commit`
- `go build ./...`
- operator unit and integration tests
- GitHub Actions AMD64 build: https://github.com/formancehq/ledger/actions/runs/31206517378
- 20-minute Antithesis quorum-recovery run: https://formance.antithesis.com/report/VTJCtCJZZ2ywxDtqi3uWuwCc/D7y6vL4BOXfeUwJEmnR-aXa5ChYMbU-J5sjFPcOtRms.html

The targeted run reduced stale-membership / exit-code-1 counterexamples from 282 to 0. Force-remove and single-voter properties passed 780/780. A remaining exit-code-2 signal is caused by Kubernetes terminating a briefly recreated pod during StatefulSet scale-down; it is not a Raft membership violation. Four other red entries have 0 examples because their unrelated drivers are intentionally absent from this targeted run.

### DIFF
--- a/internal/storage/dal/store.go
+++ b/internal/storage/dal/store.go
@@ -10,7 +10,9 @@ import (
 	"reflect"
 	"runtime"
 	"strconv"
+	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/cockroachdb/pebble/v2"
@@ -28,6 +30,11 @@ import (
 // ErrStoreClosed is returned when a store operation is attempted after the
 // Pebble database has been closed. This prevents panics during shutdown races.
 var ErrStoreClosed = errors.New("store closed")
+
+const (
+	storeOpenLockRetryTimeout = 5 * time.Second
+	storeOpenLockRetryDelay   = 50 * time.Millisecond
+)
 
 const (
 	liveDir = "live"
@@ -507,7 +514,7 @@ func NewStore(
 
 	openStart := time.Now()
 
-	db, err = pebble.Open(liveDir, opts)
+	db, err = openPebbleAfterPreviousProcess(liveDir, opts)
 	if err != nil {
 		return nil, fmt.Errorf("opening pebble database: %w", err)
 	}
@@ -561,6 +568,39 @@ func NewStore(
 	store.cleanupIncomingRestore()
 
 	return store, nil
+}
+
+// openPebbleAfterPreviousProcess tolerates the short overlap between a
+// terminating process and its replacement. Kubernetes normally waits for the
+// old container to exit, but fault injection can expose a window where the new
+// process starts before Pebble's LOCK is released. Retrying only lock-contention
+// errors preserves fail-fast startup for corruption and configuration errors.
+func openPebbleAfterPreviousProcess(dirname string, opts *pebble.Options) (*pebble.DB, error) {
+	return openPebbleWithLockRetry(
+		func() (*pebble.DB, error) { return pebble.Open(dirname, opts) },
+		storeOpenLockRetryTimeout,
+		storeOpenLockRetryDelay,
+	)
+}
+
+func openPebbleWithLockRetry(open func() (*pebble.DB, error), timeout, delay time.Duration) (*pebble.DB, error) {
+	deadline := time.Now().Add(timeout)
+	for {
+		db, err := open()
+		if err == nil {
+			return db, nil
+		}
+		if !isPebbleLockContention(err) || time.Now().Add(delay).After(deadline) {
+			return nil, err
+		}
+		time.Sleep(delay)
+	}
+}
+
+func isPebbleLockContention(err error) bool {
+	return errors.Is(err, syscall.EAGAIN) ||
+		errors.Is(err, syscall.EACCES) ||
+		strings.Contains(err.Error(), "lock held by current process")
 }
 
 // DataDir returns the base data directory path for this store.
--- a/misc/operator/internal/controller/raft_scaledown_test.go
+++ b/misc/operator/internal/controller/raft_scaledown_test.go
@@ -94,6 +94,16 @@ func TestIsPodCrashed_NotFound(t *testing.T) {
 	require.True(t, isPodCrashed(ctx, cs, ns, "nonexistent-pod"))
 }
 
+func TestIsPodNeverReady_NotFound(t *testing.T) {
+	ns := createTestNamespace(t)
+	cs := newTestClientset(t)
+
+	// A missing replica may have joined Raft before it was deleted. The
+	// scale-down path must classify it as crashed so removal is attempted with
+	// --force; remove-node is idempotent when it never joined.
+	require.False(t, isPodNeverReady(ctx, cs, ns, "nonexistent-pod"))
+}
+
 func TestIsPodCrashed_Pending(t *testing.T) {
 	ns := createTestNamespace(t)
 	cs := newTestClientset(t)
--- a/misc/operator/internal/controller/raft_scaledown_test.go
+++ b/misc/operator/internal/controller/raft_scaledown_test.go
@@ -94,16 +94,6 @@ func TestIsPodCrashed_NotFound(t *testing.T) {
 	require.True(t, isPodCrashed(ctx, cs, ns, "nonexistent-pod"))
 }
 
-func TestIsPodNeverReady_NotFound(t *testing.T) {
-	ns := createTestNamespace(t)
-	cs := newTestClientset(t)
-
-	// A missing replica may have joined Raft before it was deleted. The
-	// scale-down path must classify it as crashed so removal is attempted with
-	// --force; remove-node is idempotent when it never joined.
-	require.False(t, isPodNeverReady(ctx, cs, ns, "nonexistent-pod"))
-}
-
 func TestIsPodCrashed_Pending(t *testing.T) {
 	ns := createTestNamespace(t)
 	cs := newTestClientset(t)
@@ -127,7 +117,11 @@ func TestIsPodCrashed_Pending(t *testing.T) {
 	_, err = cs.CoreV1().Pods(ns).UpdateStatus(ctx, pod, metav1.UpdateOptions{})
 	require.NoError(t, err)
 
-	require.False(t, isPodCrashed(ctx, cs, ns, "pending-pod"))
+	// The current Pending pod may be a replacement for a replica that was a
+	// committed voter before its predecessor was deleted. Kubernetes pod state
+	// cannot prove Raft membership history, so scale-down must attempt a force
+	// removal. That operation is idempotent if the node truly never joined.
+	require.True(t, isPodCrashed(ctx, cs, ns, "pending-pod"))
 }
 
 func TestIsPodCrashed_OOMKilled(t *testing.T) {
--- a/tests/antithesis/k8s/workload.yaml
+++ b/tests/antithesis/k8s/workload.yaml
@@ -108,7 +108,7 @@ spec:
         # Bound the model command below the shortest supported platform run so
         # Test Composer can observe at least one clean command completion.
         - name: MODEL_MAX_SECONDS
-          value: "600"
+          value: "180"
         # Restore-cycle rendezvous with the restore-orchestrator sidecar (only
         # the model driver writes requests; other templates leave it idle).
         # The interval and lease are k8s-sized: one cycle is a full cluster
--- a/tests/antithesis/k8s/workload.yaml
+++ b/tests/antithesis/k8s/workload.yaml
@@ -108,7 +108,7 @@ spec:
         # Bound the model command below the shortest supported platform run so
         # Test Composer can observe at least one clean command completion.
         - name: MODEL_MAX_SECONDS
-          value: "900"
+          value: "600"
         # Restore-cycle rendezvous with the restore-orchestrator sidecar (only
         # the model driver writes requests; other templates leave it idle).
         # The interval and lease are k8s-sized: one cycle is a full cluster
--- a/tests/antithesis/k8s/workload.yaml
+++ b/tests/antithesis/k8s/workload.yaml
@@ -105,6 +105,10 @@ spec:
         # MODEL_DUMP_BATCHES env at config-image build.
         - name: MODEL_DUMP_BATCHES
           value: "__MODEL_DUMP_BATCHES__"
+        # Bound the model command below the shortest supported platform run so
+        # Test Composer can observe at least one clean command completion.
+        - name: MODEL_MAX_SECONDS
+          value: "900"
         # Restore-cycle rendezvous with the restore-orchestrator sidecar (only
         # the model driver writes requests; other templates leave it idle).
         # The interval and lease are k8s-sized: one cycle is a full cluster
@@ -114,7 +118,7 @@ spec:
         - name: MODEL_RESTORE_RESP
           value: /var/run/restore/resp
         - name: MODEL_RESTORE_INTERVAL
-          value: "180"
+          value: "30"
         - name: MODEL_RESTORE_TIMEOUT
           value: "900"
         volumeMounts:
--- a/tests/antithesis/workload/bin/cmds/main/singleton_driver_quorum_recovery/main.go
+++ b/tests/antithesis/workload/bin/cmds/main/singleton_driver_quorum_recovery/main.go
@@ -204,5 +204,16 @@ func runRound(ctx context.Context, lsClient dynamic.ResourceInterface, clientset
 		"force-remove ConfChange applies within latency budget after scale-down (EN-1043)",
 		details.With(internal.Details{"elapsed": elapsed.String(), "budget": qrConfChangeLatencyBudget.String()}))
 
+	// WaitForVoters observes the leader's in-memory ConfState before the
+	// operator necessarily finishes updating the StatefulSet to one replica.
+	// Returning earlier runs the deferred scale-up concurrently with that
+	// in-flight scale-down, which can recreate a pod with fresh storage while
+	// the leader still has stale progress for its node ID. Pin the Kubernetes
+	// side of the transition before allowing cleanup to restore three replicas.
+	if !internal.WaitForStatefulSetReady(ctx, clientset, internal.LedgerStatefulSetName(), 1, qrScaleDownTimeout) {
+		log.Printf("quorum-recovery: StatefulSet did not settle at one replica")
+		return
+	}
+
 	sentinel.Verify(ctx, client, "after_quorum_recovery")
 }
--- a/tests/antithesis/workload/bin/cmds/main/singleton_driver_quorum_recovery/main.go
+++ b/tests/antithesis/workload/bin/cmds/main/singleton_driver_quorum_recovery/main.go
@@ -33,8 +33,9 @@ import (
 var qrSentinelLedger = internal.PrefixSentinel.WithSuffix("quorum-recovery")
 
 const (
-	qrScaleDownTimeout = 4 * time.Minute
-	qrScaleUpTimeout   = 4 * time.Minute
+	qrNormalizeTimeout = 90 * time.Second
+	qrScaleDownTimeout = 2 * time.Minute
+	qrScaleUpTimeout   = 2 * time.Minute
 
 	// qrConfChangeLatencyBudget is the "should be fast" threshold for the
 	// force-remove ConfChange to bring voters down to 1 after scale-down.
@@ -86,8 +87,14 @@ func runRound(ctx context.Context, lsClient dynamic.ResourceInterface, clientset
 		return
 	}
 	if current != 3 {
-		log.Printf("quorum-recovery: cluster not at N=3 (got %d), skipping", current)
-		return
+		log.Printf("quorum-recovery: normalizing cluster from %d to 3 replicas", current)
+		if err := internal.PatchReplicas(ctx, lsClient, internal.ClusterName, 3); err != nil {
+			log.Printf("quorum-recovery: cannot normalize replicas: %s", err)
+			return
+		}
+		if !internal.WaitForVoters(ctx, clusterClient, 3, qrNormalizeTimeout, internal.Details{"phase": "normalize"}) {
+			return
+		}
 	}
 
 	sentinel, err := internal.PreCommitSentinel(ctx, client, qrSentinelLedger)
--- a/tests/antithesis/workload/bin/cmds/main/singleton_driver_quorum_recovery/main.go
+++ b/tests/antithesis/workload/bin/cmds/main/singleton_driver_quorum_recovery/main.go
@@ -33,8 +33,8 @@ import (
 var qrSentinelLedger = internal.PrefixSentinel.WithSuffix("quorum-recovery")
 
 const (
-	qrScaleDownTimeout = 8 * time.Minute
-	qrScaleUpTimeout   = 15 * time.Minute
+	qrScaleDownTimeout = 4 * time.Minute
+	qrScaleUpTimeout   = 4 * time.Minute
 
 	// qrConfChangeLatencyBudget is the "should be fast" threshold for the
 	// force-remove ConfChange to bring voters down to 1 after scale-down.
@@ -49,7 +49,7 @@ const (
 func main() {
 	log.Println("composer: singleton_driver_quorum_recovery")
 
-	ctx, cancel := internal.SingletonContext()
+	ctx, cancel := internal.PlatformSingletonContext()
 	defer cancel()
 	dynClient, err := internal.NewK8sClient()
 	if err != nil {
@@ -136,13 +136,16 @@ func runRound(ctx context.Context, lsClient dynamic.ResourceInterface, clientset
 	// operator scale-down and every subsequent driver runs against a broken
 	// cluster for the rest of the experiment.
 	defer func() {
-		if err := internal.PatchReplicas(context.Background(), lsClient, internal.ClusterName, 3); err != nil {
+		cleanupCtx, cancel := context.WithTimeout(context.Background(), qrScaleUpTimeout)
+		defer cancel()
+
+		if err := internal.PatchReplicas(cleanupCtx, lsClient, internal.ClusterName, 3); err != nil {
 			log.Printf("quorum-recovery: cleanup PatchReplicas(3) failed: %s", err)
 		}
 		// Best-effort wait for the cluster to settle back to N=3 voters before
 		// releasing the singleton slot. If it doesn't recover we let the next
 		// driver iteration deal with it.
-		_ = internal.WaitForVoters(context.Background(), clusterClient, 3, qrScaleUpTimeout, details)
+		_ = internal.WaitForVoters(cleanupCtx, clusterClient, 3, qrScaleUpTimeout, details)
 	}()
 
 	deleted := 0
--- a/tests/antithesis/workload/bin/cmds/main/singleton_driver_quorum_recovery/main.go
+++ b/tests/antithesis/workload/bin/cmds/main/singleton_driver_quorum_recovery/main.go
@@ -33,7 +33,6 @@ import (
 var qrSentinelLedger = internal.PrefixSentinel.WithSuffix("quorum-recovery")
 
 const (
-	qrCooldown         = 5 * time.Minute
 	qrScaleDownTimeout = 8 * time.Minute
 	qrScaleUpTimeout   = 15 * time.Minute
 
@@ -77,15 +76,7 @@ func main() {
 		log.Printf("cannot create sentinel ledger: %s", err)
 	}
 
-	for {
-		select {
-		case <-ctx.Done():
-			return
-		case <-time.After(qrCooldown):
-		}
-
-		runRound(ctx, lsClient, clientset, clusterClient, client)
-	}
+	runRound(ctx, lsClient, clientset, clusterClient, client)
 }
 
 func runRound(ctx context.Context, lsClient dynamic.ResourceInterface, clientset kubernetes.Interface, clusterClient clusterpb.ClusterServiceClient, client servicepb.BucketServiceClient) {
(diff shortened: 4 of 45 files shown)


tests/antithesis/k8s/workload.yaml
```diff
@@ -105,6 +105,10 @@ spec:
         # MODEL_DUMP_BATCHES env at config-image build.
         - name: MODEL_DUMP_BATCHES
           value: "__MODEL_DUMP_BATCHES__"
+        # Bound the model command below the shortest supported platform run so
+        # Test Composer can observe at least one clean command completion.
+        - name: MODEL_MAX_SECONDS
```
NumaryBot, line 110: 🟠 **[major] Avoid expiring the model during a claimed restore**

With restores starting after roughly 30–45 seconds and potentially taking up to the configured 900-second lease, this 180-second deadline can expire after the sidecar has claimed the request. `fileTrigger.Fire` returns on `ctx.Done()` even when `withdraw()` fails, so the model command exits while the cluster may still be torn down or restoring, allowing Test Composer to start another command against that incomplete restore. Either allow claimed restores to finish before command exit or size the command bound beyond the restore lifecycle.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1681#issuecomment-5221202745

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1681#issuecomment-5221202745

internal/storage/dal/store.go
```diff
@@ -563,6 +570,39 @@ func NewStore(
...
+		storeOpenLockRetryDelay,
+	)
+}
+
+func openPebbleWithLockRetry(open func() (*pebble.DB, error), timeout, delay time.Duration) (*pebble.DB, error) {
+	deadline := time.Now().Add(timeout)
+	for {
+		db, err := open()
+		if err == nil {
+			return db, nil
```
paul-nicolas, line 591: [P1][blocking] The new lock-retry window can silently create a fresh empty database

**What.** `openPebbleWithLockRetry` re-invokes `pebble.Open(dirname, opts)` on a path whose existence was decided once by the `os.Stat(liveDir)` at store.go:486. The closure at :580 captures `dirname` only; nothing re-checks that `live/` still exists between attempts, and `ErrorIfNotExists` is never set anywhere in `store.go`.

**Why it matters.** The required interleaving is temporal, not simultaneous, so the "the peer either holds the lock or has renamed, never both" intuition does not save it — the retry loop spans 5 s of wall clock:

```
t=0     incoming: os.Stat(live/) -> EXISTS        (store.go:486)
t=0+    incoming: pebble.Open(live/) -> EAGAIN    => retry loop ENTERED
t=T     peer:     RestoreCheckpoint -> closeDBSafe(oldDB)   (store.go:1327) => lock released
t=T+    peer:     os.Rename(live/, live.discard/)           (store.go:1342) => live/ GONE
t=T++   incoming: retry (<=50ms later) -> no live/ -> fresh empty DB, err == nil
```

Pebble makes this silent rather than loud: `prepareAndOpenDirs` (open.go:655-707) re-creates the missing tree, then open.go:310-316 finds no MANIFEST and, with `ErrorIfNotExists=false`, calls `versions.create(...)` — a brand-new empty database returned with `err == nil`. `live/` stays absent for the whole of restore steps 4-6, which is many multiples of the 50 ms retry period.

**Reproduced**, cross-process, two real OS processes, against the production entry point `openPebbleAfterPreviousProcess` with the PR's own 5 s / 50 ms constants — 3 of 5 runs:

```
PROBE D: openPebbleAfterPreviousProcess succeeded after 539.21825ms
PROBE D RESULT: keys=0 ; peer had already renamed live/ aside=true
PROBE D CONFIRMED: cross-process retry returned a FRESH EMPTY database
                   (business-key lost, checkpoint 0 never adopted)
```

The other 2/5 failed loudly with `ENOENT` from `objstorageprovider.Open` — and left a freshly `mkdir`'d `live/` behind, which can then be renamed to `live.discard/` and clobber the peer's rollback target.

**This is a regression, not a pre-existing gap.** Pre-PR, lock contention was fail-fast: the process died and k8s restarted it cleanly. The empty-DB outcome requires `live/` present at `os.Stat` and absent at a *later retry* — a state that could not exist before this PR, because there was no later retry.

**Nothing downstream catches it.** An empty Pebble store is indistinguishable from a first boot: `LoadPersistedConfig` returns `(nil, nil)` on `ErrNotFound` (bootstrap/persisted_config.go:31-47) and `config_validation.go:65-81` then takes the `persisted == nil` "First boot detected" branch, skipping node-id / cluster-id / schema-version validation and writing the current config in. The rename destroys only the FSM state — the etcd WAL and `checkpoints/` are siblings — so the node boots with the full Raft WAL intact but applied index reading 0, joins, and serves reads. Against AGENTS.md that breaks invariant 1 (cache divergence at the same applied index), invariant 7 (impossible states fail loudly) and invariant 8 (audit is business truth).

Reachability is the PR's own stated premise: the doc comment at store.go:573-577 describes exactly the two-process overlap, and `RestoreCheckpoint` is the Raft snapshot-restore path on a follower (capabilities.go:196 -> state/synchronizer.go:32) — a node being killed and restarted is precisely a node that has fallen behind. Routine under fault injection.

**Fix.** Pass `dirname` into `openPebbleWithLockRetry` and `os.Stat` it before each *retry*; if it vanished, return a hard error instead of retrying. That keeps the first attempt's semantics untouched (the legitimate "creating new database" branch at :507-509 still works) and also removes the `ENOENT`/leftover-directory variant. Setting `opts.ErrorIfNotExists` unconditionally is not safe — `opts` is reused by `RestoreCheckpoint` at :1409 and :1376. A deterministic regression test on the Probe A shape (no sleeps, no subprocess) pins it.

tests/antithesis/workload/internal/assertion_semantics_test.go
```diff
@@ -0,0 +1,172 @@
+package internal
```
paul-nicolas, line 1: [P2][blocking] This module's tests run in no CI job, so every regression barrier this PR adds is inert

**What.** `tests/antithesis/workload/go.mod` declares a separate module (`github.com/formancehq/ledger/v3/tests/antithesis/workload`), no `go.work` exists anywhere in the repo, and nothing runs `go test` against it:

```
$ grep -rn "antithesis/workload" .github/workflows/ Justfile scripts/
rc=1   (no matches)

$ find . -name "go.work*" -not -path "*/node_modules/*"
(no output)

$ GOROOT= go list ./... | grep -c "antithesis/workload"
0
```

`just test` (Justfile:80-81, `go test -race ./... -timeout 20m`) is rooted in the root module and cannot descend into a nested one.

**It compiles, but it is never tested.** The module's non-test code *is* built on every PR — Tests-Model -> `run_model_test.sh:388` (`go build -o "$DRIVER_BIN" ./bin/cmds/model/singleton_driver_model`) and the image build's `Dockerfile:34` (`go build -race ./internal/...`). `go build` ignores `_test.go` entirely, so the new test files are not even type-checked there. The new `.github/workflows/antithesis.yml` is `on: workflow_dispatch:` only (lines 3-20), never runs on a PR, and contains no `go test`.

**This is an omission, not a convention** — the repo's other nested module has two hooks:
```
Justfile:365-366             operator-test: cd misc/operator && go test ./...
scripts/agent-check-pr:132   go -C misc/operator test ./... -timeout 20m
```

**Impact.** `assertion_semantics_test.go`, `driver_test.go`, `singleton_driver_scaling/main_test.go` and `singleton_driver_model/restore_test.go` are the guardrail this PR is built on — the assertion-semantics catalogue, the "singletons must be finite" AST checks, the platform-bound checks, and the claimed-restore test. A later change that reverts `AlwaysOrUnreachable` to `Always`, re-adds a `for {}` to a singleton `main`, or drops the `ctxDone = nil` guard merges green. Note `assertion_semantics_test.go:53-68` even reads root-module server files, so it is guarding `internal/infra/**` too.

**Fix is cheap** — the tests already pass in ~6 s:
```
$ cd tests/antithesis/workload && GOROOT= go test -count=1 ./...
ok  .../bin/cmds/main/singleton_driver_scaling  0.766s
ok  .../bin/cmds/model/singleton_driver_model   1.638s
ok  .../internal                                3.283s
```
Add a `just` recipe mirroring `operator-test`, call it from a `main.yml` job, and gate it in `scripts/agent-check-pr`.

misc/operator/internal/controller/raft_scaledown_test.go
```diff
@@ -117,7 +117,11 @@ func TestIsPodCrashed_Pending(t *testing.T) {
 	_, err = cs.CoreV1().Pods(ns).UpdateStatus(ctx, pod, metav1.UpdateOptions{})
```
paul-nicolas, line 117: [P2][blocking] The raftScaleDown fix has no executable coverage — this file never runs, and the assertion it flips is a proxy

Two independent legs; either alone leaves the fix unguarded.

**Leg 1 — the flipped assertion does not guard the fix.** The actual fix is the deletion of the caller-side gate in `raftScaleDown`:
```go
-  if neverJoined := isPodNeverReady(ctx, clientset, ledger.Namespace, pod); neverJoined {
-      logger.Info("pod was never ready, skipping Raft removal (node never joined cluster)", ...)
-      continue
-  }
```
`isPodCrashed` was already correct for the NotFound case *before* this PR — `git show 93876adf6:...raft_scaledown_test.go` shows `TestIsPodCrashed_NotFound` asserting `require.True(...)` and passing. The bug shipped anyway, because `isPodNeverReady` also returned true for NotFound and short-circuited ahead of `isPodCrashed` in the caller. Nothing exercises the caller:
```
$ grep -rln "raftScaleDown\|removeNode(" --include="*_test.go" misc/operator
rc=1   (no matches)
```
Proven by mutation: reinstating an equivalent caller skip gate (keeping the new `isPodCrashed` Pending behaviour) leaves the package **byte-for-byte identical** — `ok 0.946s` baseline, `ok 0.946s` mutant; tagged run shows the same 3 pre-existing failures and not one new one. The regression is fully reintroduced and no test changes state.

**Leg 2 — this file is excluded from every run.** `raft_scaledown_test.go:1` is `//go:build integration` (12 files in the package are), and no recipe or CI job passes the tag:
```
$ cd misc/operator && GOROOT= go test ./internal/controller/ -run TestIsPodCrashed
ok  github.com/formancehq/ledger/misc/operator/internal/controller  0.858s [no tests to run]

$ grep -rn "integration" .github/ Justfile scripts/
Justfile:87:# Run all tests across all modules (unit + integration + e2e)
```
The sole hit in the entire CI surface is a comment. `main.yml:209-211`, the only operator job, is `go build ./...`; `just operator-test` (untagged) is called by no CI job at all.

**Corroboration:** the tagged suite has already rotted to red, which is only possible because nothing runs it — `TestReconcile_EvenReplicas`, `TestReconcile_IngressEnabledNoHosts`, `TestReconcile_IngressGrpcEnabledNoHosts` all FAIL at HEAD. These are pre-existing and unrelated (`reconcile_validation_test.go` last touched at 76862ea0b, outside this PR), **but they must be repaired first** — simply adding `-tags integration` to CI turns the pipeline red immediately.

**Fix.** A test that drives `raftScaleDown` itself over a fake clientset + cluster client and asserts the voter is removed for a recreated-Pending pod, in a file that actually runs. Reinstating the skip gate must then fail.

Minor: the body of the original finding referenced `TestIsPodCrashed_NotFound`; the assertion this PR actually flips is here in `TestIsPodCrashed_Pending` (lines 117-124). The proxy argument is unaffected.

internal/storage/dal/store.go
```diff
@@ -563,6 +570,39 @@ func NewStore(
...
+		}
+		if !isPebbleLockContention(err) || time.Now().Add(delay).After(deadline) {
+			return nil, err
+		}
+		time.Sleep(delay)
+	}
+}
+
+func isPebbleLockContention(err error) bool {
```
paul-nicolas, line 600: [P2][non-blocking] The errno clauses are the only ones reachable in production, and no test exercises them

`isPebbleLockContention` has three clauses, but the new `TestOpenPebbleWithLockRetryWaitsForPreviousProcess` (store_operations_test.go:720) opens both databases in a *single* process. Pebble short-circuits on a process-local map before ever reaching `FcntlFlock` — `vfs/file_lock_unix.go:41-49` returns `errors.New("lock held by current process")` — so the test can only reach the `strings.Contains` clause.

Proven by mutation, not inspection:
- deleting **both** `errors.Is` clauses -> whole package still green: `ok .../internal/storage/dal 5.962s`
- deleting only the `strings.Contains` clause -> FAIL, `Wraps: (2) lock held by current process`

A two-process probe confirms the shipped predicate is **correct today** — real cross-process contention yields a bare `syscall.Errno`:
```
CHILD_ERR=resource temporarily unavailable
CHILD_TYPE=syscall.Errno
CHILD_IS_EAGAIN=true   CHILD_IS_EACCES=false   CHILD_IS_LHBCP=false
CHILD_PREDICATE=true
```
So nothing is broken; this is regression protection for the exact mechanism the PR exists to ship. There is no other coverage — `grep -rn 'isPebbleLockContention'` returns only the call site (:593) and the definition (:600).

**Fix.** `isPebbleLockContention` is an unexported free function over a plain `error`; a ~15-line table test needs no second process. Cover `syscall.EAGAIN`, `syscall.EACCES`, the same-process string, and a non-lock error. Mutating away either errno clause must then fail.

tests/antithesis/workload/internal/k8s_test.go
```diff
@@ -126,6 +129,47 @@ func TestWaitForStatefulSetReady_ReportsReady(t *testing.T) {
...
+			ReadyReplicas:   1,
+			CurrentRevision: "rev-1",
+			UpdateRevision:  "rev-1",
+		},
+	}
+
+	if statefulSetReady(sts, 1) {
+		t.Fatal("one ready pod with spec.replicas=3 must not be treated as a completed scale-down to one")
+	}
+}
```
paul-nicolas, line 171: [P2][non-blocking] Re: `TestStatefulSetReady_RequiresDesiredReplicas` (lines 155-171) — the fixture does not discriminate on the guard the test is named for

The fixture omits `Status.Replicas`, so it defaults to 0 while `expected` is 1. The input is therefore rejected **twice over**: by the `*sts.Spec.Replicas == expected` guard (the clause under test) *and*, independently, by `sts.Status.Replicas == expected`. The redundant rejection is what hides the regression.

Reproduced — deleting `sts.Spec.Replicas != nil && *sts.Spec.Replicas == expected &&` from k8s.go:351-352 leaves the whole package green:
```
--- PASS: TestStatefulSetReady_RequiresObservedReplicas (0.00s)
--- PASS: TestStatefulSetReady_RequiresDesiredReplicas (0.00s)
--- PASS: TestWaitForStatefulSetReady_ReportsReady (2.00s)
--- PASS: TestWaitForStatefulSetReady_NotFoundFailsFast (2.00s)
ok  .../tests/antithesis/workload/internal  2.887s
```
(Overlay liveness was proven first — this package AST-scans at run time, so a silently-ignored overlay would look identical to a surviving mutation. Replacing the body with `return true` correctly FAILs both tests.)

**The guard is genuinely load-bearing**, so this is a test gap and not dead code: it rejects a StatefulSet whose observed replica count already matches the target while `spec.replicas` still holds the pre-scale-down value — the quorum-recovery race the comment at lines 151-154 describes, and reachable in production because Antithesis kills pods so `Status.Replicas` transiently drops to 1 before the operator writes the new spec.

**Fix (one line).** Add `Replicas: 1,` to the fixture's `Status` block. Verified: the test still passes against current `k8s.go`, and FAILs when the clause is removed —
```
k8s_test.go:170: one ready pod with spec.replicas=3 must not be treated as a completed scale-down to one
--- FAIL: TestStatefulSetReady_RequiresDesiredReplicas (0.00s)
```

tests/antithesis/workload/bin/cmds/main/singleton_driver_quorum_recovery/main.go
```diff
@@ -195,5 +204,16 @@ func runRound(ctx context.Context, lsClient dynamic.ResourceInterface, clientset
...
+	// WaitForVoters observes the leader's in-memory ConfState before the
+	// operator necessarily finishes updating the StatefulSet to one replica.
+	// Returning earlier runs the deferred scale-up concurrently with that
+	// in-flight scale-down, which can recreate a pod with fresh storage while
+	// the leader still has stale progress for its node ID. Pin the Kubernetes
+	// side of the transition before allowing cleanup to restore three replicas.
+	if !internal.WaitForStatefulSetReady(ctx, clientset, internal.LedgerStatefulSetName(), 1, qrScaleDownTimeout) {
+		log.Printf("quorum-recovery: StatefulSet did not settle at one replica")
+		return
+	}
```
paul-nicolas, line 216: [P2][non-blocking] The new StatefulSet-timeout path returns without verifying the sentinel, unlike its sibling

The new guard returns with only a log, while the voters-timeout path ~24 lines above verifies under a comparably unsettled cluster:
```go
// main.go:189-196
if !internal.WaitForVoters(ctx, clusterClient, 1, qrScaleDownTimeout, details) {
    sentinel.Verify(ctx, client, "after_quorum_recovery_timeout")   // verifies
    return
}
// main.go:213-216
if !internal.WaitForStatefulSetReady(ctx, clientset, internal.LedgerStatefulSetName(), 1, qrScaleDownTimeout) {
    log.Printf("quorum-recovery: StatefulSet did not settle at one replica")
    return                                                          // does not
}
sentinel.Verify(ctx, client, "after_quorum_recovery")
```

**Reachable.** This PR cuts `qrScaleDownTimeout` 8m -> 2m and `qrScaleUpTimeout` 15m -> 2m, and moves the command to `PlatformSingletonContext` (6 min, driver.go:27). Worst case to reach :213 is normalize (<=90 s) + `WaitForVoters(1, 2m)` ~3.5 m, then a 2 m STS wait lands ~5.5 m against a 6 m budget. Both "STS deadline fired, ctx alive" and "ctx expired first" are reachable.

**Neither defence for skipping holds.** `WaitForStatefulSetReady` returns false on its own 2 m deadline and on a NotFound STS (k8s.go:321-348), not only on `ctx.Done`, so the ctx-alive case is real and a verify there is meaningful — and the sibling at :194 has the identical ambiguity and verifies anyway. Nor can the call misfire: `Sentinel.Verify` (sentinel.go:66-91) hard-asserts only on `codes.NotFound`; a cancelled/deadline-exceeded error is not NotFound and falls through harmlessly.

**The `return` also does not do what the comment implies.** The deferred cleanup at main.go:145-156 builds `cleanupCtx` from `context.Background()` and re-patches replicas to 3 on **every** return path, so the scale-up races the in-flight scale-down whether you return at :215 or fall through. The only net effect of the early return is losing the verification — on the path where the cluster just went through a force-remove, i.e. where the file's own strongest correctness assertion matters most.

**Fix.** Insert `sentinel.Verify(ctx, client, "after_quorum_recovery_sts_timeout")` before the return, mirroring :194. A distinct label keeps the two timeout paths separable in triage.

paul-nicolas (commented):

Reviewed at head `392c0682e`. Four independent reviewers (Codex, plus three Claude reviewers split by area: core production code / operator / antithesis harness + CI), then **one adversarial verifier per finding**, each instructed to refute rather than confirm and to default to "invalid" on ambiguous evidence.

**Verification outcome: 12 findings entered, 6 survive and are posted, 3 were rejected outright, 3 were downgraded to P3.** Every posted finding below was independently reproduced — by mutation, by a cross-process probe, or by literal command output.

### Posted inline

| Severity | Blocking | Location | Finding |
|---|---|---|---|
| P1 | yes | `internal/storage/dal/store.go:591` | Lock-retry window can silently create a fresh empty database |
| P2 | yes | `tests/antithesis/workload/internal/assertion_semantics_test.go:1` | This module's tests run in no CI job |
| P2 | yes | `misc/operator/.../raft_scaledown_test.go:117` | raftScaleDown fix has no executable coverage |
| P2 | no | `internal/storage/dal/store.go:600` | Errno clauses of the lock predicate are untested |
| P2 | no | `tests/antithesis/workload/internal/k8s_test.go:171` | Fixture does not discriminate on the guard under test |
| P2 | no | `singleton_driver_quorum_recovery/main.go:216` | STS-timeout path skips the sentinel verify |

The P1 is the one to read first: it is a **regression introduced by this PR** (lock contention was fail-fast before; it is now a 5 s retry, and the retry is precisely where a peer's `RestoreCheckpoint` deletes `live/` under it), it was reproduced with two real OS processes against the production entry point in 3 of 5 runs, and no boot-time check catches the result — an empty Pebble store is indistinguishable from a first boot.

The two blocking P2s share a theme worth calling out: **this PR's regression barriers do not execute.** The antithesis workload module and the operator's integration-tagged tests are both invisible to CI, so the tests added here would not catch the regressions they were written for. Related: the operator's tagged suite has already rotted to 3 pre-existing failures, which must be repaired before `-tags integration` can be switched on.

### Rejected during verification (not posted)

- **`sentinel.go:80` "module-wide Reachable registration"** — refuted. The antithesis property id is the bare message string, so 61 per-binary registrations are *one* property needing *one* example run-wide, not 61 demanding examples each. And the PR removes a 5-minute startup cooldown plus a never-returning `for {}` that previously made `Verify` unreachable inside a 20-minute run — it makes the property strictly easier to satisfy, not harder. Direction inverted.
- **`store.go:476` "hold the lock across the pre-open reconcile"** — refuted. `RestoreCheckpoint` closes the DB and releases the lock *before* mutating the directory (`closeDBSafe` at :1327, rename at :1342), so in the window described the incoming `pebble.Open` already succeeded on its first attempt pre-PR. The retry loop is never entered; the PR changes nothing here. Pre-existing and not worsened.
- **`block.go:109-141` "bound by remaining time, not a fixed attempt count"** — refuted. The premise misreads its own grep (21 `ErrSkip` *statements* across 16 functions, not 21 of 26 blocks), and it omits the change that matters: selection moved from random to deterministic round-robin in declaration order, and every group declares its seeder before its dependents — so the balance chain resolves inside pass 1. The proposed fix would make the command always burn its full 6-minute budget, which is exactly the behaviour this PR removes.

### Re-review classification

One prior NumaryBot finding existed (`k8s/workload.yaml:110`, "avoid expiring the model during a claimed restore"): **FIXED** at this head. The `ctxDone = nil` change in commit `392c0682e` resolves it, confirmed by mutation — the new `TestFileTriggerWaitsForClaimedRestoreAfterCancellation` fails against the pre-fix `restore.go` with `claimed restore returned on cancellation: context canceled`.

Fixed 1 / still valid 0 / outdated 0.

### P3 (not posted inline)

- `tests/antithesis/workload/internal/block/block.go:98` — when ctx expires inside the setup retry loop, `setupErr` stays nil and a passing "should be able to setup scenario" `Always` is emitted with `error: null` for a setup that never ran. *(Downgraded from blocking: the 52 block-property registrations that follow are identical to what healthy executions emit and are aggregated per message across the run, so only the misleading `Always` record is a real defect. Note `TestScenarioBlockRunnerIsFinite` now rejects restoring the old `for {}`, so attribute the cancellation instead: `if setupErr == nil && ctx.Err() != nil { setupErr = ctx.Err() }`.)*
- `tests/antithesis/workload/restore-orchestrator.sh:256` — the restore claim is not durable: `$REQ.claimed` is deleted before any work, so a sidecar *container* restart (its own 128Mi limit, `workload.yaml:164`) orphans the cycle and `Fire` blocks until the run ends. *(Downgraded from blocking. The unbounded post-claim wait is real and was reproduced — 6 lease resets after cancellation, `Fire` still blocked — but it is deliberate and correct while the sidecar lives, since the orchestrator always answers on both branches. Fix it in the orchestrator by keeping the marker until the response is written; do **not** add a driver-side deadline, which would revert this PR and reintroduce the state-poisoning it fixes.)*
- `internal/infra/node/applier.go:410` — the per-future `lifecycle.SendEvent` loses the tracker dedup that `assert.Reachable` had, so a term truncation with many stragglers does one `json.Marshal` per straggler on the committer goroutine. *(Downgraded from P2: `assert.Reachable` already paid a `runtime.Caller` every iteration before the dedup check, so the delta is small, and the path only fires on term truncation. Cleanliness, not latency risk.)*
- `internal/storage/dal/store.go:573-577` — the comment claims "retrying only lock-contention errors preserves fail-fast startup for configuration errors", but `errors.Is(err, syscall.EACCES)` also matches the `*fs.PathError` from `os.Create` in `vfs.defaultFS.Lock`, so a wrong-fsGroup / read-only-volume misconfiguration now takes 5 s and ~100 attempts to surface.
- `tests/antithesis/workload/internal/assertion_semantics_test.go:85-102` — five of the eight converted sites were renamed to snake_case in this PR, but the `optionalDiagnostics` fence still lists the old message strings, so it no longer pins them.
- `internal/infra/state/machine.go:1075` and `:1081` — two overlapping events fire for one predicted-index rejection, with the same payload; the cache-epoch branch below emits only one.
- `pkg/scenario/marketplace.go:105` — `findFundedMarketplaceAccount` issues up to `count` sequential `GetAccount` RPCs (50 for purchase) and never checks `ctx.Err()`.
- `misc/operator/internal/controller/raft_scaledown.go:315-319` — the deleted `isPodNeverReady`'s doc comment is stranded above `isPodCrashed`'s own comment, documenting a function that no longer exists. Line 321 also says "Pending without a started container" while :339 treats `PodPending` as crashed unconditionally.
- `singleton_driver_scaling_structured/main.go:4` and `singleton_driver_quorum_recovery/main.go:15-17` — header comments left stale by de-looping: the first still describes "3 -> 5 -> 3 -> 7 -> 3" while `scalingCycle = []int64{5, 3}`; the second still says "runs sparsely (>= 5 minutes between rounds)" after the cooldown and round loop were deleted.
- *Concern, not a finding* — `tests/antithesis/Justfile:118-152` uses `docker build` + separate `docker push`, and `check-k8s-image-freshness` reads `docker image inspect`, both of which need the image in the local daemon store. Every other docker path in this repo running under `nscloud-setup-buildx-action` uses `docker buildx build --push`. If that action installs a remote/container-driver builder by default, `docker build` without `--load` leaves no local image. I could not determine the driver from the repository — running the workflow once, or adding `--load`, settles it.

### Checked and clean

Verified as correct with no finding: the `Always` -> `AlwaysOrUnreachable` / `Unreachable` conversions are failure-equivalent (they differ only in `MustHit`); removing the two stats relations is sound (`PostingCount`/`RevertCount` come from the usage snapshot, `TransactionCount`/`LogCount` from the primary store); the marketplace block reorder is load-bearing and does not turn reverts into block failures (insufficient funds maps to `FailedPrecondition`, handled separately); randomness consumption is unchanged; no FSM determinism impact (all converted sites are output-only); the operator's widened voter-removal cannot target a surviving member and is strictly skip -> force, never consensus -> force; voter removal provably precedes PVC deletion on every path; and `.github/workflows/antithesis.yml` has no script-injection vector — all three `workflow_dispatch` inputs are routed through job-level `env:` and quoted, `permissions: contents: read` is set, both third-party actions are SHA-pinned, `duration_minutes` is regex-validated, and the registry key reaches `docker login` via `--password-stdin`.

---

**DECISION: REQUEST CHANGES** — 3 blocking (1x P1, 2x P2), 3 non-blocking P2 posted inline, 9 P3 + 1 concern listed above. Residual risk: MEDIUM (the P1 is a genuine data-loss regression, but it needs the two-process overlap). CI is green on all 15 checks, which is consistent with the two coverage findings: the tests that would catch these do not run.

internal/storage/dal/store.go
```diff
@@ -563,6 +570,39 @@ func NewStore(
 	return store, nil
 }
 
+// openPebbleAfterPreviousProcess tolerates the short overlap between a
+// terminating process and its replacement. Kubernetes normally waits for the
+// old container to exit, but fault injection can expose a window where the new
+// process starts before Pebble's LOCK is released. Retrying only lock-contention
+// errors preserves fail-fast startup for corruption and configuration errors.
+func openPebbleAfterPreviousProcess(dirname string, opts *pebble.Options) (*pebble.DB, error) {
```
NumaryBot, line 578: 🔴 **[blocker] Prevent lock retries from creating an empty database**

If another process renames `live/` after an initial lock-contention error, the next `pebble.Open` retry recreates the now-missing directory and succeeds with a fresh empty database. This can silently discard the existing persisted state rather than failing startup, contrary to the persisted-configuration safety requirement in [AGENTS.md:15](AGENTS.md#L15); recheck existence on retries or configure Pebble not to create a database when `live/` originally existed.

tests/antithesis/workload/bin/cmds/main/singleton_driver_quorum_recovery/main.go
```diff
@@ -195,5 +204,16 @@ func runRound(ctx context.Context, lsClient dynamic.ResourceInterface, clientset
 		"force-remove ConfChange applies within latency budget after scale-down (EN-1043)",
 		details.With(internal.Details{"elapsed": elapsed.String(), "budget": qrConfChangeLatencyBudget.String()}))
 
+	// WaitForVoters observes the leader's in-memory ConfState before the
+	// operator necessarily finishes updating the StatefulSet to one replica.
+	// Returning earlier runs the deferred scale-up concurrently with that
+	// in-flight scale-down, which can recreate a pod with fresh storage while
+	// the leader still has stale progress for its node ID. Pin the Kubernetes
+	// side of the transition before allowing cleanup to restore three replicas.
+	if !internal.WaitForStatefulSetReady(ctx, clientset, internal.LedgerStatefulSetName(), 1, qrScaleDownTimeout) {
```
NumaryBot, line 213: 🟠 **[major] Verify the sentinel on StatefulSet timeout**

When voter convergence succeeds but the StatefulSet does not settle before this new timeout, the driver returns without checking the previously committed sentinel. Data loss during this unsettled recovery path therefore produces no sentinel assertion, unlike the preceding voter-timeout path, weakening the quorum-recovery test precisely when recovery is incomplete.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1681#issuecomment-5221202745

PR #1683, opened 2026-08-10
fix(chapters): archive chapters oldest-first

Azorlogh:
Archival was unordered: `ArchiveChapter` accepted any `CLOSED` chapter, so chapter 3 could be archived while chapter 2 was still merely closed.

## Why that is a problem

The purge that follows a confirm deletes **everything** up to the archived chapter's close sequence. Archiving out of order therefore leaves an older, un-archived chapter's logs retained *below* the archive boundary, and that has two consequences:

- The checker's replay skips every log at or below the boundary (the baseline checkpoint and the audit chain are authoritative down there), so those retained logs become a **permanently unverified window** in the middle of history.
- Cold storage keeps a **hole no later archive ever fills** — nothing ever comes back for chapter 2.

It also silently violates an assumption the archive-aware checker passes already make, namely that "below the archive boundary" and "archived" are the same statement. That mismatch is not hypothetical: it is the live-path trigger of the false `EXCLUSION_RECORD_MISMATCH` fixed separately in `fix/exclusion-check-archived-ranges` — the same false positive is reachable on a never-restored cluster purely by archiving out of order.

## The change

`processArchiveChapter` now admits only the chapter immediately after the archived prefix, rejecting anything else with a new `CHAPTER_ARCHIVE_OUT_OF_ORDER` reason whose metadata carries `blockingChapterId`. Archived chapters are consequently always a contiguous prefix of history.

**The prefix is tracked explicitly, not inferred from the tracker.** This is the substance of the second review round, and it turns on what each structure means:

- `ChapterTracker.allChapters` is the FSM's working set of chapters **whose data is still in hot storage** — it exists so the apply path never reads Pebble (invariant #3). The confirm-archive purge drops archived chapters from it (`WriteSet.Merge` → `DeleteChapter`), so *absence there means "purged", not "archived"*.
- `ZoneGlobal/SubGlobChapters` is the durable, complete registry — every chapter that ever existed, never deleted — and it is what recovery, the query path, cold-read routing, the checker and the backup backfill read.

So asking the tracker "is chapter N archived?" asks a history question of a residency structure. The first version of this PR did exactly that and was wrong twice over: it rejected every sequential archive past the first (by then the predecessor is purged), and it made the verdict restart-dependent, since recovery reloads archived rows as `ARCHIVED` — a restarted replica would accept an order a non-restarted one rejects and diverge the FSM (invariant #2).

`ChapterTracker.archivedThroughID` records the fact directly. `ConfirmArchiveChapter` advances it through a `Scope` method rather than at merge time, so a later order in the *same* proposal sees it — a batched `ApplyRequest` can carry confirm N and archive N+1, and `WriteSet.GetChapterByID` already made that pair work. `Merge` propagates it to the FSM tracker beside `NextChapterID`, and both recovery paths (boot and follower checkpoint-sync, which share `RecoverState`) derive it from the persisted rows as the longest run of `ARCHIVED` from chapter 1 — stopping at a gap or a missing row rather than counting past it. No new persisted state: the marker is derived from rows the registry already keeps forever, so it adds no projection for invariant #8 to cover.

Requiring the predecessor to be fully `ARCHIVED` rather than merely `ARCHIVING` matters: an in-flight upload must not let a successor's confirm, and therefore its purge, land first. A side effect is that at most one chapter is ever in `ARCHIVING`, which shrinks the archiver's state space.

The gate lives in the FSM rather than the CLI because the checker's correctness assumption now rests on it and it must hold identically on every node. It is deterministic: a scalar comparison, and the recovery derivation scans a fixed id range rather than iterating a map, so the accept/reject verdict the audit chain records cannot depend on map order.

## Trade-off, stated plainly

This removes an escape hatch. A chapter whose upload cannot complete — persistent cold-storage failure, or an object that fails the identity check added in #1679 — now blocks archival of newer chapters instead of being stepped over, so hot storage grows until the underlying problem is fixed. That is intentional: a loud, recoverable stall beats a silent permanent gap in cold storage. If a step-over is ever genuinely needed it should be an explicit flag on the order rather than the default behaviour.

## Blast radius

Nothing relied on the old freedom. The chapter scheduler only ever proposes `CloseChapter`; archival is exclusively operator-driven (`ledgerctl chapters archive <id>`) or driver-driven in tests, and both go oldest-first in practice. No automated component picks a chapter id.

## Tests

- `TestProcessArchiveChapter_RejectsOutOfOrder` — table-driven over one-ahead, several-ahead, skipping the first chapter, and re-archiving inside the prefix; asserts the typed error, the blocking id, and that no log is emitted.
- `TestProcessArchiveChapter_AllowsPrefixSuccessor` — the prefix successor archives normally, for an empty prefix and a long one.
- `TestChapterTrackerArchivedThroughDerivation` — the recovery side, which mocks cannot reach: derivation from persisted rows via both `NewChapterTracker` and `Reset`, including a status gap and a missing row stopping the prefix.
- `TestChapterTrackerCloneCarriesArchivedThrough` — a clone that dropped the marker would read 0 and reject every archive.
- `tests/e2e/business/audit_purge_test.go` archives **two** chapters sequentially through the real tracker, which is what caught the first round's bug. Verified failing both on that version and on a marker that never advances; passing now, along with the rest of the chapter/cold-storage e2e specs and the full `./internal/...` unit suite.

Docs: the ordering rule and the residency-vs-history distinction are documented in the chapter lifecycle page and the `chapters archive` CLI reference. While confirming the tracker's contract I also found and corrected two stale claims in `lifecycle.md`: `ListChapters` reads Pebble (not `Machine.AllChapters()`), and the documented `MemorySnapshot` message does not exist — chapter state is rebuilt from the persisted rows, including on follower checkpoint-sync.

### DIFF
--- a/internal/domain/processing/processor_chapter.go
+++ b/internal/domain/processing/processor_chapter.go
@@ -119,9 +119,9 @@ func processArchiveChapter(order *raftcmdpb.ArchiveChapterOrder, ctx *Context) (
 	// failures, which are hash-chained, would diverge across replicas.
 	archivedThrough := s.GetArchivedThroughChapterID()
 	if order.GetChapterId() <= archivedThrough {
-		return nil, &domain.ErrChapterArchiveOutOfOrder{
-			ChapterID:         order.GetChapterId(),
-			BlockingChapterID: archivedThrough + 1,
+		return nil, &domain.ErrChapterAlreadyArchived{
+			ChapterID:                order.GetChapterId(),
+			ArchivedThroughChapterID: archivedThrough,
 		}
 	}
 
@@ -181,6 +181,13 @@ func processConfirmArchiveChapter(order *raftcmdpb.ConfirmArchiveChapterOrder, c
 	// gate, so refusing it surfaces the anomaly (audited failure, and the
 	// archiver retries visibly) instead of silently papering over it.
 	archivedThrough := s.GetArchivedThroughChapterID()
+	if order.GetChapterId() <= archivedThrough {
+		return nil, &domain.ErrChapterAlreadyArchived{
+			ChapterID:                order.GetChapterId(),
+			ArchivedThroughChapterID: archivedThrough,
+		}
+	}
+
 	if order.GetChapterId() != archivedThrough+1 {
 		return nil, &domain.ErrChapterArchiveOutOfOrder{
 			ChapterID:         order.GetChapterId(),
--- a/internal/domain/processing/processor_chapter.go
+++ b/internal/domain/processing/processor_chapter.go
@@ -101,14 +101,6 @@ func processSealChapter(order *raftcmdpb.SealChapterOrder, ctx *Context) (*commo
 // sequence range the worker needs).
 func processArchiveChapter(order *raftcmdpb.ArchiveChapterOrder, ctx *Context) (*commonpb.LogPayload, domain.Describable) {
 	s := ctx.Scope
-	chapterReader, ok := s.GetChapterByID(order.GetChapterId())
-	if !ok {
-		return nil, &domain.ErrChapterNotFound{ChapterID: order.GetChapterId()}
-	}
-
-	if chapterReader.GetStatus() != commonpb.ChapterStatus_CHAPTER_CLOSED {
-		return nil, &domain.ErrChapterNotClosed{ChapterID: order.GetChapterId()}
-	}
 
 	// Archived chapters must form a contiguous prefix of history, so the only
 	// archivable chapter is the one right after the archived prefix. The purge
@@ -118,11 +110,34 @@ func processArchiveChapter(order *raftcmdpb.ArchiveChapterOrder, ctx *Context) (
 	// them, leaving a permanently unverified window, and where cold storage has
 	// a hole no later archive ever fills.
 	//
-	// The prefix is read from the tracker's explicit marker rather than inferred
-	// from which chapters are still resident: the purge drops archived chapters
-	// from the tracker, so residency answers "does this chapter still have hot
-	// data", not "is it archived".
-	if archivedThrough := s.GetArchivedThroughChapterID(); order.GetChapterId() != archivedThrough+1 {
+	// The prefix comes from the tracker's explicit marker, and chapters inside it
+	// are rejected BEFORE the residency lookup below: an archived chapter is
+	// absent from the tracker on a running node (the purge drops it) and present
+	// with status ARCHIVED after a restart (recovery reloads its row), so a
+	// lookup-first order would answer the same Raft order with two different
+	// reasons — CHAPTER_NOT_FOUND or CHAPTER_NOT_CLOSED — and the audited
+	// failures, which are hash-chained, would diverge across replicas.
+	archivedThrough := s.GetArchivedThroughChapterID()
+	if order.GetChapterId() <= archivedThrough {
+		return nil, &domain.ErrChapterArchiveOutOfOrder{
+			ChapterID:         order.GetChapterId(),
+			BlockingChapterID: archivedThrough + 1,
+		}
+	}
+
+	// Past the prefix the lookup is representation-independent: a chapter that is
+	// not archived is never purged, so it is resident on every node, and one that
+	// never existed has no row anywhere.
+	chapterReader, ok := s.GetChapterByID(order.GetChapterId())
+	if !ok {
+		return nil, &domain.ErrChapterNotFound{ChapterID: order.GetChapterId()}
+	}
+
+	if chapterReader.GetStatus() != commonpb.ChapterStatus_CHAPTER_CLOSED {
+		return nil, &domain.ErrChapterNotClosed{ChapterID: order.GetChapterId()}
+	}
+
+	if order.GetChapterId() != archivedThrough+1 {
 		return nil, &domain.ErrChapterArchiveOutOfOrder{
 			ChapterID:         order.GetChapterId(),
 			BlockingChapterID: archivedThrough + 1,
@@ -149,6 +164,30 @@ func processArchiveChapter(order *raftcmdpb.ArchiveChapterOrder, ctx *Context) (
 // from the ConfirmedArchiveChapterLog by deriveSignals.
 func processConfirmArchiveChapter(order *raftcmdpb.ConfirmArchiveChapterOrder, ctx *Context) (*commonpb.LogPayload, domain.Describable) {
 	s := ctx.Scope
+
+	// The confirm is what extends the archived prefix and triggers the purge, so
+	// it must land exactly on prefix + 1 — checked before any mutation, and
+	// before the residency lookup for the same reason as processArchiveChapter
+	// (an already-archived chapter reads as absent on a running node and as
+	// ARCHIVED after a restart, which would audit two different reasons for one
+	// order).
+	//
+	// A confirm past the prefix is not merely redundant: the recovery derivation
+	// stops at a gap while DispatchArchiveRequests republishes every ARCHIVING
+	// chapter, so a store holding {1 ARCHIVED, 2 un-archived, 3 ARCHIVING} would
+	// otherwise have reconciliation confirm 3 and carry the prefix past 2 —
+	// re-authorising archives while chapter 2's logs sit un-archived below the
+	// resulting archive boundary. That state is impossible through the ordering
+	// gate, so refusing it surfaces the anomaly (audited failure, and the
+	// archiver retries visibly) instead of silently papering over it.
+	archivedThrough := s.GetArchivedThroughChapterID()
+	if order.GetChapterId() != archivedThrough+1 {
+		return nil, &domain.ErrChapterArchiveOutOfOrder{
+			ChapterID:         order.GetChapterId(),
+			BlockingChapterID: archivedThrough + 1,
+		}
+	}
+
 	chapterReader, ok := s.GetChapterByID(order.GetChapterId())
 	if !ok {
 		return nil, &domain.ErrChapterNotFound{ChapterID: order.GetChapterId()}
@@ -179,11 +218,10 @@ func processConfirmArchiveChapter(order *raftcmdpb.ConfirmArchiveChapterOrder, c
 	chapter.Status = commonpb.ChapterStatus_CHAPTER_ARCHIVED
 	s.UpdateChapter(chapter)
 
-	// Extend the archived prefix. The ordering rule (processArchiveChapter) only
-	// admits the chapter right after the prefix, so a confirm always lands on
-	// prefix + 1 and the prefix stays contiguous. Advancing here rather than in
-	// the purge means a later order in the same proposal sees it.
-	s.SetArchivedThroughChapterID(chapter.GetId())
+	// Advancing here rather than at purge time means a later order in the same
+	// proposal sees it. The Scope exposes only a one-step advance, so no caller
+	// can carry the prefix over a gap.
+	s.AdvanceArchivedThroughChapterID()
 
 	return &commonpb.LogPayload{
 		Type: &commonpb.LogPayload_ConfirmArchiveChapter{
--- a/internal/domain/processing/processor_chapter.go
+++ b/internal/domain/processing/processor_chapter.go
@@ -110,6 +110,25 @@ func processArchiveChapter(order *raftcmdpb.ArchiveChapterOrder, ctx *Context) (
 		return nil, &domain.ErrChapterNotClosed{ChapterID: order.GetChapterId()}
 	}
 
+	// Archived chapters must form a contiguous prefix of history, so the only
+	// archivable chapter is the one right after the archived prefix. The purge
+	// deletes everything up to the archived chapter's close sequence, so
+	// archiving out of order would leave an older, un-archived chapter's logs
+	// retained *below* the archive boundary — where the checker's replay skips
+	// them, leaving a permanently unverified window, and where cold storage has
+	// a hole no later archive ever fills.
+	//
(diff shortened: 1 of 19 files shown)


internal/domain/processing/processor_chapter.go
```diff
@@ -150,3 +163,24 @@ func processConfirmArchiveChapter(order *raftcmdpb.ConfirmArchiveChapterOrder, c
...
+// tracker. A predecessor absent from the tracker blocks too: it cannot be shown
+// to be archived, and rejecting the order is the conservative, deterministic
+// outcome (rows are never deleted, so absence is already anomalous).
+//
+// The result is the minimum over a fixed id range, so it does not depend on map
+// iteration order — the rejection, which the audit chain records, stays
+// identical on every node.
+func oldestUnarchivedBefore(s Scope, target uint64) (uint64, bool) {
+	for id := uint64(1); id < target; id++ {
+		predecessor, ok := s.GetChapterByID(id)
```
NumaryBot: 🔴 **[blocker] Treat purged predecessors as archived**

After confirming chapter 1, `WriteSet.Merge` removes it from the in-memory tracker, so this `!ok` branch permanently rejects archiving chapter 2 until restart. Recovery reloads the retained archived row, meaning restarted and uninterrupted nodes can then produce different FSM results for the same order, violating [AGENTS.md:18-19](AGENTS.md#L18-L19). Track the archived prefix independently or distinguish a purged predecessor from a genuinely missing chapter.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1683#issuecomment-5239343113

internal/domain/processing/processor_chapter.go
Azorlogh: Confirmed and fixed in 178b6a81a — the purge does remove chapters from the tracker, so absence now reads as archived; treating it as blocking rejected every sequential archive and made the verdict restart-dependent. The two-chapter archival e2e catches it (verified failing on the previous behaviour); the mocked processor tests could not, since they encoded the same assumption.

internal/domain/processing/processor_chapter.go
```diff
@@ -150,3 +163,37 @@ func processConfirmArchiveChapter(order *raftcmdpb.ConfirmArchiveChapterOrder, c
...
+// accept an order a non-restarted one rejects, diverging the FSM. Treating
+// absence as blocking would also reject every sequential archive past the
+// first, since by then the predecessor is purged.
+//
+// The flip side is that a chapter row missing for any other reason reads as
+// archived here. Nothing else deletes it, and detecting a corrupt tracker is
+// not this gate's job — it exists to order archival, not to audit state.
+func oldestUnarchivedBefore(s Scope, target uint64) (uint64, bool) {
+	for id := uint64(1); id < target; id++ {
+		predecessor, ok := s.GetChapterByID(id)
```
NumaryBot: 🔴 **[blocker] Distinguish purged chapters from missing predecessors**

When an unarchived predecessor is unexpectedly absent from the tracker, this branch treats it as archived and permits a later chapter's purge, recreating the history hole this change is intended to prevent. Purged archived chapters are also absent, so the gate needs an independent archived-prefix marker or another way to distinguish those cases rather than silently accepting both; the current skip also conflicts with [AGENTS.md:26](AGENTS.md#L26), which requires impossible FSM states to surface loudly.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1683#issuecomment-5239343113

internal/domain/processing/processor_chapter.go
Azorlogh: Fixed in c919c0bec (branch squashed and retitled — the `!` was wrong: nothing is removed, no wire/storage format changes, and v3 is unreleased). The gate no longer interprets tracker residency at all: `ChapterTracker.archivedThroughID` records the archived prefix, advanced by ConfirmArchiveChapter (via Scope, so a batched confirm-N + archive-N+1 still works), propagated in Merge beside NextChapterID, and derived at recovery from the persisted rows as the longest run of ARCHIVED from chapter 1 — stopping at a gap or a missing row rather than counting past it. The gate is now `target == prefix + 1`, so purged and anomalously-missing chapters are no longer conflated. New tracker tests cover the derivation (including gap and missing-row cases), which the mocked processor tests could not reach.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1683#issuecomment-5239343113

internal/domain/processing/processor_chapter.go
```diff
@@ -108,6 +108,25 @@ func processArchiveChapter(order *raftcmdpb.ArchiveChapterOrder, ctx *Context) (
...
+	// archiving out of order would leave an older, un-archived chapter's logs
+	// retained *below* the archive boundary — where the checker's replay skips
+	// them, leaving a permanently unverified window, and where cold storage has
+	// a hole no later archive ever fills.
+	//
+	// The prefix is read from the tracker's explicit marker rather than inferred
+	// from which chapters are still resident: the purge drops archived chapters
+	// from the tracker, so residency answers "does this chapter still have hot
+	// data", not "is it archived".
+	if archivedThrough := s.GetArchivedThroughChapterID(); order.GetChapterId() != archivedThrough+1 {
```
gfyrag: 🔴 **[blocker] Traiter les chapitres déjà archivés avant le lookup de résidence**

Le lookup/statut exécuté plus haut reste dépendant de la résidence mémoire. Après purge, un nœud resté en ligne ne trouve plus le chapitre et renvoie `CHAPTER_NOT_FOUND`; après redémarrage, `RecoverState` recharge sa ligne `ARCHIVED` et le même ordre renvoie `CHAPTER_NOT_CLOSED`. Comme cet échec est audité, un même ordre Raft peut produire deux raisons — et deux hashes — différents. J’ai reproduit le différentiel avec les deux représentations.

Lis le préfixe avant le lookup et traite `chapterID <= archivedThroughID` de façon identique sur tous les nœuds, puis garde le lookup pour les chapitres au-delà du préfixe. Un test restart/no-restart devrait verrouiller ce cas.

internal/domain/processing/processor_chapter.go
```diff
@@ -142,6 +161,12 @@ func processConfirmArchiveChapter(order *raftcmdpb.ConfirmArchiveChapterOrder, c
 	chapter.Status = commonpb.ChapterStatus_CHAPTER_ARCHIVED
 	s.UpdateChapter(chapter)
 
+	// Extend the archived prefix. The ordering rule (processArchiveChapter) only
+	// admits the chapter right after the prefix, so a confirm always lands on
+	// prefix + 1 and the prefix stays contiguous. Advancing here rather than in
+	// the purge means a later order in the same proposal sees it.
+	s.SetArchivedThroughChapterID(chapter.GetId())
```
gfyrag: 🔴 **[blocker] Vérifier la continuité aussi lors du Confirm**

Ici le marqueur saute directement à `chapter.GetId()` sans vérifier `id == archivedThroughID + 1`. La dérivation au recovery s’arrête volontairement sur un trou, mais `DispatchArchiveRequests` republie tout chapitre `ARCHIVING`; avec `{1 ARCHIVED, 2 absent/CLOSED, 3 ARCHIVING}`, la réconciliation confirme 3 et positionne le préfixe à 3, contournant le garde-fou et réautorisant les archives suivantes malgré le trou.

Le confirm doit vérifier la continuité avant toute mutation et faire échouer bruyamment cet état impossible (invariant #7) ; idéalement le setter devrait aussi refuser tout saut. Ajouter un test de confirm au-delà d’un trou.

gfyrag (changes requested):
Deux violations de l’invariant de préfixe archivé restent possibles : une divergence restart/no-restart sur les re-archives, et un `ConfirmArchiveChapter` capable de sauter un trou. Les deux sont détaillées inline. Je demande des changements avant merge.

internal/domain/processing/processor_chapter.go
Azorlogh: Corrigé dans fe76c1eba. Le préfixe est maintenant lu **avant** le lookup de résidence, et les ids à l'intérieur du préfixe y sont rejetés — même raison auditée sur un nœud resté en ligne (chapitre purgé, absent) et sur un nœud redémarré (ligne rechargée en ARCHIVED). Au-delà du préfixe le lookup est indépendant de la représentation : un chapitre non archivé n'est jamais purgé, donc résident partout, et un chapitre inexistant n'a de ligne nulle part — `CHAPTER_NOT_FOUND` / `CHAPTER_NOT_CLOSED` restent donc atteignables et déterministes.

Même correctif appliqué à `ConfirmArchiveChapter`, qui avait le même trou (absent → NotFound vs ARCHIVED → NotArchiving).

Le test que tu demandais est là : `TestProcessChapter_ArchivedChapterRejectedIdenticallyAcrossRepresentations` rejoue les deux ordres sous les deux représentations et compare les *raisons auditées*, pas seulement le type d'erreur. Vérifié en échec sur l'ancien ordre des vérifications.

internal/domain/processing/processor_chapter.go
Azorlogh: Corrigé dans fe76c1eba. Le confirm vérifie maintenant `id == archivedThroughID + 1` **avant toute mutation**, donc ton scénario `{1 ARCHIVED, 2 non archivé, 3 ARCHIVING}` fait échouer le confirm de 3 au lieu de faire sauter le préfixe par-dessus le trou.

Et j'ai suivi ta suggestion sur le setter, en rendant le saut non représentable plutôt qu'en le vérifiant : le `Scope` n'expose plus que `AdvanceArchivedThroughChapterID()`, sans argument — un seul pas, pas d'id — donc aucun appelant ne peut porter le préfixe au-delà d'un trou, même par erreur.

Échec choisi : rejet métier (audité, déterministe, le chapitre reste ARCHIVING et l'archiver retente de façon visible) plutôt qu'un `assert.Unreachable` + panic, qui tuerait toutes les répliques sur une anomalie de données. L'état cesse d'être silencieusement contourné, ce qui est le point de l'invariant #7.

Test ajouté : `TestProcessConfirmArchiveChapter_RefusesToJumpAGap` — exactement ton état, en vérifiant qu'aucune mutation n'a lieu et que le préfixe ne bouge pas. Vérifié en échec sans le garde-fou.

internal/domain/processing/processor_chapter.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/6d9e3d94fbd0db27d634acbe44b73a08a5a9aff9 — the archived-prefix check now runs before the residency lookup, so the reason no longer depends on whether the chapter is still resident.

internal/domain/processing/processor_chapter.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/6d9e3d94fbd0db27d634acbe44b73a08a5a9aff9 — the confirm now requires `id == archivedThrough + 1` before any mutation, and the Scope only exposes a one-step advance.

gfyrag (commented):
Independent review at head 88269aa894f77dec5bffa23577a2c368809cb9d6. One blocking finding and one documentation concern remain.

[P2][blocking] Return an actionable result for an already archived chapter
Location: internal/domain/processing/processor_chapter.go, internal/domain/errors.go, chapter lifecycle/CLI docs
Evidence: with archivedThroughID=3, retrying ArchiveChapter(1) returns blockingChapterId=4 and says chapter 4 must be archived first. Chapter 4 is newer, and archiving it can never make chapter 1 valid again; each retry only advances the claimed blocker.
Impact: the public remediation metadata is impossible to follow and automated clients can loop indefinitely.
Resolution: represent the already-archived case separately, or redefine and consistently document the field as nextArchivableChapter rather than an older blocker.

[P3][non-blocking] Add CHAPTER_ARCHIVE_OUT_OF_ORDER and its metadata to the canonical public-error table in docs/technical/contributing/api-comparison.md.

Targeted domain/processing/state/check tests and CI passed.

Azorlogh:
@gfyrag both addressed at d3b652a9e.

**[P2]** New reason `CHAPTER_ALREADY_ARCHIVED` (enum 72) for a chapter inside the prefix, carrying `archivedThroughChapterId` instead of a blocker — the order is already carried out, and nothing archivable can make an older chapter archivable again. `blockingChapterId` keeps its meaning in the only case where it is one (a chapter beyond the prefix).

Two notes on scope:
- Classified `KindConflict` rather than `KindPrecondition`, following `TRANSACTION_ALREADY_REVERTED` — same shape, an operation whose effect is already in place. Wire result is unchanged (`FAILED_PRECONDITION`).
- `processConfirmArchiveChapter` had the same defect through its single `!= archivedThrough+1` check, so a confirm inside the prefix was also handed a newer blocker. Split there too — one situation, one reason across both orders that can reach it.

**[P3]** Added `CHAPTER_ARCHIVE_OUT_OF_ORDER`, `CHAPTER_ALREADY_ARCHIVED` and `CHAPTER_ARCHIVE_IDENTITY_MISMATCH` to the table in `api-comparison.md` — the identity-mismatch row was missing since #1679. `lifecycle.md` and `cli.md` documented only the blocker case and now cover both.

A test case of mine had pinned the old behaviour (`"already inside the prefix"` expecting `wantBlockingID: 4`); it is now its own test asserting the new reason and `NotErrorAs` on the old one.

gfyrag (approved):
Re-reviewed at d3b652a9e9d77fd9bd0dc59e473365d2d30c540d. The archive-prefix ordering, recovery behavior, same-proposal confirm/archive sequencing, and stale sealing-hash protection are consistent. All current CI checks pass.

gfyrag (approved):

PR #1689, opened 2026-08-12
fix(EN-1773): reconstruct last_mirror_v2_log_id on restore, reconcile mirror workers on deletion

paul-nicolas:
Jira: https://formance-team.atlassian.net/browse/EN-1773
Defect filed and fixed here: https://formance-team.atlassian.net/browse/EN-1776

EN-1773 was chartered as test debt, with the instruction *"no production-code change is expected; if either test exposes a real defect, file it separately"*. Grounding its two coverage gaps in the code surfaced **two defects**. Both are fixed here, and EN-1776 carries the record for the critical one.

## EN-1776 — `RebuildDelta` did not reconstruct `last_mirror_v2_log_id`

`LedgerBoundaries.last_mirror_v2_log_id` is the per-ledger high-water mark of applied source v2 log ids. It has exactly one writer, the FSM apply path (`processor_mirror.go:152`). The restore rebuild seeded the boundary row from the checkpoint and wrote it back otherwise untouched, while the delta's mirror ingests *were* replayed — so a restored store held the data for source logs it claimed never to have applied.

Both the worker (`worker.go:274`) and the FSM (`processor_mirror.go:51`) read that one row, so after a restore they agreed on a stale position. The re-fetched logs arrived as `v2LogID == last+1` and took the **apply** branch rather than the idempotent `v2LogID <= last` skip.

This inverts the assumption recorded in EN-1773: a rolled-back boundary is harmless when only the worker's in-memory position is ahead of the durable row. Here the rollback hits the row *both* sides read.

**The damage is invisible in transaction counts.** Mirrored transactions reuse the source transaction id, so a re-apply overwrites the same rows; `applyPosting` runs again with `force=true`, so account volumes silently double. The volume comparison, not the row count, is the assertion that cannot be masked.

### Fix

| File | Change |
|---|---|
| `internal/domain/replay/order_effects.go` | Carry `MirrorV2LogID`; **widen the guard** — it discarded every non-fill-gap mirror ingest, dropping the ledger name with it, and `CreatedTransaction` is the common kind |
| `internal/infra/backup/rebuild.go` | Fold it as a **max** in `applyAuditOrderEffects`, matching `recordMirrorIngestMutations` (`checker.go:3533-3538`) so the rebuild and the checker's oracle agree by construction |

## Second defect — `DeleteLedger` raised no reconcile signal

`Manager.reconcile` runs on a leadership change or a `ConfigChanged` notification, and that notification fires only for `sinkConfigChanged || mirrorConfigChanged` (`machine.go:903-904`). `DeleteLedger` set neither, so a deleted mirror ledger's worker kept polling its v2 source and proposing ingests the FSM rejects, until an unrelated reconcile happened to run.

`reconcile` itself was already correct for both promotion and deletion — `ReadMirrorLedgers` filters `mode == MIRROR && DeletedAt == nil`, so both converge on the same removal branch. The bug was purely in the trigger.

Set unconditionally rather than only for mirrors: the mode is not reachable in `Absorb` without an ungated overlay read, which is exactly what that case was deliberately moved away from (invariant #9). A non-mirror deletion costs one idempotent reconcile.

## Coverage

Every regression below was **demonstrated to fail without its fix**, not merely to pass with it.

| Test | Level |
|---|---|
| `DecodeOrderEffects` table tests | `internal/domain/replay` had **no test file at all** |
| `RebuildDelta` — fill-gap, `CreatedTransaction`, max fold | unit |
| `Manager.reconcile` — promotion, deletion, leadership loss | `manager.go` had **no test file at all** |
| `Worker.publishIdleStatus` — idle publication + dedup, error recovery at an unchanged head, empty-source recovery, all four propose-failure stages | unit |
| `restore_mirror_test.go` — cursor survives backup/restore, resume without gap, single ingestion via volumes, tx ids aligned | e2e `s3` |
| `mirror_test.go` — worker stops polling once its ledger is deleted | e2e |

## A premise in the ticket that turned out to be wrong

EN-1773 asked for a *"recreate under the same name restarts from source log 1"* case. That lifecycle is unreachable: the retained tombstone makes `CreateLedger` fail with `ErrLedgerDeleted` — CLAUDE.md notes this as *"the retained tombstone that makes that lifecycle unreachable"*. The spec now pins that guard instead. The nil-boundary resume path the ticket targeted is already covered by `TestWorker_FreshLedgerFetchesFromSourceIDOne`, where it *is* reachable.

## Verification

- `bash scripts/agent-check` — PASS, 0 lint issues on both modules
- `go test -race` on mirror / replay / backup / state / check / query / adapter-v2 — all pass
- `go test -tags e2e` focused on mirror — 21/21 specs
- `go test -tags "e2e,s3" ./tests/e2e/cluster/` focused on the restore suite — 8/8 specs
- Falsifications, each confirmed to fail before its fix:
  - the two error-clearing unit cases, against the previous head-only guard;
  - `TestRebuildDelta_FoldsHighestMirrorV2LogID`, with the max fold mutated to a plain assignment (proving it still guards the fold after `IsNumscript` removal);
  - the new `SourceLogCount` e2e assertions, against the previous ascending head probe (head read 1 instead of 4);
  - with the `rebuild.go` fold disabled, the e2e fails at the high-water-mark assertion and both `RebuildDelta` regressions fail.

## Follow-ups worth considering (not in scope here)

`processMirrorIngest` has **no detector for a rewound high-water mark**. `<= last` is a soft no-op and `> last+1` is a loud gap error, but a mark rewound to 0 makes every replayed log look like the legitimate next one. That asymmetry is why EN-1776 was silent; this PR removes the instance without removing the blind spot. A monotonicity check (checker, or the worker's boundary read) may be warranted.

Separately, surfaced while fixing the e2e fixture: `HTTPSource.FetchLogs` and `HTTPSource.GetLatestLogID` make **opposite ordering assumptions about the same request shape**. `GetLatestLogID` sends `pageSize=1` with no `after` and reads `Data[0]` as the newest log; `FetchLogs` also omits `after` when resuming from 0 and needs the page ascending (`TranslateBatch` enforces source-log-id contiguity). Both cannot be right against the real v2 API. `GetLatestLogID`'s intent is settled — `PostgresSource` implements the same interface method as `SELECT MAX(id)` — so the fixture was fixed to match that contract, but which side is wrong against a live v2 ledger is a pre-existing production question deserving its own ticket.

## Third defect — an idle mirror could not publish its own status

*(This section replaces what earlier revisions of this description listed as an out-of-scope follow-up. Review made it in scope, and the head now fixes it.)*

`SubPLMirrorSourceHead` is not reconstructed by the rebuild, and it is written only as a side effect of a mirror ingest proposal. A correctly restored mirror with nothing left to ingest therefore reported `SYNCING` with zero remaining logs, and it took a new source log to clear — so an idle source left a correct restore reporting `SYNCING` indefinitely, and an operator treating `FOLLOWING` as the restore-verification signal waited forever.

`Worker.publishIdleStatus` now proposes a standalone `MirrorSyncUpdate` when the worker is caught up. Review then found that the first version of that guard, keyed on the source head alone, introduced a defect of its own: the same proposal carries `ClearError`, so a mirror that recorded an error and recovered **without producing a new log** never issued the clear — the head had not moved, so the guard suppressed it and the API kept serving a stale error. An empty source could never clear at all, not even across a worker restart, because a head of `0` was indistinguishable from "never observed".

The two conditions are now tracked separately:

| Bit | Meaning |
|---|---|
| `sourceHeadObserved` | `GetLatestLogID` has answered, so an observed empty source is distinguishable from "never asked" |
| `statusClearConfirmed` | a `ClearError` update was confirmed applied since the last error report; its zero value is deliberately `false`, so a freshly started worker publishes once rather than assuming the status is clean |

`publishIdleStatus` and `reportError` were also collapsed onto a shared `proposeMirrorSync(ctx, update, label) bool` — they were the same ~55 lines of Build / Run / Raft-wait / FSM-wait / business-error ladder twice, differing only in payload and log strings.

### Fourth: `OrderEffects.IsNumscript` was dead

`310ba1f4` added it with two consumers; `5a2f8892` (EN-1420 counter cutover) deleted both when the usage counters moved to usagestore and left the field behind. It was not inert: its guard term was the sole reason a script-sourced `CreateTransaction` reached the mirror fold with `MirrorV2LogID == 0`, which is what the 15-line comment in `applyAuditOrderEffects` and `TestRebuildDelta_NumscriptOrderDoesNotClobberMirrorMark` existed to justify. Removing the field collapses both; the max fold stays (it matches the checker's own fold, and `compareMirrorV2LogID` is strict equality against it) and remains pinned by `TestRebuildDelta_FoldsHighestMirrorV2LogID`.

## Automated coverage boundary for EN-1776

`restore_mirror_test.go` is `//go:build e2e && s3`, and **no CI job runs the `s3`-tagged suites**. `Tests-E2E` runs `just test-e2e-coverage`, which is `-tags e2e` only; `s3` arrives solely through `just test-e2e-full`, which no workflow invokes. That suite therefore runs only locally, with Docker.

That gap was raised in review against the new `publishIdleStatus` path specifically, and is now closed at the unit level (see the coverage table above) rather than resting on the s3 suite. For the EN-1776 regression the committed automated coverage is likewise the unit layer:

- `TestRebuildDelta_AdvancesLastMirrorV2LogIDForCreatedTransaction`
- `TestRebuildDelta_AdvancesBoundariesForMirrorFillGap`
- `TestRebuildDelta_FoldsHighestMirrorV2LogID`
- `TestRebuildDelta_NumscriptOrderDoesNotClobberMirrorMark`
- the `DecodeOrderEffects` table cases in `internal/domain/replay/order_effects_test.go`

This gap is pre-existing for all six `restore*_test.go` suites, not introduced here. Adding an s3-tagged e2e CI lane is worth a follow-up ticket of its own rather than a rider on this PR.


paul-nicolas (commented):

**DECISION: REQUEST CHANGES** · Head reviewed: `f7439a5d0` · Blocking: 1 · Non-blocking: 12 · First review (no previous findings)

The core fix is correct. Both production changes do what they claim, and the unit tests genuinely fail without them (verified by mutation runs via `go test -overlay`). The widened `DecodeOrderEffects` guard is inert for its only other caller (`checker.go:6129` skips on empty `Ledger` and folds only `SkippedTransactionIDs`), the zero-id short-circuit matches `processMirrorIngest`'s explicit rejection, and the unconditional `mirrorConfigChanged` introduces no FSM nondeterminism (invariant #2) and no invariant #9 regression. Every finding below concerns test robustness or comment accuracy, not shipped behaviour.

**Posted inline:** 6 findings (1 P2 blocking, 5 P2 non-blocking). **Withheld:** 7 non-blocking P3, listed below.

### Verification
13 findings after dedup. Three entered a formal adversarial gate (verifiers instructed to refute, defaulting to invalid): **2 confirmed valid**, **1 adjusted**, **0 rejected**. The remaining three P2s were confirmed directly against source.

- The docs "single-writer" finding was **downgraded P2-blocking to P3**: `applyAuditOrderEffects` already wrote `NextTransactionId` before this PR, so `RebuildDelta` was already a `LedgerBoundaries` writer and the imprecision pre-dates this change. Two of its three citations were dropped (`audit-vs-technical-state.md` and `docs/ops/backup-restore.md` make no exclusivity claim).
- **Conflict resolved:** Codex rated the e2e teardown assertion *blocking*, arguing the surviving worker escalates into exponential backoff (`maxBackoff = 60s`) so the assertion could never fail. Refuted at `worker.go:268` — `w.boundariesLoaded` stays `true`, the deleted boundary is never re-read, `FetchLogs` returns empty, and `processBatch` returns `(false, nil)` with no error. No backoff; cadence stays a clean 5s. Downgraded to non-blocking.

### Verified negatives
No checker blast radius from the widened guard; no symmetric sink bug (sink configs are global, not ledger-scoped, so `sinkConfigChanged` correctly stays false); `Manager.reconcile` is synchronous under lock, so the new unit tests are not racy and need no `Eventually`; no port, bucket or clusterID collisions (15250/15150/14150 appear nowhere else; multi-node bases top out at 15003/15103/15203/15303); the `expectSingleIngestion` volume assertion does pin the regression; Ginkgo phase ordering, `DeferCleanup` lifetime for the mock v2 server, and shared-var sequencing are all correct.

### P3 (not posted inline)
- `docs/technical/architecture/subsystems/events-mirror/mirror.md:177` — scope "only ever written by the FSM" to the live path; the restore rebuild is now a second, offline writer.
- `internal/infra/backup/rebuild.go:791-793` + `internal/application/check/checker.go:6098-6099` — the "same fold `backup.RebuildDelta` performs on restore" equivalence is now false; the checker derives the mirror mark separately via `recordMirrorIngestMutations`.
- `internal/infra/state/write_set_absorb_test.go:105` — the "cleanup signal only" name and its "records only the cleanup signal now (EN-1522)" comment are now wrong; the new subtest at :173 also duplicates the `deletedLedgers` assertion already covered at :115.
- `internal/infra/state/write_set.go:998-1007` — "one read handle" undercounts: `NotifyConfigChanged` fans out to both the mirror and the events manager, so each deletion costs a read handle in each.
- `internal/infra/backup/rebuild.go:861-874` — `SubPLMirrorSourceHead` is not reconstructed; "cosmetic and self-healing" holds only once a new source log arrives, so an idle source leaves a correctly restored mirror reporting `SYNCING` indefinitely. An operator using `FOLLOWING` as a restore-verification signal waits forever.
- `tests/e2e/cluster/mirror_test.go:658-663` — assert `ErrorInfo.reason` (the wire contract) rather than the error message substring; the file's own convention is 20 lines below at :685-687.
- `tests/e2e/cluster/restore_mirror_test.go:51-58` — the "so this suite can run alongside them" rationale invents a harness capability (`go test -p 1`, no `ginkgo -p` anywhere), and this is the only restore suite deviating from the shared `testutil.TestSingle*Port` convention.

_Reviewed with Claude (3 parallel area reviewers + 3 adversarial verifiers) and Codex, from a dedicated worktree at the PR head._

internal/application/mirror/manager_test.go
```diff
@@ -0,0 +1,181 @@
...
+	session := store.OpenWriteSession()
+	require.NoError(t, state.SaveLedger(session, info.GetName(), info))
+	require.NoError(t, session.Commit())
+}
+
+func newTestManager(t *testing.T, store *dal.Store, builder *plan.Builder) *Manager {
+	t.Helper()
+
+	m := NewManager(
+		store, nil, builder,
```
paul-nicolas: [P2][blocking] Give the test Manager a stub Proposer instead of nil

`newTestManager` passes `nil` as the `Proposer`. `manager.go:183` hands it straight into `NewWorker`, and `reconcile` calls `w.Start()`, so a real worker loop runs. Any `processBatch` error reaches `reportError` (`worker.go:231`) with **no** `ctx.Err()` re-check, and `plan/runner.go:215` calls `proposer.Propose` with **no** nil guard. `Builder.Build` does not error first: a missing `SubAttrLedger` row is an ordinary cache miss, not an error.

The trigger is teardown itself. `reconcile`'s removal branch (`manager.go:145`) and `teardown` (`manager.go:192`) call `worker.Worker.Stop()`, which closes `stopCh` and cancels the loop ctx; cancellation landing inside `source.FetchLogs` makes `processBatch` return an error, and `reportError` then runs with the dead ctx.

**Reproduced**, not theorised:

```
go test -race -count=200 -run 'TestManager_' ./internal/application/mirror/
-> SIGSEGV at internal/infra/plan/runner.go:215, nil proposer {0x0, 0x0}
   mirror.(*Worker).reportError (worker.go:552) <- processLogs (worker.go:231) <- loop (worker.go:176)
```

Failed 2/2 at `-count=200`; ~0.5-1% per iteration (100 iterations in batches of 20 all passed). 0 panics without `-race` -- the detector's slowdown widens the window, and CI runs `-race` (`justfile:81`).

**Impact:** a panic from a background goroutine is unrecoverable. It aborts the whole `internal/application/mirror` test binary, taking every other parallel test with it, with a stack pointing at `plan/runner` rather than at the test.

**Not pre-existing:** `helpers_test.go:56 newWorkerForTest` also passes `nil`, but those workers never `Start()` -- `worker_test.go` calls `w.processBatch(...)` directly, so `reportError` is never entered. This is the first test to run a real worker loop against a nil proposer.

**Resolution:** pass a stub, e.g. a package-local `errProposer` whose `Propose` returns `nil, errors.New("no proposer in tests")`; `reportError` then just logs and returns. An early `if ctx.Err() != nil { return }` at the top of `reportError` would also be defensible on production merit, but the test is what this PR introduces.

internal/application/mirror/manager_test.go
```diff
@@ -0,0 +1,181 @@
...
+	require.Empty(t, workerNames(m), "a promoted ledger is no longer a mirror; its worker must be stopped and dropped")
+}
+
+// Deletion soft-deletes the ledger, which also drops it from ReadMirrorLedgers
+// (it filters DeletedAt == nil), so the same removal branch must fire.
+//
+// Note this test drives reconcile directly. On a live cluster the trigger is
+// missing: DeleteLedger sets no mirrorConfigChanged flag, so no ConfigChanged
+// notification fires and the worker survives until an unrelated reconcile.
+// That gap is covered end to end in the cluster suite.
```
paul-nicolas: [P2][non-blocking] Delete the comment claiming the DeleteLedger trigger is still missing

This says "On a live cluster the trigger is missing: DeleteLedger sets no mirrorConfigChanged flag, so no ConfigChanged notification fires and the worker survives until an unrelated reconcile." The same commit closes exactly that gap -- `write_set.go` sets `b.mirrorConfigChanged = true` in the `LogPayload_DeleteLedger` case, pinned by `write_set_absorb_test.go:173`. The cluster spec it defers to asserts the request count *stops rising*, i.e. it proves the trigger fires, not that it is absent.

**Impact:** the one file a maintainer opens to understand mirror-worker teardown states the opposite of the code in the same commit. A future reader either re-fixes `write_set.go` or removes the flag believing it was never wired up.

**Resolution:** reword to describe what this test actually covers -- reconcile's removal branch for a soft-deleted ledger, driven directly -- and note that the live trigger is `WriteSet.Absorb`'s `DeleteLedger` case, covered by `write_set_absorb_test.go` and the cluster suite.

Kept non-blocking because a test comment is explanatory rather than authoritative, and `mirror.md:168` already documents the intended behaviour correctly. Worth fixing in the same pass as the nil-proposer finding above -- same file.

tests/e2e/cluster/mirror_test.go
```diff
@@ -556,6 +569,101 @@ var _ = Describe("Mirror", Ordered, func() {
...
+			Eventually(func(g Gomega) {
+				current := mockV2.requestCount()
+				defer func() { previous = current }()
+				g.Expect(current).To(Equal(previous))
+			}, 30*time.Second, 2*time.Second).Should(Succeed(),
+				"mirror worker is still polling the source after its ledger was deleted")
+
+			// And it stays stopped rather than merely pausing between polls.
+			Consistently(mockV2.requestCount, 5*time.Second, 500*time.Millisecond).
+				Should(Equal(previous))
```
paul-nicolas, line 680: [P2][non-blocking] Probe the teardown counter at an interval larger than the 5s poll interval

The comment claims "two consecutive probes a poll interval apart", but the probe interval here is 2s and the real interval is 5s (`internal/application/mirror/worker.go:29`, `defaultPollInterval`; ticker at `:172`). It is a package-private const with no flag, env var or test override, so every e2e suite runs the worker at exactly 5s. Two probes 2s apart observe an unchanged count on a live worker much of the time, and `Eventually` retries for 30s -- so the `Eventually` clause succeeds on a live worker with near-certainty and proves nothing on its own.

The spec is not toothless *today*: its success condition forces the previous tick to have happened at or before the window start, so the next tick lands within 3s of the `Consistently` start and is observed by a poll at <=4.5s. I checked the unfixed cadence to confirm this holds -- the surviving worker does **not** enter backoff, because `w.boundariesLoaded` stays true (`worker.go:268`), the deleted boundary is never re-read, `FetchLogs` returns empty and `processBatch` returns `(false, nil)`. Clean 5s.

But the margin is the inequality `pollInterval - probeInterval (3s) <= window - pollGranularity (4.5s)` -- about 1.5s. Raise `defaultPollInterval` above ~6.5s and this spec silently becomes a no-op that passes with the `write_set.go` fix reverted.

**Impact:** a latent, silent regression-test loss, and the comment currently documents a guarantee the code does not provide, so a future reader trusts the wrong clause.

**Resolution:** `ProbeEvery(6 * time.Second)`, or hold the `Consistently` for more than two poll intervals (as `restore_mirror_test.go:441` already does at 12s), and correct the comment to name the real 5s interval.

Note for triage: Codex rated this blocking on the theory that the worker escalates into exponential backoff (`maxBackoff = 60s`), which would make the assertion unable to fail at all. That path is not reached -- see the `boundariesLoaded` trace above.

tests/e2e/cluster/restore_mirror_test.go
```diff
@@ -0,0 +1,474 @@
...
+			defer cancel()
+			Expect(server.Stop(stopCtx)).To(Succeed())
+			_ = os.RemoveAll(restoreWalDir)
+			_ = os.RemoveAll(restoreDataDir)
+		})
+
+		It("rebuilds the mirror high-water mark from the replayed delta", func() {
+			// EN-1776: RebuildDelta folded every other boundary field but left
+			// last_mirror_v2_log_id at 0, so this read returned 0.
+			Expect(restoredCursor).To(Equal(preBackupCursor),
```
paul-nicolas, line 436: [P2][non-blocking] Sample the restored cursor so a re-ingest cannot satisfy the assertion

The `Eventually` at :381 has no assertion on the sampled value -- it succeeds as soon as `GetLedger` answers -- and the read is not synchronised against the mirror worker. The comment argues the read is taken before leadership, but nothing *proves* the node is still non-leader when `GetLedger` first answers.

On an unfixed build the rebuilt mark is 0, the worker restarts at source log 1 and re-applies logs 1..4. `processMirrorCreatedTransaction` reuses the **source** transaction id (`processor_mirror.go:213,228`) and `processMirrorIngest` advances `LastMirrorV2LogId` again, so once the re-ingest completes the cursor reads 4 -- exactly `preBackupCursor`. `ReadMirrorSyncProgress` is pure Pebble state, so whichever writer got there first is what the read returns.

**Impact:** "rebuilds the mirror high-water mark from the replayed delta" (:405) and "resumes ingestion without error" (:410) can both pass on a buggy build. The regression is actually caught only by "does not re-apply source logs it had already applied" (:433), via the doubled volumes. The two specs that name the fix are non-deterministic evidence; if the volume spec is ever weakened or the account/asset drift, the suite goes green on the bug.

There is no false-failure risk in the other direction -- the value is written by `RebuildDelta` in Phase 2 and read straight from Pebble, so if `GetLedger` answers, the boundary is there.

**Resolution:** make the sample provably pre-worker. The counter this PR adds is available in the same package: record `before := mockV2.requestCount()` before `server.Start`, and require `mockV2.requestCount() == before` in the same `Eventually` body that samples the cursor -- no source request implies no ingest, so the value came from `RebuildDelta`. Alternatively read the rebuilt boundary in Phase 2 after `FinalizeRestore`, where no mirror manager exists at all.

tests/e2e/cluster/restore_mirror_test.go
```diff
@@ -0,0 +1,474 @@
+//go:build e2e && s3
```
paul-nicolas, line 1: [P2][non-blocking] No CI job runs the `s3`-tagged cluster suites, so this restore regression is only pinned locally

`.github/workflows/main.yml:95` (job `Tests-E2E`) runs `just test-e2e-coverage`, which is `go test -race -tags e2e -p 1 ./tests/e2e/business/... ./tests/e2e/cluster/...` (`justfile:189`) -- `e2e` only. Only `just test-e2e-full` passes `-tags "e2e,{{all_tags}}"` with `s3` (`justfile:39,96`), and no workflow invokes it. `docs/technical/contributing/testing.md` documents `-tags e2e` only and never mentions the s3 e2e suites.

**Impact:** the EN-1776 double-ingestion regression -- the primary evidence for this PR's headline fix -- has no automated gate. It will only be caught by someone running `just test-e2e-full` with Docker locally.

This is pre-existing for all six `restore*_test.go` suites, so fixing it is arguably out of scope here. It is reported because this change's load-bearing regression proof lives in that lane, which is exactly the dependency the review contract asks to surface.

**Resolution:** either add an s3-tagged e2e CI job (MinIO comes from testcontainers, so there is no external dependency), or state explicitly in the PR description that this suite is local-only and that the committed automated coverage for EN-1776 is the `RebuildDelta` and `DecodeOrderEffects` unit tests.

internal/infra/backup/rebuild.go
```diff
@@ -857,6 +857,21 @@ func (w *attributeReplayWriter) applyAuditOrderEffects(reader dal.PebbleReader,
...
+		// live path, and the ledger-log stream does not carry the source id
+		// except for fill-gaps, so without this fold a restored mirror keeps
+		// the checkpoint's value while the delta's ingests are replayed. Worker
+		// and FSM both read that one row, so they agree on a stale position and
+		// the re-fetched logs arrive as v2LogID == last+1 — the APPLY branch,
+		// not the idempotent skip (EN-1776).
+		//
+		// Max, not assignment: audit items are folded in key order, which is
+		// audit sequence rather than v2 log id, and it matches the checker's
+		// own fold in recordMirrorIngestMutations.
```
paul-nicolas: [P2][non-blocking] Correct the max-fold rationale and pin the regression it actually prevents

The stated reason -- "audit items are folded in key order, which is audit sequence rather than v2 log id" -- describes a hazard that cannot occur. `processMirrorIngest` no-ops `v2LogID <= last` and rejects `v2LogID > last+1` (`processor_mirror.go:86-90`), so the *applied* ingests of one ledger have strictly increasing `v2_log_id`; key order is ascending by audit sequence (big-endian `[audit_seq BE8][order_index BE4]`, `dal/key_builder.go:59-64`). I closed the escape routes too: failed proposals carry `LogSequence == 0` and are skipped at `rebuild.go:826`; idempotency replays return before any audit write; delete-then-recreate under the same name is rejected with `ErrLedgerDeleted`; FillGap shares the same gate. The only constructible descending case requires an already-rewound mark, i.e. the EN-1776 bug itself.

So `TestRebuildDelta_FoldsHighestMirrorV2LogID` (`rebuild_test.go:385-413`) does pin the max branch -- verified, it fails 9 vs 4 under a plain assignment -- but with a physically impossible input.

The **reachable** reason the max is load-bearing is untested. `DecodeOrderEffects` returns a non-empty `Ledger` with `MirrorV2LogID == 0` for a script-sourced `CreateTransaction` (`order_effects.go:74-85`: `isNumscript` alone defeats the zero guard). `boundaryFor` (`rebuild.go:1023-1039`) seeds `b` from the existing checkpoint row, and promotion never zeroes the mark (`processPromoteLedger` does not touch boundaries). So a promoted ex-mirror ledger with `last_mirror_v2_log_id = 9` plus one numscript order in the delta would, under an assignment, have its mark rewound to 0 -- which `compareMirrorV2LogID` then flags as `CHECK_STORE_ERROR_TYPE_MIRROR_V2LOGID_MISMATCH`, since it is strict equality against the checker's own max fold. Verified with a probe test: passes on shipped code, fails 9 -> 0 under assignment. `rebuild_test.go` has no numscript case.

**Impact:** no production impact -- the shipped code is a max and is correct. But the comment's stated reason is the thing a future simplifier would check before reverting the max, and it points at a hazard that cannot happen while the real one goes uncovered.

**Resolution:** reword to cite the real reason (a same-ledger effect carrying no source id -- e.g. a numscript `CreateTransaction` on a promoted ex-mirror ledger -- must not clobber the checkpoint-seeded mark), and add that case as a test alongside the existing one.

internal/application/mirror/manager_test.go
paul-nicolas: Fixed in 3ee4afc09 — `newTestManager` now passes a package-local `errProposer` stub whose `Propose` returns an error, so `reportError` stays on its logging path.

Reproduced independently before fixing, in a worktree at the PR head: clean at `-race -count=200` (37s), then SIGSEGV at `-race -count=600` (94s) with exactly the stack you described — `plan.(*Builder).Run` at `runner.go:215` receiving the proposer as `{0x0, 0x0}`, via `reportError` (`worker.go:552`) ← `processLogs` (`worker.go:231`) ← `loop` (`worker.go:176`). Verified after the fix: `-count=600` clean.

Deliberately did not take the optional `if ctx.Err() != nil { return }` guard at the top of `reportError`. It is defensible on production merit, but it only narrows the window — the nil-proposer hazard would remain on any non-cancellation error path — and the stub is the complete fix for what this PR introduces.

internal/application/mirror/manager_test.go
paul-nicolas: Fixed in 3ee4afc09. Confirmed: the same change sets `b.mirrorConfigChanged = true` in the `DeleteLedger` case (`write_set.go`), asserted in `write_set_absorb_test.go`. The comment now describes what the unit test covers — reconcile's removal branch, driven directly — and points at `WriteSet.Absorb` as the live trigger.

tests/e2e/cluster/mirror_test.go
paul-nicolas: Fixed in ce914e74f. Confirmed the mismatch: probe interval 2s against `defaultPollInterval = 5s` (`worker.go:29`, ticker at `:172`), a package-private const with no flag, env var or test override.

`ProbeEvery` is now 6s, so two equal consecutive counts mean a tick was due and did not happen. `Consistently` now holds 12s — more than two poll intervals, matching `restore_mirror_test.go`. Comment corrected to name the real 5s interval and say why the probe must exceed it.

Also accepted your `boundariesLoaded` trace (`worker.go:268`) over the Codex backoff theory: `FetchLogs` returns empty and `processBatch` returns `(false, nil)`, so there is no error and no escalation. Non-blocking was the right call.

tests/e2e/cluster/restore_mirror_test.go
paul-nicolas: Fixed in ce914e74f, with one adjustment to the placement.

The finding is right: the `Eventually` only required `GetLedger` to answer, and on an unfixed build a completed re-ingest leaves the cursor at 4 — the same value a correct rebuild leaves — so the read could not distinguish them.

Adjustment: putting `requestCount() == before` **inside** the `Eventually` body inverts the flake. Once the restored node's worker makes its first fetch the guard can never match again, so on a slow CI a *correct* build would spin to the 30s timeout. Instead the counter is sampled alongside the cursor (read after `GetLedger` returns, so the witness is "no fetch completed before this read returned") and asserted in the "rebuilds the mirror high-water mark" spec, where a failure reads as "the sample was not provably pre-worker" rather than as a timeout.

Floor captured with `preRestoreRequests = mockV2.requestCount()` immediately before `server.Start` — the mock is shared across all three phases, so it cannot be assumed zero.

tests/e2e/cluster/restore_mirror_test.go
paul-nicolas: Confirmed, and half-addressed — deliberately not the CI half.

Verified every citation: `main.yml` job `Tests-E2E` runs `just test-e2e-coverage` = `-tags e2e` only; `s3` reaches the suite solely through `just test-e2e-full` (`-tags "e2e,{{all_tags}}"`, `all_tags` at `justfile:39`); grepping `.github/workflows/*.yml` for `test-e2e-full`, `test-all` and `s3` returns nothing. `testing.md` documents `-tags e2e` only.

Done in 3eb15f113 + the PR description: a new "Automated coverage boundary for EN-1776" section states the suite is local-only and names the committed automated coverage — the three `RebuildDelta` mirror-mark tests, the new numscript one, and the `DecodeOrderEffects` cases.

Not done: adding an s3-tagged e2e CI lane. It affects all six `restore*_test.go` suites, brings its own runtime and MinIO-container cost, and is pre-existing rather than introduced here — that belongs in its own change with its own review, not as a rider on this PR. Recommend a follow-up ticket covering the whole s3 e2e lane.

Side effect of reading this closely: the PR description's claim that the unreconstructed `SubPLMirrorSourceHead` is "cosmetic and self-healing" was wrong for the same reason your P3 note gives, so that paragraph is corrected too.

internal/infra/backup/rebuild.go
paul-nicolas: Fixed in ee0313dfe (test) and 3eb15f113 (comment).

Verified both halves of the finding. The stated hazard is indeed unreachable — `processMirrorIngest` no-ops `v2LogID <= last` and rejects `> last+1`, so applied ingests of one ledger have strictly increasing ids while audit key order ascends by audit sequence. And the reachable one checks out: `order_effects.go` short-circuits on `fillGap == nil && !isNumscript && mirrorV2LogID == 0`, so `isNumscript` alone yields a non-empty `Ledger` with `MirrorV2LogID == 0`, and `boundaryFor` seeds from the checkpoint row.

`TestRebuildDelta_NumscriptOrderDoesNotClobberMirrorMark` seeds a promoted ex-mirror ledger (NORMAL mode, `last_mirror_v2_log_id = 9`) from the checkpoint — not via a delta `CreateLedger`, which would reset the boundaries to genesis — then puts one numscript order in the delta and requires the mark is still 9. Mutation-checked with `go test -overlay`: fails `expected 0x9, actual 0x0` with the max replaced by an assignment.

The comment now cites that reason and the `compareMirrorV2LogID` strict-equality consequence, keeping the accurate `recordMirrorIngestMutations` clause. `FoldsHighestMirrorV2LogID` is kept but its comment no longer implies descending audit order is reachable — it pins the fold's shape, the new test pins the regression.

gfyrag:
Re-review on head `f6222986`. I still have two blocking concerns; the head has not changed since the previous analysis.

### [P2][blocking] Clear the persisted mirror error when the source recovers without advancing

After a successful ingest, `processBatch` records `lastPublishedSourceHead = sourceLogCount` (`worker.go:512-514`). If a later fetch fails, `reportError` persists a `MirrorStatus` error but does not mark the published status as dirty.

When the source recovers without a new log:

1. `FetchLogs` returns an empty page;
2. `processBatch` calls `publishSourceHead`;
3. the guard at `worker.go:539` returns because `sourceLogCount == lastPublishedSourceHead`;
4. the proposal carrying `ClearError: true` is never applied.

The API therefore keeps exposing the stale error until the source head advances. The empty-source case is stronger: a successfully observed head of zero is indistinguishable from “not observed”, so `sourceLogCount == 0` prevents the error from ever clearing, including after a worker restart.

Please track “status needs clearing” independently from “head needs publishing”. Once `reportError` is confirmed applied, a later successful empty fetch should issue a clear even when the head is unchanged. A separate “head observed” bit is also needed if zero is a valid observed head.

### [P2][blocking] Add automated coverage for the new source-head publication path

`publishSourceHead` adds a production Build → preload → propose → Raft wait → FSM wait path with several error branches, but no unit test references `publishSourceHead` or `lastPublishedSourceHead`. The only apparent happy-path exercise is in `restore_mirror_test.go`, behind `e2e && s3`, and no CI job runs the `s3` suites.

The current Codecov result confirms the gap: patch coverage is `24.52%`, `worker.go` has 40 changed lines missing coverage, and `codecov/patch` fails.

At minimum, please cover:

- successful idle publication and deduplication;
- recovery after a persisted error with an unchanged non-zero head;
- recovery for an empty source;
- Build/Run/Raft/FSM failures and retry eligibility.

### Non-blocking concerns

- `restore_mirror_test.go:376-426` still starts the server before sampling the rebuilt cursor. The leadership callback can start the worker before or during the first `GetLedger`, so a correct but slower run can observe a completed source request and fail the assertion. Reading the restored boundary directly before starting the node would remove the election race.
- The PR description still presents `SubPLMirrorSourceHead` reconstruction and the idle `SYNCING` state as an out-of-scope follow-up, while this head adds `publishSourceHead` specifically to change that behavior.

Targeted mirror/replay/backup/state tests pass under `-race`, the `e2e,s3` cluster package compiles, and `git diff --check` passes. Those checks do not exercise the recovery scenario above.

tests/e2e/cluster/restore_mirror_test.go
```diff
@@ -0,0 +1,498 @@
...
+			// proposal — never runs. To keep an idle restored mirror from
+			// reporting SYNCING forever, the caught-up worker publishes the
+			// refreshed source head on its own (EN-1773), so FOLLOWING becomes
+			// observable without waiting for a new source log.
+			Eventually(func(g Gomega) {
+				progress := syncProgress(g, client)
+				g.Expect(progress.GetError().GetMessage()).To(BeEmpty())
+				g.Expect(progress.GetCursor()).To(Equal(preBackupCursor))
+				g.Expect(progress.GetRemainingLogs()).To(BeZero())
+				g.Expect(progress.GetState()).To(Equal(commonpb.MirrorSyncState_MIRROR_SYNC_STATE_FOLLOWING))
```
Azorlogh, line 466: The mock reports the *oldest* log as the source head, so this assertion and `RemainingLogs == 0` are satisfied by any non-zero published head.

`mockV2Server.handler` always returns ascending logs and truncates from the front. `HTTPSource.GetLatestLogID` sends `pageSize=1` with no `after` and takes `Data[0]` — its doc says "newest first". Running the mock's handler against the real `HTTPSource` with logs 1..4 present returns head = **1**.

So here `sourceHead == 1` while `cursor == 4`, and `FOLLOWING` (`sourceHead > 0 && cursor >= sourceHead`) plus `RemainingLogs == 0` hold for any published head from 1 to 4. The spec still goes red without `publishSourceHead` (head 0 → SYNCING), so it does cover the new path — it just does not pin that the published value is the real head.

Adding `g.Expect(progress.GetSourceLogCount()).To(Equal(uint64(sourceLogCount)))` catches it. Worth settling whether the mock should reverse the no-`after` page or `GetLatestLogID` is wrong against the real v2 API.

This also blocks one of the tests requested in the re-review: because `GetLatestLogID` never returns 0 against this mock, the empty-source error-clearing case cannot be covered until the fixture is fixed.

internal/domain/replay/order_effects.go
```diff
@@ -45,17 +63,24 @@ func DecodeOrderEffects(serializedOrder []byte) (OrderEffects, error) {
...
+	// v2 log ids are 1-based, so zero means "no mirror ingest here" rather
+	// than a legitimate id — matching processMirrorIngest, which rejects a
+	// zero id outright instead of recording it.
+	mirrorV2LogID := entry.GetV2LogId()
+
 	ct := ls.GetApply().GetCreateTransaction()
 	isNumscript := ct != nil && (ct.GetScript().GetPlain() != "" || ct.GetNumscriptReference().GetName() != "")
 
-	if fillGap == nil && !isNumscript {
+	if fillGap == nil && !isNumscript && mirrorV2LogID == 0 {
```
Azorlogh: `IsNumscript` is dead, and the `!isNumscript` term here is what makes the hazard the new max fold defends against reachable.

`310ba1f4` added the field with two consumers, both doing `b.NumscriptExecutionCount++`. `5a2f8892` (EN-1420 counter cutover) deleted both — in `rebuild.go`'s `applyAuditOrderEffects` and `checker.go`'s `collectAuditOrderBoundaryEffects` — when the per-ledger usage counters moved to usagestore. The field and this guard term were left behind. Dropping `!isNumscript` and running all of `./internal/...` breaks only the new test case that asserts on the field.

Knock-on: this term is the sole reason a script-sourced `CreateTransaction` returns a non-empty `Ledger` with `MirrorV2LogID == 0`, which is exactly the input `TestRebuildDelta_NumscriptOrderDoesNotClobberMirrorMark` and the 15-line comment at `rebuild.go:862-882` exist for. Removing the field collapses that justification — the max fold is still the right shape since it matches the checker, but it needs one line of comment rather than fifteen.

The field's doc comment also still describes `NumscriptExecutionCount` on `LedgerBoundaries`; `raft_cmd.proto:686-688` records where it went.

internal/application/mirror/worker.go
```diff
@@ -508,6 +524,92 @@ func (w *Worker) drainPrefetch(ch chan prefetchResult) {
...
+// proposal, but a caught-up worker fetches no logs and would otherwise propose
+// nothing — leaving SubPLMirrorSourceHead at 0 after a restore, so
+// ReadMirrorSyncProgress reports SYNCING indefinitely until a new source log
+// arrives (EN-1773).
+//
+// It proposes only when the observed head changed since the last publish, so an
+// idle mirror does not re-propose the same value on every poll tick. A source
+// head of 0 (never observed, e.g. refreshSourceHead has not yet succeeded)
+// carries no information and is skipped.
+func (w *Worker) publishSourceHead(ctx context.Context) {
```
Azorlogh: This and `reportError` below are the same ~55 lines twice: identical `Build` → `Run` → `ReleaseLoaders` → Raft wait → FSM wait → business-error ladder and identical `SubAttrLedger` coverage, differing only in the `MirrorSyncUpdate` payload and four log strings.

A `proposeMirrorSync(ctx, update, label) bool` helper collapses both — and halves the error branches needing test coverage.

docs/technical/architecture/subsystems/events-mirror/mirror.md
```diff
@@ -174,7 +176,7 @@ There is no automatic "skip the broken log" mode. Operators investigate, fix the
 
 - **Async prefetch**: the next batch is fetched from the source while the previous batch is still applying through Raft + FSM. This overlaps source latency with consensus latency.
 - **Coverage pre-declaration**: the worker pre-computes the per-order `plan.Coverage` for the whole batch in one pass (`extractMirrorNeeds`, `worker.go:594-684`), so the per-proposal preload work is amortised.
```
Azorlogh: Stale anchor: `extractMirrorNeeds` was at `worker.go:594` at the merge base, and `publishSourceHead` moved it to 696.

tests/e2e/cluster/restore_mirror_test.go
paul-nicolas: Confirmed, and settled toward the fixture. Fixed in `405ee2d69`.

`PostgresSource.GetLatestLogID` implements the same interface method as `SELECT COALESCE(MAX(id), 0)` (source_postgres.go:121-135), and `TestHTTPSource_GetLatestLogID_Success` bakes in the newest-first premise (source_http_test.go:148-172), so "newest" is the contract and the handler was wrong.

The handler now answers the head-probe shape newest-first, and both the Phase-1 and Phase-3 specs assert `GetSourceLogCount()` as you suggested (plus `postRestoreLogID` on the fresh-ingest spec). Falsified: with the previous ascending probe the Phase-1 assertion fails after the full 60s timeout, head reading 1 instead of 4.

`pageSize` is the discriminator rather than the presence of `after`, because `FetchLogs` also omits `after` when resuming from 0. Which brings up your open question, which is bigger than the fixture: the two production call sites make **opposite** ordering assumptions about the same request shape — `GetLatestLogID` wants newest-first, `FetchLogs` wants ascending (`TranslateBatch` enforces contiguity). Both cannot be right against a live v2 ledger. That is pre-existing in `source_http.go`, so I left production alone and recorded it as a follow-up in the PR description rather than riding it here.

On the empty-source error-clearing case you flagged as blocked: it is now covered at the unit level instead (`TestWorker_IdleStatusClearsErrorForEmptySource`), which pins it more tightly than the e2e could and runs in CI, which the s3 suite does not.

internal/domain/replay/order_effects.go
paul-nicolas: Right on every point. Removed in `04d8288ba`.

Verified the field has no production reader: `grep -rn IsNumscript` finds it set at order_effects.go:83 and read nowhere; both consumers of `DecodeOrderEffects` use only `SkippedTransactionIDs` and `MirrorV2LogID` (rebuild.go:830-886, checker.go:6131-6149). The checker is unaffected — it verifies the mirror mark separately through `compareMirrorV2LogID` against `recordMirrorIngestMutations`.

And the knock-on landed exactly as you described: with `!isNumscript` gone, `TestRebuildDelta_NumscriptOrderDoesNotClobberMirrorMark`'s input no longer reaches the fold, so it goes vacuous and is deleted along with the `numscriptOrder` helper. The 15-line comment shrinks to the reason that survives — the checker's fold is a max and `compareMirrorV2LogID` is strict equality against it, so an assignment would diverge from the oracle.

I checked that this does not cost regression coverage before deleting: mutating the fold to a plain assignment makes `TestRebuildDelta_FoldsHighestMirrorV2LogID` fail (`expected: 0x9, actual: 0x4`). Its doc comment previously delegated the reachable case to the deleted test, so it now carries the checker-parity justification itself.

Also dropped the `NumscriptExecutionCount` bullet from the `OrderEffects` type doc, and re-pointed the `order_effects_test.go` case at what it now proves: a script-sourced `CreateTransaction` has no boundary effect, which pins that the guard keys off the mirror source id and the fill gap rather than off how the content was expressed.

internal/application/mirror/worker.go
paul-nicolas: Agreed — extracted as `proposeMirrorSync(ctx, update, label) bool` in `dd1a2080b`.

The helper owns the whole spine: coverage declaration, `Build` (with `ReleaseLoaders` on failure), `Run`, `Guard.ReleaseLoaders`, Raft wait, FSM wait, business-error ladder, all logging keyed on `label`. Both call sites are now just their payload plus their bookkeeping.

One divergence worth naming, since the `bool` resolves it: the two originals did not actually behave the same at the end. `publishSourceHead` returned early on `result.Error != nil` so it would not advance its bit; `reportError` only logged and fell through. Returning "was this confirmed applied" puts that decision at the call site, where it belongs — and it is load-bearing now that both callers have state to advance.

Your point about halving the error branches to cover paid off directly: `TestWorker_IdleStatusRetriesAfterFailedPropose` covers all four failure stages (propose call, Raft reject, FSM error, business error) against one implementation, asserting neither bit advances so the next tick retries.

docs/technical/architecture/subsystems/events-mirror/mirror.md
paul-nicolas: Fixed in `a8c909c5e`, along with a second one the comment did not name.

`extractMirrorNeeds` is now at `worker.go:701-791` (it moved again with `proposeMirrorSync`). The write-stall anchor on line 171 had also gone stale from the same PR — the new field and its comment shifted that block from `208-217` to `240-250` — so both are recomputed against the final head.

The idle-publication prose you saw at line 88 is rewritten too: it named `Worker.publishSourceHead`, which is now `publishIdleStatus`, and it described only the head-publication half of what the method does.

paul-nicolas:
Both blocking findings were real. Fixed and pushed; head is now `a8c909c5e`.

## [P2] Stale error at an unchanged head — `dd1a2080b`

Your trace was exactly right, including the restart argument for the empty-source case. Confirmed step by step: `processBatch` sets `lastPublishedSourceHead` on a confirmed ingest (worker.go:514), `reportError` touches no publication bit, and the recovery path then hits the `sourceLogCount == lastPublishedSourceHead` guard and returns before proposing. The only other `ClearError` carrier is the TU bundled into a data proposal, which needs a new source log — so the error survived until the head moved, and with `sourceLogCount == 0` it survived forever, a fresh worker starting at zero too.

Implemented as you asked, with both bits:

| Bit | Set | Cleared |
|---|---|---|
| `sourceHeadObserved` | `refreshSourceHead` on a successful `GetLatestLogID` | never — an observed head of `0` is a real value |
| `statusClearConfirmed` | confirmed apply of a `ClearError` update (idle publish **and** the bundled ingest TU) | confirmed apply of an error report |

The guard is now "return unless the head was observed; propose when the status needs clearing **or** the head moved". `statusClearConfirmed`'s zero value is deliberately `false`: a worker that has just started cannot know whether an error is persisted, so it publishes once rather than assuming a clean status. That costs one idempotent proposal per worker start, and it is what makes the empty-source case recover across a restart.

`SourceLogCount: 0` is a safe payload for a pure clear — write_set.go:197-198 documents a zero as "no write" for that field — so an empty source clears without inventing a head.

Renamed `publishSourceHead` to `publishIdleStatus`, since publishing the head is now half of what it does.

## [P2] Coverage — same commit

Confirmed the gap: nothing referenced either symbol in any `_test.go`. Now covered at the unit level, mapped to your four bullets:

- `TestWorker_IdleStatusPublishesSourceHeadOnce` — successful idle publication, then dedup on the next tick;
- `TestWorker_IdleStatusClearsErrorWhenHeadUnchanged` — recovery after a persisted error at an unchanged non-zero head;
- `TestWorker_IdleStatusClearsErrorForEmptySource` — empty source, including that an *unobserved* head still publishes nothing;
- `TestWorker_IdleStatusRetriesAfterFailedPropose` — table over propose-call / Raft-reject / FSM-error / business-error, asserting neither bit advances so the next tick retries.

The seam is a `stubProposer` that resolves both the Raft-acceptance future and the FSM future in-process and records the decoded `MirrorSyncUpdate`, so the assertions are on payloads rather than call counts. No Raft, no FSM.

Falsified rather than assumed: cases 2 and 3 both fail against the previous guard — "an observed empty source must still publish a clear" (0 proposals, wanted 1) and "a recovered source must clear the persisted error" (2 proposals, wanted 3).

This also removes the reliance you flagged on an s3-tagged suite that no CI job runs.

## Non-blocking: the Phase 3 election race — `405ee2d69`

Real, and fixed — though not by reading Pebble directly. That would mean importing `internal/storage/dal` and `internal/infra/attributes` into the e2e cluster suite, which no e2e test does today; a new pattern for a non-blocking flake felt like the wrong trade.

Instead the mock takes a pass-through barrier at the handler entry, *before* the request counter moves. Phase 3 pauses it, starts the node, samples, then resumes — so "no source fetch has happened" holds by construction rather than by beating the election, and the request-count assertion becomes a genuine invariant. The barrier is deliberately a separate mutex from the one guarding the log slice, so `requestCount()` stays answerable while requests are parked.

## Non-blocking: the PR description

Corrected. That section now describes `publishIdleStatus` as a third defect fixed here, including the error-clearing bug your review found in its first version, and notes that the rebuild still does not reconstruct the head row itself. The rewind-detector item stays as a genuine follow-up, joined by a new one: `FetchLogs` and `GetLatestLogID` make opposite ordering assumptions about the same no-`after` request shape, which surfaced while fixing the fixture and is pre-existing production behavior.

## Validation

- `bash scripts/agent-check` — PASS, 0 lint issues on both modules
- `go test -race` on mirror / replay / backup / state / check / query / adapter-v2 — all pass
- `-tags e2e` mirror specs — 21/21; `-tags e2e,s3` restore suite — 8/8

One further finding of my own while in here: `OrderEffects.IsNumscript` had been dead since the EN-1420 counter cutover, and its guard term was what made the rewind hazard reachable at all. Removed in `04d8288ba`, with the max fold's remaining justification (checker parity) confirmed still pinned by mutation.

internal/application/mirror/worker.go
```diff
@@ -508,28 +547,86 @@ func (w *Worker) drainPrefetch(ch chan prefetchResult) {
...
+		},
+	}
+
+	if !w.proposeMirrorSync(ctx, update, "mirror error report") {
+		return
+	}
+
+	// An error is now persisted, so the next idle tick must issue a clear even
+	// if the source head has not moved.
+	w.statusClearConfirmed = false
```
Azorlogh, line 610: `proposeMirrorSync` returning false means "I did not get confirmation", not "the update was not applied" — so bailing out before this line can leave the flag `true` while an error is persisted, which is the same stale-status class the head-unchanged fix closed.

Both waits inside `proposeMirrorSync` abandon on context cancellation (`futures/future.go` returns `ctx.Err()` when the future has not resolved), and cancelling a wait does not cancel a committed Raft entry — every node still applies it. So: healthy ingest sets the flag `true`, a poll fails, `reportError`'s entry commits, the worker's wait is interrupted, `false` is returned, the flag stays `true`. The source then recovers with no new log and `publishIdleStatus`'s `statusClearConfirmed && head unchanged` guard suppresses the clear indefinitely.

Mostly self-healing, which is why this is not blocking: the realistic cause of the interrupted wait is worker teardown (leadership change, ledger deletion), and a replacement worker starts at the zero value `false` and publishes a clear on its first idle tick. It needs the same worker to survive the interrupted wait. The other `false` returns — stale proposal, business rejection — genuinely mean "not applied", so leaving the flag alone is right for those.

Suggest setting `w.statusClearConfirmed = false` *before* the propose call, reading it as "an error may now be persisted": being wrong that way costs one idempotent idle publish, being wrong the current way costs a permanent error on a healthy mirror — the same asymmetry the zero-value comment on the field already resolves this direction.

docs/technical/architecture/subsystems/events-mirror/mirror.md
```diff
@@ -166,15 +173,15 @@ This applies to any mode; it is documented here because mirror ledgers are the p
 | Source unreachable | `FetchLogs` returns an error → the worker writes the error into `MirrorStatus` via a small technical-update proposal, then retries with exponential backoff (`worker.go:225-237`). |
```
Azorlogh: Stale anchor missed by the same pass: the `reportError` call is now `worker.go:247-259`.

internal/application/mirror/worker.go
paul-nicolas: Valid, and taken in `9fad63d3c`.

Your mechanism checks out end to end. `futures.Future.Wait` returns `ctx.Err()` on an unresolved future (`internal/pkg/futures/future.go:71-79`), both waits in `proposeMirrorSync` go through it, and abandoning a wait does not un-commit a Raft entry — so a committed error report can return `false`, and `publishIdleStatus`'s guard then suppresses the clear until the head moves.

One correction to the framing: the scenario is currently unreachable, not latent-live. The only cancellation source is the loop context — `reportError` is called from `processLogs` with the raw loop ctx and no per-call timeout — and once that ctx is cancelled, `processLogs` and `loop` both return, so the worker holding the stale bit dies. You reach the same conclusion via the replacement worker; it is stronger than that, the surviving-worker case has no path at all today.

I took the change anyway, for two reasons beyond the asymmetry you name:

1. It makes the invariant local. Today its correctness rests on "ctx cancellation implies worker death", which nothing in `reportError` states or enforces.
2. It is one refactor away from live. `refreshSourceHead` already wraps its call in a `context.WithTimeout`; the day `reportError` gets the same treatment, the bug is real.

So the bit is now written before the propose, read as "an error may now be persisted", and the bool return is dropped there. `publishIdleStatus` is unchanged on purpose: an unconfirmed clear leaves both bits unadvanced and retries next tick, which is already the safe direction. I updated `proposeMirrorSync`'s doc comment, since "the bool return is what makes the caller's bookkeeping safe" is now true of one caller only.

Covered by `TestWorker_ReportErrorMarksStatusDirtyWithoutConfirmation`, built on a new `proposeWaitAbandoned` stub outcome — the only one that models a committed entry whose confirmation never arrives (it records the proposal and resolves neither future, so a cancelled-ctx caller gets `ctx.Err()`). The test asserts both the bit and the consequence: the source recovers with no new log and the clear is issued at an unchanged head.

Falsified rather than assumed — reverted to the previous ordering:

```
--- FAIL: TestWorker_ReportErrorMarksStatusDirtyWithoutConfirmation (0.08s)
    worker_test.go:426:
        Error:      Should be false
        Messages:   an unconfirmed error report may still have been applied, so the status must be marked dirty
```

Validation: `bash scripts/agent-check` PASS with 0 lint issues on both modules; `go test -race` green on mirror, state and query.

docs/technical/architecture/subsystems/events-mirror/mirror.md
paul-nicolas: Right that the anchor is stale, and it is this branch's fault: the new field comments above `worker.go:94` shifted everything below, and `git diff 93876adf6...HEAD -- .../mirror.md` shows the refresh pass touched exactly two rows (write-stall `208-217`→`240-250`, `extractMirrorNeeds` `594-684`→`701-791`) and missed this one.

Your replacement is off, though: `247-259` is base+22 applied mechanically, and it lands on the write-stall block, stopping before the backoff computation that the row's second clause names. At `d5ccfd8b5` the `reportError` call is at `worker.go:255`, and the row describes "writes the error … then retries with exponential backoff", which is the whole `if err != nil` branch.

Two more anchors in the file were stale from the same shift and would have been the next round of this comment, so I finished the pass instead of fixing one row. Fixed in `248c4729a`:

| Row | Was | Now |
|---|---|---|
| Source unreachable | `worker.go:225-237` | `worker.go:253-275` |
| Prefetch | `worker.go:425-447` | `worker.go:464-490` |
| "The Worker (…) is a polling loop" | `worker.go:27-167` | `worker.go:27-175` |

Numbers are against the final tree, so they account for the ~9 lines `9fad63d3c` adds to `reportError` — that also moved `extractMirrorNeeds` to `710-800`. Every anchor in the file now lands on the exact opening and closing line of the construct it names, verified per anchor rather than by arithmetic.

Left alone: the two `manager.go` anchors. This branch does not touch that file, so whatever their accuracy, it is unchanged here and correcting it belongs in its own pass.

Same commit adds a sentence to the idle-status section recording why the error report marks the status dirty before proposing rather than after a confirmed apply — the behaviour from your other thread.

gfyrag (approved):
Approved — no blocking code findings at head 248c4729a45c2e165fee8cff6ceabda5b957ea9f.

Non-blocking follow-ups:
- Refresh the PR description section “Automated coverage boundary for EN-1776”: Tests-E2E now includes the s3 tag, and TestRebuildDelta_NumscriptOrderDoesNotClobberMirrorMark was removed.
- Track the pre-existing HTTP v2 source contract mismatch separately. The release/v2.4 logs endpoint defaults to descending order and does not consume the after parameter, while the fixture intentionally models different ordering for the head probe and forward fetch.

Validation considered: local scripts/agent-check, focused race-enabled tests for mirror/replay/backup/state/checker, and the green GitHub CI matrix.

PR #1690, opened 2026-08-12
test(EN-1772): assert the AuditFailure message derivation for every Describable

paul-nicolas:
## What

`AuditFailure.Message` is derived from a `domain.Describable` by one unconditional
assignment in `buildAuditFailure`, and that derivation was asserted for exactly **one**
of the 74 `Describable` types that can reach it. Deleting the assignment left 13 of the
14 subtests of `TestBuildAuditFailure` green.

- `TestBuildAuditFailure` is now table-driven — **83 rows over 74 types** — with reason,
  message and context asserted in a **shared body**, so no row can assert a subset. The
  per-row assertion drift is what left the message unpinned while `Reason` was pinned
  everywhere.
- `TestBuildAuditFailureCoversEveryDescribable` discovers `Describable` implementations
  **by method set** (`Reason()` + `Metadata()` receivers) across `internal/domain` and
  `internal/infra/state`, and fails if any lacks a row. It catches
  `domain.ReplayedFailure`, `domain.RemoteError` and `domain.BusinessError`, which the
  name-prefix scan in `internal/domain/errors_test.go:318` misses — and `ReplayedFailure`
  is the type the FSM builds on **every idempotent retry** and feeds straight back into
  `buildAuditFailure`, so that blind spot is the replay path, not a hypothetical.
- The wrapped-`CoverageMiss` no-unwrap control moves to its own test rather than becoming
  a row: it exercises two errors and asserts a negative, so a row would register
  `ErrStorageOperation` in the coverage set on the strength of a test that is not about
  that type's projection.

## Why there is a production change in a tests-only ticket

`d.Error()` was written into the audit chain by `buildAuditFailure` and into the
`SubIdempKeys` projection by `recordIdempotencyFailure` — two derivations, two files —
while the checker's `idempotencyMismatch` (reached from `compareIdempotencyOutcomes`)
requires them byte-equal. So the failure mode was never "a lost message": it was a
**healthy store reporting `CHECK_STORE_ERROR_TYPE_IDEMPOTENCY_MISMATCH`** on every frozen
failure, unrepairable after the fact because one side lives inside the hash chain.

Both sites now consume one `describeFailure` helper, making that equality structural
rather than coincidental. No behavioural change: same values, same protos, no wire change.

Metadata is deliberately **not** folded into the helper — the two sites treat it
asymmetrically on purpose, and on the checker's real read path the asymmetry never even
shows (a proto3 map with no entries emits no bytes, so both sides read back nil).

## Verification: by mutation, not by green

All 83 rows passing proves nothing on its own — a row whose expectation was copied out of
a failure diff also passes, forever, while asserting nothing. So acceptance was mutation-based.
24 injected bugs, every one caught by the row or test that should own it:

| Mutation | Caught by |
|---|---|
| message dropped in `buildAuditFailure` | **all 83 rows** (was 13/14 green before this PR) |
| `ErrCoverageMiss` reverted to snake_case — the literal EN-1379 regression | `CoverageMiss` |
| `ErrInsufficientFunds` gate `ColorKnown` → `Color != ""` | the two rows that distinguish them |
| `ErrNumscriptNotFound` / `ErrTransactionReferenceConflict` gates dropped | the ungated rows |
| `ErrIdempotencyCheckFailed` / `ErrStorageOperation` leak `Cause` into context (#326) | those rows |
| `BusinessError` / `RemoteError` / `ReplayedFailure` stop delegating | those rows |
| a new `Describable` added with no row | the forcing function |
| one derivation site returns `message + "MUTANT"` | the A == B pin |

The four **gate** mutations are the decisive ones: a diff-pasted expectation matches
whatever the gate currently does, so a pasted table would have passed all four. All four
failed. Every row also cites the `Metadata()` line it was authored from; 96 citations were
verified against source, 0 wrong.

## Known limitation

Message **wording** is not pinned. `wantMessage` is derived as `tc.err.Error()`, so
rewording a domain error still changes hash-chained content without a failing test. This is
deliberate: 74 literal strings would invite blind bulk-updates that destroy the signal.
`wantContext` **is** literal, because the sorted context keys and values are folded into the
hash pre-image and a renamed key is exactly what shipped green in EN-1379. 13 field-name-to-key
divergences are now frozen, 5 of them semantic (e.g. `ErrInvalidReceipt.Detail` → `"reason"`).

`RemoteError` is documented server-side-forbidden, so it cannot reach `buildAuditFailure`
today. It has a row because the method-set scan is deliberately conservative — not a
reachability claim.

## Follow-ups (deliberately not in scope)

1. `ErrMetadataKeyValidation` / `ErrAccountValidation` seed a wrapper key (`"key"` /
   `"account"`) then `maps.Copy` the cause's metadata **over** it, so a cause carrying the
   same key would overwrite the wrapper's and the chain would record the inner value as the
   offending one. Unreachable today, so the table uses non-colliding causes rather than
   freezing the bug as the contract. Fix: copy the cause first, seed the wrapper key last.
2. Unify `internal/domain/errors_test.go:318` on method-set discovery, so the tree has one
   discovery convention instead of two.

## Gates

`just pre-commit` (0 issues) · `just test` (33 packages ok, 0 FAIL) ·
`go test ./internal/infra/state/... ./internal/domain/... ./internal/application/check/...`

## Ticket

https://formance-team.atlassian.net/browse/EN-1772


gfyrag (approved):
LGTM. The production refactor preserves behavior and makes the audit/idempotency failure contract structural. The AST scanner blind spot for generic or embedded receivers can be addressed separately as a non-blocking follow-up.

PR #1691, opened 2026-08-12
test(EN-1775): pin every REST route to its required scope

paul-nicolas:
Jira: https://formance-team.atlassian.net/browse/EN-1775

## Problem

`internal/adapter/http/handler.go` expresses the scope a REST route requires **only** by which `RequireScope(...)` `Group`/`Route` block lexically encloses the route's line. Nothing records that decision, so moving a route between two adjacent groups is a one-line diff that silently changes its authorization and keeps CI green.

That is an authorization defect class, not a cosmetic one, and it has already happened: EN-1508 found `ListLogs` demanding `ledger:OpsRead` over gRPC where HTTP demanded `ledger:LedgerRead`. The aggregate `ledger:read` scope expands to **both**, so the drift was invisible to every caller using aggregate scopes and only surfaced for callers on granular ones.

The gRPC side has had a guard for this since EN-1508 (`TestRequiredScopeForRequest_ProtoExhaustive`). The HTTP side had none.

## What this adds

A declared route-to-scope contract for the REST router, enforced two ways.

**Reconciliation** (`TestRouteScopes_Exhaustive`) — the declared table vs `chi.Walk`, in both directions. A registered route with no table row fails ("no scope decision"); a table row matching no route fails ("stale row"). This is what stops a new route under the `/v3/_/…` subtree from silently inheriting `ledger:OpsRead`, which the comment there actively invites.

**Probes** (`TestRouteScopes_GuardedRoutesRequireExactlyOneScope`) — for each of the 68 guarded routes, a token carrying the declared scope must be admitted, and a token carrying **every one of the other 13 granular scopes** must be refused with an exact 403. That proves exactly one scope opens each route, rather than the weaker "some scope is rejected".

Granular scopes only, never the aggregate `ledger:read` — the aggregate expands to both `LedgerRead` and `OpsRead` and would mask precisely the EN-1508 drift.

### Measured shape

| | |
|---|---|
| `chi.Walk` entries | 122 |
| Declared table rows | 74 (68 guarded + 5 public + 1 per-element) |
| Probes per run | 958 (68 positive, 884 negative, 5 public, 1 per-element) |
| Runtime | 0.32s for the full sweep; `-race` and `-count=3` green |

The 6 pprof endpoints registered with `r.Handle` answer all 9 HTTP methods, so they carry an explicit `anyMethod` flag that expands to 9 walk entries each — declaring the fan-out rather than hiding it behind 54 rows.

## Proof that the tests can fail

Every assertion here was mutation-tested. The two that matter most:

- **EN-1508 replayed.** Moving `GET /{ledgerName}/logs` from `requireLedgersRead` into the `requireTransactionsRead` group fails with two messages naming the route: *"must admit its declared scope ledger:LedgerRead but answered 403"* and *"must refuse scope ledger:TransactionRead — only ledger:LedgerRead opens it — but answered 500"*. Reconciliation stays green (the route set did not change), which is what proves the probes are load-bearing rather than decorative.
- **An undeclared route.** Adding a route without a table row fails reconciliation: *"registered by handler.go but have no row in expectedRouteScopes — add one declaring the granular scope each requires"*.

Also verified to fail: a duplicate table row, a `kindGuarded` row with an empty or non-existent scope, and a probe target that routes to the wrong pattern.

## Two things worth a reviewer's attention

**The middleware chain is not a usable oracle, and the test deliberately does not use it.** `chi.Walk` hands you the middleware slice, which invites inferring the scope from it. That is unsound: `With(mw).Route(prefix, …)` attaches the middleware to the sub-mux's *handler* rather than to `Middlewares()`, so `GET /v3/_/chapters` reports no `RequireScope` while returning 401 to a tokenless request. All `RequireScope` closures also share one code pointer, so reflection cannot distinguish them. A chain-inspecting test would report the entire 9-route `/_` subtree as unguarded. Only a real request through `ServeHTTP` reveals the scope.

**Probes carry an already-cancelled context.** Without it, `/debug/pprof/profile` runs for 30s (`seconds<=0` falls back to 30, so `?seconds=0` is no escape) and `/trace` for 1s — measured at 30.099s vs 205ms. This cannot make the positive probes vacuous, and that was verified rather than argued: all 63 non-pprof positive probes were run with cancelled vs live contexts and the statuses diffed, **0 of 63 differ**. Nothing in the middleware chain reads `ctx.Err()`, and `StaticKeySet.VerifySignature` takes a ctx and never reads it.

## Notes for review

- The positive assertion is `NotEqual(401) && NotEqual(403)`, which also tolerates 404. It is sound because the paired negatives assert an exact 403, unreachable unless chi matched a route and ran its gate — and because each probe now also asserts it hit the pattern it claims. **Do not weaken the negatives to "not 200"**; that deletes half the argument. The file says so at the assertion.
- **The token fixture is package-local on purpose.** `internal/adapter/auth` has an equivalent `testKeyPair`/`signToken` in its own test files, and this PR does *not* share them. It cannot: those tests are `package auth`, so nothing outside the package can import them, and the only way to share would be a package that exists solely for tests. An earlier revision of this PR did exactly that (`internal/adapter/auth/authtest`) and it was rejected — a ~40-line fixture over an API that does not churn is the cheaper duplication. `internal/adapter/auth` is therefore byte-identical to `release/v3.0`; the reasoning is recorded on `testKeyPair` so a future reader does not "fix" it by re-extracting.
- No production file is modified, and no existing test file is touched. Purely additive test code plus one docs section.

## Follow-ups filed

- **EN-1777** — `/clusterz` is public by omission of a scope guard rather than by declaration, and is absent from `HTTPAuthMiddleware`'s exempt list, so a tokenless request reaches the handler while an invalid-token one gets 401. Current behaviour is pinned by this PR; the asymmetry is not fixed here because the change had to stay additive. Also covers `docs/ops/authentication.md:94` omitting `/_info`.
- **EN-1778** — filed to consolidate the remaining token-minting duplication (`tests/e2e/cluster/auth_test.go:81,102` and the in-package EdDSA helpers) onto a shared package. Its premise is void now that the shared-package approach is rejected; commented and left for closure.


internal/adapter/http/route_scope_exhaustiveness_test.go
```diff
@@ -0,0 +1,885 @@
...
+
+		probed++
+
+		method := row.probeMethod()
+
+		status := probe(t, probeRequest{
+			handler:         fixture.handler,
+			method:          method,
+			pattern:         row.pattern,
+			token:           fixture.tokens[internalauth.ScopeClusterRead],
```
NumaryBot: 🟠 **[major] Probe per-element routes without a scope**

If the bulk route is accidentally wrapped in `RequireScope(..., ScopeClusterRead)`, this token is admitted and the test still passes while the route now has a route-level gate. Checking that the declared table does not use ClusterRead is circular because the table remains `kindPerElement` in exactly that regression; use a tokenless request (with the default mapping) so every `RequireScope` yields 401.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1691#issuecomment-5266959428

internal/adapter/http/route_scope_exhaustiveness_test.go
paul-nicolas: Fixed in e5f83fbc4 — the probe now sends no token at all.

Confirmed the gap before changing anything: with `r.With(internalauth.RequireScope(authCfg, internalauth.ScopeClusterRead))` injected around the bulk route in handler.go, the previous form of this test **passed**. The row stays `kindPerElement` so `scope` is `""` and the premise loop at 866-871 was satisfied, while the gate admitted the foil so the 403 assertion was too.

Tokenless is the right oracle: `HTTPAuthMiddleware` falls through to `cfg.ScopeMapping.AnonymousScopes()` with `AuthPresented=false` (http_middleware.go:46-56), `DefaultMapping` declares no `anonymous` key so that set is empty (scopes.go:118-124), and `RequireScope` then answers 401 whatever scope it demands (http_middleware.go:110-125). No foil to pick, no scope left uncovered — the same injected gate now fails the test.

The premise assertion moved rather than disappeared: it now requires `DefaultMapping("ledger").AnonymousScopes()` to be empty, read off the mapping the router is built from instead of off the table under test, so it is no longer circular.

paul-nicolas:
Addressed in e5f83fbc4. The per-element probe no longer relies on a foil scope: it sends no token, so any route-level `RequireScope` answers 401 regardless of which scope it requires (`DefaultMapping` grants nothing anonymously, so `HasScope` fails on the anonymous set). Verified both directions against an injected `RequireScope(authCfg, ScopeClusterRead)` on the bulk route — the previous form passed, the new one fails.

internal/adapter/http/route_scope_exhaustiveness_test.go
```diff
@@ -0,0 +1,895 @@
...
+
+		switch {
+		case row.kind == kindGuarded && row.scope == "":
+			malformed = append(malformed, fmt.Sprintf(
+				"%s %s is kindGuarded but declares no scope", row.method, row.pattern))
+		case row.kind == kindGuarded && !granular:
+			malformed = append(malformed, fmt.Sprintf(
+				"%s %s is kindGuarded but declares %q, which is not in internalauth.AllGranularScopes",
+				row.method, row.pattern, row.scope))
+		case row.kind != kindGuarded && row.scope != "":
```
NumaryBot, line 516: 🟠 **[major] Reject unknown route kinds**

A row with an undefined `routeKind` and an empty scope passes this validation, but is skipped by all three probe tests. Such a row still satisfies route reconciliation while leaving that endpoint's authorization unverified; explicitly reject kinds other than `kindGuarded`, `kindPublic`, and `kindPerElement`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1691#issuecomment-5266959428

internal/adapter/http/route_scope_exhaustiveness_test.go
paul-nicolas: Fixed in 3be261c99 — `TestRouteScopes_TableComposition` now rejects any row whose kind is none of the three.

Confirmed the gap first: swapping the bulk row to `routeKind(7)` passed every check before this commit. The switch only related kind to scope, so an unknown kind with an empty scope matched no case; reconciliation compares `(method, pattern)` only; and each probe loop opens with `if row.kind != kindX { continue }` while its `require.NotZero(probed)` stays satisfied by the other rows.

The new case fires first so the message is about the kind rather than the scope, and it names the remedy: `POST /v3/{ledgerName}/bulk declares routeKind 7, which no probe test covers — add its probe loop`. So a fourth kind cannot be introduced without its own probe.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1691#issuecomment-5266959428

gfyrag (approved):

PR #1698, opened 2026-08-12
fix(EN-1288): require an explicit selector on executable numscript references

Azorlogh:
## Summary

Follow-up to #1529, which did not implement the [EN-1288](https://formance-team.atlassian.net/browse/EN-1288) rule for **executable** numscript references. The ticket is explicit:

> For executable `NumscriptReference` values:
> * `version` is required;
> * accepted values are an exact full semver or the literal `"latest"`;
> * partial selectors such as `1` and `1.2` are not accepted for execution in v3.0;

#1529 shipped the empty string as a synonym for `latest` on the executable path:

```go
// admission.go (before)
if v := ref.GetVersion(); v != "" && v != "latest" {
    if _, perr := semver.Parse(v); perr != nil { /* NUMSCRIPT_INVALID_VERSION */ }
}
```

That `v != ""` short-circuit meant an empty selector never reached `semver.Parse`. It also propagated into `refIsLatest`, `resolveNumscriptReference`, and the FSM's apply-time resolution, all three of which treated `""` as `latest`.

## Change

The gate is now `v != "latest"`, letting `semver.Parse` reject everything else, so an empty selector returns `NUMSCRIPT_INVALID_VERSION` (400 / `InvalidArgument`) alongside the partials. The FSM resolves the latest pointer only for the literal `"latest"`.

`GetNumscript` and the other read-only APIs are untouched — the ticket explicitly keeps omitted and partial selectors there, since they resolve outside the FSM where a Pebble scan is allowed.

The two rejected forms fail for **different** reasons, which the docs now spell out rather than lumping together:

- An **empty** selector is resolvable (it would follow the same latest pointer as `"latest"`), but the submitted selector is what the audit chain records, so an executable reference must name the version it intends.
- A **partial** selector additionally resolves through a Pebble range scan the FSM apply path cannot make (invariant #3).

## Also in here

- **The antithesis numscript driver would have silently died.** `parallel_driver_numscripts` built its `ScriptReference` with no version, so every reference transaction would now be rejected and that driver would stop exercising numscript references entirely. Set to `"latest"`, preserving the exact path it was covering (FSM latest-resolution plus the stale-retry race).
- **Two docs were already wrong before this change.** `numscript-library.md` claimed partial selectors were "valid for `GetNumscript` and transaction references" — admission has rejected executable partials since #1529, so the doc contradicted the code and its own e2e test. `api-comparison.md` still described the *superseded alpha* save model (`version` optional, `"latest"` overwrites content, defaults to `"latest"`) that #1529 was meant to delete.
- **Dead code from #1529 removed.** With the only caller now passing a full semver, the partial-scan half of the overlay lookup was unreachable (`ParsePartial` → `depth == 3` → early return), along with a `key.Version == "latest"` guard left over from the removed mutable-latest slot. The exact-match `Get` that remains is inlined at its one call site.
- `openapi.yml` marks `ScriptReference.version` required; the proto comments on `NumscriptReference` / `ScriptReference` state the contract.

## Testing

- New `TestExecutableReferenceVersionSelector` pins the contract: `""`, `1`, `1.2`, `bogus`, `1.0.0-rc1`, `01.0.0`, `LATEST` rejected with the typed error and reason; `latest` and `1.0.0` accepted with the selector left unmodified on the audited order. **Verified red against pre-fix `admission.go`, and only on the `version=""` case** — the partials were already caught.
- The e2e reference test now asserts the `NUMSCRIPT_INVALID_VERSION` reason, not just the gRPC code. Confirmed the mapping really produces it (`convertToGRPCError` → `describableToGRPCStatus`): `code=InvalidArgument reason=NUMSCRIPT_INVALID_VERSION`.
- `CreateTransactionOrder` is constructed in exactly one place, inside the same `Admit` pipeline that runs the gate, so no mirror, replay, or restore path reaches the FSM unvalidated. The FSM change is defensive only.
- Full unit suite green (except three pre-existing `internal/pkg/tarutil` cases that chmod a dir to `0444` and cannot fail as uid 0); `GOROOT= just pre-commit` clean, 0 lint issues on both modules; `e2e` and `scenario` tags vet clean.

## Known gap (not addressed here)

Nothing in `internal/domain/` constructs a `NumscriptReference`, so the FSM-side resolution — the `"latest"` pointer read, the coverage check, and the `ErrStaleProposal` path — has no unit test, before or after this change. Happy to add that here if you'd rather not carry it.

[EN-1288]: https://formance-team.atlassian.net/browse/EN-1288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/application/admission/admission.go
```diff
@@ -2526,8 +2527,9 @@ func (a *Admission) resolveNumscriptReference(overlay *bulkOverlay, ledgerName s
 		}
 
 		version = greatest
-	} else if content, resolvedVersion, found := a.resolveNumscriptFromOverlay(overlay, ledgerName, name, version); found {
-		return content, resolvedVersion, nil
+	} else if content, ok := overlay.numscriptEntries.Get(numscriptEntryKey{Ledger: ledgerName, Name: name, Version: version}); ok {
```
gfyrag, line 2523: [P2] Cette nouvelle branche de résolution exacte dans l’overlay n’est pas couverte : le test same-bulk existant passe uniquement `latest`, et `codecov/patch` échoue actuellement à 66,67 %. Peux-tu ajouter un cas où une version est sauvegardée puis référencée explicitement dans le même batch ? Cela vérifierait que l’on utilise bien l’overlay au lieu de retomber sur Pebble avant que la sauvegarde soit persistée.

gfyrag (commented):
[P2] La documentation d’idempotence reste désynchronisée avec ce changement. `docs/technical/architecture/subsystems/admission/idempotency.md` classe encore à quatre endroits (`L72`, `L84`, `L91`, `L104`) les sélecteurs `latest` et vide ensemble. Puisque cette PR rejette désormais le sélecteur vide à l’admission, ces passages et le diagramme devraient mentionner uniquement le littéral `latest`.

internal/application/admission/admission.go
Azorlogh: Fixed in 166f6b9d375cdd285b33529e387fbb58533a8d95.

`TestResolveScripts_IntraBatchNumscriptExactReference` saves `1.0.0` into the overlay with nothing in Pebble, then references it explicitly — so resolution can only succeed through the overlay. Contrast case with an empty overlay asserts `NUMSCRIPT_NOT_FOUND`.

Coverage profile confirms `admission.go:2530.129,2533.3` went 0 → 1 (the missing + partial codecov reported); the contrast case also covered the `NOT_FOUND` return at `2540.17,2542.3`. The only line left uncovered in that function is the `query.ReadNumscript` storage-error branch, which isn't in this diff.

Azorlogh:
> [P2] La documentation d’idempotence reste désynchronisée avec ce changement. `docs/technical/architecture/subsystems/admission/idempotency.md` classe encore à quatre endroits (`L72`, `L84`, `L91`, `L104`) les sélecteurs `latest` et vide ensemble.

Fixed in 166f6b9d375cdd285b33529e387fbb58533a8d95 — `/empty` dropped at all four spots, including the two mermaid nodes (`FRZLATEST`, `LATEST`). Only the literal `latest` reaches that forwarding decision now.

I re-ran the sweep untruncated to check for others: the remaining `empty`/`latest` pairings are all read-only surface (`numscript-library.md` resolution table + HTTP row, `api-comparison.md` GET rows, `cli.md` `numscripts get --version`), which this PR doesn't change.

gfyrag (commented):
Independent review at head 166f6b9d375cdd285b33529e387fbb58533a8d95. One blocking finding remains.

[P2][blocking] Validate executable selectors before skipped-order short-circuiting
Location: internal/application/admission/admission.go (resolveScriptsAndEnrichNeeds)
Evidence: predictOrderSkip runs before NumscriptReference.version validation and immediately continues for a predicted TRANSACTION_REFERENCE_CONFLICT. A caller can opt into that skippable reason, provide an existing or same-batch duplicate reference, and attach an empty/partial selector. Apply checks reference uniqueness first and emits a successful skipped order without resolving the invalid script.
Impact: the API accepts and audits a selector that the protobuf/OpenAPI/docs say must be rejected with NUMSCRIPT_INVALID_VERSION.
Resolution: structurally validate all executable selectors before predictOrderSkip, with persisted and same-batch duplicate regression tests.

Targeted admission/FSM/query tests, diff check, and CI passed.

Azorlogh:
> [P2][blocking] Validate executable selectors before skipped-order short-circuiting

Confirmed and reproduced. Fixed in e81f3bb9fac69b1d1a5b2dca349bccc9b251d40a.

Your read of the mechanism is exact: `predictOrderSkip` runs at `admission.go:1683`, the `CreateTransaction` branch does `if skip { continue }` at line 1705, and the selector gate sat at line 1800 — later in the same iteration. `processCreateTransaction` checks reference uniqueness in its dry prologue (before script resolution), so the FSM emits an `OrderSkippedLog` and the selector is never examined by anyone.

It is cheaper to trigger than the report suggests. A **single** bulk request on an **empty** ledger is enough via the same-batch route — entry 1 claims a reference, entry 2 reuses it with `skippableReasons: [TRANSACTION_REFERENCE_CONFLICT]` and `version: ""`. No prior state, and the script name need not exist (`NUMSCRIPT_NOT_FOUND` is bypassed too). The client gets a success with `OrderSkippedResponse` instead of a 400. Worth noting `skippableReasons` is the documented continue-on-failure path — the `409` on the single-transaction endpoint actively recommends it — so this is on a supported flow, and the practical shape is a client that omits `version` (legal before this PR) retrying a reference-idempotent batch: rejected on first submit, silently accepted on the retry where the reference now exists.

Fix follows your resolution. The selector is a pure function of the request, so it moved to `validateOrderContent` — the structural gate in `validateOrder`, which every accepted order passes through during `requestToOrder`, well before the fold loop. The gate at 1800 is gone, so there is exactly one validation site. Script *existence* deliberately stays behind the skip guard: it is state-dependent, and short-circuiting it for an order that will not apply is correct (and avoids folding phantom effects into the resolution hash).

Regression tests as requested, both duplicate routes, driving `Admit` rather than the resolution helper: `TestExecutableReferenceVersionSelector_SkippedOrder` covers persisted and same-batch duplicates across `""`, `"1"`, `"1.2"`, `"bogus"`. I verified they fail against the pre-fix source and pass after. The selector table now targets the structural gate — worth flagging that on the old code a `"1"` selector *resolved* through the partial-scan path once resolution was reached, which is the other reason this belongs in a structural gate.

Full unit suite, `-tags e2e` / `-tags scenario` vet, and `just pre-commit` all clean.

Azorlogh:
Merged `release/v3.0` in (68 commits) and pushed c470693e6f83fb73f79d31db00633504d98b1168. Clean merge, no conflicts.

**Heads-up: this touches a test #1791 just added.** `Tests` went red on e81f3bb9f, and the cause is a contract collision rather than a defect on either side. `3a56b108c` (#1791) added `processor_transaction_numscript_reference_test.go` — the FSM-side coverage I flagged as missing in the PR description — and its table includes:

```go
{name: "empty selector", selector: "", resolvedVersion: "2.0.0", resolveLatest: true},
```

That asserts apply resolves an empty selector through the latest pointer, which is exactly the behaviour this PR removes. CI tests the PR merged into base, so the row met the change and failed: `GetNumscriptLatestVersion` never called, `ResolveNumscriptContent` invoked with `""` instead of `"2.0.0"`.

Per EN-1288 an empty executable selector is rejected at admission, so it cannot reach apply and that row asserts an unreachable state. I moved it to the sibling rejection table as `empty selector is not resolved as latest`, shaped like the existing `content not found` row: it sets **no** `GetNumscriptLatestVersion` expectation, so gomock strictness now pins the load-bearing half — the pointer is never consulted for an empty selector — and the read falls through to the exact-content lookup, yielding `NUMSCRIPT_NOT_FOUND` with an empty version. Net effect is that the FSM-side coverage #1791 added is kept, with the semantics the ticket specifies.

Flagging it explicitly since it is someone else's test and I would rather you confirm the reading than have it slip through in a merge commit. If you would prefer the row simply deleted, or kept as an explicit `ErrNumscriptInvalidVersion` assertion at apply instead, say so and I will change it.

`bash scripts/agent-check` PASS (including `check-repo-invariants`, 0 lint issues on both modules, no proto regen drift), full unit suite green, `-tags e2e` and `-tags scenario` vet clean.

gfyrag (approved):

PR #1700, opened 2026-08-13
fix(EN-1622): serialize transactions/chapters/log responses through their MarshalJSON

paul-nicolas:
Fixes [EN-1622](https://formance-team.atlassian.net/browse/EN-1622).

## The problem

One Go type, `commonpb.Transaction`, was rendered by two different encoders depending on which route you asked. The detail route (`writeOK` → sonic) honours `Transaction.MarshalJSON`; the list route (`writeProtoListOK` → `protojson.Marshal`) works off protobuf reflection and **ignores `json.Marshaler`**, so the marshaller was silently bypassed:

| Field | List (before) | Detail | `TransactionResponse` declares |
|---|---|---|---|
| `amount` | `{"v0":"12345"}` | `12345` | `type: integer` |
| `timestamp` | `{"data":"1786540255458491"}` | `"2026-08-12T13:10:55.458491Z"` | `format: date-time` |
| `id` | `"7"` (quoted) | `7` | `type: integer` |
| revert link | `revertedByTransaction` | `revertedByTransactionId` | `revertedByTransactionId` |
| `color` | omitted | always present | `required` |

The list's OpenAPI schema was a bare `type: object`, which every JSON object satisfies — so Speakeasy generated `z.object({})` and stripped every key, which is the symptom the frontend reported.

## Scope: the ticket was wrong in both directions

Re-verified at HEAD:

- The **logs list is healthy** — it already used `writeOK`. The ticket's "no clean `MarshalJSON`" premise is false; `Log`, `LogPayload`, `ApplyLedgerLog`, `LedgerLog` and `LedgerLogPayload` all have one.
- The actually-broken log route is **`GET /v3/_/logs/{sequence}`**, and its defect is worse than cosmetic: protojson cannot emit the synthetic `type` discriminator, so the two log routes disagreed for the same type.
- **`GET /v3/_/chapters`** was a third instance — shipping base64 hashes where `Chapter.MarshalJSON` hex-encodes them.
- `GET /v3/_/signing-keys` and `GET /v3/_/events-sinks` are **correct** on protojson and were left alone: their types have no marshaller, so sonic would emit snake_case.

The rule that separates them: protojson is correct for a type with **no** custom `MarshalJSON`, and wrong for a type that has one. Exactly 3 of the 12 protojson call sites carried a marshaller — the 3 broken routes.

## What changed

**Encoders (3 routes → `writeOKChecked`).** Buffered rather than streaming is deliberate: `writeOK` commits the 200 header before encoding, so a mid-encode failure appends an error object to a sent body — and these marshallers marshal a metadata map and can fail.

**Nullability.** `Transaction.MarshalJSON` emitted `"postings":null` / `"metadata":null` (no `omitempty`, nil sources) against non-nullable schema types, and omitted `id` at zero while the schema requires it. Normalised to always-emitted empty collections, mirroring the existing `Account.MarshalJSON`. Policy adopted: **`required` ⟺ always emitted.**

**Schemas.** Transactions-list items now `$ref: TransactionResponse` (+ `required` extended to the four always-emitted fields); new `Chapter` component; three descriptions corrected that claimed "protobuf JSON representation" for sonic-served routes.

**The guard test** (`internal/adapter/http/encoder_contract_test.go`) — the load-bearing piece. Asserts the rule itself: no type implementing `json.Marshaler` may be routed through protojson, and vice versa for the three fixed routes. Plus an import gate, because matching call *shapes* is unsound — `protojson.MarshalOptions{…}.Marshal(m)`, an aliased import, or a variable receiver all evade a shape matcher, and an import cannot be disguised. The gate immediately surfaced a previously-undeclared protojson importer (`handlers_create_ledger.go`, decode-side oneof parsing — legitimate, now declared).

**The documented rule** (`http-api.md`). The page had no serialization section at all; the repo's only stated rule ("camelCase") cannot arbitrate because *both* encoders satisfy it. That absence is the root-cause class. Also corrected `events.md`, which claimed sinks use protojson when they use sonic.

## Why CI never caught this

Three compounding causes, each sufficient alone — which is why the gate passed **72/72 with the leak live**, and passed again with a *correct* schema against the *unfixed* handler:

1. Response schemas were bare `type: object` — any object conforms.
2. No seed data, so lists returned `[]`, and an empty array satisfies any `items` schema.
3. **`DELETE /v3/{ledgerName}` is itself fuzzed**, and `test_api.py`'s `ledgerName` coercion aimed it at the fixture ledger. With `workers=1` operations run in declaration order and this one precedes the transaction endpoints, so the fuzzer deleted the fixture seconds into every run — every later response then validated against an error schema.

Cause 3 was invisible to static reading and to `MAX_EXAMPLES=30` (deterministic, not sampling luck). Fixed by seeding a sacrificial ledger and coercing only ledger-level DELETE to it; nested DELETEs still target the fixture. It is deliberately not exempted from coercion outright — it would then only ever 404, trading one vacuous check for another.

**The gate's first genuinely load-bearing run failed**, on two pre-existing defects outside this ticket's routes: `analyze-accounts` and `analyze-transactions` emit `metadataKeys: null` against a non-nullable array. Same class as the `metadata`/`postings` nulls above, invisible for the same reason. Fixed rather than suppressed — excluding the checks would re-introduce deliberate vacuity, the direction `9bf16816b` already took and had reverted. A sweep of those DTOs found four such fields (`Assets`, `Examples`, both `MetadataKeys`).

## Evidence

| Check | Result |
|---|---|
| `just pre-commit` (root + operator) | 0 issues, nothing regenerated |
| `go test ./...` full suite | clean |
| `golangci-lint` on all touched packages | 0 issues |
| e2e list/detail deep equality | PASS with live data |
| schemathesis, clean schema | **64/64, `Skipped: 0`** |
| schemathesis, `reverted` corrupted to `string` | **1 failure** — the transactions list |
| operator module build + `TestParseActualIndexes` | PASS |
| Guard probe: throwaway `Index.MarshalJSON` | RED, 4 rows, actionable message |
| Guard probe: disguised `MarshalOptions{…}.Marshal` | RED via the import gate |

The RED/GREEN pair is the point: before this branch, corrupting a response schema could not turn the gate red.

e2e body — list item and detail item deeply equal, `amount` a real number:

```
{id:1, metadata:{category:list-shape-test},
 postings:[{amount:12345 (float64), asset:USD, destination:..., source:world}],
 reference:list-shape-ref-001, reverted:true, revertedByTransactionId:2, ...}
```

## Acceptance criteria

AC1 (list shape == detail shape) is met and asserted by unit + e2e tests.

**AC2 as written is not verifiable in this repository** — there is no `.speakeasy/`, no `gen.yaml`, and no SDK job in either workflow, so "the generated SDK types `data[]` as `TransactionResponse`" has no exit condition here. Restated as two locally checkable facts: the affected 200 responses declare concrete `$ref` schemas, and `response_schema_conformance` passes against them with seeded rows. The SDK regeneration needs tracking wherever that pipeline lives.

One consequence to note: `amount` is now **usable but lossy** above 2^53 in JavaScript. protojson rendered the `fixed64` limbs as quoted strings, so the list was previously unusable but lossless. The detail, create, revert and bulk routes were already on the number shape, so this removes an outlier rather than creating exposure — but AC2 should not be read as "the SDK is correct for amounts"; consumers need a BigInt-aware reviver. EN-1469 declined quoting on 2026-07-07 and that decision stands.

## Deliberately out of scope

- **`Log` OpenAPI schema** (~39 recursive oneof variants) — still needs a ticket. Note this PR makes the two log routes emit an identical shape for the first time, which is the precondition for typing them with one component.
- **`LedgerLog` round-trip.** `LedgerLogPayload.MarshalJSON` wraps the payload in its oneof field-name key while `HydrateLog` expects the bare inner type, so rehydration fails for every variant except `OrderSkipped`. Pre-existing, encoder-independent, equally live on the logs-list route. Not fixed here because the encode side is also the **events-sink wire** (Kafka/NATS/ClickHouse/Databricks/HTTP) plus the prepared-query `logData`; the decode side is wire-safe but has no production callers, which is how it survived. Filed as [EN-1790](https://formance-team.atlassian.net/browse/EN-1790).
- **The index/signing/sinks surface** (9 routes, 5 marshaller-less types). Giving those types a `MarshalJSON` is **not** HTTP-local: the CLI prefers a custom marshaller when one exists (`cmd/ledgerctl/cmdutil/output.go:123-165`), so it would silently reshape four CLI commands and break the operator's index reconciler — which parses `ledgerctl indexes list --json` with a struct hard-coding the protojson shape, in a separate Go module that root `go build ./...` never compiles, with tests that feed literal JSON and would stay green. Filed as [EN-1791](https://formance-team.atlassian.net/browse/EN-1791), which leads with a do-NOT-add-MarshalJSON warning and the missing operator contract test.

## Review notes

- `writeProtoOK`/`writeProtoListOK` survive: 8 routes still use them legitimately. Their doc comments lost three false claims (sonic-leaks-snake_case as an unqualified rule, a nonexistent gRPC gateway, byte-identity) and the blanket `MUST` that directed the original author into this bug.
- The guard's allowlist is **per-file**, so an allowlisted file could later add a disguised response-side protojson call. Documented in the test; closing it needs symbol-level analysis, which reopens the unsoundness the import gate replaced.

[EN-1622]: https://formance-team.atlassian.net/browse/EN-1622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EN-1790]: https://formance-team.atlassian.net/browse/EN-1790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/technical/architecture/subsystems/api/http-api.md
+++ b/docs/technical/architecture/subsystems/api/http-api.md
@@ -559,6 +559,25 @@ func (r *RoutedController) Apply(ctx context.Context, requests ...*servicepb.Req
 }
 ```
 
+## Response serialization
+
+Two encoders serve HTTP response bodies, and the choice is **not** a matter of taste:
+
+| Writer | Encoder | Use for |
+|---|---|---|
+| `writeOK` / `writeOKChecked` | sonic (`internal/adapter/json`) | Anything whose type has a custom `MarshalJSON`, and all hand-written DTOs |
+| `writeProtoOK` / `writeProtoListOK` | `protojson` | Proto messages with **no** custom `MarshalJSON` |
+
+**The rule: when a type has a hand-written `MarshalJSON`, that method is the public contract.** Route it through `writeOKChecked`. `protojson` works off protobuf reflection and ignores `json.Marshaler`, so sending such a type through it silently discards the intended shape.
+
+The camelCase convention cannot arbitrate between the two — both encoders satisfy it. The deciding fact is whether a marshaller exists. `internal/adapter/http/encoder_contract_test.go` enforces this in both directions.
+
+Prefer `writeOKChecked` over `writeOK` when the marshaller can fail (e.g. it marshals a metadata map): `writeOK` streams, so a mid-encode failure appends an error object to an already-committed 200.
+
+This was EN-1622: the transactions list, chapters and single-log routes sent marshaller-carrying types through `protojson` and shipped `amount: {"v0":"12345"}`, `timestamp: {"data":"1786540255458491"}`, base64 hashes where the contract is hex, and quoted numeric ids.
+
+**Before adding a `MarshalJSON` to a proto type, check the blast radius.** `cmd/ledgerctl/cmdutil/output.go` also prefers a custom marshaller when one exists, so adding one changes CLI output too — and `misc/operator` parses `ledgerctl indexes list --json` with a struct that hard-codes the protojson shape, in a separate Go module that a root `go build ./...` never compiles. For types that need a clean HTTP shape without moving the CLI, use an HTTP-local response DTO instead.
+
 ## OpenAPI Documentation
 
 The OpenAPI specification is available in `openapi.yml`. It can be used for:
--- a/docs/technical/architecture/subsystems/events-mirror/events.md
+++ b/docs/technical/architecture/subsystems/events-mirror/events.md
@@ -61,7 +61,7 @@ The event carries the full `Log` entry, which contains the typed payload (transa
 
 ### JSON Format
 
-When `format=json` in the events config, the event is serialized as JSON using `protojson`. Property names follow protobuf JSON mapping conventions (camelCase):
+When `format=json` in the events config, the event is serialized as JSON via `internal/adapter/json` (sonic), which honors each proto type's hand-written `MarshalJSON` rather than protobuf reflection — see [Response serialization](../api/http-api.md#response-serialization) for the encoder rule. Property names follow protobuf JSON mapping conventions (camelCase):
 
 ```json
 {
--- a/internal/adapter/http/encoder_contract_test.go
+++ b/internal/adapter/http/encoder_contract_test.go
@@ -2,7 +2,13 @@ package http
 
 import (
 	"encoding/json"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
 	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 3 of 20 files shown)


gfyrag:
Non-blocking follow-up idea for team discussion:

This PR demonstrates that maintaining two response-encoding paths is difficult to keep exhaustive. Since JSON is intended to be the same public representation across the HTTP API, `ledgerctl`, and JSON event sinks, we could converge on a canonical generated Ledger JSON format for protobuf messages.

A possible direction would be a `protoc-gen-ledger-json` plugin integrated into `just generate-proto`:

- Generate `MarshalJSON` (and ideally `UnmarshalJSON`) for JSON-facing proto packages.
- Generate direct encoders from protobuf descriptors rather than merely delegating to `protojson`, so nested types also use the Ledger representation.
- Define the common rules once: protobuf `json_name` / camelCase, enum names, RFC3339 timestamps, decimal `Uint256`, non-null empty collections, and the chosen zero-value presence policy.
- Support annotations for exceptions such as hex-encoded hashes, oneof/discriminator layouts, and a temporary/manual escape hatch for complex types such as logs.
- Once coverage is complete, HTTP, CLI, and event sinks could all use the same ordinary JSON path, removing the protojson/Sonic route split and its allowlist.
- Generated direct encoders would also give us a single place to optimize allocations and reflection costs later.

This should be treated as a follow-up architecture discussion, not as a blocker for this PR. The current change fixes the live contract mismatch and adds the right protection for the implementation we have today.

gfyrag (approved):
Reviewed. This fixes the current serialization mismatch correctly and adds strong contract coverage across unit, E2E, OpenAPI, and Schemathesis. The generated canonical JSON proposal in my comment is a follow-up discussion, not a blocker for this PR.

PR #1703, opened 2026-08-14
refactor(EN-1791): own the HTTP response wire shape with adapter-local DTOs

paul-nicolas:
https://formance-team.atlassian.net/browse/EN-1791

> **Base note.** This work was originally built stacked on `EN-1622-fix-transactions-list-openapi` (#1700), because two files it changes (`internal/adapter/http/encoder_contract_test.go` and the `http-api.md` response-serialization section) did not exist on `release/v3.0` at the time. #1700 has since been squash-merged as `bb0297cce`, so the branch has been **rebased onto `release/v3.0`** and now contains only the 14 EN-1791 commits. All 14 applied with zero conflicts and the full gate was re-run afterwards.

## Summary

Nine HTTP response routes were serialized by `protojson`. They are now served by hand-written DTOs local to `internal/adapter/http`, which own their wire shape. `writeProtoOK` and `writeProtoListOK` are deleted, and all nine responses are typed in `openapi.yml`.

`protojson` is descriptor-driven: it emits quoted `uint64`, renders timestamps as `{"data":<micros>}`, drops meaningful zero values, and ignores `json.Marshaler`. The one remaining `protojson` import in the package is request-side `Unmarshal` in `handlers_create_ledger.go`.

## This fixed live defects, not just cosmetics

- **Metadata indexes were unidentifiable over HTTP.** The response was `{"id":{"metadata":{"key":"color"}}}` — `target` was dropped because `TARGET_TYPE_ACCOUNT` is enum `0`. Callers could not tell an account-metadata index from a transaction-metadata one with the same key.
- **The documented poll signal was unobservable.** Per-entry status returned `{"ledger":"ledger1"}` with `currentVersion` omitted at `0`, which is exactly the "never built" state clients are supposed to poll for.
- **Sink credentials were returned in plaintext.** `saslPassword`, HTTP `secret`, ClickHouse `dsn`, Databricks `token` and OAuth `clientSecret` all reached any caller holding ops-read. Redaction existed only client-side in `cmd/ledgerctl/events/redact.go`. The DTOs declare no field for them, so they cannot be serialized.

The CLI was unaffected throughout because it sets `EmitUnpopulated: true`. That asymmetry is why these survived unnoticed.

## Two further defects found by the new gate

Seeding the schemathesis suite turned it red twice on genuine bugs:

- **A 500 on caller input.** `GET /v3/{ledgerName}/indexes/{canonicalId}/inspect?cursor=<garbage>` returned `INTERNAL_ERROR`; a bare `fmt.Errorf` around a base64 decode fell through to the 500 sanitizer. Now `400 VALIDATION`, fixed at both decode sites.
- **A route violating its own schema.** `MetadataFieldStatus` required a `status` field (`COMPLETE`/`CONVERTING`) that was deleted from the proto in `3c06c63d6`; `openapi.yml` was never updated.

Both were unreachable before this branch, because both routes were among the vacuous ones.

## The gate was passing vacuously

`response_schema_conformance` could not fail on these routes: the list routes returned `[]`, which satisfies any `items:` schema, and the detail routes 404'd on a generated `{canonicalId}`. `tests/schemathesis/run.sh` now seeds a metadata index (deliberately the arm whose rendering was broken), a signing key, and six sinks covering every `SinkConfig` variant and both Databricks credential kinds. `test_api.py` coerces `{canonicalId}`, routing the fuzzed `DELETE` to a separate sacrificial index so it cannot consume the fixture.

Proven to bite by corrupting the spec: retyping `IndexStatus.lag` to `string` and `KafkaSinkConfig.brokers` items to `integer` each fail the suite.

## Known gaps

- **gRPC still returns all five sink credentials in plaintext** to any caller with `ScopeOpsRead`. This branch fixes the HTTP surface only; `api-comparison.md` now records the divergence rather than claiming parity. Filed as **EN-1796** (`https://formance-team.atlassian.net/browse/EN-1796`).
- **`GET /v3/_/indexes/{canonicalId}` and its `/status` sibling remain unvalidated.** They look up `(ledger="", canonical)` and no production path writes such a key, so they 404 whatever is seeded. Documented rather than faked.
- **gRPC `sinkStatuses` ordering is still nondeterministic** (`query.BuildSinkStatuses` ranges a map). The HTTP DTO now sorts both arrays; the shared function was left alone as out of scope.
- **Commits are unsigned.** The SSH signing key is passphrase-locked and non-interactive signing fails. The base branch is likewise unsigned.

## Validation

Run at `c867e07d0` — that is **after** the rebase onto `release/v3.0`, so these results describe the commits actually proposed here. The same gate was run before the rebase at `8081dcaaf` and produced identical numbers.

| Gate | Command | Result |
| --- | --- | --- |
| Pre-commit | `nix develop --command bash -c "just pre-commit"` | exit **0** |
| Unit suite | `nix develop --command bash -c "just test"` (= `go test -race ./... -timeout 20m`) | exit **0** — 81 `ok`, 0 `FAIL` |
| Changed packages | `GOROOT= go test -count=1 ./internal/adapter/http/... ./internal/application/ctrl/... ./internal/query/... ./cmd/ledgerctl/cmdutil/...` | exit **0** — 4/4 `ok` |
| Build | `GOROOT= go build ./...` | exit **0**, no output |
| E2E vet | `GOROOT= go vet -tags e2e ./tests/e2e/business/` | exit **0**, no output |

Detail:

- **`just pre-commit`** ran `go generate` (root + proto), `controller-gen` (operator), the jsonnet dashboards, `go mod tidy` (root + operator), and `golangci-lint` on both modules: **`0 issues.`** each. `git status --porcelain` was empty afterwards — nothing was reformatted or regenerated, so no follow-up commit was needed and no re-run was required. The run emits one pre-existing config-level warning, `The linter 'gomodguard' is deprecated (since v2.12.0) … Replaced by gomodguard_v2`; it is not a finding in a file this branch touches and was deliberately left alone.
- **`just test`** output was captured in full to a file rather than judged from its tail. 115 lines: **81 `ok` packages, 33 `[no test files]`, zero** `FAIL`, `--- FAIL`, `panic:` or `build failed` — `grep -c FAIL` over the whole log returns `0`.
- **Changed packages**, re-run individually with `-count=1`: `internal/adapter/http` `ok`, `internal/application/ctrl` `ok`, `internal/query` `ok`, `cmd/ledgerctl/cmdutil` `ok`.
- **After the rebase**, `just pre-commit` and `just test` were both re-run against the new base: `0 issues.` on both modules, and 81 `ok` / 33 `[no test files]` / 115 lines / zero `FAIL` again. `git status --porcelain` was empty after the pre-commit run, so the rebase onto the newer base required no regeneration.

Not run, and not claimed: `just test-all`, the e2e suite, and the schemathesis suite itself (it needs a live server plus Docker). The schemathesis changes in this branch were exercised during development, including the two deliberate spec-corruption runs described above, but they are not part of this gate.


docs/technical/architecture/subsystems/api/http-api.md
```diff
@@ -599,22 +599,54 @@ func (r *RoutedController) Apply(ctx context.Context, requests ...*servicepb.Req
...
 ## Response serialization
 
-Two encoders serve HTTP response bodies, and the choice is **not** a matter of taste:
+**One encoder serves HTTP response bodies: sonic, through `internal/adapter/json`.** The three writers are `writeOK`, `writeOKChecked` and `writeCreated`, all in `internal/adapter/http/response.go`.
 
-| Writer | Encoder | Use for |
-|---|---|---|
-| `writeOK` / `writeOKChecked` | sonic (`internal/adapter/json`) | Anything whose type has a custom `MarshalJSON`, and all hand-written DTOs |
-| `writeProtoOK` / `writeProtoListOK` | `protojson` | Proto messages with **no** custom `MarshalJSON` |
+**No response body in this package is serialized by `protojson`.** This is the invariant, and it is stated deliberately about the *encoder*, not about the payload types. It does **not** say that no `proto.Message` reaches a response body: `handlers_list_numscript_versions.go` embeds a `*commonpb.Timestamp` in a DTO served by `writeOK`, and that is correct — sonic calls that type's own `MarshalJSON`, which emits RFC3339Nano.
```
paul-nicolas: [P2][blocking] State the protojson invariant at the scope the gate actually enforces

Location: docs/technical/architecture/subsystems/api/http-api.md:604; internal/adapter/http/encoder_contract_test.go:98

Evidence: This line asserts "No response body in this package is serialized by `protojson`", and the section that follows presents quoted `uint64`, `{"data":<micros>}` timestamps and "ignores `json.Marshaler`" as consequently impossible on response bodies. All three ship today on `GET /v3/_/logs/{sequence}`: `handlers_get_log.go:30` passes a `*commonpb.Log` to `writeOKChecked`, sonic invokes `Log.MarshalJSON`, which (a) routes `responseSignature` through `protoFieldJSON` -> `protojson.Marshal` (`internal/proto/commonpb/common.pb.json.go:46` -> `:25`) on **every** variant, and (b) delegates 23 of the 26 `LogPayload` oneof variants to `protojson.Marshal` in its default branch (`common.pb.json.go:70-71`). Executed against the real writer, `LogPayload_SealChapter` renders `{"chapter":{"id":"42","start":{"data":"1787063922136952"},"sealingHash":"3q2+7w=="}}`, where the same `commonpb.Chapter` through its own `MarshalJSON` gives `{"id":42,"start":"2026-08-18T14:38:42.136952Z","sealingHash":"deadbeef"}` — so one of the three types line 617 calls "their contract" is rendered by the wrong marshaller on a shipped route. `auditpb.AuditEntry.MarshalJSON` does the same (`internal/proto/auditpb/audit.pb.json.go:50`) on the audit routes, which line 620 and `response.go:68-72` already describe as protojson rendering — so :604 contradicts :620 inside one document. The encoder hedge ("about the *encoder*, not about the payload types") makes the narrow reading defensible, but `encoder_contract_test.go:98-99` repeats the claim unhedged as "the primary enforcement", while its scan is `os.ReadDir(".")` (`:123`) and cannot see either call site. Nothing else enforces it: `.golangci.yaml` has no protojson rule, and no test covers `internal/proto/**`.

Impact: A contributor reads newly added authoritative guidance saying the `{v0}`/`{data}` regression class is eliminated in this package, backed by a green gate over an unscanned surface, while the class is live on `GET /v3/_/logs/{sequence}`.

Resolution: :604 and the :98 comment state the real scope — no protojson **call site in this package**, and a payload type's own `MarshalJSON` may still delegate to protojson — naming `Log.MarshalJSON`'s `responseSignature`, `LogPayload.MarshalJSON`'s default branch and `AuditEntry.MarshalJSON` with the routes they serve; or the gate is extended to the payload types the package's routes write.

internal/adapter/http/dto_indexes.go
```diff
@@ -0,0 +1,219 @@
...
+	// No omitempty: the target is always emitted so the (target, key) pair is
+	// complete. Value 0 is TARGET_TYPE_ACCOUNT, but omitempty would not have
+	// dropped it either — the stringer never yields "" — so it is a no-op here.
+	Target string `json:"target"`
+	// No omitempty: a metadata index is identified by its key, so the key is
+	// always emitted — an absent one would make the entry unidentifiable.
+	Key string `json:"key"`
+}
+
+type indexDTO struct {
```
paul-nicolas, line 54: [P2][blocking] Update indexes.md, which still documents these seven routes as protobuf-JSON

Location: docs/technical/architecture/subsystems/indexer/indexes.md:141 (finding anchored here because that file is not in the diff; the contract it describes is the one these DTOs now own)

Evidence: The HTTP row of the indexer subsystem doc enumerates exactly the routes this PR converted — `GET /v3/{ledger}/indexes`, `.../indexes/{canonicalId}`, `.../status`, `GET /v3/_/indexes`, `GET /v3/_/indexes/status`, `GET /v3/_/indexes/{canonicalId}`, `.../status` — and ends "All responses serialize the protobuf message in protobuf-JSON camelCase, wrapped in the `{data:…}` envelope." None of them serialize a protobuf message any more: they return `indexDTO` / `indexEntryDTO` / `indexStatusDTO`, and the same PR's `TestNoProtojsonMarshalInResponsePath` now forbids protojson in the package. The shapes differ observably — unquoted `uint64` (dto_indexes.go:75), RFC3339Nano instead of `{"data":<micros>}` (dto_indexes.go:104-112), emitted zero values, and a new `canonicalId` with no proto counterpart (dto_indexes.go:58).

Impact: The authoritative indexer doc describes the pre-EN-1791 wire format for seven routes. A reader deriving a client from it, or a reviewer checking a future change against it, gets the wrong encoder and the wrong shape. `CLAUDE.md` requires the matching subsystem documentation to be updated when behavior changes, and this PR updated `http-api.md` and `api-comparison.md` but not this one.

Resolution: That sentence points at the DTOs (or at `http-api.md#response-serialization`) instead of claiming protobuf-JSON.

docs/technical/architecture/subsystems/api/http-api.md
```diff
@@ -599,22 +599,54 @@ func (r *RoutedController) Apply(ctx context.Context, requests ...*servicepb.Req
...
+### Choosing a writer
+
+`writeOK` streams the body, so it writes the 200 header before the encoder finishes. `writeOKChecked` marshals to a buffer first and writes no header until the marshal succeeds, so a failure becomes a clean 500 through `handleError` instead of an error object appended to an already-committed 200.
+
+Use `writeOKChecked` when the marshaller can genuinely fail:
+
+- the audit DTOs, which render chain-bound submessages whose marshalling can fail on invalid UTF-8 (invariant #7 — a truncated record that still looks valid must not be served);
+- the EN-1622 types, whose `MarshalJSON` marshals a metadata map.
+
+The remaining routes keep `writeOK`: plain struct marshalling cannot fail, so buffering would only add an allocation. The reasoning is also on the doc comment of `writeOKChecked`.
```
paul-nicolas: [P2][non-blocking] Reconcile the writer-selection rule with the nine routes this PR ships

Location: docs/technical/architecture/subsystems/api/http-api.md:628; internal/adapter/http/response.go:82-83

Evidence: The new section says to use `writeOKChecked` "when the marshaller can genuinely fail", lists exactly two categories (the audit DTOs and the EN-1622 types), then states here that "The remaining routes keep `writeOK`: plain struct marshalling cannot fail". All nine routes this PR converts are plain-struct DTOs in neither category, and all nine call `writeOKChecked` — verified handler by handler: `handlers_get_index.go:40`, `handlers_get_bucket_index.go:37`, `handlers_list_ledger_indexes.go:32`, `handlers_list_bucket_indexes.go:57`, `handlers_get_index_status.go:26`, `handlers_get_index_entry_status.go:41`, `handlers_get_bucket_index_entry_status.go:34`, `handlers_list_signing_keys.go:66`, `handlers_get_events_sinks.go:22`. The `writeOKChecked` doc comment at `response.go:82-83` ("The remaining list/get handlers keep writeOK: their struct marshaling cannot fail") is falsified the same way and was left untouched by the diff.

Impact: Contributor guidance only, and the inconsistency runs in the safe direction (buffered). The concrete risk is a contributor "restoring consistency" by moving these routes back to the streaming `writeOK`, or picking `writeOK` for the next DTO route on the strength of this paragraph — while `dto_indexes.go` is nominated in the same document as "the reference example".

Resolution: Either state that hand-written response DTOs use `writeOKChecked` by default (and why), or record the nine EN-1791 routes as a deliberate exception. Update the `writeOKChecked` doc comment in the same pass.

tests/schemathesis/test_api.py
```diff
@@ -92,6 +102,33 @@ def before_call(context, case):
...
+    # those look the entry up under an empty ledger, and no production path ever
+    # writes a registry entry with an empty ledger (see the comment on
+    # handleListBucketIndexes). They stay 404 for a structural reason, not a
+    # seeding gap, and their 200 schemas therefore stay unvalidated until
+    # bucket-scoped indexes exist.
+    if case.path_parameters and "canonicalId" in case.path_parameters:
+        if case.method == "DELETE":
+            case.path_parameters["canonicalId"] = DELETE_INDEX_SACRIFICE_CANONICAL
+        else:
+            case.path_parameters["canonicalId"] = FIXTURE_INDEX_CANONICAL
```
paul-nicolas, line 142: [P2][non-blocking] Coerce the one uncovered destructive route, and assert the fixtures survived the run

Location: tests/schemathesis/test_api.py:96-130 (the coercions); tests/schemathesis/run.sh:327-331 (end of file)

Evidence: The fixtures are verified only BEFORE the fuzz run — `seed()`/`ctl()` status checks and the `currentVersion > 0` poll at `run.sh:184-198`. `run.sh` ends at line 331 with the `test_api.py` pipeline; nothing re-checks the fixtures afterwards, and `test_api.py:348-410` exits on `has_failures`/`has_errors` alone, never asserting that any route was observed returning 200. Non-vacuity therefore rests entirely on `before_call`'s coercion being complete. It covers two of the three destructive paths reachable from the fuzzed surface (`:96-98` for `DELETE /v3/{ledgerName}`, `:126-130` for `DELETE .../indexes/{canonicalId}`). The third is uncovered: `DELETE /v3/{ledgerName}/metadata-schema/{targetType}/{key}` (openapi.yml:1487) is fuzzed, its `ledgerName` is coerced to the fixture ledger by the `elif` at `:100-103`, and `processRemoveMetadataFieldType` cascade-drops the attached index (`internal/domain/processing/processor_metadata_schema.go:127-147`) — which would kill `metadata:TARGET_TYPE_ACCOUNT:color`. It is unreachable today (`key` is a bare `type: string` with no `example:` and no link feeding it; `derandomize=True` at `:344` plus green CI proves the fixed input set misses it) but becomes reachable the moment anyone adds an `example:` to that parameter or picks a guessable fixture key.

Impact: The comments at `test_api.py:87-95` and `run.sh:161-167` record that this failure already happened twice on this branch — a fuzzed DELETE destroyed a fixture and every later request validated against an error schema while the suite reported green. The gate has no mechanism to detect a third recurrence, which is the exact failure mode this PR exists to eliminate.

Resolution: (a) extend the coercion so metadata-schema deletes cannot target the fixture's `account`/`color` field; (b) after the `test_api.py` invocation, re-assert positively and fail hard — e.g. `GET /v3/test-ledger/indexes/metadata:TARGET_TYPE_ACCOUNT:color` still returns 200 with `"target":"TARGET_TYPE_ACCOUNT"`.

Not filed: the events-sinks credential check is *not* vacuous-on-empty (the poll at `run.sh:270-283` hard-exits if no sink reports an `"error"`), and the sink/signing-key fixtures need no post-run re-assertion — `handler.go:151-152` exposes GET only, so the fuzzer cannot touch them.

internal/query/executor.go
```diff
@@ -188,8 +188,10 @@ func executeList(
 		var err error
 
 		afterEntity, err = decodeCursor(req.GetCursor())
+		// Caller input, not a server fault — same reasoning as the inspect
+		// route in ctrl.InspectIndex. A bare error here maps to 500 / Internal.
 		if err != nil {
-			return nil, fmt.Errorf("invalid cursor: %w", err)
+			return nil, domain.NewValidationSentinel(fmt.Sprintf("invalid cursor: %v", err))
```
paul-nicolas, line 194: [P2][non-blocking] Pin the malformed-cursor fix at this second decode site

Location: internal/query/executor.go:194 (executeList)

Evidence: The PR fixes a real defect — a bad base64/JSON cursor fell through to HTTP 500 / `codes.Internal` (`internal/adapter/http/error_handler.go:118`) — at two decode sites: `ctrl.InspectIndex` (`controller_default.go:1093`) and here. Only the first is pinned: `tests/e2e/business/inspect_index_test.go:157` asserts `codes.InvalidArgument`. Nothing exercises this site at any level. `grep -rn "invalid cursor\|decodeCursor" --include='*_test.go' .` returns nothing; the only prepared-query pagination spec (`tests/e2e/business/prepared_queries_test.go:225-260`) feeds back a valid `cursor.Next`; and `handlers_execute_prepared_query_test.go:120` does send `?cursor=abc` but against a `MockBackend`, so it never reaches the decode. The schemathesis gate cannot cover it either: `run.sh` seeds no prepared query, `createPreparedQuery` returns 204 with no body so `openapi.yml` exposes no link to the execute route, and `before_call` never coerces `queryName` — every fuzzed call 404s at `internal/query/executor.go:78-83` before the decode. The sentinel -> 400 / `"VALIDATION"` hop itself is already covered (`internal/adapter/http/error_handler_test.go:133-137`), so what is unverified is only that this site returns a validation `Describable` at all.

Reachable from: HTTP `POST /v3/{ledgerName}/prepared-queries/{queryName}/execute` (`handler.go:208`; cursor from the JSON body or `?cursor=`, unvalidated at `handlers_execute_prepared_query.go:63-65,:87`) and gRPC `BucketService.ExecutePreparedQuery` (`server_bucket.go:1355`). Both need an existing prepared query and `QUERY_MODE_LIST`.

Impact: Reverting these two lines to the `fmt.Errorf` they replaced restores the 500 on caller input at the prepared-query execution route with no failing test.

Resolution: One assertion driving a malformed cursor through `executeList` — cheapest as an `It` in `tests/e2e/business/prepared_queries_test.go` (a prepared query and client already exist there) asserting `codes.InvalidArgument`, or an executor-level test following the `internal/query/executor_ledger_not_found_test.go` precedent.

internal/adapter/http/dto_sinks.go
```diff
@@ -0,0 +1,268 @@
...
+
+	dto := &sinkConfigDTO{
+		Name:         cfg.GetName(),
+		Format:       cfg.GetFormat(),
+		BatchSize:    cfg.GetBatchSize(),
+		BatchDelayMs: cfg.GetBatchDelayMs(),
+		EventTypes:   eventTypes,
+	}
+
+	switch t := cfg.GetType().(type) {
```
paul-nicolas, line 185: [P2][non-blocking] Add an exhaustiveness guard for the hand-dispatched SinkConfig oneof

Location: internal/adapter/http/dto_sinks.go:168 (newSinkConfigDTO)

Evidence: The PR replaces `protojson` — descriptor-driven, so it would render a new oneof variant automatically — with a hand-written type switch that has no `default` branch and nothing tying it to the descriptor. Adding a sixth arm to `SinkConfig.type` compiles cleanly, passes `dto_sinks_test.go` (per-variant table rows, not a descriptor enumeration), and passes the schemathesis gate, because `openapi.yml:5884-5895` pre-blesses the output: the description says "all five are absent when the stored configuration names no kind" and `required` lists only `name`/`format`/`batchSize`/`batchDelayMs`/`eventTypes`, with no `minProperties` or `oneOf`. `sinkConfigDTO` (`:83-97`) carries no discriminator, so the element becomes `{"name":"pubsub","format":"json","batchSize":0,"batchDelayMs":0,"eventTypes":[]}` — an ops read-out reporting a configured sink as having no kind, indistinguishable from the genuinely kind-less state. No mechanical gate covers it: `exhaustive` is opt-in (`.golangci.yaml:100-106`) and every tagged site in the tree is an enum switch, and `scripts/check-repo-invariants.go` does not look at `dto_*.go`. The repository already has this exact gate for three other oneofs — `internal/application/admission/wrapper_exhaustiveness_test.go:21` walks `LedgerScopedOrder.payload` via `Oneofs()` and states the pattern ("this test is the compile-time gate; the default is the runtime gate"), plus `internal/adapter/auth/request_scope_exhaustiveness_test.go`, which `http-api.md:59-60` (edited by this PR) cites by name. Invariant 7 reaches this surface: the `http-api.md` hunk added here invokes "invariant #7" for the audit response DTOs.

Narrower than it first looks: the sibling switches degrade with an in-band marker rather than silently. `newIndexIDDTO` yields `{}` but `canonicalId` still emits `"unknown"` (`internal/domain/indexes/id.go:147`), documented at `dto_indexes.go:28-29`; `newDatabricksSinkDTO` yields `authMethod:""`, enumerated and documented at `openapi.yml:5868-5878`. Only the `SinkConfig` case produces an unmarked object.

Impact: Forward-looking, not a live defect. A sink kind added without touching this file is reported by `GET /_/events-sinks` as a kindless sink, with no failing test and no schema violation. The adjacent silent `continue` skips at `:240-245` were reasoned about in writing; this arm was not.

Resolution: A descriptor-driven test enumerating `SinkConfig.type` (and, for symmetry, `DatabricksSinkConfig.auth` and `IndexID.kind`) via protoreflect `Oneofs()` and asserting each field has a DTO branch, so adding a variant without one fails.

internal/adapter/http/handlers_list_signing_keys.go
```diff
@@ -1,9 +1,48 @@
...
+// would also reshape CLI output, and misc/operator parses that in a separate Go
+// module a root `go build ./...` never compiles. See dto_indexes.go.
+type signingKeyDTO struct {
+	KeyID     string `json:"keyId"`
+	PublicKey string `json:"publicKey"`
+	// No omitempty: "" means this key is a root key, which differs from absent.
+	ParentKeyID string `json:"parentKeyId"`
+}
+
+func newSigningKeyDTOList(src []*commonpb.SigningKey) []signingKeyDTO {
```
paul-nicolas, line 27: [P2][non-blocking] Sort the signing-key array, as this same PR does for sinks

Location: internal/adapter/http/handlers_list_signing_keys.go:25 (newSigningKeyDTOList)

Evidence: The upstream source ranges a Go map. `query.ReadSigningKeys` returns `map[string]SigningKeyEntry` (`internal/query/signing.go:43`) and `ReadSigningKeysCursor` builds its slice with `for keyID, entry := range keys` (`internal/query/signing.go:112`), so element order differs per call. `newSigningKeyDTOList` appends in that order and does not sort — whereas this same PR added `slices.SortFunc` for exactly this class of problem on the sibling route (`dto_sinks.go:260-266`) and documented it in `api-comparison.md` ("both arrays sorted by name"). The `openapi.yml` description for `ListSigningKeysResponse.data` makes no ordering claim, unlike the sinks one which states "sorted by sinkName".

Impact: `GET /v3/_/signing-keys` returns the same set in a different order on successive calls, so clients that diff or display the list see spurious churn. Not a schema-conformance failure, and it will not flake CI either — `run.sh` seeds exactly one key, so it stays invisible there.

Resolution: `newSigningKeyDTOList` sorts by `keyId` (or `ReadSigningKeysCursor` emits in key order), and `openapi.yml` states the ordering the way the sinks schema does.

paul-nicolas (commented):

Two independent reviewers ran over the diff at `c867e07d0`; every finding below then went through an adversarial verification pass whose instruction was to refute it against the working tree. Three findings were rejected there and are not posted.

The conversion itself is faithful. Field-by-field against the source protos, no field is dropped, renamed or mistyped; `openapi.yml` matches what the DTOs marshal property for property, including every `required` entry and every enum member (all zero-valued members included); the nil paths are guarded and the one observable nil shape is unreachable through the real controller; the credential omission is backed by a structural reflection gate plus three negative assertions rather than a comment; and both regressions the PR claims to fix are pinned by wire-level tests. What is posted is documentation accuracy, gate self-protection, one test-placement gap and one ordering inconsistency.

**Blocking (2)** — both are authoritative documentation falsified by this change, each a one-sentence fix:
- `http-api.md:604` + `encoder_contract_test.go:98` assert a protojson invariant the tree contradicts, and the document contradicts itself at :620.
- `indexes.md:141` still documents the seven converted index routes as protobuf-JSON.

**Non-blocking (5)**: the writer-selection rule vs the nine shipped routes; the schemathesis gate's missing post-run non-vacuity assertion plus one uncoerced destructive route; the malformed-cursor fix unpinned at its second decode site; no exhaustiveness guard on the hand-dispatched `SinkConfig` oneof; the signing-key array left unsorted while the sibling sinks array is sorted.

### Rejected in verification (not posted)

- **`additionalProperties: false` on the new schemas** — `openapi.yml` uses it in zero places today, so this asks for a new repository-wide convention; and the leak-regression risk it cited is already covered by three explicit negative assertions.
- **Fail loudly instead of skipping nil list elements** — the author reasoned about each skip in writing, degrading a read-only ops list beats a 500, and `handlers_bulk.go` already uses the pattern.
- **Bucket-scoped index routes still unvalidated** — explicitly disclosed in the PR body, `run.sh` and `test_api.py`, and the disclosed count (two routes) is exactly right.

### Noted, deliberately not filed as a finding on this PR

`GET /v3/_/logs/{sequence}` returns all five sink credentials in plaintext at the same `ledger:OpsRead` scope as `/v3/_/events-sinks`. Confirmed by executing the real writer against the real payload: `{"data":{"sequence":7,"payload":{"addedEventsSink":{"config":{"name":"kafka","kafka":{...,"saslPassword":"SUPERSECRETVALUE"}}}}}}`. `AddedEventsSinkLog.config` is a full `SinkConfig` (`misc/proto/common.proto:388-390`), stored unredacted, and `LogPayload.MarshalJSON`'s default branch renders it via protojson.

This is **not** a defect of this PR: the path is entirely outside the diff, this PR strictly *reduces* exposure by closing `/events-sinks`, and the PR's claims are route-scoped and accurate on inspection — `dto_sinks.go:26` says "**this route**", `api-comparison.md:719` says "**on this route**", and the log route's `openapi.yml` 200 body is a bare `type: object` with no `$ref` to `SinkConfig`. It is raised only because it belongs on a ticket beside EN-1796, and because naming it in `dto_sinks.go` would cost one line.

### P3 (not posted inline)

- `docs/technical/architecture/subsystems/read-path/typed-metadata.md:197` — the new sentence ("the `CONVERTING`/`COMPLETE` status field went away with the value converter") is the accurate half, but :172-193 and :305 still document that lifecycle as current and name `internal/infra/state/metadata_converter.go`, which does not exist at this HEAD. The staleness is pre-existing; this edit makes the file self-contradicting rather than resolving it.
- `docs/technical/architecture/subsystems/api/http-api.md:464` — the index-detail field list is snake_case and omits `canonicalId`, the one field added so a list item can address its own detail route.

---

```text
DECISION: REQUEST CHANGES
Head reviewed: c867e07d0a5188742c5a9708d893822524ceb346
Blocking findings: 2
Non-blocking findings: 5 (plus 2 P3 not posted inline)
Validation considered: CI green on this head (Build, Tests, Tests-E2E, Tests-Schemathesis, Tests-Model, Tests-Scenarios, Dirty, Coverage); internal/adapter/http and cmd/ledgerctl/cmdutil unit tests re-run locally; bash -n on run.sh and py_compile on test_api.py; field-by-field DTO/openapi.yml comparison; go test -overlay probe of the log-route response body
Residual risk: LOW
```

paul-nicolas (commented):
## Multi-model re-review (Claude + Codex) — head unchanged since the last review

**`c867e07d0` is the same head I reviewed on 2026-08-18.** No fix commits have landed, so this run is not a re-review of new work: all seven previously posted findings were re-verified against the current tree and **all seven remain still-valid and unaddressed** (0 fixed, 0 outdated). They are not re-posted — the existing inline comments still stand at their original lines, including the two blocking documentation items (`http-api.md:604` and the `indexes.md:141` protobuf-JSON claim).

The value of this pass was therefore hunting what the first one missed. Four Claude reviewers (DTO wire-shape semantics, error handling and cross-transport parity, the test/validation surface, documentation drift) and Codex ran independently, each explicitly barred from re-filing the seven known findings and the earlier rejected set. That produced 19 new candidates; 6 met the posting threshold and entered adversarial verification, where each was handed to a fresh agent instructed to refute it.

**Two survive as posted findings.** Both are `openapi.yml` accuracy on the events-sinks route, and neither is blocking. The conversion itself continues to hold up: no new correctness defect was found in the DTOs, the error-classification change was confirmed correct on both transports, and the `MetadataFieldStatus` schema fix was confirmed complete.

### Verification results (6 entered, 2 posted, 4 removed)

| Candidate | Verdict |
|---|---|
| Six sink fields emit a default-sentinel openapi describes as the effective value | **posted** — adjust: five -> six fields; Codex's *blocking* rejected |
| `SinkConfig` "every credential-bearing setting is omitted" over-claims | **posted** — adjust: the NATS half holds strongly, the HTTP half is weaker; the `api-comparison.md` half of the charge was wrong and dropped |
| `uint64` precision loss above 2^53 (Codex, P2 **blocking**, two findings) | **rejected** |
| schemathesis fixture literals duplicated between `run.sh` and `test_api.py` | **rejected** — same root cause as an already-posted finding |
| `canonicalId` not percent-encoded for its own detail route | **downgraded to P3** |
| 400 body echoes Go-internal cursor decode text | **downgraded to P3** |
| 33 of 57 new `json:` tags have no wire-level assertion | **downgraded to P3** |

### Why the two rejections matter

**Codex's only blocking findings were the `uint64` precision pair, and they do not hold.** Codex argued `cursor`, both sequences, `lag` and `indexFileSize` became bare JSON numbers that cannot round-trip above 2^53, and asked for decimal strings. Traced to producers, all six are one `+1` counter over committed log entries, a guarded delta of two of those, or an `os.Stat` byte count — reaching 2^53 needs 9.0e15 committed logs or a 9 PB read-index file. The `lag` underflow that would have made Codex accidentally right does not exist: one assignment, behind `if lastLog > lastIndexed` (`controller_default.go:1222-1225`). Against that, the repo convention is unambiguous — 37 `format: uint64` integer response fields, **zero** string-typed sequences, and the identical log sequence already emitted bare by `Log.MarshalJSON` on the shipped `GET /v3/_/logs/{sequence}`. Adopting the fix would make one client see the same sequence as `12345` on the log route and `"12345"` here. The wire type did change from string to number, but `openapi.yml` is updated in the same PR and `CLAUDE.md` is explicit that v3 carries no backward-compatibility burden.

**The fixture-duplication finding was rejected as a duplicate**, not as false — every factual claim checked out (no env-var binding; `run.sh` passes only `--base-url`, `--max-examples`, `--workers`), but the root cause is the already-posted gate finding, and this contract keeps one root cause per finding.

### P3 (not posted inline)

- `internal/adapter/http/dto_indexes.go:56` — `canonicalId` is advertised as usable directly as `{canonicalId}`, but a namespaced key like `formance.com/reviewed` needs `%2F`/`%3A` first (chi refuses a param span containing `/`). Downgraded: the same promise already exists unchanged at `openapi.yml:2630-2633`, and the "emit it percent-encoded" half of the fix would break `handleCreateIndex`, which calls `ParseCanonical` with no unescaping. Documentation-only fix.
- `internal/adapter/http/dto_indexes_test.go` — never calls `json.Marshal`; every row compares Go struct fields. Downgraded because the finding's rationale was inverted: 19 of the 32 unasserted tags lack `omitempty`, are `required`, and *are* caught by `Tests-Schemathesis`. The 13 genuinely unguarded ones are all `omitempty` (`txBuiltin`, `logBuiltin`, `accountBuiltin`, `createdAt`, `index`, `nats`, `clickhouse`, `kafka`, `http`, `databricks`, `oauthM2m`, `occurredAt`, `error`). The five variant keys are worth one wire assertion each — a typo on `kafka` silently un-validates the whole nested `required` schema beneath it.
- `internal/query/executor.go:194` — the 400 body now carries `invalid cursor: invalid cursor data: json: cannot unmarshal number into Go struct field cursorData.after of type []uint8` (reproduced on Go 1.26.5), where the old path sanitized it away. Downgraded: `handlers_execute_prepared_query.go:45` already returns raw `encoding/json` decoder errors in 400s on this same handler, and the `after` tag is visible to anyone who base64-decodes a legitimate cursor.
- `internal/adapter/http/dto_indexes_test.go:42` — `TestNewIndexIDDTO_AllOneofVariants`' comment calls the table "the only exhaustiveness guard", but it is four hand-written rows; a fifth `IndexID.kind` member passes everything and renders `"id":{}`. The misleading claim is the defect (the missing guard itself is already-posted finding 6).
- `internal/application/ctrl/controller_default.go:1093` and `internal/query/executor.go:194` — `NewValidationSentinel` is documented for package-init use only (the pointer is the `errors.Is` identity) but is called per-request with a dynamic message, so no `errors.Is` can ever match; `%v` also drops `base64.CorruptInputError`/the json error from the chain.
- `internal/adapter/grpc/server_bucket.go:1145` — the comment "gRPC and HTTP return identical data from one snapshot (EN-1472)" is now false in two ways (five omitted credentials, plus HTTP sorts both arrays). The HTTP-side comment and `api-comparison.md` were updated; this one was not, and it is what the next person adding a field will read.
- `internal/adapter/http/error_handler.go` — the `400 VALIDATION` outcome holds only when the serving node reads locally; on a forwarded read (`ConsistencyLeader`, or the `ErrNotLeader` fallback) the raw gRPC status fails the `Describable` branch and yields `400 INVALID_REQUEST`. Pre-existing and systemic, but the PR is what makes the two paths disagree — before it, both returned 500.
- `tests/schemathesis/test_api.py:377` — connection-reset/broken-pipe errors are counted as transient unconditionally, and the summary subtracts them, so a server that dies mid-run can print `RESULT: ALL CHECKS PASSED` and exit 0. Pre-existing; load-bearing now that the nine routes actually execute handler code.
- `docs/technical/architecture/subsystems/api/http-api.md:600` — "One encoder serves HTTP response bodies ... The three writers are ..." is falsified by `handlers_info.go:27`, which writes `GET /_info` with stdlib `encoding/json` and no `BaseResponse`. The pre-PR text omitted it too, but this PR promotes it to a categorical count.
- `docs/technical/architecture/subsystems/events-mirror/events.md:392` — records neither the new HTTP credential-omission invariant nor the gRPC exposure EN-1796 tracks; an auditor asking what `ledger:OpsRead` exposes lands here and gets neither half.
- `tests/schemathesis/run.sh:217` — "fifteen component schemas hang off it" is nine (ten with the wrapper). The count is the stated justification for the seeding effort.

---

```text
DECISION: REQUEST CHANGES
Head reviewed: c867e07d0a5188742c5a9708d893822524ceb346 (unchanged since the 2026-08-18 review)
Blocking findings: 2 (both carried over unaddressed from the previous review; no new blocker found)
Non-blocking findings: 7 carried over + 2 new posted (plus 11 P3 not posted inline)
Previous findings: 0 fixed / 7 still valid / 0 outdated
Validation considered: CI green on this head (Build, Tests, Tests-E2E, Tests-Schemathesis, Tests-Model, Tests-Scenarios, Dirty, Coverage); 4 independent Claude reviewers + Codex; 6 candidates through adversarial verification (2 upheld with corrections, 3 downgraded, 4 rejected incl. both Codex blockers)
Residual risk: LOW
```

Note: the PR is still marked **draft**, so CodeRabbit skipped it.

internal/adapter/http/dto_sinks.go
```diff
@@ -0,0 +1,268 @@
...
+	Name string `json:"name"`
+	// Exactly one of these is non-nil (the SinkConfig.type oneof).
+	Nats       *natsSinkDTO       `json:"nats,omitempty"`
+	Clickhouse *clickhouseSinkDTO `json:"clickhouse,omitempty"`
+	Kafka      *kafkaSinkDTO      `json:"kafka,omitempty"`
+	Http       *httpSinkDTO       `json:"http,omitempty"`
+	Databricks *databricksSinkDTO `json:"databricks,omitempty"`
+	Format     string             `json:"format"`
+	BatchSize  int32              `json:"batchSize"`
+	// int64 in the proto, not uint64. Emitted unquoted.
```
paul-nicolas, line 110: [P2][non-blocking] Six sink fields now emit a sentinel that means "server default", while `openapi.yml` describes each as the effective value

Raised independently by both reviewers this round, which is why it is the one code-adjacent item posted.

Location: `internal/adapter/http/dto_sinks.go:91-94` (`sinkConfigDTO.Format`/`.BatchSize`/`.BatchDelayMs`), `:73-74` (`databricksSinkDTO.Table`/`.Port`), `:41` (`clickhouseSinkDTO.Table`); `openapi.yml:5863, 5867, 5918, 5923, 5927` — **six** fields, not five.

Evidence: the proto documents these as sentinel-for-default (`misc/proto/common.proto:566-568, 591, 620-621`) and nothing defaults them at write time — `processAddEventsSink` stores the config verbatim (`internal/domain/processing/processor_events_sink.go:29-35`); resolution is per-emitter-start (`internal/application/events/manager.go:176-186`, `emitter.go:47-51`, `sink_databricks.go:116-118` and `:165-167`, `sink_clickhouse.go:133-135`). The deleted handler used the bare `protojson.Marshal` with no `MarshalOptions`, so these were previously **omitted**; they are now stated as a wrong value.

Reachability differs sharply, and only two matter in practice:

- **`batchSize`, `batchDelayMs` — hit on the ordinary path.** `ledgerctl events add-sink` defaults both flags to `0` (`cmd/ledgerctl/events/add_sink.go:99-100`), so a normal sink stores `0` and the response now says `"batchSize":0,"batchDelayMs":0` for a sink that is actually batching 64 events with a 10 ms flush.
- **`format`, databricks `table`/`port`, clickhouse `table` — CLI-unreachable.** Those flags default to `"json"` / `"ledger_events"` / `443` (`add_sink.go:78, 95-97`), so the sentinel only arrives from a client calling `AddEventsSink` over gRPC directly.

Impact: an operator reading `GET /v3/_/events-sinks` to verify a sink sees `batchSize: 0, batchDelayMs: 0` against descriptions that read "Maximum number of events per published batch." / "Maximum delay in milliseconds before a partial batch is flushed." (`openapi.yml:5923, 5927`). No field is exempted in the schema text. This is the inverse of the zero-emission improvement the PR is built around — and note `dto_sinks.go:51` and `:106` already reason explicitly about why `tls:false` and `cursor:0` *are* real values, then skip exactly the six fields where zero is not.

Not blocking: no code is wrong, the DTO faithfully reports stored configuration, `0`/`""` are never legitimate configured values so no information is destroyed, emitting them is the right direction (they are in the schema's `required` set, which the old omitting body violated), and the surface is read-only ops with no audit/FSM/persistence exposure.

Resolution: doc-only — append the sentinel to each of the six descriptions, e.g. `batchSize` -> "Maximum number of events per published batch. `0` means the server default (64)." This file already uses that exact wording twice: `openapi.yml:449` and `:4396` ("0 = server default 100, capped by ..."), the latter on `MirrorSourceConfig`, which is referenced from a response schema at `:3124`. So documenting a response-side sentinel is the established convention here, not reviewer taste.

(Codex rated this blocking; rejected on verification — nothing in the code accepts incorrect behavior, and the fix touches no Go, no test and no schema.)

openapi.yml
```diff
@@ -5583,3 +5507,548 @@ components:
...
+        oauthM2m:
+          $ref: '#/components/schemas/DatabricksOAuthM2M'
+
+    SinkConfig:
+      type: object
+      description: >
+        An event sink configuration. At most one of `nats`, `clickhouse`,
+        `kafka`, `http` and `databricks` is present, naming the sink kind; all
+        five are absent when the stored configuration names no kind. Every
+        credential-bearing setting is deliberately omitted — see the individual
```
paul-nicolas: [P2][blocking] The new completeness claim on `SinkConfig` contradicts the file's own definition of "credential-bearing"

**Edit:** re-tagged from non-blocking to blocking for consistency with the two blocking findings already on this PR. All three are the same class under `docs/technical/contributing/ai-review.md` — "authoritative guidance materially false" — and this one is a security guarantee. Nothing else in the finding changed. (Note that contract post-dates this PR's base `c0c5177ab`; it landed in #1714.)

Location: `openapi.yml:5889-5890` (added by this PR)

Evidence: this line states "Every credential-bearing setting is deliberately omitted — see the individual sink schemas." Twelve lines earlier, `openapi.yml:5751-5752` justifies dropping the ClickHouse DSN because "it embeds credentials (`clickhouse://user:pass@host/db`)". By that definition `NatsSinkConfig.url` is credential-bearing too, and it is returned verbatim:

- `misc/proto/common.proto:584-587` gives `NatsSinkConfig` only `url` and `topic` — there is no credential field.
- `internal/application/events/sink_nats.go:46` authenticates solely via `nats.Connect(cfg.URL)`, and nats.go v1.49.0 (`nats.go:2787-2789`) takes user/token straight from the URL userinfo. So `nats://user:pass@host:4222` is the **only** way to authenticate a NATS sink.
- `internal/adapter/http/dto_sinks.go:171` copies the URL verbatim into the response, and `cmd/ledgerctl/events/redact.go:42-58` has no `SinkConfig_Nats` case at all, so the CLI does not mask it either.

The "see the individual sink schemas" pointer does not scope the claim: `NatsSinkConfig` (`openapi.yml:5731-5746`) says only "NATS JetStream sink settings" and `url` is "NATS server URL."; neither carries a userinfo caveat.

`HttpSinkConfig.endpoint` has the same shape but is materially weaker: `secret` is an HMAC message-signing key (`internal/application/events/sink_http.go:88-91`), i.e. it authenticates the payload to the receiver rather than the client to the endpoint, so basic-auth userinfo in `endpoint` does reach the wire (`net/http` turns `URL.User` into an `Authorization` header) — but the signed-payload pattern is the primary documented mechanism, so this is a secondary pattern, not the only channel. Only NATS has no alternative.

No admission-side validation rejects or strips userinfo: there is no `url.Parse` anywhere in `internal/application/events/`, and `cmd/ledgerctl/events/add_sink.go:225` checks only non-emptiness.

Impact: the openapi spec is the authoritative security contract for this route. An auditor reading "every credential-bearing setting is deliberately omitted" can conclude `GET /v3/_/events-sinks` cannot expose any sink credential and grant `ledger:OpsRead` broadly — at which point a `nats://svc:pw@nats.internal:4222` sink hands that password to every holder. The route is still strictly safer than before this PR, which shipped all five secrets in plaintext; that makes the exposure smaller, but it does not make the sentence true, which is what the blocking test turns on.

Resolution: scope the sentence to what the DTO actually guarantees, e.g. "The five secret-only settings are deliberately omitted — see the individual sink schemas. Note that `nats.url` and `http.endpoint` are returned in full, including any userinfo they embed."

Verified clean and needing no change: `internal/adapter/http/dto_sinks.go:14-21` (the SECURITY RULE header) and `docs/technical/contributing/api-comparison.md:722` are both correctly scoped to the five named fields. The over-claim exists in exactly one place, which supports reading it as a generalization slip rather than a considered claim.

paul-nicolas:
**Correction to the summary above.** I tagged the `openapi.yml:5890` credential over-claim non-blocking; it is now re-tagged **blocking**, and the comment records the edit.

Reason: this repo's review contract makes a finding blocking when merging would "make authoritative guidance materially false", and the two blocking findings already standing on this PR were filed on exactly that basis. The new `SinkConfig` sentence is the same class — and it is a security guarantee, not a mechanism description. "Every credential-bearing setting is deliberately omitted" is false for a NATS sink authenticated via URL userinfo, which is the only auth channel `NatsSinkConfig` has. That the route is net safer than before this PR reduces the exposure but does not make the sentence true.

Revised counts: **3 blocking** (the two carried over from 2026-08-18, plus this one), 1 new non-blocking P2 (the six sink default-sentinel fields), 11 P3 not posted inline. Decision is unchanged at REQUEST CHANGES.

One process note for the record: `docs/technical/contributing/ai-review.md` landed in #1714 (`3055d1ac9`), **after** this PR's base `c0c5177ab`, so it is absent from this branch. The verifiers that tried to read it from the PR head fell back to their own judgment; I applied the contract from `release/v3.0` at synthesis, and re-checking the blocking clause against it is what surfaced the inconsistency corrected here.

docs/technical/architecture/subsystems/api/http-api.md
paul-nicolas: Fixed in baf2d0a8a — confirmed valid, and both halves of the charge held.

`handlers_get_log.go` passes a raw `*commonpb.Log` to `writeOKChecked`; `Log.MarshalJSON` routes `responseSignature` through `protoFieldJSON` → `protojson.Marshal` on every variant, and `LogPayload.MarshalJSON` falls through to a bare `protojson.Marshal` for every oneof variant except `createLedger`/`deleteLedger`/`apply`. `AuditEntry.MarshalJSON` does the same on the audit reads. So the `{data:<micros>}` / quoted-`uint64` class is live on `GET /v3/_/logs/{sequence}`, and the AST scan is `os.ReadDir(".")` so it cannot see any of it.

`:604` now states the enforced scope — no protojson **call site in this package** — and splits out two things the invariant does not say: a `proto.Message` may sit inside a DTO, and a payload type's own `MarshalJSON` may delegate, naming all three marshallers with the routes they serve. The "Why protojson is not used" bullets are now framed as properties of the encoder, which hold wherever it runs. The `TestNoProtojsonMarshalInResponsePath` doc comment carries the same scoping, and the import-allowlist failure message was reworded too — it asserted the same over-broad claim.

Narrowing those three marshallers is left as separate work rather than folded in here.

internal/adapter/http/dto_indexes.go
paul-nicolas: Fixed in baf2d0a8a — valid.

`indexes.md:141` claimed "All responses serialize the protobuf message in protobuf-JSON camelCase", and that row enumerates exactly the routes converted here. Worth noting the sentence was wrong for one more route than the finding counted: `GET /v3/{ledger}/indexes/{canonicalId}/inspect` is in the same row and returns hand-written `inspectXxxJSON` types, so it was never protobuf-JSON either.

The row now points at `dto_indexes.go` / `handlers_inspect_index.go`, lists the four observable shape differences (unquoted integers, RFC3339Nano, emitted zero values, `canonicalId` with no proto counterpart) and cross-links `http-api.md#response-serialization`. The `{data:…}` envelope claim was accurate and is kept.

docs/technical/architecture/subsystems/api/http-api.md
paul-nicolas: Fixed in 857c34945 — valid, and verified handler by handler: all nine converted routes call `writeOKChecked`, so both `http-api.md:628` and the `response.go` doc comment were falsified by the change that shipped them.

Resolved in the direction of the code rather than the document: hand-written response DTOs now use `writeOKChecked` **by default**, with the reasoning recorded — a uniform rule removes a per-route judgement call whose wrong answer is an error object appended to a committed 200, and the buffer costs one allocation on a low-frequency ops read. `writeOKChecked` stays *mandatory* for the two categories where the marshaller can genuinely fail. Both the document and the doc comment were updated in the same pass.

tests/schemathesis/test_api.py
paul-nicolas: Fixed in 50603e45f — both halves confirmed.

(a) The metadata-schema DELETE is now coerced. `{key}` goes to a `schema-sacrifice` field seeded by `run.sh` on all three `targetType` enum values — declared for all three deliberately, since a sacrifice present for only one leaves the other two 404-ing instead of exercising the route. The comment names the cascade in `processRemoveMetadataFieldType` and records that the route is unreachable today only because `{key}` has no `example:`.

(b) `run.sh` now captures the fuzz exit status instead of letting the pipeline abort the script, re-asserts the fixture index afterwards — 200 **and** `TARGET_TYPE_ACCOUNT` present, the exact field the original defect dropped — and re-raises the captured status so a conformity failure still fails the job. The error message points at `before_call` as the thing to extend, since that is what a future de-seeding will mean.

Verified locally with `bash -n`, `ast.parse`, the `curl` status-capture idiom against 200/404 and trailing-newline bodies, and the coercion path string checked byte-for-byte against `openapi.yml`. The suite itself needs a running cluster, so it is exercised on CI rather than here.

internal/query/executor.go
paul-nicolas: Fixed in 9946fc0e6 — valid. `grep -rn "invalid cursor\|decodeCursor" --include='*_test.go' .` returned zero matches tree-wide, so this site had no coverage at any level.

Added `internal/query/executor_invalid_cursor_test.go`, driven through the exported `query.Execute` rather than `executeList` directly, so it exercises the path a client actually takes (prepared query + `QUERY_MODE_LIST` — which is exactly what made this site hard to reach and is why it was missed). Both `decodeCursor` failure branches are covered: bad base64, and valid base64 that is not the `cursorData` object. The assertion is on the classification (`ErrorAs` a `domain.Describable`), not the message, since that is the actual contract.

Mutation-checked: restoring the bare `fmt.Errorf` fails both cases. One implementation note worth recording — seeding had to go through `attrs.PreparedQuery.Set`, not `state.SavePreparedQuery`: the read path is `ReadPreparedQuery` → `attr.Get` on the attributes zone, while `SavePreparedQuery` writes the per-ledger zone, so the obvious seeding call leaves `Execute` reporting the query as not found.

internal/adapter/http/dto_sinks.go
paul-nicolas: Fixed in f8bd49c6f — valid, and gated for all three switches rather than just `SinkConfig`.

Added `dto_oneof_exhaustiveness_test.go` covering `SinkConfig.type`, `DatabricksSinkConfig.auth` and `IndexID.kind`. Deviated from the `wrapper_exhaustiveness_test.go` / `request_scope_exhaustiveness_test.go` precedents on one point: both of those keep a hand-maintained name list cross-checked against the descriptor, and the double-write is the protection. These are fully descriptor-driven instead — each variant is populated reflectively, pushed through the real converter, and the rendered JSON must carry it — so there is no second list to drift. The databricks case asserts on `authMethod` rather than the wire field, since the PAT variant is entirely secret by design.

No `default` arm added to the production switches: as you noted, the sibling switches degrade in-band and a runtime default here would have nothing useful to emit, so the test is the gate.

Also corrected the comment on `TestNewIndexIDDTO_AllOneofVariants`, which claimed that hand-written table was "the only exhaustiveness guard" — it now points at the descriptor test for the fifth-variant case it cannot see.

All three gates mutation-checked: dropping the `http` arm, the databricks `token` arm and the `logBuiltin` arm each fails exactly its own subtest.

internal/adapter/http/handlers_list_signing_keys.go
paul-nicolas: Fixed in 2fd7bd471 — valid. `ReadSigningKeys` returns `map[string]SigningKeyEntry` and `ReadSigningKeysCursor` builds its slice by ranging over it, so order was map-iteration order.

`newSigningKeyDTOList` now sorts by `keyId`, mirroring `dto_sinks.go`. Sorted in the DTO rather than in `ReadSigningKeysCursor`, for the reason recorded in the comment: the gRPC surface has its own bidirectional-cursor ordering contract, and this PR already establishes ordering as an HTTP wire-shape concern the DTO owns. `openapi.yml` and `api-comparison.md` now state the guarantee the way the sinks schema does.

Two assertions added — one on the converter, one on the response body so the ordering is pinned through the handler — and both mutation-checked against removal of the sort.

internal/adapter/http/dto_sinks.go
paul-nicolas: Fixed in d96652960 — valid, and the count of six and every default value check out: `DefaultEmitterConfig()` gives 64 / 10 ms / `json`, `sink_databricks.go` defaults port to 443 and table to `ledger_events`, `sink_clickhouse.go` the same for its table. Nothing defaults them at write time, so the stored sentinel is what the DTO reports.

Doc-only as proposed. All six descriptions now name the sentinel in the `"0 = server default (N)"` wording the file already uses at `MirrorSourceConfig.batchSize` and the audit `limit` parameter. `batchSize` and `batchDelayMs` additionally record *why* it matters in practice — `ledgerctl events add-sink` leaves both flags at `0`, so an ordinary sink reports `0` while actually batching 64 with a 10 ms flush.

openapi.yml
paul-nicolas: Fixed in d96652960 — valid, and treated as blocking per your re-tag.

Verified on every leg: `NatsSinkConfig` has only `url` and `topic`, `sink_nats.go:46` authenticates solely via `nats.Connect(cfg.URL)`, `dto_sinks.go` copies the URL verbatim, `redact.go` has cases for Kafka/HTTP/ClickHouse/Databricks and none for Nats, and `grep -rn 'url.Parse' internal/application/events/` returns nothing — so no admission-side stripping either. The file defining the ClickHouse DSN as credential-bearing twelve lines earlier is what makes the sentence self-contradictory.

The `SinkConfig` description now names the five secret-only settings that *are* omitted and states plainly that `nats.url` and `http.endpoint` are returned in full including embedded userinfo, with the `ledger:OpsRead` consequence spelled out. `NatsSinkConfig.url` and `HttpSinkConfig.endpoint` each carry the caveat, so "see the individual sink schemas" is now true; the NATS one records that userinfo is the only auth channel it has, the HTTP one that its `secret` signs the payload rather than authenticating the client.

Went one step beyond the proposed resolution: the same caveat is now on the SECURITY RULE header in `dto_sinks.go`, which is where a contributor actually reads the rule, with an explicit "do not shorten this back to 'every credential-bearing setting is omitted'". `api-comparison.md` got the parity note too.

No redaction behaviour changed here — that would be a separate change on both transports.

PR #1704, opened 2026-08-14
chore: make agent definition of done executable

gfyrag:
## Summary

This PR turns the agent Definition of Done into an executable repository contract.

- adds `scripts/agent-check` as the canonical baseline validation for agent-authored code changes
- adds `scripts/agent-check-full` for baseline validation plus the full root-module unit suite
- updates `AGENTS.md` with an explicit Definition of Done and handoff requirements
- keeps task-specific tests separate so agents must still select tests based on the touched subsystem and risk
- runs canonical validation inside the repository-pinned Nix environment rather than depending on host tooling
- validates tracked, staged, and non-ignored untracked files for whitespace errors

## Baseline contract

`bash scripts/agent-check` re-execs under `nix develop` when needed, clears inherited `GOROOT`, and runs:

1. the existing `just pre-commit` pipeline
2. `go build ./...`
3. whitespace validation across tracked/staged changes relative to `HEAD`
4. whitespace validation for non-ignored untracked files

`bash scripts/agent-check-full` runs the same pinned baseline and then `just test` in that same Nix environment.

## Why

The goal is to reduce human review load by ensuring agents complete deterministic, reproducible repository checks before handing work to another agent or a human reviewer. A task is no longer considered complete merely because the implementation appears finished.

The full unit suite is intentionally not mandatory for every narrow change. Agents must run subsystem-appropriate tests and use the full command for broad/high-risk work where it is justified.

## Scope

This PR does not add new architectural lints, CI gates, or validation logic. It composes existing repository checks into a canonical agent workflow and makes the handoff validation reproducible and complete for the current working tree.

## Review focus

Please check:

- whether the pinned Nix re-exec is robust for normal developer/agent environments
- whether the baseline validation is sufficient without being unnecessarily expensive
- whether tracked/staged/untracked whitespace coverage matches the intended handoff scope
- whether `agent-check-full` selects the right existing test target
- whether the Definition of Done is precise enough to prevent premature handoff without forcing unrelated work

### DIFF
--- a/scripts/agent-check
+++ b/scripts/agent-check
@@ -1,13 +1,47 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+# Canonical agent validation must run in the repository-pinned Nix environment.
+# Re-exec once under `nix develop` instead of relying on whatever tools happen
+# to be installed or loaded by the caller's shell/direnv session.
+if [[ "${LEDGER_AGENT_CHECK_IN_NIX:-}" != "1" ]]; then
+    exec env -u GOROOT nix develop --command \
+        env LEDGER_AGENT_CHECK_IN_NIX=1 bash "$0" "$@"
+fi
+
+# Never inherit a host GOROOT into pinned Go commands.
+unset GOROOT || true
+
 echo "==> agent-check: repository pre-commit validation"
 just pre-commit
 
 echo "==> agent-check: compile all root-module packages"
-GOROOT= go build ./...
+go build ./...
 
-echo "==> agent-check: verify diff whitespace"
-git diff --check
+echo "==> agent-check: verify tracked/staged diff whitespace"
+git diff --check HEAD --
+
+echo "==> agent-check: verify untracked file whitespace"
+untracked_failed=0
+while IFS= read -r -d '' file; do
+    # `git diff --no-index` returns 1 whenever files differ, even when there are
+    # no whitespace errors. Treat output from --check as the failure signal;
+    # reserve exit statuses >1 for actual command errors.
+    output=""
+    status=0
+    output=$(git diff --no-index --check -- /dev/null "$file" 2>&1) || status=$?
+    if [[ -n "$output" ]]; then
+        printf '%s\n' "$output"
+        untracked_failed=1
+    elif (( status > 1 )); then
+        echo "agent-check: failed to inspect untracked file: $file" >&2
+        exit "$status"
+    fi
+done < <(git ls-files --others --exclude-standard -z)
+
+if (( untracked_failed != 0 )); then
+    echo "agent-check: whitespace errors found in untracked files" >&2
+    exit 1
+fi
 
 echo "==> agent-check: PASS"
--- a/scripts/agent-check
+++ b/scripts/agent-check
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+echo "==> agent-check: repository pre-commit validation"
+just pre-commit
+
+echo "==> agent-check: compile all root-module packages"
+GOROOT= go build ./...
+
+echo "==> agent-check: verify diff whitespace"
+git diff --check
+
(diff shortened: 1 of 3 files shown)


scripts/agent-check
```diff
@@ -0,0 +1,13 @@
...
+set -euo pipefail
+
+echo "==> agent-check: repository pre-commit validation"
+just pre-commit
+
+echo "==> agent-check: compile all root-module packages"
+GOROOT= go build ./...
+
+echo "==> agent-check: verify diff whitespace"
+git diff --check
```
NumaryBot: 🟠 **[major] Include staged changes in whitespace validation**

When an agent stages its edits before running the required check, plain `git diff --check` examines only unstaged changes and can report success while staged files contain whitespace errors. Use a comparison that includes the index, such as `git diff --check HEAD`, so the baseline validates the work being handed off.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1704#issuecomment-5294463390

scripts/agent-check
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+echo "==> agent-check: repository pre-commit validation"
+just pre-commit
```
gfyrag, line 16: 🟠 **[major] Run the canonical check inside the pinned Nix toolchain**

This replaces the explicit `nix develop` validation path with a direct `just pre-commit` call. In a worktree where direnv had not loaded the flake, running `agent-check-full` selected Homebrew `protoc 35.1` instead of the pinned Nix `34.2`, rewrote 18 generated `.pb.go` headers, and still reported `PASS`. The contributor docs declare Nix required, and the previous contract also cleared inherited `GOROOT`. Re-exec the whole workflow under `nix develop` when necessary and clear `GOROOT` for every Go-based phase, including pre-commit and the full tests, so the canonical result is reproducible.

scripts/agent-check
gfyrag: 🟠 **[major] Validate new untracked files as part of the handoff**

In addition to the existing staged-change finding, the suggested `git diff --check HEAD` still ignores untracked files. I verified that a new untracked file containing trailing whitespace returns success from both `git diff --check` and `git diff --check HEAD`. New files are a normal part of agent-authored changes, and this Definition of Done does not require staging before validation, so the baseline can still pass an invalid handoff. Include non-ignored untracked files in the whitespace check, or make staging all handoff files an explicit enforced precondition.

gfyrag (commented):
Two additional actionable findings from the manual review.

scripts/agent-check-full
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Keep the baseline and the full test suite in the same pinned toolchain.
+if [[ "${LEDGER_AGENT_CHECK_IN_NIX:-}" != "1" ]]; then
+    exec env -u GOROOT nix develop --command \
+        env LEDGER_AGENT_CHECK_IN_NIX=1 bash "$0" "$@"
+fi
```
NumaryBot, line 8: 🔴 **[blocker] Run the canonical check inside the pinned Nix toolchain**

When `agent-check-full` starts outside direnv, `agent-check` runs in a child `nix develop` process, but this subsequent `just test` runs back in the caller's host environment and inherits its `GOROOT` and tools. Thus the documented full check can use incompatible tooling or fail after the baseline passes; re-exec the full workflow under Nix as required by [getting-started.md:5](docs/technical/contributing/getting-started.md#L5).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1704#issuecomment-5294463390

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1704#issuecomment-5294463390

paul-nicolas (approved):

PR #1707, opened 2026-08-17
fix(EN-1781): stop pterm spinners leaking goroutines that race with global config

paul-nicolas:
## Summary

The E2E suite failed under `-race` on pterm's package-level `RawOutput` flag.

**The cause is not parallel specs.** The cluster suite runs strictly serially (`go test -p 1` at `justfile:96`, a bare `RunSpecs` at `tests/e2e/cluster/suite_test.go:14`). Both racing goroutines live in one process. The race came from a spinner goroutine that outlived its spec.

Mechanism, verified against `pterm v0.12.82`:

| Fact | Evidence |
|---|---|
| `Start` spawns its animation goroutine unconditionally, even in `RawOutput` mode | `spinner_printer.go:157` |
| That goroutine reads the global `RawOutput` every tick | `spinner_printer.go:163` |
| `Stop` early-returns **without clearing `IsActive`** when `RawOutput` is set | `spinner_printer.go:184-186` |
| `Success`/`Fail`/`Warning`/`Info` all terminate via `Stop` | `:246`, `:263`, `:280`, `:229` |

Our own `cmdutil` `init()` set `RawOutput = true` for non-TTY stdout — always true under `go test` — so **no spinner could ever be stopped** and every in-process `ledgerctl` command leaked an immortal goroutine polling `RawOutput`. A later spec calling `pterm.DisableStyling()` wrote that flag (`pterm.go:62`) and the detector fired.

That `init()` came from commit `9f13b3a3b`, "disable pterm spinner animation in non-TTY to avoid data race" — a prior partial fix for this same race class. Its comment claimed it disabled the goroutine; it did not, and it is what made the goroutine immortal.

Because pterm reads `IsActive` without synchronisation, such a goroutine **cannot be stopped race-free — only never started**. That constraint drives the design.

## Changes

- **`cmdutil.Spinner` / `StartSpinner`** — creates no pterm spinner when stdout is not a terminal, so no goroutine exists. Output is byte-identical: in raw mode `Start` prints nothing, `fClearLine` is a no-op (`print.go:194`), and the terminal methods print through `pterm.Fprinto` with the same prefix printers.
- **Write-once configuration** — interactivity is decided once at package init; the misleading comment is corrected.
- **70 call sites migrated** across 62 files in `cmd/ledgerctl`.
- **`forbidigo` rule** banning direct `pterm.DefaultSpinner`, following the existing `OpenWriteSession` precedent.
- **E2E spec no longer reconfigures pterm globals** (`ledgerctl_typed_metadata_test.go`).
- **Tests:** a `goleak` regression test and a pterm output-equivalence test.

## Behaviour change

`DisableStyling()` also disables colour, where the previous code set only `RawOutput`. Non-interactive output (pipes, CI, files) no longer contains ANSI escape codes. `--json`/`--yaml` payloads come from encoders, not pterm, so machine-readable output is unaffected.

## Scope: `misc/operator` excluded, deliberately

`misc/operator` (16 spinner sites) is a separate Go module that cannot import `cmd/ledgerctl/cmdutil`. It contains **zero** pterm global writes — no `DisableStyling`, `DisableColor`, `RawOutput`, or `SetDefaultOutput` anywhere — and no in-process suite running its commands. A race needs a write concurrent with a read; there is no writer. `kubectl-ledger` therefore has the latent pattern, not this bug, in a short-lived plugin process. It is excluded from the lint rule with that rationale recorded inline, rather than duplicating the wrapper across a module boundary to fix a non-bug.

## Validation

Run locally:

- `just pre-commit` — exit 0, both golangci passes `0 issues`, no files modified.
- `just test` — 81 `ok`, 0 FAIL, 0 `DATA RACE`.
- `go test -race ./cmd/ledgerctl/...` — 8 `ok`, 0 FAIL, 0 `DATA RACE`.
- `goleak` test asserting a start/Success cycle leaves zero goroutines. Verified to genuinely guard: forcing `startSpinner` to always create a real pterm spinner fails it with the predicted stack at `spinner_printer.go:164`, created at `:157`.
- Lint rule verified to genuinely fire: a temporary `pterm.DefaultSpinner` reference produces a `forbidigo` finding; a deliberately non-matching exclusion anchor surfaces all 16 operator sites, confirming the exclusion is what suppresses them rather than the files not being linted.

**`just test-e2e-full` is NOT yet confirmed green — please rely on CI for this one.** The local run reported **0 `DATA RACE`** with the detector active, and the migrated spinners demonstrably executed. But it exited 1 for three environmental reasons unrelated to this change, so not all cluster specs ran and the result is inconclusive rather than passing:

1. Azurite rejects API version `2026-04-06` (`backup_azure_test.go:106`) — a pre-existing environment problem, independent of this branch.
2. `bind: address already in use` on port 15200, from a concurrent E2E run in another workspace on the same machine.
3. The cluster suite aborted as a cascade of (2), with no `Ran N of N Specs` summary.

CI runs E2E in an isolated environment and is the authoritative check here.

## Known limitation

A pre-existing blanket `_test\.go$` `forbidigo` exclusion (present for `OpenWriteSession` test seeding) makes the new rule inert in test files repo-wide. Narrowing it would disturb unrelated enforcement, so it is left alone and recorded here. Also: the `^misc/operator/` anchor works only because golangci matches paths repo-root-relative; adding a `misc/operator/.golangci.yaml` would silently break it.

Jira: https://formance-team.atlassian.net/browse/EN-1781

### DIFF
--- a/cmd/ledgerctl/cmdutil/spinner.go
+++ b/cmd/ledgerctl/cmdutil/spinner.go
@@ -1,0 +1,127 @@
+package cmdutil
+
+import (
+	"io"
+
+	"github.com/pterm/pterm"
+)
+
+// Spinner reports progress for a long-running command step.
+//
+// When stdout is a terminal it wraps a real pterm spinner and behaves exactly
+// as pterm does. When stdout is not a terminal it wraps nothing: no pterm
+// spinner is created, so no animation goroutine exists.
+//
+// That distinction is the fix for EN-1781. pterm's SpinnerPrinter.Start always
+// launches a goroutine that reads pterm's package-level configuration on every
+// tick, and its Stop returns early while pterm.RawOutput is set, leaving
+// IsActive true so the goroutine outlives the command. In a long-lived
+// non-interactive process — notably the in-process E2E suite, which runs many
+// ledgerctl commands in one binary — those goroutines accumulate and race with
+// any later write to pterm's configuration. Because pterm reads IsActive
+// without synchronisation, such a goroutine cannot be stopped race-free; it can
+// only be never started.
+//
+// The non-interactive path emits the same bytes pterm would: in raw mode Start
+// prints nothing, fClearLine is a no-op, and the terminal methods print through
+// pterm.Fprinto with the same prefix printers.
+type Spinner struct {
+	inner  *pterm.SpinnerPrinter
+	writer io.Writer
+	text   string
+}
+
+// StartSpinner starts a spinner showing text.
+//
+// This is the only sanctioned way to create a spinner in ledgerctl; direct use
+// of pterm.DefaultSpinner is rejected by the forbidigo rule in .golangci.yaml.
+func StartSpinner(text string) *Spinner {
+	return startSpinner(text, interactiveOutput, pterm.DefaultSpinner.Writer)
+}
+
+// startSpinner is the injectable form used by tests.
+func startSpinner(text string, interactive bool, writer io.Writer) *Spinner {
+	if !interactive {
+		return &Spinner{writer: writer, text: text}
+	}
+
+	inner, _ := pterm.DefaultSpinner.Start(text)
+
+	return &Spinner{inner: inner, writer: writer, text: text}
+}
+
+// UpdateText replaces the message shown by the spinner.
+func (s *Spinner) UpdateText(text string) {
+	s.text = text
+
+	if s.inner != nil {
+		s.inner.UpdateText(text)
+
+		return
+	}
+
+	pterm.Fprintln(s.writer, text)
+}
+
+// Success terminates the spinner with the success prefix.
+func (s *Spinner) Success(message ...any) {
+	if s.inner != nil {
+		s.inner.Success(message...)
+
+		return
+	}
+
+	s.finish(&pterm.Success, message)
+}
+
+// Fail terminates the spinner with the error prefix.
+func (s *Spinner) Fail(message ...any) {
+	if s.inner != nil {
+		s.inner.Fail(message...)
+
+		return
+	}
+
+	s.finish(&pterm.Error, message)
+}
+
+// Warning terminates the spinner with the warning prefix.
+func (s *Spinner) Warning(message ...any) {
+	if s.inner != nil {
+		s.inner.Warning(message...)
+
+		return
+	}
+
+	s.finish(&pterm.Warning, message)
+}
+
+// Info terminates the spinner with the info prefix.
+func (s *Spinner) Info(message ...any) {
+	if s.inner != nil {
+		s.inner.Info(message...)
+
+		return
+	}
+
+	s.finish(&pterm.Info, message)
+}
+
+// Stop terminates the spinner without printing a result.
+func (s *Spinner) Stop() error {
+	if s.inner != nil {
+		return s.inner.Stop()
+	}
+
+	return nil
+}
+
+// finish reproduces pterm's terminal-method behaviour for the non-interactive
+// path: default the message to the spinner text, then print through Fprinto.
+func (s *Spinner) finish(printer *pterm.PrefixPrinter, message []any) {
+	if len(message) == 0 {
+		message = []any{s.text}
+	}
+
+	pterm.Fprinto(s.writer, printer.Sprint(message...))
+}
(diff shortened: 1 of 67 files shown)


cmd/ledgerctl/cmdutil/spinner.go
```diff
@@ -0,0 +1,127 @@
...
+func (s *Spinner) UpdateText(text string) {
+	s.text = text
+
+	if s.inner != nil {
+		s.inner.UpdateText(text)
+
+		return
+	}
+
+	pterm.Fprintln(s.writer, text)
```
NumaryBot, line 67: 🟠 **[major] Keep non-interactive UpdateText silent**

When stdout is non-interactive, pterm's raw-mode `UpdateText` only changes the spinner text; it does not print a line. This replacement emits every update, so progress-heavy commands such as `ledgerctl upgrade` now write a line for every 32 KiB downloaded and pollute piped/CI output. Store the updated text without calling `Fprintln`, matching the claimed output-equivalence behavior.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1707#issuecomment-5342887882

cmd/ledgerctl/cmdutil/spinner.go
paul-nicolas: Not applying this -- the premise is inverted. pterm's raw-mode `UpdateText` is the branch that *does* print a new line (`pterm@v0.12.82/spinner_printer.go:135-142`):

```go
func (s *SpinnerPrinter) UpdateText(text string) {
	s.Text = text
	if !RawOutput {
		Fprinto(s.Writer, "\033[K"+...)  // interactive: overwrite in place
	} else {
		Fprintln(s.Writer, s.Text)       // raw mode: new line
	}
}
```

`Fprintln` is not suppressed by `RawOutput` either -- it goes through `Fprint` (`print.go:151`), which writes unconditionally. `RawOutput` gates `Printo`, `fClearLine` and `sClearLine`, not `Fprint`.

There is also no regression. Before this PR, `cmdutil/pterm.go` set `pterm.RawOutput = true` for non-interactive stdout and `upgrade/download.go` called `UpdateText` on a real `*pterm.SpinnerPrinter` inside the same 32 KiB read loop, so one line per 32 KiB is exactly what already shipped. `pterm.DisableStyling()` sets `RawOutput = true` (`pterm.go:58-62`), so the new `init()` preserves that state.

Verified by mutation: applying the suggested change (store the text, drop the `Fprintln`) makes the wrapper diverge from pterm --

```
--- FAIL: TestSpinnerNonInteractiveUpdateTextMatchesPterm
    expected: "still working\n"     <- pterm's own raw-mode output
    actual  : ""
```

So dropping the `Fprintln` would break the documented output equivalence, not restore it.

The coverage gap behind the finding is real, though: nothing pinned `UpdateText`'s output. Added `TestSpinnerNonInteractiveUpdateTextMatchesPterm` and `TestSpinnerNonInteractiveUpdateTextRetargetsDefaultMessage` in 3689c8b1a so the equivalence is enforced by a test rather than asserted in a comment, plus a note on `UpdateText` recording which pterm branch it mirrors.

Separately: one progress line per 32 KiB in piped output is genuinely noisy, but it is pre-existing pterm behaviour and throttling it is a user-visible change that belongs in its own ticket, not in a goroutine-leak fix that commits to output equivalence.

paul-nicolas:
The output regression this summary reports does not exist -- details in the inline thread.

Short version: pterm's raw-mode `UpdateText` is the branch that calls `Fprintln` (`spinner_printer.go:135-142`); the in-place `Fprinto` branch is the interactive one. Before this PR the same path already printed one line per `UpdateText` call, because `cmdutil/pterm.go` set `pterm.RawOutput = true` and `upgrade/download.go` drove a real `pterm.SpinnerPrinter` through the same 32 KiB loop. Output is byte-identical before and after.

Confirmed by mutation: applying the suggested fix makes the wrapper diverge from pterm's own output. Added tests in 3689c8b1a pinning that equivalence.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1707#issuecomment-5342887882

gfyrag (dismissed):

gfyrag (approved):

PR #1708, opened 2026-08-17
fix(EN-1780): green the full-tag e2e suite and gate it in CI

paul-nicolas:
## Summary

The restored balance was correct; the E2E fixture was stale. The fixture writes an additional `bank -> eve 500` posting before the second full backup, so the restored bank output is `5500`, not `5000`. The expectations are now derived from named posting amounts so future fixture changes cannot silently leave a frozen total behind.

This PR also makes the full feature-tagged E2E suite a required CI gate and fixes the pre-existing failures that gate exposed.

## CI gap

Before this change, `test-e2e-coverage` ran only with the `e2e` tag. The S3, Azure, NATS, ClickHouse and Databricks specs therefore had no runtime coverage in the required workflow. The gate now runs:

```
e2e,kafka,nats,clickhouse,databricks,s3,azure,pyroscope
```

Databricks specs continue to skip themselves when the environment is not configured.

`just lint` does not pass the `e2e` tag and therefore does not compile these packages. The required full-tag run is responsible for both compilation and runtime assertions.

## Defects fixed while enabling the gate

| Defect | Ticket | Fix |
|---|---|---|
| Stale restored bank balance | EN-1780 | Derive the expected total from the fixture postings |
| pterm global race / leaked spinner lifecycle | EN-1781 | Configure pterm once per suite with `DisableOutput`; keep styling enabled because `RawOutput` makes `SpinnerPrinter.Stop` return without stopping its goroutine |
| Azurite rejects the SDK API version | EN-1782 | Run the blob emulator with `--skipApiVersionCheck` |
| S3 empty-prefix assertion compared `*int32` with `int` | EN-1783 | Assert `listOut.Contents` is empty, which also reports leaked objects directly |
| Deterministic port collision on 15200 | EN-1784 | Reject HTTP, gRPC and derived Raft ports reserved by the shared node before startup; cover the helper with table tests |

The port collision was deterministic arithmetic, not a teardown race: `grpcPort = 16200` derives `raftPort = 15200`, which is the shared single-node HTTP port.

## Reproducible containers

All MinIO and Azurite E2E consumers now use centralized, immutable image pins in `tests/e2e/testutil/containers.go`:

- `minio/minio:RELEASE.2024-06-13T22-53-53Z`
- `mcr.microsoft.com/azure-storage/azurite:3.36.0`

The MinIO version matches the Antithesis fixtures. Emulator upgrades are now explicit code changes validated by the complete gate.

## Validation

- `nix develop --command bash -c "just pre-commit"` — pass, 0 lint issues and no generated-file drift
- `GOROOT= go build ./...` — pass
- `just test-full` — pass with all optional features and the race detector
- Full-tag business suite with coverage and `-race` — pass
- Full-tag cluster suite with coverage and `-race` — pass (329 specs); an initial run hit the existing transient follower-election assertion in `rolling_config_test.go:43`, then the focused test and the complete cluster suite rerun with the exact failing seed both passed
- Focused Azure, S3 credentials and rejected incremental-backup specs against the pinned containers — pass
- GitHub Actions on commit `43b64b694`: Build, Dirty, Tests, Tests-E2E, Tests-Model, Tests-Scenarios, Tests-Schemathesis, Coverage and repository invariants — all pass

## Documentation

`docs/technical/contributing/testing.md` documents the required full-tag gate, the distinction from the light local target, the actual lint coverage, the pinned emulator images and the macOS Docker socket trap.

## Tickets

Fixes https://formance-team.atlassian.net/browse/EN-1780  
Also fixes https://formance-team.atlassian.net/browse/EN-1781  
Also fixes https://formance-team.atlassian.net/browse/EN-1782  
Also fixes https://formance-team.atlassian.net/browse/EN-1783  
Also fixes https://formance-team.atlassian.net/browse/EN-1784

### DIFF
--- a/tests/e2e/testutil/server_test.go
+++ b/tests/e2e/testutil/server_test.go
@@ -1,0 +1,69 @@
+//go:build e2e
+
+package testutil
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestNodePortConflict(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name        string
+		httpPort    int
+		grpcPort    int
+		raftPort    int
+		errContains string
+	}{
+		{
+			name:     "shared node owns its reserved ports",
+			httpPort: TestSingleHTTPPort,
+			grpcPort: TestSingleGRPCPort,
+			raftPort: TestSingleGRPCPort - 1000,
+		},
+		{
+			name:     "auxiliary node uses distinct ports",
+			httpPort: 18000,
+			grpcPort: 18100,
+			raftPort: 17100,
+		},
+		{
+			name:        "HTTP port collides",
+			httpPort:    TestSingleHTTPPort,
+			grpcPort:    18100,
+			raftPort:    17100,
+			errContains: "HTTP port 15200",
+		},
+		{
+			name:        "gRPC port collides",
+			httpPort:    18000,
+			grpcPort:    TestSingleGRPCPort,
+			raftPort:    17000,
+			errContains: "gRPC port 15100",
+		},
+		{
+			name:        "Raft port collides",
+			httpPort:    16100,
+			grpcPort:    16200,
+			raftPort:    TestSingleHTTPPort,
+			errContains: "Raft port (derived from gRPC port 16200 minus 1000) 15200",
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			err := nodePortConflict(test.httpPort, test.grpcPort, test.raftPort)
+			if test.errContains == "" {
+				require.NoError(t, err)
+
+				return
+			}
+			require.ErrorContains(t, err, test.errContains)
+		})
+	}
+}
--- a/tests/e2e/testutil/server_test.go
+++ b/tests/e2e/testutil/server_test.go
@@ -1,5 +1,3 @@
-//go:build e2e
-
 package testutil
 
 import (
@@ -38,10 +36,12 @@ func TestNodePortConflict(t *testing.T) {
 			errContains: "HTTP port 15200",
 		},
 		{
+			// raftPort matches what SetupSingleNode derives (grpcPort - 1000),
+			// so the tuple is one the production call site can actually build.
 			name:        "gRPC port collides",
 			httpPort:    18000,
 			grpcPort:    TestSingleGRPCPort,
-			raftPort:    17000,
+			raftPort:    TestSingleGRPCPort - 1000,
 			errContains: "gRPC port 15100",
 		},
 		{
@@ -50,6 +50,28 @@ func TestNodePortConflict(t *testing.T) {
 			grpcPort:    16200,
 			raftPort:    TestSingleHTTPPort,
 			errContains: "Raft port (derived from gRPC port 16200 minus 1000) 15200",
+		},
+		{
+			// The shared node's derived Raft port is reserved too, and nothing
+			// else in this table ever passes 14100 as a port value. An auxiliary
+			// spec picking it for HTTP would collide with the shared node's Raft
+			// listener.
+			name:        "shared node's derived Raft port is reserved",
+			httpPort:    TestSingleGRPCPort - 1000,
+			grpcPort:    18100,
+			raftPort:    17100,
+			errContains: "HTTP port 14100 is reserved for the shared single-node server (the shared node's derived Raft port)",
+		},
+		{
+			// The other gRPC value that slips past the HTTP and gRPC checks and
+			// is only caught on its derived Raft port. 16100 is the live value of
+			// the business suite's s3CredsHTTPPort, so transposing that spec's
+			// HTTP and gRPC constants is the plausible next collision.
+			name:        "Raft port collides via the second uncaught gRPC value",
+			httpPort:    18000,
+			grpcPort:    16100,
+			raftPort:    TestSingleGRPCPort,
+			errContains: "Raft port (derived from gRPC port 16100 minus 1000) 15100",
 		},
 	}
 
--- a/docs/technical/contributing/testing.md
+++ b/docs/technical/contributing/testing.md
@@ -520,7 +520,7 @@ plain `-tags e2e` run neither builds nor runs them:
 | `s3` | backup, restore, bootstrap, cold storage | MinIO (Testcontainers) |
 | `azure` | Azure Blob backup | Azurite (Testcontainers) |
 | `clickhouse` | ClickHouse event sink | ClickHouse (Testcontainers) |
-| `nats` | NATS event sink | NATS (Testcontainers) |
+| `nats` | NATS event sink | none (embedded in-process server) |
 | `databricks` | Databricks event sink | real workspace credentials; skips itself when unset |
 
 `just test-e2e` runs the light set for a fast local loop. **CI runs the full tag
--- a/docs/technical/contributing/testing.md
+++ b/docs/technical/contributing/testing.md
@@ -509,6 +509,44 @@ ginkgo run -tags e2e ./tests/e2e
 go test -tags e2e ./tests/e2e/...
 ```
 
+#### Feature-gated suites
+
+Some e2e files carry an extra build tag on top of `e2e` (for example
+`//go:build e2e && s3`). They are compiled only when that tag is passed, so a
+plain `-tags e2e` run neither builds nor runs them:
+
+| Tag | Coverage | External dependency |
+|-----|----------|---------------------|
+| `s3` | backup, restore, bootstrap, cold storage | MinIO (Testcontainers) |
+| `azure` | Azure Blob backup | Azurite (Testcontainers) |
+| `clickhouse` | ClickHouse event sink | ClickHouse (Testcontainers) |
+| `nats` | NATS event sink | NATS (Testcontainers) |
+| `databricks` | Databricks event sink | real workspace credentials; skips itself when unset |
+
+`just test-e2e` runs the light set for a fast local loop. **CI runs the full tag
+set** through `just test-e2e-coverage`, so a green `just test-e2e` does not
+guarantee a green pipeline. Reproduce the CI set locally with:
+
+```bash
+just test-e2e-full
+```
+
+Linting already parses these files (`just lint` passes every tag), so a
+compile error is caught without the tag. Only a real run catches a stale
+assertion — which is why the full set belongs in CI.
+
+#### Docker socket on macOS
+
+Testcontainers looks for `/var/run/docker.sock`. When Docker Desktop is
+installed but stopped, that path exists as a symlink to its inactive socket,
+and container startup fails with `Cannot connect to the Docker daemon` even
+though the `docker` CLI works against another runtime. With OrbStack, point
+Testcontainers at the right socket:
+
+```bash
+export DOCKER_HOST=unix://$HOME/.orbstack/run/docker.sock
+```
+
 ### Operator Integration Tests
 
 ```bash
--- a/docs/technical/contributing/testing.md
+++ b/docs/technical/contributing/testing.md
@@ -531,9 +531,13 @@ guarantee a green pipeline. Reproduce the CI set locally with:
 just test-e2e-full
 ```
 
-Linting already parses these files (`just lint` passes every tag), so a
-compile error is caught without the tag. Only a real run catches a stale
-assertion — which is why the full set belongs in CI.
+`just lint` does not pass the `e2e` tag, so it excludes these packages even
+when feature tags are configured. The full CI run is therefore responsible
+for both compiling these files and exercising their assertions.
+
+The MinIO and Azurite images used by this required gate are version-pinned in
+`tests/e2e/testutil/containers.go`. Update those pins deliberately and validate
+the complete full-tag suite when upgrading either emulator.
 
 #### Docker socket on macOS
 
--- a/justfile
+++ b/justfile
@@ -180,13 +180,17 @@ test-coverage:
     GOTOOLCHAIN=$(go env GOVERSION) go test -race -coverprofile={{coverage_dir}}/unit.out -coverpkg={{coverage_pkgs}} ./... -timeout 20m
     echo "Coverage profile: {{coverage_dir}}/unit.out"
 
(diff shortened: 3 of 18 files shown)


tests/e2e/testutil/server_test.go
```diff
@@ -0,0 +1,69 @@
+//go:build e2e
```
paul-nicolas: [P2][blocking] Compile and run the new port-guard test — nothing in the repository does

Location: tests/e2e/testutil/server_test.go:1 (the `//go:build e2e` constraint) — `TestNodePortConflict`

Evidence: this new file is `//go:build e2e`, and no invocation in the repository compiles the `tests/e2e/testutil` package with that tag:
- `just test` (justfile:81) and `just test-coverage` (justfile:180, the CI `Tests` job) run `./...` with no tag, so the constraint excludes it.
- `just test-full` (justfile:85) adds only `{{all_tags}}`, not `e2e`.
- `just test-e2e` (justfile:92), `test-e2e-full` (:96) and `test-e2e-coverage` (:193) do pass `e2e`, but restrict the package list to `./tests/e2e/business/... ./tests/e2e/cluster/...`. `tests/e2e/testutil` is in neither.
- `just lint` (justfile:26) passes `--build-tags it,local,{{all_tags}}` and `.golangci.yaml` sets no build tags, so golangci-lint does not type-check it either. `go build ./...` never compiles `_test.go` files.

Reproduced:
```
$ go test -tags "e2e,kafka,nats,clickhouse,databricks,s3,azure,pyroscope" -run '^TestNodePortConflict$' ./tests/e2e/business/... ./tests/e2e/cluster/...
ok  .../tests/e2e/business  [no tests to run]
ok  .../tests/e2e/cluster   [no tests to run]
$ go test -run '^TestNodePortConflict$' ./tests/e2e/testutil/
?   .../tests/e2e/testutil  [no test files]
$ go test -tags e2e -run '^TestNodePortConflict$' ./tests/e2e/testutil/
ok  .../tests/e2e/testutil   (5/5 subtests PASS)
```
`go list -f '{{.TestGoFiles}}'` on that package returns `[]` for both the untagged and the `just lint` tag sets, and `[server_test.go]` only with `e2e`. Confirmed against the real `Tests-E2E` run for this head: the log has exactly two package result lines (business, cluster) and no testutil line.

This is not an existing convention. `tests/e2e` has three packages, and `server_test.go` is the only `*_test.go` outside `business`/`cluster`; the other ~100 `//go:build e2e` files under `tests/` all live in packages the recipes do cover.

Impact: the only regression test for `nodePortConflict` provides no protection and is not even compile-checked. A permissive degradation — an entry dropped from the `reserved` map at server.go:350-354 — stays invisible: the nil path returns nil and every e2e spec stays green. That is precisely the class this test exists to catch, and the one the running suites cannot catch themselves. A rename or signature change breaks the file with no command in the repository reporting it. The justfile comment added by this same PR says running the feature-gated suites in CI "is what stops them rotting unnoticed", so shipping a never-compiled test cuts against the PR's own purpose.

Resolution: deleting line 1 is sufficient and preferable — server.go, helpers.go and containers.go carry no build constraint and the test imports only `testing` and `testify/require`. Verified with a `go test -overlay` run that strips only that line: 5/5 subtests pass with no tags, which places it in the fast `Tests` job rather than the 20-minute e2e gate. Alternatively append `./tests/e2e/testutil/...` to justfile:92, :96 and :193. Resolved when `TestNodePortConflict` appears in a CI job log.

paul-nicolas (commented):

Reviewed at head `43b64b694`. First review of this PR — no prior findings to reconcile.

The PR does what it claims, and the parts that matter most check out:
- **The 5500 fix is correct, and for the right reason.** Spec order in `bootstrap_test.go`: Phase 1 funds bank 10000 and posts alice 3000 + bob 2000 (171-180), full backup #1 (201), eve 500 (219-222), incremental (225), **full backup #2 (238)**. Phase 2 restores from the single manifest key that full backup #2 overwrote, so eve is inside the restored checkpoint — consistent with the pre-existing eve assertion at 379-383. `bankOutput = 5500` at line 370 is right, and `charlieTransfer` is correctly excluded because charlie is posted after the balance assertions and no later spec re-asserts bank volumes.
- **The pterm reasoning is technically accurate.** `DisableStyling()` does set `RawOutput = true` (pterm@v0.12.82 pterm.go:61-63), and `SpinnerPrinter.Stop` returns at spinner_printer.go:184 before clearing `IsActive`, leaving the goroutine at :158 spinning. The old code genuinely had that leak. Moving the config out of a spec `BeforeAll` into `TestCluster` is a strict improvement.
- **The feature-gated suites genuinely execute, not just compile.** Coverage delta versus the two most recent `release/v3.0` baselines: business 32.1/32.2% -> 33.8%, cluster 25.0/26.0% -> 31.2%. Databricks is compile-only, which the justfile comment and the docs both state plainly.
- **The Azurite version skip is a real necessity, not a workaround.** `azblob v1.7.0` hard-codes `ServiceVersion = "2026-04-06"`; Azurite 3.36.0 predates it, so the pin alone cannot fix the 400 InvalidHeaderValue. The skipped check only validated emulator-vs-SDK alignment, never a production contract.
- **The `s3CredsGRPCPort` fix is correct.** 17200 derives Raft 16200, and none of 16100/17200/16200 appears as any other business-suite node's HTTP, gRPC or Raft port. I enumerated all 12 business and all cluster call sites; the guard's three-entry `reserved` map {15200, 15100, 14100} is complete for the shared node, and no caller is falsely rejected.
- **The MinIO/Antithesis claim is true** (`RELEASE.2024-06-13T22-53-53Z` matches all three Antithesis references), all 10 MinIO/Azurite call sites route through the new constants, and no `:latest` remains under `tests/`.
- `-timeout 20m` is not at risk: the step took 7m04s at this head, with the cluster suite at 231s inside a 199-242s baseline spread.

One blocking finding, posted inline.

### Counts
- Posted inline: **1** (1x P2 blocking)
- P3 non-blocking, withheld from inline posting: **6** (listed below)
- Entered verification: 4. Confirmed valid: 1. Corrected (`adjust`): 2. Rejected (`invalid`): 1.

### Rejected during verification (not posted)
- *`pterm.DisableOutput()` makes the ledgerctl text-mode assertion vacuous* — **invalid**. The mechanism is real (`Output=false` empties the capture), but `Expect(output).To(Or(BeEmpty(), Not(ContainSubstring("FAIL"))))` was already a tautology: the literal `FAIL` is unreachable in `ledgers get-schema` output (pterm's prefixes are INFO/WARNING/SUCCESS/ERROR/FATAL/DEBUG; the only `pterm.Red("FAIL")` is in `cmd/ledgerctl/audit/list.go:171`, a command the test root does not register), and error paths are caught by the `Expect(err).To(Succeed())` on the preceding line. The PR converts one always-true branch into another and removes no detection capability.

### P3 (not posted inline)
- `tests/e2e/cluster/events_sinks_clickhouse_test.go:37` — pin the ClickHouse image too. This PR is what first puts `//go:build e2e && clickhouse` on a required check, and it pulls `clickhouse/clickhouse-server:24-alpine`, a mutable series tag. `containers.go:4-5` states the rule ("Mutable latest tags would let an external release break that gate"); two of the three gate images now follow it. Downgraded from P2 to P3: `24-alpine` can only move within a superseded 24.x series, so the practical risk is low.
- `tests/e2e/cluster/suite_test.go:16` — the comment's last clause is wrong. `cmd/ledgerctl/cmdutil/pterm.go:11-17` sets `pterm.RawOutput = true` whenever stdout is not a terminal, that package is linked into this test binary, and `go test` never hands the binary a terminal (verified for single-package, multi-package, `-v`, and under a pty). So the resulting state is `RawOutput=true, Output=false`: styling is not kept enabled and the spinner goroutine still leaks once per spinner start. The API choice is right and the code needs no change — only "preserving the normal spinner lifecycle" is inaccurate. No behavioural regression (the removed `DisableStyling()` set `RawOutput` too), and the predicted `-race` failure is unreachable through any command the repo runs.
- `docs/technical/contributing/testing.md:523` — the `nats` row says "NATS (Testcontainers)", but `events_sinks_nats_test.go:38-53` starts an in-process `natsserver.NewServer` with JetStream and tears it down with `ns.Shutdown()`. It is the one feature-gated suite that needs no Docker, and the table is the authoritative answer to "what must I install for tag X".
- `tests/e2e/testutil/server.go:346` — the shared node is exempted by matching its port pair rather than by caller identity. Latent both ways: an auxiliary business spec passing `(TestSingleHTTPPort, TestSingleGRPCPort)` would take the exemption and hit the exact opaque bind panic the guard exists to prevent, while the cluster suite starts no shared node (no `SynchronizedBeforeSuite` in its `suite_test.go`), so those three values are genuinely free there. No caller triggers either case today.
- `tests/e2e/testutil/server_test.go:44` — the "gRPC port collides" case pairs `grpcPort: TestSingleGRPCPort` with `raftPort: 17000`, but the production call site always passes `grpcPort-1000` (14100 here), so that tuple is unreachable from `SetupSingleNode`. Separately, exactly two gRPC values reach the Raft branch without being caught earlier: 16200 (covered) and 16100 (not) — and 16100 is the live value of `s3CredsHTTPPort`, so transposing the two constants is the plausible next mistake.
- `tests/e2e/cluster/bootstrap_test.go:53` — "adding or changing a posting cannot leave a stale expectation behind" holds for *changing* an amount but not for *adding* one: `bankOutput` at line 65 is a hand-curated sum, so a new `bank -> dave` posting before the backup still needs a separate manual edit — which is exactly the EN-1780 regression class. Either soften the claim to what constants actually buy, or derive both sides from one list the spec posts in a loop and sums.

### Notes
- `rolling_config_test.go:43` is genuinely flaky — `Should(BeFollower())` on a just-restarted node can never flip inside its 30s window if that node wins the election — but it is plain `//go:build e2e` and was already in this gate, so this PR does not make it worse. Worth its own ticket.
- Outside this PR's scope, but adjacent to its purpose: `internal/infra/coldstorage/s3_test.go:39` uses `minio/minio:latest` and is `//go:build s3`, which no CI job compiles (`Tests` runs `./...` with no tags, and no workflow runs `just test-full`).

```
DECISION: REQUEST CHANGES
Head reviewed: 43b64b69422055968bbb72f470e48c803f6717e7
Blocking findings: 1
Non-blocking findings: 6
Validation considered: Tests-E2E job at this head (SUCCESS, 7m04s) vs baselines on release/v3.0; all 14 PR checks green; local go test/go list over every recipe tag set; go test -overlay proof that the new test compiles untagged; pterm v0.12.82 and azblob v1.7.0 module sources; full port-constant enumeration across tests/e2e
Residual risk: LOW
```

Models: Claude (3 area reviewers + 3 adversarial verifiers) and Codex. Codex independently raised the same blocking finding.

tests/e2e/testutil/server_test.go
paul-nicolas: Fixed in d6976b6d7 — took the delete, for the reason you gave: it puts the test in the fast `Tests` job and under `just lint`'s type-checker instead of behind the 20-minute e2e gate, and it needs no change to three recipes that would have to stay in sync.

Reproduced every step of your analysis before changing anything. The package's non-test files carry no build constraint, the test imports only `testing` and `testify/require`, and `nodePortConflict` (server.go:345) is pure — so the constraint bought nothing. Confirmed `go test -race -run '^TestNodePortConflict$' ./tests/e2e/testutil/` now passes 6/6 untagged, and golangci-lint reports 0 issues with the file in scope.

One thing your Impact paragraph made me check, which turned out worse than expected. You wrote that an entry dropped from the `reserved` map at server.go:350-354 stays invisible. I mutation-tested all three entries by deleting each and re-running — and after the delete alone, one was *still* invisible:

```
DETECTED   : drop TestSingleHTTPPort
DETECTED   : drop TestSingleGRPCPort
UNDETECTED : drop TestSingleGRPCPort - 1000   <- the shared node's derived Raft port
```

No case passed 14100 as any port value, so the entry protecting the shared node's Raft listener had no coverage. Added that case; all three now report DETECTED.

Also folded in the P3s from the review body:

- **server_test.go:44** — `raftPort` is now `TestSingleGRPCPort - 1000`, matching what `SetupSingleNode` derives, so the tuple is reachable from production. Added the 16100 case; as you noted it is the live value of `s3CredsHTTPPort`, which makes transposing that spec's two constants the plausible next collision.
- **ClickHouse image** — pinned via a new `testutil.ClickHouseImage`. Worth flagging: my first attempt used `24.8.14.39-alpine`, which would have quietly downgraded the gate. `24-alpine` resolves to `v24.10.2.80-stable` (read from the image's own `com.clickhouse.build.version` label), so I pinned `24.10.2.80-alpine` instead — identical amd64 manifest, so the pin changes nothing CI already runs.
- **cluster/suite_test.go:16** — reworded. Confirmed your mechanism with `go list -deps -test`: `cmd/ledgerctl/cmdutil` is linked into this binary, so its init sets `RawOutput` and the spinner-lifecycle clause was wrong. No code change; `DisableOutput` is still the right call.
- **bootstrap_test.go:53** — softened to what the constants actually buy. Left Phase 1 as it is rather than restructuring it into a posted-in-a-loop list; that felt out of scope here.
- **testing.md:523** — `nats` now reads "none (embedded in-process server)".

**Deferred: server.go:346**, the port-pair exemption. Your own analysis concludes it is latent both ways with no caller triggering it today, and I confirmed the cluster suite starts no shared node (no `SynchronizedBeforeSuite` in its `suite_test.go`). Fixing it properly means threading caller identity into the guard, which is a design change rather than a fix. Left alone deliberately.

The two "Notes" items are untouched and want their own tickets: `rolling_config_test.go:43` (already in this gate, not made worse here) and `coldstorage/s3_test.go:39` (`minio:latest` under a tag no CI job compiles).

`bash scripts/agent-check` passes. I did not run the 20-minute full-tag e2e suite locally, so the ClickHouse pin is verified by manifest identity rather than by a local suite run — CI's `Tests-E2E` job is the actual check on it.

paul-nicolas: Criterion met. `Tests` job log, same package line either side of the fix:

```
43b64b694  ?   github.com/formancehq/ledger/v3/tests/e2e/testutil   [no test files]
d6976b6d7  ok  github.com/formancehq/ledger/v3/tests/e2e/testutil   1.884s  coverage: 0.2% ...
```

Resolving.

gfyrag (approved):

PR #1713, opened 2026-08-17
chore: mechanically enforce low-noise repository invariants

gfyrag:
## Summary

This PR converts a small set of existing high-confidence repository rules into deterministic checks.

- adds `scripts/check-repo-invariants`
- runs it from the canonical `scripts/agent-check` baseline
- adds a lightweight CI workflow so the same rules apply to every contribution, not only agent-authored changes

## Enforced invariants

The checker intentionally stays narrow and low-noise:

1. **No `time.Sleep` in Go tests** — use deterministic synchronization or `require.Eventually` instead.
2. **No direct environment reads from deterministic FSM paths** — production code under `internal/infra/state`, `internal/infra/plan`, `internal/infra/preload`, and `internal/domain/processing` may not call `os.Getenv`, `os.LookupEnv`, or `os.Environ`; node-local policy belongs on the admission side of the boundary documented in the FSM docs.
3. **No protobuf `reserved` declarations under `misc/proto`** — v3 is unreleased and `AGENTS.md` explicitly requires field removal/renumbering without reintroducing reserved declarations.

The script inspects both tracked files and non-ignored untracked files so the local agent handoff check cannot miss newly-created files.

## Already enforced elsewhere

This PR deliberately does **not** duplicate existing `.golangci.yaml` enforcement for `OpenWriteSession` / `NewWriteSessionFromDB` and related storage capabilities.

## Why

These rules are deterministic enough that asking agents or human reviewers to remember them adds cognitive load without adding useful judgment. The repository should reject violations automatically and explain the reason.

This is intentionally not a broad architecture linter. More semantic invariants (checker coverage, audit projections, coverage/preload correctness, etc.) remain out of scope until they can be encoded with similarly low false-positive rates.

## Review focus

Please check primarily for:

- false-positive risk in the three patterns
- whether the selected FSM directories correctly represent the deterministic apply boundary
- whether tracked + non-ignored untracked file enumeration is sufficient
- whether a dedicated lightweight workflow is preferable to folding this into the existing large CI workflow


scripts/check-repo-invariants
```diff
@@ -0,0 +1,61 @@
...
+}
+
+# Inspect every file that could be part of the handoff: tracked files plus
+# non-ignored untracked files. Keep the checks deliberately narrow so this
+# script remains a low-noise architectural guardrail rather than a style linter.
+while IFS= read -r -d '' file; do
+    [[ -f "$file" ]] || continue
+
+    case "$file" in
+        *_test.go)
```
NumaryBot: 🔴 **[blocker] Remove existing sleeps before enabling this check**

This check makes the new workflow and `scripts/agent-check` fail immediately on the current HEAD: five tracked tests already call `time.Sleep`, including `internal/infra/state/wait_for_applied_test.go:51`. Convert or exempt those existing calls before making this repository rule mandatory (`AGENTS.md:67`).

scripts/check-repo-invariants
```diff
@@ -0,0 +1,61 @@
...
+            report_matches \
+                "tests must not use time.Sleep; use deterministic synchronization or require.Eventually" \
+                'time\.Sleep[[:space:]]*\(' \
+                "$file"
+            ;;
+    esac
+
+    case "$file" in
```
NumaryBot: 🟠 **[major] Check nested deterministic-path packages recursively**

These shell patterns match only files directly inside the listed directories, so production subpackages such as `internal/domain/processing/numscript/*.go` and `internal/infra/plan/planerr/*.go` can read environment state without detection. Since node-local environment state is forbidden throughout the FSM apply path (`AGENTS.md:34`), the directory matching needs to include descendants.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1713#issuecomment-5315685612

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1713#issuecomment-5315685612

internal/infra/node/applier_async_storage_test.go
```diff
@@ -311,9 +311,12 @@ func TestApplierRunCommitterDoesNotDeadlockOnFullSink(t *testing.T) {
...
 
-	// Give runCommitter time to fire the response — it will block on the
-	// unbuffered send. If the deadlock regressed, the test hangs here.
-	time.Sleep(200 * time.Millisecond)
+	// CommitPreparedBatch publishes the ledger immediately before runCommitter
+	// tries the response send. Once the ledger is visible, the unbuffered sink
+	// guarantees that runCommitter is blocked in fireResponses until stop closes.
+	require.Eventually(t, func() bool {
+		return listLedgerContains(setup.store, "async-full-sink")
+	}, 3*time.Second, 10*time.Millisecond, "entry should commit before the response send blocks")
```
gfyrag: 🟠 **[major] Wait until the response send is actually blocked before closing `stop`**

`listLedgerContains(...)` only proves that `CommitPreparedBatch` has completed. In `runCommitter`, the response send happens *after* that commit. There is therefore a valid window where this condition becomes true, the test closes `stop`, and `fireResponses` observes an already-closed stop channel without ever blocking on the unbuffered sink.

That means the rewritten test can pass without exercising the original regression scenario: `runCommitter` already blocked on the response send when shutdown begins. Please use a deterministic signal that proves the send path has been entered/blocked (for example an instrumented test sink/hook) before closing `stop`, rather than using ledger visibility as a proxy.

gfyrag (commented):
One important regression-test concern remains. The mechanical invariant work itself looks good, but this test rewrite weakens the shutdown/deadlock scenario it was protecting.

paul-nicolas (dismissed):

paul-nicolas (approved):

PR #1717, opened 2026-08-18
chore: add bounded AI review and fix loop

gfyrag:
## Summary

Add a provider-agnostic local orchestrator for bounded AI review → fix → re-review loops.

- add `bash scripts/review-loop` running in the repository-pinned Nix environment
- add a small Go state machine under `scripts/reviewloop/`
- consume a strict reviewer JSON contract and reject inconsistent results
- auto-fix only when every blocking finding is explicitly marked `auto_fixable`
- run `bash scripts/agent-check` after every successful fix before re-review
- cap the loop at 3 review passes by default
- escalate product/design/invariant ambiguity and non-auto-fixable blockers to humans
- persist each review/fix payload under `build/ai-review-loop/` for inspectability
- document the reviewer/fixer environment contract and exit codes

## Why

The repository now standardizes author handoff and AI review output. This PR adds the missing orchestration layer while deliberately keeping GitHub writes out of scope. The goal is to prove a deterministic, inspectable local state machine before allowing bots to comment, push, resolve threads, or merge.

## Safety boundaries

The loop never decides product behavior. `auto_fixable` is reserved for local/reversible fixes whose intended outcome is already determined by repository evidence. Any non-auto-fixable blocker, explicit human-decision result, validation failure, or pass-budget exhaustion stops/escalates the loop.

## Validation

The new Go package is covered by root-module build/tests. Unit tests cover approval, auto-fix selection, human escalation, inconsistent review decisions, and enum/schema validation.

## Review focus

Please focus on:

- whether `auto_fixable` is a sufficiently safe boundary for autonomous fixes
- state/exit-code semantics and maximum-pass behavior
- JSON contract strictness vs. flexibility for different reviewers
- whether validation occurs at the correct point between fix and re-review
- whether any GitHub write behavior has accidentally leaked into this local-only layer


scripts/reviewloop/main.go
```diff
@@ -0,0 +1,245 @@
...
+		return "", nil, fmt.Errorf("unknown review decision %q", result.Decision)
+	}
+}
+
+func loadReviewResult(path string) (reviewResult, error) {
+	content, err := os.ReadFile(path)
+	if err != nil {
+		return reviewResult{}, fmt.Errorf("review command did not produce %s: %w", path, err)
+	}
+	var result reviewResult
```
NumaryBot, line 223: 🔴 **[blocker] Require explicit blocking flags before approving**

When a reviewer omits or misspells `blocking`, Go silently defaults it to `false`, so an `APPROVE` result containing a real blocking finding is accepted as ready. This undermines the documented strict reviewer contract and inconsistency check; decode required booleans distinctly from false and reject unknown fields. See [reviewer contract](docs/technical/contributing/ai-review-loop.md#L33-L58).

scripts/reviewloop/main.go
```diff
@@ -0,0 +1,245 @@
...
+	if err := json.Unmarshal(content, &result); err != nil {
+		return reviewResult{}, fmt.Errorf("decoding review result %s: %w", path, err)
+	}
+	if result.Head == "" {
```
NumaryBot: 🔴 **[blocker] Verify the reported head before accepting approval**

The loader only checks that `head` is nonempty, so a cached or stale reviewer payload can approve a different revision and produce `READY_FOR_HUMAN_REVIEW` without reviewing the current HEAD. Compare the reported SHA with the revision passed to the reviewer/current repository state before accepting the result.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1717#issuecomment-5325962291

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1717#issuecomment-5325962291

gfyrag (commented):
One additional correctness concern on the worktree fingerprint exclusion logic.

scripts/reviewloop/main.go
```diff
@@ -0,0 +1,457 @@
...
+	}
+	if maxPasses < 1 {
+		fatal(errors.New("--max-passes must be at least 1"))
+	}
+	repositoryRoot, err := gitRepositoryRoot()
+	if err != nil {
+		fatal(err)
+	}
+	runStateDir, err := createRunStateDir(stateDir)
+	if err != nil {
```
gfyrag, line 86: 🟠 **[P2][blocking] Exclude only the per-run state directory from the fingerprint**

`stateParent := filepath.Dir(runStateDir)` means `captureWorkspaceState(..., stateParent)` ignores the entire parent directory, not just this loop run. With the default `build/ai-review-loop/run-*` this happens to be reasonable, but `--state-dir` is configurable. For example, `--state-dir .` creates `/repo/run-*`, making `stateParent` equal to the repository root; every untracked file in the repo is then excluded from `AI_REVIEW_WORKTREE_FINGERPRINT`.

That weakens the new guarantee that approvals are bound to all tracked changes plus non-ignored untracked contents. Please exclude only `runStateDir` (or otherwise reject unsafe state-dir placements) and add a regression test showing that an untracked source file still changes the fingerprint when the state directory is near the repository root.

paul-nicolas (approved):

PR #1722, opened 2026-08-18
fix(EN-1784): allocate e2e node ports and fail startup on bind errors

paul-nicolas:
https://formance-team.atlassian.net/browse/EN-1784

## What

Two defects. `tests/e2e/business` panicked with `listen tcp4 0.0.0.0:15200: bind: address already in use`, and the panic destroyed the report of the spec that was actually failing.

**G1 — hand-picked ports collided.** `backup_s3_credentials_test.go` chose gRPC `16200`; `SetupSingleNode` derived its Raft port as `grpcPort - 1000` = `15200`, which is the HTTP port of the node `SynchronizedBeforeSuite` keeps alive for the whole business suite. Deterministic, not a race. Two further duplicates were latent: `15702/15802` (`check_store` vs `get_transaction_cold`) and `15810/15910` (`auth_writes_only` vs `ed25519_auth`).

Ports now come from `testserver.AllocateNodePorts`. `NodePorts` has unexported fields and `TestNodeConfig.Ports` replaces the three `int` fields, so the setup helpers every suite goes through cannot be handed a chosen port. The `- 1000` derivation and its six local copies are gone. The allocator bind-probes each candidate and gives each process its own sub-band of 15000-30000, below the OS ephemeral floor.

Scope note: the five `restore_*` cluster specs build their own instrument lists and still call the low-level `testserver.WithHTTPPort(ports.HTTP())` etc., which take a plain `int`. Those remain the one place a literal would still compile, so this is a strong convention plus a type-enforced main path, not a total compile-time ban.

**G2 — a bind failure panicked a goroutine.** `baseServer.Start` is split into `Listen` (synchronous bind, returns the error) and `Serve`; the three duplicated fx hooks collapse into `grpcServerHook`, whose `OnStart` returns the wrapped error. A node with an occupied port now fails boot with a clear error instead of a panic stack.

## Review findings fixed in this PR

Splitting `Start` introduced a regression that review caught: `Start` passed `net.Listen`'s result to `serveSingle`/`serveDual` as a local, whereas `Serve` re-read `s.listener`, which `closeListener` nils. fx's `withRollback` calls `lifecycle.Stop` immediately on the same goroutine when any `OnStart` errors, so the freshly spawned serve goroutine is still on the local runq and `Serve` saw `nil`. It returned `"Serve called before a successful Listen"`, which `OnStop` surfaced — so on a port collision the operator got the real bind error *plus* a fabricated invariant-looking error naming the wrong server. Reproduced at 2000/2000 iterations for a same-tick stop.

Fixed with a `listened` flag that `Listen` sets and `closeListener` never clears: a genuine contract violation stays loud, the benign shutdown race returns nil.

## Out of scope

- The `testservice.Stop` budget window, where `Stop` can return while the `OnStop` listener-close hooks are still running.
- `node.Run`'s `panic(err)` in `module.go`, which has the same shape as the bug fixed here but cannot take the same fix (there is no bind to split off; it needs an error channel).
- CI coverage of the tagged suites. CI runs `-tags e2e`, so it cannot compile the spec this bug lives in.

## Verification

- `bash scripts/agent-check` — PASS, 0 lint issues in both modules.
- `go test ./... -race` — 82 packages ok, 0 failures.
- `go vet` clean under `e2e,kafka,nats,clickhouse,databricks,s3,azure,pyroscope` and under `scenario`.
- `go test -race -tags e2e -p 1 ./tests/e2e/business/... ./tests/e2e/cluster/...` — pass, zero `bind: address already in use`. Business suite: `Ran 556 of 556 Specs`, `SUCCESS! -- 556 Passed | 0 Failed`.
- `go test -race -tags scenario ./tests/scenarios/...` — 8 packages ok. This recipe has no `-p 1`, so its packages run concurrently in separate processes; it is what the per-process port sub-band exists for.

**Not executed locally:** the four `s3`/`azure`-tagged business specs, including `backup_s3_credentials_test.go` — the spec that originally panicked. They need a MinIO testcontainer and no Docker daemon was available on the machine this branch was validated on. They are compile-verified (`go vet -tags "e2e,s3"` clean) and contain no hardcoded ports, but they have not been run. `just test-e2e-full` on a Docker host is still worth doing before merge.


paul-nicolas (commented):

Reviewed at head `07e28cea5`. Four reviewers (Codex, plus three scoped Claude reviewers for the gRPC lifecycle, the port allocator, and the ~60-file suite migration), then one adversarial verifier per finding instructed to refute rather than confirm.

The change is a clear net improvement: the synchronous `Listen()` error is strictly better than the old `panic(err)` on a serve goroutine, the `Allocated()` guard genuinely closes the hand-picked-port path for every `DefaultTestInstruments` caller, restart port identity is preserved everywhere, and the `grpcPort - 1000` derivation is gone with no `Port ± 1000` expression left in `tests/` or `pkg/testserver/`.

**Verification outcome:** 4 findings entered adversarial verification, 0 confirmed as filed, 4 corrected, 0 rejected outright. Two filed as blocking were demoted to non-blocking P3 on evidence:

- *"Tagged specs were never compiled"* — false. The PR body's own `go vet` bullet type-checks all 13 under every tag (reproduced clean), and the "Out of scope" section already states the CI tag gap. The "bypasses the `Allocated()` guard" sub-claim was wrong: all six s3-tagged cluster files go through `DefaultTestInstruments`. Demoted to P3.
- *"Post-bind `Serve` failure is an availability regression"* (Codex rated this P1 blocking) — no realistic post-bind error reaches `serveErr`. grpc-go retries every temporary `Accept` error and `serveSingle`/`serveDual` tolerate the rest; grpc-go also closes the listener when `Serve` returns, so the port is released rather than black-holed. Restoring `panic` would re-break the Ginkgo report this PR exists to fix. Demoted to P3 (observability only).

Codex additionally asked for cross-process port coordination as a blocker. Rejected: a 10-run measurement showed `go test` launches package binaries back to back with consecutive pids, so sibling binaries land in distinct buckets; `test-e2e` also runs `-p 1`. Only the documentation half of that concern survives, as the second inline comment.

**Counts:** 2 posted inline (both P2, both non-blocking). 4 P3 withheld from inline. 0 blocking.

### P3 (not posted inline)

- `pkg/testserver/ports_test.go:18` — the load-bearing property, cross-process disjointness, has no test; `TestPortAllocatorNeverReturnsTheSamePortTwice` would still pass if `processPortBase` were deleted.
- `internal/bootstrap/grpc_server_hook.go:59` — a post-bind `Serve` failure is assigned to `serveErr` and read only at `:95`, so it produces no output until shutdown. No `fx.Shutdowner` exists and `/readyz` only checks `Node.IsStarted()`. Suggested fix is a log line in the goroutine, explicitly **not** a restored `panic`.
- `internal/bootstrap/grpc_server_hook.go:53` — `OnStart` ignores the fx start context, so on a start-timeout landing between `Listen()` and `return nil`, `AfterListen` (`membership.Start`) runs after fx has abandoned the hook, and `OnStop` never runs to tear the listener down.
- `tests/e2e/cluster/restore_test.go:98` — the "Not executed locally" paragraph names only the four business specs; nine touched cluster specs are equally unexecuted (six `e2e && s3`, plus `events_sinks_{clickhouse,nats,databricks}`). All are compile-verified and the prescribed `just test-e2e-full` already covers both directories, so this is a completeness edit to the description. No spec count is reported for `tests/e2e/cluster` — `Ran 556 of 556` is the business suite.

### Checked and found sound

Restart port identity across every stop/restart path; peer/join/gateway address wiring; HTTP vs gRPC vs Raft role assignment across all migrated call sites; the zero-value `NodePorts` hazard (closed by the panic at `pkg/testserver/testserver.go:31`); `membership.Start` ordering equivalence with the old `<-listening` gate; data races on `serveErr`/`listener`/`listened`; goroutine leaks and shutdown deadlock; band exhaustion (~327 allocations against 2400 candidates in the worst bucket — not reachable). `go test -race` on `pkg/testserver`, `internal/bootstrap` and `internal/adapter/grpc` passes.

```text
DECISION: APPROVE
Head reviewed: 07e28cea5c0c8df3d37113e280f41c01ef6f9834
Blocking findings: 0
Non-blocking findings: 6 (2 P2 posted inline, 4 P3 listed above)
Validation considered: CI all green including Tests-E2E/Tests-Scenarios/invariants; go test -race on the three touched packages; go vet under all build tags
Residual risk: LOW
```

pkg/testserver/ports.go
```diff
@@ -0,0 +1,153 @@
...
+	// portAllocatorBase is the first candidate port. It sits below the OS
+	// ephemeral range (Linux 32768+, macOS 49152+) so an allocated port cannot
+	// be taken by an outbound connection in the window between the probe in
+	// allocate and the moment the node actually binds it.
+	portAllocatorBase = 15000
+
+	// portAllocatorCeiling bounds the walk. Without a ceiling an exhausted
+	// range would spin past 65535 forever, where every net.Listen fails and
+	// every failure is skipped.
+	portAllocatorCeiling = 30000
```
paul-nicolas: [P2][non-blocking] Move the phantom peer addresses out of the new allocator band

Location: pkg/testserver/ports.go:20 (portAllocatorCeiling) — affects tests/e2e/cluster/learner_test.go:53,163,214,297,306,318 and tests/e2e/cluster/remove_node_test.go:103,348

Evidence: Widening the band to [15000,30000] (ports.go:15,20) with processPortBase() = 15000 + (pid%64)*200 (ports.go:97) pulls sixteen pre-existing dead-peer literals inside it: 127.0.0.1:19000/19001/19002 and :19100/19101/19102, registered through AddLearner. Before this PR every real e2e port was a literal in 14500-16400, so 19xxx was structurally free; the literals are untouched by the diff, but the diff moves the band onto them. 15000 + 20*200 = 19000 exactly, so a tests/e2e/cluster binary with pid%64==20 hands out http=19000, grpc=19001, raft=19002 as its first AllocateNodePorts(); bucket 19 (base 18800) walks onto the same window around allocation 200-302 of ~327 that the package performs. The bind probe cannot prevent this — the literal is a consumer of the port, not a binder, so nothing is bound at probe time. Nothing rejects the misdirected dial either: internal/infra/node/transport.go:596-598 validates only the ClusterID, which every e2e node shares as "test-cluster", and StreamMessages never checks msg.To.

Impact: In roughly 2/64 of runs a live node holds an address the leader also has registered as a phantom learner, so the specs asserting "a phantom learner never catches up" stop exercising that path. The usual symptom is a silently weakened test rather than a red build (in the most likely case the phantom points at node 0's HTTP port and the gRPC dial fails the HTTP/2 preface, so the assertions still pass); the rarer variant at remove_node_test.go:348 points :19002 at a live Raft port of the same cluster. It also contradicts this PR's own new guidance in docs/technical/contributing/testing.md ("Do not write a port number in a test") — AddLearnerRequest.RaftAddress/ServiceAddress are plain strings and bypass the NodePorts unexported-field guard entirely.

Resolution: The phantom addresses come from the allocator, or move to a named constant provably outside [portAllocatorBase, portAllocatorCeiling], such that no test-owned literal in [15000,30000] remains under tests/e2e/**.

docs/technical/contributing/testing.md
```diff
@@ -94,34 +94,65 @@ var _ = Describe("Simple cluster", func() {
...
+  old Raft address.
+- **Never compute one port from another.** The old convention derived the Raft
+  port as `grpcPort - 1000`. That is what made two suites collide on 15200:
+  one spec's gRPC port mapped onto another node's live HTTP port (EN-1784).
+
+A spec that builds its own instrument list instead of using
+`DefaultTestInstruments` still passes allocated values through the low-level
+instruments — `testserver.WithHTTPPort(ports.HTTP())`,
+`WithRaftPort(ports.Raft())`, `WithGRPCPort(ports.GRPC())`. Those take a plain
+`int`, so they are the one remaining place a literal would compile. Feed them
```
paul-nicolas: [P2][non-blocking] Name every API that still accepts a literal port, or drop "the one remaining place"

Location: docs/technical/contributing/testing.md:146

Evidence: The paragraph tells contributors that WithHTTPPort/WithRaftPort/WithGRPCPort "take a plain `int`, so they are the one remaining place a literal would compile". That is an exhaustiveness claim and it is false. testserver.WithJoin(raftAddr string) (pkg/testserver/testserver.go:194) appends its argument straight to --join, and testserver.NewGateway(logger, ports []int, nodes []string) (pkg/testserver/gateway.go:56) takes raw ints; a literal compiles in both. Every current call site is allocator-fed (tests/e2e/cluster/join_auth_test.go:65, restore_test.go:865, tls_test.go:219, tests/e2e/testutil/server.go:249), so nothing is broken through these two APIs today. The AddLearner addresses in the companion finding on pkg/testserver/ports.go are the case where a literal does reach a live port.

Impact: WithJoin carries a Raft address, which is exactly the value whose hand-picking caused EN-1784. A contributor who trusts this sentence has an incomplete inventory of the ways a hand-picked port can still enter a test node, which is how in-band literals get added.

Resolution: The sentence names WithJoin and NewGateway alongside the three instruments, or drops the "one remaining place" phrasing.

pkg/testserver/ports.go
paul-nicolas: Fixed in fbbbb2655 — the phantom peer addresses now come from `AllocatePort` instead of literals.

Added `phantomPeer()` beside `waitForLearner` in `tests/e2e/cluster/learner_test.go` and routed all seven `AddLearner` call sites through it (5 in `learner_test.go`, 2 in `remove_node_test.go`). `AllocatePort` reserves the number without binding it, which is what a peer that never starts needs, and it guarantees no live node in the process can be handed the same port.

Two things worth noting from the fix:

- No assertion depended on the literal values — they appear only as `AddLearner` request fields, and `waitForLearner` matches on node id plus `Suffrage`. So the swap is behaviour-preserving.
- The "should reject duplicate add-learner" spec keeps a *single* allocated pair across both of its calls. Allocating twice there would have quietly changed what it tests.

Verified: `grep -rn "127.0.0.1:19" tests/` returns nothing, and no test-owned port literal remains anywhere in `[15000,30000]`.

Validation: `bash scripts/agent-check` PASS (0 lint issues, no codegen drift); `go test -v -race -tags e2e -p 1 ./tests/e2e/cluster/... -ginkgo.focus="Learner node|Remove node"` → `Ran 28 of 266 Specs`, `SUCCESS! -- 28 Passed | 0 Failed`.

docs/technical/contributing/testing.md
paul-nicolas: Fixed in aa9b61066 — dropped the exhaustiveness claim and named the other surfaces.

The paragraph now says a literal would still compile in the three low-level instruments, and equally in `testserver.WithJoin(raftAddr)`, `testserver.NewGateway(logger, ports, nodes)`, or any RPC field carrying an address (`AddLearnerRequest.RaftAddress` is a plain string), and states that the `NodePorts` guard reaches none of them.

Also added the rule the companion fix follows, since it was the gap that let the in-band literals in: a peer that is deliberately never started still needs an allocated address, because a hand-picked one is a number inside the allocator's band and can name a port the allocator later hands to a live node.

gfyrag:
Adversarial follow-up on head `aa9b61066`. I think this PR is a net improvement and I am not treating the points below as blockers, but two guarantees are currently stronger in the prose than in the implementation.

### 1. `AllocatePort` probes a port; it does not reserve it

`portAllocator.allocate` binds a candidate and closes the listener before returning the number. There is therefore still a TOCTOU window between allocation and the server's actual bind. `processPortBase` reduces the probability across test binaries, but it is not coordination:

- PIDs congruent modulo 64 start in the same bucket;
- a process that consumes more than the 200-port stride deliberately walks into the next bucket;
- the Linux ephemeral range is configurable, so `[15000,30000]` is not universally below the OS allocator.

The current canonical recipes mitigate this (`test-e2e` uses `-p 1`, and scenario binaries are few), so this may work well in practice. Still, names/comments such as “reserves” and “keeps different processes apart” describe a guarantee the implementation does not provide. The unit tests also cover uniqueness only within one allocator/process; they would still pass if `processPortBase` disappeared.

### 2. A post-bind `Serve` error is deferred until shutdown

In `grpcServerHook`, the serve goroutine only assigns `serveErr`; that error is returned from `OnStop`. If `Serve` exits unexpectedly after startup, the application can remain alive with a dead Raft or service gRPC endpoint, and `/readyz` remains based on `Node.IsStarted()`. The fake-server test demonstrates this shape: an immediate `Serve` error does not make `OnStart` fail.

I agree that restoring the previous panic would be the wrong fix. At minimum the goroutine should log the error immediately; ideally it should request an Fx shutdown or feed a runtime health signal.

### Alternative: reserve listeners, not port numbers

The cleaner model is to bind `127.0.0.1:0`, keep each `net.Listener` open, derive the advertised ports from `listener.Addr()`, and inject the listeners into the application. That removes the fixed range, PID buckets, bind probe, and allocation/bind race.

Cobra does not need to carry sockets through flags or context. `testservice.New` already accepts a command factory, so the factory can capture explicit runtime options:

```go
listeners := testserver.MustAllocateNodeListeners()

service := testservice.New(
    func() *cobra.Command {
        return cmdserver.NewRunCommandWithOptions(
            cmdserver.WithNetworkBindings(listeners),
        )
    },
    testservice.WithInstruments(
        testserver.WithHTTPPort(listeners.HTTPPort()),
        testserver.WithGRPCPort(listeners.GRPCPort()),
        testserver.WithRaftPort(listeners.RaftPort()),
    ),
)
```

`NewRunCommand()` can remain unchanged and delegate to an option-aware constructor:

```go
func NewRunCommand() *cobra.Command {
    return NewRunCommandWithOptions()
}
```

`runServer` then injects a typed `NetworkBindings` into Fx. Production uses the zero/default bindings and keeps the normal `net.Listen` path; tests provide pre-bound HTTP, service-gRPC, and Raft listeners. `go-libs` already has `httpserver.WithListener`, so HTTP supports this directly; the gRPC adapter would need the equivalent `WithListener` server option.

The ownership rule should be explicit: once injected, the lifecycle owns and closes the listener; test cleanup closes only unconsumed listeners. For node restarts, listeners are one-shot because gRPC closes them, so the command factory should ask a node lease for a fresh generation rebound to the same advertised addresses rather than capturing a single `NetworkBindings` forever. A stable front proxy would be needed only if we wanted to eliminate even the stop/rebind window, which is probably excessive for these tests.

This is a somewhat deeper change, but the resulting mechanism is smaller conceptually and gives the property the current allocator is trying to approximate.

pkg/testserver/listeners.go
```diff
@@ -0,0 +1,281 @@
...
+// It panics on a rebind failure for the same reason AllocateNodeLease does, and
+// because the caller is a testservice command factory that cannot report an
+// error. A failure here means the node's previous generation is still bound:
+// stop the node before starting it again.
+func (l *NodeLease) NextGeneration() network.Bindings {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+
+	if l.consumed {
+		l.current = network.Bindings{
```
NumaryBot, line 231: 🔴 **[blocker] Release the unused Raft listener after restore mode**

When a lease starts a restore-mode server, `NextGeneration` binds all three listeners, but `RestoreModule` never consumes or closes the Raft listener. After restore shutdown, the next generation therefore panics while rebinding the still-open Raft port. This deterministically breaks the restore tests that run normal → restore → normal phases, such as `restore_idempotency_test.go`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1722#issuecomment-5344258917

paul-nicolas:
Both points accepted, and the listener model is implemented rather than deferred. `a2d23e39c` and `1478d9d41`.

### 1. The allocator is gone

You were right that the names described a guarantee the code did not provide, so rather than rewording them the mechanism changed. `testserver.AllocateNodeLease` binds three loopback listeners up front and keeps them open; `lease.NewService(cmdserver.NewRunCommandWithBindings, ...)` hands those very sockets to the run command, which supplies a `network.Bindings` into fx, where the gRPC adapter adopts them through a new `WithListener` option and the HTTP server through `httpserver.WithListener`. No probe, no fixed 15000-30000 range, no pid buckets, no allocation/bind window. Production supplies the zero value and each server binds its configured port exactly as before.

Deltas from your sketch, all in the same direction:

- **Restart generations.** As you predicted, listeners are one-shot, so the lease is the unit tests hold, not a `NodeListeners` value. `NextGeneration` rebinds the same advertised ports for every start and panics if the previous generation is still bound, i.e. if the node was not stopped. `RestartNode`, `RestartNodeWithInstruments` and `ScenarioCluster.Restart` all go through it.
- **Ownership.** Made explicit in `network.Bindings`: the lifecycle owns what it adopts and closes it on stop. A test fixture that had to remember which listeners the app consumed would leak the ones it forgot, so nothing tracks consumption; the restore-mode Raft listener simply stays held for the test, which is the behaviour you want anyway.
- **Dead peers.** A phantom peer cannot come from `:0`-then-close alone: the kernel can hand that number to a later node, which is the same confusion in a new costume. Ports stay claimed for the process lifetime even once released, so `AllocateDeadAddress` refuses connections and no later allocation reuses the number. `phantomPeer()` is built on it.
- **No options carrier.** `NewRunCommandWithBindings(bindings)` next to `NewRunCommand()`, rather than `NewRunCommandWithOptions(WithNetworkBindings(...))`. Bindings are the only runtime option there is today, and two constructors read better than an option type with one member.
- **`internal/pkg/network`.** `Bindings` cannot live in `internal/bootstrap`: `cert_reloader_test.go` imports `pkg/testserver`, and the fixture has to name the type, so it would be an import cycle in the test binary. It is a leaf package for that reason.
- **Gateway.** `NewGateway` no longer takes port numbers; it binds its own.

The one window left is a node's own stop-to-restart gap, where only an unrelated process could interpose. As you said, closing that means a proxy in front of every node, which these suites do not justify — it is documented as such rather than implied away.

On the test gap: `bindFresh`'s skip is now tested against a chosen sequence of ports (the injected `bind` exists for that), and the lease test asserts the ports are *bound*, not merely distinct — an independent `net.Listen` on each must fail. That is the cross-process property; the old tests could not state it.

### 2. Serve failure

Stronger than non-blocking, and thank you for catching it: the PR had regressed this. The panic it removed did report the failure, by killing the process. What replaced it reported nothing until shutdown. The serve goroutine now logs at error level and calls `fx.Shutdowner.Shutdown` the moment `Serve` returns non-nil; `OnStop` still returns the error. Non-nil is unambiguous — `serveSingle`/`serveDual` swallow `ErrServerStopped` and `net.ErrClosed`, and `Serve` returns nil when `Stop` won the race — so a normal shutdown never requests one.

`RequestShutdown` is required, not optional: a nil requester would silently disable the path, so `OnStart` refuses to bind without one. `fx` broadcasts the signal without blocking and go-libs's `service.App` selects on `app.Wait()` before `app.Stop()`, so it cannot deadlock against the `OnStop` that joins the same goroutine. `docs/ops/deployment.md` now states that no probe covers a dead listener and that the process exits instead.

### Validation

`bash scripts/agent-check` PASS, root unit suite green, `tests/e2e/cluster` 265 of 266 specs (1 skipped) and `tests/e2e/business` 556 of 556 on the committed tree, all 8 scenario packages green — each of those restarts its node through the lease, so the rebind path is exercised. The two new mechanisms were mutation-checked: neutering `WithListener` fails the adoption test, and dropping the shutdown request fails the serve-failure test.

pkg/testserver/listeners.go
paul-nicolas: Confirmed and fixed in `368cc9095`. The analysis was exactly right, including the reason nothing caught it: the restore specs are `//go:build e2e && s3`, and neither the default e2e suite nor the CI e2e job (`just test-e2e-coverage`, tags `e2e`) runs them, so my local and CI runs both stayed green.

Reproduced first, at the level where it can be pinned without Docker or the s3 tag: `TestRestoreModeReleasesEveryInjectedListener` starts the real restore-mode fx app on three injected listeners and requires all three ports to rebind after stop. Before the fix it failed on the Raft port with `bind: address already in use`.

The fix makes ownership total rather than per-adoption: both modules register a release hook first — so it stops last, after the servers — and it closes every listener in the bindings, treating `net.ErrClosed` from one a server already closed as a non-error. A fixture never has to work out which listeners an application actually served, which was the fragile half of the contract.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1722#issuecomment-5344258917

gfyrag (dismissed):

Azorlogh (approved):

PR #1723, opened 2026-08-18
fix: expose Pebble target garbage ratio

gfyrag:
## What changed

- rename the server flag to `--pebble-value-separation-target-garbage-ratio`
- expose `spec.pebble.valueSeparation.targetGarbageRatio` in the operator CRD
- propagate it as `PEBBLE_VALUE_SEPARATION_TARGET_GARBAGE_RATIO`
- regenerate both operator CRD copies and update CLI, deployment, storage, and operator documentation
- add server, operator environment, and `kubectl-ledger explain` coverage

## Why

The internal DAL and Pebble policy use `TargetGarbageRatio`, while the public CLI and operator CRD exposed the shortened `garbageRatio` name. This made the documented `targetGarbageRatio` setting unavailable on Cluster resources and obscured the fact that the threshold applies to aggregate unreferenced blob values.

Ledger v3 is unreleased, so the old field and flag are removed without a compatibility alias.

## Operational impact

Operators can configure blob rewrite pressure with:

```yaml
spec:
  pebble:
    valueSeparation:
      enabled: true
      targetGarbageRatio: "0.10"
```

Changing the value updates the StatefulSet environment and causes the normal rolling restart. Lower values reclaim blob garbage sooner at the cost of additional compaction I/O.

## Validation

- `go test ./cmd/server`
- `go test ./...` in `misc/operator`
- `bash scripts/agent-check`
- generated CRDs checked for removal of the old `garbageRatio` field
- GitNexus change-impact analysis: low risk, no affected execution flows

### DIFF
--- a/cmd/server/server.go
+++ b/cmd/server/server.go
@@ -122,7 +122,7 @@ func NewRunCommand() *cobra.Command {
 	bytesize.ByteSizeVar(runCmd, new(bytesize.ByteSize), "pebble-value-separation-min-size", 256, "Minimum value size for separation (default: 256)")
 	runCmd.Flags().Int("pebble-value-separation-max-depth", 4, "Max blob reference depth per SSTable (default: 4)")
 	runCmd.Flags().Duration("pebble-value-separation-rewrite-age", time.Hour, "Minimum blob file age before rewrite (default: 1h)")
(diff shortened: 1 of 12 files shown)


gfyrag:
Closing this PR: the setting was already exposed end-to-end as spec.pebble.valueSeparation.garbageRatio. This PR only renamed the existing configuration and is not needed.

PR #1727, opened 2026-08-19
feat(EN-1779): opt-in string encoding for posting amounts

paul-nicolas:
[EN-1779](https://formance-team.atlassian.net/browse/EN-1779)

## Problem

`Posting.amount` is a 256-bit unsigned integer and the only money field on the v3 HTTP surface serialized as a **bare JSON number**. Volume fields (`input`, `output`, `balance`) are already quoted decimal strings.

`JSON.parse` decodes every JSON number into an IEEE-754 double, which represents integers exactly only up to 2^53 − 1. So `9007199254740993` silently becomes `9007199254740992` in any JavaScript client, with no error raised anywhere.

## Approach

An **opt-in** request header, `Formance-Bigint-As-String` (v2 parity). When a request carries a truthy value, every posting amount in the response is a quoted decimal string. Without it, **response bodies are byte-identical to before this change**.

Each marshaller in the posting chain moves its field list into `buildAux(amountsAsString bool) any`; `MarshalJSON` delegates `false`, and a value-receiver wrapper type delegates `true`. A parent types its child field `any` and stores the child's wrapper, so the mode rides the existing encoder recursion rather than a parallel type tree. Ten levels, from `Posting` up through `PreparedQueryCursor`.

Independently of the header, `Uint256.UnmarshalJSON` now accepts **both** a bare number and a quoted decimal, so a client can post back what it received. Non-breaking relaxation; hex stays rejected.

`Uint256.MarshalJSON` is deliberately untouched — `ledgerctl` and `misc/operator` consume it, so the CLI wire must not move. No `docs/ops/cli.md` change.

## Scope

Eight operations accept the header: create / revert / get / list transactions, get log, list ledger logs, bulk, execute prepared query.

## The two invariants, and how they are enforced

**1. The default wire does not move.** `TestDefaultWireGolden` pins the exact default bytes of every affected payload type across 19 fixtures, under both encoder configurations. The expectations were validated by running the same table against `release/v3.0` in a detached worktree — 19/19 pass there, which is what makes them a description of the pre-change wire rather than a transcript of current output. 14 mutants were built against it; all 14 fail the gate, including a field-order swap that a structural (`JSONEq`) comparison would have missed.

**2. The header changes the amount's format, never the response shape.** A nil child renders `null` and an empty collection renders `[]` identically in both modes, pinned per wrapper. Each of the eight routes pins both bodies byte-for-byte and asserts they differ at the posting amounts and nowhere else.

## A trap worth knowing about

This package has **two** production JSON writers over different sonic configurations: `writeOK`/`writeCreated`/`writeJSONResponse` use `ConfigStd` (HTML-escapes, appends a trailing newline), while `writeOKChecked` uses `ConfigDefault` (neither). The two log routes sit on opposite sides.

So **a route must never change which writer it uses** — a swap silently changes every response body on that route. This PR adds `writeCheckedBody`/`writeCheckedStatus` (buffered, `ConfigStd`) so the marshal-before-header guarantee is available *without* moving bytes, rather than reusing `writeOKChecked` and changing the encoder.

Nothing previously enforced this. `TestResponseWriters_TableIsComplete` now pins the route→writer mapping by AST-walking the package and comparing `file: writer` pairs. Its acceptance test: flipping `handlers_list_chapters.go` between the two encoders used to pass the entire suite, and now fails.

Test fixtures deliberately contain `<` and `&` in account addresses. Those characters are the only thing that makes an encoder swap visible in a body — without them, the assertions pass under the defective wiring too.

## Testing

- `TestDefaultWireGolden`: 19 fixtures, both encoders, exact byte comparison, validated against `release/v3.0`.
- Eight route tests, each pinning both bodies with an above-2^53 amount asserted on **raw bytes** (the repo's sonic wrapper has no `UseNumber`; decoding truncates).
- Wrapper contract: all ten wrappers asserted to implement `json.Marshaler` **as values** (which is what pins the value-receiver requirement) and to render `null` over a nil embedded pointer.
- `TestWriteCheckedStatus`: covers both checked writers' error paths, including the previously untested `writeOKChecked` path. A mutant hoisting `WriteHeader` above the marshal produced a `201 Created` carrying an error body and passed the whole suite before this test existed.
- Schemathesis: seeds an above-2^53 amount and sets the header deterministically on a subset of cases, with an `after_call` assertion that the requested mode actually took effect. `oneOf` conformance is request-blind, so it cannot catch "server ignored the header" on its own.
- e2e (`transactions_amount_wire_test.go`): exact-value round-trip on the detail and list routes in both modes, plus a non-truthy header and a `"  YES  "` case. Verified to fail under both an always-true and an always-false boundary helper.

Validation: `go build ./...`, `go vet`, `golangci-lint` with the justfile tag set (0 issues, no `--fix`), the `commonpb` and `adapter/http` suites, and the full 562-spec e2e business suite under `-race`.

## Notes for the reviewer

- **Based on `release/v3.0` at `97a488a1c`.** EN-1791 (#1703) is still open and overlaps these files; the plan is to rebase once it merges.
- **Commits are unsigned.** ssh-agent holds no identities in this environment and the key is passphrase-protected. `release/v3.0` lands work as GitHub squash-merges, which GitHub signs, so this does not affect the target branch.
- `"responseSignature":{}` appears in the pinned log bodies. That is **pre-existing** behaviour, not introduced here: `protoFieldJSON` guards `msg == nil` but receives a `proto.Message` interface holding a typed nil `*Signature`, so the guard misses and `protojson` renders `{}`. Identical on `release/v3.0`. Pinned deliberately with the reason inline — fixing it is a separate ticket.

## Follow-ups (not in this PR)

1. **Platform-UI SDK regeneration** for the `oneOf` amount type — agreed as a separate change.
2. `protoFieldJSON`'s typed-nil guard, per the `"responseSignature":{}` note above.
3. `_is_network_error()` in the schemathesis suite treats a `"network_other"` token as a network error, so a **client-side** failure can turn a run green. Pre-existing; it masked a real bug twice during this work.
4. Only 2 of the 8 header-declaring operations produce amount-bearing responses under fuzzing — fuzzed `transactionId` values never match a seeded id, so the detail route returned 404 on all 50 examples. Pre-existing vacuity; the fix is a `transactionId` coercion mirroring the existing `ledgerName` one, which changes what those routes fuzz.
5. `writeOK`/`writeCreated` could be reimplemented over the buffered writer, retiring the streaming error path package-wide. Measured byte-identical on every value tried. Real correctness upside, deliberately out of scope here.

[EN-1779]: https://formance-team.atlassian.net/browse/EN-1779?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/technical/architecture/subsystems/api/http-api.md
+++ b/docs/technical/architecture/subsystems/api/http-api.md
@@ -172,6 +172,86 @@ Clients should:
 - **Idempotency**: Ensure write operations are idempotent to safely retry after leader election
 - **Monitoring**: Track `503` responses to monitor cluster health and leader election frequency
 
+### Bigint-As-String Header
+
+`Posting.amount` is a 256-bit unsigned integer, and it is the only money field on the v3 HTTP surface serialized as a bare JSON number — volume fields (`input`, `output`, `balance`) are already quoted decimal strings. A JavaScript `JSON.parse` decodes every JSON number into a `Number`, an IEEE-754 double that represents integers exactly only up to 2^53 - 1: the amount `9007199254740993` (2^53 + 1) silently becomes `9007199254740992`, with no error raised anywhere.
+
+`Formance-Bigint-As-String` is the opt-in request header that removes that truncation. When a request carries it with a truthy value, every posting amount in the response body is a quoted decimal string, which `JSON.parse` returns verbatim for the client to hand to `BigInt`. Without the header, response bodies are byte-identical to what the API emitted before the header existed.
+
+#### Header Value
+
+The truthy values are `true`, `yes`, `y` and `1`, compared case-insensitively with surrounding whitespace ignored. Any other value — including an unrecognised one — is treated exactly as if the header were absent. **A malformed header is never an error**: it never turns a readable response into a `400`. The header name and this lenient parsing match Ledger v2, so a client can send the same header to both versions (`internal/adapter/http/string_amounts.go`).
+
+#### Operations That Accept It
+
+| Operation | Route |
+|---|---|
+| Create a transaction | `POST /v3/{ledgerName}/transactions` |
+| Revert a transaction | `POST /v3/{ledgerName}/transactions/{transactionId}/revert` |
+| Get a transaction | `GET /v3/{ledgerName}/transactions/{transactionId}` |
+| List transactions | `GET /v3/{ledgerName}/transactions` |
+| List ledger logs | `GET /v3/{ledgerName}/logs` |
+| Get a log | `GET /v3/_/logs/{sequence}` |
+| Bulk operations | `POST /v3/{ledgerName}/bulk` |
+| Execute a prepared query | `POST /v3/{ledgerName}/prepared-queries/{queryName}/execute` |
+
+These are exactly the responses that can carry a posting — directly, or nested inside a log payload, a bulk result or a prepared-query cursor page. On any other route the header is inert. `openapi.yml` expresses the same set as the reusable `BigintAsString` header parameter, referenced from those eight operations; both `PostingRequest.amount` and `PostingResponse.amount` are typed as a `oneOf` of an integer and a quoted decimal string (`pattern: '^[0-9]+$'`, deliberately not `format: int64` — the value can exceed 2^64).
+
+#### Response Format
+
+Default (header absent), `GET /v3/{ledgerName}/transactions/{transactionId}`:
+
+```json
+{
+  "data": {
+    "transaction": {
+      "postings": [
+        {"source": "world", "destination": "alice", "amount": 9007199254740993, "asset": "USD/2", "color": ""}
+      ],
+      "metadata": {},
+      "id": 42,
+      "reverted": false
+    },
+    "receipt": ""
+  }
+}
+```
+
+With `Formance-Bigint-As-String: true`, the same request returns the same document with the amount quoted:
+
+```json
+{
+  "data": {
+    "transaction": {
+      "postings": [
+        {"source": "world", "destination": "alice", "amount": "9007199254740993", "asset": "USD/2", "color": ""}
+      ],
+      "metadata": {},
+      "id": 42,
+      "reverted": false
+    },
+    "receipt": ""
+  }
+}
+```
+
+**Only the amount's format changes; the response shape never does.** Field names, field presence, ordering and nesting are identical in both modes at every level of the payload — a nil child still renders `null`, and an empty collection still renders `[]`. That is mechanically enforced, not merely intended: each of the eight routes pins both bodies byte-for-byte and asserts they differ at the posting amounts and nowhere else. The implementation mechanism (how the mode travels down the marshaller chain, and what a new amount-bearing level must do) is specified in `internal/proto/commonpb/string_amounts.go`.
+
+#### Request Side
+
+Independently of the header, a posting amount is accepted **both** as a bare JSON number and as a quoted decimal string on input. A client that read string amounts can therefore post them back unchanged, with no conversion step and without sending the header on the write. This is a relaxation only — numeric input behaves exactly as before — and it is not gated by anything. Hexadecimal (`0x...`) is rejected in both forms.
+
+#### Known Limitation: Integer Metadata
+
+The header covers posting amounts and nothing else. Metadata values of integer type are **bare JSON numbers in both modes**, and an unsigned metadata value can reach 2^64 - 1, far above `Number.MAX_SAFE_INTEGER`. A JavaScript client that opts in will still truncate those silently. This is by design — the header's scope is the amount field, not every integer on the wire — but it means the header must not be read as "no truncation anywhere". A client that has to read large integer metadata losslessly needs a big-integer-aware JSON parser, or must take that field from the raw response text.
+
+#### Best Practices
+
+- **Send the header on writes as well as reads**: the response of a create, revert or bulk request carries postings too.
+- **Parse with `BigInt`**: `BigInt(posting.amount)` accepts the quoted decimal directly. Never route it through `Number` first — the truncation happens in `JSON.parse`, before any check the client could make.
+- **Do not branch on the mode when writing**: input accepts both forms unconditionally, so a client can pick one form and keep it.
+- **Skip the header when the client does not decode into a float**: Go, Java, Python and other arbitrary-precision decoders read the default numeric wire losslessly, and that wire is unchanged.
+
 ## Main Endpoints
 
 ### Ledgers
--- a/docs/technical/architecture/subsystems/api/http-api.md
+++ b/docs/technical/architecture/subsystems/api/http-api.md
@@ -235,7 +235,7 @@ With `Formance-Bigint-As-String: true`, the same request returns the same docume
 }
 ```
 
-**Only the amount's format changes; the response shape never does.** Field names, field presence, ordering and nesting are identical in both modes at every level of the payload — a nil child still renders `null`, and an empty collection still renders `[]`. That is mechanically enforced, not merely intended: each of the eight routes pins both bodies byte-for-byte and asserts they differ at the posting amounts and nowhere else. The implementation mechanism (how the mode travels down the marshaller chain, and what a new amount-bearing level must do) is specified in `internal/proto/commonpb/string_amounts.go`.
+**Only the amount's format changes; the response shape never does.** Field names, field presence, ordering and nesting are identical in both modes at every level of the payload — a nil child renders identically in both modes, `null` where the field is always present and omitted where it is optional, and an empty collection still renders `[]`. That is mechanically enforced, not merely intended: each of the eight routes pins both bodies byte-for-byte and asserts they differ at the posting amounts and nowhere else. The implementation mechanism (how the mode travels down the marshaller chain, and what a new amount-bearing level must do) is specified in `internal/proto/commonpb/string_amounts.go`.
 
 #### Request Side
 
--- a/internal/adapter/http/handlers_bulk.go
+++ b/internal/adapter/http/handlers_bulk.go
@@ -102,7 +102,7 @@ func (s *Server) serveBulk(w http.ResponseWriter, r *http.Request) {
 	results := s.runBulk(r.Context(), ledgerName, elements, opts)
 
 	// Write response
-	writeBulkResponse(w, elements, results, opts.continueOnFailure)
+	writeBulkResponse(w, r, elements, results, opts.continueOnFailure)
 }
 
 // bulkOptions contains options for bulk processing.
@@ -231,10 +231,14 @@ func (s *Server) runBulkSequential(ctx context.Context, requests []*servicepb.Re
 //
 // Any 5xx/429 status from infra/retryable/rate-limit errors always surfaces,
 // with `Retry-After: 1` on 503 to match handleError.
-func writeBulkResponse(w http.ResponseWriter, elements []*servicepb.BulkElement, results []bulkResult, continueOnFailure bool) {
+func writeBulkResponse(w http.ResponseWriter, r *http.Request, elements []*servicepb.BulkElement, results []bulkResult, continueOnFailure bool) {
 	worstBusiness := 0 // highest 4xx from a per-element business failure
 	worstInfra := 0    // highest ≥429 from a retryable/infra/rate-limit error
 	apiResults := make([]bulkAPIResult, len(results))
+
+	// Resolve the EN-1779 amount wire once for the whole batch rather than per
+	// element: the header applies to the response, not to an entry.
+	amountsAsString := wantsBigintAsString(r)
 
 	for i, result := range results {
 		responseType := servicepb.GetLedgerActionType(elements[i].Action)
@@ -284,7 +288,20 @@ func writeBulkResponse(w http.ResponseWriter, elements []*servicepb.BulkElement,
 			if payload := log.GetData(); payload != nil {
 				switch p := payload.GetPayload().(type) {
 				case *commonpb.LedgerLogPayload_CreatedTransaction:
-					data = p.CreatedTransaction.GetTransaction()
+					// EN-1779: the amount wire is applied here, in the only arm
+					// that carries an amount, and never at the write site — the
+					// OrderSkipped arm below must stay byte-identical in both
+					// modes. The value is a bare *commonpb.Transaction, so the
+					// wrapper is StringAmountTransaction; a nil inner pointer
+					// renders `"data":null` in both modes (a typed nil behind an
+					// `any` is a non-nil interface, so `omitempty` does not fire,
+					// and the wrapper's nil guard emits the same null).
+					transaction := p.CreatedTransaction.GetTransaction()
+
+					data = transaction
+					if amountsAsString {
+						data = commonpb.StringAmountTransaction{Transaction: transaction}
+					}
 				case *commonpb.LedgerLogPayload_OrderSkipped:
 					data = OrderSkippedResponse{
 						Skipped: true,
@@ -315,8 +332,13 @@ func writeBulkResponse(w http.ResponseWriter, elements []*servicepb.BulkElement,
 		statusCode = worstBusiness
 	}
 
+	// writeCheckedBody, not writeJSONResponse: bulk builds its own top-level
+	// envelope, so it must not be wrapped in BaseResponse. It is the same
+	// ConfigStd encoder writeJSONResponse used — no response byte moves — and a
+	// marshal failure on a per-element transaction now becomes a clean 500
+	// instead of an error object appended to an already-committed status.
 	response := bulkResponse{Data: apiResults}
-	writeJSONResponse(w, statusCode, response)
+	writeCheckedBody(w, r, statusCode, response)
 }
 
 // perElementStatus returns the HTTP status a given per-element error would
--- a/docs/technical/contributing/api-comparison.md
+++ b/docs/technical/contributing/api-comparison.md
@@ -415,6 +415,17 @@ date coercion (EN-1544), AND-combination, and the removed `prefix`/`reference`
 aliases (EN-1540) — is documented once in
 [query-filter.md](../architecture/subsystems/read-path/query-filter.md).
 
+### Amount output formats (dual-format contract, EN-1779)
+
+`Posting.amount` is emitted as a bare JSON number by default and as a quoted
+decimal string when the request carries the opt-in `Formance-Bigint-As-String`
+header, so JavaScript clients do not truncate above 2^53; on input both forms
+are accepted unconditionally. The full contract — truthy header values and
+lenient parsing, the eight operations that accept it, the unchanged default
+wire, the response-shape guarantee and the integer-metadata limitation — is
+documented once in
+[http-api.md](../architecture/subsystems/api/http-api.md#bigint-as-string-header).
+
 ### 10. Prepared Queries and User-Configurable Indexes
 
 Prepared queries are reusable, named filter queries stored per-ledger. They can be executed in two modes: `LIST` (returns matching entity IDs with cursor pagination) and `AGGREGATE_VOLUMES` (returns aggregated volumes per asset for matched accounts).
--- a/internal/adapter/http/encoder_contract_test.go
+++ b/internal/adapter/http/encoder_contract_test.go
@@ -7,6 +7,7 @@ import (
 	"go/token"
 	"os"
 	"reflect"
+	"sort"
 	"strings"
 	"testing"
 
@@ -67,11 +68,20 @@ func TestProtojsonRoutes_PayloadHasNoCustomMarshalJSON(t *testing.T) {
 	}
 }
 
-// TestSonicRoutes_PayloadHasCustomMarshalJSON is the converse: the three routes
-// fixed by EN-1622 rely on their type's marshaller being the contract. If a
-// marshaller is ever deleted, sonic falls back to the protoc-gen `json:` tags,
-// which are snake_case — a silent wire regression. The handler tests assert the
-// resulting shape; this asserts the mechanism they depend on.
+// TestSonicRoutes_PayloadHasCustomMarshalJSON is the converse: every payload
+// served through one of the sonic writers relies on its type's marshaller being
+// the contract. If a marshaller is ever deleted, sonic falls back to the
+// protoc-gen `json:` tags, which are snake_case — a silent wire regression with
+// no compile error. The handler tests assert the resulting shape; this asserts
+// the mechanism they depend on.
+//
+// Transaction, Chapter and Log are the three EN-1622 routes. The other three
+// arrived with EN-1779's opt-in string amounts, which routed them through the
+// checked writers as well: PreparedQueryCursor on the prepared-query execute
+// route (an exposure that predates EN-1779 — the cursor already carried its own
+// marshaller), CreatedTransaction on create-transaction and RevertedTransaction
+// on revert-transaction. Their opt-in wrappers delegate to the same buildAux
+// the marshaller uses, so deleting a marshaller breaks BOTH modes at once.
 func TestSonicRoutes_PayloadHasCustomMarshalJSON(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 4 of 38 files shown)


paul-nicolas (commented):
## Multi-model review (Claude + Codex) — PR #1727

**DECISION: APPROVE** · Head reviewed `e6a04c5faf1dc52c6b3825a23d85eb1b30052a0f`

Blocking findings: **0**. Posted inline: **1** (P2, non-blocking). Withheld as P3: **2**. Rejected in verification: **1**.

### Verification outcome

| Finding | Verdict |
|---|---|
| Stale `Retry-After` on the 500 from the new checked writer (`handlers_bulk.go`) | **valid** — reproduced: `status=500 retry-after="1"` |
| Default path loses streaming on two read routes (P3) | **valid** — documented trade-off, withheld from inline |
| `"+42"` accepted although the schema declares `^[0-9]+$` (P3) | **valid** — reproduced, withheld from inline |
| `Uint256.UnmarshalJSON` rejects escape-encoded decimal strings (`"\u00342"`), Codex, P2 blocking | **rejected** |

Rejection reason: the claim is factually true (reproduced: `strconv.ParseUint: parsing "\\u00342": invalid syntax`), but `release/v3.0` rejected *every* quoted form — `SetFromDecimal(string(data))` on the raw token — so this PR strictly **widens** accepted input and does not make the limitation worse. No standard serializer escapes ASCII digits (`json.Marshal("42")` → `"42"` in Go, and likewise for `JSON.stringify` / `json.dumps` / Jackson), and the code documents the narrowness deliberately. Both exclusions in `docs/technical/contributing/ai-review.md` apply.

### What was verified rather than assumed

The two invariants the PR claims are load-bearing, not decorative. Four mutants were built against the suite in a throwaway worktree; **all four fail**:

| Mutant | Result |
|---|---|
| Flip `handleGetTransaction` from `writeCheckedStatus` to `writeOKChecked` | `TestResponseWriters_TableIsComplete` FAILS |
| `wantsBigintAsString` hardwired `false` (header ignored) | all 8 route tests + `TestWantsBigintAsString` FAIL |
| `wantsBigintAsString` hardwired `true` | all 8 route tests + `TestWantsBigintAsString` FAIL |
| Drop the nil guard in `StringAmountTransaction.MarshalJSON` | `TestStringAmountWrappers_NilInnerPointerIsNull`, `TestTransaction_NilElementStaysNullInBothModes`, `TestHandleBulk_StringAmountWire` FAIL |

Also confirmed independently:

- **Type-graph coverage is complete.** Both `LedgerLogPayload` posting arms are rebuilt (`CreatedTransaction` *and* `RevertedTransaction`), `LogPayload_Apply` is the only `LogPayload` arm reaching a posting, and `PreparedQueryCursor` wraps both `transactionData` and `logData`. `AggregatedVolume.Input/Output` are the only other `*Uint256` fields on the surface and `toAggregatedVolumeJSON` already renders them as strings — so the docs' premise that `Posting.amount` is the only bare-number money field holds.
- **No ninth amount-bearing route.** Audit entries, chapters and `analyzeTransactions` carry no posting amount (`NormalizedPosting` has no `amount` field).
- **Byte-parity of the writer swaps.** Only four routes actually change writer for a request that would have been served before: `handleGetTransaction` and `handleListLedgerLogs` (`writeOK` → `writeCheckedStatus`), and `writeBulkResponse` and `handleExecutePreparedQuery` (`writeJSONResponse` → `writeCheckedBody`). Create-transaction and revert-transaction are **not** swaps — they keep `writeCreated` on the default branch and take `writeCheckedStatus` only on the opt-in branch, so nothing moves for a default client there. In all cases `writeCheckedBody` marshals through the same `json.MarshalWrite` (ConfigStd) as `writeJSONResponse`/`writeOK`, and neither `writeOK` nor `writeCreated` sets any extra header, so no byte and no header is lost. Confirmed by running the pre-PR and post-PR writer on the same payload: identical status, `Content-Type`, `Content-Length` (absent on both — both stay chunked) and body; the control pair `writeOKChecked` vs `writeCheckedStatus` **does** differ, so the `<`/`&` in the fixture account names are load-bearing.
- **`[]` vs `null` parity.** `drainCursor` returns `[]T{}`, never nil, matching `wrapAll`'s always-non-nil result.
- **openapi.yml declares the header on exactly the 8 operations the code honors** (verified by parsing the file, not by reading the diff).
- **Golden tests are real.** `TestDefaultWireGolden` compares raw bytes under both encoder configurations against hardcoded literals — not values generated by the code under test — and the route tests assert on raw body bytes without decoding, so the above-2^53 amount cannot be silently truncated by the assertion itself.

`go test ./internal/proto/commonpb/... ./internal/adapter/http/...` passes; PR CI is green including `Tests-E2E` and `Tests-Schemathesis`.

### P3 (not posted inline)

- `internal/adapter/http/handlers_get_transaction.go:81` and `internal/adapter/http/handlers_list_ledger_logs.go:121` — both routes move from streaming `writeOK` to buffered `writeCheckedStatus` on the **default** path too, so every request (including the majority that never send the header) now materialises the whole body in a `bytes.Buffer`. For the logs list that is a full drained page. Response bytes are unchanged; this is the stated cost of routing a `MarshalJSON` failure to a clean 500, and it is documented in the code comments. Filed for visibility only.
- `internal/proto/commonpb/uint256.go:115` — the hand-rolled quote strip passes the inner text straight to `SetFromDecimal`, which accepts a leading `+`: `"+42"` parses as `42`, while the `oneOf` string branch this PR added to `openapi.yml` declares `pattern: '^[0-9]+$'`. Harmless in value terms and normal server leniency, but it is a small divergence between the schema and the implementation the same PR published. (`"007"` → `7` is inside the pattern and fine; `"-1"`, `"0x2a"`, `"1e3"` and `" 42 "` are all correctly rejected.)

### One question for a human, not a finding

The "matches Ledger v2" parity claim is asserted in three authoritative places — `internal/adapter/http/string_amounts.go:10-11`, the `BigintAsString` parameter description in `openapi.yml`, and `docs/technical/architecture/subsystems/api/http-api.md` — and it cannot be substantiated from this repository or the module cache (no v2 source is vendored; `grep -rin 'bigint.as.string'` over `~/go/pkg/mod/github.com/formancehq/` returns nothing). If v2's truthy set is not exactly `{true, yes, y, 1}` case-insensitively, the documentation is materially false for the cross-version client it explicitly invites. Worth one confirmation against the v2 source before merge. Not filed as a finding because it cannot be shown to be wrong from here.

### Conflicts between reviewers, and how they were resolved

- **Codex vs Claude** on the `Uint256.UnmarshalJSON` escaped-digit finding: resolved against Codex, rejected on the two contract exclusions above (pre-existing, not made worse; no plausible client path).
- **Severity of the `Retry-After` finding**: the area reviewer rated it P3 residual; I post it as **P2 non-blocking**, because the stale hint has a behavioural consequence for a client (a retry of already-committed bulk work), which puts it in the contract's "bounded reliability gap ... failure in a narrower path" band rather than "no correctness impact". Non-blocking either way, so the merge decision is unaffected.
- One correction from the area reviewer was adopted into the writer-parity item above: create/revert-transaction are not writer swaps on the default path.

### Note on method

Codex reviewed the diff externally and returned one finding (rejected above). Of three delegated Claude area-reviewers, one returned a full report; the other two produced nothing, so the commonpb-chain and test-surface areas were reviewed directly in the main session instead, with the mutation testing above standing in for their verification pass. Coverage of both areas is reflected in the findings above.

```
DECISION: APPROVE
Head reviewed: e6a04c5faf1dc52c6b3825a23d85eb1b30052a0f
Blocking findings: 0
Non-blocking findings: 3 (1 x P2 posted inline, 2 x P3 listed above) + 1 question
Validation considered: go test ./internal/proto/commonpb/... ./internal/adapter/http/... (PASS); 4 mutants, all caught; PR CI green incl. Tests-E2E and Tests-Schemathesis
Residual risk: LOW
```

internal/adapter/http/handlers_bulk.go
```diff
@@ -315,8 +332,13 @@ func writeBulkResponse(w http.ResponseWriter, elements []*servicepb.BulkElement,
...
 	}
 
+	// writeCheckedBody, not writeJSONResponse: bulk builds its own top-level
+	// envelope, so it must not be wrapped in BaseResponse. It is the same
+	// ConfigStd encoder writeJSONResponse used — no response byte moves — and a
+	// marshal failure on a per-element transaction now becomes a clean 500
+	// instead of an error object appended to an already-committed status.
 	response := bulkResponse{Data: apiResults}
-	writeJSONResponse(w, statusCode, response)
+	writeCheckedBody(w, r, statusCode, response)
```
paul-nicolas, line 341: [P2][non-blocking] Clear `Retry-After` before `handleError` answers, or set it after the status is final

Re: the `w.Header().Set("Retry-After", "1")` at line 329, in the `worstInfra == 503` branch just above.

**What happens.** That branch writes the retry hint into the header map *before* the body is marshalled. `writeCheckedBody` then marshals first and, on failure, routes to `handleError` (`response.go:112-115`), which answers **500 INTERNAL_ERROR**. `handleError` only ever *sets* `Retry-After`, never deletes it (`error_handler.go:57,70,106`), so the hint set for the 503 survives onto the 500.

Reproduced against this HEAD:

```
status=500 retry-after="1" body={"errorCode":"INTERNAL_ERROR","errorMessage":"internal server error (correlation ID: ...)"}
```

**Why it is new here.** Before this PR the route called `writeJSONResponse`, which had already committed the 503 status by the time the marshal ran, so `handleError` was unreachable from this path and the header always matched its status. Moving to the checked writer is what makes the mismatch possible. Reachability rests on the same premise this change is justified by — the comment two lines above states that a per-element transaction's marshal "can genuinely fail" — so the failure path cannot be dismissed as unreachable without also undermining the reason for the writer change.

**Impact.** A client honouring `Retry-After` retries a request that failed for a non-retryable reason, and for bulk the operations have already been applied by the time `writeBulkResponse` runs, so a retry without an idempotency key risks duplicating committed work. Bounded: it needs a marshal failure and the 503 branch together. Net behaviour is still better than before the PR, which appended a second JSON object to an already-committed 503 and produced an unparseable body — hence non-blocking.

**Resolution.** Either `w.Header().Del("Retry-After")` on the `handleError` path in `writeCheckedBody`, or set the hint inside the writer once the status is final, so the header can never outlive the status it was chosen for.

docs/technical/architecture/subsystems/api/http-api.md
```diff
@@ -172,6 +172,86 @@ Clients should:
...
+      "metadata": {},
+      "id": 42,
+      "reverted": false
+    },
+    "receipt": ""
+  }
+}
+```
+
+**Only the amount's format changes; the response shape never does.** Field names, field presence, ordering and nesting are identical in both modes at every level of the payload — a nil child still renders `null`, and an empty collection still renders `[]`. That is mechanically enforced, not merely intended: each of the eight routes pins both bodies byte-for-byte and asserts they differ at the posting amounts and nowhere else. The implementation mechanism (how the mode travels down the marshaller chain, and what a new amount-bearing level must do) is specified in `internal/proto/commonpb/string_amounts.go`.
```
paul-nicolas: [P2][non-blocking] The "a nil child still renders `null`" illustration is contradicted by the PR's own goldens

**What the sentence says.** "Field names, field presence, ordering and nesting are identical in both modes at every level of the payload — a nil child still renders `null`, and an empty collection still renders `[]`."

**What the code does.** For a child carrying `omitempty` the field is **omitted, not null**. The PR's own fixtures pin exactly that:

- `internal/proto/commonpb/string_amounts_test.go:1294-1298` — `"created transaction without a transaction"` → `{"chapterId":3}`, with the fixture's own comment saying "transaction is an `omitempty` pointer retyped to `any`: absent, not null".
- the sibling golden for a reverted transaction with no revert transaction → `{"revertedTransactionId":3}`.
- `TestLog_NilPayloadStaysOmittedInBothModes` asserts `require.NotContains(t, string(def), "payload")` and the same for the opt-in body.

`null` is correct only for the three fields that carry no `omitempty`: `getTransactionData.transaction`, `LedgerLog.data` and `bulkAPIResult.data`.

**Why it matters.** This is the authoritative HTTP-API document, and the sentence is the one a client author reads to decide how to type their model. Someone reading it will code for `"transaction": null` on a created-transaction response with no transaction and hit an absent key instead. The repo's review contract lists "misleading authoritative docs" as a P2 example.

**Scope.** The parity claim the sentence is illustrating is correct and is mechanically enforced — only the illustration is wrong, which is why this is non-blocking.

**Suggested wording.** "a nil child renders identically in both modes — `null` where the field is always present, omitted where it is optional — and an empty collection still renders `[]`".

paul-nicolas (commented):
## Follow-up to [review 4993058503](https://github.com/formancehq/ledger/pull/1727#pullrequestreview-4993058503)

A late area-reviewer report landed after the first review was posted. It re-ran the test surface independently and produced six findings, two of which I reproduced and are not in the first review. **Decision is unchanged: APPROVE, 0 blocking.** One P2 non-blocking is posted inline; the rest are listed below.

### Independent confirmation worth recording

The reviewer did something the first review did not: it extracted `TestDefaultWireGolden`'s constants, fixtures and body verbatim into a standalone test in a throwaway worktree at the merge-base `f9603e3b6` and ran it there — **19/19 pass on the pre-change code, under both encoders**. That converts "the default wire does not move" from the PR's claim into a verified fact, which is the single most load-bearing assertion in the change. It also ran two mutations the first review did not: a new route with no table row (caught), and an unregistered 11th wrapper with a pointer receiver and no nil guard (caught).

Combined with the four mutants in the first review, the score is 6 mutations attempted, 5 caught, 1 escape — reported below.

### P3 (not posted inline)

- **`internal/adapter/http/encoder_contract_test.go:334` — one reproduced escape in `TestResponseWriters_TableIsComplete`.** I confirmed it against this HEAD: declaring a new writer inside `response.go` (`func writeAccepted(w http.ResponseWriter, data any) { writeJSONResponse(w, http.StatusAccepted, BaseResponse[any]{Data: data}) }`) and adding a route file whose only writer call is `writeAccepted(w, "hello")` makes the test **pass** — `go test -run TestResponseWriters_TableIsComplete` returns `ok`. The new route's file contributes no name in `responseWriterNames`, and `response.go` already carries a `writeJSONResponse` row, so nothing is missing or extra. Every other mutation is caught (new route calling an existing writer, new route calling a new writer declared in the route's own file, a route changing writer). The failure mode is precisely what the table exists to prevent: a future route silently landing on the wrong sonic configuration, changing HTML escaping and the trailing newline for every client of that route with no test failing. Disclosed in the test's own comment, so this is a hardening request, not a misrepresentation. Fix: derive `responseWriterNames` from the functions declared in `response.go` by AST rather than hardcoding it.
- **`internal/adapter/http/string_amounts.go:19` ↔ `openapi.yml:2936` — the one hand-written list in this change with no completeness test.** The eight `wantsBigintAsString` call sites and the eight `$ref: '#/components/parameters/BigintAsString'` entries are two independently maintained lists of the same set. Both are identical today (verified in the first review by parsing the file). Nothing fails if they diverge — and the PR adds completeness tests for its three *other* hand-written lists (`protojsonRoutes`, `responseWriterCallSites`, `stringAmountWrappers`), while `TestResponseWriters_TableIsComplete` already AST-walks this exact package. If they drift, `http-api.md`'s "These are exactly the responses that can carry a posting" becomes silently false in either direction.
- **`tests/schemathesis/test_api.py` — the header rotation has no non-vacuity assertion.** `_assert_amount_wire_format` derives its expectation from the header it reads back off the case, so if `_operation_declares_bigint_header` ever returned `False` for every operation, no header would be sent, `_requested_bigint_as_string` would always be `False`, and the assertion would degenerate to "amounts are bare numbers" — which the server does by default — printing ALL CHECKS PASSED while the opt-in wire is never fuzzed. The current pin is fine (confirmed: with `schemathesis==3.39.16`, enumerating `operation.headers` reports the header on exactly the 8 operations), but `requirements.txt` explicitly invites deliberate bumps. One line asserting `_bigint_header_step` is non-empty, or counting opted-in amount-bearing responses and requiring > 0, closes it.
- **`openapi.yml:3988` — the `PostingResponse` `oneOf` also weakens response-schema validation.** Beyond the SDK-regeneration consequence already tracked as follow-up 1 in the PR body, the response schema now *permits* a string amount even for a request that never sent the header — so `response_schema_conformance` no longer catches a server that starts quoting unconditionally. That specific regression is instead carried by the request-aware `_assert_amount_wire_format`, which is the right place for it; worth knowing that the two guards moved rather than added.

### On the PR's own admission that only 2 of 8 operations produce amount-bearing responses under fuzzing

Assessed and **should not block**, agreeing with the area reviewer. It describes the reach of a fuzzer, not a weakened assertion. The fuzz check is non-vacuous where it applies — the `run.sh` seed plus the alternating `BIGINT_HEADER_ROTATION` exercise both wire formats on the amount-bearing GETs at CI's `MAX_EXAMPLES=10`, `parse_float=Decimal` keeps the value exact, and the check fails both a server that ignores the header and one that quotes unconditionally. The other six operations are covered by byte-pinned unit tests that were mutation-verified to fail in *both* directions, which is stronger evidence than a fuzz sample. The proportionate response is the one-line non-vacuity assertion above, not a blocker.

### Coverage gaps assessed and deliberately not filed

Mixed bulk success+failure under the header (the ERROR arm provably sets no `Data`); bulk `REVERT_TRANSACTION` elements (no `RevertedTransaction` arm in `writeBulkResponse`'s switch, unchanged from base — so "bulk honours the header" means "for create-transaction elements"); error bodies with the header set (no amount, `handleError` untouched); the `OrderSkipped` variant in `TestLog_NonPostingPayloadIsByteIdentical` (falls through the same `default:` delegation the two covered variants prove); and a `revertedTransaction` default-wire golden (the arm is covered in both modes by `TestLog_StringAmountReachesPostingThroughRevertedTransaction`). None has a plausible failure path.

```
DECISION: APPROVE (unchanged)
Head reviewed: e6a04c5faf1dc52c6b3825a23d85eb1b30052a0f
Blocking findings: 0
New non-blocking findings: 5 (1 x P2 inline, 4 x P3 above)
Validation added since the first review: TestDefaultWireGolden re-run on merge-base f9603e3b6 (19/19 pass); 2 further mutations caught, 1 escape reproduced; schemathesis 3.39.16 $ref-resolution probe; go vet -tags e2e ./tests/e2e/business/ clean
Residual risk: LOW
```

internal/adapter/http/handlers_bulk.go
paul-nicolas: Fixed in 550d95187 — valid, and reproduced independently against `e6a04c5fa` before touching anything: `status=500 retry-after="1"`.

Took the first of your two options, placed one level up. Both buffered writers now route their marshal failure through a shared `failCheckedWrite` (`response.go:70-87`) that drops `Retry-After` before delegating to `handleError`. Applied to `writeOKChecked` as well as `writeCheckedBody`: the two have the identical shape — marshal, then hand a *different* status to `handleError` — so the former is safe today only because no caller happens to set the header yet. Only `Retry-After` is cleared; the header map also carries middleware-set entries (CORS, correlation) that must survive the status change.

Tests, both directions:

- The shared error-path subtest in `TestWriteCheckedStatus` now pre-sets the hint and asserts it is gone, so it runs for all five writer rows.
- `TestCheckedWriters_RetryableFailureKeepsRetryAfter` is the positive control: a payload failing with an error wrapping `commonpb.ErrNoLeader` must still answer 503 **with** `Retry-After: 1`. Without it, a `Del` moved after the delegation would pass the negative test while silently suppressing the hint `handleError` legitimately owes.

Both mutations verified to fail rather than assumed: removing the `Del` breaks the five negative rows; moving it after `handleError` breaks the three control rows.

One deviation worth stating, since it bounds the coverage. I intended a handler-level test driving `writeBulkResponse` through the 503 rollup with a failing per-element payload, and it is not constructible. No real bulk payload type has a reachable marshal failure — I checked the obvious candidate directly, and sonic replaces invalid UTF-8 with `U+FFFD` rather than erroring (`{"metadata":{"k":"��"}}`, nil error, under both encoder configurations) — and a synthetic failing marshaler cannot be injected because `writeBulkResponse` builds `bulkAPIResult.Data` itself from the log payload. So the pin is at the writer level, exercising `writeCheckedBody`, the exact writer bulk uses, with the hint pre-set. Worth noting this also narrows your reachability argument: the failure needs a payload type that can actually fail, which on the bulk arm today is none — the fix stands on the shape of the path rather than on a live trigger.

docs/technical/architecture/subsystems/api/http-api.md
paul-nicolas: Fixed in 325936238 — valid. Adopted your wording, so `http-api.md:238` now reads "a nil child renders identically in both modes, `null` where the field is always present and omitted where it is optional, and an empty collection still renders `[]`".

Confirmed against the fixtures you cite and against the mechanism: `string_amounts.go:84-88` states outright that the `childValue` guard exists so a nil child on an `omitempty` field stays omitted rather than emitting null.

One correction to your list of always-present fields, which does not change the verdict or the fix. You name three; there are two. `bulkAPIResult.data` carries `omitempty` (`handlers_bulk.go:400`, `json:"data,omitempty"`), so that one is omitted, not null. The `null` cases are `getTransactionData.transaction` (`handlers_get_transaction.go:24`) and `LedgerLog`\s `data` (`commonpb/log.go:125`). The new wording enumerates no fields, so nothing downstream of this depends on it — flagging it so the record is right.

Agreed on the severity read: the parity claim being illustrated is true and mechanically enforced, only the illustration was wrong.

gfyrag:
I think the product behavior here is the right one: keep the current numeric representation as the default, and only return string amounts when the client explicitly requests it through `Formance-Bigint-As-String`. Making strings the new global default would disrupt more consumers than necessary.

However, I would avoid merging the current wrapper-based architecture as-is if we have no urgency. The implementation has to propagate `amountsAsString` through most of the response graph and introduces a parallel set of transaction/log/cursor types for what is ultimately a leaf-level rendering option on `Posting.amount`.

Go 1.27 looks like a much better fit for this. It is expected this month and RC3 is already available. With `encoding/json/v2`, the HTTP writer can install a request-scoped, type-specific marshaler using `jsonv2.WithMarshalers`:

- no header: use the canonical numeric `Posting` representation;
- header enabled: override only the marshaling of `*commonpb.Posting` so that `amount` is quoted;
- parent transactions, logs, cursors, slices, etc. stay unchanged.

I would override `Posting`, not `Uint256`, and would not use `StringifyNumbers`: both alternatives would affect numbers outside this specific contract.

There is one integration caveat: our current custom `MarshalJSON` methods recurse through Sonic, so a json/v2 option installed at the top-level writer would not automatically reach nested postings. The clean version probably belongs with EN-1792 / `protoc-gen-ledger-json`: generate streaming codecs using `jsontext.Encoder`, and have parent codecs encode children through `jsonv2.MarshalEncode` so request-scoped marshalers propagate through the graph.

My suggestion would therefore be:

1. keep the opt-in header and the tolerant `Uint256.UnmarshalJSON`;
2. keep the >2^53 tests and the byte-for-byte default-wire goldens from this PR;
3. wait for the stable Go 1.27 release;
4. reimplement the response-side option centrally in the HTTP writer, on top of the generated codec, without the boolean propagation and wrapper hierarchy.

We can already validate the design against Go 1.27 RC3, then adopt it only once the stable toolchain is available.

References: [Go 1.27 release notes](https://go.dev/doc/go1.27), [current Go downloads (RC3)](https://go.dev/dl/), [encoding/json/v2](https://pkg.go.dev/encoding/json/v2).

paul-nicolas:
Closing per @gfyrag's review.

The product decision is confirmed: the numeric wire stays the default, and string amounts are returned only when the client asks through `Formance-Bigint-As-String`. What is rejected is the wrapper-based implementation in this PR — propagating `amountsAsString` through ten levels of the response graph and maintaining a parallel transaction/log/cursor type tree for what is a leaf-level rendering option on `Posting.amount`.

The agreed direction is to reimplement the response side centrally in the HTTP writer once Go 1.27 is stable, installing a request-scoped `jsonv2.WithMarshalers` override on `*commonpb.Posting` only (not on `Uint256`, and not `StringifyNumbers`). That requires the generated streaming codecs from EN-1792 / `protoc-gen-ledger-json` so a top-level marshaler actually reaches nested postings instead of being swallowed by the current custom `MarshalJSON` chain.

Kept for the reimplementation: the opt-in header contract, the tolerant `Uint256.UnmarshalJSON`, the above-2^53 tests and the byte-for-byte default-wire goldens. This branch stays available as the reference for those.

EN-1779 moves to Blocked, waiting on the stable Go 1.27 toolchain.

PR #1730, opened 2026-08-19
docs(openapi): describe the schema fields in the spec

ariel-formance:
## Why

The Ledger OpenAPI spec declares its schemas without saying what any of the fields mean, so the generated SDKs ship bare field names and the API reference renders row after row of `none` in the description column.

## What

- `openapi/v1.yaml`, `openapi/v2.yaml` — `description` added to the account, transaction, posting, volume, balance, log, query-params and exporter schemas, plus the cursor wrappers.
- `openapi.yaml` — re-merged from v1 + v2 + overlay, as `just openapi` does.
- `docs/api/README.md` — regenerated by widdershins. This is where the change is most visible: the description column stops saying `none`.

No structural changes: no fields added or removed, no types or required-ness touched, so generated clients keep the same shape and only pick up doc comments.

## Draft: one step outstanding

`just pre-commit` also runs `generate-client`, which regenerates `pkg/client` through Speakeasy — and spec descriptions do reach `pkg/client/docs/`. I could not run that locally (the CLI needs interactive browser auth), so **the Dirty check will fail on the missing `pkg/client` regen** until someone with Speakeasy access runs:

```
cd pkg/client && speakeasy run --skip-versioning
```

Everything else in `just pre-commit` was reproduced locally, and I verified the regenerated `docs/api/README.md` matches byte-for-byte what the committed one regenerates to on an unmodified tree — so there is no toolchain-version drift hiding in the diff.

Part of a sweep doing the same across the module specs.


ariel-formance:
Added 469752bbb while sweeping a defect NumaryBot caught on the sibling flows PR ([flows#202](https://github.com/formancehq/flows/pull/202)).

These specs are `openapi: 3.0.3`, where siblings of a `$ref` are ignored — so a `description` next to a bare `$ref` never reaches generated docs or SDK comments. flows had 19 such cases, all mine. I scanned every spec in the sweep; ledger has **2**, both in `V2LogDataRevertedTransaction`:

```yaml
revertedTransaction:
  description: The original transaction that was reverted
  allOf:
    - $ref: "#/components/schemas/V2LogTransaction"
```

**These two predate this PR** — they are on `main` already. I fixed them anyway since they are the same defect this PR exists to fix, in a file it already edits. Shapes are unchanged: a single-member `allOf` validates identically to the bare `$ref`.

Clean on the other seven specs in the sweep (auth, gateway, wallets, webhooks, reconciliation, payments, and the archived search patch) — zero introduced.

---

Two things for whoever picks this up:

1. **Still draft, still needs the Speakeasy regen.** `just pre-commit` runs `generate-client`, and descriptions do reach `pkg/client/docs/`, so the Dirty check will fail until someone with Speakeasy access runs `cd pkg/client && speakeasy run --skip-versioning`. I can't complete the CLI's browser auth.

2. **Unrelated latent bug, not touched here.** `LedgerInfo.storage.migrations[].sequence` is declared `type: string` with `minimum: 0` and `example: 11`. kin-openapi rejects it (`invalid example: value must be a string`) — on `main` too, identically, so this PR neither causes nor worsens it. `just pre-commit` doesn't run a spec validator here, which is why it has gone unnoticed. Looks like it should be `type: integer`. Happy to fix in a separate PR.

flemzord (approved):

ariel-formance:
Thanks @shipfox-ai — the `V2Metadata` finding was correct and is fixed in #1841.

Verified before acting: `pit` appears on exactly nine operations in the v2 spec, all under `accounts`, `transactions`, `logs` and `aggregate/balances`. None of `GET /v2`, `GET /v2/{ledger}` or `PUT /v2/{ledger}/metadata` declares it, so the inherited bi-temporal sentence was indeed over-claiming for `V2CreateLedgerRequest`, `V2Ledger` and `V2UpdateLedgerMetadataRequest`.

Took the suggested fix: the shared schema description is generic again, and the bi-temporal sentence sits on `V2Transaction.metadata` via the `allOf` wrapper. `V2Account.metadata` already declared it inline and correctly scoped.

Separately — your review arrives as a top-level PR comment rather than inline threads, so our `address-pr-comments` tooling queried `reviewThreads`, got zero, and reported this PR clean. That is our bug, not yours; fixed in formancehq/internal-skills#17.

PR #1732, opened 2026-08-19
fix: require local validation before AI approval

gfyrag:
## Summary

- require local validation after an agent approval and before `READY_FOR_HUMAN_REVIEW`
- select deterministic local gates from the immutable base-to-worktree diff (unit, E2E, scenarios, Schemathesis, operator)
- reject approvals when validation fails or mutates the reviewed workspace
- use the base-pinned validator in guarded publish mode
- keep GitHub Actions entirely outside the readiness decision

## Validation

- `AI_REVIEW_BASE_SHA=$(git rev-parse origin/release/v3.0) bash scripts/agent-check-pr`
- `go test ./scripts/reviewloop ./scripts/aiprloop ./scripts/agentcheckpr -count=1`
- `nix develop --command bash -c "just pre-commit"`
- `GOROOT= go build ./...`
- GitNexus `detect-changes`: low risk, no affected execution flow

PR #1734, opened 2026-08-20
chore: add trusted PR validation tools

gfyrag:
## Summary

- add the dormant `agent-check-pr` gate selector and base-pinned `agent-just` wrapper
- pin model-checker and Schemathesis harnesses while keeping reviewed source inputs explicit
- keep `review-loop` and `ai-pr-loop` unchanged so this bootstrap does not authorize itself
- prepare the trusted base required to activate local approval validation in #1732

## Validation

- `go test ./scripts/agentcheckpr ./scripts/agentjust -count=10`
- shell syntax checks for all changed scripts
- `AI_REVIEW_BASE_SHA=77d1473a... bash scripts/agent-check-pr --list`
- `bash scripts/agent-check`
- GitNexus impact/detect-changes: LOW, no affected execution flow

## Follow-up

After this PR is merged, #1732 will be rebased onto it and will contain only the policy/launcher activation. This guarantees that guarded publish resolves `agent-check-pr` from an already trusted base commit.

### DIFF
--- a/scripts/agent-check-pr
+++ b/scripts/agent-check-pr
@@ -100,7 +100,7 @@ classify_path() {
 
 while IFS= read -r -d '' path; do
     classify_path "$path"
-done < <(git diff --name-only --diff-filter=ACDMRTUXB -z "$base_sha" --)
+done < <(git diff --no-renames --name-only --diff-filter=ACDMRTUXB -z "$base_sha" --)
 while IFS= read -r -d '' path; do
     classify_path "$path"
 done < <(git ls-files --others --exclude-standard -z)
--- a/scripts/agent-check-pr
+++ b/scripts/agent-check-pr
@@ -1,0 +1,158 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+usage() {
+    cat <<'EOF'
+Usage: AI_REVIEW_BASE_SHA=<commit> bash scripts/agent-check-pr [--list]
+
+Runs the local validation gates required by the changes since the immutable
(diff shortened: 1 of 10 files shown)


scripts/agent-check-pr
```diff
@@ -0,0 +1,158 @@
...
+        misc/proto/cluster.proto|\
+        misc/proto/cluster_bootstrap.proto|\
+        misc/proto/proposal.proto|\
+        misc/proto/raft_cmd.proto|\
+        misc/proto/raft_transport.proto|\
+        misc/proto/restore.proto|\
+        misc/proto/snapshot.proto|\
+        tests/antithesis/*|\
+        tests/oracle/*)
+            needs_model_cluster=true
```
NumaryBot, line 96: 🟠 **[major] Classify both sides of renamed files**

When a production file is renamed outside a gated path—for example, `internal/domain/foo.go` to `docs/foo.md`—`git diff --name-only` reports only the destination, so this selector skips the E2E gate despite removing production code. Disable rename detection or consume both paths from `--name-status`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1734#issuecomment-5352861034

paul-nicolas (dismissed):

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1734#issuecomment-5352861034

paul-nicolas (approved):

PR #1736, opened 2026-08-20
fix(indexbuilder): rebuild the index config after a snapshot install

Azorlogh:
## What changed

A leader `InstallSnapshot` now also reaches the indexbuilder, which rebuilds its
per-ledger index config at the next point in its loop where it owns that config.

## Why

Catching a follower up on a leader snapshot reopens its Pebble DB with the
leader's checkpoint (`processReady` → `applier.SyncSnapshot` → `RestoreCheckpoint`).
Anything derived from the old store has to be rebuilt across that swap, and the
only callback wired for it refreshed the peer-address cache.

The builder reads the registry into `indexConfig` in `bootInit` and never again.
On a restart the two are milliseconds apart — the node boots, then Raft
discovers how far behind it is — so the builder ends up describing the pre-swap
registry: it holds indexes the restored registry dropped, leaves their
read-store rows unpurged, reports no index to drop when the declaration is
removed, and misses indexes created while the node was behind, which then never
build there. Measured on a 3-node cluster: the builder loaded its config 19ms
before the install, and answered 7 removals across 3 indexes from the stale copy.

## Risk

**LOW** — one flag and a rebuild at a point that already owns the config; no
change to what is indexed.

## Validation

- [x] Targeted: `TestIndexConfigFollowsTheStoreAcrossASnapshotInstall`, red-checked
      (fails with the production symptom when the rebuild is disabled)
- [x] `internal/application/indexbuilder`, `internal/bootstrap`

The applier's post-install callback now notifies the builder as well as
membership. No DI cycle: the builder's constructor takes only store, read store,
attributes, logger, meter and config.

## Review focus

That the rebuild runs on the builder goroutine. `OnSnapshotInstalled` is called
from the applier and only raises a flag, because `indexConfig` is not guarded by
a lock.

### DIFF
--- a/internal/application/indexbuilder/builder.go
+++ b/internal/application/indexbuilder/builder.go
@@ -46,6 +46,10 @@ type Builder struct {
 	metricsRegistration     metric.Registration // tailworker gauge triplet
 	logsIndexedRegistration metric.Registration // builder-specific logs_indexed_total gauge
 
+	// Raised when a leader snapshot install has swapped the store this cache
+	// was built from, and lowered by the loop once it has rebuilt.
+	configReloadPending atomic.Bool
+
 	// Per-ledger index configuration cache.
 	indexConfig map[string]*ledgerIndexConfig
 
@@ -539,6 +543,13 @@ func (b *Builder) loop(ctx context.Context) {
 			return
 		case <-b.notifications.LogCommitted.C():
 		case <-ticker.C:
+		}
+
+		// Ahead of touching the store: a snapshot install may have replaced it
+		// with the leader's checkpoint, leaving the config describing a store
+		// that is gone.
+		if err := b.reloadConfigIfRequested(ctx); err != nil {
+			b.logger.Errorf("%v", err)
 		}
 
 		// Fast path: skip Pebble iterator + batch commit when the FSM
--- a/internal/application/indexbuilder/builder.go
+++ b/internal/application/indexbuilder/builder.go
@@ -505,6 +505,22 @@ func (b *Builder) loop(ctx context.Context) {
 		default:
 		}
 
+		// Catch-up can run for hours, so a snapshot install landing mid-way
+		// cannot wait for the main loop: the config would describe the
+		// pre-swap store for the rest of it. The stripped view is rebuilt
+		// from the reloaded config, since restoreIndexes closes over the old.
+		if b.configReloadPending.Load() {
+			restoreIndexes()
+
+			if err := b.reloadConfigIfRequested(ctx); err != nil {
+				b.logger.Errorf("%v", err)
+
+				break
+			}
+
+			restoreIndexes = b.stripBuildingIndexes()
+		}
+
 		before := cursor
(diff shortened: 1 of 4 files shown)


internal/application/indexbuilder/builder.go
```diff
@@ -541,6 +545,13 @@ func (b *Builder) loop(ctx context.Context) {
 		case <-ticker.C:
 		}
 
+		// Ahead of touching the store: a snapshot install may have replaced it
```
NumaryBot, line 564: 🔴 **[blocker] Reload during the initial catch-up loop**

If the snapshot is installed during initial catch-up, this reload is not reached until catch-up finishes, which the surrounding code notes may take hours. Meanwhile `processLogs` and `processBackgroundTasks` continue using the pre-snapshot configuration, so dropped indexes can still receive rows and new indexes can be skipped—the inconsistency this patch intends to prevent ([AGENTS.md:33](AGENTS.md#L33)). Check the pending flag at each catch-up iteration before processing more logs.

internal/application/indexbuilder/builder.go
```diff
@@ -541,6 +545,13 @@ func (b *Builder) loop(ctx context.Context) {
...
+		// with the leader's checkpoint, leaving the config describing a store
+		// that is gone.
+		if err := b.reloadConfigIfRequested(ctx); err != nil {
```
NumaryBot, line 567: 🟠 **[major] Stop processing after a failed configuration reload**

When a transient Pebble or read-store error makes the reload fail, `initIndexConfig` has already cleared and potentially partially repopulated the configuration and task slices. This branch logs the error but then processes logs and background work with that incomplete state before retrying on the next iteration, potentially skipping required index writes; the iteration should not proceed after a failed reload.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1736#issuecomment-5353733276

Azorlogh:
Both addressed in cdfca4aae.

- Reload during initial catch-up: the catch-up loop now reloads before each iteration, restoring the stripped view and re-stripping from the reloaded config (`restoreIndexes` closes over the old one).
- Failed reload: the tick is skipped instead of proceeding, since `initIndexConfig` clears the config before repopulating it. The request stays raised, so the next tick retries.

https://github.com/formancehq/ledger/commit/cdfca4aae

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1736#issuecomment-5353733276

Azorlogh:
Closing: the premise was wrong. The builder's index config is log-synchronous state — every create/drop is itself a ledger-log entry, so folding reaches each one in order and converges the config naturally after a snapshot install (drops purge by range delete when reached, creates schedule backfills that cover the skipped range). Reloading the config to head-state mid-fold desynchronizes it from the cursor and makes entries fold against a config from their future. The assertion that motivated this compared apply-time facts against config-now, which fires on healthy catch-up.

PR #1737, opened 2026-08-20
fix(wal): recreate a missing snapshot directory, stop when the WAL directory is gone

Azorlogh:
## What changed

`Snapshotter.Save` recreates its directory when only that directory is gone, and
refuses when its parent went with it. Creating a directory now fsyncs the parent
of every level it makes, in `NewSnapshotter` and `wal.New` too.

## Why

`Save` creates its temp file directly, so a snapshot directory that went away
after `NewSnapshotter` made it fails the write. That failure surfaces as
`applying snapshot to storage: saving snapshot file: creating temp snap file:
… no such file or directory`, which fails `processReadies`; `Node.Run` picks it
up from `tasks.err()` and the bootstrap goroutine panics on it
(`internal/bootstrap/module.go:1155`). A follower catching up on a leader
snapshot dies instead, and dies again on every retry. Seen killing a node
repeatedly in a cluster e2e that drives snapshot installs.

Losing only `WalDir/snap` is recoverable, so that case is recreated — loudly: an
error log plus an `assert.Unreachable`, because what removed it is **not
established**. Losing `WalDir` itself is not recoverable: it holds the etcd WAL,
`WAL_CREATION_COMPLETED` and `INSTANCE_ID`, so etcd is fsyncing unlinked inodes
and the terms, votes and entries this node acknowledges as persisted are already
gone. The next restart finds no marker, rebuilds an empty WAL and rejoins as a
new member — and the existing guard at `wal_default.go:344` cannot catch it,
since it only trips while `*.wal` files are still there. `Save` reports that case as
`ErrWALDirectoryMissing` rather than recreate the tree, and background
maintenance — the most frequent `Save` on a healthy node, which used to log and
drop the error — now exits its loop on it, which stops the node. What this bounds
is how long the loss goes undetected: one snapshot attempt instead of forever.

Recreation is also made durable per level: `MkdirAll` plus one fsync of the
immediate parent leaves the outer levels' entries unsynced, which is the same
failure one level up.

## Risk

**LOW** — the crash-safe write (temp file → fsync → rename → dir fsync) is
untouched. `mkdirAllSynced` replaces the two `MkdirAll` calls in the constructors,
on paths that are normally already present, where it is a stat and no more. It is
stricter than `MkdirAll` in one direction only: a path occupied by a file or a
dangling symlink is now an error instead of a `Snapshotter` over a non-directory.

## Validation

- [x] `internal/storage/wal`, `internal/infra/node`, plus `./internal/...`
- [x] Targeted: recreation of a missing snapshot directory, refusal when the
      parent is missing, construction over a non-directory, every failure branch
      reachable from `Save`, and the maintenance loop exiting on the refusal —
      all red-checked against the guard removed

## Documentation

- `docs/technical/architecture/subsystems/consensus/raft-consensus.md` — the two
  outcomes under Snapshot Creation
- `docs/ops/deployment.md` — a Troubleshooting entry for the node stopping
- `wal.WAL` — the three methods that can return `ErrWALDirectoryMissing`

`Snapshotter.Save` gained a documented sentinel (`wal.ErrWALDirectoryMissing`)
and `doMaintenance` returns an error so the maintenance task can exit on it.

Two conditions that used to be silent are now declared: a vanished snapshot
directory is an `assert.Unreachable` that recovers, and a vanished WAL directory
is an `assert.Unreachable` that stops the node. Antithesis runs will report
either rather than pass silently. That is the intent — the removal is
unattributed and worth chasing.

## Review focus

Whether stopping the node on a missing WAL directory is the right call, given
that the node is unrecoverable at that point either way.


internal/storage/wal/snapshotter.go
```diff
@@ -47,6 +47,14 @@ func (s *Snapshotter) Save(snap *raftpb.Snapshot) error {
 	path := filepath.Join(s.dir, name)
 	tmpPath := path + ".tmp"
 
+	// The directory is created in NewSnapshotter, so reaching here without it
+	// means it went away underneath a running node. Recreate it: the snapshot
+	// this call is persisting is how a follower catches up, and failing the
+	// save propagates as a task-pool panic that takes the node down.
+	if err := os.MkdirAll(s.dir, 0755); err != nil {
```
NumaryBot: 🔴 **[blocker] Make the recreated directory durable before saving the WAL record**

When `s.dir` is recreated, syncing only that directory after the rename does not make its new entry in the parent directory durable. A crash after `Save` returns and the WAL snapshot record is persisted can therefore leave the WAL referencing a snapshot directory/file lost during recovery, violating the ordering relied on by `CreateSnapshot`. Fsync the newly created directory's parent before allowing `Save` to report success.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1737#issuecomment-5353733153

Azorlogh:
Addressed in 4c4b32c1b: the parent directory is fsynced after the recreate, so the new entry is durable before the WAL snapshot record can depend on it.

https://github.com/formancehq/ledger/commit/4c4b32c1b

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1737#issuecomment-5353733153

internal/storage/wal/snapshotter.go
```diff
@@ -86,6 +92,85 @@ func (s *Snapshotter) Save(snap *raftpb.Snapshot) error {
...
+	details := map[string]any{
+		"dir":          s.dir,
+		"parentExists": parentErr == nil,
+	}
+
+	assert.Unreachable("snapshot directory disappeared underneath a running node", details)
+
+	s.logger.WithFields(details).Errorf("Snapshot directory is missing, recreating it")
+
+	if err := mkdirSynced(s.dir); err != nil {
```
NumaryBot, line 143: 🔴 **[blocker] Stop when the WAL parent directory is missing**

When deletion includes `filepath.Dir(s.dir)`, this recreates the WAL data directory and lets the node continue while its open WAL files and persisted identity remain unlinked; subsequent acknowledgements can therefore be lost on restart, which then sees no marker and initializes a new identity. Only recreate the snapshot subdirectory when its parent still exists; otherwise fail loudly, as required by [AGENTS.md:13-15](/data/worktrees/formancehq__ledger-pr-1737/AGENTS.md#L13-L15).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1737#issuecomment-5353733153

gfyrag (commented):
Independent review at head 4d67e467fd9b8d7892ebd11191663fe34f49e0da. One high-impact blocking finding remains.

[P1][blocking] Fail closed when the WAL root has disappeared
Location: internal/storage/wal/snapshotter.go (Snapshotter.ensureDir)
Evidence: the code detects that the WAL parent is missing but only calls assert.Unreachable, which is a production no-op, then recreates the hierarchy. The new test explicitly removes the entire WAL root and expects Save to succeed. That root also contains the WAL/HardState, creation marker, instance identity, and cluster-joined state.
Impact: the node can acknowledge work after its durable consensus/identity paths are unlinked, then restart with lost work or a fresh identity; this is the documented double-vote/acknowledged-write-loss class.
Resolution: recreate only a missing snap child when the WAL root is demonstrably intact. Return a terminal error when the parent or required root evidence is missing, and change the ancestor-removal test to require failure without hierarchy recreation.

Targeted WAL/node tests passed. Functional CI is green; NumaryBot and patch coverage are failing.

internal/storage/wal/snapshotter.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/4d67e467fd9b8d7892ebd11191663fe34f49e0da — creating a directory now fsyncs the parent of every level it makes, so the recreated entry is durable before `Save` returns.

internal/storage/wal/snapshotter.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/cf2b63e707a061a93736588a895b08026da6b546 — `Save` returns `ErrWALDirectoryMissing` when the parent is gone instead of recreating it, and background maintenance exits its loop on that error rather than logging and ticking again.

Azorlogh:
@gfyrag P1 resolved at cf2b63e70:

- `Save` returns a terminal `ErrWALDirectoryMissing` when the parent is gone — nothing is recreated.
- The recreate covers the snap child only (`mkdirSynced`, single level), with the parent checked immediately above, so it cannot walk up.
- The ancestor-removal test now requires failure and asserts no hierarchy recreation (`require.NoDirExists(walDir)`).
- Beyond the review: background maintenance logged and dropped that error while ticking again — the most frequent `Save` on a healthy node — so it now exits the maintenance loop, which stops the node. Every other `CreateSnapshot`/`UpdateSnapshotConfState`/`ApplySnapshot` caller already propagated.

Patch coverage is green; both NumaryBot threads are resolved.

gfyrag (changes requested):
Independent review at exact head `cf2b63e707a061a93736588a895b08026da6b546`.

[P1][blocking] Preserve the terminal WAL-missing classification across the parent-check/create race
Location: `internal/storage/wal/snapshotter.go:124-144,153-163`; `internal/storage/wal/wal_default.go:768-772`; `internal/infra/node/node.go:758-768`
Evidence: the WAL parent can disappear after `os.Stat(parent)` succeeds but before `os.Mkdir(s.dir)`. `mkdirSynced` then returns a generic wrapped ENOENT rather than `ErrWALDirectoryMissing`, so `doMaintenance` logs and absorbs it. `CreateSnapshot` has already advanced its in-memory snapshot before `Snapshotter.Save` fails, allowing later maintenance ticks to take the no-new-index fast path.
Impact: the node can continue acknowledging entries after its durable consensus directory has disappeared.
Resolution: classify the create-time parent disappearance as `ErrWALDirectoryMissing`, avoid publishing/retain a retryable in-memory snapshot on non-terminal save failure, and add a deterministic race regression.

Targeted WAL/node unit tests pass, confirming that this interleaving is currently untested.

gfyrag:
Review checkpoint for tomorrow: please rebase this PR onto the current `release/v3.0` (`d095a8b085553f6045573d0317003b605d420193`). The existing P1 blocker is still valid at `cf2b63e`: if the WAL parent disappears between the parent `Stat` and `mkdirSynced`, the error is no longer classified as `ErrWALDirectoryMissing`; meanwhile `CreateSnapshot` has already published the new in-memory snapshot, so later maintenance can skip retrying. Please preserve the terminal classification across that race, avoid publishing or retain a retryable snapshot on save failure, and add a deterministic regression.

NumaryBot (commented):

A WAL-directory deletion racing the preliminary directory check still bypasses the terminal sentinel, so the node can continue after consensus state is lost.

### Findings outside the diff

🔴 **[blocker] Classify deletion after the directory check as terminal** — `internal/storage/wal/snapshotter.go:62`

If the WAL directory is removed after `ensureDir()` succeeds but before `os.Create`, the resulting `ENOENT` is returned as a generic snapshot-save error rather than `ErrWALDirectoryMissing`. Background maintenance therefore absorbs it and the node continues acknowledging unrecoverable writes until another snapshot attempt, contrary to the required immediate stop behavior documented in [raft-consensus.md:364-376](/data/worktrees/formancehq__ledger-pr-1737/docs/technical/architecture/subsystems/consensus/raft-consensus.md#L364-L376). Classify missing-parent errors from the actual create/write/rename path, not only from `mkdirSynced`.




Summary: https://github.com/formancehq/ledger/pull/1737#issuecomment-5353733153

gfyrag:
There is still a correctness gap in the recovery classification at HEAD `5ce7b9f4`: the code verifies pathname existence, not the identity of the WAL root that the running etcd WAL opened.

If the WAL directory is removed/renamed and another directory is recreated at the same pathname (with a `snap/` child), `ensureDir` succeeds and `Save` writes into the replacement while etcd can still be fsyncing descriptors in the old unlinked tree. The node then keeps acknowledging state that will not exist after restart. A replacement by a regular file similarly yields a generic `ENOTDIR`, which maintenance logs and retries instead of treating as terminal. This is distinct from the already-filed absent-parent race.

Please make the runtime decision capability/identity-based: pin the opened WAL root identity and reject a path that no longer resolves to it, or use a root-scoped directory capability and validate at the actual create/rename/fsync boundary. The simpler policy is to make any runtime loss of `snap/` fail-stop and only recreate it during startup after the WAL root and its markers have been validated. I would also add deterministic barriers for (a) replacement after the preliminary check, (b) replacement by a directory and by a file, and (c) crash/reopen. The startup side should prove that a fresh WAL cannot be bootstrapped over a non-empty existing FSM store after marker loss.

PR #1748, opened 2026-08-21
chore: gate AI PR loop on legitimacy triage

gfyrag:
## Summary

Integrate the advisory PR legitimacy triage directly before the expensive AI review/fix loop.

- run legitimacy triage before Codex review / Claude fix
- load the triage adapter and policy from the exact verified PR base commit, never from the proposed head
- bind triage to the same immutable base/head SHAs already resolved by `ai-pr-loop`
- continue only on `KEEP`
- map `QUESTION` to `HUMAN_DECISION_REQUIRED`
- map advisory `REJECT` to `LEGITIMACY_REJECTED`
- never comment, close, merge, or otherwise mutate PR metadata
- fail closed if GitHub base/head moves between launcher verification and triage
- add launcher regression coverage proving QUESTION/REJECT stop before technical review

## Why

The review loop is now capable of spending substantial agent and human attention on technically reviewable changes. The standalone legitimacy triage exists to answer an earlier question: whether a PR has enough documented product/operational need and proportionality to deserve that attention at all.

This integration makes that gate the default entry point while preserving human authority over ambiguous or rejected proposals.

## Trust boundary

A PR must not provide the policy that legitimizes itself. `ai-pr-loop` therefore creates a base-pinned trusted tool worktree for every invocation, not only `--push` mode, and executes `scripts/ai-pr-triage` from that base.

The triage adapter receives `AI_PR_TRIAGE_EXPECT_BASE_SHA` and `AI_PR_TRIAGE_EXPECT_HEAD_SHA`; after resolving live GitHub metadata it refuses any mismatch. This prevents triaging one state and reviewing another if the PR moves concurrently.

## Outcomes

```text
KEEP      -> normal review/fix loop
QUESTION  -> HUMAN_DECISION_REQUIRED, no review/fix
REJECT    -> LEGITIMACY_REJECTED, no review/fix
```

REJECT remains advisory; no GitHub action is taken.

## Review focus

Please focus on:

- whether the base-pinned triage policy can ever be replaced by PR-controlled content
- races between initial metadata/fetch and the triage's metadata resolution
- whether QUESTION/REJECT can reach any review/fix/write path
- worktree lifecycle now that trusted tools are created for non-push runs too
- exit-code semantics for QUESTION vs REJECT
- whether keeping REJECT advisory/non-destructive is preserved
- test fixture realism and any regressions to existing push/no-push behavior

## Follow-up

Product-to-technical traceability remains a separate follow-up. This PR only makes the existing legitimacy decision an automatic pre-review gate.

### DIFF
--- a/docs/technical/contributing/ai-pr-triage-loop.md
+++ b/docs/technical/contributing/ai-pr-triage-loop.md
@@ -19,3 +19,5 @@ The triage result is persisted in the launcher run directory and its path is pri
 The PR head may not choose or modify the triage policy that authorizes its own review. `ai-pr-loop` therefore uses a base-pinned trusted triage adapter for every invocation, not only guarded publish mode.
 
 `ai-pr-triage` accepts optional launcher-provided expected base/head SHAs. It still resolves current GitHub metadata itself, but refuses to run if those SHAs no longer match. This turns a concurrent PR update into a safe restart instead of triaging one state and reviewing another.
+
+The base-pinned triage adapter may predate that binding, so `ai-pr-loop` does not rely on the tool enforcing it: before accepting any decision, the launcher requires the result's `base_sha` and `head` to equal the SHAs it fetched and verified. A result describing another PR state is a target mismatch and stops the run.
--- a/docs/technical/contributing/ai-pr-triage-loop.md
+++ b/docs/technical/contributing/ai-pr-triage-loop.md
@@ -1,0 +1,21 @@
+# PR legitimacy gate in `ai-pr-loop`
+
+`ai-pr-loop` runs the advisory legitimacy triage before the expensive review/fix loop.
+
+The triage policy and adapter are always loaded from the exact verified PR base commit, never from the proposed head. The launcher binds triage to the same immutable base/head SHAs that will be reviewed.
+
+## Outcomes
+
+- `KEEP`: continue into the normal Codex review / Claude fix / validation loop.
+- `QUESTION`: stop before review with `AI_PR_LOOP_RESULT: HUMAN_DECISION_REQUIRED`.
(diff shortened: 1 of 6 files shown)


gfyrag:
### Non-blocking follow-up: Long-term legitimacy/CI architecture

The current integration of legitimacy triage inside `ai-pr-loop` is a sensible bootstrap, but the long-term boundary should be:

- CI is the authoritative producer of the legitimacy verdict as soon as a PR is created or changes.
- `ai-pr-loop` consumes and verifies that verdict before starting expensive technical review; it should not run a second independent triage.
- Missing, stale, mismatched, or untrusted results must fail closed.

The CI result should be bound to the exact inputs used for the decision, including at least `pr_number`, `base_sha`, `merge_base_sha`, `head_sha`, a digest of the PR title/body, and the policy/runner version. Binding PR metadata matters because the legitimacy evidence can change without a new commit.

The existing trust boundary must be preserved in CI: policy and runner come from the trusted target revision, while the PR head and metadata remain untrusted evidence. The job must never execute PR-supplied scripts with Codex credentials. Structured output, read-only isolation, and target verification remain mandatory.

Suggested flow:

1. PR event triggers the standalone legitimacy CI job.
2. CI publishes a structured, attributable result for the exact input fingerprint.
3. `KEEP` allows `ai-pr-loop` to proceed.
4. `QUESTION` requires human/author context and a rerun.
5. `REJECT` stops automation but remains advisory; it does not close the PR automatically.
6. Any human override is explicit, attributable, and bound to the same fingerprint.

This is not a requested scope expansion for the current PR. The remaining product decision is whether the future CI check stays advisory or becomes a required merge protection.

paul-nicolas (approved):

PR #1750, opened 2026-08-21
feat(EN-1827): replicated cluster policy (PR1)

Azorlogh:
## What

PR 1 of the [EN-1827](https://formance-team.atlassian.net/browse/EN-1827) plan: a **Raft-replicated `ClusterPolicy`** (`revision`, `idempotency_ttl_micros`, `query_checkpoint_limit`) as the deterministic home for configuration that must apply identically on every node, rather than being read from node-local startup config in the FSM apply path.

This PR establishes the policy plumbing only. Later PRs consume it:
- **PR 2** — enforce the query-checkpoint limit inside the FSM (after the idempotency gate).
- **PR 3** — per-outcome idempotency `expiresAt`.
- **PR 4** — control-plane / operator rollout.

## How

- **Audited `SetClusterPolicy` system-scoped order** (sibling of `SetSigningConfig`), with deterministic revision validation in `processSetClusterPolicy`:

  | Committed revision vs applied | Outcome |
  |---|---|
  | Higher | apply + emit `SetClusterPolicyLog` |
  | Equal, identical payload | idempotent no-op (no log) |
  | Equal, different payload | rejected `CLUSTER_POLICY_REVISION_CONFLICT` |
  | Lower | rejected `STALE_CLUSTER_POLICY` |

  The decision reads only committed state (`FSMState.ClusterPolicy`) + the committed order, so every replica resolves a given entry identically and a stale proposal can't overwrite a newer policy.
- **Persistence** — `ZoneGlobal`/`SubGlobClusterPolicy`, carried in `FSMState`, recovered on boot, replayed in audit rebuild. Snapshots and cold backups include it automatically (ZoneGlobal is inside the Pebble checkpoint).
- **Checker** — `clusterPolicyVerifier` re-derives the policy from chain-bound `SetClusterPolicy` orders (live + cold-storage-archived chapters, max-revision fold) and reports `CLUSTER_POLICY_VERIFICATION_INCOMPLETE` when the archive can't be read. Governance-truth projection under invariant #8, modeled on `signingVerifier`.
- **Reconciler + readiness** — `proposeClusterPolicyIfNeeded` runs on `LeaderReadyEvent` and proposes the desired policy via `Admit` (system actor) when its revision exceeds the applied one. Write readiness (`grpc_health_updater`) reports not-serving until a policy has been committed at least once; the `SetClusterPolicy` proposal is exempt so it can't deadlock on its own gate.
- **Config** — `--cluster-policy-revision` (default 1) and `--query-checkpoint-limit` (default 10) feed the desired policy; both validated `> 0`. Node-local for now; PR 4 sources them from the operator resource.
- **Docs** — `deterministic-fsm.md` §3.5 (config boundary + rollout), `audit-vs-technical-state.md` (verified governance projection).

## Decisions worth a look

1. **Audited `SystemScopedOrder`, not a `TechnicalUpdate`.** The ticket requires checker coverage, and the checker folds expectations from the audit chain — which `TechnicalUpdate`s bypass (that's why `ClusterConfig` is a tracked checker gap). So the policy must be an audited order.
2. **Same-revision/different-payload → deterministic rejection, not a hard apply-abort.** Aborting apply on a committed entry would wedge every replica permanently on a control-plane mistake; a deterministic typed rejection (`KindInternal`, failure-audited) is loud and safe.
3. **PR 1 step 8 (remove node-local idempotency TTL) deferred to PR 3.** The persisted-config TTL feeds the checker's `resolveIdempotencyTTLMicros` fallback (CLI/restore paths); PR 3 replaces the whole TTL-window verification model with per-outcome `expiresAt`. Removing the field here would break the checker or need throwaway re-sourcing.

## Testing

- Revision logic (all four outcomes + structural validation), checker verifier (consistent / injected / missing / mismatch / max-revision / incomplete), recovery roundtrip, readiness gate, config validation, plus every exhaustiveness/forcing test.
- `go build ./...` + `go vet` + targeted package tests all green. (The full `agent-check` isn't runnable in this env: golangci-lint OOMs and `just pre-commit` deletes tracked `.pb.go`.)

v3 is unreleased: no migration/compat shims, no reserved proto fields.

[EN-1827]: https://formance-team.atlassian.net/browse/EN-1827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/bootstrap/module.go
```diff
@@ -1111,6 +1119,7 @@ func Module() fx.Option {
 						go handleLeadershipChangeEvent(e, eventsManager, mirrorManager, logger)
 					case node.LeaderReadyEvent:
```
NumaryBot, line 1125: 🔴 **[blocker] Close the leader-ready gate before calling Admit**

On every leadership acquisition, `LeaderReadyEvent` is emitted synchronously before `node.go` closes the `leaderReady` channel. This handler calls `proposeClusterPolicyIfNeeded`, which synchronously calls `Admission.Admit`; `Admit` waits for that same channel, so the handler never returns and the channel is never closed. The initial policy therefore cannot be committed and readiness remains blocked. Dispatch this proposal asynchronously or close the gate before emitting the event.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1750#issuecomment-5369426653

internal/bootstrap/module.go
Azorlogh: Fixed in d4b4c80a1 — the reconciler is now dispatched off the observer thread (with a bounded context), so the `LeaderReadyEvent` handler returns and `node.go` closes the leader-ready channel before `Admit`'s `WaitLeaderReady` runs. Confirmed against the `Tests-Scenarios` timeout, whose goroutine was stuck in `WaitLeaderReady` right after the policy-propose log.

https://github.com/formancehq/ledger/commit/d4b4c80a12c36d1c418248f3dacb48abd175f260

internal/bootstrap/module.go
```diff
@@ -1709,6 +1722,59 @@ func proposeClusterConfigIfNeeded(n *node.Node, builder *plan.Builder, store *da
...
+	defer cancel()
+
+	if _, err := admission.Admit(
+		internalauth.WithSystemActor(ctx, commands.ComponentClusterPolicy),
+		servicepb.UnsignedApplyRequest("", &servicepb.Request{
+			Type: &servicepb.Request_SetClusterPolicy{
+				SetClusterPolicy: &servicepb.SetClusterPolicyRequest{Policy: desired},
+			},
+		}),
+	); err != nil {
```
NumaryBot, line 1779: 🔴 **[blocker] Retry a failed initial policy proposal**

If this one-shot `Admit` fails transiently (for example, proposal timeout, temporary write gate, or leadership churn), no retry occurs until another `LeaderReadyEvent`. On a fresh cluster the policy remains absent and gRPC readiness stays `NOT_SERVING` indefinitely despite the node recovering. Add periodic/event-driven reconciliation as required by the rollout behavior in [deterministic-fsm.md:143](docs/technical/architecture/subsystems/fsm/deterministic-fsm.md#L143).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1750#issuecomment-5369426653

internal/bootstrap/module.go
Azorlogh: Fixed in 5459fbc2d — replaced the one-shot proposal with a leader-gated `ClusterPolicyReconciler` that reconciles on a 5s ticker, so a transient failure self-heals on the next tick instead of waiting for another leadership event. Reconciliation is idempotent (once the applied revision reaches the desired one, every tick is a no-op), and running on its own worker also removes the observer-thread coupling.

https://github.com/formancehq/ledger/commit/5459fbc2dff2b0e3dd33c157e019d050fd839b93

cmd/server/server.go
```diff
@@ -224,6 +224,10 @@ func NewRunCommandWithBindings(bindings network.Bindings) *cobra.Command {
 	runCmd.Flags().Duration("idempotency-ttl", 24*time.Hour, "Idempotency key time-to-live (0 = never expire)")
 	runCmd.Flags().Duration("idempotency-eviction-interval", 60*time.Second, "How often the leader proposes idempotency eviction")
 
+	// Cluster policy (replicated via Raft; see EN-1827)
+	runCmd.Flags().Uint64("cluster-policy-revision", 1, "Desired revision of the replicated cluster policy (must be greater than zero)")
```
NumaryBot, line 228: 🟠 **[major] Document the new server flags**

These operator-facing flags are absent from `docs/ops/cli.md`, leaving the policy rollout controls undiscoverable in the canonical CLI reference. Add both flags and their defaults as required by [AGENTS.md:79](AGENTS.md#L79).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1750#issuecomment-5369426653

cmd/server/server.go
Azorlogh: Documented in a978d5e58 — added a "Server Cluster Policy Flags" section to docs/ops/cli.md covering `--cluster-policy-revision` and `--query-checkpoint-limit` (defaults, the >0 constraint, and the revision/readiness rollout behavior).

https://github.com/formancehq/ledger/commit/a978d5e58

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1750#issuecomment-5369426653

gfyrag (commented):
Independent review at head 1b3f6293202f2bb385a5bb472919b650edeba940. One blocking finding remains.

[P2][blocking] Prove ClusterPolicy reconstruction through the real incremental-restore lifecycle
Location: internal/infra/backup/rebuild_test.go (TestRebuildDelta_ReplaysClusterPolicy)
Evidence: the regression manually inserts a SetClusterPolicy log and invokes RebuildDelta directly. It does not exercise checkpoint seeding, non-empty export creation, ApplyExportsAndRebuild, logical source/destination comparison, CheckStore, or a consuming post-restore decision. The authoritative incremental-restore contract explicitly says an isolated RebuildDelta test is insufficient.
Impact: export, orchestration, checker, or checkpoint integration regressions can leave a restored node on stale governance policy while tests remain green.
Resolution: add a checkpoint + post-checkpoint policy revision + real restore lifecycle test, compare restored policy, require clean CheckStore, and exercise readiness/reconciliation.

Targeted tests, diff check, and current CI passed.

internal/bootstrap/module.go
```diff
@@ -1710,6 +1725,64 @@ func proposeClusterConfigIfNeeded(n *node.Node, builder *plan.Builder, store *da
...
+		if applied != nil && !applied.EqualVT(desired) {
+			logger.Errorf("cluster policy revision %d is already applied with a different payload; bump --cluster-policy-revision to change it", appliedRev)
+		}
+
+		return
+	}
+
+	logger.Infof("Proposing cluster policy revision %d", desired.GetRevision())
+
+	if _, err := admission.Admit(
```
NumaryBot, line 1772: 🔴 **[blocker] Allow the reconciler through signature enforcement**

When `RequireSignatures` is enabled, this unsigned request is always rejected by `authorizeUnsignedBatch`, which does not exempt system actors. Consequently, any later policy revision is retried forever and never applied, contrary to the documented reconciler rollout behavior in [deterministic-fsm.md:143](docs/technical/architecture/subsystems/fsm/deterministic-fsm.md#L143).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1750#issuecomment-5369426653

internal/bootstrap/module.go
Azorlogh: Confirmed. Fixed generically in #1774 (off release/v3.0): system-actor proposals are exempt from the unsigned-batch signature check, so the reconciler — and the chapter/checkpoint workers — can propose on a `RequireSignatures` cluster. #1750 picks it up on rebase once #1774 merges.

Azorlogh:
@gfyrag re the restore-lifecycle finding: addressed in 5c01ec720 — added `tests/e2e/cluster/restore_cluster_policy_test.go`. A post-checkpoint `SetClusterPolicy` lands in the exported delta; after download/finalize/rebuild the restored node reconstructs it, passes `CheckStore`, and opens the write gate (readiness/reconciliation).

internal/bootstrap/module.go
```diff
@@ -1710,6 +1725,64 @@ func proposeClusterConfigIfNeeded(n *node.Node, builder *plan.Builder, store *da
...
+	applied, err := query.ReadClusterPolicy(store)
+	if err != nil {
+		logger.WithFields(map[string]any{"error": err}).Errorf("Reading applied cluster policy for reconciliation")
+
+		return
+	}
+
+	desired := &commonpb.ClusterPolicy{
+		Revision:             cfg.ClusterPolicyRevision,
+		IdempotencyTtlMicros: uint64(cfg.IdempotencyTTL.Microseconds()),
```
NumaryBot, line 1749: 🟠 **[major] Reject negative idempotency TTLs before unsigned conversion**

When `--idempotency-ttl` is negative, converting its microseconds to `uint64` wraps it into a huge positive TTL, which is then committed as cluster policy instead of rejecting invalid configuration. Validate the duration as nonnegative before persisting it, consistent with the boot-validation requirement in [AGENTS.md:13-15](AGENTS.md#L13-L15).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1750#issuecomment-5369426653

internal/bootstrap/module.go
Azorlogh: Fixed in 03dae47ef — `Config.Validate()` now rejects a negative `--idempotency-ttl` at boot (before the reconciler's unsigned uint64 conversion), with a regression test. https://github.com/formancehq/ledger/commit/03dae47ef

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1750#issuecomment-5369426653

flemzord (commented):
One actionable issue found during the latest review pass.

misc/proto/raft_cmd.proto
```diff
@@ -114,6 +114,7 @@ message SystemScopedOrder {
     DeleteQueryCheckpointOrder delete_query_checkpoint = 14;
     SetQueryCheckpointScheduleOrder set_query_checkpoint_schedule = 15;
     DeleteQueryCheckpointScheduleOrder delete_query_checkpoint_schedule = 16;
+    SetClusterPolicyOrder set_cluster_policy = 17;
```
flemzord, line 117: This new system-scoped order also needs to be added to `systemScopedOrderType` in `internal/domain/audit_order_type.go`. Without that mapping, it falls through to `"unknown"`, so audit indexing cannot retrieve these changes with `order_type == set_cluster_policy`, and admission metrics use the same ambiguous label. Please return `"set_cluster_policy"` for this variant and add it to `TestAuditOrderType` (ideally with exhaustiveness coverage for future oneof arms).

gfyrag (dismissed):
Independent re-review at exact head `46b045017b1eeaee4e5316f930527109a4f68272`.

The replicated ClusterPolicy establishes a deterministic, audited configuration boundary with monotonic revision handling, durable recovery/rebuild, checker verification, admission readiness gating and retrying leader reconciliation. Node-local values only drive the desired proposal; committed policy controls the replicated state. The latest commit correctly maps `SetClusterPolicy` to `set_cluster_policy` and adds proto-exhaustive audit-order coverage. Current CI is green.

misc/proto/raft_cmd.proto
Azorlogh: Fixed in 46b045017 — added the `set_cluster_policy` arm to `systemScopedOrderType`, so audit indexing (`order_type == set_cluster_policy`) and the admission `order_type` metric both label it correctly instead of `unknown`. Also added a proto-oneof exhaustiveness test (`TestSystemScopedOrderType_ProtoExhaustive`) that walks `SystemScopedOrder.payload` and fails if any future arm falls through to `unknown`. https://github.com/formancehq/ledger/commit/46b045017

gfyrag (approved):
Independent re-review at exact head `6be6e5e9f19bee76035a661e7bc4567791903f92`.

The stack was rebased onto the current `release/v3.0`. Range-diff shows commits 2-8 patch-identical; commit 1 is also patch-identical outside regenerated protobuf output, whose descriptor shifts correctly incorporate the newer base fields. The deterministic policy boundary, checker, restore lifecycle, readiness gate and reconciler semantics reviewed previously remain intact.

No prior finding remains and no new finding was discovered. Targeted admission, checker, bootstrap, processing, backup, health and state tests pass; current functional CI is green.

DECISION: APPROVE
Head reviewed: 6be6e5e9f19bee76035a661e7bc4567791903f92
Blocking findings: 0
Non-blocking findings: 0
Validation considered: range-diff, protobuf alignment, targeted subsystem tests + current CI
Residual risk: MEDIUM

PR #1751, opened 2026-08-21
fix: reclaim WAL after follower snapshot restore

gfyrag:
## Summary

- compute each snapshot file SHA-256 while streaming and verify it on followers
- redownload snapshot files for every new session because manifests no longer bind file contents
- reject duplicate and staging-colliding manifest paths before parallel downloads begin
- serialize snapshot file and WAL-record persistence to prevent concurrent ConfState updates from racing on the same temporary file
- persist received snapshots with a crash-safe snapshot-file → guard-record → HardState → sync-barrier sequence
- release WAL segment locks immediately after installing a follower snapshot

## Compatibility

- no mixed-version snapshot compatibility by design
- deployment requires a coordinated non-rolling upgrade

## Validation

- official AI dev/review loop: approved repeatedly with no findings; residual risk MEDIUM
- `bash scripts/agent-check-full`: PASS on the reviewed functional diff (generation, invariants, build, lint, and full unit suite with race detector)
- E2E business suite with race detector: 560/560 PASS
- E2E cluster suite with race detector: 269/269 PASS
- 3-node model cluster: 4 leader/follower restart cycles, no findings: PASS
- snapshot-persistence race regression: reproduced 5/5 before the fix; PASS 10/10 after the fix under the race detector
- final-base selector regression: reproduced on the base, fixed upstream, and targeted package PASS on final head
- final isolated race run: snapshot/WAL and cold-storage packages PASS; remaining GitHub CI is running on head `8d55f1af8`

Jira: [EN-1835](https://formance-team.atlassian.net/browse/EN-1835)

## Bugfix evidence

DISCOVERY: DIRECT_FIX
BEFORE_FIX: BUG_REPRODUCED
AFTER_FIX: PASS

[EN-1835]: https://formance-team.atlassian.net/browse/EN-1835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/storage/wal/wal_default_test.go
```diff
@@ -104,10 +104,12 @@ func TestPurgeOldWALSegments(t *testing.T) {
...
 	require.GreaterOrEqual(t, segmentsAfterWrite, 3, "writing ~200MB should create at least 3 WAL segments")
 
-	// Create a snapshot at a high index and compact to release locks on old segments.
-	cs := &raftpb.ConfState{Voters: []uint64{1}}
-	require.NoError(t, w.CreateSnapshot(numEntries, cs, nil))
-	require.NoError(t, w.Compact(numEntries))
+	// Applying a received snapshot clears all cached entries. It must release
+	// old segment locks itself because a later Compact call has no entries on
+	// which to operate.
+	require.NoError(t, w.ApplySnapshot(&raftpb.Snapshot{
```
paul-nicolas: [P2][non-blocking] Restore the regression test for `Compact`-driven WAL segment reclamation

**What changed.** This test's trigger was replaced, not added to: `CreateSnapshot(numEntries, cs, nil)` + `Compact(numEntries)` became a single `ApplySnapshot(...)`. `TestPurgeOldWALSegments` is the only test in the repository that asserts a WAL segment file is actually *deleted* (`countWALFiles(...) < segmentsAfterWrite`, line 116). Every other `countWALFiles` assertion in this file is a "must be preserved" equality (lines 194, 214, 245, 317, 340), and the other seven `w.Compact(...)` call sites only assert in-memory entry state and `ErrCompacted`.

**Mutation-proven, both directions.**
- At this head, replacing `s.releaseLockTo(compactIndex)` (`wal_default.go:1009`) with `_ = compactIndex` and running the whole tree -> `go test ./internal/... -count=1` exits 0 with **70 `ok` packages and 0 `FAIL`**, including `ok .../internal/storage/wal` and `ok .../internal/infra/node`. Verbose run confirms the tests really executed (66 `--- PASS`, 0 `--- FAIL`), so this is not a "no tests to run" artifact.
- The only other `.Compact(` sites are production (`node.go:683`, `:761`) and `recovery_test.go:39,:89`, which use `Compact(3)` merely to build a compacted-WAL fixture and assert on `node.Run` error strings -- they never look at segment files. `grep` over `tests/` finds no e2e spec asserting reclamation, so the tagged-vs-CI question does not apply.
- At the merge base `da669fe63`, the equivalent mutation (`err := s.wal.ReleaseLockTo(compactIndex)` -> `var err error`) makes `TestPurgeOldWALSegments` FAIL: `Condition never satisfied ... old WAL segments should be purged` (10.37s).

So the guard existed before this PR and does not exist after it. To be clear, the new path *is* genuinely covered: removing `s.releaseLockTo(snap.GetMetadata().GetIndex())` from `ApplySnapshot` also makes this test fail. The problem is that one trigger displaced the other.

**Why it matters.** `Compact` is the steady-state reclamation path: `doMaintenance` (`node.go:761`) does `CreateSnapshot` then `Compact(capturedIndex - CompactionMargin)` on every node on every `MaintenanceInterval` tick, plus an early compaction at boot (`node.go:683`); `ApplySnapshot` only runs on a follower that fell behind. The PR refactored that exact `Compact` line into the new shared `releaseLockTo` helper, so the less-common path is now the only tested one, and a later change that drops or misroutes the `Compact` call ships green. The failure mode is unbounded WAL segment growth -- the same class this PR exists to fix.

The uncovered class is also wider than "someone deletes line 1009": the two call sites pass different indices (`compactIndex` vs the snapshot index), and `Compact` has an early `return nil` when `len(s.entries) == 0` plus two `ErrCompacted` returns *before* reclamation -- a refactor that moves the release behind any of those, or that changes the unlock/reclaim ordering, is invisible today.

**Resolution.** Keep both triggers: extract the setup into a helper and add a sibling case that drives `CreateSnapshot` + `Compact` and asserts the segment count drops. Resolved when removing `s.releaseLockTo(compactIndex)` from `Compact` makes `./internal/storage/wal/` fail. The 200 MB write costs well under a second per case locally.

docs/technical/architecture/subsystems/storage/follower-sync.md
```diff
@@ -0,0 +1,53 @@
...
+every manifest file again: without a content digest in the manifest, a file
+left by an earlier checkpoint cannot be trusted from path and size alone.
+
+The final rename is the trust boundary. Truncated content, excess content, a
+missing or malformed digest, and a digest mismatch all leave the final path
+untouched.
+
+## WAL reclamation after restore
+
+Installing a received Raft snapshot durably writes the snapshot record and
```
paul-nicolas: [P2][non-blocking] This paragraph reverses the two steps whose ordering is the crash-safety argument

The doc says installing a received snapshot "durably writes the snapshot record and HardState, clears cached entries through the snapshot index, then releases etcd WAL segment locks". The code does the opposite for the first two, and the clear is wider than stated:

- `internal/storage/wal/wal_default.go:896-897` runs `s.snapshot = snap; s.entries = nil` **first**;
- `s.snapshotter.Save(snap)` is at :922, `s.wal.SaveSnapshot(walSnap)` at :933, `s.wal.Save(s.hardState, nil)` at :939;
- `s.releaseLockTo(...)` at :952, after the unlock.

`s.entries = nil` also discards the *entire* entry cache unconditionally, not "through the snapshot index".

**Why this is worth a comment rather than a nit.** `ApplySnapshot` carries a long comment block (:908-919) explaining that this ordering is deliberate and crash-safety relevant, so authoritative subsystem docs stating the reverse order will actively mislead the next reader of that block. It also hides the real error semantics: if any of the three durable writes fails, `ApplySnapshot` returns an error with `s.snapshot` already replaced and the entry cache already emptied, and neither is rolled back. Someone auditing the failure path from this paragraph would conclude in-memory state is untouched on failure, which is false.

**Resolution.** State the actual sequence -- in-memory snapshot pointer and full entry-cache clear, then the snap file / WAL snapshot record / HardState durable writes, then lock release after the mutex is dropped -- and say that a failed durable write leaves the in-memory clear in place. The new inline comment at `wal_default.go:949-950` ("clears every cached entry through its index") needs the same correction.

internal/application/ctrl/snapshot_fetcher_test.go
```diff
@@ -219,6 +219,93 @@ func TestGRPCSnapshotFetcher_HashMismatch(t *testing.T) {
...
+			}},
+		},
+		nil,
+		streams,
+	)
+
+	fetcher := &grpcSnapshotFetcher{client: client, parallelism: 1, retryCount: 1, fileRetryCount: 1}
+	_, err := fetcher.FetchSnapshot(t.Context(), t.TempDir(), nil, 0)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "size mismatch")
```
paul-nicolas: [P2][non-blocking] This assertion passes with the mid-stream size guard deleted

`TestGRPCSnapshotFetcher_RejectsExcessBytes` declares `Size: len(data)-1` (6) and scripts a single 7-byte chunk with `Eof: true`. Two different branches produce an error containing `"size mismatch"`:

- the mid-stream guard, `file_fetcher.go:96-98` -- `"size mismatch for %s: expected %d bytes, got at least %d"`;
- the post-loop check, `file_fetcher.go:124-126` -- `"size mismatch for %s: expected %d bytes, got %d"`.

`require.Contains(err.Error(), "size mismatch")` cannot tell them apart, so the test passes either way. Mutation-proven: replacing the whole `if received > entry.GetSize() { ... }` block with a no-op still gives `ok github.com/formancehq/ledger/v3/internal/application/ctrl 0.697s` for `-run TestGRPCSnapshotFetcher`. `TestGRPCSnapshotFetcher_SizeMismatch` asserts the same shared substring at :277, so it does not pin the guard either.

**Why it matters.** The mid-stream guard is new in this PR and is the only thing bounding how many bytes a peer can make the follower write into the `.tmp` when it streams more than the manifest declares -- the post-loop check fires only after everything is already on disk. The guard itself is correct; this is a test-adequacy gap, so nothing here is a code defect. But an unpinned guard can be removed or reordered silently in a later refactor, and the fallback is "write it all first, reject after".

**Resolution.** Assert the branch-unique substring `"got at least"`, and drive a multi-chunk stream where the excess arrives on a non-EOF chunk so only the in-loop branch can satisfy it. Asserting the `.tmp` size never exceeds the declared size would pin the bound directly.

docs/technical/architecture/subsystems/storage/follower-sync.md
```diff
@@ -0,0 +1,53 @@
...
+
+This protocol revision deliberately does not negotiate checksum behavior and
+does not support mixed-version snapshot synchronization. Deploy it as a
+coordinated, non-rolling upgrade of every node in the cluster: every peer must
+run a version that sends and verifies the EOF digest before normal operation
+resumes.
+
+## Retries and staging
+
+A transient `FetchFile` failure retries that file within the current session.
```
paul-nicolas, line 34: [P2][non-blocking] "The retry rewrites its `.tmp` file from the beginning" is an untested claim, and the mock cannot express the case

The claim is true today, but it rests on exactly two things inside `fetchFileOnce`: `os.Create(tmpPath)` truncating (`file_fetcher.go:75`) and `hash`/`received` being per-call locals (`file_fetcher.go:80-81`). Nothing observes either.

No test exercises a file fetch that fails and then succeeds. The only `failAt` other than `-1` in `snapshot_fetcher_test.go` is line 404 (`CloseSessionAlwaysCalled`, which never recovers), and the reason is structural: `newFileStream` (`snapshot_fetcher_test.go:39-52`) returns `script.failErr` when `idx == script.failAt` **without advancing `idx`**, so a scripted stream that fails once fails forever. "Fail, then succeed" is not expressible with the current helper.

**Why it matters.** This PR is what makes partial-transfer state fatal -- the manifest digest that used to be a second line of defence is gone. Hoisting `hash` or `received` out of `fetchFileOnce`, or switching `os.Create` to an append-mode open, would make every retried file fail permanently with `"hash mismatch"`, turning a transient network blip into a follower that can never sync. No existing test would fail.

**Resolution.** Extend `fileStreamScript` so a scripted failure is consumed once (advance `idx` past `failAt`), then add a test where attempt 1 delivers a partial chunk and fails, attempt 2 succeeds, and assert both the final file content and `fetchFileCalls == 2`.

internal/adapter/grpc/server_snapshot.go
```diff
@@ -97,7 +97,7 @@ func (s *SnapshotServiceServerImpl) PrepareSnapshot(ctx context.Context, req *sn
 	}).Infof("Temporary checkpoint created for follower sync")
 
 	manifestStart := time.Now()
-	manifest, err := buildManifest(checkpointPath)
+	manifest, err := buildManifest(ctx, checkpointPath)
```
paul-nicolas, line 100: [P0][non-blocking for this PR -- PRE-EXISTING, please file a separate ticket]

**Re: `FetchFile` at lines 139-146, not the line this comment is pinned to** (that is the nearest line in this PR's diff). Nothing here is caused by #1751 -- `git blame -L 132,153` puts every guard line on `5a10939b3` (2026-07-02), and this PR's only touch to this file is `buildManifest(ctx, ...)`. Raising it because it is the security contract of the very RPC this PR hardens, and because the diff moves the integrity story onto the fetch path.

**The path guard does not contain `..`.**
```go
relPath  := req.GetPath()
fullPath := filepath.Join(session.checkpointPath, relPath)
resolved, err := filepath.Rel(session.checkpointPath, fullPath)
if err != nil || resolved != relPath { reject }
```
`filepath.Rel` is *defined* to return a `..`-prefixed result when the target is not a descendant, so `Join` -> `Rel` is an identity for any `..` chain `Clean` does not have to clamp, and the guard passes. `streamOneFile` (`file_streaming.go:65`) then `os.Open`s `filepath.Join(dirPath, relPath)` with no second check and streams the bytes. Probe output, base `/var/lib/ledger/data/tmp/follower-sync-1`:

| relPath | guard | opens |
|---|---|---|
| `MANIFEST-000005` | pass | .../follower-sync-1/MANIFEST-000005 |
| `a/../b` | **reject** | (Clean rewrites it) |
| `../x` | **pass** | /var/lib/ledger/data/tmp/x |
| `../../../../../etc/passwd` | **pass** | /var/etc/passwd |
| `../../../../../../secret.key` | **pass** | /secret.key |

Only chains deeper than the base has components get clamped. **With the default relative `--data-dir ./data` (`cmd/server/server.go:114`) the base is `data/tmp/follower-sync-N` and there is no clamp at any depth** -- the escape is unbounded. With an absolute data-dir the caller just needs the right `..` count, and `InvalidArgument` vs a successful stream is a clean oracle.

**Reachability.** The service is registered on the Raft server (`internal/bootstrap/module.go:997`), which binds `0.0.0.0:7777` by default with `host == ""` (`server.go:770`, `cmd/server/server.go:110`). The only gate is `raftAuthInterceptors(cfg.ClusterSecret)`, and `raft_auth.go:28-30` returns **nil interceptors when the secret is empty**. `--cluster-secret` is forbidden without TLS (`config.go:270-273`), so the default `--tls-mode=disabled` posture is necessarily secret-less, i.e. unauthenticated. `RequireClientCert` defaults to false. A caller mints its own session via `PrepareSnapshot`, which carries no additional authorization, and `snapshotSession` (`snapshot_session.go:21-25`) stores only `syncName`/`checkpointPath`/`lastAccess` -- the manifest is never retained, so `FetchFile` cannot and does not check `relPath` against it.

**Impact.** Default posture: unauthenticated remote arbitrary file read over the Raft port (TLS private key, Raft WAL, audit store, host files) -- P0. Hardened posture (TLS + operator-minted secret + network policy): an already-trusted peer reads files it is not entitled to -- still a security-boundary bypass. Read-only; no write primitive.

**Why this repo should treat it as a gap rather than an accepted risk:** the sibling restore path already does it correctly. `server_restore.go:56-90` `safeStagingPath` rejects empty names, absolute paths and `..` after `Clean`, then re-verifies with `Rel` as defence in depth, with a comment saying it "must reject any input that does not land inside stagingDir". `grep` finds no `safeStagingPath` or `filepath.IsLocal` on the snapshot path, and `server_snapshot_test.go` has no `..` case.

**Resolution.** Reuse the `safeStagingPath` shape, or better, validate `relPath` against the session's manifest entry set -- that is the actual contract of the RPC -- plus a table-driven rejection test covering `../x` and a matched-depth `../../../..../etc/passwd`.

internal/storage/wal/wal_default.go
```diff
@@ -930,15 +931,25 @@ func (s *DefaultWAL) ApplySnapshot(snap *raftpb.Snapshot) error {
...
 	}
 
 	// Safe to clean up old snap files now — the WAL record is persisted.
 	s.snapshotter.CleanupOlderThan(snap.GetMetadata().GetIndex())
+	s.mu.Unlock()
+
+	// Applying a received snapshot clears every cached entry through its
+	// index. Release the matching etcd WAL segment locks immediately; this
+	// node may not apply local entries soon enough for Compact to do it.
+	s.releaseLockTo(snap.GetMetadata().GetIndex())
```
paul-nicolas, line 974: [P2][non-blocking] Document or enforce the cross-file durability precondition this release depends on

**This is not a live bug -- the current path is safe.** I chased a suspected "released segments + non-durable HardState = unbootable follower" regression here and refuted it. Recording the evidence so it does not have to be re-derived:

- `s.wal.Save(s.hardState, nil)` at :939 does **not** fsync on its own. `WAL.Save` syncs only when `raft.MustSync(st, prev, len(ents))` is true (`server/v3@v3.7.0/storage/wal/wal.go:973,989-996`), and `MustSync = entsnum != 0 || st.Vote != prev.Vote || st.Term != prev.Term` (`raft/v3@v3.7.0/rawnode.go:197`) -- it ignores `Commit`, which is the only field this path changes when the term is unchanged.
- It is durable anyway, for two reasons outside this file. `internal/infra/node/node.go:1182` calls `node.wal.Append(rd.HardState, rd.Entries)` **before** `:1229` calls `node.wal.ApplySnapshot(rd.Snapshot)`, on the same `rd`; for a Ready carrying a snapshot, `rd.HardState.Commit == snapshot.Index` (`raft/v3 log.go:468` `restore` sets `l.committed = s.Metadata.Index`, and `rawnode.go:152` includes the HardState because it now differs from `prevHardSt`). And `wal.SaveSnapshot` at :936 syncs unconditionally (`wal.go:1023`), where `WAL.sync()` **flushes the encoder first** and then `Fdatasync`s (`wal.go:832-836`) -- so it makes the *earlier* buffered HardState record durable too.

Verified by probe: with the production `Append` -> `ApplySnapshot` sequence, a crash image taken after the purger runs boots clean and `ValidSnapshotEntries` retains the new snapshot record. Drop the preceding `Append` and the same image fails to boot with `wal: file not found which matches the snapshot index '5'`, because `ValidSnapshotEntries` (`wal.go:648-656`) discards snapshot records newer than the last durable HardState commit and `wal_default.go:320` turns `wal.Open`'s error into a hard `New()` failure with no fallback.

**The finding is the fragility, not the behaviour.** The safety of this line rests entirely on an ordering invariant that lives in a different package and is stated nowhere here. Two things make that easy to break:

1. A second `ApplySnapshot` call site, or a reordering inside `processReady` that moves the `Append` after the install, silently produces the unbootable-follower failure above. The blast radius is a node that fails every subsequent boot with no automatic recovery -- `recoverConfStateFromWALRecords` (:275) recovers the ConfState for the older index and `wal.Open` still fails on the purged segment; there is no `wal.Repair` call anywhere in `internal/`.
2. The `if s.hardState.GetCommit() < snap.GetMetadata().GetIndex()` branch at :900-905 is unreachable on the production path, precisely because `Append` already raised the commit. As written it reads as if `ApplySnapshot` were responsible for establishing that durability itself, which is the opposite of the truth and points a future reader away from the real invariant.

**Resolution** (any one of these): state the precondition in a comment on this line -- "safe because processReady appends rd.HardState, whose Commit == snap.Index, before this call, and SaveSnapshot's sync flushes it" -- or make the file self-sufficient by swapping the `Save`/`SaveSnapshot` order at :936-943 so `SaveSnapshot`'s own sync covers both records under etcd's `w.mu`. Both were verified to also close the synthetic hole. Prefer the reorder over calling `WAL.Sync()` directly: `WAL.Sync()` is `return w.sync()` with no `w.mu` (`wal.go:859`), so it would race with a concurrent `CreateSnapshot` -> `SaveSnapshot` from `doMaintenance`.

paul-nicolas (commented):
## Multi-model review: Claude + Codex

**Head reviewed:** `7d6ce3f296f0c216675e50bcbc36e7ea78a274e9` (first review -- no prior findings on this PR)

The core of the change is sound and, importantly, genuinely tested. I mutation-proved both halves: removing `s.releaseLockTo(snap...)` from `ApplySnapshot` makes `TestPurgeOldWALSegments` fail, and removing the EOF digest from `streamOneFile` fails the streaming tests. The reclamation fix is a real regression test, not a proxy. I also found no way to get a `.tmp` file renamed into place without a verified digest: a stream that ends without `Eof`, an empty or malformed digest, truncation, and excess bytes are all rejected before `os.Rename`, and `os.Create` truncating plus per-attempt locals make retries restart cleanly. The proto change is clean -- no renumbering actually occurred (`FileEntry` only lost its trailing field 3; `FetchFileResponse` gained a fresh field 3), the three generated artifacts are mutually consistent, and no hand-rolled `protowire` site touches this proto.

**6 findings posted inline: 5 x P2 non-blocking, 1 x P0 non-blocking (pre-existing).** No finding blocks this PR.

The P0 needs a word of explanation: `FetchFile`'s path guard at `server_snapshot.go:139-146` is bypassable (`filepath.Rel` returns `..`-prefixed results, so `Join` -> `Rel` is an identity for un-clamped `..` chains), giving arbitrary file read over the Raft port, which is unauthenticated in the default plaintext posture. It is **pre-existing** -- `git blame` puts every guard line on `5a10939b3` (2026-07-02) and this PR does not touch them -- so it does not block here, but it should get its own ticket rather than staying buried in a review thread.

### Rejected at the verification gate

Nine findings entered adversarial verification (one refuter per finding, instructed to refute rather than confirm). Six were confirmed, one was corrected, and three were dropped. The notable rejection:

- **"Released segments + non-durable HardState makes a killed follower permanently unbootable" (proposed P1, blocking) -- INVALID.** Every library fact was right and the crash image reproduced, but the precondition fails: `node.go:1182` appends `rd.HardState` before `:1229` calls `ApplySnapshot` on the same `rd`, and for a snapshot Ready `rd.HardState.Commit == snapshot.Index` (`raft/v3 log.go:468`); `SaveSnapshot`'s unconditional sync then flushes that buffered record (`wal.go:832-836`). A crash image taken after the purger runs boots clean on the production sequence and only fails on a synthetic `ApplySnapshot` with no preceding `Append`. What survives is the *fragility* of that unstated cross-file invariant, filed as the P2 on `wal_default.go:952`.
- **"Progress counters over-count across session retries" -- corrected, downgraded to P3.** The mechanism is real, but the claimed ">100% progress" impact is false: both CLI display paths clamp (`cmd/ledgerctl/cluster/status.go:276,299`). Residual impact is the raw API field and a `filesCompleted > filesTotal` log line.
- **"Clear the staging directory at the start of each session" -- downgraded to P3, pre-existing.** A failed session leaves exactly one `.tmp` before and after this change, so the PR does not make it worse.

Codex reviewed the diff independently with full repo access and returned **0 findings** (`PASS`, risk LOW) across architecture, correctness, concurrency, security, proto contract, performance, error handling, docs, and tests. Every finding below therefore comes from the Claude reviewers; there were no Claude/Codex conflicts to resolve, only Codex silence where Claude found test-adequacy and documentation gaps. Worth noting that Codex ran `scripts/agent-check` inside my review worktree, which deleted 32 generated files mid-review; I restored the tree and re-verified, and no finding rests on that window.

### P3 (not posted inline)

- `internal/application/ctrl/snapshot_fetcher.go:205` -- sync progress counters are add-only and re-count the whole checkpoint on each session retry; display is clamped, so impact is the raw API field and the `filesCompleted`/`filesTotal` log.
- `internal/adapter/grpc/file_streaming.go:88` -- the `(n>0, io.EOF)` branch is unreachable for `os.File`, so the digest assignment there is dead and the `// Empty file:` comment at :104 actually describes the normal termination path for *every* file.
- `internal/adapter/grpc/file_streaming.go:27` -- `ctx.Err()` is checked before the walk error, so a real stat/permission failure during `PrepareSnapshot` is reported as `context canceled`.
- `internal/adapter/grpc/file_streaming.go:35` -- the walk filters only directories while the comment and `follower-sync.md:12` say "regular file"; a symlink would be listed with its link-target length as `Size` and then stream the target's bytes, a guaranteed permanent size mismatch. Harmless today (Pebble checkpoints hard-link).
- `internal/application/ctrl/file_fetcher.go:70` -- `fetchFileOnce` never removes `tmpPath` on failure, so a `.tmp` for a file dropped from a later session's manifest is promoted into the activated checkpoint by `ActivateIncomingRestore`. Pre-existing and bounded.
- `internal/adapter/grpc/server_snapshot_test.go:188` -- `TestSnapshotService_FullRoundTrip`'s digest assertion is now self-consistent (both sides come from `streamOneFile`), where it previously compared against an independently computed manifest digest. It no longer proves delivered bytes equal source bytes, and the written contents are never asserted.
- `internal/application/ctrl/snapshot_fetcher_test.go:219` -- the four rejection tests assert only the error substring, never `require.NoFileExists` on the final path, so nothing observes the trust boundary that `follower-sync.md:39-41` states.
- `internal/storage/wal/wal_default.go:1017` -- `releaseLockTo` only `Errorf`-logs its one possible error, with no metric and no signal to the caller; a permanent failure degrades silently to unbounded WAL growth, which is the class this PR fixes. Consider `assert.Unreachable` (the pattern at :355) or a counter.

```text
DECISION: APPROVE
Head reviewed: 7d6ce3f296f0c216675e50bcbc36e7ea78a274e9
Blocking findings: 0
Non-blocking findings: 14 (1 x P0 pre-existing, 5 x P2, 8 x P3)
Validation considered: targeted grpc/ctrl/snapshotpb/wal suites and -race runs pass; 6 mutation experiments; a path-traversal probe; a crash-image boot probe; repo-wide `go test ./internal/...` under the Compact mutation (70 ok, 0 FAIL)
Residual risk: LOW
```

The two things I would actually gate on before merge are cheap: restore the `Compact` purge test, and fix the `follower-sync.md` ordering sentence. Neither is blocking.

internal/storage/wal/wal_default.go
```diff
@@ -929,16 +937,31 @@ func (s *DefaultWAL) ApplySnapshot(snap *raftpb.Snapshot) error {
...
 		ConfState: snap.GetMetadata().GetConfState(),
 	}
-	if err := s.wal.SaveSnapshot(walSnap); err != nil {
-		return fmt.Errorf("saving snapshot to WAL: %w", err)
+	// Buffer HardState before the snapshot record. A commit-only HardState
+	// update does not make etcd WAL Save sync by itself; SaveSnapshot always
+	// syncs and therefore durably flushes both records in this order. This keeps
+	// ApplySnapshot self-sufficient instead of depending on processReady having
+	// appended the same HardState first.
+	if err := s.wal.Save(s.hardState, nil); err != nil {
```
NumaryBot, line 955: 🔴 **[blocker] Avoid persisting HardState before its snapshot record**

If `Save` succeeds but `SaveSnapshot` fails (for example due to disk exhaustion), the commit-only HardState record has already been written and may later become durable through close, another WAL sync, or a crash flush, while no matching WAL snapshot record exists. Restart can then load a commit index beyond the old recognized snapshot and retained entries, leaving Raft unable to recover. The two records need failure-safe atomic ordering rather than relying solely on `SaveSnapshot` to flush the preceding write.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1751#issuecomment-5370148053

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1751#issuecomment-5370148053

Azorlogh (approved):
Approving.

I verified the two crash-safety commits independently rather than taking them on the comment text:

- `059d82877`: replayed the intermediate durable prefix of the *previous* ordering (snap file + durable `HardState.Commit=N`, no WAL snapshot record at N) and handed the storage to `raft.NewRawNode` — it panics with `state.commit 10 is out of range [2, 3]`, permanently, since nothing can advance afterwards. The guard-record prefix boots cleanly instead, because `wal.ValidSnapshotEntries` drops records above the durable commit. Real hole, correctly closed.
- `dc919ff3f`: red without `snapshotPersistenceMu` in 4 of 5 runs — concurrent `UpdateSnapshotConfState` collide on `<term>-<index>.snap.tmp` and fail with `renaming temp snap file: ... no such file or directory`. Lock order is `snapshotPersistenceMu` → `mu` at all three sites and nowhere reversed.

**Non-blocking:** `SyncProgress` now over-reports. `bytesReceived`/`filesCompleted` are add-only with no reset (`sync_progress.go:22,26`), so once each session refetches the whole manifest (`snapshot_fetcher.go:207`), a session retry re-counts every file the previous session already downloaded. `BytesReceived` can exceed `BytesTotal` and `FilesCompleted` exceed `FilesTotal`, surfaced unclamped at `node.go:2104`. Resetting both next to the existing `SetTotal`/`SetFilesTotal` at `snapshot_fetcher.go:211-212` would cover it, and would also fix the smaller pre-existing double count from the per-file retry loop.

PR #1758, opened 2026-08-21
fix: harden local AI PR validation

gfyrag:
## Summary

- accept trusted base-pinned review adapters without an executable bit, since the loop invokes them explicitly through Bash
- make the trusted Schemathesis runner configure historical balances when the candidate supports them and retry only `HISTORY_BUILDING` / `HISTORY_BEHIND` for a bounded interval
- isolate API conformance from host disk occupancy and add focused regression tests and documentation

## Validation

- `nix develop --command env -u GOROOT bash -c "just pre-commit"`
- `nix develop --command env -u GOROOT go build ./...`
- `go test ./scripts/aiprloop ./scripts/reviewloop ./scripts/aireviewcodex ./scripts/aifixclaude -count=1`
- `pytest -q tests/schemathesis/test_test_api.py` (4 passed)
- `nix develop --command env -u GOROOT bash tests/schemathesis/run.sh` (64/64 endpoints passed)

## Safety

The retry is restricted to the historical volumes GET and the two documented replica-local transient error codes. It is bounded; persistent lag and all other 5xx remain blocking. A cross-run against PR #1662 still fails on its persistent `HISTORY_BUILDING` state, confirming that this change does not mask that underlying projection issue.

### DIFF
--- a/docs/technical/contributing/ai-review-loop.md
+++ b/docs/technical/contributing/ai-review-loop.md
@@ -216,8 +216,16 @@ PR worktree as the recipe working directory. A PR therefore cannot replace a
 required recipe with a no-op. Script-backed gates are split the same way:
 Schemathesis and model-checker runners come from the trusted base, while their
 server build, OpenAPI input, and other reviewed sources come from the candidate
-worktree. The model driver/oracle is also base-pinned. The loop does not query
-or wait for GitHub Actions checks.
+worktree. When the candidate exposes historical balances, the trusted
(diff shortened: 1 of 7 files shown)


gfyrag:
SUPERSEDED_BY_SPLIT

The still-valid scope was rebuilt from the current `release/v3.0` and split into focused PRs:

- #1810 — base-pinned adapter file mode
- #1811 — Schemathesis host disk isolation

The obsolete historical/tooling scope from #1758 was not carried forward.

PR #1764, opened 2026-08-24
fix(EN-1850): preserve kubectl rollout-restart annotation on the cluster StatefulSet

sylr:
## What changed

The operator now carries the `kubectl.kubernetes.io/restartedAt` pod-template annotation over from the live StatefulSet when rebuilding the template from the Cluster CR, in both `CreateOrUpdate` mutate paths (main pass and the scale-down pre-update). Adds unit tests for the helper, an envtest integration test proving the stamp survives a CR-driven template rewrite, and a README note.

## Why

`kubectl rollout restart statefulset` stamps `restartedAt` on the pod template. The reconciler rebuilt the template from scratch, so the reconcile triggered by kubectl's own patch erased the stamp and cancelled the rolling restart after the first (highest-ordinal, `-2`) pod — leaving `-1`/`-0` on the old revision. Same class of bug as formancehq/operator#353.

Jira: [EN-1850](https://formance-team.atlassian.net/browse/EN-1850)

Need: operators must be able to perform a full rolling restart of a ledger cluster with the standard kubectl verb.
Current limitation: the restart silently stops after one pod, which reads as a completed restart.
Requirement / constraint: a `kubectl rollout restart` must roll every pod; operator-owned annotations (`spec-hash`, `auth-keys-hash`) must remain CR-derived.
Evidence: EN-1850; prior art formancehq/operator#353.
Durable repository evidence: `misc/operator/internal/controller/labels.go` (annotation constant comment), `misc/operator/README.md` restart section.

## Technical decision

Decision: treat the kubectl stamp as user-owned state and carry only that one key over from the existing template.
Why now / why proportionate: one-line semantic change per mutate path; no new CRD surface.
Alternatives considered: documenting `kubectl rollout restart` as unsupported (rejected — a half-completed rollout is a silent failure mode); preserving all `kubectl.kubernetes.io/*` annotations (rejected — only `restartedAt` has defined template semantics, a blanket carry-over could mask drift).

## Risk

**LOW** — scoped to the operator's StatefulSet template reconciliation; the spec-hash rolling-update mechanism is untouched.

## Validation

- [x] `bash scripts/agent-check` (PASS, root + operator lint 0 issues)
- [x] Targeted tests: `go test ./internal/controller/` (operator unit suite) and `go test -tags integration ./internal/controller/` (envtest) including the new `TestWithKubectlRestartedAt` and `TestReconcile_PreservesKubectlRestartedAt`
- [x] External review: Codex (gpt-5.5) — APPROVE, 0 blocking findings, 1 non-blocking P3 (see Known concerns)

Operator-only. The StatefulSet pod template may now carry one non-CR-derived annotation; it does not participate in the spec hash. On StatefulSet recreation (VolumeClaimTemplate change) the stamp is still lost — the README recommends `kubectl ledger restart` (CR-recorded) for that reason.

## Review focus

The interaction between the carried-over annotation and the two mutate paths in `reconcile_statefulset.go` — especially the scale-down pre-update.

## Known concerns

Codex P3 (non-blocking): if a GitOps flow manages the exact key `kubectl.kubernetes.io/restartedAt` through `spec.podAnnotations` after a prior direct `kubectl rollout restart`, the live stamp wins over the CR value until StatefulSet recreation (the rollout itself is unaffected — the spec hash still changes). Can be addressed by preserving only when the desired template lacks the key, if that usage ever materializes.

[EN-1850]: https://formance-team.atlassian.net/browse/EN-1850?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/misc/operator/internal/controller/reconcile_statefulset.go
+++ b/misc/operator/internal/controller/reconcile_statefulset.go
@@ -507,7 +507,10 @@ func buildPodTemplate(ledger *ledgerv1alpha1.Cluster, specHash string, credentia
 //
 // A value the CR explicitly provides through spec.podAnnotations wins over the
 // live stamp: the CR is the source of truth, and only a key it does not manage
-// is treated as user-owned StatefulSet state.
+// is treated as user-owned StatefulSet state. Consequently, when the CR manages
+// this key a direct `kubectl rollout restart` is reverted on the next reconcile;
+// that configuration must restart through the CR (documented in the operator
+// README).
 //
 // The annotations map is copied before mutation: the desired template built by
 // buildStatefulSetSpec may be shared across retries of a mutate function.
--- a/misc/operator/internal/controller/reconcile_statefulset.go
+++ b/misc/operator/internal/controller/reconcile_statefulset.go
@@ -505,11 +505,18 @@ func buildPodTemplate(ledger *ledgerv1alpha1.Cluster, specHash string, credentia
 // reconcile, so dropping the stamp would revert kubectl's patch and cancel the
 // rolling restart after the first (highest-ordinal) pod (EN-1850).
 //
+// A value the CR explicitly provides through spec.podAnnotations wins over the
+// live stamp: the CR is the source of truth, and only a key it does not manage
+// is treated as user-owned StatefulSet state.
+//
 // The annotations map is copied before mutation: the desired template built by
 // buildStatefulSetSpec may be shared across retries of a mutate function.
 func withKubectlRestartedAt(existing, desired corev1.PodTemplateSpec) corev1.PodTemplateSpec {
 	restartedAt, ok := existing.Annotations[annotationKubectlRestartedAt]
 	if !ok {
+		return desired
+	}
+	if _, crManaged := desired.Annotations[annotationKubectlRestartedAt]; crManaged {
 		return desired
 	}
 
--- a/misc/operator/internal/controller/reconcile_statefulset.go
+++ b/misc/operator/internal/controller/reconcile_statefulset.go
@@ -90,7 +90,7 @@ func (r *ClusterReconciler) reconcileStatefulSet(ctx context.Context, ledger *le
 					sts.Spec = desired
 				} else {
 					sts.Spec.Replicas = desired.Replicas
-					sts.Spec.Template = desired.Template
+					sts.Spec.Template = withKubectlRestartedAt(sts.Spec.Template, desired.Template)
 					sts.Spec.UpdateStrategy = desired.UpdateStrategy
 					sts.Spec.PersistentVolumeClaimRetentionPolicy = desired.PersistentVolumeClaimRetentionPolicy
 					sts.Spec.MinReadySeconds = desired.MinReadySeconds
@@ -155,7 +155,7 @@ func (r *ClusterReconciler) reconcileStatefulSet(ctx context.Context, ledger *le
 			// ServiceName, Selector, PodManagementPolicy and
 			// VolumeClaimTemplates are immutable after creation.
 			sts.Spec.Replicas = desired.Replicas
-			sts.Spec.Template = desired.Template
+			sts.Spec.Template = withKubectlRestartedAt(sts.Spec.Template, desired.Template)
 			sts.Spec.UpdateStrategy = desired.UpdateStrategy
 			sts.Spec.PersistentVolumeClaimRetentionPolicy = desired.PersistentVolumeClaimRetentionPolicy
 			sts.Spec.MinReadySeconds = desired.MinReadySeconds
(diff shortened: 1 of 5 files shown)


misc/operator/internal/controller/reconcile_statefulset.go
```diff
@@ -498,6 +498,29 @@ func buildPodTemplate(ledger *ledgerv1alpha1.Cluster, specHash string, credentia
...
+// buildStatefulSetSpec may be shared across retries of a mutate function.
+func withKubectlRestartedAt(existing, desired corev1.PodTemplateSpec) corev1.PodTemplateSpec {
+	restartedAt, ok := existing.Annotations[annotationKubectlRestartedAt]
+	if !ok {
+		return desired
+	}
+
+	annotations := make(map[string]string, len(desired.Annotations)+1)
+	maps.Copy(annotations, desired.Annotations)
+	annotations[annotationKubectlRestartedAt] = restartedAt
```
NumaryBot, line 528: 🟠 **[major] Let CR-provided restart stamps take precedence**

When `spec.podAnnotations` explicitly supplies `kubectl.kubernetes.io/restartedAt`, this assignment replaces that desired value with the old live StatefulSet value. After a direct rollout stamp exists, a GitOps/CR update of the same annotation therefore cannot trigger the requested new rollout. Preserve the live stamp only when the desired template does not already define this key.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1764#issuecomment-5394910796

misc/operator/internal/controller/reconcile_statefulset.go
sylr: Done — fixed in commit 5f80c47f5: the live stamp is now carried over only when the desired template does not already define the key, so a restartedAt explicitly managed through spec.podAnnotations wins (unit test added). One correction to the finding: a CR update of this annotation could always trigger a rollout even before this fix, because spec.podAnnotations participates in computeSpecHash (hash.go) and the resulting spec-hash bump changes the template — only the annotation value itself was masked.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1764#issuecomment-5394910796

gfyrag:
## Codex review — REQUEST CHANGES

Reviewed HEAD `5f80c47f500bef226f983fe967815da9fa82d24c` on the synthetic merge result with current `release/v3.0` at `45684f4a55530ea14b5de38d4feef4fe489fbcbf` (stale-base gate explicitly bypassed for this review).

[P2][blocking] Document the CR-managed `restartedAt` exception

Location: `misc/operator/README.md:271`, `misc/operator/internal/controller/reconcile_statefulset.go:519`

Evidence: the README states that plain `kubectl rollout restart` works, but `withKubectlRestartedAt` deliberately restores the CR value whenever `spec.podAnnotations` already contains `kubectl.kubernetes.io/restartedAt`. In that supported configuration, the direct kubectl stamp is reverted on reconcile.

Impact: the restart can be cancelled or remain partial—the same silent operational failure class this PR is intended to fix.

Resolution: document this exact exception and direct users to `kubectl ledger restart` in that case, or support the ownership conflict in code.

Previous finding: the CR-precedence finding is FIXED at HEAD.

Validation:
- `bash scripts/agent-check`: PASS
- `go -C misc/operator test ./...`: PASS
- targeted envtest (`TestWithKubectlRestartedAt`, `TestReconcile_PreservesKubectlRestartedAt`): PASS
- full envtest has 3 unrelated validation-test failures, reproduced unchanged on current `release/v3.0`

Blocking findings: 1
Non-blocking findings: 0
Residual risk: MEDIUM

sylr:
Re Codex review [P2] — Done in commit 1e1f650b6. The README restart section now states the exception explicitly: when the Cluster CR itself sets `kubectl.kubernetes.io/restartedAt` in `spec.podAnnotations`, the CR value is authoritative, a direct `kubectl rollout restart` stamp is reverted on the next reconcile, and that configuration must restart through the CR (`kubectl ledger restart` or bumping the annotation in the CR). The `withKubectlRestartedAt` doc comment carries the same note. Chose documentation over resolving the ownership conflict in code (e.g. newest-timestamp-wins): it would compete with the CR-as-source-of-truth rule for an edge-case configuration.

gfyrag (approved):
Re-review complete at HEAD 1e1f650b6ae0e7ccd01dd0eca74830127a76fd50. Previous P2 finding is fixed; no blocking or non-blocking findings remain. Baseline validation passed on the synthetic merge with the current release/v3.0 tip.

PR #1767, opened 2026-08-24
feat(EN-1501): enforce the query-checkpoint limit in the FSM (stacked on #1750)

Azorlogh:
## What

PR 2 of [EN-1827](https://formance-team.atlassian.net/browse/EN-1827): move the **query-checkpoint limit + existence enforcement into the FSM apply path** (after the idempotency gate), using the Raft-replicated `ClusterPolicy.query_checkpoint_limit` introduced in PR 1.

This replaces the soft-admission approach of #1660 (which enforced at admission and, as EN-1827 notes, broke idempotent retries). **Supersedes #1660.**

> ⚠️ **Stacked on #1750** (PR 1 — the replicated cluster policy). Base is `feat/cluster-policy`; the diff shown here is PR 2 only. Retarget to `release/v3.0` once #1750 merges.

## How

- **FSM enforcement.** `processCreateQueryCheckpoint` rejects when `live_count >= ClusterPolicy.query_checkpoint_limit` (`ErrCheckpointLimitReached`); `processDeleteQueryCheckpoint` rejects a non-live id (`ErrCheckpointNotFound`), keeping the structural `id != 0` check. Both run after the idempotency gate, so a committed order resolves identically on every node and a keyed retry replays the first apply's frozen outcome. `limit == 0` (the unset default before a policy is committed) is uncapped.
- **Deterministic live state.** `FSMState.LiveQueryCheckpointIDs` — a set seeded at boot from the stored rows (`ReadLiveQueryCheckpointIDs`) and kept in lockstep by the handlers and `WriteSet.Merge`. Apply enforces the cap/existence with **no Pebble read** (invariant #3). It's a function of the already-checker-verified rows, so it adds no new persisted projection.
- **`ErrCheckpointLimitReached` is `KindPrecondition` (freezable), not `ResourceExhausted`.** EN-1827 requires a limit failure to stay frozen under an idempotency key ("remains the same outcome even after a slot becomes free"); a retryable kind wouldn't be frozen, so a keyed retry could re-execute. `FailedPrecondition` is the closest freezable fit. This is a client-facing change from #1660's `ResourceExhausted`.
- **Checker (invariant #8).** `compareQueryCheckpoints` re-derives the live set (with each row's `max_sequence` / `created_at`) from the chain-bound `CreatedQueryCheckpoint` / `DeletedQueryCheckpoint` logs (baseline-seeded under archiving) and diffs it against the stored rows both ways → `QUERY_CHECKPOINT_MISMATCH`. Closes a pre-existing gap on the checkpoint rows.
- **Rebuild / baseline.** Audit rebuild recreates the rows and advances the monotonic next-id counter; baseline snapshots copy the checkpoint rows for the checker.

## Testing

- Unit: enforcement (create under/at limit/uncapped, delete live/non-live/zero-id), checker (×7), rebuild replay, recovery live-set roundtrip.
- **e2e** (`Query Checkpoints live cap`): boots a node with a low `--query-checkpoint-limit`, then create-to-cap → reject `FailedPrecondition`/`CHECKPOINT_LIMIT_REACHED` → delete frees a slot → `CHECKPOINT_NOT_FOUND`.
- `go build ./...` + `go vet` (incl. `-tags e2e`) + targeted package tests green. (Full `agent-check` not runnable here: golangci-lint OOMs, `just pre-commit` deletes `.pb.go`.)

## Notes / follow-ups

- Physical **orphan-dir reclamation deferred** (needs a `ListQueryCheckpointDirs` capability port; disk hygiene, per-delete cleanup already exists).
- Built fresh on the cluster-policy branch rather than replaying #1660, so there's no admission enforcement to remove.

v3 is unreleased: no migration/compat shims, no reserved proto fields.

[EN-1827]: https://formance-team.atlassian.net/browse/EN-1827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/application/check/checker.go
+++ b/internal/application/check/checker.go
@@ -401,6 +401,15 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 		// SubAttrNumscriptVersion (latest pointer = greatest stored semver).
 		expectedNumscriptContent = make(map[domain.NumscriptEntryKey]*commonpb.NumscriptInfo)
 		expectedNumscriptLatest  = make(map[domain.NumscriptVersionKey]string)
+
+		// derivedLiveCheckpoints is the audit-derived set of live query checkpoints
+		// (cluster-global), keyed by id with the CreatedQueryCheckpoint log as the
+		// value so max_sequence / created_at can be verified. A create sets the
+		// entry, a later delete removes it. Seeded from the baseline under archiving
+		// (the create logs are purged with the chapter), then advanced by the
+		// replayed create/delete logs. compareQueryCheckpoints diffs it against the
+		// stored rows both ways, contents included.
+		derivedLiveCheckpoints = make(map[uint64]*commonpb.CreatedQueryCheckpointLog)
 	)
 
 	// excluded is built incrementally as SimulateEphemeralPurge decides to
@@ -490,6 +499,13 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 		// Pre-populate the reversion tracking from the baseline's transaction
 		// states so reversion invariant checks cover pre-archive transactions.
 		if err := c.seedTxTrackingFromBaseline(baselineDB, ledgerKnownTxIDs, ledgerRevertedTxIDs); err != nil {
+			return err
+		}
+
+		// Seed the live query-checkpoint set from the baseline: pre-archive
+		// CreatedQueryCheckpoint logs are purged with the archived chapter, so
+		// replay alone would under-derive the set and false-flag live rows.
+		if err := c.foldBaselineQueryCheckpoints(baselineDB, derivedLiveCheckpoints); err != nil {
 			return err
 		}
 	}
@@ -645,6 +661,14 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 					if cur, ok := expectedNumscriptLatest[vk]; !ok || numscriptVersionGreater(info.GetVersion(), cur) {
 						expectedNumscriptLatest[vk] = info.GetVersion()
 					}
+				}
+			case *commonpb.LogPayload_CreatedQueryCheckpoint:
+				if cp := payload.CreatedQueryCheckpoint; cp != nil {
+					derivedLiveCheckpoints[cp.GetCheckpointId()] = cp
+				}
+			case *commonpb.LogPayload_DeletedQueryCheckpoint:
+				if cp := payload.DeletedQueryCheckpoint; cp != nil {
+					delete(derivedLiveCheckpoints, cp.GetCheckpointId())
 				}
 			case *commonpb.LogPayload_Apply:
 				if payload.Apply != nil {
@@ -895,6 +919,10 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 	}
 
 	c.compareNumscripts(snap, expectedNumscriptContent, expectedNumscriptLatest, deletedInReplay, pendingCleanupLedgers, callback)
+
+	if err := c.compareQueryCheckpoints(snap, derivedLiveCheckpoints, callback); err != nil {
+		return err
+	}
 
 	return nil
 }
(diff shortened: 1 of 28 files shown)


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1767#issuecomment-5397313936

internal/infra/state/write_set.go
```diff
@@ -1848,6 +1858,47 @@ func (b *WriteSet) DeleteQueryCheckpoint(checkpointID uint64) {
...
+func (b *WriteSet) LiveQueryCheckpointCount() uint64 {
+	n := len(b.fsm.State.LiveQueryCheckpointIDs)
+
+	for _, cp := range b.pendingQueryCheckpointSaves {
+		if _, ok := b.fsm.State.LiveQueryCheckpointIDs[cp.GetCheckpointId()]; !ok {
+			n++
+		}
+	}
+
+	for _, id := range b.pendingQueryCheckpointDeletes {
```
NumaryBot, line 1893: 🟠 **[major] Subtract deletes of checkpoints staged in the same proposal**

When a proposal creates a checkpoint and then deletes that new ID, the save increments `n`, but this loop only subtracts IDs present in the pre-proposal live set. A later create in the same proposal therefore sees an inflated count and can incorrectly hit the cap even though the staged checkpoint will be deleted. Account for staged saves when applying staged deletes, as required by the live-set contract in [query-checkpoints.md:45](docs/technical/architecture/subsystems/read-path/query-checkpoints.md#L45).

Azorlogh: Fixed in 4b831d55e — `LiveQueryCheckpointCount` now computes the effective live set directly (recovered set + staged creates − staged deletes), so a checkpoint created and then deleted in the same proposal nets to zero and a later create isn't spuriously capped. Added a regression test (TestWriteSet_LiveQueryCheckpointCount_FoldsStagedChanges).

https://github.com/formancehq/ledger/commit/4b831d55eaacaca9215a283c4b367560cb0d0f9d

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1767#issuecomment-5397313936

gfyrag (commented):
Independent re-review at head f2a96e22cc57c569e088444554ef633490cd0c5f. One blocking finding remains.

[P2][blocking] Gate business writes before the first replicated policy revision
Location: internal/adapter/grpc/server_cluster.go (CreateQueryCheckpoint), internal/application/admission/admission.go (Admit), tests/e2e/cluster/query_checkpoint_test.go (waitForServing)
Evidence: the latest commit only waits for SERVING in the E2E. GRPCHealthUpdater publishes NOT_SERVING but does not reject RPCs, and the production admission WriteGate has no committed-policy revision check. A direct or already-connected client can still submit before revision 1; the FSM then treats limit 0 as uncapped. The prior E2E failure reproduced this window.
Impact: checkpoints beyond the configured limit can be committed permanently during startup.
Resolution: gate business admission until a replicated policy is committed, explicitly exempt SetClusterPolicy, and retain a deterministic pre-policy write regression test.

Targeted unit tests, E2E package compilation, and functional CI passed; patch coverage is failing.

Azorlogh:
@gfyrag re the pre-policy write window: resolved by a hard admission gate in PR 1 (5c01ec720) — `Admit` holds business writes until a policy is committed (`CreateQueryCheckpoint` flows through `Admit`, so it can't precede revision 1); `SetClusterPolicy` is exempt. #1767 now inherits it via rebase, and the e2e `waitForServing` workaround is dropped.

internal/application/check/checker.go
```diff
@@ -896,6 +920,10 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 
 	c.compareNumscripts(snap, expectedNumscriptContent, expectedNumscriptLatest, deletedInReplay, pendingCleanupLedgers, callback)
 
+	if err := c.compareQueryCheckpoints(snap, derivedLiveCheckpoints, callback); err != nil {
```
flemzord, line 932: This comparison is never reached on the zero-log fast path: `Check` returns early when `lastSequence == 0`. In that state the derived set is provably empty, so an injected `SubGlobQueryCheckpoint` row should still be reported, but currently passes unchecked and is later loaded into `LiveQueryCheckpointIDs`. Please run this comparison with an empty derived set before the early return and add a full `Checker.Check` regression for a stored checkpoint with no logs.

flemzord (commented):
One correctness gap found during the latest review pass.

gfyrag (dismissed):
Independent re-review at exact head `bc60f05a0fdeba5facbddbfaf7806ec33189058a`.

The replicated checkpoint cap and existence checks run deterministically in FSM state after the idempotency gate, with no Pebble reads in apply. Recovery, rebuild, checker and same-proposal create/delete folding are aligned. The new zero-log checker wiring correctly rejects unaudited stored checkpoint rows, and the cluster-policy audit-order type is now exhaustively mapped.

Approve as the PR2 stack on #1750. Functional CI is green; patch coverage remains the only red status.

gfyrag (dismissed):
Independent re-review at exact head `de43da87ca3576f66818f5b889431fa917fb06d7`.

The PR2 stack was rebased onto the current #1750. Range-diff shows commits 2-5 patch-identical; commit 1 is patch-identical outside regenerated protobuf output, which correctly follows the updated base schema. FSM enforcement remains after the idempotency gate, uses the replicated policy and live set without Pebble reads, and keeps rebuild/checker/zero-log coverage aligned.

No prior finding remains and no new finding was discovered. Targeted checker, processing, backup, state and query tests pass. Functional CI is green; patch coverage remains red but does not expose an uncovered correctness branch in the reviewed rebase.

DECISION: APPROVE
Head reviewed: de43da87ca3576f66818f5b889431fa917fb06d7
Blocking findings: 0
Non-blocking findings: 0
Validation considered: range-diff, protobuf alignment, targeted subsystem tests + current CI
Residual risk: MEDIUM

internal/application/check/checker.go
Azorlogh: Fixed in de43da87c — the zero-log `Check` block now runs `compareQueryCheckpoints(snap, nil, callback)` (mirroring the signing and cluster-policy passes there), so an injected `SubGlobQueryCheckpoint` row on a zero-log store is reported instead of being loaded into `LiveQueryCheckpointIDs` unchecked. Added a full `Checker.Check` regression (`TestCheck_QueryCheckpointProjection_EmptyAuditWiring`) for a stored checkpoint with no logs. https://github.com/formancehq/ledger/commit/de43da87c

gfyrag (approved):
Independent re-review at exact head 3fec23beead6a15684cdd4ee38a179cce92393d5.\n\nThe five EN-1501 commits are patch-identical to the previously approved stack after the cluster-policy base was merged into release/v3.0. FSM enforcement remains deterministic and after the idempotency gate, with aligned recovery, rebuild, checker, and zero-log coverage. No prior finding remains and no new finding was found. Functional CI is green; patch coverage is the only failing status.\n\nDECISION: APPROVE\nBlocking findings: 0\nNon-blocking findings: 0\nResidual risk: MEDIUM

PR #1766, opened 2026-08-24
refactor(indexes): drop the dead BuildStatus from the registry row (EN-1858)

Azorlogh:
[EN-1858](https://formance-team.atlassian.net/browse/EN-1858)

## Why

`Index.build_status` is dead state, and misleading on the API:

- EN-1323 (#553) replaced the cluster-wide `IndexReady` flip with per-replica `IndexVersionState` and deleted the only READY writer (`applyIndexReady`, `TechnicalUpdate.index_ready`, the indexer proposal plumbing).
- #553 meant to freeze the field ("informational only") and removed the `setMetadataFieldType` BUILDING flip — but the registry-extraction PR #453, merged the same day, carried the flip into the new bucket-scoped store.
- Net result since June: creation writes BUILDING, retype re-flips BUILDING, **nothing ever writes READY**. No reader gates on the field, and `GetIndexStatus`/`ListIndexes` report fully built indexes as building forever (EN-1748 already patched one symptom of this).

## What

- `common.proto`: `Index.build_status` and the `IndexBuildStatus` enum removed, remaining fields renumbered (pre-release, no reservation per project policy).
- `processCreateIndex`: the unreachable READY short-circuit and its registry probe are gone — the handler is a straight Put keyed off the command envelope, identical behavior for every reachable input.
- `processSetMetadataFieldType`: the retype keeps bumping `forward_encoding_version`; the BUILDING flip is gone.
- `indexes.IsReady` / `indexes.Status`: deleted (no callers).
- indexbuilder `loadIndexRegistry`: the status filter is gone (every row passed it); the per-replica `IndexVersionState` guard remains the sole scheduling gate.
- Comments across checker / query compile / controller / ledgerctl updated; readiness is documented solely as `IndexEntry.current_version > 0`.

## Notes for review

- The one behavioral surface: API responses no longer carry `buildStatus`. Clients already had to use `current_version` (the ledgerctl status column, `pkg/actions.WaitFor*IndexReady`, and the query gate all do).
- Wire/store compatibility is intentionally broken (field renumbering) — v3 pre-release wipe policy.
- The `Restore typed account metadata` e2e suite and friends are untouched; index e2e suites (`indexes`, `metadata_index`, `log_date_index`) pass. Two `tarutil` unit tests fail as root on the clean base too (permission-bit tests) — unrelated.

[...]

PR #1769, opened 2026-08-24
fix(backup): rebuild the index registry from CreateIndex/DropIndex logs

Azorlogh:
## The bug

`RebuildDelta` → `replay.ReplayLedgerLog` discarded `CreateIndex`/`DropIndex` logs (`// Index operations — no state to track`), so a restored store kept the metadata schema while losing **every index registry row**. The next `removeFieldType` on an indexed field found nothing to drop, `dropped_index` stayed empty, and the read-side index (which folds logs itself) survived a field that no longer existed.

This is the root cause of the Antithesis `singleton_driver_model: remove-field-type dropped-index mismatch` finding (~30–57/run): every composer restore cycle silently emptied the registry. Traced end-to-end on run `6991af2a…`: k0's CreateIndex = log seq 351, restore-cycle backup `lastLogSequence: 386` (the create **was** in the backup), restored cluster at term 1, rmFT at seq 429 → empty drop. Replaying the same committed stream sequentially on one node (no restore) behaves correctly — the op stream is clean, the restore rebuild was the divergence.

## The fix

`ReplayLedgerLog` routes index activity through the `replay.Writer`:

- `CreatedIndexLog` → put the row `processCreateIndex` writes (BUILDING, version 1, created_at = the log's date; overwrite on duplicate, same as live).
- `DroppedIndexLog` → delete the row.
- `RemovedMetadataFieldType` cascade → delete the row named by the log's `dropped_index` (carried on the log — no schema rederivation).
- Retype (`SetMetadataFieldType`) → bump `forward_encoding_version` on a covered index, mirroring the live cascade. The bump matters: the registry version and the readstore's refolded `IndexVersionState` must agree on which versioned keyspace is current.

Backup writer details: same-window reads resolve through a `pendingIndexes` map with nil deletion markers (the write session is non-indexed, cf. `GetVolume`); checkpoint rows come from the committed store, covering incremental restores where the create predates the delta. The checker's `replayStore` implements the new interface methods as documented no-ops — the checker folds `expectedIndexes` itself. `DeleteLedger` needs nothing here: the pending-cleanup marker the rebuild already records drives the covering purge that range-deletes the ledger's `SubAttrIndex` zone.

## Tests

- `tests/e2e/cluster/restore_index_registry_test.go` — mirrors `restore_metadata_type_test.go` (which pinned the same class of rebuild loss for typed metadata): create index → full backup → incremental → restore → the registry entry must survive and a `removeFieldType` must report it dropped. **Red on base, green with the fix** (verified both directions).
- `TestRebuildDelta_ReplaysIndexRegistry` — unit fold: create/retype/cascade-drop/drop + a surviving index with exact field assertions (date stamp, version).
- `TestRebuildDelta_BumpsCheckpointIndexOnRetype` — incremental path: the retype bump reads the checkpoint row from the committed store.

Note: `BuildStatus` handling here mirrors the current live writes; #1766 (EN-1858) deletes the field, and whichever lands second rebases trivially (compile error on the removed field).

### DIFF
--- a/internal/infra/backup/rebuild.go
+++ b/internal/infra/backup/rebuild.go
@@ -512,6 +512,7 @@ func RebuildDelta(
 			writer.batch = batch
 			clear(writer.pendingVolumes)
 			clear(writer.pendingTx)
+			clear(writer.pendingIndexes)
 
 			logger.WithFields(map[string]any{
 				"logsProcessed": count,
@@ -905,10 +906,12 @@ type attributeReplayWriter struct {
 	pendingVolumes map[string]*raftcmdpb.VolumePair
 	pendingTx      map[string]*commonpb.TransactionState
 
-	// Index registry rows touched by the replay, keyed by the canonical
-	// IndexKey bytes. w.batch is non-indexed, so same-window reads (the
-	// retype version bump) resolve here first; a nil value marks a row the
-	// replay deleted, shadowing a committed checkpoint row.
+	// Index registry rows touched by the replay window, keyed by the
+	// canonical IndexKey bytes. w.batch is non-indexed, so same-window reads
+	// (the retype version bump) resolve here first; a nil value marks a row
+	// the replay deleted, shadowing a committed checkpoint row. Cleared at
+	// every batch commit alongside pendingVolumes/pendingTx — the committed
+	// rows are then visible through the direct store read.
 	pendingIndexes map[string]*commonpb.Index
 
 	// LedgerInfo per ledger, carrying the evolving metadata schema and account
--- a/internal/infra/backup/rebuild.go
+++ b/internal/infra/backup/rebuild.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/formancehq/ledger/v3/internal/domain"
 	"github.com/formancehq/ledger/v3/internal/domain/accounttype"
+	"github.com/formancehq/ledger/v3/internal/domain/indexes"
 	"github.com/formancehq/ledger/v3/internal/domain/replay"
 	"github.com/formancehq/ledger/v3/internal/infra/attributes"
 	"github.com/formancehq/ledger/v3/internal/infra/state"
@@ -50,8 +51,10 @@ func RebuildDelta(
 		ledger:          attrs.Ledger,
 		references:      attrs.References,
 		boundary:        attrs.Boundary,
+		index:           attrs.Index,
 		pendingVolumes:  make(map[string]*raftcmdpb.VolumePair),
 		pendingTx:       make(map[string]*commonpb.TransactionState),
+		pendingIndexes:  make(map[string]*commonpb.Index),
 		ledgerInfos:     make(map[string]*commonpb.LedgerInfo),
 		boundaries:      make(map[string]*raftcmdpb.LedgerBoundaries),
 		reversions:      make(map[string]*bitset.Bitset),
@@ -186,7 +189,7 @@ func RebuildDelta(
 
 			ledgerName := p.Apply.GetLedgerName()
 
-			if err := replay.ReplayLedgerLog(ledgerName, seq, p.Apply.GetLog().GetData(), writer, rawLedgerTypes, ledgerAccountTypes, ephemeralPurgeBuffer); err != nil {
+			if err := replay.ReplayLedgerLog(ledgerName, seq, p.Apply.GetLog().GetData(), p.Apply.GetLog().GetDate(), writer, rawLedgerTypes, ledgerAccountTypes, ephemeralPurgeBuffer); err != nil {
 				_ = batch.Cancel()
 
 				return fmt.Errorf("replaying ledger log %d: %w", seq, err)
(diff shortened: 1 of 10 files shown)


internal/infra/backup/rebuild.go
```diff
@@ -722,7 +725,81 @@ func (w *attributeReplayWriter) SetMetadataFieldType(ledger string, target commo
...
+	row.ForwardEncodingVersion++
+
+	return w.putIndex(ledger, id, row)
+}
+
+// CreateIndex writes the registry row the live processCreateIndex writes: a
+// fresh BUILDING entry at forward-encoding version 1, stamped with the log's
+// apply date. A duplicate CreateIndex overwrites the row, matching the live
+// handler.
+func (w *attributeReplayWriter) CreateIndex(ledger string, id *commonpb.IndexID, createdAt *commonpb.Timestamp) error {
```
NumaryBot, line 757: 🔴 **[blocker] Preserve READY indexes when replaying duplicate creates**

When a checkpoint contains a READY index and the delta contains a duplicate CreateIndex, the live handler leaves the existing row untouched, but this replay replaces it with BUILDING, resets its encoding version to 1, and changes its creation date. This violates restore parity and can select the wrong versioned keyspace after restore; read the existing row and preserve it when READY, matching `processCreateIndex` and the invariant in `docs/technical/architecture/subsystems/chapters/incremental-restore-contract.md`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1769#issuecomment-5397481191

internal/infra/backup/rebuild.go
Azorlogh: A READY checkpoint row cannot exist: nothing has written READY since EN-1323 (#553) deleted applyIndexReady — the short-circuit in processCreateIndex is dead code, and #1766 (EN-1858) removes it plus the field. For every reachable state (duplicate create over a BUILDING row) live and replay both overwrite with a fresh v1 row, so live(prefix+delta) == restore(checkpoint,delta) holds.

gfyrag (commented):
Independent review at head 8472f82cd3a86bee1bd9fed90744eae289ed1e74. Blocking findings remain.

[P2][blocking] Complete the mandatory incremental-restore lifecycle proof
Location: tests/e2e/cluster/restore_index_registry_test.go (Phase 1 / Phase 3)
Evidence: the test checkpoints an empty store, places the complete registry lifecycle in the delta, compares only index-ID presence, and never runs CheckStore. It does not exercise checkpoint-seeded registry update/deletion or compare full logical rows as required by incremental-restore-contract.md.
Impact: regressions in committed-row fallback, retype version folding, deletion/tombstones, created_at, or residual rows can pass.
Resolution: seed meaningful index state before the checkpoint, mutate/create/delete after it, verify the export, compare complete source/restored registry rows, and require a clean CheckStore result.

[P2][non-blocking] Bound pendingIndexes memory
Location: internal/infra/backup/rebuild.go (batch commit and index overlay)
Evidence: pendingVolumes and pendingTx are cleared every 5,000 logs, but pendingIndexes retains every distinct touched index until rebuild completion.
Resolution: clear it after successful batch commits or provide another bounded visibility mechanism.

Targeted backup/replay/checker tests and diff check passed. Functional CI is green; patch coverage and the NumaryBot status are currently failing.

internal/infra/backup/rebuild.go
```diff
@@ -50,8 +51,10 @@ func RebuildDelta(
 		ledger:          attrs.Ledger,
 		references:      attrs.References,
 		boundary:        attrs.Boundary,
+		index:           attrs.Index,
 		pendingVolumes:  make(map[string]*raftcmdpb.VolumePair),
 		pendingTx:       make(map[string]*commonpb.TransactionState),
+		pendingIndexes:  make(map[string]*commonpb.Index),
```
NumaryBot, line 57: 🟠 **[major] Clear replayed indexes at batch boundaries**

When a rebuild processes many distinct indexes, `pendingIndexes` retains every row for the entire replay because the 5,000-log commit path clears only `pendingVolumes` and `pendingTx`. This defeats the batching mechanism's bounded-memory goal and can exhaust memory on large histories; clear this overlay after each successful commit, since the direct read handle can then see the committed rows.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1769#issuecomment-5397481191

tests/e2e/cluster/restore_index_registry_test.go
```diff
@@ -0,0 +1,331 @@
...
+			client, clusterClient, grpcConn, err = testutil.NewGRPCClient(ports.GRPC())
+			Expect(err).To(Succeed())
+
+			Eventually(func(g Gomega) bool {
+				state, err := clusterClient.GetClusterState(ctx, &clusterpb.GetClusterStateRequest{})
+				g.Expect(err).To(Succeed())
+				return state.Leader != 0
+			}).Within(10 * time.Second).ProbeEvery(100 * time.Millisecond).Should(BeTrue())
+
+			// Full checkpoint on the EMPTY store: everything after it lands in
```
NumaryBot: 🟠 **[major] Exercise a non-empty checkpoint in the restore regression**

This takes the full backup while the store is explicitly empty, so the end-to-end test only covers rebuilding an index created entirely in the delta and cannot catch failures when folding delta operations over checkpoint state. The restore contract requires meaningful state before the checkpoint ([incremental-restore-contract.md:66-69](docs/technical/architecture/subsystems/chapters/incremental-restore-contract.md#L66-L69)); seed an index-related state before this backup and verify the checkpoint-plus-delta composition through restore.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1769#issuecomment-5397481191

internal/infra/backup/rebuild.go
Azorlogh: Fixed in e126cf7a0 — pendingIndexes is cleared at every 5,000-log batch commit alongside pendingVolumes/pendingTx.

tests/e2e/cluster/restore_index_registry_test.go
Azorlogh: Fixed in e126cf7a0 — the e2e now seeds two rows before the full checkpoint, retypes one and cascade-deletes the other in the delta, creates a third delta-only row, compares complete restored rows field by field against the source (including created_at), and requires a clean CheckStore.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1769#issuecomment-5397481191

Azorlogh:
@gfyrag Both findings addressed in e126cf7a0 — the lifecycle now spans the checkpoint (seed → retype/cascade-delete/create in the delta → full row comparison + clean CheckStore), and pendingIndexes is cleared at every batch commit. CI is green (the Tests-E2E failures were the known GetLedger snapshot-consistency writer-deadline flake; full cluster suite passes locally).

internal/infra/backup/rebuild.go
```diff
@@ -50,8 +51,10 @@ func RebuildDelta(
 		ledger:          attrs.Ledger,
 		references:      attrs.References,
 		boundary:        attrs.Boundary,
+		index:           attrs.Index,
```
flemzord, line 54: [P2][blocking] Document the index-registry restore lifecycle

Evidence: this adds the index registry to the durable `RebuildDelta` fold, including create, retype/version bump, drop, and removal-cascade semantics. `AGENTS.md` requires architecture/behavior changes to update the owning documentation, and the incremental-restore contract explicitly requires new non-obvious restore rules to be documented. However, `docs/ops/backup-restore.md` still enumerates the projections rebuilt during restore without the index registry, and the index subsystem documentation does not describe this checkpoint-plus-delta fold.

Impact: the authoritative restore guidance remains materially incomplete about a persisted projection whose omission caused the original divergence, making the same class of regression easier to reintroduce.

Resolution: update the backup/restore and index-registry documentation to describe the replay evidence and the create/retype/drop/cascade lifecycle across the checkpoint boundary.

flemzord (changes requested):
DECISION: REQUEST CHANGES

Head reviewed: `e126cf7a0b41cec10e1c059f63d6d7ba370d45ee`
Blocking findings: 1
Non-blocking findings: 0
Previous findings: 2 fixed / 0 still valid / 1 outdated
Validation considered: all GitHub checks green; `go test -count=1 ./internal/domain/replay ./internal/infra/backup ./internal/application/check` passed in the pinned Nix environment; `bash scripts/agent-check` passed.
Residual risk: LOW once the authoritative restore documentation covers the newly implemented index-registry fold.

internal/infra/backup/rebuild.go
Azorlogh: Addressed in 2eefae2eb: indexes.md gains a "Restore Lifecycle" section describing the checkpoint-plus-delta fold — the per-log replay evidence for create / retype bump / drop / removal cascade, the deletion-marker overlay that keeps a same-window read from resurrecting a checkpoint row, and what is deliberately per-replica (IndexVersionState, read-store keyspaces) and rebuilt via backfill instead of restored. backup-restore.md's rebuilt-projections list now names the index registry and links there.

flemzord (commented):
The implementation review is clean, but the required patch-coverage check is still failing.

internal/infra/backup/rebuild.go
```diff
@@ -722,7 +726,79 @@ func (w *attributeReplayWriter) SetMetadataFieldType(ledger string, target commo
...
+	// covers this field, bump its forward_encoding_version so the restored
+	// registry row matches the one the live apply wrote.
+	id := indexes.MetadataID(target, key)
+
+	row, err := w.getIndex(ledger, id)
+	if err != nil {
+		return fmt.Errorf("reading index registry row for retype cascade: %w", err)
+	}
+
+	if row == nil {
```
flemzord, line 743: `codecov/patch` is currently failing at 79.245% against the repository's 80% target. Could we close the gate with focused tests around the new replay branches here—particularly the missing-row path, same-window tombstone shadowing, and persistence/read error paths? These cases exercise the restore-parity behavior introduced by this change rather than only increasing the percentage.

Azorlogh: Closed in 6ba582fb8 with the focused tests you named rather than filler: the ledger- and transaction-target arms of the replayed schema fold, the cascade's missing-row path (declaration with no covering index writes no registry row), same-window tombstone shadowing (a checkpoint row dropped earlier in the window must not be bumped back into existence by a later retype), and the registry read-failure path (a failing read fails the fold loudly). The new code in rebuild.go is now at ~93% — the only lines left are three batch-write error wraps, which the concrete dal.WriteSession offers no way to fail in a unit test.

gfyrag (approved):
Independent re-review at exact head `6ba582fb8695a6f167a502037dc0d8c81e9e80f4`.

CreateIndex/DropIndex and retype cascades are reconstructed during delta replay, same-window tombstones shadow checkpoint rows, read failures surface, and the restored registry remains aligned with schema/index lifecycle semantics. The new tests pin ledger/transaction target arms, missing rows, tombstones, and read failures.

The prior request for changes was against an older head and its missing authoritative restore documentation is fixed on this head. Current CI is green.

flemzord (approved):

PR #1771, opened 2026-08-25
fix(bloom): enforce Stop ownership boundary

gfyrag:
Fixes EN-1861.

DISCOVERY: NO_EXISTING_WORK
Searched open/closed PRs, branches, commits, and Jira references; no direct or partial Bloom lifecycle fix found.

BEFORE_FIX: BUG_REPRODUCED
Base SHA: 0213c42236ab5a7ba9cfb2b14a900138caa8e814
Evidence: after CacheSnapshotter.Stop returned, a late StartAsyncBloomPopulate could start new Bloom work.
Invariant: after terminal Stop returns, no new Bloom-owned background work may start.

AFTER_FIX: PASS
The regression rejects late starts; Pause/Resume preserves temporary quiescence and restart semantics, with targeted race repetitions passing.

The earlier swiss failure was caused by host Go 1.27; validation uses Go 1.26.5 Nix with GOROOT neutralized.

Fix: separate temporary Pause/Resume quiescence from terminal Stop ownership.

### DIFF
--- a/internal/infra/state/synchronizer.go
+++ b/internal/infra/state/synchronizer.go
@@ -44,12 +44,6 @@ func (s *Synchronizer) SynchronizeWithLeader(ctx context.Context, snapshotFetche
 	// Stop background tasks (bloom restore, etc.) that may hold Pebble iterators.
 	// RestoreCheckpoint closes and reopens the DB — outstanding references cause a panic.
 	s.apply.PauseBackgroundTasks()
-	paused := true
-	defer func() {
-		if paused {
-			s.apply.ResumeBackgroundTasks()
-		}
-	}()
 
 	// Drop any background-request messages enqueued by the FSM hot path
 	// pre-sync: their payloads reference chapter IDs / sequence ranges /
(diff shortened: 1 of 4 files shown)


internal/infra/state/synchronizer.go
```diff
@@ -59,11 +59,13 @@ func (s *Synchronizer) SynchronizeWithLeader(ctx context.Context, snapshotFetche
 	if err := s.restoreCheckpoint(ctx, snapshotFetcher, progress, s.apply.State.SnapshotIndex); err != nil {
 		return 0, fmt.Errorf("restoring checkpoint from leader: %w", err)
 	}
-
 	// Restore cache from Pebble (the checkpoint contains the leader's cache data)
 	if err := s.recovery.RestoreCacheFromStore(); err != nil {
 		return 0, fmt.Errorf("restoring cache after sync: %w", err)
 	}
+	// Pebble replacement and synchronous cache restoration are complete; only
```
NumaryBot, line 66: 🔴 **[blocker] Start deferred Bloom population after resuming**

When the installed checkpoint has no persisted Bloom blocks or has config drift, `RestoreCacheFromStore` calls `StartAsyncBloomPopulate` while the snapshotter is paused, so the call is silently discarded. Resuming here does not re-dispatch it, leaving the Bloom filter unpopulated (or stale from the previous store) after a successful follower sync; resume and explicitly trigger the deferred population after cache restoration.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1771#issuecomment-5442935136

PR #1772, opened 2026-08-25
fix: validate incremental restore keyspace

gfyrag:
Fixes EN-1865.

DISCOVERY: NO_EXISTING_WORK
Searched open/closed PRs, branches, commits, and Jira references; no direct or partial existing fix found.

BEFORE_FIX: BUG_REPRODUCED
Base SHA: 0213c42236ab5a7ba9cfb2b14a900138caa8e814
Evidence: a framed incremental export containing a foreign key was accepted by ApplyExports and reached Pebble.
Invariant: imported segments may write only keys belonging to their declared keyspace/type and sequence range.

AFTER_FIX: PASS
The same regression rejects the foreign key before Pebble Set; targeted restore tests and race repetitions pass.

The earlier swiss failure was caused by host Go 1.27; validation uses Go 1.26.5 Nix with GOROOT neutralized.

Fix: validate segment type, prefix, key shape, and inclusive sequence range at restore ingestion.

### DIFF
--- a/internal/infra/backup/restore.go
+++ b/internal/infra/backup/restore.go
@@ -15,36 +15,34 @@ import (
 	"github.com/formancehq/ledger/v3/internal/storage/dal"
 )
 
+func exportKeyShape(segmentType string) ([]byte, int, error) {
+	switch segmentType {
+	case "log":
+		// [ZoneCold][SubColdLog][log_seq BE 8]
+		return []byte{dal.ZoneCold, dal.SubColdLog}, 10, nil
+	case "audit":
+		// [ZoneCold][SubColdAudit][audit_seq BE 8]
+		return []byte{dal.ZoneCold, dal.SubColdAudit}, 10, nil
+	case "auditItem":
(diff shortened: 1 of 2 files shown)


internal/infra/backup/restore.go
```diff
@@ -66,6 +112,13 @@ func ApplyExports(
 				return fmt.Errorf("reading entry from segment %s: %w", seg.Key, err)
 			}
 
+			if err := validateExportKey(seg, key); err != nil {
```
NumaryBot, line 119: 🟠 **[major] Validate segment type before reading entries**

When a framed segment contains no entries, `validateExportKey` is never called, so an unsupported `seg.Type` is accepted and committed as an empty batch. Validate the segment metadata before opening or consuming the stream so malformed manifests cannot bypass the newly added type check.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1772#issuecomment-5440523169

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1772#issuecomment-5440523169

PR #1774, opened 2026-08-25
fix(admission): exempt leader-internal system proposals from signature enforcement

Azorlogh:
## Problem

`authorizeUnsignedBatch` rejects **every** unsigned request when `RequireSignatures` is enabled (only a narrow signing-key bootstrap exception). But all leader-internal proposals — chapter archiver, sealer, scheduler, and the query-checkpoint scheduler — travel through admission as **unsigned** batches under a `WithSystemActor` context:

```go
admissionHandler.Admit(internalauth.WithSystemActor(ctx, commands.ComponentChapterArchiver),
    servicepb.UnsignedApplyRequest("", ...))
```

So on a **signed cluster** these components can never propose: chapter rotation/sealing/archiving and query-checkpoint creation are permanently rejected. This is pre-existing on `release/v3.0`.

It was surfaced by NumaryBot on #1750 (the cluster-policy reconciler, same pattern), where the upcoming write-readiness gate escalates the reconciler case from advisory `NOT_SERVING` to a permanent write block — but the root gap is general, hence this standalone fix.

## Fix

Exempt system actors from the unsigned-batch check. This is safe:

- `WithSystemActor` is set **only** by server-internal code (`internal/bootstrap/module.go`), never derived from client input — a client request can't forge it.
- System proposals are already audited under a `system_component` caller source.
- `RequireSignatures` exists to authenticate **client** writes; the leader's own audited internal proposals are inside the trust boundary.

Adds `auth.IsSystemActor(ctx)` and threads the context through `resolveBatch` → `authorizeUnsignedBatch`.

## Tests

- `TestAuthorizeUnsignedBatch_ExemptsSystemActor`: under `RequireSignatures`, a client's unsigned business write is rejected, the same batch from a system actor is admitted.
- Existing `authorize_unsigned_batch_test.go` (bootstrap exception) updated for the new signature and still passing.

Validation: `go build ./...`, admission + auth suites green, gofmt clean.

### DIFF
--- a/internal/application/admission/admission.go
+++ b/internal/application/admission/admission.go
@@ -543,7 +543,7 @@ func (a *Admission) Admit(ctx context.Context, req *servicepb.ApplyRequest) (log
 
 	ctx, sigSpan := tracer.Start(ctx, "admission.verify_signatures")
 	resolveBatchStart := time.Now()
-	batch, err := a.resolveBatch(req)
+	batch, err := a.resolveBatch(ctx, req)
 	a.resolveBatchDurationHistogram.Record(ctx, time.Since(resolveBatchStart).Microseconds())
 
 	sigSpan.End()
@@ -941,7 +941,7 @@ type verifiedBatch struct {
 // signature is verified against the payload bytes and the trusted batch is
 // unmarshaled from them — the server never re-serializes. An unsigned batch is
 // admitted only when signatures are not required (or for signing bootstrap).
-func (a *Admission) resolveBatch(req *servicepb.ApplyRequest) (verifiedBatch, error) {
+func (a *Admission) resolveBatch(ctx context.Context, req *servicepb.ApplyRequest) (verifiedBatch, error) {
 	switch v := req.GetVariant().(type) {
 	case *servicepb.ApplyRequest_Signed:
 		sb := v.Signed
@@ -971,7 +971,7 @@ func (a *Admission) resolveBatch(req *servicepb.ApplyRequest) (verifiedBatch, er
 			return verifiedBatch{}, fmt.Errorf("%w: empty unsigned batch", signing.ErrMissingSignature)
 		}
 
-		if err := a.authorizeUnsignedBatch(batch.GetRequests()); err != nil {
+		if err := a.authorizeUnsignedBatch(ctx, batch.GetRequests()); err != nil {
 			return verifiedBatch{}, err
 		}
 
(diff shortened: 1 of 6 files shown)


internal/application/admission/admission.go
```diff
@@ -990,7 +990,16 @@ func (a *Admission) resolveBatch(req *servicepb.ApplyRequest) (verifiedBatch, er
...
 //   - all other signing-management requests require a signature once keys exist
 //   - regular requests check the requireSignatures flag
-func (a *Admission) authorizeUnsignedBatch(reqs []*servicepb.Request) error {
+func (a *Admission) authorizeUnsignedBatch(ctx context.Context, reqs []*servicepb.Request) error {
+	// Leader-internal proposals (chapter archiver, sealer, schedulers, the
+	// query-checkpoint scheduler) travel through admission unsigned under a system
+	// actor set only by server-internal code. RequireSignatures authenticates
+	// client writes; the leader's own audited proposals are inside the trust
+	// boundary and would otherwise be permanently rejected on a signed cluster.
+	if auth.IsSystemActor(ctx) {
```
flemzord, line 999: [P2][blocking] Document the system-actor signing exception

Evidence: this early return changes an authoritative authentication/signing boundary, but `docs/technical/architecture/subsystems/admission/signing.md` still states that mandatory signing rejects unsigned batches, and `pipeline.md` repeats the same policy without this exception. `AGENTS.md` requires the relevant architecture documentation to change when behavior or a non-obvious invariant changes.

Impact: operators and future maintainers receive materially false guidance about which unsigned writes a signed cluster accepts.

Resolution: update `signing.md` and `pipeline.md` to describe the system-actor exception, its server-internal origin, audit attribution, and trust-boundary assumptions.

flemzord (changes requested):
DECISION: REQUEST CHANGES

Head reviewed: `3c4a355d1c4dfc9dfcb8f91895cbb11e18e85e77`
Blocking findings: 1
Non-blocking findings: 0
Validation considered: all GitHub checks green; `go test -count=1 ./internal/application/admission ./internal/adapter/auth` passed in the pinned Nix environment; `bash scripts/agent-check` passed.
Residual risk: MEDIUM until the authoritative signing contract documents the new trust-boundary exception.

internal/application/admission/admission.go
Azorlogh: Done in 7b2d915b6. Documented the exception in `signing.md` (new *Leader-internal proposals* subsection — server-internal origin, never client-forgeable, `system_component` audit attribution, and the trust-boundary rationale) and updated `pipeline.md`'s signature-verification step to note the exemption and link to it. Intro caveat in signing.md updated too.

gfyrag (approved):

flemzord (approved):

PR #1775, opened 2026-08-25
feat(EN-1859): expose server-side request timing on QueryProfile

gfyrag:
Closes EN-1859.

## Problem

`QueryProfile` reported query **execution** time only — `index_duration_us + enrichment_duration_us` — so a caller could not tell a slow network from a slow server. Measured against a real 87.7M-transaction ledger over an established gRPC connection: **56 ms client-observed against 2.3 ms reported, ~54 ms unaccounted for**. Read-performance work could not be targeted, no server-side latency SLO could be set, and "why is this read slow for us" had no answer beyond "2.3 ms of query execution and 54 ms we cannot account for".

## What this adds

Seven new `QueryProfile` fields (proto 8–14), populated on all four profiled read endpoints of **both** transports:

| Field | Window |
|---|---|
| `server_duration_us` | consumer-independent server cost |
| `prepare_duration_us` | auth, decode, validation, filter compilation |
| `execute_duration_us` | executor call (index + enrichment are sub-phases) |
| `barrier_duration_us` | caller-requested read-consistency wait |
| `deliver_duration_us` | row serialisation + transport hand-off |
| `first_row_duration_us` | entry → first row accepted by the transport |
| `forwarded` | the breakdown is not local (see below) |

The profile clock now starts as the **first statement** of each handler. Previously `WithProfile` ran after auth and filter parsing, which is why `prepare` was structurally invisible — that positioning was the real cause of the unmeasured gap, not a missing field.

Barrier instrumentation sits at two choke points: `waitMinLogSequence` and `RoutedController.readCtrl`.

## Design decisions

**A breakdown, not a single total.** A bare total says "slow" without saying where, which is what makes optimisation untargetable. The phases are attributions inside a directly measured total, so the unattributed residual stays visible (`ledgerctl` renders it as "Other server work") rather than silently vanishing.

**Streaming ambiguity resolved by splitting the send loop.** `cur.Next()` is row *production* and is charged to execute; `stream.Send()` is *delivery* and is excluded from the server total. This matters concretely: on a follower routing to the leader, cursors are lazy, so each `Next()` is an upstream receive — real server cost that would otherwise disappear. Time-to-first-row and total-to-last-row are separate fields, so a slow consumer shows as `deliver >> first_row` with `server_duration_us` unchanged.

**Barrier excluded from the server total** because it is latency the caller opted into; charging it to the server blames the server for the caller's consistency requirement and makes the number track cluster RTT.

**Measurement is unconditional.** An earlier revision gated the per-row split on whether the caller requested the profile, which made *what is measured* depend on *whether someone asked to see it*. Those are different questions, and the gate broke the one consumer that matters — the slow-query log, which runs precisely in the ungated regime. Two `time.Now()` per row is tens of µs across a 1000-row page, against a proto marshal and a transport write per row. Unprofiled handlers still pay nothing (`profile == nil` skips).

**`--query-profile-threshold` now compares `ServerDuration + DeliverDuration`.** This is a behaviour change on an existing operator-facing flag, and it is deliberate: thresholding on execution alone meant the slow requests the log exists to surface were the ones it could not see. This trades in consumer back-pressure, accepted because the logged breakdown says which side was slow, and a threshold that fires with an explanation beats one that never fires. Flag help, the operator CRD and both generated CRDs are updated; `0 to disable` is now actually true.

## What was deliberately not done

**No `PreparedMsg` split to measure marshalling.** It would route the hot streaming path through a different, less-exercised codec and compression path and abandon the generated `Send` every other list handler uses — real behavioural risk for a measurement refinement. Serialisation therefore stays inside `deliver_duration_us`, and the **SLO claim is withdrawn** from the proto, `openapi.yml` and the docs. Both totals are biased on a streaming read: `server_duration_us` omits a cost that grows with page size, `server+deliver` admits client behaviour. The docs now say to pick one and state which, rather than pretending either is clean.

**No nested remote breakdown on forwarded reads.** It means opting every routed read into upstream profiling, plumbing an EOF-only trailer out through `cursor.Cursor`, and deciding how two breakdowns compose. Deferred and documented as deferred; the `forwarded` flag makes "no barrier" distinguishable from "barrier not measured here", which was the load-bearing part.

## Review history

Three rounds, each of which found something real.

**Round 1 (adversarial, pre-PR)** found a **critical** defect: the per-row execute/deliver split was gated on whether the caller requested the profile, so the entire send loop was charged to delivery and excluded from the total — a 2-second forwarded read reported a sub-millisecond `server_duration_us`, in the only configuration that consumes the profile. Its own test asserted the defect. Fixed by removing the gate, which also resolved a separate gRPC/HTTP incomparability (HTTP never had the gate), plus four majors and four minors.

**Round 2 (NumaryBot, on the PR)** found two majors:

- `forwarded=true` with a non-zero `barrier_duration_us` is reachable — `readCtrl` records the local `ReadIndex` attempt before deciding to forward — contradicting the documented "forwarded implies 0". Fixed by correcting the contract rather than zeroing the attempt: the caller really waited for that quorum round-trip, and dropping it would relocate consensus latency into the server total instead of deleting it.
- HTTP started its profile in the handler, so authentication was outside the measured window while gRPC counts it. Fixed at two depths, because the scope guard is chi middleware but JWT validation is router-wide: a `startRequestClock` stamp before `HTTPAuthMiddleware`, adopted by `withQueryProfile` on the four profiled routes via the new `query.WithProfileStartingAt`.

**Round 3 (adversarial, on the pushed branch)** found five more, all fixed in `958902b3`. The two that mattered: the `forwarded` field comment still carried the absolute round 2 had just removed from the barrier field forty lines above, so the wire contract contradicted itself; and a **fourth** forwarded/barrier combination exists that nothing documented — `waitMinLogSequence` charges its catch-up regardless of consistency level and runs before routing, so `--consistency leader --min-log-sequence N` yields a non-zero barrier on a perfectly healthy cluster. Documenting the non-zero case as "an attempt failed" invited reading leadership instability off an ordinary read-your-writes wait. It also caught the round-2 rationale naming the wrong phase (`readCtrl` runs inside the execute bracket, so an uncharged wait lands in `execute_duration_us`, not `prepare_duration_us` — same conclusion, wrong mechanism), a test of mine that passed only because `clampPhase` floored two assertions it never checked, and the fact that `prepare_duration_us` never included gRPC protobuf decode and cannot, since grpc-go unmarshals before dispatching.

## Validation

- `bash scripts/agent-check` — PASS, 0 lint issues across both modules, clean tree after generation
- `go test -race` on `internal/query`, `internal/adapter/grpc`, `internal/adapter/http`, `internal/bootstrap`, `cmd/ledgerctl/...` — PASS
- Full root-module unit suite — PASS
- CI on the branch: `Tests`, `Tests-E2E`, `Tests-Model`, `Tests-Scenarios`, `Tests-Schemathesis`, `Build`, `invariants`, `Dirty`, `Coverage`, codecov — all green

The e2e suite cannot run on this workstation (`ResourceExhausted: writes blocked` — the disk is at 97%, tripping the server's own write guard, reproduced identically on the clean base), so CI is the only e2e signal here. It is green.

## Note for reviewers

The `958902b3` fixes have not themselves been through an adversarial pass. The phase arithmetic, idempotent `Finish`, absence of data races and `Finish` placement were verified clean in round 3 and are unchanged in substance; what is new since then is the HTTP middleware split and the contract wording.

One deliberate non-fix, reported in round 3 and left alone: `clampPhase` records an `Anomaly` whose only emitter is gated on the slow-query threshold, and clamping `ServerDuration` to zero is what can pull the compared total below it — so a clamped profile may never be logged. The comment claiming it "cannot pass unnoticed" has been corrected to say the fuzzing assertion is the reliable detector. No reachable trigger was found; making the anomaly a wire field is a larger change than this PR should carry.

Unrelated pre-existing defect spotted and left alone: `openapi.yml` has a duplicate `'400'` response key around lines 585/591 (`promoteLedger`), which strict YAML parsers reject.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/adapter/http/handlers_list_accounts.go
+++ b/internal/adapter/http/handlers_list_accounts.go
@@ -9,6 +9,10 @@ import (
 
 // handleListAccounts handles GET /{ledgerName}/accounts to list accounts.
 func (s *Server) handleListAccounts(w http.ResponseWriter, r *http.Request) {
+	// See handleListTransactions: the profile clock starts before decode.
+	ctx, profile := query.WithProfile(r.Context(), wantsHTTPProfile(r))
+	r = r.WithContext(ctx)
+
 	ledgerName, ok := requireLedgerName(w, r)
 	if !ok {
 		return
@@ -33,9 +37,10 @@ func (s *Server) handleListAccounts(w http.ResponseWriter, r *http.Request) {
 
 	reverse := r.URL.Query().Get("reverse") == "true"
 
-	ctx, profile := query.WithProfile(r.Context())
-
+	profile.EnterExecute()
 	cursor, err := s.backend.ListAccounts(ctx, ledgerName, pageSize, afterAddress, filter, reverse)
+	profile.LeaveExecute()
+
 	if err != nil {
 		handleError(w, r, err)
 
@@ -47,9 +52,6 @@ func (s *Server) handleListAccounts(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if wantsHTTPProfile(r) {
-		writeProfileHeader(w, profile)
-	}
-
+	finishProfile(w, r, profile)
 	writeOK(w, accounts)
 }
--- a/internal/adapter/http/handlers_list_accounts.go
+++ b/internal/adapter/http/handlers_list_accounts.go
@@ -4,14 +4,13 @@ import (
 	"net/http"
 
 	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
-	"github.com/formancehq/ledger/v3/internal/query"
 )
 
 // handleListAccounts handles GET /{ledgerName}/accounts to list accounts.
 func (s *Server) handleListAccounts(w http.ResponseWriter, r *http.Request) {
-	// See handleListTransactions: the profile clock starts before decode.
-	ctx, profile := query.WithProfile(r.Context())
-	r = r.WithContext(ctx)
+	// See handleListTransactions: the profile clock started in the routing layer.
+	ctx := r.Context()
+	profile := profileFromRequest(r)
 
 	ledgerName, ok := requireLedgerName(w, r)
 	if !ok {
--- a/internal/adapter/http/handlers_list_accounts.go
+++ b/internal/adapter/http/handlers_list_accounts.go
@@ -10,7 +10,7 @@ import (
 // handleListAccounts handles GET /{ledgerName}/accounts to list accounts.
 func (s *Server) handleListAccounts(w http.ResponseWriter, r *http.Request) {
 	// See handleListTransactions: the profile clock starts before decode.
-	ctx, profile := query.WithProfile(r.Context(), wantsHTTPProfile(r))
+	ctx, profile := query.WithProfile(r.Context())
 	r = r.WithContext(ctx)
 
 	ledgerName, ok := requireLedgerName(w, r)
--- a/internal/bootstrap/controller_routed.go
+++ b/internal/bootstrap/controller_routed.go
@@ -88,7 +88,9 @@ func (b *RoutedController) readCtrl(ctx context.Context) (ctrl.Controller, *node
 	// The ReadIndex quorum round-trip plus the local WaitForApplied catch-up is
 	// consensus latency incurred to honour the caller's linearizable-read
 	// request, not query work. Charge it to the profile's barrier phase so it is
-	// visible but excluded from the server-cost total (EN-1859).
+	// visible but excluded from the server-cost total (EN-1859). Charged whether
+	// or not the barrier succeeds — a failed attempt is still time the caller
+	// waited (see the fallback branch below).
 	barrierStart := time.Now()
 	barrier, err := b.ReadIndexAndWait(ctx)
 	query.ProfileFromContext(ctx).AddBarrierWait(time.Since(barrierStart))
@@ -107,7 +109,12 @@ func (b *RoutedController) readCtrl(ctx context.Context) (ctrl.Controller, *node
 			span.SetAttributes(attribute.String("route", "leader_fallback"))
 
 			// Same as the explicit-leader branch: the read leaves this node, so
-			// the phase breakdown below describes the local hop only.
+			// the phase breakdown describes the local hop only. Unlike that
+			// branch, the barrier already recorded above is KEPT: the caller
+			// really did wait for a quorum attempt that then failed, and that
+			// wait belongs in the barrier phase (excluded from the server total)
+			// rather than in prepare. Hence forwarded=true with a non-zero
+			// barrier_duration_us is a valid, documented combination.
 			query.ProfileFromContext(ctx).MarkForwarded()
 
 			c, leaderErr := b.getLeaderCtrl()
--- a/internal/bootstrap/controller_routed.go
+++ b/internal/bootstrap/controller_routed.go
@@ -111,10 +111,12 @@ func (b *RoutedController) readCtrl(ctx context.Context) (ctrl.Controller, *node
 			// Same as the explicit-leader branch: the read leaves this node, so
 			// the phase breakdown describes the local hop only. Unlike that
 			// branch, the barrier already recorded above is KEPT: the caller
-			// really did wait for a quorum attempt that then failed, and that
-			// wait belongs in the barrier phase (excluded from the server total)
-			// rather than in prepare. Hence forwarded=true with a non-zero
-			// barrier_duration_us is a valid, documented combination.
+			// really did wait for a quorum attempt that then failed. Dropping it
+			// would not delete the time — readCtrl runs inside the caller's
+			// EnterExecute/LeaveExecute bracket, so an uncharged wait stays in
+			// execute_duration_us and from there in the server total. Hence
+			// forwarded=true with a non-zero barrier_duration_us is a valid,
+			// documented combination.
 			query.ProfileFromContext(ctx).MarkForwarded()
 
 			c, leaderErr := b.getLeaderCtrl()
--- a/internal/bootstrap/controller_routed.go
+++ b/internal/bootstrap/controller_routed.go
@@ -72,6 +72,14 @@ func (b *RoutedController) readCtrl(ctx context.Context) (ctrl.Controller, *node
 	case grpcadp.ConsistencyLeader:
 		span.SetAttributes(attribute.String("route", "leader"))
 
+		// Forwarded: the leader runs its own ReadIndex barrier (x-consistency is
+		// not propagated, so it defaults to linearizable there) and its own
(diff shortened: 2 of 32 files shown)


internal/bootstrap/controller_routed.go
```diff
@@ -71,12 +72,27 @@ func (b *RoutedController) readCtrl(ctx context.Context) (ctrl.Controller, *node
...
 		c, err := b.getLeaderCtrl()
 
 		return c, nil, err
 	}
 
+	// The ReadIndex quorum round-trip plus the local WaitForApplied catch-up is
+	// consensus latency incurred to honour the caller's linearizable-read
+	// request, not query work. Charge it to the profile's barrier phase so it is
+	// visible but excluded from the server-cost total (EN-1859).
+	barrierStart := time.Now()
```
NumaryBot, line 94: 🟠 **[major] Clear the failed local barrier when forwarding**

When `ReadIndexAndWait` returns `ErrNodeSyncing` or `ErrNotLeader`, this duration is recorded before the request falls back to the leader. The resulting profile has `forwarded=true` but a nonzero local `barrier_duration_us`, contradicting the new wire contract that forwarded reads report zero because their barrier is remote; it also excludes the failed local attempt from `server_duration_us`. Record the barrier only for the local path, or explicitly redefine the forwarded contract.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1775#issuecomment-5412610270

internal/bootstrap/controller_routed.go
gfyrag: Confirmed reachable, fixed in 227fb527 — but by correcting the contract, not by zeroing the attempt.

The finding is right that the code and the documented absolute disagree: `readCtrl` records the local `ReadIndex` attempt before it decides to forward, so `forwarded=true` with a non-zero barrier is reachable. It is worth splitting by cause, because the two are not equivalent in magnitude: `ErrNodeSyncing` is returned at the top of `ReadIndexAndWait` before any wait, so that attempt costs nanoseconds; `ErrNotLeader` can arrive via `failAllPendingReads` on leadership loss, which resolves a future the caller was already blocked on — that one can account for the whole quorum attempt.

I took the second option you offered rather than the first. Clearing the barrier on the fallback path would not delete that time, it would relocate it: `EnterExecute` computes `prepare = since(requestStart) - BarrierDuration`, so an uncharged quorum wait lands in `prepare_duration_us` and, through it, in `server_duration_us`. That labels a failed consensus round-trip as request preparation and puts caller-requested consensus latency back into the server total — the exact distortion the phase split exists to remove. A failed barrier is no more server work than a successful one, and the caller really did wait for it.

On the second half of the finding — "excludes the failed local attempt from `server_duration_us`" — that exclusion is deliberate, and it applies to `WallDuration()` too, since both totals subtract the barrier. Neither total is meant to answer "how long did the caller wait"; that is `barrier + server + deliver`, and every term is on the wire. What must not happen is the attempt vanishing from the breakdown, and it does not.

So `barrier_duration_us` is now specified as the local attempt, successful or not, with a three-case table keyed on the `forwarded`/`barrier` pair in `bucket.proto`, `openapi.yml`, `query-profile.md` and `profile.go`. Both `readCtrl` call sites carry the reasoning, and `TestQueryProfile_ForwardedKeepsFailedLocalBarrier` pins it so the combination cannot be "fixed" back into a zero.

internal/adapter/http/handlers_list_accounts.go
```diff
@@ -9,6 +9,10 @@ import (
 
 // handleListAccounts handles GET /{ledgerName}/accounts to list accounts.
 func (s *Server) handleListAccounts(w http.ResponseWriter, r *http.Request) {
+	// See handleListTransactions: the profile clock starts before decode.
+	ctx, profile := query.WithProfile(r.Context())
```
NumaryBot: 🟠 **[major] Start HTTP profiling before authentication**

For all four profiled HTTP routes, `RequireScope` runs before the handler, so creating the profile here excludes authentication from `prepare_duration_us`. This contradicts the documented phase contract that preparation includes auth (`docs/technical/architecture/subsystems/read-path/query-profile.md:38`) and makes HTTP preparation timings systematically incomplete when authentication is enabled. Start the profile in middleware before the scope guard, then propagate it into these handlers.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1775#issuecomment-5412610270

internal/adapter/http/handlers_list_accounts.go
gfyrag: Confirmed and fixed in 958902b3 — with a second half the finding did not reach.

`RequireScope` was the cheap half. The expensive half is `internalauth.HTTPAuthMiddleware`, applied router-wide via `r.Use`, which extracts the bearer token and runs `validateToken` — a remote JWKS fetch on a cache miss. That one cannot be moved under a per-route profile, so starting the profile above the scope guard alone would still have left the costly part outside the window. gRPC counts all of it, since `internalauth.Authenticate` does token extraction, validation, scope expansion and the scope check inside the handler, after `WithProfile`.

Fixed at the two depths separately:

- `startRequestClock`, immediately before `HTTPAuthMiddleware` in the router-wide chain, stamps an instant on every request — one `time.Now()` and one context value, no profile object. Deliberately *not* first in the chain: gRPC starts its profile after its own interceptors, so stamping at the very top would fold `RequestID`, `otelhttp` and the request logger into `server_duration_us` on HTTP only, trading one asymmetry for another.
- `withQueryProfile`, on the four profiled routes, builds the profile from that stamp through a new `query.WithProfileStartingAt`. It still sits outside the scope guard, which chi would otherwise run first — so the profiled routes now live in their own block with the scope groups nested inside. That is also the only shape `chi.Walk` can see; a route-level `r.With(guard).Get(…)` is invisible to it, which I found out by writing the test first and watching it fail to locate the guard.

The ordering is asserted structurally rather than behaviourally: the consequence of getting it wrong is a duration that is merely *smaller*, and no assertion on a duration can distinguish "auth was fast" from "auth was not measured". `TestProfiledReadRoutes_ClockPrecedesAuthPrecedesScopeGuard` walks the real router and pins clock < auth and profile < guard; `TestUnprofiledRoutes_HaveNoQueryProfile` pins the set closed from the other side, since a profile on a route that never reports one would leave barrier waits accumulating into a record nobody reads.

One related claim the finding did not mention but which was wrong in the same direction: `prepare_duration_us` was documented as including request decode. On gRPC it never did and never can — grpc-go unmarshals the request before dispatching to the handler. Six surfaces said otherwise; they now say HTTP query/body decode is inside and gRPC protobuf decode is in neither total.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1775#issuecomment-5412610270

PR #1776, opened 2026-08-26
feat(operator): generate standalone JSON Schema for CRDs

sylr:
## What changed

`just generate` in `misc/operator` now also emits standalone JSON Schema files under `config/crd/schemas/`: `v1alpha1_<kind>.json` (full CRD resource schema) and `v1alpha1_<kind>.spec.json` (spec-only) for Cluster, Ledger, Backup, BackupRun, and Credentials. A new `go run ./scripts/generate-json-schemas` step extracts the `openAPIV3Schema` from the generated CRD manifests and writes it as draft-04 JSON Schema. README and justfile updated to document and wire the new step.

## Why

Standalone JSON Schema files let IDEs (VS Code YAML, JetBrains) and external tooling validate operator CRD manifests (e.g. via `# yaml-language-server: $schema=...`) without needing the full `apiextensions.k8s.io` CRD wrapper.

N/A

## Technical decision

N/A

## Risk

**LOW** — additive, generator-only change scoped to `misc/operator`. Generated output is committed (following the existing pattern for `config/crd/bases/`), does not touch reconciliation logic, FSM, or persisted state.

## Validation

- [x] Targeted tests: `go test ./scripts/...` in `misc/operator` (new generator test), `just generate` re-run end-to-end to confirm output matches
- [ ] Full suite / broader validation: N/A

N/A — build-time/dev-tooling change only, no runtime behavior change.

## Review focus

Whether draft-04 is the right JSON Schema dialect for downstream consumers, and whether committing generated schemas (vs. generating on demand) is the right tradeoff.

## Known concerns

None


sylr:
Ran an independent Codex review (`omc ask codex`) before taking this out of draft.

**Verdict: REQUEST CHANGES** — 1 blocking, 2 non-blocking. All addressed in 6889ae35c:

- **[blocking]** README's full-manifest example pointed `$schema` at `v1alpha1_cluster.spec.json`, whose root has no `spec` property — `spec.replicas` etc. went unvalidated. Fixed: full-manifest example now uses the full `v1alpha1_cluster.json` schema; `.spec.json` usage is now scoped to bare spec fragments.
- **[non-blocking]** `crdOpenAPISchema` silently fell back to `Versions[0]` when no version was marked `storage`. Fixed: now fails loudly instead (no repo CRD currently hits this path, all are single-version).
- **[non-blocking]** Generated output directory was never pruned, so a removed/renamed CRD could leave stale committed schemas. Fixed: `run()` now clears existing `*.json` in the output dir before writing.

Added `TestRunFailsWhenNoStorageVersion` and `TestRunPrunesStaleSchemaFiles` for the two behavior changes. Re-ran `just generate` — regenerated schema output is byte-identical to before, `bash scripts/agent-check` passes.

PR #1777, opened 2026-08-26
fix: resolve S3 multipart InvalidPart, migrate to s3/transfermanager

sylr:
## Summary

Two related fixes to the S3 backup/cold-storage upload path, landed as separate commits:

1. **Fix the checksum mismatch (`3f642d97`).** `feature/s3/manager v1.16.15` predates the `RequestChecksumCalculation` option and never declares a checksum algorithm on `CreateMultipartUpload`, while the paired `service/s3 v1.98.0` client defaults to auto-attaching a CRC32 checksum on `UploadPart`. That mismatch surfaced in production as `CompleteMultipartUpload` `InvalidPart` for any backup/archive file crossing the 5 MiB multipart threshold — MinIO's laxer checksum validation masked it in CI. Bumping `feature/s3/manager` to v1.22.34 (with the coherent `aws-sdk-go-v2` core/config/credentials/`service/s3`/`sts`/`smithy-go` versions it requires) sets `ChecksumAlgorithm` on `CreateMultipartUpload` up front, so create/upload/complete agree again. `RequestChecksumCalculation` stays at its `WhenSupported` default rather than forcing `WhenRequired` — the version bump alone resolves the skew without disabling checksums SDK-wide.
2. **Migrate off the deprecated manager (`a968ae95`).** `feature/s3/manager.Uploader` is deprecated upstream in favor of `feature/s3/transfermanager` (`aws/aws-sdk-go-v2#3306`). This follow-through does that migration in both `internal/infra/backup/s3.go` and `internal/infra/coldstorage/s3.go` — `Uploader.Upload` becomes `Client.UploadObject`. `feature/s3/manager` is now gone entirely from `go.mod`. Only the upload path changes; reads (`GetObject`/`HeadObject`/`ListObjectsV2`) are untouched. Note `transfermanager` is pre-1.0 (v0.3.16) — pinned deliberately, worth re-checking its changelog before any further bump.

## Test plan

- [x] `bash scripts/agent-check` (proto/mock generation, `go mod tidy`, `golangci-lint`, full build, whitespace checks) — clean after both commits, no lint suppressions needed after the transfermanager migration
- [x] `go test -tags s3 ./internal/infra/backup/... ./internal/infra/coldstorage/...` against MinIO — full suite green after both commits
- [x] Manual multipart upload/download round-trip against real AWS S3 (not just MinIO) — done after the manager bump, and again after the transfermanager migration
- [x] `ginkgo run -tags "e2e s3"` — full `S3 Backup` and `Cold Storage S3` e2e specs (checkpoint manifest, incremental export/accumulation, export cleanup on new full backup, credential handling) against MinIO — green after both commits
- [x] Manual re-run of the same full+incremental backup and cold-storage archive flow against a real AWS S3 bucket (scoped under a disposable bucket-ID, verified fully cleaned up afterward via before/after object listing diff) — done after the manager bump


gfyrag (dismissed):
Reviewed at 3f642d973adc9769d9fe22c2ece6ea22247f6cb5. No blocking or non-blocking findings. The AWS SDK dependency set is coherent, and the updated uploader initializes CRC32 consistently across CreateMultipartUpload and UploadPart. Local agent-check, module verification, and targeted compilation passed.

internal/infra/backup/s3.go
```diff
@@ -19,15 +19,15 @@ import (
...
+	uploader *transfermanager.Client
 	bucket   string
 }
 
 // NewS3Storage creates a new S3Storage backed by the given S3 client and bucket.
 func NewS3Storage(client *s3.Client, bucket string) *S3Storage {
 	return &S3Storage{
 		client:   client,
-		uploader: manager.NewUploader(client),
+		uploader: transfermanager.New(client),
```
flemzord, line 30: [P2] transfermanager v0.3.16 defaults to a 16 MiB multipart threshold, while the backup and cold-storage multipart regression tests still upload 12 MiB. After this constructor change, those tests use PutObject and no longer exercise CreateMultipartUpload / UploadPart / CompleteMultipartUpload—the exact path that produced InvalidPart, including the io.Pipe case. Please force multipart in the tests or raise their payloads, and assert the multipart request sequence and checksum algorithm for both drivers.

gfyrag:
**[P2][blocking] Restore actual multipart regression coverage**

`feature/s3/transfermanager` defaults to a 16 MiB multipart threshold and 8 MiB parts, but the three Ledger multipart tests still upload only 12 MiB and assume the old manager's 5 MiB threshold:

- `internal/infra/backup/s3_test.go` — `TestS3Storage_PutFileMultipartLargeObject`
- `internal/infra/backup/s3_test.go` — `TestS3Storage_PutFileStreamsUnknownLengthReader`
- `internal/infra/coldstorage/s3_test.go` — `TestS3Storage_ArchiveMultipartLargeObject`

After the migration these tests use `PutObject`, not `CreateMultipartUpload` / `UploadPart` / `CompleteMultipartUpload`. They therefore stay green without exercising the exact `InvalidPart` regression this PR fixes, and their comments are now incorrect.

Please either preserve the previous multipart threshold explicitly when constructing the transfer manager, or increase the test payloads above the configured threshold and assert that the multipart operation sequence is actually reached.

internal/infra/backup/s3.go
sylr: Done — fixed in commit 1d2640fa1. Confirmed transfermanager's `defaultMultipartUploadThreshold` is 16 MiB (`api_client.go`), above the old manager default of 5 MiB, and traced `nextReader` — it reads up to that threshold and returns `io.EOF` early for anything smaller, routing it through `singleUpload`/plain `PutObject` instead of multipart. All three affected tests (backup's `TestS3Storage_PutFileMultipartLargeObject`, its io.Pipe/unknown-length sibling, and cold-storage's `TestS3Storage_ArchiveMultipartLargeObject`) were at 12 MiB, so they'd silently stopped exercising the multipart path after this migration. Bumped all three to 20 MiB and added a `recordingS3Client` wrapper (embeds `*s3.Client`, overrides `CreateMultipartUpload`/`UploadPart`/`CompleteMultipartUpload`/`PutObject`) so each test now asserts the actual call sequence (`CreateMultipartUpload` fires, `UploadPart` fires at least twice, `CompleteMultipartUpload` fires, `PutObject` never fires) and that `CreateMultipartUpload` declares `ChecksumAlgorithmCrc32` up front — the exact property this PR's checksum fix depends on. Ran the updated suite against MinIO to confirm all assertions pass.

gfyrag (approved):
Re-reviewed at 1d2640fa1148624cd9bc16050ff87db62bf15229. The previous multipart coverage concern is fixed: test payloads now exceed transfermanager's 16 MiB threshold and assert the multipart call sequence. No remaining findings.

PR #1779, opened 2026-08-26
fix(storage): date the account-metadata delete's history revision (EN-1872)

Azorlogh:
## What changed

`DeleteAccountMetadata` now stamps `updated_at` on the account row — the log date on the replay path, `transaction_date()` when none is given — so the metadata-history revision the trigger writes for a delete carries the delete's own instant. Deleting an absent key now leaves the row untouched entirely (no `updated_at` bump, no history revision), matching the guard the no-op set already has. New e2e test covering metadata reads and filters as of a pit between a set and its delete, and the no-op delete.

## Why

Deleting an account metadata key erased it retroactively: the history trigger dates each revision at `updated_at`, which the delete left untouched, so the delete's revision inherited the previous write's date. Any point-in-time read at or after the set then resolved to the delete's revision — on `/accounts`, account counts, aggregated balances and `/volumes`. `DeleteTransactionMetadata` already handles this correctly; the account variant was the omission. [EN-1872](https://formance-team.atlassian.net/browse/EN-1872)

## Risk

**LOW** — one store method gains a date parameter; its two callers pass the same values the transaction-metadata delete already passes; no schema change. Only the bugged behavior changes observably.

## Validation

- [x] Targeted tests: new e2e specs are red without the fix, green with it; a third spec pins that a no-op delete leaves `updatedAt` untouched
- [x] Baseline validation: `internal/storage/ledger`, `internal/controller/ledger` and the pit-metadata-filter e2e suite green

`accounts_metadata` revisions written by future deletes are dated correctly; revisions already written by past deletes remain misdated — this fixes the write path, not existing history. Replayed `DELETED_METADATA` logs (`importLog`) date the revision at the log's date, so an import reconstructs the same history.

## Review focus

The absent-key guard (`metadata -> ? is not null`): it must not change anything a client can observe. The controller ignores the row count, so the HTTP status stays 2xx; the only suppressed effects are the `updated_at` bump my change would otherwise have introduced and a duplicate history revision.

## Known concerns

None

[EN-1872]: https://formance-team.atlassian.net/browse/EN-1872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/controller/ledger/controller_default.go
+++ b/internal/controller/ledger/controller_default.go
@@ -348,7 +348,7 @@ func (ctrl *DefaultController) importLog(ctx context.Context, store Store, log l
 					}
 				case ledger.MetaTargetTypeAccount:
 					logging.FromContext(ctx).Debugf("Deleting metadata of account %s", payload.TargetID)
-					if err := store.DeleteAccountMetadata(ctx, payload.TargetID.(string), payload.Key); err != nil {
+					if err := store.DeleteAccountMetadata(ctx, payload.TargetID.(string), payload.Key, log.Date); err != nil {
 						return nil, fmt.Errorf("failed to delete account metadata: %w", err)
 					}
 				}
@@ -662,7 +662,7 @@ func (ctrl *DefaultController) DeleteTransactionMetadata(ctx context.Context, pa
 }
 
 func (ctrl *DefaultController) deleteAccountMetadata(ctx context.Context, store Store, schema *ledger.Schema, parameters Parameters[DeleteAccountMetadata]) (*ledger.DeletedMetadata, error) {
-	err := store.DeleteAccountMetadata(ctx, parameters.Input.Address, parameters.Input.Key)
+	err := store.DeleteAccountMetadata(ctx, parameters.Input.Address, parameters.Input.Key, time.Time{})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


gfyrag:
Two items are still missing before merge:

1. **Rebase onto current `main`.** The PR is currently marked `CONFLICTING` / `DIRTY` because `pkg/client/.speakeasy/gen.lock` was also updated by #1783. The dependency changes are already present on `main`; rebasing and retaining/regenerating the current lock metadata should remove that unrelated churn and resolve the conflict.

2. **Cover the replay timestamp path.** The new e2e specs exercise the public delete path, where `DeleteAccountMetadata` receives a zero time and uses `transaction_date()`. They do not exercise the non-zero `at` branch used by `importLog`, even though preserving the original delete date during replay is part of this PR guarantee. An export/import round-trip test that sets account metadata, deletes it, imports the logs, and queries a PIT between the set and delete would cover this. This also appears to be the main reason the Codecov patch/project checks currently fail (70% patch coverage).

The implementation itself looks correct, and the targeted controller/storage tests pass.

Azorlogh:
Both addressed: rebased onto main (the dependency churn is gone — #1783 already carries it), and added an export/import round-trip spec covering the replay path — including a pit at the delete's original instant, which discriminates `log.Date` from an import-time date.

gfyrag (approved):

PR #1781, opened 2026-08-26
fix: preserve state tracking in atomic bulks

gfyrag:
## What changed

Atomic bulks retain ledger state tracking and transactional event buffering while acquiring the ledger lock. Post-import sequence resets now handle dry runs and imports ending at ID 0, failed bulks keep the ledger importable, and generated ID collisions fail loudly instead of panicking or being ignored.

Nested transaction callbacks and state transitions are deferred to the root commit. Event buffering is represented by a dedicated per-transaction buffer: `BeginTX` creates a child buffer and `LockLedger` reuses the current one, without linking controller wrappers together. Multi-instance test services are isolated and stopped after every spec.

## Why

After importing v2 logs, an atomic bulk could bypass the initializing-to-in-use transition, collide with imported transaction/log IDs, panic, and return an internal error. A successful element followed by a failing element could also publish an event for data later rolled back.

## Risk

**MEDIUM** — the production changes are confined to controller transaction wrappers and storage error handling, but they affect the write path. Unit, integration, rollback-event, dry-run, import, nested-transaction, and real multi-instance concurrency tests cover the critical paths.

## Validation

- [x] Baseline validation: `GOROOT= go build ./...`
- [x] Targeted tests: controller event commit/rollback and nested transactions, bulker, transaction/log collisions
- [x] E2E: post-import dry run, ID-0-only import, atomic rollback events, and two consecutive three-instance specs; combined run also passes with `-race`
- [x] Full suite / broader validation: `nix develop --command bash -c "just pre-commit"`; generated output remains stable
- [x] GitNexus `detect_changes --scope compare --base-ref origin/main`: LOW, no unexpected process impact

No API or persistence-format change. Ledger advisory locks remain PostgreSQL-wide, so concurrent instances serialize the initializing-to-in-use transition and sequence reset. Dry runs use a transaction/savepoint to roll back state while retaining PostgreSQL sequence updates. Events and in-memory state created in nested transactions remain pending until the root commit.

## Review focus

Please focus on wrapper composition between `controllerFacade`, `ControllerWithEvents.BeginTX`, and `ControllerWithEvents.LockLedger`; dry-run behavior during import; nested savepoint rollback; and concurrent post-import writes.

## Known concerns

None.


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1781#issuecomment-5423368194

internal/controller/system/state_tracker.go
```diff
@@ -17,31 +18,63 @@ import (
...
+	c.state.mu.RLock()
+	l := c.state.ledger
+	c.state.mu.RUnlock()
 
 	if l.State == ledger.StateInUse {
 		return fn(c.Controller)
 	}
 
-	ctrl, tx, err := c.BeginTX(ctx, nil)
+	if c.tx != nil {
```
NumaryBot: 🟠 **[major] Preserve initializing state for atomic dry runs**

When an initializing ledger receives an atomic bulk containing only dry-run operations, this branch ignores `dryRun`, performs the state transition in the bulk transaction, and the bulker subsequently commits it. Unlike the non-atomic path, the ledger becomes `InUse` and its sequences are reset even though no real operation occurred, preventing further imports.

gfyrag: Fixed in 68b34c02c: dry-run operations now return before the initializing-to-in-use transition, including when the facade already wraps a caller-owned transaction. The focused import regression and build/pre-commit checks pass.

internal/controller/system/state_tracker.go
```diff
@@ -17,31 +18,69 @@ import (
...
 	if l.State == ledger.StateInUse {
 		return fn(c.Controller)
 	}
 
-	ctrl, tx, err := c.BeginTX(ctx, nil)
+	// A dry run must not participate in the initializing-to-in-use transition,
+	// including when the caller already opened a surrounding transaction.
+	if dryRun {
+		return fn(c.Controller)
+	}
```
NumaryBot: 🔴 **[blocker] Reset sequences temporarily for dry runs after import**

When an initializing ledger already contains imported IDs, directly running a dry-run operation uses the still-unreset database sequences, so a generated transaction or log ID can collide with imported rows and make the dry run fail. The previous transactional path could reset the sequences and then roll the state change back; dry runs still need that temporary setup without committing the ledger transition.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1781#issuecomment-5423368194

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1781#issuecomment-5423368194

Azorlogh (approved):

internal/controller/ledger/controller_with_events.go
```diff
@@ -202,15 +227,19 @@ func (c *ControllerWithEvents) Commit(ctx context.Context) error {
...
 
 	return nil
 }
 
 func (c *ControllerWithEvents) Rollback(ctx context.Context) error {
-	c.atCommit = nil
+	if c.eventBuffer != nil {
+		c.eventBuffer.rollback()
+		c.eventBuffer = nil
+	}
```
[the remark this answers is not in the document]
gfyrag: Not changing this behavior: every production caller terminates the operation after a rollback error; none commits the same transaction afterward. A failed rollback leaves the database outcome uncertain, so retaining callbacks and allowing a later commit to publish them would be less safe than discarding them when rollback is requested.

PR #1786, opened 2026-08-27
test: cover transaction state metadata persistence

flemzord:
## Summary

- assert Numscript-produced transaction metadata is present in the `TransactionState` passed to `Put`
- keep the existing CreatedTransaction log assertions, covering both persisted state and returned log
- leave production behavior unchanged

## Mutation evidence

Temporarily inverted the production gate from `len(finalMetadata) > 0` to `len(finalMetadata) == 0`.

The focused race test failed at the new persisted-state assertion with `expected: "payment"` and `actual: ""`, while the existing log-only coverage would still have observed metadata. Restoring the original gate made the test pass.

## Validation

- `go test -race ./internal/domain/processing -run '^TestProcessCreateTransaction_Numscript_SetTxMeta$' -count=1`
- `go test -race ./internal/domain/processing -count=1`
- `bash scripts/agent-check`


gfyrag (approved):
The new assertion distinguishes persisted TransactionState metadata from the already-checked returned log, and catches regressions in the persistence gate. No blocking findings.

PR #1787, opened 2026-08-27
test(processing): cover revert boundary equality

flemzord:
## Summary

- Add focused coverage for a revert request whose transaction ID equals LedgerBoundaries.NextTransactionId.
- Assert the unallocated ID returns ErrTransactionNotFound.
- Leave reverted-status and transaction-state access unconfigured so any read after the boundary guard fails the test.

## Mutation evidence

Temporarily changed the production guard from >= to >. The focused race test failed on an unexpected MockScope.GetReverted call for transaction ID 5, proving the equality case reaches state access under the mutation. Restored >= before commit; the final tree is test-only.

## Checks

- GOMAXPROCS=2 env -u GOROOT nix develop --command go test -race ./internal/domain/processing -run ^TestProcessRevertTransaction_ -count=1

### DIFF
--- a/internal/domain/processing/processor_revert_transaction_test.go
+++ b/internal/domain/processing/processor_revert_transaction_test.go
@@ -318,6 +318,48 @@ func TestProcessRevertTransaction_NotFound(t *testing.T) {
 	require.Equal(t, uint64(99), txNotFound.TransactionID)
 }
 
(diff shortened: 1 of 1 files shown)


gfyrag (approved):
The equality case is a distinct boundary of the >= guard, and the mock expectations prove the processor returns before reverted/state reads. No blocking findings.

PR #1788, opened 2026-08-27
test: cover partial posting volume guards

flemzord:
## Summary

- Add table-driven regression coverage for partially populated posting volumes.
- Cover source and destination when either `Input` or `Output` is missing.
- Assert `ErrBalanceNotPreloaded`; production guard behavior remains unchanged.

## Mutation evidence

Both guards were temporarily changed from `||` to `&&`. The new test failed in all four cases: the mutated paths either returned no error or panicked instead of returning `ErrBalanceNotPreloaded`. The `||` guards were restored before commit.

## Validation

- `go test -race ./internal/domain/processing -run "^TestApplyPosting_" -count=1`
- Final diff inspection: only `internal/domain/processing/processor_posting_test.go` changed.

### DIFF
--- a/internal/domain/processing/processor_posting_test.go
+++ b/internal/domain/processing/processor_posting_test.go
@@ -175,6 +175,87 @@ func TestApplyPosting_NotPreloaded(t *testing.T) {
 	require.Contains(t, err.Error(), "balance not preloaded")
 }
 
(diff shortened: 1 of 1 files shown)


gfyrag (approved):
The table covers source and destination partial volumes with each missing component, asserting ErrBalanceNotPreloaded rather than a panic or silent success. No blocking findings.

PR #1789, opened 2026-08-27
test(processing): assert transaction ID boundaries

flemzord:
## Summary

- Assert that a successful transaction creation persists `LedgerBoundaries.NextTransactionId` from 1 to 2.
- Assert that a successful transaction reversion persists `LedgerBoundaries.NextTransactionId` from 5 to 6.
- Keep the regression coverage focused on the existing happy-path `Put` hooks; production semantics are unchanged.

## Mutation evidence

- With `nextTransactionID + 1` temporarily changed to subtraction, `TestProcessCreateTransaction` failed with expected `2`, actual `0`.
- With `revertTxID + 1` temporarily changed to subtraction, `TestProcessRevertTransaction_Success` failed with expected `6`, actual `4`.
- Both production expressions were restored to addition before commit.

## Validation

- `nix develop -c go test -race ./internal/domain/processing -run '^(TestProcessCreateTransaction|TestProcessRevertTransaction_Success)$' -count=1`
- Final diff inspection: only the two intended `NextTransactionId` assertions are present.


gfyrag (approved):
The existing happy-path tests now pin the persisted NextTransactionId values for both create and revert, catching arithmetic or assignment regressions. No blocking findings.

PR #1790, opened 2026-08-27
test: cover posting validation guards

flemzord:
## Summary

- add focused table-driven coverage for `validatePostings`
- exercise one valid posting plus invalid source, destination, asset, and color cases
- keep production code unchanged because no independent defect was found

## Mutation evidence

After adding the test, each `err != nil` validation guard was inverted independently and the focused test was rerun. Every mutation was killed: the invalid source, destination, asset, and color cases failed for their respective guard inversion. All temporary mutations were then restored, and `processor_transaction.go` has no diff.

## Validation

- `go test -race ./internal/domain/processing -count=1`
- final diff inspection: one test file only; no generated or dashboard drift

### DIFF
--- a/internal/domain/processing/processor_transaction_test.go
+++ b/internal/domain/processing/processor_transaction_test.go
@@ -12,6 +12,81 @@ import (
 	"github.com/formancehq/ledger/v3/internal/proto/servicepb"
 )
 
+func TestValidatePostings(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name    string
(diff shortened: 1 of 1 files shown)


gfyrag (approved):
Les tests ajoutent une couverture utile des quatre branches d’erreur de validatePostings, qui n’étaient pas exercées auparavant. Aucun finding bloquant. Le cas multi-postings avec une erreur après le premier élément reste un renforcement optionnel.

PR #1791, opened 2026-08-27
test: cover Numscript reference resolution

flemzord:
## Motivation

The owning processing-package mutation run did not exercise the NumscriptReference resolution path in processCreateTransaction, leaving version selection and early failure contracts unprotected.

## What changed

- Cover exact-version content resolution with a successful referenced-script transaction.
- Cover latest and empty selectors resolving the greatest version through the coverage gate.
- Pin latest lookup failure, missing latest version, stale content coverage, content lookup failure, and nil content behavior.
- Keep production behavior unchanged. A temporary selector-guard perturbation made the empty-selector regression case fail and was restored before validation.

## Validation

- nix develop -c go test -race ./internal/domain/processing -run ^TestProcessCreateTransaction_NumscriptReference_ -count=1

### DIFF
--- a/internal/domain/processing/processor_transaction_numscript_reference_test.go
+++ b/internal/domain/processing/processor_transaction_numscript_reference_test.go
@@ -1,0 +1,201 @@
+package processing
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+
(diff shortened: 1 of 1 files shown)


gfyrag (approved):
The tests cover exact, latest, and empty selectors plus each early resolution failure, including stale coverage, with concrete posting and error assertions. No blocking findings.

PR #1792, opened 2026-08-27
test(processing): cover account type enforcement

flemzord:
## Summary

- add focused regression coverage for strict account-type rejection in `processCreateTransaction`
- add focused regression coverage for strict account-type rejection in `processRevertTransaction`
- keep production behavior unchanged

## Mutation evidence

I temporarily changed both production guards from `len(compiled) > 0` to `len(compiled) <= 0` and ran:

```console
nix develop -c go test -race ./internal/domain/processing -run 'TestProcess(Create|Revert)TransactionRejectsAccountOutsideConfiguredTypes' -count=1
```

Both new tests failed as expected because create and revert returned successful, non-nil `LogPayload` values instead of `ErrAccountNotMatchingType`. Restoring the original guards made both tests pass. The mutation was fully reverted before commit.

## Validation

- `GOMAXPROCS=2 nix develop -c go test -race ./internal/domain/processing -run 'TestProcess(Create|Revert)TransactionRejectsAccountOutsideConfiguredTypes' -count=1`
- `nix develop -c go test -race ./internal/domain/processing -count=1`
- final diff inspection: one test file only; no production or generated changes

### DIFF
--- a/internal/domain/processing/processor_account_type_enforcement_test.go
+++ b/internal/domain/processing/processor_account_type_enforcement_test.go
@@ -1,0 +1,165 @@
+package processing
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+
+	"github.com/formancehq/ledger/v3/internal/domain"
+	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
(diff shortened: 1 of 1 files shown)


gfyrag (approved):
The focused tests verify strict account-type enforcement is wired into both create and revert processing and assert the rejected address. No blocking findings.

PR #1793, opened 2026-08-27
test(query): cover log horizon filtering

flemzord:
## Summary

- add focused table-driven coverage for the LOGS arm of MainHorizonKeep
- cover malformed log entity bytes, generic index lookup failure, Pebble not-found, malformed index values, and the equal/above main-sequence boundary
- assert the exact ledger-log lookup key and closer release for every successful index Get
- keep production behavior unchanged; the final diff is test-only

## Mutation evidence

- changed the horizon comparison from <= to <: the equal-to-main-horizon case failed with expected true, actual false
- weakened the 8-byte index-value guard: the malformed-value case failed with an out-of-range panic in binary.BigEndian.Uint64
- restored both production mutations before validation

## Validation

- go test -race ./internal/query -run ^TestMainHorizonKeep_ -count=1

### DIFF
--- a/internal/query/aligned_snapshot_logs_test.go
+++ b/internal/query/aligned_snapshot_logs_test.go
@@ -1,0 +1,130 @@
+package query
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/cockroachdb/pebble/v2"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+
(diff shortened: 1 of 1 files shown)


gfyrag (approved):
The LOGS horizon branch is covered across malformed inputs, lookup errors, exact key construction, closer release, and the inclusive sequence boundary. No blocking findings.

PR #1794, opened 2026-08-27
fix(EN-1874): make Raft scale-down postcondition-driven

gfyrag:
## Summary

- carry the committed Raft ConfChange index through RemoveNode and wait for the matching FSM apply before confirming success
- expose structured gRPC statuses for an already-absent node and for a committed removal whose tombstone is still pending
- make operator scale-down verify cluster membership before removal and after any opaque removal error
- preserve the removed-member admission barrier until the durable tombstone is verified

Jira: https://formance-team.atlassian.net/browse/EN-1874

## Validation

- env -u GOROOT nix develop --command go test ./internal/application/membership ./internal/infra/node ./internal/adapter/grpc
- env -u GOROOT nix develop --command sh -c "cd misc/operator && go test ./internal/controller"
- bash scripts/agent-check (with an isolated golangci-lint cache)


internal/infra/node/node.go
```diff
@@ -2556,48 +2593,98 @@ func (node *Node) RemoveNode(ctx context.Context, nodeID uint64) error {
 	// TOCTOU race where a concurrent JoinAsLearner acquiring the mutex
 	// next would otherwise re-check the blacklist and miss the tombstone
 	// still queued behind the applier's async submit.
-	postApplyFn := func(ctx context.Context) error {
+	postApplyFn := func(ctx context.Context, committedIndex uint64) error {
 		if !capturedBlacklistableID {
```
NumaryBot: 🟠 **[major] Wait for apply even when the member has no instance ID**

When removing a phantom or bootstrap member without a blacklistable instance ID, this branch returns immediately after Raft commit, before the asynchronous FSM batch durably deletes its peer row. This contradicts the new postcondition that every removal waits for its committed index ([docs/ops/cluster-operations.md:438](../../docs/ops/cluster-operations.md#L438)) and can report success despite `WaitForApplied` subsequently failing or being canceled. Always wait for `committedIndex`; only tombstone verification and the admission barrier should be conditional on having an instance ID.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1794#issuecomment-5437489818

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1794#issuecomment-5437489818

internal/infra/node/node.go
```diff
@@ -2551,53 +2588,119 @@ func (node *Node) RemoveNode(ctx context.Context, nodeID uint64) error {
...
+		"removedNodeID": nodeID,
+		"raftIndex":     committedIndex,
+	}).Infof("RemoveNode committed; waiting for durable FSM application")
 
-	for {
-		removed, err := node.membership.IsRemoved(nodeID, instanceID)
-		if err != nil {
-			return fmt.Errorf("polling removed-member visibility for %d: %w", nodeID, err)
-		}
+	if err := node.fsm.WaitForApplied(ctx, committedIndex); err != nil {
```
NumaryBot, line 2766: 🟠 **[major] Clear canceled removal barriers after durable apply**

When the RPC context is canceled before apply, this returns while leaving `pendingRemovals` permanently populated; no background waiter later verifies the tombstone, and retrying removal returns `ErrNodeNotInCluster` without cleanup. Consequently, even after `cluster forget-removed` deliberately deletes the tombstone, rejoining the same `(nodeID, instanceID)` remains blocked until the leader restarts, contrary to the documented re-admission flow ([removed-member-registry.md:243-247](../../docs/technical/architecture/subsystems/consensus/removed-member-registry.md#L243-L247)). Arrange context-independent cleanup once `committedIndex` applies while retaining the barrier until then.

internal/infra/node/node.go
```diff
@@ -1351,7 +1378,10 @@ func (node *Node) finishReady(result readyResult, stop chan struct{}) error {
 		// of transition type.
 		for _, change := range cc.GetChanges() {
 			if f, ok := node.pendingConfChanges.LoadAndDelete(change.GetNodeId()); ok {
```
NumaryBot: 🔴 **[blocker] Correlate committed ConfChanges with the proposal**

When a ConfChange caller is canceled and its proposal commits later, a subsequent different operation for the same node ID can occupy this map entry and be resolved with the stale change's index. For example, a delayed promotion can resolve an identity-less removal, causing it to wait for the promotion batch and report successful removal while the member remains present. Key the future with a proposal-specific identity and verify the change type rather than matching only `nodeID`.

internal/infra/node/node.go
```diff
@@ -2551,53 +2588,167 @@ func (node *Node) RemoveNode(ctx context.Context, nodeID uint64) error {
...
+			node.logger.WithFields(map[string]any{
+				"error":         err,
+				"removedNodeID": nodeID,
+				"raftIndex":     pending.appliedIndex,
+			}).Infof("Stopped waiting to clear pending removal barrier")
+
+			return
+		}
+
+		if node.pendingRemovals.CompareAndDelete(nodeID, pending) {
```
NumaryBot: 🔴 **[blocker] Verify the tombstone before clearing the background barrier**

After a canceled removal, this path clears the admission barrier solely because the applied index advanced, without checking that `(nodeID, instanceID)` is actually tombstoned. If the index was mis-correlated or the expected tombstone is missing, the removed live member can rejoin; the foreground path correctly treats this as an invariant failure. Perform the same tombstone verification before deletion, consistent with [AGENTS.md:39](../../../AGENTS.md#L39).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1794#issuecomment-5437489818

internal/infra/node/node.go
```diff
@@ -2230,34 +2323,46 @@ func (node *Node) SetObserver(obs *Observer) {
...
 	if err != nil {
-		return false, err
+		return 0, false, err
 	}
 
 	timeoutCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
-	_, err = future.Wait(timeoutCtx)
+	committedIndex, err := pending.future.Wait(timeoutCtx)
```
NumaryBot, line 2365: 🔴 **[blocker] Retain removal protection when cancellation wins the commit race**

If the RPC context is canceled after the removal commits but before `future.Wait` observes its resolution, this branch returns the cancellation and deletes the pending future without ever calling `waitForRemovalApplied`. The committed removal therefore gets no `pendingRemovals` barrier or background cleanup, allowing the same identity to rejoin before its tombstone is durable. This contradicts the cancellation guarantee in [removed-member-registry.md:100](../../../docs/technical/architecture/subsystems/consensus/removed-member-registry.md#L100); committed removals need protection established independently of whether the caller wins this wait race.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1794#issuecomment-5437489818

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1794#issuecomment-5437489818

misc/operator/internal/controller/raft_scaledown.go
```diff
@@ -175,8 +185,36 @@ func raftScaleDown(ctx context.Context, cfg *rest.Config, clientset kubernetes.I
...
+		podExecWithTimeout,
+	)
+}
+
+func removeNodeWithExec(ctx context.Context, cfg *rest.Config, clientset kubernetes.Interface,
+	namespace, pod0, container, serverAddr, tlsMode string, nodeID int32, force bool,
+	exec ledgerctlExec,
 ) error {
 	logger := log.FromContext(ctx)
+	present, err := raftNodePresent(
```
NumaryBot, line 203: 🔴 **[blocker] Preserve force removal when no leader is available**

When `force` is true, this mandatory pre-check prevents removal if quorum loss leaves the cluster without a leader, because `cluster status` is leader-routed and fails before `remove-node --force` runs. This breaks the documented primary use case for force removal—recovering from quorum loss ([docs/ops/cli.md:2711-2715](../../docs/ops/cli.md#L2711-L2715)). Skip the leader-dependent pre-check for force mode or obtain membership from a local source.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1794#issuecomment-5437489818

misc/operator/internal/controller/raft_scaledown.go
gfyrag: Verified against the current HEAD: this pre-check does not require quorum. raftScaleDown first makes pod-0/node 1 the leader; Node.TransferLeader is an explicit no-op when node 1 is already the local leader. GetClusterState then takes the leader-local path and reads rawNode.Status directly, without ReadIndex or any quorum round-trip. If pod-0 is not leader, force removal is already invalid independently of this pre-check: both the gRPC handler and Node.ForceRemoveNode reject it. Skipping the membership check would therefore weaken the fail-closed/idempotent postcondition without enabling a valid quorum-loss recovery path. Resolving as a false positive.

PR #1804, opened 2026-08-28
fix: join reconciliation workers during shutdown

gfyrag:
Fixes EN-1863.

## Discovery

DISCOVERY: NO_EXISTING_WORK

`NO_EXISTING_WORK`: searches across open/closed PRs, local and remote branches,
GitHub commits, Jira references, and recent Archiver/Sealer shutdown work found
no direct or partial fix to reuse.

## BEFORE_FIX

BEFORE_FIX: BUG_REPRODUCED

- Base: `b327aeaf8a0232c70842ec573e93721f0604e720`
- Evidence: `BACKGROUND_RECONCILIATION_OUTLIVES_OWNER`
- Deterministic channel/barrier tests held the Archiver and Sealer
  reconciliation goroutines active while `Stop` ran.
- Both tests failed because `Stop` returned before the held reconciliation work
  was released.

Invariant: shutdown must not return and release owned resources while
Archiver/Sealer reconciliation work can still mutate or access them.

## Root cause

The lifecycle worker joined only the main request-drain goroutine. Each
component launched its reconciliation loop as an untracked nested goroutine,
so cancellation was signaled but shutdown did not wait for reconciliation to
finish.

## Fix

- Track and join each owned reconciliation goroutine before the main worker
  completes.
- Keep reconciliation on the component-scoped stop channel.
- Make component shutdown safe for concurrent/repeated `Stop` calls.
- Document the shutdown ownership contract for chapters.

## AFTER_FIX

AFTER_FIX: PASS

- Regression tests pass for Archiver and Sealer shutdown while reconciliation
  is blocked.
- Concurrent/double shutdown passes.
- Existing archive/seal retry-error paths remain cancellation-aware and run in
  the joined request drain.
- Targeted package tests pass.

## Test sensitivity

TEST_SENSITIVITY: PASS

`TEST_SENSITIVITY=PASS`: temporarily removing the two reconciliation joins made
both regression tests fail with `Stop returned while owned reconciliation work
was still active`. The mutation was restored and never committed.

## Validation

- Go 1.26.5 from the pinned Nix environment: PASS
- Environment gate and hermetic cache isolation: PASS
- Targeted shutdown tests under race, 50 repetitions: PASS
- `go test -race ./internal/infra/state`: PASS
- `bash scripts/agent-check-full`: PASS
- `go test -race ./... -timeout 20m` (via `agent-check-full`): PASS
- Trusted pre-commit and clean-worktree pre-push gate: PASS

### DIFF
--- a/docs/technical/architecture/subsystems/chapters/README.md
+++ b/docs/technical/architecture/subsystems/chapters/README.md
@@ -2,6 +2,10 @@
 
 Chapters partition a ledger's transaction history into discrete sealed segments. Once sealed, a chapter can be archived to cold storage (S3 or filesystem) — logs and audit entries get exported and purged from Pebble while attributes stay hot. Receipt-based reverts allow undoing archived transactions without rehydrating cold data.
 
+The background Sealer and Archiver own both their request drains and periodic
+reconciliation loops. Shutdown cancels and joins all of that work before the
+chapter runtime releases its stores.
+
 ## Documents
 
 | Document | Description |
--- a/docs/technical/architecture/subsystems/chapters/lifecycle.md
+++ b/docs/technical/architecture/subsystems/chapters/lifecycle.md
@@ -152,6 +152,11 @@ The seal checkpoint exists on disk but the Sealer never proposed the `SealChapte
 
 The Sealer uses exponential backoff (100ms → 10s max) to retry on transient errors. The checkpoint remains on disk until the hash is successfully computed, ensuring no data is lost.
 
+On shutdown, the Sealer cancels both its request drain and reconciliation loop,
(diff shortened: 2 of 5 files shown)


gfyrag:
Closing as SUPERSEDED_BY_EN_1945. EN-1945 is implemented by #1884, which deletes Archiver, Sealer, their archive/seal reconciliation workers, request channels, and lifecycle wiring. QueryCheckpointScheduler survives but uses one directly tracked worker and is not an equivalent nested reconciliation path. No EN-1863 runtime surface survives the removal, and carrying this fix would only create modify/delete conflicts and an orphaned regression test. If #1884 is abandoned or chapters are retained, EN-1863 should be reassessed before release.

PR #1805, opened 2026-08-28
fix: keep active ColdReader databases alive

gfyrag:
Fixes EN-1862.

## Discovery

DISCOVERY: NO_EXISTING_WORK

Exact searches across open/closed PRs, branches, commits, and Jira references found no existing EN-1862 work. PRs #1623, #1633, and #1673 use or extend `ColdReader` for archived queries and backup backfill, but do not change its ownership or eviction lifetime; they are RELATED_BUT_DIFFERENT.

## BEFORE_FIX

BEFORE_FIX: BUG_REPRODUCED

- Base SHA: `b327aeaf8a0232c70842ec573e93721f0604e720`
- Command: `env -u GOROOT nix develop --command go test ./internal/infra/coldstorage -run '^TestColdReaderCapacityEvictionKeepsAcquiredReaderAlive$' -count=1 -v`
- Deterministic interleaving: a goroutine acquires chapter 1, a channel barrier holds its operation, loading chapter 2 at `maxCached=1` evicts chapter 1, then the operation resumes.
- Observed failure: `panic: pebble: closed` in `pebble.(*DB).getInternal`.

Invariant: A ColdReader backing DB must not be closed while an active reader still owns a valid lease/use of it.

## Root cause

`GetReader` returned the cache-owned `*pebble.DB` directly. No refcount, lease, or release operation represented an active user, so LRU/TTL eviction and shutdown could close Pebble while a consumer was still reading.

## Fix

- Return an explicit closeable read-handle lease and count active readers per cached chapter.
- Mark active entries evicted and defer Pebble close/cache-directory removal until the last lease closes.
- Allow a concurrently reacquired immutable chapter to revive the still-live cached DB.
- Make `ColdReader.Close` idempotently refuse new leases, join the TTL sweeper, wait for active handles, and then release owned databases.
- Close leases on every production path. Backup backfill scopes one lease per chapter so close errors can be joined to export errors without retaining all chapter DBs for the whole backup.
- Document the lifetime contract in the chapters subsystem README and cold-storage documentation.

## AFTER_FIX

AFTER_FIX: PASS

- Capacity eviction with an active reader passes.
- Deterministic TTL eviction, multiple readers, last release, concurrent reacquire, repeated eviction, explicit/double shutdown, close ordering, and resource cleanup pass.
- Focused adversarial tests passed normally and for 20 repetitions under the race detector after the final rebase on `1d13a0e82`; an earlier candidate also passed 20 normal repetitions.

TEST_SENSITIVITY: PASS

Temporarily removing the active-reader guard in LRU eviction makes the regression test fail again with `panic: pebble: closed`. The mutation was restored and was never committed.

## Validation

- Go 1.26 environment gate: PASS (`go1.26.5`, Nix store)
- Bugfix evidence gate: PASS
- Targeted package tests: PASS
- Targeted race repetitions: PASS (`-count=20`)
- Hermetic `bash scripts/agent-check-full`: PASS
- Full `go test -race ./...`: PASS (included by `agent-check-full`)
- Trusted `agent-just pre-commit`: PASS
- Pre-commit fixpoint: PASS
- Worktree: clean
- Exact technical review: APPROVE (`e92f3b0e69b9c94332f91b8c0780d0bdd314526a`, findings 0, blockers 0, residual risk LOW)

An initial non-hermetic lint run saw cache entries from sibling worktrees; the required hermetic-cache rerun was clean. A first manually provisioned module cache under `build/` was also rejected by `go mod tidy` because it was inside the module; moving the writable validation caches outside the worktree resolved it without tooling changes.

### DIFF
--- a/docs/technical/architecture/subsystems/chapters/README.md
+++ b/docs/technical/architecture/subsystems/chapters/README.md
@@ -12,6 +12,10 @@ Chapters partition a ledger's transaction history into discrete sealed segments.
 | [backup.md](backup.md) | Full-database backup (Pebble checkpoint + incremental segments) to S3, Azure, or filesystem; restore for disaster recovery. |
 | [incremental-restore-contract.md](incremental-restore-contract.md) | Required parity between live apply and checkpoint-plus-delta restore for persisted state. |
 
+Cold-storage reads use explicit handles: capacity/TTL eviction and shutdown
+must retain an archived chapter's Pebble database until its last active handle
+closes. See [cold-storage.md](cold-storage.md#reading-from-cold-storage).
+
 ## Related
 
 - [Storage](../storage/) — the Pebble zones that hold logs / audit entries before archival.
--- a/docs/technical/architecture/subsystems/chapters/cold-storage.md
+++ b/docs/technical/architecture/subsystems/chapters/cold-storage.md
@@ -59,7 +59,13 @@ Using SST directly (rather than re-encoding into a portable format like Parquet 
 `internal/infra/coldstorage/reader.go` exposes the read side: open an archived chapter, iterate its log + audit ranges, and feed them back into queries that span archived data. Concretely:
 
 - **Audit-chain verification** by the [checker](../checker/checker.md) walks past archive boundaries by joining each chapter's sealed `last_audit_hash` to the start of the next live chain (see [audit-chain.md § Verification](../checker/audit-chain.md#verification)).
-- **Log lookups** by ID across the archived range go through `coldstorage.Reader.Open(chapterID)` and stream the SST.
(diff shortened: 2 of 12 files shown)


gfyrag:
Closing as SUPERSEDED_BY_EN_1945. EN-1945 is implemented by #1884, which deletes ColdReader and the entire cold-storage package, removes archived-read fallback and cold backup backfill, and makes query/checker/backup paths read permanent primary-store history. No equivalent cached Pebble reader or replacement ownership model survives. Carrying this stale lease/refcount fix would conflict across all of its touched paths and would require additional adaptation for newer cold-reader consumers that #1884 also deletes. If #1884 is abandoned or cold reads are retained, EN-1862 should be reassessed before release.

PR #1806, opened 2026-08-28
fix: route snapshot fetcher and gRPC auth logs through the app logger

sylr:
## Problem

Process output mixed two log formats:

```
time="2026-08-27T16:15:11Z" level=info msg="Downloading snapshot file" path=001424.sst size=33523537
2026-08-27T16:15:12.603Z    INFO    IndexTracker updated in finishReady    {"node-id": 2, ...}
```

Both call sites use the same `logging.Logger` interface, but go-libs' `logging.FromContext(ctx)` silently falls back to a fresh **logrus text logger on stderr** when the context carries no logger. The snapshot fetcher resolved its logger from the applier's sync context (never has one attached), and gRPC auth failures (`logAuthFailure`) from the raw request context.

## Fix

- `internal/application/ctrl/snapshot_fetcher.go`: provider/fetcher take an injected `logging.Logger` (tagged `cmp=snapshot-fetcher`) and use `logger.WithContext(ctx)` instead of `FromContext`.
- `internal/bootstrap/module.go`: pass the app logger into `GRPCSnapshotFetcherProvider`.
- `internal/adapter/grpc/logger_interceptor.go` (new) + `server.go`: unary/stream interceptors inject the logger into service gRPC request contexts, mirroring the existing HTTP `loggerMiddleware`. Fixes the same leak for gRPC auth-failure logs.
- Test fixtures set `logger: logging.Testing()`.

Remaining `FromContext` callers (HTTP handlers/middleware, e2e testutil) already run under a context with the logger attached.

## Validation

- `GOROOT= go build ./...`, `golangci-lint` on touched packages (0 issues), `git diff --check`
- `go test ./internal/application/ctrl/ ./internal/adapter/grpc/ ./internal/adapter/auth/ ./internal/bootstrap/` — pass
- `scripts/agent-check` lint stage reports pre-existing issues in generated `*.pb.go` files only; nothing in touched files.

### DIFF
--- a/internal/adapter/grpc/logger_interceptor.go
+++ b/internal/adapter/grpc/logger_interceptor.go
@@ -1,0 +1,40 @@
+package grpc
+
+import (
+	"context"
+
+	ggrpc "google.golang.org/grpc"
+
+	logging "github.com/formancehq/go-libs/v5/pkg/observe/log"
+)
+
+// loggerInterceptor injects the configured logger into every unary call
+// context so downstream code (auth failure logging, error sanitization) can
+// resolve it via logging.FromContext. Without this, FromContext degrades to
+// go-libs' bare logrus stderr fallback and emits lines in a different format
+// than the rest of the process. This mirrors the HTTP loggerMiddleware.
+func loggerInterceptor(logger logging.Logger) ggrpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req any, info *ggrpc.UnaryServerInfo, handler ggrpc.UnaryHandler) (any, error) {
+		return handler(logging.ContextWithLogger(ctx, logger), req)
+	}
+}
+
+// loggerStreamInterceptor is the streaming counterpart of loggerInterceptor.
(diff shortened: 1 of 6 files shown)


sylr:
Addressed the P2 test-coverage finding in 9e2d29540: added `TestGRPCSnapshotFetcher_LogsThroughInjectedLogger` and `TestNewServiceServer_InjectsLoggerIntoRequestContexts`. Both drive the real code path with a buffered app logger (the latter through the actual `NewServiceServer` interceptor chain, unary + stream) and were mutation-checked: each fails when its fix is reverted.

gfyrag (approved):
Reviewed head 4f447afcb9b0bf1831651734807f6642fda8488a.

The existing HTTP middleware already injects the logger, while the pre-existing gRPC logging interceptor only emitted request logs and did not enrich request contexts. The new unary/stream interceptors correctly cover gRPC auth logging, and explicit injection remains necessary for the background snapshot fetcher. Focused gRPC and snapshot-fetcher tests pass. No blocking findings.

PR #1809, opened 2026-08-28
fix(grpc): stop fabricating clean pages from canceled forwarded streams

Azorlogh:
## What changed

The routed read cursors (`GRPCStreamCursor`, `upstreamPeekCursor`) no longer map every `Canceled` stream error to `io.EOF`. `normalizeStreamEnd` maps it only when the caller-supplied context is itself done (the consumer stopped reading and tore the stream down); any other cancellation propagates as an error.

## Why

On a forwarded read — a syncing follower falling back to the leader — a mid-flight stream cancellation (connection churn during partitions, pool teardown on leadership change) read as clean exhaustion, so the follower served the empty or truncated page with **OK status**. A linearizable `ListAccounts` could return zero rows for long-committed data, and a leader-side error reply (e.g. `index not found`) could surface as an empty page. Found by the Antithesis model workload as `* query outside model` findings, all in sync/heal windows with `route=leader_fallback`.

Need: linearizable reads must never return fabricated results, under any fault schedule.
Current limitation: the forwarded-read path converted transfer failures into clean pages.
Requirement / constraint: a stream that dies under a live caller surfaces an error; only consumer-side teardown reads as a clean end.
Evidence / durable repository evidence: `internal/adapter/grpc/cursor_stream_cancel_test.go` pins the contract against a real in-process stream.

## Technical decision

Decision: gate the `Canceled → io.EOF` mapping on the caller-supplied context, threaded into both cursors at construction.
Why now / why proportionate: minimal mechanism that keeps the one legitimate mapping (consumer teardown) and removes the fabrication.
Alternatives considered: dropping the mapping entirely (breaks the legitimate consumer-teardown end); gating on `ClientStream.Context()` (ineffective — gRPC cancels the stream-derived context on every termination, verified in grpc-go `stream.go` and pinned by the test).

## Risk

**LOW** — error propagation replaces silently wrong data; clients already treat `Canceled` as transient and retry.

## Validation

- [x] Targeted tests: `TestUpstreamPeekCursorStreamDeath` (bufconn, real stream: red on the old mapping and on a stream-context gate, green now), `TestNormalizeStreamEnd`, existing `internal/adapter/grpc` + `internal/bootstrap` suites
- [ ] Antithesis model-only verification runs on the instrumented branch are still in progress — draft until they complete

Forwarded list RPCs that lose their upstream stream now fail with `Canceled` instead of returning an empty/truncated OK page. No persisted state, FSM, or storage impact.

## Review focus

The `normalizeStreamEnd` discriminator, and the test's pinned assertion that the stream-derived context is already canceled at failure time — that is the reason the caller context is the only valid signal.

## Known concerns

The forwarded list requests also drop `ReadOptions.MinLogSequence` (`BucketGrpcClient` builds `ListOptions` without `Read`). Separate latent issue, currently masked by the leader's read barrier; not addressed here.

### DIFF
--- a/internal/adapter/grpc/cursor.go
+++ b/internal/adapter/grpc/cursor.go
@@ -1,6 +1,7 @@
 package grpc
 
 import (
+	"context"
 	"errors"
 	"io"
 
@@ -14,6 +15,7 @@ import (
 
 // GRPCStreamCursor implements cursor.Cursor[To] by reading from a gRPC server stream.
 type GRPCStreamCursor[Res, To any] struct {
+	ctx    context.Context
 	client grpc.ServerStreamingClient[Res]
 	mapper func(*Res) (To, error)
 }
@@ -21,9 +23,7 @@ type GRPCStreamCursor[Res, To any] struct {
 func (c GRPCStreamCursor[Res, To]) Next() (To, error) {
 	next, err := c.client.Recv()
 	if err != nil {
-		if status.Code(err) == codes.Canceled {
-			err = io.EOF
-		}
+		err = normalizeStreamEnd(c.ctx, err)
 
 		var zero To
 
@@ -33,6 +33,27 @@ func (c GRPCStreamCursor[Res, To]) Next() (To, error) {
 	return c.mapper(next)
 }
 
+// normalizeStreamEnd maps a Canceled Recv error to io.EOF ONLY when the
+// cancellation is the consumer's own — the caller-supplied context is done
+// because the consumer stopped reading and tore the stream down, so the
+// results gathered so far form a complete answer for it. A Canceled status
+// while that context is still live is a failed transfer (connection churn,
+// remote teardown): surfacing it as EOF would fabricate a clean-looking
+// empty or truncated page that the caller then serves with OK status.
+//
+// The verdict MUST come from the caller-supplied context, never from
+// ClientStream.Context(): gRPC derives the stream context with WithCancel
+// and cancels it in finish() on EVERY stream termination, so by the time
+// Recv returns a terminal error that context is always done and gates
+// nothing.
+func normalizeStreamEnd(ctx context.Context, err error) error {
+	if status.Code(err) == codes.Canceled && ctx.Err() != nil {
+		return io.EOF
+	}
+
+	return err
+}
+
 func (c GRPCStreamCursor[Res, To]) Close() error {
 	return c.client.CloseSend()
 }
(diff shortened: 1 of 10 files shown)


internal/adapter/grpc/cursor.go
```diff
@@ -33,20 +33,41 @@ func (c GRPCStreamCursor[Res, To]) Next() (To, error) {
...
+// ClientStream.Context(): gRPC derives the stream context with WithCancel
+// and cancels it in finish() on EVERY stream termination, so by the time
+// Recv returns a terminal error that context is always done and gates
+// nothing.
+func normalizeStreamEnd(ctx context.Context, err error) error {
+	if status.Code(err) == codes.Canceled && ctx.Err() != nil {
+		return io.EOF
+	}
+
+	return err
```
flemzord: At this point `ctx.Err() == nil`, so this is a transport failure, not a caller cancellation. Returning the raw `Canceled` status means repository clients do not retry it (they classify only `Unavailable` as transient), and routed HTTP list calls turn it into a generic 500 without `Retry-After`. Please map this branch to an unavailable-semantic error that becomes gRPC `Unavailable` and HTTP 503 + `Retry-After`, then add boundary-level tests for both transports.

flemzord (commented):
One public error-contract issue found during the latest review pass.

gfyrag (dismissed):
Independent review at exact head `db049ed62f5f5ab387aa41f5623277d06cf735a0`.

Forwarded stream cancellation now uses the caller context: cancellation becomes EOF only when that caller actually ended the request, while an upstream mid-stream cancellation remains visible as an error. The bufconn regressions exercise both branches. Functional CI is green; the remaining patch-coverage status is non-functional.

internal/adapter/grpc/cursor.go
Azorlogh: Done in 4911d227e. normalizeStreamEnd now re-codes the Canceled-under-live-context branch as Unavailable (original message preserved), and the HTTP error handler maps status-carrying Unavailable to 503 + Retry-After — the same contract as the KindUnavailable Describables. Boundary tests on both transports: the bufconn stream-death test asserts Unavailable end-to-end through a real gRPC stream, and the HTTP handler table asserts 503/Retry-After for Unavailable while a bare Canceled stays a sanitized 500. Side benefit: the interceptor's log demotion and the model driver's isShutdownError no longer misfile a mid-transfer failure as a clean shutdown.

gfyrag (changes requested):
Independent re-review at exact head `4911d227e4288466fa79a41964d32e89dd993b4b`.

[P2][blocking] Synchronize the widened HTTP retry contract
Location: `internal/adapter/http/error_handler.go:102-136`; `docs/technical/architecture/subsystems/api/http-api.md:81-90,112-155`; `docs/technical/contributing/api-comparison.md`; `openapi.yml:3054-3065`
Evidence: this head now adds `Retry-After: 1` to every domain `KindUnavailable` response and translates any gRPC `codes.Unavailable` into HTTP `503 UNAVAILABLE`. The authoritative HTTP documentation and shared OpenAPI response still describe 503/Retry-After as the no-leader-only `NO_LEADER` contract, and the API comparison does not record the new forwarded-stream error shape. The repository instructions require these public HTTP behavior changes to be synchronized.
Impact: clients and operators are told an incorrect retry/error-code contract for forwarded stream failures and the broader unavailable class.
Resolution: update the HTTP API docs, API comparison and OpenAPI response to describe the actual 503/Retry-After/error-code behavior, or narrow the implementation back to the documented contract.

The stream-side recoding itself is correct and targeted gRPC/HTTP tests pass. This is a genuinely new finding introduced after the previous approved head; there were no previous findings to carry forward.

DECISION: REQUEST CHANGES
Head reviewed: 4911d227e4288466fa79a41964d32e89dd993b4b
Blocking findings: 1
Non-blocking findings: 0
Validation considered: targeted gRPC + HTTP tests + current CI
Residual risk: LOW once documentation is synchronized

Azorlogh:
Addressed the P2 in 48e31644c: http-api.md's status table and Retry-After section now enumerate the four 503 families (NO_LEADER, CACHE_HORIZON_EXCEEDED, the KindUnavailable reasons, bare gRPC UNAVAILABLE), api-comparison.md gains the bare-status translation table with the forwarded-stream shape on both surfaces, and the shared OpenAPI ServiceUnavailable response describes the widened contract. Implementation untouched.

gfyrag (approved):
Independent re-review at exact head 48e31644c9042f7f235e6fe40304821563fe78f5.\n\nThe previous documentation blocker is fixed. The HTTP API documentation, API comparison, and shared OpenAPI 503 response now describe the widened Unavailable plus Retry-After contract and the forwarded-stream UNAVAILABLE shape. The implementation and public contract are synchronized. Targeted gRPC and HTTP tests pass, and current CI is green.\n\nDECISION: APPROVE\nBlocking findings: 0\nNon-blocking findings: 0\nPrevious findings: 1 fixed / 0 still valid / 0 outdated\nResidual risk: LOW

PR #1817, opened 2026-08-28
fix: server correctness fixes surfaced by the model-checking driver

Azorlogh:
## What changed

The server-side correctness fixes the model-checking driver (EN-1625, #1659) surfaced, peeled out of the test-coverage branch so that PR stays test-only. Each fix ships with the test that pins it (unit + the four e2e index-registry specs).

## Why

The model workload found these as `outside model` classes; they were fixed on the coverage branch during the hunts and now need to land independently. #1659's model runs are red without them.

Need: reads and index lifecycle must match a serializable state under churn.
Current limitation: retypes could serve mixed semantics, duplicate CreateIndex orphaned keyspaces, degenerate bounds mis-resolved, the abya scan admitted future members, a vanished WAL snapshot dir killed the node.
Requirement / constraint: every fix is pinned by a committed test.
Evidence / durable repository evidence: the tests in this PR; the model driver in #1659 exercises them continuously.

## Technical decision

Decision: land the accumulated fixes as one reviewable unit, tests included.
Why now / why proportionate: they block #1659 (test-only) from validating green.
Alternatives considered: one PR per fix (more churn, the fixes interlock through the index-version machinery); keeping them inside #1659 (mixes product and test changes).

## Risk

**MEDIUM** — touches the index-version machinery (EN-1724), query compile bounds/stamps, checker, and WAL snapshot recovery. Mitigated by the pinned tests and continuous model-run coverage.

## Validation

- [x] Targeted tests: indexbuilder, query compile (bounds/asset/retype), checker, WAL, readstore version-activation; e2e specs included
- [x] Continuous: the model workload (#1659) exercises all of these; a 3h model-only Antithesis run from the combined branch is in flight

Index retypes bind declared types to versions (per-replica `IndexVersionState`); degenerate query ranges resolve to the empty match; the account-by-asset index rows carry first-touch stamps; WAL snapshot saves recreate a missing directory.

## Review focus

The EN-1724 version-binding semantics (`index_config.go`, `backfill.go`) and the abya stamp-gating in `compile.go`.

## Known concerns

Ordering: #1659 (test-only) validates green only with these fixes present; merge this first or together.

### DIFF
--- a/internal/application/indexbuilder/index_config.go
+++ b/internal/application/indexbuilder/index_config.go
@@ -408,6 +408,30 @@ func (b *Builder) getOrCreateLedgerConfig(ledger string) *ledgerIndexConfig {
 // backfill scheduling so the builder does not redo work that has already
 // completed — and, more importantly, does not knock a live index back into
 // ErrIndexBuilding.
+// boundTypeAtCreation resolves the declared type an index's first version is
+// bound to: the schema entry in force when the CreateIndex log folds. Only
+// metadata indexes carry one; for every other kind — and for a key declared
+// after the index — the version is bound to no type and rows keep each
+// value's natural encoding.
+func (b *Builder) boundTypeAtCreation(ledgerName string, id *commonpb.IndexID) (commonpb.MetadataType, bool) {
+	meta, ok := id.GetKind().(*commonpb.IndexID_Metadata)
+	if !ok || meta.Metadata == nil || b.batchSchema == nil {
+		return 0, false
+	}
+
+	schema, err := b.batchSchema.For(ledgerName)
+	if err != nil || schema == nil {
+		return 0, false
+	}
+
+	_, fs := commonpb.SchemaFieldForTarget(schema, meta.Metadata.GetTarget(), meta.Metadata.GetKey())
+	if fs == nil {
+		return 0, false
+	}
+
+	return fs.GetType(), true
+}
+
 func (b *Builder) handleCreatedIndexLog(ledgerName string, log *commonpb.CreatedIndexLog) {
 	id := log.GetId()
 	if id == nil {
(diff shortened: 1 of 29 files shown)


internal/application/indexbuilder/index_config.go
```diff
@@ -408,6 +408,30 @@ func (b *Builder) getOrCreateLedgerConfig(ledger string) *ledgerIndexConfig {
...
+// metadata indexes carry one; for every other kind — and for a key declared
+// after the index — the version is bound to no type and rows keep each
+// value's natural encoding.
+func (b *Builder) boundTypeAtCreation(ledgerName string, id *commonpb.IndexID) (commonpb.MetadataType, bool) {
+	meta, ok := id.GetKind().(*commonpb.IndexID_Metadata)
+	if !ok || meta.Metadata == nil || b.batchSchema == nil {
+		return 0, false
+	}
+
+	schema, err := b.batchSchema.For(ledgerName)
```
NumaryBot, line 422: 🔴 **[blocker] Propagate schema lookup failures**

When the schema read fails, this treats the index as undeclared and persists that binding, causing subsequent rows to use natural encodings and queries to reject the index instead of retrying the batch. Return the error through `handleCreatedIndexLog` so no incorrect version state is committed; this also violates [AGENTS.md:64](/data/worktrees/formancehq__ledger-pr-1817/AGENTS.md#L64).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1817#issuecomment-5451988551

Azorlogh:
Closing: this bundled unrelated server fixes without prior review. Each bug is being re-prepared on its own branch with a red e2e test and will be PR'd individually.

PR #1821, opened 2026-08-28
fix: raise S3 multipart part size to 32 MiB

sylr:
## Problem

#1631 tried to harden S3 multipart uploads against `CompleteMultipartUpload InvalidPart` by bumping the part size to 32 MiB and forcing `RequestChecksumCalculation=WhenRequired`. Before it could be merged, #1777 (9b376a386) landed on `release/v3.0` fixing the same `InvalidPart` bug via a different, more targeted root cause: a version skew between `feature/s3/manager` and `service/s3` producing an incoherent checksum declaration on `CreateMultipartUpload`. That fix migrated to `feature/s3/transfermanager` and deliberately kept `RequestChecksumCalculation` at its `WhenSupported` default rather than forcing `WhenRequired`.

#1631 is closed as superseded. This PR re-applies the one part of it that's still independently valuable.

## Change

- 32 MiB multipart part size (`s3UploadPartSize`) on both the backup (`internal/infra/backup/s3.go`) and cold-storage (`internal/infra/coldstorage/s3.go`) uploaders, applied via `transfermanager.Options.PartSizeBytes` (transfermanager's default is 8 MiB). At the 10,000-part ceiling this lifts the single-object cap from ~78 GiB to ~312 GiB and cuts part count 4x for multi-GB objects, shrinking the failure surface for any single part.
- Does **not** touch `RequestChecksumCalculation` or the manager/transfermanager choice -- both already fixed by #1777.
- Bumped the two packages' multipart regression tests to apply the same `PartSizeBytes` option and raised their payload size (20 MiB -> 40 MiB) so they actually exercise >=2 parts at the real configured part size instead of only transfermanager's default threshold.

## Verification

- `GOROOT= go build -tags s3 ./...` and `go vet -tags s3 ./internal/infra/backup/... ./internal/infra/coldstorage/...` clean.
- `git diff --check` clean.
- ⚠️ MinIO-backed multipart tests (testcontainers) require Docker, unavailable in this sandbox -- rely on CI.

## Notes

- Operational: with 32 MiB parts, a non-seekable upload can buffer ~160 MiB (concurrency x part size), up from ~40 MiB at the transfermanager default.

Closes: superseded #1631


sylr:
All four findings verified against source and addressed.

**Major — unbounded upload memory** — confirmed exactly: `newDefaultSlicePool(PartSizeBytes, Concurrency+1)` (6 x 32 MiB = 192 MiB steady-state), plus the unpooled `nextReader` first-part buffer (~32 MiB) for a ~224 MiB transient peak; `BackupJobsState` excludes only per destination, and the Archiver can overlap with backups, so nothing bounded aggregate concurrency. Fixed in c2b57d0fd: `AcquireS3UploadSlot` (`internal/infra/coldstorage/s3_uploadlimit.go`) is a process-wide semaphore shared by `backup.PutFile` and `coldstorage.Archive`, fixed at 1 in-flight upload — the doc comment walks through why (production-baseline profile leaves <400 MiB of headroom above the ~3.2 GiB budget at GOMEMLIMIT=3600MiB; a single upload already consumes most of it). Covered directly by two new tests (`TestAcquireS3UploadSlot_*`) that don't need Docker.

**Minor — tests bypass the production constructor / weak assertion** — fixed in c2b57d0fd: both packages now share a `newUploader` helper between `NewS3Storage` and the recorder-based tests, and the `UploadPart` count assertion is now `Equal(2)` instead of `GreaterOrEqual(2)`.

**Minor — CI never runs these tests** — confirmed: `test-coverage` compiles with no tags at all, `test-e2e-coverage` has the s3 tag but only targets `./tests/e2e/...` (`-coverpkg` doesn't make `go test` execute other packages' `_test.go` files). Fixed in dc9e406a3: new `just test-s3-coverage` + `Tests-S3` CI job, wired into `Coverage` the same way `Tests-E2E` is. Scoped to s3 only — the same gap exists for kafka/nats/clickhouse/azure-tagged internal tests but fixing those needs their own external services in CI, out of scope here.

**Minor — 78/312 GiB ceiling comments wrong for the file paths they describe / stale 5 TB PutFile comment** — confirmed: `initSize`'s `io.Seeker` branch auto-upsizes `PartSizeBytes` for any seekable body, and both `manager.go:843` (full backup) and `archiver.go`'s `os.Open` (chapter archives) upload from `*os.File`. Only the unknown-length `io.Pipe` path (`manager.go:788`) has a real fixed ceiling from `PartSizeBytes x MaxUploadParts` (~312 GiB, not 5 TB). Fixed in c2b57d0fd: both `s3UploadPartSize` doc comments and the `PutFile` comment now scope the ceiling claim correctly to the unknown-length path and describe the seekable path as auto-upsizing with no hard ceiling.

PR #1823, opened 2026-08-28
fix(EN-1888): persist a chapter's audit anchor with the close

Azorlogh:
Fixes [EN-1888](https://formance-team.atlassian.net/browse/EN-1888).

## What changed

`CloseChapter` records the chapter's `last_audit_hash` in its own batch. It used
to be assigned to the in-memory registry after the batch had committed.

## Why

That anchor is the chain input `store check` uses for the first audit entry
surviving the chapter's purge, and one of the four fields its sealing hash
commits to. The close left it unset and `applyProposal` stamped it post-commit,
so it reached Pebble only if a later apply rewrote the row from the still-warm
object — normally the seal.

The Sealer reads that same in-memory object and folds the anchor into the sealing
hash it proposes. A replica that rebuilds its registry from Pebble between the
close and the seal — a restart, or a snapshot install — therefore applies the
seal onto a row that cannot reproduce the hash being written to it:

```
ok       id=48 ... lastAudit=2a3f9962…  sealing=92e84e88…  recomputed=92e84e88…
MISMATCH id=49 ... lastAudit=          sealing=1e8b9431…  recomputed=900cbba1…
ok       id=50 ... lastAudit=b3b15254…  sealing=71cf939e…  recomputed=71cf939e…
```

`store check` then fails on that chapter forever, and since the anchor also seeds
verification past the archive boundary, one bad row cascaded into 210
`VOLUME_MISMATCH`, 11 `REVERTED_MISMATCH` and 5 `EXCLUSION_RECORD_MISMATCH`
errors in the run that found it. It also made the persisted result of applying a
committed entry depend on node-local memory, so replicas could disagree.

The placement was not an oversight: `368b7c932` ("perf: hash orders in parallel
with ProcessOrders") made the value this proposal's own audit hash, which did not
exist during order processing. `0f13e1ecf` (#211) later changed it to the
predecessor's hash so the checker could chain from it — available before the
batch is built — and the constraint disappeared while the code stayed put.

## Risk

**LOW.** Same value, one apply earlier, inside the batch. `WriteSet.Reset` seeds
it from FSM state, which costs a slice-header assignment per proposal and no
allocation (`Compute` returns `hasher.Sum(nil)`, a fresh slice, so the aliasing is
safe and the parallel hasher is untouched); the close path loses its `make`+`copy`
and the `chapterClosing` flag it was the only consumer of.

The restore path is the one behavior change beyond that: `RebuildDelta` no longer
re-derives the anchor from the audit entry at `CloseAuditSequence` — the close log
now carries it — and refuses a close log without one rather than restoring a
registry the checker cannot verify.

## Validation

- [x] `./internal/...` (only the pre-existing root-user `tarutil` failures, which
      expect `chmod 0444` to block `MkdirAll`)
- [x] `tests/e2e/cluster` — `ChapterAuditAnchor`
- [x] Both FSM-level tests red-checked against the fix reverted

## Tests

- `TestApplyProposal_CloseChapterPersistsTheAuditAnchor` — the persisted row
  carries the same anchor as the registry, forbidding "filled in by a later write".
- `TestApplyProposal_SealAfterRecoveryReproducesItsSealingHash` — close →
  `RecoverState()` → seal, then the row must reproduce its own sealing hash. This
  is the production sequence, deterministically, in ~0.2s.
- `ChapterAuditAnchor` (e2e, 3 nodes) — across a follower restart, every replica's
  sealed chapter carries the anchor, reproduces its sealing hash, and holds the
  same value.
- `TestRebuildDelta_CloseOnlyDelta_KeepsTheAuditAnchor` and
  `..._CloseWithoutTheAuditAnchorIsRefused` — the restore path.

## Review focus

The e2e is a contract guard, not the detector, and says so: the interleaving is
not reachable through the public API because the seal completes inside the close's
own `Apply` call, leaving no observable `CLOSING` window (measured 261ms / 347ms /
602ms to CLOSED at 0 / 3k / 30k transactions). The FSM-level tests drive that
ordering directly. Worth a look at whether that split is the right place to draw
the line.

Existing damaged rows are repairable and this PR does not repair them: the archive
carries the audit range too (`archiver.go:28-35`), so the entry whose hash is the
anchor sits in the chapter's own cold object, and a re-derivation validates itself
against the stored sealing hash. Filed as a separate concern rather than folded in
here.

[EN-1888]: https://formance-team.atlassian.net/browse/EN-1888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


NumaryBot (commented):

The implementation appears coherent, but the restore-impact change lacks the repository-mandated cross-lifecycle regression and therefore does not establish incremental restore parity.

### Findings outside the diff

🟠 **[major] Add the required end-to-end restore regression** — `internal/infra/backup/rebuild_chapters_test.go:248`

This only invokes `RebuildDelta` on hand-written rows, so it cannot prove that a real post-checkpoint close is exported, restored, and checker-valid. The restore contract explicitly requires checkpoint-plus-nonempty-delta coverage through `ApplyExportsAndRebuild` or the restore service and a clean `CheckStore` result ([incremental-restore-contract.md:60-74](/data/worktrees/formancehq__ledger-pr-1823/docs/technical/architecture/subsystems/chapters/incremental-restore-contract.md#L60-L74)); add that regression for the newly changed close-log anchor.




Summary: https://github.com/formancehq/ledger/pull/1823#issuecomment-5453341930

Azorlogh:
Restore regression added in https://github.com/formancehq/ledger/commit/585ea5ac0eceb6b85bdd726612e05682d698d101 — `tests/e2e/cluster/restore_chapter_audit_anchor_test.go`.

The chapter is closed after the full checkpoint (so the anchor exists only in the delta), the delta is restored through the real restore service, the restored row is compared field by field with the source, and the chapter is then archived on the restored store — which is what puts it through `verifySealingHash`, since that decomposition is only checked for `ARCHIVED` chapters — followed by a clean `CheckStore`.

Red-checked: with the close no longer stamping the anchor, the spec fails. It guards the new contract rather than detecting the original defect, which the rebuild's re-derivation used to mask; the two FSM-level tests cover that.

internal/infra/backup/rebuild.go
```diff
@@ -882,25 +882,14 @@ func (w *attributeReplayWriter) storeChapterRow(chapter *commonpb.Chapter) error
...
-		// purges, so no purge reaches CloseAuditSequence of a delta close),
-		// and below-floor entries ride the checkpoint SSTs.
-		entry, err := query.ReadAuditEntry(ctx, w.readHandle, closed.GetCloseAuditSequence())
-		if err != nil {
-			return fmt.Errorf("reading audit entry %d to recover closed chapter %d's last audit hash: %w",
-				closed.GetCloseAuditSequence(), closed.GetId(), err)
-		}
-
-		closed.LastAuditHash = entry.GetHash()
+		return fmt.Errorf("closed chapter %d carries no audit anchor at close audit sequence %d: refusing to rebuild a registry the checker cannot verify",
```
flemzord: Checking only for an empty anchor is not sufficient here. `ClosedChapterLog.last_audit_hash` is a projection carried by the log payload, but those payload bytes are not themselves bound to the audit chain. A non-empty corrupted value would therefore be persisted and could later be sealed self-consistently, leaving the chain unverifiable after archival removes the predecessor range. Please read the authoritative audit entry at `CloseAuditSequence` and either derive the anchor from `entry.Hash` or reject a mismatch; add a regression with a non-empty incorrect anchor.

tests/e2e/cluster/restore_chapter_audit_anchor_test.go
```diff
@@ -0,0 +1,376 @@
...
+				ClusterID: clusterID,
+				Ports:     ports,
+				WalDir:    GinkgoT().TempDir(),
+				DataDir:   GinkgoT().TempDir(),
+				Debug:     testutil.Debug,
+				Output:    GinkgoWriter,
+			})
+			instruments = append(instruments, testserver.WithBootstrap(), withColdStorage())
+
+			sourceServer = testservice.New(cmdserver.NewRunCommand, testservice.WithInstruments(instruments...))
```
flemzord: This test keeps a `NodeLease` alive but starts the service with `testservice.New`, so the reserved sockets are not transferred to the process and the test reopens the probe-to-bind race. Please start this and the two later phases through `lease.NewService(cmdserver.NewRunCommandWithBindings, ...)`, as the neighboring restore tests do.

flemzord (commented):
Two blocking issues found during the latest review pass.

gfyrag (dismissed):
Independent review at exact head `585ea5ac0eceb6b85bdd726612e05682d698d101`.

The closing chapter captures the audit anchor before persistence, recovery/rebuild reject a missing durable anchor, and the state mutation order now matches the documented durability contract. Restore/checker coverage is present and functional CI is green.

internal/infra/backup/rebuild.go
Azorlogh: Fixed in 13c1a1373: the rebuild now reads the audit entry at `CloseAuditSequence` and refuses a mismatched anchor, with the non-empty-incorrect-anchor regression added. The entry is always resident when a close log is replayed — archival purges a chapter's logs and audit entries together, so a purge that removed the entry would have removed the close log with it.

tests/e2e/cluster/restore_chapter_audit_anchor_test.go
Azorlogh: Fixed in 7a30fbed4: all three phases now start through `lease.NewService(cmdserver.NewRunCommandWithBindings, ...)`. It was slightly worse than a probe-to-bind race: the spec captured only `ports`, so the lease itself was garbage-collected after tree construction and the netFD finalizers released the reserved sockets — the reservation was gone from the first GC onward.

Azorlogh:
@gfyrag I fixed flemzord's findings about the tests, can you recheck?

gfyrag (approved):
Independent re-review at exact head 7a30fbed497e671de0c464f930032fdcb8b26ec9.\n\nThe restore rebuild now verifies the close-log audit anchor against the authoritative audit entry at CloseAuditSequence before persisting the chapter row. The audit entry is resident whenever the close log is replayed, and the new mismatch regression pins the failure path. The E2E phases now use the repository node-lease binding contract. No prior finding remains and no new finding was found. Targeted backup, processing, and state tests pass, and current CI is green.\n\nDECISION: APPROVE\nBlocking findings: 0\nNon-blocking findings: 0\nResidual risk: MEDIUM

PR #1826, opened 2026-08-28
fix(EN-1901): normalize degenerate query range bounds to the empty match

Azorlogh:
## What changed

The query compiler normalizes degenerate range bounds — crossed bounds and empty exclusive intervals like `(x, x)` — to the canonical empty match instead of emitting them as pebble iterator bounds. `not()` over such a range now matches the whole universe.

## Why

Jira: EN-1901. An inverted iterator range panics pebble under the invariants build and has undefined matching on the default build. Found by the EN-1625 model-checking driver query hunts.

N/A (bug fix)

## Technical decision

N/A

## Risk

**LOW** — normalization at compile time; no storage or wire change.

## Validation

- [x] Targeted tests: `internal/query/compile_bounds_test.go` (unit); `tests/e2e/business/logs_degenerate_bounds_test.go` — RED on base under `-tags e2e,invariants` (pebble bound panic), green with the fix. On the default build the e2e is green on base too (pebble tolerates the range there); the invariants build is where the panic is observable.
- [x] Full `internal/query` package suite green.

N/A

## Review focus

The bound-comparison logic in `compile.go` — the empty-interval and min>max cases must collapse to the same canonical empty match so the complement stays exact.

## Known concerns

None

### DIFF
--- a/internal/query/compile.go
+++ b/internal/query/compile.go
@@ -649,6 +649,15 @@ func resolveIntBounds(cond *commonpb.IntCondition, params map[string]*commonpb.P
 		}
 	}
 
+	// max is the exclusive upper, so [min, max) holds nothing once the
+	// adjusted bounds meet or cross — e.g. `(x, x)` resolves to [x+1, x).
+	// Every consumer must see that as the empty match, never as Pebble
+	// iterator bounds: LowerBound above UpperBound is an iterator
+	// invariant violation, not an empty scan.
+	if b.hasMin && b.hasMax && b.min >= b.max {
+		b.empty = true
+	}
+
 	return b, nil
 }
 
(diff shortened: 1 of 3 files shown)


Azorlogh:
Addressed the arbitration's blocking item: the Dirty failure was the predeclared linter on the new test helper's min/max params — fixed in b548240d0. The codecov/patch gap was real, in compile.go itself: the txID and timestamp-range guards had no unit coverage — d7a89ec9d adds a compile-level test driving all three consumer guards directly (a zero compileCtx makes reaching past a guard a nil dereference, so the short-circuit is pinned). agent-check green locally.

flemzord (approved):
Reviewed head d7a89ec9d15eb961ad1dabdf9a46c14ebef9c9e7 on Standards and Spec. The adjusted-bound normalization is correct, every resolver consumer short-circuits before constructing Pebble iterator bounds, and the regression coverage pins both the empty match and not(empty) semantics. Current CI is green; targeted standard/invariants tests and the full internal/query suite on a simulated merge with current release/v3.0 also pass.

gfyrag (approved):
Independent review at exact head `d7a89ec9d15eb961ad1dabdf9a46c14ebef9c9e7`.

The half-open bound normalization correctly collapses degenerate int/uint ranges to the empty iterator, including negated expressions, without changing valid-range semantics. Focused tests and functional CI are green.

PR #1827, opened 2026-08-28
fix(EN-1902): recover WAL snapshotting after the snapshot directory vanishes

Azorlogh:
## What changed

WAL snapshot saves recreate the snapshot directory when it has vanished under a live node, and fsync the parent directory before the rename so the directory entry is durable before any WAL record references it.

## Why

Jira: EN-1902. Once the snapshot directory disappeared (operator mistake, cleanup job), every subsequent save failed permanently — no snapshot ever persisted again, and the follower-sync save path propagated the failure as a task-pool panic that took the node down.

N/A (bug fix)

## Technical decision

N/A

## Risk

**LOW** — one extra parent fsync per save; the new error path only fires on a genuine durability failure, where failing the save is correct. (An automated arbitration pass on the closed bundle PR #1817 challenged the unconditional parent fsync and ruled it correct-as-written: `NewSnapshotter` creates the directory without fsyncing its parent, and `Save` promises crash-safe durability.)

## Validation

- [x] Targeted tests: `snapshotter_missing_dir_test.go` (unit, incl. save-failure injection via a file-occupied parent); `tests/e2e/business/wal_snapshot_dir_recovery_test.go` — RED on base (directory never recreated within 30s), green with the fix.
- [x] Full `internal/storage/wal` package suite green.

Snapshot `Save` gains a parent-directory fsync; no format or wire change.

## Review focus

The recreate-then-rename ordering in `snapshotter.go` — the parent must be synced before a WAL record can reference the directory.

## Known concerns

None

### DIFF
--- a/internal/storage/wal/snapshotter.go
+++ b/internal/storage/wal/snapshotter.go
@@ -47,6 +47,14 @@ func (s *Snapshotter) Save(snap *raftpb.Snapshot) error {
 	path := filepath.Join(s.dir, name)
 	tmpPath := path + ".tmp"
 
+	// The directory is created in NewSnapshotter, so reaching here without it
+	// means it went away underneath a running node. Recreate it: the snapshot
+	// this call is persisting is how a follower catches up, and failing the
+	// save propagates as a task-pool panic that takes the node down.
+	if err := os.MkdirAll(s.dir, 0755); err != nil {
+		return fmt.Errorf("recreating snapshot directory: %w", err)
+	}
+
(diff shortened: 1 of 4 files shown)


Azorlogh:
Closing: duplicate of #1737, which additionally handles the unrecoverable case (WalDir itself gone → refuse with ErrWALDirectoryMissing) that this PR's unconditional MkdirAll would silently mask. The full-node red e2e from this branch (wal_snapshot_dir_recovery_test.go) is available to donate to #1737 if wanted.

PR #1828, opened 2026-08-28
fix(EN-1903): bound the checker's stored-exclusion scan to the archived audit boundary

Azorlogh:
## What changed

The checker's stored-exclusion scan is bounded to the archived audit boundary: AppliedProposal transient/purged rows at or below the boundary no longer enter the stored exclusion set.

## Why

Jira: EN-1903 (possibly related: EN-1329). The replay side derives exclusions only past the archive, so a store with pre-archive transient traffic could fail its own consistency check after a segment-level restore — a checker false positive, not data corruption.

N/A (bug fix)

## Technical decision

N/A

## Risk

**LOW** — checker-only change; no server write path touched.

## Validation

- [x] Targeted tests: `checker_exclusion_archived_test.go` pins the boundary (the pre-fix code does not compile against it).
- [x] Full `internal/application/check` package suite green.
- No red e2e is possible: a live node keeps both exclusion derivations consistent (verified empirically green-on-old); the divergence needs a raw-segment restore the e2e harness cannot stage.

N/A

## Review focus

That the boundary comparison matches the replay side's archive cutoff exactly — off-by-one here would just move the false positive.

## Known concerns

None

### DIFF
--- a/internal/application/check/checker.go
+++ b/internal/application/check/checker.go
@@ -213,6 +213,7 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 	var (
 		hasArchivedChapters  bool
 		archiveEndSeq        uint64 // max close_sequence among archived chapters
+		archiveAuditEndSeq   uint64 // max close_audit_sequence among archived chapters
 		archiveLastAuditHash []byte // last_audit_hash from the latest archived chapter
 	)
 
@@ -231,6 +232,10 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 			if p.GetCloseSequence() > archiveEndSeq {
 				archiveEndSeq = p.GetCloseSequence()
 				archiveLastAuditHash = p.GetLastAuditHash()
+			}
+
+			if p.GetCloseAuditSequence() > archiveAuditEndSeq {
+				archiveAuditEndSeq = p.GetCloseAuditSequence()
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


internal/application/check/checker.go
```diff
@@ -783,7 +788,17 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 	// projection. Combined with the LedgerLog.PurgedVolumes already
 	// accumulated above, `stored` now represents the full per-ledger
 	// exclusion set the index builder will consume.
-	if err := c.collectStoredTransientVolumes(ctx, snap, addStored); err != nil {
+	//
+	// The scan starts strictly above the archived audit boundary, mirroring
+	// the replay loop's `seq <= archiveEndSeq` skip on the log side: the
+	// derived set only covers post-boundary history, and rows at or below the
```
NumaryBot, line 794: 🟠 **[major] Document the archived exclusion coverage boundary**

This intentionally leaves retained `AppliedProposal.TransientVolumes` rows at or below the archive boundary unverified, but the authoritative checker documentation still states that pass 7 verifies the projection without this limitation. Update the checker subsystem documentation to describe the post-boundary-only coverage and restore trade-off, as required by [AGENTS.md lines 78-85](/data/worktrees/formancehq__ledger-pr-1828/AGENTS.md#L78-L85).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1828#issuecomment-5453794158

Azorlogh:
Closing: line-for-line duplicate of #1682 (same fix and test, re-derived from the same model-hunt work).

PR #1829, opened 2026-08-28
fix(EN-1724): bind declared metadata types to index versions

Azorlogh:
## What changed

Declared metadata types are bound to index versions instead of read live from the schema:

- `CreatedIndexLog` gains `bound_type` / `bound_type_declared`, stamped by the FSM at mint time — the schema entry in force at exactly the log's sequence. The indexbuilder binds the first version from the stamp; the batch-final schema read (and its silently swallowed error path) is gone from the creation fold.
- `IndexVersionState` carries `CurrentType`/`PendingType` (+ declared flags); live dual-writes encode each value once per version under that version's bound type (`coerceForVersion`), and queries validate/encode conditions under `CurrentType` until the atomic switch promotes `PendingType` with the version.
- Resumed schema rewrites read `toType` from `IndexVersionState.PendingType` (committed in the same batch by `bumpPendingVersion`); the cursor's leading type byte is a cross-check that restarts the scan on mismatch. `resolveResumedToType`, which guessed STRING on any read failure, is deleted.
- `handleCreatedIndexLog` returns an error: a failed version-state persist aborts the indexer batch so the log is re-folded.
- Indexer docs (`indexer.md`, `indexes.md`) rewritten to the per-version binding contract.

## Why

Jira: EN-1724. Mid-rewrite queries mixed encodings: the schema flips at FSM apply, but `v_current`'s rows still carry the old type — a condition compiled under the new declared type over old-encoded rows sees only rows written after the retype (typed encodings occupy disjoint key ranges), i.e. silently partial results. Found by the EN-1625 model-checking driver retype-window oracle.

This PR also resolves the three blocking findings from the automated reviews (NumaryBot + shipfox arbitration) on the closed bundle PR #1817: resume guessing STRING on a read fault, `boundTypeAtCreation` swallowing schema-read errors into a persisted false "not declared", and the creation binding leaking the batch-final schema.

Need: queries must never silently drop rows during or after a metadata retype.
Current limitation: the binding was derived from whatever schema was readable at fold time — batch-final at best, arbitrarily newer when a log folds during a backfill or rebuild replay.
Requirement / constraint: every replica folding a given CreateIndex log derives the same type binding, at any replay distance; the old view stays complete until the atomic switch.
Evidence: EN-1724; model-driver retype-window oracle; review findings on #1817.
Durable repository evidence: `docs/technical/architecture/subsystems/indexer/indexer.md` + `indexes.md` (updated in this PR), field comments on `CreatedIndexLog.bound_type` and `IndexVersionState`.

## Technical decision

Decision: stamp the binding into the log at FSM mint time (`processCreateIndex` has the at-sequence schema in hand; `validateIndexTarget` already guarantees the field is declared for metadata indexes).
Why now / why proportionate: any fold-time schema read is wrong at replay distance — only the log itself can carry the at-sequence binding. v3 is unreleased, so adding proto fields is free.
Alternatives considered: an in-batch schema overlay in the indexbuilder (fixes only the earlier-in-batch leak, not later-in-batch or replay); propagating the schema-read error without stamping (keeps a read that cannot be made correct, only loud).

## Risk

**MEDIUM** — touches the FSM mint path, the indexer fold path, and persisted per-replica state semantics. Mitigated by: no compat surface (v3 unreleased, wiped per release/v3.0 policy), deterministic stamp (proposal-carried, not re-derived per replica), and the test set below.

## Validation

- [x] Targeted tests: processor stamping (at-sequence binding + builtin-carries-no-binding), fold binding from the stamp (initial + backfilling), resume-before-first-batch from `PendingType`, cursor-type mismatch restart — the two resume specs verified RED on the pre-fix code; retype-window and version-binding suites green.
- [x] Full `internal/domain/processing` + `internal/application/indexbuilder` package suites green.
- The model driver's retype-window oracle covers the serving window continuously in Antithesis runs (no red e2e: the window is transient and needs builder pause control the harness lacks).

New proto fields `CreatedIndexLog.bound_type` / `bound_type_declared` (v3 unreleased — no reservation or compat shim per repo policy). `IndexVersionState` header is 22 bytes with per-version type bindings. The declared-type flip becomes visible to queries atomically at the version switch.

## Review focus

1. The stamp's determinism: it is minted once in `processCreateIndex` and carried by the log — confirm no replica-side re-derivation remains.
2. The idempotent re-create path (READY short-circuit) also stamps at the duplicate's sequence; the fold-side `current != 0 / pending != 0` guards make that stamp inert on replicas that saw the original.
3. The cursor-mismatch restart in `scheduleResumedRewrites` (can't-happen by same-batch commit; restart-from-zero is the safe recovery).

## Known concerns

A pre-existing `IndexVersionState` written before this change carries no binding — irrelevant on release/v3.0 (state wiped every push), noted for completeness.

### DIFF
--- a/docs/technical/architecture/subsystems/indexer/indexer.md
+++ b/docs/technical/architecture/subsystems/indexer/indexer.md
@@ -90,7 +90,7 @@ flowchart TB
 
 ### Dual-write while a rewrite is in flight
 
-When `IndexVersionState.PendingVersion != 0`, metadata-index handlers go through `dualWriteMetadataIndex` (`builder.go:194-226`) and write the **same encoded value** to **both** `v_current` and `v_pending` keyspaces. The encoding always uses the **current declared type** in `LedgerInfo` — which, during a `SetMetadataFieldType` rewrite, is already the *new* type (the FSM has applied the schema change before the indexer dispatches the log). The two versions differ only by the version embedded in their keys, not by the encoding of the value.
+When `IndexVersionState.PendingVersion != 0`, metadata-index handlers go through `dualWriteMetadataIndex` (`builder.go`) and write to **both** `v_current` and `v_pending` keyspaces, encoding the value once per version under **that version's bound type** (`coerceForVersion`): `v_current` rows keep `CurrentType`'s encoding, `v_pending` rows carry `PendingType` — the retype's target (EN-1724). During a retype the two versions therefore differ both by the version embedded in their keys and by the encoding of the value; the per-version binding is what keeps `v_current`'s view complete and correctly typed until the atomic switch. The `CreatedIndexLog` itself carries the first version's binding (`bound_type`, stamped by the FSM at mint time), so a replica folding it at any replay distance binds the same type.
 
 This is what allows the atomic switch (see below) to flip the served version with zero rebuild cost at the moment of the switch — `v_pending` is already fully consistent with live writes by the time the rewrite cursor reaches the head. See the [Changing a Metadata Key's Type](#changing-a-metadata-keys-type-setmetadatafieldtype) section for the transient query semantics this design produces during the rewrite window.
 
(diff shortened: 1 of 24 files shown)


flemzord (approved):
Reviewed at f9a5d811d3718c1201a6bbee868d1c1ce194f8ac across Standards and EN-1724 Spec axes. The per-version type binding is propagated consistently through FSM stamping, persisted state, dual-writes, rewrite resume/switch, and query compilation. CI and isolated canonical/targeted validation are green; no blocking findings.

gfyrag (approved):
Independent review at exact head `f9a5d811d3718c1201a6bbee868d1c1ce194f8ac`.

The FSM stamps the creation-time declared type, the index builder persists the per-version bindings, live dual writes coerce against each version's bound type, and query compilation consumes the served version consistently. Proto numbering and generated artifacts are aligned.

Non-blocking follow-up: add a focused second-retype regression asserting that a non-zero `ActivationSequence` is preserved.

PR #1830, opened 2026-08-28
fix(EN-1904): stop the index-registry preload serving stale absence

Azorlogh:
## What changed

Apply-side index-registry preloads no longer trust anything but pebble for index keys: the bloom veto and the cache-hit shortcut are bypassed for `SubAttrIndex`, a loader-cached absence is arbitrated against a fresh pebble read, and a failed arbitration read (or marshal) fails the resolve instead of degrading to a coverage-only plan.

## Why

Jira: EN-1904. Three holes let the apply path read an existing index registry row as absent — deterministically on every replica, since the plan is replicated — driving wrong CreateIndex/DropIndex lifecycle decisions. Found by the EN-1625 model-checking driver hunts (Antithesis assert class "index preload served stale absence").

The error-propagation fix resolves the blocking finding from the automated reviews on the closed bundle PR #1817: both `freshErr` and the arbitration marshal error now land in `firstErr`, matching the primary load and marshal paths — coverage-only authorizes the key with no seed, so apply would route back through the very cache the arbitration exists to distrust.

N/A (bug fix; index registry keys are rare, so the exempted pebble reads cost nothing)

## Technical decision

N/A

## Risk

**LOW** — scoped to `attrCode == dal.SubAttrIndex`; other attribute types keep the bloom/cache shortcuts unchanged.

## Validation

- [x] Targeted tests: `resolve_arbitration_test.go` pins all four behaviors (fresh-read failure fails the resolve, arbitrated value seeds the plan, bloom-veto exemption, cache-hit fallthrough) — every spec verified RED against the pre-fix resolver.
- [x] Full `internal/infra/plan` package suite green.
- No red e2e: these are race-window bugs needing adversarial scheduling (a rotation+restart recipe stays green on old, verified empirically); the windows stay covered continuously by the Antithesis assert class.

Index-key preloads always perform a pebble read on the miss/hit paths; plans for index keys are seeded with the durable row rather than coverage-only whenever pebble holds one.

## Review focus

The arbitration block in `resolve.go` — the error contract must match the primary load path exactly, and the confirmed-absence case must still fall through to coverage-only.

## Known concerns

None

### DIFF
--- a/internal/infra/plan/resolve.go
+++ b/internal/infra/plan/resolve.go
@@ -7,7 +7,9 @@ import (
 
 	logging "github.com/formancehq/go-libs/v5/pkg/observe/log"
 
+	"github.com/antithesishq/antithesis-sdk-go/assert"
 	"github.com/formancehq/ledger/v3/internal/infra/attributes"
+
 	"github.com/formancehq/ledger/v3/internal/infra/bloom"
 	"github.com/formancehq/ledger/v3/internal/infra/cache"
 	"github.com/formancehq/ledger/v3/internal/infra/preload"
@@ -213,7 +215,14 @@ func resolveCoverage[T interface {
 			// Bloom filter short-circuit: when the key is definitely not
 			// in Pebble, skip the goroutine + Pebble Get and emit Declare
 			// (coverage-only, no value to seed).
-			if bloomFilter != nil && !bloomFilter.MayContain(id) {
+			//
+			// Index registry keys are exempt: they are rare enough that the
+			// Pebble read costs nothing, and a false negative here would make
+			// every replica read an existing index as absent — the removal
+			// then reports nothing dropped while the row sits in Pebble.
+			// The load below both corrects it and proves the filter wrong.
+			bloomAbsent := bloomFilter != nil && !bloomFilter.MayContain(id)
+			if bloomAbsent && attrCode != dal.SubAttrIndex {
 				mu.Lock()
 				plans = append(plans, slab.appendCoverage(id, tag, attrCode))
 				mu.Unlock()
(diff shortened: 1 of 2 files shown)


internal/infra/plan/resolve.go
```diff
@@ -261,6 +282,17 @@ func resolveCoverage[T interface {
...
+					}).Errorf("Bloom filter denied a key Pebble holds")
+					assert.Unreachable("index preload bloom false negative", map[string]any{
+						"type": typeName,
+						"key":  hex.EncodeToString(canonicalKey),
+					})
+				}
+
 				// Track bloom false positives: MayContain said "maybe" but Pebble
 				// had nothing. Only counts loads we actually performed (FromLoad).
 				if result.FromLoad && !hasValue && bloomFilter != nil {
```
NumaryBot: 🟠 **[major] Exclude bloom-negative loads from false-positive metrics**

When an index key is absent and `MayContain` returned false, the new exemption still loads Pebble and this condition records a false positive. A false positive requires `MayContain` to have returned true, so ordinary absent index lookups now inflate the metric; also require `!bloomAbsent` before recording it.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1830#issuecomment-5453794297

internal/infra/plan/resolve.go
Azorlogh: Fixed in 382764bbd — the recording now also requires `!bloomAbsent`, so a bloom-negative index-key load that finds nothing counts as the true negative it is.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1830#issuecomment-5453794297

gfyrag (approved):
Independent review at exact head `382764bbd6f3e1e3e0b1c3a274467905fa9c621f`.

The index-specific preload arbitration is outside the FSM apply path, propagates read/marshal failures, bypasses unsafe bloom/cache shortcuts, and correctly seeds durable Pebble truth. Targeted `internal/infra/plan` and `internal/infra/preload` tests pass.

Non-blocking follow-ups:
- update the enclosing resolver contract and `docs/technical/architecture/subsystems/fsm/preload.md` for the `SubAttrIndex` exception;
- add the explicit loader-absence + fresh-Pebble-absence regression proving the coverage-only fallthrough.

PR #1834, opened 2026-08-28
test(model): model the chapter lifecycle in the oracle

Azorlogh:
**Depends on** #1850 and #1854 at run time, not at build time: the model test
compiles and runs against `release/v3.0`, but a combined `--archive --restore`
run's teardown `store check` needs #1850 (residency purges — otherwise restored
stores legitimately hold re-ingested archived ranges the checker now reports) and
#1854 (baseline bound to the archived boundary — otherwise the teardown check
passes or fails by pacing luck).

History note: this branch previously carried a duplicate of #1682's scan-bound
commit as a workaround. #1682 is closed — the bound silenced a true signal — and
the commit is dropped, so the checker runs unbounded here.

Draft until the combined run on top of #1850 + #1854 is green.

## What changed

The model test's oracle now models the chapter lifecycle, and the driver's chapter
orders are paced from a log-normal instead of a fixed interval.

### The oracle owns which chapter orders are valid

The driver used to decide for itself which chapter orders were legal and only send
those, which put the specification in the generator: an order the server accepted
but the driver believed invalid was never sent, and one the server rejected for a
reason the driver did not model looked like a business finding. Chapter lifecycle
is observable behavior, so the oracle models it.

`Chapters` is a normal form over the registry — `archivedThrough`, an `archiving`
flag, a list of live CLOSED/CLOSING chapters, and whether one is open. It cannot
represent a hole in the archived prefix or a CLOSED chapter below an ARCHIVING
one, and `nextChapterID` is derived rather than stored. `applyClose` and
`applyArchive` mirror the server's check order, including the prefix check running
before the residency lookup.

Two transitions carry no request — the Sealer's CLOSING→CLOSED and the Archiver's
ARCHIVING→ARCHIVED with its purge — so a chapter can be one step ahead of the
model. The search folds those as ordinary moves bounded by `chaptersInPlay`, and
the generator now guides distributions only: it emits any chapter order the
emission gate allows, out-of-order archives included, and the oracle decides
whether the server was right to refuse.

### Pacing is drawn from a log-normal

Fixed gaps held the registry in one shape. The successor was archived seconds
after the Sealer sealed it, so it was the only sealed chapter above the archived
prefix, and an archive request past it could only be answered
`CHAPTER_NOT_CLOSED` or `CHAPTER_NOT_FOUND`. `CHAPTER_ARCHIVE_OUT_OF_ORDER` needs
two sealed chapters at once, which needs a stretch with no accepted archive in it
— so that reason had never been observed, and the coverage guard passed only
because a re-archive carried the same reason before #1683 split them.

Each gap is now drawn from a log-normal whose median is the configured value, so
`MODEL_ARCHIVE_INTERVAL` keeps meaning what it says while the tail supplies those
stretches. Sampling inverts the normal CDF over a uniform Antithesis draw, which
makes the pause length a value the platform can steer rather than a constant it
cannot see. Measured on the 2.5s archive base: median 2.49s, p90 7.9s, p99 20.5s,
6.1% past 4×, 1.07% past 8×. The ceiling on a single draw is run economics — an
unbounded one idles chapter orders for most of a short run.

The coverage guard splits in two, one per rule, so neither rejection can stand in
for the other.

## Why it matters

The pacing change found [EN-1888](https://formance-team.atlassian.net/browse/EN-1888)
on its first run: a chapter whose stored sealing hash could not be reproduced from
its own fields, plus 226 cascade errors after the archive boundary. The
business-level oracle reported nothing — only the teardown `store check` caught
it. #1823 fixes it.

## Validation

- [x] `tests/oracle`, `internal/infra/state`, and the workload module's driver tests
- [x] Previous run (before the fix): 55 archival cycles, 20 restart cycles, zero
      model findings, both coverage guards satisfied, `store check` red with EN-1888
- [x] Confirming run on this stack with the fix: 44 archival cycles, 20 restart
      cycles, zero model findings, both coverage guards satisfied, and
      `store check: clean` — the run that found EN-1888 failed it with the
      chapter's `HASH_MISMATCH` plus 226 cascade errors

## Review focus

Whether folding the two unrequested transitions as search moves is the right call
versus forking candidate bases, and whether the log-normal's tail is fat enough
for short local runs while staying honest on the platform.

[EN-1888]: https://formance-team.atlassian.net/browse/EN-1888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gfyrag (changes requested):
Independent review at exact head `d6606b19099e5be9d58b105d39efa723aa01c751`.

[P1][blocking] Drop the checker scan bound superseded by #1850
Location: `internal/application/check/checker.go:213-239,795-801,902-915`
Evidence: this head carries `ab69ec30f9`, bounding stored `AppliedProposal` receipts at `archiveAuditEndSeq`. #1682 was explicitly closed in favor of #1850 because that bound silences a valid residency detector instead of repairing the hot/cold restore divergence.
Impact: archived-range residue in hot storage can pass `store check`.
Resolution: remove that commit and rebase the stack onto the causal residency fix.

[P2][blocking] Reject a non-empty chapter registry with no OPEN successor
Location: `tests/oracle/chapters.go:120-160`
Evidence: `ChaptersFrom` validates the position of an OPEN chapter but never requires one. Every committed close atomically creates its OPEN successor; only the completely empty pre-bootstrap registry may lack one.
Impact: the oracle accepts loss of the open chapter as a valid normal form.
Resolution: reject `len(observed) > 0 && !c.hasOpen` and add all-CLOSED/all-ARCHIVED regressions.

Azorlogh:
The 1h Antithesis run (`model-archival-combo-1h`, 2a269ed1) found one real hole — in the test itself, not the ledger. `chapterAdvanceExplained` was a hand-written per-chapter delta table, and it rejected a legal composition: a chapter created, closed and sealed entirely between two observations appears CLOSED on first sight (`observed: 2:ARCHIVING 3:CLOSED 4:OPEN` from `model: 2:OPEN` with two closes and archive(2) outstanding). Fault injection stretched the observation gap; the table knew every individual step but not their product.

Fixed in 447f6eaa6 by deleting the table: reachability is now a memoized fold over the outstanding chapter orders plus the two autonomous transitions, using the oracle's own moves (`WithClosed`/`WithArchived` join `WithSealed`/`WithConfirmed`), so the lifecycle is stated exactly once. Skipping an order in the fold covers both not-yet-applied and applied-but-rejected; the response validation still judges actual outcomes. The finding's composition is pinned in the test table (red-checked: the old table fails it), plus a confirm-landed extension and an unreachable negative.

Run summary otherwise: model business asserts all quiet, both archival-rejection coverage asserts hit, in-cluster store checks clean. The run's other NEW finding (ledger container exit 1) is pre-existing: learner registration aborts on a transient gRPC failure under network faults instead of retrying, and cobra dumps usage on the way out — separate issue, not this branch.

gfyrag:
Review checkpoint for tomorrow: please rebase this draft onto the current `release/v3.0` (`d095a8b085553f6045573d0317003b605d420193`) and resolve the current merge conflict. One blocker remains at the current head: `ChaptersFrom` accepts a non-empty registry with no `OPEN` chapter, although every committed close atomically creates its successor. Please reject `len(observed) > 0 && !c.hasOpen` and add all-`CLOSED` / all-`ARCHIVED` regressions. `Tests-E2E-Cluster` is also red and should be re-evaluated after the rebase.

Azorlogh:
closed since chapters are removed

PR #1844, opened 2026-08-28
feat(operator): reconcile NATS event sinks

flemzord:
## Summary

- add a validated `Cluster.spec.sinks.nats` API for Ledger v3 NATS JetStream sinks
- reconcile sink configuration at runtime through Ledger's Raft-backed `ledgerctl events` API, without restarting the StatefulSet
- track operator-owned sink names in status so externally managed sinks are preserved and name conflicts fail visibly
- include the `nats` build tag in direct/PR Ledger images and the Operator E2E image
- add documentation, a sample manifest, unit coverage, and a permanent Kind/Chainsaw scenario

## Declarative behavior

- absent `spec.sinks`: runtime sinks remain unmanaged
- present `spec.sinks` (including `{}`): the Operator maintains the declared set and removes only sinks it previously created
- `SinksSynced=True`: the Raft-replicated sink configuration matches the CR
- NATS stream provisioning and delivery monitoring remain external; the stream must capture `<topic>.>`
- this first CRD surface covers NATS only, and rejects credentials embedded in the URL because CR specs are not secret storage

## Validation

- `nix develop --command bash scripts/agent-check` — PASS
- `nix develop --command bash -c 'unset GOROOT; cd misc/operator && go test ./...'` — PASS
- `nix develop --command bash -c 'unset GOROOT; go test -tags nats ./internal/application/events ./cmd/ledgerctl/events'` — PASS
- `chainsaw test --config e2e/chainsaw-test.yaml --test-dir e2e/tests/event-sinks-nats --kube-context kind-ledger-e2e` — PASS

## Kind + NATS evidence

The permanent Chainsaw scenario built the local Ledger image with `BUILD_TAGS=s3,nats`, installed this Operator into Kind, provisioned NATS JetStream, and configured the sink only by updating the `Cluster` CR.

- CR status converged to `appliedSinks: [primary]` and `SinksSynced=True`
- sink-only update did not roll Ledger: pod UID remained `160f13d7-e04f-4260-ada1-647524ac9465` and spec hash remained `3ce71741975bbad7f72580b58b22399ff8c1331f4aa24a52ff5be3ccd5c99f13`
- NATS received JSON payloads for `CREATED_LEDGER` and `COMMITTED_TRANSACTION`, both identifying `operator-sink-e2e`
- `ledgerctl events list --json` reported sink `primary`, cursor `3`, and `error: null`
- JetStream `LEDGER_EVENTS` reported `messages: 2` across two subjects
- test cleanup completed successfully

Images used for the proof:

- `nats:2.12.14-alpine@sha256:7cef1bd3fed6034e95cf6e6bc9c28c5afa6dc58e9fb778dd7924a1ac62569f2d`
- `natsio/nats-box:0.19.2@sha256:8031d190c7ee24081f3f27cc939fb647a1eeb29ebb5c60fef9b5b6c7a846d6a2`

## Notes for reviewers

Updating an operator-owned sink is intentionally a two-pass remove/recreate because Ledger currently rejects duplicate sink names. Ledger retains the per-name cursor, so committed events remain eligible for at-least-once delivery while the configuration converges. `SinksSynced` proves configuration convergence; the E2E scenario separately proves delivery and cursor advancement.


misc/operator/internal/controller/cluster_sinks_reconcile.go
```diff
@@ -0,0 +1,195 @@
...
+
+	changed := false
+	for _, sink := range diff.toCreate {
+		if _, err := exec(addNATSSinkArgs(sink)...); err != nil {
+			return false, err
+		}
+		created = append(created, sink.name)
+		changed = true
+	}
+
```
NumaryBot, line 175: 🔴 **[blocker] Preserve ownership when status persistence fails**

If `add-sink` succeeds but the deferred status update transiently fails, the next reconcile sees an identical existing sink with no `appliedSinks` entry and intentionally does not adopt it. The operator then permanently treats the sink it created as external, so removing it from the spec will not remove the runtime sink. Ownership must remain recoverable or sink reconciliation must not report successful creation until the status record is durably persisted.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1844#issuecomment-5482521463

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1844#issuecomment-5482521463

misc/operator/internal/controller/cluster_sinks.go
```diff
@@ -0,0 +1,287 @@
...
+	var diff eventSinkDiff
+	for _, sink := range desired {
+		desiredByName[sink.name] = sink
+		actualSink, exists := actual[sink.name]
+		if !exists {
+			diff.toCreate = append(diff.toCreate, sink)
+
+			continue
+		}
+		if eventSinksEqual(sink, actualSink) {
```
NumaryBot, line 211: 🔴 **[blocker] Do not adopt matching externally managed sinks**

When a pre-existing external sink exactly matches the CR, this branch adds it to `status.appliedSinks`, making it indistinguishable from a sink created by the operator. If the sink is later removed from `spec.sinks`, reconciliation deletes that external sink, violating the documented ownership boundary that removals apply only to sinks the operator created (`misc/operator/api/v1alpha1/cluster_types.go:242-244`).

misc/operator/internal/controller/cluster_sinks_reconcile.go
```diff
@@ -0,0 +1,218 @@
...
+	actual, err := parseActualEventSinks(stdout)
+	if err != nil {
+		return false, err
+	}
+
+	diff := diffEventSinks(desired, actual, cluster.Status.AppliedSinks)
+
+	changed := false
+	for _, sink := range diff.toCreate {
+		if !slices.Contains(cluster.Status.AppliedSinks, sink.name) {
```
NumaryBot, line 159: 🔴 **[blocker] Do not treat a reservation as proof of sink creation**

If `add-sink` fails before committing, this persisted name still marks any subsequently created same-name sink as operator-owned. A later mismatch or spec removal can therefore delete an external sink, violating the ownership contract in `misc/operator/api/v1alpha1/cluster_types.go:242-244`. The status must distinguish pending reservations from confirmed creations.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1844#issuecomment-5482521463

misc/operator/internal/controller/cluster_sinks.go
flemzord: Fixed in af94b64e0. The reconciler now persists the name in status.appliedSinks before invoking add-sink and stops without creating anything if that status write fails. An exact matching sink without the ownership record remains a conflict and is never adopted. Added regression tests for both the persistence-failure ordering and matching external sinks.

misc/operator/internal/controller/reconcile_dnsendpoint.go
```diff
@@ -19,37 +19,70 @@ var dnsEndpointGVK = schema.GroupVersionKind{
...
+
+	return r.pruneDNSEndpoints(ctx, ledger, desired)
+}
 
-		endpoints := make([]any, 0, len(spec.Endpoints))
-		for _, ep := range spec.Endpoints {
-			endpoints = append(endpoints, buildEndpointEntry(ep))
+func desiredDNSEndpoints(ledger *ledgerv1alpha1.Cluster) []desiredDNSEndpoint {
+	if len(ledger.Spec.DNSEndpoints) == 0 {
+		legacy := ledger.Spec.DNSEndpoint //nolint:staticcheck // Reconcile the deprecated field for backwards compatibility.
```
Azorlogh, line 50: This commit (704e274) doesn't belong in a NATS-sinks PR. It re-adds `spec.dnsEndpoint`, which #1617 deliberately replaced with `dnsEndpoints`; nothing in the repo consumes it, the PR body never mentions DNS, and the commit message has no body explaining what broke. Please split it out with a stated reason — and if this is only about CRs that already exist in a cluster, updating those CRs beats carrying a compat shim in an unreleased v3.

misc/operator/internal/controller/cluster_sinks_reconcile.go
```diff
@@ -0,0 +1,218 @@
...
+		meta.SetStatusCondition(&cluster.Status.Conditions, metav1.Condition{
+			Type:               conditionSinksSynced,
+			Status:             metav1.ConditionFalse,
+			Reason:             "Error",
+			Message:            err.Error(),
+			ObservedGeneration: cluster.Generation,
+		})
+		ctrl.LoggerFrom(ctx).Error(err, "event sink reconciliation failed", "cluster", cluster.Name)
+
+		return earlierRequeue(baseResult, ctrl.Result{RequeueAfter: sinkRequeueInterval})
```
Azorlogh, line 82: A name conflict can never self-heal, but it lands here and requeues at `sinkRequeueInterval` (5s) forever — roughly 17k `ledgerctl events list` execs per day into pod-0 for one misconfigured Cluster. Worth backing off to `sinkDriftCheckInterval` when the failure is a conflict rather than a transient exec error.

misc/operator/internal/controller/cluster_sinks.go
```diff
@@ -0,0 +1,284 @@
...
+		return strings.Compare(a.name, b.name)
+	})
+
+	return desired
+}
+
+// listedEventSinksResponse mirrors only the stable, non-secret fields emitted
+// by `ledgerctl events list --json`. Other sink variants are retained as raw
+// JSON solely so name conflicts are detected and never overwritten.
+type listedEventSinksResponse struct {
```
Azorlogh, line 82: The same response also carries `sinkStatuses[].cursor` and `.error`, which this struct drops. A Ledger image built without the `nats` tag stores the config fine and then fails at `createSink` with "unsupported events sink type: nats (not compiled in this build)" — the CR still reports `SinksSynced=True` and shows nothing. I know the docs scope delivery monitoring out, but the data is already in hand here, so surfacing it is close to free.

Azorlogh (commented):
Reviewed the sinks logic closely and it holds up. I verified the parts most likely to be silently wrong:

- The `events list --json` parser matches protojson exactly — I probed `protoJSONOpts` and confirmed camelCase keys, `batchDelayMs` as a *quoted* string (hence `protoJSONInt64`), `eventTypes` as enum names, and unset oneof members omitted so `listedSinkKind` works.
- CRD `batchSize` max `100000` == `domain.MaxSinkBatchSize`, so no CRD-valid value can be rejected by the FSM.
- Reserve-ownership-before-create plus verify-after-failed-remove is idempotency-correct, and the deliberate asymmetry between the two is covered by tests.
- `ledgerctlCommand` single-quotes every arg, so CRD strings can't reach `/bin/sh`.

One unrelated commit to split out and three smaller points inline.

One more, pre-existing but load-bearing for this PR: `cmd/ledgerctl/events/add_sink.go` advertises "If a sink with the same name already exists, it is replaced (upsert)" and aliases the command `upsert`, while `processAddEventsSink` returns `ErrSinkAlreadyExists`. That wrong help text is exactly why the two-pass remove/recreate exists here — worth correcting while you're in the area.

misc/operator/e2e/tests/event-sinks-nats/chainsaw-test.yaml
```diff
@@ -0,0 +1,206 @@
...
+              set -eu
+              STREAM_INFO=$(kubectl exec nats-toolbox -n "$NAMESPACE" -- \
+                nats --server nats://nats:4222 stream info LEDGER_EVENTS --json)
+              echo "$STREAM_INFO"
+              MESSAGES=$(echo "$STREAM_INFO" | jq -r '.state.messages')
+              test -n "$MESSAGES" && test "$MESSAGES" -ge 2 || {
+                echo "FAIL: JetStream persisted only ${MESSAGES:-0} messages"
+                exit 1
+              }
+              echo "PROOF: JetStream LEDGER_EVENTS persisted $MESSAGES messages"
```
Azorlogh, line 206: The scenario covers create-and-deliver only. Removal and the two-pass update are exercised solely against the fake exec, which is where the `ErrSinkAlreadyExists` / `ErrSinkNotFound` assumptions could actually diverge from real Ledger. A final step flipping to `sinks: {}` and asserting `appliedSinks` empties plus the sink disappearing from `events list` would close that.

gfyrag:
I think this API would be more coherent as a dedicated `EventSink` CRD rather than embedding the lifecycle under `Cluster.spec.sinks`.

One Kubernetes resource per sink would provide a natural identity and lifecycle, independent status/conditions, isolated conflicts, simpler GitOps/RBAC, and a cleaner path for future transports and `Secret` references. It would also remove the subtle absent-vs-empty `spec.sinks` semantics and the overloaded `Cluster.status.appliedSinks` bookkeeping.

For example:

```yaml
kind: EventSink
metadata:
  name: primary
spec:
  clusterRef:
    name: sink-cluster
  nats:
    url: nats://nats:4222
    topic: ledger.events
  format: json
status:
  conditions: []
  cursor: 42
```

The dedicated CRD alone would not fully solve the current ownership blocker, however. Ownership should also be represented in Ledger's Raft-replicated `SinkConfig` with an opaque controller identity derived from the CR UID (or an equivalent durable token), and the reconciler should verify that identity before updating or deleting a runtime sink. A finalizer can then drive cleanup safely. This avoids treating either a Kubernetes status reservation or a matching runtime name as proof that the operator created the sink after an ambiguous/failed `add-sink` call.

Since Ledger v3 is still unreleased and has no compatibility burden between development revisions, I recommend establishing this API/lifecycle now rather than carrying the name-only ownership protocol forward.

PR #1850, opened 2026-08-31
fix: make archival residency the same on every path, and check it

Azorlogh:
Combines #1850 and #1851. Supersedes #1682.

Archival's purge and the restore's rebuild disagreed about which hot-storage
ranges an `ARCHIVED` chapter may keep, in two independent ways, and nothing
detected either. This closes both and adds the check.

## What changed

1. `WriteSet.executePurge` deletes the `SubColdAuditItem` range at
   `ConfirmArchiveChapter` apply, alongside the log, audit-entry and
   applied-proposal ranges it already purged.
2. `ApplyExportsAndRebuild` purges all four ranges once `RebuildDelta` has
   consumed them, so a restored store holds the split the source held.
3. A new checker pass, `verifyArchivedChapterResidency`, reports hot keys resident
   inside any `ARCHIVED` chapter's ranges as `UNPURGED_ARCHIVED_DATA` — one event
   per (chapter, sub-prefix) with a count.

## Why these belong together

They are the two producers of one condition, plus its detector, and each is
incomplete alone:

- **The residency pass without the restore purge** reports every restored store
  as corrupt: an export whose window overlaps an archived chapter is backfilled
  from cold storage, so the delta carries entries the source had purged, and the
  rebuild needs them. NumaryBot said so on #1851 — *"either the restore path must
  purge all four archived ranges in this change, or this check must distinguish
  that valid state"*.
- **The restore purge without the item purge** goes stale the moment the confirm
  starts deleting items: the restore would leave the one sub-prefix the source no
  longer keeps, and the pass would fire on it.
- **Either purge without the pass** leaves the class silent. Both defects existed
  for months; the audit-item one was found only by reading `executePurge`, and the
  restore one only because `store check` reported a healthy store as corrupt and
  the first attempt (#1682) bounded the checker's scan to make the report go away.

## Why each purge

**Audit items.** `AuditItem` rows hold the deterministic serialized business
intent, one per order per proposal — the bulk of the audit payload — and were
never deleted by any production path. Hot storage never shrank for that component
however much history was archived, and whole-prefix item scans grew without bound.
They are already in the chapter's cold object and in incremental backups, so the
hot copies inside archived ranges are redundant. The gap looks unintentional: the
`purgeChapterRanges` test helper, commented as mirroring the FSM purge, deletes
items, and a `checker.go` comment already claimed archived items were purged.

**The restore.** `ReadLogBySequenceWithCold` tries hot storage first, so a
restored store answered reads from history its own registry calls `ARCHIVED`, kept
a second copy of it on disk, and left transient-volume receipts the checker's
replay never derives — replay skips everything at or below the archive boundary.
Backup plus restore should be identity, and residency is part of the state.

## Risk

**MEDIUM.** The item purge is FSM-side, inside apply, so it changes what a
confirmed archival deletes on every node. Reader audit for the newly purged range:

1. `GetAuditEntry` reads the entry header first, which the existing purge already
   deleted — archived sequences return `NOT_FOUND` before items are touched.
2. `usagebuilder` / `auditindexer` iterate hot audit entries and fetch items per
   entry; entries in archived ranges are already gone.
3. The checker's and signing verifier's cold passes read items from the cold
   object; the live folds skip at or below the archive boundary.
4. `RebuildDelta` consumes re-ingested items before any purge runs.

The restore-side purge only touches ranges the rebuilt registry marks `ARCHIVED`,
and runs on a store no node is serving from yet.

## Validation

- [x] `internal/infra/backup`, `internal/application/check`, `internal/infra/state`
- [x] `go vet` with CI's full tag set over `./tests/...`
- [x] `tests/e2e/cluster` — `Restore chapter registry`, `Restore idempotency keys`
- [x] Red-checked in both directions:
      disabling the restore purge fails the e2e with four `UNPURGED_ARCHIVED_DATA`
      events (one per sub-prefix) plus the `staging:tx1` receipt; reverting the
      cursor close makes `FinalizeRestore` panic, which is how the leak was found
      in the first place.

Restore now reproduces the source's hot/cold split, not just its logical content —
recorded in the invariant section of `incremental-restore-contract.md`, which
stated the logical half only. `backup.md` said restored stores keep the backfilled
rows and the checker tolerates them; that was the contract #1682 was written
against, and it is now wrong. `checker.md` documents the new pass, and
`lifecycle.md` the widened purge.

`ApplyExports` is unchanged: the export still carries the archived ranges, and
`TestRunIncrementalBackup_BackfillsArchivedRangesFromCold` still asserts it. The
purge is the step after the rebuild consumes them.

## Review focus

Whether the item purge's reader audit above is complete — it is the one FSM-side
behaviour change here, and a missed reader means archived audit reads start
returning nothing.

Second, whether driving the restore purge off the rebuilt registry's `ARCHIVED`
status is the right authority, versus the manifest's view of what was backfilled.
Registry status is what the source used.

## Known concerns

1. The confirm also runs the deferred cleanup for ledgers whose `DeleteLedger` log
   falls inside the purged range (`State.PendingLedgerCleanups`). For a ledger
   deleted before the checkpoint and archived inside the delta, that cleanup is
   neither in the checkpoint nor replayed by the rebuild, so its data can outlive
   the restore. Not closed here: it needs FSM-side state the restore path does not
   hold.
2. Separate and pre-existing, found while writing the e2e: `store check` on a
   restored store reports `VOLUME_MISMATCH` with the replay-derived side exactly
   one transaction high, on a store that is correct. Present with and without
   everything in this PR (verified by disabling the purge), so the e2e asserts the
   two residency classes rather than `result.Errors` being empty. Being reported
   separately — it looks like baseline/boundary misalignment, the same shape as the
   bug this PR fixes.


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1850#issuecomment-5479307101

internal/infra/backup/purge_archived.go
```diff
@@ -0,0 +1,139 @@
...
+// archivedChapters reads the rebuilt registry and returns the ARCHIVED rows.
+func archivedChapters(ctx context.Context, store *dal.Store) ([]*commonpb.Chapter, error) {
+	handle, err := store.NewReadHandle()
+	if err != nil {
+		return nil, fmt.Errorf("creating read handle: %w", err)
+	}
+
+	defer func() { _ = handle.Close() }()
+
+	cursor, err := query.ReadChapters(ctx, handle)
```
NumaryBot: 🟠 **[major] Close the chapter cursor after scanning**

Every restore containing exports leaks the Pebble iterator created by `query.ReadChapters`; closing the read handle does not close iterators, and the cursor contract requires an explicit `Close`. Add a deferred cursor close after successful creation so SST and iterator resources are released.

internal/infra/backup/purge_archived.go
```diff
@@ -0,0 +1,139 @@
...
+
+	if len(archived) == 0 {
+		return nil
+	}
+
+	batch := store.OpenWriteSession()
+
+	for _, chapter := range archived {
+		if err := deleteChapterRange(batch, chapter); err != nil {
+			return fmt.Errorf("purging archived chapter %d: %w", chapter.GetId(), err)
```
NumaryBot, line 73: 🟠 **[major] Cancel the purge session on failure**

If any `DeleteRange` or the final commit fails, this write session is returned without `Cancel`, leaking its underlying Pebble batch. Defer cancellation immediately after opening the session; `Cancel` is harmless after a successful commit.

internal/infra/backup/purge_archived.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/47de46cd12d7963934823972b4af2d81cd6e0b49 — `cursor.Collect` closes the cursor (matching `recovery.go` and `checker.go`), and the write session is cancelled via defer. Both were real: `Tests-E2E-Cluster` failed on `Restore idempotency keys` because the pinned iterator made `Store.Close` panic inside `FinalizeRestore`. The composition test now asserts a real `Close` on the restored store, which the existing `_ = store.Close()` helpers were hiding.

internal/infra/backup/purge_archived.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/47de46cd12d7963934823972b4af2d81cd6e0b49 — `cursor.Collect` closes the cursor (matching `recovery.go` and `checker.go`), and the write session is cancelled via defer. Both were real: `Tests-E2E-Cluster` failed on `Restore idempotency keys` because the pinned iterator made `Store.Close` panic inside `FinalizeRestore`. The composition test now asserts a real `Close` on the restored store, which the existing `_ = store.Close()` helpers were hiding.

Azorlogh:
Addressed in https://github.com/formancehq/ledger/commit/47de46cd12d7963934823972b4af2d81cd6e0b49 — thanks, the leak finding was the important one and it was worse than a leak.

**Finding 1 (cursor leak) — fixed.** `cursor.Collect` closes it, matching `recovery.go:81-89` and `checker.go:155-163` as you pointed out. Your predicted impact was right in shape and understated in severity: it is not only `ledgerctl store bootstrap` treating the close error as fatal. `Tests-E2E-Cluster` failed on this PR's first push — `Restore idempotency keys / downloads and finalizes the backup` — because the pinned iterator makes Pebble panic in `Store.Close` (`element has outstanding references`), which `closeStagingStore` hits inside `FinalizeRestore`. So **every** restore through the service failed at the last step. The composition test now asserts a real `Close` on the restored store; as you noted, `_ = store.Close()` in the helpers hides this whole class.

**Finding 2 (backup.md contradiction) — fixed.** That paragraph described the contract this PR changes, and it is worth noting it is also where the previous approach came from: #1682 bounded the checker's scan because the doc said retained rows were tolerated.

**Finding 3 (no checker/backfill coverage) — fixed at the e2e level**, which is stronger than the unit fixture you suggested. `tests/e2e/cluster/restore_chapter_registry_test.go` now drives a transient account to zero inside the chapter it archives, in a ledger of its own (declaring an account type makes the whole ledger typed), then asserts `store check` reports no `EXCLUSION_RECORD_MISMATCH` on the restored store, with a guard against the vacuous pass `Check()` gives when an archived store has no baseline. Verified both directions: the receipt for `staging:tx1` appears with the purge disabled and is gone with it.

That run also surfaced something separate and pre-existing: four `VOLUME_MISMATCH` errors on the same restored store, replay-derived side exactly one transaction high, present with and without this purge. Not asserted here so the spec is not coupled to it; being reported separately.

**Finding 4 (deferred ledger cleanup) — confirmed as scoped out**, documented in the godoc, the commit body and the PR's Known concerns. Ticket to follow.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1850#issuecomment-5479307101

internal/infra/backup/purge_archived.go
```diff
@@ -0,0 +1,144 @@
...
+// confirm produced on the source. Audit items are outside it there and stay
+// outside it here.
+//
+// Not covered: the confirm also runs the deferred cleanup for ledgers whose
+// DeleteLedger log falls inside the purged range (State.PendingLedgerCleanups).
+// For a ledger deleted before the checkpoint and archived inside the delta, that
+// cleanup is neither in the checkpoint nor replayed by the rebuild, so its data
+// can outlive the restore. Tracked separately; it needs FSM-side state this path
+// does not hold.
+func purgeArchivedRanges(ctx context.Context, logger logging.Logger, store *dal.Store) error {
```
flemzord, line 46: `codecov/patch` is currently failing at 58.49% against the repository's 80% target, with most uncovered lines in this new file. Please add targeted coverage for the no-archived-chapter path, filtering non-archived rows, the inverted audit-range branch, and the read/delete/commit error paths. These are meaningful restore failure semantics, not only a numeric coverage gap.

internal/infra/backup/restore.go
```diff
@@ -195,5 +195,11 @@ func ApplyExportsAndRebuild(ctx context.Context, logger logging.Logger, storage
 		return fmt.Errorf("rebuilding derived state: %w", err)
 	}
 
+	// The rebuild is the last consumer of the archived ranges the export carried;
+	// past it they belong in cold storage alone, as they did on the source.
+	if err := purgeArchivedRanges(ctx, logger, store); err != nil {
```
flemzord, line 200: Please also cover propagation of a `purgeArchivedRanges` failure from `ApplyExportsAndRebuild`; Codecov reports this new call site as uncovered/partial, and callers need the restore to fail rather than silently finalize after an incomplete purge.

flemzord (commented):
The implementation review is clean, but the required patch-coverage check is still failing.

gfyrag (dismissed):
Independent review at exact head `a768eafea2a07ab6fdd2a82a45eb52f7d5dc1780`.

Reviewed the expanded #1850/#1851 scope. `ConfirmArchiveChapter`, post-rebuild restore cleanup, and `verifyArchivedChapterResidency` now agree on all four canonical cold ranges: logs by log sequence and audit entries/items/applied proposals by audit sequence. Audit-item suffix bounds are correct, empty audit ranges are guarded, the checker runs before the zero-log fast path, and the authoritative docs accurately state both the residency contract and the chapter-row trust boundary. Proto numbering/generated output are aligned and all functional CI jobs are green.

Non-blocking known concerns remain correctly disclosed: deferred deleted-ledger cleanup parity on restore and the separate restored-store `VOLUME_MISMATCH` report. The patch-coverage status is still red but does not expose an uncovered correctness branch in the reviewed mechanism.

internal/infra/backup/restore.go
Azorlogh: Covered in https://github.com/formancehq/ledger/commit/8d0f225a06793fa2555e9619756adb9cc0c7ef50 — `TestApplyExportsAndRebuild_PropagatesAPurgeFailure` plants a chapter row the registry scan cannot decode (the rebuild writes rows for the chapters its logs describe and never touches that id, so the purge's scan is the first step to reach it) and asserts `ApplyExportsAndRebuild` returns `purging archived ranges`.

Red-checked by swallowing the error at the call site, so it pins the propagation rather than just the inner failure. You were right that it was untested — the propagation itself was there, nothing proved it.

internal/infra/backup/purge_archived.go
Azorlogh: Covered in https://github.com/formancehq/ledger/commit/8d0f225a06793fa2555e9619756adb9cc0c7ef50, taking your framing that these are failure semantics — each one asserts behaviour, not a line.

- **Inverted audit range** — `TestPurgeArchivedRanges_SkipsAnInvertedAuditRange`. A chapter that closed with no audit entries of its own carries `close_audit_sequence < start_audit_sequence`; the confirm skips the audit-keyed ranges for it, and so does this. The test pins that the successor's entries at the start sequence survive.
- **Read error path** — `TestPurgeArchivedRanges_PropagatesAReadFailure`.
- **Delete error paths** — all four ranges, via a one-method `rangeDeleter` interface. Nothing makes `DeleteRange` fail on a healthy Pebble batch, so the seam is the only way to reach those returns; it follows the `ledgerLogWithPurgedVolumes` precedent in `indexbuilder`. Say the word if you would rather not have the seam and accept the lines uncovered.
- **No-archived-chapter path and non-archived filtering** were already covered (`..._NoArchivedChaptersIsANoOp`, `..._LeavesUnarchivedChaptersAlone`); the first exists so an empty registry cannot become a range delete over everything.

`deleteChapterRange` is now 100%. Still uncovered, so you have the full picture: `batch.Commit`'s error return and the loop's propagation of it — reaching those needs the write session injected rather than a range deleter, which is a wider seam than I wanted to add unasked. Happy to do it if you want the file clean.

Azorlogh: Finished in https://github.com/formancehq/ledger/commit/8d122f478178f51aa19bea07974186ff40dc64a5 — I was wrong to leave the commit path out. `purgeChapters` now takes the session as an interface, so the commit failure and the per-chapter propagation are covered too; a failed delete names its chapter.

`purgeArchivedRanges`, `purgeChapters` and `deleteChapterRange` are at 100%. Only `query.ReadChapters`' error return is left: it needs the reader injected as well and cannot fail on a live handle.

gfyrag (dismissed):
Independent re-review at exact head `8d122f478178f51aa19bea07974186ff40dc64a5`.

The two new commits make every delete-range and commit failure observable in tests without changing the purge contract. The extracted `purgeChapters` still commits all archived chapters atomically, while `purgeArchivedRanges` retains ownership of session cancellation. The restore integration, inverted audit range, read failure, per-range failure, per-chapter attribution, and commit failure paths are now pinned.

No prior finding remains and no new finding was discovered. Targeted `internal/infra/backup` tests pass; current functional CI is green.

DECISION: APPROVE
Head reviewed: 8d122f478178f51aa19bea07974186ff40dc64a5
Blocking findings: 0
Non-blocking findings: 0
Validation considered: targeted backup tests + current CI
Residual risk: MEDIUM

gfyrag (dismissed):
Independent re-review at exact head 67ee2a119435adea33588f2263d4916b2894481d.\n\nThe merge from release/v3.0 resolves the CheckStoreErrorType collision by keeping the base cluster-policy values at 24/25 and moving UNPURGED_ARCHIVED_DATA sequentially to 26; generated protobuf output is aligned. The follow-up documentation correctly points to the proto as the enum source of truth. No prior finding remains and no new finding was found. Targeted backup, checker, and state tests pass, and current CI is green.\n\nDECISION: APPROVE\nBlocking findings: 0\nNon-blocking findings: 0\nResidual risk: MEDIUM

gfyrag (approved):
Independent re-review at exact head 07c130c4b40452a97d122651ec58929c0e4505ca.

The force-rebase preserves the reviewed feature patches. The only semantic rebase adjustment is the sequential UNPURGED_ARCHIVED_DATA enum value after the current base additions; generated output is aligned. The stray probe file was removed. No prior finding remains and no new finding was discovered. Targeted backup, checker, and state tests pass, and current CI is green.

DECISION: APPROVE
Head reviewed: 07c130c4b40452a97d122651ec58929c0e4505ca
Blocking findings: 0
Non-blocking findings: 0
Validation considered: targeted backup/check/state tests + current CI
Residual risk: MEDIUM

PR #1851, opened 2026-08-31
fix: purge audit items at archival confirm and report unpurged archived residue

Azorlogh:
## What changed

`WriteSet.executePurge` now deletes the `SubColdAuditItem` range at `ConfirmArchiveChapter` apply, alongside the log, audit-entry and applied-proposal ranges it already purged. A new checker pass, `verifyArchivedChapterResidency`, reports hot-store keys resident inside any ARCHIVED chapter's purge ranges (all four cold sub-prefixes) as `CHECK_STORE_ERROR_TYPE_UNPURGED_ARCHIVED_DATA`.

## Why

Archival purged three of the four cold sub-prefixes. `AuditItem` rows — the deterministic serialized business intent, one per order per proposal, the bulk of the audit payload — were never deleted by any production path: hot storage never shrank for that component however much history was archived, and whole-prefix item scans (`checker.go` boundary-effects pass) grew without bound on fully archived stores.

The items are already exported to the chapter's cold object (`IterateColdKVPairs` scans them with the audit range) and backfilled into incremental backups, so the hot copies inside archived ranges are redundant. The gap looks unintentional: the `purgeChapterRanges` test helper — commented as mirroring the FSM purge — deletes items, and a `checker.go` comment already stated "archived audit items are purged".

Reader audit for the newly purged range:

1. `GetAuditEntry` reads the entry header first, which the existing purge already deletes — archived sequences return `NOT_FOUND` before items are touched.
2. `usagebuilder` / `auditindexer` iterate hot audit entries and fetch items per entry; entries in archived ranges are already gone.
3. The checker's and signing verifier's cold passes read items from the cold object; the live folds skip at/below the archive boundary.
4. Restore (`RebuildDelta`) consumes re-ingested items before any purge runs — unaffected by an FSM-side confirm purge.

Need: hot-store disk usage must be reclaimable by archival; `store check` cost must not grow without bound on archived history.
Current limitation: the largest cold sub-prefix was exempt from the purge, permanently.
Requirement / constraint: after a confirmed archival, hot storage holds no keys in the chapter's ranges — now enforced by the checker.
Durable repository evidence: `docs/technical/architecture/subsystems/checker/checker.md` (pass 15), `docs/technical/architecture/subsystems/chapters/lifecycle.md` step 2, `TestWriteSetMergePurgesConfirmedChapterRanges`.

## Technical decision

Decision: one `DeleteRange` in `executePurge` over `[start_audit_sequence, close_audit_sequence + 1)` — the item key's 4-byte order-index suffix is covered by the sequence-prefix bounds. Checker side: a residency pass per ARCHIVED chapter over the exact ranges the purge deletes, one event per (chapter, sub-prefix) with a count, new enum value per the checker contract (extend, never reuse).
Why now / why proportionate: the purge fix is one range delete; the pass makes this class (lost purge, unpurged re-ingest) visible instead of silent.
Alternatives considered: items-only checker rule (leaves purge regressions of the other three sub-prefixes undetected); a global archive-boundary rule instead of per-chapter ranges (false-positives on legitimately retained out-of-order ranges).

## Risk

**LOW** — the purge is deterministic (bounds come from the confirmed chapter payload, `DeleteRange` is idempotent on replay) and rides the same apply batch as the status transition. v3 is unreleased with stores wiped per push, so there is no mixed-version or migration exposure.

## Validation

- [x] `bash scripts/agent-check` — PASS
- [x] Red-on-old: `TestWriteSetMergePurgesConfirmedChapterRanges` fails on `release/v3.0` ("item at start, order 0" survives Merge), green with the fix
- [x] Targeted tests: `internal/infra/state`, `internal/application/check`, `internal/infra/backup` package suites
- [x] Proto regenerated with the pinned toolchain (`nix develop --command just generate-proto`); only the new enum value in the diff

FSM apply now deletes one more range per confirmed chapter (persisted-state purge scope widened). New checker error type `UNPURGED_ARCHIVED_DATA = 24`. `store check` on a store that re-ingested archived history without purging (today: any incremental restore overlapping an archived chapter) reports this finding — such stores already fail the exclusion pass today, which #1850 fixes.

## Review focus

1. The `DeleteRange` bounds in `executePurge`: the suffix-free `close + 1` upper bound must cover every order-index suffix of every sequence in `[start, close]`.
2. The residency pass scoping: per-ARCHIVED-chapter ranges (not a global boundary), and the `close < start` empty-range skip mirroring the purge's guard.

## Known concerns

#1850 (restore purges archived ranges) deliberately mirrors the pre-fix purge scope: it keeps audit items on the restore path and its test asserts they remain. Once this merges, #1850 needs `SubColdAuditItem` added to its purge loop and those assertions flipped — otherwise restored stores keep failing `store check` on this finding after #1850 lands. Same-repo follow-up, one line plus test updates.


internal/application/check/checker.go
```diff
@@ -235,6 +235,10 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 		}
 	}
 
+	if err := verifyArchivedChapterResidency(snap, chapters, callback); err != nil {
```
NumaryBot: 🟠 **[major] Run archived residency check before the empty-store return**

When `ReadLastSequence` returns zero, `Check` returns from the earlier fast path before this invocation. Consequently, a store containing archived chapter rows and residual audit-item, audit-entry, or applied-proposal keys—but no live log keys—will incorrectly pass this new residency check, despite the documented requirement that all four archived ranges be empty. Move this pass before the `lastSequence == 0` return.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1851#issuecomment-5480520559

internal/application/check/checker.go
Azorlogh: Fixed in 84901ba4fc2ade5691aba750eceffd15036fab91 — the pass now runs right after the chapter collection, before the zero-log fast path, with a full-Check regression test (`TestVerifyArchivedChapterResidency_RunsOnZeroLogStore`).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1851#issuecomment-5480520559

internal/application/check/checker.go
```diff
@@ -162,6 +162,13 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
 		return fmt.Errorf("collecting chapters: %w", err)
 	}
 
+	// Runs before the zero-log fast path: the pass needs only the registry
+	// rows and the snapshot, and a store with no logs can still hold ARCHIVED
+	// chapter rows with residue in their purge ranges.
+	if err := verifyArchivedChapterResidency(snap, chapters, callback); err != nil {
```
NumaryBot, line 168: 🔴 **[blocker] Do not flag valid incremental restores as unpurged**

When an incremental restore overlaps an archived chapter, `RebuildDelta` legitimately re-ingests the cold-backfilled audit-item rows and currently performs no corresponding purge, so this unconditional pass reports a successfully restored store as corrupt. Either the restore path must purge all four archived ranges in this change or this check must distinguish that valid state; otherwise `store check` violates the [incremental restore parity contract](docs/technical/architecture/subsystems/chapters/incremental-restore-contract.md).

Azorlogh:
Combined into #1850 (retitled "fix: make archival residency the same on every path, and check it"), which now carries all three commits from this branch plus the restore-side purge. Closing here so there is one PR.

The reason for combining is NumaryBot's blocker on this PR: the residency pass reports every restored store as corrupt, because an export overlapping an archived chapter is backfilled from cold and `RebuildDelta` needs those entries. Its prescription was "either the restore path must purge all four archived ranges in this change, or this check must distinguish that valid state" — #1850 does the former, so the two changes resolve each other. The coupling runs both ways: without the item purge from this branch, the restore purge would leave the one sub-prefix the source no longer keeps, and this pass would fire on it.

Verified in the combined branch: with the restore purge disabled, the restore e2e fails with four `UNPURGED_ARCHIVED_DATA` events — one per sub-prefix — plus the transient volume's `EXCLUSION_RECORD_MISMATCH`. With both purges, clean.

PR #1852, opened 2026-09-01
fix(wal): preserve the running leader's retained terms

gfyrag:
## Summary

- decouple the running Raft log's compacted boundary from the latest snapshot index
- keep `FirstIndex`, `Entries`, and `Term` consistent with the canonical `raft.Storage` contract
- preserve the configured catch-up margin while the leader process is running
- intentionally reset that margin to the latest durable snapshot on restart, matching etcd
- document both lifecycle semantics and add focused regressions

## Root cause

`CreateSnapshot` advanced `snapshot.Metadata.Index`, while `termLocked` used that index as the compaction boundary. Entries retained by `Compact(capturedIndex - CompactionMargin)` remained physically cached and visible through `FirstIndex`/`Entries`, but `Term` returned `raft.ErrCompacted`.

Raft could therefore not match a retained previous entry and unnecessarily fell back to a full snapshot.

This was reproduced on `ohbajrwzuthm-rfbp`: after the new leader had restarted and accumulated entries `374099–374102`, maintenance published snapshot `374102`. Node 0 was only three entries behind, yet received `MsgSnap` and entered a 234.5 GiB checkpoint sync.

## Lifecycle contract

The compaction margin is process-local, as in etcd:

- while the process is running, later snapshots do not invalidate retained terms or entries;
- on restart, WAL replay begins after the latest durable full snapshot, which becomes the new compacted boundary;
- no sentinel `walpb.Snapshot`, compatibility format, or separate persisted compaction marker is introduced.

## Validation

- regression proving a follower inside the live retained window receives `MsgApp`, not `MsgSnap`
- restart regression proving the latest durable snapshot becomes the new boundary
- `go test -race ./internal/storage/wal ./internal/infra/node`
- `bash scripts/agent-check`
- synchronized with the current `release/v3.0` before review-loop execution

Jira: [EN-1925](https://formance-team.atlassian.net/browse/EN-1925)

Related but distinct: EN-1552 covers count-based WAL sizing and a different follower-sync retry storm.

[EN-1925]: https://formance-team.atlassian.net/browse/EN-1925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/storage/wal/wal_default.go
```diff
@@ -453,6 +462,12 @@ func New(dataDir string, logger logging.Logger, meter metric.Meter, opts ...Opti
 	if s.hardState == nil {
 		s.hardState = &raftpb.HardState{}
 	}
+	// wal.Open starts replay immediately after the selected snapshot, so that
```
NumaryBot: 🔴 **[blocker] Preserve the retained boundary across WAL restarts**

After any leader restart following a newer snapshot, this resets the compaction boundary to that snapshot and `wal.Open` has already omitted all earlier retained entries. The configured margin therefore disappears on restart, so a follower within the intended retained window again receives `MsgSnap` and triggers a full checkpoint sync. Persist or reconstruct the actual compaction boundary and replay from it rather than treating the latest snapshot as compacted.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1852#issuecomment-5491637961

internal/storage/wal/wal_default.go
```diff
@@ -872,24 +891,26 @@ func (s *DefaultWAL) termLocked(i uint64) (uint64, error) {
...
 
-	if i == firstIndex-1 {
-		return s.snapshot.GetMetadata().GetTerm(), nil
+	if i == s.compactedIndex {
+		return s.compactedTerm, nil
 	}
 
-	if len(s.entries) == 0 {
-		return 0, raft.ErrUnavailable
+	if i == s.snapshot.GetMetadata().GetIndex() {
```
Azorlogh: Redundant arm: it always shadows an identical entries lookup, and a snapshot cannot land in a leading gap either, since `CreateSnapshot` has to resolve its term through `termLocked` first. Replacing it with an equivalence assertion (panic if entries are empty, if `i` falls outside the window, or if the two terms differ) keeps `./internal/storage/wal` and `./internal/infra/node` green. Can be dropped.

internal/storage/wal/wal_default.go
```diff
@@ -604,7 +619,10 @@ func (s *DefaultWAL) firstIndexLocked() uint64 {
 		return s.snapshot.GetMetadata().GetIndex() + 1
```
Azorlogh: Nit: this branch ignores `compactedIndex` while the one below uses it. Equivalent today — whenever `entries` is empty the two are always equal — but `max(snapshot.Index, compactedIndex) + 1` would make that local instead of implicit.

internal/storage/wal/wal_default.go
```diff
@@ -51,6 +51,15 @@ type DefaultWAL struct {
 	// Snapshot stores the most recent snapshot
 	snapshot *raftpb.Snapshot
 
+	// compactedIndex is the last entry incorporated into the compacted log
+	// prefix. Its term remains available through compactedTerm so raft can
+	// match an AppendEntries request at FirstIndex()-1. This boundary is
+	// deliberately independent from snapshot.Metadata.Index: a newer snapshot
```
Azorlogh: "deliberately independent" overstates it: the boundary can only lag, since `Compact` rejects `compactIndex > snapshot.Index`.

Azorlogh (commented):
Reviewed at `677e95f7` against merge base `46b21f68b`.

**The fix is real and the regression test earns its place.** With the PR's tests applied to the merge base, `TestRaftFollowerWithinRetainedMarginReceivesAppend` fails with message type 7 (`MsgSnap`). Root cause matches the description: `raftLog.term(pr.Next-1)` hit `termLocked`'s `firstIndex = snapshot.Index + 1`, so `Entries` served the retained window while `Term` refused to match it.

### Confirming NumaryBot's blocker, with numbers

300 entries, `CreateSnapshot(300)`, `Compact(44)`: in-process `FirstIndex()==45` and `Term(100)==1`; after close/reopen `FirstIndex()==301`, `Term(100)==ErrCompacted`, and `compactedIndex` is reset to 300. The whole 256-entry margin collapses, so the reported incident recurs for the window after a leader restart.

I agree with the shipfox review that this is not a regression — but `storage.md` point 3 ("Advancing `Snapshot().Metadata.Index` must not make retained entries or their terms unavailable") states it as an absolute invariant, and after a restart that is false. At minimum the doc needs qualifying as process-local. If you want to actually close the gap: `wal.Open` from the newest WAL snapshot *record* at or below the compaction target rather than from the newest snap file, and seed `compactedIndex` from it — the records survive `CleanupOlderThan`, which only deletes snap files.

### One correction to the shipfox review

The boot gate in `node.go` (`applied+1 < walFirstIdx`) is not "unaffected". `FirstIndex()` moving from `C` to `C+1` tightens it by exactly one index. The conclusion holds — the tightening is correct and unreachable in practice — but for a different reason than stated: with the dummy entry hidden, `applied == C-1` would panic raft, because `newLogWithSize` initialises `applied = firstIndex - 1` and `newRaft`'s `appliedTo(c.Applied)` panics when `i < l.applied`.

Three small inline notes otherwise. Nothing blocking from me beyond the restart gap.

internal/storage/wal/wal_default.go
```diff
@@ -453,6 +469,14 @@ func New(dataDir string, logger logging.Logger, meter metric.Meter, opts ...Opti
 	if s.hardState == nil {
 		s.hardState = &raftpb.HardState{}
 	}
+	if created {
+		// Persist the initial replay cursor before returning the fresh WAL. Every
+		// later open requires an explicit compaction marker instead of guessing
+		// the logical boundary from the newest FSM snapshot.
+		if err := s.wal.SaveSnapshot(newCompactionMarker(0, 0)); err != nil {
```
NumaryBot: 🔴 **[blocker] Persist the initial marker before declaring creation complete**

If the process crashes, or `SaveSnapshot` fails, after the creation-completion file is written but before this initial marker is persisted, the next `New` treats the WAL as existing and fails with `existing WAL has no persisted compaction marker`. This leaves a newly created, otherwise empty WAL permanently unable to start without manual cleanup; persist the marker before publishing the completion marker.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1852#issuecomment-5491637961

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1852#issuecomment-5491637961

Azorlogh (approved):
Approving at `511a9bf92` (reviewed at `1b1786e12`; only change since is a clean upstream merge).

All three of my inline comments are applied. `TestTerm_GapBeforeCachedEntriesIsCompacted` is a good catch — that sparse-window arm is genuinely live.

Two non-blocking notes, kept in the body so they don't hold the merge on the unresolved-threads rule:

1. `storage.md:407` and the `newWALWithRetainedMargin` comment (`wal_default_test.go:93`) still call a below-boundary `Compact` "a no-op"; it returns `raft.ErrCompacted`.
2. Three commits are unrelated to EN-1925: `48ab651f6`, `ff799dfd4`, and `1b1786e12` — the last sets `withHealthDiskThresholds(1, 0.99)` on every default e2e node, disabling the disk-pressure gate outside the dedicated health tests. Easier to find later as its own PR.

PR #1854, opened 2026-09-01
fix(check): bind the checker baseline to the archived boundary

Azorlogh:
Independent of #1850 (verified: the repro fires identically on clean
`release/v3.0`), found while writing its restore e2e.

## What changed

The checker's baseline snapshot is staged per closing chapter and promoted to
live only when that chapter's archival confirm moves the archived boundary onto
it. `Check()` refuses a baseline whose chapter does not match the boundary it is
verifying.

## Why

`store check` seeds expected volumes with the baseline and replays everything
above the archived boundary: `expected = baseline + replay(boundary..now)`. That
is only sound when the baseline is the state at exactly the boundary — and it was
overwritten at **every** chapter close with the state at that close.

So on a store with a closed-but-unarchived chapter — the lifecycle's normal
intermediate state between any close and its archival — every transaction between
the boundary and the newest close sat in the baseline *and* in the replay window,
and was counted twice. Result: a healthy store reported corrupt, one
`VOLUME_MISMATCH` per touched account, replay side exactly one transaction high.
Repro is three requests on a live single node:

```
tx → archive chapter 1 → tx → close chapter 2 → store check
  VOLUME_MISMATCH | input mismatch for acc:post/USD: expected 20, got 10
  VOLUME_MISMATCH | output mismatch for world/USD:   expected 30, got 20
```

Archive chapter 2 and the same check is clean; close chapter 3 over a new
transaction and it fires again. Whether a given run reports a healthy store as
corrupt is purely whether it lands in the aligned window — which is why the model
runs' teardown checks pass or fail by pacing luck, and why the 210
`VOLUME_MISMATCH` errors in the run that surfaced EN-1888 began at exactly the
first closed-unarchived chapter's start sequence: they were this bug, not that
one.

## Technical decision

- **Stage at close, promote at confirm.** The state at a chapter's close can only
  be captured at its close (the applier stages it inside the seal-checkpoint
  maintenance gate, as before); whether it is the boundary is only known at its
  confirm. The FSM's post-commit hook promotes it (`PromoteStagedBaseline`),
  next to the query-checkpoint file deletes — the established home for
  node-local file lifecycles bound to applied orders, wired through the same
  capability-interface pattern (`dal.FSMFileArtifacts`) so the Machine still
  holds no raw `*dal.Store`.
- **Boot reconcile** covers the crash window between the confirm's commit and the
  rename; replayed confirms promote through the hook itself.
- **Refuse, never serve stale.** A boundary whose own staging failed (staging
  stays non-fatal) cannot be caught up by any older snapshot, so promotion
  removes the stale live baseline; `Check()` independently refuses a baseline
  whose chapter id differs from the snapshot's boundary (promotion racing the
  check). Both degrade to the existing skip of entry-by-entry verification —
  the checker already treats a missing baseline that way.
- **Alternatives considered:** recomputing the boundary state at confirm time
  (impossible from the live store — later transactions are already folded in);
  having the checker pick among per-close baselines at check time (keeps every
  close's snapshot forever, and still needs the staleness rules).

## Risk

**MEDIUM.** The promotion is a new FSM post-commit step — node-local file
bookkeeping like the query-checkpoint deletes beside it, nothing in the applied
row, non-fatal on error. Coverage narrows in one honest way: between a confirm
whose close failed to stage and the next full cycle, the checker skips
entry-by-entry verification instead of comparing against wrong state.

## Validation

- [x] `bash scripts/agent-check`; `go vet` with CI's full tag set over `./tests/...`
- [x] `internal/storage/dal`, `internal/infra/state`, `internal/infra/node`,
      `internal/application/check`
- [x] `TestPromoteStagedBaseline_*` — the promote rule: boundary close promoted,
      multi-confirm batches, idempotency, lost-promotion recovery, and the
      stale-removal when the boundary's staging failed
- [x] `CheckStoreBaselineBoundary` e2e — a healthy store verifies cleanly in all
      three alignments (closed chapter above the boundary, boundary on the newest
      close, and again above). It asserts the full pass **ran**, not merely that
      no errors came back: the id refusal turns the old misalignment into a
      silent skip, so a zero-error assertion cannot tell the fix from the guard.
      Red-checked by restoring promote-at-close — fails on the degrade guard.

Baseline layout changes from a single `checker` directory to
`staged-<chapterID>` / `checker-<chapterID>` (v3 unreleased, no migration).
`BaselineCheckpointPath` now returns the chapter id; `BaselineSnapshotDir` is
replaced by `StagedBaselineDir(chapterID)`. Documented in `checker.md` ("The
baseline snapshot and the archived boundary").

## Review focus

The promote rule's refusal cases in `PromoteStagedBaseline` — the crash
interleavings are enumerated in its tests, and the dangerous one (promoting an
older staged snapshot when the boundary's own staging failed) is refused rather
than approximated. Second opinion welcome on degrading instead: it trades checker
coverage for never comparing against wrong state.

## Known concerns

A restored store has no staged snapshots (they are not part of the backup), so
the checker degrades there until the first close-and-confirm cycle after the
restore. Before this change it compared against a wrong baseline instead; #1850's
restore e2e keeps its own scoped assertions until both PRs land, after which its
store-check assertion can be widened to `result.Errors` empty.

### DIFF
--- a/internal/infra/node/applier.go
+++ b/internal/infra/node/applier.go
@@ -589,6 +589,17 @@ func (a *Applier) RecoverAndReplay(ctx context.Context) (bool, error) {
 		}
 	}
 
+	// Reconcile the checker baseline with the archived prefix. Promotion
+	// normally rides the confirm's post-commit hook, but a crash between that
+	// commit and the rename loses it, and nothing re-applies the confirm on
+	// restart. Idempotent; replayed confirms below promote through the same
+	// hook. Non-fatal: the checker refuses a baseline whose chapter does not
+	// match the boundary, so a missed promotion degrades rather than misleads.
+	if err := a.store.PromoteStagedBaseline(a.fsm.Chapters.ArchivedThroughID()); err != nil {
+		a.logger.WithFields(map[string]any{"error": err}).
+			Errorf("Failed to reconcile staged checker baselines at boot (checker will degrade gracefully)")
+	}
+
 	storeLastAppliedIndex, err := query.ReadLastAppliedIndex(a.store)
 	if err != nil {
 		return false, fmt.Errorf("getting store last applied index: %w", err)
@@ -1683,8 +1694,8 @@ func (a *Applier) handleCheckpointRequired(
 				applyResult.OnCheckpointDone(path)
 			}
 
-			// Create compact baseline snapshot for the checker (non-fatal on error).
-			if err := a.createBaselineSnapshot(); err != nil {
+			// Stage the checker baseline for this close (non-fatal on error).
+			if err := a.createBaselineSnapshot(applyResult.CheckpointChapterID); err != nil {
 				a.logger.WithFields(map[string]any{"error": err}).
 					Errorf("Failed to create baseline snapshot (checker will degrade gracefully)")
 			}
@@ -2062,9 +2073,9 @@ func (a *Applier) createReplayCheckpoint(result *state.ApplyEntriesResult) error
 		result.OnCheckpointDone(checkpointPath)
 	}
 
-	if err := a.createBaselineSnapshot(); err != nil {
+	if err := a.createBaselineSnapshot(result.CheckpointChapterID); err != nil {
 		a.logger.WithFields(map[string]any{"error": err}).
-			Errorf("Failed to create baseline snapshot during replay (checker will degrade gracefully)")
+			Errorf("Failed to stage baseline snapshot during replay (checker will degrade gracefully)")
 	}
 
 	if lastResult, deferred := a.extractDeferredFuture(result); deferred != nil {
@@ -2092,11 +2103,19 @@ func (a *Applier) createMainStoreCheckpoint(checkpointID uint64) error {
 	return nil
 }
 
-// createBaselineSnapshot creates a compact attribute-only snapshot for the checker.
-// Unlike a full Pebble checkpoint, this contains only computed attribute values
-// (volumes, metadata, transactions), making it orders of magnitude smaller.
-func (a *Applier) createBaselineSnapshot() error {
-	destPath, err := a.store.BaselineSnapshotDir()
+// createBaselineSnapshot stages a compact attribute-only snapshot for the
+// checker. Unlike a full Pebble checkpoint, this contains only computed
+// attribute values (volumes, metadata, transactions), making it orders of
+// magnitude smaller.
+//
+// It runs inside the seal-checkpoint maintenance gate, so what it captures is
+// the state at exactly this chapter's close. It stays staged until the
+// chapter's archival is confirmed: the checker replays everything above the
+// archived boundary and seeds the sums with the baseline, so making this live
+// at close time double-counts every transaction between the boundary and the
+// close — the FSM promotes it on the confirm instead (PromoteStagedBaseline).
+func (a *Applier) createBaselineSnapshot(chapterID uint64) error {
+	destPath, err := a.store.StagedBaselineDir(chapterID)
 	if err != nil {
 		return err
 	}
--- a/internal/infra/node/applier.go
+++ b/internal/infra/node/applier.go
@@ -586,6 +586,19 @@ func (a *Applier) RecoverAndReplay(ctx context.Context) (bool, error) {
 
 			// The checkpoint is now on disk. The Sealer's recoverPendingSeal
 			// and periodic reconciliation will pick it up once started.
+
+			// A missing seal checkpoint is the proof that nothing after the close
+			// was applied — Pebble is at the close boundary, which is what makes
+			// the checkpoint above valid. The same proof makes staging the
+			// checker's baseline valid here, so the crash window between the
+			// close's commit and its staging closes too. When the checkpoint
+			// already exists that proof is gone, and a chapter whose staging
+			// failed stays unstaged: its promotion degrades rather than serves
+			// state captured at the wrong boundary.
+			if err := a.createBaselineSnapshot(chapter.GetId()); err != nil {
+				a.logger.WithFields(map[string]any{"error": err}).
+					Errorf("Failed to stage recovery baseline snapshot (checker will degrade gracefully)")
+			}
 		}
 	}
 
--- a/internal/infra/state/machine.go
+++ b/internal/infra/state/machine.go
@@ -51,11 +51,12 @@ type Machine struct {
 	// WriteSessionFactory as a parameter to PrepareEntries / ApplyEntries, not
 	// as a field.
 
-	// queryCheckpoints lets the FSM delete query-checkpoint files when an
-	// apply removes their metadata. Two methods only: create + delete. The
-	// surface is not a read capability and does not give access to Pebble
-	// contents.
-	queryCheckpoints dal.QueryCheckpoints
+	// fileArtifacts lets the FSM drive the node-local file lifecycles bound to
+	// applied orders: deleting query-checkpoint files when an order removes
+	// their metadata, and promoting the checker's staged baseline when an
+	// archival confirm advances the archived prefix. The surface is not a read
+	// capability and does not give access to Pebble contents.
+	fileArtifacts dal.FSMFileArtifacts
 
 	// sentinel runs scoped post-commit reader-based checks in debug mode.
 	// The Reader never escapes the callback, so even in sentinelMode the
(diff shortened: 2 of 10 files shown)


Azorlogh:
Fixed in https://github.com/formancehq/ledger/commit/246c13e78de28269736635dbda65e828721cdea4 — the temp-sibling finding was real and exactly as described; I verified the `Sscanf` claim first (`staged-…2.tmp-1234-5678` parses as `n=1, id=2`).

- **Anchored parse** (`parseBaselineName`: `strings.CutPrefix` + `strconv.ParseUint`, no remainder) in both `PromoteStagedBaseline` and `BaselineCheckpointPath`, and promotion sweeps `.tmp-*` leftovers. Red-checked: restoring the Sscanf parse fails `TestPromoteStagedBaseline_IgnoresAndSweepsInterruptedWrites`; the shadowing case and crash-then-retry case (clean snapshot beside its own leftover temp) are covered too.
- **Close-recovery staging**: taken, with the scope your Codex point implied — the baseline is staged only in the branch where the seal checkpoint was missing, because that absence is the proof Pebble sits at the close boundary; when the checkpoint exists that proof is gone, and an unstaged chapter keeps degrading rather than serving state captured at the wrong boundary. The commit-crash-restart-archive composition is covered at that level rather than end-to-end.
- The stale `// QueryCheckpoints` comment is fixed as well — cosmetic, but it was mine and one word.

internal/infra/state/machine.go
```diff
@@ -897,6 +898,17 @@ func (fsm *Machine) CommitPreparedBatch(ctx context.Context, pb *PreparedBatch)
...
 			// Coalescent signal — safe to drop, next purge will re-signal.
 		}
+
+		// A purge means an archival confirm advanced the archived prefix, so the
+		// staged baseline of the newly archived close becomes the checker's live
+		// one. Node-local file bookkeeping like the query-checkpoint deletes
+		// above; non-fatal, because the checker degrades honestly without a
+		// baseline while a stale one makes it report a healthy store as corrupt —
+		// which is also why promotion happens here and not when a chapter closes.
+		if err := fsm.fileArtifacts.PromoteStagedBaseline(fsm.Chapters.ArchivedThroughID()); err != nil {
```
NumaryBot: 🟠 **[major] Promote only the boundary committed by this batch**

Because preparation is pipelined, batch N+1 may already have advanced `fsm.Chapters` while batch N is committing. Reading the live tracker here can therefore promote and consume N+1's staged baseline before its confirm is durable; if the node crashes then, boot reconciliation removes that future baseline before replaying N+1, permanently degrading checker coverage. Capture the archived-through ID in `PreparedBatch`, as already done for other post-commit state. This also violates the documented requirement that promotion belongs to the confirm's own post-commit hook ([checker.md:149-152](docs/technical/architecture/subsystems/checker/checker.md#L149-L152)).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1854#issuecomment-5493758658

internal/infra/state/machine.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/ac4df9edaf7d0823c55f07a76ad8819145e84e70 — you were right, and I verified the premise before fixing: `buffer.Merge` runs at prepare (`applyProposal`, machine.go:1581) and it is what copies the tracker onto `fsm.Chapters`, so a pipelined later prepare can advance the live tracker under an earlier batch's post-commit hook. The boundary is now captured in `PreparedBatch` at prepare time — which is the struct's own stated contract ("capture all post-commit data now, so CommitPreparedBatch does not need to read mutable fsm fields") that the hook was violating.

internal/infra/node/applier.go
```diff
@@ -586,9 +586,33 @@ func (a *Applier) RecoverAndReplay(ctx context.Context) (bool, error) {
...
 		}
 	}
 
+	// Reconcile the checker baseline with the archived prefix. Promotion
+	// normally rides the confirm's post-commit hook, but a crash between that
+	// commit and the rename loses it, and nothing re-applies the confirm on
+	// restart. Idempotent; replayed confirms below promote through the same
+	// hook. Non-fatal: the checker refuses a baseline whose chapter does not
+	// match the boundary, so a missed promotion degrades rather than misleads.
+	if err := a.store.PromoteStagedBaseline(a.fsm.Chapters.ArchivedThroughID()); err != nil {
```
NumaryBot, line 611: 🔴 **[blocker] Clear baseline artifacts when replacing the live store**

When follower sync or checkpoint restore replaces `live/`, the sibling `baseline/` directory is retained. If an old staged or live baseline has the same chapter ID as the restored store's archived boundary, this reconciliation accepts or promotes it even though its contents belong to the previous store, causing healthy restored data to report corruption. Baseline artifacts must be invalidated as part of restore activation, then checker coverage can degrade until a new close/confirm cycle.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1854#issuecomment-5493758658

internal/infra/node/applier.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/eb40c53acb29c691e38c161e6af7e5a3fc6df46f — every path that installs foreign live/ content clears baseline/ now: `RestoreCheckpoint` (before the publish rename, so each crash window lands on either the original live/ with its matching baselines or a rollback with none — honest degrade), and `NewStore`'s adopt-checkpoint branch. The boot reconcile needed nothing: it only ever rolls back to the original live/, whose baselines still describe it. Both paths regression-tested (`TestRestoreCheckpoint_ClearsBaselines`, `TestNewStore_AdoptingACheckpointClearsBaselines`).

internal/storage/dal/store.go
```diff
@@ -1412,6 +1567,14 @@ func (s *Store) RestoreCheckpoint(checkpointID uint64) error {
 
 	s.db = nil
 
+	// Before the publish rename, so every crash window is safe: before this,
+	// the rollback restores the original live/ whose baselines still match it;
+	// between this and the rename, the rollback comes back with no baselines and
+	// the checker degrades honestly; after, the foreign live/ starts clean.
+	if err := clearBaselines(s.dataDir); err != nil {
```
NumaryBot, line 1577: 🟠 **[major] Roll back when baseline cleanup fails**

If `clearBaselines` fails, the original database has already been closed and moved to `live.discard`, but this direct return bypasses `rollback`, leaving `s.db` nil and the Store unusable until restart. Route this failure through `rollback` so a failed restore preserves the existing live store as other pre-publish failures do.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1854#issuecomment-5493758658

gfyrag (changes requested):
Independent re-review at exact head eb40c53acb29c691e38c161e6af7e5a3fc6df46f.\n\n[P2][blocking] Roll back when baseline cleanup fails\nLocation: internal/storage/dal/store.go:1567-1574, Store.RestoreCheckpoint\nEvidence: clearBaselines runs after the original live database has been closed and renamed to live.discard and after the staging database has been closed. Returning its error directly bypasses the existing rollback closure, leaving s.db nil and live absent even though the pre-publish restore should preserve the installed store on failure.\nImpact: a filesystem error while deleting baseline artifacts leaves the Store unusable until restart instead of restoring the previous live database; follower synchronization can therefore turn a recoverable cleanup failure into node downtime.\nResolution: route the clearBaselines failure through rollback and add a deterministic failure regression proving that the original bytes remain readable and the Store stays open.\n\nThe previous cross-store baseline finding is fixed: cleanup now happens before the publish rename, so every crash window degrades safely. The active NumaryBot inline finding identifies the same remaining rollback gap.\n\nDECISION: REQUEST CHANGES\nBlocking findings: 1\nNon-blocking findings: 0\nPrevious findings: 1 fixed / 0 still valid / 0 outdated\nResidual risk: MEDIUM

internal/storage/dal/store.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/b66adf04c080315c7511ecfe48530540316bde33 — routed through `rollback` like every other pre-publish failure. One honest note: I did not add a test for this path, because forcing `os.RemoveAll` to fail requires non-root permission tricks the suite (running as root) cannot use; the rollback closure itself is exercised by the existing pre-publish failure tests.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1854#issuecomment-5493758658

Azorlogh:
@gfyrag your P2 is resolved — the rollback routing landed in b66adf04c (three minutes after your review; NumaryBot raised the same finding and its re-review is green), and the regression you asked for is in https://github.com/formancehq/ledger/commit/b61694557b550ec000b06b3aff5564f182c81434: `TestRestoreCheckpoint_BaselineCleanupFailureRollsBack` proves the original bytes are still served, the store stays open, and the same restore succeeds once the failure clears. The failure is injected through a `clearBaselines` seam because `os.RemoveAll` cannot be made to fail deterministically across environments — permission tricks do not bind as root.

gfyrag (approved):
Independent re-review at exact head b61694557b550ec000b06b3aff5564f182c81434.

The prior rollback blocker is fixed: a baseline-cleanup failure now goes through rollback, and the deterministic regression proves the original bytes remain readable, the Store remains open, and a retry succeeds. No new finding was discovered. The targeted DAL package and all functional CI jobs pass; the remaining Codecov patch threshold is not a correctness failure.

DECISION: APPROVE
Head reviewed: b61694557b550ec000b06b3aff5564f182c81434
Blocking findings: 0
Non-blocking findings: 0
Previous findings: 1 fixed / 0 still valid / 0 outdated
Validation considered: go test ./internal/storage/dal + current CI
Residual risk: MEDIUM

PR #1857, opened 2026-09-02
perf(indexer): skip reverse-map reads for new transactions

gfyrag:

Need: Keep the per-replica read index close enough to the primary cursor for aligned filtered reads to complete during sustained ingestion.
Current limitation: On the sylvain development cluster on 2026-09-01, aligned reads timed out while one restarted replica was 167,642 sequences behind. Ten-thousand-log folds took roughly 40–50 seconds. A 2.18-hour process-CPU profile attributed 1.16 hours to reverseMapValue below the indexbuilder 1.20-hour processing branch.
Requirement / constraint: Remove committed reverse-map reads only when lifecycle rules prove the transaction metadata entry absent; preserve authoritative lookup behavior for overwrites, deletes, accounts, and rewrites.
Evidence: EN-1931 and the measured profile/cursor data recorded in docs/technical/architecture/subsystems/indexer/indexer.md.
Durable repository evidence: docs/technical/architecture/subsystems/indexer/indexer.md, section Operational motivation and scope (EN-1931).

## Technical decision

Decision: Add an explicit known-absent insertion path for CreatedTransaction and RevertedTransaction metadata while still writing forward/existence/reverse-map rows and the batch overlay. Keep the replacement path everywhere absence is not proven. Ensure restart recovery cannot schedule a schema rewrite over a CurrentVersion=0 initial backfill.
Why now / why proportionate: Pebble reverse-map point reads were the dominant measured indexbuilder CPU cost. The shortcut is restricted to two new-ID production triggers. The restart guard is part of the same safety proof because a competing rewrite could pre-populate the pending version. The requested documentation audit challenged this lifecycle against code and exposed that prerequisite; the remaining documentation changes align authoritative contracts rather than introduce separate behavior.
Alternatives considered: Keep one uniform replacement path and its measured negative reads; optimize lower-level Pebble reads; or skip the documentation audit. The uniform path preserved the observed bottleneck, lower-level work would retain unnecessary reads, and the audit was necessary to validate the fresh-version premise.

## Summary

- bypass committed reverse-map point reads for newly created and reverted transaction metadata
- preserve reverse-map and same-batch overlay writes for later mutations
- prevent schema-rewrite recovery from competing with an initial backfill
- align typed-metadata, indexer, checker, protobuf, OpenAPI, and E2E documentation with current code

Jira: https://formance-team.atlassian.net/browse/EN-1931

## Validation

The requirement is satisfied when both transaction creation triggers emit correct current/pending index and reverse-map state without consulting committed old values, same-batch mutations still replace correctly, retype replay uses a fresh version, and restart with Current=0/Pending=2 restores only the historical backfill.

Completed before review:
- bash scripts/agent-check
- targeted indexbuilder, readstore, query, controller, and processing tests
- mutation probe reverting CreatedTransaction to replacement path
- protobuf generation
- E2E business package compilation
- OpenAPI parse and git diff --check


internal/application/indexbuilder/backfill.go
```diff
@@ -274,25 +274,34 @@ func (b *Builder) addSchemaRewriteTask(cfg *ledgerIndexConfig, ledgerName string
...
+		// every log below the stale cursor.
+		batch := b.wb.Batch()
+		if batch == nil {
+			// Same contract as bumpPendingVersion below: addSchemaRewriteTask
+			// is only ever called with an active batch. Surface it loudly per
+			// CLAUDE.md invariant #7.
+			return errors.New("invariant: addSchemaRewriteTask called without an active write batch")
+		}
+
+		if err := b.readStore.DeleteBackfillProgressInBatch(batch, bt.bbKey); err != nil {
```
flemzord, line 294: [P1][blocking] Keep the in-memory reset atomic with the fold batch

The cursor deletion and pending-version write are staged in this fold batch, but bumpPendingVersion updates indexVersions immediately and the task cursor is reset before the batch commits. If a later handler, progress write, or Flush fails, processLogs cancels the batch while Builder.loop still runs background tasks with the new in-memory state. The backfill can then persist an advanced cursor for the fresh version while durable state still names the old pending version; after a crash, recovery resumes the old keyspace past rows that were never written and can promote an incomplete index. Please defer or roll back the local reset until commit succeeds, and add a fail-then-retry/restart regression.

openapi.yml
```diff
@@ -1500,7 +1500,7 @@ paths:
       summary: Remove metadata field type
       description: |
         Removes the type declaration for a metadata key on a ledger.
-        Existing typed values are converted back to strings. Requires `ledger:write` scope.
+        Existing metadata values remain unchanged and an attached index is dropped. Requires `ledger:write` scope.
```
flemzord: [P2][blocking] Synchronize the public metadata-type contract

The DELETE description now matches the immutable-primary-value contract, but the PUT description above still says new values are converted and existing values are converted in the background. The ledgerctl set/remove help repeats the same enforcement/conversion semantics. The implementation, updated protobuf contract, and E2E coverage leave primary values unchanged and only rewrite attached index encoding. Please align the PUT OpenAPI description and CLI help/docs so clients are not promised validation or a conversion process that does not exist.

flemzord:
@gfyrag I posted two blocking inline findings: one crash-consistency issue in the retyped-backfill reset path and one public-contract mismatch in the metadata-type documentation.

openapi.yml
gfyrag: Corrigé dans 14fe6c59e9 : OpenAPI, aides `ledgerctl` set/remove et docs CLI décrivent maintenant le type comme contrat d’encodage d’index, sans validation ni conversion des valeurs primaires. Un test protège l’aide CLI ; Schemathesis passe sur les 64 endpoints.

internal/application/indexbuilder/backfill.go
gfyrag: Corrigé dans 14fe6c59e9 : les mutations mémoire du retype sont désormais journalisées par batch et restaurées avant le retour sur toute erreur de handler/progress/commit. La régression `TestRetypeDuringBackfill_FailedFoldRollsBackThenRetriesAndRestarts` force un vrai échec tardif, vérifie l’état durable + mémoire inchangé, rejoue avec succès puis reconstruit après restart. Le package indexbuilder passe sous `-race`.

flemzord (approved):
Re-reviewed on 09b68e11b1360bdb7b5e32cf310bb60bcc24a0d2. Both previous blocking findings are fixed, targeted race tests pass, and all current checks are green.

PR #1860, opened 2026-09-02
fix(test): isolate audit assertions from chapter sealing

gfyrag:
## Summary

- identify the success and failure audit entries by their unique idempotency keys instead of relying on the global audit count or the last entry
- add a deterministic unrelated audit entry so the regression reproduces the original +2 cardinality condition
- wait for scheduled chapter rotation sealing to finish before the shared E2E server proceeds to another spec

## Root cause

The chapter schedule E2E spec stopped the cron after a CloseChapter proposal but did not wait for the asynchronous SealChapter proposal. The following audit spec snapshots the bucket-wide audit count and expected exactly one new entry. When the sealer committed during that window, both the transaction and SealChapter were present.

## Validation

- focused Chapter Schedule and Audit Log run with seed 1788345778, race detector, and all E2E tags: 5 consecutive passes before the deterministic regression entry, plus a final pass after it
- complete E2E Business suite with the failing seed, race detector, coverage, and all tags: pass
- bash scripts/agent-check: pass before and after synchronization with the latest release/v3.0

### DIFF
--- a/tests/e2e/business/audit_test.go
+++ b/tests/e2e/business/audit_test.go
@@ -61,12 +61,21 @@ func decodeOrder(item *auditpb.AuditItem) *raftcmdpb.Order {
 	return order
 }
 
+func auditEntryWithIdempotency(entries []*auditpb.AuditEntry, key string) *auditpb.AuditEntry {
+	matches := make([]*auditpb.AuditEntry, 0, 1)
+	for _, entry := range entries {
+		if entry.GetIdempotency().GetKey() == key {
+			matches = append(matches, entry)
+		}
+	}
+
+	Expect(matches).To(HaveLen(1), "expected exactly one audit entry for idempotency key %q", key)
+
+	return matches[0]
+}
+
 var _ = Describe("Audit Log", Ordered, func() {
 	const ledgerName = "audit-test"
-
-	// entriesBeforeTest tracks the number of audit entries before each test
-	// so assertions can be relative to the test's own actions.
-	var entriesBeforeTest int
 
 	BeforeAll(func() {
 		// Create the test ledger (generates 1 audit entry)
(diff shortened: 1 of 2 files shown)


flemzord (dismissed):

Azorlogh (commented):
Approach is right, and the `Eventually` mirrors the existing idiom at `chapter_test.go:84`. Two gaps, both in `audit_test.go`, both the same failure mode this PR fixes.

**1.** Four specs still read `entries[len(entries)-1]` — `:224`, `:247`, `:323`, `:395`, `:471`. A leaked `SealChapter` entry is system-scoped with `order_count=1` and empty `ledgers` (`extractLedgers` at `audit.go:111` skips non-ledger orders), so they fail on nil `GetLedgerScoped()` (`:233`, `:255`), `order_count` 1 vs 2 (`:324`), 1 item vs 3 (`:400`), and `ContainElement` (`:473`). Same `UnsignedApplyRequest(key)` + `auditEntryWithIdempotency` treatment as the two specs fixed here.

**2.** `:192` (`after_sequence` pagination) and `:523` (`options.reverse`) assert cardinality *across two* list calls — `len(afterEntries) == len(allEntries)-1` and `len(desc) == len(asc)` — so a key lookup doesn't help. Bounding the second read with an `AUDIT_FIELD_SEQUENCE` max filter taken from the first read's head (the filter shape already at `:551`) pins both to the same window.

Worth doing because the new barrier is the only thing standing between those six specs and a leaked seal: it is level-triggered on a transient state, so a `CloseChapter` that reaches Raft *after* the delete — cron tick racing the `scheduleChanged` notify, which is async — is invisible to it. Very unlikely on current timings, and every robust version costs a full cron period of wall clock, so I'm not asking for a change there.

**3.** Nit: `auditEntryWithIdempotency` asserts without `GinkgoHelper()`, so a mismatch reports `:72` instead of the spec.

Azorlogh (dismissed):
Non-blocking, to be clear on the review above: nothing in the diff is wrong, and the specs I listed are pre-existing fragility that this PR makes *less* likely to fire, not more. Approving.

The follow-up is ~35 test-only lines if you'd rather fold it in here than track it.

Azorlogh (approved):

PR #1868, opened 2026-09-03
docs(consensus): clarify partition safety and read guarantees

gfyrag:
## Summary

- document Ledger v3 as CP under normal Raft membership and explain partition, quorum, retry, and force-removal boundaries
- scope read guarantees to the actual consistency mode and serving data, including leader-local, stale, chapter, audit-index, and usagestore exceptions
- document conditional read-index alignment, query-profile semantics, and mixed-provenance ledger statistics
- update the HTTP/OpenAPI descriptions for audit filtering and asynchronous usage projections

## Stack

This documentation-only PR is stacked on #1876, which contains the runtime fixes and regression tests. Merge #1876 first, then retarget this PR to `release/v3.0`.

## Validation

- `git diff --check fix/routed-read-consistency...HEAD`
- final diff contains only `docs/**` and `openapi.yml`

### DIFF
--- a/docs/technical/architecture/subsystems/consensus/raft-consensus.md
+++ b/docs/technical/architecture/subsystems/consensus/raft-consensus.md
@@ -27,7 +27,6 @@ stateDiagram-v2
     Candidate --> Leader: Majority Votes
     Candidate --> Follower: Another Leader Elected
     Leader --> Follower: Higher Term Detected
-    Leader --> Follower: Network Partition
 ```
 
 ## Single Raft Architecture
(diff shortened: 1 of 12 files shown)


flemzord (dismissed):

flemzord (commented):
Documentation review: four blocking consistency findings are attached inline.

docs/technical/architecture/subsystems/consensus/raft-consensus.md
```diff
@@ -251,10 +251,88 @@ If two nodes become candidates simultaneously, neither can obtain a majority. Th
...
+- **Linearizability**: Committed mutations are seen in the same order by all nodes; default live reads use the quorum-backed `ReadIndex` barrier described below
 - **Durability**: Once committed, an entry is guaranteed to be persisted
 - **Consistency**: All nodes see the same data once synchronized
 
+## Safety and Availability During Partitions
+
+Ledger v3 is **CP rather than AP** in the CAP sense. During a network
+partition, only a partition containing a majority of the configured voters can
+elect a leader, commit writes, or complete the quorum barrier required by a
+default live read. A minority partition does not accept a divergent history;
```
flemzord: [P2][blocking] This CP guarantee excludes the documented remove-node --force path. ForceRemoveNode applies and persists a reduced ConfState locally without consensus; if an isolated former leader force-removes still-live voters, it and the original majority can both form a quorum and commit divergent histories. Please scope the guarantee to normal consensus operations and state the force-remove precondition that removed members must be permanently unreachable and unable to rejoin.

docs/technical/architecture/subsystems/consensus/raft-consensus.md
```diff
@@ -251,10 +251,88 @@ If two nodes become candidates simultaneously, neither can obtain a majority. Th
...
+requests that require consensus wait or fail instead. Availability is therefore
+preserved on the majority side and deliberately sacrificed on every side that
+cannot form a quorum.
+
+This guarantee applies to committed state, which is the boundary visible to a
+successful write response. A request that times out while consensus is being
+lost has an unknown outcome from the client's perspective and must be resolved
```
flemzord: [P2][blocking] This recovery guidance can duplicate an ambiguously committed write. Ledger only replays the original outcome when the original request carried an idempotency key and the same key/content is retried within its TTL; an unkeyed or expired retry is a new operation. Please state those preconditions or name the endpoint-specific reconciliation procedure.

docs/technical/architecture/subsystems/consensus/raft-consensus.md
```diff
@@ -251,10 +251,88 @@ If two nodes become candidates simultaneously, neither can obtain a majority. Th
...
+
+A voter grants at most one vote per term, a candidate needs a majority to
+become leader, and a log entry needs majority replication to commit. Any two
+majorities intersect in at least one voter. Raft's voting and log-matching rules
+therefore prevent two conflicting entries at the same log position from both
+becoming committed, even if nodes on different sides of a partition temporarily
+disagree about who the leader is.
+
+Ledger v3 enables Raft `PreVote` but does not enable `CheckQuorum`. Consequently,
+an isolated former leader may continue to report its local role as leader until
```
flemzord, line 298: [P2][blocking] This paragraph contradicts the state diagram above, which still says Leader --> Follower: Network Partition. With CheckQuorum disabled, a partition alone does not cause that transition; the isolated leader stays locally Leader until it observes a higher term. Please remove or relabel that diagram edge.

docs/technical/architecture/subsystems/consensus/raft-consensus.md
```diff
@@ -251,10 +251,88 @@ If two nodes become candidates simultaneously, neither can obtain a majority. Th
...
+it hears from a higher term. It still cannot commit or acknowledge new writes,
+and a `ReadOnlySafe` `ReadIndex` cannot complete without quorum confirmation.
+That temporary role disagreement is not a split brain at the committed-state
+level. Explicit `stale` reads are the exception: they intentionally bypass the
```
flemzord: [P2][blocking] This exception conflicts with the later ReadIndex section, which still says All reads use ReadIndex, No gRPC forwarding, and No stale reads. RoutedController.readCtrl supports local stale reads, explicit leader forwarding, and sync fallback. Please scope that section to default linearizable reads and document the supported exceptions.

docs/technical/architecture/subsystems/consensus/raft-consensus.md
gfyrag: Fixed in 1ca1e1a597: the CP guarantee is now explicitly scoped to normal consensus membership changes. The section documents that `remove-node --force` is outside the guarantee, requires removed members to be permanently unreachable and unable to run/rejoin, and can create divergent writable configurations if used from an isolated former leader while the original voters remain live.

docs/technical/architecture/subsystems/consensus/raft-consensus.md
gfyrag: Fixed in 1ca1e1a597: the timeout guidance now states all replay preconditions explicitly—an idempotency key on the original request, identical key and content on retry, and retry before TTL expiry (or TTL=0). It also warns that a retry outside those conditions is a new operation that may duplicate the original.

docs/technical/architecture/subsystems/consensus/raft-consensus.md
gfyrag: Fixed in 1ca1e1a597: removed the `Leader --> Follower: Network Partition` edge. The only documented transition from leader to follower is now observing a higher term, consistent with `CheckQuorum` being disabled.

docs/technical/architecture/subsystems/consensus/raft-consensus.md
gfyrag: Fixed in 1ca1e1a597: the ReadIndex section is now scoped to default live reads and documents `stale`, explicit `leader` routing, sync/local-barrier fallback, possible forwarding failure, `ListChapters`, and write routing. The benefits are likewise limited to the normal local linearizable path.

docs/technical/architecture/subsystems/consensus/raft-consensus.md
```diff
@@ -446,7 +538,8 @@ The system can pipeline requests:
 
 ### Linearizable Reads via ReadIndex
 
-All reads use the etcd/raft **ReadIndex** mechanism to provide linearizable consistency on every node:
+Default live reads use the etcd/raft **ReadIndex** mechanism to provide
```
flemzord: [P2][blocking] This still overstates the default read guarantee for filtered ListAuditEntries. RoutedController performs the ReadIndex barrier, but the handler resolves filters through the independently asynchronous audit index; with the default minLogSequence=0, waitFilteredAuditConsistency returns immediately, so the index can omit already committed audit entries. Please either scope this guarantee and the “No stale results” benefit to reads whose secondary indexes are aligned, documenting filtered audit reads as an exception, or wait for audit progress after the barrier before serving them.

flemzord (commented):
Re-review of the updated head: the four previous findings are fixed; one new blocking consistency finding is attached inline.

docs/technical/architecture/subsystems/consensus/raft-consensus.md
gfyrag: Fixed in 5af64df48d: the ReadIndex section now scopes linearizability to FSM-backed data and aligned indexes, removes the blanket “No stale results” claim, and documents filtered ListAuditEntries as an asynchronous-index exception when minLogSequence is zero (including the non-zero bound and unfiltered behavior).

docs/technical/architecture/subsystems/consensus/raft-consensus.md
```diff
@@ -446,27 +538,50 @@ The system can pipeline requests:
...
+- **Linearizable reads on leaders, followers, and caught-up learners** that can reach quorum
+- **Read load distributed** across the cluster in the normal local path
+- **No gRPC forwarding in the normal local path** (lower latency than routing to the leader)
+- **No stale FSM-backed results in linearizable mode** for reads whose serving
+  data and indexes are aligned with the applied-state horizon
+
+**Consistency and routing exceptions**:
+- `x-consistency: stale` bypasses `ReadIndex` and reads the local store directly;
+  it may return an older view.
+- `x-consistency: leader` forwards the read to the leader. The forwarded call
```
flemzord: [P2][blocking] This is not true when `x-consistency: leader` reaches a node that still considers itself leader. `readCtrl` calls `getLeaderCtrl`, which returns `localController` directly for `IsLeader()` and therefore skips `ReadIndex`; with `CheckQuorum` disabled, an isolated former leader can serve stale local state through this mode. Please either run the quorum barrier on the local-leader path or document `leader` consistency as another partition exception instead of saying it always performs a barrier.

flemzord (commented):
Re-review of the updated head: the previous audit-index finding is fixed; one new partition-safety finding is attached inline.

docs/technical/architecture/subsystems/consensus/raft-consensus.md
gfyrag: Fixed in 1445075e20: the consensus docs now describe x-consistency: leader as a partition exception, including the local IsLeader shortcut that skips ReadIndex when CheckQuorum is disabled. I also aligned the CLI/data-flow docs and the relevant code comments so they no longer promise quorum-confirmed freshness for that path.

PR #1869, opened 2026-09-03
perf(tooling): make pre-commit input-sensitive

gfyrag:
Implements INPUT_SENSITIVE_PRECOMMIT_AND_FIXPOINT.

- selects only normalization components whose complete trusted input/output map intersects the exact candidate diff
- runs the full set for unknown paths or unavailable/mismatched identities
- forces a producer when committed generated output changes without source changes
- keeps changed Go on the applicable tidy/lint paths
- reaches a component-level in-memory fingerprint fixpoint without the duplicate full replay
- base-pins CI and AI normalization selection so candidate dependency maps cannot weaken it

Validation:
- bash scripts/agent-check (isolated lint cache): PASS; full fallback, one clean pass
- targeted race-enabled script suites: PASS
- adversarial selector/fixpoint suite: PASS
- proto input-edge mutation: regression failed as expected, then restored
- generated-proto output-edge mutation: regression failed as expected, then restored
- git diff --check and exact review: PASS

Measured selection/fixpoint totals:
- docs-only: 2.62s, zero components
- tooling-only: 2.09s, zero components
- root Go: selects root tidy and lint
- proto: selects proto generation
- operator API: selects operator generation, operator tidy, operator lint
- unknown: full fallback

Full fallback observed component timings included root lint 4m11s, operator lint 26.8s, mock generation 29.7s, dashboards 12.0s, proto 3.2s, and operator generation 3.5s.

This does not implement proportional validation and does not merge automatically.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,8 +42,24 @@ jobs:
           dissociate: true
       - name: Setup Env
         uses: ./.github/actions/default
-      - run: >
-          nix develop --command just pre-commit
+      - name: Run trusted input-sensitive pre-commit
+        env:
+          PR_BASE_SHA: ${{ github.event.pull_request.base.sha }}
+        shell: bash
+        run: |
+          if [[ -z "$PR_BASE_SHA" ]]; then
+            nix develop --command just pre-commit
+            exit
+          fi
+          trusted_tools=$(mktemp -d -t ledger-precommit-tools.XXXXXX)
+          cleanup() {
+            git worktree remove --force "$trusted_tools" >/dev/null 2>&1 || true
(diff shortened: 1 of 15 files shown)


.github/workflows/main.yml
```diff
@@ -42,8 +42,24 @@ jobs:
           dissociate: true
       - name: Setup Env
         uses: ./.github/actions/default
-      - run: >
-          nix develop --command just pre-commit
+      - name: Run trusted input-sensitive pre-commit
```
flemzord, line 45: I don't think we should change that

flemzord (approved):

PR #1877, opened 2026-09-03
feat(chart): allow several variable segments per chart-of-accounts level

ariel-formance:
## What changed

A chart-of-accounts level could declare at most one `$variable` child. It may now declare several, as long as each carries a `.pattern`. Candidates resolve in label order with backtracking, so a candidate whose pattern matches but whose subtree rejects the rest of the address does not hide its siblings.

This is a pure relaxation — charts with zero or one variable segment per level take the identical code path, and `V2ChartSegment` is already a recursive free-form map in the OpenAPI spec, so sibling wildcards need no API or SDK change.

## Why

A segment whose subtree *shape* depends on its own value had no representation. The motivating case: an account level where suffixed "barrier" accounts (`LSER_TRADING`, `LSER_TOPUP`, …) must never hold a `client` subaccount, while bare stubs (`LSER`) may.

The example the client gave is:

bank:BC:branch:U K:account:LSER:client:001 = Valid, 
bank:BC:branch:U K:account:LSER_TRADING = Valid, 
bank:BC:branch:U K:account:LSER_TRADING:client:001 = Invalid.

That can't be a regex. `.pattern` is matched against one segment string in isolation and has no visibility on whether a `client` child was addressed below it, so a single `$accountId` with an optional suffix silently accepts `account:LSER_TRADING:client:<uuid>` — verified against the parser before this change. Splitting the level into `$accountId` (bare stub, has `client`) and `$barrierAccountId` (suffixed, leaf) expresses the rule structurally, with no bespoke validation code.

## Risk

**LOW**

Contained to `internal/chart.go`; `VariableSegment` had no references outside that file. Existing charts are unaffected — the change only admits schemas that were previously rejected at parse time.

## Validation

- [x] Baseline validation: `go build ./...`, `golangci-lint run --build-tags it,local` — 0 issues
- [x] Targeted tests: `TestBarrierAccountValidation` and `TestBarrierPostingValidation` cover plain+client, plain alone, barrier alone, every barrier type, barrier+client rejected (both posting sides), unknown suffix, invalid client id. `TestChartValidation` gains a valid multi-wildcard case with roundtrip assertion, plus a missing-`.pattern` rejection.
- [x] Full suite / broader validation: all 23 unit-test packages pass. The `-tags it` suite was **not** run (needs a local docker daemon) — please let CI cover it.

- `ChartSegment.VariableSegment *ChartVariableSegment` becomes `VariableSegments []ChartVariableSegment`. Internal type, no exported-API consumers outside `chart.go`.
- One previously-rejected schema class now parses: the `cannot have two variable segments with the same prefix` error is gone, replaced by a requirement that each sibling wildcard carry a `.pattern`. Its test is updated accordingly.
- No OpenAPI, SDK, storage, or migration impact. Serialization is unchanged for single-wildcard charts, and variable segments are sorted by label so output stays deterministic.

## Review focus

- **Resolution semantics.** Where sibling patterns overlap, the lexicographically first label wins. Deterministic, but arbitrary from the schema author's point of view — worth deciding whether overlap should instead be an error, and if so whether at parse time (regex disjointness isn't decidable with stdlib `regexp`) or at resolution time.
- **`descendInto` path handling.** It uses `append(path[:len(path):len(path)], …)` to force a copy. `ErrInvalidAccount` retains the path slice, and now that candidates are tried in a loop, the previous shared-backing-array `append` would corrupt an earlier candidate's error message.
- **Requiring `.pattern` on siblings.** An ungated wildcard matches every value and would make its siblings ambiguous, so this seemed like the right guard — but it does mean a level can't mix one gated and one catch-all wildcard.

## Known concerns

Two follow-ups, neither blocking and both out of scope here:

1. **Docs now contradict the engine.** `ledger-schema.mdx` states "One variable per level" and shows this exact restructure as an *invalid* example (plus the `docs-mint` / `product-content-ref` mirrors and generated `llms-full.txt`). Separate repo — needs its own PR alongside this one.
2. **`coa-architect`'s `validate_schema.py`** has `_var_key()` returning the *first* `$` key it finds across 6 call sites, so it would validate a multi-wildcard chart against one arbitrary branch.

Also noticed while in here, pre-existing and unrelated: unknown dot-prefixed properties (`.patern`, `.forbiden`, anything) parse successfully and are silently discarded. Worth tightening, but it's a behavior change that could reject charts already stored, so it wants its own PR.

### DIFF
--- a/docs/events/InsertedSchema.json
+++ b/docs/events/InsertedSchema.json
@@ -44,8 +44,11 @@
     },
     "ChartSegment": {
       "properties": {
-        "VariableSegment": {
-          "$ref": "#/$defs/ChartVariableSegment"
+        "VariableSegments": {
+          "items": {
+            "$ref": "#/$defs/ChartVariableSegment"
+          },
+          "type": "array"
         },
         "FixedSegments": {
           "additionalProperties": {
@@ -60,15 +63,18 @@
       "additionalProperties": false,
       "type": "object",
       "required": [
-        "VariableSegment",
+        "VariableSegments",
         "FixedSegments",
         "Account"
       ]
     },
     "ChartVariableSegment": {
       "properties": {
-        "VariableSegment": {
-          "$ref": "#/$defs/ChartVariableSegment"
+        "VariableSegments": {
+          "items": {
+            "$ref": "#/$defs/ChartVariableSegment"
+          },
+          "type": "array"
         },
         "FixedSegments": {
           "additionalProperties": {
@@ -89,7 +95,7 @@
       "additionalProperties": false,
       "type": "object",
       "required": [
-        "VariableSegment",
+        "VariableSegments",
         "FixedSegments",
         "Account",
         "Pattern",
--- a/internal/README.md
+++ b/internal/README.md
@@ -335,7 +335,7 @@ func SpecMetadata(name string) string
 
 
 <a name="ValidateSegment"></a>
-## func [ValidateSegment](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L46>)
+## func [ValidateSegment](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L51>)
 
 ```go
 func ValidateSegment(addr string) bool
@@ -447,7 +447,7 @@ type BalancesByAssetsByAccounts map[string]BalancesByAssets
 ```
 
 <a name="ChartAccount"></a>
-## type [ChartAccount](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L18-L21>)
+## type [ChartAccount](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L19-L22>)
 
 
 
@@ -459,7 +459,7 @@ type ChartAccount struct {
 ```
 
 <a name="ChartAccount.DefaultMetadata"></a>
-### func \(\*ChartAccount\) [DefaultMetadata](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L318>)
+### func \(\*ChartAccount\) [DefaultMetadata](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L355>)
 
 ```go
 func (c *ChartAccount) DefaultMetadata() metadata.Metadata
@@ -468,7 +468,7 @@ func (c *ChartAccount) DefaultMetadata() metadata.Metadata
 
 
 <a name="ChartAccountMetadata"></a>
-## type [ChartAccountMetadata](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L14-L16>)
+## type [ChartAccountMetadata](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L15-L17>)
 
 
 
@@ -479,7 +479,7 @@ type ChartAccountMetadata struct {
 ```
 
 <a name="ChartAccountRules"></a>
-## type [ChartAccountRules](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L12>)
+## type [ChartAccountRules](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L13>)
 
 
 
@@ -488,7 +488,7 @@ type ChartAccountRules struct{}
 ```
 
 <a name="ChartOfAccounts"></a>
-## type [ChartOfAccounts](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L42>)
+## type [ChartOfAccounts](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L47>)
 
 
 
@@ -497,7 +497,7 @@ type ChartOfAccounts map[string]ChartSegment
 ```
 
 <a name="ChartOfAccounts.FindAccountSchema"></a>
-### func \(\*ChartOfAccounts\) [FindAccountSchema](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L298>)
+### func \(\*ChartOfAccounts\) [FindAccountSchema](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L335>)
 
 ```go
 func (c *ChartOfAccounts) FindAccountSchema(account string) (*ChartAccount, error)
@@ -506,7 +506,7 @@ func (c *ChartOfAccounts) FindAccountSchema(account string) (*ChartAccount, erro
 
 
 <a name="ChartOfAccounts.MarshalJSON"></a>
-### func \(ChartOfAccounts\) [MarshalJSON](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L188>)
+### func \(ChartOfAccounts\) [MarshalJSON](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L208>)
 
 ```go
 func (s ChartOfAccounts) MarshalJSON() ([]byte, error)
@@ -515,7 +515,7 @@ func (s ChartOfAccounts) MarshalJSON() ([]byte, error)
 
 
 <a name="ChartOfAccounts.UnmarshalJSON"></a>
-### func \(\*ChartOfAccounts\) [UnmarshalJSON](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L50>)
+### func \(\*ChartOfAccounts\) [UnmarshalJSON](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L55>)
 
 ```go
 func (s *ChartOfAccounts) UnmarshalJSON(data []byte) error
@@ -524,7 +524,7 @@ func (s *ChartOfAccounts) UnmarshalJSON(data []byte) error
 
 
 <a name="ChartOfAccounts.ValidatePosting"></a>
-### func \(\*ChartOfAccounts\) [ValidatePosting](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L306>)
+### func \(\*ChartOfAccounts\) [ValidatePosting](<https://github.com/formancehq/ledger/blob/main/internal/chart.go#L343>)
 
 ```go
 func (c *ChartOfAccounts) ValidatePosting(posting Posting) error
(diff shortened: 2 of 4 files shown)


ariel-formance:
## Follow-up 1 (only if this lands): public docs contradict the engine

`formancehq/docs` — `apps/docs/content/pages/modules/ledger/working-with/ledger-schema.mdx`

Line 162 states the constraint this PR removes, and lines 165–174 present the exact restructure this PR enables as an **invalid** example:

> - **One variable per level** — each level in your chart can have at most one variable segment
>
> For example, this is invalid because it has two variable segments at the same level:
> ```json
> { "users": { "$userId": {}, "$username": {} } }
> ```

Note the example is still *correctly* invalid after this PR, but for a different reason — neither sibling carries a `.pattern`. So the fix isn't a deletion; the rule needs restating and the example needs a reason:

- Replace the "One variable per level" bullet with something like **Patterns gate sibling variables** — a level may declare several variable segments, but then each one needs its own `.pattern`; a level with a single variable segment can omit it.
- Keep the two-`$userId`/`$username` block as the invalid case, re-captioned to "invalid because neither variable segment has a `.pattern`", and add a valid patterned pair beside it.
- The barrier/client case from this PR's tests is a good motivating example for why the feature exists: `.pattern` only ever sees one segment in isolation, so a structural rule spanning two levels can't be a regex on either.

`apps/docs/public/llms-full.txt:8941` carries the same sentence but is generated — it'll pick this up on the next build, no manual edit.

Two other copies exist and I'd **skip both**: `formancehq/docs-mint` (`modules/ledger/working-with-the-ledger/ledger-schema.mdx:157`) is the dormant Mintlify repo, and `formancehq/product-content-ref` (`apps/docs/content/pages/.../ledger-schema.mdx:170`) is a reference copy. `formancehq/docs` is the one that serves docs.formance.com.

ariel-formance:
## Follow-up 2 (only if this lands): `coa-architect` validates one wildcard branch only

`formancehq/coa-architect` — `skills/coa-designer/scripts/validate_schema.py`

`_var_key()` (line 121) returns the **first** `$`-prefixed key it finds at a level:

```python
def _var_key(children: dict) -> str | None:
    for k in children:
        if isinstance(k, str) and k.startswith("$"):
            return k
    return None
```

Five call sites depend on it: lines 147, 155, 188, 195, 467. Given a chart with sibling wildcards, each one follows a single arbitrary branch — arbitrary in the sense that it's whichever key came first in the JSON document. So `chart_resolves()` would report an address as resolvable when it only matches the *other* sibling, and vice versa.

The encouraging part is that the traversal is already multi-candidate — `chart_resolves()` maintains a `frontier` list and `nxt.extend(...)` elsewhere — so `_var_key` is the only thing collapsing it. Suggested shape:

```python
def _var_keys(children: dict) -> list[str]:
    return [k for k in children if isinstance(k, str) and k.startswith("$")]
```

then at each call site `nxt.extend(children[k] for k in _var_keys(children))` instead of appending one. Line 467's `vk != seg` guard becomes a filter over the list; line 147's `elif children:` fallback stays as-is for the `*` case.

Worth checking the eval fixtures under `skills/coa-designer/evals/` at the same time — none of them use sibling wildcards today (they couldn't), so a regression test with two patterned siblings would be new coverage.

Not urgent: it only misfires on charts that this PR makes expressible in the first place, so nothing in the wild hits it yet.

aalloiteau:
Thanks for the writeup — the rejected-alternatives log made this quick to review, and the `descendInto` path-aliasing catch is a real latent bug that's worth keeping whatever happens to the rest.

I checked out the branch: `go build ./...` is clean, all 23 `internal/...` packages pass, `VariableSegment` has no references outside `chart.go`, and the marshal round-trip is byte-stable with sibling wildcards. The diagnosis is right too — I confirmed against the pre-change parser that a single `$accountId` with an optional suffix accepts `account:LSER_TRADING:client:<uuid>`.

But I think the DSL already expresses this rule, via a fixed segment. The intended way to keep colliding variable segments apart is to introduce a fixed level that discriminates, so each variable sits under its own parent and they can never compete:

```
bank:
  foo:
    $foo_id:
      bar:
  foo_baz:
    $foo_bar_id:
```

I ran three variants of the rule against `main` at e56b43cc9, no engine change. All three hold the full invariant — including the cases this PR exists to reject (`…_TRADING:client:<uuid>`, unknown suffix, malformed client id):

**A — barrier type as its own segment** (`account:LSER:TRADING`). One variable per level throughout:

```json
"account": {
  "$accountId": {
    ".self": {},
    ".pattern": "^[A-Z]{4}$",
    "client": { "$clientId": { ".pattern": "^[0-9a-f-]{8,}$" } },
    "$barrierType": { ".pattern": "^(TRADING|TOPUP|UNIDENTIFIED|CLEARED)$" }
  }
}
```

`client` is fixed and `$barrierType` is the level's only variable, so this parses today. `LSER:TRADING` is a leaf, so `LSER:TRADING:client:<uuid>` is rejected structurally.

**C — your fixed discriminator** (`barrier:LSER_TRADING`):

```json
"account": { "$accountId": { ".self": {}, ".pattern": "^[A-Z]{4}$",
  "client": { "$clientId": { ".pattern": "^[0-9a-f-]{8,}$" } } } },
"barrier": { "$barrierAccountId": { ".pattern": "^[A-Z]{4}_(TRADING|TOPUP|UNIDENTIFIED|CLEARED)$" } }
```

**B — keep `LSER_TRADING` addresses unchanged**, enumerate the barrier accounts as fixed segments (`"LSER": {".self": {}, "client": {…}}, "LSER_TRADING": {}, …`). Also holds; cost is a chart version per new stub.

## Why the fixed-segment route is blocked here

Nothing in the engine. The only obstacle is that these addresses encode the discriminator *inside* a segment string (`LSER_TRADING`) rather than at a segment boundary. Fixed segments match whole segments, so they discriminate on `:` but not on `_`.

Which means the case for sibling wildcards reduces to wanting `_` to act as a sub-segment separator — and that's what `:` already is, and what the one-variable-per-level rule is shaped around. A is arguably the better chart anyway: it makes the barrier type a first-class segment, prefix-queryable and aggregatable over `account:LSER:*`, and it needs no regex at all. `LSER_TRADING` is a compound key hidden in a string, which is the smell that produced this PR.

## What the engine change costs

Worth weighing against "no code at all":

1. **Label order becomes a contract, and it moves persisted data.** `FindAccountSchema` feeds `DefaultMetadata()` (`internal/transaction.go:253`, `internal/controller/ledger/controller_default.go:614`). Two overlapping siblings with different `.metadata`, and renaming `$alpha` → `$omega` silently changes the metadata written onto new accounts. I reproduced this.

2. **Unanchored `.pattern` goes from "too permissive" to "picks the wrong subtree."** `regexp.Match` is a substring search and nothing in the DSL says so. Your chart with the `^…$` anchors dropped — a plausible thing for a schema author to write:

   ```
   ACCEPTED  …:account:LSER:client:c73bcdcc          ← correct
   ACCEPTED  …:account:LSER_TRADING                  ← correct
   ACCEPTED  …:account:LSER_TRADING:client:c73bcdcc  ← the invariant this PR enforces
   ACCEPTED  …:account:LSER_SETTLED                  ← unknown barrier type
   ```

   `[A-Z]{4}` matches inside `LSER_TRADING`, `$accountId` sorts first, and swallows the address. No parse error, no diagnostic.

3. **The `.pattern`-required guard doesn't provide the disjointness it's justified by.** `""`, `"."`, `"(?s).*"` and `"a|"` all satisfy it and match everything — I checked all four. It blocks only the literal absence of the key, while forbidding the one honest way to write a catch-all.

4. **Resolution goes O(address depth) → O(chart size)** worst case, on the per-transaction path (`log_process.go:117`), where the chart is also re-fetched and re-parsed per transaction. An adversarial 4.8 MB chart took 110 ms per address. Not a real-chart concern, but there's no chart size limit.

5. **It invalidates the documented rule**, as you noted — plus `coa-architect`'s `_var_key()`.

## Suggestion

Go with A (or C) and no code. If the addresses are already live and immutable, B covers them today. And if the stub set turns out open-ended enough that B doesn't scale, that's the case to argue this change on explicitly — with anchoring and overlap detection settled first, since sibling dispatch on unanchored patterns is the combination that bites.

Separately and worth their own PRs regardless: `regexp.Match` recompiles the pattern on every call (4.59 µs/op vs 144 ns precompiled, ~32×, and it's already compiled at unmarshal then thrown away), and the silently-dropped unknown dot-keys you spotted.

Happy to be wrong on the addressing constraint — if there's a reason `LSER:TRADING` or `barrier:LSER_TRADING` is off the table on the client's side, that changes the calculus and I'd rather know it than guess.

ariel-formance:
## Shelving this — the DSL already expresses the rule

Closing unmerged, not because the change is wrong but because it buys something the chart can already do. @aalloiteau's review is the reason; recording the state here so this doesn't need re-deriving from scratch later.

**The invariant does not need an engine change.** Three variants hold the full rule against `main` with no code at all, including the cases this PR exists to reject (`…_TRADING:client:<uuid>`, unknown suffix, malformed client id):

- **A** — barrier type as its own segment (`account:LSER:TRADING`). One variable per level throughout, no regex needed for the structural part, and `account:LSER:*` stays prefix-queryable and aggregatable. Probably the better chart independent of this PR.
- **C** — a fixed discriminator level (`barrier:LSER_TRADING`).
- **B** — barrier accounts enumerated as fixed segments, which keeps the existing `LSER_TRADING` addresses byte-identical. Costs a chart version per new stub.

The only thing blocking the fixed-segment route is that these addresses encode the discriminator *inside* a segment string (`LSER_TRADING`) rather than at a segment boundary. Fixed segments discriminate on `:`, not on `_`. So the case for sibling wildcards reduces to wanting `_` to behave as a sub-segment separator — which is what `:` already is, and what the one-variable-per-level rule is shaped around.

**What the engine change would have cost**, for the record: label order becomes a silent contract over persisted data (overlapping siblings with different `.metadata` mean renaming `$alpha` → `$omega` changes the metadata written onto new accounts); unanchored `.pattern` degrades from "too permissive" to "silently resolves the wrong subtree", since `regexp.Match` is a substring search; the `.pattern`-required guard doesn't deliver the disjointness it was justified by (`""`, `"."`, `"(?s).*"` and `"a|"` all satisfy it and match everything); resolution goes O(address depth) → O(chart size) worst case on the per-transaction path, with no chart size limit; and it invalidates both the documented one-variable-per-level rule and `coa-architect`'s `_var_key()`.

## What would revive this

Two open questions, either of which changes the calculus:

1. **Is `LSER:TRADING` or `barrier:LSER_TRADING` actually off the table on the client's side?** Not established either way — nobody has asked them. If the addresses are already live and immutable, B covers them today without engine work, so this only matters if B also fails.
2. **Is the barrier stub set open-ended enough that B doesn't scale?** B costs a chart version per new stub. If stubs are added often enough for that to hurt, that is the case to argue this change on — explicitly, on those grounds.

If it comes back, **anchoring semantics and overlap detection should be settled first**. Sibling dispatch over unanchored patterns is the combination that bites, and it bites silently.

## Salvaged separately

Three things from this review are independent of the sibling-wildcard question and are now filed against `main`, all verified there at 99c372e3f:

- #1901 — unknown dot-prefixed properties (`.patern`, `.metadatas`) parse successfully and are silently discarded. A typo'd `.pattern` silently produces an ungated wildcard.
- #1902 — `regexp.Match` recompiles the pattern on every segment match (~46× slower, 152 allocs/op) while the compiled form is already built at parse time and thrown away.
- #1903 — the `path` slice aliasing in `findAccountSchema`. Not a live bug on `main` (resolution never backtracks), but it's the trap this PR had to work around, and it goes back to being unmarked once this closes.

The two conditional follow-ups I posted above — the `ledger-schema.mdx` docs contradiction and `coa-architect`'s `_var_key()` — were both scoped to "only if this lands", so they drop with it. Nothing to do there.

## Recovering this

Branch `feat/chart-multiple-variable-segments` is **left in place** at `eab3de3b6`, so reopening this PR restores the diff and this thread intact. Should it ever get pruned, `git fetch origin pull/1877/head` still works — GitHub keeps that ref on a closed PR permanently — and the branch can be restored from this page to re-enable reopening.

PR #1880, opened 2026-09-04
refactor(tooling): establish cooperative PR safety boundary

gfyrag:
## Summary

- delete the Git interceptor and duplicate worktree-binding representation
- reduce primary-root protection from seven full-tree snapshots to one Git-native before/after boundary
- retain base-pinned trusted tools, exact candidate identity, final target/head freshness, and leased publication
- simplify nested fixture sanitation while preserving bounded peer synchronization and process-group reaping
- document the measured post-cleanup performance baseline and future measured-ROI policy

## Measured result

- root snapshots: 7 -> 2
- rootguard Git processes: 63 -> 12
- ignored entries/content hashed: 219,330 / 14.28 GiB -> 0
- warm snapshot benchmark: >65.662s -> 235ms
- final two-snapshot boundary: 328ms snapshot compute, 0.49s wall
- Git wrapper: removed
- diff: 679 additions, 1,655 deletions (net -976)

## Validation

- focused rootguard, reviewloop, launcher, nested synchronization, freshness, and lease regressions
- peer-reaping regression: 20 consecutive passes after deterministic fixture synchronization
- go test ./scripts/... -timeout 10m
- go test -race ./scripts/... -timeout 10m (once, through exact-base proportional selection)
- bash scripts/agent-check
- exact-base scripts/agent-check-pr
- git diff --check
- exact base-pinned final technical review: APPROVE, zero findings, LOW residual risk

Target base: 85fa82f94da62a707496a04920812cb9a7caae9b.
Final candidate: 6e04140c3ae21f8355c0a4b427c837720a8b4657.


scripts/internal/rootguard/rootguard.go
```diff
@@ -163,143 +140,60 @@ func (worker *snapshotter) capture(root string) (Snapshot, error) {
...
-	infoExclude, err := worker.runGit(root, "rev-parse", "--path-format=absolute", "--git-path", "info/exclude")
-	if err != nil {
-		return Snapshot{}, fmt.Errorf("resolving repository exclude file: %w", err)
-	}
-	globalExcludes, err := worker.runOptionalGit(root, "config", "--null", "--path", "--get-all", "core.excludesFile")
-	if err != nil {
-		return Snapshot{}, fmt.Errorf("resolving configured exclude files: %w", err)
-	}
 
 	hasher := sha256.New()
```
NumaryBot, line 148: 🟠 **[major] Include index-only state in the root snapshot**

When a child makes an index-only change that produces no patch, such as `touch new && git add -N new`, both diffs are empty and the path is no longer reported by `ls-files --others`; consequently the before/after fingerprints match despite a persistent primary-checkout mutation. Retain a porcelain-status digest or otherwise capture these index states, consistent with the documented persistent-mutation boundary in [ai-worktree-isolation.md](../../docs/technical/contributing/ai-worktree-isolation.md#root-protection).

scripts/rootguard/main.go
```diff
@@ -112,3 +122,31 @@ func run(arguments []string, stdin io.Reader, stdout, stderr io.Writer) int {
...
+
+	done := make(chan error, 1)
+	go func() {
+		done <- command.Wait()
+	}()
+
+	select {
+	case err := <-done:
+		return err
+	case forwarded := <-received:
```
NumaryBot, line 150: 🟠 **[major] Bound cancellation when the child ignores the signal**

If the child traps or ignores SIGTERM/SIGINT, this branch forwards one signal and then waits forever; subsequent signals are queued but never processed. This can leave `ai-pr-loop` and its process group stuck without taking the required after-snapshot, so cancellation needs escalation or continued signal handling with a bounded wait.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1880#issuecomment-5536923032

scripts/internal/rootguard/rootguard.go
```diff
@@ -321,22 +220,10 @@ func (worker *snapshotter) hashPath(
...
-			case "ignored":
-				worker.metrics.IgnoredRegularFiles++
-				worker.metrics.IgnoredLogicalBytes += logicalBytes
-			}
-		}
-	case info.IsDir():
-		writeHashField(hasher, []byte("directory"))
+		worker.metrics.UntrackedRegularFiles++
+		worker.metrics.UntrackedLogicalBytes += logicalBytes
 	default:
```
NumaryBot, line 229: 🔴 **[blocker] Support untracked nested repositories in root snapshots**

When the primary checkout contains a non-ignored nested Git repository or worktree, `git ls-files --others` reports its boundary as a directory, so this branch returns an error and the initial rootguard snapshot aborts the entire PR workflow. The cooperative boundary explicitly permits an already-dirty primary checkout; handle directory boundaries without traversing the nested repository. See [root protection](../../docs/technical/contributing/ai-worktree-isolation.md#root-protection).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1880#issuecomment-5536923032

scripts/internal/rootguard/rootguard.go
gfyrag: Addressed in 76e926a4a9: porcelain status is now part of the deterministic snapshot digest, with an explicit intent-to-add regression. Ignored paths remain excluded. The exact final reviewer classified this finding FIXED.

scripts/rootguard/main.go
gfyrag: Addressed in 76e926a4a9: cancellation now forwards the first signal, allows a two-second grace period, then terminates the whole child process group on timeout or a second signal before taking the after snapshot. The regression uses a child that traps TERM and INT. The exact final reviewer classified this finding FIXED.

scripts/rootguard/main.go
```diff
@@ -79,8 +79,10 @@ func run(arguments []string, stdin io.Reader, stdout, stderr io.Writer) int {
 	command.Stdin = stdin
 	command.Stdout = stdout
 	command.Stderr = stderr
-	childErr := command.Run()
+	command.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+	childErr := runChild(command)
```
NumaryBot, line 85: 🔴 **[blocker] Reap surviving child processes before the final snapshot**

If a validation or review command starts a background process and then exits, `runChild` returns when the direct child finishes, but the remaining process group is left alive. The after-snapshot can therefore pass before that process later mutates the primary checkout, allowing a persistent mutation to escape the boundary. Reap or terminate surviving group members before capturing the after state, as required by [root protection](../../docs/technical/contributing/ai-worktree-isolation.md#root-protection).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1880#issuecomment-5536923032

scripts/rootguard/main.go
```diff
@@ -112,3 +126,81 @@ func run(arguments []string, stdin io.Reader, stdout, stderr io.Writer) int {
 
 	return exitError
 }
+
+func runChild(command *exec.Cmd) error {
+	if err := command.Start(); err != nil {
```
NumaryBot: 🟠 **[major] Register cancellation handling before starting child**

If SIGINT or SIGTERM arrives after `command.Start()` but before `signal.Notify`, the default handler terminates rootguard immediately, leaving the child running and skipping the final snapshot. Register signals before starting the child so cancellation cannot bypass the required cleanup and snapshot boundary documented in [ai-worktree-isolation.md:46-52](../../docs/technical/contributing/ai-worktree-isolation.md#L46-L52).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1880#issuecomment-5536923032

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1880#issuecomment-5536923032

PR #1883, opened 2026-09-04
chore(release): migrate docker images to goreleaser dockers_v2

sylr:
## Summary

- Migrate `.goreleaser.yml` from the deprecated `dockers`/`docker_manifests` blocks to the unified `dockers_v2` config, fixing goreleaser 2.17's deprecation warnings (`dockers and docker_manifests are being phased out...`).
- `dockers_v2` builds and pushes the multi-arch manifest directly instead of pushing per-arch images and merging them afterward, so `build.Dockerfile` now reads binaries from the per-platform context layout (`$TARGETPLATFORM/<binary>`) that `dockers_v2` provides, instead of a flat binary layout.
- `sbom` is explicitly set to `false` since `dockers_v2` defaults it to `true`, which the old config never did — this avoids silently changing published image behavior.

## Behavior note

The per-arch `<version>-amd64`/`<version>-arm64` image tags that the old `docker_manifests` step pushed as a side effect are no longer published; only the multi-arch manifest tags (`v<version>`, `latest`) are. No CI workflow, docs, or scripts reference those per-arch tags, so this is not expected to be a regression.

## Test plan

- [x] `goreleaser check` passes with no deprecation warnings
- [x] `goreleaser release --snapshot --clean` run locally, successfully building both `linux/amd64` and `linux/arm64` images via buildx
- [x] Smoke-tested the built amd64 image (`ledger --help`, `ledgerctl completion bash` baked-in step) works correctly
- [x] `bash scripts/agent-check` passes

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -104,50 +104,28 @@ changelog:
     - "^chore:"
     - "^ci:"
 
-dockers:
-- image_templates: ["{{ .Env.IMAGE_REPOSITORY }}:{{ if not .IsNightly }}v{{ end }}{{ .Version }}-amd64"]
-  goarch: amd64
-  ids:
+dockers_v2:
+- ids:
   - ledger-server
   - ledgerctl
   dockerfile: build.Dockerfile
-  use: buildx
-  build_flag_templates:
+  images:
+  - "{{ .Env.IMAGE_REPOSITORY }}"
+  tags:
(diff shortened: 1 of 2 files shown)


gfyrag (approved):
Reviewed at 03589441d80ed5b76018c70a8adcb90fda23566f. No actionable findings.

PR #1881, opened 2026-09-04
refactor(api): remove public minimum log sequence (EN-1946)

gfyrag:
## Summary

This final slice removes the unreleased client-facing minimum-log-sequence surface after the preceding stack establishes Raft applied index as the common projection horizon.

- remove min_log_sequence from protobuf, HTTP/OpenAPI, CLI, routing, errors, tests, and authoritative docs;
- sequentially realign unreleased protobuf field and enum numbers without reserved declarations;
- retain AuditSuccess min/max log sequence and all internal native log/audit cursors;
- retain stale; removing stale remains a separate product decision;
- remove the superseded audit min-log workaround while preserving certified audit alignment;
- document the deliberate eventual-consistency exception for usage projections.

## Stack

EN-1946 stack 8/8. Depends on #1892 (test/en-1946-antithesis-projection-horizons).

Merge/review order: #1897 → #1889 → #1890 → #1894 → #1891 → #1893 → #1892 → #1881.

## Validation

Final base: 9a054e4e912e1f1eaaf0a1af3bcbea0e19321faf.
Final head: 38a53f2de5712005d83a983b3a91519af981cc9c.
Canonical PR validation: PASS, including full race validation, business E2E (118.855s), cluster E2E (242.580s), and Schemathesis 62/62.
Independent exact diff review: APPROVE (residual risk LOW).

Jira: https://formance-team.atlassian.net/browse/EN-1946. Do not merge automatically.

### Service protocol compatibility

Incompatible wire change: service protocol revision `3` → `4`.

PR #1890, opened 2026-09-04
feat(reads): align live queries to Raft horizons (EN-1946)

gfyrag:
## Summary

- consume the preceding projection certificates at a fixed main-store Raft horizon;
- align only the read/audit projections actually used by each query;
- propagate the local ReadIndex horizon into controller execution;
- distinguish disabled audit projection (permanent FailedPrecondition) from rebuilding (retryable Unavailable);
- preserve main-horizon trimming and snapshot lifecycle guarantees.

## Stack

EN-1946 stack 3/8. Depends on #1889 (feat/en-1946-projection-certificates).

Merge/review order: #1897 → #1889 → #1890 → #1894 → #1891 → #1893 → #1892 → #1881.

## Validation

Final base: 01f5df313ad6cc82e6cfd84fcbd75ad621d721f0.
Final head: de64081212f63bb07407a5c1eb53fbb4b533f9db.
Canonical PR validation: PASS, including full race validation, business/cluster E2E, and Schemathesis 62/62.
Independent exact diff review: APPROVE (residual risk MEDIUM).

Jira: EN-1946. Do not merge automatically.

### Service protocol compatibility

Incompatible semantic change: service protocol revision `1` → `2`.

### DIFF
--- a/docs/technical/architecture/subsystems/api/grpc-api.md
+++ b/docs/technical/architecture/subsystems/api/grpc-api.md
@@ -688,13 +688,14 @@ for {
 }
 ```
 
-For a live request whose filter contains any field other than `seq`, a non-zero
-`MinLogSequence` makes every gRPC node that receives or serves the routed
-request wait for its log index to reach the bound and for its local audit index
-to reach the live audit head sampled afterward. With a zero bound, those
-index-backed results are best-effort. Unfiltered and `seq`-only conjunctions
-scan the audit zone directly and need only the log-sequence wait. Checkpoint
-reads ignore the bound.
+For a live request whose filter contains any field other than `seq`, the server
+automatically fixes a main-store Raft horizon and waits for its local audit
+projection to certify that horizon before querying the projection snapshot. A
+non-zero `MinLogSequence` additionally waits for the native log index before
+that aligned read. Unfiltered and `seq`-only conjunctions scan the authoritative
+audit zone directly and need only the optional log-sequence wait. Checkpoint
+reads ignore the request bound because checkpoint readiness already certifies
+both frozen projections at creation time.
 
 ## Store Metrics
 
--- a/docs/technical/architecture/subsystems/read-path/query-pipeline.md
+++ b/docs/technical/architecture/subsystems/read-path/query-pipeline.md
@@ -2,22 +2,14 @@
 
 ## Overview
 
-Entity-list reads can pass through four distinct mechanisms: a Raft
-**`ReadIndex`** barrier for default live consistency, an optional
-**`min_log_sequence`** wait for read-side freshness, Pebble snapshots, and a
-composable iterator pipeline. A query waits for cross-store alignment only when
-`query.AlignmentOwed` says its shape actually consults an index leaf: filtered
-account/transaction queries and every log query. Unfiltered account/transaction
-queries use only the main-store handle and do not wait for the read index. Point
-reads likewise use their own main-store path. `stale` skips only the Raft
-barrier, while the leader-local and checkpoint exceptions are described below
-and in the [consensus contract](../consensus/raft-consensus.md#linearizable-reads-via-readindex).
+Indexed entity-list reads go through four stages: a **Raft `ReadIndex`** barrier
+for linearizability, a fixed main-store horizon, per-projection **Raft progress**
+waits, coordinated **Pebble snapshots**, and a **composable iterator pipeline**.
+The result is streamed back through gRPC with cursor-based pagination. Point
+reads and main-store-only queries use their own single main-store handle and do
+not wait for projections they do not consult.
 
-The list dispatcher is shared across `ListAccounts`, `ListTransactions`, and
-`ListLogs`, but the selected execution path depends on the query shape. Point
-reads such as `GetAccount` and `GetTransaction`, plus main-store-only queries
-such as `ListLedgers`, use the controller layer but not the read-store iterator
-algebra.
+The indexed-list pipeline is deliberately uniform across `ListAccounts`, `ListTransactions`, and `ListLogs`. Point reads such as `GetAccount` and `GetTransaction`, plus main-store-only queries such as `ListLedgers`, use the controller layer but not the read-store iterator algebra. The differences between indexed lists are which read-store prefix is scanned and how iterators are composed.
 
 ```mermaid
 sequenceDiagram
@@ -31,18 +23,19 @@ sequenceDiagram
     participant Main as Pebble<br/>(main store)
     participant Index as Pebble<br/>(read store)
 
-    C->>G: ListAccounts(filter, min_log_seq, cursor)
-    G->>G: Auth + extract min_log_seq
+    C->>G: ListAccounts(filter, cursor)
+    G->>G: Auth + consistency selection
     G->>Ctrl: ListAccounts(...)
     Ctrl->>N: ReadIndexAndWait(ctx)
     N->>Raft: ReadIndex
     Raft-->>N: commitIndex
     N->>FSM: WaitForApplied(commitIndex)
     FSM-->>N: applied up to commitIndex
-    N-->>Ctrl: ReadBarrierInfo
+    N-->>Ctrl: ReadBarrierInfo(R)
     Ctrl->>Main: store.NewReadHandle()
-    Ctrl->>Index: NewSnapshot / AlignedIndexSnapshot
-    Note over Ctrl,Index: Separate snapshots coordinated to the main-store horizon
+    Ctrl->>Main: Read durable LastAppliedIndex H; assert H >= R
+    Ctrl->>Index: Wait RaftProgress >= H; open snapshot; re-read certificate
+    Note over Ctrl,Index: H is fixed; the wait never chases the moving head
     Ctrl->>Ctrl: Resolve ledger + schema
     Ctrl->>Ctrl: Compile filter → iterator tree
     Ctrl->>Index: Iterate read store
@@ -87,37 +80,62 @@ was used.
 1. Call `node.ReadIndex(ctx)` — Raft sends a heartbeat round-trip to confirm quorum and returns the current commit index.
 2. `fsm.WaitForApplied(commitIndex)` — block until the local FSM has applied every entry up to that commit index.
 
-For a default live request, once both steps succeed the local main-store snapshot
-reflects state at least as fresh as the moment the request reached the cluster.
-FSM-backed results served from that snapshot inherit this linearizable horizon;
-secondary-index results do so only after their separate alignment wait. The
-barrier alone does not advance independently asynchronous projections such as
-the audit index or usagestore.
+Once both succeed, the local Pebble snapshot reflects state at least as fresh as the moment the request reached the cluster. This guarantees **linearizable reads on any node**: a read started after a successful write returns at least that write's effects, regardless of which node serves the read.
 
-If the node is syncing or otherwise unable to confirm `ReadIndex`, the call
-fails. The router may resolve a remote leader and forward; if leadership moved
-local after the failed barrier, it returns the failure rather than serving the
-local store without a fresh barrier. Explicit `stale` skips this barrier, an
-explicit leader-local read does not perform it, and checkpoint reads use their
-frozen snapshot instead.
+If the node is syncing or otherwise unable to confirm `ReadIndex`, the call fails — callers either retry or forward to the leader. The explicit `leader` consistency mode remains available at this stage: a remote hop defaults to the linearizable path, while a node that already considers itself leader serves locally without `R` and still aligns every projection it uses to the fixed local `H` below.
 
-## Freshness — `min_log_sequence`
+## Projection alignment — fixed Raft horizon
 
-The client can pass `min_log_sequence` in the request's `ReadOptions` to demand
-a *read-side* freshness guarantee. `ReadIndex` only guarantees the **FSM** has
-caught up; `min_log_sequence` additionally waits for the **read store** to have
-indexed up to that log sequence. This wait is independent of consistency mode
-and can therefore also block a `stale` read.
+A projection-backed read now aligns automatically against the exact main
+snapshot it will use, independently of the temporary client-facing
+`min_log_sequence` gate:
 
-`waitMinLogSequence()` (`server_bucket.go:434`) calls `readStore.WaitForSequence(ctx, minLogSequence)`, which blocks until the indexer's persisted progress cursor (see [indexer / Progress Cursors](../indexer/indexer.md#progress-cursors)) has caught up.
+1. A linearizable read obtains Raft horizon `R` through `ReadIndexAndWait`.
+   `stale` deliberately skips this step.
+2. The controller opens one main-store snapshot, reads its durable
+   `LastAppliedIndex` as fixed horizon `H`, and verifies `H >= R` when `R` is
+   present.
+3. It waits only for projections the query actually uses. Each must publish a
+   Raft progress certificate `>= H`.
+4. It opens each projection snapshot and re-reads the certificate from that
+   same snapshot before compiling or iterating the query.
 
-Important nuance, repeated from the indexer page: `min_log_sequence` pins **log application** on this replica, **not local rewrite completion**. A read against an index undergoing schema rewrite will keep serving `v_current` until the atomic switch lands — see [indexer / Changing a Metadata Key's Type](../indexer/indexer.md#changing-a-metadata-keys-type-setmetadatafieldtype).
+The indexer captures its own bounded main-store snapshot and publishes `H` only
+after processing every native item visible in it. Intermediate batches advance
+only the native cursor; the terminal projection writes and certificate are one
+atomic Pebble batch. Raft entries that emit no log or audit item still advance
+the certificate. Native log/audit cursors remain separate because folds,
+history resolution, and trimming use them.
 
-Checkpoint reads (see below) ignore `min_log_sequence` — a frozen checkpoint is by definition at a single past sequence.
+For account and transaction queries, `AlignmentOwed` walks the complete
+boolean filter tree. Main-store-only leaves (transaction ID, reverted status,
+and account-target address matching) do not acquire a read-index wait; an AND,
+OR, or NOT tree containing any indexed leaf does. LOGS always uses the read
+index because even its unfiltered universe is projected.
+
+`stale` therefore means “no quorum barrier”, not “permit torn projections”: it
+uses the local main snapshot's fixed `H` and performs the same projection waits.
+Per-index build/rewrite readiness remains explicit through
+`IndexVersionState`; a Raft certificate does not promote an unfinished build.
+
+The legacy `min_log_sequence` request field is still honored as an additional
+client-selected native-sequence floor during this staged rollout. It is no
+longer the causal proof used to align projections: the fixed Raft horizon above
+provides that proof automatically. Checkpoint reads continue to ignore the
+field because they address an immutable past snapshot.
 
 ## Pebble snapshot
 
-`store.NewReadHandle()` returns a Pebble snapshot. Within one controller request, main-store leaves and enrichment (volumes, metadata, transaction bodies) all read through the **one** main-store handle `query.OpenQueryHandle` opened. A query for which `AlignmentOwed` is false uses `readStore.NewSnapshot()` only as an implementation handle and executes `listWithoutIndex`; it does not wait for the read-store fold. A query that actually consults an index uses `query.AlignedIndexSnapshot`, which returns a separate index snapshot whose fold cursor covers the main handle's sequence. The two aligned snapshots are therefore *coordinated*, not identical — the index view may be **ahead** of the main handle, and `query.MainHorizonKeep` trims that excess back to the main-store horizon for TRANSACTIONS and LOGS (ACCOUNTS are served as folded). The contract and its per-target exceptions live in [read-snapshot-consistency.md](read-snapshot-consistency.md#cross-store-alignment-en-1748). The main handle — with the reclaim hold taken alongside it — is owned by the returned cursor and closed when that cursor is closed; the index snapshot and its read lease are released once the page's index iteration ends, before enrichment reads the handle. A paginated API call is streamed from that one cursor; when a client requests a subsequent page using the `x-next-cursor`, the server creates a new controller request and therefore fresh snapshot state.
+`store.NewReadHandle()` returns a Pebble snapshot. Within one controller request,
+main-store leaves and enrichment all use that **one** handle. Read-index
+iterators use a separate snapshot certified at the main handle's applied-index
+horizon. The projection may be ahead, so `query.MainHorizonKeep` still trims by
+the main snapshot's native log sequence for TRANSACTIONS and LOGS (ACCOUNTS are
+served as folded). The Raft certificate does not replace this native trimming
+cursor. The detailed per-target rules live in
+[read-snapshot-consistency.md](read-snapshot-consistency.md#cross-store-alignment-en-1748).
+The main handle and reclamation reservation live with the returned cursor; the
+projection snapshot and read lease are released after index iteration.
 
 The cursor carries only the exclusive resume position (for example, an account address or transaction ID); it does not identify or retain the Pebble snapshot. Within one request/page, results are served under the coordinated consistency contract described above: main-store leaves and enrichment reflect that request's single pin, subject to the per-target cross-store exceptions — ACCOUNTS membership is served as folded, so a page may include index members absent from the pinned main store. Because the cursor does not retain that snapshot state, there is no general snapshot-consistency guarantee across separate pages. Inserts, deletes, or updates committed between requests may therefore affect later pages according to the documented cursor ordering and filtering semantics. Duplications or omissions across pages under concurrent writes are not, by themselves, evidence of a product defect unless an API contract explicitly promises a cross-page snapshot.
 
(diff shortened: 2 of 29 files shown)


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

internal/application/ctrl/controller_default.go
```diff
@@ -1774,25 +1786,93 @@ func (ctrl *DefaultController) ListAuditEntriesFrom(ctx context.Context, store *
...
+	}
+	mainAuditSequence, err := query.ReadLastAuditSequence(handle)
+	if err != nil {
+		return nil, fmt.Errorf("reading main-store audit horizon: %w", err)
+	}
+
+	var auditSnap *pebble.Snapshot
+	indexReader := query.AuditIndexReader(rs)
+	if query.AuditFilterNeedsIndex(filter) {
+		disabled, rebuilding := rs.AuditProjectionState()
```
NumaryBot, line 1829: 🟠 **[major] Validate audit filters before checking projection state**

When the audit projection is disabled or rebuilding, any malformed or unsupported filter is classified by `AuditFilterNeedsIndex` as needing the index, so this early return reports `ErrIndexBuilding` instead of the `InvalidArgument` that `CompileAuditFilter` would produce. Clients can therefore receive a transient availability error for permanently invalid input; validate the filter before applying the projection-state gate.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

internal/application/ctrl/controller_default.go
```diff
@@ -1772,13 +1784,70 @@ func (ctrl *DefaultController) ListAuditEntriesFrom(ctx context.Context, store *
...
 	mainAuditSequence, err := query.ReadLastAuditSequence(handle)
 	if err != nil {
 		return nil, fmt.Errorf("reading main-store audit horizon: %w", err)
 	}
 
-	seqs, loSeq, hiSeq, narrowed, err := query.CompileAuditFilter(rs, filter)
+	var auditSnap *pebble.Snapshot
+	indexReader := query.AuditIndexReader(rs)
+	if query.AuditFilterNeedsIndex(filter) {
+		disabled, rebuilding := rs.AuditProjectionState()
```
NumaryBot, line 1829: 🔴 **[blocker] Validate audit filters before checking projection state**

When the audit projection is disabled or rebuilding, malformed or unsupported filters are conservatively classified as needing the index, so this early return reports `ErrIndexBuilding` instead of the `InvalidArgument` that filter validation would produce. Validate the filter shape before applying the projection-state gate.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

docs/technical/architecture/subsystems/read-path/query-filter.md
```diff
@@ -49,7 +49,7 @@ account, transaction, or log is now on the canonical surface.
 | `GET /v3/_/indexes` | — | — | — | `scope` | option-only |
 | `.../indexes/{id}/inspect` | — | — | `cursor`, `pageSize` | `mode` | option-only |
 | `analyze-accounts` / `analyze-transactions` | — | — | — | `variableThreshold` | option-only |
-| prepared-query `execute` | — (stored; see note) | — | `cursor`, `pageSize` | `parameters`, `mode`, `minLogSequence` | body-driven |
+| prepared-query `execute` | — (stored; see note) | — | `cursor`, `pageSize` | `parameters`, `mode` | body-driven |
```
NumaryBot: 🟠 **[major] Restore minLogSequence to the execute API documentation**

The prepared-query execute handler still accepts and forwards `minLogSequence`, and the executor enforces it, but this changed table and note claim the body contains only parameters, pagination, and mode. This gives API users an incorrect request shape and violates the documentation-maintenance requirement in [AGENTS.md:84](/data/worktrees/formancehq__ledger-pr-1890/AGENTS.md#L84); retain `minLogSequence` in both places.

NumaryBot: 🟠 **[major] Restore `minLogSequence` to the prepared-query execute docs**

The table now omits `minLogSequence`, although the HTTP handler still accepts it, the protobuf still declares it, and `ExecutePreparedQuery` still enforces it as a freshness floor. This also contradicts `query-pipeline.md`, which says the field remains honored, leaving the canonical API documentation inaccurate contrary to [AGENTS.md:82-89](../../../AGENTS.md#L82-L89).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

docs/technical/architecture/subsystems/read-path/query-filter.md
NumaryBot: 🟠 **[major] Restore `minLogSequence` to the prepared-query API docs**

The execute request still accepts `minLogSequence` in `handlers_execute_prepared_query.go`, carries it in `bucket.proto`, and enforces it in `executor.go`, so removing it here and claiming the body contains only parameters, pagination, and mode misdocuments a supported field. Restore the field in the table and note, consistent with the documentation-maintenance requirement in [AGENTS.md:82-89](../../../../../../AGENTS.md#L82-L89).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

docs/technical/architecture/subsystems/read-path/query-pipeline.md
```diff
@@ -87,37 +80,56 @@ was used.
...
-a *read-side* freshness guarantee. `ReadIndex` only guarantees the **FSM** has
-caught up; `min_log_sequence` additionally waits for the **read store** to have
-indexed up to that log sequence. This wait is independent of consistency mode
-and can therefore also block a `stale` read.
+1. A linearizable read obtains Raft horizon `R` through `ReadIndexAndWait`.
+   `stale` deliberately skips this step.
+2. The controller opens one main-store snapshot, reads its durable
+   `LastAppliedIndex` as fixed horizon `H`, and verifies `H >= R` when `R` is
+   present.
+3. It waits only for projections the query actually uses. Each must publish a
```
NumaryBot, line 98: 🟠 **[major] Avoid aligning main-store-only filtered queries**

For filtered account/transaction reads that use only main-store leaves (for example, a transaction-ID condition), `AlignmentOwed` still returns true for every non-nil filter, so these requests wait for the read projection and can stall when that unused projection is lagging. This contradicts the selectivity guarantee documented here; either make the gate inspect the filter leaves or narrow the stated behavior. See [AGENTS.md:84-89](/data/worktrees/formancehq__ledger-pr-1890/AGENTS.md#L84-L89).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

docs/technical/architecture/subsystems/read-path/query-filter.md
gfyrag: Fixed in 4ce2a1ee7e1c720297a6de0f5777ce4afabaeafa: the minLogSequence documentation is restored at this layer. Its removal is intentionally isolated in the final API-cleanup PR #1881, matching the stack boundary.

docs/technical/architecture/subsystems/read-path/query-pipeline.md
gfyrag: Fixed in a73c761053d08310735c962d205f2996aed29f6a. AlignmentOwed now walks AND/OR/NOT recursively and waits only when a read-index-backed leaf is present. Transaction ID/reverted and account-target address filters stay main-store-only; mixed trees still align. Added a table test for those shapes, and updated the authoritative read-path docs. internal/query and scripts/agent-check pass.

internal/query/aligned_snapshot.go
```diff
@@ -13,19 +13,108 @@ import (
...
 // (compileUniverse iterates the main reader) and enriches from that same
 // reader, so it never observes the two stores at different fold points —
 // waiting for the fold would gate it on data it does not read, and a lagging
-// or stopped builder would stall a read that is independent of it. LOGS is
+// or stopped builder would stall a read that is independent of it. The same
+// holds for filters whose complete tree is served by the main store, such as
+// transaction ID/reverted predicates and account-address predicates. LOGS is
 // not in that class even unfiltered: its universe is the read index's
 // per-ledger log limb.
 func AlignmentOwed(filter *commonpb.QueryFilter, target commonpb.QueryTarget) bool {
```
NumaryBot, line 64: 🔴 **[blocker] Preserve filters on reverse main-store-only listings**

When this returns false for a main-store-only filter, `listEntities` dispatches to `listWithoutIndex`, whose reverse branch calls `listDescUnfiltered` and never compiles the filter. Consequently default-order transaction queries using ID/reverted filters, and reverse-order account queries using address filters, return unfiltered rows; malformed filters on those paths can also succeed instead of being rejected.

NumaryBot, line 64: 🔴 **[blocker] Keep the reclaim reservation for filtered aggregation**

For an account-address-only filter, `AlignmentOwed` now returns false, so `OpenQueryHandle` takes no reclaim-floor reservation, but `AggregateVolumes` still unconditionally calls `AlignedIndexSnapshot` for every non-nil filter. If event GC advances before that call, the lease pin is rejected as an invariant violation; even without that race, the supposedly main-store-only aggregation unnecessarily waits for the projection. The aggregation path must follow the same aligned/unaligned decision as listing and prepared-query execution.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

internal/query/aligned_snapshot.go
gfyrag: Fixed in 2bf8222cea50cd8e710854d9b1a13bdbbc9e9142. The unaligned reverse path now calls listDescFiltered whenever a filter is present, so main-store-only filters are compiled and malformed filters still fail. Added a regression that seeds two transactions and proves a reverse ID filter returns only the selected row, plus the malformed-filter case.

gfyrag: Fixed in 2bf8222cea50cd8e710854d9b1a13bdbbc9e9142. AggregateVolumes now branches on the same AlignmentOwed decision as OpenQueryHandle: indexed filters perform the reservation-to-lease handoff; main-store-only filters compile against an unaligned snapshot with no wait, lease, or horizon trim. The regression uses a cancelled context and a main H ahead of the empty projection to prove an address-only aggregate remains available. The full ctrl package and scripts/agent-check pass.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

internal/application/ctrl/controller_default.go
```diff
@@ -1772,13 +1808,73 @@ func (ctrl *DefaultController) ListAuditEntriesFrom(ctx context.Context, store *
...
+		if disabled || rebuilding {
+			state := "disabled"
+			if rebuilding {
+				state = "rebuilding"
+			}
+
+			return nil, &domain.BusinessError{Err: &domain.ErrIndexBuilding{Index: "audit (" + state + ")"}}
```
Azorlogh: `disabled` is node configuration, not a transient state, but it surfaces as `INDEX_BUILDING`, which `reason.go` maps to Unavailable. A client hitting a node started with `--disable-audit-index` retries a permanently failing filtered audit read forever. Suggest keeping `INDEX_BUILDING` for `rebuilding` and returning a distinct reason (e.g. `AUDIT_INDEX_DISABLED`) mapped to FailedPrecondition for `disabled`. Same split applies to the admission preflight in #1889.

internal/query/audit_filter.go
```diff
@@ -87,6 +112,11 @@ func auditFilterNeedsIndex(filter *commonpb.QueryFilter, depth int) bool {
 		}
 
 		return false
+	case *commonpb.QueryFilter_Or:
+		// OR() compiles to the empty set without consulting an index. Every
+		// non-empty OR remains conservative here: validation/compilation owns
+		// the precise grammar error or index dispatch after the barrier.
+		return f.Or == nil || len(f.Or.GetFilters()) > 0
```
NumaryBot, line 119: 🟠 **[major] Treat a nil OR payload as index-independent**

When an OR oneof is present with a nil message payload, `CompileAuditFilter` accepts it as an empty OR and returns an empty result, but this predicate returns true. Consequently, if the audit projection is disabled or rebuilding, this valid index-free query now fails with a projection availability error instead of returning an empty page. Return false whenever `len(f.Or.GetFilters()) == 0`, including a nil payload.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

internal/bootstrap/controller_routed.go
```diff
@@ -487,8 +503,9 @@ var _ ctrl.Controller = (*RoutedController)(nil)
 
 func NewRoutedController(localController ctrl.Controller, node *node.Node, servicePool *transport.ConnectionPool) *RoutedController {
 	return &RoutedController{
-		Node:            node,
-		servicePool:     servicePool,
-		localController: localController,
+		Node:             node,
+		servicePool:      servicePool,
+		localController:  localController,
+		readIndexAndWait: node.ReadIndexAndWait,
```
NumaryBot: 🔴 **[blocker] Handle a nil node before binding ReadIndexAndWait**

When `NewRoutedController(nil, nil, nil)` is called by the existing `TestNewRoutedController`, evaluating this method value dereferences the nil node and panics before the constructor returns. Guard the assignment or rely on the existing fallback in `readCtrl` so the bootstrap test suite remains runnable.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

internal/application/ctrl/controller_default.go
gfyrag: Done — fixed in commit 42c85e9934. Disabled audit projection now returns `ErrAuditDisabled`, whose `AUDIT_DISABLED` reason maps to `FailedPrecondition`; `ErrIndexBuilding`/Unavailable is reserved for rebuilding. The same split is used by the shared checkpoint admission preflight introduced in #1889.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

internal/query/aligned_snapshot.go
```diff
@@ -77,22 +164,21 @@ func AlignedIndexSnapshot(ctx context.Context, rs *readstore.Store, mainReader d
...
-	// meaningless: the pair is consistent by construction at the checkpoint's
-	// sequence (the builder materializes the read index as its fold crosses
-	// the CreatedQueryCheckpoint log). Serve it as-is; the horizon filter
-	// still bounds the index-ahead direction. No GC runs against a frozen
-	// store, so there is no lease to hold.
-	if rs.Frozen() {
-		releaseHold()
-
-		return rs.NewSnapshot(), mainSeq, func() {}, nil
+	mainAppliedIndex, err := mainAppliedHorizon(ctx, mainReader)
```
NumaryBot, line 167: 🟠 **[major] Release the reclaim-floor hold on horizon errors**

When reading the newly added applied horizon fails, or the snapshot is rejected as behind the routed ReadIndex barrier, this returns without invoking `releaseHold`. Since no release closure is returned, the reservation remains permanently registered and can prevent event-history GC from advancing. Release the hold on this error path.

gfyrag: Not applicable on the current head. The hold is owned by the caller, not exclusively by AlignedIndexSnapshot: every production call site either defers releaseHold immediately after OpenReservedQueryHandle/OpenQueryHandle (for example ExecutePreparedQuery, AggregateVolumes, InspectIndex) or transfers it to listEntities, whose error paths release it. AlignedIndexSnapshot releases only when the pin handoff succeeds; on the cited mainAppliedHorizon/read-progress errors the caller defer/error cleanup releases the reservation. The exact final review classified this claim as unsupported, and the alignment tests pass.

gfyrag (commented):
All inline findings on the final head are now answered and resolved. Please refresh the automated review status against 81726d580682278002dc15643839e905919345af.

gfyrag:
Final ownership pass on 81726d580682278002dc15643839e905919345af: fixed authoritative consistency docs/comments and projection selectivity; all inline threads are answered and resolved. The latest reclaim-hold claim was rejected with code evidence: callers retain ownership and release on every error or defer path. Canonical full+E2E+Schemathesis validation PASS; exact final review APPROVE (MEDIUM).

Azorlogh (dismissed):

internal/bootstrap/controller_routed.go
NumaryBot: 🟠 **[major] Preserve nil-node construction**

When `node` is nil, evaluating `node.ReadIndexAndWait` panics during construction. This breaks the existing `TestNewRoutedController`, which explicitly calls `NewRoutedController(nil, nil, nil)`; only bind the method when a node is present.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

Azorlogh (dismissed):

internal/bootstrap/controller_routed.go
gfyrag: Pushback — verified on the exact head 889f6f669b3722a0e1d690536c5f0154a65357a7: `go test ./internal/bootstrap -run ^'^TestNewRoutedController$' -count=1` passes. In Go, taking the method value `node.ReadIndexAndWait` on a nil `*node.Node` captures the receiver without dereferencing it; a panic would occur only if that method value were invoked. The nil path is constructor-only test coverage, while the production call passes the injected `raftNode` at `internal/bootstrap/module.go:653-656`. No code change is required.

NumaryBot: 🔴 **[blocker] Avoid dereferencing a nil node in the constructor**

When `node` is nil, evaluating the method value `node.ReadIndexAndWait` panics during construction. The existing `TestNewRoutedController` explicitly calls `NewRoutedController(nil, nil, nil)`, so the bootstrap test suite now fails before its assertions; assign this dependency only when the node is non-nil.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

internal/bootstrap/controller_routed.go
gfyrag: This is not a nil dereference at construction time. In Go, taking a method value from a nil pointer receiver is valid; the nil receiver is captured and would only panic if the method body dereferenced it when invoked. On this exact rebased head (bc923b06180da925fe4e0af8ad6b589569abab36), `go test ./internal/bootstrap -run "^TestNewRoutedController$" -count=1` passes, including `NewRoutedController(nil, nil, nil)`. Production construction also injects the real raft node. No code change is required.

Azorlogh (dismissed):

internal/application/ctrl/controller_default.go
```diff
@@ -1696,13 +1733,63 @@ func (ctrl *DefaultController) ListAuditEntriesFrom(ctx context.Context, store *
...
+			if progress >= mainAppliedIndex {
+				indexReader = readstore.NewAuditIndexSnapshot(auditSnap)
+
+				break
+			}
+			_ = auditSnap.Close()
+			if rs.Frozen() {
+				return nil, &domain.BusinessError{Err: &domain.ErrIndexBuilding{Index: "audit checkpoint"}}
+			}
+			if err := rs.WaitForAuditRaftProgress(ctx, mainAppliedIndex); errors.Is(err, readstore.ErrAuditProjectionUnavailable) {
```
NumaryBot, line 1802: 🟠 **[major] Map concurrent audit projection failure to Unavailable**

If the audit indexer fails after the initial state check while this request is waiting, `WaitForAuditRaftProgress` returns `ErrAuditProjectionFailed`, which falls through here as a generic wrapped error instead of the retryable `ErrIndexBuilding`/Unavailable used when the same failed state is observed before waiting. This makes the client-visible status depend on a readiness-transition race; translate `ErrAuditProjectionFailed` to the rebuilding error as well.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

internal/application/ctrl/controller_default.go
gfyrag: Fixed on `e0597e5c07a599380ea6d20ec412aa6eb9bd1525`. `ErrAuditProjectionFailed` returned from the wait now maps to the same typed `ErrIndexBuilding{"audit (rebuilding)"}` / Unavailable contract as the pre-check. Added a fail-while-blocked regression that asserts the exact typed error, reason, and kind; focused controller test passes.

internal/application/ctrl/controller_default.go
```diff
@@ -1696,13 +1733,65 @@ func (ctrl *DefaultController) ListAuditEntriesFrom(ctx context.Context, store *
...
-	seqs, loSeq, hiSeq, narrowed, err := query.CompileAuditFilter(rs, filter)
+		for {
+			auditSnap = rs.NewSnapshot()
+			progress, err := rs.ReadAuditRaftProgressFrom(auditSnap)
+			if err != nil {
+				_ = auditSnap.Close()
+
+				return nil, fmt.Errorf("reading audit projection progress: %w", err)
+			}
+			if progress >= mainAppliedIndex {
```
NumaryBot, line 1777: 🟠 **[major] Recheck audit readiness before accepting a certificate**

When an audit rebuild starts after the initial `AuditProjectionState` check but before this certificate check, `Rebuild` has already marked the projection rebuilding while the old certificate may still satisfy the horizon until the reset batch commits. The query then succeeds instead of returning the promised retryable `ErrIndexBuilding`; verify the lifecycle generation/state when accepting the snapshot, as checkpoint publication already does.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

internal/application/ctrl/controller_default.go
gfyrag: Fixed on `b84a9e783a7a4f1c563804e5fa8697fdcb262aba`. Each audit-snapshot attempt now captures `AuditProjectionStateWithGeneration()` before opening the snapshot and rechecks state plus generation before accepting a covering certificate. Disabled/rebuilding transitions return their typed errors; a ready-but-changed generation closes the snapshot and retries, so a completed rebuild cannot leak the previous generation. The lifecycle contract documentation was updated, and the focused audit-alignment controller tests pass.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

Azorlogh (dismissed):

Azorlogh (dismissed):
Approved at c3586f788. Delta since reviewed 81726d580: maps ErrAuditProjectionFailed to the retryable INDEX_BUILDING contract (with test), adds the audit lifecycle generation recheck around the certified snapshot, drops the now-obsolete alignment e2e tweak; CI green, no open threads.

Azorlogh (dismissed):

internal/application/admission/admission.go
```diff
@@ -873,16 +873,14 @@ func (a *Admission) checkQueryCheckpointProjectionReady(reqs []*servicepb.Reques
 		}
 
 		disabled, rebuilding := a.auditProjectionState()
```
NumaryBot, line 875: 🔴 **[blocker] Increment the gRPC protocol revision**

Returning `ErrAuditDisabled` changes disabled-audit requests from retryable `Unavailable` to permanent `FailedPrecondition`, so existing clients interpret the operation differently while `pkg/grpcprotocol.Version` remains `"1"`. Increment the revision as required by [AGENTS.md lines 13-18](/data/worktrees/formancehq__ledger-pr-1890/AGENTS.md#L13-L18).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

docs/technical/architecture/subsystems/api/grpc-api.md
```diff
@@ -688,13 +688,14 @@ for {
...
 ```
 
-For a live request whose filter contains any field other than `seq`, a non-zero
-`MinLogSequence` makes every gRPC node that receives or serves the routed
-request wait for its log index to reach the bound and for its local audit index
-to reach the live audit head sampled afterward. With a zero bound, those
-index-backed results are best-effort. Unfiltered and `seq`-only conjunctions
-scan the audit zone directly and need only the log-sequence wait. Checkpoint
-reads ignore the bound.
+For a live request whose filter contains any field other than `seq`, the server
```
NumaryBot, line 691: 🔴 **[blocker] Increment the service protocol revision**

Live gRPC audit queries now gain automatic projection alignment, and disabled audit projections return a different permanent error, so existing clients observe incompatible operation semantics while `pkg/grpcprotocol.Version` remains `"1"`. The protocol rules explicitly require a same-PR increment for semantic changes ([protocol-compatibility.md:91-98](docs/technical/architecture/subsystems/api/protocol-compatibility.md#L91-L98)); bump the revision and synchronize its documented examples.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

internal/bootstrap/controller_routed.go
```diff
@@ -487,8 +503,9 @@ var _ ctrl.Controller = (*RoutedController)(nil)
...
 	}
```
NumaryBot, line 509: 🔴 **[blocker] Preserve nil-node construction**

`TestNewRoutedController` still calls `NewRoutedController(nil, nil, nil)`, but evaluating `node.ReadIndexAndWait` dereferences the nil node and panics. Guard the assignment when `node == nil` so the existing bootstrap test and supported construction path continue to work.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

Azorlogh (dismissed):

Azorlogh (approved):

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1890#issuecomment-5539561427

Azorlogh (dismissed):

Azorlogh (dismissed):

Azorlogh (approved):

PR #1898, opened 2026-09-04
fix(EN-1956): refuse index bindings outside their declaration's serving window

Azorlogh:
> **Rewritten (2026-09-07).** This PR originally gated on schema-revision *distance* alone. Review surfaced a gap that approach cannot close — a removal restarts the revision counter, so a rewound read store holding a dropped incarnation's binding compares favourably against the re-declared schema — and the follow-up was tracked as EN-1962. Since v3 is unreleased there is no reason to ship the partial fix first, so the branch was rebuilt around declaration incarnations and squashed to one commit. Review rounds against the old design are hidden as outdated.

## What changed

Metadata field declarations now carry two mint-time stamps: an **incarnation** naming the declaration lineage (the ledger-log id of the declaration that opened it, carried unchanged through retypes, necessarily fresh after a removal) and the **revision** counting declarations within that lineage. Index bindings carry both, and `query.ClassifyBindingWindow` — shared by the query compile gate and `InspectIndex` — compares the incarnation for equality first, then allows at most one revision of distance. Anything else reads as `INDEX_BUILDING`, the retryable class `RoutedController` forwards from a follower to the leader. Inspect responses also carry the served binding's type so HTTP and CLI render values under the encoding actually scanned.

## Why

A metadata index's rows are encoded under the type declared when the version was built, so a query must validate and encode its conditions against that binding rather than the live schema. The read store is a WAL-less Pebble database: a hard kill rewinds it to its last flush, and the re-fold re-walks the retype chain from the creation-era binding — folding a retype only bumps the pending version and starts a rewrite, it does not switch. A node can therefore be fully fold-aligned while still serving an ancient binding, and answer a reader past the retypes under pre-retype semantics. Found by the Antithesis model driver as a retype-window divergence. EN-1956.

Need: reads through a metadata index must never silently apply superseded type semantics after a crash.
Current limitation: nothing bounds how far behind the schema a rewound replica's serving binding can be — the stale window reaches back to index creation, and across a remove/re-declare it cannot be measured at all.
Requirement / constraint: a serving binding belongs to the schema's current declaration and is at most one revision behind it; anything else reads as retryable `INDEX_BUILDING`, and a follower forwards such refusals to the leader so converged replicas answer immediately.
Evidence: Antithesis model-driver retype-window divergence (Tests-Model), reproduced and root-caused on a local model-cluster run.
Durable repository evidence: `docs/technical/architecture/subsystems/indexer/indexes.md` documents both checks and the asymmetry that motivates the pair; `ClassifyBindingWindow`'s contract comment and the compile-gate tests pin them.

## Technical decision

Decision: mint-time declaration incarnations (`MetadataFieldSchema.incarnation`, stamped into `SetMetadataFieldTypeLog.incarnation` and `CreatedIndexLog.bound_incarnation`) alongside the existing revisions, tracked per binding in `IndexVersionState`, and a two-step gate — incarnation equality, then one-revision distance — shared by both read paths.

Why now / why proportionate: both checks are load-bearing and neither substitutes for the other. Distance catches the rewound chain walk *within* one lineage, which is the original finding. Equality catches the drop/re-declare discontinuity that resets the counter: there, a dropped incarnation's binding at revision 2 compares favourably against a re-declared schema at revision 1 and would serve a keyspace the removal purged. Comparing identities rather than distances also makes the decision independent of the caller's snapshot — the schema and the binding come from two stores, and skew between the samples can only make the ids differ, which refuses. Fold alignment therefore no longer carries this gate's correctness; it remains required for cross-store leaf skew (EN-1748), including on `InspectIndex`, which takes its handle through `OpenIndexHandle` and its snapshot through `AlignedIndexSnapshot`.

Alternatives considered: (1) revision distance alone — what this PR originally did; cannot express the removal discontinuity, and EN-1962 existed only to track that hole. (2) Refusing a binding *ahead* of the schema — rejects a healthy state, since a retype's atomic switch legitimately lands between a reader's schema sample and its state sample. (3) Non-resetting revisions — actively worse: the dropped incarnation's binding would then sit exactly one revision behind the re-declaration and read as a *legal window*. (4) A compile-time witness forcing callers to fold-align — considered as an interim mitigation; makes the precondition enforceable but still leaves a gate that is only correct given an external precondition. (5) Do nothing and let the model driver tolerate the window — rejected, it blesses user-visible stale semantics.

## Risk

**MEDIUM** — touches the query compile gate, the read-store version-state encoding (header 22→46 bytes), the FSM mint path, restore/replay, and the checker. The new refusal class maps onto the existing retryable contract (`KindUnavailable` → gRPC `Unavailable` → HTTP 503 + `Retry-After`). v3 is unreleased and every push to `release/v3.0` wipes stores, so the header change needs no migration.

## Validation

- [x] Targeted tests: compile-gate cases (two-behind, undeclared-behind-re-declared, dropped incarnation refused, direct predecessor serves, one-revision-ahead serves); `InspectIndex` gate (behind, foreign incarnation, predecessor serves stamped with its own type, fold-alignment wait); mint-side stamping (retype carries the lineage; a declaration with no prior opens one at its own log id, asserted equal to the log's own id); indexbuilder seed → bump → promote carries the lineage; read-store encode/decode round trip; production restore fold persists both stamps; checker flags revision *and* incarnation divergence; forwarding wrappers each fail-once-then-succeed with pinned attempt counts; CLI/HTTP served-type rendering
- [x] Red-verified: gate ignoring the incarnation, mint not opening a fresh lineage after removal, and replay dropping the stamp each fail a test
- [x] Full suite / broader validation: unit sweep green (`tarutil` and `ledgerctl/auth` fail identically on the base — running as root defeats their permission/keyring assumptions); both e2e suites green standalone on this commit; e2e restore spec extended with a post-checkpoint remove/re-declare so `CheckStore` exercises the incarnation across a real restore; review loop run locally against this commit: **APPROVE, 0 findings, residual risk LOW**

- Proto: `MetadataFieldSchema.incarnation`, `SetMetadataFieldTypeLog.incarnation`, `CreatedIndexLog.bound_incarnation`, `InspectIndexResponse.served_type`, plus the revisions from the original design (v3 unreleased — no reservations, no compat shims per policy).
- Read store: `IndexVersionState` gains `Current/PendingRevision` and `Current/PendingIncarnation`; encoded header 22→46 bytes; the atomic switch promotes the whole binding through `Promoted()`.
- Reads: a binding from a replaced declaration lineage, or more than one revision behind, refuses as `INDEX_BUILDING` instead of serving stale semantics.
- Routing: a follower's local `INDEX_BUILDING` refusal forwards to the leader — the whole error class, so initial backfills and activation-pending indexes included. Explicitly-stale reads are exempt, the leader never forwards, and a read that already ran remotely is not re-sent.
- Restore/audit: `replay.Writer` carries the whole minted field; `compareSchema` compares type, revision and incarnation both ways, so a rebuild that drops either stamp fails the store check and the restore service's validate step as `SCHEMA_MISMATCH`.
- Inspect: responses carry `served_type`; HTTP and CLI render with it instead of re-reading the live declared type.

## Review focus

1. `ClassifyBindingWindow` in `internal/query/compile.go` — the verdict ladder and the argument for why equality-then-distance is the right shape.
2. The incarnation's source: `processSetMetadataFieldType` reads `ctx.Boundaries.GetNextLogId()`, which is the id `processApply` stamps on the log it returns *after* the handler. `TestProcessSetMetadataFieldType_OpensLineageAfterRemoval` pins that coupling.
3. `populateInitialSchema` — declarations made with the ledger take its first log id, so a value can coincide across *different* fields. Harmless because the comparison is always one field's schema against the same field's binding, but worth confirming that reading.
4. The restore/checker path: `replay.Writer` signature change and `fieldTypesEqual`.

## Known concerns

The e2e rewind spec is a contract guard, not the detector: on an idle in-process store the post-rewind re-walk completes before the first observable read, so mid-walk serving states are pinned by the compile-gate unit tests and covered continuously by the Antithesis model driver.


internal/domain/processing/processor_metadata_schema.go
```diff
@@ -32,7 +32,11 @@ func processSetMetadataFieldType(ledger string, order *raftcmdpb.SetMetadataFiel
 		info.MetadataSchema = &commonpb.MetadataSchema{}
 	}
 
-	field := &commonpb.MetadataFieldSchema{Type: order.GetType()}
+	// The revision counts this key's declarations; a redeclaration after a
+	// removal restarts at 1 (the removal cascade dropped any attached index,
+	// so revisions never compare across incarnations).
+	_, prior := commonpb.SchemaFieldForTarget(info.GetMetadataSchema(), order.GetTargetType(), order.GetKey())
+	field := &commonpb.MetadataFieldSchema{Type: order.GetType(), Revision: prior.GetRevision() + 1}
```
NumaryBot, line 39: 🔴 **[blocker] Preserve schema revisions during incremental replay**

When a post-checkpoint `SetMetadataFieldTypeLog` is replayed during incremental restore, `domain/replay` passes only its type to `attributeReplayWriter.SetMetadataFieldType`, which reconstructs `MetadataFieldSchema{Type: fieldType}` with revision 0. The restored ledger then loses the monotonic revision introduced here, so subsequent declarations restart at 1 and the stale-binding compile gate can no longer detect revision distance. Pass and restore the stamped log revision as required by [AGENTS.md:47](../../../AGENTS.md#L47).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1898#issuecomment-5540594773

internal/domain/processing/processor_metadata_schema.go
Azorlogh: Fixed in 0d94f919c — the log's mint-time revision is now threaded through `replay.Writer.SetMetadataFieldType` into the rebuilt `MetadataFieldSchema`, and the checker's `compareSchema` now derives and compares revisions too, so a restore that dropped them is detectable (AGENTS.md item 8). Both halves red-verified against the old behavior.

internal/bootstrap/controller_routed.go
```diff
@@ -163,6 +164,48 @@ func (b *RoutedController) markForwardedIfRemote(ctx context.Context, selected c
...
+// semantics — a rebuild in progress. The refusal is per-index and retryable,
+// but a converged replica can answer NOW, and the leader is the replica
+// least likely to be mid-rebuild: forward instead of bouncing the client.
+//
+// The leader never forwards (no loops); if the leader itself is mid-rebuild,
+// the client keeps the retryable INDEX_BUILDING. Explicitly-stale reads are
+// exempt — they ask for THIS node's view, and forwarding would silently
+// substitute another node's.
+func retryOnStaleBinding[T any](b *RoutedController, ctx context.Context, out T, err error, run func(ctrl.Controller) (T, error)) (T, error) {
+	if !shouldForwardIndexBuilding(err, b.IsLeader(), grpcadp.ConsistencyFromContext(ctx)) {
```
NumaryBot: 🟠 **[major] Avoid retrying an already-routed leader**

When `readCtrl` already selected a remote leader—such as for `x-consistency: leader` or the syncing-node fallback—and that leader returns `INDEX_BUILDING`, this helper still sees the originating node as a follower and sends the identical read to the leader a second time. Track whether the initial controller was local so only local stale-binding refusals are forwarded.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1898#issuecomment-5540594773

internal/bootstrap/controller_routed.go
Azorlogh: Fixed in 60a4bb437 — `shouldForwardIndexBuilding` now requires the refusal to have been served by the local controller, so a read readCtrl already routed to the leader (explicit leader consistency, or the syncing fallback) is never re-sent. Decision-table test extended. (Today the remote's refusal arrives as a raw gRPC status, so `errors.As` couldn't match it anyway — the guard makes the contract explicit rather than relying on that.)

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1898#issuecomment-5540594773

gfyrag:
Review checkpoint for tomorrow: please rebase this PR onto the current `release/v3.0` (`d095a8b085553f6045573d0317003b605d420193`). The stale-base gate currently stops `ai-pr-loop` before technical review. The latest `60a4bb4` fix correctly prevents re-sending a read that was already routed to a remote leader, so that finding is resolved. After the rebase, please rerun CI and the review loop.

Azorlogh:
Rebased onto `release/v3.0` @ d095a8b08 (EN-1945 chapters/cold-storage removal absorbed cleanly — all touched packages green, agent-check PASS). CI rerunning.

Azorlogh:
Addressed the Shipfox review at 25fe4ee2a:

1. **[P1] forwarding retry untested** — `retryOnStaleBinding` now resolves its forward target through an injectable resolver; `TestRetryOnStaleBinding` executes the real retry (local INDEX_BUILDING refusal re-runs on a mock leader, forward profiled) plus every non-forwarding branch, and `TestRoutedController_ListAccounts_StaleRefusalNotForwarded` drives a real wrapped method. Red-verified by muting the retry.
2. **[P2] restore-fold revision untested** — `TestRebuildDelta_BumpsCheckpointIndexOnRetype` now folds a revision-stamped `SetMetadataFieldTypeLog` through `RebuildDelta` → `ReplayLedgerLog` and asserts the restored `LedgerInfo` schema revision. Red-verified by dropping the revision in the fold.
3. **[P2] promotion unasserted at the switch** — both schema-rewrite switch tests and `TestBindingRevisionWalk` (now driven through `completeBackfill`) assert the promoted type/declaredness/revision. Red-verified by zeroing the promotion.
4. **[Spec P2] broader forwarding contract** — deliberate; now documented in the PR body (What changed / Architecture impact) and in `retryOnStaleBinding`'s contract note: the forward keys on the whole INDEX_BUILDING class, initial builds included.
5. **[Spec P3] coverage claims** — the Validation section now lists exactly what is pinned and how each was red-verified.

Also added the bugfix-evidence block to the PR body for the `ai-pr-loop` gate.

Azorlogh:
Addressed the second Shipfox review at f5908d48a:

1. **[Spec P1] InspectIndex stale-binding leak — confirmed and fixed.** InspectIndex gated only on `CurrentVersion != 0`; it now applies the same serving-window decision as the compile gate via a shared predicate (`query.BindingWithinServingWindow`, used by both gates so they cannot drift). Regression test red-verified: two-revisions-behind refuses as INDEX_BUILDING, the direct predecessor serves.
2. **[Spec P2] InspectIndex refusal now forwards** — `RoutedController.InspectIndex` routes through `retryOnStaleBinding` like the other reads; forward and stale-consistency branches pinned at the wrapped-method level (the remote-served and leader branches are pinned by the helper/predicate tests shared with the other five methods).
3. **[Standards 2] Wrapped-method retry now runs end to end** — a `readBarrier` seam lets the real `ListAccounts` and `InspectIndex` run fail-once-then-succeed: local refuses once, leader serves, gomock `Times(1)` pins both attempt counts with argument-exact expectations, and the caller gets the leader's page with the forward profiled.
4. **[Standards 1] Cross-lifecycle restore regression** — the e2e restore spec (`tests/e2e/cluster/restore_test.go`) now declares `grade` BEFORE the full checkpoint (revision 1 in the checkpoint's LedgerInfo) and retypes it after (revision 2 in the incremental delta). The existing validate step runs `CheckStore` on the restored store, and this PR makes revision divergence a SCHEMA_MISMATCH there, so a reseed or refold that drops revisions fails the restore; Phase 3 additionally reads back the composed declared type.

Azorlogh:
Addressed the third Shipfox review at 895dddca7:

1. **[Standards 1] routing docs** — raft-consensus.md and query-pipeline.md now describe the post-barrier INDEX_BUILDING forward, its exemptions, and the `forwarded=true`-with-successful-barrier profile shape.
2. **[Standards 2] promotion clump** — the three atomic-switch sites now promote through `IndexVersionState.Promoted`, which carries the whole pending binding; a future switch site can't drop the revision piecemeal.
3. **[Spec 1] cross-incarnation staleness** — confirmed. Pinned for now with a documented compile-gate subtest (`KNOWN GAP: a binding ahead of the schema serves`) covering the drop→re-declare rewind shape, including why a plain binding>schema refusal can't close it (the atomic-switch race the review noted). Exposure is stale-consistency reads only — aligned reads wait for the fold, which re-applies the DropIndex before serving — and the served state is a genuine flush-era prefix. Closing it fully needs an incarnation discriminator observed consistently with the schema; that design (likely another mint-time stamp) is being taken separately rather than bolted on here.

Azorlogh:
Addressed the final Shipfox review at b6961c8d3:

**[Spec 1] InspectIndex unaligned snapshot — confirmed and fixed.** InspectIndex now opens its main-store handle through `query.OpenIndexHandle` (an index read by construction, so the fold floor is always reserved) and reads the version state through `AlignedIndexSnapshot`, exactly as the executor aligns filtered reads. By the time the fold covers the handle's sequence, the DropIndex tombstone and any re-declaration have been re-applied, so a rewound store can no longer expose a dropped incarnation's binding to the gate at any consistency. `TestDefaultController_InspectIndex_WaitsForFoldAlignment` pins the wait (red-verified against the unaligned code), and the compile-gate pin test's comment now states the alignment story — with `AlignmentOwed` true for every filtered read, the cross-incarnation predicate boundary is defense-in-depth, not a reachable window.

Azorlogh:
Addressed the fifth Shipfox review at 70e5a85a9:

1. **[Standards, blocking] stale docs** — all three passages updated: `read-snapshot-consistency.md` now states InspectIndex aligns (and why the version-state summaries GetIndexStatus/GetIndexEntryStatus deliberately do not — they report fold progress itself); `typed-metadata.md` describes the reserve→align→gate flow and the served-type rendering; `indexes.md` documents both gates and the shared predicate.
2. **[Spec] served-binding rendering — confirmed and fixed.** `InspectIndexResponse` now carries `served_type`, stamped from the version state the scan used; the HTTP adapter renders with it and the separate `declaredMetadataType` schema lookup is gone (it was the mistyping source and a second, unaligned read). Regressions: `TestHandleInspectIndex_RendersWithServedBinding` (INT64-served window renders raw integers, DATETIME-served renders RFC3339, non-empty values) and the controller gate test asserts the predecessor window is stamped with its own binding type.

Azorlogh:
Addressed the sixth Shipfox review at HEAD:

**[Spec] CLI served-type rendering** — `ledgerctl indexes inspect` now renders with `resp.GetServedType()`; the second `GetLedger` lookup (`declaredMetadataType`) is deleted, so there is no declared-type path left to confuse with the served binding. The existing `TestFormatMetadataValue_Datetime` regressions pin both render directions (int-typed → raw integer, datetime-typed → RFC3339) against the type now sourced from `served_type`; the structured/JSON output already serialized the raw response including `servedType`.

Azorlogh:
Addressed the seventh Shipfox review (approve with comments) at HEAD:

1. **[P2] API contract docs** — `api-comparison.md` (all three InspectIndex rows) and `openapi.yml` (inspect path description + min/max schema descriptions) now document `served_type`, the served-binding rendering during the retype window, and the one-revision `INDEX_BUILDING` refusal.
2. **[P3] `docs/ops/cli.md`** — the `indexes inspect` section describes the served-binding rendering semantics.
3. **[P3] Subsystem READMEs** — indexer, read-path, and consensus READMEs summarize the serving window and the forwarding contract with links to the leaf pages.

Azorlogh:
Addressed the eighth Shipfox review at 47abf0773:

1. **[Spec 1, High] read-store format migration — declined by repo policy.** Ledger v3 is unreleased: every push to `release/v3.0` wipes all stores, so no 22-byte-era read store can reach this commit, and pre-release on-disk compatibility/migrations are explicitly out of scope (no compat shims, no reserved fields). The malformed-record→absent decode contract and the skip in `ReadAllIndexVersionStates` are pre-existing base behavior from EN-1323, unchanged by this PR beyond the header constant. The PR body's Risk section already states the no-migration premise.
2. **[Spec 2] alignment contract** — now stated on `BindingWithinServingWindow` itself: the predicate accepts binding-ahead states and cannot distinguish incarnations, and the fold-alignment prerequisite (with why a binding>schema refusal is impossible) lives in the function's contract comment where a new caller will read it.
3. **[Standards 1] deadline-based proof** — the blocking half now also asserts the error wrap `waiting for read index alignment`, so the test cannot pass by timing out in an earlier read; the deadline provably expired inside the alignment wait.
4. **[Standards 2/3, Low judgement calls]** — leaving the six wrapped-read closures and the binding triple as they are, per the same verification rationale rounds 5–7 applied; `Promoted()` already centralizes the copy-risk site.

Azorlogh:
Addressed the ninth Shipfox review at 4858c6677:

1. **[P2] CLI wiring unguarded** — the render step is extracted into `renderInspectResult` (the exact hunk that owns the served-type decision) and `TestRenderInspectResult_UsesServedBinding` drives it with a discriminating `served_type`, capturing the rendered output: DATETIME-served renders RFC3339, INT64-served renders the raw integer. Red-verified by pinning the type to a constant.
2. **[P3] stale `declaredType` names** — the CLI render helpers and `formatMetadataValue` (plus the test comments) now say `servedType`, matching the HTTP twin.
3. **[P3] checker docs** — `checker.md`'s verification-passes table gains the `compareSchema` row (declared type AND mint-time revision, compared both ways) with a note naming the revision comparison the serving-window gate's detective control; the checker README links it.
4. **[Low spec] residual caller-side alignment invariant** — tracked as EN-1962 (make the gate incarnation-aware — mint-time incarnation discriminator — so safety no longer rests on caller discipline).

Azorlogh:
**ai-pr-loop status on this PR** — the loop was run repeatedly on this machine per the review-checkpoint request; it passes both gates (base revalidation UNCHANGED, bugfix evidence PASS — the required `DISCOVERY/BEFORE_FIX/AFTER_FIX` block is in the PR body) and its normalize/lint stage is clean, but its proportional validation cannot reliably complete here: the full `-race` e2e's fold-drain-budget specs (`cross_store_snapshot_alignment`, `cross_store_value_skew`, the mirror v2 sync) intermittently exceed their 180s budgets on this ARM VM. Evidence that the diff is not implicated: every affected suite is green in CI on the same heads, the full cluster suite passes locally in isolation on both base `e3ef77087` and this head, and the failing specs pass individually.

Two harness issues found while running it, for whoever owns the loop:
1. **Shared golangci cache poisoning**: `agent-validation-env` shares `GOLANGCI_LINT_CACHE` across per-run worktrees; entries recorded against a deleted run worktree make the generated-file exclusion fail open on the next run — 44 phantom iface/unconvert/gocritic findings in `.pb.go` files, reproducible until `rm -rf ~/.cache/ledger-ai/golangci-lint`. A per-run (or path-keyed) lint cache would fix it.
2. **Leaked run worktrees SIGPIPE the launcher**: killed loop runs leave their worktrees registered; once `git worktree list` output grows past a pipe buffer, the launcher's `| sed -n '...;q'` pipeline under `pipefail` dies with exit 141 before printing anything. `git worktree remove` + `prune` recovers.

The independent Shipfox review has converged to approve-with-comments over nine rounds, all findings addressed or tracked (EN-1962); CI is fully green. Happy to rerun the loop here on request, but a machine with more headroom will get a cleaner validation verdict.

Azorlogh:
Addressed the tenth Shipfox review at 5536f7caf:

1. **[Low, Standards] duplicated threshold** — `ClassifyBindingWindow` replaces the boolean predicate: it returns the verdict (`BindingServes` / `BindingPredeclaration` / `BindingBehindWindow`) and both gates translate to their error types at the call site, so the one-revision and first-declaration thresholds exist in exactly one function. The compile-gate tests (including the red-verified refusal cases and the KNOWN GAP pin) and the InspectIndex regressions all pass unchanged.
2. **[Low, Spec] runtime alignment enforcement** — folded into EN-1962, which now covers both the incarnation discriminator and the interim option of a runtime guard (aligned-snapshot type threaded through the classifier's callers); the CONTRACT comment on the classifier references the ticket.

Azorlogh:
Addressed the eleventh Shipfox review at HEAD:

1. **[Standards 1] wrapper coverage** — all six wrapped reads now have per-method fail-once-then-succeed tests with argument-exact leader expectations and pinned attempt counts (`TestRoutedController_ForwardingWrappers` covers the four that lacked one).
2. **[Standards 2] binding value type** — declining, consistent with rounds 7 and 10 which examined and rejected the same clump as non-actionable style: `Promoted()` owns the switch-site copy risk, `ClassifyBindingWindow` owns the decision thresholds, and the serialized fields stay explicit at the storage boundary by design. If a binding struct is wanted, it belongs in EN-1962's rework rather than another rename pass here.
3. **[Spec 1] 22→30 B upgrade note** — per the review's own closing clause: the index subsystem has not shipped (v3 unreleased, stores wiped on every push to `release/v3.0`), so no migration or release note applies.

Azorlogh:
Addressed the twelfth Shipfox review at HEAD:

1. **[Standards 1] e2e guard's stated mechanism** — confirmed; comments-only fix. The `It` description and the stale-read comments now state the post-alignment contract the loop actually pins (probes fold-align regardless of consistency; convergence + no creation-era page), with the mid-walk detectors (compile-gate unit tests, Antithesis driver) named in the header as before.
2. **[Spec 1, minor, self-declared]** — the cross-incarnation note remains tracked as EN-1962; no change.

Azorlogh:
**Rebuilt around declaration incarnations; force-pushed as `c97f26506` (single commit).**

The revision-distance gate this PR shipped could not close one case: a removal restarts the revision counter, so a read store rewound past a `RemoveMetadataFieldType` holds the dropped incarnation's binding at a *higher* revision than the re-declared schema, reads as comfortably converged, and serves a keyspace the removal purged. That was tracked as EN-1962 with the intent of following up. Since v3 is unreleased there was no reason to ship the partial gate first, so the design is now incarnation-based:

1. Declarations carry an **incarnation** — the ledger-log id of the declaration that opened the lineage, carried unchanged through retypes, necessarily fresh after a removal — alongside the revision that counts declarations within it. Both are mint-time stamped (`SetMetadataFieldTypeLog`, `CreatedIndexLog`) and tracked per binding in `IndexVersionState`.
2. `ClassifyBindingWindow` compares the incarnation for **equality** first, then allows one revision of distance. Both checks are load-bearing: distance catches the rewound chain walk within a lineage (the original finding), equality catches the drop/re-declare discontinuity.
3. Because it compares identities rather than distances, the decision is fail-safe under sampling skew — the schema and the binding come from two stores, and any skew can only make the ids differ, which refuses. **Fold alignment no longer carries this gate's correctness**; it stays for cross-store leaf skew (EN-1748), which is why `InspectIndex` still aligns. The `KNOWN GAP` subtest is now a closed-gap assertion.
4. Everything from the previous rounds is kept: served-binding rendering (HTTP + CLI), the six forwarding wrappers with per-method attempt-count tests, `Promoted()`, the restore/checker parity (now covering both stamps), and the doc sweep.

Validation: agent-check PASS; unit sweep green (`tarutil` / `ledgerctl/auth` fail identically on the base under root); both e2e suites green standalone on this commit; the e2e restore spec now performs a post-checkpoint remove/re-declare so `CheckStore` exercises the incarnation across a real restore; three mutations red-verified (gate ignoring the incarnation, mint not opening a fresh lineage, replay dropping the stamp). The review loop run against this commit returned **APPROVE, 0 findings, residual risk LOW**.

EN-1962 is obsolete — the gate is now self-sufficient rather than relying on caller-side alignment. Review rounds against the old design are hidden as outdated; the rewritten PR body describes the current one.

Azorlogh:
Closing in favour of #1931.

This PR gated served bindings by schema-revision distance (then incarnation + distance). That is the wrong layer for the finding: distance measures nothing about monotonicity — two back-to-back retypes give distance 2 with nothing lost — and a depth-1 rewind after a completed switch is a real regression the gate cannot see, which is exactly what the model driver detected. The actual defect is durability: the read store has no WAL, so a promotion committed after the last flush is undone by a kill and the node reopens serving the binding it had already superseded. #1931 flushes each promotion before it is served and drops the revision/incarnation machinery entirely.

Three independent bugs surfaced during this review are not part of #1931 and remain open as follow-ups: `InspectIndex` takes an unaligned read-store snapshot; HTTP and CLI render inspected values under the live declared type instead of the served binding's type during a retype window; a follower's local `INDEX_BUILDING` refusal is not forwarded to the leader.

PR #1884, opened 2026-09-04
Ledger v3: remove chapters and cold storage (EN-1945)

Azorlogh:
## What changed

Removes the chapter subsystem and cold storage entirely: wire model (proto arms renumbered), FSM/runtime (archiver, sealer, chapter scheduler, cold reader), storage layout (`ZoneCold` → `ZoneHistory`, sub-prefixes renumbered, schema version 3 → 4), config/deployment surface (server flags, operator CRD `ColdStorageConfig`/`ColdCache`, antithesis cold volumes), chapter HTTP/gRPC APIs, and all chapter tests/docs. History is now permanent in the main store.

`DeleteLedger` cleanup no longer rides the chapter purge — its range tombstones land in the same apply (`state.DeleteLedgerData`), mirrored on the backup rebuild path and in the checker's replay store. The checker replays the full history from genesis; every baseline fold, archive-boundary escape, and idempotency-TTL carve-out is gone, and two residual carve-outs now flag instead of skipping (stored reference rows surviving a deleted ledger, idempotency entries dated before the verified range). `CreateQueryCheckpoint` and its scheduler are untouched; receipts stay minus `ChapterID`; MinIO/S3 remains for backups.

## Why

Chapters/archiving cost more in complexity than they return right now (the archiver/sealer/restore interplay was a steady source of Antithesis findings, e.g. EN-1940); we may reintroduce the feature later on a released base. EN-1945.

## Product / operational motivation

N/A (executes the removal decided in EN-1945).

## Technical decision

N/A

## Risk

**MEDIUM** — large blast radius across FSM, storage layout, checker, backup/restore, and operator; mitigated by the schema-version bump (old stores refuse to boot), full-history checker verification, and the suite runs below.

## Validation

- [x] `bash scripts/agent-check`
- [x] Targeted tests: full unit run of the root module green (the two failures in `tarutil` / `ledgerctl/auth` are pre-existing chmod-as-root artifacts, present on base); new regressions: populated-ledger deletion leaves CheckStore clean (verified red on the unpurged checker), stranded reference row after `DeleteLedger` flags, pre-range idempotency entry flags
[...]

PR #1899, opened 2026-09-04
fix(EN-1953): persist usagebuilder progress across restarts

gfyrag:
DISCOVERY: NO_EXISTING_WORK
BEFORE_FIX: BUG_REPRODUCED
AFTER_FIX: PASS

## Summary

- keep the usage-store WAL disabled while establishing explicit durable flush watermarks
- flush progressed boot catch-up slices, periodically flush steady-state progress, and flush on controlled shutdown
- durably flush rollback resets before publishing the rewind
- add close/reopen, suffix-resume, periodic retry, reset durability, and abrupt-stop regressions
- align usage subsystem documentation with the permanent-history contract introduced by EN-1945

Need: Normal pod restarts must not replay the complete audit history or expose counters rebuilt from zero.
Current limitation: usagestore disables its WAL, commits with NoSync, and previously closed without flushing; Pebble therefore reopened at the last automatically flushed cursor, commonly zero.
Requirement / constraint: graceful restart preserves all acknowledged usage projection progress; abrupt restart replays only work after the last periodic flush; counters and cursor always share the same durable boundary.
Evidence: [EN-1953](https://formance-team.atlassian.net/browse/EN-1953) and the reported pod-restart behavior.
Durable repository evidence: `docs/technical/architecture/subsystems/usage/usagebuilder.md`.

## Technical decision

Decision: Keep the rebuildable side-store WAL disabled and establish explicit flush boundaries after progressed catch-up slices, every 30 seconds after progress, after rollback reset, and during controlled close.
Why now / why proportionate: EN-1945/#1884 permanently retains primary history, so bounded suffix replay is correct without paying per-batch WAL writes.
Alternatives considered: enable Pebble WAL; flush every batch; do nothing and continue unbounded replay. WAL was unnecessary for a reconstructible projection, per-batch flush would create avoidable L0 churn, and doing nothing preserves the operational regression.

## Dependency

Built on EN-1945 / #1884, merged into `release/v3.0`.

## Validation

The requirement is satisfied when close/reopen preserves the head cursor, an unchanged restart processes no historical entries, abrupt termination recovers from the explicit flush watermark and exactly replays the suffix, and reset state survives reopen.

- `go test ./internal/application/usagebuilder ./internal/storage/usagestore`
- `go test -race ./internal/application/usagebuilder ./internal/storage/usagestore`
- `bash scripts/agent-check`
- `AI_REVIEW_BASE_SHA=afdd58395fd4689c624286893c2c618e38db8148 bash scripts/agent-check-pr`

Rebased and validated against exact `release/v3.0` base `afdd58395fd4689c624286893c2c618e38db8148` on 2026-09-08. The targeted PR gate includes generation/tidy, both lints, baseline invariants/build, and race tests for usagebuilder, bootstrap, and usagestore.

The existing load benchmarks were rerun with the pinned Nix toolchain (Apple M5 Pro, Go 1.27.1, three samples): the 30-second cadence again produced 10 flushes, 10 L0 tables, 97,249 L0 bytes and 4 compactions, versus 300/300/2,917,036/149 at one second. Replaying 3,000 audit entries took 29.2–32.7 ms. The healthy steady-state replay bound includes up to one batch beyond the flush interval.

[EN-1953]: https://formance-team.atlassian.net/browse/EN-1953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Azorlogh (dismissed):

docs/technical/architecture/subsystems/usage/usagebuilder.md
```diff
@@ -107,17 +127,19 @@ stats.PostingCount, _ = snap.GetCounter(ledger, usagestore.CounterPosting)
...
+| Abrupt process or pod stop | Counters, templates, and cursor through the last successful flush watermark | Only the committed mutable suffix after that watermark (at most the steady-state flush interval under a healthy store). |
+| Usagebuilder crash mid-batch | The previous atomic batch and flush watermark | The incomplete batch plus any committed mutable suffix. |
 | Ledger deletion (audit log `DeleteLedger`) | Usagestore range-deletes every counter + template row keyed on the ledger via `DeleteLedger(batch, ledgerName)` | Re-created ledgers start at zero counters; audit-chain history for the old incarnation is idempotent to re-process. |
-| Primary store rolled back beneath the persisted cursor | Nothing — `usagestore.Reset()` drops every counter + template row and clears the cursor | The next boot/tick replays from audit sequence 0 into the freshly-reset store. |
+| Primary store rolled back beneath the persisted cursor | `usagestore.Reset()` atomically drops every counter + template row, clears the cursor, and flushes that reset before cursor zero is published | The next boot/tick replays from audit sequence 0 into the durably reset store. |
```
flemzord: [P2][blocking] Document the residual state when durability flushes fail

The table describes only the successful `Close` and `Reset` outcomes, but `AGENTS.md` requires durability/recovery docs to state mutation order, durable-write order, and residual failure state. `Reset` commits the visible delete/cursor-clear batch before `Flush`; if that flush fails, the live store is already empty while the builder still publishes the old cursor, and an abrupt restart may reopen the previous SST before the retry. `Close` likewise closes the DB after a failed flush, so a controlled shutdown is not guaranteed to retain every committed batch. Please qualify these rows and document both failure states.

docs/technical/architecture/subsystems/usage/usagebuilder.md
```diff
@@ -107,17 +127,19 @@ stats.PostingCount, _ = snap.GetCounter(ledger, usagestore.CounterPosting)
...
 
 ## Cutover semantics
 
 The migration that introduced this subsystem (EN-1420 / EN-1422) moved every non-ID-generator counter off `LedgerBoundaries` and onto the usagestore. On production upgrade, each ledger's counters **reset to 0** and repopulate from the earliest audit entry still reachable in Pebble. Historic pre-upgrade values are lost — accepted trade-off. Fresh ledgers boot with a genesis-derived count that matches the FSM.
```
flemzord: [P2][blocking] Remove the obsolete lossy-history contract

This paragraph still says reconstruction starts at the “earliest audit entry still reachable” and accepts loss of older counter history. EN-1953 explicitly builds on EN-1945's permanent history and requires the documentation to state that a complete replay from zero is valid; line 146 now says the source is always complete. Please update or remove this stale cutover contract without adding a compatibility path.

internal/application/usagebuilder/process_audit_integration_test.go
```diff
@@ -377,6 +379,264 @@ func TestResetIfRolledBack(t *testing.T) {
...
+
+	progress, err := usage.ReadProgress()
+	require.NoError(t, err)
+	assert.Equal(t, uint64(1), progress, "the unflushed cursor must roll back with its counters")
+	counter, err := usage.GetCounter(ledger, usagestore.CounterPosting)
+	require.NoError(t, err)
+	assert.Equal(t, uint64(1), counter)
+
+	b := newTestBuilder(primary, usage, 200)
+	require.NoError(t, b.boot(context.Background()))
```
flemzord, line 669: [P2][blocking] Assert the abrupt-recovery attempt count required by EN-1953

The test proves the persisted cursor/counter roll back together and that the final counter is correct, but it never observes how many audit entries boot attempted. EN-1953 explicitly requires the abrupt-stop regression to assert the exact suffix replay attempt count. Please instrument the processing seam and assert one attempted entry (sequence 2), so a full-history scan that discards the durable prefix cannot satisfy this test.

internal/application/usagebuilder/builder.go
```diff
@@ -44,6 +45,12 @@ const DefaultBatchSize = 200
 // failures that emit no log), so a ticker is what guarantees pickup.
 const TickInterval = 100 * time.Millisecond
 
+// flushInterval bounds how much already-processed audit history an abrupt
+// stop can require the WAL-less usage store to replay. The interval is
+// deliberately internal: it is a durability/performance policy, not an
+// operator-facing consistency setting.
+const flushInterval = 30 * time.Second
```
flemzord, line 52: [P2][blocking] Supply the required load evidence for the 30-second policy

EN-1953's acceptance criteria require this interval to be checked under representative load for L0/compaction churn and catch-up cost. The PR currently records unit, race, and script validation only; there is no reproducible load result in the diff or PR description. Please attach the measurement evidence (or add a representative benchmark) supporting 30 seconds.

internal/application/usagebuilder/builder.go
```diff
@@ -193,9 +249,16 @@ func (b *Builder) boot(ctx context.Context) error {
 		}
 
 		before := cursor
-		deadline := time.Now().Add(catchUpBudget)
+		deadline := b.currentTime().Add(catchUpBudget)
 
 		cursor, err = b.processAuditEntries(ctx, cursor, deadline)
+		if cursor > before {
+			if flushErr := b.flush(cursor, b.currentTime()); flushErr != nil {
```
flemzord, line 259: [P3][non-blocking] Preserve both catch-up errors

`processAuditEntries` can return an advanced cursor plus a later processing/read error after an earlier batch committed. If this flush also fails, this branch logs only `flushErr` and breaks before the `err` handling below, suppressing the underlying cause during boot. Please log or join both errors so the audit failure is not discarded.

flemzord (changes requested):
Requesting changes for four EN-1953 and durability-documentation gaps. The core implementation paths look sound, and the exact current candidate passes the repository's canonical PR validation. One additional non-blocking error-reporting issue is noted inline.

docs/technical/architecture/subsystems/usage/usagebuilder.md
gfyrag: Addressed in f7532c9be9: the failure table now states the actual commit/flush/close ordering and the residual live/durable state for failed Close and Reset flushes.

docs/technical/architecture/subsystems/usage/usagebuilder.md
gfyrag: Addressed in f7532c9be9: the obsolete lossy cutover paragraph is replaced with the unreleased-v3 permanent-history reconstruction contract; replay from sequence 0 is complete and there is no compatibility path.

internal/application/usagebuilder/builder.go
gfyrag: Addressed in f7532c9be9: the boot catch-up branch now joins the underlying processing/read error with the contextualized flush error before logging, preserving both causes.

internal/application/usagebuilder/builder.go
gfyrag: Addressed in f7532c9be9 with reproducible benchmarks plus recorded load evidence. Over five simulated minutes, the 30-second cadence reduced flushes and L0 tables from 300 to 10 and compactions from 149 to 4; replaying a 3,000-entry suffix representing 30 seconds at 100 audits per second took 37.9 to 78.2 ms across three samples. The command, fixture, environment, and results are documented in usagebuilder.md.

internal/application/usagebuilder/process_audit_integration_test.go
gfyrag: Addressed in f7532c9be9: the abrupt-stop regression now observes each attempted audit entry immediately after cursor read and asserts that sequence 2 is the sole attempted entry, so a hidden full-history scan cannot pass.

internal/application/usagebuilder/builder.go
```diff
@@ -242,9 +309,13 @@ func (b *Builder) tick(ctx context.Context) error {
 		}
 	}
 
-	_, err := b.processAuditEntries(ctx, cursor, time.Time{})
+	cursor, processErr := b.processAuditEntries(ctx, cursor, time.Time{})
```
NumaryBot: 🟠 **[major] Flush progress during long steady-state passes**

When a steady-state snapshot contains enough backlog for `processAuditEntries` to run longer than 30 seconds, the flush is deferred until the entire pass returns, despite batches and cursors being committed throughout. An abrupt stop can therefore replay substantially more than the intended 30-second suffix. Apply the interval check between committed batches or time-bound steady-state passes like boot catch-up.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1899#issuecomment-5540964980

internal/application/usagebuilder/builder.go
gfyrag: Addressed in cefe4af1c2. A steady-state pass now receives the next durability deadline and checks it after each committed batch, so an overdue flush is delayed by at most one batch instead of the complete backlog. The new deterministic regression seeds three one-entry batches, makes the deadline overdue, and asserts that only sequence 1 is attempted and flushed on that tick.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1899#issuecomment-5540964980

gfyrag:
Current status — 2026-09-08, after flemzord's approval at 14:38 UTC:

- Candidate remains `02dff7238cfd0e545f4beaf7f6b63664d169800b`. flemzord explicitly approved this exact SHA and independently confirmed that the returned-flush-error scenario is unreachable with pinned Pebble v2.1.4. GitHub reports `APPROVED` and `MERGEABLE`; all seven inline threads are resolved. All 12 executed Actions checks and both Codecov checks passed.
- The remaining red `NumaryBot review` is a legacy commit status authored by NumaryBot at 13:53:26 UTC (`53739804577`, description `Changes requested: 1 finding`). It predates the evidence-based reply and thread resolution, and no replacement status has been published. The earlier review-request event did not produce a new evaluation. It is not a GitHub Actions job/check run, has no target URL, and no supported retry command/workflow is exposed in this repository.
- **This stale status is not a required merge gate.** The effective rules returned by `GET /repos/formancehq/ledger/rules/branches/release%2Fv3.0` come from active organization ruleset [Protected branches](https://github.com/organizations/formancehq/settings/rules/19700368), which requires one approval, code-owner review and resolved threads, but contains no `required_status_checks` rule. There is no classic branch-protection rule (`baseRef.branchProtectionRule = null`), and `gh pr checks 1899 --required` reports no required checks. The failure explains `mergeStateStatus: UNSTABLE`; it does not make NumaryBot a required gate.

No product change, synthetic push, status override, or merge was performed. The historical investigation below is retained for traceability; the prior statements that flemzord's review remains blocking are superseded by the approval above.

---

Addressed the remaining Shipfox finding from https://github.com/formancehq/ledger/pull/1899#issuecomment-5542306557 in 02dff7238cfd0e545f4beaf7f6b63664d169800b: the abrupt-stop failure table now includes the flush interval plus at most one batch, matching the deadline check after each committed batch.

Rebased the five-commit series without conflicts onto `release/v3.0` at afdd58395fd4689c624286893c2c618e38db8148 and pushed the exact branch with an explicit lease. The later base change b2b7e7d033 only touches the Schemathesis test harness; CI will check integration with the moving base.

Reverified all review surfaces, including the six resolved inline threads and the top-level reviews/comments. The four blocking requests and one minor request in flemzord's review 5114135992 remain addressed in rebased commit 98b1eb6cfb (failure-state documentation, permanent-history reconstruction, exact suffix-attempt assertion, load evidence, and joined catch-up errors). The NumaryBot backlog-flush finding remains addressed in fbaf5b6f65. No unresolved code finding remains after independent review of the final candidate.

Validation passed in pinned Nix: targeted race tests; `bash scripts/agent-check`; and `AI_REVIEW_BASE_SHA=afdd58395fd4689c624286893c2c618e38db8148 bash scripts/agent-check-pr`, including generation/tidy, both lints, invariants/build and usagebuilder/bootstrap/usagestore race tests. The final PR check used `GOMAXPROCS=4 GOFLAGS=-p=2` to reduce host resource pressure without removing checks. The worktree is clean.

The three-sample load rerun reproduced 10 versus 300 flushes/L0 tables and 4 versus 149 compactions; replay of 3,000 entries took 29.2–32.7 ms. The PR description now records the exact validation base and these results.

@flemzord, the `CHANGES_REQUESTED` decision is still active although its threads are resolved. Please re-review the corrected candidate so the review gate can be cleared. Fresh CI is being monitored; this PR has not been merged.


Final CI outcome: all 12 executed GitHub Actions checks passed on 02dff7238cfd0e545f4beaf7f6b63664d169800b, including both E2E suites, the model, Dirty, Tests, and Coverage ([run](https://github.com/formancehq/ledger/actions/runs/34234694325)). Release/image jobs were correctly skipped.

NumaryBot subsequently raised a reset-flush retry claim. It was checked against pinned Pebble v2.1.4 and rejected with the concrete synchronous/background-flush control flow in [this reply](https://github.com/formancehq/ledger/pull/1899#discussion_r3958632635); the thread is resolved. All seven inline threads are now resolved. The bot status still reports FAILURE, and flemzord's earlier CHANGES_REQUESTED review is still active. Re-review requests were submitted to both accounts. GitHub reports MERGEABLE (no conflicts) but BLOCKED by review state. No merge was performed.

internal/storage/usagestore/store.go
```diff
@@ -222,6 +241,10 @@ func (s *Store) Reset() error {
 		return fmt.Errorf("committing usage store reset: %w", err)
 	}
 
+	if err := s.Flush(); err != nil {
```
NumaryBot, line 244: 🔴 **[blocker] Keep rollback reset pending after a failed flush**

If this flush fails, the delete batch is already visible while the builder retains the old cursor. The next tick retries only while that cursor remains above the audit head; if new entries bring the head back to the old cursor first, rollback detection stops, processing resumes after the old cursor, and a later successful flush can durably preserve counters missing the entire deleted prefix. Track the reset as pending until its flush succeeds rather than inferring retry solely from `cursor > auditHead`; this projection must continue to reflect authoritative audit history ([AGENTS.md:44](AGENTS.md#L44)).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1899#issuecomment-5540964980

internal/storage/usagestore/store.go
gfyrag: Verified against the pinned Pebble v2.1.4 source (`go.mod:25`): this failure sequence is not reachable for our writable WAL-less usage store.

`Store.Flush()` delegates directly to `DB.Flush()` (`internal/storage/usagestore/store.go:147`). In Pebble [db.go:2034–2065](https://github.com/cockroachdb/pebble/blob/v2.1.4/db.go#L2034), Flush waits on `flushDone` and returns nil; AsyncFlush only returns ErrReadOnly or the result of `makeRoomForWrite(nil)`. Our store is writable with DisableWAL=true (`store.go:80`), and the nil-batch path of [makeRoomForWrite](https://github.com/cockroachdb/pebble/blob/v2.1.4/db.go#L2687) returns nil (`db.go:2759`). A closed database panics rather than returning a recoverable flush error.

SST IO errors take a different path: [compaction.go:1484–1496](https://github.com/cockroachdb/pebble/blob/v2.1.4/compaction.go#L1484) reports the background error and schedules retry. The memtables enter `flushed` only when err==nil (`compaction.go:1830`), and only those successful entries have their wait channels closed (`compaction.go:1872`). Thus an IO failure keeps this Reset blocked inside Flush; the single builder goroutine cannot run the proposed next tick while the audit head catches up. On eventual success it publishes cursor zero before continuing.

No pending-reset state is needed to close the claimed path for this pinned dependency. The defensive error branch/documentation describes a hypothetical returned error, not a recoverable SST IO error from Pebble. Resolving as unsupported by the production execution path; this should be reassessed if the dependency's flush contract changes.

flemzord (approved):
Reviewed commit 02dff7238cfd0e545f4beaf7f6b63664d169800b. No actionable findings remain. Verified flush boundaries, shutdown ordering, atomic cursor/counter persistence, and exact suffix recovery. Local agent-check and race tests for usagebuilder, usagestore, and bootstrap passed in the pinned Nix environment. Build, test, and coverage checks are green. The remaining NumaryBot failure refers to a resolved finding whose returned-flush-error path is unreachable with the pinned Pebble v2.1.4 writable store; verified against the dependency source.

PR #1904, opened 2026-09-04
feat(indexbuilder): key reverse metadata map by field

gfyrag:
## What changed

- Re-key account and transaction metadata reverse maps as field/version/entity.
- Bound metadata rewrite, old-version GC, and field purge to exact Pebble ranges.
- Preserve ordered read-your-writes semantics when range tombstones and later point writes coexist in one batch.
- Add architecture documentation, boundary/regression tests, and hot-path/maintenance benchmarks. Mark the old monolithic agent instructions as a historical snapshot and link to the current reverse-map contracts.

## Why

Implements [EN-1957](https://formance-team.atlassian.net/browse/EN-1957). The former entity-first layout forced maintenance for one metadata field to scan the complete reverse-map namespace and plan individual deletes.

Need: metadata index rewrites and removals must remain bounded by the field being maintained without increasing preload cost.

Current limitation: for 10,000 entities, 12 fields, and 2 versions, the old layout scans 240,000 reverse-map rows to rewrite 10,000 relevant rows, and purge plans 20,000 point tombstones.

Requirement / constraint: rewrite reads only the selected field/version; purge and old-version GC use bounded range tombstones; normal mutations still recover the old encoded value and emit append-only DEL/ADD events; complete metadata API reads and preload remain unchanged.

Evidence: EN-1957 and the committed reverse-map benchmarks.

Durable repository evidence: `docs/technical/architecture/subsystems/indexer/indexes.md` and `docs/technical/architecture/subsystems/indexer/indexer.md`.

## Technical decision

Decision: encode reverse-map keys as `ledger/namespace/field\0/version/entity` for both accounts and transactions.

Why now / why proportionate: Ledger v3 is unreleased, so the storage layout can be replaced directly without compatibility machinery. The key order turns rewrite and cleanup into prefix/range operations while retaining complete point keys for live writes.

Alternatives considered: keeping entity-first preserves entity-local write locality but retains namespace-wide maintenance scans; adding a second projection doubles storage and write amplification; preloading metadata increases startup and memory cost; removing the reverse map loses the authoritative old encoding required for DEL events.

## Risk

**MEDIUM** — persisted IndexBuilder projection and same-batch overlay ordering change. v3 state is disposable, and targeted boundary/race tests cover both namespaces, rewrite, purge, GC, and delete-then-set ordering.

## Validation

All local checks below passed.

Candidate: `bbd559a229af4027e1afe1c7bdfbcf54e273da26`, rebased onto `release/v3.0` at `8a2a7c85bc3122f64f637edd0720c3e94618ea5f`.

- Targeted race tests: `go test -race ./internal/storage/readstore ./internal/application/indexbuilder ./internal/application/check -timeout 10m`.
- `bash scripts/agent-check`.
- `AI_REVIEW_BASE_SHA=8a2a7c85bc3122f64f637edd0720c3e94618ea5f bash scripts/agent-check-pr` (with `GOFLAGS=-p=2 GOMAXPROCS=2` to bound host resource use; all selected gates retained).
- The canonical selector includes full normalization, root-module race tests and business/cluster E2E tests because the diff includes the protobuf documentation and its generated output.
- Independent review of implementation candidate `4f0862f0`: zero findings. The final two commits only clarify a source comment and mark obsolete legacy guidance; their diff was self-reviewed. Earlier Shipfox and NumaryBot feedback was verified against the code; Paul’s follow-up adds prefix-sibling isolation, malformed zero-version detection and committed overlay-parity coverage.
- Mutation checks: removing the field terminator fails both prefix/purge regressions; removing the zero-version guard fails parser/checker regressions. Production source was restored, then focused race tests passed again.
- The hot-account benchmark smoke check reports the derived `analytical_overlay_hit_ratio`; its single iteration is not new timing evidence.
- Historical benchmark evidence: synthetic 10,000-entity × 12-field × 2-version fixture; rewrite scans 10,000 relevant field-first rows instead of 240,000 entity-first rows. Paired fold measurements are recorded in EN-1957; they were not remeasured during this rebase.

The IndexBuilder's disposable Pebble reverse-map format changes. No primary-store, API, Raft/FSM, preload, or cache-coverage behavior changes. No migration or compatibility reader is included because v3 is unreleased.

## Review focus

- Ordering semantics between reverse-map range tombstones and later point writes in the same batch.
- Exact lower/upper bounds for account and transaction rewrite, GC, and field purge.
- Checker parsing and protection against cross-field/cross-version deletion.

## Known concerns

The field-first layout trades some entity-local write locality for bounded maintenance. The paired fold results recorded in EN-1957 measured no SavedMetadata latency regression and a +4.9% worst-case CreatedTransaction cost at 12 indexed fields; typical usage has fewer than 12 fields and the ticket's accepted envelope is <=7%.

Residual format limitation: malformed reverse-map bytes containing an embedded NUL can be re-split as field/version/entity. `ParseReverseMapKey` rejects zero encoding versions and applies the production metadata/account invariants, so invalid re-splits are rejected, but detection is best-effort: a re-split made entirely of valid components is indistinguishable without an authenticated length or checksum. This is accepted for the disposable readstore format of unreleased v3.

[EN-1957]: https://formance-team.atlassian.net/browse/EN-1957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gfyrag:
Merci — remarques challengées point par point.

Corrigé dans `e25e1c8` :
- les trois pages autoritatives décrivent maintenant le layout field/version/entity, les `DeleteRange` bornés et le rôle du checker comme défense corruption/rupture de lifecycle ;
- les commentaires obsolètes de `gc.go` et `backfill_test.go` sont alignés ;
- `ParseReverseMapKey` valide désormais la metadata key et l’account address avec les mêmes invariants que les écritures. Le test utilise `0x01010101`, donc sa détection ne dépend plus des octets nuls d’une petite version. La limite intrinsèque est explicitement documentée : sans longueur authentifiée/checksum, une séquence corrompue qui se redécoupe entièrement en composants valides reste ambiguë.

Non retenu comme défaut : mutualiser les fixtures de benchmark. Elles sont volontairement package-locales et mesurent deux niveaux différents (opérations Pebble de maintenance vs fold IndexBuilder complet). Les largeurs d’ID différentes font partie de ces workloads ; extraire un helper partagé imposerait une dépendance de test inter-package ou du code de production sans renforcer le contrat fonctionnel.

La description de PR dit maintenant explicitement « synthetic fixture » pour les 10k/240k lignes. Les mesures `+4.9%` et l’enveloppe `<=7%` restent traçables dans EN-1957, qui contient les valeurs, la significativité et les critères d’acceptation.

Validations ciblées après correction : `go test ./internal/storage/readstore ./internal/application/indexbuilder ./internal/application/check` PASS ; `git diff --check` PASS. La boucle PR complète est relancée sur ce SHA.

gfyrag:
Deuxième passe challengée point par point.

Corrigé dans `60e24b2e61` :
- le diagnostic ne prétend plus connaître la cause ; il dit désormais `classified against the … lifecycle`, conformément au rôle de classification du schéma rejoué ;
- la documentation de `ErrReverseMapKeyTruncated` est ramenée aux deux cas réellement couverts (header fixe et bloc version). Les erreurs de terminateur et d’entité restent respectivement `ErrReverseMapKeyMetadataKey` et `ErrReverseMapKeyEntityID`.

Non retenu : rejeter le transaction ID `0`. Cette conclusion est vraie pour la création native v3 (`LedgerBoundaries.NextTransactionId` démarre à 1 et `validateTransactionTarget` traite 0 comme une cible absente), mais pas pour l’espace persistant complet. Le mirror v2 conserve les IDs sources et importe explicitement l’ID 0 : `internal/adapter/v2/translator.go::translateNewTransaction` copie `data.Transaction.ID`, et `TestTranslateBatch_NewTransaction` / `TestTranslateBatch_MultipleLogs` pinent toutes deux `ID: 0` comme cas valide. `processMirrorCreatedTransaction` persiste ensuite cet ID sans remappage. Rejeter 0 dans `ParseReverseMapKey` ferait donc classifier comme corrompues des reverse-map rows légitimes de ledgers mirror.

Validation ciblée : `go test ./internal/application/check ./internal/storage/readstore` PASS ; `git diff --check` PASS. La boucle PR complète repart sur ce SHA.

gfyrag:
Troisième passe challengée : les trois remarques sont fondées, non bloquantes et corrigées dans `450d397d2c` sans changement de comportement.

- le commentaire du garde nil-batch parle maintenant de `reverse-map range tombstone` ;
- les deux `Cancel()` volontairement ignorés documentent leur caractère best-effort et l’erreur qui reste prioritaire ;
- `missedBy` / `orphanMissedBy` sont renommés `lifecycleLabel` / `orphanLifecycleLabels`, ce qui retire la causalité implicite devenue fausse.

Validation ciblée : `go test ./internal/application/indexbuilder ./internal/storage/readstore ./internal/application/check` PASS ; `git diff --check` PASS. Nouvelle boucle complète en cours sur ce SHA.

gfyrag:
Quatrième passe challengée.

Corrigé dans `2e66994900` : l’exemple obsolète du commentaire `wantErrContains` cite désormais deux cas réels qui partagent `ErrReverseMapKeyMetadataKey` (clé vide vs alphabet invalide).

Non retenu comme défaut : le couplage du parseur aux invariants de production est intentionnel et déjà documenté dans le contrat de `ParseReverseMapKey` (`Parsed strings are checked with the same production invariants as writes`). Il garantit précisément la symétrie lecture/écriture qui a corrigé le cas de corruption précédent. La dépendance est versionnée et le readstore concerné est une projection secondaire reconstruisible ; recopier l’alphabet dans `keys.go` créerait deux sources de vérité pouvant dériver. Un futur changement d’alphabet devra, comme toute évolution d’un invariant persistant, traiter explicitement son impact au moment où il sera proposé — ce n’est pas un défaut du diff courant.

Validation : `go test ./internal/storage/readstore` PASS ; `git diff --check` PASS. Nouvelle revue canonique sur le SHA final en cours.

gfyrag:
Cinquième passe challengée.

Point benchmark : la prémisse d’une comparaison cross-suite est rejetée. Chaque suite compare `entity_first_baseline` et `field_first_production` à workload constant dans son propre package ; les ratios entre suites ne sont ni calculés ni revendiqués, et les largeurs d’ID différentes sont intentionnelles. Mutualiser les fixtures ferait perdre cette indépendance. J’ai néanmoins ajouté dans les deux benchmarks une note croisée qui rend ce contrat explicite et empêche une future lecture erronée.

Point `ReverseMapPrefix` : confirmé. Le dernier caller de production ayant disparu, conserver un helper exporté de scan namespace-wide était inutile et pouvait encourager le pattern que cette PR supprime. `d6ac0fa29c` retire cette API de `keys.go` et remplace ses seuls usages par des helpers strictement test/benchmark-locales.

Validation : `go test ./internal/storage/readstore ./internal/application/indexbuilder` PASS ; `git diff --check` PASS. Une nouvelle boucle canonique va relire ce SHA.

gfyrag:
Cinquième passe challengée.

Point benchmark : la prémisse d’une comparaison cross-suite est rejetée. Chaque suite compare `entity_first_baseline` et `field_first_production` à workload constant dans son propre package ; les ratios entre suites ne sont ni calculés ni revendiqués, et les largeurs d’ID différentes sont intentionnelles. Mutualiser les fixtures ferait perdre cette indépendance. J’ai néanmoins ajouté dans les deux benchmarks une note croisée qui rend ce contrat explicite et empêche une future lecture erronée.

Point `ReverseMapPrefix` : confirmé. Le dernier caller de production ayant disparu, conserver un helper exporté de scan namespace-wide était inutile et pouvait encourager le pattern que cette PR supprime. `d6ac0fa29c` retire cette API de `keys.go` et remplace ses seuls usages par des helpers strictement test/benchmark-locales.

Validation : `go test ./internal/storage/readstore ./internal/application/indexbuilder` PASS ; `git diff --check` PASS. Une nouvelle boucle canonique va relire ce SHA.

gfyrag:
Sixième passe challengée : finding confirmé et corrigé dans `ce61d3357e`.

Le commentaire public de `CHECK_STORE_ERROR_TYPE_REVERSE_MAP_ORPHAN` décrit désormais le layout field/version/entity, la purge field-bounded par `DeleteRange`, et le rôle conservé du checker contre la corruption ou une rupture de lifecycle. Les semantics d’agrégation et de coexistence des versions restent inchangées.

Conformément au workflow protobuf, `just generate-proto` a été exécuté immédiatement. Le seul fichier généré modifié est `internal/proto/servicepb/bucket.pb.go`, avec le même commentaire ; aucun champ, numéro, descripteur ni wire format ne change.

Validation : `go test ./internal/application/check ./internal/proto/servicepb` PASS ; `git diff --check` PASS. Nouvelle boucle canonique sur ce SHA.

docs/technical/architecture/subsystems/checker/checker.md
```diff
@@ -52,8 +52,8 @@ Notes:
...
-  - **The registry alone decides the verdict**: a row is an orphan when its field has no registered index. Both `DropIndex` and `RemovedMetadataFieldType` now purge all three metadata-index limbs, so an aligned residual row is a missed purge. The audit-replayed schema is diagnostic only: a declared orphan points to `DropIndex`, while an undeclared orphan points to `RemovedMetadataFieldType`. It must remain replay-derived, never read from stored `LedgerInfo`.
+- **`compareReverseMapOrphans` is the only pass that reads a peer store**, and the only exception to invariant #8's main-store scope. Reverse-map keys use `[field\x00][version][entity]`, so `purgeReverseMapForKey` now removes a field across all versions with one bounded `DeleteRange`, like the `0x01` forward index and `0x02` existence limbs. The pass remains as defense against read-store corruption or a broken lifecycle path; cheaper atomic maintenance does not make those states impossible.
```
NumaryBot, line 55: 🟠 **[major] Update the linked audit-chain description**

The authoritative linked page `docs/technical/architecture/subsystems/checker/audit-chain.md:5` still says the reverse map is entity-first, requires row-by-row point deletion, and that findings identify which purge missed rows. This contradicts the new field-first `DeleteRange` behavior documented here and can mislead future checker work; update it per the documentation-maintenance rule in [AGENTS.md:82-86](../../../../../AGENTS.md#L82-L86).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1904#issuecomment-5542402681

gfyrag:
Septième passe challengée : les trois remarques sont confirmées, avec une qualification importante sur la troisième, et corrigées dans `981712a58e`.

1. La table du checker ne présente plus le lifecycle propriétaire comme la cause prouvée d’un purge miss ; elle décrit une classification neutre.
2. Les commentaires de `renderReverseMapEntity` et de la borne hexadécimale reflètent désormais le vrai contrat : une longueur tx différente de 8 produit `ErrReverseMapKeyEntityID`, et 64 octets sont seulement un échantillon diagnostic, pas un header/composant complet.
3. La PR divulgue maintenant la limite inhérente de détection des NUL : les re-splits invalides sont rejetés par les invariants de production, mais une séquence qui se recompose entièrement en composants valides reste ambiguë sans longueur authentifiée ni checksum. C’est documenté comme risque résiduel du format readstore v3 jetable, pas comme garantie de détection totale.

Validation ciblée : `go test ./internal/application/check` PASS ; `gofmt` et `git diff --check` PASS. La CI complète repart sur ce SHA. Les tentatives locales de gate complet ont été arrêtées uniquement par la capacité disque de la machine (`no space left on device`), avec `ROOT_UNCHANGED=PASS`; aucune erreur fonctionnelle n’a été trouvée.

docs/technical/architecture/subsystems/checker/checker.md
gfyrag: Confirmed against ce61d335: audit-chain.md still described the former namespace scan/point-delete path and attributed residue to a specific missed purge. Fixed in ef64abc7ba63b90ae771fe162b1026013b550fac: field/version/entity keying, field-bounded DeleteRange, checker retained for corruption or lifecycle violations, and neutral lifecycle classification. I inspected the rest of the page and the linked checker/indexer/audit-state docs; no other stale reverse-map passage remains.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1904#issuecomment-5542402681

gfyrag:
Huitième passe challengée : le finding P3 est rejeté, sans changement de code.

La duplication mécanique du petit encodeur legacy existe, mais le risque contractuel allégué est déjà traité exactement par l’alternative proposée dans la review : les commentaires croisés réciproques aux lignes 68–72 des deux benchmarks nomment l’autre suite, indiquent que les fixtures restent package-locales, que workloads et largeurs d’ID peuvent différer, et que seules les comparaisons A/B intra-suite sont valides.

Cette indépendance est intentionnelle : readstore mesure lookup/rewrite/purge sur Pebble (comptes 10 chiffres, versions 7/8), tandis qu’indexbuilder mesure fold/commit/overlay (comptes 12 chiffres, version 1 et cardinalités hot/seed propres). Chaque baseline et son candidat partagent localement les mêmes générateurs d’entité/champ/version ; aucun ratio cross-suite n’est revendiqué.

Extraire l’encodeur entre deux packages imposerait de promouvoir un helper historique benchmark-only dans un package importable ou d’ajouter une infrastructure de test dédiée. Cela augmenterait le couplage sans renforcer la comparaison mesurée. L’invariant commun — `[ledger][namespace][entity][NUL compte][version][field]` — reste explicite dans les deux helpers.

Conclusion : remarque de maintenabilité recevable comme préférence, mais aucun défaut ni affaiblissement de la preuve benchmark ne subsiste; zéro modification justifiée.

gfyrag:
CI complète verte sur `ef64abc7ba` et reviews automatisées traitées (`NumaryBot: No findings`; dernier point Shipfox challengé et rejeté avec justification ci-dessus).

@Azorlogh @paul-nicolas @flemzord — la PR est prête pour approbation humaine.

gfyrag:
Reprise complète sur `d4f2e1564db67d696bccefc703ac3610ee359b37`.

- Rebase sans conflit sur `release/v3.0` à `afdd58395fd4689c624286893c2c618e38db8148`, dans un worktree détaché. Les dix commits existants sont inchangés selon `git range-diff`. Push effectué avec un lease explicite sur l'ancien HEAD `ef64abc7ba`.
- Les huit synthèses Shipfox, toutes les reviews et l'unique fil inline NumaryBot ont été relus. Les corrections précédentes sont conservées ; aucun finding restant. Les pushbacks restent fondés : le mirror v2 peut importer l'ID transaction 0 (`translator.go::translateNewTransaction` et ses tests), et les benchmarks documentent leurs comparaisons A/B indépendantes aux lignes 68–72 des deux suites.
- Une dernière phrase de `docs/technical/architecture/subsystems/indexer/indexes.md` attribuait encore les orphelins à une purge manquée. Corrigée dans `d4f2e1564d` : le schéma rejoué classe le lifecycle sans prétendre identifier la cause. Aucun autre changement de code ajouté pendant cette reprise.
- Revue indépendante de ce candidat et des interactions avec les changements d'index mergés : zéro finding. Aucun shim ni migration v3 ajouté.
- Validation PASS dans Nix : tests ciblés `readstore`, `indexbuilder`, `check` sous `-race` ; `bash scripts/agent-check` ; `AI_REVIEW_BASE_SHA=afdd58395fd4689c624286893c2c618e38db8148 bash scripts/agent-check-pr`. Ce dernier a exécuté la normalisation complète, le lint, tous les tests racés du module principal et les E2E business/cluster. `GOFLAGS=-p=2 GOMAXPROCS=2` limite seulement le parallélisme ; aucune gate n'a été supprimée. Diff final propre, sans dérive générée.

Le premier run complet a reçu un signal externe pendant une forte pression disque ; la relance ci-dessus a terminé intégralement. Les caches partagés ont été conservés et les temporaires du validateur nettoyés automatiquement.

La base a ensuite reçu `b2b7e7d033` (#1940), qui ne modifie que le harness Schemathesis ; son diff a été vérifié sans interaction avec cette PR. La CI du nouveau HEAD est maintenant suivie. La règle de protection exige une approbation CODEOWNER ; aucune fusion effectuée.

gfyrag:
État final sur `d4f2e1564db67d696bccefc703ac3610ee359b37` : validation locale complète PASS et [CI entièrement verte](https://github.com/formancehq/ledger/actions/runs/34235507599), y compris les deux checks Codecov (15 checks réussis, 2 jobs de publication non applicables).

NumaryBot : aucun finding. Zéro fil non résolu ; les huit reviews Shipfox ont été vérifiées et traitées, avec les preuves détaillées dans le commentaire précédent. Worktree propre.

GitHub indique `MERGEABLE` ; la fusion reste protégée par l'approbation CODEOWNER requise (`REVIEW_REQUIRED`). La PR est prête pour cette approbation. Aucune fusion effectuée.

internal/storage/readstore/keys_test.go
```diff
@@ -153,6 +154,49 @@ func TestParseReverseMapKey_TransactionRoundTrip(t *testing.T) {
...
+			versionPrefix := ReverseMapVersionPrefix(dal.NewKeyBuilder(), "main", tt.namespace, "status", 2)
+			target := tt.key(dal.NewKeyBuilder(), "status", 2)
+			otherVersion := tt.key(dal.NewKeyBuilder(), "status", 3)
+			otherField := tt.key(dal.NewKeyBuilder(), "team", 2)
+
+			require.True(t, bytes.HasPrefix(target, fieldPrefix))
+			require.True(t, bytes.HasPrefix(target, versionPrefix))
+			require.True(t, bytes.HasPrefix(otherVersion, fieldPrefix))
+			require.False(t, bytes.HasPrefix(otherVersion, versionPrefix))
+			require.False(t, bytes.HasPrefix(otherField, fieldPrefix))
```
paul-nicolas, line 208: [P2][blocking] Pin the reverse-map field prefix terminator with a prefix-sibling field

`ReverseMapFieldPrefix` (keys.go:586) bounds the purge range solely through the `\x00` that `PutStringNull(metadataKey)` emits. Without it the range widens to every field whose name starts with the removed one — and nothing in the suite detects that.

Every "kept" field in the tests is lexically disjoint from the removed one: `"team"` vs `"status"` here at line 189 and in `process_metadata_field_removal_test.go:350`, `"other"` vs `"score"` in the check package. All five assertions at lines 191-195 hold with or without the terminator, so none of them reaches the branch.

Proven by mutation at this head, in a detached worktree. Replacing `PutStringNull` with `PutString` in `ReverseMapFieldPrefix`:

```
ok  	.../internal/application/indexbuilder	8.524s
ok  	.../internal/storage/readstore	5.224s
ok  	.../internal/application/check	9.988s
```

All three packages green. A throwaway probe that seeds `"status"` and `"status2"` for one account and purges `"status"` passes unmutated and fails mutated with `pebble: not found` on the `status2` row — the mutated tombstone range-deletes the entire sibling field, through a path shared by `RemovedMetadataFieldType` and `DropIndex` (`index_config.go:559`).

The production code is correct; the PR's central safety property just has no regression trigger, and `docs/technical/contributing/testing.md` requires assertions to uniquely identify the intended branch. `compareReverseMapOrphans` would not catch the regression either — it detects residual rows, not missing ones.

Fix: add a prefix-sibling field to the kept-rows set of `TestPurgeReverseMapForKeyIsFieldBoundedAcrossVersions`, and a `require.False(t, bytes.HasPrefix(tt.key(kb, "status2", 2), fieldPrefix))` case here. Resolved when that mutation makes at least one test fail.

Note: `ReverseMapVersionPrefix` needs nothing — dropping its terminator misaligns the fixed-width version block rather than widening the range, and it already fails `TestReverseMapPrefixesBoundFieldAndVersion` (line 192), `TestProcessSchemaRewriteIsFieldVersionBoundedAndPersistsCursor` and `TestInitIndexConfig_PurgesOrphanVersionsOnBoot`.

internal/storage/readstore/keys.go
```diff
@@ -281,46 +276,43 @@ func ParseReverseMapKey(key []byte) (ParsedReverseMapKey, error) {
...
+	if err := invariants.ValidateMetadataKey(parsed.MetadataKey); err != nil {
+		return ParsedReverseMapKey{}, fmt.Errorf("%w: %w", ErrReverseMapKeyMetadataKey, err)
+	}
+
+	rest = rest[metadataEnd+1:]
+	if len(rest) < 4 {
+		return ParsedReverseMapKey{}, fmt.Errorf("%w: version block shorter than 4 bytes", ErrReverseMapKeyTruncated)
 	}
 
 	parsed.Version = binary.BigEndian.Uint32(rest[:4])
```
paul-nicolas, line 304: [P2][non-blocking] Reject `Version == 0` here — it is unwritable by contract, and it is the whole reachable mis-split window

`parsed.Version` is read raw and never validated, yet no write path can produce 0: `effectiveCurrentVersion` (builder.go:220-231) returns `pending` or the literal `1`; the mirrored write is guarded by `pending != 0` (builder.go:290, 325); backfill fails loudly when `pendingVersion == 0` (backfill.go:596-604); and the non-V wrappers hardcode `..., 1` (keys.go:163, 183). Accepting 0 at the documented single chokepoint is invariant #7 territory.

It is not academic. Sweeping hand-built account rmap bytes whose field carries an embedded NUL — the residual limitation the PR body discloses — the acceptance window opens at real version 45 and every accepted decode in the realistic range carries version 0:

```
v=44  REJECTED: malformed entity id: ledger account address must contain only ...
v=45  ACCEPTED -> field="statu" version=0 entity="-users:1"
v=48  ACCEPTED -> field="statu" version=0 entity="0users:1"
v=301 ACCEPTED -> field="statu" version=1 entity="-users:1"

realistic v in [1,255]: accepted-with-version-0 = 64, accepted-with-version!=0 = 0
```

So for a single-NUL field this is a **complete closure** of the reachable window, not the narrowing it first looks like — residual ambiguity would need real version >= 301, i.e. an index retyped 301+ times. Two and three embedded NULs are already fully rejected in that range, and the transaction namespace is immune because the exactly-8-byte entity check kills every shift. The issue is account-only.

Impact today is read-only but real: the checker buckets by `{ledger, namespace, metaKey}` with version deliberately ignored (`reverse_map_orphans.go:155, 341-345`), so a corrupted row in that window is filed under `"statu"` instead of `"statu\x00"` and is silently missed whenever the shorter field is a registered index. Nothing mutates on the strength of the wrong bucket, hence non-blocking. Reachable only from physical storage damage, since `ValidateMetadataKey` keeps an embedded NUL out of every input path — which is precisely the class this checker exists to detect.

The fix is proven safe: adding a sentinel and `if parsed.Version == 0 { ... }` right after this line keeps readstore, indexbuilder and check green and `go build ./...` clean — no test asserts that a version-0 key parses. The "detection is best-effort" note above can then carry its bound rather than reading as unqualified.

paul-nicolas (commented):

**DECISION: FINDINGS** — 1 blocking, 7 non-blocking. Head reviewed: `d4f2e1564db67d696bccefc703ac3610ee359b37`.

The re-keying itself is sound and I could not break it. Verified against real code and by mutation, not accepted on the strength of the prose:

- **Bounds are provably exact.** `invariants.ValidateMetadataKey` restricts fields to `[a-zA-Z0-9._:/-]`, so no field can embed the `\x00` separator and `[field\x00, field\x01)` cannot reach a prefix sibling. `IncrementBytes` carries correctly out of a `0xFFFFFFFF` version block into the field terminator, and the over-reach window cannot contain a real key. `KeyBuilder.Snapshot()`/`Build()` both copy, so the reused `kb` cannot alias a live iterator bound.
- **Same-batch ordering is correct in both directions, proven against real Pebble.** Committed probes: range tombstone then point write on a covered key -> value present after commit; point write then covering tombstone -> `ErrNotFound`. That matches `ReverseMapOverlay`'s exact-map-first-then-ranges resolution, which is order-exact because `DeleteReverseMapRange` eagerly nils covered overlay entries. Confirmed in Pebble itself: `memTable.apply` assigns a monotonic seqNum per batch entry, so a later RANGEDEL outranks earlier point writes.
- **Rewrite read path loses nothing.** The narrowed `(field, v_current)` iterator yields exactly the row set the removed `MetadataKey`/`entryVersion` filters produced; the authoritative old value still comes from the FSM main store via `fetchStoredMetadataValue`, and the DEL event's old encoding still from the v_pending rmap row. Turning the mismatch into an `invariant:` error is right — it is unreachable inside a bounded range.
- **Nothing was dropped with the 219 deleted lines.** The snapshot/iterator, dedup set, overlay pass and truncation guards all existed to enumerate rows the range tombstone now covers wholesale; the truncation guards protected a partial-scan failure mode a range delete does not have. No metric, event emission or live-row guard was removed.
- **No invariant #2/#3 exposure** — the indexbuilder folds committed logs outside Raft apply, and the change moves in the safe direction (a range tombstone is a pure function of `(ledger, ns, field, version)`, where the old purge depended on Pebble iteration over live data).
- **Docs are accurate.** Full `docs/` sweep found no stale entity-first prose in any authoritative doc; the key-layout diagram matches `keys.go` byte for byte; `indexer.md`'s overlay paragraph states the real execution sequence rather than the intended end state.
- **Benchmarks are hermetic and measure what they claim** — `ResetTimer` after fixture construction everywhere, `b.Loop` accounting correct, both layout arms use exactly 104-byte account keys so the comparison isolates ordering rather than key size, and the op-count labels trace correctly through `ReplaceMetadataIndexV`.

**Re-review:** the one prior NumaryBot finding (`audit-chain.md` still entity-first) is **fixed** at this head — by `23e92e5c1`, not `ef64abc7` as stated in the thread. Not re-filed. 0 still-valid, 0 outdated.

**One finding was dropped after verification.** Three of the five reviewers independently flagged that `gcReverseMapVersion` (`gc.go:66`) and `DeleteLedgerIndexes` (`delete_ledger.go:56`) queue reverse-map range tombstones straight onto the raw session, bypassing the new overlay mirror, and hung it on the claim that this PR adds the "a reverse-map write that is not mirrored is unrepresentable" comment. **That escalation is factually wrong** — the sentence is unchanged from `afdd58395`, the `gc.go` path is unreachable (every caller flushes before any rmap read; `purgeOrphanVersions` uses its own batch with no bound `wb`), and the `DeleteLedgerIndexes` case reproduces **byte-identically on the merge base**, in a file this PR does not touch. Not charged to this PR.

That last one is worth a separate ticket though, and it is worse than the reviewers thought: a DeleteLedger followed by a same-name recreate plus `SavedMetadata` in one fold batch makes `ReplaceMetadataIndexV`'s `bytes.Equal` early return skip the rmap PUT **and** the ADD event, so the new incarnation's row is silently absent from the index — data loss, not an orphan event. Reproduced identically on head and base.

### P3 (not posted inline)

- `internal/storage/readstore/keys.go:304` — the entity-decode `switch` lost the `default` that returned `ErrReverseMapKeyNamespace` when the namespace check was hoisted; it now falls through to `return parsed, nil` with a nil `EntityID`. Unreachable today, but `reverse_map_orphans.go:546` documents that it relies on a populated entity per namespace, so a third namespace would hand the checker a nil instead of an error (invariant #7).
- `internal/storage/readstore/write_batch_reverse_map_overlay_test.go:39` — the new test asserts only the in-memory overlay model; the batch is never committed, so nothing pins the doc claim that the model matches Pebble's batch sequencing. The opposite direction is covered against real Pebble by `TestPurgeReverseMapForKeyIsFieldBoundedAcrossVersions`; tombstone-then-write across a commit is not.
- `docs/technical/architecture/subsystems/checker/checker.md:239` — the bar for admitting a new peer-store checker pass changed from a falsifiable test ("a divergence no other mechanism detects") to "a similarly explicit integrity argument". The keying change correctly invalidated the old justification, but this is the only place invariant #8's peer-store scope is defined, and the replacement cannot be failed. Same substitution in `audit-vs-technical-state.md:243` and `usage/usagebuilder.md:130`.
- `docs/technical/architecture/subsystems/indexer/indexer.md:301` — stale `backfill.go:NNN` refs in the section this PR rewrites: step 3's `ReplaceMetadataIndexV (backfill.go:781-787)` points at step 2's code (real call is 827-831); `processSchemaRewrite (545-887)` is really 581-938; `tryCommitScanCompleteSwitch (901-948)` is really 952+. Pre-existing at `afdd58395` and the PR's net effect is an improvement — flagged only because it edits the same list.
- `internal/application/indexbuilder/reverse_map_keying_fold_bench_test.go:199` — `overlay_hit_ratio` is a hardcoded literal reported through `b.ReportMetric`, indistinguishable from a measurement in benchstat output. Correct only because `DefaultBatchSize == 1000`; nothing ties it to that constant.
- `internal/storage/readstore/reverse_map_keying_bench_test.go:436` — the entity-first arm does `bytes.Clone` + `string(...)` per row (480,140 allocs/op vs 99) though the scan only needs an equality test, inflating the measured RewriteScan gap to 29.5x against an intrinsic row-count ratio of 24x. The conclusion survives comfortably; the quoted ratio is ~20% overstated relative to the keying effect.

---
Reviewers: Claude (4 area agents + 3 adversarial verifiers, all mutation-based) and Codex (0 findings). 5 findings entered verification, 2 confirmed and corrected, 1 rejected. `go test ./internal/storage/readstore/ ./internal/application/indexbuilder/ ./internal/application/check/` green at this head.

internal/storage/readstore/keys_test.go
gfyrag: Fixed in a5a1e85cfc. Both account and transaction cases now distinguish `status` from the prefix sibling `status2` in the prefix helper and committed purge-survivor tests, including sibling rows at versions 1 and 3. The existing unrelated-field, other-ledger, other-namespace and in-flight-write triggers remain. Mutation verified: replacing `PutStringNull` with `PutString` specifically in `ReverseMapFieldPrefix` makes both regressions fail; restoring production code makes them pass under `-race`.

internal/storage/readstore/keys.go
gfyrag: Fixed in 9ecd0f611b. `ParseReverseMapKey` now rejects encoding version zero with `ErrReverseMapKeyVersion`. Parser regressions cover both namespaces and the embedded-NUL/version-45 account example; checker coverage uses an actually registered `statu` field with both aligned and lagging peer cursors. Removing the guard makes the intended parser and checker regressions fail. A separate positive test preserves transaction ID zero at a nonzero encoding version. Docs still acknowledge the residual ambiguity when corrupted bytes re-split entirely into valid components (including a nonzero version); this is not claimed to authenticate the key format.

gfyrag:
Addressed the review on candidate `4f0862f0f0f011d2a64ea827b7be5139a787b3a4`, rebased onto `8a2a7c85bc3122f64f637edd0720c3e94618ea5f`.

- **Prefix-sibling regression:** `a5a1e85cfc` adds `status2` isolation in both prefix and committed purge tests, for accounts and transactions. Removing the field terminator makes both regressions fail.
- **Zero encoding version:** `9ecd0f611b` rejects it and exercises the checker with a registered field at aligned/lagging cursors. Removing the guard makes those regressions fail. Transaction ID zero remains supported; nonzero-version re-split ambiguity remains documented.
- **Defensive entity switch:** `9ecd0f611b` restores the fail-loud namespace default.
- **Committed overlay parity:** `3afd0a113e` retains the overlay test and adds real Pebble commit checks for tombstone-then-write and tombstone-then-write-then-tombstone, in both namespaces.
- **Peer-store checker admission:** `9ecd0f611b` restores a concrete divergence/effect/regression requirement in the canonical checker contract. The usagebuilder text still describes its exclusion and the reverse-map exception; it does not establish a weaker admission rule.
- **Stale documentation offsets:** `4f0862f0f0` replaces offsets with named symbols in the rewritten indexer section.
- **Overlay metric:** `325e8b8702` derives it from the batch size and fixture entity count and renames it `analytical_overlay_hit_ratio`. A smoke run reports 0.9 for both layout arms; that single iteration is not a new timing comparison.
- **Historical scan allocations:** retained deliberately in `325e8b8702`. The production parser at the exact base clones account/transaction IDs and converts the metadata key to a string (`keys.go` lines 296, 303 and 323) before the caller filters field/version. Removing those allocations would benchmark an optimized hypothetical baseline. The benchmark now explicitly labels its scope as candidate selection, not the full rewrite or a pure key-ordering effect; the row-count reduction is 240,000 to 10,000 for the stated fixture.

Validation: targeted race tests for readstore/indexbuilder/check; both mutation experiments with source restored and focused tests rerun; protobuf regeneration; canonical `agent-check` and `agent-check-pr` against the exact base (all selected normalization, race and business/cluster E2E gates retained). Independent review of the five response commits found no actionable defects.

One qualification on the excluded delete/recreate scenario: the normal processor retains `LedgerInfo.DeletedAt` on deletion and `processCreateLedger` rejects an existing tombstoned name with `ErrLedgerDeleted`. A direct fold probe can demonstrate behavior for that synthetic log sequence, but it does not establish a reachable production data-loss trigger. No unrelated lifecycle change is included here.

gfyrag:
Addressed both non-blocking documentation items from the review of `4f0862f0`:

- `e759c79f54`: the `handleDroppedIndexLog` comment now explains that the version high-water prevents keyspace reuse even if corruption introduces an old row outside the atomic cleanup path. Removed the obsolete partial-purge rationale and repaired the line wrapping.
- `bbd559a229`: marked `agent-reference-legacy.md` explicitly as a historical snapshot, with current `AGENTS.md`/subsystem docs taking precedence throughout. Added an EN-1957 warning covering the obsolete layout, scan/point-delete purge, lifecycle attribution and zero-version wording, with links to the current index/checker contracts. The historical body is preserved as intended by `AGENTS.md`.

The review’s claim that this is the only agent-instructions document defining the reverse-map exception is not accurate: `AGENTS.md` and `agent-context.md` already establish the current authority and the snapshot’s subordinate status. The new warning makes that status clear to readers who open the snapshot directly.

Candidate: `bbd559a229af4027e1afe1c7bdfbcf54e273da26`, base `8a2a7c85bc3122f64f637edd0720c3e94618ea5f`. The delta contains only comments/documentation. Canonical `agent-check` and `agent-check-pr` pass with all selected gates retained; earlier mutation and focused race evidence remains applicable to the unchanged implementation.

gfyrag:
Reviewed both comments on `bbd559a229` against the current helpers and the earlier discussion.

1. **Benchmark fixture extraction: not taken.** The mechanical duplication is real and extraction is a reasonable style preference, but this repeats the tradeoff already discussed in [the earlier response](https://github.com/formancehq/ledger/pull/1904#issuecomment-5546890192). The legacy encoders deliberately freeze the former entity-first format; future production layout changes must not be applied to that baseline. Both field-first arms already call production key builders (`reverseMapFoldBenchmarkKey` → `metadataReverseMapKeyV`; `reverseMapBenchmarkKey` → `AccountReverseMapKeyV` / `TransactionReverseMapKeyV`). There is therefore no requirement to update three current-format encoders. The fixture generators use different widths, types and cardinalities for their respective workloads. A shared testutil encoder is possible, but no current divergence or invalid comparison warrants adding that package in this PR.

2. **Namespace-prefix test helpers: no current change needed.** Both helpers are confined to `_test.go` files. The indexbuilder helper accepts an arbitrary ledger to inspect purge survivors; the benchmark helper uses its fixed fixture ledger to model the historical full-namespace scan. Neither restores `ReverseMapPrefix` to the production API. Consolidation if more consumers appear is a future maintenance option, not an existing defect.

No code changed. The exact candidate remains `bbd559a229af4027e1afe1c7bdfbcf54e273da26`; its local canonical validations and CI are green. Both review surfaces have been checked and no unresolved inline thread remains. Human CODEOWNER approval is still required before merge.

flemzord (approved):
Reviewed field/version range isolation, reverse-map maintenance, same-batch ordering, and the fixes for previous review findings. No actionable findings. Targeted readstore, indexbuilder, and checker tests pass with -race; CI is green.

PR #1905, opened 2026-09-04
fix(replication): close HTTP exporter response bodies

imtiazmangerah:
Report and reproduction: https://github.com/orgs/formancehq/discussions/2006

## Problem

`Driver.Accept` never reads or closes the response body from `httpClient.Do`, so keep-alive
connections are never reused. Each push leaks one connection, 3 goroutines and ~27 KB of heap for
the life of the worker.

## Change

- Drain `rsp.Body` in a deferred block and close it. The drain is bounded in bytes (64 KiB) and in
  time (5 s, by cancelling the request context from a timer, which interrupts the body read). A
  larger, endless or stalled body is closed unread, so the transport drops that connection instead
  of stalling the worker on a misbehaving exporter.
- `TestHTTPDriverReusesConnections`: 20 pushes against an `httptest` server must open exactly one
  connection. Fails on `main` (20), passes with the fix.
- `TestHTTPDriverBoundsResponseDrain`: the server streams an endless body; `Accept` must return and
  the next push must succeed. Fails without the byte cap, passes with it.
- `TestHTTPDriverDrainIsTimeBounded`: the server sends a few bytes then stalls; `Accept` must return
  and the next push must succeed. Fails without the timer, passes with it.

## Reproduction

From a clone of `main`, using only this branch:

```bash
git fetch https://github.com/imtiazmangerah/ledger.git fix/http-exporter-close-response-body
PKG=./internal/replication/drivers/http

# analyzer and the new test against main (test file only)
go run github.com/timakin/bodyclose@latest $PKG
#   internal/replication/drivers/http/driver.go:45:29: response body must be closed
git checkout FETCH_HEAD -- $PKG/driver_test.go
go test $PKG -run TestHTTPDriverReusesConnections -count=1
#   expected: 1   actual: 20   FAIL

# same against the branch
git checkout FETCH_HEAD
go run github.com/timakin/bodyclose@latest $PKG        # no output
go test $PKG -run TestHTTPDriverReusesConnections -count=1
#   ok
```

Measured with 200 pushes (probe in the report linked at the top): server connections 200 → 1,
goroutines
+600 → +3, heap in use +5.5 MB → +0.5 MB.

`bodyclose` is built into golangci-lint and reports nothing else in the repository once this is
applied; enabling it in `.golangci.yml` would prevent a recurrence.

The driver is identical on `release/v2.4`; the commit cherry-picks cleanly if you want a backport.

Triaged with Claude Fable 5.1 (Anthropic, model id `claude-fable-5-1`, via Claude Code). The
finding comes from a production symptom. Every number and the fix were reproduced by running the
commands above and reviewed and verified by the author.

### DIFF
--- a/internal/replication/drivers/http/driver.go
+++ b/internal/replication/drivers/http/driver.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"time"
 
 	"github.com/pkg/errors"
 
@@ -17,13 +18,19 @@ import (
 	"github.com/formancehq/ledger/internal/replication/drivers"
 )
 
-// maxResponseDrainBytes caps how much of an exporter's response body Accept reads
-// before closing it; ingest endpoints answer with a few bytes at most.
-const maxResponseDrainBytes = 64 << 10
+const (
+	// maxResponseDrainBytes caps how much of an exporter's response body Accept reads
+	// before closing it; ingest endpoints answer with a few bytes at most.
+	maxResponseDrainBytes = 64 << 10
+	// responseDrainTimeout caps how long Accept waits for those bytes. A stalled body
+	// costs the connection, not the worker.
+	responseDrainTimeout = 5 * time.Second
+)
 
 type Driver struct {
-	config     Config
-	httpClient *http.Client
+	config       Config
+	httpClient   *http.Client
+	drainTimeout time.Duration
 }
 
 func (c *Driver) Stop(_ context.Context) error {
@@ -45,6 +52,8 @@ func (c *Driver) Accept(ctx context.Context, logs ...drivers.LogWithLedger) ([]e
 	if err != nil {
 		return nil, err
 	}
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 	req = req.WithContext(ctx)
 
 	rsp, err := c.httpClient.Do(req)
@@ -52,10 +61,13 @@ func (c *Driver) Accept(ctx context.Context, logs ...drivers.LogWithLedger) ([]e
 		return nil, err
 	}
 	defer func() {
-		// Drain (bounded) and close so the underlying connection is released back to
-		// the transport instead of leaking one socket per push. A body larger than
-		// the bound is closed unread: the transport then drops that connection, which
-		// is preferable to blocking the worker on a misbehaving exporter.
+		// Drain and close so the underlying connection is released back to the
+		// transport instead of leaking one socket per push. The drain is bounded in
+		// bytes and in time: cancelling the request context interrupts a stalled body
+		// read, and the transport then drops that connection, which is preferable to
+		// blocking the worker on a misbehaving exporter.
+		stop := time.AfterFunc(c.drainTimeout, cancel)
+		defer stop.Stop()
 		_, _ = io.Copy(io.Discard, io.LimitReader(rsp.Body, maxResponseDrainBytes))
 		_ = rsp.Body.Close()
 	}()
@@ -69,8 +81,9 @@ func (c *Driver) Accept(ctx context.Context, logs ...drivers.LogWithLedger) ([]e
 
 func NewDriver(config Config, _ logging.Logger) (*Driver, error) {
 	return &Driver{
-		config:     config,
-		httpClient: http.DefaultClient,
+		config:       config,
+		httpClient:   http.DefaultClient,
+		drainTimeout: responseDrainTimeout,
 	}, nil
 }
 
--- a/internal/replication/drivers/http/driver.go
+++ b/internal/replication/drivers/http/driver.go
@@ -17,6 +17,10 @@ import (
 	"github.com/formancehq/ledger/internal/replication/drivers"
 )
 
+// maxResponseDrainBytes caps how much of an exporter's response body Accept reads
+// before closing it; ingest endpoints answer with a few bytes at most.
+const maxResponseDrainBytes = 64 << 10
+
 type Driver struct {
 	config     Config
 	httpClient *http.Client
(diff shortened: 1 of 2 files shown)


internal/replication/drivers/http/driver.go
```diff
@@ -46,6 +47,12 @@ func (c *Driver) Accept(ctx context.Context, logs ...drivers.LogWithLedger) ([]e
 	if err != nil {
 		return nil, err
 	}
+	defer func() {
+		// Drain and close so the underlying connection is released back to the
+		// transport instead of leaking one socket per push.
+		_, _ = io.Copy(io.Discard, rsp.Body)
```
NumaryBot: 🟠 **[major] Bound response draining before returning**

When an exporter returns an unbounded or very large response body, this `io.Copy` waits for EOF and prevents `Accept` from returning; the production context has no deadline and is only canceled after `Accept` completes. This regresses the previous immediate return into a stuck replication worker, so drain only a bounded amount before closing the body.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1905#issuecomment-5541403737

internal/replication/drivers/http/driver.go
```diff
@@ -46,6 +51,14 @@ func (c *Driver) Accept(ctx context.Context, logs ...drivers.LogWithLedger) ([]e
 	if err != nil {
 		return nil, err
 	}
+	defer func() {
+		// Drain (bounded) and close so the underlying connection is released back to
+		// the transport instead of leaking one socket per push. A body larger than
+		// the bound is closed unread: the transport then drops that connection, which
+		// is preferable to blocking the worker on a misbehaving exporter.
+		_, _ = io.Copy(io.Discard, io.LimitReader(rsp.Body, maxResponseDrainBytes))
```
NumaryBot, line 71: 🔴 **[blocker] Bound response draining before returning**

The byte limit does not bound read time: if an exporter sends headers and then stalls or trickles fewer than 64 KiB, this deferred `io.Copy` waits indefinitely. In the normal pipeline path the request context is canceled only after `Accept` returns, so one such response can still block replication; draining must also be time-bounded or otherwise interruptible.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1905#issuecomment-5541403737

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1905#issuecomment-5541403737

internal/replication/drivers/http/driver.go
```diff
@@ -40,12 +52,25 @@ func (c *Driver) Accept(ctx context.Context, logs ...drivers.LogWithLedger) ([]e
...
 	}
+	defer func() {
+		// Drain and close so the underlying connection is released back to the
+		// transport instead of leaking one socket per push. The drain is bounded in
+		// bytes and in time: cancelling the request context interrupts a stalled body
+		// read, and the transport then drops that connection, which is preferable to
+		// blocking the worker on a misbehaving exporter.
+		stop := time.AfterFunc(c.drainTimeout, cancel)
+		defer stop.Stop()
+		_, _ = io.Copy(io.Discard, io.LimitReader(rsp.Body, maxResponseDrainBytes))
```
flemzord, line 71: [P1] Preserve header-time completion of Accept

When an exporter flushes 202 (or 500), writes one byte, and then stalls, this deferred read postpones the success/error until the five-second timer fires. Comparing the base and this head locally, 202 changed from 7.19 ms to 5.007 s and 500 from 1.60 ms to 5.003 s, with identical return values. This delays batch completion (batcher.go:59-72), checkpoint advancement (pipeline.go:153-155), and the start of retry backoff (pipeline.go:135-140). The timer fixes the previously reported indefinite wait, but does not preserve the old behavior. Under the maintainer requirement that no existing behavior may change, this remaining latency regression blocks acceptance. Please preserve completion without waiting for the response body and add comparative stalled-body coverage for both success and error responses.

internal/replication/drivers/http/driver.go
```diff
@@ -40,12 +52,25 @@ func (c *Driver) Accept(ctx context.Context, logs ...drivers.LogWithLedger) ([]e
...
+		_ = rsp.Body.Close()
```
flemzord, line 72: [P2] Explicit compatibility decision required for the network lifecycle change

For twenty direct Accept calls with a short complete HTTP/1 response, the connection-reuse test opens 20 connections on the base and 1 on this head; I reproduced both outcomes. The driver now consumes and closes response bodies instead of leaving them unread, and interrupts unfinished responses at the read/time limit. These are exporter-visible changes in response consumption and connection reuse/closure. This is the intended resource-management fix, not an intrinsic correctness defect. However, it independently conflicts with the maintainer requirement that no existing behavior may change, even beneficially; removing the synchronous wait alone would not satisfy that requirement. Please explicitly resolve this scope conflict with the maintainer before accepting the change, documenting the permitted network-lifecycle difference if an exception is granted.

flemzord (changes requested):
Requesting changes under the maintainer requirement that existing behavior must remain unchanged. The head is still 60423f3e5826146081d6f811486bd1ad9346f2f5.

- [Synchronous response draining](https://github.com/formancehq/ledger/pull/1905#discussion_r4005867364): an exporter that flushes 202 or 500 and then stalls used to let Accept return within milliseconds; this head takes approximately five seconds. Return values are unchanged, but batch completion, checkpoint advancement, and retry scheduling are delayed. This is a reproduced latency regression, even though the previously reported indefinite wait is fixed.
- [Network lifecycle](https://github.com/formancehq/ledger/pull/1905#discussion_r4005867375): twenty direct calls now reuse one connection instead of opening twenty; response bodies are also consumed and closed. This is the intended resource-management improvement, not an intrinsic correctness defect, but it remains an acceptance blocker under the stated compatibility constraint until the maintainer explicitly permits that difference.

Please address the completion-latency regression and resolve the permitted scope of network behavior changes. Local replication tests pass with the race detector; that does not establish behavioral equivalence. Existing inline findings are referenced above rather than duplicated.

PR #1906, opened 2026-09-04
fix(EN-1958): fail-stop force removal on WAL failure

gfyrag:
## Summary

- fail-stop a node when ForceRemoveNode changes the live raft tracker but snapshot ConfState persistence reports failure
- serialize terminal publication with proposal admission, fail queued proposals and commands, and preserve the definitive force-remove result across caller cancellation
- release an admitted force-removal waiter when the node stops before execution, preserving any completed result
- preserve successful force-removal, pending-removal, normal removal, and restart behavior
- add deterministic live/durable/restart coverage for both snapshot persistence residuals and a four-to-three quorum consequence

Jira: https://formance-team.atlassian.net/browse/EN-1958
Audit: AI-AUDIT:raft-membership-leadership/force-remove-wal-failure-changes-live-quorum

## Bugfix evidence

DISCOVERY: NO_EXISTING_WORK
BEFORE_FIX: BUG_REPRODUCED
AFTER_FIX: PASS
TEST_SENSITIVITY: PASS

The pre-fix probe used the real ForceRemoveNode path and injected failure at UpdateSnapshotConfState. It observed a reduced live RawNode tracker and node ConfState, the previous durable voters, an error response, continued command service, and restart to the previous voter set.

The fixed contract is fail-stop: the irreversible live tracker transition is not rolled back. The persistence error is published as terminal; concurrent queued commands receive that error; queued proposals have both futures failed and their index prediction rolled back; Node.Run propagates the task failure to bootstrap termination.

UpdateSnapshotConfState first replaces its in-memory snapshot, then atomically replaces the snapshot file and appends the etcd WAL snapshot record. The regression drives a real temporary-file creation failure after the in-memory snapshot mutation, which restarts the old membership, and a real failure after replacement at the closed-WAL record write, which restarts the new membership. Neither residual can coexist with a continuing process. A four-to-three test stages an entry that becomes live-committed only after quorum shrinks and proves it is neither durably committed nor applied before fail-stop/restart, for both failure phases. A separate regression runs the complete Node.Run lifecycle and verifies propagation of the original terminal cause.

## Validation

Candidate: `ae204292e5685a78da4e50d225e7fd7522d6acbf`  
Rebased on: `b2b7e7d033daa481cee3f89dfd9669de5b6ed24b` (`release/v3.0`)

- `go test -race ./internal/infra/node -run 'TestForceRemoveNode|TestClusterCommandResultAfterStop' -parallel 2 -count=1 -timeout 5m` — passed under pinned Nix
- `bash scripts/agent-check` — passed
- `AI_REVIEW_BASE_SHA=b2b7e7d033daa481cee3f89dfd9669de5b6ed24b bash scripts/agent-check-pr` — passed, including clean normalization, both lint targets, baseline, and the full node package race suite
- Independent review of this exact candidate: no remaining findings

GitHub CI is the broad clean validation gate. Both blocking review threads are resolved with evidence; a fresh review has been requested from flemzord. No merge has been performed.

### DIFF
--- a/internal/infra/node/force_remove_durability_test.go
+++ b/internal/infra/node/force_remove_durability_test.go
@@ -1,0 +1,542 @@
+package node
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"go.etcd.io/raft/v3"
+	"go.etcd.io/raft/v3/raftpb"
+	"go.opentelemetry.io/otel/metric/noop"
+	"go.uber.org/mock/gomock"
+	"google.golang.org/protobuf/proto"
+
+	logging "github.com/formancehq/go-libs/v5/pkg/observe/log"
+
+	"github.com/formancehq/ledger/v3/internal/infra/membership"
+	"github.com/formancehq/ledger/v3/internal/storage/wal"
+)
+
+type forceRemoveFailureStage int
+
+const (
+	forceRemoveNoFailure forceRemoveFailureStage = iota
+	forceRemoveBeforeSnapshotPersistence
+	forceRemoveAfterSnapshotFilePersistence
+)
+
+type forceRemoveWAL struct {
+	wal.WAL
+
+	mu           sync.Mutex
+	failureStage forceRemoveFailureStage
+	updateErr    error
+	updates      int
+	started      chan struct{}
+	startOnce    sync.Once
+	release      chan struct{}
+	releaseOnce  sync.Once
+}
+
+func (w *forceRemoveWAL) UpdateSnapshotConfState(cs *raftpb.ConfState) error {
+	w.mu.Lock()
+	w.updates++
+	w.mu.Unlock()
+
+	w.startOnce.Do(func() { close(w.started) })
+	if w.release != nil {
+		<-w.release
+	}
+
+	switch w.failureStage {
+	case forceRemoveNoFailure:
+		return w.WAL.UpdateSnapshotConfState(cs)
+	case forceRemoveBeforeSnapshotPersistence:
+		return w.updateErr
+	case forceRemoveAfterSnapshotFilePersistence:
+		// Closing the real etcd WAL makes its SaveSnapshot call fail while
+		// leaving Snapshotter.Save operational. This drives the production
+		// UpdateSnapshotConfState sequence through its atomic snapshot-file
+		// replacement and fails at the following WAL-record write.
+		if err := w.Close(); err != nil {
+			return fmt.Errorf("closing WAL for failure injection: %w", err)
+		}
+
+		err := w.WAL.UpdateSnapshotConfState(cs)
+		if err == nil {
+			return errors.New("invariant: closed WAL accepted snapshot record")
+		}
+
+		return fmt.Errorf("%w after snapshot-file persistence: %w", w.updateErr, err)
+	default:
+		return errors.New("invariant: unknown force-remove failure stage")
+	}
+}
+
+func (w *forceRemoveWAL) updateCount() int {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	return w.updates
+}
+
+type forceRemoveHarness struct {
+	n               *Node
+	rawNode         *raft.RawNode
+	durableWAL      wal.WAL
+	injectedWAL     *forceRemoveWAL
+	membership      *membership.Membership
+	walDir          string
+	baselineCommit  uint64
+	stagedIndex     uint64
+	stop            chan struct{}
+	orchestrateDone chan error
+	stopped         bool
+}
+
+func newForceRemoveTransport(t *testing.T) Transport {
+	t.Helper()
+
+	transport := NewMockTransport(gomock.NewController(t))
+	transport.EXPECT().Unreachable().AnyTimes().Return((<-chan uint64)(nil))
+	transport.EXPECT().RecvHighPriority().AnyTimes().Return((<-chan []*raftpb.Message)(nil))
+	transport.EXPECT().RecvMediumPriority().AnyTimes().Return((<-chan []*raftpb.Message)(nil))
+	transport.EXPECT().RecvLowPriority().AnyTimes().Return((<-chan []*raftpb.Message)(nil))
+	transport.EXPECT().Send(gomock.Any()).AnyTimes()
+
+	return transport
+}
+
+func newForceRemoveHarness(
+	t *testing.T,
+	voters []uint64,
+	failureStage forceRemoveFailureStage,
+	updateErr error,
+	blockUpdate bool,
+) *forceRemoveHarness {
+	t.Helper()
+
+	setup := newTestApplierSetup(t)
+	initial := &raftpb.ConfState{Voters: voters}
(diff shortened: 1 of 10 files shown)


internal/infra/node/node.go
```diff
@@ -1946,6 +2023,18 @@ func (node *Node) Propose(ctx context.Context, proposal *Proposal) (*futures.Fut
 	select {
```
NumaryBot, line 2091: 🟠 **[major] Reject proposals atomically with terminal publication**

If `terminalCh` closes while `proposeCh` still has capacity, both select cases are ready and Go may choose the send, returning a successful future after the orchestrator has exited; that proposal is never handled or resolved. `execClusterCommand` has the same check-then-select race, leaving commands queued until their contexts expire. This violates the documented rejection contract in [raft-consensus.md:278-281](docs/technical/architecture/subsystems/consensus/raft-consensus.md#L278-L281); admission must be synchronized with terminal publication rather than relying on competing select cases.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1906#issuecomment-5541502285

internal/infra/node/force_remove_durability_test.go
```diff
@@ -0,0 +1,290 @@
...
+	updates   int
+}
+
+func (w *forceRemoveWAL) UpdateSnapshotConfState(cs *raftpb.ConfState) error {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	w.updates++
+	if w.updateErr != nil {
+		return w.updateErr
```
flemzord: **[P2][blocking] Exercise the real WAL failure phases**

This injects the error before calling the real `UpdateSnapshotConfState`, so the restart assertions only cover a pre-write failure. Production first swaps the in-memory snapshot, then atomically overwrites `<term>-<index>.snap`, then calls `wal.SaveSnapshot`. If that last write fails, restart can still match the already-replaced file to the previous WAL record by term/index and load the reduced `ConfState`, contradicting the unconditional “old voters restored” contract. Please add phase-accurate injection (especially after the snapshot-file save) and either recover or document/test every safe residual state.

internal/infra/node/force_remove_durability_test.go
```diff
@@ -0,0 +1,290 @@
...
+	stop            chan struct{}
+	orchestrateDone chan error
+	stopped         bool
+}
+
+func newForceRemoveHarness(t *testing.T, updateErr error) *forceRemoveHarness {
+	t.Helper()
+
+	setup := newTestApplierSetup(t)
+	initial := &raftpb.ConfState{Voters: []uint64{1, 2, 3}}
```
flemzord: **[P2][blocking] Exercise a quorum that actually shrinks**

`{1,2,3} → {1,2}` keeps quorum at 2, so this fixture cannot prove EN-1958's criterion that no work commits because only the live quorum was reduced. It also stages no uncommitted entry or concurrent proposal. Use a transition such as 4→3 (old quorum 3, new quorum 2) or 2→1, stage an entry that is committable only after removal, then assert no acknowledgement/application/persistence occurs before restart and that restart follows the proven durable configuration.

flemzord (changes requested):
Requesting changes for two additional durability/test-contract gaps, detailed inline. I did not duplicate the existing terminal-admission race thread, which remains valid. The current head passes the repository's canonical `agent-check-pr`; these findings concern failure-phase fidelity and the missing reduced-quorum scenario.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1906#issuecomment-5541502285

internal/infra/node/force_remove_durability_test.go
gfyrag: Fixed in ae204292e5685a78da4e50d225e7fd7522d6acbf. Both failure cases now execute the real WAL UpdateSnapshotConfState: the pre-file case blocks Snapshotter.Save's temporary-file creation, after the WAL has already replaced its in-memory snapshot; the post-file case closes the real etcd WAL, allows atomic snapshot-file replacement, and fails the following snapshot-record write. The tests distinguish the changed in-memory ConfState from the reopened durable state: old voters before file replacement, reduced voters after replacement. Both paths assert fail-stop and skipped peer/tombstone cleanup. See internal/infra/node/force_remove_durability_test.go (forceRemoveWAL.UpdateSnapshotConfState and TestForceRemoveNodeDurabilityFailureContract), and the corrected write-order/restart contract in removed-member-registry.md. Targeted race tests and canonical PR validation pass.

internal/infra/node/force_remove_durability_test.go
gfyrag: Verified and strengthened in ae204292e5685a78da4e50d225e7fd7522d6acbf. The real RawNode fixture starts with four voters and stages an entry acknowledged only by nodes 1 and 2: commit stays at the old baseline with quorum 3. Removing node 4 reduces quorum to 2 and advances the live commit to the staged entry. The regression then proves that orchestrate exits, FSM application and durable HardState.Commit remain at their baselines, and reopening reconstructs the expected membership. This now runs for both real WAL failure phases: four voters after a pre-file failure, three after snapshot-file replacement. The separate concurrent-admission test proves that queued proposal/command callers receive the terminal error and proposal index reservations are rolled back. See TestForceRemoveNodeDurabilityFailureContract. Targeted race tests and canonical PR validation pass.

gfyrag:
All inline and top-level review feedback has been checked against ae204292e5685a78da4e50d225e7fd7522d6acbf, rebased on b2b7e7d033daa481cee3f89dfd9669de5b6ed24b.

- The two blocking durability findings are covered by real pre-/post-snapshot-file failure injection and a four-to-three staged-commit regression for both restart outcomes; replies are attached to the original threads.
- Terminal proposal admission is serialized with publication and draining. Queued commands observe the terminal cause; cancellation after force-removal admission cannot hide its persistence result. The existing concurrent regression verifies both futures and index rollback.
- The remaining Shipfox findings are addressed: restart closes are checked without double-closing the injected WAL; the generated transport mock is used; the consensus README and file-scoped lifecycle exception wording are updated; completion mode is a parameter of execClusterCommand.
- TestForceRemoveNodeDurabilityFailureStopsRun now exercises actual Node.Run and verifies propagation of the original terminal cause. An additional shutdown regression catches an admitted command that never executes: it returns raft.ErrStopped once Run exits, releases confChangeMu, and preserves an already completed result.

Validation: focused race regressions, bash scripts/agent-check, and AI_REVIEW_b2b7e7d033daa481cee3f89dfd9669de5b6ed24b_SHA=b2b7e7d033daa481cee3f89dfd9669de5b6ed24b bash scripts/agent-check-pr passed. Independent review of this candidate found no remaining findings. The [GitHub CI run](https://github.com/formancehq/ledger/actions/runs/34235061850) completed successfully on this exact head, including all test suites and Coverage. The PR remains unmerged and needs a fresh human approval to replace the existing changes-requested review.

flemzord (approved):
Reviewed ae204292e. No actionable findings remain. Verified terminal proposal admission, cancellation and shutdown behavior, both real WAL failure phases, and the four-to-three quorum regression. The previous review findings are addressed. Targeted node regression tests pass with the race detector under pinned Nix, scripts/agent-check passes, and all executed CI checks are green.

PR #1907, opened 2026-09-04
feat(EN-1827): freeze idempotency expiry per outcome (PR3)

Azorlogh:
## What changed

The FSM no longer reads a node-local idempotency TTL when deciding whether a committed outcome has expired. Each apply freezes the outcome's absolute `expires_at` once, from the TTL in the cluster policy committed *before* that proposal, stores it on `IdempotencyKeyValue`, and chain-binds it in the audit header. `PersistedConfig.idempotency_ttl_seconds` and its boot backfill/mismatch check are removed; `--idempotency-ttl` now only seeds the desired cluster policy.

## Why

Reading the node-local `--idempotency-ttl` inside the apply-time expiry check (`IsExpired`) let two nodes configured with different TTLs diverge on whether the *same committed outcome* was still live — a deterministic-FSM boundary violation (EN-1797). This is PR3 of the EN-1827 replicated-cluster-policy umbrella: PR1 made the TTL replicated; this closes the remaining node-local read.

- **Need:** a committed idempotency outcome's lifetime must be identical on every replica, so a retry replays the same result regardless of which node serves it.
- **Current limitation:** the expiry check read the node-local TTL, so a rolling upgrade or a misconfigured node could commit an outcome and then disagree with its peers on whether it was still live.
- **Requirement / constraint:** the expiry decision for any committed outcome must be a pure function of committed state.
- **Evidence:** EN-1797; `docs/technical/architecture/subsystems/fsm/deterministic-fsm.md` §3.5.
- **Durable repository evidence:** `deterministic-fsm.md`, `audit-vs-technical-state.md`, `subsystems/admission/idempotency.md`, `subsystems/checker/audit-chain.md`.

## Technical decision

- **Decision:** freeze `expires_at = created_at + committed-policy TTL` per outcome at apply time (`0` = never; overflow saturates), store it on the outcome, and bind it in the audit header so restore and the checker read it back from the chain. Eviction keys its time index by `expires_at`; the leader scan takes a wall-clock cutoff.
- **Why now / why proportionate:** PR1 made the TTL a replicated cluster-policy field; this is the minimal change that removes the last node-local expiry read while keeping eviction leader-scheduled and Raft-replicated.
- **Alternatives considered:** (a) re-derive expiry from a policy timeline in the checker rather than storing it — rejected as more complex and still requiring the apply-time freeze; (b) do nothing — leaves the determinism violation.

## Risk

**MEDIUM** — touches the audit hash chain (a new bound field) and a persisted projection. v3 is unreleased so there is no compatibility burden (fields deleted + renumbered), and the checker verifies the new `expires_at` against the chain-derived value.

## Validation

- [x] Targeted tests: `internal/infra/state/...`, `internal/application/check/...`, `internal/bootstrap/...`, `internal/adapter/grpc/...`, `internal/infra/node/...`
- [ ] Full suite / broader validation: N/A (CI)

New tests: per-outcome freeze from committed policy (+ non-retroactivity + audit-chain binding), TTL=0 never-expires, `expires_at` round-trip through recovery, `IdempotencyExpiresAt`/`IdempotencyExpired` unit tests, and a checker `expires_at`-tamper case.

- **Audit hash chain:** `Idempotency.expires_at` is now bound into the header pre-image (tamper-evident; golden test updated).
- **Persisted state:** `IdempotencyKeyValue.expires_at` added; `PersistedConfig.idempotency_ttl_seconds` removed and field numbers realigned (v3 unreleased, per RULE/CLAUDE.md). New replicated `FSMState.LastIdempotencyEvictionCutoff` (persisted under `SubGlobLastIdempotencyEvictionCutoff` `0x14`, advanced in `applyIdempotencyEviction`, reloaded in `RecoverState`), the eviction high-water mark the preload gate consults.
- **FSM/Raft:** expiry is frozen from committed policy; no node-local TTL read in apply. The preload gate reads the stored `expires_at`, the replicated `LastIdempotencyEvictionCutoff` (an eviction never advances the HLC, so the gate cannot rely on it), and the current outcome's `created_at` (freshness); eviction reads only the stored `expires_at`.
- **Eviction:** time index re-keyed by `expires_at`. Eviction of the main key switched from `SingleDelete` to a plain `Delete`: an expired key is legitimately re-`Set` on reuse, so it is not write-once and `SingleDelete` over two `Set`s is undefined (could resurrect a stale outcome at compaction). This also rewrites the scheduler's no-timeout rationale (a duplicate apply is now a map-gated no-op rather than an undefined double-`SingleDelete`).
- **Compatibility (v3 unreleased):** the `PersistedConfig` encoding change (field renumber + removed `idempotency_ttl_seconds`) and the audit-header pre-image change are not backward-compatible with stores/backups written by pre-PR `release/v3.0` builds. There is no migration path by design (RULE 12 / CLAUDE.md: state is wiped on `release/v3.0` pushes) — such stores must be wiped. `CurrentStorageSchemaVersion` is intentionally not bumped (a pre-PR store's old field 3 decodes as the version and fails boot fail-safe, never corrupting).

## Review focus

- `machine.go`: the single `idempotencyExpiresAt` freeze point feeding the audit stamp, both success/failure freeze paths, and the preload gate — confirm it always reads the pre-apply committed policy.
- `audit_envelope.go` + golden test: the new chain-bound field.
- `checker.go` `idempotencyMismatch`: the `expires_at` verification.

## Known concerns

None. This branches off the chapters-removed base (EN-1945): the checker's archived idempotency re-derivation is already gone, so PR3's checker delta reduces to the `expires_at` verification.


internal/infra/state/machine.go
```diff
@@ -890,8 +890,15 @@ func (fsm *Machine) Preload(executionPlan *raftcmdpb.ExecutionPlan, batch *dal.W
...
+		//
+		// Skip a value already past its expiry as of the last applied HLC: a
+		// committed IdempotencyEviction between the leader's plan-build and this
+		// apply has already removed it from Pebble and the in-memory map, and
+		// re-injecting the stale copy would leave the map ahead of Pebble and let
+		// a later eviction double-SingleDelete the main key.
 		v := ik.GetValue()
-		if v != nil && (v.GetFirstLogSequence() > 0 || v.GetFailure() != nil) {
+		if v != nil && (v.GetFirstLogSequence() > 0 || v.GetFailure() != nil) &&
+			!IdempotencyExpired(v.GetExpiresAt(), fsm.State.LastAppliedTimestamp) {
```
NumaryBot, line 901: 🔴 **[blocker] Do not use the order HLC to detect intervening eviction**

When an idle cluster evicts an outcome after its wall-clock `expires_at`, the technical eviction does not advance `LastAppliedTimestamp`, so this comparison can still consider the stale preloaded value live. A proposal planned before that eviction can therefore reinsert the deleted value into memory, diverging the cache from Pebble and exposing the later double-`SingleDelete` risk described above. This violates the cache-consistency invariant in [AGENTS.md](../../../AGENTS.md#invariants); the guard must identify whether the value was evicted rather than infer that from the order HLC.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1907#issuecomment-5541641673

internal/infra/state/machine.go
Azorlogh: Fixed in 1554467b: the preload gate now reads a committed high-water mark, `FSMState.LastIdempotencyEvictionCutoff` (advanced and persisted by `applyIdempotencyEviction`, re-read in `RecoverState`), instead of the HLC — so an evicted outcome is identified by the eviction that removed it, not inferred from the order clock. Added a load→evict→stale-preload regression asserting map/Pebble parity and no second `SingleDelete`.

https://github.com/formancehq/ledger/commit/1554467bb75e6f9a2f3cb90ca1be4e71574bfee4

Azorlogh:
@shipfox-ai thanks — all three addressed in 1554467b:

1. **Preload staleness guard (blocker):** the gate now consults a committed high-water mark, `FSMState.LastIdempotencyEvictionCutoff` (advanced + persisted by `applyIdempotencyEviction`, re-read in `RecoverState`), instead of the HLC an eviction never advances. Added the missing `load → evict → stale-preload` regression asserting map/Pebble parity and no second `SingleDelete`.
2. **Restore parity (invariant #11):** added `TestRebuildDelta_IdempotencyExpiresAtRoundtrips` — the rebuild reconstructs the frozen `expires_at` from the audit chain *and* its eviction time-index entry (via `SaveIdempotencyKey`), so a restored cluster evicts on the same schedule.
3. **Doc:** the `deployment-profiles.md` "Long client retry window" section now describes the cluster-policy revision-bump model instead of the removed persisted-TTL validation.

https://github.com/formancehq/ledger/commit/1554467bb75e6f9a2f3cb90ca1be4e71574bfee4

Azorlogh:
@shipfox-ai — both items from the re-review addressed in `6a998f0f`:

1. **Restore parity (invariant #11):** the `RebuildDelta` unit test stays as the lowest-level check, and the cross-lifecycle proof now lives in the e2e suite — `restore_idempotency_test.go` runs `CheckStore` on the restored node, which re-derives and verifies each outcome's `expires_at` from the audit chain, proving the frozen retention deadline (and its eviction time-index entry) survive restore. Reworded the unit test's comment so it no longer claims #11 parity on its own, and recorded the *Rebuilt* classification in `rebuildIdempotency`'s doc.
2. **Doc precision:** `audit-chain.md` now notes `IdempotencyExpiresAt` is stamped on *every* keyed proposal, including conflict/non-freezable ones that freeze nothing (the checker derives no expectation from those).

https://github.com/formancehq/ledger/commit/6a998f0f740f5387c879f1451b33969d5a982d07

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1907#issuecomment-5541641673

internal/infra/state/machine.go
```diff
@@ -890,8 +890,17 @@ func (fsm *Machine) Preload(executionPlan *raftcmdpb.ExecutionPlan, batch *dal.W
...
+		// leader's plan-build and this apply: re-injecting it would leave the map
+		// ahead of Pebble and let a later eviction double-SingleDelete the main
+		// key. Eviction status is read from the replicated eviction cutoff, NOT
+		// the HLC: an eviction is a technical-only proposal that never advances
+		// LastAppliedTimestamp, so on an idle cluster the HLC lags the eviction's
+		// wall-clock cutoff and would wrongly consider the removed value live.
 		v := ik.GetValue()
-		if v != nil && (v.GetFirstLogSequence() > 0 || v.GetFailure() != nil) {
+		if v != nil && (v.GetFirstLogSequence() > 0 || v.GetFailure() != nil) &&
+			!IdempotencyEvicted(v.GetExpiresAt(), fsm.State.LastIdempotencyEvictionCutoff) {
```
gfyrag: 🔴 **[blocker] A stale expired preload can overwrite a newer live outcome**

The eviction cutoff proves that this plan-carried value was not already evicted; it does not prove that it is still the newest value for the key. A and B can both build while Pebble contains the same expired outcome V0 and the cutoff is still below V0.expires_at. A then applies, treats V0 as expired, executes, and installs a live V1. When B applies, this Put overwrites V1 with its stale V0. The idempotency gate then sees V0 as expired and executes again: with identical content I observed a second created log (sequences 3 then 4); with different content I observed no idempotency conflict and another created log. Recovery first restored V1 correctly, but the stale preload overwrote it again, so restart does not close the race.

The existing load → eviction → stale-preload regression covers the case fixed by LastIdempotencyEvictionCutoff. Please also protect the distinct expired-but-not-yet-evicted replacement interleaving; otherwise at-most-once and conflict semantics remain bypassable.

internal/infra/state/idempotency_store.go
```diff
@@ -247,9 +277,9 @@ func (s *IdempotencyStore) Evict(batch *dal.WriteSession, cutoffMicros uint64, l
 			continue
 		}
 
-		if value.GetCreatedAt() > cutoffMicros {
-			// Defensive: the leader scan returned this hash with a stale
-			// cutoff. Don't evict a still-live entry.
+		if value.GetExpiresAt() == 0 || value.GetExpiresAt() > cutoffMicros {
```
gfyrag, line 276: 🔴 **[blocker] Expired-key reuse violates the claimed SingleDelete lifecycle**

The production path still permits Set(K,A) → Set(K,B) → SingleDelete(K). Once A is logically expired, apply treats the key as reusable and SaveIdempotencyKey writes B with Set to the same main key while adding a second expiry-index row. When B later reaches this cutoff, Evict emits SingleDelete on that twice-written main key. The in-memory map gate prevents a duplicate SingleDelete for the repeated hash, but it does not undo the two Sets.

A deterministic DAL/Pebble fixture confirmed the reachability: Set(A), flush, Set(B), flush produced two scanned time-index rows and one production Evict SingleDelete. This particular run remained absent after delete, CompactAll, close, and reopen, so I did not observe a resurrection; however, that does not establish safety for a lifecycle Pebble documents as undefined. The write-once/delete-once guarantee in the comment below is therefore not actually enforced.

internal/infra/state/machine.go
Azorlogh: Fixed in ee2434f6 — the preload re-injection now also gates on freshness: it re-injects only when the plan value is not older than the map's current value (`created_at`, the strictly-monotonic apply HLC), so a stale plan value can't clobber a newer live outcome. Regression `TestPreload_DoesNotOverwriteNewerOutcomeWithStalePlan` fails on the old unguarded code.

https://github.com/formancehq/ledger/commit/ee2434f61a7ca4b34f5aec2cc8b142b862a7643d

internal/infra/state/idempotency_store.go
Azorlogh: Fixed in ee2434f6 — `Evict` now uses a plain `Delete` instead of `SingleDelete`. The main key is legitimately re-`Set` on reuse (an expired key reused by a fresh proposal), so it is NOT write-once; a plain `Delete` tombstones every prior `Set` and is correct — matching `backup_jobs.go`'s existing `DeleteKey` pattern for rewritten keys. Updated the now-false write-once comments in `Evict` and the scheduler. Regression `TestEviction_ReusedKeyDeletesCleanly`.

https://github.com/formancehq/ledger/commit/ee2434f61a7ca4b34f5aec2cc8b142b862a7643d

Azorlogh:
@shipfox-ai — both re-review items addressed:

1. **Preserved-path parity** (`7d1810613`): added `TestRebuildDelta_PreservesCheckpointExpiresAt` — a checkpoint-carried outcome with a finite `expires_at`, folded past an unrelated delta key, keeps its `expires_at` and eviction time-index entry, asserted against a **literal** expected value (an oracle independent of the shared audit decoder). The e2e suite continues to cover the Rebuilt/delta path + `CheckStore`.
2. **PR body**: amended *Persisted state* and *FSM/Raft* to enumerate the new replicated `LastIdempotencyEvictionCutoff` the preload gate reads.

Separately, gfyrag found two further blockers in the preload/eviction path (a stale-preload overwrite of a newer outcome, and `SingleDelete` over a legitimately re-`Set` key), fixed with a freshness guard + a switch to plain `Delete` in `ee2434f6`.

https://github.com/formancehq/ledger/commit/ee2434f61a7ca4b34f5aec2cc8b142b862a7643d

gfyrag:
Follow-up on `7d181061`: the production changes for both reported defects look directionally correct (`created_at` freshness ordering and plain `Delete`). I do not think the current tests are strong enough evidence for approval yet.

## Methodology concern

Several test comments describe a production lifecycle that the fixture itself does not execute. A white-box assertion immediately around the changed helper is useful as a unit test, but it must not be presented as proof of the cross-component race, restart behavior, or storage lifecycle that motivated the fix. The regression should fail for the original externally observable reason, not merely because the new conditional or API call is absent.

### 1. Exercise the complete stale-preload interleaving

`TestPreload_DoesNotOverwriteNewerOutcomeWithStalePlan` puts an artificial live value directly in the map and calls `Machine.Preload` directly. Its values have no finite `ExpiresAt`; it does not build two proposals against the same expired Pebble outcome, apply A, resume B, run the idempotency gate, or verify persisted state/recovery. It therefore proves only the local `created_at` comparison.

Please add a deterministic FSM-level regression which:

- stores one finite, expired-but-not-yet-evicted outcome in Pebble and the map;
- constructs both A and B execution plans before A applies, with valid production-equivalent index/coverage metadata;
- applies A and verifies that it installs a newer live outcome;
- resumes B with its stale plan;
- for identical content, asserts a reference to A and no second created log/audit/business mutation;
- for different content, asserts `ErrIdempotencyKeyConflict` and no second execution;
- repeats the relevant boundary after recovery/restart and asserts both the map and Pebble still hold A.

This should test the actual at-most-once and conflict contracts. Avoid inserting an unrelated replay between A and B unless the reserved `PredictedIndex` sequence makes that interleaving production-valid.

### 2. Finish the storage-lifecycle proof

`TestEviction_ReusedKeyDeletesCleanly` reaches the real eviction handler, but `require.NotEmpty(hashes)` does not prove that both expiry-index rows from A and B exist, and the test checks the main key only immediately after `Delete`. It does not exercise the compaction/reopen boundary that exposed the original risk.

Please assert the exact precondition (two index rows for the reused hash and one logical eviction), then verify absence immediately after delete, after `CompactAll`, and after close/reopen, with the expiry index empty as well. Plain `Delete` should make this straightforward and turns the test into evidence for the claimed durable lifecycle rather than an implementation-shaped smoke test.

### 3. The Preserved restore path is still not cross-lifecycle

I agree with the current Shipfox finding. `TestRebuildDelta_PreservesCheckpointExpiresAt` does not create a checkpoint, export a non-empty delta, copy SSTs, or perform a real restore. It seeds the projection and audit rows in one store and calls `RebuildDelta`; that only proves an unrelated fold leaves an existing key untouched. Meanwhile the e2e suite still checkpoints an empty store, so it covers only the Rebuilt path.

Please put a finite-expiry keyed outcome before a real full checkpoint, add a meaningful non-empty post-checkpoint delta, restore through the production path, compare the literal `expires_at` and time-index state between source and restored stores, and run `CheckStore`. The comments claiming invariant #11 cross-lifecycle proof should match what the fixture actually executes.

Green CI cannot close these gaps because the current fixtures do not traverse the failure surfaces above. I am not approving at this point.

Azorlogh:
@shipfox-ai — the [High] Preserved-path finding is addressed in `35f68bd6`:

- The e2e restore suite now freezes a keyed outcome **before** the full checkpoint (Preserved via the raw-SST copy) alongside the post-checkpoint delta keys (Rebuilt), and asserts both dedup after the real restore + `CheckStore` — so both halves of the `expires_at` projection are proved cross-lifecycle.
- Corrected the `rebuild_test.go` comment to scope its claim to the Rebuilt-path unit test and point at both unit tests + the e2e for the full proof.

Also rebased onto the current `release/v3.0` (Go 1.27.1 + tooling commits).

https://github.com/formancehq/ledger/commit/35f68bd6522ac01b2a56c728dbd128075c74a99b

gfyrag:
@Azorlogh quick reconciliation on the current HEAD `35f68bd6`:

- The PR is now `CONFLICTING`/`DIRTY` against the current `release/v3.0` base (`ce72d6df`). Please rebase it before the next review pass.
- `35f68bd6` closes the cross-lifecycle **Preserved** restore gap, but it does not address the two remaining evidence gaps from my previous review.

After the rebase, I still need:

1. A deterministic FSM-level two-proposal regression for the stale-preload race: both proposals preload the same finite expired outcome; A installs the newer live result; B resumes with the stale plan; verify identical-content replay without a second business/audit mutation, different-content conflict without a second execution, and the same result after recovery/restart. The current direct `Preload` helper test proves only the local `created_at` comparison.
2. A complete reused-key Pebble lifecycle regression: prove the exact precondition of two expiry-index rows for `Set(K,A) -> Set(K,B)`, then verify the main key and expiry rows are absent immediately after production eviction, after `CompactAll`, and after close/reopen. `require.NotEmpty(hashes)` plus an immediate lookup is not enough evidence for the lifecycle that motivated replacing `SingleDelete`.

Please keep these tests contract-shaped: they should fail for the original externally observable behavior, not only when the new conditional or `Delete` call is removed. I am not approving until the rebase and these two proofs are in place.

Azorlogh:
@gfyrag rebased onto the current `release/v3.0` (`404ac87a`), and both contract-shaped regressions are in (`ed867cde`):

1. **Stale-preload race** — `TestApplyProposal_StalePreloadCannotResurrectSupersededOutcome`: FSM-level, driven through `ApplyEntries`. Both proposals preload the same finite expired V0; A installs a newer live V1; B resumes with its stale plan → **identical content replays A with no second log/business mutation, different content conflicts with no second execution**, and after a real restart (`recoverMachineOnStore` rebuilds the map from Pebble) the outcome is still V1. Confirmed it **fails on the pre-fix unconditional `Put`** (a second committed log at seq 4).
2. **Reused-key lifecycle** — `TestEviction_ReusedKeyLifecycleThroughCompaction`: asserts the **two expiry-index rows** for `Set(K,A)→Set(K,B)` and **one logical eviction**, then absence of the main key and expiry rows immediately after the production eviction, **after `CompactAll`, and after close/reopen**.

Also added `TestBackup_IdempotencyExpiresAtRestoreParity` (full backup→delta→`ApplyExportsAndRebuild`, literal source-vs-restored `expires_at`/time-index for the Rebuilt path), and relabeled the three white-box unit tests.

https://github.com/formancehq/ledger/commit/ed867cdeb7a8b68b7529ca7c9e3dad4dd8e99441

Azorlogh:
@shipfox-ai all three Minor items addressed:
1. Relabeled `LastIdempotencyEvictionCutoff` as **wall-clock** micros (not HLC) in `fsmstate.go`, `store.go`, `query/config.go` — the self-contradictory comment included (`1ea18eab`).
2. PR body: pre-PR `release/v3.0` stores/backups must be wiped (no migration path; v3 unreleased), and why `CurrentStorageSchemaVersion` isn't bumped.
3. PR body: documented the `SingleDelete`→`Delete` eviction change.

Also rebased onto the current `release/v3.0`.

https://github.com/formancehq/ledger/commit/1ea18eabfc1032f49648afe39a92c4ef3425c057

Azorlogh:
@gfyrag rebased onto current `release/v3.0`, and both regressions are now contract-shaped — each reds on the original externally-observable behavior, not just on removing the new conditional/`Delete`:

1. **Stale-preload race** — `TestApplyProposal_StalePreloadCannotResurrectSupersededOutcome`: two proposals preload the same finite expired outcome; A installs the newer live result, B resumes on the stale plan. Reverting to the unconditional `Put` reds it (a second business/audit log at seq 4). Identical-content replays without a second log, different-content conflicts without a second execution, and V1 holds after recovery/restart.

2. **Reused-key Pebble lifecycle** — `TestEviction_ReusedKeyLifecycleThroughCompaction`: pins the exact two expiry-index rows for `Set(K,A)→Set(K,B)`, then asserts the main key and both expiry rows are absent after production eviction, `CompactAll`, and close/reopen. The earlier version wrote A and B into the same level, so one `SingleDelete` merged them away and passed on the buggy path too; it now compacts A to a lower level before the reuse, so reverting `Delete`→`SingleDelete` reds it — the `SingleDelete` cancels only B and a read resurrects the older expired A (`FirstLogSequence:1, ExpiresAt:expA`), the exact resurrection the switch prevents. Verified red under `SingleDelete`, green under `Delete`.

c17db164f

Azorlogh:
@shipfox-ai on the two Standards [High]s: both are cross-binary storage/backup compatibility — a store or backup written by the merge-base binary, opened by the new one. V3 is unreleased, and persisted/backup formats are explicitly **not** supported across development revisions (`docs/technical/architecture/subsystems/api/protocol-compatibility.md` L113-114 and L131-132; state is wiped on `release/v3.0` pushes). Both cases fail safe — boot is refused and the chain check flags rather than silently corrupts — which is the intended behavior for an old store that cannot exist in a supported deployment. Out of scope, not blocking.

Spec [Low] (docs claim the checker "re-derives" `expires_at`): fixed in c17db164f — reworded to "verifies the stored projection against the chain-bound value," matching `audit-chain.md`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1907#issuecomment-5541641673

docs/technical/architecture/subsystems/fsm/deterministic-fsm.md
```diff
@@ -155,6 +155,10 @@ keeps retry, replay, and restart behavior deterministic (EN-1860).
...
+
+The preload re-injection gate is the one place that must not use the HLC. An `IdempotencyEviction` is a technical-only proposal, so it never advances `LastAppliedTimestamp`; on an idle cluster the HLC lags the eviction's wall-clock cutoff, and a proposal whose plan predates the eviction would re-inject the already-removed outcome into the cache (map ahead of Pebble). The gate therefore consults a separate committed high-water mark, `FSMState.LastIdempotencyEvictionCutoff` (advanced by `applyIdempotencyEviction`, persisted at `ZoneGlobal`/`SubGlobLastIdempotencyEvictionCutoff`): an outcome whose `expires_at` is at or below it has been evicted and is not re-injected. Because the cutoff is committed FSM state, replay and snapshot restore reach the same decision on every node.
```
NumaryBot: 🟠 **[major] Describe the cutoff as a scan horizon, not an eviction frontier**

When more than 10,000 outcomes are expired, the scheduler scans only one bounded batch, but `applyIdempotencyEviction` advances the cutoff to the full wall-clock horizon. Unscanned outcomes below that cutoff therefore remain in both Pebble and the map; the statement that they have been evicted is false. Preload safety additionally depends on every still-persisted outcome already being in the map, as the implementation comments correctly explain. Document that invariant rather than attributing safety to the cutoff alone, consistent with the subsystem-documentation requirement in [AGENTS.md:95](../../../../../AGENTS.md#L95).

Azorlogh: Fixed in dd631aad5 — reworded to describe the cutoff as a scan horizon and name the map-superset invariant the gate's safety rests on, rather than calling everything below the cutoff evicted. https://github.com/formancehq/ledger/commit/dd631aad5ca992e682b09cf2a4b83479fb350f64

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1907#issuecomment-5541641673

paul-nicolas (approved):
Reviewed at `dd631aad5ca992e682b09cf2a4b83479fb350f64`.

The per-outcome expiry is frozen from committed cluster-policy state, carried consistently through success and frozen-failure paths, chain-bound, checker-verified, and reconstructed across both preserved and rebuilt restore paths. The preload cutoff, freshness guard, and plain-Delete reuse lifecycle are covered by contract-shaped regressions.

The persisted-format and audit-envelope incompatibility is explicitly within Ledger v3's unreleased reset boundary and is disclosed in the PR, so it is not a blocking compatibility defect.

Local validation: `bash scripts/agent-check` and targeted tests for state, backup, checker, bootstrap, and query all passed. No actionable findings.

PR #1908, opened 2026-09-04
fix(restore): join downloads during Fx shutdown

gfyrag:
## What changed

Restore-mode services now own asynchronous download jobs for the full Fx application lifetime. Shutdown rejects new restore RPC work, cancels and joins an active download, waits for admitted restore RPCs, then closes retained staging Pebble state.

Terminal jobs release their child context before signaling completion, so successful or failed attempts do not remain retained by the service lifetime context.

RPC detachment remains intentional: cancelling the initiating StartDownload RPC does not cancel the background job; explicit CancelDownload remains the job-level cancellation API.

## Why

The current service creates downloads from a background context and has no Fx stop hook. A deterministic production-module regression showed Fx shutdown returning while an S3 manifest request and restore job remained alive, leaving service-owned staging state without teardown ownership.

Need: restore work may outlive its initiating RPC, but must never outlive the restore-mode application that owns it.
Current limitation: Fx shutdown neither cancels nor joins active download work and never closes retained staging Pebble state.
Requirement / constraint: once shutdown begins, no new job starts; active work receives cancellation and is joined; admitted RPCs finish; service-owned staging state closes last.
Evidence: internal/bootstrap/restore_download_lifecycle_test.go and internal/adapter/grpc/server_restore_lifecycle_test.go.
Durable repository evidence: docs/ops/backup-restore.md and docs/technical/architecture/subsystems/api/grpc-api.md.

## Technical decision

Decision: add a restore-service lifetime context, an admission/stopping gate, active-request and active-job joins, and phased Fx stop hooks that order admission closure before transports and resource closure after gRPC draining.
Why now / why proportionate: this is the smallest ownership boundary matching the existing single-job service and Fx lifecycle.
Alternatives considered: tying work to the initiating RPC would break intentional asynchronous behavior; closing staging state directly from Fx would race active restore operations; a broader restore subsystem redesign is unnecessary.

## Risk

**MEDIUM**: shutdown ordering and concurrency change in restore mode; deterministic synchronization tests and focused race validation cover the affected paths.

## Validation

- [x] `bash scripts/agent-check` (standalone, then repeated by the final PR gate)
- [x] `AI_REVIEW_BASE_SHA=afdd58395fd4689c624286893c2c618e38db8148 bash scripts/agent-check-pr` with shared caches and reduced local build concurrency (`GOMAXPROCS=4 GOFLAGS=-p=2`)
- [x] Focused race suites: `./internal/adapter/grpc` and `./internal/bootstrap`
- [x] Production Fx regression: `go test -race -tags s3 ./internal/bootstrap -run '^TestRestoreDownloadStopsWithFxApplication$' -count=1 -timeout 2m`
- [x] CI `Tests` explicitly runs the tagged Fx regression against the in-process S3 endpoint; no MinIO or Docker is required for this case
- [x] Regression sensitivity: restoring the background parent made the production Fx regression fail deterministically because shutdown did not cancel the job
- [x] Final diff and fresh review: no blocking findings or generated drift
- [x] Mutation checks: removing either Fx phase, reversing phase order, forwarding RPC cancellation, removing the job join, or omitting terminal child cancellation each fails its intended regression assertion
- [x] GitHub CI on `3ef3bec1280754c19e8475201901c87326de5c08` (15 successful checks; 2 expected release/image skips)

No wire, storage-format, FSM, Raft, or compatibility change. Restore-mode shutdown now owns asynchronous work and retained staging-store teardown. Explicit cancellation still leaves the application available for another download.

## Review focus

Please check lock/join ordering, the double admission check at job registration, and the phased Fx hook order around HTTP/gRPC shutdown and staging-store closure.

## Known concerns

The current service runner supplies no deadline to Fx Stop. Synchronous storage-client initialization, local staging cleanup, or a backend that ignores cancellation can delay shutdown. The existing staging-close policy logs close failures without returning them from the stop hook. The runbook states these limits and the residual state after forced termination.

### DIFF
--- a/docs/ops/backup-restore.md
+++ b/docs/ops/backup-restore.md
@@ -377,6 +377,14 @@ The server-side job:
 If the job fails or is cancelled, the staging directory is wiped so the
 operator can retry with a fresh state without restarting the server.
 
+The transfer is detached from the initiating RPC, not from the restore-mode
+application. Fx shutdown first rejects new restore RPCs and cancels any active
+download, then stops the gRPC server, joins admitted RPCs and the download job,
+and finally closes the retained staging Pebble store. Explicit
+`CancelDownload` keeps its client-facing behavior: it cancels only the selected
+job, waits for a bounded drain, and leaves the restore-mode application running
+so the operator can retry.
+
 ### Step 2: Validate
 
 ```bash
--- a/docs/ops/backup-restore.md
+++ b/docs/ops/backup-restore.md
@@ -378,12 +378,23 @@ If the job fails or is cancelled, the staging directory is wiped so the
 operator can retry with a fresh state without restarting the server.
 
 The transfer is detached from the initiating RPC, not from the restore-mode
-application. Fx shutdown first rejects new restore RPCs and cancels any active
-download, then stops the gRPC server, joins admitted RPCs and the download job,
-and finally closes the retained staging Pebble store. Explicit
+application. After any configured shutdown grace period, restore shutdown
+rejects new restore RPCs and cancels any active download, stops the HTTP health
+server and then gRPC, joins admitted RPCs and the download job, and finally
+closes the retained staging Pebble store. Explicit
 `CancelDownload` keeps its client-facing behavior: it cancels only the selected
 job, waits for a bounded drain, and leaves the restore-mode application running
 so the operator can retry.
+
+The join includes failed/canceled-job staging cleanup. Storage-client
+initialization and local filesystem operations do not accept the download
+context, so shutdown waits for them to return. The current service runner
+calls Fx `Stop` with a context without a deadline; `--total-stop-timeout` does
+not bound this path. A slow cleanup or backend can therefore delay process
+exit. Forced termination can interrupt cleanup and leave staging files or an
+uncleanly closed staging store; size the process termination grace period
+accordingly. Staging-store close errors are logged under the existing close
+policy and are not returned by the restore stop hook.
 
 ### Step 2: Validate
 
--- a/internal/adapter/grpc/server_restore_lifecycle_test.go
+++ b/internal/adapter/grpc/server_restore_lifecycle_test.go
@@ -1,0 +1,435 @@
+package grpc
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/formancehq/ledger/v3/internal/infra/backup"
+	"github.com/formancehq/ledger/v3/internal/proto/restorepb"
+	"github.com/formancehq/ledger/v3/internal/storage/dal"
+)
+
+type cancellationBlockingReader struct {
+	ctx      context.Context
+	started  chan struct{}
+	canceled chan struct{}
+	release  chan struct{}
+	closed   chan struct{}
+	start    sync.Once
+	cancel   sync.Once
+	unblock  sync.Once
+	close    sync.Once
+}
+
+func newCancellationBlockingReader(ctx context.Context) *cancellationBlockingReader {
+	return &cancellationBlockingReader{
+		ctx:      ctx,
+		started:  make(chan struct{}),
+		canceled: make(chan struct{}),
+		release:  make(chan struct{}),
+		closed:   make(chan struct{}),
+	}
+}
+
+func (r *cancellationBlockingReader) Read(_ []byte) (int, error) {
+	r.start.Do(func() { close(r.started) })
+	<-r.ctx.Done()
+	r.cancel.Do(func() { close(r.canceled) })
+	<-r.release
+
+	return 0, r.ctx.Err()
+}
+
+func (r *cancellationBlockingReader) Close() error {
+	r.close.Do(func() { close(r.closed) })
+
+	return nil
+}
+
+func (r *cancellationBlockingReader) allowUnwind() {
+	r.unblock.Do(func() { close(r.release) })
+}
+
+func startBlockedRestoreDownload(t *testing.T) (*RestoreServiceServerImpl, *downloadJob, *cancellationBlockingReader) {
+	t.Helper()
+
+	ctrl := gomock.NewController(t)
+	storage := NewMockStorage(ctrl)
+	readerReady := make(chan *cancellationBlockingReader, 1)
+	storage.EXPECT().
+		GetFile(gomock.Any(), gomock.Any()).
+		DoAndReturn(func(ctx context.Context, _ string) (io.ReadCloser, error) {
+			reader := newCancellationBlockingReader(ctx)
+			readerReady <- reader
+
+			return reader, nil
+		})
+
+	server := newServerForTest(t, staticFactory(storage))
+	_, err := server.StartDownloadBackup(context.Background(), &restorepb.StartDownloadBackupRequest{})
+	require.NoError(t, err)
+
+	reader := <-readerReady
+	<-reader.started
+
+	server.mu.Lock()
+	job := server.job
+	server.mu.Unlock()
+	require.NotNil(t, job)
+	t.Cleanup(func() {
+		reader.allowUnwind()
+		server.Shutdown()
+	})
+
+	return server, job, reader
+}
+
+func TestRestoreShutdownCancelsAndJoinsActiveDownload(t *testing.T) {
+	t.Parallel()
+
+	server, job, reader := startBlockedRestoreDownload(t)
+	server.BeginShutdown()
+	<-reader.canceled
+
+	shutdownDone := make(chan struct{})
+	go func() {
+		server.Shutdown()
+		close(shutdownDone)
+	}()
+
+	select {
+	case <-shutdownDone:
+		t.Fatal("shutdown returned before the canceled download unwound")
+	default:
+	}
+	select {
+	case <-reader.closed:
+		t.Fatal("the download resource closed before its owner unwound")
+	default:
+	}
+
+	_, err := server.StartDownloadBackup(context.Background(), &restorepb.StartDownloadBackupRequest{})
+	require.Equal(t, codes.Unavailable, status.Code(err))
+
+	reader.allowUnwind()
+	select {
+	case <-shutdownDone:
+	case <-time.After(10 * time.Second):
+		t.Fatal("shutdown did not join the canceled download")
+	}
+
+	select {
+	case <-reader.closed:
+	default:
+		t.Fatal("download resource was not closed during unwind")
+	}
+	select {
+	case <-job.done:
+	default:
+		t.Fatal("shutdown returned before the download job terminated")
+	}
+
+	server.mu.Lock()
+	state := job.state
+	server.mu.Unlock()
+	require.Equal(t, restorepb.DownloadState_DOWNLOAD_STATE_CANCELED, state)
+}
+
+func TestRestoreDownloadOutlivesInitiatingRPC(t *testing.T) {
+	t.Parallel()
+
+	ctrl := gomock.NewController(t)
+	storage := NewMockStorage(ctrl)
+	readerReady := make(chan *cancellationBlockingReader, 1)
+	storage.EXPECT().
+		GetFile(gomock.Any(), gomock.Any()).
+		DoAndReturn(func(ctx context.Context, _ string) (io.ReadCloser, error) {
+			reader := newCancellationBlockingReader(ctx)
+			readerReady <- reader
+
+			return reader, nil
+		})
+
+	server := newServerForTest(t, staticFactory(storage))
+	rpcCtx, cancelRPC := context.WithCancel(context.Background())
+	started, err := server.StartDownloadBackup(rpcCtx, &restorepb.StartDownloadBackupRequest{})
+	require.NoError(t, err)
+
+	reader := <-readerReady
+	<-reader.started
+	t.Cleanup(func() {
+		reader.allowUnwind()
+		server.Shutdown()
+	})
+	cancelRPC()
+
+	select {
+	case <-reader.canceled:
+		t.Fatal("initiating RPC cancellation reached the detached download")
+	default:
+	}
+
+	cancelDone := make(chan error, 1)
+	go func() {
+		_, err := server.CancelDownload(context.Background(), &restorepb.CancelDownloadRequest{JobId: started.GetJobId()})
+		cancelDone <- err
+	}()
+	<-reader.canceled
+	reader.allowUnwind()
+
+	select {
+	case err := <-cancelDone:
+		require.NoError(t, err)
+	case <-time.After(10 * time.Second):
+		t.Fatal("explicit cancellation did not join the download")
+	}
+
+	server.Shutdown()
+}
+
+func TestRestoreShutdownWaitsForAdmittedRequestBeforeClosingStagingStore(t *testing.T) {
+	t.Parallel()
+
+	server := NewRestoreServiceServer(t.TempDir(), "test-cluster", 1, noopLogger{})
+	store, err := dal.OpenDirect(server.stagingDir(), noopLogger{})
+	require.NoError(t, err)
+
+	server.mu.Lock()
+	server.stagingStore = store
+	server.downloaded = true
+	server.mu.Unlock()
+	require.NoError(t, server.beginRequest())
+
+	server.BeginShutdown()
+	shutdownDone := make(chan struct{})
+	go func() {
+		server.Shutdown()
+		close(shutdownDone)
+	}()
+
+	select {
+	case <-shutdownDone:
+		t.Fatal("shutdown returned while an admitted restore request was active")
+	default:
+	}
+
+	handle, err := store.NewDirectReadHandle()
+	require.NoError(t, err, "staging store must remain valid while the request unwinds")
+	require.NoError(t, handle.Close())
+
+	server.endRequest()
+	select {
+	case <-shutdownDone:
+	case <-time.After(10 * time.Second):
+		t.Fatal("shutdown did not finish after the request drained")
+	}
+
+	server.mu.Lock()
+	retained := server.stagingStore
+	server.mu.Unlock()
+	require.Nil(t, retained)
+
+	_, err = store.NewDirectReadHandle()
+	require.Error(t, err, "shutdown must close the retained staging store")
+
+	// Repeated shutdown is a no-op: it must not double-close or panic.
+	server.BeginShutdown()
+	server.Shutdown()
+}
+
+func TestRestoreShutdownWithoutActiveWorkClosesStagingStore(t *testing.T) {
+	t.Parallel()
+
+	server := NewRestoreServiceServer(t.TempDir(), "test-cluster", 1, noopLogger{})
+	store, err := dal.OpenDirect(server.stagingDir(), noopLogger{})
+	require.NoError(t, err)
+
+	server.mu.Lock()
+	server.stagingStore = store
+	server.downloaded = true
+	server.mu.Unlock()
+
+	server.Shutdown()
+
+	server.mu.Lock()
+	retained := server.stagingStore
+	server.mu.Unlock()
+	require.Nil(t, retained)
+
+	_, err = store.NewDirectReadHandle()
+	require.Error(t, err)
+}
+
+func TestRestoreDownloadRemainsAsynchronousAndSucceeds(t *testing.T) {
+	t.Parallel()
+
+	manifest, objects := restoreCheckpointFixture(t)
+	manifestData, err := json.Marshal(manifest)
+	require.NoError(t, err)
+
+	ctrl := gomock.NewController(t)
+	storage := NewMockStorage(ctrl)
+	manifestRequested := make(chan struct{})
+	releaseManifest := make(chan struct{})
+	var requested sync.Once
+
+	storage.EXPECT().
+		GetFile(gomock.Any(), gomock.Any()).
+		AnyTimes().
+		DoAndReturn(func(_ context.Context, key string) (io.ReadCloser, error) {
+			if strings.HasSuffix(key, "/backups/manifest.json") {
+				requested.Do(func() { close(manifestRequested) })
+				<-releaseManifest
+
+				return io.NopCloser(bytes.NewReader(manifestData)), nil
+			}
+
+			data, ok := objects[key]
+			if !ok {
+				return nil, fmt.Errorf("unexpected storage key %q", key)
+			}
+
+			return io.NopCloser(bytes.NewReader(data)), nil
+		})
+
+	server := newServerForTest(t, staticFactory(storage))
+	t.Cleanup(server.Shutdown)
+	started, err := server.StartDownloadBackup(context.Background(), &restorepb.StartDownloadBackupRequest{})
+	require.NoError(t, err)
+	require.NotEmpty(t, started.GetJobId())
+	<-manifestRequested
+
+	server.mu.Lock()
+	job := server.job
+	server.mu.Unlock()
+	select {
+	case <-job.done:
+		t.Fatal("StartDownloadBackup did not leave the transfer asynchronous")
+	default:
+	}
+
+	close(releaseManifest)
+	select {
+	case <-job.done:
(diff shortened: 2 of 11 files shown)


gfyrag:
Addressed the four items in the Shipfox review after rebasing onto `afdd58395fd4689c624286893c2c618e38db8148`:

1. Cleanup now asserts the `app.Stop` error in `internal/bootstrap/restore_download_lifecycle_test.go`.
2. The isolated Fx test now calls `t.Parallel()`.
3. Kept admission at the service-method boundary. All six restore RPCs call `beginRequest`/`endRequest`; `beginRequest` and `BeginShutdown` synchronize the stopping flag and WaitGroup registration through the same mutex, and `StartDownloadBackup` rechecks at job registration. An interceptor alone would miss direct service calls, including the production-module regression. The proposed error sentinel/shared fixture are optional refactors without a correctness defect, so the focused implementation is retained.
4. The CI coverage gap is fixed: the `Tests` job now explicitly runs `go test -race -tags s3 ./internal/bootstrap -run '^TestRestoreDownloadStopsWithFxApplication$' -count=1 -timeout 2m`. The tag must remain: `internal/infra/backup/s3.go` requires it, and `s3_disabled.go` otherwise rejects S3 storage creation. The regression uses an in-process HTTP endpoint and needs no MinIO/Docker. The backend-start wait is also bounded so a setup failure reports its actual phase.

No unresolved inline threads or pending reviews were present. The S3 Fx regression and both affected package race suites passed locally. Canonical validation and the final candidate are recorded in the PR description and follow-up status.

Fix commit: `8797af37b73b3619e6408c396234f1f478146cf6`. Final rebase base: `afdd58395fd4689c624286893c2c618e38db8148`; range-diff confirms the PR commits are unchanged by the subsequent HTTP/operator test-only base advances. After disk recovery, the complete exact-base `agent-check-pr` passed, including normalization, baseline and tooling race suites. The tagged Fx regression and both affected package race suites also passed. The worktree is clean. CI passed on this SHA, including the explicit S3 Fx lifecycle step. The remaining merge gate is the required human/CODEOWNER approval; no merge was performed.

gfyrag:
Follow-up to the latest Shipfox review: both optional maintainability suggestions repeat the scope decision in https://github.com/formancehq/ledger/pull/1908#issuecomment-5586127833. The method-level admission guard protects direct calls as well as gRPC calls (`server_restore.go:187`, `server_restore_download.go:69`); the additional registration check remains necessary. The shared sentinel and blocked-reader fixture extraction are optional and do not justify changing this validated candidate.

The note about pending local validation is stale: the current PR description records the completed exact-base `agent-check-pr`, including normalization, baseline and tooling race suites. The final log ends with `agent-check-pr: PASS`, and all checks on `8797af37b73b3619e6408c396234f1f478146cf6` are green. No technical blocker remains; required human/CODEOWNER approval is still pending.

internal/bootstrap/restore_download_lifecycle_test.go
```diff
@@ -0,0 +1,124 @@
...
+	case <-requestCanceled:
+	case <-time.After(10 * time.Second):
+		t.Fatal("Fx shutdown did not cancel the restore job")
+	}
+
+	select {
+	case err := <-stopDone:
+		require.NoError(t, err)
+	case <-time.After(10 * time.Second):
+		t.Fatal("Fx shutdown did not join the canceled restore job")
```
paul-nicolas, line 158: [P2][non-blocking] The dedicated CI regression does not guard the join or the phase order

**What is wrong.** This test is the one the PR adds a dedicated CI step for, and it is named after the fix ("stops with the Fx application"), but its assertions only prove that *some* cancellation happens at Fx stop. Mutation testing in an isolated worktree at this head:

| mutation in `internal/bootstrap/module_restore.go` | result |
|---|---|
| baseline | `--- PASS (0.01s)` |
| delete only the `restoreServer.Shutdown()` OnStop block | `--- PASS (0.01s)` |
| delete only the `restoreServer.BeginShutdown()` OnStop block | `--- PASS (0.01s)` |
| delete both | `--- FAIL: restore_download_lifecycle_test.go:111: Fx shutdown did not cancel the restore job` |
| swap the two registrations (join phase now runs *before* the gRPC stop) | `--- PASS (0.01s)` |

**Why it passes either way.** `Shutdown()` opens with `s.BeginShutdown()` (`server_restore.go:233`), so either hook alone produces the `lifetimeCancel()` that closes `requestCanceled` (line 109) and flips `stopping`, satisfying the `codes.Unavailable` assertion at line 122. Nothing here orders "Stop returned" after "job terminated": `stopDone` is buffered, so a value already queued before the first select is read unconditionally by the second.

**Scope.** The gap is narrow, not severe: the join, the resource close and the `DOWNLOAD_STATE_CANCELED` terminal state *are* discriminated by the untagged `TestRestoreShutdownCancelsAndJoinsActiveDownload` (50/50 failures when `<-job.done` is removed and it runs alone). What no test covers is the Fx wiring of the two-phase split -- which hook runs in which phase, and that the join happens after the gRPC stop. That is exactly what the comments at `module_restore.go:66-69` and `:132-135` assert, and a future refactor can drop or reorder either hook with this CI step green.

**Fix.** Either rename this test to what it covers (e.g. `TestRestoreDownloadIsCanceledByFxShutdown`) and add a deterministic ordering test that supplies a custom `fxevent.Logger`, records `*fxevent.OnStopExecuting.FunctionName`, and asserts `BeginShutdown` first / gRPC stop before `Shutdown` / `releaseBindingsHook` last; or make this test assert the join directly by requiring that once `stopDone` fires the job is terminal and the staging store is closed. Note the httptest handler cannot be used to hold the join open: the AWS SDK aborts the request on client-side cancellation without waiting for the handler to return.

internal/adapter/grpc/server_restore_lifecycle_test.go
```diff
@@ -0,0 +1,435 @@
...
+		reader.allowUnwind()
+		server.Shutdown()
+	})
+	cancelRPC()
+
+	select {
+	case <-reader.canceled:
+		t.Fatal("initiating RPC cancellation reached the detached download")
+	default:
+	}
```
paul-nicolas: [P2][non-blocking] The RPC-detachment assertion cannot fail

**What is wrong.** This non-blocking `select` is the only assertion in `TestRestoreDownloadOutlivesInitiatingRPC` for the property the test is named after, and it is vacuous: cancellation reaches `cancellationBlockingReader.Read` only after a goroutine hop, so the `default` arm wins deterministically. The later `<-reader.canceled` wait comes *after* an explicit `CancelDownload`, which an already-canceled reader satisfies just as fast, so it adds nothing.

**Proof by mutation** (isolated worktree at this head):
- Derive the job context from the initiating RPC context (`server_restore_download.go:99` -> `context.WithCancel(ctx)`): `20/20 PASS`, flake rate 0/20. The named test does not discriminate the regression it exists for. That crude mutation is caught elsewhere, but only indirectly and only as a 60s deadlock in `TestRestoreShutdownCancelsAndJoinsActiveDownload:106`, because it also severs the lifetime linkage.
- Surgical detachment-only regression -- keep `context.WithCancel(s.lifetimeCtx)` and add `go func() { <-ctx.Done(); cancel() }()`: **all 15 restore tests PASS**. Nothing in the package catches it. The detachment invariant, which `docs/ops/cli.md` and `grpc-api.md` both describe as deliberate, is currently guaranteed only structurally by `StartDownloadBackup(_ context.Context, ...)` discarding its ctx.

**Fix (verified).** Replace the select with a bounded negative assertion:

```go
require.Never(t, func() bool {
	select {
	case <-reader.canceled:
		return true
	default:
		return false
	}
}, 200*time.Millisecond, 10*time.Millisecond, "initiating RPC cancellation reached the detached download")
```

Measured: `20/20 PASS` at head (~0.2s per run), `5/5 FAIL` under the parent-swap mutation, and FAIL under the surgical mutation with `Condition satisfied`. `require.Never` is the repo-sanctioned form (no `time.Sleep`). Pairing it with a positive liveness check -- `GetDownloadStatus` still reporting `DOWNLOAD_STATE_RUNNING` after `cancelRPC()` -- would make the test prove detachment in both directions.

The identical vacuous shape at lines 114-118 and 119-123 is harmless: those properties are really asserted by the post-`allowUnwind` checks at 135-144, which is what catches a removed join.

docs/ops/backup-restore.md
```diff
@@ -377,6 +377,14 @@ The server-side job:
 If the job fails or is cancelled, the staging directory is wiped so the
 operator can retry with a fresh state without restarting the server.
 
+The transfer is detached from the initiating RPC, not from the restore-mode
+application. Fx shutdown first rejects new restore RPCs and cancels any active
+download, then stops the gRPC server, joins admitted RPCs and the download job,
+and finally closes the retained staging Pebble store. Explicit
+`CancelDownload` keeps its client-facing behavior: it cancels only the selected
+job, waits for a bounded drain, and leaves the restore-mode application running
+so the operator can retry.
```
paul-nicolas, line 387: [P2][non-blocking] State the stop-timeout bound and the residual state, per the durability-docs rule

**What is wrong.** This paragraph presents the teardown as a sequence that always completes ("...joins admitted RPCs and the download job, and finally closes the retained staging Pebble store"), and `docs/ops/cli.md:3602-3604` adds "RPC detachment never lets the transfer outlive the server application". Both omit the bound and the residual state when the bound is hit. `CLAUDE.md` is explicit that durability and recovery documentation "must state the actual in-memory mutation order, durable-write order, and residual state after failure; do not describe the intended end state as if it were the execution sequence", and this is the restore runbook an operator follows.

**The actual bound.** `Shutdown()` ignores the OnStop context (`module_restore.go:72`), but the stop as a whole does not: go-libs applies `fx.StopTimeout(--total-stop-timeout)`, defaulting to `fx.DefaultTimeout` = 15s (`go-libs/v5@v5.7.2/pkg/service/app.go:30,115`; `fx@v1.24.0/app.go:45`), and nothing in this repo, its docs, the helm charts or `misc/operator` overrides it. `fx.withTimeout` (`fx/app.go:787-820`) abandons the blocked hook goroutine at the deadline, so `app.Stop` returns `context.DeadlineExceeded`, `service.Run` propagates it and the process exits 1 (`pkg/service/execute.go:12-14`). Residual state on that path: `closeStagingStore()` never ran, so the staging Pebble handle is not cleanly closed, and the remaining OnStop hooks are skipped.

**Why it is reachable on the ordinary path.** An Fx stop cancels the job through `lifetimeCtx`, so `finishJob` runs `os.RemoveAll(s.stagingDir())` (`server_restore_download.go:454`, and `:471` for any other error) *inside* the `<-job.done` join, because `defer close(job.done)` runs after `finishJob`. Deletion cost is bounded by file count rather than bytes (`maxRestoreManifestFiles = 100_000`), so exceeding 15s is a storage-dependent tail risk rather than the routine outcome -- but it does not need a misbehaving backend to happen.

This is not a regression: at merge base `afdd58395` Fx stop cancelled nothing, joined nothing and exited 0 with the download still running. The gap is that the new runbook prose promises more than the wiring guarantees.

**Fix.** Add a sentence here and in `docs/ops/cli.md` stating that the join is bounded by `--total-stop-timeout` (15s by default), that exceeding it aborts teardown with a non-zero exit and an unclosed staging store, and that operators should size the termination grace period accordingly when a download may be in flight. Reword the `Shutdown` comment at `server_restore.go:227-231` to match (its "remain visibly stuck" describes an unbounded hang that cannot occur). Gating the staging wipe on the application lifetime in both `finishJob` branches would remove the main way the budget is consumed -- `prepareDownload` already wipes at `:284-290`, so the shutdown-time wipe is redundant.

paul-nicolas (commented):
## Multi-model review (Claude + Codex), head `8797af37b`

First review of this PR (no prior inline comments or reviews existed).

**Verification gate:** 4 findings entered adversarial verification (one subagent per finding, instructed to refute, mutation experiments run in isolated detached worktrees). 0 rejected as invalid, 4 corrected (`adjust`): 2 were downgraded P2 -> P3 after base comparison and now sit below the inline threshold, 2 were confirmed by mutation and kept at P2.

**Posted inline:** 3 findings, all P2 and all non-blocking. No P0/P1 survived verification.

**Conflict resolved:** Codex filed a single finding, `[P1][blocking] Shutdown can block forever on a canceled download`, on the premise that a storage backend ignoring context cancellation parks the Fx stop hook forever. Verification downgraded it to P3 non-blocking on three grounds: (1) go-libs applies `fx.StopTimeout(--total-stop-timeout)`, default `fx.DefaultTimeout` = 15s (`go-libs/v5@v5.7.2/pkg/service/app.go:30,115`; `fx@v1.24.0/app.go:45`), and `fx.withTimeout` (`fx/app.go:787-820`) abandons the blocked hook goroutine at the deadline, so `app.Stop` returns `DeadlineExceeded` and the process exits 1 through `service.Execute` (`execute.go:12-14`) rather than hanging; (2) the failure is therefore loud, which is what the code comment's substantive claim asks for; (3) at merge base `afdd58395` Fx stop cancelled nothing, joined nothing, closed no staging store and exited 0, so every timeout scenario here is strictly better than base. The real trigger is not a non-compliant backend but `finishJob`'s staging wipe running inside the join - see the P3 list.

**Verified as correct, no finding (recorded so it is not re-litigated):**
- The two hook-ordering comments in `internal/bootstrap/module_restore.go` are accurate. Real append order is releaseBindings, `Shutdown`, cert-reloader, gRPC, HTTP, `BeginShutdown`; fx walks OnStop backward, so the sequence is `BeginShutdown` -> HTTP stop -> gRPC stop -> cert reloader -> join + staging close -> release listeners. No competing Fx hook owns the staging Pebble handle, and `releaseBindingsHook` still closes last.
- `sync.WaitGroup` use is sound: `Add` only happens under `s.mu` with `!s.stopping`, and `BeginShutdown` sets `stopping` under the same mutex before `Wait`. No Add-after-Wait window. The recheck at `server_restore_download.go:79` is load-bearing for job registration, not for the WaitGroup.
- All six `restorepb.RestoreServiceServer` methods carry `beginRequest`/`endRequest`, and `defer s.endRequest()` is registered first in each handler so it runs after the inner handle/cursor closes.
- No use-after-close from the download path: `close(job.done)` is the outermost defer, so `finishJob`'s store publish and both wipes happen-before `Shutdown`'s `<-job.done` and `closeStagingStore`.
- No deadlock: `CancelDownload`'s drain is capped at 5s and holds no lock, `BeginShutdown` calls `cancel()` after releasing `s.mu`, no RPC holds `s.mu` across a blocking call, and `checker.Check` honors the stream context which the 2s-capped `stopGraceful` force-cancels.
- `codes.Unavailable` does not trigger a retry storm: `actions.GRPCRetryPolicy` (50 attempts at 0.2s) is installed only by `cmdutil.GetClient` and the cluster clients; `getRestoreClient` (`cmd/ledgerctl/restore/command.go:42-45`) does not use it, so the operator sees the error on the first attempt.
- `docs/technical/contributing/testing.md` is accurate: the test carries `//go:build s3`, points at an in-process `httptest` server with static credentials, and needs no MinIO or Docker. The CI step is in the `Tests` job with the nix env, the tag and `-run` regex both match, and the test genuinely runs (it is the only s3-tagged file in `internal/bootstrap`, so nothing is newly compiled-but-unrun).
- No protocol revision bump owed: `pkg/grpcprotocol` does not exist at this PR's base, and a shutdown-time `Unavailable` is availability behavior, not a change to request/response interpretation.
- Test helpers are reused, not duplicated (`newServerForTest`, `staticFactory`, `noopLogger`, `NewMockStorage`); no `time.Sleep` anywhere in the new tests; private-field reads are correctly guarded by `server.mu`.

### P3 (not posted inline)
- `internal/adapter/grpc/server_restore.go:227-231` -- the comment's "shutdown must remain visibly stuck" describes an unbounded hang that cannot occur. `fx.withTimeout` abandons the hook at `--total-stop-timeout` (15s default, never overridden in this repo), so `closeStagingStore` never runs and the process exits 1. Reword to say the join is bounded by that flag and that exceeding it aborts teardown, mirroring the 5s cap already documented on `CancelDownload`.
- `internal/adapter/grpc/server_restore_download.go:454` and `:471` -- an Fx stop now unwinds the job into `finishJob`, and both terminal branches run `os.RemoveAll(s.stagingDir())` inside the `<-job.done` join. Cost is bounded by file count (`maxRestoreManifestFiles = 100_000`), so overrunning 15s is a storage-dependent tail risk rather than routine, but the work is pure waste on the stop path because `prepareDownload` already wipes at `:284-290`. Gate it on the application lifetime in *both* branches (`if s.lifetimeCtx.Err() == nil { ... }`) -- a branch swap will not do, since a lifetime cancellation surfacing as an S3/Azure API error lands in `default`. Verified safe: `dal.ValidateFreshRestoreTarget` inspects only `live*/` and `checkpoints/`, never `restore-staging`, and the SUCCEEDED-then-SIGTERM path already leaves the tree on disk today.
- `internal/adapter/grpc/server_restore_download.go:205-222` -- `runDownloadJob` has no `defer job.cancel()`. `job.cancel` is invoked only by `CancelDownload`, so the SUCCEEDED and FAILED paths leak a `cancelCtx` child. This is newly reachable: the parent used to be `context.Background()` (not cancellable, no child registered), and is now `s.lifetimeCtx`, which retains every uncancelled child for the process lifetime. A FAILED job leaves `downloaded == false`, so each retry overwrites `s.job` and drops the previous `cancel`.
- `internal/adapter/grpc/server_restore_download.go:26` -- `storageFactory` takes no context, so `prepareDownload`'s first step structurally cannot observe `jobCtx`: `defaultStorageFactory` -> `backup.NewStorage` -> `newS3Client` -> `awsconfig.LoadDefaultConfig(context.Background(), ...)` (`internal/infra/backup/s3.go:138`), which does IMDS/STS network I/O under the AWS SDK's own timeouts whenever credentials are not static. `BeginShutdown` cannot cancel a job in its pre-transfer window; the join just waits it out, which contradicts the model stated at `:97-98`.
- `internal/adapter/grpc/server_restore.go:232-247` -- `Shutdown()` returns nothing and `closeStagingStore` only logs a failed `Close()`, so the OnStop hook always returns nil. The same module deliberately does the opposite (`grpc_server_hook.go:126-132` returns `serveErr`, `releaseBindingsHook` returns `errors.Join`). A Pebble close failure on the staging store is the last durability-relevant action in restore mode and currently exits zero.
- `internal/bootstrap/module_restore.go:132-135` -- "Registered last so it is the first OnStop hook" holds only within `RestoreModule`. go-libs appends a grace-period OnStop hook *after* the application options when `--grace-period` is non-zero (`pkg/service/app.go:128-142`), making that hook first and consuming part of the stop budget. Comment-only; qualify it as "the first OnStop hook registered by this module".
- `docs/ops/backup-restore.md:380-382` -- the stated sequence omits the HTTP `/health` server stop, which actually runs between `BeginShutdown` and the gRPC stop. Immaterial operationally, but the paragraph presents itself as the actual teardown order.
- `internal/adapter/grpc/server_restore_lifecycle_test.go:135-144` -- the post-`allowUnwind` join assertions are the ones that really discriminate a removed `<-job.done` (9/10 runs alone), but they escaped the same mutation in 2 of 5 and 1 of 10 runs when executed alongside the other parallel restore specs. `require.Eventually`-style synchronization would make them deterministic.

### Pre-existing, not attributed to this PR (no action required here)
`FinalizeRestore` calls `closeStagingStore()` (`server_restore.go:442`), then hard-links and `os.RemoveAll`s the staging directory, while a concurrently admitted `ValidateRestore`/`PreviewRestore` still holds a handle captured at `:276`/`:338`. `activeRequests` counts admitted RPCs, it does not serialize them, and `dal.Store` guards only *new* handle creation with `ErrStoreClosed`. Present at merge base (`afdd58395:internal/adapter/grpc/server_restore.go:354`), so this PR does not introduce it -- noted only because it is the one staging-store lifetime hole the new ownership model does not close. Exclusive access (an `RWMutex` around the handle, or refusing `FinalizeRestore` while a reader RPC is admitted) would close it in a follow-up.

internal/bootstrap/restore_download_lifecycle_test.go
gfyrag: Fixed in 3ef3bec1280754c19e8475201901c87326de5c08. The existing S3 regression now records Fx OnStopExecuting events and requires BeginShutdown → HTTP → gRPC → Shutdown. The two restore registrations use equivalent fx.StopHook bound methods so the observer does not depend on anonymous closure numbers. The original real-S3 cancellation trigger, test name and dedicated CI command are retained.

Verified independently with the tagged race test: deleting BeginShutdown, deleting Shutdown, or swapping the registrations each fails at the exact phase-order assertion. The gRPC join regression was also strengthened with synctest.Wait; removing the job join is now caught deterministically before the blocked reader is released.

internal/adapter/grpc/server_restore_lifecycle_test.go
gfyrag: Fixed in 3ef3bec1280754c19e8475201901c87326de5c08. The test now runs inside testing/synctest and calls synctest.Wait after cancelRPC, allowing all runnable cancellation work to complete before asserting that the reader has not been canceled. It also checks the live job status: PENDING is the correct state here because the fake reader is deliberately still blocking manifest preparation, before the RUNNING transition.

A surgical mutation retaining the lifetime parent but forwarding RPC cancellation through context.AfterFunc fails at the intended detachment assertion in all 3 runs. The corresponding missing-join mutation also fails in all 3 runs. No wall-clock negative window or time.Sleep is needed.

docs/ops/backup-restore.md
gfyrag: Partially addressed in 3ef3bec1280754c19e8475201901c87326de5c08: the runbook, CLI/API docs and Shutdown comment now spell out the actual sequence, blocking initialization/filesystem cleanup, and residual state after forced termination. HTTP stop and the optional preceding grace period are included.

The proposed 15-second bound does not apply to this executable. go-libs/v5@v5.7.2/pkg/service/app.go:87-90 calls Fx app.Stop with context.Background wrapped only with context values. Fx v1.24.0 app.go:716-731 passes the supplied context directly to withTimeout; its StopTimeout-derived deadline is created in Fx's own App.run at app.go:617-620, which this service runner does not call. Therefore withTimeout's ctx.Done arm cannot fire on the production path merely because --total-stop-timeout is configured.

Verified against the pinned dependency with a bounded experiment: fx.StopTimeout(1ms) plus app.Stop(context.Background()) remained blocked beyond 20ms and completed only after explicitly releasing the hook. A custom caller supplying its own deadline can return early; the API docs distinguish that case. The existing staging wipe remains inside the join, and the docs state that synchronous initialization and filesystem work can delay exit. I have not added a misleading 15s/non-zero-exit guarantee or changed the existing cleanup policy.

gfyrag:
Addressed the review on 8797af37 in 3ef3bec1280754c19e8475201901c87326de5c08:

- The Fx regression now verifies actual shutdown phase order. Removing either hook or swapping them fails the assertion. The original S3 trigger and dedicated CI step remain.
- Detachment and joining use testing/synctest to eliminate the scheduler gap. Surgical RPC-cancellation forwarding and missing-join mutations fail in all 3 runs each.
- Terminal jobs now defer job.cancel before job.done closes, releasing child contexts after both success and failure. Both terminal paths have additive regression assertions; removing cancellation fails both tests.
- The module comment now accounts for go-libs' preceding grace hook, and the docs include HTTP stop, blocking client initialization/filesystem cleanup, forced-termination residue, and the existing log-only staging-close policy.

The timeout claim is refuted by the actual service runner: go-libs calls Fx Stop with context.Background, while Fx applies StopTimeout in its own App.run path, which is not used here. A pinned-dependency probe confirmed StopTimeout(1ms) does not terminate a direct Stop(context.Background()) after 20ms. The detailed inline reply cites the source locations.

The remaining P3 suggestions are acknowledged scope choices, not classified as invalid: skipping the pre-existing staging wipe is an optional cleanup-policy change; threading cancellation through shared storage constructors and propagating the existing close helper's logged errors are broader lifecycle/error-policy changes. Their current limitations are now explicit in the docs. The pre-existing FinalizeRestore/read concurrency issue is unchanged by this PR.

gfyrag:
CI is green on `3ef3bec1280754c19e8475201901c87326de5c08`: 15 successful checks and the two expected release/image skips. The validation checklist is updated.

The latest Shipfox review repeats the two optional extractions already dispositioned in https://github.com/formancehq/ledger/pull/1908#issuecomment-5598867818; the same scope decision applies. Its additional terminal-state observation does not affect cancellation, joining, or cleanup and does not require a behavior change here. All three inline threads are resolved. Required human/CODEOWNER approval remains pending.

paul-nicolas (approved):

PR #1912, opened 2026-09-07
fix(EN-1526): verify audit-only histories and bound stored logs by the audit chain

paul-nicolas:
Jira: https://formance-team.atlassian.net/browse/EN-1526

## Problem

`Check()` returned early whenever `query.ReadLastSequence` reported 0, above the only call site of `verifyAuditHashChain`. A store whose whole history is rejected proposals holds a complete audit chain, frozen idempotency outcomes, and no `Log` rows, so none of it was verified. `ReadLastSequence` reads the `sequence` field inside the last Log row's value, which is not hash-bound, so editing that one field on a populated store also took the early return and skipped every pass.

Separately, no pass could see a deleted log tail: drop the top N logs and every projection rebuilt from the survivors is self-consistent. That is the restore failure mode the audit chain exists to witness.

## Changes (three commits)

1. `docs(EN-1526)`: `audit-chain.md` claimed a Failure writes zero `AuditItem` rows. The FSM writes one per order with `LogSequence = 0`, hashed into the chain like success items. Corrected the table, prose, and the matching comment in `internal/infra/backup/manager.go`.
2. `fix(EN-1526)`: remove the `lastSequence == 0` early return. The four comparators the branch duplicated already run on the normal path; their rationale comments move there. Emit the final progress position unconditionally once. Report a Log row whose value `sequence` disagrees with its key as `CHECK_STORE_ERROR_TYPE_LOG_SEQUENCE_MISMATCH` (new, 27).
3. `feat(EN-1526)`: `logBoundsVerifier` derives the expected highest log sequence from chain-verified `AuditSuccess` ranges, asserts the ranges are contiguous from 1 (invariant #7 on violation), and compares against the highest stored Log key. Deleted tail: one aggregated `SEQUENCE_GAP`. Logs above the audited bound: `CHECK_STORE_ERROR_TYPE_LOG_UNAUDITED` (29). Chain break or inverted/discontinuous range: one `CHECK_STORE_ERROR_TYPE_LOG_VERIFICATION_INCOMPLETE` (28) and the comparison is suppressed, following the signing/cluster-policy pattern.

Scope deltas against the ticket text: chapters and cold storage were removed by EN-1945, so the archive-boundary items are moot. The expected log set is a contiguous interval derived from ranges, not a per-item bijection; per-item duplicate and failure-associated sequences cannot exist by construction. Coordinated truncation of both the top logs and the top audit entries remains undetectable and is now documented in `audit-chain.md`.

## Invariants

Strengthens #8 (audit is business truth) and complies with #7 (the removed clean-returning gate, the invariant reports on inverted or discontinuous ranges). No FSM, cache, or apply-path change. No persisted-format change.

## Tests

17 new tests in `internal/application/check` (`checker_audit_only_test.go`, `checker_log_bounds_test.go`, `clusterpolicy_test.go`), all hermetic and parallel. Mutation checks: re-adding the gate fails the audit-hash, idempotency, and log-sequence tests; the restore-shaped lost-tail fixture reports nothing on the parent commit and exactly one finding on this branch; reverting the inverted-range and duplicate-progress fixes fails their tests.

## Validation

- `bash scripts/agent-check`: PASS
- `golangci-lint run --max-same-issues=0 ./internal/application/check/...`: 0 issues
- `go test ./internal/application/check/... -count=1`: ok, 17 new tests, all with mutation proofs
- `AI_REVIEW_BASE_SHA=e3ef77087 bash scripts/agent-check-pr`: every stage passes except one pre-existing flake in `scripts/internal/testenv` (`TestRunSynchronizedTimeoutTerminatesAndReportsEveryPeer`, a 200ms fixture budget missed under full-suite load; untouched by this branch, passes 6/6 alone with `-race` on both branch and base). Because that failure aborts `just test` before the `test-e2e` stage, the e2e gates were run separately:
- `go test -race -tags e2e -p 1 ./tests/e2e/cluster/...`: ok (light set)
- `go test -race -tags "e2e,s3" ./tests/e2e/cluster/... -ginkgo.focus='Restore|Bootstrap'` against OrbStack: 95/95 specs, no `SEQUENCE_GAP`, `LOG_UNAUDITED`, `LOG_VERIFICATION_INCOMPLETE` or `LOG_SEQUENCE_MISMATCH` on any restored store
- `go test -race -tags e2e ./tests/e2e/business/... -ginkgo.focus=CheckStore`: 21/21 specs
- Full business suite: fails identically on this branch and on base `e3ef77087` in the cross-store value-skew / snapshot-alignment specs (index-builder backlog not drained within 180s on this machine, then `server.Stop` exceeds its 10s cleanup timeout and the temp dir is removed under the running builder, which panics on the vanished `.sst`). Pre-existing, environment-bound, not on the checker path.

## Docs

- `docs/technical/architecture/subsystems/checker/checker.md`
- `docs/technical/architecture/subsystems/checker/audit-chain.md`


internal/application/check/checker.go
```diff
@@ -361,6 +346,22 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
...
+		// projection is keyed on, so replaying it would fold unverifiable data
+		// into the expected state.
+		if log.GetSequence() != seq {
+			callback(errorEvent(servicepb.CheckStoreErrorType_CHECK_STORE_ERROR_TYPE_LOG_SEQUENCE_MISMATCH,
+				fmt.Sprintf("log at key sequence %d carries sequence %d in its stored value", seq, log.GetSequence()),
+				seq, "", "", ""))
+
+			continue
+		}
+
```
paul-nicolas, line 489: [P2][non-blocking] The replay-skip after LOG_SEQUENCE_MISMATCH has no regression trigger, and its deterministic cascade is neither pinned nor documented

Evidence: deleting only the `continue` here leaves `go test ./internal/application/check/... -count=1` green, so the skip itself is untested (TestCheck_LogSequenceFieldMismatch at checker_audit_only_test.go:407-416 asserts one LOG_SEQUENCE_MISMATCH and its key sequence, nothing about the rest of the event set). On this head every case in that table cascades: "last log claims sequence zero" yields 6 events (1 mismatch + 2 VOLUME_MISMATCH + 1 TRANSACTION_UPDATE_MISMATCH "unexpected transaction in live store for tx 3 (no matching log)" + 2 BOUNDARY_MISMATCH), "middle log claims another sequence" yields 8, and hitting the create_ledger log yields 11 including UNKNOWN_LEDGER and UNAUDITED_LEDGER. The collateral messages point at volumes, boundaries and transactions that are consistent with the audit; the log exists, the checker chose not to read it.

The value's `sequence` field is consumed nowhere after this check: ReplayLedgerLog, compareTransactionPostCommitVolumes, checkReversionInvariants, verifySkippedOrder, dispatchElisionCheck and progress all take the key-derived `seq`. With the `continue` removed, each case reports exactly one LOG_SEQUENCE_MISMATCH and nothing else. The rationale in the comment above ("replaying it would fold unverifiable data") applies equally to every Log payload, none of which is hash-bound, so skipping this one row buys no verification.

Impact: operator triage noise on the exact store shape this finding was added for, plus an untested production branch. No ledger-correctness effect.

Fix: drop the `continue` so the row is reported and replayed under its key, tighten the test to `require.Len(t, errs, 1)` per case, and reword the comment. If the skip is intentional, pin the full cascade per case (require.Len plus per-type expectations, including the 4 -> 0 headline case) and document it in checker.md row 14.

internal/application/check/checker.go
```diff
@@ -2046,6 +2102,7 @@ func (c *Checker) verifyAuditHashChain(
 			// reports the gap instead of those false positives.
 			signing.markLiveTruncated()
 			policy.markLiveTruncated()
+			bounds.markLiveTruncated()
```
paul-nicolas: [P2][non-blocking] Two of the three `bounds.markLiveTruncated()` exits have no regression trigger

Evidence: deleting this call (smuggled-items exit) or the one at :2146 (header error) leaves the package green; only the hash-mismatch exit at :2192 is covered (TestCheck_LogBounds_ChainBreakAboveTruncationSuppressesGap fails when it is removed). checker.md (Stored log bounds) states that "each of verifyAuditHashChain's three non-error exits calls bounds.markLiveTruncated()", and docs/technical/contributing/testing.md requires an independent case per production trigger. The signing verifier already meets that bar for all three exits through TestVerifyAuditHashChain_MarksSigningFoldTruncatedOnEveryBreak (checker_tamper_test.go:482-528, one table case per exit including smuggled items and a nil outcome); the bounds verifier is at 1/3 and, as a pre-existing aside, clusterPolicyVerifier is at 0/3.

Impact: a future refactor of these exits can drop the suppression silently, and the bound pass would then compare a partial expectedMax against the stored head and emit a false LOG_UNAUDITED or SEQUENCE_GAP instead of one LOG_VERIFICATION_INCOMPLETE. Head behaviour is correct; the exposure is regression-only.

Fix: runChainVerifierWithSigning (checker_tamper_test.go:439-466) already constructs newLogBoundsVerifier() inline and discards it; return it and assert incompleteReason is non-empty in all three existing table cases (and empty on the clean run). Returning the clusterPolicyVerifier the same way closes its gap for free.

docs/technical/architecture/subsystems/checker/audit-chain.md
```diff
@@ -193,41 +193,55 @@ Two independent monotone counters, bridged per successful non-replayed proposal:
...
 
-The `internal/infra/backup/manager.go` incremental export is the canonical example: each of the three companion segments (`audit`, `auditItem`, `appliedProposal`) is guarded on its own count when appended to the manifest. Failure-only ranges produce an `audit` segment with `count > 0`, an `auditItem` segment with `count == 0` (skipped from the manifest), and no `appliedProposal` segment at all.
+What the outcome does **not** change is the `AuditItem` count: every non-replayed proposal writes one item per order regardless of outcome, so a range of audit entries whose proposals each carried at least one order always has `auditItem.count > 0`. The remaining way to reach an empty `AuditItem` range is a proposal that carried no orders. Nor does `LogSequence` identify the outcome: an item reads `0` on a failure, but also on a success whose order neither created a log nor resolved to a positive in-batch reference sequence. Only `AuditEntry.outcome` distinguishes the two.
```
paul-nicolas: [P2][non-blocking] The "proposal that carried no orders" empty-AuditItem case does not exist: order-less proposals write no AuditEntry at all

Evidence: applyProposal returns before the audit path when `len(proposal.GetOrders()) == 0` (internal/infra/state/machine.go:1183-1192, "the audit-entry path is skipped entirely"), and the apply loop skips entries with no orders and no technical updates (:614). The only AuditEntry writer is writeAuditEntry (:1343-1417, sole appendAuditEntries caller at :1407); its items come from marshalOrdersForAudit(orders) with no filtering, and buildAuditItems (internal/infra/state/audit.go:65-100) returns one item per serialized order unconditionally. All three callers (:1424 failure, :1473 transient-validation failure, :1567 success) sit below the :1183 return. So every persisted entry has OrderCount >= 1 and at least one AuditItem, and an empty auditItem range over a non-empty audit range cannot come from the FSM.

The same claim is repeated in the internal/infra/backup/manager.go:533-534 comment ("The range can still be empty (a proposal that carried no orders)"), and line 184 above calls idempotent replay "the one exception" to one-entry-per-proposal while the technical-only proposal is a second no-entry path (machine.go:1149-1150 says so explicitly).

Impact: this section is the stated reference for code that scans the audit range alongside a companion stream; a reader will classify an empty auditItem range as legitimate rather than as corruption, and the manager.go guard is justified by a cause that cannot occur.

Fix (wording only): state that SubHistoryAuditItem is dense for every audit_seq with no exception; call an empty auditItem range over a non-empty audit range a corruption signal; describe the manager.go guard as defence in depth; name the technical-only proposal as the second no-entry case at line 184.

paul-nicolas (commented):
Multi-model review (Claude + Codex) of head 78a150b62.

Findings that survived adversarial verification: 3 posted inline, all P2 non-blocking. 0 P0, 0 P1, 0 blocking. Verification gate: 4 candidates entered, 1 confirmed as filed, 3 corrected, 0 rejected; one candidate (Codex P2 blocking) was downgraded to P3 on evidence that it is pre-existing and tamper-only, so it is listed below instead of inline. Validation considered: CI green on all 14 running checks; `go build`, `go vet`, and `go test ./internal/application/check/... -count=1` pass at head; the PR's mutation claims were re-run and hold (re-adding the `lastSequence == 0` gate fails 8 tests, disabling the bounds comparison fails all 7 bounds tests). Contiguity premise re-verified against the two `IncrementNextSequenceID` producer sites, `WriteSet.Merge`, the idempotent-replay short-circuit, and the overlay commit path. Residual risk: LOW.

P3 (not posted inline):
- internal/application/check/checker.go:669 -- reverseMapOrphanScope.lastSequence (and TotalLogs at :583/:609) still come from query.ReadLastSequence, the value field this PR documents as not hash-bound; with a peer attached, lowering the last row's value emits a misattributed REVERSE_MAP_ORPHAN ("read index is ahead of the verified log range") beside the correct LOG_SEQUENCE_MISMATCH. Pre-existing on base, tamper-only; storedMaxLogSeq covers :609/:669 but the in-loop :583 needs a key-derived head read before the loop.
- internal/application/check/log_bounds.go:213 -- LOG_UNAUDITED "(%d logs)" is a position-range size, not a row count (an injected row at key 1000 over a chain ending at 10 reports "990 logs"); the proto comment for value 29 promises a row count.
- misc/proto/bucket.proto:884 -- value 28 comment names two causes (chain break, non-contiguous); observeSuccess also fires on an inverted range.
- internal/application/check/checker.go:604 -- progress dedup is pinned only for head == progressInterval; the in-loop-plus-distinct-final-event shape (e.g. 150 logs) is uncovered.
- internal/application/check/log_bounds.go:104 -- first-cause-wins / fold stops after incomplete is uncovered by tests.
- internal/application/check/checker_audit_only_test.go:305 -- idempotency tamper cases assert NotZero instead of an exact count.
- internal/application/check/checker_log_bounds_test.go:585 -- vacuous for-loop where require.Empty is meant.
- internal/application/check/checker_log_bounds_test.go (persistSuccessAuditEntries / RestoreLostTail fixture) -- success entries with OrderCount 0 and a non-empty log range, a shape the FSM never produces; the "nothing else notices" single-finding claim is proven only for item-less tail entries.
- docs/technical/architecture/subsystems/checker/checker.md:34 -- interior SEQUENCE_GAP scan attributed to pass 1 (verifyAuditHashChain); it lives in Check()'s log loop (also :125, :145).
- docs/technical/architecture/subsystems/checker/checker.md:171 -- flowchart and "independent passes" paragraph omit the bound pass, its two feeds (observeSuccess in the chain walk, storedMaxLogSeq from the loop), and the in-loop LOG_SEQUENCE_MISMATCH check.
- docs/technical/architecture/subsystems/checker/checker.md:5 -- progress contract (one final event guaranteed, logs_checked key-derived vs total_logs value-derived) documented only in code comments.
- docs/technical/architecture/subsystems/checker/audit-chain.md:223 -- "Where the chain itself is cut short" reads as tail truncation in a section about tail truncation; the code emits INCOMPLETE only on a hash-mismatch stop or a bad range.
- docs/ops/cli.md:1982 -- `store check` error-name list does not include LOG_SEQUENCE_MISMATCH / LOG_UNAUDITED / LOG_VERIFICATION_INCOMPLETE, which the CLI prints.

Design questions (no defect claimed): reusing SEQUENCE_GAP for the aggregated tail while the symmetric case got a dedicated type (LOG_UNAUDITED); LOG_VERIFICATION_INCOMPLETE carrying both the benign chain-break cause and the "invariant:" contract-breach cause under one type.

Conflict resolved: Codex rated the value-derived lastSequence alignment P2 blocking and coupled it with the replay-skip cascade; verification split them, kept the cascade at P2 (posted) and downgraded the alignment to P3 (pre-existing, tamper-only, correct finding emitted in the same run).

internal/application/check/checker.go
paul-nicolas: Fixed in 5215b469 -- the finding holds and the fix is applied as proposed.

The `continue` is gone, so the divergent row is reported and replayed under its key. Confirmed the value's `sequence` field is not an input to replay: every in-loop consumer takes the key-derived `seq` (checker.go:459, :473, :531, :533, :559, :582, :608). Also confirmed the skip was jumping past `dispatchElisionCheck` at :559, whose own comment says it runs at every sequence precisely so it fires for tampered projections that never reach the well-formed Apply branch -- so the skip was suppressing a check documented as unconditional.

`require.Len(t, errs, 1)` added, plus the create_ledger row (`{key 1, value 7}`) as a fourth case. Mutation proof: restoring the `continue` now fails all four cases with 11 / 6 / 8 / 6 events -- your predicted counts reproduce exactly, and the create_ledger shape is the 11-event one. The cascade included `"unexpected transaction in live store for tx 3 (no matching log)"` for a log that exists and is readable, which is what settled it: those events describe the checker's own hole, not the store.

Took the primary fix rather than the alternative, for the reason you gave: pinning the cascade would freeze into a regression test a set of events that make false statements about the store, and would churn on every unrelated change to the volume, boundary and ledger-liveness passes.

One correction to the finding as posted: "neither pinned nor documented" is too strong on the docs half. Row 14 does exist at checker.md:47 and describes the check; the skip surfaced at checker.md:139 in a subordinate clause justifying a different design point ("a row skipped as LOG_SEQUENCE_MISMATCH still counts as a position the store occupies"). That sentence contradicted the new behaviour, so both it and row 14 are updated.

Not addressed here, and worth its own ticket: a tampered last-log `sequence` field still moves `lastSequence` (checker.go:128) and therefore the reverse-map alignment gate (reverse_map_orphans.go:223) -- value 0 trips the AHEAD `assert.Unreachable` branch, value 99 silently downgrades the pass to key-decoding only. Pre-existing on base and unaffected by either variant of the `continue`. The clean fix is for that gate to take `storedMaxLogSeq` as `logBoundsVerifier.compare` already does.

internal/application/check/checker.go
paul-nicolas: Fixed in b0ab162e -- confirmed by mutation, including the exact 1/3, 3/3, 0/3 split.

Reproduced before the change: deleting `bounds.markLiveTruncated()` at the smuggled-items or header-error exit left the package green; only the hash-mismatch exit was covered. `clusterPolicyVerifier` was indeed 0/3, so it is closed in the same change as you suggested.

`runChainVerifierWithSigning` is now `runChainVerifierWithFolds`, returning a `chainVerifierFolds{signing, policy, bounds}` struct rather than four return values. The existing table asserts all three folds per exit, and the existing `TestCheck_LogBounds_ChainBreakAboveTruncationSuppressesGap` is untouched and still fires on the hash-mismatch exit. Mutation matrix after the change: all nine calls RED.

Two corrections where the proposal was incomplete:

1. The clean-run half ("assert `incompleteReason` empty on the untampered run") fails at head for an unrelated reason. `newRichAuditEntry` declares a success range of `100..101` (checker_tamper_test.go:195-196), which leaves a hole below the range and violates `logBoundsVerifier`'s contiguity-from-1 premise -- so `observeSuccess` already marked the bound incomplete on the baseline run and would have masked the wiring under test. Moved to `1..2`; no assertion depended on the old values.

2. `require.NotEmpty` on `incompleteReason` would have passed for the wrong reason, since `observeSuccess`'s contiguity guard writes the same field -- a different production branch. Used `require.Contains(..., "cut short by a hash chain break")` instead, per testing.md's requirement that failure-path assertions uniquely identify the intended branch.

No doc change needed: checker.md:113 and :137 already assert what the widened test now pins.

Methodology note worth propagating, since it nearly inverted this result: the first mutation pass used `sed -i ''`, and this host's `sed` is GNU sed, which parses that as an empty suffix and silently mutates nothing -- reporting a false GREEN on all three sites. Any mutation run here needs an asserted non-zero `git diff --numstat` before its result means anything.

docs/technical/architecture/subsystems/checker/audit-chain.md
paul-nicolas: Fixed in 04b2fbfb -- the defect is real and the doc is corrected, but three of the supporting statements needed adjusting and I did not adopt the proposed wording verbatim.

Confirmed: `applyProposal` returns before the audit path at machine.go:1183-1193; `orders` is bound at :1222, below that return, so `OrderCount` (:1357) is always >= 1; `buildAuditItems` (audit.go:82-105) is structurally `len(items) == len(serializedOrders)` with no filtering for any order kind. So the named case cannot occur.

Corrections:

1. **"The only AuditEntry writer is writeAuditEntry" is wrong.** `backup.ApplyExports` (internal/infra/backup/restore.go:64, sole production caller :181) writes `SubHistoryAudit` and `SubHistoryAuditItem` rows as raw KV -- whitelisted at restore.go:23-28 -- bypassing `appendAuditEntries` entirely, and full restore replays a Pebble checkpoint at file level. Neither invents a row, so the conclusion survives, but the doc must not claim the FSM is the only writer. Both writers are now named.

2. **There are four no-entry paths, not two:** idempotent replay (machine.go:1290), technical-only (:1183), a TU-phase plan-invariant rejection (:1147-1163), and a no-op skipped in the apply loop before `applyProposal` is called (:614). The `:184` line now lists all four. (The "audit chain is not extended here" sentence is at :1154-1156, not :1149-1150.)

3. **Did not write "dense for every audit_seq with no exception".** That unqualified absolute is false of a store restored from a manifest that lost its `auditItem` segments -- which is precisely the state the section should teach readers to recognise -- and stating it invites the reader to conclude they need not handle the case, the opposite of the intent and against invariant 7. Phrased instead as a property of the writers plus the detection that already exists: no writer produces an entry without its items, and the chain already reports the state as `HASH_MISMATCH`, because the per-item payloads are in the hash pre-image (machine.go:1385-1391) and `OrderCount` is bound into the hashed header (audit_envelope.go:107).

"Corruption signal" is kept, on that stronger footing: it names enforced behaviour rather than a new obligation on downstream code.

manager.go reworded as suggested, with the threat named -- a corrupt or partially-restored source store -- rather than described as redundant, since the branch runs under `currentAuditSeq > afterAuditSeq` and the structural guard is inherited from `exportEntries`. Also noted at :205 that `appliedProposal` is the one stream legitimately empty over a failure-only range.

internal/application/check/checker.go
```diff
@@ -322,13 +287,33 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
...
+		lastProgressSeq uint64
+		progressEmitted bool
+	)
+
 	for logIter.First(); logIter.Valid(); logIter.Next() {
 		if ctx.Err() != nil {
 			return ctx.Err()
 		}
 
 		// Extract sequence from key: [ZoneHistory(1)][SubHistoryLog(1)][sequence(8)]
```
NumaryBot, line 379: 🟠 **[major] Reject log rows at sequence zero**

When a forged Log row uses key sequence 0, `storedMaxLogSeq` remains indistinguishable from an empty store. The interior gap scan starts at 1, and `logBoundsVerifier.compare` sees stored and audited maxima both as 0, so a sequence-0 row with a matching value and inert payload can pass without any finding. Explicitly report sequence 0 as unaudited/invalid; audit-backed log sequences form an interval starting at 1.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1912#issuecomment-5571574532

internal/application/check/checker.go
paul-nicolas: Fixed in 856a3c667.

Confirmed. Sequence 0 is never allocated: `NewFSMState` seeds `NextSequenceID: 1` (internal/infra/state/fsmstate.go:74), recovery only ever raises it (fsmstate.go:159, guarded by `lastSeq > 0`), and `observeSuccess` pins the first audited range's minimum at 1 (log_bounds.go:160), so no success range can account for a row there.

Reproduced at 04b2fbfb2 with a forged row at key 0 whose `sequence` field matches its key and whose payload is nil: `Check()` reported zero findings, both alongside a healthy audited history (`rows=5`, `storedMax=4`) and as the store's only Log row (`rows=1`, `storedMax=0`). Your diagnosis of the coverage hole is exactly right — `compare` was only ever bounding the interval from above, and the interior gap scan is seeded at `expectedSeq` 1 so it never looks beneath it.

The log loop now reports a row at key sequence 0 as `LOG_UNAUDITED`, read off the key and emitted unconditionally: the audit fold has not run at that point, and a row at 0 is unaudited whether or not the chain later verifies. No new enum, so no field renumbering. The row is still replayed, on the same grounds as a divergent `sequence` field. No double-reporting — 0 is below `expectedMax`, so `compare`'s above-bound branch is untouched, and in the lone-row case both sides read 0 and it stays silent.

Covered by `TestCheck_LogBounds_ReservedSequenceZeroIsUnaudited` over both fixtures, verified to fail on the unfixed tree ("[]" should have 1 item(s), but has 0) and pass with the fix.

While in the same enum block I also corrected the `LOG_SEQUENCE_MISMATCH` proto comment, which still said the divergent row is skipped rather than replayed — 5215b4692 reversed that.

internal/application/check/checker.go
```diff
@@ -322,13 +287,55 @@ func (c *Checker) Check(ctx context.Context, callback func(*servicepb.CheckStore
...
```
NumaryBot, line 379: 🔴 **[blocker] Include the maximum uint64 log key in verification**

When a forged Log uses sequence `math.MaxUint64`, the iterator's exclusive upper bound is exactly that key, so the row is never visited and `storedMaxLogSeq` remains unchanged. `ReadLastSequence` uses the same bound, allowing this unaudited row to pass without any finding, contrary to the checker requirement in [AGENTS.md:44](../../../AGENTS.md#L44). Use a prefix-successor upper bound so every uint64 sequence is included.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1912#issuecomment-5571574532

internal/application/check/log_bounds.go
```diff
@@ -0,0 +1,231 @@
...
+			fmt.Sprintf("log sequences %d..%d are missing: the audit chain accounts for logs up to %d "+
+				"but the highest stored log is %d (%d logs)",
+				first, v.expectedMax, v.expectedMax, storedMax, v.expectedMax-storedMax),
+			first, "", "", ""))
+	case storedMax > v.expectedMax:
+		first := v.expectedMax + 1
+
+		callback(errorEvent(
+			servicepb.CheckStoreErrorType_CHECK_STORE_ERROR_TYPE_LOG_UNAUDITED,
+			fmt.Sprintf("log sequences %d..%d have no audited origin: the audit chain accounts for logs "+
```
Azorlogh, line 372: The count here is `storedMax - expectedMax` (the range *width*) and the message reads `sequences expectedMax+1..storedMax have no audited origin` — exact only for a contiguous injected tail (the one shape tested). For a sparse injection (a forged row above the bound with a gap beneath it) it over-counts and labels the empty positions unaudited, while the interior gap scan separately reports those same positions as *missing*. Counting actual rows in the range would avoid the over-count. (The `SEQUENCE_GAP`/tail path is fine — the chain proves that range was fully populated.)

internal/application/check/log_bounds.go
```diff
@@ -0,0 +1,231 @@
...
+
+	v.incompleteReason = reason
+}
+
+// observeSuccess folds one chain-verified SUCCESS audit entry into the expected
+// bound. Callers MUST only pass entries whose hash has just been verified: an
+// unverified range carries no statement about which logs the FSM produced.
+//
+// Entries with min == max == 0 carry no freshly created log (an all-idempotent
+// replay, or a proposal whose orders produced no log) and are ignored — they
```
Azorlogh: Nit: an idempotency-key replay returns before writing any audit entry (`machine.go:1290`) — as commit 6's own `audit-chain.md` correction states — so a `min == max == 0` success entry can't come from "an all-idempotent replay". Only the second cause ("a proposal whose orders produced no log") actually reaches here.

gfyrag (changes requested):
I think the implementation direction should be reconsidered before merge.

The checker already reads every hash-bound AuditItem while verifying the chain, but the new bounds pass ignores the exact per-order LogSequence evidence and keeps only AuditSuccess min/max, then compares that interval with one stored maximum. On the current v3 writers, positive AuditItem log sequences are the exact fresh-log positions; failure and no-log outcomes carry 0. This PR also documents that no supported store needs the old pre-release reference-item tolerance.

That lossy max-only model creates avoidable ambiguity:

- A sparse injection above the audited head is described as a contiguous unaudited range and its interval width is reported as a row count. I reproduced an audited head of 4 plus one row at 7 being reported as "5..7 (3 logs)", while 5 and 6 are separately reported as missing.
- The log iterator excludes the exact math.MaxUint64 key. I reproduced a store containing only that forged row: Check returned no LOG_UNAUDITED finding.
- Fixing only the upper bound is unsafe because the existing per-sequence gap loop would try to emit almost 2^64 findings before reaching the row.
- The checker still keeps two competing heads: the key-derived storedMaxLogSeq for this pass, but the value-derived ReadLastSequence result for progress totals and reverse-map alignment, even though this PR establishes that the value field is untrusted.

A simpler and stronger shape would be:

1. During the already-required audit walk, derive the expected positive log sequence stream from the verified AuditItems and validate that AuditSuccess min/max agrees with that stream.
2. Scan the full log prefix using the prefix successor as the exclusive upper bound.
3. Compare actual keys against the audited sequence stream, aggregating missing ranges and counting only rows actually observed above the audited range.
4. Use the resulting key-derived head everywhere, including progress and reverse-map alignment.

This removes the implicit contiguity guess, covers sparse and maximum-key corruption without an event explosion, validates the redundant range fields against the item evidence, and leaves one source of truth for the observed store head.

REQUEST CHANGES: the exact MaxUint64 bypass is a checker blind spot, and a direct bound-only fix would make the same input effectively non-terminating. Please address the range traversal and oracle choice together, with regression tests for a sparse injected key and an exact MaxUint64 key that completes with bounded findings.

internal/application/check/log_bounds.go
```diff
@@ -0,0 +1,231 @@
...
+// neither advance nor interrupt the interval. An entry with min > 0 and max ==
+// 0 is NOT that shape: it is an inverted range, and it is reported as one.
+//
+// The three range shapes are therefore tested in a fixed order — inversion,
+// then log-less, then discontinuity — because the log-less short-circuit would
+// otherwise absorb every inverted range whose max happens to be 0.
+func (v *logBoundsVerifier) observeSuccess(entry *auditpb.AuditEntry) {
```
gfyrag: [P2][blocking] Validate the success range against the audited items

observeSuccess only sees the AuditEntry header, so it cannot prove that any audited order actually produced the positions declared by min/max. I reproduced a chain-valid Success with range 1..4, OrderCount=0, no AuditItem rows, and Log rows 1..4: Check returns zero findings. The PR test helper persistSuccessAuditEntries builds this exact impossible writer shape, which is why the bounds tests do not expose the gap.

The same issue covers duplicate or missing positive AuditItem.LogSequence values inside an otherwise contiguous header range. The hash proves that the header and items were persisted as supplied; it does not prove their semantic relationship. Trusting min/max alone therefore lets a writer bug create a hash-valid history where logs have no audited producing orders, despite invariant #7 requiring impossible states to fail loudly.

verifyAuditHashChain already has the verified items in hand. Pass them to this fold, derive the positive fresh-log sequence range from those items, and require it to agree with AuditSuccess min/max before using the bound as an oracle.

internal/application/check/checker.go
gfyrag: There is a shared root cause underneath this local iterator. dal.MaxUint64Bytes is documented as an upper-bound sentinel, but Pebble bounds are exclusive; the same value is used by dal.ReadLastEntry and query.ReadLogsSince / ReadLogsSinceRaw. Fixing only Checker.Check would leave recovery, index building, event emission, and backup replay with a different view of the log range.

I would fix the range contract once in DAL with the successor of the zone/sub prefix, migrate all sequence-prefix scans to it, and add the MaxUint64 regression at that layer. This PR should also decide which copy of the sequence is authoritative: it correctly treats the key as authoritative for bounds, but ReadLastSequence still exposes the untrusted protobuf field to progress, reverse-map alignment, recovery, and other callers. Keeping both notions of the head is the architectural inconsistency that made the original early-return bypass possible.

paul-nicolas: Fixed in d8ad5314c (and e20d7faba), and taken DAL-wide as @gfyrag asked rather than patched locally.

Confirmed the hole first: a raw `Log` row at `math.MaxUint64` produced **zero findings of any type**, both alongside a healthy audited 1..4 history and as the only row on a failure-only history, while a raw point lookup proved the value was present. `checker.go:280` was that key.

One correction on the mechanism: `ReadLastSequence` was not what let the row pass. The verdict comes from the key-derived head, not from that function — the bound claim about it is right, the causal one is not. It is a real problem for a different reason (it fed the progress total and the reverse-map oracle off a non-hash-bound field), and that is fixed too, in the same push.

Three things had to move before the bound could be widened at all:

- **The interior gap scan enumerated.** Widening alone hands `seq = MaxUint64` to `for expectedSeq < seq`, which emitted 780M events in 90s at 8.68M/s before the test timed out — ~67,000 years extrapolated. Runs are now aggregated into one event each. Worth flagging separately: **this is reachable on `release/v3.0` today**, with no bound change. A forged in-range row at 1<<24 costs 1.8s / 16.8M events; at 1<<40 it never finishes. `MaxUint64` was in fact the one forged position that did *not* already deny the checker. `TestCheck_LogBounds_WideInteriorGapStaysBounded` pins it — it hangs under mutation rather than merely failing.
- **A row above the bound closes no hole.** The run beneath it is the audited tail, which `logBoundsVerifier` owns; letting the scan report it too described the same tail twice and called never-allocated positions lost rows.
- **`compare` switched on one head.** With a genuine truncation (audit accounts for 1..10, only 1..4 survive) plus one forged `MaxUint64` row, raising the head flipped the branch and the `SEQUENCE_GAP` for `5..10 are missing` **disappeared**, replaced by `11..18446744073709551615 (18446744073709551605 logs)`. One forged row would have hidden a genuine restore failure. The tail is now measured from the highest row still inside the bound, and both findings are emitted when both hold — `TestCheck_LogBounds_InjectedMaxUint64DoesNotEraseLostTail`.

On the DAL-wide part, @gfyrag: agreed, and done. `dal.PrefixUpperBound` / `dal.ZonePrefixUpperBound` now own the rule; `ReadLastEntry`, `ScanZone`, `ReadLogsSince`/`Raw`, `ReadAuditEntries`, `readAuditPageFromZone`, `ReadAppliedProposals` and the two ledger-scoped fallbacks in `compile.go` all use them. `readstore.prefixUpperBound` and `state.buildLedgerScopedPrefixSuccessor` were separate definitions of the same rule and now delegate. `dal.MaxUint64Bytes` is deleted, so the pattern cannot be reintroduced by reaching for a named constant.

Widening it turned three latent `+1` wraps into live ones, all handled explicitly: `ReadAuditItems` built `[seq, seq+1)`, which wraps to a bound *below* the lower one and made the entry read as carrying no items; `backup.exportEntries` had the same wrap and would have exported an empty segment (invariant #11); and `LoadFSMStateFromStore` now **refuses to boot** on a `MaxUint64` head rather than wrapping `lastSeq + 1` to 0 and allocating on top of the planted row (invariant #7). That last one is why I had initially proposed deferring — it is a recovery-path change — but it is a small, well-pinned one, so it lands here.

Contract written up in `docs/technical/architecture/subsystems/storage/range-bounds.md`.

internal/application/check/log_bounds.go
paul-nicolas: Confirmed and fixed in e20d7faba. Reproduced exactly as described — audited 1..4 plus one forged row at 7:

```
SEQUENCE_GAP    seq=5  "log sequence 5 is missing"
SEQUENCE_GAP    seq=6  "log sequence 6 is missing"
LOG_UNAUDITED   seq=5  "log sequences 5..7 have no audited origin: ... (3 logs)"
```

It was also a coverage gap: the only test pinning that message asserts the count on *contiguous* injections, where width and row count coincide, so the whole pre-existing suite stayed green under the bug. Proven by mutation — reverting the count to the width now fails the new test and two others.

Counting rows is nearly free. `verifyAuditHashChain` runs at `checker.go:171`, before the log loop opens, so `expectedMax` is already final when the loop reads its first row: one `uint64` compare per row, no second iterator. `(3 logs)` becomes `(1 logs)`.

Your carve-out is right and I kept it: the `SEQUENCE_GAP` side still reports the width, because there the chain proves every position in the range was allocated. Only the injected side lacks a density proof. That asymmetry is now stated in `compare`s doc comment.

One place I went further than you suggested. You noted the interior scan reports 5 and 6 as *missing* while the bound pass encloses them in an unaudited range — two findings contradicting each other. Making the count truthful fixes the arithmetic but leaves the contradiction: nothing allocated 5 and 6, so calling them missing rows is not just imprecise, it is unsupported. The audit chain accounts for logs up to 4. So the interior scan is now capped at the audited bound when the chain derived one, and the whole contradiction goes away — that fixture reports exactly one finding, `(1 logs)`, and nothing claims 5 and 6 were lost. `TestCheck_LogBounds_SparseInjectedTailCountsRowsNotWidth` pins both halves.

Two residuals left deliberately: the leading clause still names the range searched (`5..7`) rather than the offending rows — with a truthful count beside it the message is coherent, and rewording breaks a pinned prefix assertion for no correctness gain; and `(%d logs)` renders `(1 logs)`, which is pre-existing and already pinned on the deleted-tail path. Say the word if you want either.

internal/application/check/log_bounds.go
paul-nicolas: Half right, and the comment is corrected in e20d7faba either way.

Your premise is exactly right: `machine.go:1290` (`if replayed { ... return }`) is proposal-level idempotency-key dedup, it returns before the pipeline, and it writes no `AuditEntry` of any shape. `audit-chain.md` says so, and nothing from that path reaches `observeSuccess`.

But there is a separate **per-order** idempotent replay, and it does reach here. `processor.go:337` (`if payload == nil { continue }`) skips the log slot without consuming a sequence id; `processor_mirror.go:87` (`case v2LogID <= last: return nil, nil`) is that path today. The proposal still succeeds, still runs the pipeline, and `machine.go:1543` builds the success entry unconditionally from the accumulated range. I checked it rather than argued it: a proposal of two mirror-ingest orders both at or below `LastMirrorV2LogId` returns `CreatedLogs=0 MinLogSequence=0 MaxLogSequence=0` with a nil error. So the shape is reachable, and the comments first cause stands.

The nit lands anyway. Unqualified "an all-idempotent replay" reads naturally as the idempotency-*key* replay — which is how you read it, holding the correct facts — while this files own header at `:35` and `:41-43` draws the distinction precisely. That asymmetry inside one file is what made the misreading reasonable, so the comment now names the mechanism and explicitly excludes the proposal-level replay.

Two things I found while checking, both folded into the same commit:

- There is a **second** live cause the comment and I both missed: `processor_cluster_policy.go:46-48`, a `SetClusterPolicy` re-proposal at the applied revision. Also named now.
- `audit-chain.md:195` still described the *dead* cause ("a success in which every order is an in-batch idempotent reference contributes `[0, 0]`") and omitted both live ones — `processor.go:181-184` records that mechanism no longer exists. Corrected.

Adjacent catch, also in that commit: the sequence-0 comment a few lines away in `checker.go` justified reporting unconditionally with "the audit fold has not run at this point", which is false — the fold runs at `checker.go:171`. The conclusion was right, the reason was not.

internal/infra/state/fsmstate.go
```diff
@@ -155,6 +156,17 @@ func LoadFSMStateFromStore(reader dal.RecoveryReader, handle *dal.ReadHandle, cl
 		return nil, fmt.Errorf("reading last sequence: %w", err)
 	}
 
+	// A head at MaxUint64 cannot be advanced past: lastSeq+1 wraps to 0, and
+	// the FSM would start allocating at a sequence the checker reports as
+	// impossible, on top of whatever row is already there. The FSM cannot reach
+	// that head on its own — NextSequenceID is seeded at 1 and only ever
+	// incremented — so a store holding one was corrupted or restored from a
+	// tampered stream. Refuse to boot rather than wrap (invariant #7).
+	if lastSeq == math.MaxUint64 {
```
NumaryBot, line 174: 🔴 **[blocker] Prevent allocation when the recovered head is MaxUint64-1**

When a restored or corrupted store ends at `math.MaxUint64-1`, this check permits boot and initializes the next counter to `math.MaxUint64`. The next log allocation or audit append then returns MaxUint64 and increments the counter to 0, so subsequent writes reuse reserved/old sequence space instead of failing loudly. Guard the increment path or reject this penultimate head as required by [AGENTS.md:43](../../../AGENTS.md#L43).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1912#issuecomment-5571574532

internal/application/check/log_bounds.go
paul-nicolas: Valid, and fixed in e20d7faba. I reproduced your shape independently before touching anything: Log rows 1..4 under a chain-valid Success declaring 1..4 with `OrderCount=0` and no `AuditItem` rows, `Check` returns **zero** error events. And you are right that the items are sitting there unused — `checker.go:2142` reads them, `:2154-2158` hashes them into the pre-image, and `observeSuccess` is called one statement above `collectExpectedSkippable(items, ...)`.

You are also right about `persistSuccessAuditEntries`. It hashed the header alone and passed `nil` items, so every fixture built through it expressed a shape no writer produces. It now writes the per-order items a real apply would have written, and `persistItemlessSuccessAuditEntry` builds the itemless shape deliberately, for the test that pins this finding.

I checked the producer relationship against real pipeline output rather than assuming it, because the check would false-positive if it were wrong:

```
entry=1  SUCCESS range=1..1   orderCount=1 items=1 itemSeqs=[1]
entry=4  SUCCESS range=4..5   orderCount=2 items=2 itemSeqs=[4 5]   <- multi-order
entry=5  FAILURE range=0..0   orderCount=2 items=2 itemSeqs=[0 0]
entry=8  SUCCESS range=8..8   orderCount=1 items=1 itemSeqs=[8]     <- skip-log path
```

The rule as shipped: for a chain-verified SUCCESS with `max > 0`, the **distinct** item sequences falling inside `[min, max]` must be exactly `{min, ..., max}`. Stated over distinct in-range values on purpose, so it tolerates the three legitimate shapes that are not one-item-per-position — an item at `LogSequence` 0 (the no-log outcome inside an otherwise log-producing proposal, `processor.go:327-332`, which is also why I did **not** compare `OrderCount` to the positive-item count as you suggested; that would fire on every such proposal), an item below `min` (the legacy pre-`f9ee1e829` back-reference `checker.go:2564-2591` still filters), and two items sharing one in-range sequence (`foldedLogSeqs`, `checker.go:2526-2537`). The range is never walked — a forged entry can declare `1..MaxUint64`, so enumerating it to find the first hole would hang the fold; the walk is over the items.

Two pushbacks, neither affecting the fix.

**I would not call it blocking.** The audit hash key is `blake3.Sum256("audit-hash:blake3:v1:" + clusterID)` (`hash_blake3.go:12-22`) — no secret. Anyone able to forge a chain-valid itemless Success can equally forge four matching `AuditItem`s, so against the corruption/restore model this pass serves the cross-check raises the forgery cost rather than closing a hole this PR opens. Checked concretely: deleting the whole `SubHistoryAuditItem` key space on a real history already reports `HASH_MISMATCH` plus all three `*_VERIFICATION_INCOMPLETE` events, and the itemless header alone still catches tail truncation correctly. Worth having as hardening — which is why it is in — but it is not the severity that gates the merge.

**"which is why the bounds tests do not expose the gap" is not what happened.** The five load-bearing bounds tests run on real pipeline output via `newTestEngine`/`processAndCommit` (`:260`, `:378`, `:443`, `:487`, `:544`). `persistSuccessAuditEntries` was only used for shapes the pipeline cannot produce — reserved seq 0, log-less success, discontinuous and inverted ranges, large truncation — which is its documented purpose. The helper was wrong; it was not what hid this.

One thing worth resolving separately, since you raised the oracle question: `log_bounds.go:64-75` argues the legacy reference-item shape cannot exist on any v3 store, while `checker.go:2526-2537` and `:2564-2591` still code defensively for it. Per AGENTS.md ("v3 is unreleased, no compatibility burden") one of the two should go. I kept the tolerances so this change stays behaviour-preserving on that axis, but happy to delete them in a follow-up if you agree.

paul-nicolas:
@shipfox-ai — replying to both review comments together.

**Spec 1 (LOG_SEQUENCE_MISMATCH docstring) — stale.** Fixed earlier in this PR; you reviewed an older head. At `856a3c667` and since, `misc/proto/bucket.proto:871-875` reads "the divergent row is reported and still replayed under that key sequence", and the regenerated comment matches in `internal/proto/servicepb/bucket.pb.go` — both files you named. `grep -rn "skips replaying"` over `misc/proto`, `internal/proto`, `docs` and `internal/application` returns nothing.

**Spec 2 (row at reserved key sequence 0) — stale, and the symmetric case was real.** `checker.go` reports a row at key sequence 0 as `LOG_UNAUDITED`, off the key, pinned by `TestCheck_LogBounds_ReservedSequenceZeroIsUnaudited`. Your reasoning independently matched NumaryBots; it just landed before this review posted. One footnote: the justification *comment* that commit added was itself wrong ("the audit fold has not run at this point" — it runs at `checker.go:171`), and it is corrected in `e20d7faba`.

The hole at the **other** end of the key space was not stale. Pebbles iterator upper bound is exclusive, so the eight-byte `0xFF` run this loop used was byte-identical to the key at `math.MaxUint64` and excluded exactly that row — the mirror of the sequence 0 case you describe. Closed in `e20d7faba`, and taken DAL-wide in `d8ad5314c`, where `dal.MaxUint64Bytes` is deleted and `dal.PrefixUpperBound` owns the rule.

**Standards 1 (the coverage-verifier clump) — agreed, done in `e20d7faba`.** `chainVerifierFolds` now lives in `checker.go` next to the verifiers it groups; `verifyAuditHashChain` takes one `folds` parameter instead of three, and each of the three early exits calls a single `folds.markLiveTruncated()`. The rationale comment is stated once and reworded — it still opened "The signing fold below stops here..." while governing all three. The test-only copy in `checker_tamper_test.go` is deleted in favour of the production type, and no assertion changed: the per-exit wiring stays pinned by `TestVerifyAuditHashChain_MarksSigningFoldTruncatedOnEveryBreak`. For the record on scope, you were right that the clump pre-dates this diff — base had two verifiers and six calls — but this PR is what took it to three and nine.

The two spec-text drifts you noted (test count, commit count) are in the ticket prose rather than the code; left as-is, and thanks for separating them out.

internal/infra/state/fsmstate.go
paul-nicolas: Valid, and fixed in 5b38f49c7. You are right that the boot guard I added was necessary but not sufficient.

Confirmed the mechanism before fixing it: both allocators are plain post-increments on replicated state — `WriteSet.IncrementNextSequenceID` (`write_set.go:1376-1381`) and `FSMState.AppendAuditEntry` (`fsmstate.go:122-128`) each return the pre-bump value and then increment. At a head of `MaxUint64-1` the boot check passes, `NextSequenceID` becomes `MaxUint64`, the next allocation returns it and leaves the counter at 0, and every allocation after that addresses a row that already exists.

I took the first of your two options rather than the second. Rejecting the penultimate head is a one-line change, but the threshold is arbitrary — the same question applies to `MaxUint64-2`, and a boot-only check covers only heads that arrive through recovery. The guard that holds for every head is at the allocator, so both now return an error and fail with `domain.ErrSequenceSpaceExhausted` instead of wrapping. `Scope.IncrementNextSequenceID` carries the error through the order overlay and the skip-safe scope; `ProcessOrders` surfaces it, so the proposal fails, a Failure audit entry is written, and the chain continues. Mocks regenerated for the signature change.

Determinism, since this is the apply path (invariant #2): both counters are replicated state, so every node reaches the refusal on the same entry and fails the same proposal. Nothing node-local enters the decision, and the node is not stopped.

One implementation note worth recording. The order overlay allocates ahead of the parent and drains the delta in `Commit`, so the check has to be against the position it is about to hand out, not against the parent counter, which has not moved yet. The drain then replays allocations the overlay already validated against the same base, so it cannot legitimately fail — it reports rather than ignores an error there, which would mean the two views disagreed.

`TestAllocatorsRefuseToWrap` pins both counters at `MaxUint64-1`: the last representable sequence is still allocatable, the one after it fails, and a refused allocation moves neither the counter nor the audit chain head. The boot guard and its test stay, since refusing a store that can never allocate again is still the right answer at recovery.

Contract updated in `docs/technical/architecture/subsystems/storage/range-bounds.md`, including why the boot guard alone is not enough.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1912#issuecomment-5571574532

gfyrag:
I think we should change the direction of this PR rather than keep extending the current log-driven checker.

The audit chain must be the checker's sole source of business truth. Logs are now a derived projection, like volumes, metadata, balances, and the other persisted views; they must be verified against the audit, not used as the source from which the checker reconstructs expected state.

Concretely, the checker should:

1. verify the audit chain and decode the audited AuditEntry/AuditItem payloads;
2. replay the audited orders to derive the complete expected state, including the expected log stream;
3. compare the persisted logs and every other projection against that audit-derived state.

With that model, missing, extra, reordered, or malformed log rows are ordinary projection mismatches. We also avoid creating two competing heads (audit head vs. log head) and avoid trusting audit range metadata without proving it against the audited items.

So my recommendation is to rework #1912 around an audit-driven replay, instead of adding bounds around the existing log-driven replay.

gfyrag:
Rechecked at `5929235` against the current `release/v3.0` head `66a9b23`.

The previously reported local correctness issues are resolved at this head: the audit success range is checked against its verified items, sparse/zero/MaxUint64 log keys are covered with bounded findings, and the observed log head is key-derived. I am not reopening those findings.

The architectural hold remains, independently of those fixes. `verifyAuditHashChain` runs first, but the main expectation-building path below it still consumes the stored, non-hash-bound `LogPayload` (`ReplayLedgerLog`, ledger/schema/index/boundary/numscript/checkpoint folds). `logBoundsVerifier` proves which log *positions* the audit accounts for; it does not prove that the payload at a position is the payload implied by `AuditItem.SerializedOrder`. For example, if an audited metadata order says value X while both the derived Log payload and stored metadata projection are changed to Y, the checker derives Y from the log and compares the projection to Y. The audit remains valid but never acts as the oracle for that projection. That is exactly the circularity behind the requested pivot: logs must be compared as a projection, not consumed as business truth.

The rebase conflict is also now meaningful rather than incidental. There are 13 content-conflict files, almost all on the sequence-allocation paths that `6080ed9b1f` (EN-1860) has since replaced with the canonical `domain.CheckedNextSequence` / public `ErrSequenceExhausted` handling across apply, mirror, recovery, checker and rebuild. The rebase should take the base implementation for those counter/error paths and drop the branch-local variant. The DAL prefix-successor migration is still distinct and useful: current base fixed `ReadLastEntry`, but still has sequence scans using `dal.MaxUint64Bytes`, so that part should be reconciled rather than discarded.

Suggested rework shape:

1. Rebase and remove the now-superseded counter-wrap implementation in favor of EN-1860.
2. Keep the independent zero-log audit verification, audit-cardinality documentation corrections, and any still-needed DAL full-prefix bound fixes.
3. Build the expected log/effect stream from the verified audit entries/items, then compare stored Logs and every other projection against that audit-derived stream. The stored Log iterator should only collect actual rows for comparison; it should not drive expected business state.

So the current `CHANGES_REQUESTED` remains an architectural blocker, not because the fixed local findings regressed, but because the branch still reinforces a log-driven replay model that contradicts the repository's “audit is business truth” contract.

PR #1916, opened 2026-09-07
fix(EN-1959): fence leadership worker generations

gfyrag:
## What changed

Leadership transitions for the events and mirror managers are now recorded synchronously and reconciled by their lifecycle-owned loops. Monotonic generations fence stale scans, startups, teardowns, and callbacks after Fx shutdown. Deterministic regression tests cover superseded leadership loss and post-Stop callbacks.

## Why

Fixes [EN-1959](https://formance-team.atlassian.net/browse/EN-1959).

Out-of-order asynchronous leadership callbacks could stop current-generation workers or recreate leader-only workers after leadership loss or Fx teardown.

Need: leader-only event and mirror workers must match the current Raft leadership generation.
Current limitation: callback serialization did not preserve transition order or establish generation ownership.
Requirement / constraint: work from generation N must never mutate generation N+1 lifecycle state or survive shutdown.
Evidence: EN-1959, deterministic stale-callback regression tests, and the lifecycle contracts in docs/technical/architecture/subsystems/events-mirror/.
Durable repository evidence: docs/technical/architecture/subsystems/events-mirror/events.md and mirror.md.

## Technical decision

Decision: record leadership transitions inline, reconcile slow work in the existing manager loops, and fence each reconciliation with a monotonically increasing generation. Stop closes the leadership gate before draining the loop.
Why now / why proportionate: this directly fixes the confirmed stale-callback execution path without redesigning the worker lifecycle.
Alternatives considered: mutex-only callback serialization was rejected because it cannot preserve transition order; callback-owned reconciliation goroutines were rejected because they lack lifecycle generation ownership.

## Risk

**MEDIUM** — concurrency and leader-only worker lifecycle change, bounded to events, mirror, and bootstrap wiring; no persisted format, API, or FSM semantics change.

## Validation

- [x] AI_REVIEW_BASE_SHA=ce72d6df10142a48fa3193ed369831d5f0252854 bash scripts/agent-check-pr
- [x] Race tests: ./internal/application/events, ./internal/application/mirror, ./internal/bootstrap
- [x] Deterministic stale-callback regression and regression sensitivity evidence
- [ ] Full suite / broader validation: GitHub CI

Leader-only events and mirror workers now use explicit leadership-generation ownership. No persisted state, wire format, API, or deterministic FSM behavior changes.

## Review focus

Please focus on generation checks around slow reconciliation, worker retention, superseded leadership loss, and the Stop gate/drain ordering.

## Known concerns

None.

[EN-1959]: https://formance-team.atlassian.net/browse/EN-1959?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/application/events/manager.go
+++ b/internal/application/events/manager.go
@@ -37,10 +37,15 @@ type Manager struct {
 	logger         logging.Logger
 	notifications  *signal.Notifications
 
-	mu       sync.Mutex
-	isLeader bool
-	emitters map[string]*managedSink
-	retries  map[string]struct{}
+	mu                  sync.Mutex
+	emitters            map[string]*managedSink
+	retries             map[string]struct{}
+	resourcesGeneration uint64
+
+	leadershipMu         sync.Mutex
+	leadershipGeneration uint64
+	isLeader             bool
+	stopped              bool
 
 	w worker.Worker
 }
@@ -68,6 +73,18 @@ func (m *Manager) Start() {
 
 // Stop gracefully stops the Manager and tears down any active emitters/sinks.
 func (m *Manager) Stop() {
+	m.leadershipMu.Lock()
+	if m.stopped {
+		m.leadershipMu.Unlock()
+
+		return
+	}
+
+	m.leadershipGeneration++
+	m.isLeader = false
+	m.stopped = true
+	m.leadershipMu.Unlock()
+
 	m.w.Stop()
 
 	m.mu.Lock()
@@ -76,13 +93,24 @@ func (m *Manager) Stop() {
 	m.teardown()
 }
 
-// OnLeadershipChange is called when the node's leadership status changes.
+// OnLeadershipChange records the latest leadership generation and wakes the
+// lifecycle-owned reconciliation loop. It deliberately does not reconcile on
+// the caller: bootstrap invokes it synchronously from the Raft observer so
+// transitions are recorded in order without blocking the Raft processing loop
+// on Pebble reads or worker startup.
 func (m *Manager) OnLeadershipChange(isLeader bool) {
-	m.mu.Lock()
-	defer m.mu.Unlock()
+	m.leadershipMu.Lock()
+	if m.stopped {
+		m.leadershipMu.Unlock()
 
+		return
+	}
+
+	m.leadershipGeneration++
 	m.isLeader = isLeader
-	m.reconcile()
+	m.leadershipMu.Unlock()
+
+	m.notifications.NotifyConfigChanged()
 }
 
 func (m *Manager) loop(stop <-chan struct{}) {
@@ -99,11 +127,30 @@ func (m *Manager) loop(stop <-chan struct{}) {
 			m.mu.Lock()
 			defer m.mu.Unlock()
 
-			if m.isLeader {
-				m.reconcile()
-			}
+			m.reconcile()
 		},
 	)
+}
+
+func (m *Manager) leadershipSnapshot() (uint64, bool, bool) {
+	m.leadershipMu.Lock()
+	defer m.leadershipMu.Unlock()
+
+	return m.leadershipGeneration, m.isLeader, m.stopped
+}
+
+func (m *Manager) isCurrentLeader(generation uint64) bool {
+	m.leadershipMu.Lock()
+	defer m.leadershipMu.Unlock()
+
+	return !m.stopped && m.isLeader && m.leadershipGeneration == generation
+}
+
+func (m *Manager) isCurrentGeneration(generation uint64) bool {
+	m.leadershipMu.Lock()
+	defer m.leadershipMu.Unlock()
+
+	return m.leadershipGeneration == generation
 }
 
 // reconcile reads the current per-sink configurations from the store and
@@ -111,7 +158,24 @@ func (m *Manager) loop(stop <-chan struct{}) {
 // were added, removed, or changed are affected — unchanged sinks keep running.
 // Must be called under lock.
 func (m *Manager) reconcile() {
-	if !m.isLeader {
+	generation, isLeader, stopped := m.leadershipSnapshot()
+	m.reconcileGeneration(generation, isLeader, stopped)
+}
+
+// reconcileGeneration applies one captured leadership generation. A transition
+// can arrive after the loop wakes but before it acquires m.mu, so reject a
+// superseded generation before either teardown or startup mutates ownership.
+// Must be called under lock.
+func (m *Manager) reconcileGeneration(generation uint64, isLeader, stopped bool) {
+	if !m.isCurrentGeneration(generation) {
+		return
+	}
+	if m.resourcesGeneration != generation {
+		m.teardown()
+		m.resourcesGeneration = generation
+	}
+
+	if stopped || !isLeader {
 		m.teardown()
 
 		return
@@ -133,6 +197,10 @@ func (m *Manager) reconcile() {
 		return
 	}
 
+	if !m.isCurrentLeader(generation) {
+		return
+	}
+
 	// Build desired state as a map keyed by sink name
 	desired := make(map[string]*commonpb.SinkConfig, len(sinkCfgs))
 	for _, sc := range sinkCfgs {
@@ -147,6 +215,10 @@ func (m *Manager) reconcile() {
 
 	// Remove sinks that no longer exist or whose config changed
 	for name, ms := range m.emitters {
+		if !m.isCurrentLeader(generation) {
+			return
+		}
+
 		sc, stillDesired := desired[name]
 		if !stillDesired || !sc.EqualVT(ms.config) {
 			m.stopSink(name, ms)
@@ -156,11 +228,21 @@ func (m *Manager) reconcile() {
 
 	// Start sinks that are new or were just removed due to config change
 	for name, sc := range desired {
+		if !m.isCurrentLeader(generation) {
+			return
+		}
+
 		if _, exists := m.emitters[name]; exists {
 			continue // already running with same config
 		}
 
 		if ms := m.startSink(sc); ms != nil {
+			if !m.isCurrentLeader(generation) {
+				m.stopSink(name, ms)
+
+				return
+			}
+
 			m.emitters[name] = ms
 			delete(m.retries, name)
 		}
@@ -245,10 +327,10 @@ func (m *Manager) scheduleStartupRetry(name string) {
 		}
 
 		delete(m.retries, name)
-		isLeader := m.isLeader
+		_, isLeader, stopped := m.leadershipSnapshot()
 		m.mu.Unlock()
 
-		if isLeader {
+		if isLeader && !stopped {
 			m.notifications.NotifyConfigChanged()
 		}
 	}()
--- a/internal/application/mirror/manager.go
+++ b/internal/application/mirror/manager.go
@@ -40,9 +40,14 @@ type Manager struct {
 	meterProvider metric.MeterProvider
 	maxBatchSize  int
 
-	mu       sync.Mutex
-	isLeader bool
-	workers  map[string]*Worker
+	mu                  sync.Mutex
+	workers             map[string]*Worker
+	resourcesGeneration uint64
+
+	leadershipMu         sync.Mutex
+	leadershipGeneration uint64
+	isLeader             bool
+	stopped              bool
 
 	w worker.Worker
 }
@@ -70,6 +75,18 @@ func (m *Manager) Start() {
 
 // Stop gracefully stops the Manager and tears down any active workers.
 func (m *Manager) Stop() {
+	m.leadershipMu.Lock()
+	if m.stopped {
+		m.leadershipMu.Unlock()
+
+		return
+	}
+
+	m.leadershipGeneration++
+	m.isLeader = false
+	m.stopped = true
+	m.leadershipMu.Unlock()
+
 	m.w.Stop()
 
 	m.mu.Lock()
@@ -78,13 +95,24 @@ func (m *Manager) Stop() {
 	m.teardown()
 }
 
-// OnLeadershipChange is called when the node's leadership status changes.
+// OnLeadershipChange records the latest leadership generation and wakes the
+// lifecycle-owned reconciliation loop. It deliberately does not reconcile on
+// the caller: bootstrap invokes it synchronously from the Raft observer so
+// transitions are recorded in order without blocking the Raft processing loop
+// on Pebble reads or worker startup.
 func (m *Manager) OnLeadershipChange(isLeader bool) {
-	m.mu.Lock()
-	defer m.mu.Unlock()
+	m.leadershipMu.Lock()
+	if m.stopped {
+		m.leadershipMu.Unlock()
 
+		return
+	}
+
+	m.leadershipGeneration++
 	m.isLeader = isLeader
-	m.reconcile()
+	m.leadershipMu.Unlock()
+
+	m.notifications.NotifyConfigChanged()
 }
 
 func (m *Manager) loop(stop <-chan struct{}) {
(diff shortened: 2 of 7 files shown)


internal/application/events/manager.go
```diff
@@ -156,11 +223,21 @@ func (m *Manager) reconcile() {
 
 	// Start sinks that are new or were just removed due to config change
 	for name, sc := range desired {
+		if !m.isCurrentLeader(generation) {
+			return
+		}
+
 		if _, exists := m.emitters[name]; exists {
```
NumaryBot, line 235: 🔴 **[blocker] Fence retained emitters across a coalesced leadership flap**

If leadership changes `true → false → true` before the notification loop processes the loss, the buffered signal coalesces the callbacks and this branch treats the generation-N emitter as already valid for generation N+2. It therefore runs continuously through the leadership loss, contrary to the documented generation ownership contract ([events.md:239-244](docs/technical/architecture/subsystems/events-mirror/events.md#L239-L244)); existing emitters need generation ownership or an equivalent forced teardown on a skipped loss.

internal/application/mirror/manager.go
```diff
@@ -149,6 +216,10 @@ func (m *Manager) reconcile() {
 
 	// Start workers for new mirror ledgers
 	for name, info := range desired {
+		if !m.isCurrentLeader(generation) {
+			return
+		}
+
 		if _, exists := m.workers[name]; exists {
```
NumaryBot, line 228: 🔴 **[blocker] Fence retained mirror workers across a leadership flap**

If a loss and subsequent gain occur before the loop handles the buffered notification, reconciliation observes only the latest leader generation and this branch retains the worker created by the previous generation. That worker survives the entire loss and becomes implicitly owned by the new generation, violating the stated lifecycle contract ([mirror.md:18-23](docs/technical/architecture/subsystems/events-mirror/mirror.md#L18-L23)); workers must carry generation ownership or be torn down when an intervening loss is superseded.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1916#issuecomment-5571928731

internal/application/events/manager.go
paul-nicolas, line 235: [P2][non-blocking] Stamp the leadership generation onto a retained emitter

`reconcileGeneration` fences a superseded *loss* (`:169`) and a superseded *startup* (`:235`), but nothing fences a superseded *retention*. `managedSink` carries no record of the generation it was started under, so when a leadership loss is superseded by a regain before the loop takes its next `leadershipSnapshot`, `sc.EqualVT(ms.config)` is true at `:218` and this line keeps the `Emitter` created under the old generation.

**Mechanism (corrected -- it is not buffered-channel coalescing).** `leadershipSnapshot` is taken *under* `m.mu` (`:118-129`, `:160`), so the loop only has to be busy for the whole interregnum, not to lose a notification. A real loss->gain pair spans an election timeout, so a literal `Notify` coalesce is implausible on its own; the reachable path is the loop holding `m.mu` across the interregnum -- a `nats.Connect` dial in `createSink` (`sink_nats.go:46`), or `stopSink` awaiting a hung publish under the 30s HTTP timeout (`sink_http.go:44`), with `scheduleStartupRetry` re-entering reconcile every second (`:20`, `:299-332`).

**Impact.** The retained `Emitter` reads its cursor from Pebble exactly once (`emitter.go:170`) and thereafter keeps it in a local variable (`emitter.go:216`); only `Stop` cancels its context (`emitter.go:145-148`, `:162`). It resumes from a pre-interregnum position, re-publishes the delta the new leader already delivered, and writes a lower cursor: `applyEventsSinkUpdate` guards only `Cursor > 0` (`machine_technical_updates.go:218`) and `SetSinkCursor` is a blind `SetBytes` with no monotonic clamp (`state/batch.go:337-351`). Reproduced at runtime against a real store, HTTP sink and direct-apply proposer: the same `*Emitter` survives, the durable cursor regresses `6 -> 4`, and post-flap deliveries duplicate. This contradicts `docs/technical/architecture/subsystems/events-mirror/events.md:13` ("Only the Raft leader emits events, avoiding duplicates") and `:203` ("Followers: Never emit events").

**Why non-blocking.** No Ledger-internal state is corrupted, the cursor converges forward and there is no path to event loss, and the duplicate-delivery class is pre-existing and documented as at-least-once (`sink_clickhouse.go:52-55`). Base is strictly worse in the opposite goroutine ordering -- verified at runtime: `(true)` then `(false)` leaves the node leader with zero emitters, and base's config-changed callback gates on `if m.isLeader` (base `manager.go:102`), so not even a config change revives it; emission stops until the next transition. Blocking this PR reinstates that. What the PR does add is amplification: the replay window grows from "the last in-flight batch" to "the whole foreign-leadership delta", which is permanent for the Databricks sink (blind `INSERT INTO`, no dedup key -- `sink_databricks.go:183-220`).

**Resolution.** Stamp the generation onto `managedSink` in `startSink` (`:251-295`) and, in this loop, restart any sink whose stamp differs from the current generation instead of `continue`-ing on config equality alone. Optionally give the `Emitter` a leader-scoped context the way `backup/orchestrator.go:200-208` does. An FSM-side monotonic clamp is *not* a cheap alternative: `applyEventsSinkUpdate` receives a write-only `*dal.WriteSession` (`dal/batch.go:124-206` exposes no `Get`) and `proposeSinkUpdateOnce` declares empty coverage on purpose (`emitter.go:486-496`), so a clamp would require a Pebble read in apply (invariant #3) or moving the cursor into the covered domain (invariant #6). Include the doc half here: if the fix intentionally preserves coalesced no-teardown, correct `mirror.md:13`'s "on every leadership change" to match. (`events.md:201` is already self-limiting -- its "; reconciliation from a superseded generation is discarded" clause covers this case.)

Re: NumaryBot's existing finding on this file -- same root cause, confirmed, and its mechanism sharpened above. Its P1/blocker rating is not supported by the base comparison.

internal/application/mirror/manager.go
paul-nicolas, line 228: [P2][non-blocking] Record the leadership edge, or state the level-triggered semantic

Same root cause as the events manager. `OnLeadershipChange` (`:102-115`) records only the latest *level*, and the loop's config handler snapshots leadership only after acquiring `m.mu` (`:127-132` -> `:159-161`), so a loss superseded before the loop observes it is never applied and this line is reached with a `Worker` that never stopped. Verified at runtime with `require.Same`: the identical `*Worker` pointer survives `true -> false -> true`, while a paired control confirms an un-coalesced loss does tear down. Note this is not flap-specific -- teardown became asynchronous in general; the flap is just the case where the window is bounded by the flap rather than by loop latency. Contradicts `docs/technical/architecture/subsystems/events-mirror/mirror.md:13` ("on the leader only ... on every leadership change").

**Durable impact is nil, and this is why the mirror case is not the events case.** The retained worker's re-proposals are absorbed by the FSM's contiguous-applied-prefix guard, evaluated before any mutation (`internal/domain/processing/processor_mirror.go:85-90`): `v2LogID <= last` no-ops, `v2LogID > last+1` rejects deterministically. Double-ingest is structurally impossible, and `last_mirror_v2_log_id` -- checker-verified for exact equality (`check/checker.go:923-1034`) -- cannot be advanced wrongly. Non-leader proposals are dropped at `rawNode.Propose` (`node.go:1961-1983`), and the source adapters are `GET`/`SELECT` only (`adapter/v2/source_http.go:44`, `source_postgres.go:56`). Residual harm is read-only source fetches during the window and a possibly rewound `SubPLMirrorSourceHead` -- a reporting projection that is not checker-verified and self-heals on the next `publishIdleStatus`.

Two corrections worth recording. First, `boundariesLoaded = false` on error (`worker.go:257-263`) is **not** the mitigation: the worker can sit parked in its 5s ticker select (`worker.go:196-206`) for the whole flap, never error, and keep a pre-flap `lastAppliedV2LogID`. The FSM guard is what bounds this. Second, `mirror.md:20-21` ("Superseded generations cannot retain *newly started* workers") is accurate and **not** violated -- it describes the `isCurrentLeader` fences at `:207`, `:219`, `:256`; only `:13` is in tension.

**Why non-blocking.** Base reconciled inline and always tore down (`manager.go:82-88` at `ce72d6df1`), but bootstrap dispatched `go handleLeadershipChangeEvent(...)`, so a reordered `(true)`/`(false)` tore workers down while the node *was* leader, and base's `if m.isLeader` loop gate (`manager.go:104`) left ingestion stalled permanently. The PR trades a permanent stall for a transient retention, which is strictly better.

**Resolution.** If level-only is not the intended semantic for a pre-existing worker, record the leadership edge and tear down on any generation whose recorded level was `false`. Otherwise state the level-triggered semantic in `mirror.md:13` in place of "on every leadership change".

Re: NumaryBot's existing finding on this file -- the retention is confirmed, but "can fetch and propose mirror ingestion while this node was not leader" with durable effect is refuted by the FSM guard above. P1/blocker is not supported.

paul-nicolas (commented):
## Multi-model review: Claude + Codex

**Reviewed head:** `7717565c0c28e796c08972c36512156f096f0b80`

**Result: 2 findings posted inline, both P2 non-blocking. No blocking finding.** The design is sound: the inline observer record is correct and cannot stall the Raft thread (`Signal.Notify` is a non-blocking buffered(1) send; `OnLeadershipChange` never takes `m.mu`), lock ordering is `mu -> leadershipMu` with no inversion, no leadership transition can be permanently lost, the new early `return`s leave no unreconciled half-state, `Stop` has no window to start a worker after teardown, and the `stopped` guard also makes `Stop` idempotent (removing a double-`close` panic on repeat).

Both posted findings are the same root cause in two subsystems: reconciliation is level-triggered, so a leadership loss superseded before the loop takes its next `leadershipSnapshot` is never applied and an already-running emitter/worker is retained across it. Both were reproduced at runtime (pointer identity on the retained instance, plus a paired control), and both were compared against base `ce72d6df1` in **both** goroutine orderings. **Base is strictly worse** -- it stalled emission/ingestion permanently on the reverse ordering -- which is why neither is blocking.

### Verification

19 candidate findings entered synthesis; 6 met the posting threshold and went to adversarial verification (one verifier per finding, in isolated throwaway worktrees, instructed to refute). Outcome: **0 confirmed as filed, 6 corrected (`adjust`), 0 rejected outright** -- but 4 of the 6 were corrected downward past the posting threshold and are listed as P3 below rather than posted inline:

- **events retention** -- P1/blocking -> **P2/non-blocking.** Mechanism corrected: it is *not* buffered-channel coalescing (a real loss->gain spans an election timeout); it is the loop holding `m.mu` across the interregnum. Severity corrected on the base comparison. Verifier also established that no FSM-side clamp on the sink cursor is possible today (`WriteSession` is write-only), which redirects the fix into the manager.
- **mirror retention** -- P1/blocking (NumaryBot, Codex) -> **P2/non-blocking.** My own proposed mitigation was refuted (`boundariesLoaded` does not help a parked worker); the real bound is the FSM contiguous-prefix guard. One doc citation (`mirror.md:20-21`) withdrawn as not violated.
- **docs overstate per-transition teardown** -- P2 -> **folded into the events finding.** It has no independent resolution: fixing the retention makes both doc lines true again. The `events.md:201` half was withdrawn -- the finding truncated its own quote and dropped the caveat clause this PR added.
- **events reconcile coverage loss** -- P2/blocking -> **P3.** All probe numbers reproduced exactly (`startSink` 20/10/20/10/10 -> 0), but the "zero fence coverage" claim is wrong (`:226` and `:235` do execute) and none of the five tests ever asserted on emitter state -- two were already always-true `require.Eventually` on base. Loss of incidental `-race` surface, not of assertion strength.
- **initial-catchup test tautologies** -- P2 -> **P3.** Confirmed by mutation, but the file is byte-identical base-to-head.
- **superseded-generation tests SIGSEGV** -- P2 -> **P3.** Both cited `testing.md` rules fail on inspection; one of them is actually satisfied. Blast-radius number corrected (14 of 16 mirror results silenced, not 15 of 16) and it only manifests under a mutation the contract forbids committing.

**Codex vs Claude conflicts.** Codex filed exactly 2 findings, both matching NumaryBot's existing comments, both rated P1/blocking. Claude's lanes agreed the retention mechanism is real -- and sharpened it -- but resolved the severity down to P2/non-blocking on evidence neither Codex nor NumaryBot produced: a runtime reproduction, and a base comparison in both orderings showing base fails worse. Codex's stated mechanism ("the buffered signal coalesces") is also imprecise. Codex reported no test or documentation findings.

**Validation considered:** CI green except `NumaryBot review` and `codecov/patch`; `go test` and `go test -race` on events/mirror/bootstrap all `ok` at head; `scripts/agent-check` deliberately not run (read-only review).

### P3 (not posted inline)

- `internal/application/events/manager_test.go:38`, `manager_http_sink_test.go:46` -- five pre-existing events tests no longer reach `startSink` (measured 20/10/20/10/10 -> 0, base vs head). Fix: adopt the `m.Start()` + `requireWorkerNames` pattern this PR already applied to the sibling mirror package, where the removal loop and both fences stay covered.
- `internal/application/events/manager_http_sink_test.go:202` -- two of four assertions no longer mean what their messages say. Mutation-proven: a full revert of the async design still passes 5/5, and a teardown no-op fails only `:235`, the sole surviving proof of teardown.
- `internal/application/mirror/manager_test.go:241`, `internal/application/events/manager_leadership_test.go:35` -- the superseded-generation tests fail by SIGSEGV rather than assertion, silencing 14 of 16 mirror and 54 of 71 events results under the prescribed fence mutation. If fixed: the events fixture needs a non-nil `sink` (`stopSink` calls `ms.sink.Close()`), and the mirror fixture needs `w.w = worker.New()` plus a started goroutine or `Stop()` panics on `close of nil channel` / hangs on a nil `doneCh`.
- `internal/bootstrap/module.go:1798` -- `leadershipChangeManager` is declared between `handleLeadershipChangeEvent`'s doc comment and the function, so godoc attributes that comment -- including the load-bearing note on why the backup orchestrator is not called there -- to the type. The interface has no consumer beyond the two parameters it displaced.
- `internal/application/mirror/manager_test.go:99` -- `requireWorkerNames` takes no `msgAndArgs`, so three call sites lost their rationale strings; on timeout it reports only "Condition never satisfied", with no expected or actual set. `require.EventuallyWithT` + `assert.ElementsMatch` restores both. The set comparison itself is correct.
- `internal/application/mirror/manager_test.go:233`, `internal/application/events/manager_leadership_test.go:27` -- the trailing `require.Empty` in the Stop-fence tests is vacuous: the fixture has a nil store, so no path could populate the map even with the guard removed. Only `require.False(t, isLeader)` bites under mutation.
- No test drives a coalesced flap through the real loop -- both superseded-generation tests call `reconcileGeneration` directly with a hand-captured tuple, so the wiring they depend on could regress undetected.

internal/application/events/manager.go
gfyrag: Done — fixed in commit 80d5bf4ef437bbd273d2aa3334a5f3a153243cbc. The manager now records the generation owning the active emitter set and tears it down before reconciling a newer generation; the deterministic regression forces the loss/regain notifications to coalesce and verifies emitter replacement.

internal/application/mirror/manager.go
gfyrag: Done — fixed in commit 80d5bf4ef437bbd273d2aa3334a5f3a153243cbc. The manager now records the generation owning the active worker set and tears it down before reconciling a newer generation; the deterministic regression forces the loss/regain notifications to coalesce and verifies worker replacement.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1916#issuecomment-5571928731

internal/application/events/manager.go
gfyrag: Done — fixed in commit 80d5bf4ef437bbd273d2aa3334a5f3a153243cbc. The events manager now records the generation owning the entire active emitter set; any newer generation tears that set down before rebuilding it. The regression blocks the real notification loop, proves the loss/regain notifications coalesce, and verifies the emitter pointer is replaced.

internal/application/mirror/manager.go
gfyrag: Done — fixed in commit 80d5bf4ef437bbd273d2aa3334a5f3a153243cbc. The mirror manager now records the generation owning the entire active worker set; any newer generation tears that set down before rebuilding it. The regression blocks the real notification loop, proves the loss/regain notifications coalesce, and verifies the worker pointer is replaced.

paul-nicolas (approved):

gfyrag:
Follow-up on the Shipfox review:\n\n- The retained-resource gap across a coalesced leadership flap is fixed in 80d5bf4ef437bbd273d2aa3334a5f3a153243cbc.\n- Events and mirror now record the generation owning their active resource set and recycle it before reconciling a newer generation.\n- Both regressions drive the real notification loop, prove the loss/regain signals coalesce, and assert resource pointer replacement.\n- The local duplication remains intentionally narrow in this PR; the Node-level generalization is tracked in [EN-1963](https://formance-team.atlassian.net/browse/EN-1963).\n\nThe exact final review, NumaryBot rerun, canonical validation, and broad CI all pass.

[EN-1963]: https://formance-team.atlassian.net/browse/EN-1963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1917, opened 2026-09-07
fix(EN-1960): keep snapshot checkpoints alive during FetchFile

gfyrag:
## What changed

Snapshot `FetchFile` now acquires explicit ownership of its session for the duration of the request. Session close, expiry, and service shutdown retire lookup authority immediately while deferring physical checkpoint cleanup until admitted users have released it.

The storage lifecycle documentation and deterministic concurrency regression coverage are updated with the same contract.

## Why

[EN-1960](https://formancehq.atlassian.net/browse/EN-1960) tracks a race where cleanup could delete a snapshot checkpoint after `FetchFile` found its session but before it opened or finished reading the requested file.

Need: admitted snapshot downloads must remain readable through completion.
Current limitation: session lookup returned an unowned pointer, allowing concurrent close, expiry, or shutdown to delete its backing checkpoint.
Requirement / constraint: retirement rejects new acquisitions and physical cleanup occurs exactly once after the final active user releases ownership.
Evidence: deterministic pre-fix reproduction and post-fix regression in `internal/adapter/grpc/snapshot_session_test.go`.
Durable repository evidence: `docs/technical/architecture/subsystems/storage/follower-sync.md` and `docs/technical/architecture/subsystems/storage/README.md`.

## Technical decision

Decision: add explicit active-use accounting and retired state to snapshot sessions; `FetchFile` acquires/releases ownership, while all cleanup paths retire first and wait for active users to drain.
Why now / why proportionate: the checkpoint filesystem lifetime must cover the complete RPC, not only registry lookup. Centralizing ownership in the session gives close, expiry, shutdown, and multiple readers one lifecycle contract.
Alternatives considered: retaining the unowned lookup leaves the race intact; holding the registry lock across file streaming would serialize unrelated operations and couple network duration to global registry progress.

## Risk

**MEDIUM** — localized gRPC snapshot lifecycle synchronization, with concurrency-sensitive cleanup semantics and deterministic race coverage.

## Validation

- [x] `env -u GOROOT -u GOTOOLDIR nix develop --command bash -c 'AI_REVIEW_BASE_SHA=ce72d6df10142a48fa3193ed369831d5f0252854 bash scripts/agent-check-pr'`
- [x] Targeted race: `go test -race ./internal/adapter/grpc` selected by the canonical validation
- [ ] Full suite / broader validation: GitHub CI

Additional existing evidence: `BEFORE_FIX=BUG_REPRODUCED`; `AFTER_FIX=PASS`; multi-user/boundaries PASS; regression sensitivity PASS; exact final review APPROVE.

Snapshot session lifecycle only. No wire-format, API-shape, FSM, Raft, or persisted-format change.

## Review focus

Please focus on acquisition/release ordering, retirement under close/expiry/shutdown, and exactly-once cleanup after the final concurrent reader.

## Known concerns

Broad CI and human review remain required before merge. No known product-code concern from the focused validation.

[EN-1960]: https://formance-team.atlassian.net/browse/EN-1960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/adapter/grpc/snapshot_session.go
+++ b/internal/adapter/grpc/snapshot_session.go
@@ -3,7 +3,7 @@ package grpc
 import (
 	"crypto/rand"
 	"encoding/hex"
-	"maps"
+	"errors"
 	"sync"
 	"time"
 
@@ -22,6 +22,8 @@ type snapshotSession struct {
 	syncName       string
 	checkpointPath string
 	lastAccess     time.Time
+	activeUsers    int
+	retired        bool
 }
 
 // snapshotSessionStore manages snapshot sessions with TTL-based expiry.
@@ -33,6 +35,8 @@ type snapshotSessionStore struct {
 	logger   logging.Logger
 	ttl      time.Duration
 	stopCh   chan struct{}
+	stopOnce sync.Once
+	stopped  bool
 }
 
 func newSnapshotSessionStore(store *dal.Store, logger logging.Logger, ttl time.Duration) *snapshotSessionStore {
@@ -56,26 +60,50 @@ func (ss *snapshotSessionStore) create(syncName, checkpointPath string) (string,
 	}
 
 	ss.mu.Lock()
+	defer ss.mu.Unlock()
+
+	if ss.stopped {
+		return "", errors.New("snapshot session store stopped")
+	}
+
 	ss.sessions[id] = &snapshotSession{
 		syncName:       syncName,
 		checkpointPath: checkpointPath,
 		lastAccess:     time.Now(),
 	}
-	ss.mu.Unlock()
 
 	return id, nil
 }
 
-func (ss *snapshotSessionStore) get(sessionID string) (*snapshotSession, bool) {
+// acquire pins a session's checkpoint until release is called. A retired
+// session cannot be acquired again, but its active users may finish.
+func (ss *snapshotSessionStore) acquire(sessionID string) (*snapshotSession, bool) {
 	ss.mu.Lock()
 	defer ss.mu.Unlock()
 
 	s, ok := ss.sessions[sessionID]
 	if ok {
+		s.activeUsers++
 		s.lastAccess = time.Now()
 	}
 
 	return s, ok
+}
+
+func (ss *snapshotSessionStore) release(s *snapshotSession) {
+	ss.mu.Lock()
+	if s.activeUsers == 0 {
+		ss.mu.Unlock()
+		panic("releasing snapshot session without an active user")
+	}
+
+	s.activeUsers--
+	cleanup := s.retired && s.activeUsers == 0
+	ss.mu.Unlock()
+
+	if cleanup {
+		ss.cleanupCheckpoint(s.syncName)
+	}
 }
 
 func (ss *snapshotSessionStore) remove(sessionID string) {
@@ -83,10 +111,12 @@ func (ss *snapshotSessionStore) remove(sessionID string) {
 	s, ok := ss.sessions[sessionID]
 	if ok {
 		delete(ss.sessions, sessionID)
+		s.retired = true
 	}
+	cleanup := ok && s.activeUsers == 0
 	ss.mu.Unlock()
 
-	if ok {
+	if cleanup {
 		ss.cleanupCheckpoint(s.syncName)
 	}
 }
@@ -120,8 +150,13 @@ func (ss *snapshotSessionStore) reapExpired() {
 	expiredSessions := make([]*snapshotSession, 0, len(expired))
 
 	for _, id := range expired {
-		expiredSessions = append(expiredSessions, ss.sessions[id])
+		s := ss.sessions[id]
 		delete(ss.sessions, id)
+		s.retired = true
+
+		if s.activeUsers == 0 {
+			expiredSessions = append(expiredSessions, s)
+		}
 	}
 
 	ss.mu.Unlock()
@@ -143,23 +178,35 @@ func (ss *snapshotSessionStore) cleanupCheckpoint(syncName string) {
 		ss.logger.WithFields(map[string]any{
 			"error":    err,
 			"syncName": syncName,
-		}).Errorf("Failed to remove temporary checkpoint for expired session")
+		}).Errorf("Failed to remove temporary checkpoint for snapshot session")
 	}
 }
 
 func (ss *snapshotSessionStore) stop() {
-	close(ss.stopCh)
+	ss.stopOnce.Do(func() {
+		close(ss.stopCh)
 
-	// Clean up all remaining sessions.
-	ss.mu.Lock()
-	remaining := make(map[string]*snapshotSession, len(ss.sessions))
-	maps.Copy(remaining, ss.sessions)
-	ss.sessions = make(map[string]*snapshotSession)
-	ss.mu.Unlock()
+		// Retire all remaining sessions. Checkpoints with active users are
+		// cleaned up by the last release.
+		ss.mu.Lock()
+		ss.stopped = true
+		remaining := make([]*snapshotSession, 0, len(ss.sessions))
 
-	for _, s := range remaining {
-		ss.cleanupCheckpoint(s.syncName)
-	}
+		for id, s := range ss.sessions {
+			delete(ss.sessions, id)
+			s.retired = true
+
+			if s.activeUsers == 0 {
+				remaining = append(remaining, s)
+			}
+		}
+
+		ss.mu.Unlock()
+
+		for _, s := range remaining {
+			ss.cleanupCheckpoint(s.syncName)
+		}
+	})
 }
 
 func generateSessionID() (string, error) {
--- a/docs/technical/architecture/subsystems/storage/README.md
+++ b/docs/technical/architecture/subsystems/storage/README.md
@@ -7,7 +7,7 @@ The persistence layer (`internal/storage/dal`, `internal/storage/wal`, `internal
 | Document | Description |
 |----------|-------------|
 | [storage.md](storage.md) | WAL, snapshot/compaction boundaries, runtime stores, persistence, and recovery. |
-| [follower-sync.md](follower-sync.md) | Checkpoint streaming, SHA-256 verification, retries, and WAL reclamation after snapshot install. |
+| [follower-sync.md](follower-sync.md) | Checkpoint streaming, session lifetime, SHA-256 verification, retries, and WAL reclamation after snapshot install. |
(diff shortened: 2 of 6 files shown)


gfyrag:
Shipfox follow-up: verified all three non-blocking observations against the current head. The repeated retirement shape and two-phase reaper are maintainability refactors, not correctness defects; changing concurrency-sensitive structure is intentionally outside EN-1960. The in-flight TTL behavior is also pre-existing and would require a separate expiry-contract decision and regression coverage. No additional product change is needed in this PR.

Azorlogh (approved):

flemzord (approved):
Reviewed session acquisition, retirement, final-reader cleanup, and shutdown integration. No actionable findings. Snapshot tests pass with the race detector and CI is green. The unrelated profiling timing test failed once locally and passed on an isolated rerun.

internal/adapter/grpc/snapshot_session.go
```diff
@@ -120,8 +150,13 @@ func (ss *snapshotSessionStore) reapExpired() {
 	expiredSessions := make([]*snapshotSession, 0, len(expired))
 
 	for _, id := range expired {
-		expiredSessions = append(expiredSessions, ss.sessions[id])
+		s := ss.sessions[id]
 		delete(ss.sessions, id)
+		s.retired = true
+
+		if s.activeUsers == 0 {
```
paul-nicolas, line 157: [P2][non-blocking] Bound the FetchFile pin so checkpoint retention is not tied to a stalled stream

**Evidence.** Pre-PR, `reapExpired` unconditionally called `cleanupCheckpoint` once a session passed the TTL, so `RemoveTemporaryCheckpoint` (`os.RemoveAll`, `internal/storage/dal/store.go:897`) unlinked every hard-linked SST in the checkpoint at TTL + <= 60s (`defaultSessionTTL` 5m, `sessionReapInterval` 60s). Verified empirically that this reclaims the space of every checkpoint file except the one the handler currently has open: with two 400MB hard-linked files whose live links were already dropped, `RemoveAll` with one FD open returned 400MB immediately and the rest only on FD close.

Post-PR, expiry only retires the session and defers `cleanupCheckpoint` to the last `release()` (`snapshot_session.go:100-106`, `:154-159`), so nothing is unlinked while a handler is parked. Nothing bounds that handler:
- `streamOneFile` takes no context and exits only when `Send` returns an error (`internal/adapter/grpc/file_streaming.go:63-125`);
- `FetchFile` has no server-side deadline (`internal/adapter/grpc/server_snapshot.go:134`);
- the follower sets no per-call deadline on the stream (`internal/application/ctrl/file_fetcher.go:78` - only `PrepareSnapshot` and `CloseSession` are capped, `internal/application/ctrl/snapshot_fetcher.go:23-33`, `:258-263`);
- `NewRaftServer` configures no keepalive (`internal/adapter/grpc/server.go:789-806`), so grpc-go v1.82.1 defaults apply (`MaxConnectionIdle`/`MaxConnectionAge` = infinity).

A follower that keeps the HTTP/2 connection alive but stops draining the stream - e.g. blocked writing its `.tmp` file to a stalled disk at `file_fetcher.go:110` - fills the flow-control window, parks the leader's `Send`, and keeps `activeUsers > 0` for the duration of the stall. grpc-go answers keepalive pings from its transport reader goroutine regardless of whether the application calls `Recv`, so the 2h ping does not break the alive-but-not-draining case (it does bound a genuinely dead peer).

**Secondary: the same event is silent.** `reapExpired` only appends to `expiredSessions` when `activeUsers == 0` (`:157-159`), so the "Reaping expired snapshot session" log (`:164-168`) never fires when a pinned session is retired. `release` -> `cleanupCheckpoint` logs only on failure, and no metric covers it. `stop()` has the same gap (`:199-201`).

**Impact.** Leader-side retention of one temporary checkpoint per stalled follower moves from "bounded by the 5-minute session TTL" to "bounded by whenever the stalled handler returns", with no operator signal. A Pebble checkpoint hard-links live SSTs, so the pinned checkpoint blocks reclamation of every SST Pebble obsoletes during the stall - up to roughly a full DB copy of extra disk on the leader. This is per stalled follower, not cumulative per retry: the errgroup ctx (`snapshot_fetcher.go:216`) cancels abandoned streams, which errors `Send` and releases the pin, so a follower cannot open a new session while its previous one is still pinned.

**Why non-blocking.** The pre-PR alternative was the removal-mid-stream race this PR exists to fix, so the trade-off is deliberate and correct in direction. The gap is the new upper bound, not the fix.

**Resolution.** Bound the pin itself - a server-side deadline on `FetchFile`, or a context-aware `streamOneFile` that observes the stream context - so `release()` is guaranteed to run within a bounded multiple of the TTL; and log (or count) when expiry/`stop` retires a session that still has active users.

paul-nicolas (commented):

**Verdict: FINDINGS (0 blocking).** Head reviewed `1aff86521762c2b2255331faf4134e9f1221ffd9`. Not a re-review (no prior review comments on this PR).

The fix is sound and the test set is genuinely regression-bearing. Independent verification: the new test file was ported into a base worktree at `e3ef77087` with a shim reproducing base semantics (`acquire` == base `get`, `release` a no-op), and **5 of 6 new tests fail on base** — one failing assertion per lifecycle trigger (CloseSession `:69`/`:141`, reaper expiry `:168`, multi-user release `:190`, shutdown `:211`, which also reproduces the base double-`stop` `close of closed channel` panic). `TestSnapshotService_FetchFileReleasesSessionUseOnCompletion` passes on base — it is a control guarding the new `defer release`, not regression coverage. The server-level racing test reaches the original EN-1960 state through `server.FetchFile`, so the production trigger is covered, and `server_snapshot_test.go` lost no assertion (2 insertions / 1 deletion, the mechanically required `get` -> `acquire` swap). Clean under `-race`, 11 PASS, no build tags, covered by CI's `go test -race ./...`.

Also verified clean, no finding: acquire/release accounting cannot under/over-count; cleanup is exactly-once across `remove`/`reapExpired`/`stop` (map-absence <=> retired holds under one critical section, and `syncName` is unique per process); the `retired` flag is not redundant (`release` holds only a pointer and cannot test map membership); the new `create()` error path does not leak a checkpoint (`PrepareSnapshot:111-118` already removes it, and `RemoveTemporaryCheckpoint` is a plain `os.RemoveAll` independent of the Pebble handle); no unlocked read of a mutable field; `stop()` vs the reaper cannot double- or lose-cleanup.

### Counts

| | |
|---|---|
| Blocking | 0 |
| P2 non-blocking (posted inline) | 1 |
| P3 non-blocking (listed below, not posted inline) | 6 |

### Verification gate

3 findings met the posting threshold and were each handed to an independent verifier instructed to refute it. **1 confirmed with corrections, 1 downgraded below the inline threshold, 1 rejected:**

- **adjust** — the checkpoint-pin finding (posted). Two over-claims removed: "repeated stalled syncs accumulate distinct pinned checkpoints" is unsupported (client-side stream cancellation releases the pin first), and pre-PR expiry freed all-but-one checkpoint file rather than the whole checkpoint.
- **adjust, P2 -> P3** — the racing test's unguarded `<-firstSend`. The hang is real (reproduced twice by mutation), but diagnosability is largely preserved: the timeout panic names the test and line 136 without `-v`, completed tests still report, sibling tests give a named assertion for the same regression, and the identical shape is pre-existing in the untouched `TestSnapshotService_ParallelFetchFile` in this package. Demoted to the P3 list.
- **rejected (invalid)** — "the new `PrepareSnapshot` store-stopped rejection is unclassified for client retry". The branch is effectively unreachable through the handler: fx stops the Raft gRPC server *before* `StopSnapshotService` (`module.go:831` vs `:904`, reverse-order `OnStop`) via `Server.Stop()`, so the only window is an in-flight RPC whose connection is already closed — the client sees `Unavailable` from the transport, not the handler's status. The proposed `codes.Unavailable` fix is also a behavioral no-op: `isUnavailableError` short-circuits to an immediate return at `snapshot_fetcher.go:95`, and both codes collapse to `syncFailed: true` at `applier.go:1050`. Failing fast to the outer loop, which re-resolves the leader, is the correct response to a dying leader.

**Codex reviewed the same diff independently and returned APPROVE with zero findings.** No contradiction with the Claude lanes — Codex refuted nothing; it surfaced none of the items above. Where the two differ, the evidenced findings are kept.

### P3 (not posted inline)

- `internal/adapter/grpc/snapshot_session_test.go:136` — bound the racing test's first-send handshake on `fetchDone`; on a future acquire/stream regression the bare `<-firstSend` stalls the package binary for the full CI `-timeout 20m`. Do **not** add a `t.Context().Done()` arm: it is cancelled only just before Cleanup functions run, so it cannot fire while the test body is blocked. Match the bounded-handshake pattern in `server_bucket_audit_consistency_test.go:195-198`.
- `internal/adapter/grpc/snapshot_session.go:97` — `panic` in `release()` is unreachable today, but `RaftServer` installs only `raftAuthInterceptors` (`server.go:799-804`); `recoveryStreamInterceptor` is wired only into `NewServiceServer`. A future mispaired defer would terminate the node process rather than one RPC. Invariant 7 is satisfied either by adding recovery to `RaftServer` or by logging the violation instead.
- `docs/technical/architecture/subsystems/storage/follower-sync.md:22` — "the last active user removes the checkpoint" overstates the in-process guarantee. `stop()` no longer removes pinned checkpoints synchronously, the Raft server is force-stopped first, and neither `stop()` nor the reaper is joined, so on shutdown the removal can still be owed when the process exits. The real backstop is `Store.cleanupTemporaryCheckpoints` wiping `tmp/` on next boot (`dal/store.go:966-975`) — worth naming. Also, when a session has zero active users `stop()` does remove it synchronously, with no "last active user" involved.
- `internal/adapter/grpc/server_snapshot.go:111` — the checkpoint-removal branch on session-creation failure was effectively unreachable on base (`crypto/rand` only) and is now reachable via `ss.stopped`. It guards exactly the leak class this PR fixes and has no test. Low value given the reachability finding above, but cheap to cover.
- `internal/adapter/grpc/snapshot_session_test.go:142` — the retired-session rejection is asserted only through the private `sessions.acquire`, never through `server.FetchFile` returning `codes.NotFound`. `follower-sync.md:19-23` documents that externally observable contract; `TestSnapshotService_InvalidSessionID` covers only an unknown id.
- `internal/adapter/grpc/server_snapshot.go:53` — `StopSnapshotService` has zero test callers (all tests call `sessions.stop()` directly), so the public shutdown entry point whose contract this PR rewrites, and its type-assertion false branch, are unexercised.

PR #1920, opened 2026-09-08
test(EN-1965): specify and cover the AddressTxIterator union

paul-nicolas:
[EN-1965](https://formance-team.atlassian.net/browse/EN-1965)

## Scope after rebase

The optimization itself landed in #1919 (`perf(query): sort address transaction unions once`), which merged first. This PR is rebased on top of it and now carries only the material that #1919 did not: the read-path specification, the regression tests, and the benchmarks.

`git log release/v3.0..HEAD` is two commits; the code delta is doc comments plus one line — `sort.Slice` with an inline closure becomes `slices.SortFunc(it.txns, bytes.Compare)`. Both are unstable pdqsort and the IDs are unique after `txSeen`, so that is a same-semantics swap, not a behavior change.

## Docs

`docs/technical/architecture/subsystems/read-path/iterator-seek-contract.md` gains a "The materialized union (`AddressTxIterator`)" section stating the **observable** requirement rather than the algorithm:

- before any positioning call returns, the exposed slice is sorted and unique — `ensureMaterialized` is the single gate in front of both `Next` and `SeekGE`, and it returns only after the sort;
- the order *during* materialization is unspecified: IDs are appended as scanned, deduplicated through a `uint64` set, and the completed slice is sorted once. Nothing outside `materialize` may observe the intermediate order.

It also records why sorting per insertion was quadratic (an address match visits accounts in address order and each `account→tx` scan is ascending, so with interleaved histories almost every new ID lands near the front: O(U²) element movement for U unique IDs, against O(U log U) for one sort), and that neither variant makes an O(pageSize) memory claim — both materialize in full, with immutable 8-byte ID copies rather than retained Pebble key buffers.

The read-path README index line is extended. The doc comments on the type, on `materialize`, and on `NewAddressTxIterator` are restated against this contract; the constructor's stale "(merge-union)" wording is dropped.

## Tests

New in `iterator_address_test.go`:

- `TestAddressTxIterator_UnionIsSortedAndUnique` — 8 subtests: single ascending account, interleaved histories, strictly descending append order, duplicate-heavy, partially overlapping histories, a matched address with no rows, an unmatched account, and no matched addresses. Each asserts the drained union *and* that a seek below the first entry sees the same sorted union, covering both entry points into `ensureMaterialized`.
- `TestAddressTxIterator_UnionIsSortedPerAddressRole` — the three buckets picked by `addressRolePrefix` (any, source, destination).
- `TestAddressTxIterator_AddressRoleBucketsAreIsolated` — rows written for one role stay invisible to another role's iterator.

Additive: the existing `TestAddressTxIterator_SeekGEIsAbsolute`, `TestAndIterator_AddressTxChildKeepsIntersection`, `TestNotIterator_AddressTxChildExcludesAfterExhaustion` and `TestAddressTxIterator_EmptyUnion` triggers are untouched.

`newTestStore` was widened from `*testing.T` to `testing.TB` so the benchmarks share the existing fixture; its 37 callers are unchanged.

**Mutation proof:** replacing the sort with a no-op fails 6 of the new subtests plus the pre-existing `TestAddressTxIterator_SeekGEIsAbsolute`. The tests detect the sort's absence rather than merely passing alongside it.

## Benchmarks

`iterator_address_bench_test.go` holds the workloads that keep the sort honest: interleaved multi-account histories, an already-ascending single account (a one-shot sort can only lose there, so any small-history regression is visible rather than implicit), and duplicate-heavy unions that report `scannedRows` alongside `uniqueIDs` — plus the source and destination role buckets. Sizes 1k/10k/50k.

Figures below are pre-#1919 base vs the merged implementation, median of 3, `-benchtime=1s`, Apple M5 Pro. Base was measured by porting this benchmark file into a detached pre-#1919 worktree, so these show direction of change rather than head-only numbers. They are what the benchmarks record, not a deliverable of this PR.

| Benchmark | pre-#1919 | with #1919 | change |
|---|---|---|---|
| Interleaved 50k | 165.9 ms | 7.30 ms | **22.7x faster** |
| Interleaved 10k | 5.89 ms | 1.31 ms | 4.5x faster |
| Interleaved 1k | 154.8 µs | 113.2 µs | -27% |
| SortedSingleAccount 50k | 4.68 ms | 3.91 ms | -16% |
| SortedSingleAccount 10k | 911 µs | 781 µs | -14% |
| SortedSingleAccount 1k | 89.3 µs | 82.1 µs | -8% |
| DuplicateHeavy 50k (200k scanned) | 9.98 ms | 9.73 ms | -2.5% |
| DuplicateHeavy 10k (40k scanned) | 1.86 ms | 1.80 ms | -4% |
| DuplicateHeavy 1k (4k scanned) | 174.1 µs | 173.0 µs | ~0 |
| Interleaved 10k, source role | 6.46 ms | 1.32 ms | 4.9x faster |
| Interleaved 10k, destination role | 5.95 ms | 1.32 ms | 4.5x faster |

No small-history regression: the 1k and already-ascending cases improve too, because `insertSorted` paid a binary search plus a bounds-checked append per element while pdqsort short-circuits an already-sorted run. Allocation counts and bytes/op are unchanged (~9.09 MB at 50k).

Direction and magnitude match the diagnostic recorded on the ticket (288.4 ms → 23.4 ms at 50k); the absolute numbers differ because that probe ran on different hardware and a different harness.

## Validation

- `bash scripts/agent-check` — PASS
- `go test -race -count=1 ./internal/storage/readstore/...` — ok; 16 `TestAddressTxIterator*` cases pass, including the 8 subtests above

No storage or wire migration, no new cache, no weakened invariant.

## Note for review

The ticket's plan listed a "reverse write order" test. That case is unreachable: Pebble scans each account bucket in key order, so write order never reaches `it.txns`. The unsorted append order comes from the address iterator visiting accounts whose ID ranges overlap, which the `descending append order` subtest encodes instead (acc:1={7,8,9}, acc:2={4,5,6}, acc:3={1,2,3}).

`SeekGE`'s `sort.Search` was left as-is. Migrating it to `slices.BinarySearchFunc` would be a clean swap and would drop the `sort` import, but it touches the seek path that EN-1597 fixed; out of scope here.

[EN-1965]: https://formance-team.atlassian.net/browse/EN-1965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/technical/architecture/subsystems/read-path/README.md
+++ b/docs/technical/architecture/subsystems/read-path/README.md
@@ -16,7 +16,7 @@ and the pipeline pages.
 | Document | Description |
 |----------|-------------|
 | [query-pipeline.md](query-pipeline.md) | End-to-end read flow: ReadIndex barrier, min_log_sequence, Pebble snapshot, iterator algebra, pagination, streaming. |
-| [iterator-seek-contract.md](iterator-seek-contract.md) | Absolute SeekGE/SeekLE semantics across the iterator algebra, and the seekFloor/seekCeil exhaustion-proof cache. |
+| [iterator-seek-contract.md](iterator-seek-contract.md) | Absolute SeekGE/SeekLE semantics across the iterator algebra, the AddressTxIterator materialized union, and the seekFloor/seekCeil exhaustion-proof cache. |
 | [read-snapshot-consistency.md](read-snapshot-consistency.md) | Single-snapshot rule for controller reads that stitch LedgerInfo with attribute data. |
 | [readstore-event-keys.md](readstore-event-keys.md) | Append-only metadata/existence event resolution, lease-bounded reclamation, and edge-triggered GC cycles. |
 | [prepared-queries.md](prepared-queries.md) | Named pre-validated query templates: lifecycle, filter DSL, execution, bloom acceleration. |
--- a/docs/technical/architecture/subsystems/read-path/iterator-seek-contract.md
+++ b/docs/technical/architecture/subsystems/read-path/iterator-seek-contract.md
@@ -44,18 +44,43 @@ The contract is enforced by unit tests per leaf (`iterator_floor_test.go`,
 contradiction specs in
 `tests/e2e/business/filter_nested_not_reposition_test.go`.
 
-## AddressTxIterator materialization
+## The materialized union (`AddressTxIterator`)
 
-`AddressTxIterator` materializes the union of transaction IDs lazily on first
-positioning call and keeps it as a stable sorted slice for the iterator's
-lifetime. Materialization deduplicates via a seen-set, appends each unseen ID
-as an owned copy (never a Pebble iterator key buffer), and **sorts the slice
-once** when the scan completes. The observable requirement is O(U log U)
-sorting work for U unique IDs rather than the O(U²) shifts of the former
-per-insert `insertSorted`, which an interleaved account history (evens in one
-account, odds in another) triggers. This is a pure cost change: the emitted
-IDs remain identical, sorted and unique, so the absolute-seek contract above is
-unchanged.
+`iterator_address.go`. An address match on the `TRANSACTIONS` target has no
+entity-ordered index to scan: it walks the matching account addresses and, per
+account, scans that account's `account→tx` bucket. Each per-account scan is
+ascending, but the accounts are visited in address order, so the transaction
+IDs arrive out of order across accounts.
+
+The iterator satisfies the absolute-seek contract by materializing the whole
+union once, on first use, and keeping it for the iterator's lifetime; `Next`
+and `SeekGE` are then cursor moves over a stable sorted slice. `SeekGE`
+binary-searches that slice, which makes it computed from `target` alone,
+idempotent, and well-defined after exhaustion for free.
+
+The observable requirement is on the *exposed* slice, not on how it is built:
+
+- Before any positioning call returns, the slice is **sorted and unique**.
+  `ensureMaterialized` is the single gate in front of both `Next` and
+  `SeekGE`, and it returns only after the sort.
+- The order **during** materialization is unspecified. IDs are appended as
+  they are scanned, deduplicated through a `uint64` set, and the completed
+  slice is sorted once (`slices.SortFunc` with `bytes.Compare`, which is the
+  numeric order of the 8-byte big-endian IDs — see `ReadStoreComparer`).
+  Nothing outside `materialize` may observe the intermediate order.
+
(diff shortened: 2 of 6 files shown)


gfyrag:
The same EN-1965 optimization was implemented in parallel in #1919, which has already been merged. This PR therefore materially overlaps with work that is now on release/v3.0. Could you rebase it and keep only the complementary documentation, regression tests, and benchmarks that are not already covered by #1919, or close it if nothing distinct remains?

paul-nicolas:
Rebased onto `release/v3.0` (`8a2a7c85b`) — head is now `b03d5156b`. #1919's merge was already an ancestor of this branch, so the optimization is no longer in the delta: `git log release/v3.0..HEAD` is two commits and the code change left is doc comments plus one line (`sort.Slice` with an inline closure → `slices.SortFunc(it.txns, bytes.Compare)`, same unstable pdqsort, IDs unique after `txSeen`).

What remains is only the complementary material, none of which is on `release/v3.0`:

- `iterator-seek-contract.md` — the "The materialized union (`AddressTxIterator`)" section: the observable requirement (exposed slice sorted and unique before any positioning call returns, order *during* materialization unspecified, `ensureMaterialized` the single gate) and why per-insert sorting was quadratic.
- `internal/storage/readstore/iterator_address.go` — type, `materialize` and constructor doc comments restated against that contract.
- `iterator_address_test.go` — `TestAddressTxIterator_UnionIsSortedAndUnique` (8 subtests), `..._UnionIsSortedPerAddressRole`, `..._AddressRoleBucketsAreIsolated`.
- `iterator_address_bench_test.go` — interleaved / already-ascending / duplicate-heavy at 1k/10k/50k, plus the source and destination buckets.

Retitled to `test(EN-1965)` and rewrote the body, which still described the perf change as this PR's deliverable; the figures are relabelled as pre-#1919 vs merged, i.e. what the benchmarks record rather than something this PR delivers.

Happy to drop the `slices.SortFunc` line too if you would rather this PR be strictly docs+tests.

Validation: `bash scripts/agent-check` PASS, `go test -race -count=1 ./internal/storage/readstore/...` ok.

gfyrag (approved):
Re-reviewed at b03d5156. The PR is now correctly scoped after #1919: the remaining production change is semantics-preserving, while the durable contract, regression tests, and benchmarks add useful coverage. Targeted race-enabled readstore tests and CI pass.

PR #1930, opened 2026-09-08
perf(EN-1966): avoid draining all matches for descending pages

paul-nicolas:

## What changed

Direction is now compiled rather than applied after the fact. `query.CompileReverse` mirrors `Compile` node for node and returns a `readstore.ReverseIterator`, so `listDescFiltered` has the same shape as `listAscending`: compile, trim to the main-store horizon, hand to `PaginateReverse`, which seeks to the cursor and stops after the page lookahead.

Nine reverse primitives fill the gaps in the descending algebra (And, Not, Bitset, EventResolve, a borrowed slice view, a stamp-gated prefix, a bounded tx range, an account prefix, the address-union view). `ReverseIterator` gains `Close()`.

Public API defaults and cursor encoding are unchanged.

## Why

Fetching the newest 100 matching transactions should not copy every matching ID first. Descending is the public default for transactions, and `listDescFiltered` drained the whole ascending result, reversed it in place, and only then applied the cursor and page size — O(M) visits and O(M) memory per page, even when every leaf in the tree was entity-ordered and could have been walked backwards directly.

**Need:** a descending page must cost the page, not the history. This is the default direction for the transactions endpoint, so the cost is paid by ordinary reads on any ledger with depth.

**Current limitation:** every filtered descending page was linear in the match count, in both time and allocations, and got worse as a ledger grew. Late pages cost the same as first pages.

**Requirement / constraint:** for filter shapes whose leaves are entity-ordered, the entities visited to serve a page must be bounded by the page size and not track the match count. Correctness is unchanged: a full descending traversal must equal the reversed ascending result, and no visibility gate may differ between directions.

**Evidence:** requested by Geoffrey in the Slack review thread; plan reviewed adversarially and recorded on EN-1966 (PASS, 0 blocking findings). Benchmark below.

**Durable repository evidence:** `docs/technical/architecture/subsystems/read-path/query-pipeline.md` (direction-aware compilation, the two materializing fallback classes and why each remains), `docs/.../iterator-seek-contract.md` (the descending algebra rules), and the tests named under Validation.

## Technical decision

**Decision:** a parallel `compileReverse` recursion sharing `compileCtx`, rather than extending `EntityIterator` with `Prev`/`SeekLE` so one tree serves both directions.

**Why now / why proportionate:** the bidirectional-interface alternative touches every iterator and every mock, rewrites the seek contract, and forces bidirectional state onto leaves that will never need it — `RangeIterator` cannot even implement forward `SeekGE`. Blast radius far exceeds the payoff. The chosen shape adds files rather than changing existing contracts.

**Alternatives considered:** bidirectional `EntityIterator` (rejected, above); a ring buffer over the existing drain (rejected — it lowers final-page memory while still walking the whole history, which is explicitly not what this ticket asks for); do nothing (rejected — this is the default read direction).

Iterator *construction* is duplicated between the two entry points, and that is more than the leaves: `compileRev` carries its own recursion shape too — the depth guard, the per-target `rejectInvalidCondition` check, the filter-type dispatch switch, and the AND/OR/NOT/universe composition with its profile-tree wiring. A change to ascending composition semantics does **not** propagate on its own. What *is* shared is what a filter means: predicate resolution, schema validation, index-readiness gating and bound computation are shared functions (`resolveFieldMetadataCtx`, `resolveTxTimestampArm`/`resolveLogDateArm`, `resolveIntBounds`/`resolveUintBounds`, `intRangeBounds`/`uintRangeBounds`/`timestampRangeBounds`/`logIDRangeBounds`), so the two directions cannot disagree about what a filter *means* — only about which way they walk it.

### Two leaf classes stay materializing, deliberately

| Fallback | Why it cannot stream backwards | Descending behaviour |
|---|---|---|
| Value-ordered ranges — int/uint metadata, transaction timestamp / inserted-at / reverted-at, log date, log-id | The scan spans several index-value buckets, so rows surface in `(value, entity)` order and "the next entity below X" is undefined without the sorted result | `materializeReverse` reuses the ascending path's single `materializeIterator` result and hands out a borrowed `ReverseSliceIterator` over it |
| `AddressTxIterator` — the account→transaction union, including the exact-address form | Members come from N per-account scans, each ascending but collectively unordered | `ReverseAddressTxIterator` shares the same `addressTxUnion` and its one sorted slice |

Neither adds a second complete-result collection for the reversal, and both stay visible in the iterator tree under their own `Kind` so `query-profile` still attributes their cost. Making them stream is separate work (EN-1965, EN-1967).

## Measured result

In-binary A/B on one seeded store (`compile_reverse_bench_test.go`), one streaming equality leaf, page size 100. `drain` reproduces the old algorithm in the same process, so the delta is the algorithm and not machine noise:

| matches | page | drain | stream | rows visited | allocs |
|---|---|---|---|---|---|
| 1,000 | first | 167 µs | 27 µs | 1,000 → 101 | 2,082 → 282 |
| 10,000 | first | 689 µs | 13 µs | 10,000 → 101 | 20,095 → 282 |
| 100,000 | first | 6.30 ms | 15 µs | 100,000 → 101 | 200,110 → 282 |
| 100,000 | late | 8.52 ms | 17 µs | 100,000 → 101 | 200,110 → 283 |

The benchmark reports raw event work (`rows_visited`) separately from output collection (`rows_out`) on purpose: it is the metric that distinguishes a real traversal fix from a smaller final-page buffer. It is flat at 101 across every dataset size.

Local diagnostic measurements, not cluster TPS forecasts.

## Risk

**MEDIUM**

Read-path iterator algebra, where a mis-implemented seek drops or repeats rows silently rather than failing. No persisted state, no wire format, no new cache, no weakened invariant, and the ascending path is refactor-only. Mitigated by the parity oracle and the conformance suite below, both of which were mutation-tested rather than assumed.

## Validation

- [x] Targeted tests: `-race` on `internal/query`, `internal/storage/readstore`, `internal/application/ctrl`; `golangci-lint --max-same-issues=0` on all three (0 issues)
- [x] Full suite / broader validation: `AI_REVIEW_BASE_SHA=72d51387e bash scripts/agent-check-pr` (escalated to full) — PASS

New tests:

- **`compile_reverse_parity_test.go`** — the acceptance oracle. For 48 target/filter cases across ACCOUNTS, TRANSACTIONS and LOGS × 6 page sizes, a full descending traversal **by pages** (each resuming from the previous page's last id, as the controller does) equals the reversed ascending reference: no duplicates, no omissions. Plus cursor boundaries and uint64 extremes. `TestDescendingParity_NoFullMaterialization` asserts `MaterializedRanges == 0` for streaming compositions, and exactly 1 — with the same `MaterializedItems` as the ascending run — for the fallback.
- **`iterator_conformance_test.go`** — a registry-driven forward/reverse suite. Every pair is checked for set parity, gate parity at a pin, the absolute `SeekLE` contract, and error propagation, so a reverse leaf added later is covered without anyone remembering to write those four tests.

`ReverseIterator` gains `Close()`, which it lacked while `EntityIterator` had it. Without it a reverse composite cannot own its children and each call site needs a closer shim; adding it removes the `reverseCloser` wrapper in `list_entities.go`. Contained to the read path.

`ReversePrefixIterator` gains the fold-sequence stamp gate `PrefixIterator` already carried — a behaviour **fix**: descending scans of the account-by-asset index previously admitted rows folded past the reader's pin that ascending scans correctly hid.

`AddressTxIterator` is restructured around a shared `addressTxUnion`. The `materialize` body is unchanged, so the overlap with EN-1965 should be a mechanical conflict.

## Review focus

- **`ReverseEventResolveIterator` group resolution.** Events are seq-ascending in key order, so walking a group backwards the *first* event with `seq <= pin` is the latest one at or below it — the event the forward pass settles on last. Worth a second pair of eyes. Its `SeekLE` bound is `prefix+target+(terminator+1)`, which relies on entity bytes never being NUL or `0x01`.
- **`ReverseAndIterator.converge`.** The comparisons are the mirror of `AndIterator.converge` and mirroring them wrongly is easy.
- **The shared/duplicated split in `compile_reverse.go`** — whether anything I left duplicated should have been shared.

## Known concerns

Two, both worth stating rather than leaving to CI:

1. **My parity oracle initially had a hole, found by mutation testing.** Latching `ReverseNotIterator`'s excluded child — the descending form of the EN-1597 bug, and the highest-likelihood risk on this change — passed the full paged parity suite green. A paged traversal rebuilds the iterator tree per page, so it never re-seeks a child that iteration already consumed. `TestReverseNotIterator_SeekLERepositionsConsumedChild` now drives that seek directly and fails under the mutation. Flagging it because it means the parity oracle alone is *not* sufficient for that class of bug, and a reviewer should not read "parity is green" as covering it.

2. **Merge order with EN-1965.** `origin/perf/en-1965-sort-address-transaction-unions-once` rewrites `AddressTxIterator.materialize`, which this PR moves into `addressTxUnion`. The two are independent by design, but whichever lands second needs a deliberate rebase rather than an automatic one.

Two mutations were run to confirm the new tests bite, not just pass: removing the `ReversePrefixIterator` stamp gate fails `TestIteratorPairs_GateParity` (descending leaks a row stamped above the pin), and an off-by-one in `ReverseSliceIterator.SeekLE` fails the parity oracle.

Also worth noting for scope: `internal/query/audit.go:134` reverses for descending audit pages too, but it reverses an already-materialized candidate-sequence set rather than draining a full result to build one. Different shape, different subsystem, left alone.


gfyrag:
The optimization target is right: compiling a descending traversal directly avoids the O(M) drain/reverse path and restores page-sized work. But I think we should challenge the implementation direction before merging, because in Ledger we try to keep equivalent paths **isomorphic by construction**.

The current design makes that isomorphism a convention rather than a property of the code. `compile_reverse.go` is a 992-line second compiler which mirrors `Compile` node-for-node. Sharing predicate resolution, schema/index validation and bound helpers is useful, but it does not eliminate semantic duplication: AST recursion and dispatch, AND/OR/NOT composition, leaf selection, error wrapping, resource cleanup and profile-tree construction still have to be maintained twice. A new filter kind or a change in composition can therefore be correct in one direction and missing/different in the other.

The same pattern appears in the iterators: `Filter`/`Tracked`, `And`, `Or` and `Not` each have a forward and reverse copy whose essential difference is the seek operation and ordering comparison. These merge/convergence algorithms are exactly where maintaining two implementations is risky. `EventResolve` legitimately has direction-specific traversal, but even there parsing and key-layout validation can be shared. `SliceIterator` is also split asymmetrically between the query compiler and readstore. Finally, now that `ReverseIterator` owns `Close`, `reverseCloser` in `list_entities.go` is redundant; for logs it even closes the raw iterator rather than the composed wrapper, which is equivalent today but bypasses the wrapper's ownership contract.

Could we pivot to a single direction-neutral compiled plan, then lower that plan through forward/reverse iterator factories? An alternative is one shared compiler/combinator implementation parameterized by a small direction policy (`seek`, comparison, extremum), with thin adapters exposing `SeekGE` and `SeekLE`. I am **not** suggesting adding `Prev`/`SeekLE` to every existing iterator: Pebble leaves whose physical traversal truly differs can stay separate. The important part is to share the filter semantics and the generic boolean combinators, while isolating only the genuinely direction-specific leaf mechanics.

The test methodology should enforce the same contract end-to-end: for every supported target (ACCOUNTS, TRANSACTIONS and LOGS), filter family, cursor and page size, exhaust both directions and assert `descending == reverse(ascending)`, including identical compilation errors, index/schema gates, pin/horizon visibility and page boundaries. Today the parity oracle is account-only even though transactions are the motivating/default descending surface; Codecov also reports only 55.31% coverage on the new compiler (227 missing lines). That is a symptom of the duplicated semantic surface, not just a coverage number to patch locally.

So my recommendation is to make direction an execution property of one compiled query semantics, then keep only the irreducible direction-specific iterator code. If we deliberately retain the two compilers, I would at minimum consider the full three-target parity/error matrix and an exhaustive shared registry mandatory before merge, but structurally sharing the plan would give us the stronger isomorphism we are aiming for everywhere else.

paul-nicolas:
Agreed on all of it, and I am taking the pivot rather than defending the two compilers. Split into two PRs, because the restructure has no behavior delta and the existing ascending suite is a complete oracle for it — putting it in the same diff as a new read path throws that away.

**#1951 is up now: the shared algebra, refactor-only.**

One interface, one positioning method:

```go
type Direction interface{ compare(a, b []byte) int } // sealed: Asc | Desc
type Iterator[D Direction] interface {
    Next() bool
    Current() []byte
    Seek(target []byte) bool // first entity at or after target IN D'S ORDER
    Err() error
    Close()
}
type EntityIterator  = Iterator[Asc]
type ReverseIterator = Iterator[Desc]
```

`And`, `Or`, `Not`, `Filter` and `Slice` are now single direction-parameterized implementations. `SliceIterator` moved to `readstore`, so the asymmetry you pointed at is gone. `reverseCloser` is deleted — including the LOGS branch that closed the raw prefix iterator instead of the wrapper it returned. All 67 `EntityIterator` references and every constructor call site are unchanged, because the named constructors are the direction-naming boundary.

I took your second option (one implementation parameterized by a direction policy) over the plan-IR one. Two reasons, both worth your pushback:

- An intermediate plan tree adds a representation and a second traversal for no extra safety, and "plan" is loaded vocabulary here — invariant #6's `plan.Coverage`, `internal/infra/plan`.
- Direction had to stay a *type* parameter, not a runtime flag. Collapsing `SeekGE`/`SeekLE` into one untyped `Seek` would let `PaginateForward` accept a descending iterator and compile; the distinct method names were the only thing preventing that. `Iterator[Asc]` is not assignable to `Iterator[Desc]`, so the guard survives the merge.

The cost of that choice, measured rather than assumed: Go cannot infer a type argument that appears only inside a generic *interface* parameter type (`Iterator[D]` from `Iterator[Asc]` fails; `[]Iterator[D]` from `[]Iterator[Asc]` succeeds), so a few internal entry points need explicit instantiation. That is the whole price.

Two things I want on the record before you read the diff.

**`AndIterator` now positions every child on its first `Next`.** The lazy form was correct ascending only by accident: an unpositioned child returns an empty `Current()`, which sorts below any candidate and reads as "behind, seek forward". Descending, empty reads as "past the end", `converge` adopts it as the candidate, and the intersection collapses to nothing. Results unchanged; only the `Next`/`Seek` mix a child sees differs, and nothing asserts those counts.

**You were right that parity-green is not sufficient, and it bit again here.** Parameterizing And/Not/Slice creates `[Desc]` instantiations that nothing in #1951 consumes, so I added `combinator_direction_test.go` and mutation-tested it. Flipping `Desc.compare` to ascending kills 9 of 10 subtests — the survivor is `Slice`, because a slice's descending *walk* is driven by its step, not the comparator. That needed a second mutation (`SliceIterator.at` ignoring direction, kills 10 of 10). A single mutation would have left me believing the suite covered something it did not.

**What #1951 deliberately does not do:** `compile_reverse.go` is untouched. Collapsing the two compilers into one recursion with per-direction leaf factories is the follow-up, on top of this. Deferred with it: sharing `EventResolve`'s key parsing and layout validation (the reverse iterator does not exist at #1951's base yet, so there is nothing to share with), and the `leafFactory` interface — a factory with one implementation is dead abstraction until the second direction exists.

**On test methodology, I am taking your bar as the requirement, not the fallback.** The follow-up carries the full three-target matrix — ACCOUNTS, TRANSACTIONS and LOGS × filter family × cursor × page size, asserting `descending == reverse(ascending)` plus identical compilation errors, identical index/schema gate rejections, identical pin/horizon visibility and identical page boundaries. You are right that account-only was the wrong scope when transactions is the motivating default surface. The registry also gets an exhaustiveness assertion that *fails* when an iterator has no registered pair, so this cannot regress by omission. On the 55.31% figure: my expectation is that most of those 227 lines disappear with the duplicated surface rather than needing tests bolted onto them, and I will report what is actually left rather than waive it.

Two open questions for you:

1. `Direction`'s seal is an unexported method, so `Asc`/`Desc` are the only implementations. Right call, or do you want it open for a future third order?
2. Merge order. #1951 touches `iterator_address.go`, which collides with `perf/en-1965-sort-address-transaction-unions-once`. I would land #1951 first, then rebase EN-1965, then rebase this PR onto both. Say if you would rather EN-1965 went first.

gfyrag:
For visibility, #1929 is a Shipfox-generated duplicate of this work: both address EN-1966 and replace the full descending-result drain with direct reverse traversal. We have asked Shipfox to close #1929 and will keep this human-authored PR, together with the preparatory work in #1951, as the canonical path forward.

paul-nicolas:
Correction to my previous comment, before you spend time on #1951.

I claimed the type parameter preserved the compile-time guard that `SeekGE`/`SeekLE` gave us — that `Iterator[Asc]` is not assignable to `Iterator[Desc]`. **That was wrong.** NumaryBot caught it on #1951 and it was right.

Go decides interface satisfaction structurally. `D` appeared in no method signature, so the two instantiations had identical method sets and satisfied each other:

```go
_, _, _ = PaginateReverse(NewSliceIterator(nil), 10, nil)        // compiled, go vet exit 0
_, _, _ = PaginateForward(NewReverseSliceIterator(nil), 10, nil) // compiled, go vet exit 0
```

So the branch as first pushed was **strictly worse** than what it replaced: an ascending iterator paginated through a descending cursor, wrong order and wrong rows filtered, with nothing to stop it at compile time. The distinct method names had been carrying real weight and I removed them on a false premise.

Fixed in 7ccda3b1c. `Iterator` carries `Direction() D`, a witness never called at runtime whose only job is to put `D` in the method set. All four crossings now fail with `wrong type for method Direction`; a positive probe confirms correct usage still compiles.

The part I think is worth your attention is *why I did not catch it*. The package is full of lines that look like exactly this assertion:

```go
var (
    _ EntityIterator  = (*SliceIterator[Asc])(nil)
    _ ReverseIterator = (*SliceIterator[Desc])(nil)
)
```

Every one of those compiled just as happily before the fix as after, because both instantiations satisfied both interfaces. They read as a guard and were silent. That is the same shape as the parity-oracle hole you flagged on #1930 and I flagged back at you: a check that passes for a reason unrelated to the property it appears to test. In this branch it appeared in a *compile-time* assertion, which I had been treating as self-evidently sound.

So the guard is now `TestIteratorMethodSetNamesDirection`, a source-level assertion that fails if `D` ever leaves the method set again. I earlier told you I was skipping a negative-compilation test because "the type checker already enforces it on every build" — that reasoning was circular and this is the counterexample. Overruled by events.

Nothing else in the plan changes: #1951 is still the shared algebra, and the compiler collapse plus the three-target parity matrix still follow it.

paul-nicolas:
Addressed the review at `a8147bc1` in 443d79423.

**Both documentation-accuracy findings are fixed.** On what the two compile entry points share: you are right that "only leaf *construction* is duplicated" understated it — `compileRev` carries its own depth guard, `rejectInvalidCondition` call, filter dispatch switch and AND/OR/NOT/universe composition including the profile wiring, and the doc now says so explicitly, with the consequence spelled out (a change to ascending composition must be mirrored by hand). On the parity oracle: the doc claimed a filter-family coverage the matrix did not have.

**The parity gap is closed by extending the matrix, not by trimming the sentence.** Every reverse leaf class reachable from a compiled plan now has a paged parity case: the account→transaction union in both match forms and on a role bucket, the `txref` reverse prefix, the reversion bitset and its complement, the stamp-gated `abya` scan, the signed and bool metadata encodings, the `exists` null OR arm, the `inserted_at` arm of the timestamp fallback, and the log-date fallback — each crossed with AND/OR/NOT compositions and the six page sizes.

Two things make those cases more than green checkmarks:

- `TestDescendingParity_LeafFixtureSizes` pins each family's exact result size. A family whose index rows are missing or written under the wrong prefix still compiles and still yields an empty reference, so `descending == reverse(ascending)` holds on `[] == []` and proves nothing about the leaf it was added for.
- One has-asset row is deliberately stamped **above** the read pin. A direction that drops the stamp gate therefore serves a row the other hides, instead of both agreeing on the same over-wide view.

I verified this by mutation rather than trusting the green: passing `0` instead of `ctx.pin` to `NewStampGatedReversePrefixIterator` in `compileAccountHasAssetConditionRev` fails the has-asset parity cases with `accounts:01` appearing descending only.

**I also closed the one clause of @gfyrag's merge condition that was still open** — identical compilation errors. `TestCompileErrorParity` drives 14 refusals (target guard, depth guard, per-target validity table, schema and coercion checks, index readiness, the "condition has no value" arms) and asserts both entry points fail with the *identical* message; `TestCompileErrorParity_AcceptanceIsAlsoShared` asserts the converse over the whole matrix. Also mutation-verified: deleting `rejectInvalidCondition` from `compileRev` fails four cases.

**On the int/uint compiler pair, I am deferring rather than dismissing.** The duplication is real, but unifying it correctly is not local to `compile_reverse.go`: it needs `resolvedIntBounds`/`resolvedUintBounds` to collapse into one `numericBounds[T int64 | uint64]`, which pulls in `resolveIntBounds`, `resolveUintBounds`, `isEquality`, `mergeFieldRanges` and all four `compile{Int,Uint}Condition{,Rev}` call sites — the ascending compiler included. Doing only the descending half would make the two entry points *less* isomorphic, which is the property this PR is being held to. That belongs in its own refactor with its own review, not folded into a perf change.

Validation: `scripts/agent-check` PASS; `agent-check-pr` against base `658dfaca` reports `ok internal/query`; `golangci-lint --max-same-issues=0` clean on `internal/query`. The gate's one red package is `tests/antithesis/TestRunModelTestRequiresVerifiedOutcome`, failing on `context deadline exceeded` under full-suite load; it passes in isolation in 6.6s and `go list -deps -test ./tests/antithesis/` shows no dependency path to `internal/query`, so this change cannot reach it.

gfyrag:
Fresh review of HEAD `aeee1608`: the reverse compiler, iterator algebra, cursor exclusion, horizon filter, and error propagation otherwise look sound. Two actionable gaps remain.

1. Log-ID ranges are naturally entity-ordered: the Pebble key is the ledger-log prefix followed directly by the big-endian log ID. However, `compile_reverse.go` still routes them through `NewRangeIterator` + `materializeReverse`, and `query-pipeline.md` incorrectly classifies them with intrinsically value-ordered ranges. This keeps that case O(M) before returning a page, despite the stated bounded-work contract for entity-ordered leaves. Please coordinate with #1922 and generalize the bounded entity iterator by direction for the ascending/descending log and transaction ranges. At minimum, correct the documentation and explicitly track this as a temporary exception rather than an ordering limitation.

2. Please port `TestListDescFilteredStopsAtLookahead` from the now-closed #1929. The current parity tests and benchmark invoke `CompileReverse`/`PaginateReverse` directly, so they would not catch a full drain reintroduced in `listDescFiltered` before or after pagination or around the horizon wrapper. The production controller path should be instrumented with cursors and horizon handling and assert no more than the page lookahead is consumed.

The published performance gain is credible for streaming leaves; these findings ensure the production boundary and the naturally ordered log-ID leaf are covered too.

paul-nicolas:
Both addressed in `9b8f1374` and `74cb3c94`.

**The lookahead test is ported, and I mutation-tested it rather than trusting the green.**

`TestListDescFilteredStopsAtLookahead` now sits on `listDescFiltered` itself, with the seven cursor cases and both instruments from #1929: `horizonKeep` as a counting predicate that fails inside the callback (catching over-consumption *outside* the compiled tree, including a drain of the wrapper), and exact equality on `profile.Root.NextCalls`/`SeekCalls` (catching it inside, and pinning `SeekCalls == 0` on the uncursored first page, `1` everywhere else). It hangs off `listDescFiltered` rather than `listEntities` deliberately — `listEntities:110` overwrites `horizonKeep` with the real `MainHorizonKeep`, which would throw away the instrument that sees a drain outside the tree.

Three mutations, because a budget assertion nobody has tried to break is exactly the check that passes for an unrelated reason:

- drain after `PaginateReverse` → **7/7 fail**;
- restore the pre-EN-1966 shape (drain everything, then paginate the materialized slice) → **7/7 fail while still returning the correct rows**. That is the one that matters: it is behaviour-preserving, every other test in the repo stays green under it, and nothing on the branch detected it before this;
- remove the horizon wrapper → **6/7 fail** (the 7th legitimately expects zero visits).

Every hard-coded count from #1929 was still correct at `aeee1608` — none needed touching across #1951's algebra refactor, because the fixture's filter compiles to a single leaf with no combinator above it.

**On log-ID ranges: you are right about the classification, and it was my line. Three corrections, two of them against your comment.**

Right: the key is `llog:<ledger>` followed directly by the big-endian log ID — `logIDRangeBounds` (`compile.go:2027`) appends nothing else, and the leaf is built with `entityOffset == len(prefix)`, `entityLen == 8`. No value bucket, byte order *is* entity order. `query-pipeline.md:185` filed it with the intrinsically value-ordered ranges and gave it that group's reason ("the scan spans several index-value buckets"), which is simply false for this leaf. Fixed: it now has its own row stating the real cause — `RangeIterator` has no reverse form yet — explicitly marked a temporary implementation exception rather than an ordering property, and pointed at #1922. I also corrected the same over-generalization in `iterator-seek-contract.md:99`, which pre-dates this branch.

But two parts of the finding do not hold, and I would rather say so than quietly accept them:

1. **It is not a descending asymmetry, and this PR does not introduce it.** The ascending compiler drains the identical leaf: `compile.go:1488`, `materializeIterator` over the same `NewRangeIterator`. That call is present unchanged at the merge base — `git show 658dfaca:internal/query/compile.go` has it at `:1417`. Both directions materialize equally; the descending page costs exactly the one drain the ascending page already paid. The bounded-work contract is not violated *relative to ascending*, which is the only claim this PR makes. I have written that into the doc too, since the old wording let it read the other way.
2. **Transaction ID ranges already stream descending.** `compileTxIDConditionRev` (`compile_reverse.go:854-875`) builds `NewPebbleReverseTxRangeIterator`, no materialization, with the reasoning in a code comment. There is no outstanding descending transaction-range work.

**On generalizing the bounded entity iterator: routing it to #1922 rather than doing it here.** #1922 *is* `perf(query): stream ordered log ID ranges through pagination` (EN-1967) and is open. It adds `iterator_bounded_entity.go` and rewrites the ascending log-ID leaf in `compile.go`, and it touches `compile.go`, `iterator_pebble.go`, `query-pipeline.md`, `iterator-seek-contract.md` and `read-path/README.md` — every one of which this PR also touches. Building a descending half here would duplicate the abstraction #1922 is introducing, conflict in five shared files, and put a perf refactor of the *ascending* leaf inside a PR whose whole safety argument is that ascending is the untouched oracle. So: doc corrected and the exception tracked here, iterator generalized by direction on top of #1922.

Validation: `scripts/agent-check` PASS; `internal/application/ctrl`, `internal/query` and `internal/storage/readstore` all `ok` at `74cb3c94`. `agent-check-pr` against `658dfaca` shows its usual red on `tests/antithesis/TestRunModelTestRequiresVerifiedOutcome` (`context deadline exceeded` under full-suite load); it passes in isolation in 7.1s, and this time the innocence is structural rather than circumstantial — the only code I added is a `_test.go` in package `ctrl`, which Go compiles solely into that package's own test binary, and `go list -deps -test ./tests/antithesis/` has no path to `internal/application/ctrl` regardless.

I also corrected the two stale PR-body claims Shipfox flagged (the body understated the compiler duplication and undercounted the parity matrix at 12 shapes when 48 shipped).

paul-nicolas:
Hi @gfyrag @Azorlogh @flemzord, could you review this PR when you have a moment? Thanks!

gfyrag (dismissed):
Approved: no blocking correctness issue found in the reviewed descending compilation/pagination paths. Local race tests passed for internal/storage/readstore, internal/query/... and internal/application/ctrl. The local single-leaf benchmark reproduces the reduction from draining all matches to page-sized traversal.

Important follow-up: converge on ONE direction-parameterized compiler using the foundations in #1955. The parallel compileRev recursion should remain an intermediate step, not become the permanent architecture. Shared leaf helpers alone do not prevent drift in dispatch, depth/target guards, boolean composition, error handling, cleanup and profile-tree construction.

Please make this an explicit tracked follow-up under EN-1966: one shared recursion with direction-specific leaf construction, preserving the ascending/descending parity tests, error-parity and absolute-seek regressions, pinned visibility checks and performance benchmarks from this PR. Those tests should be the acceptance criteria for removing the duplicated recursion. This is an architectural follow-up, not a newly demonstrated correctness blocker for this PR.

gfyrag (dismissed):

gfyrag (approved):

PR #1931, opened 2026-09-08
fix(EN-1956): flush index promotions before serving them

Azorlogh:
Replaces #1898, whose revision/incarnation gate solved the wrong problem (see the closing comment there).

## What changed

Two commits.

**1. Flush before serve (`36bc59df8`).** A promotion of a metadata or builtin index version (`current ← pending`, at the end of a backfill or a schema rewrite) is flushed to stable storage before it is served. The builder writes the promoted state into a batch, marks the promoted version in flight in the read store immediately before committing it; once the batch has committed, `finishPromotion` publishes the state to the builder's cache and calls `FlushPromotions` on the store, which flushes Pebble and releases the marks it covered. The mark's lifecycle follows the commit, not the flush: a batch that is cancelled or fails to commit withdraws its mark, while a flush that fails after the commit keeps it and the flush is re-attempted at the top of every builder tick. A tick with no mark pending triggers no flush.

**2. Serve the replaced version meanwhile (`6bd6984f3`).** A rewrite switch retains the version it replaces — `IndexVersionState.Previous*`: version, binding, activation, and the fold cursor at the switch as the last log its keyspace received — and leaves its keyspace on disk instead of GCing it in the promotion batch. `PinnedVersionResolver` serves that version, under its own binding, to any read the promoted version cannot answer yet: the promotion is still in flight, or the read is pinned below its activation. Marks are keyed by version, so a snapshot taken before the commit keeps serving the standing version. A creation promotion retains nothing and reads as `INDEX_BUILDING` while in flight, as it did throughout its backfill. The replaced version is retired by the builder (`retirePrevious`, every tick and after each promotion) once the promotion is flushed and no live read lease sits in `[PreviousActivationSequence, ActivationSequence)` — the only positions the resolver can still serve from it — in one batch that zeroes `Previous*` and range-deletes the keyspace. A second switch on the same index waits for that retirement (`promotionBlocked`), so at most one version is retained per index; `bumpPendingVersion` keeps `Previous*` across a retype that lands during retention; `purgeOrphanVersions` treats a retained version as live and the first loop wake after a restart retires it.

Docs: the indexer, indexes, query-pipeline and typed-metadata pages describe the mark-to-flush window, the retained version and its retirement; the API comparison table's `INDEX_BUILDING` row is corrected from `FAILED_PRECONDITION` to `UNAVAILABLE`, the code `KindForReason` has always mapped it to.

## Why

The read store is a WAL-less Pebble database: a hard kill rewinds it to its last flush. A promotion committed after that flush is undone by the kill, and the node reopens serving the superseded binding — at states where it had already answered queries under the promoted one (the Antithesis model driver caught this as a retype-window divergence). Transitions the fold performs inline are not exposed to this: the rewound fold re-walks their logs before alignment lets a query through. A promotion is, because the backfill or rewrite that completes it runs beside the fold and is only redone after the fold has caught up. EN-1956.

DISCOVERY: crash-recovery / read-projection regression (Antithesis model driver, Tests-Model)
BEFORE_FIX: BUG_REPRODUCED — `TestCompleteBackfill_PromotionSurvivesAKill`, `TestScanCompleteSwitch_PromotionSurvivesAKill` and `TestSchemaRewriteImmediateSwitch_PromotionSurvivesAKill` fail with the flush removed (the on-disk image reopens at the pre-promotion state)
AFTER_FIX: PASS

Need: a node must never answer a query under an index type binding it had already superseded — the served binding is monotone across crashes.
Current limitation: the promotion is a plain Pebble commit in a WAL-less store; any kill before the next memtable flush (64 MB of fold writes, or a graceful close) takes it back.
Requirement / constraint: a promotion is servable on a node only once it is durable there. Meanwhile a retyped index keeps answering from the version it replaced under that version's binding — what it served for the whole rewrite — and an initial build reads as `INDEX_BUILDING`, as it did throughout its backfill. After an index has served once on a node it does not return `INDEX_BUILDING` again.
Evidence: Antithesis model-driver retype-window divergence; `TestServingTransition_UnflushedPromotionDoesNotSurviveAKill` pins the premise (a committed, unflushed promotion is absent from the reopened image).
Durable repository evidence: `docs/technical/architecture/subsystems/read-path/typed-metadata.md` (Index versioning and schema rewrite), `docs/technical/architecture/subsystems/indexer/indexer.md` (`completeBackfill` — the switch; recovery table), `internal/storage/readstore/serving_transitions.go` contract comment.

## Technical decision

Decision: flush-before-serve with an in-process hold-off, and the replaced version retained and served through the hold-off. Marks are in-memory only; a reopened store holds exactly what was flushed, so nothing is in flight after a restart, and a version still retained at boot is retired on the first wake.

Why now / why proportionate: promotions are the only serving transitions not ordered by the fold cursor, and they happen once per index creation or retype, so the cost is one memtable flush per promotion and nothing on ordinary fold ticks. Retention costs one extra keyspace per retyped index for the flush duration plus the life of any read pinned between the two activations; the retirement gate reuses the read-lease registry the event GC already maintains.

Alternatives considered: (1) a revision/incarnation gate on the served binding (#1898) — the wrong layer: revision distance measures nothing about monotonicity (two back-to-back retypes give distance 2 with nothing lost), and a depth-1 rewind after a completed switch is a regression it cannot see; (2) enabling the WAL on the read store — pays a log write for every fold row to protect one record; (3) flushing the drop tombstone and the born-empty initial version too — unnecessary, they are fold-inline and alignment re-orders them; (4) flushing after every fold commit — hot-path cost for nothing; (5) refusing the index as `INDEX_BUILDING` for the mark-to-flush window (this PR's first head) — deterministic on every retype, for one 64 MB memtable flush, on an index that otherwise never returns `INDEX_BUILDING` after its first serve; (6) a resolver-only fallback to the old version number — insufficient while the promotion batch GCs the old keyspace, which is why the GC moved to retirement.

## Risk

**MEDIUM** — the per-replica `IndexVersionState` layout grows (22 → 43 bytes; v3 is unreleased, no compatibility path) and the old version's GC moves from the promotion batch to a later retirement batch, gated on the flush and on read leases. Confined to the three promotion sites, the version resolver, the builder's retirement pass and `InspectIndex`. The flush blocks the builder goroutine for the memtable flush duration once per promotion; a read pinned between two activations, on any ledger of the bucket, delays that index's retirement and its next switch for the read's lifetime.

## Validation

- [x] Targeted tests: kill-image tests for the three promotion paths reopen a Pebble checkpoint of the read store taken right after the promotion (no WAL, so no memtable — the image a hard kill leaves) and require the promotion to be there; premise test (unflushed promotion lost); in-flight reads as building, flushed serves; cancelled mark withdrawn; no flush when nothing is marked; `InspectIndex` refuses an in-flight promotion and serves after the flush; the mark is up when the promotion's commit runs, stays up after it, and is withdrawn by a failed commit (`TestBuilder_CommitPromotion_MarksBeforeTheCommit`, on the single helper every promotion path commits through)
- [x] Flush-failure path on all three promotion paths (injected flush error, since Pebble retries a failed memtable flush internally and never returns one): the promotion commits and is cached, the mark survives, the index stays refused, and `flushPromotions` lifts the mark once the flush succeeds
- [x] Red-verified: the three kill-image tests fail with the `FlushPromotions` calls removed from the promotion paths; the ordering test fails with the mark moved after the commit inside `commitPromotion`
- [x] Retained version: served under its own binding while in flight and to pins below the activation; refused past the last log it received, below its own activation, when itself in flight, and to a pin-less read; a snapshot predating the commit keeps serving the standing version (`readstore` tests); retain-then-retire under a lease, lease gate = the servable window `[PreviousActivation, Activation)`, second switch deferred while a version is retained, retirement at boot, `bumpPendingVersion` keeps `Previous*`, retirement leaves the pending slot alone (`indexbuilder` tests); the kill images of both switch paths still retain `v_old` since retirement is not flushed
- [x] Red-verified: 6 retention tests fail with the resolver's fallback disabled; the `bumpPendingVersion` test fails with the field-by-field rebuild restored
- [x] Full suite / broader validation: unit suites of `readstore`, `indexbuilder`, `ctrl`, `query`, `bootstrap` green; the kill-image tests stable over 25 consecutive runs; `just generate tidy lint` clean; the repository's codex review loop could not run (codex quota exhausted until 2026-10-04) — the head was reviewed with an independent Claude code-review pass instead, whose findings (registry-wide lease gate, pin-0 fallback, `bumpPendingVersion`, stale docs) are folded in

- Read store: `IndexVersionState` gains `PreviousVersion` / `PreviousType` / `PreviousTypeDeclared` / `PreviousActivationSequence` / `PreviousValidThrough`; an unexported tracker of promotions committed but not yet flushed, keyed by version, owned by `Store`, which exposes only `MarkPromotion` / `UnmarkPromotion` / `FlushPromotions` / `PromotionInFlight`. The flush the failure path needs is supplied at construction (`WithPromotionFlushForTest`), so nothing mutates a live store; `PinnedVersionResolver` and `SnapshotVersionResolver` are `Store` methods so the resolver can consult the tracker and fall back to the retained version (InspectIndex resolves through the pinned resolver on its aligned snapshot). `CreateCheckpoint`'s flush goes through the tracker (`TestServingTransition_CheckpointFlushLiftsTheMark`). `LeaseRegistry.AnyLiveIn` answers the retirement gate.
- Indexer: `writePromotion` / `commitPromotion` (mark, then commit; a failed commit withdraws the mark) / `discardBatch` / `finishPromotion` around the three promotion writes; `promotedState` retains the replaced version on the two rewrite switches; `flushPromotions` (every wake and after each promotion) flushes and runs `retirePrevious`; `promotionBlocked` defers a switch while an earlier promotion is unflushed or retained; `bumpPendingVersion` starts from the prior state; `purgeOrphanVersions` treats a retained version as live.
- Reads: a retyped index whose promotion is committed but not yet flushed is served from the version it replaced, under that version's binding, as it was during the rewrite; an initial build reads as `INDEX_BUILDING` until the flush succeeds.

## Review focus

1. The ordering argument in `PinnedVersionResolver`'s comment: the state is read through the caller's snapshot, the mark from the live store — why no interleaving lets an unflushed promotion through, and why the fallback to the retained version is bounded (`PreviousActivationSequence ≤ pin ≤ PreviousValidThrough`, retained version not itself in flight, no pin-less fallback).
2. The retirement gate in `retirePrevious`: why `[PreviousActivationSequence, ActivationSequence)` is exactly the set of leases that could still be served the retained version, and why a read that could need it is always registered before the switch (`OpenQueryHandle` reserves at the fold cursor before opening the main-store handle).
3. Retirement atomicity: `Previous*` zeroed and the keyspace range-deleted in one batch; not flushed, redone by the first wake after a kill; a drop or ledger delete purges the keyspace and evicts the state, so nothing can resurrect.
4. The mark lifecycle: withdrawn only when the batch did not commit (`discardBatch` in the cancel paths), kept across a failed flush (the replaced version keeps serving, `flushPromotions` retries); `FlushPromotions` clears only the marks captured before the flush.
5. The claim that fold-inline transitions (born-empty initial version, drop tombstone) need no flush — typed-metadata.md, "Index versioning and schema rewrite".

## Known concerns

A hard kill cannot be produced deterministically in the in-process e2e harness (a graceful stop flushes), so the kill image is produced at unit level with a Pebble checkpoint. Read leases are bucket-wide, not per ledger, so a long-lived read on an unrelated ledger whose position falls between two activations of a retyped index delays that index's retirement and its next switch for the read's lifetime; the deferred task's progress log carries `scanComplete: true` while it waits. The pre-existing alignment race (a read that stalls in `WaitForRaftProgress` while the fold leaps past an activation) is narrowed by retention but not closed for a read that arrives after retirement — separate ticket. Three independent bugs surfaced by the #1898 review are not part of this fix and remain open: `InspectIndex` takes an unaligned read-store snapshot; HTTP and CLI render inspected values under the live declared type rather than the served binding's type during a retype window; a follower's local `INDEX_BUILDING` refusal is not forwarded to the leader.


NumaryBot (commented):

Schema-rewrite flush failures remove the safety mark despite the promotion already being committed. This reopens the durability regression the patch is intended to prevent.

### Findings outside the diff

🔴 **[blocker] Keep the mark when post-commit flush fails** — `internal/application/indexbuilder/backfill.go:928`

When `flushWriteBatch` commits successfully but `FlushServingTransitions` fails, this returns before setting `committed = true`; the deferred cleanup then calls `abandonPromotion`, allowing readers to serve the committed but non-durable promotion. A subsequent hard kill can therefore rewind a binding that was already served, contradicting the failed-flush contract in [indexer.md:255](docs/technical/architecture/subsystems/indexer/indexer.md#L255). Track commit and flush success separately so only a failed/cancelled commit withdraws the mark.




Summary: https://github.com/formancehq/ledger/pull/1931#issuecomment-5583047632

Azorlogh:
Addressed in c913aabfaaed0a9b94143afeeabd17cc899d9ca8 (re: NumaryBot's [blocker](https://github.com/formancehq/ledger/pull/1931#issuecomment-5583047632) and shipfox's [Spec 1](https://github.com/formancehq/ledger/pull/1931#issuecomment-5583184229)). Confirmed: on the two schema-rewrite paths a flush failure after a successful commit returned before `committed = true`, so the deferred cleanup withdrew the mark.

The flush no longer lives inside `flushWriteBatch`. Each promotion site now sets `committed`, then calls `finishPromotion`, which publishes the state to the cache and flushes; a flush failure is logged and keeps the mark, and `retryServingFlush` re-attempts it at the top of every builder tick. `abandonPromotion` runs only when the batch did not commit. Pinned on all three promotion paths with an injected flush error (`OverrideServingFlushForTest` — Pebble retries a failed memtable flush internally, so a real one cannot be produced): the promotion is committed and cached, the mark survives, the index stays refused, and the retry lifts the mark. Docs (indexer.md, typed-metadata.md) now describe that order.

Azorlogh:
Re shipfox's [second review](https://github.com/formancehq/ledger/pull/1931#issuecomment-5583455923): item 2 taken in 6c3e2a045168ad7d3e747ec526bbef2cc9cd01da — `tryCommitScanCompleteSwitch` now guards its teardown with the same `staged` flag as `processSchemaRewrite`, so the deferred cleanup can only withdraw the mark that call staged. Item 1 (Store methods in `serving_transitions.go`) is left as is: the file carries the mechanism's contract comment, and the package already keeps `Store` methods per concern.

Azorlogh:
Re shipfox's [review of c4aa86222](https://github.com/formancehq/ledger/pull/1931#issuecomment-5584596293): both taken in addc4e331 — indexer.md now names `discardPromotion` (the doc kept the pre-rename symbol), and the `_ = batch.Cancel()` in `discardPromotion` carries its justification inline. The PR body's mechanism section already describes the `ServingTransitions` tracker and `discardPromotion`; the Spec note quotes the pre-refactor names.

internal/storage/readstore/serving_transitions.go
```diff
@@ -0,0 +1,128 @@
...
+// until Flush clears it.
+func (t *ServingTransitions) Mark(ledgerName, canonicalID string) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	if t.inFlight == nil {
+		t.inFlight = make(map[servingTransitionKey]struct{}, 1)
+	}
+
+	t.inFlight[servingTransitionKey{ledgerName, canonicalID}] = struct{}{}
```
NumaryBot: 🔴 **[blocker] Preserve earlier marks when abandoning a later promotion**

If a flush fails and another promotion for the same index is staged before retry succeeds, `Mark` merely overwrites the existing set entry. If that later batch then fails or is cancelled, `Unmark` deletes the mark belonging to the already-committed, still-unflushed promotion, allowing readers to serve state that can disappear after a hard kill. Track promotion ownership/generations or prevent a failed later promotion from clearing a pre-existing mark.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1931#issuecomment-5583047632

Azorlogh:
Head 7ca7e0b7c48d38ad34ff881020c530603cef81ab, reviewed locally before pushing this time (independent reviewer run with the `scripts/ai-review-codex` rubric; the Codex account's quota is exhausted until 2026-10-04, so the repo loop cannot complete its review stage). Changes since addc4e331:

1. NumaryBot's [blocker](https://github.com/formancehq/ledger/pull/1931#discussion_r3957551492): marks are counted per index (`ServingTransitions.inFlight map[…]int`); `Unmark` releases one, `Flush` releases only the counts it captured, so withdrawing a later staged promotion leaves an earlier committed one gated. Pinned by `TestServingTransitions_WithdrawingALaterMarkKeepsTheEarlierOne`.
2. `indexer.md`'s recovery table now splits "crash after the switch committed" into before/after the flush completed, with the actual residual state and replay path for each; the paragraph above states the execution order (`writePromotion` → same-batch GC → `markPromotion` → commit → `finishPromotion`).
3. `ServingTransitions` documents its single-writer contract; the `Flush` release comment states the implemented rule.
4. On the rewrite paths the mark is now taken after the old version's GC, immediately before the commit, so readers keep serving the standing version while the batch is assembled; the refusal window is commit-to-flush.

Re shipfox's [two minor notes](https://github.com/formancehq/ledger/pull/1931#issuecomment-5584876646), declined: (a) the two rewrite sites' `committed` defer covers the whole scan batch and they commit through `flushWriteBatch` (event-zone epochs) while `completeBackfill` commits a direct batch, so a helper owning the commit would not fit; (b) `serving_transitions_test.go` is an internal `package readstore` test and cannot import a test-support package that imports `readstore` (import cycle), so the 8-line kill-image wrapper stays duplicated.

internal/storage/readstore/serving_transitions.go
Azorlogh: Fixed in 7ca7e0b7c48d38ad34ff881020c530603cef81ab: marks are now a per-index count; `Unmark` releases one and `Flush` releases only the counts it captured, so withdrawing a later staged promotion leaves an earlier committed one gated (`TestServingTransitions_WithdrawingALaterMarkKeepsTheEarlierOne`).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1931#issuecomment-5583047632

paul-nicolas (commented):

**Head reviewed:** `2fd2418a7`. Codex reviewed the full diff with repo access and returned **no findings**. Three Claude reviewers (mark/flush lifecycle, reader/serving side, tests) raised **7 P2 candidates**, each of which was then handed to an independent adversarial verifier instructed to refute it. **6 of 7 were downgraded to P3 on base comparison or reachability grounds; 1 survived.**

**Blocking findings: 0.** The mark/commit/flush ordering is correct on all three promotion paths, the fold-inline exemptions (born-empty initial version, drop tombstone) are justified — both write into the batch carrying the fold cursor, so a rewind takes them together — and the reader sweep found no ungated consumer: all five resolver call sites were converted to the `*Store` method, and no package-level `PinnedVersionResolver`/`SnapshotVersionResolver` remains.

Two premises were checked independently and hold:
- **`pebble.Checkpoint` does not capture the memtable.** `checkpoint.go:165` gates the only memtable-touching branch on `opt.flushWAL && !d.opts.DisableWAL`, and the read store hardcodes `DisableWAL: true` (`store.go:118`). The kill-image tests are genuine.
- **Red-verification reproduces.** Mutating `ServingTransitions.Flush` to release marks without flushing fails all three `*_PromotionSurvivesAKill` tests on their load-bearing assertions.

Validation run: `go test ./internal/storage/readstore/... ./internal/application/indexbuilder/... ./internal/application/ctrl/...` — all green at head.

### Why the other six were withdrawn

- **Status RPCs ungated** (raised by two reviewers, one as blocking): the same divergence already ships on base via the `ActivationSequence > pin` arm (`store.go:638`, unchanged, pinned by `version_activation_test.go:26-48`), so `current_version > 0` was never a serving guarantee. Every affected e2e follow-up is `Eventually`-wrapped. -> P3.
- **Unbounded `pebble.DB.Flush` on the builder goroutine**: real, but `Builder.Stop()` was already unkillable under disk stall via `process_logs.go:322`, `gc.go:218`, `gc.go:282`, `backfill.go:1286`, all of which park in `maybeInduceWriteStall` with no context. The PR shortens time-to-stall on an already-wedged disk; it does not add a failure class. -> P3.
- **Version-blind mark over-refusing pre-commit snapshots**: real, but confined to the `markPromotion`->commit sliver (no work between the two on any path) on the retype paths only; `completeBackfill` is unaffected because `requireIndexReady` already refuses `Version == 0`. `Unavailable`-coded and client-retried. -> P3.
- **`retryServingFlush` loop call uncovered**: `pebble.DB.Flush` cannot return a non-nil error against a live writable store (a closed DB panics in `AsyncFlush`; `makeRoomForWrite(nil)` cannot fail), so the retry is a permanent no-op guarding an unreachable state. -> P3.
- **`discardPromotion` unreached by any test** (proven with a panic probe plus a control probe on `markPromotion`): the failure self-heals within one tick, and the dangerous direction — over-unmarking a committed promotion — is prevented by construction. -> P3.

### P3 (not posted inline)

- `internal/application/indexbuilder/builder.go:236` -- `discardPromotion` silently stops resetting the shared `WriteBatch` after a failed commit: `WriteBatch.Flush` nils the binding before returning the error, so `b.wb.Batch() == batch` is `nil == batch` and `Reset()` is skipped; the pre-PR code reset unconditionally. No live exploit (`initBatch` recreates the overlay), but it removes the only unconditional teardown of the read-your-writes state.
- `internal/storage/readstore/serving_transitions.go:84` -- `release` clamps an over-release to zero instead of failing loudly, so a mark/unmark pairing bug degrades into exactly the EN-1956 defect rather than surfacing (invariant 7).
- `internal/storage/readstore/serving_transitions.go:32` -- the single-writer contract is prose only; `Store.ServingTransitions()` hands `Mark`/`Unmark`/`Flush` to every `*Store` holder, including the ctrl request path that needs only `InFlight`.
- `internal/storage/readstore/serving_transitions.go:140` -- `OverrideFlushForTest` is exported production API on the durability path with no build tag and no lint guard; a `.golangci.yaml` rule would match the repo's `OpenWriteSession` style.
- `internal/application/ctrl/controller_default.go:1258` -- `GetIndexStatus` and `GetIndexEntryStatus` (`:1392`) report `current_version` the resolver may refuse; downgraded above, but worth a comment or a gate so the omission reads as deliberate.
- `internal/storage/readstore/store.go:634` -- version-blind mark over-refuses a pre-commit snapshot whose `v_old` keyspace is intact (the GC rides the same batch).
- `internal/application/indexbuilder/builder.go:252` -- promotion flush is uncancellable and unbounded on the builder goroutine.
- `docs/technical/architecture/subsystems/indexer/indexer.md:255` -- "the refusal window is commit-to-flush" is imprecise: it is mark-to-flush, and the mark-to-commit slice refuses reads that were servable. Same line: "run at the top of every builder tick" — `retryServingFlush` is at `builder.go:790`, inside the steady-state loop only; the pre-loop `processBackgroundTasks` (`builder.go:775`) can complete a promotion before any retry exists.
- `docs/technical/architecture/subsystems/indexer/indexer.md:314` -- the new recovery row asserts one of two possible residual states as the only one; Pebble may have rotated the memtable on its own, leaving the promotion durable. Both are safe, but the repo rule asks for the actual residual state.
- `docs/technical/architecture/subsystems/read-path/query-pipeline.md:114` -- "will keep serving `v_current` until the atomic switch lands" is now false for the mark-to-flush window.
- `docs/technical/architecture/subsystems/indexer/indexes.md:27,:35,:79,:173` -- the page reached from `indexer/README.md` still carries the pre-PR model; `:173` ("gates the inspect call on `state.CurrentVersion != 0`") is directly contradicted by this PR's own `controller_default.go:1023`.
- `docs/technical/contributing/api-comparison.md:869` -- maps `INDEX_BUILDING` to `FAILED_PRECONDITION`; the actual mapping is `KindUnavailable` -> `codes.Unavailable` / HTTP 503 + `Retry-After`. Pre-existing drift, but this PR is what makes the row consequential — a client author reading it would not retry.
- `internal/application/indexbuilder/builder.go:264` -- no metric for a pending serving transition; a stuck flush is a log line only.
- `internal/storage/readstore/serving_transitions_test.go:90` -- no end-to-end query-path test for the new refusal; only the resolver tuple and `InspectIndex` are covered.

internal/application/indexbuilder/backfill.go
```diff
@@ -1277,17 +1282,21 @@ func (b *Builder) completeBackfill(task *backfillTask) error {
...
 	batch := b.readStore.NewBatch()
-	if err := b.readStore.WriteIndexVersionState(batch, task.ledger, canonical, newState); err != nil {
-		_ = batch.Cancel()
+	if err := b.writePromotion(batch, task.ledger, canonical, newState); err != nil {
+		b.discardPromotion(batch, task.ledger, canonical, false)
 
 		return fmt.Errorf("persisting backfill atomic switch: %w", err)
 	}
 
+	b.markPromotion(task.ledger, canonical)
```
paul-nicolas: [P2][non-blocking] Nothing pins the mark-before-commit ordering the whole fix rests on

**Evidence.** Moving `markPromotion` to *after* `batch.Commit()` here — and correspondingly flipping the commit-failure `discardPromotion(..., true)` to `false`, which is what a cleanup refactor would naturally do — leaves the entire suite green:

```
ok  github.com/formancehq/ledger/v3/internal/application/indexbuilder  7.130s
ok  github.com/formancehq/ledger/v3/internal/storage/readstore        4.909s
ok  github.com/formancehq/ledger/v3/internal/application/ctrl         2.070s
```

**Why it matters.** `PinnedVersionResolver`'s own contract comment (`readstore/store.go:605-612`) states the premise: "one committed before it is either still in flight — refused — or flushed." That holds *only* because `Mark` precedes `Commit`. With the order swapped, a query whose snapshot is taken after `Commit` returns and whose `InFlight` check runs before `markPromotion` sees `CurrentVersion = pending, InFlight = false` and serves the promoted binding. A kill anywhere in the much longer commit->flush window then rewinds the read store to the pre-promotion image, and the node serves the superseded binding at sequences it already answered under the promoted one. That is EN-1956, reopened. `InspectIndex` (`controller_default.go:1023`) has the same exposure. The window is observable: `completeBackfill` runs on the builder goroutine while queries run on their own.

The existing tests cannot separate the two orderings at any point they inspect — success path: `InFlight` false after `finishPromotion` either way; flush-failure path: `InFlight` true either way; commit-failure path: correct code does Mark(+1)/Unmark(-1) = 0, mutated code does neither = 0, and `release` deletes on `<= n` so even a pre-seeded count cannot discriminate.

The same ordering exists at `:918` and `:1001`; this is one root cause, not three.

**Fix.** Assert that no committed-but-unmarked instant exists. The cheapest form: in the existing flush-failure tests, have the injected flush assert from inside itself that the promotion is already committed (`ReadIndexVersionState` shows the promoted `CurrentVersion`) **and** still `InFlight` — mark-after-commit fails that immediately.

One caveat on remediation: `b.readStore` is a concrete `*readstore.Store` and `NewBatch()` returns a concrete `*dal.WriteSession`, so there is no seam for a "assert `InFlight` at commit time" hook. A deterministic test needs a small production seam (a `beforeCommit func()` hook on `Builder`, or an interface around the batch). A goroutine-racing probe would be non-deterministic and should not be accepted as the fix.

Azorlogh: Confirmed: with the mark moved after the commit at all three sites the three suites stayed green. Addressed in b7b79ed50 (https://github.com/formancehq/ledger/pull/1931/commits/b7b79ed5046aff5b4de4d538e451c8eb9ea04f64) — the three sites now commit through one helper, `commitPromotion(ledger, canonical, commit func() error)`, which marks, runs the commit, and withdraws the mark on failure; `TestBuilder_CommitPromotion_MarksBeforeTheCommit` passes its own commit closure and asserts the index is in flight when it runs, so the ordering is pinned deterministically without a production hook (swapping the order inside the helper fails it). The flush-time assertion you suggested would not have separated mark-after-commit from mark-before-flush, which is why the seam is the helper rather than the flush. The doc P3s (refusal window, retry tick, recovery residual state, inspect gate, rewrite reads, `INDEX_BUILDING` → `UNAVAILABLE`) are fixed in the same commit.

gfyrag:
Review checkpoint on HEAD `fd7bfd7f`: I agree that EN-1946's Raft-index alignment does **not** make this change redundant. A promotion is node-local side work outside the fold cursor; an unflushed promotion can disappear on a hard kill while the durable read certificate remains high enough for alignment to pass. Some flush-before-serve mechanism is still required.

The remaining merge gate is integration rather than the core algorithm: GitHub currently reports this PR `DIRTY` against base `8c33007e`, while the EN-1946 stack has changed the surrounding read/index lifecycle. Please rebase onto the current `release/v3.0` and rerun the promotion/reopen tests against that exact base before sign-off.

One design challenge for the rebase: the narrow durability rule currently exports `Store.ServingTransitions()` and `OverrideFlushForTest`, exposing lifecycle state and a production mutation seam. Could the store instead expose only the domain operations needed by the builder/resolver, with flush failure injected at construction or through a package-private test seam? The per-index fail-closed design is reasonable; reducing the capability surface would make it harder for a future caller to bypass or manipulate its lifecycle.

Azorlogh:
Rebased onto current `release/v3.0` and did the encapsulation, both in 36bc59df8.

`Store` now exposes only `MarkPromotion`, `UnmarkPromotion`, `FlushPromotions` and `PromotionInFlight`; the tracker type is unexported and no longer handed out. `OverrideFlushForTest` is gone — the flush the failure path needs is supplied at construction (`WithPromotionFlushForTest`, nil ignored), so nothing mutates a live store. Readers reach the marks through the resolvers as before.

Verified: the promotion suites, `ctrl`, `query` and `bootstrap` green; `just lint` and agent-check clean; the three kill-image tests still red with the promotion flush removed. An independent review found no correctness defect and six hygiene items, all applied — orphaned godoc on `New`, the test-only name and nil guard above, a note on `CreateCheckpoint` under a wrapped flush, `retryServingFlush` renamed `retryPromotionFlush`, and two stale doc/comment references.

One rebase note: the only conflict was the two-row `INDEX_BUILDING` / `INDEX_INCONSISTENT` table in `api-comparison.md`; I kept the base's metadata column and this PR's corrected code mapping (`UNAVAILABLE`).

internal/application/indexbuilder/backfill.go
Azorlogh: Still fixed at the current head `36bc59df8`: the three sites commit through `commitPromotion`, and `TestBuilder_CommitPromotion_MarksBeforeTheCommit` asserts the index is in flight from inside the commit closure — swapping the order there fails it.

Azorlogh:
Fair — PR body updated to the shipped API: `MarkPromotion` / `UnmarkPromotion` / `FlushPromotions` / `PromotionInFlight` on `Store` with the tracker unexported, `WithPromotionFlushForTest` in place of the old setter, and `discardBatch` / `retryPromotionFlush` in the lifecycle bullets.

gfyrag:
@Azorlogh Design challenge on head `36bc59df8`: flush-before-serving the **new** version is necessary, but does it require refusing reads that could still use the **old** version?

The current mark-to-flush gate returns `INDEX_BUILDING` for the whole index, including a reader whose snapshot predates the promotion. That over-refusal is documented, but documentation alone does not justify the availability trade-off. The builder also waits synchronously for a Pebble memtable flush, so the impact is not necessarily limited to the one promoted index: fold progress can stall and aligned reads can wait.

Could you compare the current design with retaining the old servable binding while the replacement becomes durable, then publishing the replacement and retiring the old version? I am not suggesting a resolver-only fallback: the rewrite promotion batch also carries old-version GC, and retaining a version number is insufficient unless its rows, type binding and validity at the requested horizon remain available.

Please make the comparison concrete:
- Can existing versioning/read leases support this, or which exact invariant prevents it? Distinguish first creation (no old version) from retype/replacement.
- Spell out commit / flush / serving-publication / GC ordering and crash recovery. A naive two-phase switch can merely move the non-durable-publication problem; show why a version already served cannot regress after restart.
- Account for writes arriving during the transition, pinned snapshots, successive retypes, and failed or stalled flushes. Serving an old binding must still return correct rows at the requested horizon, not stale data.
- Quantify the current trade-off with representative memtable sizes and concurrent load: flush duration, index refusals and alignment latency. “Once per promotion” does not bound the outage duration.

A deterministic barrier test should hold the flush open while reads run and pin the chosen contract; include failure/retry and kill-image recovery.

This is a design/availability challenge, not a newly demonstrated correctness defect or a request to implement a more complex alternative blindly. If keeping the old version genuinely requires disproportionate machinery, please explain that constraint and the evidence that the simpler fail-closed window is acceptable.

Azorlogh:
@gfyrag Agreed, and done on head `6bd6984f3` as a separate commit: a rewrite switch retains the replaced version (`IndexVersionState.Previous*`: version, binding, activation, fold cursor at the switch) and leaves its keyspace on disk; the resolver serves it — under its own binding — to any read the promoted version cannot answer yet (unflushed, or pinned below its activation), so a retyped index no longer returns `INDEX_BUILDING`. Marks are keyed by version, so a snapshot predating the commit keeps serving the standing version. Retirement (`retirePrevious`, one batch: zero `Previous*` + GC) runs once the promotion is flushed and no live read lease sits in `[PreviousActivationSequence, ActivationSequence)`; a second switch on the index waits for it. First creation retains nothing and refuses as before — it never served.

Your points, briefly: existing leases do support it (`OpenQueryHandle` reserves at the fold cursor before the main-store handle, so a read that could need the old version is registered before the switch); the ordering is switch-with-retention → mark → commit → flush → serve new → retire, and every kill point is in the PR body's recovery table; successive retypes are serialised by the retirement guard (at most one retained version); a stalled flush now keeps serving the old binding instead of refusing; writes never target the retained version. The resolver-only fallback you ruled out was indeed insufficient — the GC sat in the promotion batch. Barrier/lease/kill-image tests are in `serving_transition(s)_test.go`; the codex review loop could not run (quota), an independent Claude review pass was used instead.

Azorlogh:
Both fixed in 1a9dde46c: `commitRetirement` now returns an `invariant:` error for a canonical that does not parse or names a non-metadata index (logged by the pass, test added), and discards the write session on a failed commit.

paul-nicolas (approved):

PR #1939, opened 2026-09-08
fix(EN-1636): rebuild typed errors on the leader-forwarding hop

paul-nicolas:
## Problem

On the REST surface, a business error raised by the FSM is returned as
`500 INTERNAL_ERROR` (with a correlation ID and a sanitized message) instead of
its correct status whenever the HTTP request is served by a node that is not the
Raft leader.

Reproduced in the cloud (stack `reslene-0fa5ae0`) with
`POST /api/ledger/v3/{ledger}/indexes` for a metadata field never declared via
`SetMetadataFieldType`:

```
HTTP/2 500
{"errorCode":"INTERNAL_ERROR","errorMessage":"internal server error (correlation ID: b37c1a74098abd21a956fd6ec4fc5231)"}
```

while the leader logged the correct outcome:

```
ERROR gRPC call failed {"code": "FailedPrecondition",
 "error": "rpc error: code = FailedPrecondition desc = metadata field not declared in schema: TARGET_TYPE_ACCOUNT/type"}
```

This is not an error-classification bug. The mapping is already correct:
`ErrMetadataFieldNotInSchema` -> `KindPrecondition` -> `codes.FailedPrecondition`
on gRPC and `400` on HTTP. The 500 comes purely from the transport boundary: the
leader serialises the error faithfully (status code + `errdetails.ErrorInfo`),
but a `*status.Error` is not a `domain.Describable`, and both `handleError` and
the bulk per-element mapper dispatch on that contract. Every forwarded business
error fell through to the sanitizer.

Roughly (N-1)/N of requests were affected on a multi-node cluster, depending on
which node the load balancer picked. Single-node deployments were unaffected.
The gRPC surface was unaffected too: `convertToGRPCError` returns an
already-status error as-is, so the code and its `ErrorInfo` survived the hop.

Not caught by the existing suite because no e2e test anywhere combined a
multi-node cluster with HTTP: every HTTP e2e suite is single-node, and every
multi-node write goes over gRPC asserting `Succeed()`.

## Fix

Decode once, at the seam that erased the identity, rather than adding
status-code special cases to `handleError`.

`grpcerr.NewConn` decorates the leader connection at `getLeaderCtrl` — the only
production site that builds a leader-forwarding client. `Invoke` converts unary
errors; `NewStream` returns a stream whose `RecvMsg` converts.

Wrapping the connection rather than the generated client's 37 methods (11 of
them server-streaming) is deliberate. Six of the `BucketGrpcClient` list
methods return a lazy cursor — `ListLedgers`, `ListTransactions`,
`ListAccounts`, `ListLogs`, `ListAuditEntries`, `ListIndexes` — and surface the
leader's error from a later `Recv()`, after the method itself already returned
`nil`, so per-method conversion would convert nothing there; and every method
generated in future routes through `Invoke` or `NewStream` regardless. **Bulk
is repaired by the same change**, because both HTTP mappers read the failure
through the same boundary contract.

`grpcerr` is a leaf package (gRPC, errdetails, `apierr`, `domain`, `commonpb`
only). `internal/adapter/grpc`, which the ticket suggested, transitively pulls
in the Pebble DAL, readstore, usagestore, backup and the Raft node —
`cmd/ledgerctl` must not link that. So the single `ErrorKind` → status-code
table (`CodeForKind`) lives here and `internal/adapter/grpc` imports it, which
also removes the drift risk between the encoder and the decoder that reverses
it. `BusinessErrorFromGRPC` moves here from `cmd/ledgerctl/cmdutil` as `Decode`
so the CLI and the forwarder share one implementation; the CLI gains
reason-first kinds as a side effect.

### Aligned with EN-1980

The first revision of this PR decoded into `domain.RemoteError` and reached its
wire-derived kind through a private `domain.Kind` escape hatch
(`kindOverride`). @gfyrag held the merge on that: a transport-boundary
representation does not belong in `internal/domain`, and `domain.Kind` should
not carry an override that exists for one such type. Both are now relocated.
The decoding seam, the preserved gRPC status and the reason-first derivation
are unchanged — the review's correction that **preserving the original status
is necessary** is what the whole design rests on.

**`internal/adapter/apierr` is the boundary contract.** `Describe(err)` returns
a `Descriptor` — semantic kind, stable reason, client-safe message, structured
metadata — and normalises the two provenances a failure can have:

| Provenance | Classification |
|---|---|
| Raised locally (`domain.Describable` from admission, the FSM, a read path) | `domain.Kind(d)`, a pure function of the reason |
| Decoded from a peer (`*apierr.Remote`, produced only by `grpcerr`) | the kind the wire carried |

`handleError`, the bulk per-element mapper and `ledgerctl` all read through it.
The `*apierr.Remote` branch is checked first: falling through to the
`Describable` branch would re-derive the kind from the reason, which for a
reason a newer server sent yields `KindInternal` and answers 500 for a caller
error. `apierr` imports `internal/domain` and nothing else, so HTTP is coupled
to no gRPC detail, and there is still exactly one table per direction —
`kindToHTTPStatus` on REST, `grpcerr.CodeForKind` on gRPC.

**`domain.Kind` is now strictly reason-based**, and the audit exhaustiveness
table no longer carries a row for a type with no production path to the audit
chain. The separation is stated as a dependency rule rather than claimed as a
type-system guarantee: `scripts/check-repo-invariants` rejects an `apierr`
import from `internal/domain`, `internal/application/admission`,
`internal/infra/state`, `internal/infra/plan` or `internal/infra/preload`.
Proven by mutation — a probe file importing `apierr` under
`internal/infra/state/` makes the real repository run FAIL with that message.

**Two axes, deliberately separate.** Semantic kind answers the client; the gRPC
code is transport behaviour. `READ_INDEX_NOT_CAUGHT_UP` is semantically
`KindUnavailable` (503 + `Retry-After` on REST) while travelling as
`codes.FailedPrecondition`, because `actions.GRPCRetryPolicy` retries
`codes.Unavailable` 50× at 0.2s and the semantic code would turn a read-index
lag into a ~10s client-side hang. The carrier keeps `GRPCStatus()` intact, so
both hold at once and on the next hop.

**New behaviour: a reason-specific allowed-wire-code policy.** The allowed set
for a known reason is `CodeForKind(KindForReason(reason))` plus any declared
exception — today only the `READ_INDEX_NOT_CAUGHT_UP` pair above. It is
reason-specific rather than kind-specific for exactly that reason: deriving it
from `CodeForKind` alone would reject the pair the server actually sends. A
known reason arriving outside its allowed set is a protocol fault, not a
business outcome: it decodes to an `*apierr.InvalidWireError`, which implements
neither the boundary contract nor `GRPCStatus`, so it reaches the internal-error
sanitizer on every surface (500 + correlation ID + server-side log on REST,
`codes.Unknown` + correlation ID on gRPC) and the untrusted reason, message and
metadata are never echoed. Without dropping the status, a contradicting
`codes.InvalidArgument` pair would be answered to the caller as a trusted 400
carrying the peer's own message. An **unknown** reason cannot be validated —
this build has no policy for it — so its reason, message, metadata and exact
code are preserved verbatim, and the code still supplies the classification
where it has one (unknown + `AlreadyExists` stays 409, not 500).

**Kind derivation is reason-first.** `KindForReason(ReasonCode(reason))` when
this build's enum knows the reason, falling back to the status code otherwise.
Neither source suffices alone:

- `CodeForKind` sends both `KindConflict` and `KindPrecondition` as
  `codes.FailedPrecondition`, so **code-only** derivation answers 400 for a
  `KindConflict` reason (`LEDGER_DELETED` and five others) where the leader
  answers 409;
- **reason-only** derivation collapses a reason from a newer server to
  `KindInternal`, turning a caller mistake into a 500.

**Statuses left deliberately raw.** Read the rule as the complement of the two
reconstructed shapes rather than as a list of excluded codes: everything that
is neither a ledger-domain `ErrorInfo` nor a bare `codes.NotFound` passes
through. `codes.Canceled` is the one unconditional passthrough — `cursor.go`
normalises it to `io.EOF` to end a page, so converting it breaks pagination.
A **bare** `Unavailable` already reaches 503 + `Retry-After` through
`handleError`'s own branch and carries no reason to recover; bare `Internal`,
`Unknown` and `DeadlineExceeded` are server faults; and a bare
`Unauthenticated`/`PermissionDenied` from the leader denotes a cluster-secret
failure *between nodes*, not a caller credential problem, so it correctly stays
a 500. Note the qualifier: `Unavailable` and `Internal` are exactly the codes a
`KindUnavailable`/`KindInternal` `Describable` is *sent* under, so those arrive
**with** an `ErrorInfo` and are reconstructed like any other — `INDEX_BUILDING`
and `COVERAGE_MISS` keep their reason across the hop. The first revision's
documentation of this bullet was wrong and is corrected.

## Notes for the reviewer

**The 13 `fmt.Errorf` prefixes are message hygiene, not the cause.** 21 of 32
client methods returned the leader's status unwrapped and lost the typed
identity just as thoroughly. But `status.FromError` rewrites the status message
from `err.Error()`, so a prefix leaked into the client-visible `errorMessage` on
the one path that already worked (`InvalidArgument`).

**Four premises from the ticket had moved** since it was written, and are worth
knowing while reviewing:

| Ticket said | Actual on `release/v3.0` |
|---|---|
| Bare `Unavailable` survives the fix at 500, no `Retry-After` | Already fixed; `handleError` has that branch. Remaining bare-status gap was `NotFound`, which this converts |
| Promote `READ_INDEX_NOT_CAUGHT_UP` to an enum reason | Already done — enum entry, `KindUnavailable`, and `ErrReadIndexNotCaughtUp` is a `Describable`. No proto change needed |
| 11 wrap prefixes | 13 |
| Five `KindConflict` reasons collapse | Six (`STALE_CLUSTER_POLICY` added). Note the three 409 cases the ticket lists under "why it matters" are `KindAlreadyExists` → `codes.AlreadyExists`, a collision-free code that round-trips correctly — using one as the 409 regression case gives a green test that proves nothing |

**One change I tried and reverted.** `convertToGRPCError` has a hand-written
branch emitting `codes.FailedPrecondition` for `query.ErrReadIndexNotCaughtUp`
even though its reason classifies as `KindUnavailable`, so it looks like dead
weight in front of the `Describable` branch. Deleting it regressed the
cross-store e2e specs for the 50× retry reason above. It also bought nothing —
the HTTP server is wired `NewDefaultBackend(node, ctrl)`, straight to the
controller, so that function is not on the REST path at all. Reverted;
`internal/adapter/grpc/server.go` is byte-identical to base. That pair is now
the declared wire-code exception, with the rationale written down in
`grpcerr.wireCodeExceptions` and pinned by tests on both axes.

**Scoped out:** `EXTERNAL_SERVICE_ERROR` (`server.go:681,699`) is a string
literal rather than an enum reason, but it is emitted only from smithy S3 errors
and the HTTP surface has no backup/restore/checkpoint/snapshot handlers, so it
is HTTP-unreachable and no leader/follower divergence exists. As a non-enum
reason it takes the unknown-reason path and is not validated, which is correct.
The aged #432 follow-up (`bulkErrorCode` returning `"ERROR"`) stays deferred for
the review isolation it was deferred for, though bulk is repaired for real
business reasons here.

**Base has advanced 8 commits** since the merge base (`b2b7e7d03`), including
`feat(grpc)!: require a matching service protocol revision (EN-1851)`. This
change does not touch the service contract — no `.proto` change, no RPC added
or removed, no wire-format change — so `pkg/grpcprotocol.Version` needs no
increment. The branch merges cleanly with `release/v3.0` (`git merge-tree`,
0 conflicts); a rebase is still worth doing before merge and is deliberately
not folded into this revision so the validation below stays attributable.

## Tests

- **Conversion table driven off the enum itself**: every `ErrorReason` this
  build knows, under every code its policy allows, must decode to
  `KindForReason`. One reason per `ErrorKind` is not enough coverage — kinds
  are not one-to-one with codes, so a sampled reason can pass while a sibling
  of the same kind decodes wrong. A second test walks every `KindConflict` and
  `KindPrecondition` reason across the shared `codes.FailedPrecondition` and
  asserts each lands on the right side, refusing to run unless at least two of
  each exist.
- **Unknown reason × six codes**, plus unknown reason × `codes.Aborted` (a code
  no `ErrorKind` maps to) asserting the exact code survives a second hop.
- **The wire-code exception** on both axes, at the unit level and over a real
  bufconn connection.
- **The mismatch policy**: rejected pair decodes to `*apierr.InvalidWireError`,
  is not a `Describable`, has no `GRPCStatus`, and neither its message nor its
  metadata appears in the error text — asserted at the decoder, over bufconn,
  on the HTTP surface (500 + correlation ID, nothing echoed) and on the CLI.
- **`apierr.Describe`** over local, wrapped-local, remote, wrapped-remote and
  unrecognised errors, including a branch-order test that fails if the
  `Describable` branch is checked first.
- **bufconn round trips** for the unary and stream paths against the real
  generated client, pinning that `status.Code` survives and that a clean stream
  end stays `io.EOF`.
- **HTTP table** asserting the client-visible status and `errorCode`, including
  the ticket's exact repro, the bare-auth passthrough, the 503 for the wire-code
  exception, and a bulk per-element table covering the second dispatch site.
- **The dependency rule**: positive and negative cases for
  `isBusinessCorePath`, the import finding itself, and that the boundary import
  is allowed outside the core.
- **`tests/e2e/cluster/rest_error_parity_test.go`** — the first e2e combining a
  multi-node cluster with HTTP — asserting follower/leader parity for a write, a
  bulk element and a conflict. The bulk spec now asserts `HaveLen(1)` and
  `INSUFFICIENT_FUNDS`, so a request-level rejection fails it instead of
  satisfying it (review thread).
- **ledgerctl** keeps structured formatting, with a new case for an unknown
  reason still being formatted as a business error.

Every new test was mutation-proven rather than merely observed green:

| Mutation | Result |
|---|---|
| validation never rejects (`false &&`) | exactly 4 tests fail — decoder, `FromStatusError`, bufconn, HTTP — and nothing else |
| delete the `READ_INDEX_NOT_CAUGHT_UP` wire-code exception | fails the 3 two-axes tests (unit, bufconn, HTTP 503) |
| swap `Describe`'s branch order | fails the 2 `apierr` branch tests plus the 2 "reason from a newer server" HTTP/bulk rows |
| import `apierr` from `internal/infra/state/` | `check-repo-invariants: FAIL` naming the file and line |
| make the bulk e2e body a request-level rejection | `the rejection must be per-element, not request-level; top-level errorCode was "VALIDATION"` — precisely what the old assertions accepted |
| revert reason-first derivation | fails exactly the two 409-collapse assertions |
| remove the decorator | fails the reconstruction tests; the contract-preservation tests still pass |
| remove the decorator, run the e2e | `node 2 answered 500 where the leader answered 400` |

## Validation

- `bash scripts/agent-check` — PASS
- `just test` — exit 0, **99 packages**, 0 failures
- `golangci-lint --max-same-issues=0 --max-issues-per-linter=0` on all changed
  packages — clean. The 4 remaining `SA4023` findings
  (`cmd/ledgerctl/store/bootstrap.go`, `internal/adapter/grpc/server_cluster.go`)
  are in files **byte-identical to base**, verified with `git show <base>:<file>`
- `just test-e2e` cluster — **275 Passed | 0 Failed** (1 skipped), verified from
  the `Ran 275 of 276 Specs` line rather than the `ok` exit
- `just test-e2e` business at seed 1788860003 — **535 Passed | 1 Failed**
- `just test-schemathesis` — `RESULT: ALL CHECKS PASSED`, 62 endpoints
- `AI_REVIEW_BASE_SHA=b2b7e7d03 bash scripts/agent-check-pr` — one failure,
  `scripts/rootguard TestRunnerReapsSurvivingDescendantBeforeFinalSnapshot`
  (`signal: killed` after 14.7s under full-gate load). `scripts/rootguard` is
  **byte-identical to base** and the test passes 3/3 in 1.6s in isolation — a
  load-induced timeout, not candidate-caused. The gate aborted there, so its
  `test-e2e` and `test-schemathesis` stages were run separately, above

The one business failure is `Cross-store snapshot alignment`
(`cross_store_snapshot_alignment_test.go:156`), on the
`Eventually(..., 3*time.Minute).Should(BeZero(), "fold backlog drained")`
cleanup budget whose own comment notes it "can exceed on slow runners". It is
pre-existing and spec-order sensitive — all 536 specs share one node, and Ginkgo
randomizes order per seed. At a **matched seed**:

| Tree | Seed | Result | Duration |
|---|---|---|---|
| `release/v3.0` | 1788860003 | **2 failed** | 618s |
| this branch | 1788860003 | **1 failed** | 410s |

Same assertion, same failure mode, base strictly worse and slower. Both specs
pass in isolation on both trees.

Closes EN-1636, EN-1980.

## Bugfix evidence

```text
DISCOVERY: NO_EXISTING_WORK
BEFORE_FIX: BUG_REPRODUCED
AFTER_FIX: PASS
BASELINE_CLASSIFICATION: BASELINE_FAILURE
```

- `BEFORE_FIX` — `tests/e2e/cluster/rest_error_parity_test.go` was ported
  unchanged onto the merge-base (`72d51387e`), where it compiles as-is and
  fails: *node 2 answered 500 where the leader answered 400; body: internal
  server error (correlation ID: ...)*.
- `AFTER_FIX` — the four parity specs run focused on this head:
  `Ran 4 of 276 Specs`, `4 Passed | 0 Failed`.
- `BASELINE_CLASSIFICATION` — the one `test-e2e` failure
  (`cross_store_snapshot_alignment_test.go:156`, `fold backlog drained`) also
  fails on the rebase target `47ae5b048` at the matched seed 1788874011, which
  additionally fails `cross_store_value_skew_test.go:161` (base 2 failed,
  candidate 1 failed). Not candidate-caused.


gfyrag:
The interception point looks right to me: decoding at the leader-forwarding connection boundary covers unary and streaming calls without duplicating conversion in every Ledger client method. I would keep grpcerr.NewConn.

I think the reconstructed error carrier is more complex than necessary, though. It deliberately has two identities at once: it unwraps to a domain business error while still exposing the original gRPC status through GRPCStatus. Those identities can disagree. For example, a PermissionDenied status carrying the known LEDGER_DELETED reason is classified as Conflict by domain.Kind, while status.Code still reports PermissionDenied. Which classification a caller sees then depends on how it inspects the same error.

I would suggest simplifying the contract before merge:

- Keep the connection wrapper at the forwarding seam.
- Decode business ErrorInfo directly into domain.BusinessError, and bare NotFound into commonpb.NotFoundError.
- Leave only genuine transport failures such as Canceled, Unavailable, and DeadlineExceeded as raw gRPC statuses.
- Move both encoding and decoding into grpcerr as a single leaf wire codec. Today the server mapping and its inverse live in separate packages, and the decoder tests have to reproduce the encoder mapping manually, which can drift.
- For a known reason, validate that the incoming gRPC code agrees with the expected business kind/code. Treat a contradictory pair as a malformed/protocol error. For an unknown reason, keep the code-based fallback for forward compatibility.
- Add round-trip tests over the actual encoder and decoder, plus tests for contradictory code/reason pairs.

Keeping GRPCStatus for a potential second hop does not seem necessary: once reconstructed as a BusinessError, the existing server-side converter can encode it again if it crosses another gRPC boundary.

I would not replace the wrapper with per-method conversion, and a global gRPC interceptor would be less scoped. So the seam chosen by this PR is good; I am mainly challenging the dual-purpose reconstructedError and the split codec. Since v3 is unreleased, this seems like a good moment to make that contract simpler and stricter.

paul-nicolas (commented):

**DECISION: FINDINGS** (no blocker). Head reviewed: `8f670cf2ceed324ee62b711208828ca2ee440e19`.

The mechanism is sound and I could not break it. Specifically verified against real code, not accepted on the strength of the comments:

- `FromStatusError`'s passthrough set loses nothing: `kindToGRPCCode` never emits `Aborted`/`OutOfRange`/`Unimplemented`/`Canceled`, `context.Canceled` is remapped to `Unavailable` at `server.go:579-581`, and no `Describable` returns an empty `Reason()` — so the unconditional `Canceled` passthrough and the `Reason == ""` skip are both safe.
- The `reconstructedError` shape is right. grpc-go v1.82.1 `status.FromError` type-asserts `interface{ GRPCStatus() *Status }` directly, so `status.Code`/`status.FromError` keep working; `errors.AsType[Describable]` unwraps past the carrier to the `*BusinessError`, and `domain.Kind` hits the `kindOverride` branch. The deliberate choice *not* to implement `Describable` on the carrier is correct as documented.
- **Second hop preserves the details.** Returned bare, the value takes `convertToGRPCErrorWithContext`'s "already a status" shortcut and grpc's `writeStatus` reads `GRPCStatus()`, so code, message and `ErrorInfo` are byte-identical to the leader's. Wrapped by a handler, `status.FromError`'s `errors.As` path does `p := grpcStatus.Proto(); p.Message = err.Error()` and `Proto()` is a `proto.Clone`, so `Details` survive and only the message gains a prefix.
- **The `RemoteError` containment invariant holds.** `*domain.RemoteError` is constructed in exactly one place (`grpcerr/errors.go:137`); `grpcerr` is imported by three non-test packages (`cmd/ledgerctl/cmdutil`, `internal/bootstrap`, and `internal/adapter/http` in test only) and never by admission, the FSM or `internal/infra/state`. `buildAuditFailure` is fed the locally raised `Describable`. Nothing persists or hashes a `RemoteError`.
- `clientStream`: `Header()`/`Trailer()`/`CloseSend()`/`Context()` promote correctly, the generated client makes no concrete-type or extra-interface assertion, and `normalizeStreamEnd` branches only on `status.Code(err) != codes.Canceled` — which is never reconstructed — so a reconstructed business error at `Recv` cannot be swallowed as end-of-stream.
- **The transport-prefix removal is complete and compiler-enforced.** 13/13 `fmt.Errorf` sites removed and the `fmt` import deleted from `client_bucket.go`, so re-adding a prefix requires re-adding the import. All 13 removed literals grepped repo-wide: zero stale references. `go list -deps ./cmd/ledgerctl/...` base 842 → head 843, delta exactly `internal/adapter/grpcerr`.
- The `status.FromError` rationale in the new test comments is **true**, proved in a scratch module pinned to this repo's `go.mod`: a `fmt.Errorf("prefix: %w", statusErr)` takes the `errors.As` branch, which rewrites `p.Message` from `err.Error()`, while Code and `details=1` survive.
- **The new e2e spec runs in CI** — job `Tests-E2E-Cluster` (`.github/workflows/main.yml:132-134`) → `just test-e2e-coverage './tests/e2e/cluster/...'` → `go test -race -tags "e2e,…" ./tests/e2e/cluster/...`, and its profile is a `required_profiles` entry. `go vet -tags e2e` is clean. The deleted `cmdutil` tests are re-homed name-by-name, and the CLI boundary is still integration-tested by `TestFormatGRPCError_BusinessError_ReturnsDisplayed`.
- `TestConvertToGRPCError_ReadIndexNotCaughtUp`: every claim in its long comment checks out, including "50x at 0.2s" (`pkg/actions/grpc.go:4-15`, `MaxAttempts: 50`, `InitialBackoff/MaxBackoff: 0.2s`, `RetryableStatusCodes: ["UNAVAILABLE"]`).

### Counts

| | P0 | P1 | P2 | P3 |
|---|---|---|---|---|
| blocking | 0 | 0 | 0 | 0 |
| non-blocking | 0 | 0 | 3 | 9 |

3 posted inline (all P2, all non-blocking); 9 P3 withheld from inline and listed below.

### Verification

15 candidate findings entered adversarial verification; 3 survived at P2, 9 were downgraded to P3, 3 were rejected outright:

- **rejected** — *`clientStream.Header()` bypasses reconstruction* (Codex): no production call site invokes `Header()` on a forwarded stream. Only `Trailer()` (`client_bucket.go:171,251,414`, `cursor.go:128`) and `CloseSend()` (`cursor.go:70,139`) are used, and neither returns an error needing reconstruction. No execution path.
- **rejected** — *`ListNumscriptVersions` breaks the parity claim*: the mechanism is real (`controller_default.go:1916` returns the raw `domain.ErrNotFound`, which the leader sanitizes to `codes.Unknown`, so a syncing follower answers 500 where the leader answers 404 via `error_handler.go:87`) — but it is wholly pre-existing, `controller_default.go` is not in this diff, and the doc's parity claim is scoped both by its `### Forwarded Writes` heading and by the `codes.Unknown` passthrough bullet three lines above it. Worth a standalone ticket, not a finding against this PR.
- **rejected** — *forwarded `NotFound` arms a new error-level log*: proved by execution that `errors.Is(err, &commonpb.NotFoundError{})` at `controller_routed.go:231` now matches on the forwarded path and did not at base — but the guard already fires for **every local 404** at base, the log already records `forwarded` (line 239) so it misattributes nothing, and on HTTP the pre-PR forwarded case emitted an error-level `"HTTP unmapped handler error"` *plus* a wrong 500. Net HTTP log volume unchanged, status code fixed. Nitpick, below the P3 bar.

### P3 (not posted inline)

- `docs/…/api/http-api.md:145-146`, `internal/adapter/grpcerr/conn.go:15`, `internal/bootstrap/controller_routed.go:55` — the method counts are wrong. `BucketServiceClient` has **37** methods, not 32 (`BucketGrpcClient` has 30 — neither is 32), and **10** `BucketGrpcClient` methods surface the error at a later `Recv()`, not six (11 `ServerStreamingClient` methods in the generated interface). The design conclusion holds a fortiori; only the stated evidence is false.
- `internal/adapter/grpcerr/errors.go:13-16` — the leaf rationale is 3/5 false. Measured on the **base** checkout, `go list -deps ./cmd/ledgerctl/...` already contains `internal/storage/dal` (+58 pebble packages), `internal/storage/readstore` and `internal/infra/node`; `grpcerr` and `internal/domain` return the *same* formance set, because `internal/domain` itself imports `internal/storage/dal`. Only `usagestore` and `application/backup` are genuinely avoided. The conclusion (do not import `internal/adapter/grpc` from the CLI) stands — that import would add 16 internal packages.
- `internal/bootstrap/controller_routed.go:58` — "the only production site that builds a leader-forwarding client" overstates it. `ClusterServiceServerImpl.leaderClient()` (`server_cluster.go:230-244`) and `GetClusterState`'s `nodeId == 0` branch (`:111-126`) also build a client aimed at the current leader. Neither needs the decorator today (gRPC-to-gRPC only; HTTP calls `b.Node.GetClusterState` in-process), but the claim is the kind a contributor adding an HTTP surface over cluster ops would trust instead of re-checking — which is the exact shape of the bug this PR fixes.
- `internal/adapter/grpcerr/errors_test.go:215` — `TestBusinessErrorRoundTrip`'s comment claims the replicated codes "prove the two directions agree". Its only assertion is `require.Equal(t, tt.err.Error(), reconstructed.Err.Error())`, and `RemoteError.Error()` returns `Message`, set from `st.Message()`, which `serverSideConvert` builds as `inner.Error()` — so the value is independent of both the code and the reason chosen in the replica. Every row still passes if a code is changed. Asserting `domain.Kind` and `Reason()` equality would make the comment true.
- `internal/adapter/grpcerr/errors.go:177` — only one row of `kindForCode` is asserted anywhere. For a *known* reason, drift is structurally impossible (both sides route through `domain.KindForReason`, pinned by `TestKindForReasonMatchesTypedErrors`), so the exposure is the unknown-reason fallback only: a change to `kindToGRPCCode` would silently disagree with no failing test. A `kindForCode(kindToGRPCCode(k))` table over every `ErrorKind`, minus the documented `FailedPrecondition` collapse, would close it.
- `internal/adapter/grpc/server_bucket.go:532,588,702,937,1003,1037,1066,1374` and `stream_helper.go:133` — the no-prefix contract the new tests state as absolute is re-violated one frame above on the follower's own gRPC surface. `server_bucket.go:1066` re-adds `fmt.Errorf("listing signing keys: %w", err)` to the error `client_bucket.go:245` just stopped prefixing, so a gRPC caller hitting a follower sees `desc = "listing signing keys: ledger deleted: foo"`. Pre-existing, and this PR strictly improves it (one layer removed, `"rpc error: code = …"` noise stripped, Code and `details=1` unaffected). Flagged only because the unary write path — the one the e2e spec covers — *is* clean end to end, which can read as full coverage.
- `docs/…/api/README.md:3` and `docs/technical/agent-context.md:33-35` — see the inline comment on `grpcerr/errors.go`.
- `openapi.yml:2659,2745` — the PR adds `'503'` to the two index paths but not `'429'`, which `KindResourceExhausted` makes reachable on every write (`error_handler.go:37-38`; `WRITES_BLOCKED_DISK_FULL`, `SEQUENCE_EXHAUSTED`). `'429'` is documented on `POST /v3/{ledgerName}/bulk` only (`:1837`, inline — there is no shared 429 component, though `ServiceUnavailable` exists at `:2987`), and `'503'` is still absent from ten read paths reachable through `commonpb.ErrNoLeader`. Pre-existing and unchanged by this PR — the per-path union of reachable statuses is the same, since every code the PR makes reachable on a *follower* was already reachable on the *leader*. Noted only because the PR did audit these two paths and left the identical class of gap on them.
- `internal/adapter/grpc/server.go:730-737`, `internal/adapter/http/response.go:188-193` — the two sanitizer log lines carry only `trace_id`/`span_id`/`correlation_id`, no method or route, so the removed prefixes were their only in-line call-site identity. Well mitigated: `loggingInterceptor` (`server.go:423-431`) always names the RPC on an adjacent line, and `recordGRPCInternalError` does not even fire for a forwarded status. The residual is an operator reading HTTP logs *without* traces for a forwarded status `grpcerr` deliberately leaves untyped.

Also noted, below the bar: `controller_routed.go:231-246` fires an error-level `"GetTransaction returned NotFound for committed transaction"` for any `*commonpb.NotFoundError`, including a genuinely nonexistent id and a missing ledger. Pre-existing on every local 404; this PR extends it to the forwarded route (reachable in a healthy cluster via `x-consistency: leader`).

### Models and conflicts

Claude (three area reviewers: reconstruction correctness, transport/CLI, tests+docs) and Codex, then five adversarial verifiers. No substantive conflict: Codex found nothing the Claude pass missed, and both of its findings failed verification or were downgraded. The two reviewers that reached the `http-api.md` passthrough bullet independently, from different angles (reconstruction order vs. omitted auth codes), are merged into the single inline comment below.

docs/technical/architecture/subsystems/api/http-api.md
```diff
@@ -110,6 +110,56 @@ Every path logs the raw cause **server-side** with a `correlation_id` field. Whe
...
+so that reconstruction happens once, at the transport seam:
+
+- a status carrying a ledger-domain `ErrorInfo` becomes a
+  `*domain.BusinessError` wrapping a `*domain.RemoteError`;
+- a bare `codes.NotFound` becomes a `*commonpb.NotFoundError`;
+- `codes.Canceled`, `Unavailable`, `Internal`, `Unknown` and `DeadlineExceeded`
+  are left untouched — `Canceled` because the cursor layer keys end-of-stream
+  detection off it, `Unavailable` because `handleError` already answers it with
+  `503` + `Retry-After`, and the rest because they denote server faults with
+  nothing typed to restore.
```
paul-nicolas: [P2][non-blocking] Qualify the passthrough bullet — `Unavailable` and `Internal` **are** reconstructed when they carry a ledger `ErrorInfo`

This bullet says `codes.Canceled`, `Unavailable`, `Internal`, `Unknown` and `DeadlineExceeded` "are left untouched". Only `Canceled` is unconditional.

`FromStatusError` checks the `ErrorInfo` **before** any code except `OK`/`Canceled` (`internal/adapter/grpcerr/errors.go:75-85`):

```go
if st.Code() == codes.OK || st.Code() == codes.Canceled { return err }
if bizErr := BusinessErrorFromGRPC(err); bizErr != nil { ... }
```

And `kindToGRPCCode` routes `KindUnavailable`/`KindInternal` `Describable`s to exactly those two codes (`internal/adapter/grpc/errors.go:47,:57`), so both arrive **with** an `ErrorInfo` and are reconstructed. This PR's own tests assert it: `error_handler_forwarded_test.go` pins `codes.Unavailable` + `INDEX_BUILDING` to `503` with `errorCode: INDEX_BUILDING`, and `codes.Internal` + `COVERAGE_MISS` to `500` with `errorCode: COVERAGE_MISS` — "a genuine 500, but a named one", as the test row itself puts it.

The package doc gets this right ("A **bare** `codes.Unavailable` already reaches the right outcome", `grpcerr/errors.go:52`); the subsystem doc drops the qualifier.

Second, the list reads as exhaustive when the rule is the complement: `FromStatusError` reconstructs exactly two shapes and returns *everything else* unchanged. A bare `codes.Unauthenticated`/`PermissionDenied` from the leader is therefore also passed through to the 500 fallthrough — deliberate, and explained in `error_handler_forwarded_test.go:44-48` ("an auth status arriving from the leader denotes a cluster-secret failure … and correctly stays a 500"), but absent here.

**Impact:** a contributor reading this would conclude that `INDEX_BUILDING`, `READ_INDEX_NOT_CAUGHT_UP` and `BALANCE_NOT_PRELOADED` lose their reason across the hop, and could add a redundant HTTP code branch or "fix" a non-bug. It also contradicts the change's own regression tests, which is the worst kind of doc drift to leave in an authoritative subsystem document.

**Resolution:** the bullet reads "a **bare** `codes.Unavailable`/`Internal`/`Unknown`/`DeadlineExceeded`", names `Canceled` as the one unconditional passthrough and why, and states the rule as its complement — everything that is neither a ledger-domain `ErrorInfo` nor a bare `NotFound` — naming the auth codes with the rationale the test already gives.

tests/e2e/cluster/rest_error_parity_test.go
```diff
@@ -0,0 +1,271 @@
...
+			if len(decoded.Data) > 0 && decoded.Data[0].ErrorCode != "" {
+				return decoded.Data[0].ErrorCode
+			}
+
+			return decoded.ErrorCode
+		}
+
+		onLeader := elementCode(leader)
+		Expect(onLeader).NotTo(BeEmpty(), "the element must fail for this spec to mean anything")
+		Expect(onLeader).NotTo(Equal("ERROR"), "the leader must name the business reason")
```
paul-nicolas: [P2][non-blocking] Give the bulk parity spec an expected code — a request-level rejection satisfies all three assertions

`elementCode` returns `data[0].errorCode` when the array is non-empty and otherwise falls back to the **top-level** `errorCode` (lines 250-258, "Return whichever fired so a mismatch between nodes is visible either way"). The three assertions are then `NotTo(BeEmpty())`, `NotTo(Equal("ERROR"))` and follower == leader — with no expected value anywhere.

A request-level rejection sets a non-empty top-level `errorCode` **identically on every node**: `serveBulk` answers a parse failure with `writeBulkErrorResponse(w, http.StatusBadRequest, "VALIDATION", err)` (`internal/adapter/http/handlers_bulk.go:50`), before any element reaches admission and therefore before anything crosses the forwarding seam. `"VALIDATION"` is non-empty, is not `"ERROR"`, and is the same on the leader and every follower. All three assertions pass and the spec proves nothing about the per-element mapper it exists to cover.

Today the body does reach the FSM on the leader — it is byte-for-byte the shape of the known-good `bulkWriteBody` at `internal/adapter/http/handlers_bulk_test.go:24`, `BulkElement.UnmarshalJSON` accepts `action`/`data`, and `no-such-account:a` as a posting source yields a zero-balance `INSUFFICIENT_FUNDS` (`KindPrecondition` → 400). So the spec is correct as of this head. But nothing in it *pins* that, so a later change to bulk parsing or admission would make the PR's only bulk regression test silently stop testing the seam while staying green.

This is the "verify that a rewritten test still reaches the original regression state rather than only a nearby proxy state" clause of `docs/technical/contributing/ai-review.md`, applied to a new test.

**Resolution:** assert the expected per-element code and that the failure really was per-element:

```go
Expect(decoded.Data).To(HaveLen(1))
Expect(onLeader).To(Equal("INSUFFICIENT_FUNDS"))
```

so a request-level rejection fails the spec instead of satisfying it.

internal/adapter/grpcerr/errors.go
```diff
@@ -0,0 +1,198 @@
...
+// that dispatches on the Describable contract — the HTTP error handler, the
+// bulk per-element mapper, the CLI formatter — falls through to its generic
+// "unknown error" branch. This package turns the wire representation back into
+// a Describable so those consumers keep working across a network hop.
+//
+// It is deliberately a leaf: it imports only gRPC, errdetails, internal/domain
+// and internal/proto/commonpb. internal/adapter/grpc, the intuitive home,
+// transitively pulls in the Pebble DAL, the readstore, the usage store, backup
+// and the Raft node, which cmd/ledgerctl must not link.
+package grpcerr
```
paul-nicolas, line 70: [P2][non-blocking] Register `internal/adapter/grpcerr` in the api README and the `agent-context.md` routing table

Re: `docs/technical/architecture/subsystems/api/README.md:3` and `docs/technical/agent-context.md:33-35` — neither file is in this diff, so I am commenting here, on the package this concerns.

This PR adds a new production package in the api adapter layer (198 + 68 lines, plus the seam applied at `internal/bootstrap/controller_routed.go:59`) and documents it only as a section inside `http-api.md`. It appears in neither index:

- `docs/technical/architecture/subsystems/api/README.md:3` still describes the layer as "Client-facing transport layers (`internal/adapter/grpc`, `internal/adapter/http`, `internal/adapter/auth`)" — `grpcerr` is not listed.
- `agent-context.md` maps `internal/adapter/grpc/**` (:33) and `internal/adapter/http/**`, `openapi.yml` (:35) to the api subsystem docs. `internal/adapter/grpcerr/**` matches neither pattern — `grpc/**` does not match `grpcerr/`.

**Impact:** `CLAUDE.md`'s documentation-maintenance rule requires a new technical mechanism to update "the matching `docs/technical/architecture/` subsystem documentation **and its README**". Beyond the rule, `agent-context.md` is the repository's authoritative context-routing contract, and the omission defeats the documented loading procedure: an agent following steps 2-3 of "How to load context" for a change under `internal/adapter/grpcerr/**` finds no row, so it never loads `http-api.md` — the only place the seam, the reason-first kind derivation and the passthrough contract are written down. That is precisely the context a change to this package must not be made without.

**Resolution:** `grpcerr` appears in the README's package list, and `agent-context.md` gets a row for `internal/adapter/grpcerr/**` (or the existing `internal/adapter/grpc/**` row is widened to cover it).

gfyrag:
I went through the PR again from scratch.

The EN-1636 fix itself is sound: rebuilding typed errors at the leader-forwarding boundary restores follower/leader HTTP parity, covers unary and streaming calls, preserves reason-first business semantics, and keeps the original gRPC status/details across a second hop.

However, I would hold the merge and align this with EN-1980: https://formance-team.atlassian.net/browse/EN-1980

RemoteError represents a network-decoded error, is only constructed by the gRPC adapter in production, and must never be raised by admission/FSM/processing or become audit business state. It should therefore not live in the domain package or require a private kindOverride in domain.Kind. The decoded remote representation and its transport behavior belong at the adapter boundary.

One correction to my earlier feedback: preserving the original gRPC status is necessary. Semantic kind and transport code are separate axes. READ_INDEX_NOT_CAUGHT_UP is intentionally semantically unavailable while using FailedPrecondition on the wire to avoid the automatic Unavailable retry policy. The better direction is therefore to relocate and validate this representation, not to re-encode everything from BusinessError alone.

Before merge, I would also address the still-valid resolved threads:
- fix the passthrough documentation, since a Ledger ErrorInfo is reconstructed even for statuses such as Internal or Unavailable;
- make the bulk forwarding test assert one result and INSUFFICIENT_FUNDS so it cannot pass on a top-level rejection;
- document the new grpcerr package in the API routing docs and correct the method/stream counts.

I ran the complete PR validation locally: race-enabled unit tests, business and cluster e2e suites, and Schemathesis all pass. So this is an architectural hold, not a newly found EN-1636 runtime regression.

paul-nicolas:
Addressed in `6849a7639` (+ a one-line README follow-up). Summary of what changed, in the order you raised it.

**EN-1980 relocation.** `domain.RemoteError` is gone and `domain.Kind` is strictly reason-based — no `kindOverride`, no escape hatch. The decoded representation now lives in a new transport-neutral leaf, `internal/adapter/apierr`, whose `Describe(err)` returns a `Descriptor` (semantic kind, reason, client-safe message, metadata) reading either a locally raised `domain.Describable` or an `*apierr.Remote` decoded from a peer. `handleError`, the bulk per-element mapper and `ledgerctl` all read through that one contract; `apierr` imports `internal/domain` and nothing else, so HTTP links no gRPC detail. The `Remote` branch is checked first, and there is a test that fails if the order is swapped.

There is now one table per direction rather than two that can drift: `kindToGRPCCode` moved out of `internal/adapter/grpc` into `grpcerr.CodeForKind`, so the encoder and the decoder that reverses it are the same table, and `allowedWireCodes` is derived from it.

The audit exhaustiveness test no longer carries the unreachable `RemoteError` row. Per your "do not claim a type-system guarantee that does not exist": the separation is a dependency rule, enforced by `scripts/check-repo-invariants` — an `apierr` import from `internal/domain`, `internal/application/admission`, `internal/infra/state`, `internal/infra/plan` or `internal/infra/preload` fails the check. I proved it fires by dropping a probe file under `internal/infra/state/` and getting a real `check-repo-invariants: FAIL` naming the line.

**Your correction is the load-bearing part.** Preserving the original status is kept exactly as it was — the carrier still answers `GRPCStatus()` with the received status. `READ_INDEX_NOT_CAUGHT_UP` stays semantically `KindUnavailable` (503 + `Retry-After` on REST) while travelling as `FailedPrecondition` on every hop, and that is now pinned on both axes at the unit level, over a real bufconn connection, and at the HTTP surface. Nothing is re-encoded from `BusinessError` alone.

That pair is also the one declared entry in `grpcerr.wireCodeExceptions`, because the new validation is reason-specific as EN-1980 requires: allowed set = `CodeForKind(KindForReason(reason))` ∪ per-reason exceptions. Deriving it from `CodeForKind(kind)` alone would have rejected the pair the server actually sends. A known reason arriving outside its allowed set is a protocol fault, not a business outcome — it decodes to an `*apierr.InvalidWireError` that implements neither the boundary contract nor `GRPCStatus`, so it reaches the internal-error sanitizer (500 + correlation ID + server-side log) and the received reason, message and metadata are never echoed. Dropping the status matters: otherwise a contradicting `InvalidArgument` pair would be answered to the caller as a trusted 400 carrying the peer's own message.

Conversion coverage is now driven off the `ErrorReason` enum itself rather than sampling one reason per kind — every reason, under every code its policy allows — plus a second test that walks every `KindConflict` and `KindPrecondition` reason across the shared `FailedPrecondition` and refuses to run unless at least two of each exist.

**The three threads.**

1. **Passthrough documentation** — the bullet now qualifies `Unavailable`/`Internal`/`Unknown`/`DeadlineExceeded` as **bare**, names `Canceled` as the one unconditional passthrough with the `cursor.go` reason, states the rule as its complement (everything that is neither a ledger-domain `ErrorInfo` nor a bare `NotFound`), and names the bare auth codes with the cluster-secret rationale the test already gave. It also spells out that `Unavailable`/`Internal` **are** reconstructed when they carry an `ErrorInfo`, which is what the old text got wrong.
2. **Bulk forwarding test** — now `Expect(decoded.Data).To(HaveLen(1))` and `Expect(onLeader).To(Equal("INSUFFICIENT_FUNDS"))`. Mutation-proven: pointing the spec at a request-level rejection fails it with `the rejection must be per-element, not request-level; top-level errorCode was "VALIDATION"` — exactly the case the old three assertions accepted.
3. **Registration and counts** — `apierr` and `grpcerr` are in the api README package list and have a row in `agent-context.md`. Counts corrected: the generated `BucketServiceClient` has **37** methods, **11** server-streaming (was "32"). The "six list methods" claim was right but unstated; the doc now names them — the six `BucketGrpcClient` methods returning a lazy cursor, `ListSigningKeys` excluded because it drains eagerly.

**Validation.** `agent-check` PASS; `just test` 99 packages, 0 failures; cluster e2e `Ran 275 of 276 | 275 Passed | 0 Failed`; business e2e at seed 1788860003 `535 Passed | 1 Failed` (the pre-existing `cross_store_snapshot_alignment` fold-drain budget, base strictly worse at the matched seed); schemathesis `ALL CHECKS PASSED`. `agent-check-pr` had one failure, `scripts/rootguard TestRunnerReapsSurvivingDescendantBeforeFinalSnapshot` (`signal: killed` under full-gate load) — that file is byte-identical to base and the test passes 3/3 in 1.6s in isolation, so its `test-e2e`/`test-schemathesis` stages were run separately. Every new test was mutation-proven; the table is in the PR body.

**One thing I did not do.** Base has advanced 8 commits, including `feat(grpc)!: require a matching service protocol revision (EN-1851)`. This change does not touch the service contract, so `pkg/grpcprotocol.Version` needs no increment, and the branch merges cleanly with `release/v3.0` (`git merge-tree`, 0 conflicts). I left the rebase out of this revision so the validation above stays attributable — say the word and I will rebase and re-run.

gfyrag:
I went through the current head again. The EN-1980 architectural change is now in the right direction: the remote representation is out of `internal/domain`, `domain.Kind` is strictly reason-based again, and decoding lives at the adapter boundary.

I still see two blockers before merge:

1. **The candidate no longer builds against the current `release/v3.0`.** The branch is still based on `b2b7e7d03`, while `release/v3.0` is now `22b2ddde36`. The newer base removed `ERROR_REASON_READ_INDEX_NOT_CAUGHT_UP`, but `internal/adapter/grpcerr/errors.go` still references it in `wireCodeExceptions`. I built the synthetic merge tree and got:

   ```text
   internal/adapter/grpcerr/errors.go:129:11: undefined: commonpb.ErrorReason_ERROR_REASON_READ_INDEX_NOT_CAUGHT_UP
   ```

   The GitHub checks are green because they ran against the old base. Please rebase and remove/adapt the obsolete exception, tests and documentation before revalidating.

2. **The invalid-wire sanitization contract is not enforced on every surface.** `FormatGRPCError` calls `grpcerr.Decode(err)` only to detect `*apierr.Remote`; when decoding returns `*apierr.InvalidWireError`, it discards that result and formats the original gRPC status, including its untrusted message. `TestFormatGRPCError_InvalidWirePairIsNotABusinessError` currently pins the leak by expecting `ledger deleted: foo`. The bulk writer has the same class of gap: it returns `result.err.Error()` as `errorDescription`, so a bare `InvalidWireError` exposes the received reason/code rather than a generic correlated internal error. This contradicts the PR's stated guarantee that an invalid pair is logged server-side and its received reason/message/metadata are never echoed. Please make the safe representation shared or handle `InvalidWireError` explicitly in both consumers, with negative assertions over the client-visible output.

One smaller contract mismatch: `FromStatusError` says an `ErrorInfo` from another domain is left untouched, but a foreign-domain `NotFound` falls through to the bare-NotFound conversion and becomes a ledger `commonpb.NotFoundError`. Either require that there are no `ErrorInfo` details before taking that branch, or qualify and test the exception.

Local validation on the PR head remains green (`agent-check` and focused race tests), but the current-base merge compilation failure is decisive. I have not approved the PR.

gfyrag:
Re-review of `93d94ed84`: the stale-base compilation failure is fixed, and the explicit `InvalidWireError` handling now closes the previously reported ledgerctl and bulk leaks for the tested codes. The synthetic merge with current `release/v3.0` passes `agent-check` and the targeted race suites.

Two blocking edge cases remain:

**[P2][blocking] Validate ledger ErrorInfo before the unconditional Canceled passthrough**

`internal/adapter/grpcerr/errors.go:155-170` returns every `codes.Canceled` status before calling `Decode`. A status carrying `reason=LEDGER_DELETED`, `domain=ledger`, `code=Canceled`, and message `ledger deleted: secret-ledger` therefore remains a raw status instead of becoming `InvalidWireError`. In the bulk path (`internal/adapter/http/handlers_bulk.go:266-289`) this produces status 500 with `errorCode=ERROR` and a client-visible description containing `secret-ledger`. This bypasses the central guarantee that a known reason/code mismatch is sanitized. Decode the ledger detail first; keep the passthrough only for a bare Canceled status so normal cursor termination remains unchanged. Add a regression covering the bulk response.

**[P2][blocking] Do not reinterpret a foreign-domain NotFound as a Ledger NotFound**

`Decode` correctly ignores an `ErrorInfo` whose domain is not `ledger`, but `FromStatusError` then applies its generic `codes.NotFound` fallback and wraps it as `commonpb.NotFoundError` (`internal/adapter/grpcerr/errors.go:159-175`). Extending `TestFromStatusError_ForeignErrorDomain` with `codes.NotFound` reproduces the failure: the original status becomes `*grpcerr.reconstructedError`. This contradicts the documented contract that foreign-domain details are left untouched. Apply the NotFound fallback only when the status carries no `ErrorInfo` at all, and cover both FailedPrecondition and NotFound in the foreign-domain test.

No other blocking finding found on this head.

paul-nicolas:
Rebased onto `bc3031e6e` (current `release/v3.0`) and both blocking edge cases are fixed. Head is `c7479199c`.

**Rebase.** Two conflicts, plus one the auto-merge did not raise.

- `internal/adapter/grpc/errors.go`: EN-1623 rewrote `describableToGRPCStatus` to encode `domain.PublicErrorDetails`, this PR replaced `kindToGRPCCode` with the shared `grpcerr.CodeForKind`. Resolution keeps both — the redacted message and metadata, encoded through the one table the decoder reverses.
- `internal/adapter/http/handlers_coverage_test.go`: `writeBulkResponse` gained an `*http.Request`; kept this PR's `testBulkRequest()` helper over the eight inlined constructions.
- The silent one: `handleError` passed an `apierr.Descriptor` to `domain.PublicErrorDetails`, which does not compile. The selection belongs in the contract both surfaces read, not re-derived after it, so `Descriptor.Message`/`Metadata` are now the public presentation and `Descriptor.PublicOverride` reports that the type owns one — that flag is what tells a surface to render `Message` instead of the wrapped chain the presentation exists to withhold. `bulkErrorDescription` reads it through `apierr.Describe` too, so it no longer keeps a second `domain.Describable` dispatch beside the one the rest of the mapper uses. A decoded remote failure needs no selection: the sender applied its own before serialising, so `PublicOverride` is false for an `*apierr.Remote` and a consumer keeps rendering its own outer context, exactly as before the hop.

**1. Ledger `ErrorInfo` validated before the `Canceled` passthrough.** Fixed. The detail is decoded first; the passthrough now covers a *bare* `Canceled` only. Your example — `reason=LEDGER_DELETED`, `domain=ledger`, `code=Canceled`, `ledger deleted: secret-ledger` — becomes an `InvalidWireError` and the bulk element answers `500` / `INTERNAL_ERROR` with a correlation ID and none of the payload. The regression is a new row on `TestBulkPerElementForwardedInvalidWirePairIsSanitized`, asserting the client-visible element body, alongside the `InvalidArgument` row rather than replacing it.

Pagination is unaffected in both directions, which is the part worth stating explicitly: a reconstructed value keeps answering `GRPCStatus()` with the received status, so `status.Code(err)` still reports `Canceled` for the unknown-reason path that cursor.go reads — pinned by `TestFromStatusError_CanceledWithUnknownReasonKeepsItsCode`. Only the contradiction loses the status, which is the intent.

**2. Foreign-domain `NotFound`.** Fixed by your first option: the fallback is gated on `hasErrorInfo(st)`, so a status carrying an `ErrorInfo` of any domain is a typed failure of whichever contract stamped it and is left untouched. `TestFromStatusError_ForeignErrorDomain` is now a table over `FailedPrecondition` and `NotFound` — the single code it had could never have caught this.

Both fixes are proven by mutation. Restoring the unconditional `Canceled` return fails the new bulk row with `secret-ledger` in the client-visible description and fails the two grpcerr tests; dropping the `hasErrorInfo` guard turns the foreign `NotFound` into a `*grpcerr.reconstructedError` while the `FailedPrecondition` control row stays green.

**Validation on the rebased head.** `agent-check` PASS, golangci-lint clean on both modules, `go test -race ./...`, `test-e2e` (business + cluster) and Schemathesis (62 endpoints, 0 failed) all pass. The e2e parity specs are verified to have executed, not merely compiled: `Ran 4 of 287 Specs … 4 Passed`.

One caveat, reported rather than hidden: inside the full gate, `tests/antithesis/TestRunModelTestRequiresVerifiedOutcome` fails under load — it wraps a shell script that runs `go build` in a hard 15s `context.WithTimeout` (`run_model_test_test.go:138`), which a concurrent `-race` build of the whole module cannot meet. It is not this diff: a matched `go test -race ./... -timeout 20m` on a clean `bc3031e6e` worktree fails the same test, and additionally fails `scripts/rootguard`, which the PR head does not. That gate abort is also why `test-e2e` and Schemathesis had to be run as separate steps above.

internal/adapter/grpcerr/errors.go
```diff
@@ -0,0 +1,340 @@
...
+//
+// A known reason is also validated against the code it arrived under, because
+// a reason this build knows is a reason whose legitimate codes it knows too. A
+// contradiction there is a protocol fault and nothing received is trusted.
+// An unknown reason cannot be validated — this build has no policy for it — so
+// its exact code, reason, message and metadata are preserved verbatim and the
+// code supplies the classification.
+func decodeReason(info *errdetails.ErrorInfo, st *status.Status) error {
+	rc := domain.ReasonCode(info.GetReason())
+	if rc == commonpb.ErrorReason_ERROR_REASON_UNSPECIFIED {
```
NumaryBot: 🟠 **[major] Reject the UNSPECIFIED wire reason**

When a peer sends `reason=UNSPECIFIED`, `ReasonCode` returns `ERROR_REASON_UNSPECIFIED`, but this branch treats it like an unknown future reason and trusts its code, message, and metadata. Since `UNSPECIFIED` is a known sentinel that no ledger business error may emit, a malformed pair can bypass invalid-wire sanitization and expose the peer's arbitrary payload; distinguish a missing enum lookup from the explicit `UNSPECIFIED` value and reject the latter.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1939#issuecomment-5586554895

internal/adapter/grpcerr/errors.go
paul-nicolas: Fixed in `957a9ddb3`. The conflation is real: `domain.ReasonCode` is a bare map index, so a name absent from the enum and the enum's own `ERROR_REASON_UNSPECIFIED` member both come back as the zero value, and `decodeReason` keyed its forward-compatibility branch on that value.

- `domain.LookupReasonCode` now reports whether the enum declares the name, and the passthrough keys on that miss instead of the zero value.
- `allowedWireCodes` returns an empty set for the sentinel, so the existing single validation branch rejects it as an `*apierr.InvalidWireError` under **every** code — including the `codes.Internal` that `CodeForKind(KindForReason(UNSPECIFIED))` would otherwise have licensed — and the peer's message and metadata are dropped on every surface.
- Tests: `TestDecode_ExplicitUnspecifiedReasonIsRejected` (five codes, plus the assertion that a reason from a newer server under the same code still decodes verbatim), `TestFromStatusError_ExplicitUnspecifiedReachesTheSanitizer`, and `TestLookupReasonCode*` in `internal/domain`. Reverting the decoder to the zero-value branch fails both new grpcerr tests (`Should be empty, but was [Internal]`), so they pin the defect rather than the code.

One correction on the framing: this was not a sanitization bypass. A peer that wants the same effect sends any reason string this build's enum does not declare and gets exactly the same verbatim treatment through the same branch — that is the documented forward-compatibility policy (`http-api.md`, "An unknown reason cannot be validated"). What was actually wrong is narrower: `UNSPECIFIED` is a value this build *knows* no ledger error emits, and it was answered as a business outcome (`400` with `errorCode: "UNSPECIFIED"` and the peer's message) instead of failing loudly, per invariant 7.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1939#issuecomment-5586554895

gfyrag (approved):
Re-reviewed at 957a9ddb3. The explicit UNSPECIFIED sentinel is now distinguished from an unknown future reason and rejected under every transport code, with regression coverage for both decoding and sanitizer routing. The previous Canceled and foreign-domain NotFound blockers remain fixed. All review threads are resolved; CI is green; targeted race tests and the canonical agent-check pass.

PR #1944, opened 2026-09-08
refactor(telemetry): remove in-process tail sampling (EN-1588)

gfyrag:
## What changed

Remove Ledger's in-process error-aware tail sampler, its two CLI flags, custom environment wiring, and the embedded Operator's sampling configuration/schema. Ledger uses the standard OpenTelemetry exporter; error status, attributes, exception events and trace relationships remain intact.

For collector-side sampling, keep OTLP export enabled and set `OTEL_TRACES_SAMPLER=always_on`. No Ledger-specific sampling setting remains. Rewrite deployment/CLI/architecture documentation and remove the deleted package from the lifecycle audit manifest.

## Why

[EN-1588](https://formance-team.atlassian.net/browse/EN-1588): local tail sampling retains spans for 30 seconds inside Ledger and competes with the application for CPU and memory. Sampling decisions and retention belong in the collector.

Need: export the Ledger span stream to the collector while preserving error instrumentation, without local tail-sampling cost or duplicate configuration.
Current limitation: the custom exporter buffers successful spans for later error discovery and requires settings separate from the SDK sampler.
Requirement: use standard SDK sampling/export configuration, retain emitted error signals and trace relationships, and remove local tail-retention state.
Durable contract and acceptance experiment: [trace sampling](docs/technical/architecture/subsystems/monitoring/trace-sampling.md).

## Technical decision

Delete the custom exporter and its wiring instead of maintaining another tail-sampling implementation inside Ledger. Keep the standard SDK sampler configurable; collector deployments must explicitly use `always_on` because the SDK otherwise defaults to parent-based sampling.

Remove `TracesConfig.Sampling` / `TraceSamplingConfig` from the embedded Operator, regenerate deepcopy and Cluster CRD/Helm schemas, and use the existing `extraEnv` path for the standard SDK setting. Unreleased v3 configuration gets no compatibility shim.

## Risk

**MEDIUM**: deployment configuration must remove the old controls and configure collector policies before relying on selective trace retention. Exhaustive export increases downstream traffic. No persisted business state, FSM/Raft, or authentication decision changes.

## Validation

- [x] `AI_REVIEW_BASE_SHA=4d4798f8db4fdc11705313a210d042540322dfb5 bash scripts/agent-check-pr` — full normalization/lint, baseline, full root-module `go test -race ./...`, and complete embedded Operator tests.
- [x] Targeted `go test -race ./cmd/server ./internal/adapter/auth -count=1`.
- [x] Operator generation and targeted controller environment tests with `-race`; CRD and Helm copies match.
- [x] Regression: standard SDK batch export delivers the child before the parent error, honors `always_on` under an incoming unsampled parent, and preserves IDs/error status/attributes/exception event.
- [x] Mutation check: removing the auth error status makes that regression fail at its status assertion.
- [x] Final diff/reference inspection and independent read-only review: no Ledger defects; external Operator adoption follow-up documented.

Validation base: `4d4798f8db4fdc11705313a210d042540322dfb5`.

- No local 30-second tail buffer, per-trace error cache, success-ratio filtering, or custom sampling exporter remains.
- Children may be exported in earlier batches before a parent ends with an error. Error instrumentation remains available to the collector.
- Remove `--trace-sampling-enabled`, `--trace-sampling-success-ratio`, `TRACE_SAMPLING_ENABLED`, `TRACE_SAMPLING_SUCCESS_RATIO`, and `monitoring.traces.sampling` from deployment configuration.
- SDK recording, batching, transport queues, and delivery limitations still apply. This does not guarantee complete distributed traces after upstream/transport losses or a collector's late drop decision.

## Review focus

Complete removal of custom controls, preservation of standard error signals, and accuracy of collector/Operator adoption requirements.

## Known concerns

The main Formance Operator is a separate repository. Its dependency bump must update [tests referencing the removed Sampling API](https://github.com/formancehq/operator/blob/c136147425376b7f1744b5aef4be73ee3b95025e/internal/resources/ledgers/v3_spec_test.go#L25) and regenerate LedgerConfiguration CRD/Helm schemas and copied Cluster test fixtures. These cross-repository changes are not included here.

The collector must provide trace-ID affinity, error/probabilistic-success policies, and span metrics before sampling, with sizing and long-trace handling validated. EN-1588's 8,000 TPS A/B, ≤10% P99 overhead, long-run resource behavior and complete error-trace acceptance remain external and unproven by local unit tests. Keep the ticket open.

[EN-1588]: https://formance-team.atlassian.net/browse/EN-1588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/adapter/auth/grpc_auth_traces_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"go.opentelemetry.io/otel/sdk/trace/tracetest"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// This test changes SDK environment configuration, so it cannot run in parallel.
+func TestAuthFailureExportsChildrenBeforeParentError(t *testing.T) {
+	t.Setenv("OTEL_TRACES_SAMPLER", "always_on")
+	exporter := tracetest.NewInMemoryExporter()
+	provider := sdktrace.NewTracerProvider(sdktrace.WithBatcher(exporter))
+	t.Cleanup(func() { require.NoError(t, provider.Shutdown(context.Background())) })
```
NumaryBot, line 21: 🔴 **[blocker] Configure the sampler on the test provider**

`sdktrace.NewTracerProvider` does not read `OTEL_TRACES_SAMPLER`; without `sdktrace.WithSampler(sdktrace.AlwaysSample())`, its default parent-based sampler inherits the incoming unsampled parent and records neither span. Consequently `require.Len(t, before, 1)` fails, so this newly added test cannot pass.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1944#issuecomment-5586586000

paul-nicolas (approved):

internal/adapter/auth/grpc_auth_traces_test.go
gfyrag: This finding does not apply to the pinned OpenTelemetry SDK v1.43.0 (go.mod). NewTracerProvider calls applyTracerProviderEnvConfigs before applying explicit options (trace/provider.go:99); tracerProviderOptionsFromEnv calls samplerFromEnv and installs its result with WithSampler (provider.go:476–487). samplerFromEnv reads OTEL_TRACES_SAMPLER and maps always_on to AlwaysSample() (trace/sampler_env.go:48–61). Source: https://github.com/open-telemetry/opentelemetry-go/blob/sdk/v1.43.0/sdk/trace/provider.go#L99 and https://github.com/open-telemetry/opentelemetry-go/blob/sdk/v1.43.0/sdk/trace/sampler_env.go#L48 . Re-ran `go test -race ./internal/adapter/auth -run '^TestAuthFailureExportsChildrenBeforeParentError$' -count=1` on c0c7455: PASS. The existing regression explicitly asserts export under an unsampled remote parent. Adding an explicit sampler would override the environment and remove coverage of the deployment setting being tested; retaining the test unchanged.

gfyrag:
Review feedback checked on both surfaces. The current NumaryBot finding is a false positive for pinned OTel SDK v1.43.0: source evidence and a fresh passing race test are recorded in https://github.com/formancehq/ledger/pull/1944#discussion_r3959865047 ; the thread is resolved. `bash scripts/agent-check` also passes. No code change is needed. The earlier Shipfox findings on c43dc13 are superseded by c0c7455, which deletes the custom sampling package and rewrites its contract; Shipfox's current-head review reports no material findings.

PR #1945, opened 2026-09-08
fix(EN-1979): invalidate preload loaders at commit; drop the index-key special cases

Azorlogh:
## What changed

The admission preload loader is now invalidated by the event that makes its memoized values stale: the FSM brackets every proposal's commit with a fence on the keys its plan covered (`state.PreloadInvalidator`, implemented by `plan.Builder.FencePreloaded` / `UnfencePreloaded`, wired in bootstrap). The fence drops the memo immediately before `pb.batch.Commit()`, and no load that starts, runs or completes while it is up is memoized until it lifts once the commit has returned, so a load whose store read may sit on either side of the write is returned to its caller but never kept, and no preload is served a memoized value from before a committed write to a key the proposal covered. The one write outside a proposal's covered keys is `DeleteLedger`'s ledger-wide range delete, whose keys no plan can enumerate; the commit of a batch that deletes a ledger is bracketed by `FenceAllPreloaded` / `UnfenceAllPreloaded`, which fence every loader. Loader entries are also keyed on a local cache reset count (`Cache.ResetSeq`, part of the new `preload.CacheStamp`), so a value read before a snapshot install or restore is never reused after one. `AttributeCache.CheckCache` compares the resident's tag and mirrors `Get`'s shadowing (a Gen0 resident decides; Gen1 is consulted only when Gen0 holds nothing), so a hash-colliding resident of another canonical key is a miss. With those in place the three index-registry special cases EN-1904 added to `resolveCoverage` are removed: index keys take the cache-hit shortcut, honour the bloom veto, and load through the loader like every other attribute; the arbitration re-read and its `assert.Unreachable("index preload served stale absence")` are gone.

## Why

Antithesis run `5572460b288ddd058df2b44d04d9f991-59-13` fired that assertion. Two `CreateIndex` bulks for the same index were in flight; the first admission memoized "absent", applied, and the second admission preloaded before the first's handler returned — the only point at which the loader entry was released — so it was served the stale absence and the arbitration had to correct it. The outcome was right (the arbitration seeded the row, and the apply reads the cache regardless), but the path is reachable by design under concurrent same-key admissions, so every such race fails a run. The root cause is the loader's late invalidation; EN-1904 worked around it for one key class by distrusting every cache except the loader, which is itself a cache. EN-1979.

DISCOVERY: race between concurrent admissions of one key and the apply-side write (Antithesis model driver)
BEFORE_FIX: BUG_REPRODUCED — `TestMachine_CommitFencesCoveredKeysAcrossTheCommit`, `TestMachine_DeleteLedgerCommitFencesEveryLoader`, `TestAttributeLoader_FenceBlocksMemoizationUntilUnfence`, `TestAttributeLoader_LoadStartedUnderFenceIsNotMemoizedAfterUnfence` and `TestAttributeCache_CheckCache_ForeignTagIsMiss` fail with the fence hooks, the fence's memo blocks and the tag check removed
AFTER_FIX: PASS

Need: admission must never build a plan from a value that a committed proposal has already changed, for any attribute type, without per-type escape hatches.
Current limitation: loader entries live until the proposer's handler returns or the generation advances, which is later than the FSM apply that changes the key; index keys were exempted from the cache shortcuts and re-read Pebble on absence to compensate.
Requirement / constraint: no preload is served a memoized value from before a committed write to its key (the memo is dropped before the commit and nothing loaded until the commit has returned is kept), and no memoized value outlives a cache reset; index keys follow the same resolution path as every other attribute.
Evidence: the Antithesis counterexample above (moment `4548305566712468551 / 147.138`, key `model-cc1c67cdb80b3ca9-2 / tx_builtin:TX_BUILTIN_INDEX_REVERTED_AT`, bulks at seq 213 and 216).
Durable repository evidence: `docs/technical/architecture/subsystems/fsm/preload.md` (loader section), `docs/technical/architecture/subsystems/fsm/deterministic-fsm.md` (fence bracket), `preload.CacheStamp` and `state.PreloadInvalidator` contract comments.

## Technical decision

Decision: a fence bracketing the commit, from the committer — dropping the memo before `pb.batch.Commit()` and holding memoization off until after it, because a plain release on either side of the commit leaves a window (before: a reload can memoize the pre-write value; after: the memo is still served until the release lands). A batch that deletes a ledger range-deletes keys no plan can enumerate, so it fences every loader for the same window (fence-all) — the one write path outside a proposal's covered keys. The proposer's handler-return release stays as the fallback for proposals that never applied. Tag-aware `CheckCache` makes the hit shortcut trustworthy for every key class, which is what let the index exemption go.

Why now / why proportionate: the model driver now churns index lifecycles and will keep tripping the assertion; the fix is smaller than the workaround it removes and closes the same window for all twelve attribute types.

Alternatives considered: (1) read index keys straight from Pebble, bypassing the loader — fixes the symptom for one class and keeps the late release for the rest; (2) reclassify the assertion as a coverage signal — leaves the double read and the special case; (3) release loader entries right after `fsmFuture.Wait` in admission — narrows the window without closing it.

## Risk

**LOW–MEDIUM** — touches the admission plan resolver for every attribute type (hit shortcut now tag-checked; index keys lose their exemptions), the loader's validity key, and brackets the FSM committer's `pb.batch.Commit()` with a fence hook on either side. No persisted format or proto change. The invalidator is nil on nodes without a plan builder. Per covered key the commit path does one fence (memo delete plus a count increment) before the commit and one unfence (count decrement) right after it returns. A ledger-deleting batch instead fences every loader: twelve loaders times 256 shards, each a mutex acquisition and a memo-map clear, twice per such batch — a rare, operator-driven proposal.

## Validation

- [x] Targeted tests: the fence goes up before the commit and lifts after it, with the proposal's write read back from the store absent at every fence and present at every unfence (`TestMachine_CommitFencesCoveredKeysAcrossTheCommit`); a fence drops the memo and blocks memoization until unfenced, fences count (`TestAttributeLoader_FenceBlocksMemoizationUntilUnfence`, `TestBuilder_FencePreloaded_DropsAndBlocksTheMemo`); a load that starts under a fence and completes after it lifts is not memoized (`TestAttributeLoader_LoadStartedUnderFenceIsNotMemoizedAfterUnfence`); a fence-all drops every memo in every loader and blocks memoization until lifted (`TestAttributeLoader_FenceAllBlocksEveryKey`, `TestBuilder_FenceAllPreloaded_DropsEveryLoader`); a batch that deletes a ledger fences every loader before its commit and lifts it after, with no per-key fence beside it (`TestMachine_DeleteLedgerCommitFencesEveryLoader`); a release landing mid-load skips the memo (`TestAttributeLoader_LoadOrWait_ReleaseDuringLoadSkipsTheMemo`); tag mismatch is a miss, a foreign Gen0 resident shadows a matching Gen1 entry, tombstone with the right tag a hit (`TestAttributeCache_CheckCache_ForeignTagIsMiss`); reset sequence counted by local and replicated resets (`TestCache_ResetSeq_CountsEveryClear`); loader reloads across a reset (`TestAttributeLoader_LoadOrWait_ResetSeqMismatchReloads`); index key on cache hit is coverage-only, foreign-tag resident loads, bloom veto honoured (`resolve_index_keys_test.go`)
- [x] Red-verified: the FSM fence test fails with the fence moved after the commit (the probe finds the row present at the fence) and with the fence's memo block removed; the started-under-fence test fails with the registration-time fence check removed; the three fence-all tests fail with the ledger-deletion flag never set and the loader's fence-all checks removed; the tag and mid-load tests fail with the tag check and the in-flight mark neutered
- [x] Full suite / broader validation: `internal/infra/{preload,cache,plan,state}`, `internal/bootstrap`, `internal/application/admission` suites green

- Admission: `preload.LoadOrWait` takes a `CacheStamp` (boundary, epoch, reset sequence); `CheckCache` takes the key's tag; `resolveCoverage` has no per-attribute branches.
- FSM: `Preload` records the plan's covered keys on the prepared batch (keys left by a failed prepare are discarded when the next prepare starts); the `PreloadInvalidator` hook is captured onto the prepared batch under `fsm.mu` as well, and `CommitPreparedBatch` fences the keys through it before the commit and unfences them as soon as the commit returns, reading no mutable Machine field; a batch that deletes a ledger fences every loader for the same window.
- Docs: `preload.md`, `deterministic-fsm.md`, `admission-cache-horizon.md` and `plan-intent-verification.md` state the fence bracket, the handler-return fallback and the tag-aware `CheckCache`.
- Removed: the EN-1904 arbitration read, the `index preload served stale absence` and `index preload bloom false negative` assertions, and `resolve_arbitration_test.go`.

## Review focus

1. The fence bracket in `CommitPreparedBatch` — memo dropped before the commit, memoization held off until the commit has returned — and the argument in its comment for why a one-sided release leaves a window either way; and the fence-all branch for a ledger-deleting batch, whose range delete no plan can enumerate (`WriteSet.HasLedgerDeletions` read before `Merge`).
2. `CheckCache` treating a tombstone with the right tag as a hit: the delete removed the Pebble row in the same apply, so coverage-only is what the apply should read.
3. Whether any write path to an attribute row bypasses `WriteSet.Merge` (and therefore the bloom filter) — that would be the one reason to keep the index bloom exemption.

## Known concerns

None.


internal/infra/cache/cache.go
```diff
@@ -246,11 +253,11 @@ func (a *AttributeCache[T]) CheckCache(at uint64, k attributes.U128) CacheStatus
 		// gen holds the key: the FSM apply path reads via Get which
 		// falls back to Gen1, so the caller only needs to know "in
 		// cache anywhere".
-		if _, ok := a.gen0.Load().Get(k); ok {
+		if e, ok := a.gen0.Load().Get(k); ok && e.Tag == tag {
```
NumaryBot: 🟠 **[major] Treat a foreign Gen0 entry as shadowing Gen1**

When Gen0 contains the same U128 with a foreign tag and Gen1 contains the requested tag, this falls through and reports `CacheHit` from Gen1. At apply, `AttributeCache.Get` returns the Gen0 entry without falling back, so `KeyStore.Get` reports a collision rather than reading Gen1. Return `CacheMiss` whenever Gen0 is present but its tag differs, allowing admission to seed the requested value.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1945#issuecomment-5586346765

internal/infra/cache/cache.go
Azorlogh: Fixed in e3eb8e4ab3d3757c05b4c2d75b88dbabf3539b24: in the same-generation case a Gen0 resident now decides (hit if its tag matches, miss otherwise) and Gen1 is consulted only when Gen0 holds nothing, mirroring `AttributeCache.Get`. Pinned by the new assertion at the end of `TestAttributeCache_CheckCache_ForeignTagIsMiss` (foreign Gen0 + matching Gen1 → miss), red on the previous logic. Note this is a consistency fix: on such a collision the seed cannot displace the live Gen0 resident either (`putAttributeIfAbsent`), so the apply still ends in `KeyStore`'s collision path — `CheckCache` just no longer promises a value the apply cannot read.

Azorlogh:
Head e3eb8e4ab3d3757c05b4c2d75b88dbabf3539b24 — addresses NumaryBot's [Gen0 shadowing finding](https://github.com/formancehq/ledger/pull/1945#discussion_r3958692535); `CheckCache` now mirrors `Get` exactly (Gen0 resident decides, Gen1 only when Gen0 has nothing). Reviewed locally before pushing (independent reviewer, APPROVE, no findings).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1945#issuecomment-5586346765

Azorlogh:
Head 4bb7810b1c9f22f5e67e1225c68b666400203181: adds the `MockPreloadReleaser` that `go generate` derives from the new `state.PreloadReleaser` interface (`internal/infra/state/notifier_generated_test.go`); the Dirty job caught the missing regeneration. No other change.

Azorlogh:
Re shipfox's [note](https://github.com/formancehq/ledger/pull/1945#issuecomment-5586735791): taken in 32bb09f29c6894def78e40bc0c810a277b3780ec — the `Cache.resetSeq` field comment now states it is process-local, admission-side only and never an input to the apply path (invariant #2). Comment-only change.

internal/infra/state/machine.go
```diff
@@ -756,6 +783,16 @@ func (fsm *Machine) CommitPreparedBatch(ctx context.Context, pb *PreparedBatch)
 
 	pb.batch = nil // committed, prevent double-close
 
+	// The writes are in the store: a preload built from here on must read
```
NumaryBot: 🔴 **[blocker] Close the post-commit invalidation window**

When an admission calls `LoadOrWait` after `pb.batch.Commit()` has made the write visible but before this loop acquires the loader shard, it can still return the pre-commit memoized value. Thus the stated guarantee that no memoized load outlives a commit is not established; invalidation needs synchronization/versioning spanning the commit rather than only a subsequent delete.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1945#issuecomment-5586346765

internal/infra/state/machine.go
Azorlogh: Addressed in a12b238af (https://github.com/formancehq/ledger/pull/1945/commits/a12b238af1acdb18e8a0c9cde684dccee0b3a4a7): the commit is bracketed by a fence on the covered keys, dropped before `pb.batch.Commit()` and lifted as soon as it returns, and no load that starts, runs or completes under the fence is memoized, so the post-commit window is gone. A ledger-deleting batch fences every loader, since its range delete covers keys no plan enumerates. Pinned by `TestMachine_CommitFencesCoveredKeysAcrossTheCommit`, `TestAttributeLoader_FenceBlocksMemoizationUntilUnfence` and `TestAttributeLoader_LoadStartedUnderFenceIsNotMemoizedAfterUnfence`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1945#issuecomment-5586346765

Azorlogh:
Re shipfox's [review of a12b238af](https://github.com/formancehq/ledger/pull/1945#issuecomment-5587948756):

1. The mockgen directive runs in `-source machine.go` mode, which mocks every interface in the file regardless of the trailing `. Notifier` argument. `MockPreloadInvalidator` is generated output: `go generate ./internal/infra/state/` reproduces the file byte-exact, and the Dirty job (which runs that generation) is green. It is unused because `machine_test.go` needs an ordering probe at each fence, which a gomock recorder cannot express.
2. A filter denial has always been authoritative for the eleven other attribute types; index keys now take the same path. Observing a false negative in-band would require the Pebble read the filter exists to skip, and a Bloom filter has no false negatives once every key is added, which `AddCanonicalKeys` does before `batch.Commit()` (verified in the review above). No change.

gfyrag (changes requested):
Putting a reject to make sure I have the opportunity to review before merge ;)

internal/bootstrap/module.go
```diff
@@ -703,6 +703,12 @@ func Module() fx.Option {
 			func(lc fx.Lifecycle, bindings network.Bindings) {
 				lc.Append(releaseBindingsHook(bindings))
 			},
+			// Loader coherence: a proposal's commit is bracketed by a fence on
+			// its covered keys in the admission loaders (see
+			// state.PreloadInvalidator).
```
paul-nicolas, line 708: [P2][blocking] Cover the fence wiring — the whole mechanism can be disconnected from production with no failing test

Re: the `fx.Invoke` at lines 706-711.

**What's wrong.** This invoke is the only thing that installs the fence in production, and nothing tests it. I deleted the entire block in a throwaway worktree at `a12b238af`:

```
go build ./...                                                   -> rc=0
go test -count=1 ./internal/bootstrap/... ./internal/infra/state/... ./internal/infra/plan/...
  ok  internal/bootstrap    3.316s
  ok  internal/infra/state  35.912s
  ok  internal/infra/plan   1.131s
```

Green. With that block gone `fsm.preloadInvalidator` stays nil on every node, `CommitPreparedBatch` skips the fence entirely (machine.go:802), and the PR is a complete no-op in production — the EN-1979 race is fully back.

**Why it matters.** The two halves are each well covered: `TestMachine_CommitFencesCoveredKeysAcrossTheCommit` proves the Machine calls fence/unfence around the commit against a real store probe, and `TestBuilder_FencePreloaded_DropsAndBlocksTheMemo` proves Builder->loader behaviour. Only the seam between them is untested, and it is a single unreferenced `fx.Invoke` — the easiest thing in the change to lose in a future refactor of `Module()`, with the loudest consequence. `*plan.Builder` satisfying `state.PreloadInvalidator` is compile-checked; the invoke's *presence* is not. There is no `Module()` wiring test in `internal/bootstrap` (the only `fx` test there is `join_preflight_test.go`).

**Fix.** Either an fx test that builds `Module()` and asserts the machine came out with a non-nil invalidator, or an integration test that drives a real `plan.Builder` + `Machine`: memoize an absent index through the builder's loader, commit a `CreateIndex` proposal through the machine, then resolve the same key again and assert a real store load. Codex raised the same gap independently.

**Resolved when.** Deleting this invoke makes a test fail.

internal/infra/preload/loader.go
```diff
@@ -146,15 +189,95 @@ func (al *AttributeLoader[T]) LoadOrWait(key attributes.U128, boundary, cacheEpo
...
+// range-deletes keys no plan enumerates (a ledger deletion).
+func (al *AttributeLoader[T]) FenceAll() {
+	for i := range al.shards {
+		s := &al.shards[i]
+
+		s.mu.Lock()
+		clear(s.loaded)
+
+		for _, inflight := range s.loading {
+			inflight.released = true
```
paul-nicolas, line 249: [P2][blocking] Cover FenceAll's in-flight marking loop — it is the only guard on the DeleteLedger path and no test touches it

**What's wrong.** These three lines are the sole reason a load that started *before* `FenceAll` and completes *after* `UnfenceAll` is not memoized: at completion the guard at loader.go:176 sees `s.fencedAll == 0` (UnfenceAll already ran), so only `!inflight.released` refuses the memo. Removing just the marking loop leaves the suite green — verified in a throwaway worktree at `a12b238af`:

```
go test -count=1 ./internal/infra/preload/... ./internal/infra/plan/... ./internal/infra/cache/...
  ok  internal/infra/preload  0.743s
  ok  internal/infra/plan     1.053s
  ok  internal/infra/cache    1.078s
```

The per-key equivalent *is* covered — `TestAttributeLoader_LoadStartedUnderFenceIsNotMemoizedAfterUnfence` lifts the fence from inside `loadFn`. `TestAttributeLoader_FenceAllBlocksEveryKey` never calls `FenceAll` with a load in flight: it calls `UnfenceAll` from inside a `loadFn` whose inflight was already marked at registration by `released: ... || s.fencedAll > 0` (loader.go:160), so the loop is never the thing under test. Every *other* fence mutation is caught (drop `dropLocked` from `Fence` -> 2 failures; drop `s.fencedAll++` -> 2; drop the `UnfenceAll` decrement -> 2; drop `!inflight.released` from the memoize guard -> 3).

**Why it matters.** This is the ledger-deletion path. `batch.DeleteLedgerData`'s range delete purges Pebble and leaves cache-resident rows in place (`processDeleteLedger`, processor_ledger.go:136-152), so a covered key is a `CacheMiss` at admission and goes through the loader. If this loop regresses, a load that read Pebble before the range delete committed gets memoized and served to every later preload, seeding the deleted row back. Behaviour is correct today; the regression is simply untested, which is the blocking criterion in `docs/technical/contributing/ai-review.md`.

**Fix.** Add the FenceAll counterpart of loader_test.go:551 — a `loadFn` that calls `FenceAll()` then `UnfenceAll()`, then assert the next `LoadOrWait` still has `FromLoad == true` and re-invokes `loadFn`.

**Resolved when.** Deleting the marking loop makes a test fail.

docs/technical/architecture/subsystems/fsm/preload.md
```diff
@@ -126,15 +126,15 @@ The bloom filter is consulted **here**, at the resolution layer, not below it: `
...
...y allows. Invalidation happens at two points. The FSM brackets every proposal's commit with a fence on its covered keys (`state.PreloadInvalidator`, wired to `plan.Builder.FencePreloaded` / `UnfencePreloaded` in bootstrap): `Fence` drops the memo immediately before `pb.batch.Commit()`, and no load that starts, runs or completes while the fence is up is memoized — its Pebble read may sit on either side of the write — until the matching `Unfence`, which runs once the commit has returned (or failed). So no preload is served a memoized value from before a committed write to a key the proposal covered. The one write outside a proposal's covered keys is `DeleteLedger`'s ledger-wide range delete (`batch.DeleteLedgerData`), whose keys no plan can enumerate; the commit of a batch that deletes a ledger is bracketed by `FenceAllPreloaded` / `UnfenceAllPreloaded`, which fence every loader. The proposer's `CleanupToken` releases again when its handler returns, which covers proposals that never applied. Every attribute type, the index registry included, takes the same path: a cache hit (same tag) is coverage-only, a bloom-absent key is coverage-only, and everything else loads through the loader.
```
paul-nicolas, line 137: [P2][blocking] The fence's stated coverage is wrong — writes are not restricted to the plan's covered keys

**What's wrong.** This paragraph says "The one write outside a proposal's covered keys is `DeleteLedger`'s ledger-wide range delete (`batch.DeleteLedgerData`), whose keys no plan can enumerate", and concludes "no preload is served a memoized value from before a committed write to a key the proposal covered". `deterministic-fsm.md:430` makes the same claim. The exclusion set is materially larger than that.

The fence set is built exclusively from the plan's `AttributeCoverage` entries (machine.go:1011-1015) and fenced around the commit (machine.go:793-822). But writes are deliberately *not* bound to that set — `gatedAccessor` gates `Get` and `Delete` on `CheckCoverage` and explicitly leaves `Put` ungated (accessor.go:85-91: "Put is not gated because in-batch writes to keys admission did not preload are still covered by the batch's own mutation overlay"). Two real uncovered writes to cache-attribute rows:

1. **Every `CreateTransaction`.** `processor_transaction.go` reads `boundaries.GetNextTransactionId()`, builds `txKey` from it, and does `s.TransactionStates().Put(txKey, txState)`. The id is allocated at apply, so admission structurally cannot declare it — and does not: the `LedgerApplyOrder_CreateTransaction` needs case (admission.go:1333-1350) declares `SubAttrReference`, `SubAttrMetadata` and `SubAttrLedger`, never `SubAttrTransaction` for the new id.
2. **Mirror ingest.** `processor_mirror.go` writes `TransactionStates()` and `TransactionReferences()` for the mirrored txID, while the `MirrorIngest` needs case (admission.go:1229-1293) declares only Ledger, Boundary, volumes, account metadata and `addTransactionTargetNeeds`.

`FencePreloaded` is never called for those keys.

**Why it matters.** The code still looks correct, but for a different reason than the doc gives: `loadedEntry.validFor` (loader.go:35-37) requires `stamp.Boundary >= requested.Boundary`, so a memo never crosses a generation boundary, and within one boundary a committed write is Gen0-resident (write_set.go:567) so `CheckCache` answers `CacheHit` and the memo is never consulted. The fence is belt-and-braces over that. This matters because the doc's claim is the whole stated justification for deleting the three EN-1904 exemptions **and both `assert.Unreachable` calls**. A reader who trusts "the only uncovered write is DeleteLedger" will conclude plan enumeration is a write-side invariant, which `gatedAccessor.Put` explicitly does not enforce — and the next handler that writes an attribute row outside its plan lands in the uncovered set with no assertion left to catch it. Two of my reviewers flagged this independently.

**Fix.** State the real contract in both docs: the fence covers the keys the plan enumerated; writes are not restricted to that set; memo staleness across the uncovered writes is blocked by `CacheStamp.Boundary` monotonicity in `validFor` plus Gen0 residency, not by the fence. Name `DeleteLedger` as the case where even that reasoning fails (a range delete leaves nothing resident), which is exactly why it needs `FenceAll`.

**Resolved when.** Both docs describe the mechanism that actually holds for the uncovered writes.

paul-nicolas (commented):

**Verdict: FINDINGS.** 3 blocking, 3 non-blocking posted inline; 6 non-blocking P3 listed below. Head reviewed `a12b238af` (base `47ae5b048`).

This is a strong change and I could not break the fence itself. I specifically failed to construct a fence leak (`defer unfence()` backs the panic path, the `lifted` flag makes the lift idempotent, and it runs on the error path too), a counter underflow (`Unfence` deletes at `<= 1`, `UnfenceAll` decrements only above 0), a lock cycle (the loader shard mutex is a leaf; `loadFn` runs outside it and `CommitPreparedBatch` holds no `fsm.mu`), or a determinism violation (the fence set derives from the replicated `ExecutionPlan`; `resetSeq` is admission-only and never read by apply). `HasLedgerDeletions()` really is read before `Merge` — `deletedLedgers` is populated in `Absorb` during `ProcessOrders` and cleared only by `WriteSet.Reset`. The covered-key handoff is correct in a load-bearing detail: `preparedPreloadKeys` is set to `nil`, not `[:0]`, so the next prepare cannot alias the in-flight batch's array. `CacheStamp.ResetSeq` closes a real pre-existing hole — before this PR `Cache.Reset()` did not bump the epoch, so memos survived a full cache clear. All four touched packages pass `go test -count=1` (`preload`, `cache`, `plan`, `state`).

The exemption removal is justified **for index keys specifically**: every production write to a `SubAttrIndex` row goes through `processor_index.go` or `processor_metadata_schema.go`, and all four order types declare the matching key at admission.go:1420-1456; the tag-aware `CheckCache` genuinely mirrors `AttributeCache.Get`'s Gen0-shadows-Gen1 semantics; and the bloom veto is nil-gated on `snap.Ready()`. The three deleted `resolve_arbitration_test.go` exemption tests died with the branches they pinned.

The blocking findings are not about the mechanism being wrong — they are about it being *undefended*. Two are mutation-proven test gaps in the new code, and one is an authoritative doc claim that is materially false and is the stated basis for deleting two `assert.Unreachable` calls.

### Verification

Every posted finding was verified adversarially before posting; findings that did not survive were dropped, not downgraded.

**Dropped as invalid (1):** "the deleted `resolve_arbitration_test.go` lost coverage of the surviving `firstErr`/partial-tracker error contract, violating testing.md's additive rule." It does not. I applied the same mutation (swallow every `getValue` failure so `firstErr` is never set) to **base** `47ae5b048` and `./internal/infra/plan/...` stayed green — that branch was already untested before this PR. Breaking only the *arbitration* `freshErr` branch at base does fail the deleted test:
```
--- FAIL: TestResolveCoverage_IndexArbitration_FreshReadFailureFailsResolve
    Expected error with "injected pebble read failure" in chain but got nil.
```
So the deleted test covered only the branch this PR removes; the primary load-error gap is pre-existing and unchanged, which `ai-review.md` excludes. Deleting the file was correct.

**Re-review (2 prior findings, both FIXED):**
- `cache.go` "Treat a foreign Gen0 entry as shadowing Gen1" — fixed; a Gen0 resident now decides in case 0 (cache.go:249-263) and Gen1 is consulted only when Gen0 holds nothing, mirroring `Get`. Pinned by the last assertion in `TestAttributeCache_CheckCache_ForeignTagIsMiss`. (The *case-1* half of the same change is still untested — posted inline as non-blocking.)
- `machine.go` "Close the post-commit invalidation window" — fixed; the fence now brackets the commit on both sides. I walked all four interleavings (load before/under the fence, completing before/after the lift) and each is refused memoization.

Neither was re-posted. Models: Claude (3 parallel reviewers — FSM/commit, loader/cache, resolve/docs) + Codex. No conflicts to resolve: Codex's single finding is the wiring-seam gap, which one of my reviewers reached independently, and my two reviewers agreed independently on both the missing-resolver no-op and the doc-coverage overstatement.

### P3 (not posted inline)

- `internal/infra/state/machine.go:833` — no test drives a failed `pb.batch.Commit()`, so nothing pins the explicit `unfence()` before the error return. A leaked fence-all would disable preload memoization on that node until restart (perf only; an unmemoizable load still returns the right value). `Applier.runCommitter` does not exit on commit error, so the leak would persist for the process lifetime.
- `internal/infra/state/machine.go:119-125` — the `preloadInvalidator` comment says "plan.Builder on the leader; nil on nodes without one", but bootstrap installs it from an unconditional `fx.Invoke` with a non-optional `*plan.Builder`, so every node has it and `nil` occurs only in tests. The leader-only reading implies followers commit unfenced, which would be a real hole.
- `internal/infra/plan/resolve.go` — no in-process detector survives for the stale-absence / bloom-false-negative class; the only remaining signal is the checker's async `INDEX_MISMATCH` (checker.go:790). Consider a `fsm.sentinelMode`-only probe that re-reads the key when a coverage-only entry comes off a memoized (`!result.FromLoad`) absence, keeping the Antithesis signal without a per-proposal Pebble read.
- `docs/technical/architecture/subsystems/fsm/deterministic-fsm.md:430-433` — the surviving "Both are safe because" bullets ("On success... / On error: stale boundary") were written to justify a post-apply `Release()`. Neither describes the fence, which lifts identically on success and failure; boundary staleness is `validFor`'s job. Attach them to `CleanupToken.Release()` explicitly.
- `internal/infra/state/machine.go:1611` — `HasLedgerDeletions()` is read before only one of the two `Merge` call sites (:1618, not the technical-only path at :1301). Correct today, since `deletedLedgers` is only appended by `Absorb` under `ProcessOrders`; latent if a technical update ever queues a ledger deletion. Folding the flag into `Merge`'s own contract would make it unforgettable.
- `internal/infra/state/machine.go:1012` and `:806`/`:825` — a fresh `[]preloadKey` per batch on the prepare path under `fsm.mu`, plus `2 x len(preloadKeys)` loader-shard mutex acquisitions on the serial `runCommitter` goroutine, both unbounded in the batch's total coverage size (`applyEntriesPipelined` splits only at checkpoint triggers). Worth a `batchCommitHistogram` before/after on a large-batch benchmark. Fence-all is bounded and rare by comparison (12 x 256 locks, twice per ledger-deleting batch).

internal/infra/cache/cache.go
```diff
@@ -263,7 +274,7 @@ func (a *AttributeCache[T]) CheckCache(at uint64, k attributes.U128) CacheStatus
 		// If Gen0 has the key now, post-rotation it sits in new Gen1 only.
 		// Get's gen0→gen1 fallback still surfaces it (and lazy Del promotes
 		// on tombstone if the handler deletes).
-		if _, ok := a.gen0.Load().Get(k); ok {
+		if e, ok := a.gen0.Load().Get(k); ok && e.Tag == tag {
```
paul-nicolas, line 277: [P2][non-blocking] Cover the new tag check on the case-1 branch — only the case-0 half of the changed predicate is tested

**What's wrong.** The PR adds `&& e.Tag == tag` in both `CheckCache` branches, but only case 0 is exercised. `TestAttributeCache_CheckCache_ForeignTagIsMiss` (cache_test.go:589-620) sets `currentGeneration = 0` and calls `CheckCache(5, ...)` with threshold 10, so `Gen(5,10) = 0` and every assertion lands in case 0. The only next-generation test, `TestAttributeCache_CheckCache_NextGeneration`, stores `attributes.Entry[...]{}` (Tag 0) and queries with tag 0, so the comparison is trivially satisfied and cannot distinguish tag-aware from tag-blind.

Verified by reverting this line to its pre-PR tag-blind form (`if _, ok := a.gen0.Load().Get(k); ok { return CacheHit }`) in a throwaway worktree at `a12b238af`:

```
go test -count=1 ./internal/infra/cache/... ./internal/infra/plan/... ./internal/infra/preload/... ./internal/infra/state/...
  ok  internal/infra/cache    0.663s
  ok  internal/infra/plan     1.060s
  ok  internal/infra/preload  1.127s
  ok  internal/infra/state    35.843s
```

The same revert applied to case 0 fails `TestAttributeCache_CheckCache_ForeignTagIsMiss` and `TestResolveCoverage_ForeignTagResidentLoads` — so the coverage asymmetry is real, not an artifact.

**Why it matters.** Half the changed predicate carries no coverage, so a later edit can silently restore the tag-blind verdict on the one-rotation-ahead horizon. In that state admission emits a coverage-only entry for a U128 whose resident belongs to another canonical key; at apply `KeyStore.Get` returns `ErrCollisionDetected` and `GetEntry` reports "not cached" — precisely the read this method's new doc comment promises to mirror. Non-blocking because the outcome is a loud collision error at apply, not a silent wrong read.

**Fix.** Two lines in the existing test (threshold is already 10, so index 15 is one generation ahead of `currentGeneration = 0`):

```go
assert.Equal(t, CacheHit, ac.CheckCache(15, key, 7))
assert.Equal(t, CacheMiss, ac.CheckCache(15, key, 9),
    "a colliding Gen0 resident of another key is not this key at the next-gen horizon")
```

**Resolved when.** Reverting either branch to tag-blind makes a test fail.

internal/infra/plan/builder.go
```diff
@@ -134,6 +134,43 @@ func NewBuilder(tracker *node.IndexTracker, c *cache.Cache, attrs *attributes.At
...
 }
 
+// FencePreloaded drops the memoized load of one covered key and blocks
+// re-memoization until UnfencePreloaded. The FSM calls it for every key of a
+// proposal's plan immediately before committing the batch that writes them,
+// so no preload built from then on is served a value from before the write.
+// An attrCode without a resolver is a no-op: the plan validator has already
+// rejected such entries at apply.
+func (p *Builder) FencePreloaded(attrCode byte, id attributes.U128) {
+	if r, ok := p.resolvers[attrCode]; ok {
```
paul-nicolas, line 144: [P2][non-blocking] Fail loudly when a plan attrCode has no resolver instead of silently skipping the fence

**What's wrong.** `FencePreloaded`, `UnfencePreloaded` and `FenceAllPreloaded` (lines 143-176) all silently do nothing when `p.resolvers` has no entry for the code, justified by "An attrCode without a resolver is a no-op: the plan validator has already rejected such entries at apply." That holds only while two independently maintained registries stay equal:

- `state.cacheAttrKinds` (state/scope.go:449-462), the set `validatePlan` enforces via `coverageSlotIndex[kind] < 0` (scope.go:165-170);
- `plan.Builder.resolvers` (attribute_resolvers.go:95-190).

I checked: they match exactly today — 12 codes each (Volume, Metadata, Reference, Ledger, Boundary, SinkConfig, NumscriptVersion, Transaction, NumscriptContent, PreparedQuery, LedgerMetadata, Index). Nothing enforces that equality: no test compares the two sets, and `.golangci.yaml` cannot.

**Why it matters.** Add a 13th cache attribute to `cacheAttrKinds` — so `validatePlan` starts accepting plans for it — without adding its resolver, and the commit fence is silently disabled for that attribute. That reintroduces exactly the loader-coherence hole this PR closes, with every test green and, after this PR, no `assert.Unreachable` left in `resolveCoverage` to catch the consequence. CLAUDE.md invariant #7 requires a contract-unreachable branch to surface an invariant failure rather than silently skip.

Note this file already does exactly that for the *same lookup on the same map*: `buildPreloadsAt` at lines 378-391 fires `assert.Unreachable("plan builder: no resolver registered for attrCode — extend buildAttrResolvers")` and returns an error, with the comment "Silent no-op would leave the FSM without seeded values on the apply side, violating invariant #6." The new fence hooks take the silent path that comment warns against.

**Fix.** `assert.Unreachable` (plus a logged error) on the missing-resolver branch — returning an error is not an option on a void commit-path hook, but the assertion is. And add a small test asserting set equality between the resolver keys and the FSM's gated attribute kinds, so adding a code to one registry only fails.

**Resolved when.** The branch fails loudly and a test pins the two registries together.

internal/infra/plan/resolve.go
```diff
@@ -200,41 +199,22 @@ func resolveCoverage[T interface {
...
 			// (coverage-only, no value to seed).
-			//
-			// Index registry keys are exempt: they are rare enough that the
-			// Pebble read costs nothing, and a false negative here would make
-			// every replica read an existing index as absent — the removal
-			// then reports nothing dropped while the row sits in Pebble.
-			// The load below both corrects it and proves the filter wrong.
-			bloomAbsent := bloomFilter != nil && !bloomFilter.MayContain(id)
-			if bloomAbsent && attrCode != dal.SubAttrIndex {
+			if bloomFilter != nil && !bloomFilter.MayContain(id) {
```
paul-nicolas, line 217: [P2][non-blocking] Question, not a confirmed defect: index keys now depend on bloom durability that is flushed only at rotation

This is a **question about a pre-existing mechanism** that this PR newly depends on — I could not close it, and I am explicitly not calling it a defect.

**Context.** Index keys now honour the bloom veto (this line). The comment this PR removed justified the old exemption on exactly this risk: "a false negative here would make every replica read an existing index as absent — the removal then reports nothing dropped while the row sits in Pebble." So the veto's soundness for index keys is worth stating.

**What I confirmed.** The *feed* is sound. The only production caller of `FilterSet.AddCanonicalKeys` is machine.go:1645-1647, after `buffer.Merge` and before `batch.Commit()`; `BloomUpdates.Indexes` is populated inside `Merge` (write_set.go:567). No write to an index row bypasses `Merge` on the apply path — this answers the PR's own review focus #3 in the affirmative. Restore/snapshot paths set readiness false (cache_snapshotter.go:411-418) and `Builder.bloomFilter` returns nil on a non-ready snapshot (builder.go:323-334), so nothing vetoes while repopulating, and incremental restore purges the persisted bloom range (attributes/prepare.go:122-138).

**What I could not close.** `PersistDirtyBlocks` has exactly one production call site — the rotation branch at machine.go:598-604 — and it clears the dirty bitmap after writing (bloom.go:125-127). The applied index, by contrast, is written every batch (`SetAppliedIndex`, machine.go:735). On boot, `restoreBloomFilters` (cache_snapshotter.go:734-782) takes `RestoreFromStore` whenever blocks exist and the config has not drifted, and `RestoreFromStore` (bloom.go:521-538) ORs the persisted blocks back with no applied-index watermark comparison. On that reading a crash between two rotations leaves the restored filter missing every key written since the last rotation, while the applied index says those entries need no replay. I did not find a compensating mechanism, but I also did not prove there is none.

**Bounding today's exposure.** It is pre-existing and identical for the eleven types that already honour the veto (references and volumes among them) — so if the reading were right, those would already be affected, which is decent circumstantial evidence that something compensates. And the index filter is off unless an operator sets `bloom-indexes-expected-keys`: `cmd/server/server.go:1062-1064` returns nil at 0, and nothing in the repo sets `BloomIndexes` outside `resolve_index_keys_test.go:100`.

**What would resolve this.** A note in the PR discussion (not necessarily code) confirming a post-crash `RestoreFromStore` cannot publish a filter missing keys written after the last rotation. If it can, that is a separate bloom-subsystem ticket — and this PR should say so rather than silently extend the veto to a twelfth type with the assertion that caught it now deleted.

internal/infra/plan/resolve.go
```diff
@@ -316,54 +283,10 @@ func resolveCoverage[T interface {
...
-				// concurrent write populated the cache between admission
-				// and apply, Get's gen0→gen1 fallback will surface it at
-				// apply time (bounded by CacheUnreachable at ≥2 rotations).
+				// Pebble had no value — coverage-only entry. If a concurrent
+				// write populated the cache between admission and apply,
+				// Get's gen0→gen1 fallback will surface it at apply time
+				// (bounded by CacheUnreachable at ≥2 rotations).
 				plans = append(plans, slab.appendCoverage(id, tag, attrCode))
 			})
 		}
```
paul-nicolas, line 292: [P3][non-blocking] One test would restore the error-propagation coverage that `resolve_arbitration_test.go` took with it

Re: the `if firstErr != nil { return &resolveResult{tracker: tracker}, firstErr }` block at lines 297-305 — snapped here, the nearest commentable line in this file's diff.

Correcting my own earlier report on this: I first dropped this as invalid, on a base-vs-head mutation that tested where `firstErr` is *set* rather than where it is *returned*. Redone properly, mutating the return to yield a `nil` error is caught at base (`TestResolveCoverage_IndexArbitration_FreshReadFailureFailsResolve` fails) and silent at head. So this return statement did lose its only test.

Two things keep it at P3 rather than blocking:
- the trigger the deleted test used was the arbitration re-read, which this PR removes from production. The *surviving* trigger — a `getValue`/`LoadOrWait` failure at lines 250-256 — was never covered, at base or head, so no production regression trigger was replaced and `testing.md`'s additive rule holds in its own terms.
- the behavior the comment at lines 298-303 calls load-bearing (returning the partial tracker so the caller's cleanup token still releases keys that loaded before the failure, "otherwise transient Pebble or marshal failures leave loader entries pinned for the rest of the process's lifetime") was **never** asserted: mutating the return to `&resolveResult{}` is green at base and at head. The deleted test discarded the result with `_, err := resolveCoverage(...)`.

**Fix.** One case in `resolve_index_keys_test.go` alongside `countingGetValue`: a `getValue` that fails on the Nth key, asserting `require.ErrorIs` on the injected error **and** that the returned `resolveResult.tracker` still carries the ids that loaded before it. That closes the pre-existing gap and the lost coverage in one test, and pins the guard the comment documents.

**Resolved when.** Both mutations above fail a test.

paul-nicolas (commented):
## Correction to my previous review — one dropped finding was partly right

In [my review above](https://github.com/formancehq/ledger/pull/1945#pullrequestreview-5151296955) I reported dropping a finding about `resolve_arbitration_test.go`'s deletion losing error-path coverage, and I cited a base-vs-head mutation as proof it was invalid. **That proof tested the wrong site.** I mutated where `firstErr` is *set*; the finding was about the shared `firstErr != nil` *return*. Re-running it properly at both revisions splits three ways:

| Property | Base `47ae5b048` | Head `a12b238af` | |
|---|---|---|---|
| `firstErr` set from a `getValue`/`LoadOrWait` failure — the surviving production trigger | untested | untested | pre-existing, unchanged |
| the shared `if firstErr != nil` return propagates the error | **covered** | **untested** | **coverage lost** |
| the partial tracker returned alongside the error (the unbounded-loader-growth guard at resolve.go:298-303) | untested | untested | pre-existing, unchanged |

Middle row, mutating `return &resolveResult{tracker: tracker}, firstErr` to return `nil` error:

```
head -> ok  internal/infra/plan  0.740s
base -> --- FAIL: TestResolveCoverage_IndexArbitration_FreshReadFailureFailsResolve
             Expected error with "injected pebble read failure" in chain but got nil.
```

Bottom row, mutating the same return to `&resolveResult{}` (drop the tracker, keep the error): green at **both** revisions — the deleted test discards the first return value (`_, err := resolveCoverage(...)`), so it never asserted the tracker at all.

**Net:** the deleted test did incidentally cover the surviving error-propagation return, reachable at base only through the arbitration branch this PR removes. So `testing.md`'s additive rule is not violated in its own terms — no *production trigger* lost its regression test, and the specific behavior that rule protects (the partial-tracker guard) was never covered either way. But a surviving error path did go from mutation-caught to mutation-silent, which is a real if narrow regression.

**Revised classification: P3, non-blocking** — not the blocking finding my reviewer filed, and not the clean rejection I posted. It does not change the merge decision; the three blocking findings in my previous review stand unchanged. The fix is one test, noted inline.

Nothing else in that review is affected. Credit to the reviewer that pushed back.

PR #1947, opened 2026-09-08
fix(logs): align ledger-log JSON output with v2 (EN-1790)

gfyrag:
## What changed

Ledger-log JSON output follows the v2 layout: one `type` discriminator and the payload directly in `data`, without a second oneof wrapper. Each of the 13 v3 variants has a distinct name. Metadata logs use `targetType` and `targetId` (account string or uint64 transaction ID, including zero).

HTTP get/list logs, prepared-query `logData`, JSON events and ledgerctl JSON/YAML share this output projection. Tests assert the emitted envelopes, payloads, exact large numeric values, null metadata and colored volumes. Documentation and OpenAPI describe the output and remaining v2/v3 differences.

## Scope and implementation

EN-1790 is scoped to emitted JSON. Re-reading responses into internal Go log types is not a requirement: this PR adds no payload decoders or round-trip guarantee. The unused HydrateLog helper and its sole caller LedgerLog.UnmarshalJSON are removed, together with their decoding-only tests and fuzz target. Log-type constants now live next to LogType; output exhaustiveness checks still cover all 13 variants.

The selected protobuf oneof message is serialized directly through its custom marshaler or protojson. Existing explicit discriminator switches remain, covered by all 13 names and complete oneof coverage. No additional codec dependency or representation layer is introduced.

## Differences from v2 and risk

MEDIUM: intentional JSON format change for unreleased v3. V3 retains typed metadata, colored volumes, `revertedTransactionId` plus `revertTransaction`, and the outer global-log/event envelopes. No protobuf definitions, persisted state, audit hashing, response signatures or FSM behavior change in this PR. The output is not an audit replay/backup format. ClickHouse/Databricks keep their separate analytical projection.

## Bugfix evidence

DISCOVERY: DIRECT_FIX — emitted v3 logs had redundant wrappers, shared discriminators and omitted zero transaction targets.
BEFORE_FIX: BUG_REPRODUCED — three output-contract cases fail on base b9d8ccbbb: created-transaction direct payload, metadata targetId zero, and the distinct FILL_GAP discriminator. This evidence tests emitted JSON, not hydration.
AFTER_FIX: PASS — the output matrix and complete codec/HTTP/events/CLI suites pass after removing the added decoders; final rebased-candidate results follow below.
BASELINE_CLASSIFICATION: UNKNOWN — the historical intermittent Sonic/profiling issue remains separately investigated; no independence or fix is claimed.

## Validation

On candidate de2886a83289cd5545ba80044d41117ba04f3cf9, based on 8d38782c44b012101961cfb732322729a5163208:

- PASS: canonical agent-check (invariants, root compilation, whitespace).
- PASS: complete commonpb, HTTP, events and ledgerctl/cmdutil suites with the concurrency detector in pinned Nix (GOMAXPROCS=2, GOFLAGS=-p=2).
- Historical PASS before the final rebase and unused-helper removal: exact-base normalization, generation/tidy clean and both linters report zero issues. The current removal passed baseline (including fuzz inventory) and all four affected suites; fresh normalization is part of the pending native run.
- Static self-review: output numeric assertions strengthened after removing hydration checks; no outstanding finding in the scoped simplification.
- Pending: CI and native final-review workflow on this exact candidate. The run for the preceding 7717b2d42 was intentionally stopped during validation after the user requested removal of the unused helper; it ended VALIDATION_FAILED by termination with ROOT_UNCHANGED PASS. Older bot approvals are not approval of this SHA.

Historical evidence: CI34372122681 passed all applicable checks on 84795956c. Its native local loop passed agent-check-full, then was intentionally stopped during E2E compilation under an operational disk threshold the user has since removed; the loop ended VALIDATION_FAILED before technical review. A prior candidate exposed an intermittent synchronization issue involving Sonic and runtime profiling, under investigation separately. This PR does not claim to fix that issue or adopt a replacement JSON library.

## Review focus

Direct payload shape; unique discriminators; targetId type and zero preservation; unchanged transaction/color projection; independent output assertions across consumers. The contract is documented in docs/technical/architecture/subsystems/api/http-api.md#ledger-log-json-output, with the v2/v3 comparison in docs/technical/contributing/api-comparison.md.


gfyrag:
Addressed Shipfox's malformed-payload assertion finding in baeaaa0581. The three cases now require distinguishing error fragments: `uint64` for the transaction ID decoder, `invalid value for enum field type: "NOT_A_TYPE"` for the schema enum, and `[]*commonpb.VolumeEntry` for the post-commit volume-list shape. An unrelated envelope/type rejection can no longer satisfy those assertions.

Validation: `TestHydrateLogRejectsMalformedPayload` with `-race` in the pinned Nix environment and `bash scripts/agent-check` both passed. This follow-up changes only three expected error strings; production code and emitted JSON are unchanged. The preceding HEAD's remote CI was fully green, including business E2E. CI is rerunning for this test-only commit.

gfyrag:
Reviewed both non-blocking maintainability comments on `4c92607e`.

- The four explicit discriminator mappings follow the existing implementation. `TestLogTypeJSONNames` pins all 13 public names, and the round-trip matrix checks complete oneof coverage and unique discriminators. Keeping these explicit avoids adding a registry/reflection abstraction for this bounded change, consistent with the requested implementation simplicity.
- The consumer fixtures intentionally pin each surface's concrete envelope independently. HTTP, prepared-query, event and CLI envelopes differ; sharing the decoding shape would reduce that independent contract coverage and require cross-package test plumbing.

No code change for these two comments. The accepted JSON contract is documented in `docs/technical/architecture/subsystems/api/http-api.md#ledger-log-json-hydration` and the PR body, including direct `data`, 13 types and `targetType`/`targetId`.

Validation status remains separate from bot approval: the native loop stopped at `VALIDATION_FAILED` in Antithesis runner fixtures before invoking its reviewer. CI run 34330355043 also failed the unit job on a Sonic JIT/pprof data race; all E2E, Model and Schemathesis jobs passed. Both validation failures are being investigated; this is not a ready-to-merge claim.

gfyrag:
Reviewed both comments against `84795956c`.

- Explicit discriminator switches: retaining the existing pattern for this bounded change. The independent name pins and exhaustive oneof matrix cover drift; the implementation-simplicity decision remains documented in the PR body.
- Decoder strictness: confirmed that custom nested/leaf decoders can ignore unknown fields. The PR body overstated the scope with “malformed payloads are rejected”; it now names the actual checks and explicitly says hydration is not recursive schema or business validation. This response-codec round-trip change does not introduce a general validation layer.
- The specific `SET_METADATA` / `data: {}` example does not match the current code: `SavedMetadata.UnmarshalJSON` calls `logMetadataTargetJSON.target()`, which rejects an absent `targetId` with `metadata targetId is required`. The existing missing-target regression pins that branch. Unknown fields inside `Transaction.UnmarshalJSON` are a confirmed limitation and are now disclosed.

No production-code change for these non-blocking comments. Verification here was source inspection plus the existing current-candidate suite results, not a new executable reproduction. Current CI 34372122681 is entirely green. The local native loop passed `agent-check-full`, then was stopped during E2E compilation because free disk fell below 50 GiB; its final `VALIDATION_FAILED` termination and pending native technical review remain disclosed.

internal/proto/commonpb/common.pb.json.go
```diff
@@ -298,50 +298,43 @@ func (x *RevertedTransaction) MarshalJSON() ([]byte, error) {
...
+	return json.Marshal(&struct {
+		TargetType string `json:"targetType"`
+		TargetID   any    `json:"targetId"`
+		Key        string `json:"key,omitempty"`
+	}{TargetType: x.GetTarget().AsConst(), TargetID: targetID, Key: x.GetKey()})
+}
+
+func logMetadataTargetID(target *Target) (any, error) {
+	switch v := target.GetTarget().(type) {
 	case *Target_Account:
```
NumaryBot, line 331: 🟠 **[major] Reject a nil account target instead of emitting an empty ID**

When a target selects `Target_Account` but its nested `Account` message is nil, `GetAddr()` silently returns `""`, so the new encoder emits a seemingly valid `targetType: "ACCOUNT"` with an empty `targetId`. Treat this malformed oneof like the missing-target case rather than concealing the invalid state, as required by [AGENTS.md:50](/data/worktrees/formancehq__ledger-pr-1947/AGENTS.md#L50).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1947#issuecomment-5588097907

internal/proto/commonpb/common.pb.json.go
gfyrag: Fixed in a3f1ff37a. Both SavedMetadata and DeletedMetadata now reject a selected account target whose Account message is nil, with a distinct error. The regression fails for both operations before the fix; the complete commonpb suite passes with -race after the fix. Canonical agent-check passes. Broader exact-base validation is still running.

gfyrag:
Addressed the current Shipfox documentation finding in 1f4d5bb76: the HTTP contract and OpenAPI now explicitly describe FILL_GAP.originalId as a decimal uint64 JSON string, versus JSON-number transaction IDs and metadata integers, including the need to preserve integer precision. Encoding behavior is unchanged. NumaryBot account-target finding is fixed separately in a3f1ff37a with before/after regression evidence. Canonical agent-check and the full commonpb suite with -race pass; exact-base extended validation and new-head CI remain pending.

internal/proto/commonpb/common.pb.json.go
```diff
@@ -298,50 +298,47 @@ func (x *RevertedTransaction) MarshalJSON() ([]byte, error) {
...
-		aux.AccountId = v.Account.GetAddr()
+		if v == nil || v.Account == nil {
+			return nil, errors.New("missing metadata account target")
+		}
+
+		return v.Account.GetAddr(), nil
 	case *Target_TransactionId:
```
NumaryBot, line 337: 🟠 **[major] Reject a nil transaction target instead of panicking**

When `Target.Target` contains a typed-nil `*Target_TransactionId`, this branch matches and dereferences `v`, causing JSON serialization to panic. Handle `v == nil` like the malformed account and missing-target cases so the encoder returns an invariant error instead.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1947#issuecomment-5588097907

internal/proto/commonpb/common.pb.json.go
gfyrag: Fixed in 46128d3c2. A typed-nil transaction target now returns the distinct missing metadata transaction target error. Added independent SavedMetadata and DeletedMetadata cases while preserving the account-target regressions. Complete commonpb tests with -race and canonical agent-check pass. The preceding extended local check encountered actual ENOSPC; it is not claimed as passing.

gfyrag:
Reviewed Shipfox’s low-severity observation on the empty-payload test: agreed that this individual case does not distinguish the two guards. Retaining it as input coverage; the separate nil-Data and nil-selected-message cases independently pin each guard, as the review confirms. No production change is needed for this non-blocking observation. Latest follow-up 46128d3c2 fixes the typed-nil transaction-target finding; commonpb -race and agent-check pass. Extended local validation of the preceding head encountered ENOSPC, while remote CI passed; fresh CI and bot reviews are pending.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1947#issuecomment-5588097907

paul-nicolas (approved):

PR #1950, opened 2026-09-09
revert(EN-1904): drop the index-registry preload special cases and assertions (#1830)

Azorlogh:
## What changed

Reverts #1830 (EN-1904, squash `f36913ac3`): the index-registry special cases in `resolveCoverage` — the bloom-veto exemption, the cache-hit bypass, the arbitration Pebble re-read — and the two assertions it introduced, `index preload served stale absence` and `index preload bloom false negative`, together with `resolve_arbitration_test.go`. Index keys resolve like every other attribute again.

## Why

The assertion fired in Antithesis run `5572460b288ddd058df2b44d04d9f991-59-13` (two overlapping `CreateIndex` bulks for one index). Investigating it under EN-1979 showed the race it guards is harmless: at apply the FSM cache is authoritative, a key written at generation g stays cache-resident through g+1, a preload memo is valid only for plans at or before the boundary it was loaded for, and seeds never displace a resident. So a stale memoized absence is always ignored at apply, the outcome is correct on every replica, and no client-visible or checker-visible effect exists. The other two "holes" #1830 closed are not reachable either: every `SubAttrIndex` write goes through `WriteSet.Merge` and the bloom filter, and the cache-hit case needs a 128-bit hash collision. The workaround cost a Pebble read on every index admission and made a benign, reachable race fail every run it occurred in. EN-1979 (PR #1945, closed) is superseded by this revert.

DISCOVERY: EN-1979 root-cause analysis of the Antithesis assertion class
BEFORE_FIX: N/A — no bug; the assertion's claim was wrong
AFTER_FIX: PASS — `internal/infra/plan` suite green without the special cases

N/A (removal of a workaround for a non-bug)

## Technical decision

Decision: plain revert. Alternatives considered: (1) keep the special cases and delete only the assertions — leaves the double read and per-type branches for no benefit; (2) make the loader coherent at commit time (#1945) — machinery to make an unneeded assertion true.

## Risk

**LOW** — restores the pre-2026-09-01 resolver for index keys; two files, no persisted format or proto change.

## Validation

- [x] `internal/infra/plan` suite green
- [x] Clean revert, no conflicts; no other file references the removed code

- Admission: index keys take the cache-hit shortcut, honour the bloom veto and load through the loader like every other attribute; no arbitration read.
- Removed: both EN-1904 assertions.

## Review focus

That the revert applied cleanly onto the current `resolveCoverage` (the surrounding code moved since `f36913ac3`).

## Known concerns

None.


gfyrag (approved):
Approved at 11f7281499fb51497c03e82603ceda9746b7df18. The diff cleanly reverts the index-specific cache/bloom bypasses, arbitration read, assertions, and associated tests introduced by #1830. It restores the generic resolver path without adding the #1945 invalidation machinery. The FSM cache remains authoritative for the reported concurrent CreateIndex scenario. No blocking findings. Validation considered: all executed CI checks passed, including unit, business/cluster E2E, scenario, model, and Antithesis workload tests; no local tests rerun.

PR #1951, opened 2026-09-09
refactor(query): parameterize the iterator algebra by direction (EN-1966)

paul-nicolas:

Preparatory refactor for #1930, in response to [gfyrag's review](https://github.com/formancehq/ledger/pull/1930#issuecomment-5598704784). **No behavior change.**

## What changed

There is now one iterator interface, `Iterator[D Direction]`, with one positioning method:

```go
type Direction interface{ compare(a, b []byte) int } // sealed: Asc | Desc
type Iterator[D Direction] interface {
    Next() bool
    Current() []byte
    Seek(target []byte) bool // first entity at or after target IN D'S ORDER
    Err() error
    Close()
}
type EntityIterator  = Iterator[Asc]
type ReverseIterator = Iterator[Desc]
```

`Direction` is a sealed pair whose only content is a comparator that orders two entities *along the direction of travel*. `And`, `Or`, `Not`, `Filter` and `Slice` are single direction-parameterized implementations in `combinator_*.go`. Only leaves whose physical Pebble traversal genuinely differs — a cursor walked `First`/`Next` versus `Last`/`Prev` — keep two implementations.

`SliceIterator` moves from `internal/query` to `readstore`, where both directions can share it.

## Why

#1930 compiles descending traversals directly, which is the right optimization, but it does so with a second 992-line compiler and a forward/reverse copy of every combinator. The review's objection is that this makes forward/reverse isomorphism *a convention rather than a property of the code*: a new filter kind can be correct in one direction and missing in the other, and the merge/convergence algorithms are exactly where maintaining two implementations is risky.

This PR removes that class of drift for the boolean algebra. It is deliberately separate from #1930 because it has **no behavior delta**, so the existing ascending suite is a complete oracle for it — reviewing a restructure and a new read path in one 5k-line diff would give up that leverage.

## Technical decision

**Decision:** direction is a *type parameter*, not a runtime flag.

**Why not one untyped `Iterator` with a plain `Seek`:** collapsing `SeekGE`/`SeekLE` into one method name would let `PaginateForward` accept a descending iterator and compile. The distinct method names were the only thing preventing that.

**Correction — a phantom parameter alone does not keep that guard.** This PR originally claimed it did, and that was wrong; NumaryBot caught it. Go decides interface satisfaction *structurally*, so with `D` absent from every method signature `Iterator[Asc]` and `Iterator[Desc]` had identical method sets and `PaginateReverse(NewSliceIterator(nil), 10, nil)` compiled cleanly. `Iterator` therefore carries `Direction() D` — a witness never called at runtime whose only job is to put `D` in the method set. The four crossings (two interface assignments, both `Paginate` calls) now fail with `wrong type for method Direction`, and a positive probe confirms correct usage still compiles.

**Cost, measured:** Go cannot infer a type argument that appears only inside a generic *interface* parameter type (`Iterator[D]` from `Iterator[Asc]` fails; `[]Iterator[D]` from `[]Iterator[Asc]` succeeds), so generic entry points taking a bare `Iterator[D]` need explicit instantiation. That is a handful of internal call sites; the named constructors (`NewOrIterator`/`NewReverseOrIterator`, …) are that boundary, and **every existing caller is unchanged**.

**Alternative considered and rejected:** an intermediate direction-neutral plan tree lowered by two walkers. Same de-duplication, but it adds a representation and a second traversal for no extra safety, and "plan" is loaded vocabulary in this repo (invariant #6 `plan.Coverage`, `internal/infra/plan`).

Two consequences worth stating rather than leaving to CI.

**`ReverseIterator` now carries `Close`,** because the unified interface does. That deletes `reverseCloser` from `list_entities.go` — and with it the LOGS branch that returned `&reverseCloser{rev, it.Close}`, closing the raw prefix iterator rather than the `FilterReverseIterator` wrapper it handed back. Equivalent today, but it bypassed the wrapper's ownership contract, which the review flagged.

**`AndIterator` positions every child on its first `Next`,** instead of advancing `children[0]` and letting `converge` seek the rest. The lazy form works ascending only by accident: an unpositioned child returns an empty `Current()`, which sorts below any candidate and so reads as "behind, seek forward". Descending, empty reads as "past the end", `converge` adopts it as the candidate, and the intersection collapses to nothing. Results are unchanged; the only observable difference is the mix of `Next`/`Seek` calls a child receives, and no test asserts those counts on a real traversal.

## Validation

- [x] `bash scripts/agent-check` — PASS
- [x] `go test -race ./...` — 98 packages ok
- [x] `golangci-lint run --max-same-issues=0` on `internal/query/...`, `internal/storage/readstore/...`, `internal/application/ctrl/...` — 0 issues

**The refactor-only claim is checkable from the diff.** The only test-file changes outside the new suites are: `SeekGE`/`SeekLE` → `Seek` renames, the regenerated mock's explicit `[readstore.Asc]` instantiation (7 lines), one `Close()` and five one-line `Direction()` witnesses on fakes the interface now requires, and one import. No test logic changed.

New test — **`combinator_direction_test.go`**. Parameterizing And/Not/Slice creates `[Desc]` instantiations that nothing in this PR consumes, so without it the PR would ship unexercised code. For five compositions (Slice, And, Or, Not, And-over-Not) it asserts the descending traversal is the exact reverse of the ascending one, and that `Seek` is absolute in both directions — idempotent, non-consuming, and still repositioning after exhaustion.

Both halves were mutation-verified rather than assumed:

| Mutation | Effect |
|---|---|
| `Desc.compare` returns `bytes.Compare(a, b)` (ascending) | kills 9 of 10 subtests |
| `SliceIterator.at` ignores `step` | kills 10 of 10 subtests |
| `Direction() D` removed from `Iterator` | kills `TestIteratorMethodSetNamesDirection` |

The first survives on `DescendingIsReversedAscending/Slice` on purpose: a slice's descending *walk* is driven by `step`, not the comparator — which is why the second mutation exists.

## Known concerns

1. **One unrelated flake aborts `agent-check-pr` locally.** `tests/antithesis/TestRunModelTestRequiresVerifiedOutcome` fails every subtest at `building (already in nix shell)... context deadline exceeded` after exactly 15.2s. It is a fixed build-timeout starved under full-suite load, not a read-path failure. Proven on the base commit itself: `go test -race ./...` at `66a9b23a5`, with none of this PR's changes, fails the same test the same way — 98 packages ok, identical single failure. It passes alone on this branch with and without `-race`, and alone on base. Called out so a reviewer does not read the gate abort as this PR's.
2. **Merge order with #1930 and EN-1965.** This PR touches `iterator_address.go`, so it needs to land before #1930 is rebased onto it. The follow-up rebases #1930 onto this shape and deletes `compile_reverse.go`.

## Review focus

- **`combinator_and.go` `converge`.** The comparisons now read from `it.cmp` rather than `bytes.Compare`. Ascending behavior must be byte-identical to the deleted `iterator_and.go`.
- **`combinator_slice.go` `at`/`Seek`.** One `sort.Search` serves both directions because `cmp(at(i), target)` is monotone over travel positions. Worth confirming that argument.
- **Whether `Direction`'s sealed unexported method is the right seal,** versus exporting it for a future third order.

## What this does not do

`compile_reverse.go` is untouched and still exists on #1930. Collapsing the two compilers into one recursion with per-direction leaf factories, plus the three-target (ACCOUNTS/TRANSACTIONS/LOGS) parity and error matrix the review asks for, is the follow-up. This PR is the algebra it will be built on.


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1951#issuecomment-5599665891

internal/storage/readstore/iterator.go
```diff
@@ -0,0 +1,94 @@
...
+// raw bytes) in D's order.
+//
+// Direction is a type parameter and not a runtime flag on purpose. It keeps
+// the ascending and descending contracts distinct types, so an ascending
+// consumer cannot be handed a descending iterator — the guard the separate
+// SeekGE/SeekLE method names used to provide before the two algebras were
+// merged (EN-1966).
+type Iterator[D Direction] interface {
+	// Next advances to the next entity. Returns false when exhausted OR when
+	// the underlying storage reported an error. Callers MUST call Err() after
```
NumaryBot, line 50: 🟠 **[major] Encode direction in the iterator method set**

Because `D` never appears in any method signature, Go's structural typing makes `Iterator[Asc]` and `Iterator[Desc]` interchangeable: for example, `var _ ReverseIterator = NewSliceIterator(nil)` compiles. Consequently an ascending iterator can be passed to `PaginateReverse` or a reverse combinator and produce incorrectly ordered/cursor-filtered results, whereas the former `SeekGE`/`SeekLE` interfaces rejected this at compile time.

paul-nicolas: Confirmed, and the finding is exactly right — including the example. Fixed in 7ccda3b1c.

I verified it before changing anything, because this claim was load-bearing in the PR description:

```go
_, _, _ = PaginateReverse(NewSliceIterator(nil), 10, nil)        // compiled cleanly
_, _, _ = PaginateForward(NewReverseSliceIterator(nil), 10, nil) // compiled cleanly
```

`go vet` exit 0. So the branch was strictly worse than the `SeekGE`/`SeekLE` names it replaced, which rejected both at compile time — and the PR body asserted the opposite.

`Iterator` now carries `Direction() D`: a witness never called at runtime, whose only job is to put `D` in the method set. Direction-parameterized types get it once as `func (it *T[D]) Direction() (d D) { return }`; concrete leaves declare the direction they actually walk.

The same four crossings now fail with `wrong type for method Direction`, and a positive probe confirms correct usage still compiles.

One thing worth flagging beyond the fix: **no ordinary compile-time assertion could have caught this.** Every `var _ EntityIterator = (*T)(nil)` line in the package compiled just as happily before as after, so all of them were silent on the property they looked like they were testing. The guard is now `TestIteratorMethodSetNamesDirection`, a source-level assertion that fails if `D` ever leaves the method set again; deleting `Direction() D` fails it with the explanation attached.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1951#issuecomment-5599665891

gfyrag (approved):
Reviewed at 7ccda3b1. The direction-parameterized iterator algebra removes the duplicated composite implementations while preserving static direction separation; targeted race tests, agent-check, CI, and integration with the current release/v3.0 base are green.

PR #1952, opened 2026-09-09
refactor(grpc)!: route audited mutations through BucketService.Apply (EN-1954)

paul-nicolas:
Closes [EN-1954](https://formance-team.atlassian.net/browse/EN-1954). Follow-up to EN-1950 / #1888.

## Problem and behavior

The five prepared-query and query-checkpoint mutation RPCs rebuilt unsigned requests and entered admission outside `BucketService.Apply`. This duplicated authentication and forwarding logic and prevented batching and request signing. All five operations now use their existing `ledger.Request` variants through Apply; ClusterService retains only its three checkpoint read RPCs and has no admission dependency.

The common checkpoint readiness wait also needs to distinguish fresh execution from an idempotent replay. Otherwise, create with key K → delete checkpoint C → retry K waits forever for C's deleted marker. See [gfyrag's finding](https://github.com/formancehq/ledger/pull/1952#issuecomment-5631242951).

## Implementation

- Remove the five dedicated RPCs and unused messages without compatibility shims. Migrate CLI, actions helpers, E2E callers, and the Antithesis checkpoint driver to Apply. CLI commands reuse action builders and verify response signatures before consuming results.
- Preserve the FSM's replay decision through admission and controllers as `domain.ApplyResult{Logs, Replayed}`. Every successful Apply response carries a server-produced `ledger-apply-replayed` trailer. Forwarders require exactly one valid value; incoming client metadata cannot suppress a readiness wait.
- A new checkpoint creation waits for the local read-index marker before response signing and payload stripping. With payloads enabled, a forwarding follower also waits for its own marker. With `skip_response`, the leader strips the checkpoint ID, so only the leader waits.
- A replay returns the original outcome immediately, including after deletion and while the first call still materializes. It allocates no new log and recreates nothing. Historical success is not a current existence/readiness guarantee.
- Concurrent deletion supersedes a fresh creation's wait. A consistent main-store snapshot proves deletion only when the row is absent and the next-ID counter has passed that ID; absence before allocation is follower lag. Progress notifications plus bounded lifecycle polling prevent a stalled index builder from stranding the waiter. Creation success is retained, and deletion cleanup may still be in progress.

The replay flag and trailer are transient execution metadata. No persisted format, audit intent, FSM coverage, or restore-state changes are introduced. A marker existence check alone was rejected because it cannot distinguish replay, replication lag, and deletion.

## Protocol and documentation

Service protocol revision **6 → 7** relative to the current target branch: RPC removal and required forwarding provenance are incompatible contract changes. EN-2009 already consumed revision 6 on `release/v3.0`. Communicating clients and servers must use the matching revision.

Durable contracts are in `docs/technical/architecture/subsystems/read-path/query-checkpoints.md`, `subsystems/admission/idempotency.md`, and `subsystems/api/protocol-compatibility.md`. API comparison, CLI documentation, subsystem navigation, and code comments are updated. HTTP response bodies and OpenAPI are unchanged.

## Validation

Validated candidate `666cc98c4`, incorporating target commit `a69f72449`:

| Check | Result |
|---|---|
| `bash scripts/agent-check` | PASS |
| Regeneration / tidy / root and operator lint | PASS; no remaining generated drift; 0 lint issues |
| Root `go test -race ./...` | All packages pass except the known Antithesis fixture below |
| `just test-e2e` (race, light feature set) | PASS: full business and cluster suites |
| `just test-schemathesis` | PASS: 62 endpoints tested; one recovered transient network error |
| Fresh independent source review | 0 blocking or non-blocking findings |

`AI_REVIEW_BASE_SHA=a69f72449 bash scripts/agent-check-pr` selected the full
normalization/baseline/root-race gate plus E2E and Schemathesis. Its root-race
stage still fails the six 15-second deadline subtests in the previously
documented `tests/antithesis/TestRunModelTestRequiresVerifiedOutcome` shell
fixture. No checks were weakened. E2E and Schemathesis were run separately and
passed because that existing failure stops the selector before those stages.

The fix is commit `7606c95f7`; target integration and the two new upstream HTTP
fixture adaptations are included in the final candidate. The target later
advanced to `76894b3da` with an unrelated query optimization: a merge-tree check
is clean and its protocol remains revision 6.

Regression coverage includes real three-node leader/follower keyed create → ready → delete → marker absent → replay, unchanged checkpoint/log IDs, no additional creation log or resource recreation, replay while the original call waits, and deletion during a fresh wait without index progress. Existing authentication, signature, multi-log, and `skip_response` regression triggers are retained. Structural AST tests enforce the single gRPC write entry point.

[EN-1954]: https://formance-team.atlassian.net/browse/EN-1954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/ledgerctl/querycheckpoint/create.go
```diff
@@ -43,7 +45,26 @@ func runCreate(cmd *cobra.Command, _ []string) error {
...
+		if spinner != nil {
+			spinner.Fail("Failed to sign request")
+		}
+
+		return cmdutil.Displayed(err)
+	}
+
+	// Apply does not return until the read index checkpoint is materialized on
+	// the serving node, so a read at the returned ID succeeds immediately there.
+	resp, err := client.Apply(ctx, applyReq)
```
NumaryBot, line 58: 🔴 **[blocker] Verify the checkpoint Apply response before using it**

When `--response-verify-key` is configured, this migrated command extracts and reports the checkpoint ID from unverified Apply logs, so a tampered or incorrectly signed response is accepted despite the global verification option. Call `cmdutil.VerifyResponseSignatures` before reading the payload; the other four newly migrated query/checkpoint commands likewise discard their Apply responses without verification.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1952#issuecomment-5619489461

cmd/ledgerctl/querycheckpoint/create.go
paul-nicolas: Done — fixed in commit de2b66322. All five migrated commands now call VerifyResponseSignatures before consuming checkpoint data or reporting success. Added command-level gRPC regression tests for missing, invalid, valid, and disabled verification across all five commands (20 cases). Baseline and PR checks, lint, and focused race tests pass.

paul-nicolas:
Addressed the response-signature finding in de2b66322 and resolved the corresponding inline thread. All five migrated query/checkpoint commands verify Apply response signatures before reporting success or consuming checkpoint data. The 20 regression cases pass, including with the race detector; agent-check and agent-check-pr also pass.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1952#issuecomment-5619489461

gfyrag:
Follow-up on reviewed HEAD `55eb50a`: the new readiness wait needs to distinguish a newly executed creation from an idempotent replay.

**[P2][blocking] Preserve checkpoint-create replay after deletion.**

Concrete sequence, before the idempotency key expires:
1. `Apply(K, CreateQueryCheckpoint)` creates checkpoint C and succeeds.
2. Another request deletes C, including its read-index `.ready` marker.
3. The client retries the identical batch with K.
4. The FSM returns the original log references without re-executing creation; admission resolves them to the historical `CreatedQueryCheckpoint` log.
5. `waitCreatedQueryCheckpoints` (`server_bucket.go:225`) waits for C's marker, which will never be recreated. The retry times out, or waits indefinitely without a deadline.

The original operation's recorded success must not become permanently pending merely because the resource was subsequently deleted.

**Suggested direction:** preserve execution provenance in the Apply result rather than infer it from log payloads or filesystem existence. The FSM already knows whether an outcome was replayed; propagate that information through admission/controller, conceptually as `ApplyResult{Logs, Replayed}`.

- A newly executed checkpoint creation retains the local-readiness wait.
- An idempotent replay returns the historical outcome without re-waiting for materialization or recreating the resource.
- Preserve this distinction across the leader→follower hop. Otherwise the leader skips the wait but the follower receives historical logs and starts waiting again. This must be server-produced information, not a client-controlled instruction. Choose the concrete transport representation deliberately; the example above is not a requirement to expose an unnecessary public API field.

Document the distinction explicitly: initial creation waits for readiness; replay returns a historical result, not a guarantee that the resource still exists or is currently ready. In particular, pin the case where a replay arrives while the first call is still waiting for materialization.

A one-time registry/filesystem existence check is weaker: follower lag can look like absence, and deletion can race the check. A timeout alone does not preserve idempotent semantics. Concurrent deletion during a genuinely new creation's wait also needs deliberate lifecycle handling; a replay flag alone does not solve that separate race.

Expected regression coverage:
- keyed create → ready → delete → marker absent → identical replay returns the original success/log IDs, without recreation or a new creation log;
- the same sequence through a follower;
- initial creation still waits for readiness;
- replay while the original creation is still materializing follows the explicitly chosen replay contract.

Validation so far: source-path tracing and independent counter-review confirm the mechanism; existing local race tests passed for grpc, auth, ledgerctl and actions. The complete create/delete/replay regression has not yet been executed.

gfyrag (approved):
Approved on this reviewed HEAD. The checkpoint replay blocker is resolved: server-produced replay provenance survives admission and leader/follower forwarding, so historical success no longer waits for a deleted marker. Added regressions cover deletion followed by replay through leader and follower, replay during initial materialization, and concurrent deletion during a fresh wait.

The documented skip_response limitation remains: initial creation guarantees readiness on the leader only when payloads are stripped. This approval does not imply identical follower-readiness guarantees across response modes.

Validation for this re-review: source and regression-test inspection plus passing CI test/coverage checks; no fresh local execution on this HEAD. The NumaryBot status was in error during the re-review, independently of the passing test checks.

PR #1943, opened 2026-09-08
fix(readstore): flush the memtable before checkpointing the read index (EN-1978)

Azorlogh:
## What changed

`readstore.Store.CreateCheckpoint` flushes the read-index memtable before `db.Checkpoint`. The comment on it and step 5 of `docs/technical/architecture/subsystems/read-path/query-checkpoints.md` now state the WAL-less checkpoint contract.

`ListAuditEntriesFrom` reads the audit head of the main-store handle it materializes from and clamps the compiled audit upper bound to it, so index rows for sequences past that head (a live handle opened before the audit projection folded, or a frozen pair whose read index was cut after its main store) are trimmed instead of failing the page.

Regressions: `tests/e2e/cluster/query_checkpoint_readindex_test.go` (3 nodes, 40 transactions and a destination-address builtin index; `ListLogs` and filtered `ListTransactions` through the checkpoint must equal the leader's live pre-checkpoint pages on every replica, and stay equal after a post-checkpoint write), `TestCreateCheckpointIncludesUnflushedRows` (a row committed just before `CreateCheckpoint` is in the frozen store), and `TestListAuditEntriesTrimsProjectionAheadOfMainSnapshot` (an audit index row past the stored head is trimmed). `listAllTransactionsFromCheckpoint` takes a nil-able filter.

## Why

The read index runs with `DisableWAL`, so a Pebble checkpoint of it contains only the flushed SSTs: every row still in the memtable when the index builder materialized a query checkpoint was silently absent from the frozen read index, while the frozen main store was complete. The amount lost was whatever each replica had not flushed yet, so the frozen read indexes of one checkpoint id diverged across replicas — Antithesis model runs reported it as `query checkpoint served a mutated page`. On the local repro, checkpoint 16's frozen read progress was 228269 / 223493 / 228454 on the three nodes against a marker at 228515, with all three frozen main stores at the identical applied index. Checkpoint reads are served locally on whichever node receives them, and the client round-robins.

With the flush, the frozen read index also carries whatever the audit indexer (its own 200 ms tick) landed between the applier's main-store checkpoint and the builder's crossing of the checkpoint log — audit sequences the frozen main store does not hold. `ReadAuditEntriesPage` treats a missing sequence as a consistency failure for the whole page, so without the clamp an index-backed audit read through such a checkpoint would fail permanently on that replica.

All three regressions are red on `release/v3.0` @ 4fba96b4f (the e2e: node 1 serves 0 of 41 logs) and green with the fix.

[...]

PR #1933, opened 2026-09-08
perf(indexer): skip backfills for empty ledgers

gfyrag:
## What changed

- Add an indexbuilder-owned durable EMPTY/NON_EMPTY tracker per active ledger incarnation.
- Promote indexes created on EMPTY ledgers immediately, even across proposal boundaries and restarts.
- Generate exhaustive HISTORY/CONTROL classification from protobuf annotations and remove CreatedIndexLog.initial/FSM BornEmptyLedgers.
- Increment the exposed service protocol revision from 7 to 8 because the remaining CreatedIndexLog fields are renumbered.
- Keep tracker, IndexVersionState, task, and cursor changes atomic, including delete/recreate, backfill, checkpoint, and restore paths.

## Why

EN-1771: creating indexes on newly created but still empty ledgers previously scheduled a replay from global cursor zero whenever CreateLedger and CreateIndex were separate proposals. On buckets with large unrelated histories this repeatedly read and decompressed millions of foreign log entries per index and replica.

## Product / operational motivation

Need: prevent unnecessary CPU saturation when provisioning indexes for fresh ledgers in a busy bucket.
Current limitation: CreatedIndexLog.initial only represented emptiness within one Apply, so later CreateIndex proposals were conservatively treated as historical.
Requirement / constraint: an index created before the ledger emits a HISTORY payload must perform zero global backfill reads; missing or incoherent tracker state must fail loudly.
Evidence: Jira EN-1771 and deterministic regressions in internal/application/indexbuilder/ledger_history_test.go.
Durable repository evidence: docs/technical/architecture/subsystems/indexer/indexes.md and docs/technical/architecture/subsystems/indexer/indexer.md.

## Technical decision

Decision: persist an EMPTY/NON_EMPTY byte in the peer read store and classify every LedgerLogPayload arm through a build-enforced protobuf annotation.
Why now / why proportionate: one byte per active ledger avoids the measured repeated global scans without redesigning log locality.
Alternatives considered: retain CreatedIndexLog.initial, rejected because proposal boundaries lose the fact; redesign per-ledger log storage, rejected as disproportionate to EN-1771.

## Risk

**MEDIUM** — changes indexbuilder boot/replay state, peer read-store lifecycle, and the unreleased v3 protobuf layout. State/version/cursor mutations share one batch and commit-failure/restart/delete-recreate paths are covered.

## Validation

[...]

PR #1889, opened 2026-09-04
feat(reads): certify projection Raft progress (EN-1946)

gfyrag:
## Summary

- add independent Raft applied-index certificates for the read and audit projections;
- publish certificates atomically with the terminal projection batch;
- make audit readiness explicit for checkpoint admission and materialization;
- preserve checkpoint applied-index provenance across incremental restore;
- wait through rebuilds, stop on explicit disabled/failed states, and generation-guard ready publication.

## Stack

EN-1946 stack 2/8. Depends on #1897 (test/en-1946-bound-cross-store-alignment).

Merge/review order: #1897 → #1889 → #1890 → #1894 → #1891 → #1893 → #1892 → #1881.

## Validation

Final base: b9080b6270942237cb2f98fe7aa4b6f90d83fc4c.
Final head: 01f5df313ad6cc82e6cfd84fcbd75ad621d721f0.
Canonical PR validation: PASS, including full race validation and business/cluster E2E.
Independent exact diff review: APPROVE (residual risk MEDIUM).

Jira: EN-1946. Do not merge automatically.

PR #1954, opened 2026-09-09
fix(dal): materialize the query-checkpoint main store atomically (EN-1978)

Azorlogh:
## What changed

**Atomic main-store materialization.** `dal.Store.CreateQueryCheckpoint` builds the query checkpoint's main store into `main.tmp`, fsyncs it, writes the `.ready` marker into the temp directory, then renames and fsyncs the parent: the final path never exists without its marker. `openCheckpointStores` requires both directories' markers before opening either store, and a marked main store that fails to open surfaces the error (damage, not lag). The marker primitives move from `readstore` to `dal`; the read index keeps its audit-generation-guarded marker path.

**Both halves finish an interrupted materialization on restart.**
- Main store: `Applier.RecoverAndReplay` rebuilds every registered checkpoint whose main store lacks its marker and whose `applied_index` equals the live store's applied index, before WAL replay. That equality holds after any crash inside the materialization: the FSM batch registering the checkpoint commits first, the applier is gated for the whole copy (later entries wait in the spool) and the spool is reset on boot, so the live store reopens exactly at `H`. Restored checkpoints and checkpoints applied at another index are left alone (never materialized here).
- Read index: a checkpoint log is now the only log of its builder batch; the batch that crosses it commits only the certificate, and the indexed + AppliedProposal cursors commit in a batch of their own once the checkpoint action has run. A builder that dies in between resumes at the checkpoint log, re-indexes nothing (the log writes no rows) and materializes again; a marked directory is a no-op.

Rebased onto `release/v3.0` after #1933 and #1952: the builder's in-process retry of a failed materialization (`materializePendingCheckpoint` at the top of `processLogs`) still runs first; with the cursor parked before the checkpoint log, the re-cross then finds the directory marked and no-ops. `WaitForCheckpoint` keeps #1952's deletion-supersedes wait. History rewritten from 14 commits to 4.

`ErrCheckpointNotReady`, the builder comments and `query-checkpoints.md` state the durable-write order of each half, the residual state after a crash, and the two cases that stay `Unavailable` on a replica.

## Why

`pebble.DB.Checkpoint` links the sstables, writes the MANIFEST — which is what makes a directory openable — and only then copies the WAL files. The applier ran it straight into the final `{id}/main` while readers polled the path, so a read landing between the MANIFEST write and the end of the WAL copy opened a main store missing every write still in the WAL — tens of thousands of transactions with the default memtable, including the ledger's own `LedgerInfo`. This is the state the Antithesis model driver reports as `query checkpoint served a mutated page`; #1943's read-index flush is a real, separate fix, but the same 3-node model run still reproduced the finding with it alone and passes with this change.

Review of the first head asked what happens when a node dies after the checkpoint is registered and before its marker exists: it reopened with the checkpoint registered and permanently `Unavailable` on that node, and two comments contradicted each other about it. Neither store actually moves past the checkpoint point while its half is unmaterialized, so both halves can finish the job on restart instead of documenting a limitation.

Jira: EN-1978.

## Regression coverage

- `internal/infra/node/query_checkpoint_recovery_test.go`: every crash residue of the main-store materialization — nothing, a partial temp copy (WAL file lost), an unmarked temp directory, a marked temp directory, an unmarked final directory — is produced from the production output and rebuilt into a marked store frozen at the checkpoint's applied index; a marked store is left untouched (sentinel survives); checkpoints at another index and restored checkpoints are skipped; `RecoverAndReplay` is exercised end to end.
- `internal/application/indexbuilder/process_logs_checkpoint_recovery_test.go`: a builder cancelled on the audit wait leaves its cursor before the checkpoint log and the restarted builder finishes the checkpoint with the promised certificate; a builder that died after the marker keeps the marked directory on the re-cross; checkpoint logs (create and delete) head their own batches.
- `tests/e2e/cluster/query_checkpoint_readindex_test.go`: a follower whose `{id}/main` is removed while stopped serves the checkpoint's pages again after restart.
- Behavioral regression: the query-checkpoint probe of the `singleton_driver_model` model driver (lands on `release/v3.0` with the EN-1625 coverage). `TestQueryCheckpointIsOpenableBeforeItIsComplete` pins the pebble write order deterministically.

All recovery tests fail on the previous head (`TestRecoverAndReplayRebuildsUnmaterializedQueryCheckpoint`: marker absent; `TestProcessLogsResumesAtCheckpointLogAfterDyingDuringMaterialization`: cursor already past the log; `TestProcessLogsGivesCheckpointLogsTheirOwnBatch`: mixed batch).

N/A

## Technical decision

The read index's marker stays after its rename: `MarkCheckpointReadyAtAuditGeneration` checks the audit generation atomically with the marker, and moving the marker into the temp directory would require the rename under the same lock. Its unmarked-final residue is discarded on the re-cross, so nothing is lost.

## Risk

**MEDIUM** — the read gate refuses (retryable `Unavailable`) where it previously could serve wrong data; startup recovery gains a step that materializes checkpoints (failure fails boot, like the `live.discard` reconcile); the builder's cursor commit moves after checkpoint actions and checkpoint logs no longer share a batch, so `LastIndexedSequence` reported by status RPCs lags by one log while a materialization waits on audit. Live reads align on the certificate, not the cursor. No persisted format or API change.

## Validation

On the rebased head (`release/v3.0` @ `ddae03bfd`):

- [x] `bash scripts/agent-check`; nix `just lint` (0 issues)
- [x] Unit: `internal/storage/dal`, `internal/infra/node`, `internal/application/indexbuilder`, `internal/storage/readstore`, `internal/application/ctrl`, `internal/adapter/grpc`, `internal/domain`
- [x] e2e: every checkpoint container in `tests/e2e/cluster` green (follower-restart spec included)
- [x] Red on the pre-fix code for the three recovery tests: `TestRecoverAndReplayRebuildsUnmaterializedQueryCheckpoint`, `TestProcessLogsResumesAtCheckpointLogAfterDyingDuringMaterialization`, `TestProcessLogsGivesCheckpointLogsTheirOwnBatch`
- [x] `TestProcessLogsRetriesCheckpointMaterializationWithoutReplayingCommittedBatch` (from #1933) re-pinned: after a failed materialization the cursor stays before the checkpoint log; every other assertion of that test — tracker persisted, later ledger held back, retry clears pending, frozen certificate and tracker state — is unchanged
- [x] 3-node model driver, 600 s, on the pre-rebase head: PASS, 13 restart cycles, no findings; 4-min smoke of this head merged with #1931 and the EN-1625 coverage (#2009): PASS, no findings, no panics
- [ ] Antithesis, model-only, 60 min, on that same composite: run `882ad864cdd88cb0fa38af88dbef94b1-60-11` (in progress)

Rebased onto #2032 (EN-2047): the node-local lease taken before the open means a committed deletion can no longer unlink under a read, so a read-only open that fails with both markers present is damage only; the raced-delete `NotFound` branch and its test are gone. A checkpoint read is served only once both halves are marked ready on that replica; otherwise it is refused with the retryable `Unavailable` (or `NotFound` after a barrier proves the id does not exist). A replica that dies at any point of either materialization finishes it on restart. Two cases stay `Unavailable` on a replica until the checkpoint is deleted and recreated: a replica that never applied the creation entry (joined through a later snapshot, or resynced from the leader after the crash), and a disabled or failed audit projection.

## Review focus

0. The rebase interplay in `processLogs`: a checkpoint-action batch no longer writes the cursor (`appliedProposalSeq` is hoisted so the non-checkpoint path keeps #1933's contract), and the cursor batch after the action moves `cursor`/`unpersistedCursor` together.
1. The two ordering arguments: the live store is exactly at `H` in `RecoverAndReplay` (`applier.go`, gated apply + spool reset), and the builder's durable cursor never passes a checkpoint log before the marker exists (`process_logs.go`, `carry` + the cursor batch).
2. `recoverQueryCheckpointMainStores` guards: `applied_index == live`, `restored_from_backup` skipped.
3. The carried checkpoint log keeps decoded fields only — the log cursor reuses one message.
4. Both markers gate `openCheckpointStores` against `resolveMissingMarker`'s retryable-vs-`NotFound` classification.

## Known concerns

None blocking. The two residual `Unavailable` cases above are stated in the readiness contract; a post-leader-resync rebuild is out of scope.

## Bugfix evidence

DISCOVERY: RELATED_BUT_DIFFERENT — #1943 (merged) fixed the read-index half; this fixes the main-store materialization window, the actual cause of the model finding, and makes an interrupted materialization of either half recoverable.
BEFORE_FIX: BUG_REPRODUCED — 3-node model run FAIL with the finding on the base and on the flush-only fix; `TestQueryCheckpointIsOpenableBeforeItIsComplete` shows the partial directory opens with a committed key gone; the three recovery tests fail on the previous head.
AFTER_FIX: PASS — model run PASS; `agent-check`, nix lint, seven unit packages, all checkpoint e2e containers green.
BASELINE_CLASSIFICATION: ENVIRONMENTAL — `agent-check-pr` fails only in `cmd/ledgerctl/auth` (2 tests) and `internal/pkg/tarutil` (3 tests): permission-denial tests that cannot fail as root; both fail identically at the base `37ccf5238` and neither package is touched by this diff. Every other package, the checkpoint e2e containers and the model run pass.


Azorlogh:
Re [Shipfox review](https://github.com/formancehq/ledger/pull/1954#issuecomment-5600743822):

1. Justification comments on the three failed-attempt `RemoveAll` cleanups — 1234b6cca.
2. Shared temp→rename→marker skeleton for both halves: deferred, as suggested — the two halves differ in the build step (DB lock scope vs flush + link-race retry) and #1889 is reshaping the builder side; a follow-up once that lands.
3. A marked-but-unopenable main store now surfaces its open error, like the read index's, instead of a retryable `Unavailable` that could never heal — 1234b6cca.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1954#issuecomment-5600647061

docs/technical/architecture/subsystems/read-path/query-checkpoints.md
```diff
@@ -19,6 +19,7 @@ Checkpoint IDs are assigned sequentially by the FSM (1, 2, 3, ...).
 
 1. Client sends `CreateQueryCheckpoint` request (via ClusterService RPC or BucketService Apply).
 2. The request is proposed through Raft consensus.
+<<<<<<< HEAD
```
NumaryBot: 🟠 **[major] Resolve the committed merge conflict**

Resolve these conflict markers and retain one accurate lifecycle description. The authoritative checkpoint documentation currently contains both competing versions plus Git conflict syntax, making the creation/readiness contract contradictory and causing `git diff --check` (and therefore canonical validation) to fail. This violates the repository's documentation-maintenance requirement in `AGENTS.md`.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1954#issuecomment-5600647061

internal/adapter/grpc/server_bucket.go
```diff
@@ -272,24 +272,28 @@ func (impl *BucketServiceServerImpl) openCheckpointStores(ctx context.Context, c
...
+	// resolveMissingMarker, which returns a retryable ErrCheckpointNotReady for
+	// a registered checkpoint (or NotFound after a barrier confirms it does not
+	// exist).
+	//
+	// Openability is not a completeness gate, so the markers are checked before
+	// the open rather than inferred from it: pebble writes the MANIFEST that
+	// makes a directory openable BEFORE it copies the WAL files, so an unmarked
+	// directory can open cleanly while missing every write still resident in the
+	// source memtable.
+	if !dal.CheckpointDirReady(readIndexPath) || !dal.CheckpointDirReady(mainPath) {
```
paul-nicolas, line 288: [P2][non-blocking] The PR's central correctness change is untested at every layer

**What's wrong.** Two behaviors change here versus the merge-base `37ccf5238`, and neither is pinned by any test in any package:

1. **readindex marked + main unmarked** → now `ErrCheckpointNotReady`. At base (`git show 37ccf5238:internal/adapter/grpc/server_bucket.go`, lines 283-293) the gate checked *only* the read-index marker, so an unmarked main store fell through to the open attempt and, being openable, **served the read**. That is precisely the incompleteness this PR exists to fix — pebble writes the MANIFEST before copying the WALs. This is the PR's central correctness change and nothing exercises it.
2. **both marked + main open fails** (line 296) → a bare `fmt.Errorf` that `convertToGRPCError` renders as `codes.Unknown`. Intentional per your comment, but it inverts the "never Unknown" contract the e2e itself asserts, which is exactly why it deserves a test that says so deliberately.

**Why it matters.** `grep -rn "openCheckpointStores" --include='*_test.go' .` returns nothing. The producer side (`CheckpointDirReady`/`MarkCheckpointReady`/builder materialization) is well covered; the *consumer* gate is not. The one e2e net — `tests/e2e/cluster/query_checkpoint_test.go:504`, which does run on every PR via `Tests-E2E-Cluster` — only reaches main-marked/readindex-unmarked, because the read index is always built after the main checkpoint. Case 1 essentially cannot occur naturally in a cluster, so no amount of e2e will catch a regression in it.

**Fix.** A package-internal table test in `internal/adapter/grpc` over `{mainMarked, readIndexMarked}`. This was prototyped and run against the real APIs — no mocks, ~70 lines for all three cases:

- Seed the local registry (otherwise `resolveMissingMarker` falls through to `ReadIndexAndWait` and nil-panics): build the key with `dal.NewKeyBuilder()` + `PutZonePrefix(dal.ZoneGlobal, dal.SubGlobQueryCheckpoint)` + `PutUint64(id)`, marshal a `raftcmdpb.QueryCheckpointState`, write it via `OpenWriteSession` (`_test.go$` is excluded from the forbidigo rule at `.golangci.yaml:239`).
- **Case 1:** real `dal.NewStore` + `s.CreateQueryCheckpoint(id)`, then remove *only* `mainDir/.ready`. This leaves a fully openable directory, so the test **fails on the base commit and passes on head** — a true regression trigger for the fix.
- **Case 2:** mark both, then truncate the main store's `MANIFEST`. Observed: `pebble: malformed manifest file` → `codes.Unknown`.

internal/storage/dal/query_checkpoint_wal_window_test.go
```diff
@@ -0,0 +1,131 @@
...
+	_, _, err = partial.Get([]byte("committed-key"))
+	require.Error(t, err, "the staged checkpoint serves a state missing a committed write")
+}
+
+// TestCreateQueryCheckpointRedundantCallIsNoOp pins that a second checkpoint for
+// the same id succeeds against the already-materialized directory. The applier
+// re-runs this while replaying the spool or the WAL after a restart, and pebble
+// rejects an existing destination, so the redundant call has to be recognized
+// and skipped — the same guard the read index half carries.
+func TestCreateQueryCheckpointRedundantCallIsNoOp(t *testing.T) {
```
paul-nicolas, line 81: [P2][non-blocking] Two ported branches lost the assertions that identify them

The main-store test set mirrors the read-index contract tests but drops the two assertions that actually pin the branch, so two production branches would survive deletion with the suite green. Both proven by mutation in throwaway detached worktrees at this SHA, and reproduced independently by two reviewers.

**1. This test does not test a no-op.** Delete the guard at `internal/storage/dal/store.go:943-945` (`if CheckpointDirReady(dir) { return dir, nil }`) entirely and this test still passes — as does the whole package. It asserts only that the second call returns no error, the same path, and a marker, and the *rebuild* path satisfies all three identically. The sibling gets this right at `internal/application/indexbuilder/checkpoint_test.go:118-126` (`TestCreateReadIndexCheckpointNoopWhenReady`) by dropping a sentinel file first:

```go
sentinel := filepath.Join(dir, "sentinel.marker")
require.NoError(t, os.WriteFile(sentinel, []byte("x"), 0o640))

again, err := s.CreateQueryCheckpoint(1)
require.NoError(t, err)
require.Equal(t, dir, again)
require.FileExists(t, sentinel, "a redundant call must not rebuild the directory")
```

Verified: passes at HEAD, fails with the guard removed (`unable to find file .../main/sentinel.marker`).

**2. The stale-temp-dir clearing at `store.go:958-960` has no test at all.** Delete those three lines and `go test ./internal/storage/dal/` is still `ok`. Nothing stages a leftover `{id}/main.tmp`, and on a first call `RemoveAll` on a missing path returns nil, so the line is never exercised. The read-index half does have this test — `TestCreateReadIndexCheckpointClearsStaleTempDir` at `checkpoint_test.go:143`. Without the `RemoveAll`, a `main.tmp` left by a crash makes every later create for that id fail permanently on that replica with pebble's `file already exists`.

This matters more than a normal coverage nit because the no-op guard is what protects point-in-time correctness: if a redundant call rebuilt instead of no-op'ing, the checkpoint would capture a later store state. The guard is present and correct — it just isn't pinned.

Repo rule: "Assertions must uniquely identify the intended branch" (`CLAUDE.md`, testing conventions).

paul-nicolas (commented):
## Multi-model review (Claude + Codex) — head `1ddf2873f`

The core mechanism is sound. The atomic build/rename/mark sequence is correct, the pebble MANIFEST-before-WAL premise the design rests on was verified in the vendored source (`pebble/v2@v2.1.4 checkpoint.go:381` writes the MANIFEST, `:394-408` copies the WALs afterwards), and the point-in-time regression the rebuild branch could have introduced **does not exist** — `SetAppliedIndex` rides the trigger entry's own write batch and commits before the checkpoint is built, so a re-crossed trigger can only ever rebuild at exactly `H`.

**Findings: 4 posted inline (all P2, all non-blocking), 9 P3 withheld. No blocking findings.**

Two of the four are test-coverage gaps proven by mutation: the production code is correct, but two of its branches — including the PR's *central* correctness change — would survive deletion with the suite green.

### Verification

Every finding was verified adversarially by a dedicated agent instructed to refute it. 6 findings entered the gate; 0 survived unchanged, 4 were corrected, 2 were rejected outright. Notable rejections and corrections:

- **Rejected** — "the delete race violates the documented `NotFound` contract": every bullet at `query-checkpoints.md:84-86` is conditioned on marker *absence*; in the race the markers were present at check time.
- **Rejected** — "the raw-error path makes fd exhaustion worse": directionally wrong. At base an EMFILE open became `Unavailable`, which `actions.GRPCRetryPolicy` retries 50x at 0.2s while amplifying fd pressure. Head fails fast.
- **Corrected** — Codex's only finding (post-rename branches leave `{id}/main` behind) was rated P2/blocking. Downgraded to P3: `CheckpointDirReady` is false for a missing *and* an unmarked directory, so adding the cleanup is **availability-neutral** — it buys disk space only. Demoted to the P3 list below.
- **Corrected** — "three behaviors changed in the read gate": only two did. Main-marked/readindex-unmarked is not a change (the base already gated on the read-index marker) and *is* covered by the e2e at `tests/e2e/cluster/query_checkpoint_test.go:504`, which runs on every PR via `Tests-E2E-Cluster`.
- **Corrected** — "doc line 24 contradicts line 86": not as phrased (different directory states). Reframed as a premise conflict — see the inline comment.

Also refuted, with evidence: no other path materializes `{id}/main` (backup/restore explicitly does not carry the physical directories, and the restored-cluster outcome is identical on base and head); `DeleteQueryCheckpointFiles` does reclaim `{id}/main.tmp`; `CreateQueryCheckpoint` is not on the deterministic FSM apply path, so invariant 2 is intact; and the deleted `TestStore_CreateQueryCheckpoint_ErrorOnExistingDir` was a legitimate removal — the branch it pinned no longer exists.

NumaryBot's earlier finding (committed merge-conflict markers in `query-checkpoints.md`) is **fixed** at this head.

Validation run: `go build ./...` (exit 0), `go vet` on dal/readstore/grpc/indexbuilder (exit 0), `go test ./internal/storage/dal` (ok), plus 6 production-code mutations and 6 probe tests in throwaway detached worktrees.

### P3 (not posted inline)

- `internal/adapter/grpc/server_bucket.go:296` — a checkpoint deleted between the marker check and the open now returns `codes.Unknown` instead of `NotFound`. Both codes are permanent, so the client's retry decision is unchanged; the cost is lost `NotFound` semantics plus Error-level internal-error log noise. Note the obvious fix (re-stat the marker on failure) only narrows the window — `os.RemoveAll` deletes children in readdir order, so `.ready` can outlive the MANIFEST.
- `internal/storage/dal/store.go:985,991` — the two post-rename failure branches return without reclaiming the renamed `{id}/main`, unlike the three pre-rename branches (`:966`, `:974`, `:981`). Disk-only, bounded by `QueryCheckpointLimit`, released on delete. Either reclaim `dir`, or say in the comment why it is deliberately left (as the read-index half does at `process_logs.go:604-607`) — the sentence at `store.go:933-934` currently over-claims "discarded and rebuilt".
- `internal/storage/dal/query_checkpoint_wal_window_test.go:73` — `require.Error` also passes on a corruption error, inverting the test's stated claim that the data is gone *silently*. The value is `pebble.ErrNotFound`; `require.ErrorIs` costs nothing.
- `internal/storage/dal/query_checkpoint_wal_window_test.go:100` — `TestCreateQueryCheckpointRebuildsUnmarkedDir` does not assert the stale content is discarded. The branch is pinned only indirectly (via `os.Rename` refusing a non-empty destination); the indexbuilder sibling writes a `stale.sst` and asserts it is gone.
- `internal/storage/dal/store.go:935` — `CreateQueryCheckpoint` now has no failure-path test at all; none of its seven error returns is reached by any test. One line restores it: `require.NoError(t, s.Close())` then `require.ErrorIs(t, err, ErrStoreClosed)`.
- `internal/storage/dal/checkpoint_marker.go:1` — the new file ships with no test in its own package; `dal.CheckpointDirReady`/`MarkCheckpointReady` are exercised from `package readstore`, and `dal.FsyncDir` has no direct test anywhere.
- `docs/technical/architecture/subsystems/read-path/query-checkpoints.md:54` — "(steps 4 and 5)" should be "(steps 4 and 6)". Step 5 is detection; step 6 is the read-index materialization.
- `docs/technical/architecture/subsystems/read-path/query-checkpoints.md:84` — the not-ready bullet still attributes a missing marker only to the index builder. After this PR the applier's main-store marker is the other producer, and on a lagging follower it is the one missing first.
- `docs/technical/architecture/subsystems/read-path/query-checkpoints.md:206` — the directory-layout tree omits `main.tmp/` and `readindex.tmp/`, so an operator will not recognize leftover debris as expected.

docs/technical/architecture/subsystems/read-path/query-checkpoints.md
```diff
@@ -21,7 +21,7 @@ Checkpoint IDs are assigned sequentially by the FSM (1, 2, 3, ...).
 2. The request is proposed through Raft consensus.
 3. The FSM commits pending state and records `QueryCheckpointState` metadata in
    Pebble, including the entry's Raft applied index `H`.
-4. The Applier creates a physical Pebble checkpoint of the main store at `{dataDir}/query-checkpoints/{id}/main/`.
+4. The Applier materializes the main store at `{dataDir}/query-checkpoints/{id}/main/`, **per-replica** and **atomic** on the same pattern as the read index below: build into `main.tmp/`, fsync, rename, then write `.ready` last. `pebble.DB.Checkpoint` writes the MANIFEST that makes a directory openable *before* it copies the WAL files, so a directory caught mid-copy opens cleanly while missing every write still resident in the memtable; the marker, written only after the rename, is what vouches for completeness. A redundant materialization on an already-marked directory is a no-op, which the Applier relies on when it re-crosses the checkpoint log during spool or WAL replay.
```
paul-nicolas: [P2][non-blocking] The replay rationale for the no-op guard is false, and its premise falsifies the recovery contract three paragraphs down

**What's wrong.** "...which the Applier relies on when it re-crosses the checkpoint log during spool or WAL replay" describes a path the applier cannot take. The same claim is repeated at `internal/storage/dal/store.go:940-942` and in the commit message.

`SetAppliedIndex` rides the trigger entry's *own* write batch (`internal/infra/state/machine.go:699`) and commits once (`machine.go:750`). `CreateQueryCheckpoint` runs only after that commit on both paths — hot path drains it explicitly first (`internal/infra/node/applier.go:1408-1412`, then `:1509`); replay path commits at `:1124-1126`, then `:1166-1168`. Every replay entry point starts strictly *above* the durable cursor (`replayWAL` uses `lo := afterIndex + 1` at `:1867`; `replaySpool` filters `Index > fromIndex` at `:1902`; `unspoolAndResume` re-reads the durable index at `:1790`), and `PrepareDecodedEntries` skips `entryIndex <= LastAppliedIndex` with a zero `ApplyResult` (`machine.go:497-501`), so a re-delivered trigger never sets `CheckpointRequired`. Follower sync only advances the cursor (`applier.go:553-560`, `synchronizer.go:89-93`), and cross-cluster restore replants a raft genesis snapshot containing no `CreateQueryCheckpoint` entry for a pre-restore id.

**Why it matters.** The premise contradicts line 86 of this same document. Line 86 says a node that crashes between the rename and the marker "never self-heals", and that is the stated basis for the operator's delete-and-recreate recovery. But if the applier really did re-cross the checkpoint log, that node would re-cross it too, hit the rebuild branch at `store.go:952-956`, and self-heal. Exactly one of the two statements can be true — and it is line 86.

This also runs into the `CLAUDE.md` rule that durability and recovery documentation must state the actual execution sequence rather than the intended end state.

**Fix.** Keep the guard — it is sound, and it is what replaces pebble's `ErrExist` now that `TestStore_CreateQueryCheckpoint_ErrorOnExistingDir` is gone. Just state it as defensive rather than relied upon, e.g. "A redundant materialization on an already-marked directory is a no-op; the applier's cursor commits in the same batch as the trigger entry, so a re-cross is not expected and the guard only keeps the call idempotent." Apply the same correction at `store.go:940-942`.

docs/technical/architecture/subsystems/read-path/query-checkpoints.md
```diff
@@ -83,7 +85,7 @@ The read index materializes asynchronously and **per-replica** (step 5). Readine
 - **A read for a checkpoint id that does not exist returns `NotFound`.** If there is no `.ready` marker *and* no `QueryCheckpointState` entry for the id, `openCheckpointStores` returns `NotFound` (permanent) so clients stop retrying — distinct from the retryable `Unavailable` above.
 - **Unrecoverable checkpoints degrade to `NotFound`, not wrong data.** There is no historical reconstruction: inline materialization is already exactly point-in-time. If a node crashes between the atomic rename and the marker, that node will never have a `.ready` marker for the checkpoint. Since the checkpoint is still registered, reads there return the retryable `Unavailable` and never self-heal — the operator/client recreates the checkpoint (aligned with the existing `AcquireCheckpoint` client workaround, which deletes-and-recreates on timeout). Deleting the checkpoint then makes reads return `NotFound`.
```
paul-nicolas: [P2][non-blocking] The Readiness and Error Contract omits the new terminal outcome this PR introduces

**What's wrong.** This section enumerates what a checkpoint read can return — retryable `Unavailable` when a marker is absent, permanent `NotFound` when the id does not exist, `NotFound` after an unrecoverable checkpoint is deleted. This PR adds a fourth, client-visible terminal outcome that is not listed anywhere: **both markers present and the main store fails to open** now surfaces the raw error (`internal/adapter/grpc/server_bucket.go:296`), which `convertToGRPCError` sanitizes into a permanent `codes.Unknown` with a correlation ID.

That is a deliberate design decision — a damaged directory should not be retried forever, and `Unavailable` is what triggers `actions.GRPCRetryPolicy`'s 50x retry — and you document it in the code comment at `server_bucket.go:294-295` ("damaged rather than late"). But the authoritative contract a client integrator reads is this section, and it is the section this PR rewrites (lines 54 and 56).

**Why it matters.** As written, this section tells a reader that checkpoint reads return only OK / `Unavailable` / `NotFound`. A client built against it will treat the new `Unknown` as an unexpected server fault rather than the documented "this replica's checkpoint is damaged, recreate it" signal. It also sits oddly beside the e2e's own assertion phrase at `tests/e2e/cluster/query_checkpoint_test.go:504` — "never Unknown".

**Fix.** Add a bullet to this list: both markers present but the store fails to open means damage rather than lag, so the error surfaces as a permanent sanitized `Unknown` (with a correlation ID in the node log) instead of a retryable `Unavailable`; recovery is delete-and-recreate, same as the crash-before-marker case above.

gfyrag:
I think there is an unresolved recovery question at HEAD `1ddf2873`, beyond the partial-directory safety fix.

The FSM batch containing `QueryCheckpointState` and the applied index is committed before the asynchronous main-store materialization. A crash after that commit but before `.ready` can therefore reopen with the checkpoint registered and the apply cursor already beyond its creation entry. There is no reconciler and the historical point H cannot be reconstructed after the live store advances. The comments currently conflict: `CreateQueryCheckpoint` says an unmarked directory is "discarded and rebuilt", while the index-builder lifecycle explicitly says that without a marker the checkpoint remains `Unavailable` until the client deletes and recreates it.

Please add a deterministic production-sequencing fixture with crash/reopen failpoints before the temp checkpoint, after it, after rename, and before marker publication. Then either provide a recoverable source at H/reconciliation path, or state explicitly that a successfully registered checkpoint can require client/operator deletion after a local crash and treat that as an accepted product limitation.

There is also a smaller choreography improvement: for the main-store half, create and fsync `.ready` inside `main.tmp`, then atomically rename the fully ready directory and fsync its parent. That removes the avoidable final-directory-visible-but-unmarked window. It does not solve the larger pre-rename recovery hole, but it narrows it. The current test that deletes the WAL from an already completed checkpoint proves the Pebble premise; it does not exercise this commit/materialization/restart interleaving.

Azorlogh:
@gfyrag Both halves now finish an interrupted materialization on restart (head `c0166cbb2`), so the contradiction is gone in the recovering direction:

1. **Main store.** The live store is still exactly at `H` when the node reopens: the applier is gated for the whole copy (later entries wait in the spool) and the spool is reset on boot, so nothing has advanced yet. `RecoverAndReplay` rebuilds any registered checkpoint whose main store is unmarked and whose `applied_index` equals the live index, before WAL replay. Restored checkpoints and checkpoints at another index (never applied here) stay unavailable, as before.
2. **Read index.** A checkpoint log now heads its own batch, and the indexed + AppliedProposal cursors commit only after the materialization. A builder that dies in between re-crosses the log (no rows to re-index) and materializes again; a marked directory is a no-op.
3. **Choreography.** The main-store marker is written in `main.tmp` and published by the rename.
4. **Fixtures.** `node/query_checkpoint_recovery_test.go` (nothing / partial temp copy / unmarked temp / marked temp / unmarked final, the guards, and `RecoverAndReplay` itself), `indexbuilder/process_logs_checkpoint_recovery_test.go` (dies on the audit wait → resumes; dies after the marker → re-cross keeps it; own-batch), and an e2e follower restart. All red on `1ddf2873`.
5. The readiness contract now names the two cases that stay `Unavailable` on a replica: it never applied the creation (later snapshot / leader resync), or its audit projection is disabled or failed.

internal/adapter/grpc/server_bucket.go
Azorlogh: Fixed in https://github.com/formancehq/ledger/pull/1954/commits/bb7f7da57 — table test in `internal/adapter/grpc` over the marker combinations plus the served case and a damaged main store. Confirmed red at 37ccf5238: `main store unmarked` returns the opened store instead of `ErrCheckpointNotReady`.

internal/storage/dal/query_checkpoint_wal_window_test.go
Azorlogh: Both fixed in https://github.com/formancehq/ledger/pull/1954/commits/d9e1df9ed — sentinel file on the redundant call, and a rename-into-`main.tmp` for the stale-temp branch. Each fails with its guard deleted.

docs/technical/architecture/subsystems/read-path/query-checkpoints.md
Azorlogh: Right, fixed in https://github.com/formancehq/ledger/pull/1954/commits/1d4eb0c8d — stated as idempotence, not a path replay depends on, in both the doc and `store.go`. Note the line 86 half is moot now: c0166cbb2 made the crash case self-heal via `RecoverAndReplay`, which is a real re-crossing caller.

docs/technical/architecture/subsystems/read-path/query-checkpoints.md
Azorlogh: Added in https://github.com/formancehq/ledger/pull/1954/commits/1d4eb0c8d.

Azorlogh:
@paul-nicolas P3 sweep in https://github.com/formancehq/ledger/pull/1954/commits/3cd1f18f7 — 8 of 9 done:

1. `store.go` post-rename — the parent fsync failure now reclaims `{id}/main`. (mark-in-temp had already collapsed your two branches into one)
2. `wal_window_test.go:73` — `require.ErrorIs(pebble.ErrNotFound)`, so the test fails if the read starts erroring rather than silently missing the write.
3. `RebuildsUnmarkedDir` — plants a `stale.sst` and asserts it is gone; confirmed red with the `RemoveAll(dir)` deleted.
4. `CreateQueryCheckpoint` failure path — closed store, asserts `ErrStoreClosed` and no directory left behind.
5. `checkpoint_marker_test.go` in `package dal`, `FsyncDir` included.
6. docs `steps 4 and 5` → `4 and 6` (already in c0166cbb2).
7. docs not-ready bullet — now names the applier's half, not just the builder.
8. docs layout tree — `main.tmp/` and `readindex.tmp/` listed as expected debris.

Left open: the delete race at `server_bucket.go:296` returning `Unknown` instead of `NotFound`. Your own note is why — re-stat only narrows it, since `RemoveAll` can take the MANIFEST before `.ready` — so it wants a real ordering decision rather than a patch. Happy to take it in a follow-up if you think it is worth the `NotFound` semantics.

Also folded in: the `RedundantCallIsNoOp` doc comment still carried the replay rationale you corrected, in 1d4eb0c8d.

Azorlogh:
@paul-nicolas Took the last P3 after all, in https://github.com/formancehq/ledger/pull/1954/commits/6ec595d92 — and you were more right than the P3 rating suggests: at `37ccf5238` a failed main-store open routed through `resolveMissingMarker`, so the delete race already returned `NotFound`. Turning that into a raw error was a regression in this PR, not a pre-existing wart.

The fix consults the registry rather than the marker:

```go
if exists, existsErr := impl.queryCheckpointExists(checkpointID); existsErr == nil && !exists {
    return nil, nil, commonpb.NewNotFoundError("query checkpoint %d not found", checkpointID)
}
```

That escapes the window you identified. The row is removed in the committed batch that *precedes* the unlink (`machine.go`, `deleteQueryCheckpointFiles` is dispatched as a post-commit side effect), so any open failure caused by a delete is necessarily preceded by the row disappearing in this node's own store. The filesystem has no such ordering, which is exactly why re-stat'ing `.ready` stays racy.

No barrier, unlike `resolveMissingMarker`'s absent-row case: both markers prove this replica applied the creation, so a locally missing row means it applied the delete too. Noted in the comment so it does not get "fixed" back.

Test stages the real interleaving — row deleted, `.ready` kept, MANIFEST unlinked — and asserts `NotFound`; confirmed red with the check removed. The damage case (row still present) keeps asserting the permanent error, so the two are pinned apart.

Azorlogh:
Addressed in deacf5ef2:

1. `query_checkpoint_recovery.go`: the discarded close now says why.
2. `server_bucket.go`: a failed registry probe returns its own error with the open failure attached, no longer classified as damage.
3. `process_logs.go`: both cursor-advance blocks go through `advanceCursors`.
4. `dal/store.go`: one deferred cleanup in `CreateQueryCheckpoint`.
5. Docs: the both-markers `NotFound` case is in the readiness contract.
6. Upgrade path: V3 is unreleased and state is wiped on every push to `release/v3.0`, so there are no pre-marker stores to migrate.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1954#issuecomment-5600647061

internal/storage/dal/store.go
```diff
@@ -921,26 +921,107 @@ const queryCheckpointsDir = "query-checkpoints"
...
+	}
+
+	if err := os.RemoveAll(tmpDir); err != nil {
+		return "", fmt.Errorf("clearing stale temp query checkpoint %d: %w", id, err)
+	}
+
+	// A failure from here on leaves the attempt under one of the two names;
+	// discard it so the next call starts clean. Best effort: the error being
+	// returned is the one that matters.
+	defer func() {
```
NumaryBot, line 973: 🟠 **[major] Preserve the final path for deferred failure cleanup**

If the parent fsync fails after the rename, `return "", err` assigns the empty string to the named result `dir` before this defer runs. Consequently, `os.RemoveAll(dir)` removes nothing and leaves the marked final directory behind despite the failed publication. A subsequent call treats that directory as ready and skips the failed durability step entirely. Capture the final path in a separate local variable so failure cleanup actually removes it.

Azorlogh: Fixed in https://github.com/formancehq/ledger/commit/e921da907: the path lives in a local the cleanup reads.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1954#issuecomment-5600647061

gfyrag (approved):

docs/technical/architecture/subsystems/read-path/query-checkpoints.md
```diff
@@ -45,20 +45,30 @@ Checkpoint IDs are assigned sequentially by the FSM (1, 2, 3, ...).
...
... response payloads enabled, `Apply` waits for the serving node's marker unless deletion supersedes it.** The `Apply` handler executes in this order: `ctrl.Apply` returns the committed logs and execution provenance → for a new execution, the handler scans those logs for a `CreatedQueryCheckpointLog` and, for each one, waits (`readStore.WaitForCheckpoint`) for that checkpoint's local `.ready` marker → response signing → `skip_response` payload stripping → return. The wait therefore precedes stripping, which nils out the payload carrying the id, and it locates the log by payload type rather than by position, because the checkpoint trigger is the batch's *last* action. The Applier marks the main store before it resolves the creation, so when the call returns both markers exist on that node and, if the checkpoint remains live, an immediate read at the returned `checkpoint_id` **routed back to that node succeeds**. It waits on the marker, not on the index-builder progress cursor — the cursor fast path was the EN-1460 root cause: the cursor is persisted in the batch that *precedes* the physical checkpoint creation, so it reaches the target sequence ~100-150 ms before the directory exists.
```
NumaryBot, line 71: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Limit the both-markers response guarantee to the leader**

A forwarding follower does not wait for its local applier’s checkpoint completion: `RoutedController.Apply` returns the leader’s result, and `waitCreatedQueryCheckpoints` checks only the local read-index marker. The FSM notifies index builders before main-store materialization finishes, so the follower’s read-index marker can exist while its main store is still copying. `Apply` can therefore return successfully and an immediate read on that same follower return `Unavailable`. Qualify the new guarantee to describe this follower window; the leader’s deferred future does not establish readiness on the forwarding node.

NumaryBot (commented):

PR #1955, opened 2026-09-09
refactor(query): unify the query compiler across directions (EN-1966)

paul-nicolas:

Adds the foundation for a direction-parameterized query compiler, following #1951 and the review of #1930. This PR targets `release/v3.0` and is ready for review. The compiler conversion is follow-up work: `leaves[D]` has no live callers yet.

## Changes

- Expose generic `And`, `Or`, `Not`, `Slice`, and `Filter` constructors while retaining the named direction-specific constructors.
- Add descending bitset, event-resolution, account-prefix, transaction-range, and ledger-log leaves.
- Add a stamp-gated descending prefix constructor that excludes rows folded after the reader's pin. This stages the visibility fix for the future compiler; the existing descending path in `internal/application/ctrl/list_entities.go` still uses the ungated constructor.
- Define the ten-method `leaves[D]` interface with ascending and descending factories. Compile-time assertions verify that each factory implements this interface; they do not prove that every direction-dependent leaf is represented.
- Use one `AddressTxIterator[D]` over a direction-independent account-to-transaction union. Materializing range scans remain outside the interface: both directions can consume the same sorted result.
- Share fold-stamp validation and event-key parsing between iterators, and add focused regression tests for descending visibility, bitset traversal, event resolution, and absolute seek repositioning.

## Follow-up

Convert `compile.go` to a single recursion parameterized by direction while keeping shared compilation semantics non-generic. Replace the existing `newReverseIterator` path in `internal/application/ctrl/list_entities.go` with the shared compiler. There is no `compile_reverse.go` in this branch.

Add end-to-end ascending/descending parity and error checks across accounts, transactions, and logs, including filters, cursors, page sizes, index/schema gates, and pinned visibility. Add registry exhaustiveness checks to detect omitted leaf pairs. These broader compiler guarantees are not delivered by this foundation.

## Validation

`go test -race ./internal/storage/readstore ./internal/query/... ./internal/application/ctrl` passes, as do the focused descending regression tests. Readstore lint passes with the unavailable custom `gomodguard_v2` plugin omitted; the full configured lint run requires that plugin. Query lint also reports a pre-existing `modernize` warning in unchanged `compact_account.go`. GitHub CI will rerun on the updated head; patch coverage remains subject to Codecov's result.

### DIFF
--- a/internal/query/leaves.go
+++ b/internal/query/leaves.go
@@ -1,0 +1,152 @@
+package query
+
+import (
+	"github.com/formancehq/ledger/v3/internal/pkg/bitset"
+	"github.com/formancehq/ledger/v3/internal/storage/readstore"
+)
+
+// leaves builds the leaf iterators whose PHYSICAL traversal depends on the
+// direction of the page being served — a Pebble cursor driven
+// First/Next versus Last/Prev, a bitset word walked from its lowest set bit
+// versus its highest, an event group resolved forwards versus backwards.
+//
+// It exists so that everything else about compiling a filter — predicate
+// resolution, schema validation, index-readiness gating, bound computation,
+// boolean composition, error wrapping, resource cleanup and profile-tree
+// construction — is written ONCE, in one recursion, and cannot be correct
+// ascending while being wrong or missing descending (EN-1966).
+//
+// Deliberately absent: the materializing range leaves. Draining a scan into a
+// sorted slice is direction-free, so the shared recursion builds the
+// ascending range scan and hands the one sorted result to Slice[D]. A
+// descending page over a materialized range therefore costs no second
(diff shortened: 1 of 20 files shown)


gfyrag (approved):
Approved for the explicitly scoped iterator/compiler foundations. No blocking correctness issue found in the reviewed paths; local race tests passed for internal/storage/readstore, internal/query/... and internal/application/ctrl on this exact HEAD.

The compiler conversion is follow-up work, not delivered here. Please retain the coverage follow-up: codecov/patch is still below the configured threshold (74.16% versus 80%), notably around factories and new constructors. This approval is not a request to lower or bypass that gate.

PR #1957, opened 2026-09-09
feat(metadata)!: enforce consistent metadata size limits (EN-1829)

paul-nicolas:
## Context

[EN-1829](https://formance-team.atlassian.net/browse/EN-1829). Metadata was bounded only by transport size (HTTP 4 MiB, gRPC 64 MiB) and by shape (key charset, NUL rejection). Nothing bounded entry count, key size, value size or aggregate size, so one accepted command could replicate megabytes of metadata through Raft into the audit chain, the FSM cache and the read-side projections.

## What this does

One canonical contract — `domain.MetadataLimits` — enforced at the two layers that can see the payload.

| Dimension | Default | Flag |
|---|---|---|
| Entries per entity | 128 | `--metadata-max-entries` |
| Key size | 256 B | `--metadata-max-key-bytes` |
| Value size | 16 KiB | `--metadata-max-value-bytes` |
| Total per entity | 64 KiB | `--metadata-max-entity-bytes` |
| Total per command | 256 KiB | `--metadata-max-command-bytes` |

**Admission** (`validateCommandMetadata`) covers direct HTTP, public gRPC and bulk through `requestsToOrders`. Mirror workers separately validate translated/re-written orders and aggregate bytes against the committed destination policy before proposing. Mirror FSM apply rechecks shape, per-entity limits and the proposal byte budget before mutation; rejected batches retain their durable cursor for retry.

**FSM apply** (`processCreateTransaction`) re-checks the *merged* transaction and per-account maps. Admission cannot: the union of caller metadata and `set_tx_meta` / `set_account_meta` output only exists after the script runs, and two individually legal halves can exceed the entity ceiling together.

A proposal-wide budget starts with every order's caller metadata and replaces each transaction's input contribution with its merged output. This enforces the aggregate ceiling across accounts and orders without double-counting caller overrides.

Direct account, transaction and ledger metadata saves, plus reversals, also check non-empty input against the committed policy and proposal budget before mutation. A policy tightened after admission therefore applies consistently to these paths. Bare-key metadata deletion and field-type set/remove operations also check the current key and proposal-wide ceilings before mutation, including before a missing-key delete can become a skipped order. Admission and mirror workers share `domain.ValidateCommandMetadata` for shape, entity and aggregate validation, including deterministic error selection.

The ceilings live in the Raft-replicated `ClusterPolicy`, so FSM apply never reads node-local configuration (invariant #2). Zero is never "unlimited" — it is refused by flag validation, by boot validation against a committed policy, and by the FSM before commit.

Violations return a stable typed error: gRPC `InvalidArgument` / HTTP 400, `ErrorInfo.reason = METADATA_LIMIT_EXCEEDED`, with `dimension` / `limit` / `actual` metadata. `Validation` rather than `ResourceExhausted` on purpose: the caller must send less, and a retryable code would make client retry policies re-drive a permanent rejection.

## Compatibility assessment

**`pkg/grpcprotocol.Version` "7" → "8" (incompatible).** The service-exposed `common.ClusterPolicy` gains five required metadata ceilings, so an older client's `SetClusterPolicy` request omits required configuration. Metadata payloads previously accepted can now be rejected. Revision 7 is already used on the target branch by EN-1954's Apply execution provenance; this PR increments from that revision to 8. Current examples in `protocol-compatibility.md` and `ops/cli.md` match revision 8.

No protobuf field is removed, so no renumbering. `ERROR_REASON_METADATA_LIMIT_EXCEEDED = 69` follows `ERROR_REASON_INDEX_ALREADY_EXISTS = 68` on the target branch.

## Scope decision: per command, not per stored entity

The ceilings bound what one command carries or produces. They do **not** bound an entity's accumulated stored metadata — 128 entries per command, repeated, still accumulates.

Bounding the accumulated total would need either an FSM read of the entity's whole metadata namespace, which no proposal's declared coverage authorises (invariant #6), or a new persisted per-entity counter, which would be a primary-store projection needing checker verification (invariant #8) and a restore classification (invariant #11). That is a larger change than the risk this addresses, which is what a *single accepted write* can replicate. The gap is documented in `metadata-limits.md` rather than left implicit; closing it is a candidate follow-up.

## Determinism

Both size checks run inside apply, where the rejection is hash-bound into the audit chain, so a per-node choice of error would diverge it. `ValidateMap` checks count and entity total before the per-entry rules and reports the lexicographically smallest offending key; `validateMergedAccountMetadata` does the same across accounts. Both are pinned by 100-iteration tests against Go's randomised map iteration.

## Refactor worth reviewing

Metadata traversal lives in the shared domain helper `WalkOrderMetadata` (`internal/domain/order_metadata.go`), including mirror deletion keys. The shape check, the size check and the per-command byte accounting now share a single traversal, so they cannot drift about which maps exist — a drift would either let a map through unvalidated or make the per-command total exclude bytes the per-entity check saw.

## Validation

Candidate based on `release/v3.0` at `25a2249b71d037d11bf02174be01b3735375fc49` (EN-1954), preserving exact integer decoding and using protocol revision 8.

- Added 12 apply-time policy-tightening regressions: account saves, transaction saves, reversals and ledger saves × value, entity and aggregate ceilings. Valid policy fixtures demonstrate acceptance at admission followed by rejection under the tighter committed policy. Assertions pin the dimension, limit, actual bytes and absence of writes.
- Added 19 bare-key policy-change cases covering account/transaction/ledger deletion and field-type set/remove: key and aggregate rejection, exact-boundary success, and skippable missing-key rejection. Disabling the key guard through a temporary overlay fails all 14 rejection cases.
- Added exact command-boundary success coverage and shared mixed-order accounting/deterministic error-selection regressions.
- Mutation check: disabling the apply guard through a temporary Go overlay fails all 12 rejection cases.
- Independent source review found no new correctness blockers.

Final local validation passed in the pinned Nix environment with four-CPU Go concurrency:

- `AI_REVIEW_BASE_SHA=25a2249b71d037d11bf02174be01b3735375fc49 bash scripts/agent-check-pr`: PASS, including baseline, full root race tests, E2E and Schemathesis.
- Dirty (`just pre-commit`): PASS, zero lint issues and no generated-file drift.
- CI unit coverage (`just test-coverage`), S3 lifecycle test, operator tests and workload race tests: PASS.
- Full-tag Business and Cluster E2E coverage and scenario coverage: PASS.
- `just test-model-cluster 180`: PASS, four restart cycles and no findings. An initial run hit the host disk-pressure threshold; the retry passed with the original threshold after clearing disposable Go build cache.
- Schemathesis (`MAX_EXAMPLES=10 SCHEMATHESIS_WORKERS=1`): PASS, all 62 endpoints.

The committed tree matches the validated tree exactly.

## Docs

New `docs/technical/architecture/subsystems/admission/metadata-limits.md` (contract, measurement rule, enforcement points, configuration, the accumulation gap). Updated `admission/{validation,README}.md`, `api/protocol-compatibility.md`, `contributing/api-comparison.md`, `ops/{cli,deployment}.md`, and `openapi.yml` (configurable metadata ceilings documented in descriptions; the schema retains the NUL-byte restriction on strings).

`validation.md` previously claimed a metadata key size limit that was never enforced. That claim is now true.

## Known non-blocking observations

- `openapi.yml`'s `MetadataValue` `oneOf` omits the `uint` and `datetime` variants that `common.proto` defines — pre-existing drift, reported not fixed.
- `processor_transaction.go` aliases the order's `MetadataMap` into stored account metadata (invariant #10 observation) — pre-existing, untouched.

[EN-1829]: https://formance-team.atlassian.net/browse/EN-1829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1957#issuecomment-5619596240

internal/domain/processing/processor_transaction.go
```diff
@@ -199,6 +212,15 @@ func processCreateTransaction(ledger string, order *raftcmdpb.CreateTransactionO
 		}
 	}
 
+	// Same reasoning as the transaction map above, per account. Checked in its
+	// own pass — and reporting the lexicographically smallest offending account
+	// — so the rejection does not depend on Go map iteration order: this runs
+	// inside apply, where a per-node choice of error would diverge the audit
+	// chain.
+	if metaErr := validateMergedAccountMetadata(accountMetadata, limits); metaErr != nil {
```
NumaryBot, line 220: 🔴 **[blocker] Enforce the command ceiling on Numscript output**

When Numscript produces metadata for several accounts, each merged map can pass `ValidateMap` while their aggregate exceeds `MaxTotalBytesPerCommand`; admission cannot detect these generated bytes, and apply never calls `ValidateCommandBytes`. This allows one command to replicate more metadata than the documented contract permits ([metadata-limits.md:27-31](docs/technical/architecture/subsystems/admission/metadata-limits.md#L27-L31)).

openapi.yml
```diff
@@ -3157,11 +3157,17 @@ components:
...
     MetadataValue:
-      description: A typed metadata value (string, integer, boolean, or null). String values must not contain NUL bytes.
+      description: >-
+        A typed metadata value (string, integer, boolean, or null). String
+        values must not contain NUL bytes. A value is limited to 16384 bytes by
+        default; the effective ceiling comes from the cluster policy
+        (`--metadata-max-value-bytes`). Exceeding it returns 400 with reason
+        `METADATA_LIMIT_EXCEEDED`.
       nullable: true
       oneOf:
```
NumaryBot, line 3175: 🟠 **[major] Do not expose configurable defaults as fixed schema limits**

When operators raise the replicated value or entry ceilings, the server accepts payloads that this OpenAPI schema still rejects because `maxLength: 16384` and the new `maxProperties: 128` values are hard-coded defaults. Generated clients and request validators therefore cannot use the configured API contract; keep configurable limits in descriptions rather than fixed schema constraints.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1957#issuecomment-5619596240

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1957#issuecomment-5619596240

internal/domain/processing/processor.go
```diff
@@ -223,6 +226,11 @@ func (p *RequestProcessor) ProcessOrders(orders []*raftcmdpb.Order, scopeFactory
 		AssetCache:     p.assetCache,
 	}
 
+	ctx.metadataBudget = &commandMetadataBudget{}
```
NumaryBot, line 229: 🟠 **[major] Revalidate bare metadata keys during apply**

When the policy tightens after admission but before apply, batches containing only delete-metadata or metadata-field-type orders never consume this seeded budget or revalidate their bare keys. Such a batch can therefore exceed the current per-key or proposal-wide ceiling and still commit, despite the apply-time policy guarantee documented in [metadata-limits.md:101-110](docs/technical/architecture/subsystems/admission/metadata-limits.md#L101-L110). Add apply-time validation for these bare-key handlers or validate the complete order batch before dispatch.

paul-nicolas: Done — fixed in commit b00ad2927. Apply now revalidates bare keys and the proposal-wide metadata budget for account, transaction, and ledger metadata deletion plus metadata field-type set/remove operations. The check runs before metadata lookup for deletes, so a skippable METADATA_NOT_FOUND cannot bypass a tightened key or command ceiling. Added 19 regression cases covering key/aggregate rejection, exact boundaries, state immutability, and skipped-missing-key behavior.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1957#issuecomment-5619596240

gfyrag (approved):
Re-reviewed at b00ad29270292fb2d6b13c8a58bb500d88dfc98c. No actionable findings remain; CI is green and prior Shipfox/NumaryBot feedback is resolved.

PR #1958, opened 2026-09-09
fix(transport): finish shutdown after stop cancellation (EN-1986)

gfyrag:
## What changed

Transport stop requests now survive caller cancellation. The dispatcher joins peers, closes their queues and the connection pool, then publishes completion; repeated stop callers share that completion.

## Why

Fixes [EN-1986](https://formance-team.atlassian.net/browse/EN-1986). A timeout after Fx hook entry could abandon the only dispatcher stop signal, leaving the worker and hook unfinished and skipping resource cleanup. Fx's outer `App.Stop` still returned on its deadline.

Shutdown must initiate transport termination even when its caller times out, and the existing worker join must prove peer/pool cleanup. The deterministic reproduction and durable shutdown contract are in `internal/infra/node/transport_shutdown_test.go` and `docs/technical/architecture/subsystems/consensus/raft-consensus.md`.

## Technical decision

Close the stop signal exactly once and let the dispatcher own cleanup. A buffered rendezvous would preserve termination but still abandon cleanup on timeout. Keeping cleanup in the worker's exit path reuses the existing Fx join without another background worker. Node drain and leadership-transfer ordering are unchanged.

## Risk

**MEDIUM** — changes transport shutdown ownership. No wire, persisted-state, or FSM changes.

## Validation

- `bash scripts/agent-check`: PASS.
- `AI_REVIEW_BASE_SHA=8c33007e46efaa95c6d8f02d474ac2fcea9e39ed bash scripts/agent-check-pr`: PASS (generation/tidy, root/operator lint, invariants/build, full node-package race tests).
- `go test -race ./internal/infra/node -run '^TestTransportShutdown$' -count=20`: PASS in pinned Nix.
- Native exact-base `agent-check-pr` on `658dfacacafd6a14b074c7ab2ddb484630178c56`: PASS for final candidate `f738803de754168d311a37fd48f69cf823840748`.
- Independent native final review: APPROVE, zero findings, clean worktree, `ROOT_UNCHANGED=PASS`, and final target refresh unchanged.
- Local ignored launcher/adapter copies corrected two macOS tooling issues (SIGPIPE from early-exit worktree enumeration; a portable multi-variable binding probe). All native validation, identity, sandbox, and review gates remained enabled; no tooling changes are in this PR.

DISCOVERY: NO_EXISTING_WORK
BEFORE_FIX: BUG_REPRODUCED
AFTER_FIX: PASS

On initial current base `8c33007e46efaa95c6d8f02d474ac2fcea9e39ed`, channel barriers keep the real dispatcher in peer publication while a live Fx hook begins stopping. Cancellation makes the outer App.Stop return, but the old hook/worker remain unfinished until a second Stop rescues them. Reverting the transport source to that base fails the exact missing-termination assertion in 5/5 repetitions. The fixed regression covers normal stop, cancellation after hook entry, cancellation during peer cleanup, already-cancelled callers, queue/pool closure, and repeated stops. The target advancement did not change the affected transport/bootstrap source.

`Stop` cancellation ends only that caller's completion wait. `Start` exits after peer/pool cleanup; the Fx hook joins it even if the outer `App.Stop` has already returned a context error. Pending Raft messages are not promised delivery after peer cancellation.

## Review focus

Persistent termination under cancellation; cleanup ownership and ordering; deterministic worker/hook and peer queue assertions.

## Known concerns

None identified. CI and human approval remain required before merge.

[EN-1986]: https://formance-team.atlassian.net/browse/EN-1986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/infra/node/transport.go
+++ b/internal/infra/node/transport.go
@@ -63,7 +63,10 @@ type DefaultTransport struct {
 
 	bufferSize           int
 	pendingSendQueue     chan []*raftpb.Message
-	stopCh               chan chan struct{}
+	stopCh               chan struct{}
+	stopDone             chan struct{}
+	stopOnce             sync.Once
+	stopErr              error // published by closing stopDone
 	advertiseAddr        string
 	serviceAdvertiseAddr string
 	// Metrics for recv queues (indexed by priority: 0=high, 1=medium, 2=low)
@@ -164,7 +167,8 @@ func NewTransport(
 		nodeID:               nodeID,
 		clusterID:            clusterID,
 		bufferSize:           bufferSize,
-		stopCh:               make(chan chan struct{}),
+		stopCh:               make(chan struct{}),
+		stopDone:             make(chan struct{}),
 		pendingSendQueue:     make(chan []*raftpb.Message, pendingSendCapacity),
 		advertiseAddr:        advertiseAddr,
 		serviceAdvertiseAddr: serviceAdvertiseAddr,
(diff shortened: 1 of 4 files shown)


internal/infra/node/transport.go
```diff
@@ -323,7 +328,7 @@ func (t *DefaultTransport) Stop(ctx context.Context) error {
 	t.peersMu.RUnlock()
 
 	for _, pc := range peersSnapshot {
-		err := pc.stop(ctx)
+		err := pc.stop(context.Background())
```
paul-nicolas, line 331: Why using context.Background() here instead of the ctx ?

Azorlogh (dismissed):

internal/infra/node/transport.go
gfyrag: The `ctx` passed to `Stop` limits how long that caller waits; cleanup must continue after its deadline. In `peerConnection.stop` (lines 1347–1355), we first cancel the peer's own `stopCtx`, then wait for `loopDone`. Passing the caller's expired context here would let that wait return early, skipping the peer queue closures and, because `closePeers` returns on error, the connection pool closure.

`context.Background()` makes this join uncancellable; it does not keep the peer's network operations alive, since those are cancelled through `stopCancel()`. The dispatcher publishes `stopDone` and exits only after peer and pool cleanup. `Stop(ctx)` can still return `ctx.Err()` promptly, and Fx's outer `App.Stop` can still hit its deadline while the hook finishes joining the worker.

That is the shutdown contract covered by `transport_shutdown_test.go` and documented in `docs/technical/architecture/subsystems/consensus/raft-consensus.md` (lines 180–190).

paul-nicolas (approved):

PR #1960, opened 2026-09-09
fix(EN-1987): index every log's date for a log-date index

Azorlogh:
## What changed

A log-date index now holds a row for every log of its ledger, on both paths that could leave one out:

- The backfill replaying a newly created index's history writes the date row from `backfillLogDateRow`, called for every log of the ledger being backfilled. Previously the row came from inside `indexLogEntry`, which the replay loop reaches only for a data log, so the `isDataLog` gate now covers only the entity projections a payload contributes.
- An index declared while its ledger is still born-empty no longer takes the promote-immediately shortcut when it is the log-date builtin: it takes the backfill path, which reaches the ledger's existing config-mutation logs and the `CreateIndex` log itself. That replay is bounded to the ledger's own history: the builder records the first global sequence it folds per ledger, and an initial index starts there.

## Why

Jira: EN-1987. The log-date index is opt-in and answers date filters over the same universe `ListLogs` scans, so it covers every ledger log. Two paths left the config-mutation logs — index creates and drops, account-type adds and removes, metadata field-type declarations — out of it, permanently, and nothing wrote them later:

1. **With history.** Creating the index on a ledger that already has logs schedules a backfill, but the replay only visited data logs (`isDataLog`: created/reverted transactions, saved/deleted metadata, skipped orders).
2. **Born-empty.** An index declared in the same proposal as its ledger is `initial` (EN-1564) and skips the backfill, which is sound for an entity index because there is no data history to replay. But born-empty is cleared only by an *indexable data* payload, so such a ledger can already carry schema logs, and the live fold cannot write the `CreateIndex` log's own date row because it reads the ledger config before that log registers the index.

A client saw two queries that must agree disagree: an unfiltered `ListLogs` returned every log, while a date range covering all of time silently omitted those entries for the life of the index. Whether a log was findable by date depended on its payload kind and on which side of the `CreateIndex` log it landed. The complement showed it from the other side: `not(<all-covering range>)` compiles to the log universe minus the date-index matches, so it returned exactly the unindexed logs instead of nothing.

DISCOVERY: model-checking driver (singleton_driver_model log-date index churn, PR #1659), first local cluster run
BEFORE_FIX: BUG_REPRODUCED — `tests/e2e/business/log_date_index_backfill_test.go`. With history: unfiltered `[1,2,3,4,5,6]`, date-filtered `[3,5,6]`, missing `[1,2,4]`, identical on three consecutive runs. Born-empty: unfiltered `[1,2]`, date-filtered empty.
AFTER_FIX: PASS

Need: a date-range log query returns the ledger's logs in that range, whenever and however the index was created.
Current limitation: only transaction and metadata logs are indexed from before the index existed, so date-filtered history is silently incomplete and its complement silently over-complete.
Requirement / constraint: the date index covers the same universe `ListLogs` scans; no client-visible difference between a log indexed by the live fold and one indexed by the backfill.
Evidence: EN-1987, with the deterministic e2e above and two independent model-driver findings.

## Technical decision

Decision: keep `isDataLog` for what it names — only a data log's payload has entity rows to fold — and give the per-log date row its own call site in the replay loop, restoring the live path's shape where it sits above the payload switch. `WriteLedgerLogIndex` stays inside `indexLogEntry`: the live fold writes that row unconditionally for every log, so hoisting it would only widen redundant puts.

For the born-empty case, the decision is which side decides. `initial` is minted by the FSM and replicated in the log, and its meaning ("this ledger has emitted no indexable data log") stays correct; the indexbuilder is what must stop treating it as "this ledger has no logs".

Alternatives considered: (1) admit schema logs in `isDataLog` — they would enter the entity switch, whose contract forbids it and which has no case for them; (2) widen the FSM's born-empty clearing to any log — changes a replicated field's meaning for every index kind; (3) write the `CreateIndex` log's date row inline on the live path — fixes one log and leaves the ledger's earlier logs out.

## Risk

**LOW–MEDIUM** — the backfill change adds one write per replayed non-data log during a log-date backfill, keyed exactly as the live path keys it, and is gated to a log-date task by the task-scoped `cfg`. The born-empty change means a log-date index declared with its ledger reports `INDEX_BUILDING` until its backfill completes instead of being live immediately. That backfill is bounded: an initial index's ledger was created in the proposal being folded, so this run has seen the ledger's first log, and the replay starts just below it instead of walking the whole global log. Without a recorded first log — a ledger this process never folded, which cannot be initial — it falls back to scanning from zero. No persisted format, proto or API change.

## Validation

- [x] Targeted tests: the new e2e covers both shapes — a ledger with schema logs and a data log before the index, and a ledger whose index is declared in its creating proposal — asserting that an all-covering date range returns the unfiltered listing and that the complement of a range every log satisfies is empty
- [x] Red-verified separately: the history spec fails on `release/v3.0` three times out of three with the same missing set; the born-empty spec fails with the shortcut restored (date-filtered listing empty against `[1,2]`)
- [x] Unit coverage of both halves and of the delete branch (`log_date_backfill_test.go`): the date row is written for a config-mutation log and only by a date task, an initial log-date index schedules a backfill, a replayed history of config-mutation logs is fully indexed. Mutation-verified: removing the carve-out, restoring the data-only gate, or dropping the task-ledger guard each turns a test red
- [x] Full suite / broader validation: `internal/application/...`, `internal/storage/...`, `internal/query/...`, `internal/domain/...` unit suites green; all 8 log-date e2e specs green, so the pre-existing coverage still holds
- [x] The model driver's log-date churn (PR #1659) finds no divergence on a local 3-node cluster run with this fix applied, over 5 restart cycles, with the date index created 91 times and backfilled to promotion 107 times across the three nodes — the same driver found the bug twice within seconds on the base commit

- Indexer: `backfillLogDateRow` writes the date row for every log of the ledger being backfilled; `indexLogEntry` keeps the always-on per-ledger log row and the entity projections; a log-date index always gets a backfill task.
- Reads: a date-range `ListLogs` returns the ledger's config-mutation logs whenever the index was created, and the complement of an all-covering range is empty. A log-date index declared with its ledger is briefly `INDEX_BUILDING`.

## Review focus

1. The born-empty carve-out: whether any other index kind's rows are also per-log and would need the same treatment.

## Known concerns

Two things this change does not address, both predating it and worth follow-ups:

- Nothing removes a backfill task when its ledger is deleted (`removeBackfillTask` is reached only from DropIndex and metadata-field removal), so a replay in flight keeps writing rows for a ledger that no longer exists. Those rows are unreachable — every read path gates on `DeletedAt` and `processCreateLedger` refuses a deleted name — so this is a leak rather than a correctness problem, and it belongs with backfill-task cancellation in its own lifecycle fix.
- The store checker scans neither the per-ledger log prefix nor the log date prefix, which is why a gap of exactly this shape was invisible to it and had to be found by the model driver.


internal/application/indexbuilder/backfill.go
```diff
@@ -1357,27 +1357,65 @@ func (b *Builder) processBackfill(ctx context.Context, stop <-chan struct{}, tas
...
+			// processBackfillPostings do, so a delete + same-name recreate
+			// leaves nothing of the deleted generation behind.
+			// markLedgerDeletedInBatch invalidates the in-batch dedup state so
+			// the recreate's writes, queued after the range delete, are not
+			// suppressed. A historical delete of any OTHER ledger must not be
+			// acted on: DeleteLedgerIndexes wipes every ledger-scoped prefix,
+			// which would clobber that ledger's live-maintained indexes.
+			if dl := log.GetPayload().GetDeleteLedger(); dl != nil {
+				if dl.GetName() == task.ledger {
+					if err := readstore.DeleteLedgerIndexes(b.wb.Batch(), task.ledger); err != nil {
```
NumaryBot: 🔴 **[blocker] Preserve current-generation indexes during historical deletes**

When a ledger name was deleted, recreated, and later receives a new backfilled index, replaying the old `DeleteLedger` calls `DeleteLedgerIndexes`, which removes every persisted index and version-state prefix for the recreated ledger, including unrelated indexes already maintained for the current generation. The replay then rebuilds only `task.index`, leaving those other READY indexes empty (and their persisted readiness state missing); the historical delete must clear only stale rows belonging to the backfill target rather than the ledger's current index keyspaces.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1960#issuecomment-5604632840

internal/application/indexbuilder/backfill.go
Azorlogh: The scenario this rests on is unreachable: a deleted ledger name cannot be recreated. `processCreateLedger` (`internal/domain/processing/processor_ledger.go:21-23`) refuses a name whose `LedgerInfo` carries `DeletedAt` with `ErrLedgerDeleted`, and nothing anywhere clears `DeletedAt`. So there is no "recreated ledger" whose current-generation indexes this wipe could remove, and no live generation's rows can sit under those prefixes once the name is deleted.

What the branch is for is narrower than the sibling's comment suggests, and I have made that explicit in a86217154: the replay sits behind the delete, so it has been writing rows for logs of a ledger that no longer exists. The wipe removes the replay's own leaked rows rather than leaving them; the alternative is to leak them permanently. It fires only for `task.ledger`, since a historical delete of another ledger may name one that is very much alive.

If a "purge and recreate the same name" feature ever lands, this wipe does become the hazard you describe, and the fix then is to narrow it to the task's own keyspace (`PrefixLedgerLogDate` for a log-date task) instead of every ledger-scoped prefix. I would rather keep one behaviour across both replays today than special-case this one against a scenario the processor forbids, but I am happy to narrow it if a reviewer prefers that.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1960#issuecomment-5604632840

Azorlogh:
Standards 1 and 4 fixed in a3790cc88: the indexer page's initial-index claims now say *entity* history and carry the log-date exception, and the complement spec's bound is renamed with a comment stating what the assertion rests on.

Standards 3: fair, no direct test. The branch is hygiene, not correctness — the replay sits behind the delete and removes its own leaked rows; they are unreachable either way, since every read path gates on `DeletedAt` and `processCreateLedger` refuses a deleted name.

Standards 2 left as is: the fourth site deliberately matches the sibling replay's shape, and unifying the guards is a refactor of untouched paths.

Merge note: #1659 carries an equivalent fix in `af44c449a`, so only one should land. This one adds e2e coverage in two shapes; theirs adds an indexbuilder unit regression.

internal/application/indexbuilder/index_config.go
```diff
@@ -448,7 +457,15 @@ func (b *Builder) handleCreatedIndexLog(ledgerName string, log *commonpb.Created
...
-	if log.GetInitial() {
+	// Born-empty means the ledger has emitted no indexable DATA log yet
+	// (processing.isIndexableDataPayload), which is what makes the shortcut
+	// sound for an entity index: there is nothing to replay. A log date index
+	// covers every log of the ledger, and such a ledger can already carry
+	// config-mutation logs — a schema declaration earlier in the proposal, and
+	// this CreateIndex log itself, whose date row the live path cannot write
+	// because it reads the config before this handler registers the index — so
+	// it takes the backfill path and reaches them.
+	if log.GetInitial() && !isLogDateIndex(id) {
```
paul-nicolas, line 468: [P2][non-blocking] The born-empty backfill replays the entire global log, not "a handful of logs"

**Evidence.** The carve-out routes an initial log-date index into the else branch, which calls `scheduleBackfillForIndex` → `addBackfillTaskForLogBuiltin` → `addBackfillTask`, and `addBackfillTask` hardcodes `cursor: 0` (`backfill.go:128-133`). `processBackfill` then runs `query.ReadLogsSince(ctx, handle, task.cursor, ...)` (`backfill.go:1310`), and with `afterSequence == 0` `ReadLogsSince` builds its lower bound as the bare `ZoneHistory/SubHistoryLog` prefix with no sequence suffix (`internal/query/log.go:194-201`) — the first global log in the cluster. Every log of every other ledger is decoded and discarded by the foreign-log skip at `backfill.go:1394`, and the task cannot finish before `task.cursor >= globalCursor` (`backfill.go:1149`), so the scan must reach the head. Meanwhile `CurrentVersion` stays 0 and every date-filtered read returns `ErrIndexBuilding` (`compile.go:1441-1449`).

The work is provably unnecessary for this case: `initial == true` requires the ledger to have been created in *this* proposal — `Context.BornEmptyLedgers` is "transient, per-ProcessOrders-call bookkeeping — never persisted" (`internal/domain/processing/processor.go:60-105`) — so every log the backfill must reach sits in the same apply batch, at the tail of the global log.

**Impact.** Declaring the log-date index at ledger birth goes from an O(1) promotion to an O(total cluster log) proto-decoding scan. On a cluster with real history that is minutes, during which date-filtered `ListLogs` on the brand-new ledger returns `ErrIndexBuilding`; N such ledgers created together each schedule their own full scan and divide one `backfillBudget` (`perTaskBudget = b.backfillBudget / len(b.backfillTasks)`, `backfill.go:1132`, against a 100ms tick). EN-1564 introduced the initial fast path to remove exactly this cost.

The PR's Risk section states "for such a ledger the backfill is a handful of logs". That is true of the *rows written* but not of the *scan*, and the scan is what the user waits on. It is also the sentence a future reader will trust.

**Resolution.** Either give the task a lower bound — `initial == true` guarantees the ledger's first log is in the batch being folded, so the builder can record the first global sequence it sees for that ledger and pass it to `addBackfillTask` as a starting cursor (a new `lowerBound` field consumed by `ReadLogsSince`) — or, if that is deferred, correct the Risk section and record the full-scan cost in `indexes.md` so it is a stated tradeoff rather than a silent regression of EN-1564.

Side effect worth noting either way: `WaitForLogBuiltinIndexReady` polls with a fixed 10s timeout (`pkg/actions/indexes.go:249-252`), and both `log_date_index_test.go:59` and the new spec at `:142` now wait on a full global replay whose length grows with how much of the shared business suite ran first. Fine at current scale; order-sensitive as the suite grows.

docs/technical/architecture/subsystems/indexer/indexes.md
```diff
@@ -104,10 +104,12 @@ stateDiagram-v2
 
 An index gets a fast path when it is declared in the **same atomic apply batch** as the `CreateLedger` that creates its ledger, before any indexable data log for that ledger. The FSM classifies this per-proposal — a ledger is treated as "born empty" until it emits its first indexable data log — and stamps the result on a new `CreatedIndexLog.initial` boolean.
 
-- **Initial index** (`CreatedIndexLog.initial == true`): there is no history to replay, so the indexbuilder allocates `next = HighWater + 1`, promotes it straight to `CurrentVersion=next`, and schedules **no** historical backfill. On the first incarnation `next` is 1; after a drop/recreate it is higher. `GetIndexStatus` immediately reports `current_version > 0` and carries no backfill cursor.
+- **Initial index** (`CreatedIndexLog.initial == true`): there is no *entity* history to replay, so the indexbuilder allocates `next = HighWater + 1`, promotes it straight to `CurrentVersion=next`, and schedules **no** historical backfill. On the first incarnation `next` is 1; after a drop/recreate it is higher. `GetIndexStatus` immediately reports `current_version > 0` and carries no backfill cursor.
```
paul-nicolas: [P2][non-blocking] The "Initial index" bullet still states the old contract; only its opening clause was amended

**Evidence.** The diff softened the opening to "there is no *entity* history to replay", but the rest of the bullet is unchanged and unqualified: "promotes it straight to `CurrentVersion=next`, and schedules **no** historical backfill […] `GetIndexStatus` immediately reports `current_version > 0` and carries no backfill cursor." For the log-date builtin all three clauses are now false — `index_config.go:468` gates the fast path on `log.GetInitial() && !isLogDateIndex(id)`, so it falls through to `{CurrentVersion: 0, PendingVersion: next}` plus a `backfillTask`.

The sibling bullet at line 85 *was* corrected ("An initial index, the log-date builtin aside"), and the exception is stated correctly at line 111 — but this bullet is the normative statement of the fast path and reads as authoritative on its own. It is also the natural landing spot for anyone debugging an unexpected `ErrIndexBuilding` on a fresh ledger.

Both reviewers flagged this line independently.

**Impact.** The subsystem doc tells the reader the opposite of the behaviour this PR introduces, and the correction only arrives two sentences later.

**Resolution.** Qualify line 107 the way line 85 already is — e.g. "**Initial index** (`CreatedIndexLog.initial == true`, log-date builtin excepted — see below)". While in this section, line 112 also needs the leading space dropped and "it replays an empty history and completes immediately" scoped to entity indexes, since for a log-date index on a still-empty ledger the history is precisely what the fix makes it replay.

internal/application/indexbuilder/index_config.go
```diff
@@ -448,7 +457,15 @@ func (b *Builder) handleCreatedIndexLog(ledgerName string, log *commonpb.Created
...
 	boundType, declared := log.GetBoundType(), log.GetBoundTypeDeclared()
 
-	if log.GetInitial() {
+	// Born-empty means the ledger has emitted no indexable DATA log yet
+	// (processing.isIndexableDataPayload), which is what makes the shortcut
+	// sound for an entity index: there is nothing to replay. A log date index
+	// covers every log of the ledger, and such a ledger can already carry
+	// config-mutation logs — a schema declaration earlier in the proposal, and
+	// this CreateIndex log itself, whose date row the live path cannot write
+	// because it reads the config before this handler registers the index — so
```
paul-nicolas, line 466: [P2][non-blocking] Two authoritative sites outside the diff still state the pre-carve-out contract

The carve-out's contract change was recorded in `indexes.md` only. Two other sites now say the opposite, and neither is in this diff:

**1. `docs/technical/architecture/subsystems/indexer/indexer.md:266`** — "**Index backfill** (a non-initial `CreateIndex`): the builder allocates `pending = HighWater + 1`, keeps `Current = 0`, and queries return `ErrIndexBuilding` until the switch". After this change an *initial* `CreateIndex` for the log-date builtin takes exactly this path, so the two indexer documents now disagree on when a backfill is scheduled. Per CLAUDE.md ("New technical mechanism / non-obvious invariant: update the matching subsystem documentation"), the carve-out is a non-obvious invariant and this is its second home. Suggested: "(a non-initial `CreateIndex`, or any `CreateIndex` for the log-date builtin — see indexes.md#initial-indexes-vs-later-indexes)".

**2. `internal/query/compile.go:1568`** — `requireIndexReady`'s contract comment reads: "This holds for EVERY index kind — builtin (reference, timestamp, inserted_at, address, **log_date**) and metadata alike. […] indexes declared on a born-empty ledger take the direct-ready fast path." This is the very gate that now returns `ErrIndexBuilding` for a born-empty log-date index, and the comment names `log_date` two lines above telling the reader that case cannot happen. CLAUDE.md requires relevant code comments to be updated with a behaviour change.

**Impact.** A reader of either site concludes the new state is unreachable. `compile.go:1568` is the worst of the two: it documents the exact gate whose behaviour this PR changes.

**Resolution.** Add the log-date exception to both, matching the wording already written into `indexes.md`.

(Posted here because neither file is in the diff; this is the line whose contract changed.)

internal/application/indexbuilder/backfill.go
```diff
@@ -1357,27 +1357,69 @@ func (b *Builder) processBackfill(ctx context.Context, stop <-chan struct{}, tas
...
+			// Wiping every ledger-scoped prefix is safe because the name is
+			// gone for good — processCreateLedger refuses a name whose
+			// LedgerInfo carries DeletedAt, and nothing clears it — so no
+			// live generation's rows can sit under those prefixes.
+			// markLedgerDeletedInBatch invalidates the in-batch dedup state,
+			// which reads committed Pebble and cannot see a pending range
+			// delete. A historical delete of any OTHER ledger must not be
+			// acted on: that ledger may well be live, and the wipe would
+			// clobber its indexes.
+			if dl := log.GetPayload().GetDeleteLedger(); dl != nil {
```
paul-nicolas: [P2][non-blocking] The new `DeleteLedger` branch has no test on this path — including the foreign-ledger guard

**Evidence.** The two existing `DeleteLedger` regressions — `backfill_test.go:2172 TestAccountAssetBackfillWipesDeletedLedgerGeneration` and `backfill_test.go:2242 TestAccountAssetBackfillDoesNotWipeUnrelatedLedger` — both drive `runAccountAssetBackfill`, i.e. an `ACCT_BUILTIN_INDEX_ASSET` task. `processBackfills` dispatches on `isPostingIndex(task.index)` (`backfill.go:1186-1188`): a posting index goes to `processBackfillPostings`, everything else — metadata, tx/account builtins, and the log-date builtin — goes to `processBackfill`. `grep -n makeDeleteLedgerLog internal/application/indexbuilder/*_test.go` returns only `backfill_test.go:2186`, `:2271` (both posting) and `protowire_postings_test.go:164` (a parse test). The new e2e spec never deletes a ledger either.

So **both halves** of this branch are untested on `processBackfill`: the same-name wipe (`DeleteLedgerIndexes` + `markLedgerDeletedInBatch`) and the `dl.GetName() == task.ledger` guard.

**Impact.** That guard is the exact review blocker `TestAccountAssetBackfillDoesNotWipeUnrelatedLedger` exists to pin on the posting path — dropping it silently resets a live, unrelated ledger's persisted `IndexVersionState` to `current = 0` on the next restart. On this path nothing would catch its removal. The branch is also the one NumaryBot challenged, so it is the part of the diff most likely to be edited again.

**Resolution.** Mirror the two posting tests for a non-posting task. A log-date backfill is the cheapest instance — it now replays every log of its ledger, so a `DeleteLedger` of the task ledger and of a foreign ledger both land in the replay. Assert, as the sibling does, that the foreign ledger's persisted `ReadIndexVersionState` survives.

internal/application/indexbuilder/process_logs.go
```diff
@@ -629,6 +629,32 @@ func (b *Builder) deleteReadIndexCheckpoint(checkpointID uint64) {
 	}
 }
 
+// backfillLogDateRow indexes one replayed log's date when cfg builds the log
+// date index. The date index covers every log of the ledger — a date filter
+// reads the same universe ListLogs scans — so this runs for a config-mutation
+// log too, while the entity projections in indexLogEntry stay specific to a
+// data log's payload. A no-op for any other backfill kind. The caller has
+// already established that log belongs to the ledger being backfilled.
+func (b *Builder) backfillLogDateRow(cfg *ledgerIndexConfig, log *commonpb.Log) error {
```
paul-nicolas, line 651: [P2][non-blocking] Neither new log-date code path has unit coverage; only a tagged e2e spec holds them

**`backfillLogDateRow`** — `grep -rn backfillLogDateRow --include='*_test.go'` returns nothing. The function has four early returns (index not enabled, non-`Apply` payload, nil `LedgerLog`, nil `Data`) and one side effect, and the load-bearing case is precisely the one `indexLogEntry` never reached: a config-mutation `Apply` log (`SetMetadataFieldType` / `AddedAccountType` / `CreatedIndex`). A reintroduced `isDataLog` gate above the call is caught only by the e2e suite. `backfill_test.go:254` already calls `indexLogEntry` directly, so the harness for a table test over the four guards plus one positive case (assert a `LedgerLogDateKey` row exists for a `SetMetadataFieldType` log) exists.

**The `isLogDateIndex` carve-out** (`index_config.go:468`) — every `Initial: true` unit test uses an account-asset or metadata id: `index_config_test.go:1387` and `:1494` both build `indexes.AccountBuiltinID(ACCT_BUILTIN_INDEX_ASSET)`, and `:1468` is the metadata case. None passes `indexes.LogBuiltinID(LOG_BUILTIN_INDEX_DATE)`. Deleting `&& !isLogDateIndex(id)` from line 468 leaves the entire unit suite green — I checked. The e2e spec cannot see the difference either: it calls `WaitForLogBuiltinIndexReady` (`log_date_index_backfill_test.go:141`) before asserting, which polls until `current_version > 0` and therefore steps over the new `INDEX_BUILDING` window entirely.

**Impact.** Both halves of the EN-1987 fix rest on `tests/e2e/business` (which CI does run — `main.yml:115` — so this is a fast-feedback and localisation gap, not a coverage hole). A later refactor of the initial fast path or of the replay loop drops either one with a green unit suite and a single opaque e2e failure.

**Resolution.** Add the log-date twin of `TestHandleCreatedIndexLog_InitialSkipsBackfill` — `Initial: true` with a `LOG_BUILTIN_INDEX_DATE` id must schedule exactly one backfill task and leave `{current: 0, pending: 1}`, keeping the account-asset test as the control — plus the `backfillLogDateRow` table test above.

paul-nicolas (commented):

Head reviewed: `a3790cc88e9ec22418f6e3a3594da94aa3643ad9`

**DECISION: FINDINGS** — 0 blocking, 5 non-blocking P2 posted inline, 6 non-blocking P3 listed below.

The correctness core of the fix holds under both models. Independently verified and **not** re-filed:

- `indexLogEntry` has exactly one production caller (`backfill.go:1423`); the live fold writes the date row inline at `process_logs.go:244`, untouched. Removing the write from `indexLogEntry` loses coverage on no path.
- `backfillLogDateRow` is correctly placed (after the foreign-log skip, before the `isDataLog` gate) and its predicate matches the live path byte for byte, so the two producers of the `lldt` keyspace cannot disagree about which logs get a row.
- The `CreateIndex` log **is** an `Apply` log carrying `LedgerLog{data,date,id}` (`common.proto:650-705,730`), so the born-empty backfill really does reach its own date row. The fix works.
- `lastSeq`/`batchCount` are set on all four branches; the flush gate is `batchCount > 0`, not `wb.Empty()`, so the range-delete wipe cannot be silently cancelled.
- Single-goroutine builder, unversioned `LedgerLogDateKey`, idempotent `(key, nil)` writes — no race, no pending/current keyspace mismatch.
- **NumaryBot's blocker on the `DeleteLedger` branch: OUTDATED.** @Azorlogh's refutation is correct and stronger than stated. `processCreateLedger` returns `ErrLedgerDeleted` (`processor_ledger.go:22-24`); `DeleteLedgerData` explicitly spares `LedgerInfo` (`internal/infra/state/batch.go:470-471`); `rebuild.go:1097-1105` reproduces the tombstone on restore; nothing clears `DeletedAt`; and `tests/e2e/cluster/mirror_test.go:728-736` pins the lifecycle as unreachable. Not re-filed.

The surviving findings are one behaviour/cost regression the PR body understates, three documentation sites that now state the opposite of the code, and three untested new branches.

### Conflicts between the two models

Codex reported only the `indexes.md:112` stray space (P3). Claude's per-area reviewers additionally found the cursor-0 replay extent, the two out-of-diff doc sites and the unit-test gaps; each was verified against the code before posting. One internal disagreement was resolved: one reviewer called the added per-log write "bounded and cheap" (true — one 81-byte key per replayed log of the task ledger) while another flagged the *scan extent* (the whole global log). Different scopes, both correct; the finding is filed on the extent, not the per-log cost.

### P3 (not posted inline)

- `internal/application/indexbuilder/index_config.go:361` — `isLogDateIndex` matches the enum **value**, not the kind. Every `LogBuiltinIndex` is per-log by construction, which is exactly what makes the born-empty shortcut wrong, so the next value added to the enum silently falls back into the fast path and reintroduces EN-1987. This is your own review-focus question 2. Consider `_, ok := id.GetKind().(*commonpb.IndexID_LogBuiltin); return ok`, or an exhaustive switch whose default is an invariant failure.
- `internal/application/indexbuilder/process_logs.go:653` — `b.wb.SetEventSequence(...)` is a no-op here: `WriteLedgerLogDateIndex` (`write_batch.go:398-402`) never reads `eventSeq`, unlike `WriteTransactionRevertedAtIndex`. Harmless (the next `indexLogEntry` re-stamps the same value) but it signals the date row is event-stamped when it is not.
- `internal/application/indexbuilder/backfill.go:1504` — `isDataLog`'s doc comment still says config-mutation logs "never need re-indexing on backfill". After this PR they do: their date row is written eleven lines above the gate.
- `internal/application/indexbuilder/process_logs.go:700` — the amended comment promotes the `CreateIndex`/`DropIndex` arms from "presumed live" to "documented dead" without making them loud (invariant 7), and omits the `LogPayload_DeleteLedger` arm at :668, which is now double-handled by the new branch at `backfill.go:1373`.
- `docs/technical/architecture/subsystems/indexer/indexes.md:112` — stray leading space (confirmed by both models; rendering is unharmed by CommonMark lazy continuation, and `git diff --check` does not flag it), and the sentence "it replays an empty history and completes immediately" now contradicts the sentence directly above it for the log-date kind.
- `tests/e2e/business/log_date_index_backfill_test.go:108` — the same all-covering date filter literal is rebuilt three times, with the bound spelled `everySince` at :108 and `everSince` at :144/:168. One helper plus `actions.NotFilter` would do.

### Validation considered

Read-only inspection at the head commit. `go build`/`go vet` on `internal/application/indexbuilder/...` and `go vet -tags e2e ./tests/e2e/business/...` clean. No suite executed (the e2e suite needs Docker and would contend for ports). CI does run this package: `.github/workflows/main.yml:115` invokes `just test-e2e-coverage './tests/e2e/business/...'` and the coverage gate requires `e2e-business.out`, so the new regression coverage is enforced.

Residual risk: LOW.

Azorlogh:
All five addressed in f84ae6baa.

1. Risk corrected: the rows are a handful, the scan is the whole global log. Recorded, with the lower bound (`initial == true` puts the ledger's first log in the batch being folded) as a follow-up rather than in this PR.
2 and 3. `indexes.md` bullet qualified, and the two sites outside the diff — `indexer.md` backfill paragraph and `requireIndexReady`'s contract — now carry the log-date exception.
4 and 5. `log_date_backfill_test.go` covers `backfillLogDateRow` (config-mutation log, non-date task, non-Apply, nil ledger log), the initial carve-out, a replayed config-mutation history, the delete branch, and the foreign-ledger guard. Mutation-verified: removing the carve-out, restoring the data-only gate, or dropping the task-ledger guard each turns a test red.

One fixture note for whoever reads those tests: `makeCreatedTxLog` stamps neither log id nor date, and the log cursor reuses its decoded message, so an unstamped transaction fixture reads back the previous log's date. Real logs always carry both.

internal/application/indexbuilder/index_config.go
NumaryBot, line 468: 🟠 **[major] Avoid replaying global history for born-empty date indexes**

On clusters with substantial existing history, routing an initial log-date index through the normal backfill schedules a task at cursor 0. `processBackfill` then scans every global log and merely discards logs from other ledgers, keeping this new ledger's date index unavailable until the entire cluster history is traversed even though only logs from its current proposal are relevant. Seed this special backfill at the ledger's creation boundary or otherwise replay only its short local history.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1960#issuecomment-5604632840

internal/application/indexbuilder/index_config.go
Azorlogh: Fixed in cf39a5b00 rather than deferred, since you and paul-nicolas landed on it independently.

The builder now records the first global sequence it folds per ledger, and an initial log-date index starts its replay just below that instead of at zero. An initial index means the ledger was created in the proposal being folded, so this run has necessarily seen its first log; with no recorded entry — a ledger this process never folded, which therefore cannot be initial — it still scans from zero.

Pinned by `TestHandleCreatedIndexLog_InitialLogDateBackfillsFromLedgerFirstLog` and its no-entry twin; an off-by-one on the bound fails both that test and the born-empty e2e spec, which loses exactly log id 1.

internal/application/indexbuilder/index_config.go
```diff
@@ -497,6 +514,18 @@ func (b *Builder) handleCreatedIndexLog(ledgerName string, log *commonpb.Created
 
 	b.putVersionState(ledgerName, indexes.Canonical(id), state)
 
+	// An initial log-date index only has its own ledger's logs to replay, and
+	// this run folded the first of them, so the replay starts there instead of
+	// walking the whole global log (EN-1987). Any other index either has real
+	// history behind it or is not initial, and starts from zero.
+	if log.GetInitial() {
+		if first := b.ledgerFirstSeq[ledgerName]; first > 0 {
```
NumaryBot, line 522: 🔴 **[blocker] Avoid trusting process-local first-log state after restart**

If the indexbuilder restarts after committing an earlier configuration log but before folding the initial `CreateIndex` log, `ledgerFirstSeq` is empty and records the `CreateIndex` itself as the ledger's first log. Starting the backfill at `first-1` then permanently skips those earlier ledger logs, recreating the incomplete date index this patch fixes. Only use this bound when the actual ledger-creation boundary is proven; otherwise replay from zero.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1960#issuecomment-5604632840

internal/application/indexbuilder/index_config.go
Azorlogh: Right, my justification was wrong: "initial" does not imply this process folded the ledger's first log, because the fold of that proposal can span a restart.

Fixed in 78a941623: the bound is now recorded only when the fold sees the ledger's own creation log, so an entry proves this process folded the creation itself. A restart between the creation and the CreateIndex log leaves no entry and the replay covers the whole log. The entry is dropped when the ledger is deleted, and a re-folded creation log cannot move an existing bound.

`TestRecordLedgerCreation` pins the recording rules and the two `InitialLogDate` tests pin both branches of the lookup.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1960#issuecomment-5604632840

internal/application/indexbuilder/index_config.go
Azorlogh: Fixed rather than documented: the replay is now bounded by the ledger's creation log sequence in 78a941623, with a full scan when this process did not fold that creation.

docs/technical/architecture/subsystems/indexer/indexes.md
Azorlogh: Bullet qualified in f84ae6baa, and the exception moved into its own paragraph.

internal/application/indexbuilder/index_config.go
Azorlogh: Both sites carry the exception now: the indexer.md backfill paragraph and requireIndexReady's contract (f84ae6baa).

internal/application/indexbuilder/backfill.go
Azorlogh: Covered in f84ae6baa: TestLogDateBackfillWipesDeletedLedgerGeneration and TestLogDateBackfillDoesNotWipeUnrelatedLedger, the latter red when the task-ledger guard is dropped.

internal/application/indexbuilder/process_logs.go
Azorlogh: log_date_backfill_test.go covers both paths and their early returns, mutation-verified (f84ae6baa).

gfyrag:
Re-check on current HEAD `78a94162`: the new unit coverage closes the earlier test gap around the `DeleteLedger` branch, and the core EN-1987 log-date correction still looks sound.

The remaining design challenge is whether that destructive branch belongs here at all. The author response describes it as hygiene rather than correctness: deleted ledgers are already unreadable, same-name recreation is refused, and the real outstanding lifecycle problem is that the backfill task is not removed. In that state, range-deleting every ledger-scoped index prefix expands the behavior and blast radius without contributing to the EN-1987 guarantee.

My preference is to remove the `DeleteLedgerIndexes` branch from this PR and address backfill-task cancellation as its own lifecycle fix. If the wipe is retained, please make the invariant explicit: why can no later ledger generation ever observe or depend on those prefixes, including after compact/reopen, and why is a broad wipe safer than stopping/removing the obsolete task? The added tests prove the branch executes and is scoped by ledger name; they do not establish that this is the best lifecycle boundary.

Azorlogh:
Removed the `DeleteLedger` branch in 6515e9ca0, per your preference — it was hygiene for rows nothing can read, and backfill-task cancellation is the real lifecycle fix. Known concerns now says so, and the two tests that pinned the branch went with it. That also settles shipfox's duplication point: three replay loops become two again.

Also folded the bounded cursor back into `addBackfillTask` as a parameter rather than a second method (shipfox 3).

On shipfox 2, method colocation: `backfillLogDateRow` sits in `process_logs.go` next to the live-path write it mirrors, which is where every other `Builder` fold method lives — `indexLogEntry`, `indexCreatedTransaction`, `indexPayload`. Seven files in the package define `Builder` methods, so moving this one to `builder.go` would isolate it from the code it belongs with. Leaving it, unless you read the convention more strictly than the package does.

gfyrag (approved):
@Azorlogh Approved at `6515e9ca09aa9d5e48b5ae3d57910f7f78cd7c8e`: the focused log-date correctness fix and its regression coverage look good, and current checks are green.

Please merge this as soon as the remaining merge requirements are satisfied, so #1933 can rebase cleanly on top. The intended integration is to retain these log-date completeness regressions while #1933 replaces the temporary initial-log-date backfill strategy with its durable EMPTY/NON_EMPTY tracker and proactively staged CONTROL-log dates. We should reconcile the overlapping helpers and remove the superseded initial/ledgerFirstSeq mechanism rather than stack both approaches.

PR #1963, opened 2026-09-09
fix(api)!: redact sink and mirror credentials from reads (EN-1632)

gfyrag:
## What changed

HTTP and gRPC read responses now mask sink and mirror credentials in live configuration, historical creation logs, serialized audit orders, and signed payloads. A shared deep-clone projection also hides transport diagnostics that can repeat credentials, while preserving status timestamps, pagination, and operational settings.

Unchanged audit bytes and signatures remain exact. Changed signed payloads retain key identity but omit the invalidated signature; malformed or shadowed credential fields fail the read. Service protocol revision remains 5: protobuf messages and RPC definitions are compatible. The repository policy now reserves bumps primarily for breaking protobuf/RPC changes; non-schema bumps require a concrete interoperability failure.

## Why

Fixes EN-1632, EN-1634, and EN-1635. EN-1796 was closed as a duplicate of EN-1632, with qualification evidence attached.

At base `37ccf5238ed460e0df2b9844869bdb9c7031f532`, deterministic HTTP/gRPC tests reproduced the leaks under read scopes, including decoded hex/base64 audit bytes and a transport error containing an endpoint credential. Previous PRs #1651, #1653, #1654, and #1703 were closed without merging; their proposed fixes are absent from this base.

Read scopes must expose status and history without granting reusable connector credentials, including credentials retained after rotation or removal. Persisted configuration, accepted orders, audit evidence, workers, and backups must retain their original bytes and behavior.

The requirement and its verification tradeoff are recorded in `docs/technical/architecture/subsystems/api/secret-redaction.md`. Regression tests cover transport boundaries, real checkpoint reads, production audit-hash computation, and Ed25519 verification.

## Technical decision

Apply one field policy immediately before HTTP serialization or gRPC stream emission, after controller/checkpoint selection. Deep clones avoid changing shared runtime objects or audit capture. A schema-aware wire guard detects credential fields hidden by protobuf last-value semantics before preserving opaque evidence.

Client-only formatting or live-configuration-only redaction leaves alternate transports and historical bytes exposed. Mutating storage would alter authoritative evidence and operational credentials. A new privileged raw-read API is outside this change.

## Risk

**MEDIUM** — deliberate read-contract change: returned credential-bearing audit payloads are display projections, and status diagnostics become a fixed marker. Existing revision-5 clients remain admitted and can decode responses; audit-verification consumers must account for redacted evidence.

## Validation

Policy follow-up (`f146fb085`): `bash scripts/agent-check`, normalization/lint, and the HTTP/gRPC/projection race tests pass. The exact merge-base `agent-check-pr` rerun fails on local timing fixtures in `tests/antithesis` and `scripts/internal/testenv`; both unchanged packages pass isolated under race (`-p 1 -parallel 1`, 32.261s and 1.394s). The version constant returns to 5 and the credential projection implementation is unchanged. Follow-up CI is pending; prior-candidate results below remain historical.

- [ ] `AI_REVIEW_BASE_SHA=37ccf5238ed460e0df2b9844869bdb9c7031f532 bash scripts/agent-check-pr`: normalization/lint and baseline pass; full root race suite passes except the unrelated local Antithesis launcher fixture described below. Retrying with `GOFLAGS=-p=1` removes deadline failures but still exposes its early-exit polling race.
- [x] Antithesis launcher package passes independently under race, both with its default parallelism (8.458s) and `-parallel 1` (32.139s).
- [x] Standalone local `test-schemathesis`: 62 endpoints, zero failures/errors; one recovered transient network error ignored by the canonical runner (the combined gate stops at the fixture failure).
- [x] Previous candidate `d5682feca` GitHub CI: all checks passed, including root tests, E2E business/cluster, model, Schemathesis, operator, scenarios, workload, normalization, invariants, build and coverage.
- [x] HTTP/gRPC and projection regression tests, including real checkpoint reads and pagination.
- [x] Original audit hashes/signatures still verify; redacted payloads cannot authenticate as original evidence; unchanged noncanonical payloads remain byte-exact.
- [x] Base-handler mutation checks reproduce the original leaks; malformed and shadowed protobuf credentials fail closed.
- [x] Independent final review of projection and wire policy; no blocking findings.

The default concurrent full run hit the Antithesis launcher fixture's 15-second deadline. With `GOFLAGS=-p=1`, the remaining `driver_exits_before_deadline` failure misses `DRIVER EXITED EARLY` inside its two-second polling window and reports `NO VERIFIED MODEL OUTCOMES` instead. The fixture runs fake binaries from temporary repositories; both fixture and launcher are byte-identical to the base. No test assertions or deadlines were changed or skipped. CI ran the complete suite successfully on the previous candidate; the policy follow-up will rerun CI.

Public HTTP/gRPC read semantics change; protobuf fields and HTTP response shapes do not. Protocol revision remains 5; no schema or RPC compatibility break. FSM, durable writes, checker inputs, backup, and incremental restore are unchanged because projection creates no persisted effect. Current history remains in the main store; archived/cold history was removed by EN-1945.

OpenAPI, CLI, API parity, event/mirror, and signing documentation describe the public projection and verification boundary.

## Review focus

Credential URL/DSN policy, projection after checkpoint selection, and exact-byte preservation versus rejection of shadowed protobuf credentials. Review the documented distinction between returned audit identity and verification of the original stored evidence.

## Known concerns

The combined local gate has the unrelated Antithesis and synchronization-fixture timing failures above; isolated regression reruns and the previous candidate’s full clean CI pass. This is retained as a validation concern rather than changing unrelated harness behavior in this credentials PR.

This policy covers connector credential fields and their public status diagnostics. Arbitrary business metadata, URL paths, write-response echoes, and protected operator diagnostics retain their existing semantics. Original evidence requires the protected store/backup workflow; read scopes do not provide a raw bypass.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,8 +12,12 @@ This file is the always-loaded entry point for AI agents working on Ledger v3. K
 
 ### Service protocol revision — required during pre-release
 
-Every incompatible service gRPC contract change must increment
-`pkg/grpcprotocol.Version` in the same PR, including during pre-release. Read the
+Increment `pkg/grpcprotocol.Version` primarily for breaking changes to exposed
+protobuf messages or RPC definitions, including during pre-release. Compatible
+response-value changes (such as credential redaction or sanitized diagnostics)
+do not automatically require a bump. Any non-schema bump must identify the
+concrete client/server incompatibility that requires rejecting older clients.
+Read the
 [protocol maintenance rules](docs/technical/architecture/subsystems/api/protocol-compatibility.md#maintaining-the-revision)
 before modifying this contract.
 
--- a/docs/ops/cli.md
+++ b/docs/ops/cli.md
@@ -5076,3 +5076,17 @@ ledgerctl query-checkpoint get-schedule
 # Output (no schedule):
 #  SUCCESS  No query checkpoint schedule configured (automatic creation disabled)
 ```
+
+### Sink and mirror credentials in read output
+
+The server masks reusable sink/mirror credentials in ledger, event-sink, log,
+and audit reads, including structured and result-file output. Connection status
+retains its error presence and timestamp, but its diagnostic message is masked.
+Read connection strings are display projections and must not be copied back
+into configuration. The write path still accepts the original credentials.
+
+For audit reads, an unchanged signed payload retains its signature. When a
+credential is replaced, the signature bytes are omitted and the displayed
+payload cannot reproduce the original audit hash. Use protected original
+store/backup evidence for verification; see the
+[read projection contract](../technical/architecture/subsystems/api/secret-redaction.md).
--- a/docs/technical/architecture/subsystems/admission/signing.md
+++ b/docs/technical/architecture/subsystems/admission/signing.md
@@ -100,3 +100,12 @@ The signed payload format is fixed by vtprotobuf encoding rules — `internal/pk
 | Bootstrap registration rejected | `RequireSignatures()` is true and there is no parent key — solution is the maintenance-mode workaround above. |
 | Client cannot verify a stored log | Server's response key has rotated; client must refresh via the Discovery RPC or the `--response-verify-key` flag. |
 | Audit chain hash mismatch on entries with valid `Log.Signature` | Log entry tampered after the fact — caught by the checker, not by signature verification (the signature still verifies; the hash chain doesn't). |
+
+## Credential-bearing read projections
+
+Public audit/log reads apply the [credential projection contract](../api/secret-redaction.md).
+The original stored bytes remain verifiable. A redacted read payload is a
+display projection: it cannot reproduce the original signature or audit hash,
+and its cryptographic signature bytes are omitted when the payload changes.
+Unchanged read payloads retain their exact bytes and original signatures.
+Request admission and signed write responses keep the contract described above.
--- a/docs/technical/architecture/subsystems/api/README.md
+++ b/docs/technical/architecture/subsystems/api/README.md
@@ -11,6 +11,7 @@ Client-facing transport layers (`internal/adapter/grpc`, `internal/adapter/http`
 | [protocol-compatibility.md](protocol-compatibility.md) | Mandatory service protocol revision, client/server rejection behavior, diagnostics, and revision maintenance (EN-1851). |
 | [http-api.md](http-api.md) | HTTP REST API endpoints, response formats, and error handling. |
 | [auth.md](auth.md) | Client JWT authentication (OIDC + Ed25519), scope-based authorization, and the Raft inter-node cluster-secret auth layer. |
+| [secret-redaction.md](secret-redaction.md) | Shared sink/mirror credential projections, public status diagnostics, and audit verification limits. |
 
 ## Related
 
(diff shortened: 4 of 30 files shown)


gfyrag:
This draft is superseded by the stack #1977 → #1979 → #1970, now rebased on `release/v3.0` at `1f86c44ac`.

**Delta qualification:**

The two commits unique to this PR are:

- `d5682feca` (`fix(api)!: redact sink and mirror credentials from reads`) — the credential redaction implementation is fully covered across the three-layer stack. The `docs/technical/architecture/subsystems/api/secret-redaction.md` file introduced here is not present in the stack; its content is distributed across `sensitive-projection.md` (projection mechanics), `structured-credentials.md` (field policy, audit/signature contract, validation requirements), and `protocol-compatibility.md` (version policy). No implementation delta is missing.

- `f146fb085` (`fix(grpc): reserve protocol bumps for interoperability breaks`) — the protocol version policy is documented in `protocol-compatibility.md` in the stack. The comment change in `pkg/grpcprotocol/protocol.go` (from "must change when wire format or semantics break" to "tracks breaking changes to exposed protobuf messages and RPCs") is not in the stack. The version number itself resolves to 10 from `release/v3.0` in the stack.

**Missing from the stack:** the updated comment in `pkg/grpcprotocol/protocol.go` (one line) and the consolidated `secret-redaction.md` file. Both are documentation-only. The implementation and behavioral contracts are fully covered.

This PR remains open for reference per maintainer instructions; the stack is the active implementation path.

gfyrag:
**EN-1632 stack qualification — #1963 superseded**

PR #1963 introduced credential redaction at read time via `internal/adapter/readprojection` (per-type projection with hand-rolled wire-format parsing to handle DSN-encoded credentials in the old SinkConfig/MirrorSourceConfig storage format).

The three-layer stack that replaces this approach:

- **PR #1977** (`codex/sensitive-projection`): `internal/pkg/sensitive.Clone()` helper and `(sensitive) = true` proto field annotations — field-level descriptor-driven redaction for any annotated message.
- **PR #1979** (`codex/structured-connection-configs`): Changes the storage model so credentials are never written as opaque DSN strings. Sink and mirror configs store parsed connection components (`ConnectionURL`, `ConnectionAddress`, etc.) with individual sensitive annotations. Read responses use `sensitive.Clone()` directly at the handler level.
- **PR #1970** (`codex/structured-credential-projections`): Typed public audit views (`publicauditpb.AuditEntry`) that expose business details without raw signed-batch evidence, eliminating the need to redact opaque bytes from audit reads.

**Functional coverage**: the same HTTP and gRPC endpoints (`GetEventsSinks`, `GetLedger`/`ListLedgers`, `GetLog`/`ListLogs`, `GetAuditEntry`/`ListAuditEntries`) are protected by the new stack with no code from #1963 carried forward. The `adapter/readprojection` package, `secret-redaction.md`, and the hand-rolled wire decoder are not included in the new stack; their concerns are addressed at the storage and annotation layers.

The `docs/technical/architecture/subsystems/api/structured-credentials.md` in PR #1979 explicitly notes this as the alternative to #1963 and documents the architectural decision.

**Recommendation**: This draft can remain open for reference. No new code from #1963 needs to be cherry-picked into the current stack. If there are any requirements unique to #1963 not addressed by the new stack, please flag them for targeted inclusion.

PR #1964, opened 2026-09-10
test(antithesis): model-driver prepared query coverage (EN-1629)

paul-nicolas:
Stacked on #1659 (EN-1625). **Base is `feat/model-driver-querying`, not `release/v3.0`** — this work reuses that PR's filter evaluators and window validation, so it cannot merge first.

## What this adds

The model driver never touched prepared queries. This adds generation and oracle validation for the registry lifecycle and for `ExecutePreparedQuery` in both modes.

**Oracle (`tests/oracle`)**

- `LedgerState` carries a `preparedQueries` registry, fingerprinted so two bases differing only in a stored filter cannot dedup — they predict different execution windows.
- `LedgerOf` / `logKindFor` / `applyOne` extended in lockstep, plus the three apply methods. They delegate to the functions the FSM itself calls (`ValidatePreparedQueryName`, `IsPreparedQueryExecutableTarget`, `ValidateFilterForTarget`) and surface those errors' own `Reason`, so the model and the FSM cannot drift.
- `logKindFor` returns `""` for all three. Prepared-query orders produce **top-level** `LogPayload` arms, not the `Apply` arm, so they consume no ledger-local log id. Naming a kind would shift every subsequent log id and make #1659's log-query validation report findings against a correct server. Pinned by `TestGlobalState_Apply_PreparedQueryAppendsNoLog`.

**Driver**

- Registry ops churn a small shared name pool as single-request bulks, so `ALREADY_EXISTS` and `NOT_FOUND` arise from ordinary contention rather than a dedicated probe.
- Parameterization is the exact inverse of substitution: a concrete filter from the ad-hoc generators has a random subset of leaves rewritten into parameter references at creation, and binding them back at execution reproduces it. That is what keeps the marginal cost low — the ad-hoc evaluators are reused unchanged, with no second grammar. The round trip is a property test over every parameterizable arm.
- `ListPreparedQueries` and `ExecutePreparedQuery` (LIST over all three targets plus a follow-on page driven by the server's own cursor, and AGGREGATE_VOLUMES) validate against candidate bases, each supplying its **own** stored definition. A concurrent update or delete is therefore a legal alternative, not a finding.
- Aggregates compare the bucket **set** as well as the totals: a zero-valued cell one side purged and the other kept changes the set while leaving every total identical.
- ~1-in-8 a parameter binding is deliberately omitted or mistyped; the compiler must reject it.

## Server change included (`fix(query)`, first commit)

`ExecutePreparedQuery` returned two caller errors as bare `errors.New`, so they reached clients as the sanitised `codes.Unknown` reserved for server faults. The driver caught this on its first run: the workload's standing "every RPC error must be classified" assertion rejects any code outside the transient/cancelled/business set. Both are now validation sentinels mapping to `codes.InvalidArgument`.

**Compatibility: `pkg/grpcprotocol.Version` increments from 10 to 11.** No `.proto` text, RPC, message, or field changes, but revision-10 clients observe these validation failures as sanitised `codes.Unknown` responses while revision-11 clients observe `codes.InvalidArgument`. Because the compatibility contract treats changed error semantics as a protocol-visible change, the revision bump is required even though both calls already failed.

## Divergences the harness surfaced during bring-up

All four were driver-side and are fixed here, each explained at the code:

1. LOGS was routed through the accounts/transactions index classifier, whose default arm maps every log condition to the never-built sentinel — right for an accounts filter, wrong on the LOGS target.
2. A base with an unresolvable parameter was required to produce the parameter's own compilation error, when the compiler resolves an index first and may legitimately fail there.
3. The log-date builtin index is never created by the driver, so a stored LOGS filter reading the date is permanently refused.
4. `transactionWindowRows` inverts `reverse` (`descending := !reverse`), because the ad-hoc list surface defaults to newest-first while prepared execution paginates forward — the intuitive `false` predicts a descending page the server never returns.

## Validation

| Check | Result |
|---|---|
| `bash scripts/agent-check` | PASS |
| `just test-model` (single node, 60s) | PASS, 0 findings |
| `just test-model-cluster` (3 nodes, rolling restarts) | PASS, 0 findings |
| `go test ./internal/query/... ./internal/domain/... ./internal/adapter/grpc/... ./tests/oracle/...` | PASS |
| `just test` | 91 packages ok; 8 environmental failures (below) |

Coverage measured from a `MODEL_DEBUG=1` run rather than inferred from the absence of findings:

| Trace | Count |
|---|---|
| registry listings validated | 740 |
| LIST executions validated | 665 |
| aggregates validated | 70 |
| aggregate-on-non-accounts rejections observed | 124 |
| spoiled parameter bindings | 36 (24 compilation, 11 index, 1 legally successful) |

Two invariants are pinned by mutation, not just covered: breaking `substituteUint`'s `ParamMax` branch fails four round-trip subtests, and making `logKindFor` name a prepared-query kind shifts the log ids from `1` to `1,2,3,4` and fails its test.

### The 8 `just test` failures are environmental

- Seven `scripts/*` packages shell out to `git commit` in temporary repositories and inherit the machine's `commit.gpgsign=true` / `gpg.format=ssh`, failing with `fatal: failed to write commit object` (exit 128). No file under `scripts/` is touched here.
- `tests/antithesis` `TestRunModelTestRequiresVerifiedOutcome` exceeds its 15s deadline under full-suite load; it passes in isolation in 6.7s.


gfyrag:
Review follow-up (reviewed `43a46ed`; registry finding rechecked on current HEAD `ad8960e`).

**[P2][blocking] Reject duplicate names in `registryMatches`.**
In `tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go`, the comparison checks the total length and then each entry's membership/content, but never uniqueness. For an expected registry `{a,b}`, a response `[a,a]` passes if both copies of `a` are correct: the duplicate hides the missing `b`. This is an oracle blind spot, not evidence that the server currently returns duplicates. Please reject duplicate names and add this same-length counterexample to `TestRegistryMatches`.

**Model CI divergence needs an explanation before merge.**
The run on `43a46ed` failed on `not(and(logDate[_,_], ledger=…))`: the server returned log IDs `1,2`, while the model expected no rows. This was the ordinary `ListLogs` path, not directly prepared execution, and the evidence does not yet establish whether the server or the model is wrong. Please isolate the divergence and pin the resolved semantics with a regression rather than merely rerunning the probabilistic test.
Historical failure: https://github.com/formancehq/ledger/actions/runs/34484445123/job/102895058869
The current HEAD's Tests-Model check is also failing, but I have not diagnosed that newer run and am not claiming it has the same cause.

The latest prepared-LOGS correction reviewed on `43a46ed` is a good improvement: it uses the actual date-index lifecycle and checks missing/invented rows and `hasMore`, with dedicated negative tests.

Non-blocking: please add explicit gRPC assertions for the two new execution validation errors (aggregate on a non-accounts target and unsupported mode). Observing a rejection alone does not establish the intended error classification; concurrent delete/recreate outcomes should remain distinguished from the targeted rejection.

No approval at this stage. Validation for this review was source/test inspection and CI-log analysis, not a fresh local test run.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,952 @@
...
+// the flat sum of every matching account's volume cells, one bucket per
+// (asset, color) — ExecutePreparedQuery exposes no grouping, max-precision or
+// colour-collapse option, so the executor calls AggregateVolumes with the
+// zero-valued AggregateOptions.
+//
+// Bucket PRESENCE is asserted, not just the sums: a zero-valued cell the server
+// purged but the model kept (or the reverse) changes the bucket set while
+// leaving every total identical, and comparing sums alone would miss it.
+func (c *Checker) validateExecuteAggregate(maxTicket uint64, call preparedCall, agg *commonpb.AggregateResult) {
+	if call.errKind == pqErrOther {
```
NumaryBot: 🟠 **[major] Validate aggregate target rejections against candidate bases**

If an ACCOUNTS query is deleted and recreated as LOGS or TRANSACTIONS after `preparedQuerySnapshot`, the generated aggregate request legitimately returns `ErrPreparedQueryAggregateTarget`. That error is classified as `pqErrOther`, so this branch reports a divergence before checking any candidate base. `runAggregateTargetMisuse` cannot handle this case because routing used the earlier snapshot. Model this rejection against each candidate's target and add a delete/recreate regression.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,952 @@
...
+	target commonpb.QueryTarget,
+	bound *commonpb.QueryFilter,
+	after string,
+	cur *commonpb.PreparedQueryCursor,
+) bool {
+	switch target {
+	case commonpb.QueryTarget_QUERY_TARGET_ACCOUNTS:
+		return preparedAccountPageMatches(ls, call, bound, after, cur)
+	case commonpb.QueryTarget_QUERY_TARGET_TRANSACTIONS:
+		afterID, ok := parseAfterUint(after)
```
NumaryBot: 🟠 **[major] Preserve cursor byte ordering across query recreation**

If an ACCOUNTS query returns a page ending at `acc:1` and is recreated as TRANSACTIONS before the follow-on request, the server reuses the cursor's raw account-address bytes as the seek boundary and can correctly return an empty page. The model instead parses `acc:1` as a transaction ID and unconditionally rejects that candidate. The reverse target transition also compares decimal IDs rather than their big-endian cursor bytes. Preserve the previous page's raw entity key and apply the server's bytewise boundary semantics when validating a changed target.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1964#issuecomment-5682179924

NumaryBot (commented):

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,952 @@
...
+		return
+	}
+
+	cursor := resp.GetCursor()
+	c.validateExecuteList(maxTicket, call, "", cursor)
+
+	// One follow-on page, driven by the server's own cursor: the model's
+	// after-key is the last row of the page just validated, so the opaque
+	// cursor is never decoded.
+	if call.errKind != pqErrNone || cursor.GetNext() == "" {
```
NumaryBot, line 277: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Reject missing pagination continuations**

A correct full page with `hasMore=true` but an empty `next` passes validation, then silently ends pagination here. The transaction validator additionally ignores `hasMore`, so returning only the first page with both continuation fields cleared also passes despite known remaining transactions. These regressions make later results inaccessible while the workload reports success. Validate continuation presence against `hasMore`, and validate transaction `hasMore` against the candidate’s remaining required/possible rows.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1964#issuecomment-5682179924

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1004 @@
...
+	}
+
+	out := map[aggregateBucket]oracle.VolumePair{}
+
+	for key, pair := range ls.Volumes().All() {
+		if _, ok := matched[key.Address]; !ok {
+			continue
+		}
+
+		bucket := aggregateBucket{asset: key.Asset}
```
NumaryBot: 🟠 **[major] Preserve color in model aggregate buckets**

The workload already generates colored postings through `randomColor()`, but this key folds every color into the uncolored bucket. The production aggregator preserves color with the default options used by prepared execution. Consequently, any selected colored volume makes a correct aggregate fail validation. Include `key.Color` in the bucket and add a regression containing colored postings.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries.go
```diff
@@ -0,0 +1,863 @@
...
+	return v.StringValue, true
+}
+
+func lookupUint(params preparedParams, name string) (uint64, bool) {
+	pv, present := params[name]
+	if !present {
+		return 0, false
+	}
+
+	v, isUint := pv.GetValue().(*commonpb.ParameterValue_Uint64Value)
```
NumaryBot: 🟠 **[major] Match the compiler's numeric parameter coercions**

When a concurrent update reuses `p0` for a different numeric condition, the request can supply an int64 to a uint condition or vice versa. `extractUint64` and `extractInt64` in the production compiler accept these conversions when the value fits, whereas these lookup helpers mark them unresolved. The validator then requires rejection and reports a correct successful response as a divergence. Mirror the compiler's accepted conversions and bounds checks, and test an update that changes the parameter's numeric kind.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1004 @@
...
+	if stored.GetTarget() != commonpb.QueryTarget_QUERY_TARGET_ACCOUNTS {
+		return call.errKind == pqErrOther
+	}
+
+	bound, resolved := substituteParams(stored.GetFilter(), call.params)
+	if !resolved {
+		return unresolvedParamRejectionLegal(ls, bound, stored.GetTarget(), call.errKind)
+	}
+
+	return indexedQueryOutcomeLegal(ls, stored.GetTarget(), bound, preparedNeededIndexes(bound, stored.GetTarget()),
```
NumaryBot, line 832: 🟠 **[major] Reject unexpected aggregate errors before validating results**

For an ACCOUNTS query whose filter selects no volume cells, an unexpected non-transient error is accepted here: `asIndexedErrKind(pqErrOther)` becomes `indexedErrNone`, and `aggregateMatches` treats the nil response as the expected empty aggregate. This hides server failures even without a target-recreation race. Keep the candidate-specific target rejection handling, but reject `pqErrOther` for ACCOUNTS candidates instead of passing it through the successful-result path.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1964#issuecomment-5682179924

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1002 @@
...
+			"query":  name,
+		})
+
+		return
+	}
+
+	// A success is legal only if every candidate base disagrees with the
+	// snapshot that shaped the call — the query was concurrently recreated on
+	// the ACCOUNTS target between the snapshot and the call.
+	if c.matchesModel(maxTicket, "PQAGGMISUSE", func(base oracle.GlobalState) bool {
```
NumaryBot: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Validate successful aggregates after target recreation**

When the misuse probe’s query is concurrently recreated as ACCOUNTS, this branch accepts any successful response solely because an ACCOUNTS candidate exists. It never checks parameter resolution, index readiness, or aggregate contents. Thus incorrect totals—or success despite missing required parameters—pass the oracle. Validate the complete aggregate outcome against that candidate and add a recreation regression with an incorrect result.

NumaryBot (commented):

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,993 @@
...
+		return call.errKind == pqErrNotFound
+	}
+
+	if call.errKind == pqErrNotFound {
+		return false
+	}
+
+	// The call was shaped for an ACCOUNTS-target query; a base holding another
+	// target explains only the rejection, which runAggregateTargetMisuse owns.
+	if stored.GetTarget() != commonpb.QueryTarget_QUERY_TARGET_ACCOUNTS {
```
NumaryBot, line 820: 🟠 **[major] Restrict aggregate target rejection to validation errors**

For a LOGS or TRANSACTIONS candidate, this accepts every `pqErrOther`, including an unrelated `LEDGER_NOT_FOUND` or `AlreadyExists` response. Those business codes also pass the workload's global classification interceptor, so incorrect server failures silently pass model validation. The target guard now returns `InvalidArgument` with reason `VALIDATION`; classify that outcome separately and reject unrelated errors, with a regression covering a non-ACCOUNTS candidate.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1964#issuecomment-5682179924

internal/domain/errors.go
```diff
@@ -498,6 +498,18 @@ var (
 	// defined by its filter). Rejecting nil keeps the stored query intact and the
 	// audit trail deterministic on wire-replay.
 	ErrPreparedQueryFilterRequired = NewValidationSentinel("prepared query filter is required")
+	// ErrPreparedQueryAggregateTarget guards ExecutePreparedQuery:
+	// AGGREGATE_VOLUMES folds the volume cells of the accounts a query selects,
+	// so it is only meaningful for an ACCOUNTS-target query. Returning it as a
+	// bare error sent clients a sanitised codes.Unknown — the "unclassified
+	// server fault" bucket — for what is a caller mistake; the model-checking
+	// driver surfaced that (EN-1629).
+	ErrPreparedQueryAggregateTarget = NewValidationSentinel("AGGREGATE_VOLUMES mode is only valid for ACCOUNTS target queries")
```
gfyrag: [P2][blocking] Keep query-mode validation out of the domain error set

These failures describe `ExecutePreparedQuery` request/query compatibility; they are not FSM-generated business outcomes. Adding them to `internal/domain` and returning them through `domain.BusinessError` contradicts the layer boundary documented by `internal/query.ErrAggregateOverflow`, which deliberately keeps query-only outcomes in `internal/query` while implementing `domain.Describable` for shared wire conversion. `ErrQueryModeUnsupported` is even defined entirely by a protobuf request enum.

Please keep the `VALIDATION`/`InvalidArgument` wire classification, but define these typed errors in `internal/query` (or validate/map the transport-only case in the gRPC adapter) and return them without presenting them as domain business errors. Reserve domain sentinels and `BusinessError` for actual domain outcomes.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1964#issuecomment-5682179924

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,998 @@
...
+		return call.errKind == pqErrAggregateTarget
+	}
+	if call.errKind == pqErrOther {
```
NumaryBot: 🟠 **[major] Reject target-validation errors for ACCOUNTS and LIST candidates**

For an ACCOUNTS query selecting no volume cells, an incorrect `InvalidArgument`/`VALIDATION` response still passes validation: it becomes `pqErrAggregateTarget`, bypasses this `pqErrOther` guard, and `asIndexedErrKind` converts it to success, allowing `aggregateMatches` to accept the nil response as an empty aggregate. LIST validation has the same blind spot for empty windows. Reject `pqErrAggregateTarget` for ACCOUNTS aggregate candidates and all LIST candidates, and add empty-result regressions; only a non-ACCOUNTS aggregate candidate can explain this rejection.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/main.go
```diff
@@ -210,12 +210,13 @@ func runWorker(
...
-		// in-flight bulk set, exercising cross-node freshness without needing
-		// quiescence.
+		// field types), the list queries (filtered, paginated, ordered windows
+		// over accounts, transactions and logs), and the prepared-query surface
+		// (the registry listing, and execution of a stored definition with its
+		// parameters bound). Reads validate against the in-flight bulk set,
+		// exercising cross-node freshness without needing quiescence.
 		if random.RandomChoice([]uint8{0, 1, 2, 3, 4}) == 0 {
-			switch random.RandomChoice([]uint8{0, 1, 2, 3, 4, 5, 6, 7, 8}) {
+			switch random.RandomChoice([]uint8{0, 1, 2, 3, 4, 5, 6, 7, 8, 9}) {
```
NumaryBot: 🟠 **[major] Preserve a reachable choice for single-account reads**

Every value in the random choice pool now has an explicit case, so the default branch calling `runRead` is unreachable. This removes the workload's only `GetAccount` read path, including its absent-account and absent-ledger probes, rather than adding prepared-query coverage alongside it. Add another selectable value or an explicit account-read case. This also violates the additive regression-coverage requirement in [AGENTS.md:87](AGENTS.md#L87).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1964#issuecomment-5682179924

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/main.go
```diff
@@ -230,7 +231,11 @@ func runWorker(
 				runReplay(ctx, client, c)
 			case 6:
 				runLogQuery(ctx, client, c)
+			case 7:
```
NumaryBot: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Restore the GetAccount read selection**

Every value in the random selection now has an explicit case, making the default `runRead` branch unreachable. This is its only call site, so the workload loses all GetAccount point-read validation, including absent-account and absent-ledger probes. Add a selectable case for `runRead` to preserve the existing coverage.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1001 @@
...
+	}
+
+	if !complete {
+		// Coverage trace for the deliberately spoiled binding: the rejection is
+		// what the negative path exists to observe, so record that it happened
+		// rather than inferring it from the absence of a finding.
+		dbg("PQPARAM spoiled errKind=%d", int(call.errKind))
+	}
+
+	if mode == commonpb.QueryMode_QUERY_MODE_AGGREGATE_VOLUMES {
```
NumaryBot, line 261: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Validate the response arm before extracting results**

The protobuf getters return nil when the response carries the wrong result arm or no result. Both validators accept nil as an empty result when the candidate predicts no rows or volume buckets. Consequently, a successful LIST response containing an aggregate—or an aggregate request returning a cursor—can pass validation. Check that successful responses contain the requested result arm before extracting it, including follow-on pages and misuse probes, and add empty-result counterexamples.

NumaryBot (commented):
OpenCodeReview's independent Codex pass added 2 comment-only findings.

internal/domain/errors.go
NumaryBot: 🟠 **[major] Keep query-mode validation out of the domain error set**

These errors are produced exclusively by query execution, not FSM processing. Defining them here and wrapping them in `domain.BusinessError` violates the layer boundary explicitly documented by `internal/query/errors.go:14-21`, which keeps query-only outcomes in `internal/query` while implementing `domain.Describable` for shared wire conversion. Define these errors in the query layer and return them directly, preserving the new VALIDATION/InvalidArgument classification.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1964#issuecomment-5682179924

internal/adapter/grpc/stream_helper_profile_test.go
```diff
@@ -100,10 +100,11 @@ func TestSendPagedToStream_ProfilePhaseAttribution(t *testing.T) {
 
 		profile.Finish()
 
-		require.Less(t, profile.ServerDuration, 3*spinPerItem,
-			"consumer back-pressure must not inflate the consumer-independent total")
-		require.GreaterOrEqual(t, profile.WallDuration(), 3*spinPerItem,
-			"but the slow-query threshold must still see it")
+		require.Equal(t, profile.DeliverDuration,
```
NumaryBot: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Preserve verification that delivery is excluded from server time**

`WallDuration()` returns `ServerDuration + DeliverDuration`, so the new equality is tautological and the inequality only checks that delivery is positive. If `Finish()` stops subtracting `DeliverDuration`, both assertions still pass, although consumer back-pressure incorrectly inflates server time. The removed assertion detected that regression. Retain an independent check of delivery exclusion, using controlled timing to avoid scheduler-sensitive bounds.

NumaryBot (commented):

internal/query/executor.go
```diff
@@ -84,10 +84,12 @@ func Execute(
...
 	}
 
-	// Validate mode compatibility
+	// Validate mode compatibility with query-layer typed errors. They implement
+	// domain.Describable for wire conversion without presenting read-side request
+	// validation as an FSM-generated business outcome.
 	if req.GetMode() == commonpb.QueryMode_QUERY_MODE_AGGREGATE_VOLUMES &&
 		pq.GetTarget() != commonpb.QueryTarget_QUERY_TARGET_ACCOUNTS {
-		return nil, errors.New("AGGREGATE_VOLUMES mode is only valid for ACCOUNTS target queries")
+		return nil, ErrPreparedQueryAggregateTarget
```
NumaryBot: 🟠 **[major] Bump the service protocol revision for changed error semantics**

This changes the response to the same `ExecutePreparedQuery` request from a sanitized `codes.Unknown` fault to `InvalidArgument` with a `VALIDATION` ErrorInfo, but `pkg/grpcprotocol.Version` remains `10`; therefore an existing v10 client can observe and classify the operation differently while still passing the protocol gate. Increment the revision and update its current-revision documentation. [protocol-compatibility.md:149-156](docs/technical/architecture/subsystems/api/protocol-compatibility.md#L149-L156)

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1964#issuecomment-5682179924

internal/query/executor.go
```diff
@@ -84,10 +84,12 @@ func Execute(
 		}
 	}
 
-	// Validate mode compatibility
+	// Validate mode compatibility with query-layer typed errors. They implement
+	// domain.Describable for wire conversion without presenting read-side request
+	// validation as an FSM-generated business outcome.
 	if req.GetMode() == commonpb.QueryMode_QUERY_MODE_AGGREGATE_VOLUMES &&
```
NumaryBot: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🔴 **[blocker] Increment the service protocol revision for the changed error contract**

Returning a `Describable` changes these client-visible failures from sanitized `codes.Unknown` responses to `codes.InvalidArgument` with `ErrorInfo{reason: VALIDATION}`. Existing clients can and do interpret gRPC status codes as retry/operation semantics, so this is an incompatible service contract change under the repository’s protocol-maintenance rules. `pkg/grpcprotocol.Version` remains `"10"`; bump it and update the protocol documentation’s current-version references.

NumaryBot (commented):
OpenCodeReview's independent Codex pass added 2 comment-only findings.

internal/query/executor.go
```diff
@@ -190,7 +192,7 @@ func Execute(
 		}
 
 	default:
-		return nil, fmt.Errorf("unknown query mode: %v", req.GetMode())
+		return nil, ErrQueryModeUnsupported
```
NumaryBot: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Validate unsupported query modes before compiling the stored filter**

The new `ErrQueryModeUnsupported` is reached only after `Compile`. A gRPC caller can send an unknown enum with a stored filter requiring an unavailable index or a missing parameter and receive `INDEX_NOT_FOUND`/`FILTER_COMPILATION_ERROR` instead of `InvalidArgument` for its invalid mode. Validate that the mode is LIST or AGGREGATE_VOLUMES immediately after loading the query; the added test only covers an unfiltered query and misses this path.

NumaryBot (commented):
Findings

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1053 @@
...
+}
+
+func uint64EntityKey(v uint64) []byte {
+	key := make([]byte, 8)
+	binary.BigEndian.PutUint64(key, v)
+
+	return key
+}
+
+// runAggregateTargetMisuse issues AGGREGATE_VOLUMES against a query whose
```
NumaryBot, line 384: **🟡 [minor] Stale claim that the aggregate-target rejection carries no error reason**

The comment says the rejection carries no error reason, but this same PR makes it `ErrPreparedQueryAggregateTarget`, a validation sentinel with reason `VALIDATION`, and `classifyPreparedExecError` matches on exactly that reason. A reader trusting the comment could drop the reason server-side and silently break the driver's classification. Update the comment to state the driver keys on the `VALIDATION` reason.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries.go
```diff
@@ -0,0 +1,881 @@
...
+// parameter references.
+//
+// Filters carrying a condition invalid on the target are discarded rather than
+// stored — the FSM rejects those at write time (ValidateFilterForTarget), so
+// storing one would only ever exercise the write-side rejection the ad-hoc
+// generators already cover. Returns nil when the roll produced nothing storable;
+// the caller then emits no op this round.
+func genPreparedQueryFilter(ls oracle.LedgerState, ledger string, target commonpb.QueryTarget) *commonpb.QueryFilter {
+	var concrete *commonpb.QueryFilter
+
```
NumaryBot, line 136: **🔵 [suggestion] `sampleFieldSeeds` lock contract contradicted by new lock-free caller**

`sampleFieldSeeds` documents "Caller holds c.mu", but `genPreparedQueryFilter` now calls it from the bulk generator, which runs outside the lock on a published immutable `GlobalState`. The access is safe, but the documented contract is false at this call site. Follow the `txFilterSeedsOf` pattern from this same diff: extract a lock-free core or fix the comment.

pkg/grpcprotocol/protocol.go
```diff
@@ -12,7 +12,7 @@ const (
 	// Version must change when the client-facing wire format or semantics break.
 	// See docs/technical/architecture/subsystems/api/protocol-compatibility.md.
 	// A compiled constant also identifies local builds without release ldflags.
-	Version = "10"
+	Version = "11"
```
NumaryBot, line 15: **🟠 [major] Version bump contradicts the PR's declared compatibility contract**

The PR states "no `pkg/grpcprotocol.Version` increment" and argues both calls were already errors, but the diff sets `Version = "11"` and adds a doc section conceding revision-10 clients interpret the codes differently. The compatibility doc mandates incrementing for changed error semantics and requires the PR to state the assessment, so the description is stale or the bump unnecessary. Reconcile the description with the actual revision decision.

internal/adapter/grpc/stream_helper_profile_test.go
```diff
@@ -83,7 +83,12 @@ func TestSendPagedToStream_ProfilePhaseAttribution(t *testing.T) {
 	t.Run("stream sends are charged to delivery, outside the server total", func(t *testing.T) {
 		t.Parallel()
 
-		ctx, profile := query.WithProfile(context.Background())
+		const (
```
NumaryBot, line 86: **🟡 [minor] Unrelated profiling-test change outside the declared scope**

This hunk rewrites `TestSendPagedToStream_ProfilePhaseAttribution` timing assertions via `WithProfileStartingAt`/`AddDelivery`, which the PR description does not mention. It looks like a flakiness fix riding along, possibly owned by the stacked PR. Describe it or split it out.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1053 @@
...
+		if bytes.Compare(key(row), after) > 0 {
+			kept = append(kept, row)
+		}
+	}
+
+	return kept
+}
+
+// parseAfterUint reads the model-side cursor for the id-keyed targets; "" means
+// a first page, which starts after id 0.
```
NumaryBot, line 757: **🟡 [minor] Dead helper `parseAfterUint`**

`parseAfterUint` is newly added with no call site in the package and is the file's only use of `strconv`. Delete the helper and its import, or wire it where intended, so the dead code does not drift from the byte-key cursors the validators use.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1053 @@
...
+	}
+	if call.errKind == pqErrAggregateTarget {
+		return false
+	}
+
+	bound, resolved := substituteParams(stored.GetFilter(), call.params)
+	if !resolved {
+		return unresolvedParamRejectionLegal(ls, bound, stored.GetTarget(), call.errKind)
+	}
+
```
NumaryBot, line 534: **🔵 [suggestion] Duplicated per-base prelude in LIST and AGGREGATE validators**

`preparedListOutcomeLegal` and `preparedAggregateOutcomeLegal` repeat the same prelude: wrong-result check, absent-query/`pqErrNotFound` rejection, `substituteParams`, `unresolvedParamRejectionLegal`, then `indexedQueryOutcomeLegal`, differing only in the window predicate. Extracting the shared prelude with a predicate parameter would keep outcome rules in one place.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1053 @@
...
+	pqErrNone pqErrKind = iota
+	pqErrNotFound
+	pqErrIndex
+	pqErrCompilation
+	pqErrAggregateTarget
+	pqErrLedgerNotFound
+	pqErrOther
+)
+
+// classifyPreparedExecError buckets err by its error reason — the same stable
```
NumaryBot, line 158: **🔵 [suggestion] `pqErrAggregateTarget` bucket also swallows unsupported-mode rejections**

`classifyPreparedExecError` maps any `codes.InvalidArgument` + `VALIDATION` error to `pqErrAggregateTarget`, but the executor now also returns `ErrQueryModeUnsupported` with that code and reason. No finding is masked today since the driver only sends known modes, but future mode drift would be misreported in `validateExecuteList`. Key the bucket on a distinguishing reason or assert the mode value.

NumaryBot (changes requested):
Changes requested.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1053 @@
...
+
+// aggregateBucket is one (asset, color) total.
+type aggregateBucket struct {
+	asset string
+	color string
+}
+
+// aggregateMatches folds the base's volume cells over the accounts the filter
+// selects and compares the bucket set and every total exactly.
+func aggregateMatches(ls oracle.LedgerState, bound *commonpb.QueryFilter, agg *commonpb.AggregateResult) bool {
```
NumaryBot, line 847: 🟠 **[major] Reject grouped aggregate results**

When the server attaches any `AggregateResult.Groups`, this validator still accepts the response if its flat totals match. Prepared execution calls `AggregateVolumes` with empty options, so it must return only flat `volumes`; arbitrary grouped buckets are an incorrect client-visible response that this new oracle silently accepts.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1964#issuecomment-5682179924

tests/oracle/model.go
```diff
@@ -1093,6 +1118,15 @@ func logKindFor(req *servicepb.Request) string {
 		return "create_index"
 	case *servicepb.Request_DropIndex:
 		return "drop_index"
+	case *servicepb.Request_CreatePreparedQuery,
```
NumaryBot, line 1125: 🟠 **[major] Validate prepared-query payloads in committed logs**

The new operations deliberately produce top-level log payloads, but this makes `OrderResult.LogID` zero and all existing Apply-payload checks skip them; `validateLifecycleLog` also has no prepared-query cases. Thus a committed create/update/delete can return the wrong or empty audit payload (including wrong query/filter data) and still advance the model as successful. Validate these top-level payload arms in both commit and replay checks, as required for changed response fields by [testing.md:667](docs/technical/contributing/testing.md#L667).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1964#issuecomment-5682179924

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1057 @@
...
+// Execution is forward-only (PaginateForward), so there is no reverse case.
+func preparedWindowMatches(
+	ls oracle.LedgerState,
+	call preparedCall,
+	target commonpb.QueryTarget,
+	bound *commonpb.QueryFilter,
+	after []byte,
+	cur *commonpb.PreparedQueryCursor,
+) bool {
+	if cur.GetHasMore() != (cur.GetNext() != "") {
```
NumaryBot, line 604: 🟠 **[major] Validate echoed cursor fields on prepared pages**

For every successful LIST response, this only checks that `has_more` and `next` agree; it never verifies that `page_size` echoes the requested size or that a continuation's `previous` equals the cursor that was sent. A server returning correct rows but corrupt pagination metadata therefore passes the new model driver, even though clients cannot reliably use the returned cursor. Check these fields for the first and follow-on pages, consistent with the response-field validation rule in [testing.md:667](docs/technical/contributing/testing.md#L667).

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1057 @@
...
+
+	c.mu.Lock()
+	readID := c.registerRead()
+	c.mu.Unlock()
+
+	defer c.finishRead(readID)
+
+	readCtx := metadata.AppendToOutgoingContext(ctx, "x-consistency", "linearizable")
+
+	resp, err := client.ListPreparedQueries(readCtx, &servicepb.ListPreparedQueriesRequest{Ledger: ledger})
```
NumaryBot, line 52: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Freeze the response ticket frontier for prepared-query reads**

These new read paths load `ticketSeq` after the RPC returns without holding `dispatchMu`. A writer can register after the server produced the response but before this load, and `candidateBases` will incorrectly consider that future bulk a legal predecessor. This can mask an incorrect registry/execution response, such as one reflecting a prepared-query update or deletion that was dispatched only after the read completed. Use `beginResponseFrontier()` around every prepared-query RPC (including follow-on and aggregate-misuse calls), as existing linearizable reads do.

tests/oracle/model.go
```diff
@@ -1396,6 +1430,15 @@ func (s *LedgerState) applyOne(req *servicepb.Request, touched map[VolumeKey]boo
 	case *servicepb.Request_DropIndex:
 		return s.applyDropIndex(r.DropIndex)
 
+	case *servicepb.Request_CreatePreparedQuery:
```
NumaryBot, line 1437: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Validate prepared-query Apply log payloads**

The new lifecycle handlers update only model state. The commit checker’s generic ledger-log validation is skipped because `logKindFor` returns `""`, and `validateLifecycleLog` has no prepared-query cases. A server can therefore persist the correct registry but emit a malformed created/updated/deleted prepared-query audit payload (including an incorrect previous filter) without this new model-driver coverage detecting it.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
+)
+
+// roundTripFilters covers every parameterizable arm: address prefix and exact,
+// a reference string, a ledger string, a uint range (two-sided, so each bound
+// rolls independently), a log-id range, and each FieldCondition value kind.
+func roundTripFilters() map[string]*commonpb.QueryFilter {
+	lo, hi := int64(-4), int64(9)
+	ulo, uhi := uint64(3), uint64(77)
```
NumaryBot, line 17: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Cover LOGS date-filter parameter round trips**

The claimed “every parameterizable arm” matrix never constructs `QueryFilter_LogBuiltinUint`: `filterDateRange` is the transaction `BuiltinUint` arm and the log-id case is `LogId`. `parameterizeInto`, `substituteInto`, and `collectParams` each have a separate LOGS date branch, so regressions there are untested; the workload does generate LOGS date filters, but they normally fail at the absent date-index gate before exercising parameter substitution.

NumaryBot (commented):
OpenCodeReview's independent Codex pass added 3 comment-only findings.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1057 @@
...
+type pqErrKind int
+
+const (
+	pqErrNone pqErrKind = iota
+	pqErrNotFound
+	pqErrIndex
+	pqErrCompilation
+	pqErrAggregateTarget
+	pqErrLedgerNotFound
+	pqErrOther
```
NumaryBot, line 155: **🔵 [suggestion] Error classifier conflates unsupported-mode and aggregate-target rejections**

(spec) `classifyPreparedExecError` maps any `codes.InvalidArgument` + `VALIDATION` failure to `pqErrAggregateTarget`, which `validateExecuteList` reports as an unexpected error; both `ErrQueryModeUnsupported` and `ErrPreparedQueryAggregateTarget` are `domain.NewValidationSentinel` errors with reason `VALIDATION`, so an unsupported-mode rejection would surface as a false finding if the driver ever sent a mode outside LIST/AGGREGATE_VOLUMES. Unreachable with current call shaping. Distinguish the two sentinels by their distinct messages or assert the rolled mode before widening the driver.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1057 @@
...
+			kept = append(kept, row)
+		}
+	}
+
+	return kept
+}
+
+// parseAfterUint reads the model-side cursor for the id-keyed targets; "" means
+// a first page, which starts after id 0.
+func parseAfterUint(after string) (uint64, bool) {
```
NumaryBot, line 758: **🟡 [minor] Dead helper `parseAfterUint` is never called**

(standards) `parseAfterUint` has no callers: the prepared id-keyed windows cursor uses `uint64EntityKey` bytes plus `filterPreparedRows`, never a decimal string. The unexported dead function (and its sole `strconv` use) is Speculative Generality that `unused` lint or the next reader will trip over. Delete it and the `strconv` import.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries.go
```diff
@@ -0,0 +1,881 @@
...
+	if m, ok := bc.GetValue().(*commonpb.BoolCondition_Hardcoded); ok {
+		bc.Value = &commonpb.BoolCondition_Param{Param: pz.put(pz.next(), boolParam(m.Hardcoded))}
+	}
+}
+
+// parameterizeUint moves each present bound onto a parameter independently, so
+// a two-sided range can end up half hardcoded. The bound itself is cleared: the
+// compiler reads the parameter when the param name is set, and leaving the
+// literal behind would make the inverse ambiguous.
```
NumaryBot, line 294: **🔵 [suggestion] Duplicated Code: uint/int parameterize-substitute pairs**

(standards) `parameterizeUint`/`parameterizeInt`, `substituteUint`/`substituteInt`, and `lookupUint`/`lookupInt` are line-for-line twins over `UintCondition`/`IntCondition`, differing only in the value constructor. Judgement call (no shared proto interface, so extraction needs a small generic over min/max fields), but the repo standard prefers DRY solutions. Consider one generic helper per phase parameterized by the accessor pair.

NumaryBot (commented):
Findings

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1074 @@
...
+	kept := rows[:0]
+	for _, row := range rows {
+		if bytes.Compare(key(row), after) > 0 {
+			kept = append(kept, row)
+		}
+	}
+
+	return kept
+}
+
```
NumaryBot, line 755: **🟡 [minor] Dead helper parseAfterUint is never called**

`parseAfterUint` has no caller in this diff or the package (spec+standards): id-keyed continuations use byte cursors via `lastPageKey`/`filterPreparedRows`/`uint64EntityKey` instead. The `unused` linter is enabled in `.golangci.yaml`, so it fails `scripts/agent-check`. Delete `parseAfterUint` or use it where a string cursor is decoded.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries.go
```diff
@@ -0,0 +1,881 @@
...
+	}
+
+	out := f.CloneVT()
+	ok := substituteInto(out, params)
+
+	return out, ok
+}
+
+func substituteInto(f *commonpb.QueryFilter, params preparedParams) bool {
+	switch c := f.GetFilter().(type) {
```
NumaryBot, line 378: **🔵 [suggestion] Three parallel type-switches over the QueryFilter grammar**

`parameterizeInto`, `substituteInto`, and `collectParamsInto` each re-walk the same filter grammar, and `substituteInto`'s `default: return true` silently passes unknown arms. Adding a new `commonpb.QueryFilter` arm requires touching all three, and the round-trip test only pins the arms listed in `roundTripFilters`. Consider one shared visitor seam, or make the default arm fail loudly in tests.

NumaryBot (commented):
Findings

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1074 @@
...
+	key := make([]byte, 8)
+	binary.BigEndian.PutUint64(key, v)
+
+	return key
+}
+
+// runAggregateTargetMisuse issues AGGREGATE_VOLUMES against a query whose
+// stored target is not ACCOUNTS. The executor rejects that combination
+// outright, so the only assertion is that it did reject: the rejection carries
+// no error reason, and matching its message text would pin the driver to a
```
gfyrag: [minor] Stale claim contradicted by this PR

"the rejection carries no error reason" — this PR makes the rejection `ErrPreparedQueryAggregateTarget`, a validation sentinel with `Reason() == "VALIDATION"`, and `classifyPreparedExecError` matches on exactly that reason. A future reader who trusts the comment and drops the reason server-side would silently break the driver's classification. Update to reflect the actual contract.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/prepared_queries_read.go
```diff
@@ -0,0 +1,1074 @@
...
+			kept = append(kept, row)
+		}
+	}
+
+	return kept
+}
+
+// parseAfterUint reads the model-side cursor for the id-keyed targets; "" means
+// a first page, which starts after id 0.
+func parseAfterUint(after string) (uint64, bool) {
```
gfyrag: [minor] Dead helper with orphaned import

`parseAfterUint` has no call site in this package. Because it is the only user of `strconv`, removing it also drops the import. Either wire it where intended or delete it before merge.

internal/query/errors.go
```diff
@@ -36,3 +36,11 @@ func (e *ErrAggregateOverflow) Metadata() map[string]string {
 // Compile-time assertion that ErrAggregateOverflow satisfies domain.Describable
 // so it keeps flowing through the shared error edge (gRPC/HTTP mapping).
 var _ domain.Describable = (*ErrAggregateOverflow)(nil)
+
+// Query-mode validation is a read-side concern: these failures are produced
+// only while executing a prepared query and can never be emitted by the FSM.
+// They implement domain.Describable solely to reuse the shared wire conversion.
+var (
+	ErrPreparedQueryAggregateTarget = domain.NewValidationSentinel("AGGREGATE_VOLUMES mode is only valid for ACCOUNTS target queries")
```
gfyrag: [P2][blocking] Define query-layer validation sentinels as concrete query types

`domain.NewValidationSentinel` returns a `*domain.validationSentinel` — an unexported concrete type whose runtime identity lives in `internal/domain`. The intent was to move these errors out of domain and into the query layer, but the current approach only relocates the variable declarations; the concrete type and its factory still belong to domain, so the coupling is not severed.

`ErrAggregateOverflow` in this same file shows the correct pattern: define a concrete struct in the query layer that implements `domain.Describable`, and borrow only the stable wire constant (`domain.ErrReasonAggregateOverflow`) which belongs in the shared contract. Apply the same pattern here: define a local `validationSentinel` in the query package that implements `domain.Describable` directly, so no domain factory is needed and the concrete types stay in their owning layer.

PR #1965, opened 2026-09-10
fix(validation): bound shared Go build cache

gfyrag:
## What changed

`agent-validation-env` now binds each validation or exact-review process to a shared Go build-cache generation. A later run rotates the current generation after a 32 GiB soft budget, while process-identity leases keep in-use retired generations alive and remove them once idle.

The module, GOPATH, lint, and XDG caches remain stable and shared. The PR loop now enters the same managed environment for both validation and review, and both adapters fail closed on unmanaged `GOCACHE` paths.

## Why

The persistent Ledger AI `GOCACHE` had no lifecycle or storage budget and repeatedly grew past 90 GiB after cleanup, eventually contributing to a real local `ENOSPC`. The fix bounds future build-cache generations without serializing tests or discarding warm cross-worktree reuse.

Need: local Ledger correctness campaigns must continue running many validations concurrently without exhausting developer workstations.

Current limitation: all build modes, candidate SHAs, race runs, coverage, and tags accumulated forever in one shared directory.

Requirement / constraint: retain concurrent warm-cache reuse, never remove a cache used by a live process, and reclaim retired generations automatically.

Evidence: `docs/technical/contributing/local-validation.md` and the lifecycle/concurrency regressions in `scripts/agentvalidationenv/cache_test.go`.

## Technical decision

Decision: select a current cache generation under a short filesystem lock, write PID plus process-start identity leases, rotate after a throttled size check, and prune only marked retired generations without live leases.

Why now / why proportionate: the shared cache is a major validation speedup, but its observed growth now causes recurring disk exhaustion. The mechanism is local to the existing environment wrapper and leaves production code untouched.

Alternatives considered: serializing validation would waste available CPU; per-run caches would discard reuse and increase I/O; `go clean -cache` can delete entries used by concurrent processes; TTL-only cleanup cannot establish liveness.

## Risk

**MEDIUM** — contributor tooling and PR validation entry points change, but production behavior is unaffected. Marker, containment, process-identity, stale-lock, and concurrent-initialization checks fail closed.

## Validation

- [x] Targeted race tests for rotation, concurrent initialization, live/stale leases, stale-lock recovery, invalid configuration, reviewer integration, and PR-loop quoting
- [x] `AI_REVIEW_BASE_SHA=8d38782c44b012101961cfb732322729a5163208 bash scripts/agent-check-pr` exercised generation, lint, invariants, compilation, and every changed tooling package; the final run hit two unrelated `rootguard` timing failures under host-wide contention, whose focused `-race -count=5` rerun passed

Local validation cache lifecycle only. Legacy `go-build/` is intentionally not adopted or deleted automatically because older live processes cannot advertise leases.

## Review focus

Please focus on lock/lease lifecycle under process exit and concurrent rotation, plus the deliberate soft-budget behavior while a retired generation remains leased.

## Known concerns

The budget is soft: one run can overshoot it, and active retired generations temporarily consume space in addition to the current generation. This is required to avoid deleting cache entries from live concurrent builds.


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1965#issuecomment-5615419097

scripts/agent-validation-env
```diff
@@ -81,15 +81,225 @@ if [[ "$clean_cache" == true ]]; then
...
+    fi
+fi
+if [[ "$go_cache_generation_valid" == false ]]; then
+    new_go_cache_generation "$now"
+else
+    go_cache_path="$go_cache_generation_root/cache"
+fi
+
+last_check=0
+if [[ -f "$go_cache_check_file" ]]; then
```
NumaryBot, line 332: 🟠 **[major] Reset future cache-check timestamps**

If the wall clock moves backward after writing `check-timestamp`, `now - last_check` remains negative and size checks—and therefore rotation—are suppressed until the clock catches up, potentially leaving an oversized cache indefinitely. Treat timestamps greater than `now` as stale and perform a check.

scripts/agentvalidationenv/cache_test.go
```diff
@@ -47,10 +47,158 @@ func TestSharedCacheEnvironment(t *testing.T) {
...
+	for index := range peers {
+		command := testenv.Command(t, "bash", validationEnvPath(t), filepath.Join(root, fmt.Sprintf("run-%d", index)),
+			"sh", "-c", "printf %s \"$GOCACHE\"")
+		command.Dir = repository
+		command.Env = testenv.Environment("HOME="+root, "LEDGER_AI_CACHE_ROOT="+cacheRoot)
+		command.Stdout = &outputs[index]
+		command.Stderr = &outputs[index]
+		commands = append(commands, command)
+	}
+	for _, command := range commands {
```
NumaryBot: 🟠 **[major] Synchronize concurrent initializers at a barrier**

Starting commands sequentially does not ensure they initialize the cache concurrently; on a loaded runner they may all finish serially, so this test can pass even if concurrent initialization creates multiple generations. Add a bounded barrier before invoking the wrapper, as required by [testing.md:52-55](docs/technical/contributing/testing.md#L52-L55).

paul-nicolas (dismissed):

scripts/agent-validation-env
gfyrag: Fixed in e04c75fd9801b41625c4b1f8d8178859415081f0. Future timestamps now force a fresh size measurement and budget check. Added TestGoCacheClockRollbackForcesBudgetCheck with a nonzero 300-second interval and a timestamp one day ahead. Mutation proof: removing the future-timestamp condition makes this test fail because the oversized generation is retained. Restored fix passes, including focused race tests (3 repetitions).

scripts/agentvalidationenv/cache_test.go
gfyrag: Fixed in e04c75fd9801b41625c4b1f8d8178859415081f0. The 8 initializers now rendezvous on blocking ready/release pipes through testenv.RunSynchronized with a 30-second bound and sibling abort/output collection. This exposed a real namespace-marker publication race, also fixed by atomic publication under the selection lock. The synchronized test passed 100 repetitions; focused cache tests also pass with -race -count=3.

gfyrag:
Addressed all six Shipfox findings in e04c75fd9801b41625c4b1f8d8178859415081f0:

1. Shared contract: scripts/agent-cache-contract now owns the generation layout, regex, marker and canonical path validation used by the producer and both adapters.
2. Invalid configuration tests assert the exact MAX_MIB / CHECK_INTERVAL guard messages.
3. Concurrent initializers use the bounded RunSynchronized ready/release barrier with sibling abort. Stress testing exposed an additional namespace-marker publication race; marker publication is now atomic and serialized.
4. du runs before selection locking and its result is accepted only for the still-current generation. Idle retired generations are detached under the lock and deleted afterward. Cleaner identity lives outside the payload so another run can resume interrupted deletion. Blocking du/find fixtures prove a second validator can complete before the slow operation is released.
5. Ephemeral mode launches an exec-based child wrapper, giving the workload its own PID lease. The SIGKILL regression proves rotation retains the live orphan's generation.
6. Future check timestamps force a fresh budget check.

Validation: canonical agent-check PASS; exact-base agent-check-pr PASS against 8d38782c44b012101961cfb732322729a5163208 (normalization, lint, baseline, all selected tooling suites); synchronized initializer test passed 100 repetitions; final focused cache regressions passed -race -count=3. Clock-rollback and orphan-workload mutation checks both fail when their fixes are temporarily removed. Mutations restored; worktree clean. Updated local-validation.md with the actual lock/retirement/lease sequence.

CI and bot reviews have restarted for the new head. Merge remains gated on green checks plus bot and human approval of the updated version.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1965#issuecomment-5615419097

paul-nicolas (approved):

PR #1966, opened 2026-09-10
fix(operator): preserve backup scheduling after history pruning (EN-2000)

gfyrag:
## What changed

Preserve per-type terminal completion cursors in `Backup.status`, independently of retained `BackupRun` resources. Persist cursors and successful-result summaries before pruning history, and read the parent status uncached so informer lag cannot erase the observed cursor. Incremental scheduling uses the durable successful full summary, so pruning that full run does not disable incrementals.

## Why

Fixes [EN-2000](https://formance-team.atlassian.net/browse/EN-2000). With a daily full schedule and zero history retention, completing and pruning a run caused a fresh controller to create another run immediately, before the next cron deadline. This affected both successful-only and failed-only history.

History limits must affect Kubernetes record retention without changing execution timing. The regression reproduced an extra run at 02:03 despite a next deadline of 02:00 the following day. The durable requirement and failure ordering are recorded in `docs/technical/architecture/subsystems/backup/operator-scheduling.md`.

## Technical decision

Store the greatest observed terminal completion time separately for full and incremental runs, including failures and manual executions. Evaluate the current cron expression against that cursor and retain existing Pending/Running/uninitialized concurrency guards. Write complete status before pruning so a failed status write preserves the child evidence needed on retry.

Using success summaries as cursors would miss failed-only history and conflate results with scheduling. Using next-deadline fields as cursors would conflate cron-derived output with execution progress, especially during active runs and schedule changes.

## Risk

**LOW** — scoped to Kubernetes backup scheduling and retention; deterministic failure/restart regressions cover the changed ordering.

## Validation

- [x] Current-code reproduction: both successful and failed zero-retention cases failed on an early run after restart.
- [x] `AI_REVIEW_BASE_SHA=8d38782c44b012101961cfb732322729a5163208 bash scripts/agent-check-pr`
- [x] Operator unit tests with `-race` under Nix.
- [x] Backup controller integration tests with `-race -tags integration` under Nix.
- [x] Independent review and self-review of code, tests, docs, CRD and deepcopy diff.

Deterministic regressions also reproduce stale parent informer reads after child deletion for both success and failure, and verify uncached API read fail-then-success recovery. They exercise two daily success/failure cycles, restart before and at the deadline, status/delete fail-then-success recovery, active-run exclusion, independent incremental progress, disabled/changed schedules and older retained history.

Adds optional `lastFullRunCompletionTime` and `lastIncrementalRunCompletionTime` status fields to the Backup CRD; regenerated deepcopy and both standalone/Helm CRDs. Successful-result summaries and next-deadline fields retain their meanings. Kubernetes operational state only: no service gRPC contract, protocol revision, Ledger storage or restore-format changes. Pre-release change requires no migration or compatibility fallback.

## Review focus

Uncached parent reads across child informer deletion events, persist-before-prune ordering, monotonic cursors across failed runs and restart, and incremental eligibility after successful full history is removed.

## Known concerns

None. No merge requested.

[EN-2000]: https://formance-team.atlassian.net/browse/EN-2000?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Bugfix evidence

DISCOVERY: NO_EXISTING_WORK — EN-2000 and existing branches/PRs checked before implementation.
BEFORE_FIX: BUG_REPRODUCED — deterministic daily schedule tests for Succeeded and Failed both created a run at 02:03 after pruning/restart, before the next 02:00 deadline; original scheduler logic with an explicit reconciliation clock.
AFTER_FIX: PASS — zero-retention restart regressions, operator race suite, Backup integration suite, agent-check and exact-base agent-check-pr passed.

The post-review cache-lag regression failed against commit `04a1263e6` for both outcomes before switching parent reads to the API server.

### DIFF
--- a/misc/operator/internal/controller/backup_controller.go
+++ b/misc/operator/internal/controller/backup_controller.go
@@ -38,12 +38,13 @@ const (
 // When a cron schedule fires, it creates a BackupRun whose own reconciler
 // performs the actual ledgerctl invocation. It also maintains Backup status
 // summaries from the latest Succeeded child runs and prunes runs in excess of the
-// configured history limits.
+// configured history limits. Completion cursors survive pruning; parent and child
+// reads bypass informer caches to avoid restoring stale scheduling/status data.
 type BackupReconciler struct {
 	client.Client
 
-	// APIReader reads Backup status directly from the API server. Child deletion
-	// events can reach the cache before the status write that preserved its cursor.
+	// APIReader reads Backups and child runs directly from the API server.
+	// Independent informer lag must not rewind durable cursors or summaries.
 	APIReader client.Reader
 
 	Scheme    *runtime.Scheme
@@ -340,7 +341,7 @@ func (r *BackupReconciler) listChildRuns(
 	backup *ledgerv1alpha1.Backup,
 ) ([]ledgerv1alpha1.BackupRun, error) {
 	var list ledgerv1alpha1.BackupRunList
-	if err := r.List(ctx, &list,
+	if err := r.APIReader.List(ctx, &list,
 		client.InNamespace(backup.Namespace),
 		client.MatchingLabels{ledgerv1alpha1.LabelBackup: backup.Name},
 	); err != nil {
--- a/misc/operator/internal/controller/backup_controller.go
+++ b/misc/operator/internal/controller/backup_controller.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sort"
 	"time"
@@ -41,6 +42,10 @@ const (
 type BackupReconciler struct {
 	client.Client
 
+	// APIReader reads Backup status directly from the API server. Child deletion
+	// events can reach the cache before the status write that preserved its cursor.
+	APIReader client.Reader
+
 	Scheme    *runtime.Scheme
 	Config    *rest.Config
 	Clientset kubernetes.Interface
@@ -53,8 +58,12 @@ func (r *BackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 func (r *BackupReconciler) reconcile(ctx context.Context, req ctrl.Request, now time.Time) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 
+	if r.APIReader == nil {
+		return ctrl.Result{}, errors.New("APIReader not configured")
+	}
+
 	var backup ledgerv1alpha1.Backup
-	if err := r.Get(ctx, req.NamespacedName, &backup); err != nil {
+	if err := r.APIReader.Get(ctx, req.NamespacedName, &backup); err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
--- a/misc/operator/internal/controller/backup_controller.go
+++ b/misc/operator/internal/controller/backup_controller.go
@@ -47,6 +47,10 @@ type BackupReconciler struct {
 }
 
 func (r *BackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	return r.reconcile(ctx, req, time.Now())
+}
+
+func (r *BackupReconciler) reconcile(ctx context.Context, req ctrl.Request, now time.Time) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 
 	var backup ledgerv1alpha1.Backup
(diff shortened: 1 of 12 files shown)


gfyrag:
Verified both current Shipfox comments against the code:

- **Run-type duplication:** keeping the explicit cases in this focused fix. `misc/operator/api/v1alpha1/backuprun_types.go:18-24` admits exactly Full and Incremental; both are handled in scheduling, cursor refresh, creation and pruning. A hypothetical third type would require a broader execution/API change. This is optional maintenance rather than a current correctness defect.
- **Manual-run coverage:** added `TestBackupScheduling_ManualRunAdvancesCursor` in commit `9189a6336` for both successful and failed labeled manual Full runs, using the same parent-label/BackupRef association as `kubectl-ledger backup trigger` without scheduler ownership. It verifies durable cursor advancement, zero-retention pruning, restart before the deadline, and execution at the next deadline.

The earlier informer-lag finding was reproduced for both outcomes and fixed in `8d5fdf2c3` using the uncached parent reader, with error/retry regressions. The remaining association with labels is the existing API contract; this change preserves it.

Validation: `agent-check`, exact-base `agent-check-pr`, and the scheduling regressions with `-race` all pass.

gfyrag:
Verified the remaining Shipfox notes on `d9c26ec1733c6a598e5a104235939c6ad6c8491b`:

- The struct overview remains accurate about successful-result summaries. The new durability responsibilities are documented at their implementation boundaries: the `APIReader` field comment explains the informer ordering hazard, `refreshCompletionCursors` documents failure-inclusive monotonic progress, and `docs/technical/architecture/subsystems/backup/operator-scheduling.md` specifies the full ordering and failure states. I do not consider repeating those details in the struct overview necessary to correct misleading documentation.
- The mapping suggestion remains optional: the API enum admits only Full and Incremental, and both are handled at every site. No missing supported type or current scheduling defect was found. Keeping the explicit cases for this focused fix.

Native review approves this exact HEAD against base `dfa416b8216c0418ce7c18c9a01f50c74b9c34b4`, with zero actionable findings; canonical validation and CI pass. Awaiting the required human approval before merge.

misc/operator/internal/controller/backup_controller.go
```diff
@@ -99,7 +112,11 @@ func (r *BackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, fmt.Errorf("listing child runs: %w", err)
 	}
 
-	now := time.Now()
+	// Capture terminal progress before history can be pruned, even with schedules
+	// disabled. Successful summaries remain distinct from scheduling cursors.
+	r.refreshCompletionCursors(&backup, runs)
+	r.refreshStatusSummaries(&backup, runs)
```
NumaryBot, line 119: 🟠 **[major] Prevent stale child cache from rewinding summaries**

When the uncached parent read observes a newer persisted status but the cached `BackupRun` list still contains an older successful run, `refreshStatusSummaries` unconditionally replaces `LastFullBackup`/`LastIncrementalBackup` and health conditions with that stale child's data. Kubernetes provides no cross-resource informer ordering, so the subsequent status update can durably rewind a correct summary; use a consistent child read or only advance summaries monotonically.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1966#issuecomment-5615460507

misc/operator/internal/controller/backup_controller.go
gfyrag: Confirmed and fixed in 08096148a. A deterministic regression supplies a fresh parent with a newer durable summary and a stale cached child list containing an already-pruned older success; it failed for both Full and Incremental before the fix. Child listing now uses APIReader as well, so stale informer data cannot overwrite the fresh parent summary. Updated the ordering documentation and reconciler comments. Scheduling race regressions, agent-check and exact-base agent-check-pr all pass (including operator race tests and lint).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1966#issuecomment-5615460507

gfyrag:
Verified the latest Shipfox comments on 08096148a:

1. **The stale-cache tests do detect the regression.** Keeping stale data on Client and fresh data on APIReader is intentional: these are two different read surfaces. With the fix, the stale surface is bypassed; reverting to r.Get/r.List makes that same interceptor execute. Both tests were run before their respective production fixes and failed: StaleParentCacheAfterPruning failed for Succeeded and Failed because an early run was created; StaleChildCacheDoesNotRewindSummaries failed for Full and Incremental because the latest summary timestamp was lost. The same regressions pass with the fixes, including with -race. Thus the claim that they also pass with cached reads is contradicted by execution. Moving the interceptor onto APIReader would instead simulate a stale direct API response, not informer lag. No test change warranted.
2. **Type mapping:** still an optional future refactor; both admitted types are handled.
3. **Manual-run reconciliation delay:** correct observation of the existing unowned-run watch behavior. The next reconcile reads the retained manual result and advances the cursor before scheduling/pruning. This does not invalidate the scheduling contract, and changing watches is outside this fix.

Current exact-HEAD native review: APPROVE, zero findings; CI green. Awaiting required human approval.

paul-nicolas (approved):

PR #1968, opened 2026-09-10
fix(operator): recover backup Jobs after lost creation responses (EN-2001)

gfyrag:
## What changed

Retryable provisioning errors, including foreign/unowned Job conflicts, are now visible in `BackupRun.status.message`. Identical diagnostics do not rewrite status; the message clears when the owned Job becomes available. Status-write failures preserve both the provisioning cause and write error for retry.

BackupRun now persists its Running reservation before creating a Job and retries ambiguous provisioning errors. Recovery uses the deterministic Job name and verifies the controller owner UID before consuming its outcome. Definitive construction and API validation errors remain terminal; transient Backup/Cluster reads remain retryable during recovery.

## Why

Fixes [EN-2001](https://formance-team.atlassian.net/browse/EN-2001). When the API server committed Job creation but its response was lost, the controller persisted Failed and its terminal guard ignored the eventual successful Job result.

A run must remain attached to the outcome of a potentially committed Job, and waiting sibling runs must stay excluded while that outcome is unknown. The regression reproduces commit, timeout, successful Failed-status persistence, and ignored completion on the original controller. The recovery contract and execution order are documented in `docs/ops/backup-restore.md` and linked from `misc/operator/README.md`.

## Technical decision

Persist Running before Create so the execution slot survives lost responses and controller restart. Retry the same deterministic Job identity, including AlreadyExists while the cache catches up, and require the current BackupRun controller UID. Only local construction failures and Create Invalid/BadRequest responses prove a definitive provisioning failure. Keeping the terminal guard preserves final outcomes; removing it would leave the underlying false failure and sibling-admission window unresolved.

## Risk

**LOW** — confined to BackupRun provisioning and recovery. Running now includes provisioning; persistent transient errors retain the sibling exclusion until recovery. No service protocol, storage format, FSM, or dependency changes.

## Validation

- Original behavior reproduced: committed owned Job, lost response, successful Failed status, valid completed Job ignored by a fresh controller.
- Mutation check: the recovery regression fails with the original controller restored.
- `bash scripts/agent-check`
- `AI_REVIEW_BASE_SHA=fada86bcfc1c7f228ede10151fd5f718c2a0beb8 bash scripts/agent-check-pr`
- `go -C misc/operator test -race ./internal/controller -count=1`
- `go -C misc/operator test -race -tags=integration ./internal/controller -run '^TestBackupRunCommittedCreateLostResponseIntegration$' -count=1`

The integration test uses a real API server for Job creation and Pod/Job status writes. It proves sibling exclusion, recovery after controller replacement, preserved result fields and Job UID, and exactly one Job creation. Additional tests cover a stale read followed by AlreadyExists, a failed reservation write, wrong/missing owners, transient dependency reads, and definitive versus retryable creation errors. Tests retain parallel execution.

Running is persisted before the external Job effect. Unknown effects remain reconcilable; existing Jobs are accepted only for their owning run. Final outcomes remain terminal.

## Review focus

Error classification, reservation-before-Create ordering, and recovery without consuming a foreign Job or creating a duplicate.

## Known concerns

None.

[EN-2001]: https://formance-team.atlassian.net/browse/EN-2001?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Bugfix evidence

```text
DISCOVERY: PARTIAL_FIX — the EN-2001 implementation was already present; Shipfox identified missing status diagnostics for foreign Jobs.
BEFORE_FIX: BUG_REPRODUCED — on 8d38782c44b012101961cfb732322729a5163208, committed Job Create followed by DeadlineExceeded and successful Failed status leaves completed Job result ignored by a fresh reconciler.
AFTER_FIX: PASS — original recovery regressions, additive diagnostic/recovery and fail-once status-write tests, controller race suite, real API-server integration, agent-check and exact-base agent-check-pr passed for 193c05691. The new tests failed before the follow-up fix because status.message remained empty.
```

### DIFF
--- a/docs/ops/backup-restore.md
+++ b/docs/ops/backup-restore.md
@@ -59,7 +59,14 @@ siblings continue waiting. Reconciliation retries the same name and accepts an
 existing Job only when its controller owner UID matches the run. `AlreadyExists`
 (including while the read cache catches up) is retryable; it never causes creation
 under another name. A foreign or unowned Job is left untouched and reported as a
-retryable ownership error. Definitive local construction errors and API
+retryable ownership error in `BackupRun.status.message`. The run remains Running
+and keeps siblings waiting; an operator must resolve the conflicting Job's
+ownership or remove that Job before provisioning can resume. The controller never
+adopts or deletes a foreign Job automatically. Retryable provisioning errors are
+persisted in the message without rewriting unchanged diagnostics. If that status
+write fails, the run stays nonterminal and reconciliation retries. The message is
+cleared after the owned Job becomes available, before processing its outcome.
+Definitive local construction errors and API
 `Invalid`/`BadRequest` creation rejections mark the run Failed; other provisioning
 errors retry. Transient reads of the parent Backup or Cluster also retry rather
 than ending the run during recovery. Once the owned Job completes, its outcome
--- a/docs/ops/backup-restore.md
+++ b/docs/ops/backup-restore.md
@@ -46,6 +46,26 @@ spec:
   failedRunsHistoryLimit: 1
 ```
 
+Each `BackupRun` uses one deterministic Job name and a controller owner reference
+bound to the run UID. To keep a run attached to its actual outcome (EN-2001), the
+controller first checks for a Running sibling, then persists this run as Running
+before attempting Job creation. If that status write fails, it does not submit
+Create. Running therefore includes provisioning and reserves the execution slot
+while a creation outcome is unknown.
+
+A timeout or connection failure after Create may mean the API server committed
+the Job but its response was lost. The run remains nonterminal and Running, so
+siblings continue waiting. Reconciliation retries the same name and accepts an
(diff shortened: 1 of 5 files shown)


gfyrag:
Addressed the [ownership-conflict observability comment](https://github.com/formancehq/ledger/pull/1968#issuecomment-5615682280) in 193c05691. Retryable provisioning errors now appear in `BackupRun.status.message`, remain nonterminal, and clear once the owned Job is available. Unchanged messages do not trigger redundant status writes.

Added regression coverage for owned/unowned conflicts, recovery after an operator removes the foreign Job, and a failed-then-successful diagnostic status write. The tests failed before the fix and pass now; the controller race suite, Kubernetes integration regression, `agent-check`, and exact-base `agent-check-pr` also pass.

PR #1970, opened 2026-09-10
fix(EN-1632): expose typed public audit views without raw evidence

gfyrag:
## What changed

Audit reads now return a dedicated `publicaudit.AuditEntry` with typed order details. Public responses omit serialized orders, signed batch payloads and signature bytes. Configuration fields use the normalization and sensitivity mechanisms supplied by the lower stack layers.

Stack **3/3**: #1977 → #1979 → **this PR**. The base is `codex/structured-connection-configs`, so this review contains only the audit layer. The alternative #1963 remains open for comparison. Merge bottom-up and restack after squash merges.

Review size: **46 files**, down from the combined 132-file PR; generated protobuf code accounts for about 80% of changed lines.

## Why

Raw audit bytes can contain connection credentials even after configuration endpoints are masked. A typed public schema keeps audit details usable without returning those opaque credential-bearing envelopes (EN-1632, EN-1634, EN-1635).

Clients can inspect ordering, outcome, caller, business intent and signing-key identity without receiving protected raw evidence. The server retains exact original audit and signed bytes for its integrity and recovery workflows.

Durable contract: [public audit projection](https://github.com/formancehq/ledger/blob/codex/structured-credential-projections/docs/technical/architecture/subsystems/api/structured-credentials.md).

## Technical decision

Project internal audit records into a separate public protobuf model at the controller read boundary, for live and checkpoint queries. Decode the canonical serialized order; normalize configuration-bearing orders, then apply generic masking. Malformed rejected configurations are omitted with `configurationUnavailable`, without raw-input or parser-error fallback.

Missing/nil audit outcomes and nil entries fail with an explicit integrity error rather than returning an incomplete public entry.

The displayed hash identifies original evidence and `signature.keyId` identifies its signer; neither authenticates the redacted view. Original records are never rewritten.

## Risk

**HIGH** — incompatible audit response contract spanning gRPC, HTTP and CLI. No changes to authoritative audit persistence or signature generation.

## Validation

- Restacked on release/v3.0 `12f48ef84e7a82c2a9e96784dd069bd9bf320f36` through the intermediate PR bases. Range-diff confirms that upstream changes, including PostgreSQL initialization cancellation, are preserved.
- Protobuf regeneration is clean. Focused race suites for normalization, public audit projection, admission, mirror lifecycle, v2 adapters and controllers pass after rebase.
- Fragment and invalid-outcome regressions failed before their fixes and pass afterward. An overlay confirms that malformed IAM input previously produced the wrong TLS diagnostic.
- Full canonical gates and fresh GitHub checks are being run for the new candidate; previous-head approvals and aggregate results are not treated as current approval evidence. No assertions or validation requirements were weakened.

Service protocol **11 → 12** because both audit RPCs (ListAuditEntries, GetAuditEntry) replace raw audit.AuditEntry with typed publicaudit.AuditEntry views. Internal query/storage/backup/checker evidence remains raw and unchanged. OpenAPI and CLI documentation describe the same public model.

## Review focus

Sink diagnostics are sanitized by adaptors in #1977 and remain visible through public projections; HTTP, gRPC and CLI assertions preserve useful messages. Mirror and audit diagnostics retain their independent annotations.

Exhaustive typed-order coverage, absence of paths to raw input or signature payloads, rejected-order handling, cursor lifetime, and preservation of original evidence.

## Known concerns

The audit-first checker pivot and verification of all primary-store projections remain separate in #1912. The current checker does not cryptographically verify client/server signatures or prove completeness of a public response.

Latest validation: `agent-check-pr` passed on `cd5d62ee1` against exact base `aeee0ab75` (normalization, baseline, race, business/cluster E2E and Schemathesis). `TestAuditCursorErrorsIdentifyEntryAndOrder` failed before the error-context fix and passes afterward. Current-head CI and fresh reviews are pending.


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1970#issuecomment-5619568448

internal/application/readprojection/audit.go
```diff
@@ -0,0 +1,145 @@
...
+	if entry == nil {
+		return nil, nil
+	}
+	result := &publicauditpb.AuditEntry{
+		Sequence: entry.GetSequence(), Timestamp: entry.GetTimestamp(), ProposalId: entry.GetProposalId(),
+		OrderCount: entry.GetOrderCount(), Ledgers: slices.Clone(entry.GetLedgers()),
+		Hash: slices.Clone(entry.GetHash()), HashVersion: entry.GetHashVersion(),
+		CallerSnapshot: entry.GetCallerSnapshot(), Idempotency: entry.GetIdempotency(),
+	}
+	switch outcome := entry.GetOutcome().(type) {
```
NumaryBot, line 33: 🟠 **[major] Reject audit entries without a recognized outcome**

If stored evidence is corrupted or malformed so that `outcome` is unset or unknown, this switch silently returns a public entry with neither success nor failure, violating the audit contract instead of surfacing the invariant failure. Add a default/error path (and reject nil oneof payloads) as required by [AGENTS.md:52](AGENTS.md#L52).

gfyrag: Fixed in e8a255722. The projection now rejects unset outcomes, nil success/failure wrappers, nil payloads and a nil audit entry with explicit errors. `TestAuditRejectsMissingOutcome` failed against the previous implementation and passes after the fix. Existing order-decoding regressions now use valid outcome fixtures so they still prove the original decoding failure path. Internal evidence and the public protobuf schema are unchanged.

gfyrag:
Addressed the inline outcome finding and all global NumaryBot/Shipfox feedback in e8a255722:

- Missing/nil outcomes and a nil entry fail explicitly. The regression failed before the change and passes now, alongside live/checkpoint controller projection tests. Existing malformed-order tests retain valid outcome fixtures and still reach their decoding error.
- Added `audit_view.proto` and `publicauditpb/` to the protobuf location tables and documented the integrity-error behavior.
- Corrupt serialized orders continue to fail detail reads; only non-normalizable configuration degrades to `configurationUnavailable`. This preserves the fail-loud contract and is not broadened to returning partial evidence.

Restacked on #1979; original internal audit evidence remains unchanged.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1970#issuecomment-5619568448

gfyrag:
Restacked on #1979; current HEAD `a396bf93a` includes the current release/v3.0 base `12f48ef84`. GitHub now reports MERGEABLE. The review fixes are preserved by range-diff, including upstream PostgreSQL initialization cancellation. Fresh exact-base gates and bot/CI checks are in progress; earlier approvals are not reused.

gfyrag:
Follow-up in 603c26cce6c3902f747f0b0f09c42f3d91e0bc98, restacked on #1979 at 8e10833372054b2d3a96d2f9bed8c1dc66d39bf9:

- `GetAuditEntry` now wraps public projection errors with the entry sequence and operation while retaining the cause.
- The durable public-projection contract explicitly includes undecodable order bytes and unrecognized scope as read integrity errors. Corrupt evidence is not silently converted into a valid-looking partial order.
- Both normalization arms now explain why parser error text is deliberately discarded: it may contain credentials. They retain explicit typed assignments instead of a new generic abstraction for two small branches.
- The contract explicitly records that a rejected sink creation omits the entire configuration, including its nested name; fields outside configuration remain. This takes the review's offered documentation option and avoids introducing a second public sink identity field.
- The HTTP and gRPC summaries remain near their respective endpoints for readers of either protocol; the shared structured-credentials document remains the authoritative detailed contract.

Canonical `agent-check-pr` passed on this exact HEAD/base, including normalization, baseline, race suites, business/cluster E2E and Schemathesis. New reviews and CI must assess these current heads.

gfyrag:
Rebased/restacked at `03c17f076` (base `11e627d7f`). The only manual resolution combines the events documentation index entries; upstream HTTP redirect acknowledgement policy and adaptor diagnostic sanitization are both retained. Canonical `agent-check-pr` passed on this exact HEAD/base. Please review the current HEAD; prior approvals are not used as current evidence.

gfyrag:
Fixed protocol documentation on 782065e67: the current revision and typed audit transition both say 9; structured connection messages correctly belong to revision 8. Restacked on aeee0ab75. Canonical agent-check-pr passed against that exact base, including normalization, baseline, race, business/cluster E2E and Schemathesis. The two small typed normalization arms remain explicit so each order case keeps its omission/error handling visible; no generic abstraction is introduced for two arms. Please assess the current HEAD.

gfyrag:
Addressed the list-error context finding in cd5d62ee1: decode and order-projection failures now identify both entry sequence and order index, including through NewAuditCursor. The regression test fails before the fix on both malformed bytes and missing scope, then passes with race enabled. Canonical agent-check-pr passes against exact base aeee0ab75. The shared ordering-only success type is now explicitly documented at its assignment and in the contract; changes to it require public-API review. The contract also clarifies that absent config stays absent without configurationUnavailable, which reports normalization failure. The two typed normalization arms remain explicit as previously explained. Please review the current HEAD.

gfyrag:
Rebased on `release/v3.0` at `1f86c44ac` (base: updated `codex/structured-connection-configs` at `fc497969b`). New HEAD: `cd75e79d4`.

Conflict resolutions:
- `pkg/grpcprotocol/protocol.go`: kept upstream version "10" (our commit targeted "9").
- `misc/proto/bucket.proto` `ListAuditEntries`/`GetAuditEntry`: merged HEAD's `auth_policy` annotations with our `publicaudit.AuditEntry` return types.
- `docs/technical/architecture/subsystems/api/protocol-compatibility.md`: kept upstream "10" in examples.
- `docs/technical/architecture/subsystems/api/README.md`: kept our updated `structured-credentials.md` description.
- `docs: align public audit protocol revision examples` commit (empty after resolution): skipped cleanly.

Proto regenerated with full `publicauditpb` package and auth-policy-annotated audit RPCs.

gfyrag:
@NumaryBot review

gfyrag:
Restacked onto new `codex/structured-connection-configs` (c64ada47b). Resolved: `bucket.proto` RPC return type merge (publicaudit.AuditEntry with auth_policy option from trunk), proto regeneration. Current HEAD: `108da14b8`. CI triggered.

docs/technical/architecture/subsystems/api/README.md
```diff
@@ -1,25 +1 @@
...
...on maintenance (EN-1851). |\n| [http-api.md](http-api.md) | HTTP endpoint lifecycle and failure shutdown, REST API endpoints, single-decoding metadata key paths, response formats, exact metadata number decoding, ledger-log JSON output and discriminators, error handling, and the / error boundary crossed by a forwarded write. |\n| [structured-credentials.md](structured-credentials.md) | Structured stored connection configurations, sensitive annotations and public audit views. |\n| [auth.md](auth.md) | Client JWT authentication (OIDC + Ed25519), scope-based authorization, and the Raft inter-node cluster-secret auth layer. |\n| [sensitive-projection.md](sensitive-projection.md) | Detached, descriptor-driven secret projection and its ownership boundary. |\n\nDecoded leader errors retain their exact gRPC status through routing wrappers\nand streaming cursors. See [forwarding and cancellation](http-api.md#forwarded-writes-and-the-transport-seam)\nfor the distinction between a decoded rejection and raw transport cancellation.\n\n## Related\n\n- [Admission](../admission/) — what every write request enters next.\n- [Scripting](../scripting/) — numscript library available through the API.\n\n
```
NumaryBot: README rewritten as one line of literal \n escapes

The entire `docs/technical/architecture/subsystems/api/README.md` was replaced by a single line containing literal `\n` escape sequences, with inline code formatting stripped (e.g. empty parens in "transport layers (, , )" and the bare `-to-status-code` table name). The document is unreadable as rendered Markdown. Regenerate the file with real newlines and preserved backticks.

NumaryBot (commented):
Findings

docs/technical/architecture/subsystems/api/protocol-compatibility.md
```diff
@@ -191,3 +191,6 @@ forwarding so the gate cannot make the repository's own clients incompatible.
 Structured connection input/output messages require revision 8 on top of the
 revision 7 Apply contract: existing configuration fields change message types.
 
+Typed public audit views require revision 9 on top of structured connections:
+ListAuditEntries and GetAuditEntry return the public audit message contract.
```
NumaryBot: Incompatible audit response change shipped without a protocol revision bump

The diff adds "Typed public audit views require revision 9", but the same document records revision 9 as empty conjunction semantics and revision 10 as deleted-ledger write rejection, and `pkg/grpcprotocol.Version` is "10" with no change to `pkg/grpcprotocol` in this PR. Incompatible clients declaring revision 10 pass the gate while `ListAuditEntries`/`GetAuditEntry` replace their response shape. The document also still says "The audit response contract is unchanged". Bump `pkg/grpcprotocol.Version` in this PR and correct both statements. (Consolidates duplicate spec and standards findings.)

docs/technical/architecture/subsystems/api/README.md
NumaryBot: API README replaced by one line of literal newline escapes

The hunk replaces the 25-line subsystem index with a single line of literal `\n` sequences, stripped inline-code backticks and no trailing newline, so the document renders as corrupted text with broken descriptions (e.g. "layers (, , )"). Regenerate the file as proper Markdown with real line breaks, backticked package names and links. (Consolidates duplicate spec and standards findings.)

internal/application/readprojection/audit.go
```diff
@@ -0,0 +1,158 @@
...
+				InitialSchema: create.GetInitialSchema(), Mode: create.GetMode(), MirrorSource: config,
+				AccountTypes: create.GetAccountTypes(), DefaultEnforcementMode: create.GetDefaultEnforcementMode(),
+				ConfigurationUnavailable: err != nil,
+			}}
+		} else if err := copyOrderPayload(source.ProtoReflect(), target.ProtoReflect()); err != nil {
+			return nil, err
+		}
+		result.Type = &publicauditpb.Order_LedgerScoped{LedgerScoped: target}
+	case *raftcmdpb.Order_SystemScoped:
+		source := value.SystemScoped
```
NumaryBot, line 94: Duplicated normalize-and-flag shape across configuration arms

The `CreateLedger` and `AddEventsSink` arms repeat the same shape: normalize via `connectionconfig`, drop the config on error, and set `ConfigurationUnavailable: err != nil` from a discarded diagnostic. Extracting a shared helper would keep the failure-only flag semantics in one place. Judgement call (Duplicated Code), not a hard violation.

openapi.yml
```diff
@@ -5824,16 +5828,981 @@ components:
...
+              - ERROR_REASON_WRITES_BLOCKED_DISK_FULL
+              - ERROR_REASON_WRITES_BLOCKED_CLOCK_SKEW
+              - ERROR_REASON_CHECKPOINT_NOT_READY
+              - ERROR_REASON_MIRROR_V2_LOG_ID_GAP
+              - ERROR_REASON_MIRROR_V2_LOG_ID_INVALID
+              - ERROR_REASON_STALE_INPUTS_RESOLUTION
+              - ERROR_REASON_PRELOAD_UNAVAILABLE
+              - ERROR_REASON_AGGREGATE_OVERFLOW
+              - ERROR_REASON_BALANCE_NOT_FOUND
+              - ERROR_REASON_STALE_CLUSTER_POLICY
```
NumaryBot, line 6225: AuditLedgerApplyOrder.skippableReasons enum is missing two ErrorReason values

The newly added `AuditLedgerApplyOrder.skippableReasons` enum ends at `ERROR_REASON_SEQUENCE_EXHAUSTED`, but `misc/proto/common.proto` at this head also defines `ERROR_REASON_INDEX_ALREADY_EXISTS` (68) and `ERROR_REASON_METADATA_LIMIT_EXCEEDED` (69), and `raft.LedgerApplyOrder.skippable_reasons` is typed as the full `common.ErrorReason` enum. OpenAPI-validated clients reject audit views whose apply orders carry those two skippable reasons, and generated clients cannot round-trip them. Regenerate (or extend) the enum to match `common.ErrorReason`.

NumaryBot (commented):
Findings

docs/technical/architecture/subsystems/api/README.md
NumaryBot: **🟠 [major] API README rewritten as one literal-\n line with code spans stripped**

Verified at head: the file is a single line of literal `\n` sequences with all inline code stripped (e.g. `(, , )` for the transport layers, `-to-status-code` for the error table) and no trailing newline, destroying headings, tables and links in the API subsystem index referenced from grpc-api.md and http-api.md. Regenerate the file with real newlines and preserved backtick formatting, keeping only the intended `structured-credentials.md` description update. (Consolidates duplicate spec and standards findings.)

docs/technical/architecture/subsystems/api/protocol-compatibility.md
```diff
@@ -191,3 +191,6 @@ forwarding so the gate cannot make the repository's own clients incompatible.
 Structured connection input/output messages require revision 8 on top of the
 revision 7 Apply contract: existing configuration fields change message types.
 
+Typed public audit views require revision 9 on top of structured connections:
```
NumaryBot: **🟠 [major] Revision-9 claim conflicts with the revision ledger and unchanged Version constant**

The added paragraph says typed public audit views `require revision 9`, but the same document already assigns revision 9 (empty conjunction) and revision 10 (deleted-ledger rejection), and `pkg/grpcprotocol.Version` is `"10"` at head with no bump in this diff; the PR description's `8 -&gt; 9` claim is stale. Replacing both audit RPC response types is an incompatible service contract, so per this document's maintenance rule the constant must be incremented from the target branch's current value in the same PR. Bump `pkg/grpcprotocol.Version` (to 11) or correct the paragraph to the revision actually consumed, and fix the stale 8-&gt;9 claim.

internal/application/readprojection/audit.go
```diff
@@ -0,0 +1,158 @@
...
+		source := value.LedgerScoped
+		if source == nil {
+			return nil, errors.New("audit order contains a nil ledger scope")
+		}
+		target := &publicauditpb.LedgerScopedOrder{Ledger: source.GetLedger()}
+		if create := source.GetCreateLedger(); create != nil {
+			config, err := connectionconfig.Mirror(create.GetMirrorSource())
+			if err != nil {
+				// Parser diagnostics can contain credentials; expose only the flag.
+				config = nil
```
NumaryBot: **🟡 [minor] Normalization failures are swallowed with no server-side signal**

Both configuration arms (`connectionconfig.Mirror` and `connectionconfig.Sink`) discard the normalization error and only set `ConfigurationUnavailable`; the comment justifies hiding parser diagnostics, not suppressing every server-side signal. A normalizer regression would silently mark configurations unavailable across all audit reads with no way to detect it. Add a detail-free debug/warn log keyed by entry sequence and order index.

NumaryBot (changes requested):
Changes requested.

docs/technical/architecture/subsystems/api/protocol-compatibility.md
```diff
@@ -191,3 +191,6 @@ forwarding so the gate cannot make the repository's own clients incompatible.
 Structured connection input/output messages require revision 8 on top of the
 revision 7 Apply contract: existing configuration fields change message types.
 
+Typed public audit views require revision 11 on top of structured connections (current value):
```
NumaryBot: **🟡 [minor] Protocol bump disagrees with PR description**

The PR description states "Service protocol 8 → 9", but the diff bumps `pkg/grpcprotocol.Version` from "10" to "11" and documents revision 11 in `protocol-compatibility.md`; code and docs are internally consistent. Update the PR description and release notes to revision 11 (Spec provenance).

internal/application/readprojection/audit.go
```diff
@@ -0,0 +1,158 @@
...
+				AccountTypes: create.GetAccountTypes(), DefaultEnforcementMode: create.GetDefaultEnforcementMode(),
+				ConfigurationUnavailable: err != nil,
+			}}
+		} else if err := copyOrderPayload(source.ProtoReflect(), target.ProtoReflect()); err != nil {
+			return nil, err
+		}
+		result.Type = &publicauditpb.Order_LedgerScoped{LedgerScoped: target}
+	case *raftcmdpb.Order_SystemScoped:
+		source := value.SystemScoped
+		if source == nil {
```
NumaryBot, line 95: **🔵 [suggestion] Unregistered sink type panics during audit read projection**

`connectionconfig.Sink` panics (`missing sink normalizer`) when a stored order carries a sink type absent from the compiled registry, so `publicOrder` would panic instead of failing the read. Admission guarantees a normalizer exists, but returning an integrity error here would keep the documented "integrity error, never panic" contract robust across partial upgrades (Spec provenance).

cmd/ledgerctl/audit/list.go
```diff
@@ -207,28 +206,15 @@ func printAuditEntry(entry *auditpb.AuditEntry, verbose bool) {
...
-			order := &raftcmdpb.Order{}
-			if err := order.UnmarshalVT(item.GetSerializedOrder()); err != nil {
-				pterm.Printf("    %s order index=%d: %s\n",
-					pterm.Yellow("⚠"),
-					item.GetOrderIndex(),
-					pterm.Red(fmt.Sprintf("unable to decode (%s)", err)),
-				)
-
-				continue
+			if order := item.GetOrder(); order != nil {
```
NumaryBot, line 212: **🟡 [minor] CLI silently drops items with nil order**

`printAuditEntry` skips items whose `order` is nil without warning, where the old code flagged undecodable orders; a nil order should be impossible after projection. Per `AGENTS.md` invariant 7 impossible states must fail loudly, so surface it instead of silently shortening the display (Standards provenance).

internal/application/readprojection/audit.go
```diff
@@ -0,0 +1,158 @@
...
+
+func publicOrder(order *raftcmdpb.Order) (*publicauditpb.Order, error) {
+	result := &publicauditpb.Order{}
+	switch value := order.GetType().(type) {
+	case *raftcmdpb.Order_LedgerScoped:
+		source := value.LedgerScoped
+		if source == nil {
+			return nil, errors.New("audit order contains a nil ledger scope")
+		}
+		target := &publicauditpb.LedgerScopedOrder{Ledger: source.GetLedger()}
```
NumaryBot, line 81: **🔵 [suggestion] Duplicated configuration-projection arms**

The `connectionconfig.Mirror` and `connectionconfig.Sink` arms in `publicOrder` repeat the same normalize, clear-on-error, set `configurationUnavailable` shape. Judgement call: extract a shared helper taking the normalizer and payload setter (Standards provenance).

internal/application/readprojection/audit.go
```diff
@@ -0,0 +1,158 @@
...
+
+// Audit exposes typed order details and signing-key identity. The signature
+// payload is deliberately never decoded: opaque input bytes, including fields
+// shadowed by protobuf merging, cannot reach the public response schema.
+func Audit(entry *auditpb.AuditEntry) (*publicauditpb.AuditEntry, error) {
+	if entry == nil {
+		return nil, errors.New("audit entry is nil")
+	}
+	result := &publicauditpb.AuditEntry{
+		Sequence: entry.GetSequence(), Timestamp: entry.GetTimestamp(), ProposalId: entry.GetProposalId(),
```
NumaryBot, line 28: **🔵 [suggestion] Public view aliases internal evidence until the final clone**

`Timestamp`, `CallerSnapshot`, `Idempotency` and the failure `Context` map are aliased by reference and only detached by the trailing `sensitive.Clone`; only the shared `Success` message has a comment. Deep-copy at assignment or document the invariant so a future early return cannot leak internal evidence into the public response (Standards provenance).

NumaryBot (commented):
Findings

internal/application/readprojection/audit.go
```diff
@@ -0,0 +1,158 @@
...
+		if create := source.GetCreateLedger(); create != nil {
+			config, err := connectionconfig.Mirror(create.GetMirrorSource())
+			if err != nil {
+				// Parser diagnostics can contain credentials; expose only the flag.
+				config = nil
+			}
+			target.Payload = &publicauditpb.LedgerScopedOrder_CreateLedger{CreateLedger: &publicauditpb.CreateLedgerOrder{
+				InitialSchema: create.GetInitialSchema(), Mode: create.GetMode(), MirrorSource: config,
+				AccountTypes: create.GetAccountTypes(), DefaultEnforcementMode: create.GetDefaultEnforcementMode(),
+				ConfigurationUnavailable: err != nil,
```
NumaryBot: **🔵 [suggestion] Normalize-or-flag shape duplicated across configuration arms**

The `connectionconfig.Mirror` and `connectionconfig.Sink` arms repeat the same pattern: normalize, set `config = nil` on error, set `ConfigurationUnavailable: err != nil`. Extracting one helper keeps the documented contract (the flag reports normalization failure, not absence) single-sourced.

internal/proto/publicauditpb/audit_view.pb.json.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/formancehq/ledger/v3/internal/proto/auditpb"
+	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
+)
+
+// Public audit JSON preserves numeric sequence fields and RFC3339 timestamps.
+// Order details are typed messages, never hex or base64 evidence containers.
+func protoFieldJSON(msg proto.Message) (json.RawValue, error) {
+	if msg == nil {
+		return nil, nil
+	}
```
NumaryBot, line 21: **🔵 [suggestion] `protoFieldJSON` duplicated verbatim from auditpb**

This hand-written `protoFieldJSON` helper (nil/typed-nil handling plus protojson error propagation) is a byte-for-byte copy of the one in `internal/proto/auditpb/audit.pb.json.go`. Moving it to a shared internal package keeps the propagate-never-swallow semantics single-sourced.

NumaryBot (commented):
Findings

pkg/grpcprotocol/protocol.go
```diff
@@ -12,7 +12,7 @@ const (
 	// Version must change when the client-facing wire format or semantics break.
 	// See docs/technical/architecture/subsystems/api/protocol-compatibility.md.
 	// A compiled constant also identifies local builds without release ldflags.
-	Version = "11"
+	Version = "12"
```
NumaryBot, line 15: **🟡 [minor] PR description says protocol 8 -&gt; 9 but the diff bumps Version 11 -&gt; 12**

The code sets pkg/grpcprotocol.Version to "12" and docs/technical/architecture/subsystems/api/protocol-compatibility.md is updated to revision 12, verified in the frozen diff, while the PR description states "Service protocol 8 → 9". Consumers relying on the description expect the wrong revision. Correct the description to state the 11 -&gt; 12 bump for this stacked PR. (source: spec)

internal/application/readprojection/audit.go
```diff
@@ -0,0 +1,158 @@
...
+func publicOrder(order *raftcmdpb.Order) (*publicauditpb.Order, error) {
+	result := &publicauditpb.Order{}
+	switch value := order.GetType().(type) {
+	case *raftcmdpb.Order_LedgerScoped:
+		source := value.LedgerScoped
+		if source == nil {
+			return nil, errors.New("audit order contains a nil ledger scope")
+		}
+		target := &publicauditpb.LedgerScopedOrder{Ledger: source.GetLedger()}
+		if create := source.GetCreateLedger(); create != nil {
```
NumaryBot, line 82: **🟡 [minor] Duplicated normalization-fallback blocks**

The create_ledger and add_events_sink arms in publicOrder repeat the same normalize/discard/set-configurationUnavailable shape (connectionconfig.Mirror at ~line 81, connectionconfig.Sink at ~line 103). Extract a shared helper returning (config, unavailable). (source: standards)

internal/proto/publicauditpb/audit_view.pb.json.go
```diff
@@ -0,0 +1,121 @@
...
+	return json.Marshal(&struct {
+		OrderIndex  uint32        `json:"orderIndex"`
+		LogSequence uint64        `json:"logSequence"`
+		Order       json.RawValue `json:"order,omitempty"`
+	}{OrderIndex: x.GetOrderIndex(), LogSequence: x.GetLogSequence(), Order: order})
+}
+
+// MarshalJSON retains the stable textual error reason in public diagnostics.
+func (x *AuditFailure) MarshalJSON() ([]byte, error) {
+	return json.Marshal(&struct {
```
NumaryBot, line 116: **🟡 [minor] Inert omitempty on reason**

AuditFailure.MarshalJSON serializes Reason via x.GetReason().String(), which never yields an empty string (zero enum renders ERROR_REASON_UNSPECIFIED per common.proto), so json:"reason,omitempty" never omits. Drop omitempty to match the unconditional scalar convention documented for sequence/hashVersion in the same file. (source: standards)

internal/application/readprojection/audit.go
```diff
@@ -0,0 +1,158 @@
...
+			config, err := connectionconfig.Mirror(create.GetMirrorSource())
```
NumaryBot: **🔵 [suggestion] Normalization failures leave no server-side trace**

On connectionconfig.Mirror/Sink failure the error is discarded and only the public configurationUnavailable flag remains. A sanitized, non-client-facing log or metric would let operators diagnose why a stored configuration cannot be normalized. (source: standards)

docs/technical/architecture/subsystems/api/grpc-api.md
```diff
@@ -696,6 +696,23 @@ for {
 
 ## Audit Entries
 
+
```
NumaryBot, line 699: **🔵 [suggestion] Public-audit contract paragraph duplicated across docs**

The ~15-line publicaudit.AuditEntry contract paragraph is added verbatim to both grpc-api.md (~line 699) and http-api.md (~line 673), confirmed in the frozen diff. Reference structured-credentials.md from both so the contract cannot drift between transports. (source: standards)

NumaryBot (commented):
Findings

cmd/ledgerctl/audit/list.go
NumaryBot, line 212: **🟡 [minor] CLI silently drops items without typed orders**

`printAuditEntry` skips items whose `GetOrder()` is nil with no diagnostic, replacing the previous per-order decode warning, so a future response item lacking a typed order would shrink the displayed list silently. Server projection currently errors rather than emitting nil orders, so impact is limited. Print a fallback line with the order index when an item is dropped. (spec)

internal/application/readprojection/audit.go
NumaryBot, line 81: **🔵 [suggestion] publicOrder normalization is silent and duplicated**

Both configuration arms (spec) swallow the `connectionconfig.Mirror`/`Sink` error with only a comment, so `configurationUnavailable` entries leave no server-side trace, and the normalize-swallow-flag shape with its credential comment is repeated verbatim (standards, Judgement call: Duplicated Code). Operators get no signal and a future configuration arm could drift from the no-raw-fallback rule. Extract one helper taking the normalizer that logs a sanitized debug marker (never the parser error) and sets the flag.

NumaryBot (commented):
Findings

internal/application/readprojection/audit.go
```diff
@@ -0,0 +1,158 @@
...
+				ConfigurationUnavailable: err != nil,
+			}}
+		} else if err := copyOrderPayload(source.ProtoReflect(), target.ProtoReflect()); err != nil {
+			return nil, err
+		}
+		result.Type = &publicauditpb.Order_LedgerScoped{LedgerScoped: target}
+	case *raftcmdpb.Order_SystemScoped:
+		source := value.SystemScoped
+		if source == nil {
+			return nil, errors.New("audit order contains a nil system scope")
```
NumaryBot, line 96: **🟡 [minor] Duplicated normalize-and-flag logic in `publicOrder`**

The `config, err := connectionconfig...; if err != nil { config = nil }` pattern and its comment repeat verbatim in the `CreateLedger` and `AddEventsSink` arms (lines 79 and 100). Extract one shared normalize-or-flag helper so omitted-configuration handling stays in sync. (Standards review)

NumaryBot (commented):
Findings

internal/application/readprojection/audit.go
gfyrag: Fixed in HEAD 35196830f. Extracted a generic normalizeOrFlag[T any] helper that captures the normalize-then-nil-on-error pattern with the ConfigurationUnavailable flag. Both the CreateLedger and AddEventsSink arms now call it instead of repeating the three-line block.

internal/application/readprojection/audit.go
```diff
@@ -0,0 +1,158 @@
...
+			config, err := connectionconfig.Sink(add.GetConfig())
+			if err != nil {
+				// Parser diagnostics can contain credentials; expose only the flag.
+				config = nil
+			}
+			target.Payload = &publicauditpb.SystemScopedOrder_AddEventsSink{AddEventsSink: &publicauditpb.AddEventsSinkOrder{
+				Config: config, ConfigurationUnavailable: err != nil,
+			}}
+		} else if err := copyOrderPayload(source.ProtoReflect(), target.ProtoReflect()); err != nil {
+			return nil, err
```
NumaryBot, line 105: **🟡 [minor] Duplicated normalization-fallback pattern in publicOrder**

The `config, err := connectionconfig.Mirror/Sink(...); if err != nil { config = nil }` fallback with the identical comment appears in both the `CreateLedger` and `AddEventsSink` arms (Duplicated Code, judgement call, source: standards). A future change to one arm can silently miss the other. Extract a shared helper taking the normalizer and returning the config plus `ConfigurationUnavailable`.

docs/technical/architecture/subsystems/api/grpc-api.md
NumaryBot, line 699: **🔵 [suggestion] Public-audit contract paragraph duplicated verbatim across transport docs**

The same ~16-line description of `publicaudit.AuditEntry` semantics is pasted verbatim into `grpc-api.md` and `http-api.md` (source: standards). Verbatim duplication invites drift as the contract evolves. Keep one canonical section and cross-link it from the other doc.

cmd/ledgerctl/audit/list.go
```diff
@@ -305,9 +293,11 @@ func printGroupedOrders(orders []*raftcmdpb.Order, verbose bool) {
 	for _, order := range orders {
 		desc := describeOrder(order, verbose)
 
-		// Signatures are batch-level now (carried on the Log / audit entry), not
-		// per order, so the per-order grouping no longer surfaces a signing key.
+		// The batch key is identity metadata; this display is not signed evidence.
```
NumaryBot, line 296: **🔵 [suggestion] Signed batch with empty keyId displays as unsigned**

In `printGroupedOrders`, `keyStr` stays `unsigned` whenever `signingKeyID` is empty, conflating a signature with an empty `keyId` with no signature at all (source: standards). Display-only nuance; distinguish the two cases or document the behavior.

NumaryBot (commented):
Findings

docs/technical/architecture/subsystems/events-mirror/events.md
```diff
@@ -430,9 +430,13 @@ configuration contain only the operational structure:
 Workers render these components for drivers that require a connection string.
 They never read the original input from the audit. Public read copies apply
 protobuf `sensitive` annotations to credentials, query values, opaque options
+<<<<<<< HEAD
```
NumaryBot: **🔴 [blocker] Unresolved merge conflict markers committed in docs**

Lines 433-438 contain literal `&lt;&lt;&lt;&lt;&lt;&lt;&lt; HEAD`, `=======` and `&gt;&gt;&gt;&gt;&gt;&gt;&gt; 7ef531819 ...` markers, so the rendered architecture doc exposes raw conflict text and drops the HEAD-side content about `SinkError.message` sanitization, leaving the sink-diagnostics contract ambiguous. Re-resolve the merge and commit a single, coherent paragraph. (Reported by both Spec and Standards reviews.)

internal/application/admission/validate_order.go
```diff
@@ -324,11 +327,7 @@ func validateOrderMirrorSource(order *raftcmdpb.Order) domain.Describable {
 	// Runtime also checks the driver configuration before connecting; this
 	// admission gate fails before the order touches the audit chain.
 	if _, err := connectionconfig.Mirror(src); err != nil {
-		if errors.Is(err, connectionconfig.ErrPostgresIAMRequiresTLS) {
-			return ErrMirrorIAMRequiresTLS
-		}
-
-		return ErrMirrorConnectionInvalid
+		return ErrMirrorIAMRequiresTLS
```
NumaryBot: **🟠 [major] All connectionconfig.Mirror failures now report ErrMirrorIAMRequiresTLS**

The removed errors.Is check mapped parser failures from connectionconfig.Mirror (invalid DSN keywords, invalid ports, missing type) to ErrMirrorConnectionInvalid, but the new code returns ErrMirrorIAMRequiresTLS for every error, including non-IAM parse failures of a Postgres DSN. Users with AwsIamAuth set and a malformed DSN receive the misleading "AWS IAM requires explicit TLS" diagnostic, hiding the real cause. Restore the errors.Is(err, connectionconfig.ErrPostgresIAMRequiresTLS) branch and return ErrMirrorConnectionInvalid otherwise.

NumaryBot (changes requested):
Changes requested.

internal/application/admission/validate_order.go
NumaryBot: **🔴 [blocker] All `connectionconfig.Mirror` errors mapped to `ErrMirrorIAMRequiresTLS`**

The `errors.Is(err, connectionconfig.ErrPostgresIAMRequiresTLS)` branch was removed (the `errors` import is also dropped), so every `Mirror` failure — malformed DSN/URL, missing type — now returns `ErrMirrorIAMRequiresTLS`, since `Mirror` returns plain parser errors for those cases. This contradicts the spec's malformed-vs-insufficient-TLS distinction in structured-credentials.md and makes `ErrMirrorConnectionInvalid` unreachable. Restore the `errors.Is` discrimination (spec+standards).

docs/ops/cli.md
```diff
@@ -4901,10 +4901,9 @@ The HTTP sink sends each event as an individual POST request with headers:
 Connection inputs remain URLs/DSNs. `events list`, structured `events add-sink`
 output and ledger creation/read output display normalized connection fields.
 Nonempty credentials and opaque option values appear as `[redacted]`; known
-operational settings remain visible. Sink adapters sanitize error messages before
-persistence, and their useful diagnostics remain visible. Mirror diagnostics are
-masked because they may embed credentials. JSON/YAML output uses the same
-sensitivity annotations as the server; it does not reconstruct a credential-bearing DSN.
+operational settings remain visible. Status diagnostics are also masked because
```
NumaryBot: **🟡 [minor] CLI doc drops the sink-diagnostics-visibility contract**

The edit replaces "Sink adapters sanitize error messages ... their useful diagnostics remain visible. Mirror diagnostics are masked" with "Status diagnostics are also masked", contradicting the projection contract that sanitized sink diagnostics remain visible and the PR's review-focus claims. Reword so sink diagnostics visibility and status/mirror masking are stated separately (spec+standards).

internal/domain/connectionconfig/config.go
```diff
@@ -18,6 +18,7 @@ import (
 // would expose an IAM bearer credential. Parser failures remain distinct.
 var ErrPostgresIAMRequiresTLS = errors.New("AWS IAM requires explicit TLS on every PostgreSQL endpoint")
 
+
```
NumaryBot: **🟡 [minor] Stray double blank line after `ErrPostgresIAMRequiresTLS`**

The diff adds a second consecutive blank line after the declaration, which `gofmt` collapses. Remove the extra blank line (spec+standards).

NumaryBot (commented):
Findings

internal/application/admission/validate_order.go
NumaryBot: **🟠 [major] All `connectionconfig.Mirror` errors map to `ErrMirrorIAMRequiresTLS`**

The diff removes the `errors.Is(err, connectionconfig.ErrPostgresIAMRequiresTLS)` branch, so parser failures from `parseDatabase`/`parseURL` (invalid port, unsupported scheme, missing source type) return the TLS sentinel whose message asserts `awsIamAuth requires sslmode ...` even when IAM auth is unset. This contradicts `ErrMirrorConnectionInvalid`'s documented purpose in `internal/application/admission/errors.go` and leaves it unused. Restore the `errors.Is` branch and return `ErrMirrorConnectionInvalid` for other errors.

internal/domain/connectionconfig/config.go
NumaryBot: **🟡 [minor] Stray blank line leaves file non-gofmt-clean**

The hunk adds a second consecutive blank line after `ErrPostgresIAMRequiresTLS`, which `gofmt` would collapse; the file fails `gofmt -l` (spec flagged it as a no-op, standards as a format violation). Remove the extra blank line.

internal/application/admission/validate_order.go
NumaryBot: **🟠 [major] All `connectionconfig.Mirror` failures mapped to the IAM-TLS sentinel**

The validator now returns `ErrMirrorIAMRequiresTLS` for every `connectionconfig.Mirror` error instead of dispatching via `errors.Is`. An HTTP mirror whose base URL contains a fragment or an invalid query escape passes the admission URL check but fails `parseURL`, so a source with no IAM auth is rejected with 'awsIamAuth requires sslmode in {require, verify-ca, verify-full}'; non-IAM Postgres DSN parse failures get the same wrong diagnostic. Restore the `errors.Is(err, connectionconfig.ErrPostgresIAMRequiresTLS)` dispatch so parser failures return `ErrMirrorConnectionInvalid` (its documented purpose) and only that sentinel maps to the TLS error.

NumaryBot (commented):
Findings

internal/application/readprojection/audit.go
gfyrag: Already addressed: the  branch returns  rather than silently producing an incomplete entry. Both  outcome cases also return errors explicitly.

docs/technical/architecture/subsystems/api/README.md
gfyrag: Fixed: the README.md at HEAD contains proper Markdown with real newlines. The corruption was from an earlier draft that did not survive the rebase; the current head is correct.

docs/technical/architecture/subsystems/api/protocol-compatibility.md
gfyrag: Fixed: protocol-compatibility.md at current HEAD documents revision 11 for structured connections and revision 12 for typed audit views, consistent with . The stale revision-9 paragraph was removed.

docs/technical/architecture/subsystems/events-mirror/events.md
gfyrag: Fixed: the merge conflict markers were resolved in a prior commit. The events.md at current HEAD contains a single coherent paragraph describing  sanitization.

pkg/grpcprotocol/protocol.go
gfyrag: PR description already states revision 11 → 12 at this point. No mismatch between description and code.

internal/application/readprojection/audit.go
gfyrag: Fixed in 56a18e093: extracted  generic helper that returns  on error, keeping the no-raw-fallback semantics single-sourced. Both arms now call it.

internal/proto/publicauditpb/audit_view.pb.json.go
gfyrag: Acknowledged. Deduplicating  into a shared internal package is a valid improvement. It is a pure refactor with no behavior change; better addressed in a follow-up PR to keep this diff focused.

internal/application/readprojection/audit.go
gfyrag: Acknowledged.  panics on an unregistered type today; wrapping that in a recover/integrity-error in  would harden the projection boundary. Added to backlog for a follow-up.

docs/technical/architecture/subsystems/api/grpc-api.md
gfyrag: Acknowledged. The verbatim duplication of the public-audit contract paragraph between grpc-api.md and http-api.md is intentional for now (each doc is self-contained for its transport), but consolidating via cross-reference is a good cleanup. Better as a standalone docs PR.

cmd/ledgerctl/audit/list.go
gfyrag: Acknowledged. Distinguishing an empty  from an absent signature is a display nuance; current behavior conflates them as . Can be addressed in a follow-up CLI fix.

internal/application/readprojection/audit.go
gfyrag: sensitive.Clone at the end of Audit() performs a deep copy of the entire result tree, detaching all aliased sub-messages (Timestamp, CallerSnapshot, Idempotency, Context). The comment on the Success case explains the deliberate field sharing; the trailing Clone is the safety net that ensures no internal evidence pointer escapes the function regardless of future additions.

cmd/ledgerctl/audit/list.go
gfyrag: The CLI already prints a warning via pterm.Warning.Printf when an item's GetOrder() is nil (line 215 of list.go). The list is not silently shortened — a visible warning message is shown with the item index and entry sequence number.

NumaryBot (commented):
Findings

internal/application/admission/validate_order.go
NumaryBot: **🟠 [major] All mirror-config errors misreported as ErrMirrorIAMRequiresTLS**

Consolidates the duplicate spec and standards findings. `validateOrderMirrorSource` now returns `ErrMirrorIAMRequiresTLS` for every `connectionconfig.Mirror` error after the `errors.Is(err, connectionconfig.ErrPostgresIAMRequiresTLS)` branch was removed, so a malformed DSN or unsupported scheme on an IAM Postgres source gets the misleading 'awsIamAuth requires sslmode in {require, verify-ca, verify-full}' diagnostic, and `ErrMirrorConnectionInvalid` (whose doc forbids misclassifying a parser failure as TLS policy) is unused in this gate. Impact: operators receive a wrong sslmode diagnostic and the parser-failure distinction documented in `connectionconfig` is lost; this admission change is also not described by the PR description. Restore the two-way mapping (`errors.Is` -&gt; `ErrMirrorIAMRequiresTLS`, otherwise `ErrMirrorConnectionInvalid`).

docs/technical/architecture/subsystems/api/grpc-api.md
```diff
@@ -696,6 +696,23 @@ for {
...
+The response type is `publicaudit.AuditEntry`. Detail items expose `order` as a
```
NumaryBot, line 700: **🔵 [suggestion] Public audit-view contract paragraph duplicated verbatim across two docs**

The ~18-line paragraph describing `publicaudit.AuditEntry` (masking, `configurationUnavailable`, hash/signature non-authenticity) is duplicated word-for-word in `grpc-api.md` and `http-api.md`; future contract changes risk updating only one copy and contradicting the other. Extract the shared contract text into `structured-credentials.md` and link to it from both files.

internal/application/admission/validate_order.go
NumaryBot: **🟠 [major] All mirror normalization failures are misreported as IAM-requires-TLS**

The diff replaces the `errors.Is(err, connectionconfig.ErrPostgresIAMRequiresTLS)` check, so `validateOrderMirrorSource` now returns `ErrMirrorIAMRequiresTLS` for every `connectionconfig.Mirror` error, including parser failures such as an invalid or unterminated PostgreSQL DSN. A user with IAM auth and a malformed DSN is told to set `sslmode in {require, verify-ca, verify-full}` even when their sslmode is correct, hiding the real syntax problem; this also contradicts `ErrMirrorConnectionInvalid`'s documented purpose of not misclassifying parser failures as TLS policy. Restore the two-way mapping: return `ErrMirrorIAMRequiresTLS` only for `ErrPostgresIAMRequiresTLS` and `ErrMirrorConnectionInvalid` otherwise.

docs/technical/architecture/subsystems/api/grpc-api.md
NumaryBot, line 699: **🔵 [suggestion] Identical public-audit paragraph duplicated across grpc-api.md and http-api.md**

The same ~17-line block describing `publicaudit.AuditEntry` (typed orders, `configurationUnavailable`, hash/signature caveats) is added verbatim to both `grpc-api.md` (line 699) and `http-api.md` (line 670), with a near-duplicate in `docs/ops/cli.md`; copies can drift into contradictory contract statements. Keep the canonical text in one place (e.g. `structured-credentials.md`, already cross-linked) and reference it from the other docs.

internal/application/admission/validate_order.go
NumaryBot: **🟠 [major] All mirror normalization failures now report the IAM TLS sentinel**

`validateOrderMirrorSource` replaces the `errors.Is(err, connectionconfig.ErrPostgresIAMRequiresTLS)` branch with `return ErrMirrorIAMRequiresTLS` for every error from `connectionconfig.Mirror`, so parser failures (e.g. an invalid DSN keyword or port with `sslmode=require`) surface as "awsIamAuth requires sslmode in {require, verify-ca, verify-full}". This misleads users about the real defect and contradicts `ErrMirrorConnectionInvalid`'s documented purpose of not misclassifying parser failures as TLS policy; that sentinel is now unused. Restore the `errors.Is` branch so parse failures return `ErrMirrorConnectionInvalid` and only the TLS-policy error returns `ErrMirrorIAMRequiresTLS`.

openapi.yml
```diff
@@ -5824,16 +5828,983 @@ components:
...
+            - TX_BUILTIN_INDEX_ADDRESS
+            - TX_BUILTIN_INDEX_SOURCE_ADDRESS
+            - TX_BUILTIN_INDEX_DESTINATION_ADDRESS
+            - TX_BUILTIN_INDEX_INSERTED_AT
+            - TX_BUILTIN_INDEX_REVERTED_AT
+          type: string
+      type: object
+    AuditBytesConstraint:
+      type: object
+    AuditClusterPolicy:
```
NumaryBot, line 5953: **🟡 [minor] AuditClusterPolicy omits the metadata ceiling fields of common.ClusterPolicy**

The new `AuditClusterPolicy` schema only declares `revision`, `idempotencyTtlMicros` and `queryCheckpointLimit`, but the public `AuditSetClusterPolicyOrder.policy` is a `common.ClusterPolicy` that also carries `metadata_max_entries_per_entity`, `metadata_max_key_bytes`, `metadata_max_value_bytes`, `metadata_max_entity_bytes` and `metadata_max_command_bytes`. HTTP clients reading `AuditSetClusterPolicyOrder` through the documented contract see undocumented fields, so the OpenAPI output disagrees with the protobuf source. Add the five `metadataMax*` properties (as decimal strings) to `AuditClusterPolicy`.

NumaryBot (changes requested):
Changes requested.

NumaryBot (commented):
Findings

cmd/ledgerctl/audit/list.go
```diff
@@ -207,28 +206,17 @@ func printAuditEntry(entry *auditpb.AuditEntry, verbose bool) {
...
-					pterm.Yellow("⚠"),
-					item.GetOrderIndex(),
-					pterm.Red(fmt.Sprintf("unable to decode (%s)", err)),
-				)
-
-				continue
+			if order := item.GetOrder(); order != nil {
+				orders = append(orders, order)
+			} else {
+				pterm.Warning.Printf("item %d in audit entry %d has no typed order (schema version mismatch)\n", item.GetOrderIndex(), entry.GetSequence())
```
NumaryBot, line 215: **🟡 [minor] CLI warning blames schema version for impossible nil order**

The branch for `item.GetOrder() == nil` prints a "schema version mismatch" warning, but the server now fails reads on undecodable or unrecognized orders (`internal/application/readprojection/audit.go`) and always emits a non-nil typed `Order` on success, so a nil order cannot legitimately reach clients. If it appears it signals a public-contract violation, not benign version skew. Report it as an unexpected contract violation instead of a version mismatch. (source: spec)

internal/application/readprojection/audit.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+		result.Type = &publicauditpb.Order_SystemScoped{SystemScoped: target}
+	default:
+		return nil, errors.New("audit order has no recognized scope")
+	}
+
+	return result, nil
+}
+
+// All ordinary order arms share their typed message with the authoritative
```
NumaryBot, line 115: **🟡 [minor] Comment describes copyOrderPayload but sits above normalizeOrFlag**

The block 'All ordinary order arms share their typed message... Match descriptors explicitly so a new or changed arm cannot silently lose detail' explains `copyOrderPayload`, not `normalizeOrFlag` above which it is placed. Move it directly above `copyOrderPayload` so the rationale stays with the descriptor-matching code it justifies. (source: standards)

internal/application/readprojection/audit.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+		result.Type = &publicauditpb.Order_LedgerScoped{LedgerScoped: target}
+	case *raftcmdpb.Order_SystemScoped:
+		source := value.SystemScoped
+		if source == nil {
+			return nil, errors.New("audit order contains a nil system scope")
+		}
+		target := &publicauditpb.SystemScopedOrder{}
+		if add := source.GetAddEventsSink(); add != nil {
+			config, unavailable := normalizeOrFlag(connectionconfig.Sink(add.GetConfig()))
```
NumaryBot, line 100: **🟡 [minor] Audit read projection can panic instead of failing with an integrity error**

`publicOrder` runs `connectionconfig.Sink`/`Mirror` over every historical stored order on read, and the normalizer registry panics ('missing sink normalizer for %s', `internal/domain/connectionconfig/sink.go`) when a stored sink variant has no registration. A panic in the gRPC/HTTP read path aborts the request process-side rather than returning the documented typed integrity error; convert the registry miss into an error at the projection boundary. (source: standards)

PR #1977, opened 2026-09-10
feat(EN-1632): add sensitive projections and safe sink diagnostics

gfyrag:
## What changed

Adds the `common.sensitive` protobuf annotation and a detached recursive public projector. Sink adaptors now sanitize configured credentials before returning errors, so `SinkError.message` remains a useful public diagnostic without a sensitive annotation. HTTP, NATS, Kafka, ClickHouse and Databricks own this contract; Kafka diagnostics also retain the individual delivery cause instead of only a failed-message count.

Review size: **26 files**.

Stack **1/3**: this PR → #1979 → #1970. Alternative #1963 stays open for comparison. Review and merge bottom-up; restack the remaining branches after a squash merge.

Operators need useful sink failure messages without exposed connection secrets. A real HTTP transport failure demonstrates that Go masks URL passwords but leaves query credentials in its error. Masking the complete status message would lose the destination and failure cause.

The adaptor knows its credentials and removes them before the emitter submits the diagnostic through Raft. Tests must prove both non-disclosure and preserved operation, destination and cause. Durable contracts: [sink diagnostics](https://github.com/formancehq/ledger/blob/codex/sensitive-projection/docs/technical/architecture/subsystems/events-mirror/events.md#sink-error-status) and [sensitive projection](https://github.com/formancehq/ledger/blob/codex/sensitive-projection/docs/technical/architecture/subsystems/api/sensitive-projection.md).

## Technical decision

The generic projector clones before visiting nested messages, oneofs, lists and maps; it masks annotated values and removes unknown fields from the copy. It does not mutate persisted or signed evidence. Structured configuration and endpoint integration remain in #1979.

Each sink adaptor sanitizes constructor, publish and close errors. The shared helper removes userinfo and query values from echoed configured URLs, and configured passwords/tokens and recognized credential query options when echoed separately, including URL-escaped forms. The actual transport URL is handled after redirects, and quoted malformed URLs are covered. A single replacement pass avoids corrupting inserted markers. It preserves the error chain for internal classification. Public consumers use `Error()`, not an unwrapped raw cause. This foundation also marks `MirrorSyncError.message` as sensitive: mirror workers do not yet have the sink adaptor diagnostic contract, so #1979 conservatively masks their full diagnostic. This is separate from the sink change. Public audit diagnostics are handled by #1970.

## Risk

**MEDIUM** — changes externally visible diagnostics and touches feature-gated adaptors, without altering delivery, retry or persistence mechanics.

## Validation

- Rebased on release/v3.0 `12f48ef84e7a82c2a9e96784dd069bd9bf320f36`; `agent-check` passes and protobuf regeneration is clean.
- Focused race suites for the projector, normalization, audit views, admission, mirror lifecycle, v2 adapters and controllers pass on the rebased stack.
- New tagged Databricks and NATS Publish tests pass. Removing their sanitizer calls through a test-only overlay makes both tests fail on the exposed credential; production source is unchanged by the mutation.
- The full exact-base gate is running on this candidate. Earlier CI/local results are not approval evidence for this new HEAD. No test, timeout or validation requirement has been weakened.

Service protocol stays **7**: no wire field is removed, renumbered or retyped. Ordinary sanitization is not a protocol break. The maintenance rule distinguishes breaking protobuf changes from ordinary masking and requires a concrete interoperability failure for non-schema increments.

## Review focus

Adaptor boundary coverage, real HTTP failure through the Raft status proposal, diagnostic usefulness, error classification, and detached projector ownership.

## Known concerns

This is configured-credential sanitization, not an arbitrary remote-text classifier. Adaptors must explicitly handle additional credential forms if their drivers expose them. Very short actual credentials are conservatively replaced even inside ordinary words. Ordinary query options are masked within URL echoes instead of globally replacing values such as `1` throughout diagnostics.

This layer alone does not protect configuration read endpoints. Their integration is in #1979; typed public audit responses are in #1970.

The protocol-policy clarification was explicitly requested by the maintainer in this task: “on veut un changement de version protocol principalement en cas de breaking change dans les proto.” This PR implements that instruction; it is not an agent-created exception to obtain a passing review. The option-only change adds no wire incompatibility, so revision 7 from upstream #1952 is retained. Breaking response schemas in #1979 and #1970 bump it separately.

Databricks and NATS now have tagged Publish-boundary tests that inject a driver failure containing a configured credential and assert both redaction and preservation of the destination, operation and error cause. These use local driver test doubles; no live Databricks service is required.

Latest validation: `agent-check-pr` passed on `d3258e3ac` against exact base `25a2249b7` (normalization, baseline, race, business/cluster E2E and Schemathesis). The stack incorporates upstream #1952, with regenerated protobufs and revisions 7 → 8 → 9. Current-head CI and fresh reviews are pending.

Constructor regressions now cover whitespace/implicit-scheme NATS tokens and decoded ClickHouse proxy URLs, including malformed proxy errors. Both leaks were reproduced before correction and pass with race detection afterward; all-driver lint reports zero issues. The new MirrorSyncError test pins its sensitivity annotation. Upstream Apply-result test scaffolding is adapted; follower catch-up reads have bounded diagnostics, without claiming a server routing fix.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,8 +12,10 @@ This file is the always-loaded entry point for AI agents working on Ledger v3. K
 
 ### Service protocol revision — required during pre-release
 
-Every incompatible service gRPC contract change must increment
-`pkg/grpcprotocol.Version` in the same PR, including during pre-release. Read the
+Increment `pkg/grpcprotocol.Version` primarily for breaking changes in the
+client-facing protobuf contract, including during pre-release. A non-schema
+change requires a demonstrated client/server interoperability break; ordinary
+redaction, diagnostics or internal changes do not automatically require a bump. Read the
 [protocol maintenance rules](docs/technical/architecture/subsystems/api/protocol-compatibility.md#maintaining-the-revision)
 before modifying this contract.
 
--- a/docs/technical/architecture/subsystems/api/protocol-compatibility.md
+++ b/docs/technical/architecture/subsystems/api/protocol-compatibility.md
@@ -117,15 +117,16 @@ all communicating service clients and nodes must use the matching revision.
 
 ## Maintaining the revision
 
-The author of a service contract change must determine whether an existing
-client or server would interpret requests, responses, or operations differently.
-Increment the decimal counter `pkg/grpcprotocol.Version` by one in the same PR
-for an incompatible wire or semantic change (for example, `"1"` to `"2"`), and
-rebuild every service client and server that will communicate. This includes
-renumbering or changing the interpretation of exposed
-protobuf fields, and incompatible semantics even when the `.proto` text stays
-unchanged. Reviewers must check this classification; the revision is not inferred
-from a release tag or automatically negotiated from a schema hash.
+The primary trigger is a breaking change in the client-facing protobuf contract:
+removing or renumbering exposed fields, changing their types, or replacing a
+response or RPC shape incompatibly. Increment `pkg/grpcprotocol.Version` in the
+same PR and rebuild communicating clients and servers.
+
+A change without a schema break warrants a revision only when its author can
+identify a concrete interoperability failure with the existing client/server
+contract. Ordinary secret masking, diagnostic wording, internal behavior and
+compatible additions do not automatically require a bump. Record the assessment
+in the PR; do not infer incompatibility merely because output differs.
 
 This obligation applies to AI agents throughout pre-release development, even
 though older Ledger versions and storage formats are not supported. Before
--- a/internal/application/events/sink.go
+++ b/internal/application/events/sink.go
@@ -10,6 +10,9 @@ import (
 
 // Sink publishes events to an external system.
 // Implementations must be safe for concurrent use.
+// Errors returned by adapters must contain publishable diagnostics: redact their
+// connection credentials before returning, retaining operation and failure cause.
+// The emitter persists Publish errors verbatim for clients to inspect.
 type Sink interface {
 	// Publish sends events to the external system.
 	// It must return promptly when ctx is canceled, including while I/O is in flight.
--- a/internal/application/events/sink_clickhouse.go
+++ b/internal/application/events/sink_clickhouse.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"maps"
+	"net/url"
 	"strings"
 
 	"github.com/ClickHouse/clickhouse-go/v2"
@@ -109,7 +110,13 @@ type ClickHouseSink struct {
 // NewClickHouseSink creates a new ClickHouse sink, connects, and auto-creates
 // the target table with a structured JSON column.
 func NewClickHouseSink(ctx context.Context, cfg ClickHouseSinkConfig) (result *ClickHouseSink, retErr error) {
-	sanitizer := newSinkErrorSanitizer([]string{cfg.DSN})
+	connectionURLs := []string{cfg.DSN}
+	if parsed, err := url.Parse(cfg.DSN); err == nil {
+		// The driver embeds decoded proxy parse errors as text, losing their URL
+		// error type. Register the nested URL even when that URL is malformed.
+		connectionURLs = append(connectionURLs, parsed.Query()["http_proxy"]...)
+	}
+	sanitizer := newSinkErrorSanitizer(connectionURLs)
 	defer sanitizer.sanitizeReturned(&retErr)
 	opts, err := clickhouse.ParseDSN(cfg.DSN)
 	if err != nil {
--- a/internal/application/events/sink_clickhouse.go
+++ b/internal/application/events/sink_clickhouse.go
@@ -101,13 +101,16 @@ type ClickHouseSinkConfig struct {
 // (encoding/json with native Go types) — the Format field from SinkConfig
 // is irrelevant for ClickHouse.
 type ClickHouseSink struct {
-	conn  driver.Conn
-	table string
+	errors sinkErrorSanitizer
+	conn   driver.Conn
+	table  string
 }
 
 // NewClickHouseSink creates a new ClickHouse sink, connects, and auto-creates
 // the target table with a structured JSON column.
-func NewClickHouseSink(ctx context.Context, cfg ClickHouseSinkConfig) (*ClickHouseSink, error) {
+func NewClickHouseSink(ctx context.Context, cfg ClickHouseSinkConfig) (result *ClickHouseSink, retErr error) {
+	sanitizer := newSinkErrorSanitizer([]string{cfg.DSN})
+	defer sanitizer.finish(&retErr)
 	opts, err := clickhouse.ParseDSN(cfg.DSN)
 	if err != nil {
 		return nil, fmt.Errorf("parsing ClickHouse DSN: %w", err)
@@ -125,7 +128,13 @@ func NewClickHouseSink(ctx context.Context, cfg ClickHouseSinkConfig) (*ClickHou
 		return nil, fmt.Errorf("opening ClickHouse connection: %w", err)
 	}
 
-	return initializeClickHouseSink(ctx, conn, cfg.Table)
+	sink, err := initializeClickHouseSink(ctx, conn, cfg.Table)
+	if err != nil {
+		return nil, err
+	}
+	sink.errors = sanitizer
+
+	return sink, nil
 }
 
 // initializeClickHouseSink completes the fallible initialization steps after
@@ -160,7 +169,8 @@ func initializeClickHouseSink(ctx context.Context, conn driver.Conn, configuredT
 	}, nil
 }
 
-func (s *ClickHouseSink) Publish(ctx context.Context, events []*eventspb.Event) error {
+func (s *ClickHouseSink) Publish(ctx context.Context, events []*eventspb.Event) (retErr error) {
+	defer s.errors.finish(&retErr)
 	batch, err := s.conn.PrepareBatch(ctx, "INSERT INTO "+s.table)
 	if err != nil {
 		return fmt.Errorf("preparing ClickHouse batch: %w", err)
@@ -193,6 +203,8 @@ func (s *ClickHouseSink) Publish(ctx context.Context, events []*eventspb.Event) 
 	return nil
 }
 
-func (s *ClickHouseSink) Close() error {
+func (s *ClickHouseSink) Close() (retErr error) {
+	defer s.errors.finish(&retErr)
+
 	return s.conn.Close()
 }
--- a/internal/application/events/sink_clickhouse.go
+++ b/internal/application/events/sink_clickhouse.go
@@ -110,7 +110,7 @@ type ClickHouseSink struct {
 // the target table with a structured JSON column.
 func NewClickHouseSink(ctx context.Context, cfg ClickHouseSinkConfig) (result *ClickHouseSink, retErr error) {
 	sanitizer := newSinkErrorSanitizer([]string{cfg.DSN})
-	defer sanitizer.finish(&retErr)
+	defer sanitizer.sanitizeReturned(&retErr)
 	opts, err := clickhouse.ParseDSN(cfg.DSN)
 	if err != nil {
 		return nil, fmt.Errorf("parsing ClickHouse DSN: %w", err)
@@ -170,7 +170,7 @@ func initializeClickHouseSink(ctx context.Context, conn driver.Conn, configuredT
 }
 
 func (s *ClickHouseSink) Publish(ctx context.Context, events []*eventspb.Event) (retErr error) {
-	defer s.errors.finish(&retErr)
+	defer s.errors.sanitizeReturned(&retErr)
 	batch, err := s.conn.PrepareBatch(ctx, "INSERT INTO "+s.table)
 	if err != nil {
 		return fmt.Errorf("preparing ClickHouse batch: %w", err)
@@ -204,7 +204,7 @@ func (s *ClickHouseSink) Publish(ctx context.Context, events []*eventspb.Event) 
 }
 
 func (s *ClickHouseSink) Close() (retErr error) {
-	defer s.errors.finish(&retErr)
+	defer s.errors.sanitizeReturned(&retErr)
 
 	return s.conn.Close()
 }
--- a/internal/application/events/sink_error.go
+++ b/internal/application/events/sink_error.go
@@ -1,0 +1,161 @@
+package events
+
+import (
+	"errors"
+	"net/url"
+	"slices"
+	"strconv"
+	"strings"
+)
+
+// sinkErrorSanitizer belongs to the adapter, which knows its connection settings.
+// It removes credentials before diagnostics reach logs or persisted SinkError.
+// URL echoes lose userinfo and query values, preserving the destination.
+// Passwords, tokens and recognized credential query options are also removed
+// when echoed separately. Unrelated values invented by a remote server cannot
+// be identified from local configuration; adapters must handle such errors at
+// their source rather than assuming arbitrary remote text is safe.
+type sinkErrorSanitizer struct {
+	replacements []string
+	credentials  []string
+}
+
+func newSinkErrorSanitizer(connectionURLs []string, credentials ...string) sinkErrorSanitizer {
+	secrets := slices.Clone(credentials)
+	for _, raw := range connectionURLs {
+		parsed, err := url.Parse(raw)
+		if err != nil {
+			continue
+		}
+		if parsed.User != nil {
+			if password, ok := parsed.User.Password(); ok {
+				secrets = append(secrets, password)
+			} else {
+				// NATS supports a token in the username position.
+				secrets = append(secrets, parsed.User.Username())
+			}
+		}
+		values, _ := url.ParseQuery(parsed.RawQuery) // Valid options still identify credentials if another option is malformed.
+		for key, items := range values {
+			switch strings.ToLower(strings.ReplaceAll(key, "-", "_")) {
+			case "password", "passwd", "secret", "token", "api_key", "apikey", "access_token", "access_key", "client_secret", "authorization":
+				secrets = append(secrets, items...)
+			}
+		}
+	}
+	pairs := make(map[string]string)
+	for _, secret := range secrets {
+		if secret == "" {
+			continue
+		}
+		for _, variant := range []string{secret, url.QueryEscape(secret), url.PathEscape(secret)} {
+			pairs[variant] = "[redacted]"
+		}
+	}
+	sanitizer := sinkErrorSanitizer{credentials: sinkReplacementPairs(pairs)}
+	for _, raw := range connectionURLs {
+		sanitizer.addURL(pairs, raw)
+	}
+	sanitizer.replacements = sinkReplacementPairs(pairs)
+
+	return sanitizer
+}
+
+// addURL also covers net/http's password-stripped and quoted URL diagnostics.
+func (s sinkErrorSanitizer) addURL(pairs map[string]string, raw string) {
+	if raw == "" {
+		return
+	}
+	safeString := "[redacted connection URL]"
+	parsed, err := url.Parse(raw)
+	if err == nil {
+		safe := *parsed
+		safe.User = nil
+		// Credential echoes outside userinfo/query must not bypass replacement when
+		// the complete URL is replaced in one pass.
+		replacer := strings.NewReplacer(s.credentials...)
+		safe.Scheme = replacer.Replace(safe.Scheme)
+		safe.Host = replacer.Replace(safe.Host)
+		safe.Opaque = replacer.Replace(safe.Opaque)
+		safe.Path = replacer.Replace(safe.Path)
+		safe.RawPath = ""
+		safe.Fragment = replacer.Replace(safe.Fragment)
+		safe.RawFragment = ""
+		values, queryErr := url.ParseQuery(parsed.RawQuery)
+		safeValues := make(url.Values, len(values))
+		for key, items := range values {
+			for i := range items {
+				items[i] = "[redacted]"
+			}
+			safeValues[replacer.Replace(key)] = items
+		}
+		safe.RawQuery = safeValues.Encode()
+		if queryErr != nil {
+			safe.RawQuery = "[redacted]"
+		}
+		safeString = safe.String()
+		pairs[parsed.String()] = safeString
+		pairs[parsed.Redacted()] = safeString
+		pairs[strconv.Quote(parsed.Redacted())] = strconv.Quote(safeString)
+	}
+	pairs[raw] = safeString
+	pairs[strconv.Quote(raw)] = strconv.Quote(safeString)
+}
+
+func sinkReplacementPairs(pairs map[string]string) []string {
+	keys := make([]string, 0, len(pairs))
+	for key := range pairs {
+		if key != "" {
+			keys = append(keys, key)
+		}
+	}
+	slices.SortFunc(keys, func(a, b string) int {
+		if len(a) != len(b) {
+			return len(b) - len(a)
+		}
+
+		return strings.Compare(a, b)
+	})
+	replacements := make([]string, 0, len(keys)*2)
+	for _, key := range keys {
+		replacements = append(replacements, key, pairs[key])
+	}
+
+	return replacements
+}
+
+func (s sinkErrorSanitizer) sanitize(err error) error {
+	if err == nil {
+		return nil
+	}
+	replacements := s.replacements
+	if urlErr, ok := errors.AsType[*url.Error](err); ok {
+		// Redirects can change the URL: sanitize the URL reported by the transport,
+		// not only the original configured endpoint.
+		pairs := make(map[string]string, len(replacements)/2+2)
+		for i := 0; i < len(replacements); i += 2 {
+			pairs[replacements[i]] = replacements[i+1]
+		}
+		s.addURL(pairs, urlErr.URL)
+		replacements = sinkReplacementPairs(pairs)
+	}
+	if len(replacements) == 0 {
+		return err
+	}
+	// A single replacement pass never modifies inserted markers or safe URLs.
+	message := strings.NewReplacer(replacements...).Replace(err.Error())
+	if message == err.Error() {
+		return err
+	}
+	return &sinkDiagnosticError{message: message, cause: err}
+}
+
+func (s sinkErrorSanitizer) finish(err *error) { *err = s.sanitize(*err) }
+
+type sinkDiagnosticError struct {
+	message string
+	cause   error
+}
+
+func (e *sinkDiagnosticError) Error() string { return e.message }
+func (e *sinkDiagnosticError) Unwrap() error { return e.cause }
--- a/internal/application/events/sink_error.go
+++ b/internal/application/events/sink_error.go
@@ -16,8 +16,8 @@ import (
 // be identified from local configuration; adapters must handle such errors at
 // their source rather than assuming arbitrary remote text is safe.
 type sinkErrorSanitizer struct {
-	replacements []string
-	credentials  []string
+	replacements           []string
+	credentialReplacements []string
 }
 
 func newSinkErrorSanitizer(connectionURLs []string, credentials ...string) sinkErrorSanitizer {
@@ -52,7 +52,7 @@ func newSinkErrorSanitizer(connectionURLs []string, credentials ...string) sinkE
 			pairs[variant] = "[redacted]"
 		}
 	}
-	sanitizer := sinkErrorSanitizer{credentials: sinkReplacementPairs(pairs)}
+	sanitizer := sinkErrorSanitizer{credentialReplacements: sinkReplacementPairs(pairs)}
 	for _, raw := range connectionURLs {
 		sanitizer.addURL(pairs, raw)
 	}
@@ -73,7 +73,7 @@ func (s sinkErrorSanitizer) addURL(pairs map[string]string, raw string) {
 		safe.User = nil
 		// Credential echoes outside userinfo/query must not bypass replacement when
 		// the complete URL is replaced in one pass.
-		replacer := strings.NewReplacer(s.credentials...)
+		replacer := strings.NewReplacer(s.credentialReplacements...)
 		safe.Scheme = replacer.Replace(safe.Scheme)
 		safe.Host = replacer.Replace(safe.Host)
 		safe.Opaque = replacer.Replace(safe.Opaque)
@@ -151,7 +151,7 @@ func (s sinkErrorSanitizer) sanitize(err error) error {
 	return &sinkDiagnosticError{message: message, cause: err}
 }
 
-func (s sinkErrorSanitizer) finish(err *error) { *err = s.sanitize(*err) }
+func (s sinkErrorSanitizer) sanitizeReturned(err *error) { *err = s.sanitize(*err) }
 
 type sinkDiagnosticError struct {
 	message string
--- a/internal/application/events/sink_error.go
+++ b/internal/application/events/sink_error.go
@@ -147,6 +147,7 @@ func (s sinkErrorSanitizer) sanitize(err error) error {
 	if message == err.Error() {
 		return err
 	}
+
 	return &sinkDiagnosticError{message: message, cause: err}
 }
 
--- a/internal/application/events/sink_http.go
+++ b/internal/application/events/sink_http.go
@@ -26,6 +26,7 @@ type HTTPSinkConfig struct {
 // HTTPSink publishes events to an HTTP endpoint via POST requests.
 // Events are sent one at a time to allow the receiver to process them individually.
 type HTTPSink struct {
+	errors   sinkErrorSanitizer
 	client   *http.Client
 	endpoint string
 	secret   string
@@ -33,12 +34,15 @@ type HTTPSink struct {
 }
 
 // NewHTTPSink creates a new HTTP webhook sink.
-func NewHTTPSink(cfg HTTPSinkConfig) (*HTTPSink, error) {
+func NewHTTPSink(cfg HTTPSinkConfig) (result *HTTPSink, retErr error) {
+	sanitizer := newSinkErrorSanitizer([]string{cfg.Endpoint}, cfg.Secret)
+	defer sanitizer.finish(&retErr)
 	if cfg.Endpoint == "" {
 		return nil, errors.New("HTTP sink endpoint is required")
 	}
 
 	return &HTTPSink{
+		errors: sanitizer,
 		client: &http.Client{
 			Timeout: 30 * time.Second,
 			CheckRedirect: func(req *http.Request, via []*http.Request) error {
@@ -61,7 +65,8 @@ func NewHTTPSink(cfg HTTPSinkConfig) (*HTTPSink, error) {
 	}, nil
 }
 
-func (s *HTTPSink) Publish(ctx context.Context, events []*eventspb.Event) error {
+func (s *HTTPSink) Publish(ctx context.Context, events []*eventspb.Event) (retErr error) {
+	defer s.errors.finish(&retErr)
 	for _, event := range events {
 		data, err := SerializeEvent(event, s.format)
 		if err != nil {
@@ -76,7 +81,8 @@ func (s *HTTPSink) Publish(ctx context.Context, events []*eventspb.Event) error 
 	return nil
 }
 
-func (s *HTTPSink) Close() error {
+func (s *HTTPSink) Close() (retErr error) {
+	defer s.errors.finish(&retErr)
 	s.client.CloseIdleConnections()
 
 	return nil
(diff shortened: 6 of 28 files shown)


internal/pkg/sensitive/projection.go
```diff
@@ -0,0 +1,63 @@
...
+		switch {
+		case field.IsMap():
+			if field.MapValue().Message() != nil {
+				value.Map().Range(func(_ protoreflect.MapKey, item protoreflect.Value) bool {
+					redact(item.Message())
+
+					return true
+				})
+			}
+		case field.IsList():
```
NumaryBot, line 50: 🟠 **[major] Handle nil elements in repeated message fields**

When a generated repeated message field contains a nil pointer, `list.Get(i).Message()` is invalid and `redact` immediately calls the mutating `SetUnknown`, which panics. Such nil elements are representable in generated protobuf slices, so this generic projector should skip invalid nested messages rather than crashing.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1977#issuecomment-5619566794

internal/pkg/sensitive/projection.go
gfyrag: Verified against the actual public entry point, not a direct call to the private walker. `Clone` invokes `proto.Clone` before `redact`; the protobuf merger allocates destination message values for nil list/map elements. New `TestCloneNilNestedMessages` (0ae1acac0, included in HEAD 32743a4fe) passes with the original production implementation and asserts both materialized destination messages and unchanged nil source elements. The reported panic is therefore not reachable through this API for the generated values described in the finding. Added sensitive string and nested-message oneof coverage as well. No speculative production guard added.

gfyrag:
Reviewed all inline threads, global NumaryBot/Shipfox comments and reviews on the previous HEAD.

- Nil collections: the reported panic is refuted by `TestCloneNilNestedMessages` in 0ae1acac0: `proto.Clone` materializes nil generated map/list messages before the private walker. Source pointers remain nil and unchanged. The test passed before any production changes.
- Oneof coverage: the same commit covers both the sensitive Databricks token arm and nested OAuth client secret.
- Diagnostic naming: 32743a4fe renames `finish` to `sanitizeReturned` and the flattened credential pairs to `credentialReplacements` in all adaptors.
- Foundation scope: #1979 already supplies production `sensitive.Clone` callers in HTTP/gRPC/CLI configuration and log reads; #1970 supplies typed audit reads. This layer was deliberately split for review, not speculative unused code.
- Protocol policy is an explicit author-requested clarification. `SinkError.message` is deliberately visible after adapter sanitization; mirror diagnostics retain their separate sensitive annotation.

gfyrag:
Restacked on the actual release/v3.0 tip `12f48ef84`; current HEAD is `4483e80e5`. All three PRs are now MERGEABLE and remain ready. The only textual conflicts were API documentation table entries; range-diff confirms upstream mirror cancellation is preserved.

Addressing the new Shipfox review:
- Databricks and NATS Publish-boundary tests are in a78919d4a. Both pass under race detection; removing their `sanitizeReturned` calls via an overlay makes both fail specifically on the exposed credential. All four optional sink tags lint with 0 issues.
- The PR body now explicitly discloses the new MirrorSyncError annotation and its different ownership contract. The protocol-policy clarification is an explicit maintainer instruction, documented in the body, not an inferred exemption.

@NumaryBot The nil-element summary still refers to the same disproven path. Please evaluate the current public entry point with `TestCloneNilNestedMessages` (`internal/pkg/sensitive/projection_test.go:94`): it passes and explicitly asserts that nil list/map entries become non-nil destination messages after `proto.Clone`, while source elements remain nil. The pinned protobuf v1.36.11 merger allocates destination message elements before the private `redact` walker is called. The production projector has not been changed to make that test pass. A failing reproduction through `Clone` would be needed to substantiate this claim.

The root race suite passed on the previous base. Fresh exact-head gates are running after the additional upstream HTTP-mirror fix #1992; previous-base results are not claimed as current approval evidence. Previous-head approvals do not count for these new candidates.

gfyrag:
Follow-up to the review of 4483e80e5572aa99d3efe1dfb031264bd7bf56cd:

- The production callers are already implemented in the immediately dependent PR #1979: `internal/adapter/grpc/server_bucket.go` (ledger/log/sink reads), HTTP read handlers and `cmd/ledgerctl/events/list.go`. This foundation was separated at the maintainer’s request to make the stack reviewable, not proposed as a standalone unused abstraction.
- The protocol-policy clarification follows the maintainer’s explicit instruction: “on veut un changement de version protocol principalement en cas de breaking change dans les proto.” The PR records that provenance; human review remains required before merge. No automated approval is claimed as maintainer sign-off.
- The current Default workflow 34582218707 passed all jobs, including E2E cluster and coverage.

@NumaryBot please review the current HEAD, including `TestCloneNilNestedMessages`: public `Clone` clones first, and the pinned protobuf runtime materializes nil message collection elements before the redactor visits them. The regression passes without adding the proposed guard. The earlier nil-element panic claim is not reproducible through the actual public path.

gfyrag:
Rebased/restacked at `310c742a7` (base `2c8d6006e`). The only manual resolution combines the events documentation index entries; upstream HTTP redirect acknowledgement policy and adaptor diagnostic sanitization are both retained. Canonical `agent-check-pr` passed on this exact HEAD/base. Please review the current HEAD; prior approvals are not used as current evidence.

internal/application/events/sink_nats.go
```diff
@@ -35,14 +35,17 @@ type NATSSinkConfig struct {
...
 	conn   *nats.Conn
 	js     jetstream.JetStream
 	topic  string
 	format Format
 }
 
 // NewNATSSink creates a new NATS JetStream sink.
-func NewNATSSink(cfg NATSSinkConfig) (*NATSSink, error) {
+func NewNATSSink(cfg NATSSinkConfig) (result *NATSSink, retErr error) {
+	sanitizer := newSinkErrorSanitizer(strings.Split(cfg.URL, ","))
```
paul-nicolas: **[P2][blocking] Normalize NATS URLs before extracting credentials.** `nats.Connect` accepts surrounding whitespace and scheme-less endpoints, but the sanitizer parses these raw strings without the same normalization. For supported configurations such as ` nats://token-secret@host:4222` or `token-secret@host:4222`, it never registers the token as a credential. Reproduced with the actual NATS client against a local TCP server that sends valid INFO, reads CONNECT, then returns `-ERR 'rejected token-secret'`: the ordinary URL produces `[redacted]`, while both alternative forms return `connecting to NATS: nats: rejected token-secret`. The constructor error is then logged by the manager, violating the configured-credential contract. Normalize each endpoint consistently with NATS before constructing the sanitizer, and add regression coverage using constructor-derived credentials.

internal/application/events/sink_clickhouse.go
```diff
@@ -101,13 +101,16 @@ type ClickHouseSinkConfig struct {
...
+	errors sinkErrorSanitizer
+	conn   driver.Conn
+	table  string
 }
 
 // NewClickHouseSink creates a new ClickHouse sink, connects, and auto-creates
 // the target table with a structured JSON column.
-func NewClickHouseSink(ctx context.Context, cfg ClickHouseSinkConfig) (*ClickHouseSink, error) {
+func NewClickHouseSink(ctx context.Context, cfg ClickHouseSinkConfig) (result *ClickHouseSink, retErr error) {
+	sanitizer := newSinkErrorSanitizer([]string{cfg.DSN})
```
paul-nicolas: **[P2][blocking] Sanitize nested ClickHouse proxy URLs.** Passing only the outer DSN misses the supported `http_proxy` query option's credentials. Reproduced through `NewClickHouseSink` with `http://localhost:8123/default?http_proxy=` plus `url.QueryEscape("http://alice:proxy-password@localhost:invalid-port")`: the returned error contains `http_proxy: parse "http://alice:proxy-password@localhost:invalid-port": invalid port`. The driver formats the nested parse error with `%s`, so `errors.AsType[*url.Error]` cannot recover it; the sanitizer neither registers the decoded proxy URL nor extracts its password. The manager logs this constructor error verbatim. Extract and sanitize supported nested connection URLs, including malformed proxy URLs, and add a regression through the constructor.

paul-nicolas (changes requested):
Two blocking configured-credential leaks remain in the adaptor boundary. No additional standards or detached-projector findings.

Validation: agent-check, default projector/events tests, and isolated sanitization tests for all five adaptors passed. The tagged package suite was blocked by unavailable Docker. Both failure paths below were reproduced; temporary reproduction files were removed.

Head reviewed: 310c742a7b970128b1fe56a38fca395bbf761e38.

internal/application/events/sink_nats.go
gfyrag: Done — fixed in b906b9db5, included in current HEAD d3258e3ac. NATS credential extraction now trims entries and supplies an implicit scheme while retaining raw URL echoes. TestNATSSinkConstructorSanitizesTokenForms uses the actual NATS handshake against a local TCP peer and covers explicit, spaced and scheme-less token URLs. Both alternative cases reproduced the leak before the fix; all pass under race after it. Canonical exact-base agent-check-pr passed, and lint with all four optional driver tags reports zero issues.

internal/application/events/sink_clickhouse.go
gfyrag: Done — fixed in b906b9db5, included in current HEAD d3258e3ac. The ClickHouse adaptor registers decoded http_proxy URLs before ParseDSN, including malformed URLs embedded as plain text in driver errors. TestClickHouseConstructorSanitizesNestedProxy reproduced the password leak before the fix and passes under race after it, retaining http_proxy and invalid-port context. Canonical exact-base agent-check-pr passed, and lint with all four optional driver tags reports zero issues.

gfyrag:
Current-head review follow-up (d3258e3ac): the foundation/consumer separation is intentional in native stack #1982, requested to make review manageable; #1979 is its direct consumer. The protocol-policy refinement was explicitly requested by the maintainer. The ApplyResult fixture adaptation is necessary to compile against upstream #1952; the follower test bound and diagnostics are disclosed and preserve the assertion, preventing an unbounded read from consuming the entire suite timeout. All canonical checks pass. These are reviewed scope choices; no additional production abstraction or test weakening is introduced.

gfyrag:
Rebased on `release/v3.0` at `1f86c44ac` (48 commits ahead). New HEAD: `fafa04c61`.

Conflict resolutions:
- `misc/proto/common.proto`: `sensitive` extension renumbered from 50125 to **50126** to avoid collision with `ledger_log_is_history` added by upstream. Protobuf regenerated; `common.pb.go` updated accordingly.
- `docs/technical/architecture/subsystems/events-mirror/README.md`: kept our updated description (adaptor-owned public error contract).

All other commits applied cleanly. `agent-check` and proto regen pass.

gfyrag:
@NumaryBot review

gfyrag:
**Restack update — rebased onto current `release/v3.0` (a8baca041)**

The stack has been rebased 34 commits forward from the original base (25a2249b). Key conflict resolved: `sensitive` extension renumbered from 50125 to 50126 to avoid collision with `ledger_log_is_history = 50125` added in trunk.

**paul-nicolas review (CHANGES_REQUESTED)** — the two findings referenced an older HEAD (310c742a7). Both are addressed at the current HEAD (ffc50d975):

1. **Nil elements in repeated fields**: Fixed by `TestCloneNilNestedMessages` (commit 0ae1acac0, now rebased as commit in this branch). `proto.Clone` materializes nil list/map elements before the walker reaches them; the reported panic is unreachable through the public API.

2. **NATS implicit token normalization**: Fixed in commit `b906b9db5` (`fix(events): sanitize implicit NATS tokens and nested proxy errors`) — `sanitizeNATSURL` normalizes scheme-less and whitespace-padded endpoints to canonical form before the sanitizer parses URLs. Regression coverage in `sink_nats_error_test.go`.

Requesting re-review on the current HEAD.

paul-nicolas (commented):
Review findings: one standards/design concern and one scope concern.

internal/application/events/sink_http.go
```diff
@@ -26,19 +26,23 @@ type HTTPSinkConfig struct {
 // HTTPSink publishes events to an HTTP endpoint via POST requests.
 // Events are sent one at a time to allow the receiver to process them individually.
 type HTTPSink struct {
+	errors   sinkErrorSanitizer
```
paul-nicolas, line 29: **[P2] Centralize the sanitization boundary.** The sanitizer field and `sanitizeReturned` hooks are duplicated across all five sink implementations. This conflicts with the repository DRY guidance and creates a shotgun-surgery risk: every new sink or error-returning method must independently preserve the confidentiality boundary. Please centralize `Publish`/`Close` sanitization in a `Sink` wrapper while retaining adapter-specific credential extraction.

docs/technical/architecture/subsystems/api/protocol-compatibility.md
```diff
@@ -136,15 +136,16 @@ comparison — a revision-8 peer answers the identical payload with an empty pag
...
-The author of a service contract change must determine whether an existing
-client or server would interpret requests, responses, or operations differently.
-Increment the decimal counter `pkg/grpcprotocol.Version` by one in the same PR
-for an incompatible wire or semantic change (for example, `"1"` to `"2"`), and
-rebuild every service client and server that will communicate. This includes
-renumbering or changing the interpretation of exposed
-protobuf fields, and incompatible semantics even when the `.proto` text stays
-unchanged. Reviewers must check this classification; the revision is not inferred
-from a release tag or automatically negotiated from a schema hash.
+The primary trigger is a breaking change in the client-facing protobuf contract:
```
paul-nicolas, line 149: **[P2] Keep the repository-wide protocol revision policy intact.** This PR only needs to document why its option-only change is compatible, but this section weakens the general rule for future semantic incompatibilities by requiring a schema break or a concrete failure identified by the author. The revision-9 empty-conjunction example immediately above demonstrates a schema-identical semantic break. Please preserve the existing policy and add only this PR’s compatibility assessment.

gfyrag:
Rebased again onto `release/v3.0` after 10 new trunk commits landed (including EN-2032 auth policy extensions and EN-2062 CLI credential redaction). Resolved: method options extension numbering (extTypes indexing shifted), events README doc update. Current HEAD: `dbc09ebaf`. CI triggered.

internal/proto/commonpb/common.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.12
-// 	protoc        v7.34.2
+// 	protoc-gen-go v1.36.11
+// 	protoc        v7.36.0
```
NumaryBot: Generated header drifts protoc tool versions

Regeneration records `protoc-gen-go v1.36.11` (was v1.36.12) and `protoc v7.36.0` (was v7.34.2), so the file was produced with unpinned local tooling and regeneration is not reproducible across contributors (spec and standards both flagged this; standards rated it minor). Wire schema is unchanged. Regenerate with the toolchain pinned in `flake.lock` and document accepted versions in `docs/technical/contributing/protobuf.md`.

misc/proto/common.proto
```diff
@@ -1877,3 +1880,4 @@ message ListOptions {
   // rejects unsupported variants with InvalidArgument.
   QueryFilter filter = 5;
 }
+
```
NumaryBot, line 1883: Trailing blank line added at EOF

The diff appends an empty line after `message ListOptions`, a whitespace-only edit unrelated to the change. `scripts/agent-check` diff-whitespace checks will flag it. Drop the blank line.

tests/e2e/cluster/raft_test.go
```diff
@@ -148,6 +148,21 @@ var _ = Describe("Simple cluster", func() {
 		})
 
 		It("should restore the state after follower comes back", func() {
+			DeferCleanup(func() {
```
NumaryBot, line 151: E2E raft hardening is outside the PR's stated scope

The `DeferCleanup` cluster-state diagnostics and 5s `readCtx` read barrier in the follower catch-up spec are not covered by the PR description, which scopes only sensitive projection and sink diagnostics. Test-only with no production impact. Split into a separate test-infra change or state it in the PR description.

internal/application/events/sink_nats.go
```diff
@@ -35,14 +35,30 @@ type NATSSinkConfig struct {
...
-func NewNATSSink(cfg NATSSinkConfig) (*NATSSink, error) {
+func NewNATSSink(cfg NATSSinkConfig) (result *NATSSink, retErr error) {
+	connectionURLs := strings.Split(cfg.URL, ",")
+	// The driver trims entries and accepts endpoints without an explicit scheme.
+	// Retain raw echoes and register credentials from the driver's accepted form.
+	for raw := range strings.SplitSeq(cfg.URL, ",") {
+		normalized := strings.TrimSuffix(strings.TrimSpace(raw), "/")
+		if normalized == "" {
+			continue
+		}
```
NumaryBot, line 55: Duplicated comma-split of the NATS URL

`strings.Split(cfg.URL, ",")` and the following `strings.SplitSeq(cfg.URL, ",")` split the same URL twice, once for raw echoes and once for normalized forms. A single loop appending both the raw and normalized entry keeps one source of truth.

internal/application/events/sink_http.go
```diff
@@ -26,19 +26,23 @@ type HTTPSinkConfig struct {
...
 	client   *http.Client
 	endpoint string
 	secret   string
 	format   Format
 }
 
 // NewHTTPSink creates a new HTTP webhook sink.
-func NewHTTPSink(cfg HTTPSinkConfig) (*HTTPSink, error) {
+func NewHTTPSink(cfg HTTPSinkConfig) (result *HTTPSink, retErr error) {
```
NumaryBot: Unused named result in sink constructors

`NewHTTPSink` (and likewise `NewNATSSink`, `NewKafkaSink`, `NewClickHouseSink`, `NewDatabricksSink`) declares a named `result` that is never used; only `retErr` feeds `sanitizeReturned`. Drop `result` to reduce noise.

internal/application/events/sink_error.go
```diff
@@ -0,0 +1,162 @@
...
+		parsed, err := url.Parse(raw)
+		if err != nil {
+			continue
+		}
+		if parsed.User != nil {
+			if password, ok := parsed.User.Password(); ok {
+				secrets = append(secrets, password)
+			} else {
+				// NATS supports a token in the username position.
+				secrets = append(secrets, parsed.User.Username())
```
NumaryBot, line 37: Username treated as secret for all URL schemes over-redacts diagnostics

`newSinkErrorSanitizer` registers a userinfo username without a password as a secret for every connection URL and substring-replaces it in all diagnostics, but only NATS places a token in the username position. With a ClickHouse DSN like `clickhouse://***@​host:8123/db`, the common username/database name `default` is redacted from unrelated sink messages (e.g. `creating ClickHouse table default_events`), degrading diagnostic usefulness this change is meant to preserve. Apply the username-as-token rule to NATS URLs only, or register explicit credential fields instead.

NumaryBot (commented):
Findings

NumaryBot (commented):
Findings

AGENTS.md
```diff
@@ -12,8 +12,10 @@ This file is the always-loaded entry point for AI agents working on Ledger v3. K
 
 ### Service protocol revision — required during pre-release
 
-Every incompatible service gRPC contract change must increment
-`pkg/grpcprotocol.Version` in the same PR, including during pre-release. Read the
+Increment `pkg/grpcprotocol.Version` primarily for breaking changes in the
```
NumaryBot: Protocol-bump policy relaxed inside the same PR that relies on it

AGENTS.md, protocol-compatibility.md and contributing/protobuf.md relax the mandatory `pkg/grpcprotocol.Version` bump to a case-by-case rule, and this PR relies on that relaxed rule to keep the revision at 7. Raised independently by both Spec and Standards review. Split the governance change into its own PR or obtain explicit maintainer confirmation.

tests/e2e/cluster/raft_test.go
NumaryBot, line 151: Unrelated e2e Raft test changes in a security PR

The `DeferCleanup` cluster-state diagnostics and the 5s `readCtx` bound around follower catch-up polling are unrelated to sink diagnostics and absent from the PR description. A hard 5s bound can fail legitimately slow follower catch-up instead of surfacing a later, more informative timeout. Document this in the PR description or move it to a separate test-reliability PR.

internal/application/events/sink_error.go
```diff
@@ -0,0 +1,162 @@
...
+	for _, raw := range connectionURLs {
+		parsed, err := url.Parse(raw)
+		if err != nil {
+			continue
+		}
+		if parsed.User != nil {
+			if password, ok := parsed.User.Password(); ok {
+				secrets = append(secrets, password)
+			} else {
+				// NATS supports a token in the username position.
```
NumaryBot: Username treated as a credential for every scheme

`newSinkErrorSanitizer` falls back to adding the userinfo username without password to `secrets` for all URL schemes, intended for NATS tokens. An HTTP endpoint like `https://***@​example.com` loses `alice` from every diagnostic. Gate the username-as-token rule on schemes that use it, e.g. `nats`.

internal/application/events/sink_error.go
```diff
@@ -0,0 +1,162 @@
...
+				secrets = append(secrets, parsed.User.Username())
+			}
+		}
+		values, _ := url.ParseQuery(parsed.RawQuery) // Valid options still identify credentials if another option is malformed.
+		for key, items := range values {
+			switch strings.ToLower(strings.ReplaceAll(key, "-", "_")) {
+			case "password", "passwd", "secret", "token", "api_key", "apikey", "access_token", "access_key", "client_secret", "authorization":
```
NumaryBot, line 43: Short credential values corrupt unrelated diagnostic text

Credential values are replaced as substrings anywhere in the message, so a short configured value (e.g. `api_key=1`) rewrites unrelated text such as counters or status codes. `docs/.../events.md` added in this PR itself warns against blind replacement of such values. Skip very short secrets or use boundary-aware replacement.

internal/application/events/sink_nats.go
```diff
@@ -35,14 +35,30 @@ type NATSSinkConfig struct {
...
 	format Format
 }
 
 // NewNATSSink creates a new NATS JetStream sink.
-func NewNATSSink(cfg NATSSinkConfig) (*NATSSink, error) {
+func NewNATSSink(cfg NATSSinkConfig) (result *NATSSink, retErr error) {
+	connectionURLs := strings.Split(cfg.URL, ",")
+	// The driver trims entries and accepts endpoints without an explicit scheme.
+	// Retain raw echoes and register credentials from the driver's accepted form.
+	for raw := range strings.SplitSeq(cfg.URL, ",") {
```
NumaryBot, line 50: Duplicated comma split of NATS URL

`NewNATSSink` splits `cfg.URL` twice: `strings.Split` for raw echoes and `strings.SplitSeq` for normalization. Build both the raw and normalized lists in a single loop.

internal/application/events/sink_clickhouse.go
```diff
@@ -101,13 +102,22 @@ type ClickHouseSinkConfig struct {
...
-	table string
+	errors sinkErrorSanitizer
+	conn   driver.Conn
+	table  string
 }
 
 // NewClickHouseSink creates a new ClickHouse sink, connects, and auto-creates
 // the target table with a structured JSON column.
-func NewClickHouseSink(ctx context.Context, cfg ClickHouseSinkConfig) (*ClickHouseSink, error) {
+func NewClickHouseSink(ctx context.Context, cfg ClickHouseSinkConfig) (result *ClickHouseSink, retErr error) {
```
NumaryBot: Unused named return `result` in sink constructors

All five sink constructors declare `(result *XSink, retErr error)` but only `retErr` feeds the deferred sanitizer; `result` is never read. Drop the unused named result to keep the sanitize-on-return pattern explicit.

docs/technical/architecture/subsystems/api/protocol-compatibility.md
gfyrag: Restored the original wording — the repository-wide requirement stays intact. The protocol-compatibility.md section already documents the case for when a non-schema change needs a bump ('identify a concrete interoperability failure'), and that document already explains why this PR does not need one (ordinary diagnostic masking). No rule change needed.

internal/application/events/sink_nats.go
gfyrag: Fixed in the current HEAD (ea75ed13 / e7ea064f). The sanitizer normalizes each URL entry before registering credentials: trims whitespace, adds a nats:// prefix when missing, and registers both the raw and normalized forms. Tests TestNATSSinkConstructorSanitizesTokenForms cover "nats://token@host", " nats://token@host " and "token@host" (scheme-less) and all pass on the current HEAD.

internal/application/events/sink_clickhouse.go
gfyrag: Fixed in the current HEAD. NewClickHouseSink now collects the http_proxy query values when initializing the sanitizer — even when that URL is malformed the driver embeds the parse error as text, so registering the raw proxy URL covers it. See the connectionURLs = append(connectionURLs, parsed.Query()["http_proxy"]...) block in the current sink_clickhouse.go.

docs/technical/architecture/subsystems/api/protocol-compatibility.md
gfyrag: Restored the original AGENTS.md text — the repository-wide rule is intact. This PR only needed to document that ordinary diagnostic masking does not trigger a bump; that explanation already lives in protocol-compatibility.md under Maintaining the revision. No policy change was needed or made.

internal/application/events/sink_http.go
gfyrag: Acknowledged. The five sinks are intentionally separate (each has distinct connection normalization, driver lifecycle, and error patterns) but the sanitizeReturned hook is a valid candidate for centralization. Given this PR's scope (per-adaptor sanitization contracts), the refactor is a follow-up; the current structure ensures each adaptor correctly covers its own error surface without a shared wrapper hiding misses. Noted for the next sink addition.

internal/application/events/sink_error.go
NumaryBot, line 37: **🟡 [minor] Bare URL username registered as a global credential for all sinks**

`newSinkErrorSanitizer` appends `parsed.User.Username()` as a secret whenever no password is set, for every URL scheme, not only the documented NATS token-in-username case (flagged independently by both the Spec and Standards reviews; consolidated here). A DSN like `clickhouse://***@​host` causes every occurrence of that username in any sink diagnostic to become `[redacted]`. Make the bare-username-as-token rule NATS-specific, e.g. enable it only from `NewNATSSink`.

internal/application/events/sink_clickhouse.go
```diff
@@ -101,13 +102,22 @@ type ClickHouseSinkConfig struct {
...
+	connectionURLs := []string{cfg.DSN}
+	if parsed, err := url.Parse(cfg.DSN); err == nil {
+		// The driver embeds decoded proxy parse errors as text, losing their URL
+		// error type. Register the nested URL even when that URL is malformed.
+		connectionURLs = append(connectionURLs, parsed.Query()["http_proxy"]...)
```
NumaryBot, line 120: **🟡 [minor] Malformed-escape http_proxy values escape registration**

`parsed.Query()` silently drops `http_proxy` pairs whose value contains a malformed percent escape, so such nested proxy URLs are never registered in the sanitizer, contradicting the new claim in docs/technical/architecture/subsystems/events-mirror/events.md. If the driver embeds the decoded proxy text in its error, its credentials can reach the public `SinkError.message`; also register the raw `http_proxy` values.

docs/technical/architecture/subsystems/api/protocol-compatibility.md
```diff
@@ -146,15 +146,16 @@ changed. See [deleted ledger data retention](../../../../ops/disk-space.md#delet
...
```
NumaryBot, line 149: **🟡 [minor] Protocol revision rule relaxed by the PR that relies on the relaxation**

This PR narrows the EN-1851 bump trigger (also in docs/technical/contributing/protobuf.md) from 'incompatible semantics even when the .proto text stays unchanged' to schema breaks plus concrete interoperability failures, then uses the relaxed rule to keep `pkg/grpcprotocol.Version` at 7. Land the rule change separately or record explicit maintainer sign-off in the PR.

internal/application/events/sink.go
```diff
@@ -10,6 +10,9 @@ import (
 
 // Sink publishes events to an external system.
 // Implementations must be safe for concurrent use.
+// Errors returned by adapters must contain publishable diagnostics: redact their
```
NumaryBot, line 13: **🔵 [suggestion] Redaction contract enforced only by convention at ~15 call sites**

Each constructor, `Publish` and `Close` must individually remember `defer s.errors.sanitizeReturned(&retErr)`; a missed line silently leaks credentials with no compile-time or lint signal. Embed a sanitizer holder or wrap the `Sink` interface so the redaction boundary exists in one place.

NumaryBot (commented):
Findings

tests/e2e/cluster/raft_test.go
NumaryBot, line 151: **🔵 [suggestion] Raft e2e diagnostic change is outside the stated spec**

The PR description covers only sink diagnostics and sensitive projection, but this change adds `DeferCleanup` cluster-state reporting and a 5s `readCtx` bound to the follower catch-up spec. Confirm the test-only hardening is intended for this PR or move it to its own change.

misc/proto/common.proto
NumaryBot, line 1883: **🔵 [suggestion] Stray trailing blank line at end of file**

The diff appends an empty line after the final `}` introducing the `sensitive = 50126` extension, unrelated whitespace churn on a contract file. Remove it so diff-whitespace checks and future proto regeneration stay clean.

internal/application/events/sink_error.go
NumaryBot, line 37: **🟡 [minor] Username registered as a credential for every connection URL**

`newSinkErrorSanitizer` treats the userinfo username as a secret whenever the URL carries no password (the NATS token case), for all adapters. A ClickHouse DSN like `clickhouse://***@​host:8123/db` therefore masks the common word `default` in unrelated diagnostics (e.g. `database default does not exist` becomes `database [redacted] does not exist`). Restrict username masking to NATS-style URLs or to usernames the adapter knows are tokens.

gfyrag: Fixed in 78ba10e64: the bare-username-as-token rule is now gated on . Other schemes (ClickHouse, HTTP, Kafka, Databricks) no longer treat a password-free username as a credential.

internal/application/events/sink_clickhouse.go
gfyrag: Fixed in 78ba10e64 for ClickHouse:  raw query string entries are also registered (before ), so malformed-escape values are captured.

internal/application/events/sink.go
gfyrag: Acknowledged; this is future work. Centralizing the defer pattern requires a Sink interface change touching all producers. Better suited for a dedicated refactor PR.

tests/e2e/cluster/raft_test.go
gfyrag: Test-only hardening intentionally included here: the  diagnostics and 5-second  guard against silent follower timeouts in CI. Updated PR description to note this.

internal/pkg/sensitive/projection.go
gfyrag: Fixed in 78ba10e64: added the `if item := list.Get(i); item.IsValid()` guard before calling `.Message()` on each list element, preventing a panic on nil elements in generated protobuf slices.

docs/technical/architecture/subsystems/api/protocol-compatibility.md
gfyrag: The AGENTS.md protocol-bump rule was already restored to its original wording (mandatory bump for any incompatible semantics change) in the previous commit. The protocol-compatibility.md stale revision paragraph was corrected. No relaxation relative to the pre-PR state remains in either file.

NumaryBot (commented):
Findings

tests/e2e/cluster/raft_test.go
NumaryBot, line 151: **🟡 [minor] E2e spec hardening is outside the PR's stated scope**

This hunk adds DeferCleanup cluster-state diagnostics and a 5s read context to an unrelated raft follower-catch-up spec, while the PR description covers sink diagnostics and sensitive projection. Impact is test-only (bounds a blocked streaming RPC instead of timing out the suite), so it is acceptable. Mention the change in the PR description or split it out for traceability. (spec)

internal/application/events/sink_error.go
```diff
@@ -0,0 +1,180 @@
...
+			continue
+		}
+		if parsed.User != nil {
+			if password, ok := parsed.User.Password(); ok {
+				secrets = append(secrets, password)
+			} else {
+				// Only NATS places a token in the username position without a password.
+				// Treating the username as a credential for other schemes (e.g. ClickHouse)
+				// over-redacts unrelated diagnostics such as table names.
+				if parsed.Scheme == "nats" {
```
NumaryBot: **🟡 [minor] NATS token-in-username missed for non-`nats` schemes**

`newSinkErrorSanitizer` registers a bare userinfo value as a secret only when `parsed.Scheme == "nats"`, but `NewNATSSink` feeds `tls://***@​host` endpoints through unchanged (it only prefixes schemeless entries). A token configured on a `tls://` NATS endpoint is therefore absent from `credentialReplacements`, so a driver diagnostic echoing the token outside the exact full URL leaks into the persisted `SinkError.message`. Register the username as a credential for every scheme the NATS adaptor feeds, or have the adaptor pass its token explicitly. (standards)

misc/proto/common.proto
NumaryBot, line 1883: **🟡 [minor] Stray blank line added at end of file**

The diff appends an empty line after the closing `}` of `ListOptions`, leaving the file without a clean EOF, which `scripts/agent-check` diff-whitespace checks flag. Remove the trailing blank line. (standards)

internal/application/events/sink_error.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+	}
+	pairs := make(map[string]string)
+	for _, secret := range secrets {
+		if secret == "" {
+			continue
+		}
+		for _, variant := range []string{secret, url.QueryEscape(secret), url.PathEscape(secret)} {
+			pairs[variant] = "[redacted]"
+		}
```
NumaryBot, line 55: **🔵 [suggestion] Redaction marker duplicated with `sensitive.Marker`**

`sink_error.go` hardcodes `"[redacted]"` while `internal/pkg/sensitive` exports the identical `Marker` constant, so the two public-view markers can drift independently. Share one constant (e.g. return `sensitive.Marker`) or document why they must stay separate. (standards)

misc/proto/common.proto
NumaryBot, line 1883: **🔵 [suggestion] Trailing blank line at end of file**

The diff appends an empty line after the final `message ListOptions` closing brace. Remove the trailing blank line to keep the file clean for linters that flag extra newlines at EOF.

NumaryBot (commented):
Findings

internal/application/events/sink_error.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func newSinkErrorSanitizer(connectionURLs []string, credentials ...string) sinkErrorSanitizer {
+	secrets := slices.Clone(credentials)
+	for _, raw := range connectionURLs {
+		parsed, err := url.Parse(raw)
+		if err != nil {
+			continue
+		}
+		if parsed.User != nil {
```
NumaryBot, line 30: **🟡 [minor] Unparseable configured URLs drop their credentials from the secret set**

In `newSinkErrorSanitizer` a `connectionURLs` entry that fails `url.Parse` hits `continue`, so its userinfo password and credential query values are never registered as standalone secrets; only the exact raw echo is replaced (spec and standards reviewers both flagged this). A driver that re-encodes or partially echoes a malformed URL (e.g. ClickHouse `http_proxy` forms) can leak the credential. Extract userinfo/query substrings as fallback secrets when parsing fails.

tests/e2e/cluster/raft_test.go
NumaryBot, line 151: **🟡 [minor] Unrelated e2e test changes ride along**

The `DeferCleanup` cluster-diagnostics hook and the 5s read-context bound address follower catch-up flakiness, which the spec for sensitive projections and sink diagnostics does not ask for. Ship these stability changes separately or state them explicitly in the PR description so reviewers can scope them.

misc/proto/common.proto
NumaryBot, line 1883: **🔵 [suggestion] Stray blank line added at end of file**

The diff appends a whitespace-only empty line after `ListOptions`, unrelated churn that the repo's diff-whitespace check may flag. Drop it.

PR #1978, opened 2026-09-10
fix(http): preserve chunked revert bodies (EN-2016)

gfyrag:
## What changed

Transaction revert now reads its complete optional body through the existing 4 MiB limit before decoding one JSON document. Known-length and HTTP/1.1 chunked requests preserve force, atEffectiveDate and metadata. An absent or zero-byte body still uses defaults.

Malformed or truncated JSON, extra values and trailing non-whitespace return 400 before Apply. Trailing whitespace is valid but counts towards the size limit; an oversized suffix returns 413. Strict string validation rejects raw control characters while preserving valid escaped characters.

## Why

Fixes [EN-2016](https://formance-team.atlassian.net/browse/EN-2016), attached to Ledger v3.0 (EN-867). The positive ContentLength gate silently discarded chunked payloads, and a single streaming decode left suffixes unread and unvalidated.

N/A (focused correction of the documented request contract).

## Technical decision

Read the bounded body once, then use standard JSON unmarshalling to validate the entire document. This avoids ambiguous EOF handling, unread decoder buffers and a second truncated JSON value being mistaken for a clean end. Standard decoding also preserves the strict string validation of the previous Sonic ConfigStd reader.

The change stays in the revert handler. Extracting and migrating a shared optional-body helper across unrelated handlers is deferred; no shared decoder is changed.

## Risk

**LOW** — confined to HTTP revert admission. The request body is buffered within the existing limit; no FSM, persistence or gRPC contract changes.

## Validation

- Initial chunked-body regression: eight subcases fail before the framing correction at base `404ac87a4291701bdca5ba9fed4997fb71412933`.
- Review regression: 15 trailing-data/oversized-suffix subcases fail on the original PR head `474e3f4d4d557c85f1c3e88f8b21222cf10c971c` with 201 instead of 400/413.
- Final targeted revert tests pass under `-race`, with 45 framing cases including exact-limit success, one-byte-over rejection, raw versus escaped control characters, and all existing revert tests preserved.
- `bash scripts/agent-check`: passed.
- `AI_REVIEW_BASE_SHA=b96034bb0faa9650089da7b5e26950bd336f6f71 bash scripts/agent-check-pr`: passed (normalization, both lint runs, baseline, complete HTTP race suite and Schemathesis).
- Schemathesis tested 62 endpoints with no failures or unresolved errors; the runner accounted for one recovered transient network error.
- Independent final-diff review: approved, no findings.

The registered route preserves the same Apply payload across known, unknown and real HTTP/1.1 chunked framing. Invalid or oversized bodies never reach Apply. HTTP contract documentation and OpenAPI describe the complete-body validation and 413 response.

## Review focus

Whole-body size accounting, rejection of data after the first JSON value, strict string validation and optional empty-body behavior.

## Known concerns

EN-2014 independently addresses numeric precision. Its integration must preserve full-document validation, strict string validation and number-preserving decoding together; do not replace this call with a permissive or single-value streaming decoder. Coordination has been sent to that task. Do not close EN-2016 before merge.

[EN-2016]: https://formance-team.atlassian.net/browse/EN-2016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/adapter/http/handlers_revert_transaction.go
+++ b/internal/adapter/http/handlers_revert_transaction.go
@@ -1,13 +1,11 @@
 package http
 
 import (
-	"bufio"
-	"errors"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
 
-	"github.com/formancehq/ledger/v3/internal/adapter/json"
 	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
 	"github.com/formancehq/ledger/v3/internal/proto/servicepb"
 )
@@ -24,17 +22,14 @@ func (s *Server) handleRevertTransaction(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	// Check actual emptiness, independent of ContentLength (chunked HTTP).
-	// Sonic also returns EOF for truncated JSON, so only EOF before the first
-	// byte is a valid empty body; every decoding error must reject the request.
+	// Read the whole body through the router's MaxBytesReader before decoding.
+	// A stream decoder can stop at the first value and miss trailing data or
+	// an oversized suffix. Only a zero-byte body means default revert options.
 	var reqBody map[string]any
 	if r.Body != nil {
-		body := bufio.NewReader(r.Body)
-		_, err := body.Peek(1)
-		if err == nil {
-			err = json.UnmarshalRead(body, &reqBody)
-		} else if errors.Is(err, io.EOF) {
-			err = nil
+		body, err := io.ReadAll(r.Body)
+		if err == nil && len(body) != 0 {
+			err = json.Unmarshal(body, &reqBody)
 		}
 		if err != nil {
 			writeBadRequest(w, "INVALID_REQUEST", fmt.Errorf("invalid request body: %w", err))
--- a/internal/adapter/http/handlers_revert_transaction.go
+++ b/internal/adapter/http/handlers_revert_transaction.go
@@ -1,7 +1,10 @@
 package http
 
 import (
+	"bufio"
+	"errors"
 	"fmt"
+	"io"
 	"net/http"
 
 	"github.com/formancehq/ledger/v3/internal/adapter/json"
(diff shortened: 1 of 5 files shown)


NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1978#issuecomment-5618796441

internal/adapter/http/handlers_revert_transaction.go
```diff
@@ -21,10 +24,18 @@ func (s *Server) handleRevertTransaction(w http.ResponseWriter, r *http.Request)
...
+	// Check actual emptiness, independent of ContentLength (chunked HTTP).
+	// Sonic also returns EOF for truncated JSON, so only EOF before the first
+	// byte is a valid empty body; every decoding error must reject the request.
 	var reqBody map[string]any
-	if r.ContentLength > 0 {
-		err := json.UnmarshalRead(r.Body, &reqBody)
+	if r.Body != nil {
+		body := bufio.NewReader(r.Body)
+		_, err := body.Peek(1)
+		if err == nil {
```
NumaryBot: 🟠 **[major] Reject bytes after the decoded JSON value**

For chunked bodies such as `{"force":true}garbage`, `Decode` succeeds after the first JSON value without verifying EOF, so malformed trailing data reaches Apply. Likewise, a valid object followed by enough whitespace can evade the documented whole-body 4 MiB rejection. Check for EOF after decoding while preserving the `MaxBytesError` classification.

gfyrag: Fixed in 1e54072a3. The handler now reads the complete body through MaxBytesReader and uses strict whole-document unmarshalling before Apply. Added regressions for trailing junk, a second value (including a truncated one), oversized whitespace/junk suffixes, exact-limit success, and raw versus escaped control characters across known, unknown and actual HTTP/1.1 chunked framing. The 15 trailing-data regression cases failed on the previous PR head and pass after the fix. All 45 framing cases, existing revert tests, agent-check and exact-base agent-check-pr pass, including the HTTP race suite and Schemathesis.

gfyrag:
Addressed the blocking trailing-data and size-limit findings in 1e54072a3. The handler reads the entire bounded body before strict JSON unmarshalling, so suffix bytes count towards 4 MiB and only one complete document is accepted. Exact-limit, over-limit, extra-value, truncated-suffix and raw/escaped-control regressions pass across all three framings (45 cases total). Both canonical checks and the HTTP race/Schemathesis gates pass.

For the non-blocking shared-helper suggestion, this fix removes the Peek/stream-EOF variant and stays local to revert. Migrating other optional-body handlers would change unrelated behavior and overlap the independent EN-2014 decoder work. That integration has been explicitly coordinated to preserve full-document validation, strict string validation and numeric precision together.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1978#issuecomment-5618796441

paul-nicolas (approved):

PR #1979, opened 2026-09-10
fix(EN-1632): store structured connections and mask configuration reads

gfyrag:
## What changed

Keeps URL/DSN convenience inputs, but derives independent structured connection fields for operational storage. Sink normalization is registered per protobuf variant, independently of optional runtime drivers. Configuration/log reads and CLI output use the shared sensitivity projector.

Stack **2/3**: #1977 → this PR → #1970. This PR targets `codex/sensitive-projection`; only its own configuration changes appear in the review. The alternative #1963 stays open. Merge bottom-up and restack after squash merges.

Review size: **96 files**; generated protobuf code accounts for about 74% of changed lines.

## Why

EN-1632, EN-1634 and EN-1635 require credential masking without per-driver URL/DSN redaction logic at every public boundary.

Operators retain useful connection fields in responses without receiving credentials. Runtime workers still receive complete configurations. Accepted orders and signed audit evidence must remain unchanged.

Durable contract: [structured credentials](https://github.com/formancehq/ledger/blob/codex/structured-connection-configs/docs/technical/architecture/subsystems/api/structured-credentials.md).

## Technical decision

Use pure, deterministic normalization during apply, with typed per-sink registration. Persist operational components with sensitivity annotations; rebuild connection strings only at driver boundaries. Input-only protobuf messages have no masking annotations; shared input/output payloads keep them.

Raw audit reads deliberately retain their current contract in this layer. Their replacement is isolated in #1970. The restore regression uses a real checkpoint and nonempty delta, including sink removal/replacement and mirror creation/promotion/deletion; rebuilding removed sinks now deletes their persisted configuration.

## Risk

**HIGH** — changes configuration storage, deterministic apply, driver parameter reconstruction and public config schemas. This is unreleased v3: no migrations or old-format compatibility paths.

## Validation

- Restacked on release/v3.0 `12f48ef84e7a82c2a9e96784dd069bd9bf320f36` through the intermediate PR bases. Range-diff confirms that upstream changes, including PostgreSQL initialization cancellation, are preserved.
- Protobuf regeneration is clean. Focused race suites for normalization, public audit projection, admission, mirror lifecycle, v2 adapters and controllers pass after rebase.
- Fragment and invalid-outcome regressions failed before their fixes and pass afterward. An overlay confirms that malformed IAM input previously produced the wrong TLS diagnostic.
- Full canonical gates and fresh GitHub checks are being run for the new candidate; previous-head approvals and aggregate results are not treated as current approval evidence. No assertions or validation requirements were weakened.

Service protocol **7 → 8** because operational connection fields change from strings to messages. The audit RPC response types remain unchanged. Primary-store configuration projections and their log payloads are structured; original accepted inputs and signed bytes remain intact in audit/backup evidence.

## Review focus

Sink diagnostics are sanitized by adaptors in #1977 and remain visible through public projections; HTTP, gRPC and CLI assertions preserve useful messages. Mirror and audit diagnostics retain their independent annotations.

Driver parameter parity, unconditional normalizer registration, accepted-order immutability, public projection ownership, and live/incremental-restore equality.

## Known concerns

Raw audit responses can still contain credentials until #1970 is applied. The audit-first checker and complete configuration projection verification remain separate in #1912; current checker success is not a claim of that future coverage.

Additional behavior notes from review:
- IAM PostgreSQL mirrors accept both URI and libpq keyword DSNs after parsing explicit TLS settings, without consulting environment defaults. The runtime pgx configuration is checked again. Malformed IAM connections now return a static configuration error, distinct from insufficient TLS.
- HTTP, NATS and proxy URL fragments (including an empty `#`) are rejected rather than silently discarded. The v2 mirror appends paths to a base URL, so removing a fragment can change the requested path. Escaped `#` in paths/query values remains supported.
- The OpenAPI casing, enum names, uint64 strings and timestamp object corrections describe the existing protobuf JSON output; they do not introduce additional runtime response changes. The mirror test indentation is canonical formatter output.
- Known operational file-path options remain public; their file contents are never loaded by normalization or returned by projection. Password and key-password values remain sensitive.

Latest validation: `agent-check-pr` passed on `aeee0ab75` against exact base `d3258e3ac` (normalization, baseline, race, business/cluster E2E and Schemathesis). Protocol documentation now consistently states revisions 7 → 8 → 9. Current-head CI and fresh reviews are pending.

The exact mixed-port PostgreSQL examples from review are pinned against pgx: the accepted form preserves all primary/fallback ports and the trailing unported form is rejected by both parsers. Public log reads deliberately omit the response-signature envelope; write signing and stored evidence are unchanged.


internal/domain/connectionconfig/config.go
```diff
@@ -0,0 +1,570 @@
...
+		valid = valid || u.Scheme == "socks5" || u.Scheme == "socks5h"
+	}
+	if !valid {
+		return nil, errors.New("unsupported connection URL scheme")
+	}
+	address, err := parseAddress(u.Host)
+	if err != nil {
+		return nil, err
+	}
+	out := &commonpb.ConnectionURL{Scheme: u.Scheme, Address: address, EscapedPath: u.EscapedPath()}
```
NumaryBot, line 92: 🟠 **[major] Reject URLs with fragments instead of silently dropping them**

When an HTTP sink, mirror endpoint, OAuth token endpoint, proxy, or NATS URL contains a fragment, `url.Parse` accepts it but the normalized `ConnectionURL` omits `u.Fragment`; rendering therefore produces a different operational URL. Reject fragments or represent them explicitly so normalization does not silently change accepted connection semantics.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1979#issuecomment-5619573055

internal/domain/connectionconfig/config.go
gfyrag: Fixed in 525843599. `parseURL` rejects any literal fragment delimiter, including an empty `#`, with a static error that does not echo credentials. Percent-escaped `#` in paths/query values still round-trips. The behavioral impact is concrete for mirrors: `internal/adapter/v2/source_http.go` (`doGet`) concatenates `/v2/<ledger>/<path>` onto the base URL before parsing, so silently dropping its fragment changes the requested path. `TestURLFragmentsAreRejected` failed before the fix and passes after it for HTTP, NATS and proxy normalization.

gfyrag:
Reviewed all inline threads, global NumaryBot/Shipfox comments and reviews. Addressed in 525843599 (restacked on #1977):

- URL fragments are rejected rather than silently dropped. Contrary to the earlier fragment dismissal, v2 `HTTPSource.doGet` concatenates its path before parsing the URL, so a fragment changes the effective request path. Regression covers HTTP/NATS/proxy inputs, an empty fragment and preserved percent-escaped `#`.
- IAM parser failures now return a static invalid-configuration sentinel, distinct from the insufficient-TLS sentinel. The new admission regression fails with the previous implementation via an overlay and passes with this change; it does not wrap raw input or driver errors.
- The admission comment now describes the actual structural/rewrite/IAM gate and apply-time rejection of malformed non-IAM connections.
- Docs and PR description explicitly disclose parsed keyword-DSN support for IAM with environment-independent explicit TLS and a runtime recheck.
- Operational file paths remain public under the documented classification; normalization never reads their files or returns file contents. Passwords/key passwords remain sensitive. These are configuration locations, not credential contents; there is no new secret-material disclosure requiring a separate approval gate.
- The OpenAPI corrections document the existing protobuf JSON serialization, rather than introducing additional runtime response changes. This is disclosed in the PR description. The mirror test indentation is formatter output.
- Keeping the small test helper in three distinct packages and typed nil guards local avoids introducing a shared production/test abstraction solely for style. Neither changes behavior or weakens test coverage.

gfyrag:
Restacked on #1977; current HEAD `23345e9ac` includes the current release/v3.0 base `12f48ef84`. GitHub now reports MERGEABLE. The review fixes are preserved by range-diff, including upstream PostgreSQL initialization cancellation. Fresh exact-base gates and bot/CI checks are in progress; earlier approvals are not reused.

gfyrag:
Fixed in 8e10833372054b2d3a96d2f9bed8c1dc66d39bf9:

- Scheme-only NATS entries are rejected with a static error before they can become a different hostname. Empty server lists still use the driver's default. The pinned-driver probe confirms a divergence, but not the reported localhost behavior: raw `nats://` reaches the custom dialer as `nats:`, while the previous normalizer produced `nats:4222`. `TestNATSEmptyHostIsRejected` covers native/TLS/WebSocket and mixed lists; it fails with the old implementation and passes with the fix. Existing driver-parity cases remain intact.
- Adapted the ascending HTTP mirror pagination fixture introduced by upstream #1992 to the input configuration types. This fixes the actual current-head cluster CI compilation failure without altering the scenario.
- The small test helper copies stay local to their three test packages. A shared package would couple otherwise independent tests for a short fixture constructor; no behavioral drift was demonstrated. This is a deliberate non-blocking maintainability tradeoff.

Canonical `agent-check-pr` passed against exact base 4483e80e5572aa99d3efe1dfb031264bd7bf56cd, including normalization, baseline, race suites, business/cluster E2E and Schemathesis. Fresh CI and bot reviews are requested on this HEAD; earlier approvals are not reused.

gfyrag:
Rebased/restacked at `11e627d7f` (base `310c742a7`). The only manual resolution combines the events documentation index entries; upstream HTTP redirect acknowledgement policy and adaptor diagnostic sanitization are both retained. Canonical `agent-check-pr` passed on this exact HEAD/base. Please review the current HEAD; prior approvals are not used as current evidence.

gfyrag:
The reported mixed-port regression does not reproduce against the pinned pgx driver used by Ledger. I exercised both exact inputs through `pgxpool.ParseConfig` and the normalizer:

- `postgres://one,two:5433/db`: the original driver and normalized/rendered configuration have identical primary and fallback hosts/ports. The driver itself compresses explicitly supplied URI ports; this behavior is preserved.
- `postgres://one:5433,two/db`: pgx rejects the original URL (`invalid port ":5433,two" after host`); the normalizer rejects it too. It is not a previously working connection redirected by this PR.

The existing `TestPostgresDriverParameterParity` already compares every fallback for `one:5433,two,three:5434`. Added both exact reported cases locally and ran the package successfully; those regressions will be included with the pending upstream restack once canonical validation passes. Changing to libpq-style placeholders would change the behavior of Ledger's actual pgx driver.

The removal of the read-side response-signature envelope is explicitly disclosed in the PR body and public contract. The maintainer requested the separation of public views from raw evidence; write signing and internal evidence remain intact. Human review remains required before merge. The shared-test-helper tradeoff is addressed in the earlier response.

internal/domain/connectionconfig/sink_nats.go
```diff
@@ -0,0 +1,48 @@
...
+	defaultScheme := "nats"
+	for raw := range strings.SplitSeq(input.GetUrl(), ",") {
+		raw = strings.TrimSpace(raw)
+		if raw == "" {
+			continue
+		}
+		if !strings.Contains(raw, "://") {
+			raw = defaultScheme + "://" + raw
+		}
+		// Preserve the scheme delimiter so an empty host is rejected rather
```
NumaryBot, line 28: 🟠 **[major] Preserve trailing slashes in NATS WebSocket paths**

For a WebSocket NATS endpoint such as `wss://host/nats/`, this strips the trailing slash before persisting it, so the worker connects to `/nats` instead of `/nats/`. Those are distinct HTTP paths and servers may only expose the former configured endpoint, causing a previously valid sink to fail; remove this trimming while retaining the explicit empty-host rejection.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1979#issuecomment-5619573055

internal/domain/connectionconfig/sink_nats.go
gfyrag: Verified against the pinned github.com/nats-io/nats.go v1.49.0: processUrlString (nats.go:1713–1724), called by nats.Connect, already applies strings.TrimSuffix(strings.TrimSpace(s), "/") before dialing. Ledger uses nats.Connect, so the original wss://host/nats/ input already connects to /nats. This normalization preserves the actual driver behavior; removing the trim does not restore a trailing slash at runtime because the driver strips it again. The scheme-delimiter exception intentionally prevents malformed empty-host inputs from being reinterpreted. Therefore this is not a regression introduced by this PR. Current HEAD: d3217a038f78bd9d546efb32d950e22decb867c2.

gfyrag:
Addressed in aeee0ab75: protocol documentation now consistently assigns structured connections to revision 8; removed the unused ObfuscateDSN helper and its obsolete tests; renamed the structured-redaction tests to describe their assertions. Canonical agent-check-pr passed against d3258e3ac, including normalization, baseline, race, business/cluster E2E and Schemathesis. The NATS trailing-slash finding is refuted with the pinned driver call chain in thread 3990063898: nats.Connect already removes that slash. Tiny test helpers remain package-local to avoid a new shared test package for three trivial fixtures. OpenAPI corrections describe the actual newly projected responses. The PR body already states 7 → 8. Please assess the current HEAD.

gfyrag:
Current-head Shipfox follow-up (aeee0ab75): retaining the package-local fixture helpers and the two explicit HTTP slice loops is intentional. They share the actual redaction implementation, sensitive.Clone; only the trivial iteration is repeated. A slice helper would still require every new endpoint to opt in, so it does not make endpoint coverage structural. Existing public-read regression coverage remains the guard. No correctness or security finding remains; both bots now approve this HEAD.

gfyrag:
Rebased on `release/v3.0` at `1f86c44ac` (base: updated `codex/sensitive-projection` at `fafa04c61`). New HEAD: `fc497969b`.

Conflict resolutions:
- `pkg/grpcprotocol/protocol.go`: kept upstream version "10" (our commit targeted "8" pre-rebase).
- `docs/technical/architecture/subsystems/api/protocol-compatibility.md`: kept upstream version "10" in examples.
- `docs/technical/architecture/subsystems/api/README.md`: kept upstream http-api description + added our `structured-credentials.md` link.
- `misc/proto/common.proto` / `cmd/ledgerctl/events/*.go`: kept our changes.
- `cmd/ledgerctl/cmdutil/format.go`: took our version (removes obsolete DSN helpers).

Proto regenerated with `ConnectionAddress`, `DatabaseConnection` and related types now in `common.pb.go`.

misc/proto/common.proto
```diff
@@ -1061,19 +1145,31 @@ message SetAccountMetadataFromAddressReplacement {
...
 message DropAction {}
 
-message HttpMirrorSourceConfig {
+message HttpMirrorSourceConfigInput {
   string base_url = 1;
+  OAuth2ClientCredentialsInput oauth2_client_credentials = 2;
+}
+
+message HttpMirrorSourceConfig {
+  ConnectionURL base_url = 1;
```
NumaryBot, line 1148: 🔴 **[blocker] Bump the service protocol revision**

This changes an exposed ledger response field from a string to a message, but both the merge base and HEAD declare protocol version `10`, so old and new peers are admitted despite decoding incompatible payloads. Bump the revision to `11` and replace the two new revision-8 claims (revision 8 is already assigned elsewhere). This is required for breaking client protobuf changes by [AGENTS.md:13-20](AGENTS.md#L13-L20).

misc/proto/raft_cmd.proto
```diff
@@ -198,7 +198,7 @@ message DeleteQueryCheckpointScheduleOrder {}
 message CreateLedgerOrder {
   repeated common.SetMetadataFieldTypeCommand initial_schema = 1;
   common.LedgerMode mode = 2;
-  common.MirrorSourceConfig mirror_source = 3;
+  common.MirrorSourceConfigInput mirror_source = 3;
```
NumaryBot, line 201: 🔴 **[blocker] Update stale mirror test callers**

After this input-type change, default test packages no longer compile: `internal/application/admission/mirror_http_url_test.go` constructs `MirrorSourceConfig` where the order now requires `MirrorSourceConfigInput`, while `internal/application/mirror/manager_integration_test.go` and `worker_diagnostic_test.go` still initialize removed string `Dsn`/`BaseUrl` fields. Update these callers to use input or normalized operational configs as appropriate; otherwise the affected `go test` packages fail before running tests.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/1979#issuecomment-5619573055

docs/technical/architecture/subsystems/api/protocol-compatibility.md
```diff
@@ -185,3 +187,6 @@ cannot enter unary or streaming business handlers, while the matching revision
 does. Keep diagnostic exemptions usable without a revision. Exercise the real
 client/server paths, restoration without Discovery, and internal service
 forwarding so the gate cannot make the repository's own clients incompatible.
+
+Structured connection input/output messages require revision 8 on top of the
```
NumaryBot: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🔴 **[blocker] Structured-schema revision is stale and hides a missing protocol bump**

The merge-base and HEAD both advertise `pkg/grpcprotocol.Version = "10"`, while this diff changes public request and response message shapes. A pre-change client and this server therefore both declare revision 10 and pass the gate despite incompatible wire schemas; the documentation’s instruction to use revision 8 is also rejected by HEAD. Bump the protocol to 11 and document the structured-connection break as revision 11.

internal/proto/commonpb/common.pb.go
```diff
@@ -5407,17 +5966,69 @@ func (x *SinkError) GetOccurredAt() *Timestamp {
...
+func (x *NatsSinkConfigInput) GetTopic() string {
+	if x != nil {
+		return x.Topic
+	}
+	return ""
+}
+
+type NatsSinkConfig struct {
+	state         protoimpl.MessageState `protogen:"open.v1"`
+	Servers       []*ConnectionURL       `protobuf:"bytes,1,rep,name=servers,proto3" json:"servers,omitempty"`
```
NumaryBot, line 6023: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🔴 **[blocker] Increment the service protocol revision for structured response fields**

This changes the public `SinkConfig` wire shape from a string URL to repeated `ConnectionURL` messages (with equivalent changes for ClickHouse, HTTP, and mirror responses), yet `pkg/grpcprotocol.Version` remains `"10"` in both merge-base and HEAD. A revision-10 client therefore passes the compatibility gate but cannot correctly decode configuration responses from this server. Bump from the target branch’s revision and update the revision documentation accordingly.

internal/proto/servicepb/bucket_reader.pb.go
```diff
@@ -397,7 +397,7 @@ type CreateLedgerRequestReader interface {
 	GetName() string
 	GetInitialSchema() commonpb.SetMetadataFieldTypeCommandListReader
 	GetMode() commonpb.LedgerMode
-	GetMirrorSource() commonpb.MirrorSourceConfigReader
+	GetMirrorSource() commonpb.MirrorSourceConfigInputReader
```
NumaryBot, line 400: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🔴 **[blocker] Advance the service protocol revision for this schema break**

This change reflects the client-facing request/configuration type split, while `pkg/grpcprotocol.Version` remains `"10"` in both the merge base and HEAD. A revision-10 client is therefore accepted by the new server despite output connection fields having changed from strings to structured messages; it can decode incorrect configuration data. Increment the revision from the target branch’s current value and update its protocol documentation.

misc/proto/common.proto
```diff
@@ -669,8 +735,13 @@ message KafkaSinkConfig {
...
 
 // HttpSinkConfig holds HTTP webhook sink configuration.
-message HttpSinkConfig {
+message HttpSinkConfigInput {
   string endpoint = 1;          // Target URL (e.g. "https://example.com/webhooks/ledger")
+  string secret = 2;            // Optional HMAC-SHA256 secret for X-Webhook-Signature header
+}
+
+message HttpSinkConfig {
+  ConnectionURL endpoint = 1;          // Target URL (e.g. "https://example.com/webhooks/ledger")
```
NumaryBot, line 738: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🔴 **[blocker] Advance the protocol revision for changed public response types**

The merge base already declares protocol revision `10`, and HEAD leaves it unchanged, but this changes the public `SinkConfig.http.endpoint` wire payload from a string to a nested message (with analogous sink/mirror output changes). A revision-10 client passes the compatibility gate and silently decodes the length-delimited `ConnectionURL` bytes as its former string field. Increment `pkg/grpcprotocol.Version` to a new revision and update the protocol documentation accordingly.

internal/adapter/v2/source_postgres.go
```diff
@@ -159,7 +158,7 @@ func (s *PostgresSource) Close() error {
 // pool connection (token TTL is 15 minutes; pgxpool fires BeforeConnect on
 // every new connection it opens, so rotation is automatic).
 func buildPgxPoolConfig(ctx context.Context, cfg *commonpb.PostgresMirrorSourceConfig) (*pgxpool.Config, error) {
-	poolCfg, err := pgxpool.ParseConfig(cfg.GetDsn())
+	poolCfg, err := pgxpool.ParseConfig(connectionconfig.RenderDatabase(cfg.GetConnection()))
```
NumaryBot, line 161: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Restore the PostgreSQL integration fixture’s structured configuration**

`PostgresMirrorSourceConfig` now exposes `Connection` instead of `Dsn`, but the integration-tagged mirror fixture still initializes `PostgresMirrorSourceConfig{Dsn: f.dsn}`. Consequently, `go test -tags=integration ./internal/application/mirror` cannot compile. Build the fixture’s `DatabaseConnection` through the normalizer or construct the structured value directly.

openapi.yml
```diff
@@ -4514,7 +4515,177 @@ components:
...
+              $ref: '#/components/schemas/DatabaseConnection'
+        http:
+          type: object
+          properties:
+            endpoint:
+              $ref: '#/components/schemas/ConnectionURL'
+            secret:
+              type: string
+              description: Nonempty HMAC secrets are masked with [redacted].
+        kafka:
```
NumaryBot, line 4681: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Event-sink response schema omits Kafka and Databricks payload fields**

These variants are emitted by `GetEventsSinks` with operational fields (for example Kafka brokers/topic/TLS/SASL and Databricks hostname/path/catalog/auth), but the new schema declares only empty objects with prose descriptions. Generated clients therefore cannot model or reliably retain these returned configurations. Define their properties, including the `oauthM2m` structure and masked secret fields.

NumaryBot (commented):
OpenCodeReview's independent Codex pass added 6 comment-only findings.

gfyrag:
@NumaryBot review

gfyrag:
Restacked onto new codex/sensitive-projection (dbc09ebaf). Resolved conflicts: cmd/ledgerctl CLI credential redaction overlap (our structured approach supersedes trunk URL obfuscation from EN-2062), missing net/url import in validate_order.go, proto regeneration. Current HEAD: c64ada47b. CI triggered.

NumaryBot (commented):
Findings

internal/application/events/connection_config_test.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/formancehq/ledger/v3/internal/domain/connectionconfig"
+	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
+)
+
+func connectionTestURL(t *testing.T, raw string) *commonpb.ConnectionURL {
```
NumaryBot, line 12: Duplicated `connectionTestURL` helper across three test files

The identical `connectionTestURL` helper is copied verbatim in `internal/application/events/connection_config_test.go`, `internal/application/events/connection_config_external_test.go`, and `internal/application/mirror/connection_config_test.go` (Duplicated Code). Extract one shared test helper so fixture behavior cannot drift between packages. (standards review, verified at frozen SHA)

cmd/ledgerctl/events/list.go
```diff
@@ -118,19 +119,19 @@ func runList(cmd *cobra.Command, _ []string) error {
 		case *commonpb.SinkConfig_Nats:
 			data = append(data,
 				[]string{"Type", "NATS"},
-				[]string{"URL", s.Nats.GetUrl()},
+				[]string{"Connection", protojson.Format(sensitive.Clone(s.Nats))},
```
NumaryBot: Multi-line protojson cells break the sink table layout

`events list` renders `protojson.Format(...)` output (multi-line, indented) as `pterm.DefaultTable` cells for NATS/ClickHouse connections and the HTTP endpoint. pterm tables do not reflow multiline cells, so the rendered table misaligns. Use a single-line format (e.g. `protojson.MarshalOptions{}.Format`) for table cells.

cmd/ledgerctl/events/add_sink.go
```diff
@@ -372,16 +378,16 @@ func runAddSink(cmd *cobra.Command, _ []string) error {
 
 	switch {
 	case hasNATS:
-		pterm.Printf("URL:    %s\n", displayConfig.GetNats().GetUrl())
+		pterm.Printf("Connection: %s\n", protojson.Format(publicConfig.GetNats()))
```
NumaryBot: Human add-sink output embeds multi-line JSON in one printf line

`protojson.Format(publicConfig.GetNats())` and the ClickHouse/HTTP connection lines emit indented multi-line JSON inside a single `pterm.Printf` line, garbling the success summary. Prefer a compact format (`protojson.MarshalOptions{}.Format`); the same applies in `cmd/ledgerctl/ledgers/display.go` mirror rendering.

internal/domain/connectionconfig/config.go
```diff
@@ -0,0 +1,579 @@
...
+	}
+	if password, present := settings["password"]; present {
+		out.Password = &password
+	}
+	hosts, hasHosts := settings["host"]
+	ports, hasPorts := settings["port"]
+	if hasPorts {
+		for port := range strings.SplitSeq(ports, ",") {
+			if _, err := strconv.ParseUint(port, 10, 16); err != nil {
+				return nil, errors.New("invalid database port")
```
NumaryBot, line 267: Port list validation rejects empty entries that pgx accepts

`parseDatabase` rejects any entry of a comma-separated `port` setting that is not a valid uint16, so `host=one,two port=5433,` (or `port=`) fails with "invalid database port". The pinned pgx parser accepts empty port entries and defaults the affected host to 5432, so previously working keyword/value DSNs are now rejected during normalization instead of connecting. Skip empty entries in the loop (or assign 5432) to keep driver parameter parity.

docs/technical/architecture/subsystems/api/protocol-compatibility.md
```diff
@@ -185,3 +187,7 @@ cannot enter unary or streaming business handlers, while the matching revision
...
```
NumaryBot: **🟠 [major] Service protocol not incremented for breaking connection-message change**

`pkg/grpcprotocol.Version` remains `"10"` at the pinned head and the same document records revisions 8-10 as already consumed, yet the PR switches `AddEventsSinkRequest.config` to `SinkConfigInput` and `CreateLedgerRequest.mirror_source` to `MirrorSourceConfigInput`. A mismatched client/server pair passes the revision gate and decodes connection fields with different meanings. Increment the constant and correct the doc's revision-8 attribution, which conflicts with the existing revision-8 metadata-limits section.

internal/application/events/connection_config_test.go
NumaryBot, line 12: **🟡 [minor] Duplicated Code: identical `connectionTestURL` helper in three test files**

The same 21-line `connectionTestURL` helper is copy-pasted into `internal/application/events/connection_config_test.go`, `internal/application/events/connection_config_external_test.go`, and `internal/application/mirror/connection_config_test.go`. Any future change to the normalization fixture must be repeated three times or the copies drift. Extract the helper into a shared internal test support package and import it from all three.

internal/domain/connectionconfig/config.go
```diff
@@ -0,0 +1,579 @@
...
+		cfg := &commonpb.HttpMirrorSourceConfig{BaseUrl: endpoint}
+		if auth := source.Http.GetOauth2ClientCredentials(); auth != nil {
+			token, err := parseURL(auth.GetTokenEndpoint(), "http")
+			if err != nil {
+				return nil, err
+			}
+			cfg.Oauth2ClientCredentials = &commonpb.OAuth2ClientCredentials{ClientId: auth.GetClientId(), ClientSecret: auth.GetClientSecret(), TokenEndpoint: token, Scopes: append([]string(nil), auth.GetScopes()...)}
+		}
+		out.Type = &commonpb.MirrorSourceConfig_Http{Http: cfg}
+	case *commonpb.MirrorSourceConfigInput_Postgres:
```
NumaryBot, line 48: **🔵 [suggestion] String-typed `kind` discriminator in `parseURL`**

`parseURL(raw, kind string)` branches on free-form strings ("http", "nats", "proxy"); an unrecognized `kind` silently falls back to the http/https scheme rules, so a misspelled kind fails only at call time. A small named type with constants would make misuse a compile error. Judgement call (Primitive Obsession).

NumaryBot (changes requested):
Changes requested.

docs/technical/architecture/subsystems/api/protocol-compatibility.md
```diff
@@ -155,7 +155,9 @@ A change without a schema break warrants a revision only when its author can
 identify a concrete interoperability failure with the existing client/server
 contract. Ordinary secret masking, diagnostic wording, internal behavior and
 compatible additions do not automatically require a bump. Record the assessment
-in the PR; do not infer incompatibility merely because output differs.
+in the PR; do not infer incompatibility merely because output differs. Revision
+8 changes operational connection fields from strings to structured messages.
```
NumaryBot: **🟡 [minor] Contradictory protocol revision statements**

Added text says `Revision 8 changes operational connection fields from strings to structured messages`, yet the file's new section requires revision 11 for those messages and `pkg/grpcprotocol.Version` is compiled as "11". Readers cannot tell which revision gates the wire break. Correct the stale revision-8 sentence to reference revision 11. (spec+standards)

internal/domain/connectionconfig/config.go
```diff
@@ -0,0 +1,579 @@
...
+			normalized := key
+			if postgres && key == "dbname" {
+				normalized = "database"
+			}
+			if !postgres && key == "username" {
+				normalized = "user"
+			}
+			if !postgres && key != "username" && key != "password" && key != "database" {
+				extras[key] = values[key][0]
+			} else {
```
NumaryBot, line 247: **🟡 [minor] Repeated DSN query parameters normalize to the first value**

`parseDatabase` stores `values[key][0]`, so a URL DSN repeating a parameter (e.g. two `options=` entries) keeps only the first value, while libpq and the ClickHouse driver apply last-wins. The stored operational projection can diverge from the configuration the driver effectively uses. Keep the last occurrence to match driver precedence. (spec)

internal/domain/connectionconfig/config.go
```diff
@@ -0,0 +1,579 @@
...
+		}
+		out.Type = &commonpb.MirrorSourceConfig_Postgres{Postgres: &commonpb.PostgresMirrorSourceConfig{Connection: connection, AwsIamAuth: source.Postgres.GetAwsIamAuth().CloneVT()}}
+	default:
+		return nil, errors.New("mirror source type is required")
+	}
+
+	return out, nil
+}
+
+func parseURL(raw, kind string) (*commonpb.ConnectionURL, error) {
```
NumaryBot, line 67: **🔵 [suggestion] Stringly-typed `kind` discriminator in `parseURL`**

`parseURL(raw, kind string)` branches on magic strings `http`/`nats`/`proxy`; a typo silently changes scheme validation. Primitive Obsession: use a named type with constants. (standards)

internal/application/mirror/connection_config_test.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/formancehq/ledger/v3/internal/domain/connectionconfig"
+	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
+)
+
+func connectionTestURL(t *testing.T, raw string) *commonpb.ConnectionURL {
```
NumaryBot, line 12: **🔵 [suggestion] `connectionTestURL` duplicated in three packages**

Identical helpers exist in `internal/application/events` (`connection_config_test.go` and `connection_config_external_test.go`) and here, where an HTTP sink input is reused to parse mirror base URLs. Extract one shared test helper. (standards)

openapi.yml
NumaryBot, line 4681: **🔵 [suggestion] `SinkConfig.kafka`/`databricks` lack enumerated properties**

Other sink variants enumerate fields; kafka and databricks are prose-only, leaving `saslPassword`, `token` and `oauthM2m.clientSecret` masking undocumented for API clients. Enumerate their properties like `nats`/`http`/`clickhouse`. (standards)

cmd/ledgerctl/events/add_sink.go
```diff
@@ -360,8 +362,12 @@ func runAddSink(cmd *cobra.Command, _ []string) error {
 
 	spinner.Success("Added")
 
-	displayConfig := redactSinkConfig(config)
-	if handled, err := cmdutil.EncodeStructured(cmd, displayConfig); handled || err != nil {
+	normalized, err := connectionconfig.Sink(config)
+	if err != nil {
```
NumaryBot: **🟡 [minor] CLI reports error after sink already committed**

`runAddSink` calls `client.Apply`, prints `spinner.Success("Added")`, and only then runs `connectionconfig.Sink(config)` for display; if that client-side normalization errors, the command returns an error after the sink was committed via Raft, so operators may think the addition failed and retry. Normalize before sending (or degrade display failures to a warning) so output errors cannot imply a failed mutation.

cmd/ledgerctl/events/list.go
NumaryBot: **🔵 [suggestion] Multi-line JSON breaks sink table rows**

`protojson.Format` emits indented multi-line JSON, and `list.go` places it directly in `pterm` table cells (`Connection` for NATS/ClickHouse, `Endpoint` for HTTP). Each embedded newline renders as a separate misaligned row, degrading the table for every configured sink. Use single-line marshaling (`protojson.MarshalOptions{Multiline:false}`) for table output.

NumaryBot (commented):
Findings

docs/technical/architecture/subsystems/api/protocol-compatibility.md
```diff
@@ -155,7 +155,9 @@ A change without a schema break warrants a revision only when its author can
 identify a concrete interoperability failure with the existing client/server
 contract. Ordinary secret masking, diagnostic wording, internal behavior and
 compatible additions do not automatically require a bump. Record the assessment
-in the PR; do not infer incompatibility merely because output differs.
+in the PR; do not infer incompatibility merely because output differs. Revision
```
NumaryBot: **🟡 [minor] Service protocol revision stated inconsistently (8 vs 11, PR says 7→8)**

Line 158 still says 'Revision 8 changes operational connection fields from strings to structured messages', while the wire contract, the grpcurl example, and the final section require revision 11, matching the compiled `pkg/grpcprotocol.Version` bump from 10 to 11 in the frozen diff. The PR description also claims 'Service protocol 7 → 8', contradicting the code (spec provenance; standards provenance for line 158). A revision-8 client following the stale sentence could be wrongly treated as compatible. State revision 11 consistently in the doc and correct the PR description/release notes.

internal/application/events/connection_config_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/formancehq/ledger/v3/internal/domain/connectionconfig"
+	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
+)
+
```
NumaryBot, line 11: **🔵 [suggestion] Duplicated connectionTestURL helper in three test files**

The identical `connectionTestURL` helper is copied into `internal/application/events/connection_config_test.go`, `internal/application/events/connection_config_external_test.go` and `internal/application/mirror/connection_config_test.go` (all verified identical at the frozen SHA). Diverging copies would make packages verify different parsing semantics. Extract it into a shared test-support package.

internal/domain/connectionconfig/config.go
```diff
@@ -0,0 +1,579 @@
...
+func space(ch byte) bool {
+	return ch == ' ' || ch == '\t' || ch == '\n' || ch == '\r' || ch == '\v' || ch == '\f'
+}
+
+// RenderURL constructs the driver URL exclusively from normalized components.
+func RenderURL(config *commonpb.ConnectionURL) string {
+	if config == nil {
+		return ""
+	}
+	u := &url.URL{Scheme: config.GetScheme(), Host: renderAddress(config.GetAddress())}
```
NumaryBot, line 425: **🔵 [suggestion] Render helpers panic instead of returning errors**

`RenderURL` panics on an invalid escaped path and `optionValue` panics in its default case; both run at driver boundaries (`manager.createSink`, `buildPgxPoolConfig`, `mirror.createSource`) over configurations read from the attribute store. Corrupt persisted state would crash the process instead of failing sink startup. Return errors to the callers.

cmd/ledgerctl/events/list.go
NumaryBot: **🟡 [minor] Multi-line protojson output breaks CLI table rows**

`protojson.Format` emits indented multi-line JSON, and this diff embeds it directly into `pterm.DefaultTable` cells ("Connection", "Endpoint") and single-line `pterm.Printf` output in `list.go`, `events/add_sink.go`, and `ledgers/display.go`. The embedded newlines misalign or break the terminal table and connection display introduced by this change. Render a compact single-line representation instead (e.g. `protojson.MarshalOptions{}.Marshal`) or join the formatted lines.

NumaryBot (commented):
Findings

docs/technical/architecture/subsystems/api/protocol-compatibility.md
gfyrag: Fixed in HEAD bd1a6cf7e. Removed the stale "Revision 8 changes operational connection fields" sentence; it was an artifact of an earlier draft when the bump was called revision 8. The section now refers readers to the revision-specific subsections below, which document each numbered bump.

cmd/ledgerctl/events/list.go
gfyrag: Fixed in HEAD bd1a6cf7e. Added a protoCompact helper that uses protojson.MarshalOptions{Multiline: false} and replaced all three protojson.Format calls (NATS connection, ClickHouse connection, HTTP endpoint) in both list.go and add_sink.go with it. Table cells and printf lines now emit compact single-line JSON.

internal/application/events/connection_config_test.go
gfyrag: Agreed, noted for a follow-up. The test helper is small and the three files exercise different packages (events, mirror), so a shared test utility is the right destination. Left as a follow-up to keep this PR's scope contained.

internal/domain/connectionconfig/config.go
gfyrag: These are invariant guards: the normalizer validates and rejects malformed input on write, so invalid escaped path or unrecognized option kind should never appear in a persisted ConnectionURL. Changing the signature of RenderURL to return (string, error) would propagate into every caller (manager, pgxpool, mirror). Per AGENTS.md rule 7, the panic is the correct signal for a "should not happen" branch — it will surface immediately under test rather than producing a silently wrong connection string. Noted as a future improvement to convert these to assert.Unreachable once that pattern is established across the codebase.

docs/technical/architecture/subsystems/api/protocol-compatibility.md
```diff
@@ -185,3 +185,15 @@ cannot enter unary or streaming business handlers, while the matching revision
...
 forwarding so the gate cannot make the repository's own clients incompatible.
+
+Structured connection input/output messages require revision 11 on top of the
+revision 10 gate: existing connection configuration fields changed to structured message types.
+
+
+## Structured connection input/output messages (revision 11)
+
+Revision 11 is required when the client uses structured connection configuration
+messages (SinkConfigInput, MirrorSourceConfigInput) for write requests and reads
```
NumaryBot, line 196: **🟡 [minor] PR description and doc disagree on the protocol bump**

The diff bumps `pkg/grpcprotocol.Version` to "11" and this doc records revision 11 for structured connection messages (verified), but the PR description motivates "Service protocol 7 → 8". Correct the description so the recorded compatibility assessment matches the shipped revision.

internal/application/events/connection_config_test.go
NumaryBot, line 12: **🔵 [suggestion] Identical `connectionTestURL` helper duplicated in three packages**

The same `connectionTestURL` helper is copy-pasted in `internal/application/events/connection_config_test.go`, `connection_config_external_test.go` and `internal/application/mirror/connection_config_test.go`. Extract one shared test helper and call it from all three.

cmd/ledgerctl/events/proto_compact.go
```diff
@@ -0,0 +1,21 @@
...
+// Use this in table cells and printf output instead of protojson.Format, which
+// emits indented multi-line JSON that breaks pterm table layout.
+func protoCompact(m proto.Message) string {
+	if m == nil || !m.ProtoReflect().IsValid() {
+		return ""
+	}
+	b, err := protojson.MarshalOptions{Multiline: false, EmitUnpopulated: false}.Marshal(m)
+	if err != nil {
+		return "[error]"
+	}
```
NumaryBot, line 18: **🔵 [suggestion] `protoCompact` masks marshal failures as a literal `[error]`**

On `protojson` failure the CLI silently prints `[error]` with no justification comment, contrary to the repo rule that errors are handled explicitly. Propagate or log the error, or add a justification comment for the placeholder.

internal/domain/connectionconfig/sink_nats.go
```diff
@@ -0,0 +1,48 @@
...
+		// than reinterpreted as a hostname.
+		if !strings.HasSuffix(raw, "://") {
+			raw = strings.TrimSuffix(raw, "/")
+		}
+		server, err := parseURL(raw, "nats")
+		if err != nil {
+			return nil, err
+		}
+		if len(cfg.GetServers()) == 0 && (server.GetScheme() == "ws" || server.GetScheme() == "wss") {
+			defaultScheme = "ws"
```
NumaryBot, line 38: **🟡 [minor] Implicit scheme for later NATS servers downgrades wss to ws**

When the first parsed server has scheme `wss`, `defaultScheme` is set to `ws`, so a list like `wss://first:443,second:80` normalizes the scheme-less `second` endpoint to `ws://second:80`, connecting to it over plaintext websockets while the operator specified TLS. Suggest reusing the first server's scheme verbatim (or rejecting implicit schemes that would cross the wss/ws boundary) so later servers do not silently lose TLS.

NumaryBot (commented):
Findings

PR #1981, opened 2026-09-10
fix(http): preserve exact metadata integers (EN-2014)

gfyrag:
## What changed

HTTP metadata preserves exact integers through `Apply`, including `9007199254740993`, its negative counterpart, and the int64 boundaries. Fractional and out-of-range numbers are rejected before submission. This covers direct metadata writes, transaction/account metadata at creation, reversals, and their bulk decoders.

## Why

[EN-2014](https://formance-team.atlassian.net/browse/EN-2014), under Ledger v3.0 epic EN-867. Decoding metadata into `map[string]any` first converted JSON numbers to float64, irreversibly rounding supported integers before protobuf validation.

## Product / operational motivation

Clients must be able to submit supported metadata integers without silently changing their business data. Router regressions capture native integer values at `Apply`; the durable contract is documented in `docs/technical/architecture/subsystems/api/http-api.md#metadata-number-decoding` and OpenAPI.

## Technical decision

Use explicit number-preserving JSON helpers only in consumers of typed metadata, including custom `UnmarshalJSON` methods. Convert decimal tokens exactly, with work bounded by input length rather than exponent magnitude. Integral spellings such as `1.0` and `1e3` remain accepted. The ordinary shared JSON decoders keep their existing semantics; configuring only the outer HTTP decoder would miss nested custom decoders.

## Risk

**MEDIUM** — several metadata input paths share the conversion, covered by router, bulk, and custom-codec regressions. No new dependencies.

## Validation

- Fail-before: router cases on the original decoder reproduce rounding on account metadata, creation and known-length reversal, including incorrect acceptance of rounded fractions and overflow. Assertions compare native uint64/int64 values, never float64.
- Exact base: `404ac87a4291701bdca5ba9fed4997fb71412933` (`release/v3.0`).
- `bash scripts/agent-check`: PASS, including a second baseline after normalization.
- `go test -race ./internal/adapter/http ./internal/adapter/json ./internal/proto/commonpb ./internal/proto/servicepb -count=1`: PASS.
- Root and operator lint: 0 issues; deterministic generation leaves no unrelated diff.
- `AI_REVIEW_BASE_SHA=404ac87a4291701bdca5ba9fed4997fb71412933 bash scripts/agent-check-pr`: normalization, both lints, baseline and 102 unit-test packages PASS; stops at `tests/antithesis/TestRunModelTestRequiresVerifiedOutcome` (details below).
[...]

PR #1983, opened 2026-09-10
fix(test): synchronize model runner fixture deadlines

gfyrag:
## What changed

Make the local Antithesis model-runner fixture use a controlled clock tied to driver readiness. Preserve all six scenarios and the runner's real liveness checks, and add supervision that cancels the process group and collects driver/server logs when a fixture peer fails.

## Why

The fixture's two-second wall-clock window can expire before the driver reaches its intended state or before the monitor observes its exit. This produces intermittent `NO VERIFIED MODEL OUTCOMES` instead of `DRIVER EXITED EARLY`, and can interrupt the positive/blocked scenarios before they write evidence. The early-exit failure was reproduced on the exact base of #1981, independently of its metadata changes.

Follow-up to the fixture introduced by #1941 (EN-1975), discovered during #1981 validation. This PR does not change the production runner or relax the verified-outcome guard.

N/A — local test-fixture reliability correction.

## Technical decision

N/A

## Risk

**LOW** — test fixture and contributor documentation only.

## Validation

- [x] `AI_REVIEW_BASE_SHA=f76499637a5af0d92a2cfabbdacf8247256fa4bb bash scripts/agent-check-pr` — generation/tidy, root and operator lint, canonical `agent-check`.
- [x] `go test -race ./tests/antithesis -count=10 -v` — all six scenarios and three fixture regressions pass on each repetition.
- [x] Isolated mutations: removing the outcome gate fails the four negative cases while both controls pass; bypassing readiness fails the controlled-clock regression; removing early-exit detection reaches the watchdog.
- The initial mutation batch hit infrastructure timeouts; only the focused reruns above are used as mutation evidence.

None. Runner, Ledger binaries, APIs and persisted state are unchanged.

## Review focus

Clock advancement must require driver readiness while early-exit continues to exercise the real process-liveness branch. Runner completion must drain pending fixture events before deciding success; infrastructure errors must stay distinct from expected model findings.

## Known concerns

The controlled diagnostic reproduced a sufficient missed-exit ordering, but no naturally failing scheduling trace was captured. The separate 15-second infrastructure watchdog remains in place: a later run of the old fixture reached it during fake build/server startup, before monitoring. This PR does not claim to fix that distinct timeout signature. Local execution uses macOS; Linux CI remains the cross-platform check.

### DIFF
--- a/docs/technical/contributing/testing.md
+++ b/docs/technical/contributing/testing.md
@@ -517,6 +517,16 @@ after a definitive server outcome reaches model validation. Driver liveness,
 assertion registration, and ledger-setup assertions do not satisfy that gate.
 
 It also requires every coverage sonde to have been satisfied (see below).
+
+The local runner's shell-fixture tests use a logical clock. Time remains before
+its deadline until the fake driver has written the scenario's assertions and
+installed its shutdown handler; the runner then gets one live monitor iteration
+before expiration. The early-exit case holds time before the deadline so the
+runner must detect actual process termination. Fixture events and clock replies
+use blocking pipes, with process-group cancellation and collected server/driver
+logs on infrastructure failure. The outer timeout is a watchdog, never evidence
+that the intended model state was reached.
+
 Common tunables (full list in the script header):
 
 | Variable | Meaning |
--- a/tests/antithesis/run_model_test_test.go
+++ b/tests/antithesis/run_model_test_test.go
@@ -183,25 +183,24 @@ esac
 		fixtureErr = errors.Join(fixtureErr, fmt.Errorf("expected one runner workdir, got %d", len(workDirs)))
 	}
 	var driverLog string
+	var collectedLogs strings.Builder
 	for _, workDir := range workDirs {
-		var combinedSb187 strings.Builder
 		for _, name := range []string{"driver.log", "server-0.log"} {
 			data, readErr := os.ReadFile(filepath.Join(workDir, name))
 			if readErr != nil {
-				combinedSb187.WriteString(fmt.Sprintf("\n%s: %v\n", name, readErr))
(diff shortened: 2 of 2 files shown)


gfyrag:
@NumaryBot @shipfox-ai Please review the new HEAD `748efff7fc71bf6d3f5a00d5d3d8b9981418cbcf`.

Fixed the CI Dirty failure with one stable log builder and addressed both Shipfox comments (builder name and pipe-close justification). Canonical agent-check-pr, a second lint pass without auto-fix, and the focused race tests pass. Previous approvals covered `b3ead0f` and are not being counted for this candidate.

gfyrag:
Checked the new review against HEAD `748efff7fc71bf6d3f5a00d5d3d8b9981418cbcf`: item 2 conflates `wait_healthy` (lines 347–355) with `wait_leader` (lines 307 onward). `wait_leader` uses `seq`/`grep` and no `date`; `wait_healthy` is unreachable with this fixture's `NODES=1`, `RESTORE=0`. The first two clock queries really are deadline/restart initialization at lines 491/494, so the existing comment is accurate. Item 1 is an optional maintainability suggestion; runner/fixture error separation remains covered by the dedicated tests. No code change is needed for either non-blocking comment.

Azorlogh (dismissed):

gfyrag:
@NumaryBot @shipfox-ai Please review the rebased HEAD `2f1b0bfa4870166811d5c512ff316cddc2c6ccbb`.

Rebased onto release/v3.0 at `9a6fa7d0308e2d33d57fed63c5d1b7f0411194e0`. Resolved the documentation conflict by preserving both the new coverage-sonde requirement and the fixture-clock explanation. Range-diff confirms no changes to the PR code. Exact-base agent-check-pr and focused race tests pass. Earlier approvals are not counted for this new HEAD.

paul-nicolas (approved):

PR #2002, opened 2026-09-11
fix(bootstrap): join node after startup cancellation (EN-1996)

paul-nicolas:
## What changed

Cancelled node startup now waits for readiness or completion, explicitly stops the node, and joins `Run` before returning the startup error. The run context stays live through task drain. The production Fx hook is extracted into a testable helper, with deterministic lifecycle regressions and updated consensus documentation.

## Why

[EN-1996](https://formance-team.atlassian.net/browse/EN-1996): the failed `OnStart` path previously cancelled only the run context, leaving idle tasks alive. Fx does not call `OnStop` for a hook whose startup failed.

Failed startup must retain ownership of node work until it has terminated. The observable requirement is that the failed node hook cannot return while `Run` is still initializing or draining.

Durable contract: `docs/technical/architecture/subsystems/consensus/raft-consensus.md`, “Node shutdown.” Hook and real-node regressions cover this ownership boundary.

## Technical decision

Wait for readiness or completion before calling `Stop`, because `Run` initializes the node's completion state asynchronously. Use a cleanup context independent of the startup deadline, join `Run`, then cancel its context. This keeps the fix in bootstrap without changing Node lifecycle semantics.

Calling `Stop` immediately risks racing initialization; cancellation alone does not terminate idle tasks; imposing another cleanup timeout would not establish a join. Normal shutdown and fatal `Run` error behavior are preserved.

## Risk

**MEDIUM** — changes startup lifecycle ordering. Cleanup deliberately may outlast the startup deadline; deterministic tests cover initialization, readiness/cancellation overlap, drain, and Fx rollback ordering.

## Validation

- [x] `bash scripts/agent-check` and `AI_REVIEW_BASE_SHA=25a2249b71d037d11bf02174be01b3735375fc49 bash scripts/agent-check-pr`
- [x] Dirty: `just pre-commit`, including generation, consistency, and lint
- [x] Bootstrap and node race suites; the new hook regression failed before the fix and passed afterward
- [x] `just test-coverage`, S3 restore lifecycle test, and operator tests
- [x] Full-feature business and cluster E2E coverage, scenario coverage, and workload race tests
- [x] `just test-model-cluster 180`: four restart cycles, no findings
- [x] Schemathesis with `MAX_EXAMPLES=10 SCHEMATHESIS_WORKERS=1`
- [x] Standards and Spec reviews: no findings

Final checks passed in the pinned Nix environment. E2E required the documented OrbStack socket setting. One business rerun hit a normal-shutdown timeout under concurrent E2E load; the final sequential business and cluster runs passed without changing assertions or timeouts.

The failed node hook owns termination and joining. Fx's outer `App.Start` can still return its deadline error before hook cleanup finishes, and an expired Fx rollback context can skip dependency stop hooks. No persisted-state, API, or wire-format changes.

## Review focus

Readiness as the synchronization boundary before `Stop`, and preservation of the run context until cleanup completes.

## Known concerns

A stalled initialization or task drain can retain the failed hook and its resources beyond the startup deadline. This is the approved tradeoff for guaranteeing the join.

[EN-1996]: https://formance-team.atlassian.net/browse/EN-1996?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Bugfix evidence

DISCOVERY: NO_EXISTING_WORK

The initial EN-1996 search found no existing implementation of the failed-startup hook cleanup. Related shutdown work did not establish this hook's ownership of initialization and task drain.

BEFORE_FIX: BUG_REPRODUCED

The deterministic hook regression failed against the original cancellation-only implementation: startup returned while initialization was pending, cancelled the run context, and did not call Stop. The real-node regression also confirms that cancellation alone leaves idle tasks alive.

AFTER_FIX: PASS

The hook regression passes with readiness/completion synchronization, explicit Stop, and Run joining. The selected race suites pass after rebasing onto `25a2249b71d037d11bf02174be01b3735375fc49`.

BASELINE_CLASSIFICATION: BASELINE_FAILURE

The new base's metadata integer HTTP test still returned `[]*commonpb.Log` from its mock after Apply changed to `*domain.ApplyResult`. This PR adapts that test mock to the current return type, preserving its assertions and returned logs. No production behavior changes are needed for this compatibility fix.

The full CI suite and Dirty checks passed after the rebase and mock adaptation. An initial unit run hit readiness/deadline timeouts in two unchanged script-test packages under concurrent machine load; both isolated package reruns and the final full unit coverage run passed without modifying tests or timeouts.


gfyrag (approved):
Reviewed at 1c7dbb624b6cd6396956ee2b7f36e1cbcdfc5568. The cancelled-startup cleanup now explicitly stops and joins Node.Run while preserving the run context through drain; focused race tests, mutation sensitivity, agent-check, and exact-base agent-check-pr passed.

PR #2003, opened 2026-09-12
fix(api): roll back atomic bulks when a worker panics

SashaMIT:
## Summary

`Bulker.run` submits each element to a pond worker pool. pond recovers panics and does not propagate them.

`hasError` is only set on the normal error return path. When `processElement` panics, no result is emitted and `hasError` stays false. `Bulker.Run` then commits an `atomic=true` bulk and returns success with a short results array.

The caller asked for all-or-nothing and got a partial commit reported as success.

This recovers the panic in the submit closure, sets `hasError`, and emits an error result so atomic bulks roll back.

## Test plan

- [x] `go test ./internal/api/bulking/ -count=1` green
- [x] Revert-tested: without the recover, the new spec fails (unexpected Commit, missing Rollback)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/api/bulking/bulker.go
+++ b/internal/api/bulking/bulker.go
@@ -48,6 +48,18 @@ func (b *Bulker) run(ctx context.Context, ctrl ledgercontroller.Controller, sche
 				trace.WithAttributes(attribute.Int("index", itemIndex)),
 			)
 			defer span.End()
+			// pond recovers worker panics and does not propagate them. Without
+			// this, hasError stays false and an atomic bulk commits survivors.
+			defer func() {
+				if recovered := recover(); recovered != nil {
+					hasError.Store(true)
+					err := fmt.Errorf("bulk element panicked: %v", recovered)
+					observe.RecordError(ctx, err)
+					result <- BulkElementResult{
+						Error: err,
+					}
+				}
+			}()
 
 			select {
 			case <-ctx.Done():
--- a/internal/api/bulking/bulker_test.go
+++ b/internal/api/bulking/bulker_test.go
@@ -1,6 +1,7 @@
 package bulking
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"math/big"
@@ -433,4 +434,61 @@ func TestBulk(t *testing.T) {
 			require.NoError(t, bulker.Run(ctx, bulk, results, testCase.options))
 		})
 	}
+}
+
+func TestBulkAtomicPanicRollsBack(t *testing.T) {
+	t.Parallel()
+
+	ctx := logging.TestingContext()
+	ctrl := gomock.NewController(t)
+	ledgerController := NewLedgerController(ctrl)
+
+	ledgerController.EXPECT().
+		BeginTX(gomock.Any(), nil).
+		Return(ledgerController, &bun.Tx{}, nil)
+	ledgerController.EXPECT().
+		CreateTransaction(gomock.Any(), gomock.Any()).
+		DoAndReturn(func(context.Context, ledgercontroller.Parameters[ledgercontroller.CreateTransaction]) (*ledger.Log, *ledger.CreatedTransaction, bool, error) {
+			panic("simulated worker panic")
+		})
+	ledgerController.EXPECT().
+		Rollback(gomock.Any()).
+		Return(nil)
+
+	bulker := NewBulker(ledgerController)
+	bulk := make(Bulk, 2)
+	results := make(chan BulkElementResult, 2)
+	now := time.Now()
+	bulk <- BulkElement{
+		Action: ActionCreateTransaction,
+		Data: TransactionRequest{
+			Postings: []ledger.Posting{{
+				Source:      "world",
+				Destination: "bank",
+				Amount:      big.NewInt(100),
+				Asset:       "USD/2",
+			}},
+			Timestamp: now,
+		},
+	}
+	bulk <- BulkElement{
+		Action: ActionAddMetadata,
+		Data: AddMetadataRequest{
+			TargetID:   json.RawMessage(`"world"`),
+			TargetType: "ACCOUNT",
+			Metadata: metadata.Metadata{
+				"foo": "bar",
+			},
+		},
+	}
+	close(bulk)
+
+	require.NoError(t, bulker.Run(ctx, bulk, results, BulkingOptions{Atomic: true}))
+
+	var got []BulkElementResult
+	for result := range results {
+		got = append(got, result)
+	}
+	require.GreaterOrEqual(t, len(got), 1)
+	require.ErrorContains(t, got[0].Error, "bulk element panicked")
 }


internal/api/bulking/bulker.go
```diff
@@ -48,6 +48,18 @@ func (b *Bulker) run(ctx context.Context, ctrl ledgercontroller.Controller, sche
...
 			)
 			defer span.End()
+			// pond recovers worker panics and does not propagate them. Without
+			// this, hasError stays false and an atomic bulk commits survivors.
+			defer func() {
+				if recovered := recover(); recovered != nil {
+					hasError.Store(true)
+					err := fmt.Errorf("bulk element panicked: %v", recovered)
+					observe.RecordError(ctx, err)
+					result <- BulkElementResult{
```
flemzord, line 58: [P1] Make panic-result delivery cancellation-aware

With a streaming request, `result` is unbuffered and the JSON/text stream receiver exits on `ctx.Done()`. If the client disconnects while an element is running and that element then panics, this send has no receiver: the worker never finishes, `wp.StopAndWait()` blocks, and `Run` never reaches its explicit atomic rollback. I reproduced this with the real JSON streaming handler and a mock controller: the base completes, whereas this head only reaches rollback after an artificial receiver drains the channel. Please select between sending the result and cancellation, and add a disconnect-then-panic test that requires `Run` to terminate without an extra receiver.

internal/api/bulking/bulker.go
```diff
@@ -48,6 +48,18 @@ func (b *Bulker) run(ctx context.Context, ctrl ledgercontroller.Controller, sche
 				trace.WithAttributes(attribute.Int("index", itemIndex)),
 			)
 			defer span.End()
+			// pond recovers worker panics and does not propagate them. Without
+			// this, hasError stays false and an atomic bulk commits survivors.
+			defer func() {
+				if recovered := recover(); recovered != nil {
+					hasError.Store(true)
```
flemzord, line 55: [P1] Blocking behavior change under the maintainer's compatibility rule

The maintainer's acceptance rule for this review is that no existing observable behavior may change, including intentional fixes. For a sequential success / panic / success bulk, this flag changes atomic mode from committing survivors to rolling back; without `continueOnFailure`, it also prevents the last operation from executing. This applies to non-atomic bulks too: with default options, the last mutation previously ran and now does not. With continuation enabled it still runs, but an `ERROR/INTERNAL` result replaces the omitted panic result; with no other error, HTTP 200 becomes 400. These differences were confirmed against the base with controller mocks and the JSON response writer. The atomic rollback is the intended correction, not intrinsically a bug, but these observable changes block acceptance under that rule. Please preserve existing behavior on existing request paths, or obtain an explicit maintainer exception for these semantics before accepting this change.

internal/api/bulking/bulker.go
```diff
@@ -48,6 +48,18 @@ func (b *Bulker) run(ctx context.Context, ctrl ledgercontroller.Controller, sche
...
+					err := fmt.Errorf("bulk element panicked: %v", recovered)
```
flemzord, line 56: [P2] Keep the recovered panic value out of the public error

This error is passed to `writeJSONResponse`, which copies `result.Error.Error()` verbatim into `errorDescription`. A controller panic containing `private-marker` now produces `bulk element panicked: private-marker` in the HTTP response; on the base the panic value was handled internally by Pond and no result exposed it. This introduces a public diagnostic payload for arbitrary panic values. Please record the detailed panic internally but send a generic internal-error description to the client, with a response test asserting that the panic marker is absent. No production secret exposure is claimed here; the newly public payload itself is reproduced and is also a behavior change under the maintainer's compatibility rule.

flemzord (commented):
Reviewed the current head against 7d20326887509f06e4ae1c71eaf517cce20b2f7d. The inline findings distinguish a reproduced streaming regression, the maintainer-specific no-behavior-change acceptance constraint, and newly public panic details. Targeted bulking/API v2 tests and race-enabled bulking probes passed; the diagnostic probes reproduce the differences described below. No approval is being submitted.

flemzord (changes requested):
Requesting changes on fcaf3ceb6fb18d42eef6d3b5f675a19f9b54ec21 for the existing inline findings:

- [Streaming termination regression](https://github.com/formancehq/ledger/pull/2003#discussion_r4005868194): after a client disconnects and the in-flight element panics, the base completes, but the new unbuffered error send blocks the worker and prevents Run from reaching its explicit rollback. Reproduced with the real JSON streaming handler and a mock controller.
- [Maintainer compatibility constraint](https://github.com/formancehq/ledger/pull/2003#discussion_r4005868206): for success / panic / success, atomic mode changes from committing survivors to rollback; default non-atomic mode stops executing the last operation; HTTP 200 becomes 400 with an additional error result when no other error occurs. The intended atomicity correction is not intrinsically a bug, but these changes block acceptance under the maintainer's explicit rule that existing observable behavior must not change, including intentional fixes.
- [Public panic details](https://github.com/formancehq/ledger/pull/2003#discussion_r4005868215): panic text previously handled internally is now copied into the HTTP error description. Keep detailed diagnostics internal.

The targeted tests pass, but the before/after probes confirm these issues. Please address the regression and public diagnostic payload, and preserve existing semantics or obtain an explicit maintainer exception for the compatibility changes. The inline comments are not duplicated here.

internal/api/bulking/bulker.go
NumaryBot: 🔴 **[blocker] Make panic-result delivery cancellation-aware**

If a streaming client disconnects while an element is executing and that element then panics, both streaming handlers stop receiving results on context cancellation. This unconditional send to their unbuffered channel then blocks forever, preventing `wp.StopAndWait()` from completing and `Run` from reaching atomic rollback. Select between result delivery and `ctx.Done()`, and add a disconnect-then-panic regression test that requires completion without draining the result channel.

internal/api/bulking/bulker.go
NumaryBot: 🟠 **[major] Keep the recovered panic value out of the public error**

When an element panics with internal diagnostic data, this formatting embeds the entire panic value in the result error, which `writeJSONResponse` copies verbatim into the public `errorDescription`. Previously Pond handled that value internally. Keep the detailed panic in internal telemetry but return a generic internal-error description, with a response test ensuring a private panic marker is absent.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2003#issuecomment-5646371975

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2003#issuecomment-5646371975

internal/api/bulking/bulker.go
SashaMIT: Done. Panic send now selects between the result channel and ctx.Done. TestBulkAtomicPanicDoesNotBlockOnCanceledContext cancels, then panics, on an unbuffered results channel and requires Run to return without an extra receiver.

internal/api/bulking/bulker.go
SashaMIT: Agreed on the non-atomic path. Recover no longer sets hasError and no longer sends a panic result, so later elements still run and HTTP stays 200 when nothing else failed.

Atomic rollback is the intended correction. Asking for an exception on atomic=true only. If you want that path to keep committing survivors, say so and I will drop the atomic hasError set.

internal/api/bulking/bulker.go
SashaMIT: Done. The recovered value is recorded on the span only. The public result is a generic internal error. TestBulkAtomicPanicRollsBack asserts the panic marker is absent.

SashaMIT:
Pushed 6fcb57ae for the three review notes.

Cancel-aware send is in, with a disconnect-then-panic test on an unbuffered channel. Public results are a generic internal error; the panic value stays on the span. Non-atomic paths keep the old omitted-result shape so later elements still run and HTTP stays 200 when nothing else failed.

Atomic rollback is still the intended correction (pond recovers and would otherwise commit survivors). Asking for an exception on atomic=true only.

PR #2004, opened 2026-09-12
fix(machine): expose structured accounts list on ErrInsufficientFund (#1242)

Tyagiquamar:
### Summary
Fixes #1242. When a transaction fails due to insufficient funds, the error message lists the affected accounts in string format, requiring consumers to parse raw strings to extract account IDs.

### Fix
- Add structured accounts []string field and Accounts() []string accessor to ErrInsufficientFund in internal/machine/errors.go.
- Pass lstAccounts in Funding.Take() when constructing NewErrInsufficientFundWithAccounts.
- Add unit test TestFundingTakeInsufficientFunds in internal/machine/funding_test.go.

### DIFF
--- a/internal/machine/errors.go
+++ b/internal/machine/errors.go
@@ -32,11 +32,16 @@ func NewErrInvalidScript(f string, args ...any) *ErrInvalidScript {
 }
 
 type ErrInsufficientFund struct {
-	msg string
+	msg      string
+	accounts []string
 }
 
 func (e *ErrInsufficientFund) Error() string {
 	return e.msg
+}
+
+func (e *ErrInsufficientFund) Accounts() []string {
+	return e.accounts
 }
 
 func (e *ErrInsufficientFund) Is(err error) bool {
@@ -44,10 +49,15 @@ func (e *ErrInsufficientFund) Is(err error) bool {
 	return ok
 }
 
-func NewErrInsufficientFund(f string, args ...any) *ErrInsufficientFund {
+func NewErrInsufficientFundWithAccounts(accounts []string, f string, args ...any) *ErrInsufficientFund {
 	return &ErrInsufficientFund{
-		msg: fmt.Sprintf(f, args...),
+		msg:      fmt.Sprintf(f, args...),
+		accounts: accounts,
 	}
+}
+
+func NewErrInsufficientFund(f string, args ...any) *ErrInsufficientFund {
+	return NewErrInsufficientFundWithAccounts(nil, f, args...)
 }
 
 func IsInsufficientFundError(err error) bool {
--- a/internal/machine/funding.go
+++ b/internal/machine/funding.go
@@ -97,7 +97,7 @@ func (f Funding) Take(amount *MonetaryInt) (Funding, Funding, error) {
 			return fp.Account.String()
 		})
 
-		return Funding{}, Funding{}, NewErrInsufficientFund("account(s) %s had/have insufficient funds", strings.Join(lstAccounts, "|"))
+		return Funding{}, Funding{}, NewErrInsufficientFundWithAccounts(lstAccounts, "account(s) %s had/have insufficient funds", strings.Join(lstAccounts, "|"))
 	}
 	return result, remainder, nil
 }
--- a/internal/machine/vm/machine_test.go
+++ b/internal/machine/vm/machine_test.go
@@ -2484,3 +2484,40 @@ func (s *mockStore) GetBalances(_ context.Context, query BalanceQuery) (Balances
 func (s *mockStore) GetAccount(ctx context.Context, address string) (*ledger.Account, error) {
 	panic("not implemented")
 }
+
+func TestErrInsufficientFundAccountsPreserved(t *testing.T) {
+	tc := NewTestCase()
+	tc.compile(t, `send [GEM 15] (
+		source = {
+			@users:001
+			@payments:001
+		}
+		destination = @users:002
+	)`)
+	tc.setBalance("users:001", "GEM", 3)
+	tc.setBalance("payments:001", "GEM", 4)
+
+	m := NewMachine(*tc.program)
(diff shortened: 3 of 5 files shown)


internal/machine/funding.go
```diff
@@ -97,7 +97,7 @@ func (f Funding) Take(amount *MonetaryInt) (Funding, Funding, error) {
 			return fp.Account.String()
 		})
 
-		return Funding{}, Funding{}, NewErrInsufficientFund("account(s) %s had/have insufficient funds", strings.Join(lstAccounts, "|"))
+		return Funding{}, Funding{}, NewErrInsufficientFundWithAccounts(lstAccounts, "account(s) %s had/have insufficient funds", strings.Join(lstAccounts, "|"))
```
NumaryBot, line 100: 🔴 **[blocker] Preserve accounts when propagating the funding error**

During normal script execution, `OP_TAKE` catches this error and creates a new `ErrInsufficientFund` with the legacy constructor (`internal/machine/vm/machine.go`), discarding `lstAccounts`. Consequently, API consumers still receive an error whose `Accounts()` returns nil; propagate the original error or copy its accounts into the wrapper.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2004#issuecomment-5646836167

Tyagiquamar:
Updated in 59d962d: Updated OP_TAKE in internal/machine/vm/machine.go to check rrors.As(err, &errInsufficientFund) and propagate the original ErrInsufficientFund directly instead of wrapping it in a new empty error. Added TestErrInsufficientFundAccountsPreserved to verify Accounts() returns the structured account list on VM execution errors.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2004#issuecomment-5646836167

internal/machine/errors.go
```diff
@@ -32,24 +32,34 @@ func NewErrInvalidScript(f string, args ...any) *ErrInvalidScript {
...
-	msg string
+	msg      string
+	accounts []string
 }
 
 func (e *ErrInsufficientFund) Error() string {
 	return e.msg
 }
 
+func (e *ErrInsufficientFund) Accounts() []string {
```
flemzord, line 43: **Acceptance blocker: intentional internal behavior change.** For the existing insufficient-funds failure (send 15 GEM from accounts holding 3 and 4 GEM), `err.(interface{ Accounts() []string })` previously returned `ok == false`; it now returns `ok == true` and exposes both account names, including through VM execution. This was reproduced against the base and this head. The error text and HTTP payload remain unchanged, but the Go error contract is observably extended. This is not an intrinsic bug or a demonstrated regression: it conflicts with the maintainer's explicit requirement to accept no changes to existing behavior, including intentional changes. Please remove this contract extension from the behavior-preserving scope, or obtain an explicit change to that acceptance requirement before proceeding.

internal/machine/vm/machine_test.go
```diff
@@ -2484,3 +2484,40 @@ func (s *mockStore) GetBalances(_ context.Context, query BalanceQuery) (Balances
...
+			}(),
+		}
+	}
+	require.NoError(t, m.ResolveResources(context.Background(), store))
+	require.NoError(t, m.ResolveBalances(context.Background(), store))
+	err := m.Execute()
+	require.Error(t, err)
+	var errIns *machine.ErrInsufficientFund
+	require.True(t, errors.As(err, &errIns))
+	require.NotEmpty(t, errIns.Accounts())
```
flemzord, line 2520: **[P1] The API behavior requested by #1242 is still missing.** This assertion proves that accounts survive inside the VM, but no production caller reads `Accounts()`. With this same 15 GEM / 3+4 GEM scenario, API v2 still calls `api.BadRequest` in `controllers_transactions_create.go:53-54`, which serializes only the error code and message. Before and after this PR, the response is HTTP 400 without `details.accounts`, so the Java consumer still has to parse the message. This is a separate gap from the resolved VM-propagation finding. Moreover, `openapi/v2.yaml:2947-2948` and the generated Go error model still define `details` as a string. Please stop claiming this closes #1242 unless the HTTP contract and an API-level assertion are addressed. Completing that feature would itself change observable HTTP behavior, so it must first be reconciled with the maintainer's no-behavior-change requirement.

flemzord (changes requested):
Requesting changes on 59d962d82e30849f2c97bf1ed8196f86c2a6bbc8 for the two findings already posted inline:

- [Acceptance constraint](https://github.com/formancehq/ledger/pull/2004#discussion_r4005868427): with a 15 GEM transfer from accounts holding 3 and 4 GEM, the returned error now satisfies `interface{ Accounts() []string }` and exposes the accounts; the same assertion failed on the base. This intentional extension conflicts with the maintainer's explicit no-behavior-change requirement. It is a compatibility/acceptance constraint, not an intrinsic bug or a demonstrated regression.
- [Incomplete API feature](https://github.com/formancehq/ledger/pull/2004#discussion_r4005868442): the same request still returns HTTP 400 without `details.accounts`, exactly as before. No production caller reads the accessor, and the API/OpenAPI/generated client remain unchanged, so #1242 is not fulfilled.

Please reconcile the feature scope with the acceptance requirement before proceeding. Local race-enabled tests pass, but do not remove these blockers. Inline comments are referenced rather than duplicated.

Tyagiquamar:
Thanks, I reproduced both points.\n\nThe current PR extends the Go error contract, but it still does not implement #1242 requested details.accounts HTTP response. Completing #1242 would necessarily require an observable API/OpenAPI contract change.\n\nCould you confirm which direction you prefer?\n\n1. Allow that intentional contract extension, and I will implement details.accounts end to end with API/OpenAPI/generated-model coverage; or\n2. Keep strict behavior preservation, in which case I will defer or close this feature rather than expand the PR.

PR #2007, opened 2026-09-14
test(model): cover enforcement modes and skipped orders

paul-nicolas:
## What changed

Extend the independent oracle and concurrent gRPC model workload to cover both enforcement-mode setters and all five skippable reasons. Validate rollback, continuation, logs, ledger reads, and idempotent replay through deterministic and real-server tests.

Normalize equivalent mode/chart request forms when predicting idempotency. Record driver exit time and status so restart recovery cannot misreport a normal watchdog exit as early.

## Why

Covers the enforcement and skippable-order portions of [EN-1631](https://formance-team.atlassian.net/browse/EN-1631). Signing/key-lifecycle modeling remains a separate deliverable.

The model must detect discrepancies in committed Apply outcomes, including skipped mutations and mode-dependent failures. It must also distinguish an actual early driver exit from a normal exit while cluster recovery finishes. Observable coverage assertions, real RPC regressions, and synchronized harness tests establish these requirements. Durable documentation: `docs/technical/contributing/testing.md`.

## Technical decision

Use independent oracle state and per-order savepoints for skip behavior, preserving whole-batch rollback. Compare normalized business intent for equivalent wire forms without importing the production admission converter. A supervisor records driver termination during blocked restart recovery; early and spontaneous nonzero exits still fail the run.

## Risk

**LOW** — Changes are limited to tests, the model oracle/workload, the local test harness, and contributor documentation. No production service contract or persisted-state changes.

## Validation

- [x] `AI_REVIEW_BASE_SHA=9a6fa7d0308e2d33d57fed63c5d1b7f0411194e0 bash scripts/agent-check-pr`
- [x] `just pre-commit` and final generated-file/formatting consistency checks
- [x] `just test-coverage` and `just test-internal-coverage`
- [x] Operator tests and `just test-scenarios-coverage`
- [x] Business and cluster E2E coverage suites
- [x] `go -C tests/antithesis/workload test -race ./...`
- [x] `just test-model 60` and `just test-model-cluster 180` — four restart cycles, no findings
- [x] Schemathesis with CI settings — all 62 endpoints passed
- [x] Regression tests for cross-form RPC replay, skipped-log mutations, mode/search ordering, and driver exit classification

Run locally in the pinned Nix environment (Go 1.27.1). Final broad runs used four Go execution cores and the active Docker socket. Initial host readiness timeouts and socket configuration failures were resolved and the affected checks rerun successfully.

No production behavior changes. The test oracle now represents enforcement modes and skipped outcomes; local model-run reporting uses recorded exit timing/status.

## Review focus

Independent skip-context predictions, mode/chart idempotency equivalence, and supervisor cleanup/exit classification during restart recovery.

## Known concerns

Signing/key-lifecycle coverage is intentionally outside this PR; this does not complete that part of EN-1631.

[EN-1631]: https://formance-team.atlassian.net/browse/EN-1631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gfyrag (dismissed):

Azorlogh (changes requested):
Sorry just blocking temporarily to get more time to review

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/actions.go
```diff
@@ -109,6 +109,11 @@ func sourceAddress() string {
 // snapshot (a published GlobalState is never mutated — Apply forks first).
 func generateBulk(g oracle.GlobalState, ledgers []string) oracle.Bulk {
 	picks := pickLedgers(ledgers)
+	if random.RandomChoice([]uint8{0, 1, 2, 3, 4, 5, 6, 7}) == 0 {
+		if bulk := generateSkippableBulk(picks[0], g.Ledger(picks[0])); len(bulk.Requests) > 0 {
```
Azorlogh: `generateSkippableBulk` early-returns a fixed 2-order single-ledger bulk, so an opt-in only ever lands at position 0: never at 1..4, never twice in a bulk, never across ledgers, never composed with a transient or index shape.

`skippable_reasons` is a per-order flag on an ordinary request, so it belongs as a roll inside the slot loop below (like `rollChartOp`), attaching a whitelisted reason to whatever that slot already generated. The existing whole-bulk early returns are that way because their shape is intrinsically whole-bulk; this one isn't.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/apply_surface.go
```diff
@@ -0,0 +1,127 @@
...
+	var action *servicepb.LedgerAction
+	var reason commonpb.ErrorReason
+	switch random.RandomChoice([]uint8{0, 1, 2, 3, 4, 5}) {
+	case 0:
+		ref, _, ok := pickTxRef(ls)
+		if !ok {
+			return oracle.Bulk{}
+		}
+		action = &servicepb.LedgerAction{Data: &servicepb.LedgerAction_CreateTransaction{CreateTransaction: &servicepb.CreateTransactionPayload{
+			Reference: ref, Postings: []*commonpb.Posting{commonpb.NewPosting("world", poolAddress(), "USD", big.NewInt(1))},
```
Azorlogh: Near-verbatim copy of `duplicateReferenceTransaction` (actions.go:421) — same `pickTxRef`, same `world -> poolAddress()` posting, same amount — except the asset is `"USD"`, which isn't in `assets` (config.go:41). Case 4 likewise open-codes `generateRemoveAccountType`, case 2 a delete-metadata.

The function also bypasses `applyCreate` only to attach `SkippableReasons`. Give `applyCreate` that parameter and reuse the existing generators and most of this file goes away.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/coverage_test.go
```diff
@@ -161,6 +162,9 @@ func TestCoverageHits_DecidesEveryEntitySonde(t *testing.T) {
 
 	hits := coverageHits(accounts, filterMetaExists("k1"), nil, true, 1, false)
 	for _, msg := range coverageMessages() {
+		if slices.Contains(applyCoverageMessages(), msg) {
```
Azorlogh, line 165: Exempting the Apply sondes from this invariant is what let a dead sonde through: `invalidSkipCoverageMessage` is in `applyCoverageMessages()` but `noteApplyCoverage` never sets it in `hits` — I measured 0 evaluations there against 8303 from `validateFailure`.

An `applyCoverageHits(bulk, res) map[string]bool` mirroring `coverageHits`, with this test extended to it rather than skipping it, restores the guard.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/validate.go
```diff
@@ -80,6 +81,29 @@ func (c *Checker) crossCheckCommit(bulk oracle.Bulk, resp *servicepb.ApplyRespon
 	}
 
 	logs := resp.GetLogs()
+	if len(logs) != len(res.Orders) {
+		assert.Unreachable("singleton_driver_model: committed log count mismatch", internal.Details{"expected": len(res.Orders), "actual": len(logs)})
+		return
+	}
+	for i, order := range res.Orders {
+		entry := logs[i].GetPayload().GetApply()
+		if order.LogID != 0 && (entry.GetLog().GetId() != order.LogID || entry.GetLedgerName() != oracle.LedgerOf(bulk.Requests[i])) {
```
Azorlogh, line 90: These four commit-path checks (log identity, skip payload, enforcement mode, and `chartResponseMatches` further down) are mutation-green — I disabled all four and every test in the repo still passes.

The replay path and the ListLogs path each have a red-on-corruption test; this one doesn't. `skippable_integration_test.go` already has the right shape for it (call `crossCheckCommit`, assert the model advanced) — one corrupted response per check.

Azorlogh (changes requested):
The modelling is right — I checked the oracle against production rather than against the PR's own tests: the skippable whitelist matches the `(allowed_skippable_reasons)` annotations in `bucket.proto`, `ErrInvalidSkippableReason.Reason()` is VALIDATION, all five errors' `Metadata()` key sets match `predictSkippedLog` exactly, both wire forms of the three chart/mode actions produce identical `raftcmdpb.Order`s, and the ref-conflict-before-chart / already-reverted-before-chart orderings match. `candidateBases` is used properly: the mode is in `LedgerState.Fingerprint()`, `validateLedgerRead` requires chart+metadata+mode on the same base, and `enforcement_search_test.go` pins that a revert failure is only explainable once both the chart removal and the STRICT flip fold in. Over 400k generated bulks all 9 Apply sondes are reachable and ACCOUNT_NOT_MATCHING_TYPE still fires on 5.6% of bulks despite AUDIT being live ~49% of the time.

My concern is that the Apply surface landed as a second workload alongside the existing one rather than inside it. Four blocking items below; I have a dozen smaller ones (diagnostics, dead guards, stale comments, the untested per-order rollback) that I'll send once these are settled.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2007#issuecomment-5661795552

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/validate.go
```diff
@@ -438,6 +447,15 @@ func (c *Checker) validateFailure(maxTicket uint64, failedBulk oracle.Bulk, reqE
 	})
 
 	if matched {
+		invalidOptIn := false
+		if reason == domain.ErrReasonValidation {
+			for _, req := range failedBulk.Requests {
```
NumaryBot: 🟠 **[major] Require an invalid reason before crediting invalid-opt-in coverage**

When `emptyTransaction` receives the valid reference-conflict opt-in through `maybeAddSkippableReason`, admission rejects its missing postings with `VALIDATION`. This branch nevertheless credits the invalid-opt-in sonde merely because a skip list is present. A bulk containing an empty transaction and another valid opted-in order does the same. Consequently, local coverage can pass without exercising invalid-reason rejection, contrary to the [Apply coverage contract](docs/technical/contributing/testing.md#L566-L569). Check that an opt-in is actually disallowed for its action before crediting this sonde, and add a negative regression for a valid opt-in combined with an unrelated validation failure.

paul-nicolas: Fixed in 8b1377279. Invalid-opt-in coverage now requires a reason that is actually disallowed for its action, with a negative regression for valid opt-ins combined with unrelated validation failure.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2007#issuecomment-5661795552

Azorlogh (approved):

PR #2008, opened 2026-09-14
fix(checkpoints): handle deletion races with model coverage

paul-nicolas:
## What changed

Add checkpoint lifecycle and frozen-state validation to the model driver: concurrent create/delete, retention-cap and invalid-input errors, historical account/transaction reads and lists, varied non-firing schedule configuration, and reads against the predicted checkpoint ID while its create is in flight. Capture immutable oracle states when creation responses drain in global sequence order.

Make repeated model invocations bounded: setup now resets the schedule, frees one slot when needed, and creates a probe checkpoint whose assigned ID establishes the allocation frontier without scanning historical logs.

Fix the server race exposed by this coverage: deletion between checkpoint readiness checking and opening its read index could return gRPC `Unknown`. Missing read-index files return `NotFound` only when persisted lifecycle state proves the checkpoint was deleted; missing files for a live checkpoint and unrelated corruption remain errors. A lifecycle lookup failure now records both opening stages under a correlation ID and returns retryable `Unavailable`.

Make checkpoint cleanup order explicit: remove `readindex/` before `main/`, then remove the parent directory. A DAL filesystem test proves that `.ready` can outlive read-index database files during recursive deletion and that `main/` outlives the read index.

Also fix a pre-existing model-driver selection bug used by both `runRead` and checkpoint reads: ordered-map probes now wrap to the first entry when the random seek falls beyond the final key, so non-empty maps cannot be mistaken for empty ones.

## Why

[EN-1630](https://formance-team.atlassian.net/browse/EN-1630) requires historical reads to match the oracle state at the checkpoint sequence and deleted checkpoints to return `NotFound` once local cleanup completes; reads racing deletion may still return their original frozen data. The model reproduced the race on checkpoint 259: creation sequence 5432, deletion sequence 5546, followed by a failed `ListAccounts` read opening a missing Pebble database.

The additional predicted-ID probe covers the pre-commit window: before creation commits it may return `NotFound`; if creation wins the race, the response must match the model state immediately preceding the matching create transition.

Concurrent checkpoint creation and deletion must not turn supported historical reads into opaque server errors or allow live data to masquerade as a frozen snapshot. Repeated Antithesis invocations must also start in bounded time regardless of accumulated global log history.

Durable behavior and test ownership are documented in `tests/antithesis/README.md` and `docs/technical/architecture/subsystems/read-path/query-checkpoints.md`.

## Technical decision

Use the existing ordered response buffer to capture copy-on-write oracle snapshots and extend candidate-state traversal to observe the exact predecessor of an in-flight checkpoint creation. Pin metadata reads to one node and precede them with a linearizable barrier because those APIs read local state.

Use the persisted deletion check to classify missing files after deletion. Preserve raw storage and lifecycle errors only in the correlation-ID diagnostic log; return `Unavailable` when lifecycle state cannot be read.

Use several long non-firing cron intervals so schedule reads verify the stored value without allowing background writes to escape the oracle timeline.

## Risk

**MEDIUM**: changes checkpoint deletion ordering and the error classification of missing checkpoint read-index files during concurrent deletion. Deterministic filesystem-order tests, server regressions, exact-base race gates, and model coverage constrain the change. No persisted format or Raft command changes.

## Validation

Local validation ran in the repository-pinned Nix environment with `GOFLAGS=-p=2 GOMAXPROCS=4`. Docker Desktop and OrbStack were not used; Docker-backed suites run in GitHub CI.

- [x] `AI_REVIEW_BASE_SHA=928250fe9bba8d344a0053db7f62137082a89eeb bash scripts/agent-check-pr`
- [x] Focused `go test -race` for `internal/adapter/grpc`, `internal/storage/dal`, `internal/storage/readstore`, and `tests/oracle`
- [x] `go -C tests/antithesis/workload test -race ./bin/cmds/model/singleton_driver_model`
- [x] Root unit coverage and operator unit tests
- [x] Scenario coverage
- [x] `MAX_EXAMPLES=10 SCHEMATHESIS_WORKERS=1 just test-schemathesis`: 62/62 endpoints
- [x] Root/operator builds and Windows amd64/arm64 `ledgerctl` cross-builds
- [x] Current-head GitHub CI, including Docker-backed business/cluster E2E and model suites

Checkpoint reads whose files disappear during opening return `NotFound` when persisted state confirms deletion. If that lifecycle lookup fails, callers receive retryable `Unavailable` and operators can correlate the response with a server log containing both failed stages.

The model retains bounded frozen snapshots, including the last ten known deleted checkpoints, and seeds repeated invocations through a newly allocated probe ID. Deleted-ID probes accept only `NotFound` or data matching the original frozen snapshot, allowing replica lag and cleanup races without accepting live data. Reads racing a predicted create validate success against the matching creation predecessor.

Audit rebuild restores checkpoint registry rows but cannot reconstruct physical checkpoint files. Those IDs remain retryably unavailable until deletion; checkpoints created after rebuild resume frozen-read coverage.

## Review focus

Candidate-state transition capture for predicted IDs; deterministic read-index-before-main cleanup; correlation-ID diagnostics with retryable lifecycle failure; bounded repeated-run setup.

## Known concerns

Actual cron firings and business-list cursor continuation are outside this coverage. Creation uses singleton bulks; mixed business/create batches remain covered by oracle unit tests and would require per-order snapshot capture for end-to-end model coverage. The exclusive model template requires no unrelated automatic checkpoint creator.

[EN-1630]: https://formance-team.atlassian.net/browse/EN-1630


internal/adapter/grpc/server_bucket.go
```diff
@@ -370,6 +373,24 @@ func (impl *BucketServiceServerImpl) openCheckpointStores(ctx context.Context, c
...
+		// Deletion can remove files after the readiness check and main open.
+		// Reclassify only missing files backed by a committed deletion; storage
+		// loss on a still-live checkpoint and other corruption remain errors.
+		// Pebble's missing-directory error drops the filesystem cause, unlike
+		// its missing-manifest sentinel. Check the path again rather than matching
+		// error text; neither absence signal alone proves lifecycle deletion.
+		_, pathErr := os.Stat(readIndexPath)
+		missing := errors.Is(err, pebble.ErrDBDoesNotExist) || errors.Is(err, fs.ErrNotExist) || errors.Is(pathErr, fs.ErrNotExist)
+		if !pebble.IsCorruptionError(err) && missing {
+			deleted, lifecycleErr := impl.queryCheckpointDeleted(checkpointID)
```
NumaryBot: 🟠 **[major] Do not let the post-open stat mask unrelated errors**

If `OpenReadOnly` fails for a non-missing reason and deletion removes the directory before `os.Stat`, `pathErr` makes `missing` true and the original error is converted to `NotFound`. This can hide permission, I/O, or other non-Pebble-corruption failures during concurrent deletion; use the stat result only when the open error itself indicates absence, or otherwise preserve the original error.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2008#issuecomment-5661963695

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2008#issuecomment-5661963695

Azorlogh (commented):
Two good finds. The `Unknown` leak on the read-index open is real, and `queryCheckpointDeleted` is a stricter classifier than the `!queryCheckpointExists` one I used on the main-store arm in #1954.

`e9fda70d9` is the right call on the deleted-read probe. Routing deleted IDs through the same frozen comparison is correct: the registry row is removed in the committed batch before the filesystem unlink, so a replica can legally still serve the checkpoint. The regression guard survives — a non-NotFound error from a deleted checkpoint is still a finding.

Optional: you could split the server fix into its own PR. It's small and independently reviewable, and I'd like it ahead of #1954. The coverage half has more open questions. Happy either way if you'd rather keep it as one.

Note the coverage half collides with #2009 — both add `singleton_driver_model/checkpoints.go`. I'm dropping my checkpoint probe from #2009 in favour of yours, so build on that once it's rebased.

`TestOpenCheckpointStoresReadIndexDeletion` fabricates its residues by hand. Could you add a `dal` test that runs `DeleteQueryCheckpointFiles` through an FS recording unlink order? That would show `os.RemoveAll` actually produces them — `.ready` outliving the read-index database files, and `main` outliving `readindex`. Right now it's asserted only in the PR body.

One structural gap left on the coverage half: the driver can only read an ID the server already handed back, so it never probes one before its create commits. The oracle already predicts it — `applyCheckpoint` assigns `g.nextCheckpointID` before dispatch. The validation machinery already handles the answer too, since `checkpointNotFound` consults `candidateBases` for `!base.QueryCheckpointExists(id)`, which is true while the create is in flight. Worth firing reads at the predicted ID concurrently with the create. We don't know what a not-yet-existent checkpoint read returns, and that's the kind of path this driver exists to pin.

Unit tests pass locally on `e9fda70d9`: `singleton_driver_model`, `tests/oracle`, `internal/storage/readstore`, `internal/adapter/grpc`.

internal/adapter/grpc/server_bucket.go
```diff
@@ -370,12 +372,30 @@ func (impl *BucketServiceServerImpl) openCheckpointStores(ctx context.Context, c
...
 
+func (impl *BucketServiceServerImpl) checkpointReadIndexOpenError(checkpointID uint64, err error) error {
+	// Deletion can remove files after the readiness check and main open.
+	// Reclassify only missing files backed by a committed deletion; storage
+	// loss on a still-live checkpoint and other corruption remain errors.
+	missing := errors.Is(err, pebble.ErrDBDoesNotExist) || errors.Is(err, fs.ErrNotExist)
+	if !pebble.IsCorruptionError(err) && missing {
+		deleted, lifecycleErr := impl.queryCheckpointDeleted(checkpointID)
+		if lifecycleErr != nil {
+			return errors.Join(err, lifecycleErr)
```
Azorlogh: `errors.Join(err, lifecycleErr)` drops the `opening checkpoint read index` wrap, so the correlation-ID'd log won't say which stage failed. A failed lifecycle read also says nothing about the checkpoint — Unavailable seems more honest than falling through to the sanitizer's `Unknown`.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/checkpoint_setup.go
```diff
@@ -0,0 +1,87 @@
...
+					c.modelState = c.modelState.SeedQueryCheckpoints(ids, nextID)
+					return true
+				}
+			}
+		}
+	}
+	if isShutdownError(err) {
+		return false
+	}
+	assert.Unreachable("singleton_driver_model: checkpoint setup failed", internal.Details{"error": err.Error()})
```
Azorlogh, line 90: Only `isShutdownError` is excluded, so a transient Unavailable during startup election becomes a finding; this needs the `internal.IsTransient` check too.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/checkpoint_setup.go
```diff
@@ -0,0 +1,87 @@
...
+
+// checkpointBaseline scans backwards because the largest live ID need not be
+// the last allocated ID. A deleted last checkpoint still advances the counter.
+func checkpointBaseline(listed *clusterpb.ListQueryCheckpointsResponse, sequence uint64, read func(uint64) (*commonpb.Log, error)) ([]uint64, uint64, error) {
+	ids := make([]uint64, 0, len(listed.GetCheckpoints()))
+	for _, cp := range listed.GetCheckpoints() {
+		ids = append(ids, cp.GetCheckpointId())
+	}
+	nextID := uint64(1)
+	for ; sequence > 0; sequence-- {
```
Azorlogh: One `GetLog` per sequence back to 1 when no create is ever found — unbounded on the repeated-invocation case the README advertises.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/main.go
```diff
@@ -110,6 +110,25 @@ func main() {
 	}
 
 	checker := NewChecker(names, schemas)
+	checker.modelState = checker.modelState.WithQueryCheckpointLimit(uint64(envInt("MODEL_QUERY_CHECKPOINT_LIMIT", defaultModelCheckpointLimit)))
+	dialCtx, cancelDial := context.WithTimeout(ctx, 5*time.Second)
+	checkpointNodes, err := internal.DialPerNode(dialCtx)
+	cancelDial()
+	if err != nil || len(checkpointNodes) == 0 {
+		assert.Unreachable("singleton_driver_model: checkpoint node connections unavailable", internal.Details{"error": fmt.Sprint(err)})
```
Azorlogh: `DialPerNode` only ever returns `(conns, nil)` — the `err != nil` half is dead.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/main.go
```diff
@@ -151,16 +170,11 @@ func main() {
 	// index is live everywhere. Per-node conns are lazy, so dialing never fails on
```
Azorlogh: Stale: nothing is skipped now, the run returns.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/main.go
```diff
@@ -232,6 +250,9 @@ func runWorker(
 		c.mu.Unlock()
 
 		bulk := generateBulk(state, c.ledgerNames)
```
Azorlogh: `generateBulk` is computed then discarded 10% of the time; roll first.

tests/antithesis/README.md
```diff
@@ -30,6 +30,51 @@ two templates:
...
+entity already absent from the frozen snapshot. Concurrent deletion of a live
+checkpoint is accepted only when an eligible model state explains it. The existing
+transient-error contract covers replicas whose checkpoint read index is not ready yet.
+
+`MODEL_QUERY_CHECKPOINT_LIMIT` defaults to 10 and must match the server's
+`--query-checkpoint-limit`. Creation deliberately attempts to exceed the cap and
+requires `CHECKPOINT_LIMIT_REACHED`. Live snapshots are bounded by that cap;
+only the ten most recent deleted checkpoints with known frozen snapshots are
+retained for deletion probes. Schedule set/get/delete uses a valid 100-year interval so no scheduler-generated writes
+enter the model timeline. Restore cycles retain the frozen business snapshots
```
Azorlogh: `rebuild.go:430` says the opposite — "physical checkpoint files cannot be reconstructed from the audit, so a rebuilt checkpoint reads as Unavailable until an operator deletes it". Those snapshots go permanently Unavailable and stop contributing coverage.

tests/oracle/model.go
```diff
@@ -1887,3 +1915,105 @@ func (s *LedgerState) purgeZeroBalance(touched map[VolumeKey]bool) map[VolumeKey
...
+		}
+		id := g.nextCheckpointID
+		g.nextCheckpointID++
+		g.checkpoints = g.checkpoints.Set(strconv.FormatUint(id, 10), id)
+
+		return OrderResult{OK: true, CheckpointID: id}, true
+	case *servicepb.Request_DeleteQueryCheckpoint:
+		id := r.DeleteQueryCheckpoint.GetCheckpointId()
+		if id == 0 {
+			return OrderResult{Reason: domain.ErrReasonCheckpointIDRequired}, true
```
Azorlogh, line 1978: This and the invalid-cron branch at line 1986 are unreachable: `generateCheckpointBulk` only picks IDs from `QueryCheckpointIDs()` (all >= 1) and one constant cron.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/checkpoints.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/formancehq/ledger/v3/tests/oracle"
+)
+
+// The model template owns its cluster's checkpoint timeline. This must match
+// the server's query-checkpoint-limit configuration (10 by default).
+const defaultModelCheckpointLimit = 10
+
+// A 100-year interval exercises configuration without introducing scheduler
+// writes outside the model driver's observed Apply stream during a test run.
+const modelCheckpointCron = "@every 876000h"
```
Azorlogh: One constant means the schedule read only distinguishes it from empty; a couple of different non-firing crons would exercise the stored value.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/checkpoints.go
```diff
@@ -0,0 +1,118 @@
...
+			delete(c.checkpoints, id)
+			// Inherited checkpoints have no captured business state. Do not
+			// invent one for validating successful reads after deletion.
+			if !known {
+				noteCheckpointCoverage(checkpointDeleteCoverage)
+				continue
+			}
+			c.deletedCheckpointSnapshots[id] = snapshot
+			c.deletedCheckpoints = append(c.deletedCheckpoints, id)
+			if len(c.deletedCheckpoints) > defaultModelCheckpointLimit {
```
Azorlogh: `defaultModelCheckpointLimit` now does triple duty — model cap, deleted-ID ring, and snapshot retention.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/checkpoint_reads.go
```diff
@@ -0,0 +1,242 @@
...
+	defer cancel()
+	c.mu.Lock()
+	if c.paused {
+		c.mu.Unlock()
+		return
+	}
+	readID := c.registerRead()
+	var id uint64
+	var frozen oracle.GlobalState
+	deleted := len(c.deletedCheckpoints) > 0 && percentChance(25)
```
Azorlogh: With the special branch gone, `deleted` only picks which snapshot map to read and gates one sonde; the 25% split reads as leftover structure.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/reads.go
```diff
@@ -156,7 +156,10 @@ func pickAbsentAccount(g oracle.GlobalState, ledgers []string) (ledger, addr, as
 // the probe shares addr's address iff the account has state).
 func modelKnowsAccount(ls oracle.LedgerState, addr string) bool {
```
Azorlogh, line 157: Real pre-existing bug affecting `runRead` too, not just checkpoints — worth its own commit and a line in the description.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/main.go
```diff
@@ -110,6 +111,25 @@ func main() {
...
+	checker.modelState = checker.modelState.WithQueryCheckpointLimit(uint64(envInt("MODEL_QUERY_CHECKPOINT_LIMIT", defaultModelCheckpointLimit)))
+	dialCtx, cancelDial := context.WithTimeout(ctx, 5*time.Second)
+	checkpointNodes, _ := internal.DialPerNode(dialCtx)
+	cancelDial()
+	if len(checkpointNodes) == 0 {
+		assert.Unreachable("singleton_driver_model: checkpoint node connections unavailable", nil)
+		return
+	}
+	defer checkpointNodes.Close()
+	checkpointSetupNode, err := selectCheckpointSetupNode(ctx, checkpointNodes, names[0])
```
NumaryBot, line 123: 🟠 **[major] Treat transient setup-node probe failures as retryable**

If an election or temporary partition prevents every node from completing its two-second fence probe, `selectCheckpointSetupNode` returns an error while the parent workload context remains active. This branch then records an Antithesis invariant violation for expected transient unavailability, producing a false finding before workers start. Retry node selection or exit without an assertion when all probe failures are transient; preserve the individual error classifications rather than relying on the joined diagnostic.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2008#issuecomment-5661963695

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2008#issuecomment-5661963695

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/main.go
```diff
@@ -247,9 +271,28 @@ func runWorker(
 		// key on a different body (conflict) or minting a fresh tracked one (a
 		// replayable original) — to exercise the server's dedup.
 		c.stampIdempotency(&bulk)
+		var predictedCheckpointID uint64
+		if len(bulk.Requests) == 1 && bulk.Requests[0].GetCreateQueryCheckpoint() != nil {
+			predictedCheckpointID = c.modelState.NextQueryCheckpointID()
```
gfyrag, line 283: [P2][non-blocking] Keep the predicted checkpoint probe bound to its create

`predictedCheckpointID` is read only from the drained `modelState`. With the default eight workers, two create bulks can be registered before either response drains, so both workers capture the same next ID. If both creates commit, only the first receives that ID; the later worker probes the first checkpoint instead of the ID assigned to its own create. `checkpointCreationMatches` then legitimately matches the other in-flight transition, so the probe passes while the later create’s pre-commit window is never exercised. Please either prevent overlapping create dispatches for this probe or derive/probe the candidate ID(s) in a way that remains tied to the matching create transition, and add a regression with two concurrent in-flight creates.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2008#issuecomment-5661963695

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/main.go
```diff
@@ -301,13 +354,26 @@ func runWorker(
...
 		// Block on a full channel — natural back-pressure.
 		select {
 		case <-ctx.Done():
+			if checkpointCreate {
+				c.checkpointCreateMu.Unlock()
+			}
 			return
 		case c.incoming <- obs:
 		}
+		if checkpointCreate {
```
NumaryBot, line 370: 🟠 **[major] Resume draining after failures before waiting on create completion**

If a successful create is buffered behind an outstanding bulk that subsequently fails, `handleObservation` removes that failed ticket without calling `tryDrain`, leaving this wait blocked despite the create being eligible to drain. If the remaining workers next choose creates, they all block on `checkpointCreateMu`, so no worker can issue the read or successful write needed to restart draining. Without a restore cycle, workload progress stops until context cancellation. Retry draining after processing failed observations, and cover a buffered create followed by the last outstanding bulk failing.

Azorlogh (approved):

gfyrag (approved):

PR #2009, opened 2026-09-14
test(antithesis): model-driver coverage for the remaining query surface (EN-1625)

Azorlogh:
## What changed

Follows #1659 (merged as 9a6fa7d03); rebased onto #1954, #2008, #2021, #2046, #2055 and #2007. Extends the `singleton_driver_model` driver to the rest of the query surface, each read validated against the oracle at a linearizable Raft horizon:

- `AggregateVolumes` (index-free filters, per-(asset, colour) fold, with `collapse_colors` and `use_max_precision` folded the way the server's result stage does), `ListAuditEntries`, `ListLedgers`, `GetLedgerStats`, `GetLog` by global sequence, and the index registry reads `ListIndexes` / `GetIndex` / `GetIndexEntryStatus` — every served value is judged against the oracle over the candidate bases, the way `GetAccount` and the entity lists already are:
  - an audit success entry must be exactly one committed bulk (contiguous log range, one log per order, every sequence a log the model learned, ledgers equal to those logs' ledgers); a failure entry must match a rejection the model explained by ledgers, order count and reason; entries past the committed frontier are admissible only while a bulk is outstanding, and a reverse first page must start at the newest committed bulk;
  - a `GetLog` answer must be the model's row for that sequence, field by field through the ListLogs comparator, and a sequence past every one handed out must be NotFound;
  - `ListLedgers` must be exactly the fleet with every chart and ledger metadata matching a base; `GetLedgerStats` boundaries must match a base exactly and its usage counters are bounded by the model's.
- Numeric range shapes: metadata field leaves now roll open sides, exclusive sides and (one in eight) the type's extrema, a two-sided range is sometimes split into two same-key half-range leaves under `And`, the transaction id range gets the open/exclusive variants the date builtins had, and `And`/`Or` carry one to three operands. The field leaves were inclusive-only with bounds within 64 of a committed value, so the range merger never saw an exclusive extremum — the EN-2041 wrap (#2021) was unreachable by construction.
- Audit filters: all eight audit fields (ledger, outcome, sequence, proposal id, timestamp, log sequence, caller subject, order type), alone or under `And`/`Or`, aimed at values served earlier and at the model's committed log sequences, plus the two shapes the grammar refuses (sequence bound inside `Or`, `Not`). Every served entry must satisfy the filter on its own fields; an order-type leaf is judged against the kinds the model committed under the entry's logs; a zone scan must be dense from its lower bound; a bare log-sequence page the size limit did not cut must cover every committed log the model learned in its range.
- Has-asset leaves also probe untouched cells (foreign precision, unknown base) and a precision past the one-byte index cell (rejected once the index is ready); account address leaves carry any role.
- `And`/`Or` also roll zero operands now that an empty `And` is the universe on every target (EN-2055, #2046).
- The six new read surfaces share one slot of the worker's read mix (the slot that duplicated the plain account read), so the entity-query sondes the 300 s local runner requires keep the base's share.
- Log-date filters draw their bounds from dates the model has learned: a log date is a wall-clock microsecond count, so the fixed 0–71 range produced only total or empty windows and never a partial one.
- Metadata-schema ops fill a bulk slot 1-in-48 instead of 1-in-6, and each readiness poll RPC is bounded to one poll interval. Together these let `retype window closed` and the strict post-retype checks arm: before, every indexed key was retyped about every 2 s while closure needs two consecutive polls, every chained retype reset the per-node progress, and a dead replica froze the poller for the whole outage (43 s and 104 s gaps observed).

## Why

EN-1625. The model driver covered writes and the two entity list paths; every other read RPC, the checkpoint path and the log-date index were unexercised, and the index-lifecycle oracle was stuck in its permissive regime because its strict states were unreachable at the workload's own cadence.

N/A

## Technical decision

N/A

## Risk

**LOW** — test workload and oracle only; no production code.

## Validation

- [x] `go build ./...`, `go vet`, `go test ./...` in `tests/antithesis/workload`; `go test ./tests/oracle/...`; `bash scripts/agent-check` PASS; golangci-lint clean on `tests/oracle` and on every driver file this PR touches.
- [x] Local 3-node model runs on this branch merged with #1931 and #1954: 20 min (27 restart cycles), 5 min ×3, all PASS, no findings. On the 5-min runs `retype window closed`, `index promoted to active` and `index demoted to ambiguous` each fire; before the rate/timeout change `retype window closed` never executed.
- [x] 5-min local 3-node smoke of the final branch (7 leader/follower kill cycles): no model findings; `audit page validated`, `audit failure entry matched a model rejection`, `GetLog validated`, `GetLog on an unassigned sequence returned NotFound`, `ledger listing validated`, `ledger stats usage caught up with the model`, `retype window closed`, `index promoted to active` all reached. Two earlier iterations of the audit validator produced false findings that the smoke caught (setup-era entries the oracle seeded without a sequence; a bounded history ring evicting entries the trail still serves) — both fixed, both pinned by unit tests.
- [x] Red on EN-2041: a 450 s local 3-node run of this branch against the unfixed base fails fast with `indexed account query outside model` on `or(and(k2 > MaxInt64, k2 <= 10))` — oracle empty, server served every account carrying `k2`. The same run with #2021 merged in: PASS, no findings, 10 restart cycles.
- [x] 5-min local 3-node smoke of the audit/has-asset commit with #2021 merged in (7 kill cycles): no model findings; `audit page validated`, `audit failure entry matched a model rejection`, `audit sequence bound inside or rejected`, `audit not filter rejected`, `has-asset precision overflow rejected`, `asset-index account query served results`, `retype window closed` all reached. Against the unfixed base the same commit fail-fasts on EN-2041 through `not(or(and(k3 >= 0, k3 < 0), …))`.
- [x] Rebased branch (`e576a2cae`), 5-min 3-node smoke, 7 kill cycles: no model findings; every sonde of this PR reached.
- [ ] Antithesis model-only run, 120 min, on `e576a2cae`: https://github.com/formancehq/ledger/actions/runs/35121673120 — results to follow.
- [ ] Antithesis model-only run, 60 min, on the composite of this branch + #1931 + #1954 all rebased onto `release/v3.0` (`tmp/antithesis-mdq-1931-1954` @ `00df1e3aa`), dispatched 2026-09-14: https://github.com/formancehq/ledger/actions/runs/34830680991 — results to follow.

N/A

## Review focus

0. `genFieldLeaf` / `rollOpenOrExclusive` / `genChildren` — the shape rolls; rates are 1-in-8 extremum, 1-in-3 split range, 1/2/2/3 operands.
1. `audit.go` `validateAuditPage` — the admissibility rules above; system-scoped entries (no ledger) and setup-era entries (below the first sequence the driver learned) are not judged.
2. `genLogDateCond` — bounds sampled from learned dates with ±1 ms jitter; empty sample falls back to the trivial match.
3. `rollSchemaOp` rate and the per-poll timeout in `reconcileIndexes` — the two changes that make the strict index-lifecycle states reachable; the Antithesis run of 2026-09-11 shows the three sondes at 0 hits without them.

## Known concerns

- The three index sondes were also unreached on Antithesis because the driver's restore cycle (45/h) demotes every index and fault injection rarely leaves all three replicas answering in one poll — the timeout helps, the restore cadence is untouched.
- The 300 s local runner's must-hit gate is marginal on the base itself: a pristine `release/v3.0` run misses `Apply skipped ERROR_REASON_METADATA_NOT_FOUND`, `…TRANSACTION_REFERENCE_CONFLICT` and one address-index page sonde in 300 s, and this branch misses the same set (with the destination-address variant). Not introduced here.
- `run_model_test.sh` computes `MODEL_MAX_SECONDS` as duration + 15 s, but its last restart cycle can run 30 s past the deadline; a run whose cycles land that way reports `DRIVER EXITED EARLY` on a clean driver (seen once here). Separate fix.


tests/antithesis/workload/bin/cmds/model/singleton_driver_model/audit.go
```diff
@@ -0,0 +1,855 @@
...
+
+			if !slices.ContainsFunc(entries, func(e auditEntry) bool { return !e.failed && e.minLog <= seq && seq <= e.maxLog }) {
+				return auditVerdict{finding: "audit trail misses a committed log", entry: "log sequence " + strconv.FormatUint(seq, 10), why: "no served entry covers it", missingSeq: seq}
+			}
+		}
+	}
+
+	// A reverse first page starts at the newest entry, so the newest committed
+	// bulk must be on it — unless a later rejection or an undrained bulk is the
+	// real tail.
```
NumaryBot, line 287: 🟠 **[major] Restrict the reverse-tail assertion to unfiltered pages**

When a reverse query filters out the newest committed bulk, a correct page triggers this assertion: `committedMax` is global, but `latestSeen` only covers matching entries. For example, a log-sequence filter capped at 42 legitimately returns the bulk at 42 even if the model has committed sequence 44. `genAuditFilter` generates these bounded queries as well as ledger and order-type filters. Restrict this check to unfiltered pages or compute the newest bulk that satisfies the filter.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/audit.go
```diff
@@ -0,0 +1,855 @@
...
+// is contiguous, one order per log, every sequence is a committed log, and the
+// entry's ledgers are exactly those logs' ledgers.
+func auditSuccessMismatch(e auditEntry, logs map[uint64]committedLog, committedMax uint64) (why string, missingSeq uint64) {
+	if e.minLog == 0 || e.maxLog < e.minLog {
+		return "empty or inverted log range", 0
+	}
+
+	if e.maxLog > committedMax {
+		return "log range straddles the committed frontier", 0
+	}
```
NumaryBot, line 382: 🟠 **[major] Validate audit ranges against recorded bulk boundaries**

The validator checks that the range contains known logs, but never verifies that those logs belong to exactly one committed bulk. If two one-order bulks commit adjacent sequences 10 and 11 on the same ledger, a corrupted entry claiming a two-order bulk spanning 10–11 passes every check here. Likewise, an entry representing only part of a larger bulk can pass. Record committed bulk boundaries and require an exact range match so the advertised one-entry-per-bulk oracle actually detects merged or split audit entries.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2009#issuecomment-5700879395

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/audit.go
```diff
@@ -0,0 +1,855 @@
...
+				continue
+			}
+
+			if !slices.ContainsFunc(entries, func(e auditEntry) bool { return !e.failed && e.minLog <= seq && seq <= e.maxLog }) {
+				return auditVerdict{finding: "audit trail misses a committed log", entry: "log sequence " + strconv.FormatUint(seq, 10), why: "no served entry covers it", missingSeq: seq}
+			}
+		}
+	}
+
+	// A reverse first page starts at the newest entry, so the newest committed
```
NumaryBot, line 285: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Respect the filter when checking the reverse audit tail**

This compares every reverse page against the global committed frontier, regardless of its filter. With no outstanding bulks, a ledger-A query correctly excludes a newer ledger-B bulk but triggers “audit tail misses the newest committed bulk.” Sequence, timestamp and order-type filters can produce the same false finding. Restrict this check to unfiltered pages or derive the newest matching bulk, and add a filtered reverse-page regression test.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/audit.go
```diff
@@ -0,0 +1,855 @@
...
+func auditSuccessMismatch(e auditEntry, logs map[uint64]committedLog, committedMax uint64) (why string, missingSeq uint64) {
+	if e.minLog == 0 || e.maxLog < e.minLog {
+		return "empty or inverted log range", 0
+	}
+
+	if e.maxLog > committedMax {
+		return "log range straddles the committed frontier", 0
+	}
+
+	span := e.maxLog - e.minLog + 1
```
NumaryBot, line 384: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Preserve bulk boundaries when validating audit success entries**

Checking contiguous learned sequences and their ledger union does not establish that they belong to exactly one committed bulk. For example, two separate one-order commits at sequences 10 and 11 are accepted as a fabricated two-order audit entry spanning 10–11. A subset of a larger bulk also passes. This misses the bulk-integrity failures the new validator claims to detect. Record committed bulk boundaries and require each success entry to match one complete bulk.

tests/antithesis/workload/bin/cmds/model/singleton_driver_model/main.go
```diff
@@ -230,6 +230,8 @@ func runWorker(
 				runReplay(ctx, client, c)
 			case 6:
 				runLogQuery(ctx, client, c)
+			case 7:
```
NumaryBot, line 233: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Preserve a reachable dispatch slot for GetAccount**

Adding case 7 exhausts every value in `RandomChoice({0,1,2,3,4,5,6,7,8})`, making the default branch unreachable. That branch is the only caller of `runRead`, so the driver stops exercising its model-validated `GetAccount` reads, including absent-account and absent-ledger probes. Add a selectable slot for `runRead` to preserve this existing coverage.

NumaryBot (commented):
OpenCodeReview's independent Codex pass added 3 comment-only findings.

flemzord (commented):
Scoped validation of the per-node `GetIndexStatus` deadline at `e576a2caee1729cd69f6f288e1349263d4148c2c`:

The identical deterministic regression fails on the run source `2b52ff5468d68368e4e951625048f404d27ccff3` after 8 seconds because the blocked first RPC exhausts the parent deadline. With this PR's correction, it passes in 2 seconds: the request is bounded to one poll interval, the parent stays live, the healthy second node is sampled, the index demotes, and the next successful poll restores active readiness.

The regression is now preserved in the tests-only follow-up #2106, targeting this branch (`feat/mdq-coverage`) and reusing EN-1625. The complete model package passed under the race detector; both canonical local gates passed against this exact parent SHA. The parent runtime code and branch were not changed.

This COMMENT is limited to that mechanism. It is not approval of the complete PR, does not resolve the five open review discussions or the parent CI failures, and does not turn the historical Antithesis 0/0 probes into remote coverage evidence.

PR #2010, opened 2026-09-14
fix(health): publish coherent disk usage samples (EN-2044)

gfyrag:
## What changed

Publish each filesystem volume's used bytes and total capacity as one atomic,
immutable sample. Local health evaluation and `GetDiskUsage` now load each pair
together, and the separate scalar getters are removed.

## Why

Fixes [EN-2044](https://formance-team.atlassian.net/browse/EN-2044).

Two independent atomics allowed readers to combine values from different
filesystem collection cycles during a capacity change. The fabricated fraction
could falsely trip the disk write gate and then remain blocked through the
hysteresis band.

Need: healthy clusters must keep admitting writes unless a coherent filesystem
sample crosses the configured disk threshold.

Current limitation: a capacity change could fabricate an over-threshold ratio
that no collection cycle observed.

Requirement / constraint: preserve existing write-only disk hysteresis while
making every per-volume `(used,total)` observation coherent.

Evidence: EN-2044 and the pre-fix regression output `observed fabricated disk
sample 154/100` for coherent source samples `70/100` and `154/200`.

Durable repository evidence: `docs/ops/disk-space.md` documents coherent sample
publication; `TestVolumeUsage_PublishesCoherentSample` enforces it.

## Technical decision

Decision: publish an immutable sample through `atomic.Pointer` and expose one
pair-returning `Load` operation.

Why now / why proportionate: the collector is read concurrently by health, RPC,
and metrics paths; one atomic pointer is the smallest synchronization primitive
that makes the pair indivisible without adding locks.

Alternatives considered: a mutex would add coordination to every reader; a
sequence-lock retry protocol would be more complex; retaining scalar getters
would leave the unsafe consumption pattern available.

## Risk

**LOW** — node-local monitoring publication only. No FSM, persistence, protobuf,
or external API contract changes.

## Validation

- [x] Targeted tests: `go test -race ./internal/infra/monitoring/diskusage ./internal/infra/health ./internal/adapter/grpc`
- [x] Broader validation: `AI_REVIEW_BASE_SHA=ddae03bfd3dfc72edc877c7ca6ac1cfaf76bdbff bash scripts/agent-check-pr`

Disk admission hysteresis is unchanged. Its local and peer inputs are now
coherent across filesystem capacity changes. Zero-value startup behavior remains
`0/0` until the collector's synchronous first sample.

## Review focus

Please verify the atomic sample publication contract and that every production
consumer migrated away from independent scalar reads.

## Known concerns

None.

[EN-2044]: https://formance-team.atlassian.net/browse/EN-2044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gfyrag:
@claude review

paul-nicolas (approved):

PR #2011, opened 2026-09-14
fix(EN-1531): verify metadata logs against audited orders

gfyrag:
## What changed

The checker now reconstructs exact `SavedMetadata` expectations from hash-bound `SaveMetadata` and mirror `SET_METADATA` orders. It compares persisted log payloads with that authoritative value before replay can use the log as the expected-state oracle for metadata projections.

A production-path regression covers a healthy store, coordinated log/projection corruption, and projection-only corruption.

## Why

A coordinated mutation of a stored `SavedMetadata` payload and its `SubAttrMetadata` row could previously remain mutually consistent while contradicting the intact audited serialized order. `CheckStore` and staged restore validation could therefore certify corrupted business metadata as clean.

Related: [EN-1531](https://formance-team.atlassian.net/browse/EN-1531)

Need: integrity verification must reject persisted business metadata that contradicts audited caller intent.
Current limitation: replay derived its metadata expectation from the same unchecked history-log payload whose matching projection was being verified.
Requirement / constraint: a `SavedMetadata` log and its projection must not verify clean when either conflicts with the hash-bound serialized order.
Evidence: EN-1531 and `TestCheck_SavedMetadataUsesAuditedOrderAsAuthority`.
Durable repository evidence: `docs/technical/architecture/subsystems/checker/checker.md` and the regression test.

## Technical decision

Decision: fold only the complete `SavedMetadata` payloads already present in chain-verified metadata orders, then compare them before log replay.
Why now / why proportionate: these orders carry every emitted business field, so the checker can close this P1 oracle without a second processor implementation or an on-disk format change.
Alternatives considered: full canonical log reconstruction was rejected because it requires broader authenticated replay state and is unnecessary for this payload; doing nothing preserves the confirmed P1 bypass.

## Risk

**LOW** — checker-only read path; no FSM, API, protobuf, protocol revision, or storage-format change. Healthy and mirror-derived expectations use exact protobuf equality.

## Validation

- [x] `go test ./internal/application/check -count=1`
- [x] `go test -race ./internal/application/check -run TestCheck_SavedMetadataUsesAuditedOrderAsAuthority -count=1`
- [x] `AI_REVIEW_BASE_SHA=ddae03bfd3dfc72edc877c7ca6ac1cfaf76bdbff bash scripts/agent-check-pr`
- [x] Mutation sensitivity: removing the new verification call makes the coordinated-corruption case fail with zero checker events.

The requirement is satisfied when the healthy control stays clean, attribute-only corruption remains detected, and coordinated log/projection corruption emits `METADATA_MISMATCH` while the retained audit item still decodes to `status=approved`.

The checker now treats hash-bound metadata orders as the authority for `SavedMetadata` payload contents before deriving downstream expected projections. Persisted state and write behavior are unchanged.

## Review focus

Please focus on the audit-item-to-log-sequence fold, legacy replay filtering, and the decision to reuse `METADATA_MISMATCH` for order/log divergence.

## Known concerns

None.

[EN-1531]: https://formance-team.atlassian.net/browse/EN-1531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gfyrag:
Addressed the actionable review feedback in 0ac6e9169:

- added coverage for the mirror `SET_METADATA` path, including coordinated history/projection corruption
- documented why the ignored `batch.Cancel()` error is intentional best-effort cleanup

I kept the two expected-metadata constructions inline: the regular and mirror protobuf accessor shapes differ, and extracting a helper for only these two direct cases would add indirection without removing meaningful logic.

paul-nicolas (approved):

PR #2022, opened 2026-09-14
feat(grpc): enforce RPC auth in interceptors (EN-2033)

paul-nicolas:
Moves public gRPC credential evaluation and authorization into the production unary and streaming interceptor chain, before generated handlers run. Fixed policies use the generated registry, Apply authorizes every embedded request while preserving signed-payload verification precedence, and ListIndexes authorizes its first request message before handler business logic.

Public, restore, and Raft service construction retain distinct trust modes. Handler-owned auth configuration and manual guards are removed, and the replacement tests exercise the production server chain, 401/403 behavior, cluster-secret forwarding, dynamic scope selection, mixed Apply batches, and real admission signature verification.

The public interceptor chain also stamps the query-profile request clock before authentication. This preserves the previous full-request timing boundary and allows authentication and authorization denials on profiled reads to emit the requested profile trailer and participate in slow-query telemetry. Restore services retain the handler-local fallback because they do not install the public interceptor chain.

Stacked on #2020.

Validation:
- `nix develop --command just pre-commit`
- `bash scripts/agent-check`
- full `.github/workflows/main.yml` local matrix (unit/internal coverage, operator, builds, business/cluster/scenario E2E, Antithesis race, 180s model, Schemathesis 62/62)
- `GOFLAGS=-p=1 AI_REVIEW_BASE_SHA=8cd62d1b464194ad3d7e89602a80b76c3a41c6c0 bash scripts/agent-check-pr`


internal/adapter/grpc/server_bucket.go
```diff
@@ -497,17 +441,12 @@ func (impl *BucketServiceServerImpl) ListTransactions(req *servicepb.ListTransac
 		trace.WithAttributes(attribute.String("ledger", req.GetLedger())))
 	defer span.End()
 
-	// Start the profile clock before authentication and validation so those
-	// phases land inside PrepareDuration instead of being invisible. Protobuf
-	// decode is already done: grpc-go unmarshals the request before dispatching,
-	// so no in-handler clock can reach it.
+	// Authentication and protobuf decoding finish in the interceptor/generated
```
gfyrag: [P2][blocking] Preserve cross-transport query-profile timing

Moving authentication into the interceptor makes this `WithProfile` call start after credential evaluation. The HTTP path deliberately captures its profile clock before authentication, and both `internal/query/profile.go` and `middleware_query_profile_test.go` state that this ordering exists so `prepare_duration_us`/`server_duration_us` measure the same span on HTTP and gRPC. After this change, JWT verification is included only on HTTP, so the same metric has different semantics depending on transport. Please carry a pre-auth timestamp/profile through the gRPC interceptor (or otherwise preserve the documented common boundary) and add a regression assertion for the interceptor-backed path.

gfyrag: Fixed in `8ce73fcc`. The public-service chain now captures the query-profile start immediately before the auth interceptor, and all four profiled gRPC handlers reuse that timestamp via `WithProfileStartingAt`, restoring the documented HTTP/gRPC timing boundary. The focused race tests and exact-base `agent-check-pr` pass. I found no regression in the stream context wrapping.

internal/adapter/auth/grpc_auth.go
```diff
@@ -124,11 +118,46 @@ func Authenticate(ctx context.Context, cfg AuthConfig, scopes ...Scope) (context
...
+	logAuthFailure(ctx, KeyIDFromContext(ctx), "missing_scope", fmt.Errorf("required: %v, have: %v", scopes, claimedScopes))
 
-		return ctx, status.Errorf(codes.PermissionDenied,
-			"missing required scope (required: %v)", scopes)
+	return status.Errorf(codes.PermissionDenied, "missing required scope (required: %v)", scopes)
+}
+
+// Authenticate is retained for compatibility tests and non-handler callers.
+// Public gRPC handlers are authenticated by the service interceptors.
+func Authenticate(ctx context.Context, cfg AuthConfig, scopes ...Scope) (context.Context, error) {
```
Azorlogh: `Authenticate` has no production callers left — 30 call sites, all in `grpc_auth_test.go`. V3 is unreleased, so move it into the test file as an unexported helper and drop it from production.

internal/adapter/grpc/server_bucket.go
```diff
@@ -99,65 +97,15 @@ func NewBucketServiceServer(logger logging.Logger, c ctrl.Controller, localCtrl
...
-
-				return nil, status.Errorf(codes.PermissionDenied,
-					"request %d requires scope %s", i, required)
-			}
-		}
-	}
+	// Authorization already peeked the batch in the interceptor. Keep a
+	// best-effort view solely for tracing and metrics; admission remains the
+	// authoritative parser and signature verifier.
+	batch, _ := servicepb.PeekBatch(req)
```
Azorlogh: Three `UnmarshalVT` of a signed payload per Apply now — here, in the interceptor, and in admission's `ExtractBatch` — for a trace line and the `batch_size` label. Worth carrying the count forward from the interceptor.

internal/adapter/grpc/server_bucket.go
```diff
@@ -1411,6 +1208,14 @@ func (impl *BucketServiceServerImpl) emitProfile(ctx context.Context, profile *q
 	}
 }
 
+func withTransportQueryProfile(ctx context.Context) (context.Context, *query.QueryProfile) {
```
Azorlogh, line 1212: HTTP asserts on this same invariant (`withQueryProfile` → `assert.Unreachable`); gRPC falls back silently, so a chain reorder would quietly reintroduce the under-reporting you just fixed. The Antithesis SDK is pinned no-op, so the assert won't affect tests that call handlers directly.

Azorlogh (commented):
Checked all 48 pre-PR handler guards against the generated registry mechanically — zero mismatches, and `Discovery` is the only policy method with no former guard (declared public, matching its prior state).

Three inline comments. Also stale, but outside this diff:
- `internal/adapter/http/handler.go:117` — "The gRPC handlers start their profile before `internalauth.Authenticate`".
- `docs/technical/architecture/subsystems/admission/pipeline.md:21` and `:48` — still say `Apply` authenticates the caller.
- `internal/adapter/auth/grpc_auth_test.go:611` — describes the old Apply handler.

One for #2026 rather than here: this adds `AuthenticationState` beside `expandedScopesKey`/`authPresentedKey`, and #2026 adds a third population site on HTTP while `buildCallerSnapshot` still reads the old key. Raising it there.

internal/adapter/grpc/server.go
```diff
@@ -940,6 +942,12 @@ func NewServiceServer(authPolicy ServiceAuthPolicy, host string, port int, logge
 		errorConversionStreamInterceptor(logger),
 		protocolVersionStreamInterceptor(),
 	}
+	if authPolicy == ServiceAuthPolicyPublic {
+		// Stamp the query-profile clock before authentication so profiled reads
+		// measure the same preparation span as their HTTP counterparts.
+		unaryInterceptors = append(unaryInterceptors, queryProfileClockUnaryInterceptor(), authUnaryInterceptor(authCfg))
```
NumaryBot: 🟠 **[major] Preserve profiles for auth-denied reads**

When authentication or authorization fails for any profiled RPC, these interceptors return before the handler can create and defer `emitProfile`. Previously the four handlers created their profile before `Authenticate`, so requested gRPC profiles still produced `x-query-profile-result-bin` trailers and slow auth failures fed the slow-query log. This now violates the full-profile-on-every-profiled-read contract (`docs/technical/architecture/subsystems/read-path/query-profile.md:188-193`); retain profile emission around interceptor denials and cover it through the real server chain.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2022#issuecomment-5695916729

docs/technical/architecture/subsystems/api/auth.md
```diff
@@ -71,7 +71,9 @@ Every BucketService and ClusterService RPC declares a typed `common.auth_policy`
...
 
-The generated inventory is the structural contract for the interceptor migration. During this first step, request behavior is unchanged: each RPC still calls `auth.Authenticate(ctx, cfg, scopeRequired...)` explicitly, and dynamic methods still derive their required scope in their handler.
+The public service consumes this inventory in unary and streaming interceptors before generated handlers run. Credential evaluation produces immutable authentication state, then fixed policies check their declared scope. Dynamic resolvers inspect the request: `Apply` checks every embedded request, `GetIndex` and `GetIndexEntryStatus` choose ledger-read or ops-read from the ledger field, and `ListIndexes` authorizes the first received request message before handler business logic. An unparsable signed Apply payload still reaches admission so signature verification keeps precedence over payload parsing.
```
NumaryBot, line 74: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Update the admission pipeline’s gRPC entry-point contract**

This new interceptor-based description conflicts with `docs/technical/architecture/subsystems/admission/pipeline.md`, which still says `BucketServiceServerImpl.Apply` authenticates JWTs and peeks the batch. At HEAD those actions occur in the public interceptor before the generated handler; the handler only adopts a trusted forwarded snapshot and delegates. Update that architecture flow/table so future maintainers and audits do not rely on the removed handler-owned boundary.

internal/adapter/auth/grpc_auth.go
```diff
@@ -124,14 +118,35 @@ func Authenticate(ctx context.Context, cfg AuthConfig, scopes ...Scope) (context
 	ctx = WithExpandedScopes(ctx, effective)
 	ctx = WithAuthPresented(ctx, true)
 
-	if !HasScope(effective, scopes...) {
-		logAuthFailure(ctx, keyID, "missing_scope", fmt.Errorf("required: %v, have: %v", scopes, claims.Scopes))
+	return withAuthenticationState(ctx, true, true, effective), nil
```
NumaryBot, line 121: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Preserve trace telemetry for scope denials**

`EvaluateGRPCCredentials` returns (and its deferred `span.End` runs) before the interceptor invokes `AuthorizeGRPC`. Scope-denial paths then call `logAuthFailure` with a context whose current auth span is already ended, so OTel drops the failure attributes, error, and error status. This regresses tail-sampling/trace visibility for missing-scope and anonymous-scope denials, and makes the updated auth documentation’s tracing claim false.

NumaryBot (commented):
OpenCodeReview's independent Codex pass added 2 comment-only findings.

internal/adapter/grpc/auth_interceptor.go
```diff
@@ -0,0 +1,265 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	internalauth "github.com/formancehq/ledger/v3/internal/adapter/auth"
+	"github.com/formancehq/ledger/v3/internal/proto/commonpb"
+	"github.com/formancehq/ledger/v3/internal/proto/servicepb"
+)
+
+type queryProfileClockKey struct{}
+type applyBatchSizeKey struct{}
```
NumaryBot, line 25: **🟡 [minor] Query-profile clock instrumentation rides along beyond the stated scope**

The diff adds `queryProfileClockUnaryInterceptor`/`queryProfileClockStreamInterceptor`, `withTransportQueryProfile` and an Antithesis `assert.Unreachable` fallback, changing `grpc.apply.duration`/query-profile timing behavior (verified in the frozen diff). The PR description claims only interceptor auth enforcement, so this is scope creep inside a high-risk auth change. Describe the clock change in the PR description or split it into its own PR.

internal/adapter/grpc/auth_interceptor.go
```diff
@@ -0,0 +1,265 @@
...
+	return func(srv any, stream ggrpc.ServerStream, info *ggrpc.StreamServerInfo, handler ggrpc.StreamHandler) error {
+		if isInfrastructureRPCMethod(info.FullMethod) {
+			return handler(srv, stream)
+		}
+
+		policy, err := commonpb.RPCAuthPolicyForMethod(info.FullMethod)
+		if err != nil {
+			return status.Error(codes.Internal, err.Error())
+		}
+		if public, ok := policy.GetPolicy().(*commonpb.MethodAuthPolicy_Public); ok {
```
NumaryBot: **🟡 [minor] Policy dispatch duplicated between unary and stream interceptors**

`authUnaryInterceptor` (lines 42-53) and `authStreamInterceptor` (lines 73-84) each repeat the infrastructure bypass, the `commonpb.RPCAuthPolicyForMethod` lookup, and the `Public` policy validation (verified in target file). A future edit to one copy would not be mirrored on the other transport, silently weakening one chain. Extract a shared helper that resolves the policy and reports whether it is public.

internal/adapter/grpc/auth_interceptor.go
```diff
@@ -0,0 +1,265 @@
...
+				return err
+			}
+			if err := internalauth.AuthorizeGRPC(ctx, scope); err != nil {
+				return err
+			}
+		case *commonpb.MethodAuthPolicy_DynamicResolver:
+			if typed.DynamicResolver != commonpb.DynamicAuthResolver_DYNAMIC_AUTH_RESOLVER_LIST_INDEXES {
+				return status.Errorf(codes.Internal, "dynamic resolver %s is not valid for a streaming RPC", typed.DynamicResolver)
+			}
+			stream = &listIndexesAuthServerStream{authServerStream: stream.(*authServerStream)}
```
NumaryBot: **🔵 [suggestion] Unchecked type assertion couples stream wrapping order**

`stream.(*authServerStream)` in `authStreamInterceptor` relies on the wrapper assigned two lines earlier (verified in target file); inserting another stream wrapper between them panics at runtime. Keep the `*authServerStream` value in a local variable and pass it to the `listIndexesAuthServerStream` constructor.

internal/adapter/grpc/server_bucket.go
```diff
@@ -1411,6 +1206,15 @@ func (impl *BucketServiceServerImpl) emitProfile(ctx context.Context, profile *q
...
+	if start, ok := ctx.Value(queryProfileClockKey{}).(time.Time); ok {
+		return query.WithProfileStartingAt(ctx, start)
+	}
```
NumaryBot, line 1217: **🔵 [suggestion] Profile-clock assertion assumes the public interceptor chain**

`withTransportQueryProfile` fires `assert.Unreachable` when `queryProfileClockKey` is absent, but `server.go` installs the clock interceptor only for `ServiceAuthPolicyPublic`; restore servers pass no clock and register `BucketServiceServerImpl` only in tests today. Any future registration of a profiled read handler on a restore server would trip the assertion while degrading to `query.WithProfile`. Gate the assertion on public mode or keep the fallback unannotated.

NumaryBot (commented):
Findings

docs/technical/architecture/subsystems/admission/pipeline.md
```diff
@@ -10,15 +10,17 @@ In one sentence: **admission turns external requests into preloaded, structurall
...
     participant A as Admission<br/>(admission.go)
     participant Pre as Preload<br/>(infra/preload, infra/plan)
     participant N as Node<br/>(infra/node)
     participant FSM as FSM<br/>(infra/state)
 
-    C->>G: Apply(SignedApplyBatch)
-    G->>G: Authenticate (JWT)
+    C->>I: Apply(SignedApplyBatch)
+    I->>I: Authenticate JWT + PeekBatch<br/>authorize every embedded request
+    I->>G: trusted request + caller snapshot
```
NumaryBot: 🟠 **[major] Describe the Apply trust boundary accurately**

For signed `Apply`, `PeekBatch` runs before admission verifies the signature (as the next diagram step shows), and the interceptor does not install the forwarded caller snapshot; `Apply.adoptForwardedSnapshotIfTrusted` does so in the handler. Labeling this handoff as a “trusted request + caller snapshot” misstates the security boundary for the architecture being changed. Document the request as unverified inspection data and the handler as the snapshot-adoption boundary (`AGENTS.md:91-99`).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2022#issuecomment-5695916729

docs/technical/architecture/subsystems/admission/pipeline.md
```diff
@@ -10,15 +10,17 @@ In one sentence: **admission turns external requests into preloaded, structurall
...
     participant Ctrl as Controller<br/>(controller_default.go)
     participant A as Admission<br/>(admission.go)
     participant Pre as Preload<br/>(infra/preload, infra/plan)
     participant N as Node<br/>(infra/node)
     participant FSM as FSM<br/>(infra/state)
 
-    C->>G: Apply(SignedApplyBatch)
-    G->>G: Authenticate (JWT)
+    C->>I: Apply(SignedApplyBatch)
+    I->>I: Authenticate JWT + PeekBatch<br/>authorize every embedded request
```
NumaryBot: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Do not label pre-verification Apply data as trusted**

`PeekBatch` unmarshals signed payloads without signature verification and is explicitly non-authoritative; malformed signed payloads intentionally continue to admission without per-request authorization so signature verification retains precedence. Thus neither the request nor its peeked batch size is trusted at this point, and “authorize every embedded request” needs that parseable-payload qualification. Update this flow (and the matching “trusted batch size” table entry) to reserve trust for admission’s verified extraction.

NumaryBot (commented):

internal/adapter/grpc/auth_interceptor.go
```diff
@@ -0,0 +1,305 @@
...
+			emitQueryProfile(ctx, profile, logger, slowThreshold)
+		}
+
+		return resp, err
+	}
+}
+
+func queryProfileClockStreamInterceptor(logger logging.Logger, slowThreshold time.Duration) ggrpc.StreamServerInterceptor {
+	return func(srv any, stream ggrpc.ServerStream, info *ggrpc.StreamServerInfo, handler ggrpc.StreamHandler) error {
+		clock := &queryProfileClock{start: time.Now()}
```
NumaryBot, line 47: **🟡 [minor] Stream profile clock is not injectable unlike the unary clock**

`queryProfileClockStreamInterceptor` hardcodes `time.Now()` while the unary variant exposes `queryProfileClockUnaryInterceptorAt` with an injectable `now`, so streaming clock behavior cannot be tested deterministically the way `TestQueryProfileClockIncludesInterceptorWork` does. Consolidated from spec (minor) and standards (suggestion). Mirror the injectable-clock helper for the stream interceptor.

internal/adapter/grpc/auth_interceptor.go
```diff
@@ -0,0 +1,305 @@
...
+
+		ctx, err := internalauth.EvaluateGRPCCredentials(stream.Context(), cfg)
+		if err != nil {
+			return err
+		}
+		authStream := &authServerStream{ServerStream: stream, ctx: ctx}
+		stream = authStream
+
+		switch typed := policy.GetPolicy().(type) {
+		case *commonpb.MethodAuthPolicy_FixedScope:
```
NumaryBot, line 113: **🟡 [minor] Duplicated fixed-scope authorization branch**

The FixedScope case in `authStreamInterceptor` repeats the same `authScope` conversion plus `AuthorizeGRPC` call as `authorizeUnaryRPC`, so a change to fixed-scope handling must be made twice. Extract a shared helper, e.g. `authorizeFixedScope(ctx, fixedScope)`, and call it from both interceptors.

internal/adapter/grpc/auth_interceptor.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+}
+
+func unexpectedAuthRequest(resolver commonpb.DynamicAuthResolver, req any) error {
+	return status.Errorf(codes.Internal, "dynamic resolver %s received %T", resolver, req)
+}
+
+func indexAuthScope(ledger string) internalauth.Scope {
+	if ledger != "" {
+		return internalauth.ScopeLedgersRead
```
NumaryBot, line 222: **🔵 [suggestion] Index read-scope rule expressed twice**

`indexAuthScope` (non-empty ledger) and `indexAuthScopeForList` (`SCOPE_LEDGER`) both encode the ledger-read-vs-ops-read decision with different predicates. Consolidate into one helper taking the scope selector or comment why the predicates differ, so future scope changes cannot desynchronize `GetIndex` and `ListIndexes`.

NumaryBot (commented):
Findings

internal/adapter/grpc/server_bucket.go
```diff
@@ -99,68 +97,15 @@ func NewBucketServiceServer(logger logging.Logger, c ctrl.Controller, localCtrl
...
-					return nil, status.Errorf(codes.Unauthenticated,
-						"request %d requires scope %s", i, required)
-				}
-
-				return nil, status.Errorf(codes.PermissionDenied,
-					"request %d requires scope %s", i, required)
-			}
-		}
-	}
+	batchSize, _ := ctx.Value(applyBatchSizeKey{}).(int)
```
NumaryBot, line 105: 🟡 **[minor] Apply batch_size metric silently depends on the interceptor chain**

When `BucketServiceServerImpl.Apply` is invoked without the public interceptor chain (such as existing handler-level harnesses), this context value is absent, so both the trace log and `grpc.apply.duration` record `batch_size=0` for a non-empty request. Retain a `PeekBatch` fallback or otherwise preserve accurate handler telemetry outside that chain.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2022#issuecomment-5695916729

NumaryBot (commented):
Findings

internal/adapter/grpc/auth_interceptor.go
```diff
@@ -0,0 +1,305 @@
...
+			return nil, err
+		}
+		if !protected {
+			return handler(ctx, req)
+		}
+
+		ctx, err = internalauth.EvaluateGRPCCredentials(ctx, cfg)
+		if err != nil {
+			return nil, err
+		}
```
NumaryBot, line 85: **🟡 [minor] Forwarded-snapshot diagnostic now follows scope checks**

`authUnaryInterceptor` authorizes every embedded Apply request before the handler's `adoptForwardedSnapshotIfTrusted`, inverting the old order. A forwarded snapshot on a non-cluster-internal connection lacking scopes now gets 401/403 "request N requires scope X" instead of the documented 403 forwarded-snapshot diagnostic. The write is still rejected; confirm the changed surface or pre-check forwarded snapshots.

internal/adapter/grpc/auth_interceptor.go
NumaryBot, line 47: **🟡 [minor] Stream profile clock lacks the injectable time seam**

`queryProfileClockStreamInterceptor` hardcodes `time.Now()` while `queryProfileClockUnaryInterceptorAt` (line 31) injects `now`, so streaming profile timing is only testable end-to-end (merged spec suggestion and standards minor). Add a matching `At` seam mirroring the unary variant.

internal/adapter/grpc/auth_interceptor.go
```diff
@@ -0,0 +1,305 @@
...
+		return ctx, internalauth.AuthorizeGRPC(ctx, indexAuthScope(entryReq.GetLedger()))
+	default:
+		return ctx, status.Errorf(codes.Internal, "dynamic resolver %s is not valid for a unary RPC", resolver)
+	}
+}
+
+func unexpectedAuthRequest(resolver commonpb.DynamicAuthResolver, req any) error {
+	return status.Errorf(codes.Internal, "dynamic resolver %s received %T", resolver, req)
+}
+
```
NumaryBot, line 219: **🟡 [minor] Duplicated ledger-to-scope rule in indexAuthScope and indexAuthScopeForList**

`indexAuthScope` (non-empty ledger -&gt; `ScopeLedgersRead` else `ScopeOpsRead`) and `indexAuthScopeForList` (`SCOPE_LEDGER` -&gt; same) encode the ledger-read/ops-read decision twice. Extract one shared decision point so a future scope change cannot update one path and miss the other.

internal/adapter/grpc/auth_interceptor.go
NumaryBot, line 113: **🔵 [suggestion] FixedScope authorization duplicated across unary and stream paths**

`authStreamInterceptor` repeats the `authScope` mapping plus `AuthorizeGRPC` call from `authorizeUnaryRPC`'s FixedScope branch instead of sharing one helper. Extract it so unary and streaming enforcement stay identical if scope mapping changes.

internal/adapter/grpc/auth_interceptor.go
```diff
@@ -0,0 +1,305 @@
...
+		return resp, err
+	}
+}
+
+func queryProfileClockStreamInterceptor(logger logging.Logger, slowThreshold time.Duration) ggrpc.StreamServerInterceptor {
+	return func(srv any, stream ggrpc.ServerStream, info *ggrpc.StreamServerInfo, handler ggrpc.StreamHandler) error {
+		clock := &queryProfileClock{start: time.Now()}
+		ctx := context.WithValue(stream.Context(), queryProfileClockKey{}, clock)
+		wrapped := &authServerStream{ServerStream: stream, ctx: ctx}
```
NumaryBot, line 49: **🔵 [suggestion] authServerStream doubles as profile-clock wrapper**

`queryProfileClockStreamInterceptor` wraps the stream in `authServerStream` purely to propagate the profile-clock context, with no auth state involved. Rename the wrapper (e.g. `contextServerStream`) so the auth-specific type is not reused for clock plumbing.

Azorlogh (commented):
Re-checked after the new commits: `Authenticate` is gone, the duplicate `PeekBatch` is gone, and the profile-clock assert being reverted in favour of the `claimed` flag reads fine to me.

Two new points inline, both non-blocking. Still stale and outside this diff: `internal/adapter/http/handler.go:117` — "The gRPC handlers start their profile before `internalauth.Authenticate`".

internal/adapter/auth/grpc_auth.go
```diff
@@ -124,14 +118,38 @@ func Authenticate(ctx context.Context, cfg AuthConfig, scopes ...Scope) (context
...
+}
+
+// AuthorizeGRPC checks required scopes against the immutable state produced by
+// EvaluateGRPCCredentials. Missing state is an internal wiring error and fails
+// closed.
+func AuthorizeGRPC(ctx context.Context, scopes ...Scope) error {
+	ctx, span := authTracer.Start(ctx, "auth.authorize")
```
Azorlogh, line 128: Unconditional, and the Apply resolver calls `AuthorizeGRPC` once per embedded request — so a batch of N emits N `auth.authorize` spans on the write path, with no request-count cap on gRPC Apply. Deduping the required scopes before that loop collapses it and drops the redundant `HasScope` calls.

internal/adapter/grpc/auth_interceptor.go
```diff
@@ -0,0 +1,305 @@
...
+		if err != nil && isProfiledRPCMethod(info.FullMethod) && !clock.claimed {
+			_, profile := query.WithProfileStartingAt(ctx, clock.start)
+			emitQueryProfile(ctx, profile, logger, slowThreshold)
+		}
+
+		return err
+	}
+}
+
+func isProfiledRPCMethod(method string) bool {
```
Azorlogh, line 60: This duplicates the four `withTransportQueryProfile` call sites with nothing keeping them in sync — a new profiled read loses its denial telemetry silently.

Azorlogh (approved):
Both points from my last pass are addressed: the Apply loop now authorizes distinct scopes only (keeping the first index in the error), and `isProfiledRPCMethod` is gone in favour of `wantsProfile` — no duplicated method list to drift.

Re-checked the rebase: the auth files are byte-identical to what I reviewed, and the `server_bucket.go` delta is the checkpoint-store work coming from the base, not this PR.

All CI jobs green.

Issue #1901, opened 2026-09-04
Chart: unknown dot-prefixed properties parse successfully and are silently discarded


ariel-formance:
## What happens

`ChartSegment.UnmarshalJSON` (`internal/chart.go:105-169`) dispatches dot-prefixed keys through an `if / else if` chain over `.self`, `.metadata` and `.rules`. There is no final `else`, so **any other dot-prefixed key parses successfully and is silently discarded** — no error, no warning, and it vanishes on round-trip.

Verified against `origin/main` (99c372e3f):

| input | parse result | round-trip output |
| --- | --- | --- |
| `{"bank":{".self":{},".patern":"^[A-Z]{4}$"}}` | `err=<nil>` | `{"bank":{}}` |
| `{"bank":{".self":{},".forbiden":{"client":true}}}` | `err=<nil>` | `{"bank":{}}` |
| `{"bank":{".self":{},".metadatas":{"x":{"default":"y"}}}}` | `err=<nil>` | `{"bank":{}}` |
| `{"bank":{".self":{},".rulez":{}}}` | `err=<nil>` | `{"bank":{}}` |
| `{"bank":{".self":{},".anything-at-all":[1,2,3]}}` | `err=<nil>` | `{"bank":{}}` |

## Why it matters

Every one of these is a plausible typo of a real key, and each fails in the permissive direction:

- `.patern` instead of `.pattern` — the segment silently becomes an **ungated wildcard** that accepts any value. The schema author believes they constrained the level; the engine accepts everything.
- `.metadatas` instead of `.metadata` — default metadata is silently dropped, so accounts get created without the metadata the chart promises (`FindAccountSchema` feeds `DefaultMetadata()` at `internal/transaction.go:253` and `internal/controller/ledger/controller_default.go:614`).
- `.rulez` instead of `.rules` — same shape.

There is no diagnostic anywhere in the flow. The chart uploads cleanly, validates cleanly, and is wrong.

Note that the parent frame reads a child's `.pattern` before recursing, so `.pattern` itself is handled correctly — this is specifically about keys the DSL does not define.

## Suggested fix

Reject unknown dot-prefixed keys in `ChartSegment.UnmarshalJSON` with an error naming the key, the way the root unmarshaller already rejects structurally invalid keys.

## Migration caveat

[...]

PR #2024, opened 2026-09-14
test(chart): preserve unknown-property upload and import compatibility

DivyamTalwar:
## Summary

Preserve the existing chart-decoding contract, as requested in the review of cb80817430ed5cf41490e33c67fc2088b68c96f2.

The unconditional rejection of unknown dot-prefixed properties has been removed. `internal/chart.go` is restored **byte-for-byte** to base commit `7d20326887509f06e4ae1c71eaf517cce20b2f7d` (blob `13ea84ee79bc8f2605bac52caf877e67f5db0c42`). The effective PR diff against the base is now regression tests only: no production behavior change.

Related to #1901, but this PR no longer claims to implement strict typo rejection or resolve that issue. A stricter accepted-input contract requires a separately agreed compatibility change.

## Regression coverage

- `internal/chart_unknown_properties_test.go`: six unknown names, six JSON value types and five segment positions (180 combinations). Assert successful decoding, unchanged normalized meaning and serialization. Retain coverage of known directives, metadata keys beginning with a dot, valid variable patterns, invalid patterns, invalid root properties and receiver replacement/failure behavior.
- `TestInsertSchemaUnknownChartPropertyCompatibility`: submit the exact raw payload `{"chart":{"bank":{".patern":{}}}}` twice with the same idempotency key. The mocked controller expects the same normalized parameters for an initial insertion and an idempotency hit; assertions require 204 for both and `Idempotency-Hit: true` on replay.
- `TestLogsImportUnknownChartPropertyCompatibility`: send raw `INSERTED_SCHEMA` NDJSON at the import boundary, both as a single schema log and between two valid schema logs. Assert 204 and that the mocked consumer receives every decoded log in order. The fixture is not marshalled through `ChartOfAccounts`, which would strip the unknown key before reaching the handler. Cancellation cleans up the mock consumer if a decoding regression returns early.

The HTTP tests cover handler contracts with mocked controllers; they do not claim database-backed idempotency or transaction-commit verification.

## Verification actually performed

Using the available Go 1.23.2 toolchain, a reduced-source harness containing the actual chart parser/serializer and the unchanged new parser test file produced:

| Parser source | Passed leaf cases | Failed leaf cases |
| --- | ---: | ---: |
| Base `7d203268` | 189 | 0 |
| Compatibility fix | 189 | 0 |
| Reviewed strict head `cb808174` | 7 | 182 |

`go test -race -count=25 ./...` also passed in that reduced parser harness. Both uploaded test-file Git blob hashes were verified against the locally formatted files.

**Not executed:** native repository/API package tests, full-suite tests and `just pre-commit`. The environment has Go 1.23.2, whereas this repository requires Go 1.26.0 / toolchain Go 1.26.1. Downloading Go 1.26.1 failed because outbound DNS/network access is unavailable; a direct repository clone failed for the same reason. The new API tests were source-reviewed and formatted, but have not been compiled or run in the full repository here. Reduced harness results are not a substitute for native CI.

Required native verification includes `go test ./internal ./internal/api/v2` and the applicable repository checks. This PR remains a draft pending native verification and maintainer review.

## Scope

No API/OpenAPI or SDK changes, no stricter read path, no import error-mapping changes, no idempotency implementation changes, no persisted-data changes and no deployment. The two compatibility findings are addressed by restoring the original shared decoder, not by changing 500 to 400 or moving rejection to another existing entry point.


internal/chart.go
```diff
@@ -166,6 +166,10 @@ func (s *ChartSegment) UnmarshalJSON(data []byte) error {
 			if err != nil {
 				return fmt.Errorf("invalid account rules: %v", err)
 			}
+		} else if key != PATTERN_KEY {
```
flemzord: [P1] Preserve the existing schema-upload contract

For `POST /v2/{ledger}/schemas/{version}` with `{"chart":{"bank":{".patern":{}}}}`, the base parser discards the unknown property and reaches insertion; this branch now returns 400 `VALIDATION` before the controller is called. Differential handler tests with a successful mocked controller reproduce 204 on the base versus 400 on this head. This also bypasses idempotency lookup for a retry of a previously successful request. The current generated SDK can serialize this payload through `V2ChartSegment.AdditionalProperties`, and OpenAPI still permits it. Rejecting typos is intentional, but it is blocking under the maintainer's explicit no-existing-behavior-change acceptance rule, rather than intrinsically an incorrect validation policy. Please preserve the existing accepted-input behavior in this PR; any stricter contract needs a separately agreed compatibility change. Add an API compatibility test covering this payload and an idempotent replay.

internal/chart.go
```diff
@@ -166,6 +166,10 @@ func (s *ChartSegment) UnmarshalJSON(data []byte) error {
...
+			// .pattern is validated by the parent before it descends into this segment.
+			// Other dot-prefixed keys are not part of the chart language.
+			return fmt.Errorf("unknown chart property: %v", key)
```
flemzord: [P1] Keep existing import streams decodable

This shared unmarshaller also runs through `HydrateLog` for `INSERTED_SCHEMA`. A log containing `data.schema.chart = {"bank":{".patern":{}}}` decoded successfully on the base but fails here. Differential import-handler tests with a mocked stream consumer reproduce 204 before versus 500 `INTERNAL` after, because `controllers_logs_import.go:39-52` maps decoding failures to an internal error. In a multi-log import, earlier entries can already have committed (`controller_default.go:253-272`); per-log atomicity predates this PR, but this newly rejected input now stops a previously accepted stream. Please preserve compatibility for import decoding and cover this case at the import boundary. This is a separate consumer impact of the stricter parser and is also blocking under the maintainer's no-existing-behavior-change rule; changing the response to 400 alone would not restore compatibility.

flemzord (changes requested):
Requesting changes on cb80817430ed5cf41490e33c67fc2088b68c96f2 under the maintainer's explicit requirement to preserve existing behavior.

The validated inline findings are:
- [Schema upload compatibility](https://github.com/formancehq/ledger/pull/2024#discussion_r4005867278): a chart containing `bank: {".patern": {}}` previously reached insertion but now fails with 400 VALIDATION before the controller/idempotency lookup. Differential handler tests reproduce 204 -> 400 with a successful mocked controller.
- [Import compatibility](https://github.com/formancehq/ledger/pull/2024#discussion_r4005867283): the same chart in an INSERTED_SCHEMA log previously decoded successfully but now causes 500 INTERNAL at the import boundary. Differential handler tests reproduce 204 -> 500 with a mocked consumer; earlier logs in a multi-log import may already have committed.

Rejecting unknown directives is the intended correction, but that intention does not satisfy the maintainer's no-existing-behavior-change acceptance rule. This compatibility constraint is distinct from claiming that strict validation is intrinsically wrong. Please preserve the existing upload/import contracts in this PR; a stricter contract requires a separately agreed compatibility change. Parser and API package tests pass, but they do not establish behavior equivalence. No production stored-data incompatibility is claimed.

internal/chart.go
DivyamTalwar: Addressed in c05d1eaa24182251f70ca612ae68a1ff94d03e84. `internal/chart.go` is restored byte-for-byte to the base implementation, preserving the existing accepted-input contract rather than moving rejection elsewhere.

Added `TestInsertSchemaUnknownChartPropertyCompatibility`: the exact raw payload is submitted twice with the same idempotency key; ordered mock expectations require both calls to reach the controller with identical normalized parameters, returning 204 and then 204 with `Idempotency-Hit: true`.

Strict validation is deferred to a separately agreed compatibility change, and the PR title/description now reflect that scope. The reduced parser harness passes 189 cases (including a 25-run race check); the new native API tests remain unexecuted here because the required Go 1.26.1 toolchain could not be downloaded. Keeping the PR in draft and leaving this thread for your verification.

internal/chart.go
DivyamTalwar: Addressed in c05d1eaa24182251f70ca612ae68a1ff94d03e84 by restoring the original shared decoder byte-for-byte. No import error mapping or stream-commit behavior was changed.

Added `TestLogsImportUnknownChartPropertyCompatibility` at the HTTP boundary for a single raw `INSERTED_SCHEMA` log and for an unknown-property schema between two valid schema logs. It asserts 204 and delivery of every decoded log to the mocked consumer in order. Raw NDJSON retains `.patern` until `Log.UnmarshalJSON` / `HydrateLog`; the test also cancels its consumer on an early decoding failure to avoid leaking a goroutine.

These are mocked-consumer compatibility tests, not a claim about database transaction atomicity or production stored data. Native API execution remains pending because Go 1.26.1 could not be downloaded in this environment. The PR remains a draft and this thread is left for your verification.

PR #2026, opened 2026-09-14
feat(audit): persist total caller attribution (EN-2034)

paul-nicolas:
## What changed

Caller attribution is now a discriminated principal for authenticated, anonymous, named system, and authentication-disabled writes. The persisted snapshot carries effective authorization into the audit hash, renders in ledgerctl, survives follower forwarding, and is verified after incremental restore.

## Why

EN-2034 requires every admitted write to retain explicit, tamper-evident caller attribution instead of collapsing weak cases to a missing identity.

## Risk

**MEDIUM** — this changes the persisted/public protobuf shape and advances the gRPC protocol revision from 8 to 9.

## Validation

- [x] Baseline validation: Dirty and agent-check passed
- [x] Targeted tests: auth, gRPC/HTTP forwarding, admission, audit hashing/tamper, checker, CLI, restore
- [x] Full suite / broader validation: unit and optional-feature coverage; business and cluster E2E coverage; scenarios; builds; operator; Antithesis workload; Schemathesis 62/62

The 180-second model-cluster run completed four restart cycles with no model findings, then its driver exited early. A subsequent Docker rerun was stopped at the operator's request.

The public gRPC protocol revision moves from 8 to 9. Auth-disabled followers derive the same explicit principal on the leader instead of sending a client-spoofable field on plaintext clusters.

Stacked on #2022; merge after #2020 and #2022.

## Review focus

Please focus on deterministic audit encoding, principal rendering, and the follower trust boundary.

## Known concerns

The full parallel gate can exceed fixed Antithesis harness subprocess deadlines on a loaded host; the affected race test passes in isolation.


internal/proto/commonpb/common_reader.pb.go
```diff
@@ -11991,31 +11991,294 @@ func NewCallerIdentityListReader(s []*CallerIdentity) CallerIdentityListReader {
...
+// Call Mutate() to obtain a mutable clone.
+type CallerSnapshotReader interface {
+	GetPrincipal() isCallerSnapshot_Principal
+	Mutate() *CallerSnapshot
+}
+
+type callerSnapshotReadonly CallerSnapshot
+
+func (r *callerSnapshotReadonly) GetPrincipal() isCallerSnapshot_Principal {
+	return (*CallerSnapshot)(r).GetPrincipal()
```
NumaryBot: 🟡 **[minor] Do not expose the mutable principal through CallerSnapshotReader**

`CallerSnapshotReader` promises a read-only view, but `GetPrincipal` returns the original oneof wrapper. A consumer can type-assert it to an exported wrapper such as `*CallerSnapshot_Authenticated` and mutate its exported nested message or fields, thereby changing the underlying snapshot without calling `Mutate()`. This bypasses the reader immutability contract for caller attribution.

**Suggestion:** Return a cloned principal (for example, clone the `CallerSnapshot` before calling `GetPrincipal`) or expose read-only per-variant reader interfaces.

PR #2028, opened 2026-09-14
fix(raft): refresh committed peer addresses

gfyrag:
## What changed

Refresh the pooled Raft connection when an existing peer registration is observed with a new address. Preserve the existing peer send loop and queues, retain the committed target across transient optional-TLS probe failures, and add deterministic regressions for the committed `ConfChangeUpdateNode` path.

## Why

EN-2049: a committed peer refresh updated the membership cache and eventually Pebble while `DefaultTransport` kept dialing the old Raft endpoint.

https://formance-team.atlassian.net/browse/EN-2049

Need: a committed membership address refresh must make Raft dial the refreshed endpoint.
Current limitation: cache and durable state converged, but the live Raft connection pool retained the previous address until restart.
Requirement / constraint: refresh the operational connection after commit without adding FSM side effects or changing peer identity, role, queues, or membership cardinality.
Evidence: EN-2049 and `docs/technical/audits/raft-membership-leadership.md`.
Durable repository evidence: `docs/technical/architecture/subsystems/consensus/raft-consensus.md` and the updated audit contract.

## Technical decision

Decision: route every `DefaultTransport.AddPeer` call through the address-aware connection pool before returning for an existing peer loop. Failed replacements retain the committed address on a closed entry so the Raft loop or later service registration can retry it.
Why now / why proportionate: this keeps post-commit `Membership.Set` aligned with the committed dial target without moving transport side effects into FSM apply.
Alternatives considered: removing and recreating the whole transport peer would unnecessarily replace queues and the send loop.

## Risk

**LOW** — the change is isolated to peer connection replacement/retry behavior and preserves membership identity, role, queues, and deterministic FSM apply.

## Validation

- [x] `go test -race ./internal/infra/membership ./internal/infra/node ./internal/infra/transport -count=1`
- [x] Mutation proof: the committed-refresh regression observes `old:7000` when the pool refresh is removed.
- [x] Replay proof: `Rehydrate` converges cache, Raft transport, and service pool after an out-of-band committed address refresh while preserving replayed-removal coverage.

When `finishReady` observes a committed `ConfChangeUpdateNode`, it first refreshes the membership cache and both connection pools. Pebble may still contain the old registration until the entry is submitted to and committed by the asynchronous FSM applier; committed-WAL replay followed by `Rehydrate` repairs that interval after a crash. FSM apply remains deterministic and transport-free. No protocol change.

## Review focus

Please verify the committed refresh ordering, retained retry target, and preservation of peer identity, role, send loop, and queues.

## Known concerns

None.


internal/infra/node/transport.go
```diff
@@ -342,21 +342,22 @@ func (t *DefaultTransport) closePeers() error {
...
 func (t *DefaultTransport) AddPeer(id uint64, addr string) {
 	t.peersMu.Lock()
 	defer t.peersMu.Unlock()
 
-	if _, exists := t.peers[id]; exists {
-		return
-	}
-
 	if err := t.connectionPool.AddPeer(id, addr); err != nil {
 		t.logger.WithFields(map[string]any{"peer": strconv.FormatUint(id, 16), "addr": addr, "error": err}).Errorf("Failed to add peer to client pool")
```
NumaryBot, line 352: 🟠 **[major] Preserve the refreshed endpoint when the TLS probe fails**

In optional TLS mode, if the new endpoint is temporarily unreachable during this call, `ConnectionPool.AddPeer` closes the existing connection but returns from `dialPeer` without replacing the old entry. This error is only logged, and the retained send loop subsequently calls `RestartConnection`, which retries the old address, not the committed one. The refresh therefore remains lost even after the new endpoint recovers. Retain the committed target for retries and add a fail-then-success regression for this path.

docs/technical/audits/raft-membership-leadership.md
```diff
@@ -135,6 +135,13 @@ horizon. A forwarded fallback must not become an unbarriered local read merely
 because leadership changed during resolution; explicit stale mode remains the
 only expected bypass.
 
+A committed `ConfChangeUpdateNode` with a peer-registration payload must update
```
NumaryBot: 🟠 **[major] Document the actual cache-before-FSM execution order**

`Node.finishReady` calls `membership.Set` at line 1535 before submitting the committed entries for asynchronous FSM application at line 1643, so the cache and pools are refreshed before the durable peer row is written, not afterward. The new test manually executes the opposite sequence and does not establish this claimed production ordering. Describe Raft commit observation separately from Pebble application and acknowledge the intervening state, as required by [AGENTS.md:99](../../AGENTS.md#L99).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2028#issuecomment-5666373684

internal/infra/node/transport.go
```diff
@@ -342,21 +342,22 @@ func (t *DefaultTransport) closePeers() error {
...
+// and queues, but its pooled connection is refreshed when addr changed.
 func (t *DefaultTransport) AddPeer(id uint64, addr string) {
 	t.peersMu.Lock()
 	defer t.peersMu.Unlock()
 
-	if _, exists := t.peers[id]; exists {
-		return
-	}
-
 	if err := t.connectionPool.AddPeer(id, addr); err != nil {
```
NumaryBot, line 351: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Retain the refreshed address when connection replacement fails**

With optional TLS enabled, a transient TCP failure at the refreshed endpoint makes `ConnectionPool.AddPeer` fail after closing the existing connection but before storing the new address. This newly reachable replacement path then returns without arranging a retry. The preserved send loop calls `RestartConnection` using the pool's old address, so recovery of the new endpoint does not restore replication; if that restart also fails, the loop exits altogether. Preserve the committed dial target through replacement failures and retry it, with a regression covering a failed probe followed by recovery.

NumaryBot (commented):

internal/infra/node/transport.go
gfyrag: Done — fixed in commit a7b75fd69. Replacement failures now retain the committed target on the closed pool entry, and a deterministic optional-TLS fail-then-success regression proves RestartConnection retries the recovered new endpoint.

docs/technical/audits/raft-membership-leadership.md
gfyrag: Done — fixed in commit a7b75fd69. The docs, production comments, and regression now state and exercise the actual order: commit observation refreshes cache/pools first, then asynchronous FSM apply converges Pebble; the intervening state and crash-replay repair are explicit.

internal/infra/node/transport.go
gfyrag: Done — fixed in commit a7b75fd69 (same root cause as thread 091cb9b58d337d78). Failed replacement probes retain the new desired address and repeated RestartConnection failures keep the closed entry available for the peer loop to retry.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2028#issuecomment-5666373684

internal/infra/transport/connection_pool.go
```diff
@@ -201,6 +204,7 @@ func (p *ConnectionPool) AddPeer(id uint64, addr string) error {
 	// Close any pre-existing connection for this peer.
 	if existing, ok := p.peers[id]; ok {
```
NumaryBot, line 209: 🟠 **[major] Allow failed service-pool replacements to retry**

In optional TLS mode, a failed replacement probe now leaves the service-pool entry with the new address but a closed connection and a cancelled monitor. Unlike the Raft pool, the service pool has no owning loop that calls `RestartConnection`; its consumers simply use `GetConnection`. Subsequent `Membership.Set` calls with the committed endpoint now also return successfully through AddPeer's same-address guard, whereas previously they retried the replacement. Thus a transient outage can leave forwarding permanently using a closed channel even after recovery and repeated registration. Retry closed entries on AddPeer or provide pool-owned recovery, and cover the service-pool path.

internal/infra/transport/connection_pool.go
```diff
@@ -201,6 +204,7 @@ func (p *ConnectionPool) AddPeer(id uint64, addr string) error {
...
 		p.teardownLocked(existing)
+		existing.addr = addr
```
NumaryBot, line 211: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Retry failed replacements in pools without a peer send loop**

In optional-TLS mode, if a service-address refresh encounters a transient probe failure, this assignment leaves the service pool holding a closed connection with the new address and a canceled monitor. Unlike the Raft pool, the service pool has no peer send loop calling `RestartConnection`: its consumers simply call `GetConnection`. Subsequent `Membership.Set` or `Start` calls now also return success through AddPeer's same-address early return, whereas previously they could retry the replacement. Consequently, forwarding to that peer remains broken after its endpoint recovers. Allow same-address registration to redial a closed entry, or provide pool-owned recovery, and cover the service-pool failure-then-recovery path.

NumaryBot (commented):

gfyrag:
Addressed the Shipfox review on the current branch:\n\n- The replacement-failure finding is fixed in `a7b75fd69`: the committed target survives a failed optional-TLS probe and a fail-then-success test proves retry recovery.\n- The stale `ReconcileAgainstConfState` / `Rehydrate` comments are corrected in `2173e3e62`.\n- The two non-blocking coverage notes are closed in `2173e3e62`: one test now drives the actual `finishReady` `ConfChangeUpdateNode` branch and asserts voter-role preservation; another asserts the exact peer loop and all three queue instances are retained across refresh.\n\nAll focused tests, race tests, and `scripts/agent-check` pass.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2028#issuecomment-5666373684

internal/infra/transport/connection_pool.go
NumaryBot, line 211: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🔴 **[blocker] Allow same-address registration to retry a failed replacement**

In optional TLS mode, a failed replacement probe now leaves the service-pool entry with the new address, a closed connection, and a canceled monitor. Subsequent `Membership.Set` or `Start` calls hit `AddPeer`'s same-address early return instead of retrying. Unlike the Raft pool, service consumers such as `nodeForwarder.resolve` only call `GetConnection`; no owning loop calls `RestartConnection`. Consequently, forwarding remains broken even after the endpoint recovers and registration is repeated. Retry closed entries on same-address registration and cover this service-pool fail-then-success path.

NumaryBot (commented):

internal/infra/transport/connection_pool.go
gfyrag: Done — fixed in commit 9a9764084. Same-address AddPeer is now idempotent only for a live connection; a closed entry retries the committed endpoint. The fail-then-success regression covers both Raft RestartConnection and service-pool re-registration.

internal/infra/transport/connection_pool.go
gfyrag: Done — fixed in commit 9a9764084 (same root cause as thread 48ab9d1c507f2ab9). Repeated Membership.Set/Start wiring now retries a closed same-address service connection, with deterministic outage-then-recovery coverage.

gfyrag:
Addressed the fresh Shipfox review in `34c32e88f`:

- Corrected the stale Rehydrate locking comment: optional-TLS `AddPeer` can perform a bounded network probe while the membership lock is held.
- Removed the probe-and-release port race entirely. The recovery regression now injects a deterministic first-probe outage followed by success, runs in parallel, and asserts exactly two probes of the committed target.
- Removing the temporary listener also removes both unjustified close-error discards.

On the two non-blocking coverage observations, no additional test is needed: `TestMembership_RehydrateAfterReplay` already proves replayed Pebble registrations are republished to cache and transport, while `TestCommittedPeerRefreshUpdatesRaftTransport` proves the cache-before-Pebble divergence and convergence order introduced here. The service recovery boundary is intentionally split: that membership test proves `Set` wires the service pool, and `TestConnectionPool_AddressReplacementRetriesCommittedTarget/service_re-registration` proves a later same-address registration recovers a failed service-pool replacement. Combining those established lifecycle layers would duplicate behavior without a new branch.

Focused race tests for transport, membership, and node pass.

gfyrag:
Addressed the latest Shipfox comments in `c5c1ef3ee`:

- Updated `RestartConnection`, `monitorPeer`, and `getPeerConn` comments to describe the retained-closed-entry semantics accurately for both Raft and service pools.
- Extended `TestMembership_RehydrateAfterReplay` from cache-only no-ops to recording Raft/service wiring. It now starts from old peer-7 endpoints, simulates WAL replay changing both endpoints in Pebble without touching memory, calls `Rehydrate`, and asserts cache, Raft transport, and service pool all converge to the committed endpoints.

Focused race tests for membership, transport, and node pass.

internal/infra/membership/membership_test.go
```diff
@@ -348,34 +373,43 @@ func TestMembership_OnSnapshotInstalled(t *testing.T) {
...
 	require.NoError(t, err)
 	m.Start()
 	require.Equal(t, "before:1", m.PeerAddresses()[7].RaftAddress)
+	require.Equal(t, "before:1", raftTransport.peers[7])
+	require.Equal(t, "before:2", servicePool.peers[7])
 
 	// Simulate WAL replay: WriteConfChange wrote these rows directly to
 	// Pebble without touching the cache.
-	require.NoError(t, ps.Delete(7))
+	require.NoError(t, ps.Put(7, "after:7", "after:8", nil))
```
NumaryBot, line 401: 🟠 **[major] Preserve the replayed-removal regression alongside address refresh**

Replacing `ps.Delete(7)` with an address update removes the existing regression for a peer deleted during WAL replay: this test no longer checks that post-replay `Rehydrate` removes that peer from memory. `TestMembership_OnSnapshotInstalled` exercises a different lifecycle trigger and does not preserve this replay case. Keep the original removal scenario as a sibling case, or delete another preloaded peer and assert its absence alongside the new refresh assertions, as required by [AGENTS.md:87](../../../../AGENTS.md#L87).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2028#issuecomment-5666373684

internal/infra/membership/membership_test.go
gfyrag: Done in ab12c3788. The replay regression is additive again: peer 7 changes endpoints, peer 9 is deleted, and peers 1/3 are added. After Rehydrate the test asserts cache, Raft transport, and service pool all contain the refreshed peer and all exclude the replay-deleted peer.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2028#issuecomment-5666373684

gfyrag:
Addressed the final Shipfox documentation note without changing HEAD: the PR description now states the real order—commit observation refreshes cache/pools first, then asynchronous FSM apply converges Pebble, with WAL replay + Rehydrate repairing a crash in between. I also updated the risk/decision text to reflect the retained-target retry semantics.

I am leaving the suggested `replacePeerLocked` extraction out: `AddPeer` and `RestartConnection` share teardown/redial mechanics but intentionally differ in admission (same-address live no-op) and retry ownership. The current duplication is small and keeps those distinct contracts explicit; Shipfox classified it as a non-blocking judgement call.

paul-nicolas (approved):

PR #2030, opened 2026-09-14
fix(query): reject aligned reads of a ledger deleted after the pin (EN-1991)

Azorlogh:
## What changed

`query.AlignedIndexSnapshot` gates the accepted projection snapshot on the ledger still being live, re-read through the caller's handle live view (`dal.ReadHandle.Live`) once the projection snapshot is open. A `LedgerInfo` row that is absent rather than soft-deleted is unreachable by contract and fails as an invariant instead of as a deletion. A ledger deleted after the read's pin is rejected with `ErrLedgerNotFound` (gRPC `NotFound`) on every aligned read path: `ListLogs`, prepared-query execution, filtered account/transaction lists, `AggregateVolumes`, `InspectIndex`. Builder comments about delete + same-name recreate now state that `processCreateLedger` rejects that create.

## Why

Fixes [EN-1991](https://formance-team.atlassian.net/browse/EN-1991). A LOGS read pins its main snapshot with the ledger live, then acquires the projection snapshot. A `DeleteLedger` folded in between wipes every ledger-scoped projection row, and the unfiltered LOGS universe — served from the projection alone — returned a successful empty page for a ledger the pin holds logs for. Both `ListLogs` and prepared LOGS execution were affected.

```text
DISCOVERY: NO_EXISTING_WORK
BEFORE_FIX: BUG_REPRODUCED
AFTER_FIX: PASS
```

Need: a read that races a ledger deletion must answer either the pinned logs or an explicit rejection; a client cannot tell a wiped page from a genuinely empty ledger.
Current limitation: the projection wipe and the pinned main snapshot disagree about the ledger's existence, and nothing on the LOGS path checked.
Requirement / constraint: every aligned read's response is either complete at its pin or `NotFound`.
Evidence: reproduced end to end with a throwaway spec driving fold lag by apply pressure — 12/12 reads admitted before the deletion returned 0/3 logs on both callers (ListLogs and ExecutePreparedQuery). The spec is not part of the PR: a black-box read cannot tell a gate rejection from a late admission (both are `NotFound`), so its only proof of an open window was a hardware-dependent fold-lag gate, which CI showed does not form on a 4-vCPU runner (lag stayed at the 80 in-flight logs). Deterministic regression coverage lives at the gate instead (see Validation).
Durable repository evidence: `docs/technical/architecture/subsystems/read-path/read-snapshot-consistency.md` ("A deleted ledger rejects the aligned read"), `requireLedgerLive` doc comment in `internal/query/aligned_snapshot.go`.

## Technical decision

Decision: post-alignment liveness re-read through the handle's live view. Each lookup observes whatever is committed when it runs, so it is at or ahead of the projection snapshot by construction rather than by argument — no second snapshot to take, close, or reason about. The builder folds a deletion only after the main store applied it, so a projection snapshot holding the wipe is always paired with a live view showing the deletion; a ledger live there proves the projection snapshot holds every pinned row. The FSM rejects creating a ledger under a soft-deleted name, so no identity check is needed. The read goes through the caller's handle so it reuses the lifecycle read lock already held: a second `NewReadHandle` would block behind a queued `Close`/`RestoreCheckpoint` that is itself waiting for the first handle.
Why now / why proportionate: one point read per aligned read, no persisted state, no builder or replay changes.
Alternatives considered: a readstore deletion tombstone (persisted state that restore-rebuild and backfill replay must also write); preserving the projection for pinned readers (readers hold no lease while waiting for alignment and the wipe is required, so it needs generation-scoped keys).

## Risk

**LOW** — one extra point read per aligned read, on a key the caller already read; no snapshot is taken. A deletion applied after the projection snapshot is over-rejected with the answer every later read gives.

## Validation

- [x] Targeted tests: `internal/query` — `TestAlignedIndexSnapshotRejectsLedgerDeletedAfterPin` (deletion folded / not yet folded, lease released on rejection) and `TestExecute_LogsQueryRejectsLedgerDeletedAfterPin` commit the deletion and the projection wipe between the pinned handle and alignment; both are red with the gate removed (mutation-checked). `TestAlignedIndexSnapshotFailsLoudlyWhenPinnedLedgerRowVanishes` pins the absent-row branch as an invariant rather than a deletion (mutation-checked). `internal/storage/dal` — `TestReadHandle_LiveSeesLaterCommits`. Unit suites of `internal/query`, `internal/application/ctrl`, `internal/application/indexbuilder`, `internal/storage/readstore`, `internal/storage/dal` green.
- [x] Full suite / broader validation: e2e deletion, prepared-query, checkpoint and cross-store alignment specs green, also under `-race`; golangci clean on the touched packages

API: an aligned read admitted before a deletion of its ledger now returns `NotFound` instead of a successful page. No persisted state, no FSM/Raft or storage-format change. `AlignedIndexSnapshot` takes the main handle as `*dal.ReadHandle` plus the ledger name. `dal.ReadHandle.Live` is new: a getter over the store's current committed state, pinning nothing.

## Review focus

The ordering argument in `requireLedgerLive` (why a live re-read after the projection snapshot is sufficient), and `dal.ReadHandle.Live` reusing the handle's lifecycle lock rather than taking one of its own. `ListLogs` has no opener seam (`DefaultController` holds a concrete `*dal.Store`), so its deterministic coverage is the shared gate test; the prepared-query path is covered through its own seam.

## Known concerns

The pre-alignment ledger lookup in ctrl callers returns `commonpb.NewNotFoundError` while the gate returns `*domain.ErrLedgerNotFound`; same gRPC code, different error details depending on timing. Left for a separate cleanup.

Unrelated, pre-existing: the e2e spec `AccountHasAssetIndex … applied live (processLogs)` fails with `index not found: has asset` when run together with the checkpoint/inspect/aggregate specs, on release/v3.0 too; passes alone. Also pre-existing: the cross-store alignment spec's 32-log lag gate is satisfied in CI only by one round of 80 in-flight logs, not by real fold lag.

[EN-1991]: https://formance-team.atlassian.net/browse/EN-1991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/e2e/business/ledger_delete_pinned_logs_test.go
```diff
@@ -0,0 +1,242 @@
...
+				})
+				listCh <- outcome{logs: len(logs), err: err}
+			}()
+			go func() {
+				defer GinkgoRecover()
+
+				resp, err := actions.ExecutePreparedQuery(ctx, client, victim, preparedName, commonpb.QueryMode_QUERY_MODE_LIST, victimPageSize)
+				preparedCh <- outcome{logs: len(resp.GetCursor().GetLogData()), err: err}
+			}()
+
```
NumaryBot: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Confirm both reads are pinned before deleting the ledger**

Launching these goroutines does not establish that either RPC has reached the server before the deletion commits. The lag check only observes builder progress, not read admission. If deletion wins, both reads return the pre-existing NotFound response and `classify` accepts them, so this regression test can pass without exercising the post-pin deletion gate, even with that gate removed. Synchronize through a server-side admission/alignment seam and confirm both main snapshots are pinned before proposing deletion. This follows the branch-proof requirement in [AGENTS.md](AGENTS.md#L85).

NumaryBot (commented):

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2030#issuecomment-5666403742

docs/technical/architecture/subsystems/read-path/read-snapshot-consistency.md
```diff
@@ -178,3 +178,15 @@ it that way and returns `ErrIndexNotFound`; a record present at version 0
...
 client to wait for readiness that will never arrive is the failure this
 prevents.
+
+**A deleted ledger rejects the aligned read**: `DeleteLedger` folds
+as a wipe of every ledger-scoped projection row while the mainstore handle
+that admitted the read still holds the ledger live. Index leaves then fail
+readiness as removed, and the unfiltered LOGS universe — served from the
+projection alone — has nothing left to fail on: a deletion folded between the
+two acquisitions would answer a successful empty page for a ledger the pin
+holds logs for. Every aligned read is therefore gated on the ledger still
```
NumaryBot, line 188: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Scope the claimed `ErrLedgerNotFound` result to the alignment gate**

The gate returns `*domain.ErrLedgerNotFound`, but reads beginning after deletion can fail before alignment: controller paths such as `ListLogs` convert `query.GetLedgerByName`’s sentinel into `commonpb.NewNotFoundError`. Both map to gRPC `NotFound`, but their error identity/details differ. Say `NotFound`, or qualify the typed error as the gate’s result.

NumaryBot (commented):

docs/technical/audits/read-consistency-projections.md
```diff
@@ -67,7 +69,9 @@ when target-specific gates project it back to the main pin:
 - account metadata resolves event history at the pin and has-asset rows use
   first-touch stamps;
 - folded account membership retains the documented purged-address exception;
-- schema and `IndexVersionState` come from their owning pinned views.
+- schema and `IndexVersionState` come from their owning pinned views;
+- ledger liveness is re-read from a main snapshot taken after the projection
```
NumaryBot: 🟠 **[major] Describe the liveness read as a live point lookup**

`requireLedgerLive` does not take a main-store snapshot: `ReadHandle.Live` returns a getter whose `Get` directly calls `*pebble.DB.Get`. Calling this a snapshot incorrectly describes both the mechanism and its lifetime; document it as a post-projection live point lookup instead ([AGENTS.md:91-98]).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2030#issuecomment-5666403742

docs/technical/audits/read-consistency-projections.md
Azorlogh: Fixed in 6d65265e529c2e8d9ecb052f888ce46d5908ff73 — now "a live point lookup issued once the projection snapshot is open".

docs/technical/architecture/subsystems/read-path/read-snapshot-consistency.md
Azorlogh: Fixed in 6d65265e529c2e8d9ecb052f888ce46d5908ff73 — the typed error is now scoped to the gate, with the pre-alignment `commonpb.NewNotFoundError` path named alongside it.

Azorlogh:
@shipfox-ai the P3 is moot as of 6d65265e529c2e8d9ecb052f888ce46d5908ff73: `requireLedgerLive` no longer takes a snapshot, so the discarded `live.Close()` is gone. `ReadHandle.Latest` was replaced by `ReadHandle.Live`, a getter over current committed state that pins nothing — the review above describes the previous mechanism.

internal/query/aligned_snapshot.go
```diff
@@ -232,6 +244,48 @@ func AlignedIndexSnapshot(ctx context.Context, rs *readstore.Store, mainReader d
...
+// snapshot opened above. The builder folds a deletion only after the main
+// store applied it, so a projection snapshot holding the wipe is always
+// paired with a live view showing the ledger deleted; a ledger live here
+// proves the projection snapshot still holds every row of the pinned ledger.
+//
+// A deletion applied after the projection snapshot is rejected too, although
+// that snapshot could have been served: NotFound is the answer every later
+// read gives, so the read stays linearizable.
+func requireLedgerLive(mainReader *dal.ReadHandle, ledgerName string) error {
+	kb := dal.NewKeyBuilder()
```
NumaryBot: **🔵 [suggestion] Duplicated LedgerInfo row lookup in requireLedgerLive**

`requireLedgerLive` rebuilds the `[ZoneGlobal][SubGlobLedgerInfo][ledger\0]` key with `dal.NewKeyBuilder().PutZonePrefix(dal.ZoneGlobal, dal.SubGlobLedgerInfo).PutLedgerName(...)` and reads it via `dal.ReadProto`, duplicating the lookup `query.GetLedgerByName` already encapsulates (internal/query/ledger.go); a future change to the LedgerInfo key layout must be made in two places. Extract a shared `readLedgerInfoRow(getter, name) (*commonpb.LedgerInfo, error)` helper and keep the nil-means-invariant / deleted-means-`ErrLedgerNotFound` interpretation at each call site. (provenance: standards review, verified against the frozen diff)

NumaryBot (commented):
Findings

internal/query/aligned_snapshot.go
Azorlogh: Fixed in d560672cbdc2bbb36959da90d95c6288126f20c4 — extracted `readLedgerInfoRow` in `ledger.go`; `GetLedgerByName` and `requireLedgerLive` now share it and keep their own nil/deleted interpretation.

internal/storage/dal/reader.go
```diff
@@ -113,7 +114,30 @@ func (s *Store) NewDirectReadHandle() (*ReadHandle, error) {
...
+// It reads under the lifecycle lock this handle already holds and takes none
+// of its own. A second NewReadHandle from the handle's owner would take a
+// second RLock, and a queued writer (Close, RestoreCheckpoint) blocks new
+// readers while it waits for this handle: both sides would wedge for good.
+func (h *ReadHandle) Live() PebbleGetter {
+	return liveGetter{db: h.db}
+}
+
+// liveGetter serves point lookups from a store's current committed state. It
+// pins nothing, so there is no resource to release beyond each lookup's own
```
NumaryBot, line 145: **🟡 [minor] `Live()` getter outlives the handle unsafely**

`Live()` returns a `liveGetter` over `*pebble.DB` with no lifetime tie to the `ReadHandle`; a caller retaining it past `Close()` can read a closed DB, the panic class this file documents (EN-2072). The only current caller, `requireLedgerLive`, uses it within the handle's lifetime, so impact is latent. Consider binding the getter to the handle's `mu` or stating the handle-scoped contract on the return type.

internal/query/aligned_snapshot.go
```diff
@@ -232,6 +244,45 @@ func AlignedIndexSnapshot(ctx context.Context, rs *readstore.Store, mainReader d
...
+	// stamps DeletedAt — so a caller that pinned this ledger live cannot
+	// outlive its row.
+	if live == nil {
+		assert.Unreachable("query: pinned ledger has no LedgerInfo row after alignment", map[string]any{
+			"ledger": ledgerName,
+		})
+
+		return fmt.Errorf("invariant: ledger %q was pinned live but has no LedgerInfo row", ledgerName)
+	}
+
```
NumaryBot, line 278: **🟡 [minor] Vanished-row invariant surfaces as `Unknown`, not a typed internal error**

`requireLedgerLive` returns a plain `fmt.Errorf("invariant: ...")` for the impossible absent-row case, which the gRPC adapter's default branch sanitises to `codes.Unknown`, while the deletion path uses the typed `domain.ErrLedgerNotFound`. A typed `KindInternal` Describable would give the wire a deterministic, correctly classified status. The loud fail-not-NotFound behaviour itself matches the spec.

NumaryBot (commented):
Findings

internal/query/aligned_snapshot_lifecycle_test.go
```diff
@@ -0,0 +1,194 @@
...
+	require.Equal(t, pinnedSeq, mainSeq)
+}
+
+// The prepared LOGS path pins the ledger and the definition on the main
+// handle, then aligns. A deletion committed and folded in between must
+// surface as ErrLedgerNotFound, never as a successful empty cursor.
+func TestExecute_LogsQueryRejectsLedgerDeletedAfterPin(t *testing.T) {
+	t.Parallel()
+
+	for _, tc := range []struct {
```
NumaryBot, line 146: **🔵 [suggestion] Rejection coverage ends at the gate and the prepared-LOGS caller**

The spec requires rejection on every aligned read path (`ListLogs`, `listEntities`, `AggregateVolumes`, `InspectIndex`, prepared queries); all satisfy it via the single shared `requireLedgerLive` gate, but the new tests exercise only the gate and `Execute` directly. One caller-level case per remaining controller entry point would pin the requirement where the PR states it.

internal/storage/dal/reader.go
Azorlogh: Fixed in cbcd2677cfc32fb4d0d7d92722b2200998ccf739 — `Live()` and `liveGetter` now state the handle-scoped contract explicitly: borrow the lock, do not retain past `Close`, same rule as a retained iterator.

internal/query/aligned_snapshot.go
Azorlogh: Not changing this. The adjacent pre-existing invariant in the same function (the reclaim-floor guard) returns a plain `fmt.Errorf("invariant: ...")` too, so this matches local convention; typing only mine would leave the file inconsistent, and typing both means a new wire reason constant for branches unreachable by contract — out of scope for EN-1991. Worth a follow-up covering every read-path invariant at once.

internal/query/aligned_snapshot_lifecycle_test.go
Azorlogh: Not adding these. `ListLogs`, `listEntities`, `AggregateVolumes` and `InspectIndex` all open their handle from the concrete `ctrl.store`, so there is no seam to commit the deletion between the pin and alignment — soft-deleting beforehand makes the pre-alignment lookup reject instead, which exercises a different path. `Execute` is testable precisely because it takes a `queryHandleStore`. The PR body pre-declares this; the shared gate is the coverage point.

internal/storage/dal/reader.go
```diff
@@ -113,7 +120,37 @@ func (s *Store) NewDirectReadHandle() (*ReadHandle, error) {
...
+// It reads under the lifecycle lock this handle already holds and takes none
+// of its own. A second NewReadHandle from the handle's owner would take a
+// second RLock, and a queued writer (Close, RestoreCheckpoint) blocks new
+// readers while it waits for this handle: both sides would wedge for good.
+//
+// The getter is scoped to the handle and borrows its lock without extending
+// it. Use it within the handle's lifetime and do not retain it past Close:
+// afterwards the lock is gone and the lookup races a DB that Close or
+// RestoreCheckpoint may already have closed (EN-2072), exactly as a retained
+// iterator would.
```
NumaryBot, line 139: **🔵 [suggestion] Live() getter's lifetime safety rests on caller discipline**

`Live()` returns a `liveGetter` that borrows the handle's lifecycle lock with no runtime guard, so retaining it past `ReadHandle.Close()` silently races a closed DB — the EN-2072 failure class its own doc comment warns about. The sole current caller `requireLedgerLive` uses it inline, so impact is latent. Consider a scoped form (e.g. `h.WithLive(func(PebbleGetter) error)`) that makes the borrow impossible to outlive.

NumaryBot (commented):
Findings

internal/storage/dal/reader.go
Azorlogh: Same finding as #discussion_r4039183280, already answered — declining again. `ReadHandle` states this contract for everything it hands out: "The handle does not track what it hands out: every iterator and every Get closer obtained through it must be closed BEFORE Close()" (reader.go:68). `NewIter` and `Get` carry the identical retention hazard and are not wrapped in scoped callbacks. `Live()` is not a new hazard class, so a scoped form only here would be inconsistent; applying it to `NewIter` too is a `dal` refactor well outside EN-1991.

internal/storage/dal/reader.go
```diff
@@ -113,7 +120,37 @@ func (s *Store) NewDirectReadHandle() (*ReadHandle, error) {
...
+func (h *ReadHandle) Live() PebbleGetter {
+	return liveGetter{db: h.db}
```
NumaryBot, line 141: **🔵 [suggestion] Live() hands out a getter whose lifetime is enforced only by comments**

`(*ReadHandle).Live()` returns an unowned `liveGetter` that is only safe for as long as the handle lives; retention past `Close()` races a DB that `Close`/`RestoreCheckpoint` may have closed — the same class as the EN-2072 panic documented in the `PebbleGetter` contract. Impact is latent today because the sole caller (`requireLedgerLive`) uses it inline. Consider keeping the lookup inside the handle (e.g. a `LiveGet` method) so the borrowed lifetime is structural rather than comment-enforced.

internal/application/indexbuilder/pinned_logs_lifecycle_test.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/formancehq/ledger/v3/internal/query"
+)
+
+// Observe the projection progress wait, after ListLogs has opened its main
+// snapshot and checked ledger existence. This follows checkpointWaitObservedContext
+// in the gRPC checkpoint tests and needs no production synchronization hook.
+type pinnedLogsWaitContext struct {
+	context.Context
+
+	once    sync.Once
```
NumaryBot, line 27: **🔵 [suggestion] Alignment-wait seam triggers on any first Done() consumer**

`pinnedLogsWaitContext.Done()` closes `waiting` on the first `Done()` call from any code path, and the test commits the ledger deletion as soon as it fires, assuming the alignment wait is the first consumer. A future `ctx.Done()` consumer earlier in the `ListLogs` path would commit the deletion before the read pins its snapshot, turning the straddling-window assertion into a flaky pre-alignment rejection. Consider gating the deletion on a signal that proves the read reached the alignment wait rather than any `Done()` call.

NumaryBot (commented):
Findings

internal/storage/dal/reader.go
Azorlogh: stale — third post of the same finding: #discussion_r4039183280 (🟡) and #discussion_r4039294876 (🔵), both answered and resolved.

Also note `LiveGet` conflicts with your own earlier finding #discussion_r4039079166: `Live()` returns a `PebbleGetter` precisely so `requireLedgerLive` can share `readLedgerInfoRow` instead of rebuilding the LedgerInfo key. A key-less `LiveGet` means either a local adapter (same escapable object) or duplicating that lookup again.

internal/application/indexbuilder/pinned_logs_lifecycle_test.go
Azorlogh: Fair point, but the degradation is loud, not silent: if the deletion landed before the pin, `ListLogs` rejects at the pre-alignment lookup with `*commonpb.NotFoundError`, which fails the `*domain.ErrLedgerNotFound` assertion at line 177. The early-completion `t.Fatalf` at line 127 covers the other ordering. So a future earlier `Done()` consumer turns this red, not green-but-vacuous.

Leaving as is for now; worth hardening if someone adds such a consumer.

PR #2031, opened 2026-09-14
fix(transport): fence stale peer monitors

gfyrag:
## What changed

Fence optional-TLS connection-monitor restarts by the exact peer entry that the monitor captured. Add a deterministic regression that pauses the old monitor after its final failure check, installs a replacement under the same peer ID, and proves the replacement remains current.

## Why

A cancelled monitor could previously call `RestartConnection(id)` after replacement and close or delete the successor connection.

Jira: [EN-2054](https://formance-team.atlassian.net/browse/EN-2054)

Need: preserve healthy inter-node connectivity while peers are replaced during optional-TLS migration.
Current limitation: an obsolete monitor was authorized only by a reusable peer ID and could tear down its successor.
Requirement / constraint: a monitor may restart or remove only the exact connection entry it observed.
Evidence: EN-2054 and the deterministic regression in `internal/infra/transport/connection_pool_probe_test.go`.
Durable repository evidence: `docs/technical/architecture/subsystems/consensus/raft-consensus.md` and `docs/technical/audits/concurrency-lifecycle-shutdown.md`.

## Technical decision

Decision: capture the `peerEntry` in each monitor and compare its identity with the current map entry under the pool mutation lock before restarting.
Why now / why proportionate: the pointer identity already models ownership and requires no new generation counter or protocol state.
Alternatives considered: checking cancellation before restart leaves another race window; joining each cancelled monitor would add blocking lifecycle coupling to every replacement.

## Risk

**LOW** — the change is local to optional-TLS monitoring and preserves explicit `RestartConnection` behavior.

## Validation

- [x] `go test ./internal/infra/transport -run TestConnectionPool_StaleOptionalMonitorDoesNotRestartReplacement -count=50`
- [x] `go test -race ./internal/infra/transport -count=1`

The requirement is satisfied when releasing the obsolete monitor after replacement leaves the replacement connection open and current.

Optional-TLS monitor callbacks are now fenced by local resource-entry identity. No API, protocol, persistence, or compatibility change.

## Review focus

Please verify the entry-identity check is performed atomically with restart teardown, and that the regression reaches the stale-monitor interleaving.

## Known concerns

None.

[EN-2054]: https://formance-team.atlassian.net/browse/EN-2054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gfyrag:
Addressed both Shipfox convention comments in 6d571e0d1: restored an explicit justification for the intentionally ignored re-dial error, corrected the monitor ownership comment, and added `t.Parallel()` to the deterministic regression. Revalidated the focused test (50x), transport race suite, `scripts/agent-check`, and `scripts/agent-check-pr`.

paul-nicolas (approved):

PR #2036, opened 2026-09-14
fix(query): reject aligned reads of a deleted ledger (EN-1991)

gfyrag:
## What changed

All aligned index reads now verify that the ledger is still live immediately after acquiring the projection snapshot, using the caller existing read handle to take a fresh point-in-time view of the main store. Any ledger deleted after the read main-store pin is rejected with ErrLedgerNotFound before compilation, across every aligned path: ListLogs, prepared-query execution, filtered account/transaction lists, AggregateVolumes, and InspectIndex.

This replaces the previous approach (persisted readstore lifecycle records per ledger) which was correctly identified as more complex than needed and LOGS-only.

## Why

Fixes EN-1991. A LOGS read pinned to a main snapshot with the ledger live then acquired a projection snapshot that had already folded DeleteLedger, wiping every ledger-scoped projection row. The unfiltered LOGS universe returned a successful empty page instead of an explicit rejection.

## Technical decision

Post-alignment liveness re-read on a main snapshot strictly after the projection snapshot. The builder folds a deletion only after the main store applied it, so a projection snapshot holding the wipe is always followed by a main snapshot showing the deletion; a ledger live there proves the projection snapshot holds every pinned row. The snapshot is taken through the caller existing handle (dal.ReadHandle.Latest) to reuse the lifecycle lock already held.

## Risk

**LOW** - one extra Pebble snapshot and point read per aligned read; no persisted state, no builder or replay changes.

## Validation

- bash scripts/agent-check PASS
- Targeted race tests on internal/query, /ctrl, /indexbuilder, /storage/dal, /storage/readstore - all PASS
- CI run 35222501497 fully green on HEAD d15913573


gfyrag:
Shipfox findings processed on HEAD `c3801edeb`.

1. **Upgrade/backfill:** pushed back. This repository’s authoritative `AGENTS.md` states that Ledger v3 is unreleased, state is wiped on pushes to `release/v3.0`, and explicitly forbids migrations, compatibility shims, version guards, or fallback paths for older v3 storage formats. A boot-time backfill or readstore format migration would therefore violate the active project contract. The missing-record rejection is intentional for malformed/incomplete projections; normal v3 deployment rebuilds the projection from history and writes lifecycle records while folding `CreateLedger`.

2. **Same-name recreation coverage:** fixed in commit `c3801edeb`. `TestListLogsRejectsProjectionThatDeletedPinnedLedger` now exercises both the inactive tombstone and an active projection with a different ledger ID through the production `ListLogs` caller.

`bash scripts/agent-check` and the focused controller regression pass after the change.

gfyrag:
Shipfox finding on `f0e29dfb0` is not applicable to this repository's supported lifecycle.

The authoritative root `AGENTS.md` states under **Release status — v3 is unreleased** that state is wiped on pushes to `release/v3.0`, there is no backward-compatibility burden between v3 development revisions, and agents must not add migrations, compatibility shims, version guards, or fallback paths for old v3 storage formats. A normal supported deployment therefore rebuilds `read-indexes` from sequence 0 and folds every `CreateLedger`, populating `0x0A`; pre-change persistent readstores and pre-change query checkpoints are explicitly outside the supported contract.

Adding the proposed boot repair, targeted refold, or format-version migration would violate that explicit project rule. The missing-record branch remains intentional fail-loud behavior for an incomplete current-revision projection. No code change is warranted for this finding.

gfyrag:
@shipfox-ai Please re-review the current HEAD `f0e29dfb0f12e94b13b7e771f07bcd2129a632fe`, including the repository-level `AGENTS.md` release-status contract and the response above. Confirm explicitly whether any applicable change request remains under the supported unreleased-v3 lifecycle.

Azorlogh:
Heads-up first: I accidentally opened #2030 for EN-1991 before spotting this one — near-identical scope. Happy to close mine if we go with this implementation, but two things I'd want to settle before deciding which shape we keep.

**1. `lifecycle.ID != info.GetId()` is unreachable.** `processCreateLedger` rejects a create under a soft-deleted name with `ErrLedgerDeleted`, so delete + same-name recreate can't happen — `rebuild.go:251` records the same. "Would miss delete+recreate" is the stated reason for persisting `(ID, active)` over a presence marker, so the new keyspace may not be earning itself.

**2. The gate is LOGS-only, but the bug isn't.** `listEntities` compiles filtered ACCOUNTS/TRANSACTIONS candidates from the same aligned snapshot, so a folded `DeleteLedger` erases those universes identically and still returns a successful empty page. `AggregateVolumes` and `InspectIndex` share the pattern.

**3.** `ReadLedgerLifecycle` returns a plain error on a corrupt value — invariant 7 and the adjacent history tracker (`historyReplayInvariantf`) fail loudly instead.

**4.** `0x0A`'s deliberate absence from `ledgerScopedPrefixes` isn't noted in `delete_ledger.go`, which documents exactly that exemption for `SubInternalLedgerHistory`.

**5.** All four new subtests assert `ErrLedgerNotFound` — an unconditionally-rejecting `ValidateLedgerProjection` passes every one. Nothing pins the accept direction.

**6.** `ValidateLedgerProjection` allocates its own `KeyBuilder`; both call sites build one two lines later.

**7.** It'd sit more naturally in `ledger.go` than in the alignment file.

PR #2037, opened 2026-09-14
fix(usage): avoid recounting persisted zero volumes

gfyrag:
## What changed

Distinguish persisted zero-valued volume rows from absent or purged cells. Purge and transient cleanup now retain cache tombstones, so coverage-gated reads remain cache hits while row-existence classification uses `Old.IsDefined()`. Add deterministic unit, oracle, and E2E coverage for a zero-valued commit followed by a non-zero commit and for an ephemeral-to-normal policy change before cache rotation.

## Why

Fixes [EN-2051](https://formance-team.atlassian.net/browse/EN-2051). A normal persisted `{0,0}` volume was indistinguishable from a purged/absent zero cell, so every later commit re-emitted `NewKeptVolumes` and overcounted `CounterVolume`.

Need: `GetLedgerStats.VolumeCount` must equal live persisted volume-key cardinality once usage processing catches up.
Current limitation: a valid zero-amount transaction creates normal rows that later transactions recount.
Requirement / constraint: distinguish real persisted zero rows without changing cache coverage or deterministic apply.
Evidence: EN-2051 and `docs/technical/architecture/subsystems/usage/counters.md`.
Durable repository evidence: the updated usage counter contract, oracle model, and E2E regressions.

## Technical decision

Decision: represent absent purged/transient cells with the existing cache tombstone format and classify a new persistent row only when the prior value is undefined.
Why now / why proportionate: this preserves cache-hit admission and snapshot/restore behavior without adding a persistence format or parallel presence bit.
Lifecycle clarification: a defined `{0,0}` row that later matches a transient policy is a persisted row and is purged; a tombstoned row remains steady-state transient.

## Risk

**MEDIUM** — the change affects cache representation for ephemeral/transient cleanup and usage annotations, with state, oracle, and end-to-end regressions covering both persistence-policy transitions.

## Validation

- [x] Focused state, usagebuilder, oracle, and EN-2051 E2E tests
- [x] GitHub Actions workflow `34869850883` after full rerun

No persisted primary-store format, API, or gRPC protocol change. FSM reads remain coverage-gated and deterministic. Cache tombstones preserve absence across policy changes and cache snapshot/restore; the oracle mirrors the same map-membership semantics.

## Review focus

Please verify tombstone handling in purge/transient cleanup, new-volume classification, and oracle parity.

## Known concerns

None.

[EN-2051]: https://formance-team.atlassian.net/browse/EN-2051


internal/infra/state/write_set_ephemeral_purge.go
```diff
@@ -87,22 +94,22 @@ func (b *WriteSet) partitionVolumes(
...
 			// set, so the gated read still returns it with its types intact.
-			result.kept = append(result.kept, update)
+			keep(update, !update.Old.IsDefined())
 
 			continue
 		}
 
 		compiled := entry.compiled
 
 		if len(compiled) == 0 {
```
NumaryBot, line 104: 🟠 **[major] Preserve new-row counting after removing a persistence policy**

If an ephemeral volume is purged, its cache entry remains defined as `{0,0}` through `zeroVolumeCache`. `processRemoveAccountType` allows removing that account's type without clearing those entries. A subsequent transaction under the default normal policy therefore writes a new persistent row but this check omits it from `newKept`, permanently undercounting `VolumeCount` once usage catches up. The same path exists after removing a transient policy. Distinguish persisted zeros from retained absence placeholders across policy changes, and add a regression covering purge, type removal, and recreation before cache rotation.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2037#issuecomment-5666847439

tests/e2e/business/stats_test.go
```diff
@@ -73,6 +73,42 @@ var _ = Describe("GetLedgerStats", Ordered, func() {
...
+			Expect(err).To(Succeed())
+		})
+
+		It("Should count each persisted volume only once", func() {
+			Eventually(func(g Gomega) {
+				resp, err := sharedClient.GetLedgerStats(sharedCtx, &servicepb.GetLedgerStatsRequest{
+					Ledger: ledgerName,
+				})
+				g.Expect(err).To(Succeed())
+				g.Expect(resp.TransactionCount).To(Equal(uint64(2)))
```
NumaryBot, line 100: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Wait for usage processing before asserting the final volume count**

If the usage builder has processed only the zero-valued transaction, this poll can already succeed: `TransactionCount` comes from primary-store boundaries in `GetLedgerStats`, whereas `VolumeCount` comes from the asynchronous usage store. Thus `TransactionCount == 2` does not prove usage has caught up, and the test can finish without checking the final counter value. Also require `PostingCount == 2` in this callback; it shares the usage snapshot with `VolumeCount`. The subsequent log assertion checks annotations, not whether usage processing produced the correct final count.

NumaryBot (commented):

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2037#issuecomment-5666847439

internal/infra/state/write_set_ephemeral_purge.go
gfyrag: Done in 7a004fc4f. Purged/transient cache entries are now tombstones, so a later normal-policy recreation is classified as new even before cache rotation. The added E2E covers purge, type removal, normal type recreation, and the subsequent NewKeptVolumes/VolumeCount update.

tests/e2e/business/stats_test.go
gfyrag: Done in 7a004fc4f. The regression now waits for PostingCount == 2 alongside VolumeCount, so the assertion cannot pass before the asynchronous usage snapshot has caught up.

gfyrag:
Addressed Shipfox’s blocking finding in ca420bff6: the oracle now classifies new volume rows by persistent-map membership rather than numeric zero, with an additive regression for a zero-valued commit followed by a later non-zero commit. I also made every production branch call isNewVolumeUpdate consistently. Full oracle tests and scripts/agent-check pass.

gfyrag:
@shipfox-ai Please re-review the current HEAD `ca420bff6cc5acc20afac66ecfde6f7ee7d28bf6`. The blocking oracle finding is addressed by map-membership row-existence classification plus `TestGlobalState_Apply_PersistedZeroVolumeIsNotNewAgain`; every production branch now uses `isNewVolumeUpdate` consistently. Confirm explicitly whether any applicable change request remains.

paul-nicolas (dismissed):

gfyrag:
@NumaryBot Please review the current HEAD `115b44b7d5f57e81d0035b4ba6c72da0ad070284`. The latest delta renames cache `Delete` to `Tombstone`, removes the individual-eviction API, and makes the operation explicitly write a tombstone that remains a cache hit across the bounded admission/apply horizon. Targeted tests, baseline validation, the full GitHub test matrix, coverage, and repository invariants pass. Please treat prior reviews as applying only to their previous HEADs.

docs/technical/architecture/subsystems/storage/storage.md
```diff
@@ -225,7 +225,7 @@ Each row is keyed by generation and cache type:
...
 
-Tombstones are exactly the 9-byte header with no payload -- the pre-delete value is deliberately not marshalled, since `AttributeCache.Del` only flips the `Deleted` flag and keeps the payload in memory, which would otherwise resurrect the entry on restore. On the read path `parseLeanValue` fails closed on any malformed row (short buffer, unknown flag byte, or a tombstone carrying trailing bytes) rather than restoring partial or forged cache state (EN-1527).
+Tombstones are exactly the 9-byte header with no payload -- the pre-delete value is deliberately not marshalled, since `KeyStore.Tombstone` only flips the `Deleted` flag and keeps the payload in memory, which would otherwise resurrect the entry on restore. On the read path `parseLeanValue` fails closed on any malformed row (short buffer, unknown flag byte, or a tombstone carrying trailing bytes) rather than restoring partial or forged cache state (EN-1527).
```
NumaryBot: 🟡 **[minor] Correct tombstone payload documentation**

`KeyStore.Tombstone` clears `Data` by writing a zero-value `Entry`; it does not retain the pre-delete payload. Update this explanation (and the matching source comment) so the documented recovery state matches the implementation. [AGENTS.md:93-99]

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2037#issuecomment-5666847439

gfyrag:
@NumaryBot Please review the current HEAD `2ed4fbe7b754df15d43159a7e7d83312a0c43869`. This is the same EN-2051 change rebased onto `1f86c44ac` (release/v3.0), with one test assertion corrected: the oracle test for persisted zero volumes now uses the actual `address:asset:color:` format produced by renderTouchedVolumes. agent-check, cache/attributes/state/admission/mirror/oracle tests all pass under race. Please treat prior reviews as applying only to their previous HEADs.

docs/technical/architecture/subsystems/storage/storage.md
gfyrag: Done — corrected in commit 6f9f2db7b. The doc now states that KeyStore.Tombstone writes Entry{Tag: tag, Data: zero, Deleted: true}, clearing the payload before the tombstone reaches disk rather than keeping it in memory. This commit also fixes the oracle test assertion to match the actual address:asset:color: format produced by renderTouchedVolumes.

gfyrag:
@NumaryBot Please review the current HEAD `6f9f2db7b6d6a41794cb50f59cc247baa062b382`. This amends the previous commit with two corrections: (1) the oracle test assertion now uses the actual address:asset:color: format produced by renderTouchedVolumes, and (2) storage.md correctly documents that KeyStore.Tombstone clears Data to the zero value rather than keeping the payload. All CI checks pass on 2ed4fbe7b (identical diff minus these two doc/test fixes). Please treat prior reviews as applying only to their previous HEADs.

PR #2043, opened 2026-09-15
fix(backup): rebuild next ledger ID on restore

gfyrag:
## What changed

Rebuild the monotonic ledger-ID allocator from the checkpoint seed and every post-checkpoint `CreatedLedgerLog` during incremental restore. Preserve allocated IDs across later ledger deletion, and fail closed if the ID space is exhausted.

Add unit and real S3 restore regressions proving a non-empty delta cannot cause restored or newly created ledgers to share an ID. Document the restore fold and residual staging state, and keep the persistence audit contract synchronized.

## Why

Fixes [EN-2057](https://formance-team.atlassian.net/browse/EN-2057).

A checkpoint followed by a delta ledger creation restored the ledger row but left `SubGlobNextLedgerID` at its checkpoint value. The first creation after restore could therefore reuse the restored ledger's numeric ID.

Need: Restoring a full checkpoint plus a non-empty incremental delta must preserve unique, monotonic ledger identities and the same next-create behavior as uninterrupted execution.

Current limitation: `RebuildDelta` reconstructed post-checkpoint ledger rows without advancing their durable allocator.

Requirement / constraint: The checkpoint allocator is the fold seed; every delta creation reserves its ID permanently, including when that ledger is later deleted. A failed rebuild must never activate its staging store.

Evidence: EN-2057 and `docs/technical/architecture/subsystems/backup/incremental-restore-contract.md`.

Durable repository evidence: `docs/technical/architecture/subsystems/backup/README.md` and `docs/technical/audits/persistence-restore-replay.{json,md}`.

## Technical decision

Decision: Fold `nextLedgerID = max(checkpointNextLedgerID, createdLedgerID + 1)` from authoritative `CreatedLedgerLog` entries, and persist the result in the final replay batch. Reject `MaxUint32` rather than wrap.

Why now / why proportionate: This mirrors the existing live allocation semantics with one deterministic maximum fold and reuses the state writer used by live apply.

Alternatives considered: Deriving the counter during FSM recovery would scan projections outside the restore fold and would not preserve allocation across deleted ledgers. Leaving the counter unchanged reproduces the collision.

## Risk

**MEDIUM** — changes a durable allocator during restore, bounded to the incremental replay path and covered through unit, race, and real S3 restore tests.

## Validation

- [x] `AI_REVIEW_BASE_SHA=847ead022ee25b8b5c8414ef9777bbfb1c075592 bash scripts/agent-check-pr`
- [x] `go test -race ./internal/infra/backup ./internal/infra/state -count=1`
- [x] Focused unit regression showed fail-then-success (`actual 2`, expected `5` before the fold)
- [x] `go test -v -race -tags 'e2e,s3' ./tests/e2e/cluster -run TestCluster -ginkgo.focus 'Restore' -ginkgo.v -count=1 -timeout 20m` (78 passed, 0 failed)

The durable `SubGlobNextLedgerID` projection is now classified as rebuilt during incremental restore. Its checkpoint value seeds the fold, delta creation logs advance it, deletion never rewinds it, and restore activation remains gated on successful rebuild/validation/finalization.

No API, protobuf, or service-protocol change.

## Review focus

Please review the allocator fold seed/overflow behavior and the evidence that the S3 restore scenario exercises a non-empty post-checkpoint delta before creating two new ledgers.

## Known concerns

None.

[EN-2057]: https://formance-team.atlassian.net/browse/EN-2057?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gfyrag:
@shipfox-ai Please review the current HEAD `21dea52a12e6a1e01c5cc6f3d8b97838e8124a79` for the EN-2057 incremental-restore allocator fix. Confirm explicitly whether any applicable change request remains.

gfyrag:
@shipfox-ai Please review the current HEAD `0c339be7f3c9c6ab377639ab746a510284fa584e`; the only follow-up adds regression coverage for the existing exhausted-ID safety branch. Confirm explicitly whether any applicable change request remains.

internal/infra/backup/rebuild.go
```diff
@@ -217,6 +229,13 @@ func RebuildDelta(
 			// MirrorSource, AccountTypes, and DefaultEnforcementMode, all part of
 			// the stored projection. ToLedgerInfo copies every creation-time field.
 			info := p.CreateLedger.ToLedgerInfo()
+			if info.GetId() == math.MaxUint32 {
```
paul-nicolas: [P1][blocking] Align exhausted-ID restore behavior with live apply

Live ledger creation accepts `math.MaxUint32` through `IncrementNextLedgerID()` and then wraps the `uint32` allocator. This branch instead rejects the corresponding committed `CreatedLedgerLog` during restoration. The source cluster can therefore produce and back up a history that this code cannot restore, violating the incremental restore parity contract.

Please either enforce exhaustion consistently in the live apply/admission path or reproduce the live fold during rebuild, and add cross-lifecycle coverage for the chosen behavior.

paul-nicolas (changes requested):
One blocking restore-parity issue needs to be addressed before merge.

internal/infra/backup/rebuild.go
```diff
@@ -228,6 +240,13 @@ func rebuildDelta(
...
```
NumaryBot: 🔴 **[blocker] Align exhausted-ID restore behavior with live apply**

When the source allocator reaches `math.MaxUint32`, `processCreateLedger` still accepts the creation: both `orderOverlayScope.IncrementNextLedgerID` and `WriteSet.IncrementNextLedgerID` increment without an exhaustion check. This new branch therefore rejects a committed history that the source can successfully produce and back up, violating [AGENTS.md:56](AGENTS.md#L56). Enforce exhaustion consistently in live creation and restore, and add cross-lifecycle coverage; the new synthetic rejection test only verifies the restore side.

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2043#issuecomment-5676499278

internal/infra/backup/rebuild.go
gfyrag: Done — fixed in commit d3e018d65. `CheckedNextLedgerID` now enforces the same exhaustion guard in both `processCreateLedger` (live creation) and `rebuildDelta` (restore replay). `TestProcessCreateLedger_RejectsExhaustedLedgerID` and `TestRebuildDelta_RejectsExhaustedLedgerIDWithoutCommit` provide cross-lifecycle coverage for the chosen behavior.

internal/infra/backup/rebuild.go
gfyrag: Done — fixed in commit d3e018d65. Added `CheckedNextLedgerID` in `internal/domain/sequence.go` that rejects `math.MaxUint32`, then wired it into `processCreateLedger` before `IncrementNextLedgerID` (live path) and into `rebuildDelta` before advancing `nextLedgerID` (restore path). Both lifecycles now enforce the same exhaustion boundary and surface `ErrSequenceExhausted{Counter: SequenceCounterLedgerID}`. Cross-lifecycle tests added: `TestProcessCreateLedger_RejectsExhaustedLedgerID` covers the live path and `TestRebuildDelta_RejectsExhaustedLedgerIDWithoutCommit` covers the restore path with rollback verification.

internal/domain/sequence.go
```diff
@@ -25,3 +26,15 @@ func CheckedNextSequence(current uint64, counter SequenceCounter) (uint64, *ErrS
 
 	return current + 1, nil
 }
+
+// CheckedNextLedgerID returns the next uint32 ledger identifier without
+// allowing modular wrap. MaxUint32 is deliberately not allocatable because the
+// persisted counter stores the next value; accepting it would reset that value
+// to zero and permit identifier reuse.
+func CheckedNextLedgerID(current uint32) (uint32, *ErrSequenceExhausted) {
```
NumaryBot, line 44: 🟡 **[minor] Document the uint32 ledger-ID exhaustion contract**

`CheckedNextLedgerID` now returns `ErrSequenceExhausted` for a `uint32` allocator, but the error type comment still says it only handles `uint64` counters and the FSM sequence-exhaustion documentation likewise omits ledger IDs. Update those descriptions so the new terminal behavior is accurately documented, as required for behavior changes by [AGENTS.md:93-99](AGENTS.md#L93-L99).

NumaryBot (commented):

Summary: https://github.com/formancehq/ledger/pull/2043#issuecomment-5676499278

internal/domain/sequence.go
```diff
@@ -8,6 +8,7 @@ type SequenceCounter string
 
 const (
 	SequenceCounterTransactionID SequenceCounter = "transactionId"
+	SequenceCounterLedgerID      SequenceCounter = "ledgerId"
```
NumaryBot, line 11: <sub>OpenCodeReview · independent Codex pass · comment-only</sub>

🟠 **[major] Document the new public ledger-ID exhaustion counter**

This new counter is returned as `counter: "ledgerId"` with the public `SEQUENCE_EXHAUSTED` error, but `docs/technical/contributing/api-comparison.md` exhaustively lists the allowed counter values and omits `ledgerId`. Update that API contract (and the related uint64-only exhaustion descriptions) so documented behavior matches HEAD.

NumaryBot (commented):

internal/domain/sequence.go
gfyrag: Done — fixed in commit a6dfacc54. Updated the `ErrSequenceExhausted` comment in `internal/domain/errors.go` to say "integer counter" and explicitly name both `CheckedNextSequence` (uint64) and `CheckedNextLedgerID` (uint32), so the new terminal behavior is accurately documented for both allocator widths.

internal/domain/sequence.go
gfyrag: Done — fixed in commit a6dfacc54. Added `ledgerId` to the `counter` value list for the `SEQUENCE_EXHAUSTED` error row in `docs/technical/contributing/api-comparison.md` line 986, so the API contract now matches the values returned by the `CheckedNextLedgerID` code path.

gfyrag:
Addressed in commits a6dfacc54 and 7ba58af8e.

**Finding 1 (minor) — saveNextLedgerIDForBackupTest duplicates StoreNextLedgerID:** Fixed in 7ba58af8e. The helper now calls `state.StoreNextLedgerID(batch, nextID)` directly, removing the hand-rolled key/encoding that could silently diverge from production.

**Finding 2 (minor) — comment overstates atomicity:** Fixed in 7ba58af8e. Updated the comment to reflect that the write goes into "whatever batch is current at the end of the loop" (flushed every 5000 logs), not necessarily the same batch as the last ledger row. The functional guarantee is unchanged: a failed rebuild never activates its staging store, and the counter is always monotonic.

**Finding 3 (suggestion) — merge GetNextLedgerID/IncrementNextLedgerID into a scope-level checked allocator:** Not addressed in this PR. The two-step call mirrors the existing `CheckedNextSequence` pattern used throughout the codebase (e.g. in `processApply`, `processTransaction`). Merging them would require changing the `Scope` interface, which is out of scope for this bug fix. The current code is correct.

NumaryBot (commented):
Findings

internal/infra/backup/rebuild.go
```diff
@@ -538,6 +557,18 @@ func rebuildDelta(
 		}
 	}
 
+	// Persist the allocator. The counter is written to whatever batch is current
+	// at the end of the loop (batches are flushed every 5000 logs); a successful
```
NumaryBot, line 561: Rebuilt `SubGlobNextLedgerID` has no independent checker pass

The PR classifies the ledger-ID allocator as a rebuilt projection and inventories it in the persistence audit, but `CheckStore` (`internal/application/check/checker.go`) re-derives no expectation for it, so a restore that folds the allocator to a wrong value still checks clean. `docs/technical/architecture/subsystems/backup/incremental-restore-contract.md` question 6 asks which checker pass independently validates each rebuilt value. Add a checker pass comparing the stored counter to `max(CreatedLedgerLog id)+1` over the replayed history, or document the exemption.

internal/domain/sequence.go
```diff
@@ -25,3 +26,15 @@ func CheckedNextSequence(current uint64, counter SequenceCounter) (uint64, *ErrS
 
 	return current + 1, nil
 }
+
```
NumaryBot, line 29: `CheckedNextLedgerID` name hides dual call semantics

Live callers (`processCreateLedger`) pass the persisted next-ID counter and discard the return, while `rebuild.go` passes an already-allocated ID and uses the return as the post-allocation counter, so the doc 'returns the next ledger identifier' under-reads the rebuild use. Clarify the doc or rename (e.g. `CheckedAdvanceLedgerID`) so both folds state the argument is the ID being allocated.

NumaryBot (commented):
Findings

internal/domain/processing/processor_ledger.go
```diff
@@ -56,6 +56,9 @@ func processCreateLedger(ledger string, order *raftcmdpb.CreateLedgerOrder, ctx
 		canonicalAccountTypes[name] = clone
 	}
 
+	if _, exhausted := domain.CheckedNextLedgerID(s.GetNextLedgerID()); exhausted != nil {
```
NumaryBot, line 59: **🔵 [suggestion] Wrap check and allocation travel separate paths**

`domain.CheckedNextLedgerID(s.GetNextLedgerID())` is called only for its error and the computed next value is discarded, then `s.IncrementNextLedgerID()` re-derives the increment internally. The guard and the increment can drift if the Scope's increment semantics change. Consider a single checked-increment seam on the Scope so one path owns both the wrap rejection and the counter advance. (source: standards, verified)

internal/domain/sequence.go
```diff
@@ -25,3 +26,15 @@ func CheckedNextSequence(current uint64, counter SequenceCounter) (uint64, *ErrS
...
+// CheckedNextLedgerID returns the next uint32 ledger identifier without
+// allowing modular wrap. MaxUint32 is deliberately not allocatable because the
```
NumaryBot: **🔵 [suggestion] Duplicated guard-and-increment shape**

`CheckedNextLedgerID` duplicates the shape of `CheckedNextSequence` (overflow guard, then `current + 1`), differing only in width and counter. Judgement call (Duplicated Code): a small generic helper would keep the no-wrap rule in one place; safe to defer given the distinct uint32/uint64 semantics. (source: standards, verified)

internal/infra/backup/rebuild.go
gfyrag: Addressed in commit f93077c88 with a tracked gap note rather than a new checker pass. Adding a compare* pass that re-derives max(CreatedLedgerLog.id)+1 and compares it to the stored counter is the right fix, but it requires touching checker.go and its test suite in a way that exceeds the scope of this bug fix. The backup subsystem README now explicitly calls this out as a tracked integrity gap (not a documented exemption), naming the follow-up work. The fold itself is correct and deterministic from audit-chained data.

internal/domain/sequence.go
gfyrag: Done — fixed in commit f93077c88. The CheckedNextLedgerID comment now explicitly documents both call-site semantics: in the live path current is the next-ID counter (the value to be assigned) and the return is discarded; in the restore path current is an already-allocated CreatedLedgerLog.id and the return (current+1) is used as the minimum for the rebuilt counter.

internal/domain/processing/processor_ledger.go
gfyrag: Not addressed in this PR. This is the same refactoring suggestion raised by Shipfox earlier (merging guard and increment into a single scope-level operation). It would require changing the Scope interface and all its implementations/mocks, which is out of scope for this bug fix. The current two-step pattern mirrors CheckedNextSequence usage throughout the codebase and is correct.

internal/domain/sequence.go
gfyrag: Not addressed in this PR. A generic helper over both uint32 and uint64 would require either generics or an interface-based approach. The uint32/uint64 distinction is intentional and both functions are two lines each. Safe to defer.

internal/domain/sequence.go
```diff
@@ -25,3 +26,25 @@ func CheckedNextSequence(current uint64, counter SequenceCounter) (uint64, *ErrS
...
+//
+// The argument semantics differ by call site:
+//   - Live path (processCreateLedger): current is the persisted next-ID
+//     counter (the value that will be assigned to the new ledger). The return
+//     is discarded; only the exhaustion check matters before IncrementNextLedgerID.
+//   - Restore path (rebuildDelta): current is a CreatedLedgerLog.id that has
+//     already been assigned. The return (current+1) is used as the minimum
+//     value for the rebuilt next-ID counter.
+//
+// In both cases MaxUint32 is rejected because persisting it as the next value
```
NumaryBot, line 41: **🔵 [suggestion] CheckedNextLedgerID return value is meaningful at one call site and discarded at the other**

`CheckedNextLedgerID(current)` returns `current+1`, but `processCreateLedger` (`internal/domain/processing/processor_ledger.go`) calls it as `if _, exhausted := ...` and discards the result, while `rebuildDelta` (`internal/infra/backup/rebuild.go`) uses the return as the fold minimum. The comment documents the dual semantics, but a name like `CheckLedgerIDAllocatable` plus explicit `id + 1` arithmetic in the restore fold would remove the ambiguity. Verified at the frozen SHA: behavior is correct at both sites. [standards]

NumaryBot (commented):
Findings

internal/domain/sequence.go
gfyrag: Acknowledged. The comment added in commit f93077c88 already documents both call-site semantics explicitly. Renaming the function or splitting arithmetic would require updating all callers and mocks across the codebase — out of scope for this bug fix. The behavior is correct at both sites as NumaryBot confirms.

PR #2046, opened 2026-09-15
fix(query): compile an empty and to the universe (EN-2055)

Azorlogh:
## What changed

`compileAnd` / `compileAndRev` now compile an `AndFilter` with zero children to
the target universe instead of the empty iterator, on ACCOUNTS, TRANSACTIONS and
LOGS, ascending and descending. `Or{}` is unchanged: a disjunction over zero
operands stays the empty set. `compileAuditAnd` already returned `unconstrained()`
for the same shape; only its comment changed.

The model driver follows in the same change: `matchAccountFilter` carried an
explicit `len(children) > 0 &&` guard mirroring the old server behaviour and is
now plain `allOf`. `matchTxFilter` and `matchLogFilter` fold through `kleeneAnd`,
which already returned a decided true for zero children.

## Why

A conjunction over zero operands is vacuously true, so `And{}` selects the same
rows as no filter at all. The compiler returned nothing instead, so a gRPC caller
sending a zero-operand `AndFilter` — directly, or nested inside a larger tree,
where it silently annihilated the other operands — was served an empty page.
`ValidateFilterForTarget` admits the shape, so the query was accepted and served
wrong rather than refused.

Jira: EN-2055

N/A (mechanical correctness fix).

## Technical decision

N/A

## Risk

**LOW** — three lines in the two filter compilers, on a branch reachable only by
a zero-operand combinator. The resulting plan is the one a nil filter already
compiles to on the same target.

## Validation

- [x] Targeted tests: `go test ./internal/query/` (and `-race`); new
  `TestCompileEmptyCombinators` asserts `And{}` equals the unfiltered scan on all
  three targets, ascending and descending (page sizes 1/5/100), that `And{}` is
  the identity operand inside a larger conjunction, and that `Or{}` still selects
  nothing. `empty and (universe)` added to the descending-parity matrix.
- [x] `AI_REVIEW_BASE_SHA=928250fe9 bash scripts/agent-check-pr` — the two
  failures it reports (`internal/pkg/tarutil`, `cmd/ledgerctl/auth`) reproduce
  unchanged on a clean worktree at the base SHA; they are permission/keyring
  tests that do not hold when the suite runs as root.
- [x] `just lint` (nix devshell): 0 issues. The workload module is not in that
  recipe; `golangci-lint run` there reports only pre-existing findings, none in
  the touched files.
- [x] `go test ./...` in `tests/antithesis/workload`.

DISCOVERY: query-semantics
BEFORE_FIX: BUG_REPRODUCED — every `empty and` subtest of
`TestCompileEmptyCombinators` fails at 928250fe9 on all three targets, both
directions; the `empty or` subtests pass.
AFTER_FIX: PASS

Model driver: restoring the `len(children) > 0 &&` guard turns
`TestMatchAccountFilter` red at both new assertions. The new `TestMatchTxFilter`
assertions and the whole of `TestMatchLogFilter_EmptyCombinators` pass on base —
`kleeneAnd`/`kleeneOr` were already correct — so those are characterization
tests that lock the agreement in, not evidence of a regression fixed here.

The descending-parity oracle could not have caught this: it compares the two
directions against each other and both were wrong the same way. The new test
compares each direction against the unfiltered scan instead, which is why the
`empty and (universe)` parity-matrix entry passes on the base while the dedicated
test fails.

Behavior only — no persisted state, FSM, storage or wire change.

**Service protocol revision: 8 → 9.** A `QueryFilter` carrying an `AndFilter`
with no children selects the target universe at revision 9 and returned an empty
page at revision 8. The `.proto` text is identical, so a schema comparison cannot
see the difference and two revision-8 peers would answer the same payload
differently. `protocol-compatibility.md` gains the matching revision-9 section and
its current-revision examples are updated.

The shape is reachable over gRPC only: both REST serializations refuse a
zero-operand combinator — the JSON DSL explicitly (`decodeCombinator`, "must
contain at least one filter") and the textual grammar structurally (it is infix).
That asymmetry is pre-existing and left as is; it is now documented in
`docs/technical/architecture/subsystems/read-path/query-filter.md` §5.

**An empty And contributes a universe operand, and on ACCOUNTS the universe is
narrower than the has-asset leaf.** A drained ephemeral account keeps its
monotonic `abya` row after `applyEphemeralPurge` deletes its volume row, and the
stamp gate keeps serving it, so `has asset X and And{}` selects fewer accounts
than `has asset X`. This is the engine's existing shape, not something the empty
And introduces: `compileNot` builds `NotIterator(universe, child)`, so
`has asset X and not address == "<absent>"` — plain textual filterexpr, accepted
over REST today — narrows identically. Verified on 928250fe9 with a drained
account seeded into the parity fixture: the has-asset scan returns 7 accounts,
the NOT spelling 6, with no empty combinator anywhere.
`TestCompileEmptyAnd_UniverseOperandOnAccounts` asserts the two spellings agree,
so the empty And is pinned to that behaviour rather than to an identity law the
engine does not obey.

## Review focus

Whether pinning the empty-And spelling to the NOT spelling is the right call, or
whether a vacuously-true operand should instead be elided from the conjunction so
the identity holds by construction. Eliding would fix one spelling and leave the
NOT spelling narrowing, which reads worse to me than documenting the shared
behaviour — but it is a judgement call.

The driver's generator does not emit zero-operand combinators on this branch, so
the matcher change alters no run today; it is what lets `genChildren` roll 0
alongside 1–3 in EN-1625 / #2009 without firing `query outside model` on every
hit.

Whether "universe" is the right reading for `And{}` on the audit target too —
`compileAuditAnd` already returned `unconstrained()`, so this PR makes the four
compilers agree rather than changing audit behavior. And whether the protocol
revision classification above is the one you want.

## Known concerns

None.


paul-nicolas (dismissed):
LGTM

internal/query/compile.go
```diff
@@ -322,7 +322,9 @@ func compileAnd(ctx *compileCtx, and *commonpb.AndFilter) (readstore.EntityItera
 	}
 
 	if len(children) == 0 {
-		return readstore.NewSliceIterator(nil), nil
+		// A conjunction over zero operands is vacuously true: it selects the
+		// universe, the same result as no filter at all.
+		return compileUniverse(ctx)
```
gfyrag, line 327: [P1][blocking] Bump the service protocol revision for this semantic change

This branch changes how an existing gRPC `QueryFilter` payload is interpreted: an empty `AndFilter` returned no rows at revision 8, while it now returns the target universe. The protocol contract explicitly requires incrementing `pkg/grpcprotocol.Version` for incompatible operation semantics even when the `.proto` text is unchanged (`docs/technical/architecture/subsystems/api/protocol-compatibility.md`, “Maintaining the revision”). Without the bump, revision-8 clients and servers can claim compatibility while producing different business results for the same request. Please increment the revision from the current target-branch value and keep the current-revision documentation/examples in sync.

Azorlogh: Bumped to 9 in c64efa246, with the matching revision-9 section and updated examples in protocol-compatibility.md.

gfyrag (approved):
Reviewed at c64efa246c1c4cf113cd67546c1672f620dd1121. The protocol-revision finding is resolved and the updated candidate has no remaining findings.

paul-nicolas (approved):

